Issue #580, opened 2014-06-08
[R-Forge #2622] Add command "fwrite" to faster save csv files 


arunsrinivasan:
Submitted by: Ma Jom; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2622&group_id=240&atid=978)

I really love the new command "fread" as it really speeds up the work with large files. It would be great to see a corresponding command for writing CSV files.

Currently, I use a combination of write.csv and as.matrix to speed up this process a bit:
http://stackoverflow.com/questions/10505605/speeding-up-the-performance-of-write-table

skanskan:
I agree, data.table is missing a fwrite functionality.
It would be a great addition.

minhhahl:
+ 1 for it

stanasa:
+1 for fwrite function

takje:
I would also really appreciate an fwrite function!

MichaelChirico:
Of course I'm all in for this.

arunsrinivasan:
At this point, as I briefly touched upon during UseR 2015, there's no motivation or incentive to implement `fwrite` for neither of us. To implement it efficiently is a lot of work, and it comes out of our free time. With the possibility of it being just *reimplemented* as well in another package (like `readr`) at some point in time, it doesn't seem like a good investment of our time.

If anyone else is interested in investing their time, please feel free to contact us.

oseiskar:
We have developed a faster function for writing CSV files and achieved > 2x speedups (on a 64-bit Windows machine with an SSD) See [this Gist](https://gist.github.com/oseiskar/15c4a3fd9b6ec5856c89) for a proof of concept. I can start working on a pull request with more features if you are interested.

smartinsightsfromdata:
@oseiskar code seems very promising.
@mattdowle @arunsrinivasan would you consider a pr for a fwrite on this basis?

Bohdan-Khomtchouk:
Not having an `fwrite` to complement `fread` is like not having `write.csv` to complement `read.csv`.

migueltp:
+1 for fwrite function

julja:
+1 for fwrite function

mattdowle:
[...]

Issue #633, opened 2014-06-08
[R-Forge #2110] Add . and .. symbol prefixes and inherits=FALSE for strict scope


arunsrinivasan:
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2110&group_id=240&atid=978)

inherits=FALSE would allow user to specify that j should use column names only, and if column name doesn't exist that should be error rather than possibly finding a match in calling scope.

..() is for the sub-join examples where the join goes in j and you want to refer to the columns in i not .SD.  Something like that.

mattdowle:
would be useful here : 
http://stackoverflow.com/questions/34077955/variable-usage-in-data-table

JoshOBrien:
Once version 10.2 is released (if not before), we should revisit [this answer](http://stackoverflow.com/a/13383995/980833) to remove the mention of `datatable.WhenJisSymbolThenCallingScope=TRUE`. (As always, any of you should feel free to edit as you like).

Synergist:
It'd be nice to be able to use this feature and the ! notation for excluding columns in tandem.

```r
dt <- data.table(x = 1:10, y = rnorm(10), z = runif(10))
cols <- c('y', 'z')
dt[, !..cols] # should return col x
```

ctbrown:
I followed the [SO issue](http://stackoverflow.com/questions/34077955/variable-usage-in-data-table) here looking for this feature.  Since, it is still experimental may I recommend that the syntax be changed from `..A` to `..$A` where `..` is a reference to the parent.frame.  I *believe* this is more R-like, does the same thing and conforms to user expectations.  It seems that this can be simply(?) done with:

    makeActiveBinding('..', parent.frame)

Thanks,

Chris

arunsrinivasan:
Refer to https://github.com/Rdatatable/data.table/issues/2186 for name mention under NEWS.

franknarf1:
Another SO post: https://stackoverflow.com/q/45387223/

PR #1613, opened 2016-03-27
First version of the fwrite function #580

oseiskar:
This implementation of `fwrite` (#580) aims to be faster, or at least as fast as `write.csv`, but a few things have been left out or simplified:
 
 1. When `quote=TRUE`, all column names are quoted
 
 2. When `quote=FALSE`, nothing is quoted, even if this would break the CSV
 
 3. There is no option for `row.names`. They only make sense for `data.frame`s with named rows. For `data.tables`, they would just reduce to row numbers.

The speedup compared to `write.csv` depends on column types and parameters but speedup factors from 2 to 4 are possible.


R/fwrite.R
```diff
@@ -0,0 +1,62 @@
+fwrite <- function(dt, file.path, append = FALSE, quote = TRUE,
```
MichaelChirico: should definitely use `DT` or `x`, given the `dt` function in the `stats` package.

R/fwrite.R
```diff
@@ -0,0 +1,62 @@
...
+                   sep = ",", eol = "\n", col.names = TRUE, qmethod = "double",
+                   block.size = 10000) {
+  
+  # validate arguments
+  stopifnot(is.data.frame(dt))
+  stopifnot(ncol(dt) > 0)
+  stopifnot(identical(unique(names(dt)), names(dt)))
```
MichaelChirico: maybe a warning for this, instead of an error?

R/fwrite.R
```diff
@@ -0,0 +1,62 @@
...
+    append <- TRUE
+  }
+  
+  # handle empty dt
+  if (nrow(dt) == 0) return()
+  
+  # determine from column types, which ones should be quoted
+  if (quote) {
+    column_types <- lapply(dt, class)
+    quoted_cols <- column_types %in% c('character', 'factor')
```
MichaelChirico: better `%chin%`

R/fwrite.R
```diff
@@ -0,0 +1,62 @@
...
+  }
+  
+  # write in blocks of given size to avoid generating full copies
+  # of columns in memory
+  block_begin <- 1
+  
+  repeat {
+    block_end <- min(block_begin+(block.size-1), nrow(dt))
+    
+    dt_block <- dt[c(block_begin:block_end),]
```
MichaelChirico: Not sure how easy it is to skip the last column in C, but would probably be faster to define a `block` column, then key on it and use binary search, _a la_:

    DT[.(block_no)]

Would require using `setDT` in the case of `data.frame`, but that would also take care of the one-column DF problem on L22

R/fwrite.R
oseiskar: 1) ok, I'll change this

2) This is semi-intentional. Supporting `file=""` (stdout) and connections would require passing file handles (instead of just file name strings) from R to C and I don't currently know how to do this. R's C interface is poorly documented.

R/fwrite.R
```diff
@@ -0,0 +1,61 @@
...
+  }
+  
+  # write in blocks of given size to avoid generating full copies
+  # of columns in memory
+  block_begin <- 1
+  
+  repeat {
+    block_end <- min(block_begin+(block.size-1), nrow(x))
+    
+    dt_block <- x[c(block_begin:block_end),]
```
oseiskar: As a reply to [MichaelChirico's comment](https://github.com/Rdatatable/data.table/pull/1613#discussion-diff-57540368) if it would be faster to create an extra column and use it like `x[.(block_no)]`: could be a little bit faster and not difficult to implement in C, but I dislike the idea of modifying the input data table. What if the user already has a column called `block_no`?

jangorecki: @oseiskar none I'm aware of, once #633 solved it will be easy, you can use cryptic name (prefix it with the dot), and check if it doesn't exist in a data.table. Not sure what you are referring by modifying input, but adding column is as single as `shallow(x)[, "col" := new]`, it won't copy the data, and add new column only to locally processed data.

MichaelChirico: Just to reinforce Jan's point, this sort of thing (restricting acceptable column names) is done under the hood in several places of `data.table` code already, see e.g. [here](https://github.com/Rdatatable/data.table/blob/3dea474e15c3c8ad62f41a1a5d3b450f41f7d127/R/setkey.R#L49). As a suggestion, there's `b__`, which has an analogue to `f__`, `l__`, `o__`, `zo__`, `jn__`, and `jl__`, all found in `[.data.table`

R/fwrite.R
```diff
@@ -0,0 +1,61 @@
...
+    
+    # convert data.frame row block to a list of columns
+    col_list <- lapply(dt_block, function(column) {
+      str_col <- as.character(column)
```
MichaelChirico: I'm not 100% sure I'm interpreting the output of `profvis` correctly, but I'm pretty sure this here is by far the biggest bottleneck on the code as it stands:

screenshot from 2016-03-28 09 56 00

I think all the time spent on `Cwritefile` is in the tiny slices between `lapply` blocks (AFAICT `profvis` has no way of recording anything about C code, not even that the compiler is currently in C).

If that's correct, it seems there's ample room for improvement by porting the formatting step into C.

----
Here's the code I used for that:

```
setwd("~/Desktop/")
library(profvis)
library(data.table)
NN <- 1e6
set.seed(51423)
testDT <- 
  data.table(str1=sample(sprintf("%010d",1:NN)), #ID field 1
             str2=sample(sprintf("%09d",1:NN)),  #ID field 2
             #varying length string field--think names/addresses, etc.
             str3=replicate(NN,paste0(sample(LETTERS,sample(10:30,1),T),
                                      collapse="")),
             #factor-like string field with 50 "levels"
             str4=sprintf("%05d",sample(sample(1e5,50),NN,T)),
             #factor-like string field with 17 levels, varying length
             str5=sample(replicate(17,paste0(sample(LETTERS,
                                                    sample(15:25,1),T),
                                             collapse="")),NN,T),
             #lognormally distributed numeric
             num1=exp(rnorm(NN,mean=6.5,sd=1.5)),
             #3 binary strings
             str6=sample(c("Y","N"),NN,T),
             str7=sample(c("M","F"),NN,T),
             str8=sample(c("B","W"),NN,T),
             #right-skewed integer
             int1=ceiling(rexp(NN)),
             #dates by month
             dat1=sample(seq(from=as.Date("2005-12-31"),
                             to=as.Date("2015-12-31"),by="month"),
                         NN,T),
             dat2=sample(seq(from=as.Date("2005-12-31"),
                             to=as.Date("2015-12-31"),by="month"),
                         NN,T),
             num2=exp(rnorm(NN,mean=6,sd=1.5)),
             #date by day
             dat3=sample(seq(from=as.Date("2015-06-01"),
                             to=as.Date("2015-07-15"),by="day"),
                         NN,T),
             #lognormal numeric that can be positive or negative
             num3=(-1)^sample(2,NN,T)*exp(rnorm(NN,mean=6,sd=1.5)))

profvis(fwrite(testDT, "test.csv"))
```

oseiskar: That sounds right. I also interpreted from [my performance test](https://gist.github.com/oseiskar/c8117c5eed1c4522f7e0) that the biggest bottleneck is formatting, which is especially slow for floating point numbers. However, I'm not sure how much room there is for optimization. Is R's native `as.character` (which might allocate a lot of new small string objects) much slower than using C's `fprintf` to write directly to a file? Should be easy to check...

oseiskar:
I moved number and NA formatting to C code, which resulted in less column copying and a significant perfomance boost. [See this gist](https://gist.github.com/oseiskar/c8117c5eed1c4522f7e0). Now this version seems to be between 2 and 4 times faster than `write.csv` with character and numeric columns.

@jangorecki thanks for pointing out the shortcomings of `Sys.time`. Any ideas how it should be replaced in benchmarking?

MichaelChirico:
My preference is to use `microbenchmark::get_nanotime`

MichaelChirico:
I get an error when trying to write outside current wd; modifying slightly the example in `?fwrite`:

    fwrite(data.table(first=c(1,2), second=c(NA, 'foo"bar')), "~/Desktop/table.csv")

oseiskar:
@MichaelChirico The problem is not the working directory but the `~`, which is not automatically expanded in C. I'll fix this using `path.expand`...

MichaelChirico:
inre: `list`s, just noticed that `write.csv` pops an error in this case:

> Error in `.External2(C_writetable, x, file, nrow(x), p, rnames, sep, eol,`  : 
  unimplemented type `list` in `EncodeElement`

MichaelChirico:
inre: numeric columns, for consistency with `write.csv` more digits should be printed; from `?write.csv`:

> In almost all cases the conversion of numeric quantities is governed by the option "scipen" (see `options`), but with the internal equivalent of `digits = 15`. For finer control, use `format` to make a character matrix/data frame, and call `write.table` on that.

jangorecki:
@oseiskar 
```r
elapsed.secs <- system.time(ans <- fwrite(...))[[3L]]`
```
is the easiest way
for precise timing you can use
```r
pt = microbenchmark::get_nanotime()
ans = fwrite(...)
microbenchmark::get_nanotime() - pt
```
or from the devel version of [microbenchmark**Core**](https://github.com/olafmersmann/microbenchmarkCore) a drop-in replacement for `system.time` called `system.nanotime`

No need to have list columns supported now, this can be easily and flexibly handled within data.table before passing it to fwrite.

Not sure about `as.character()` for non-(numeric/integer/characters), it might be better to use `toString` instead, but it generally depends on how it is processed further.

R/fwrite.R
MichaelChirico: 2) Gotcha. Not sure how important the consistency is, but maybe just follow `write.table` to this end:

```
if (file == "") 
  file <- stdout()
else if (is.character(file)) {
  file <- if (nzchar(fileEncoding)) 
    file(file, ifelse(append, "a", "w"), encoding = fileEncoding)
  else file(file, ifelse(append, "a", "w"))
  on.exit(close(file))
}
else if (!isOpen(file, "w")) {
  open(file, "w")
  on.exit(close(file))
}
if (!inherits(file, "connection")) 
  stop("'file' must be a character string or connection")
```

And [here](https://github.com/wch/r-source/blob/b156e3a711967f58131e23c1b1dc1ea90e2f0c43/src/library/utils/src/io.c#L1038) is the C side of the code, to the extent that it helps

oseiskar:
As a reply to [this earlier comment](https://github.com/Rdatatable/data.table/pull/1613#discussion-diff-57534709) about supporting `write.csv`-like `file` argument that could also be an R connection and would default to writing to the console: **not possible**, as far as I know. Adding support to writing in STDOUT is possible but it seems that writing to an existing R connection is not.

@MichaelChirico thanks for looking up [the C side of `write.csv`](https://github.com/wch/r-source/blob/b156e3a711967f58131e23c1b1dc1ea90e2f0c43/src/library/utils/src/io.c#L1038). Unfortunately, it does not seem to be possible do replicate this because the code is calling functions that are not part of the [public C API of R](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-R-API). It would seem that all file/connection handling is part of the _private_ / internal R code and thus unavailable.

In particular, one would need to call `getConnection` to transform a parameter `SEXP`, representing a connection number, to an `Rconnection` and trying to replicate the [implementation this function](https://github.com/wch/r-source/blob/53c4fa4b0d918b4e6ab42ea103300c455cad8c23/src/main/connections.c#L166) would not help here.

dselivanov:
Not sure, if this will help, but leave it here: [official C level API for connection handles in R 3.3.0](http://developer.r-project.org/blosxom.cgi/R-devel/2016/03/24#n2016-03-24).

oseiskar:
Actually, it is possible to bypass the private API limitation and call `getConnection` at least on some platforms: see [this branch](https://github.com/oseiskar/data.table/compare/fwrite...oseiskar:fwrite-with-rconnections) for a proof of concept. However, this brings in a whole new bunch of problems:

 1. It is not guaranteed to work or remain supported (However, as @dselivanov pointed out, official support will be available in R version 3.3.0)
 2. The first attempted implementation absolutely kills performance. The code linked above is about 4 times slower that `write.csv` with character columns.
 3. For regular files, it is possible to "fix" 2. using lower-level functions [like this](https://github.com/oseiskar/data.table/blob/82239f346134d70f0c2cb513c96c671f28b6fe2d/src/fwrite.c#L23) (speedup drops from about 2x to about 1.5x) but this will not work for all R connections, such as `stdout`. It is possible to handle both cases separately, which would significantly increase the complexity of the code.

In my opinion, supporting R connection arguments is not worth the trouble at this point.

R/fwrite.R
dselivanov: IMHO `%chin%` is overkill here =)

dselivanov:
Personally I think, connections is "nice to have" feature (which can be implemented in future), but even without connections this is a great PR. Thank you, @oseiskar.

R/fwrite.R
MichaelChirico: @dselivanov 
chin

Seems its always 2-3x faster than `%in%`, what's the harm? Two-character difference...

dselivanov: no harm! Just little bit readable. All know about `%in%`, but `%chin%` is `data.table` specific. In this particular case speedup won't be noticeable.

MichaelChirico: that's what `?"%chin%"` is for ;-)

jangorecki:
+1 dselivanov
FYI @mattdowle :)

mattdowle:
Awesome! Looks great @oseiskar :-)

arunsrinivasan:
Great PR, @oseiskar. Using the `data.dable` from the extensive benchmark Michael did on SO [here](http://stackoverflow.com/a/36465497/559784), I profiled it using Instruments -> time profiler. Here's a snapshot.

fwrite-time-profiler

I've not looked at the code, but IIUC `writefile()` takes only about 50% of the time? Also perhaps there are trivial places that could be improved on that you could spot? Hope it's of some use.

jangorecki:
@oseiskar 
Could you substitute `R_xlen_t` with something which is available in R 2.15.0?
While installing package on R 2.15.0 (data.table stated dependency) I get the following error:
```r
fwrite.c: In function ‘writefile’:
fwrite.c:31:3: error: unknown type name ‘R_xlen_t’
   R_xlen_t ncols = LENGTH(list_of_columns);
   ^
```

You can reproduce it with:
```
docker run -it docker.io/jangorecki/r-2.15.0 /bin/bash
curl -O https://Rdatatable.github.io/data.table/src/contrib/data.table_1.9.7.tar.gz
R2 CMD INSTALL data.table_1.9.7.tar.gz
```

Alternatively, if there is no substitution for that or it would decrease performance, then @mattdowle would need to decide about stated dependency upgrade. R 2.15.0 is from March 2012, in my opinion 4 years old isn't old enough to deprecate it without strong reason.

arunsrinivasan:
@jangorecki  here's the solution: https://github.com/Rdatatable/data.table/blob/9d2d71098d849c99e6eebb0e0b539eb58d723b05/src/rbindlist.c#L13-17

mattdowle:
Work continued here:  https://github.com/Rdatatable/data.table/issues/1664

PR #1622, opened 2016-04-04
Add trailing `[]` to each merge() call in examples

JoshOBrien:
That way the output will be printed (rather than silently returned, which might confuse some users).

### DIFF
--- a/man/merge.Rd
+++ b/man/merge.Rd
@@ -99,17 +99,17 @@ allow.cartesian=getOption("datatable.allow.cartesian"),  # default FALSE
 \examples{
 (dt1 <- data.table(A = letters[1:10], X = 1:10, key = "A"))
 (dt2 <- data.table(A = letters[5:14], Y = 1:10, key = "A"))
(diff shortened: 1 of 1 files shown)


jangorecki:
In my opinion it is better to handle that inside `merge` function just before returning data, but I'm not sure if `[]` inside merge would not affect performance. If it does then we need to alter visibility some other way. But lets what for @arunsrinivasan opinion.

arunsrinivasan:
@JoshOBrien, thanks for the PR, but this was recently fixed in `merge.data.table`: https://github.com/Rdatatable/data.table/commit/af47d833d1c31656ac38f050459a13d98415ac2a

Please reopen if you it doesn't work as expected on v1.9.7.

JoshOBrien:
@arunsrinivasan Aarg. Don't know why I didn't update *first*, before filing the PR. Sorry for the noise.

arunsrinivasan:
@JoshOBrien no worries :-).

PR #1630, opened 2016-04-06
Correct speed units

UweBlock:
Distances between airports are given in statute miles in the data according to description by   [Bureau of Transporation Statistics](http://www.transtats.bts.gov/DL_SelectFields.asp?Table_ID=236) and verified with [Great Circle Mapper](http://www.gcmap.com/dist?P=JFK-LAX).


arunsrinivasan:
Thank you.

PR #1640, opened 2016-04-11
fixed url in fread man

dalloliogm:
URL didn't work, for example when clicking it from http://www.inside-r.org/packages/cran/data.table/docs/fread


jangorecki:
Thanks @dalloliogm

PR #1665, opened 2016-04-20
Closes #1654; dcast is more selective on erroring for duplicated columns

### LINKED ISSUE #1654: dcast should only error more selectively on non-unique columns
Consider:

    DT <- data.table(a = 1:4, a = 1:4, id = rep(1:4, 2), V1 = 8:1)

    dcast(DT, id ~ rowid(id), value.var = "V1")

Which returns error:

> Error in `dcast.data.tabl(DT, id ~ rowid(id), value.var = "V1")` : 

> `data.table` to cast must have unique column names

This is an understandable restriction, but in this case none of the non-distinct columns will appear in the output, so it seems `dcast` should be able to proceed unabated and without risk of any problems.

I'll explore this and should be able to put in a PR later this week

MichaelChirico:
This was the most straightforward way I could find for this. I guess it could use `%chin%` but that's pretty minor


arunsrinivasan:
Looks great! Thanks again.

PR #1666, opened 2016-04-21
Closes #1653; fixes 0-rounding issue in vignette

### LINKED ISSUE #1653: Error Display in Vignettes
When I was reading the Vignettes on CRAN (https://cran.r-project.org/web/packages/data.table/vignettes/datatable-keys-fast-subset.html), I found something seeming not well put, as below
screen shot 2016-04-17 at 3 28 21 pm

The speedup factor is not apprearring. I checked the raw Rmd file, it is as below,
screen shot 2016-04-17 at 3 28 07 pm
screen shot 2016-04-17 at 3 28 14 pm

So the reason should be that the t2[3] is 0, then t1[3]/t2[3] is a invalid value, which caused the disapprearance of the speedup factor.

I'm thinking if it's necessary to use **microbenchmark** package to do more accurate benchmarking. If you find it necessary, I would be more than happy to submit a PR regarding it.

MichaelChirico:
Leaving it as `max(0, .001)` because it would be cheating to do otherwise without `microbenchmark`; from `proc.time`:

> The resolution of the times will be system-specific and on Unix-alikes **times are rounded down to milliseconds**. On modern systems they will be that accurate, but on older systems they might be accurate to 1/100 or 1/60 sec. They are typically available to 10ms on Windows.

(and even here we're _technically_ cheating on Windows systems, albeit probably only on very slow ones)

### DIFF
--- a/vignettes/datatable-keys-fast-subset.Rmd
+++ b/vignettes/datatable-keys-fast-subset.Rmd
@@ -464,7 +464,7 @@ dim(ans2)
 identical(ans1$val, ans2$val)
 ```
 
-* The speedup is **~`r round(t1[3]/t2[3])`x**!
+* The speedup is **~`r round(t1[3]/max(t2[3], .001))`x**!
 
 ### b) Why does keying a *data.table* result in blazing fast susbets?
 


jangorecki:
Maybe instead of cheating we can put a comment about rounding.

> The speedup is **~`r round(t1[3]/t2[3])`x**! (after rounding)

Even if `Inf` appears it will be still correct, and not that far from real speed-up :)

MichaelChirico:
@jangorecki Problem was `Inf` wasn't showing up, it just went blank. Maybe

> The speedup is at least `...`x!

arunsrinivasan:
Thanks!

arunsrinivasan:
Ideally, `Inf` would be great.. Perhaps using `sprintf()` instead?

```r
sprintf("%.3f", 45/0)
# [1] "Inf"
sprintf("%.3f", 45/19)
# [1] 2.368
```

MichaelChirico:
@arunsrinivasan I dunno about "ideally" since it's pretty misleading to say we've created an "infinite" speedup...

Could perhaps do an `if(require(microbenchmark))` split to give the most accurate results to users with that installed already, and use the `sprintf` approach on the branch for users without `microbenchmark`... at the risk of making the code in that section much more complicated :\

jangorecki:
+1 for [microbenchmark](https://github.com/olafmersmann/microbenchmark), I would prefer [microbenchmarkCore](https://github.com/olafmersmann/microbenchmarkCore) but it is not yet on CRAN. Both packages are already updated so microbenchmark depends on *Core, but not yet published in CRAN.
In fact CRAN policy allows to have Suggested packages from non-mainstream repo, and verify their availability (but not install) during `R CMD check` based on `Additional_repositories` field in `DESCRIPTION` file. Sample from log:
```
* this is package ‘logR’ version ‘2.1.5’
Maintainer: ‘Jan Gorecki <j.gorecki@wit.edu.pl>’
New submission

Suggests or Enhances not in mainstream repositories:
  microbenchmarkCore

Availability using Additional_repositories specification:
  microbenchmarkCore   yes   https://olafmersmann.github.io/drat
```
I would stick to the CRAN anyway, microbenchmark's heavy deps are already in data.table suggests, so it doesn't cost much. Just another package to suggests.
I would also apply it on `datatable.verbose=TRUE` output timings, but it won't measure *user* time (only *elapsed*), which starts to be more meaningful for parallell functions.

Issue #1459, opened 2015-12-04
Integer64 columns not assigned NA when doing a left merge


abielr:
Handle NA integer64 properly in:
- [x] merge
- [x] rbindlist (following @jangorecki's comment below)

``` r
require(data.table)
require(bit64)
x1 <- data.table(id=1, value=as.integer64(1))
x2 <- data.table(id=c(1,2))
merge(x2, x1, by="id", all.x=T)
```

Yields

``` r
#    id               value
#1:  1                   1
#2:  2 9218868437227407266
```

This is with data.table 1.9.6.

dlithio:
I think this is a [duplicate](https://github.com/Rdatatable/data.table/issues/1385).

arunsrinivasan:
Thanks for linking to the old issue @dlithio. Maybe we can fix it internally..

jangorecki:
The same happens when `rbindlist(fill=TRUE)`. This probably can be fixed in the same place as `merge` so I'm not creating new issue for that, let me know if I should.
```r
library(data.table)
library(bit64)
rbindlist(list(
    data.table(a=1:2, b=as.integer64(c(1,NA))),
    data.table(a=3L)
), fill=TRUE)
#   a                   b
#1: 1                   1
#2: 2                  NA
#3: 3 9218868437227407266
```

andrewkho:
I'm not sure if I should create a new issue for this, but I am having this problem when using the := operator, but not when using merge. 

```
> packageVersion("data.table")
[1] ‘1.10.4’
> packageVersion("bit64")
[1] ‘0.9.7’
> dt1 <- data.table(id=c(1,2), x=bit64::as.integer64(100,200))
> dt2 <- data.table(id=c(1,3))
> dt1
   id   x
1:  1 100
2:  2 100
> dt2
   id
1:  1
2:  3
> merge(dt1,dt2, all=T)
   id   x
1:  1 100
2:  2 100
3:  3  NA
> dt2[J(dt1), x := i.x, on=.(id)]
> dt2
   id                   x
1:  1                 100
2:  3 9218868437227407266
```

MichaelChirico:
confirming problem for current dev

Issue #1377, opened 2015-10-05
grouping sets


jangorecki:
Some keywords: GROUPING SETS, ROLLUP, CUBE, GROUPING
Some references: [postgres](http://www.postgresql.org/docs/9.5/static/queries-table-expressions.html#QUERIES-GROUPING-SETS), [Oracle](https://oracle-base.com/articles/misc/rollup-cube-grouping-functions-and-grouping-sets), [SQL Server](https://technet.microsoft.com/en-us/library/bb510427%28v=sql.105%29.aspx), [groupings combined with arbitrary functions](http://www.databasejournal.com/features/mssql/using-the-rollup-cube-and-grouping-sets-operators.html)

_Grouping sets_ and friends are useful to pre-calculate various aggregation levels, which is often desired. Api for that feature in data.table is not very friendly, see [Aggregating sub totals and grand totals with data.table](http://stackoverflow.com/q/9315258/2490497).

In case of _rollup_ those are aggregations for provided `by` from top to bottom. See description from postgres man, and example code below.

```
ROLLUP ( e1, e2, e3, ... )
```

is equivalent to:

```
GROUPING SETS (
    ( e1, e2, e3, ... ),
    ...
    ( e1, e2 )
    ( e1 )
    ( )
)
```

I wonder if there could be cheap speed-up of that process? this is potentially heavy computing task. Would be great to have computation of _grouping sets_ feature developed in C, so all the _rollup/cube_ and other features could be built on top of _grouping sets_ more easily in R still utilizing full speed.

---

Answers to update when closed:
- [ ] https://stackoverflow.com/questions/21366138/multi-level-aggregations-like-grouping-sets-via-ddply-or-other-r-function

``` r
library(plyr)
grp.cols <- c("vs", "am", "gear", "carb", "cyl")
plyr.r = do.call(
    rbind.fill,
    lapply(1:length(grp.cols), function(x) ddply(mtcars, grp.cols[1:x], summarize, agg=mean(mpg)))
)

library(data.table) # 1.9.7+
dt.r = rollup(as.data.table(mtcars), j = .(agg=mean(mpg)), by=grp.cols)
all.equal(
    as.data.table(plyr.r),
    dt.r[-.N], # exclude grand total, not present in BrodieG answer
    ignore.row.order = TRUE,
    ignore.col.order = TRUE
)
#[1] TRUE
# install.packages("data.table", type = "source", repos = "https://Rdatatable.github.io/data.table")
```
[...]

PR #1667, opened 2016-04-21
RC - Grouping Sets, rollup, cube. #1377

jangorecki:
**rebased to 1.10.1**

Working solution to solve #1377 on R side. It rbind results from multiple aggregation calls. Surely not as efficient as if it would be pushed to C. Still not trivial to achieve for end-users.

This PR contains:
- [x] 3 generic methods for rollup, cube, grouping sets
- [x] methods for data.table
- [x] optionally generates aggregation levels as grouping id (bit mask), crucial for better consumption of the data and navigation over aggregates.
- [x] unit tests vs. postgres grouping sets queries matched successfully (not included in source due to heavy dependency).

Additionally handled in second iteration:
- [x] define tests
- [x] check edge cases
- [x] detection if `.SD` was used in `j`, then set default `.SDcols`, currently provided `.SD` and no `.SDcols` is not well handled
- [x] replace `bitwAnd` function call with something that works on R 2.15.0
- [x] ~~use `verbose` argument in some useful way~~
- [x] forbid `:=` calls
- [x] double counting on duplicated sets passed to `groupingsets` function should raise warning
- [x] detect duplicated columns within single grouping - error
- [x] col used both in `by` and `j` raise error
- [x] workaround for integer64 issue on `rbindlist(fill=TRUE)` #1459 
- [x] unit tests proper handling of factors on 2.15.0

Extra TODO (won't be included in source):
- [x] test edge cases against postgres and compare match to data.table
- [x] check bigger volumes vs. postgres

---
- [x] Add info to NEWS.md!

---
Improvements:
- [ ] handle same sets of columns but in different order, now it force order and warn on double counting


mshiner:
I'd be very interested to see this implemented in data.table as I'm hoping to use it as part of a reporting front end for the open source ERP system https://github.com/uzerpllp/uzerp we develop

jangorecki:
possible speed up after #1892

PR #1669, opened 2016-04-22
Closes #1648--improvements to metadata functions tables/columns

### LINKED ISSUE #1648: Improve access to data.tables metadata - `tables()`, `columns()`
FR for functions to allow easier access to data.tables metadata and easier audit/programmatic use. Function `tables` could be improved, and new function `columns` could provide information on columns.

Follow up of Rick's old PR #845
Also related to #641

MichaelChirico:
few tweaks, add address to columns

more tidying

man page and news updates

more updates, ready for PR

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -5,7 +5,7 @@ import(methods)
 exportClasses(data.table, IDate, ITime)
 ##
 
-export(data.table, tables, setkey, setkeyv, key, "key<-", haskey, CJ, SJ, copy)
+export(data.table, tables, columns, setkey, setkeyv, key, "key<-", haskey, CJ, SJ, copy)
 export(set2key, set2keyv, key2, setindex, setindexv, indices)
 export(as.data.table,is.data.table,test.data.table,last,like,"%like%",between,"%between%",anywhere,"%anywhere%")
 export(timetaken)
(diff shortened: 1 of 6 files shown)


jangorecki:
`columns` get way more complicated, it has almost sample scope in a much simpler way in https://github.com/Rdatatable/data.table/compare/master...jangorecki:information_schema#diff-f804b88946820485d304b5ab91817450R67

- Nesting `with` in `set` does looks complex and makes no different on `info_i` data.table which just holds metadata, at worst case it has nrow of columns of all tables. I would focus on simplicity of code + performance
- Transforming call `DT[,tryUniqueN(eval(as.name(colname)))])` to conditional could better to write as `if (is.atomic(DT[[colname]])) DT[...] else DT[...]` or even `if(.) uniqueN(DT[[colname]]) else length(unique(.))` instead of nesting `if` inside `DT[, if(x<-eval(.)) ... else ... ]`.
- `DT[ , sum(is.na(eval(as.name(jj))))]` may be a slightly faster as `sum(is.na(DT[[jj]]))`

arunsrinivasan:
At the moment, improving `tables()` alone makes sense. It'd be great to split this into two, and work on `tables()` alone, adding secondary indices as a first step.

MichaelChirico:
closing this, but keeping the code around for future reference

PR #1670, opened 2016-04-23
Closes #1668; download.file in fread forced to "auto" for file:// input

### LINKED ISSUE #1668: fread tries to use wget on file:// input in RStudio
Not sure what's going on here, but two tests fail when I run them in R Studio, but not from command line:

```
library(data.table)
test.data.table()
```

> Running test id 1378.2     Test 1378.2 Error in download.file(input, tt, mode = "wb", quiet = !showProgress) : 
  (converted from warning) download had nonzero exit status
Running test id 1378.3     Test 1378.3 didn't produce correct error :
> fread(f, showProgress = FALSE) 
Expected error: 'Line ending is .*r.*r.*n. R's download.file() appears to add the extra .*r in text mode on Windows. Please download again in binary mode (mode='wb') which might be faster too. Alternatively, pass the URL directly to fread and it will download the file in binary mode for you.'
Observed warning: 'download had nonzero exit status'

However, when I run from terminal:

    R -e 'library(data.table); test.data.table()'

All is well:

> All 5385 tests (lastID=1668.5) in inst/tests/tests.Rraw completed ok in 00:01:20 on Fri Apr 22 16:48:58 2016 (endian=little, sizeof(long double)==16)

Running 3.2.4 on RStudio 0.99.878, Linux Mint 17.3 = Ubuntu 14.04.3 LTS

I've been experiencing this for a while, but ignoring it because all the PRs pass Travis, etc. I only discovered the bit about the command line today, I was a bit surprised.

MichaelChirico:


arunsrinivasan:
Thanks for the quick fix, @MichaelChirico.

PR #1682, opened 2016-04-28
Closes #1674 -- only force remove column names for appending to extant files

### LINKED ISSUE #1674: fwrite shouldn't affect col.names when append = TRUE but the file doesn't exist
Currently we've got this line in `fwrite`:

    if (append && missing(col.names)) col.names = FALSE

Basically, I would change that to

    if (append && missing(col.names) && file.exists(file.path)) col.names = FALSE

Logic being we should always add `col.names` to a new file unless the user specifically overrides it.

MichaelChirico:


arunsrinivasan:
Makes sense. Could you please rebase with latest commit so I can merge? Thanks.

MichaelChirico:
@arunsrinivasan  done

PR #1707, opened 2016-05-16
Closes #1706, library(data.table) fails if "Packaged" field is not pr…

### LINKED ISSUE #1706: library(data.table) fails if "Packaged" field is not present in DESCRIPTION
Recently, R seems to no longer add the "Packaged" field to DESCRIPTION when building from source. This prevents an end-user from loading the dev version because of the "staleness" check in `onAttach()`. However, "Built" is in there (at least on the systems I tested this against).  I will submit a that  PR attempts to capture that and use it as a substitute.  If that fails, it returns since this check is purely informational.

restonslacker:
…esent in DESCRIPTION

Recently, R seems to no longer add the "Packaged" field to DESCRIPTION when building from source. This prevents an end-user from loading the dev version because of the "staleness" check in `onAttach()`. However, "Built" is in there.  This PR attempts to capture that and use it as a substitute.  If that fails, it skips the rest of the check.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -203,6 +203,8 @@
   
   59. 55. Added `+.IDate` method so that IDate + integer doesn't revert to `Date`, [#1528](https://github.com/Rdatatable/data.table/issues/1528); thanks @MichaelChirico for FR&PR.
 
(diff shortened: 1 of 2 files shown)


jangorecki:
Looks like we don't have to `try()` as `read.dcf` has an argument `fields`. Check below
```
read.dcf(system.file("DESCRIPTION", package="data.table"), fields="Packaged")
read.dcf(system.file("DESCRIPTION", package="data.table"), fields="notexistingfield")
```
Then we can handle that with `is.na`.

PR #1784, opened 2016-07-29
Closes #1783 -- fixed bug preventing subsetting an ITime column

### LINKED ISSUE #1783: Internal Error when an ITime is used in an i-expression
The following code generates an error:

```r
library(data.table)
it <- as.ITime("16:00")
DT <- data.table(x=it)

DT[x == it]
#     Error in alloc.col(x) : 
#         Internal error: length of names (0) is not length of dt (1)
```

However, this works correctly:

```r
DT[x == unclass(it)]
#           x
# 1: 16:00:00
```

Versions:
R 3.3.1
data.table 1.9.6

MichaelChirico:
that was easy enough


arunsrinivasan:
:-) you could just add `as.data.table.ITime` to the already existing many functions like so...

```r
as.data.table.factor <- as.data.table.ordered <- 
as.data.table.integer <- as.data.table.numeric <- 
as.data.table.logical <- as.data.table.character <- 
as.data.table.Date <- as.data.table.ITime <-  # <~~~~~
function(x, keep.rownames=FALSE, ...) {
```

Isn't it?

MichaelChirico:
Oh, I misread that code initially -- didn't catch that _all_ of those methods are identical. Will push the tinker momentarily

PR #1792, opened 2016-08-02
Closes #592 -- allows for incomplete specification of new order in setcolorder

### LINKED ISSUE #592: [R-Forge #2490] Add new movecol(DT, "colname", 1) function
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2490&group_id=240&atid=978)

As subject:   would save a whole setcolorder() just to move one column.

MichaelChirico:
Basic idea is simple -- if `setcolorder` gets its new 3rd argument, manipulate that into something that the old `setcolorder` would have understood and continue.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -1442,7 +1442,6 @@ DT = data.table(a=1:2,b=3:4,c=5:6)
 test(495, setcolorder(DT,c(2,1,3)), data.table(b=3:4,a=1:2,c=5:6))
 test(496, setcolorder(DT,c(2,1,3)), data.table(a=1:2,b=3:4,c=5:6))
 test(497, setcolorder(DT,c("c","a","b")), data.table(c=5:6,a=1:2,b=3:4))
-test(498, setcolorder(DT,"a"), error="neworder is length")
 test(498.1, setcolorder(DT,c("d","a","b")), error="Names in neworder not found in x: d")
 
 
@@ -9789,6 +9788,11 @@ indices(DT, vectors = TRUE)
 test(1749.1, indices(DT), c("A__B","A","B"))
 test(1749.2, indices(DT, vectors = TRUE), list(c("A","B"),"A","B"))
 
+# allow incomplete specification in setcolorder, #592
+DT = data.table(a = 1:3, b = 2:4, c = 3:5)
+test(1750.1, names(setcolorder(DT, "b")), c("b", "a", "c"))
+test(1750.2, names(setcolorder(DT, c(2, 3))), c("a", "c", "b"))
+test(1750.3, setcolorder(DT, 1:4), error = "x has only")
 
 ##########################
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@
 
 1. `indices()` function gain new argument `vectors` default `FALSE`, when `TRUE` provided then list of vectors is returned, single vector refers to single index. Closes #1589.
 
+2. `setcolorder` allows incomplete specification of new order (chosen columns are moved to the front), [#592](https://github.com/Rdatatable/data.table/issues/592). Thanks @MichaelChirico for the PR. 
+
 #### BUG FIXES
 
 #### NOTES
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -2569,22 +2569,30 @@ setnames <- function(x,old,new) {
     invisible(x)
 }
 
-setcolorder <- function(x,neworder)
+setcolorder <- function(x, neworder)
 {
+    if (any(duplicated(neworder))) stop("neworder contains duplicates")
     # if (!is.data.table(x)) stop("x is not a data.table")
-    if (length(neworder)!=length(x)) stop("neworder is length ",length(neworder)," but x has ",length(x)," columns.")
+    if (length(neworder) != length(x)) {
+      if (length(neworder) > length(x)) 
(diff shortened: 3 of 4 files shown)


jangorecki:
@MichaelChirico can't just `neworder` move those columns to front in provided order, then all others. No new args, and no breaking changes as we put new feature in place of error `Error in setcolorder(dt, "b") : neworder is length 1 but x has 2 columns.`.

MichaelChirico:
@jangorecki that seems reasonable. Approach in this PR represents a more general new feature, but as you point out, may be tougher to merge into current paradigm. If people actually ask for that, may be worth revisiting. Given that issue has little apparent traffic, your suggestion probably covers vast majority of empirical use cases.

Will edit this PR and resubmit.

inst/tests/tests.Rraw
```diff
@@ -9789,6 +9788,11 @@ indices(DT, vectors = TRUE)
 test(1749.1, indices(DT), c("A__B","A","B"))
 test(1749.2, indices(DT, vectors = TRUE), list(c("A","B"),"A","B"))
 
+# allow incomplete specification in setcolorder, #592
+DT = data.table(a = 1:3, b = 2:4, c = 3:5)
+test(1750.1, names(setcolorder(DT, "b")), c("b", "a", "c"))
+test(1750.2, names(setcolorder(DT, 3)), c("c", "b", "a"))
```
MichaelChirico: I'll just change one of the tests to be length>1. the other tests are good but I guess aren't new to this PR...

jangorecki: yeah, but your chunk changes behavior of the next chunk, `setdiff` will both remove NAs and duplicates so later validation will not raise error when it should.

MichaelChirico: Actually `setcolorder(DT, c(2, 2, 3))` (e.g.) causes erroneous behavior. But there's a simple fix. Will push momentarily. Thanks for the extra eyes.

jangorecki:
@mattdowle why not just new argument then? we used to not create new functions for new stuff when it is not needed, `setcolorder` is still good name for function to reorder partially.

mattdowle:
Maybe new argument.  Looking more into it, what happened to my original request of not just moving to the beginning but moving a column to a particular position?  The title of [#592](https://github.com/Rdatatable/data.table/issues/592) (that I filed on R-Forge in 2013, so I remember wanting the ability for some reason at the time) is "Add new movecol(DT, "colname", 1) ".   This PR doesn't close that request because it always moves to the beginning.
So setcolfirst is too restrictive,  maybe setcolpos() then?
setcolpos(DT, "b", 1)   # move b first
setcolpos(DT, c("b", "d"), 1)  # move b and d first
setcolpos(DT, c("b","d"), 10)  # move b and d to be positions 10 and 11.
setcolpos(DT, c("b","d"), before="someCol")   # move b and d to be just before someCol.

MichaelChirico:
That was how I originally programmed it, as a new argument to `setcolorder` (so it works like `setnames`). I unfortunately just overwrote that to make jan's adjustment.

Problem with that PR was that I used new argument names which made more sense for the more robust version of `setcolorder`, but which would have broken old code.

mattdowle:
I see now - yes comparison to `setnames` is helpful.  What would have broken?  `setnames` checks `missing(new)` first and then takes `old` as if it were `new`.  Perhaps it could be done in the same way for `setcolorder` so that nothing breaks.

MichaelChirico:
Given the analogy/parallel use, perhaps we should change the `setcolorder` arg names to `old` and `new`?

Again this gets at the breaking change issue. But perhaps we can continue to accept `neworder` with warning?

mattdowle:
Sounds reasonable to me.  So it would be `setcolorder(DT,old,new,neworder)`?  If user code contains `setcolorder(DT,neworder=o)`, the warning would ask to either change to `new=` or no named argument at all.  Then in future, remove `neworder=`.   Any existing code such as `setcolorder(DT, o)`  where o is passed unnamed, will continue to work in future and need to be as long as `ncol(DT)`  just like `setnames(DT, newnames)`.
In fact,  the `new` argument could just be called `neworder=` and then no warning would need to happen.  But then we'd be stuck with the old name and inconsistency with `setnames`.  So I'd lean towards named `neworder=` being a warning and then deprecate it, yes.  Think it's quite low risk that users have named that argument and would mind changing.
Jan?

MichaelChirico:
Actually Jan said it broke some of his own code to have different argument names! Since he called `neworder` specifically.

I do like the consistency with `setnames`, I think that's why I found this issue in the first place, was trying to use `setcolorder` like I do `setnames`.

jangorecki:
Sorry, I missed that part of PR.
Change of `neworder` argument breaks (at least) two of my pkgs. Better to avoid argument name changing if it isn't really necessary.
`setcolorder(DT,old,new,neworder)` is good one

mattdowle:
Merging as-is.  Thought process ...
Looking back with fresh eyes, I was being over-cautious on this one: the convenience of the common case of moving columns to the front outweighs the loss of the (most often, inconvenient, rather than welcomed) error.  This PR itself doesn't add any functions or arguments and so won't impact any dependent packages.  It's a good low-risk first step.  Whether and how to move a subset of columns to a not-first position can be revisited in future.  I can't see that accepting this PR now will rule out any options in future.

PR #1796, opened 2016-08-03
Closes #1383; as.ITime.character parallels logic of as.POSIXlt.character to guess format

### LINKED ISSUE #1383: IDateTime should take character input
I suspect this is known, but could not find the issue posted:

```
library(data.table) # testing on the latest github build
x  = "2015-09-29 08:22:00"

IDateTime(x)
#         idate itime
#1: 2015-09-29  <NA>
IDateTime(factor(x))
#         idate    itime
#1: 2015-09-29 08:22:00
```

I would expect to see this give both a date and a time in both cases.

By the way, the [posting guidelines](https://github.com/Rdatatable/data.table/blob/master/Contributing.md) have a broken link to "How to file a bug report." Could've sworn that worked before.

MichaelChirico:


mattdowle:
Looks good making it work. But I fear the approach of repeated strptime() with different formats until it works could be very slow.
Accepting and merging as something working is better than not working, but expecting to come back to it to fix speed.

MichaelChirico:
@mattdowle right, it's not great. But of course, if `format` is specified all of that can be obviated. Would be better to have a format detector, in the long run.

PR #1804, opened 2016-08-08
Progress towards #1648; improvements to metadata function tables()

MichaelChirico:
As discussed in #1669, here we are hammering away at #1648 a bit more gradually.

Only added the `index` argument here.


MichaelChirico:
should probably update the secondary indices/auto-indexing vignette to reflect this way of exploring a `data.table`'s `indices` once this is merged

PR #1815, opened 2016-08-17
Need note about set2key name change

dracodoc:
I was confused about `Error: could not find function "setindex"` when I read this vignette first. After some search it seemed to be an new name only available in newer version of data.table.

Since many reader are using the CRAN version 1.9.6, it's better to give a note about the name change.

My edit is just a placeholder, not necessarily in the best format.

### DIFF
--- a/vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
+++ b/vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
@@ -76,7 +76,7 @@ names(attributes(flights))
 ```
 
 * `setindex` and `setindexv()` allows adding a secondary index to the data.table.
-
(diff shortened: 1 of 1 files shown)


MichaelChirico:
I would instead change this as follows:

```
* `setindex` and `setindexv()` allows adding a secondary index to the `data.table` (historical note: earlier versions of the package called this function `set2key`, see [#1442](https://github.com/Rdatatable/data.table/issues/1442)).
```

dracodoc:
Besides, if it's assumed that user don't need to create second index explicitly: 

See FAQ 3.4
>Typically you don’t need to create secondary indexes. They are created automatically and used for you automatically by using data.table normally.

Then maybe this vignette also need some revises to reflect this.

PR #1816, opened 2016-08-17
put test data in string instead of external csv

dracodoc:
The rendered html vignette linked from the wiki doesn't give reference to this vignette folder, thus reader may have difficulties in finding the external csv file.

Since the data is small, it can be written in string and read by fread. I think this way reader can copy the code and test by himself/herself easier.


jangorecki:
LGTM, commit could also remove csv files that are not being used any more.

arunsrinivasan:
Valid point, but please create issues first before creating PRs (unless they are minor) so that things can be discussed, and I/we also have some time to think about it. It'll also save your time.

The same also happens for `flights14.csv`. I'd rather that it works with files as input to `fread()`.

dracodoc:
@arunsrinivasan got it, will create issue first next time!

When I was submitting suggestions for Hadley's book `R packages`, he prefer to create PR for each suggestion. Somehow it became a habit of me and I forgot this is not that book :)

Issue #1695, opened 2016-05-09
Completing the roster of .IDate methods with companion .Date methods


MichaelChirico:
Comparing

```
methods(class = "IDate")
#  [1] as.Date     as.list     as.POSIXct  c           coerce      cut        
#  [7] initialize  mean        rep         round       seq         show       
# [13] slotsFromS3 split       unique  
```

With 

```
methods(class = "Date")
#  [1] as.character  as.data.frame as.data.table as.IDate      as.list      
#  [6] as.POSIXct    as.POSIXlt    Axis          c             coerce       
# [11] cut           -             [<-           [             [[           
# [16] +             diff          format        hist          initialize   
# [21] is.numeric    julian        Math          mean          months       
# [26] Ops           pretty        print         quarters      rep          
# [31] round         seq           show          slotsFromS3   split        
# [36] str           summary       Summary       trunc         weekdays     
# [41] weighted.mean xtfrm
```

We see there's a lot of minor stuff to take care of to get `IDate` on par with `Date` for built in methods (and to help make sure we're not accidentally converting to `Date` along the way).

Specifically, we can add the following:

 - [ ] as.character
 - [ ] as.data.frame
 - [ ] as.data.table
 - [ ] as.POSIXlt
 - [ ] Axis
 - [ ] -
 - [ ] [<-
 - [ ] [
 - [ ] [[
 - [ ] diff
 - [ ] format
 - [ ] hist
 - [ ] is.numeric
 - [ ] julian
 - [ ] Math
 - [ ] mean
 - [ ] months
 - [ ] Ops
 - [ ] pretty
 - [ ] print
 - [ ] quarters
 - [ ] split
 - [ ] str
 - [ ] summary
 - [ ] Summary
 - [ ] trunc
 - [ ] weekdays
 - [ ] weighted.mean
 - [ ] xtfrm

jangorecki:
but they should be dispatched to next method, so `Date` method, we may only need to adjust class on returning object.
Anyway I would prefer to wait for #1451

MichaelChirico:
The hope is that there's some efficiency gains, especially if the `Date` method forces numeric before we force integer again.

I'm gonna handle this myself, so no need to divert resources from more important things.

franknarf1:
Was about to post a new issue (titled "replace() returns invalid IDates") before I saw this. There are good reasons for at least some of these besides efficiency, I guess.

Here's an example for `[<-.IDate`:

    library(data.table)
[...]

PR #1829, opened 2016-08-29
Closes #1765 -- new isoweek function produces ISO 8601-consistent week numbering as an integer

### LINKED ISSUE #1765: week() does not return actual calendar week
Hi,

the week() function does not return the actual calendar week (what I think most users expect).
Please fix or rename function since a lot of users are expecting a different result.

Thank you!

Example:

```
### Packages
library(data.table)
library(lubridate)

### Thursday, 7th january 2016, calendar week 1
my_date <- as.Date("2016-01-07", format = "%Y-%m-%d")

### Week from data.table
data.table::week(my_date)
> 2 # not what most people would expect

### Week from lubridate
lubridate::isoweek(my_date)
> 1 # as expected
```

MichaelChirico:
A few notes:

* Do we want `isoyear` as well?

* Feels clunky to need `as.POSIXlt` twice, but `lubridate::isoweek` does the same essentially. Could _maybe_ get away with doing some direct manipulation of the `POSIXlt` list elements, but things are complicated since [we can't add/subtract from `POSIXlt` objects](http://stackoverflow.com/questions/8857287/how-to-add-subtract-time-from-a-posixlt-time-while-keeping-its-class-in-r). 

* Any interest in adding an `Idifftime` class? Related as well to #1695...

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -100,7 +100,7 @@ S3method(na.omit, data.table)
 
 # IDateTime support:
 export(as.IDate,as.ITime,IDateTime)
-export(second,minute,hour,yday,wday,mday,week,month,quarter,year)
+export(second,minute,hour,yday,wday,mday,week,isoweek,month,quarter,year)
 
 export(as.chron.IDate, as.chron.ITime)
 export(as.Date.IDate) # workaround for zoo bug, see #1500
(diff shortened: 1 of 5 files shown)


arunsrinivasan:
The next release will most likely have internal (efficient) implementations. So yes, better to have R versions for now..

PR #1835, opened 2016-09-01
Closes #1832; improves documentation clarity for dcast.data.table

### LINKED ISSUE #1832: Add explanation of specials in dcast to documentation; other clean-up of ?dcast
Currently, `?dcast` reads:

> ...There are two special variables: `.` and `...`. Their functionality is identical to that of `reshape2::dcast`.

Since `reshape2` is not a dependency (currently just a Suggests), there's no reason to believe all users have this installed; nor is the link to online help sufficient for users working offline. 

We should just copy-paste from the `reshape2` manual the explanation of these:

> There are a couple of special variables: "..." represents all other variables not used in the formula and "." represents no variable...

Also, there are no examples with `.` or `...` in `?dcast`.

MichaelChirico:


arunsrinivasan:
Thanks Michael.

Issue #1664, opened 2016-04-20
fwrite(): final items


mattdowle:
- [x] malloc()s and write() need a thread safe way to error() if fail
- [x] Add progress % and ETA to appear after 2 seconds if more than 2 seconds are remaining
~~Include R_CheckUserInterrupt and test closes worker team ok~~ 1a4263f
- [x] Date, IDate and update [this quesion](http://stackoverflow.com/questions/37346684/how-to-avoid-converting-date-to-idate-in-r-data-table-fwrite-1-9-7)
- [ ] POSIXct  (see #1656)
- [x] ITime
- [x] integer64
~~Add "NEW:" item to startup banner~~
- [x] Confirm fwrite() writes 10GB ok on Windows (it should do) to ensure 'big' file > 4GB ok.  Thanks to Hugh Parsonage for testing as we don't have Windows other than via AppVeyor for test suite.
- [x] quote = 'auto'
- [x] sep2  (see #806) 
- [x] match write.csv's scientific/decimal format exactly and add many tests. 6c1ed96
- [x] add dec='.' to R level and connect to already existing option at C level
- [x] add row.names for data.frames, default FALSE.
- [x] rename `setthreads` to `setDTthreads` so as not to affect other packages using OpenMP, add reference to its manual in fwrite man
- [x] refine ?fwrite
- [x] more tests qmethod= 'double' and 'escape'

mattdowle:
See #1656 for nice option to provide on base types.

eantonya:
Do people actually like having `quote=TRUE` when writing to csv? I find it to be a big nuisance and would much prefer for `fwrite` to have `quote=FALSE` by default.

MichaelChirico:
I find `quote = TRUE` to be more robust -- you never know when you have a `JAMES SMITH, JR` in a `character` column and it can be a huge pain to get a .csv read when it has nuisance commas strewn about.

mattdowle:
@eantonya  Agree.  I prefer quote=FALSE too.   The base R thinking I believe has numbers/ids with leading 0's stored as character format ... the default ensures they get read by Excel as character and the leading 0's not lost.  But fwrite could detect that situation and quote just that situation by default.  Where character columns contain letters and no embedded quotes, I really don't see why quotes are needed.  Plus we save a bit on file size if character widths are small by saving the 2 extra quotes per field.

mattdowle:
[...]

PR #1838, opened 2016-09-02
Interim reset of default for quote on fwrite

MichaelChirico:
See discussion on #1664 and [here on SO](http://stackoverflow.com/questions/39282655/r-fwrite-separates-a-dataframe-entry-that-has-commas).

Agree that mid- to long-term solution is to make `fwrite` smarter and optimize the `quote = "auto"` setup, but in the meantime, having `quote = "auto"` is only hurting, it seems.


mattdowle:
Many thanks for the PR.  Sorry I wasn't following at the time.
`quote='auto'` fully implemented now and the default.

MichaelChirico:
no worries! obviously this PR didn't consume much of my time & also clearly
having quote='auto' up and running is far superior. thanks!

PR #1839, opened 2016-09-04
Closes #1836; adds - method for IDate, with substantial speed-up

### LINKED ISSUE #1836: IDate seems to be slower than Date
Example data

```
library(data.table)
set.seed(1)
id <- 1:1e4
date_samp <- seq(as.IDate("2010-01-01"),as.IDate("2011-01-01"),"day")
n_samp <- 5e7
DT <- data.table(
  id = sample(id,size = n_samp, replace=TRUE),
  date_1 = sample(date_samp,size = n_samp, replace=TRUE)
)
setkey(DT,id,date_1)

# also with integers
DT[, idt := as.integer(date_1)]
# also with vanilla dates
DT[, dt := as.Date(date_1)]
```

(Don't ask me why the data is keyed; I'm just following the lead of this SO post: http://stackoverflow.com/a/39283484/ )

Attempts

```
# the obvious way with IDate
system.time(DT[, diff := date_1 - shift(date_1), by=id]) # 25.90 seconds

# the obvious way with Date
system.time(DT[, diff2 := dt - shift(dt), by=id]) # 1.92

# other ways
system.time(DT[, 
  r := idt - shift(idt), by=id]) # 0.69

system.time(DT[, 
  r2 := idt - shift(idt)][
   idt != shift(idt), r2 := NA_integer_]) # 0.61
```

So IDate is a lot slower than Date for this fairly common task, it looks like.

(EDIT: Oops, was accidentally making difftimes into dates there. No effect on the timing.)

MichaelChirico:
We can skirt around the `as.POSIXct.IDate` sluggishness vs. precision tradeoff and still solve the molasses-like subtraction of `IDate`s.

Running @franknarf1 's example on this branch, I find we're now _faster_ than subtracting `Date`s (which should be the case in the first place since we're subtracting integers -- but more so since I tweaked the approach of `-.Date` to be a bit faster).

One potential funkiness is that I'm secretly supplying an `integer` vector to become a `difftime` object, which everywhere else would have a `numeric` vector underneath... not sure how un-kosher that is. Time to create `Idifftime`?

Timings on my current crappy machine:

```
library(data.table)
set.seed(1)
id <- 1:1e4
date_samp <- seq(as.IDate("2010-01-01"),as.IDate("2011-01-01"),"day")
n_samp <- 5e7
DT <- data.table(
  id = sample(id,size = n_samp, replace=TRUE),
  date_1 = sample(date_samp,size = n_samp, replace=TRUE)
)
setkey(DT,id,date_1)

# also with vanilla dates
DT[, dt := as.Date(date_1)]

system.time(DT[, diff := date_1 - shift(date_1), by=id])
#    user  system elapsed 
#   1.775   0.168   1.945 

system.time(DT[, diff2 := dt - shift(dt), by=id])
#    user  system elapsed 
#   3.954   0.213   4.173 
```

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -107,6 +107,7 @@ export(as.Date.IDate) # workaround for zoo bug, see #1500
 
 S3method("[", ITime)
 S3method("+", IDate)
+S3method("-", IDate)
 S3method(as.character, ITime)
 S3method(as.data.frame, ITime)
 S3method(as.Date, IDate)
--- a/NEWS.md
+++ b/NEWS.md
@@ -218,7 +218,7 @@
   
   58. Handled use of `.I` in some `GForce` operations, [#1683](https://github.com/Rdatatable/data.table/issues/1683). Thanks gibbz00 from SO and @franknarf1 for reporting and @MichaelChirico for the PR.
   
-  59. Added `+.IDate` method so that IDate + integer doesn't revert to `Date`, [#1528](https://github.com/Rdatatable/data.table/issues/1528); thanks @MichaelChirico for FR&PR.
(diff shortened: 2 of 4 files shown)


mattdowle:
Update in the issue here: https://github.com/Rdatatable/data.table/issues/1836#issuecomment-320476994

Issue #1837, opened 2016-09-02
fwrite garbage output for NA with turbo=T


philippechataignon:
I create a data.table with a integer missing value :

```
> z1 <- data.table(x=NA_integer_)
> fwrite(z1, "/tmp/df.csv")
```

csv file's contents is :

```
x
-./,),(-*,(
```

With turbo=F, file is OK :

```
x

```

philippechataignon:
Fix:

there is a missing *else* in `writeInteger` function in `src/fwrite.c`, so that if x==NA_INTEGER and na_len==0, the default, then do nothing.

Actual:

```
static inline void writeInteger(int x, char **thisCh)
{
  char *ch = *thisCh;
  if (x == NA_INTEGER) {
    if (na_len) { memcpy(ch, na_str, na_len); ch += na_len; }
  } if (x == 0) {
    *ch++ = '0';
  } else {
    if (x<0) { *ch++ = '-'; x=-x; }
    // avoid log() call for speed. write backwards then reverse when we know how long
    int width = 0;
    while (x>0) { *ch++ = '0'+x%10; x /= 10; width++; }
    for (int i=width/2; i>0; i--) {
      char tmp=*(ch-i);
      *(ch-i) = *(ch-width+i-1);
      *(ch-width+i-1) = tmp;
    }
  }
  *thisCh = ch;
}
```

Fixed:
```
static inline void writeInteger(int x, char **thisCh)
{
  char *ch = *thisCh;
  if (x == NA_INTEGER) {
    if (na_len) { memcpy(ch, na_str, na_len); ch += na_len; }
  } else if (x == 0) {
    *ch++ = '0';
  } else {
    if (x<0) { *ch++ = '-'; x=-x; }
    // avoid log() call for speed. write backwards then reverse when we know how long
    int width = 0;
    while (x>0) { *ch++ = '0'+x%10; x /= 10; width++; }
    for (int i=width/2; i>0; i--) {
      char tmp=*(ch-i);
      *(ch-i) = *(ch-width+i-1);
      *(ch-width+i-1) = tmp;
    }
  }
  *thisCh = ch;
}
```

MichaelChirico:
thanks philippe! do you want to file a PR?

arunsrinivasan:
TODO: add tests before closing.

PR #1842, opened 2016-09-06
Fix garbage output for NA_integer in fwrite

philippechataignon:
When a data.table contains a integer missing value,
`fwrite` writes it with a garbage contents.

This patch adds a missing *else* in `writeInteger` function in
`src/fwrite.c` : if x==NA_INTEGER and na_len==0, the
default, then do nothing.

Fix issue #1837


arunsrinivasan:
Thanks!

arunsrinivasan:
TODO: add tests

PR #1853, opened 2016-09-22
Closes #1852 -- print(null.data.table()) no longer visibly returns NULL

### LINKED ISSUE #1852: data.table:::print.data.table sometimes returns visibly, however it should be invisible
```
x=data.table(1)[0]  # an empty data.table
```

Typing `x` on its own calls the data.table print method:

```
> x
Empty data.table (0 rows) of 1 col: V1
```

Typing `print(x)` should give the same result as above, however there is extra output:

```
> print(x)
Empty data.table (0 rows) of 1 col: V1
NULL
```

Very low priority I guess, but a bug nonetheless (I think).

I think it's because, in such cases, the return statement (within the code for `data.table:::print.data.table`) is as follows:

```
return()
```

However, it should return invisibly, as expected by the `?print` documentation

> ‘print’ prints its argument and returns it _invisibly_ (via
>      ‘invisible(x)’). 

The method should return the value (not `NULL`) and it should return invisibly. Something like:

```
return(invisible(x))
```

MichaelChirico:


inst/tests/tests.Rraw
```diff
@@ -9122,6 +9122,10 @@ if (.Platform$OS.type=="unix") {
     unlink(f)
 }
 
+# Fix #1852 -- calling print on null data.table output NULL
+test(1704, dput(capture.output(print(data.table:::null.data.table()))),
```
jangorecki: Better to avoid testing `:::` as this isn't user api, `data.table(NULL)` will works here. Also `dput` seems redundant.

MichaelChirico: Fixed. Thanks for the extra set of eyes @jangorecki

PR #1870, opened 2016-10-07
R code cleanups (nchar, sapply)

mllg:
While reading some source files of data.table, I've spotted you are using `sapply()` instead of `vapply()`. The latter is type safe and usually faster. Furthermore, you are using `nchar(x) == 0L` instead of the faster `nzchar()`. 
I quickly grepped over the package and replaced the respective functions. My changes probably won't have a big impact on performance, but you basically get it for free.

I've also removed an outdated backport of `provideDimnames()`.

### DIFF
--- a/R/as.data.table.R
+++ b/R/as.data.table.R
@@ -30,28 +30,10 @@ as.data.table.Date <- as.data.table.ITime <- function(x, keep.rownames=FALSE, ..
     as.data.table.list(x, FALSE)
 }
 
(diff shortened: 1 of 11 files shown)


jangorecki:
@mllg Thanks for PR. Any chance to move third commit to separate PR? I'm quite comfortable on merging `provideDimnames` and `nzchar`, but not as much about `v*apply` functions.

PR #1938, opened 2016-11-29
Minor spelling and formatting fixes to a few help pages

JoshOBrien:
Made the following four fixes:

repitition -> repetition
bioconductor -> Bioconductor
an usage -> a usage
code{yday} -> \code{yday}

Unfortunately, my Emacs seems also to have stripped trailing whitespace from a bunch of lines, so the pull request ends up looking a lot bigger than it is (and may be annoying if you want that trailing whitespace).

### DIFF
--- a/man/IDateTime.Rd
+++ b/man/IDateTime.Rd
@@ -31,13 +31,13 @@
 \alias{split.IDate}
 \alias{second}
 \alias{minute}
-\alias{hour}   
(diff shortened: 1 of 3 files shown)


MichaelChirico:
[here](https://github.com/Rdatatable/data.table/pull/1938/files?w=1)'s the diff log with those whitespace trims ignored

PR #2062, opened 2017-03-17
Closes #1973; lazyeval edge case of IDateTime caused an error in as.D…

### LINKED ISSUE #1973: IDateTime fails on POSIXct without tzone.
The function IDateTime generates an error when applied to POSIXct values without tzone attribute:

    library(data.table)
    IDateTime(Sys.time())
    # Error in if (tz == "UTC") { : argument is of length zero

Versions
R: 3.3.2
data.table: 1.10.0 and latest 1.10.1

MichaelChirico:
…ate.POSIXct

See discussion @ #1973. Quite a strange bug, not sure I fully wrap my head around how lazyeval is working here... may be that there's a better solution?


inst/tests/tests.Rraw
```diff
@@ -9910,6 +9910,11 @@ for (i in 1:100) test(1763, rbindlist(x, idcol="id"), ans)
 # as.ITime(character(0)) used to fail, #2032
 test(1764, capture.output(format(as.ITime(character(0)))), "character(0)")
 
+# IDateTime error when tzone is NULL, #1973
+x = as.POSIXct('2017-03-17')
+attr(x, 'tzone') = NULL
+test(1765, print(IDateTime(x)), output=".*idate.*itime.*1: 2017-03-17 00:00:00")
```
jangorecki: @MichaelChirico 
This test on my machine produces:
```
  Test 1765 didn't produce correct output:
  > print(IDateTime(x)) 
  Expected: '.*idate.*itime.*1: 2017-03-17 00:00:00'
  Observed: '        idate    itime1: 2017-03-16 00:00:00'
```
If timezone set to `"Poland"`. If timezone set to `"GB"` test pass fine.
Poland is +1 hour vs GB so it shifts the date already. It should be fixed so users from various timezones can run tests.
```
> sessionInfo()
R version 3.4.0 (2017-04-21)
Running under: Ubuntu 16.04.2 LTS

Matrix products: default

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8   

attached base packages:

[1] compiler_3.4.0
```

PR #2071, opened 2017-03-22
rbindlist default argument modification

jtrive84:
In man/rbindlist.Rd, changed use.names default argument in rbindlist function prototype from fill to TRUE.

### DIFF
--- a/man/rbindlist.Rd
+++ b/man/rbindlist.Rd
@@ -7,7 +7,7 @@
   Same as \code{do.call("rbind", l)} on \code{data.frame}s, but much faster. See \code{DETAILS} for more.
 }
 \usage{
-rbindlist(l, use.names=fill, fill=FALSE, idcol=NULL)
+rbindlist(l, use.names=TRUE, fill=FALSE, idcol=NULL)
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Why would you do this, when the actual default is `fill`?

```
data.table::rbindlist
function(l, use.names=fill, fill=FALSE, idcol=NULL) [...]
```

Also, please follow the [contribution guidelines](https://github.com/Rdatatable/data.table/blob/master/Contributing.md)

jtrive84:
Attempting to set `use.names=fill` within rbindlist throws an error. From the description of `use.names` from the rbindlist help page,  it can be set to :

**use.names**	
If TRUE items will be bound by matching column names. By default FALSE for rbindlist (for backwards compatibility) and TRUE for rbind (consistency with base). Columns with duplicate names are bound in the order of occurrence, similar to base. When TRUE, at least one item of the input list has to have non-null column names.

MichaelChirico:
I'm confused. if you set `use.names=fill` yourself explicitly and don't have `fill` defined in your calling scope, you'll get an error.

Issue #1465, opened 2015-12-11
Feature Request: finer control of `strip.white` in `fread`? Dealing with BOM


MichaelChirico:
I'm working with a series of files, one of which has the UTF-8 BOM marking the beginning of the file: `\0xef`  `\0xbb` `\0xbf`

As noted [here](http://stackoverflow.com/questions/21624796/read-the-text-file-with-bom-in-r), the default behavior of `read.csv` is now to detect and delete the BOM. Unfortunately, for me at least, `fread` seems to have converted the three characters into a space.

Fortunately, `strip.white` removes this before returning the `data.table`; unfortunately, my file also has lots of important trailing white space, so I need to set `strip.white = FALSE`, negating this.

Here's a link to the file I'm working with (_caveat clickor_: it's a scary executable link, and also non-trivial size, ~80 MB. For whatever reason they decided to "zip" the file with an executable. My only word of reassurance is that you can tell it's a [US government website](http://lbstat.dpi.wi.gov/lbstat_newasr)): http://lbstat.dpi.wi.gov/sites/default/files/imce/lbstat/exe/11STAFF.exe

To see the BOM, run:

    r<-readBin("11STAFF.txt",raw(),file.info("11STAFF.txt")$size)
    > r[1:10]
     [1] ef bb bf 30 30 30 30 36 37 31
    > r[1] == as.raw(0xef)
    [1] TRUE

Here's some relevant output from `fread` with `verbose = TRUE`:

    > fread("11STAFF.txt", sep = "^", header = FALSE, verbose = TRUE)
    ...
    First 10 characters: ﻿0000671

That is, it has treated the first 3 characters as being a space. With `strip.white = TRUE`, this space disappears in the output.

I compare this to the behavior of `read.csv` (also a nuisance to use because the file is on the large side):

```
> read.csv("11STAFF.txt", sep = "^", header = FALSE, stringsAsFactors = FALSE)$V1[1]
[1] "000067182Abel                Nancy           FW19554    2011R187  70 70  45880  21809            1  00070007030020530050KGKG1616N100              Abbotsford Sch Dist           Abbotsford Elementary         61010Clark County                  04PO Box A                      Abbotsford WI  54405-0901                                   510 W Hemlock St              Abbotsford WI  54405                                        Abbotsford       WI54405-0901Abbotsford       WI54405     715-223-4281      Gary Gunderson                                    NNN                                                  "
```

[...]

Issue #1087, opened 2015-03-19
Fread should exclude UTF-8 BOM from column names


nigmastar:
Currently, `fread` seems to include the utf-8 BOM inside the file.

```
library(data.table)

# Create file

f <- "test_utf-8.csv"
cat(intToUtf8(c(239, 187, 191)), 'a,b,c\n1,2,3\n', file = f, sep = '')

# Import file

dt <- fread(f)
names(dt)
## [1] "ï»¿a" "b"    "c"
```

jangorecki:
is there an easy workaround for that?

httassadar:
Not sure if the issue title makes sense, why file names?

But the post does describe the issue I really hope it's fixed.

arunsrinivasan:
The file you provide is incorrectly encoded.

```r
> readBin("test_utf-8.csv", raw(), file.info("test_utf-8.csv")$size)
#  [1] c3 af c2 bb c2 bf 61 2c 62 2c 63 0a 31 2c 32 2c 33 0a
```

whereas it should be:

```r
#  [1] ef bb bf 61 2c 62 2c 63 0a 31 2c 32 2c 33 0a
```

And performing `fread("test_utf-8.csv")` seems to work fine, although the BOM is still included:

```r
ans = fread("test_utf-8.csv")
#    ﻿a b c
# 1: 1 2 3
charToRaw(names(ans))
# [1] ef bb bf 61
```

arunsrinivasan:
With this fix, now I get:

```r
charToRaw(names(ans)[1L])
# [1] 61
```

PR #2084, opened 2017-03-28
Improvements to BOM detection: 

st-pasha:
if BOM is found then set encoding to UTF8; print a message about BOM in verbose mode; detect BOM in GB18030 and UTF16 encodings and issue warning / error appropriately


mattdowle:
Can you remove whitespace diffs please.
Please retain the nice comment with references to #1087 and #1465.
If a function is only ever called once, then I prefer it not to be a function but inside alongside where it opens the file in one place.  That saves hopping around to helper functions.  If a function is called twice or more, then it can be a function.
What's variable `size`?   Yes I see it's not compiling for that reason.

PR #2095, opened 2017-03-31
Closes #2082 -- split.data.table respects factor ordering

### LINKED ISSUE #2082: split.data.table does not respect factor ordering
Pretty self-explanatory. MWE:

```
DT = data.table(a = factor(c('a', 'b'), levels = c('b', 'a')))
split(DT, by = 'a', sorted = TRUE)
# $a
#    a
# 1: a

# $b
#    a
# 1: b
split(DT, f = DT$a, sorted = TRUE)
# $b
#    a
# 1: b

# $a
#    a
# 1: a
split.data.frame(DT, f = DT$a)
# $b
#    a
# 1: b

# $a
#    a
# 1: a
```

MichaelChirico:
Simple enough fix. When `CJ` was called with `sorted = TRUE`, it was reversing the order because `by` was being passed as an alphabetically sorted `character` instead of a `factor`.


inst/tests/tests.Rraw
```diff
@@ -9877,6 +9877,13 @@ test(1757, fread("winallquoted.csv")[c(1,2,4998,4999)],
 
 test(1758, sapply(fread("A,B\n,"),class), c(A="logical",B="logical"))
 
+# split.data.table should respect non-alphabetic order if passed a factor in by, #2082
+DT = data.table(a = factor(c('a', 'b'), levels = c('b', 'a')))
+split(DT, by = 'a', sorted = TRUE)
```
mattdowle: This line can be deleted iiuc?

inst/tests/tests.Rraw
```diff
@@ -9877,6 +9877,13 @@ test(1757, fread("winallquoted.csv")[c(1,2,4998,4999)],
...
+test(1759, split(DT, by = 'a', sorted = TRUE),
```
mattdowle: Another tests with more than one column to check this fix works in combination with other columns would be great please.

MichaelChirico:
@mattdowle done thanks. let me know if you prefer a squashed single commit

PR #2127, opened 2017-04-22
Switch to new Travis configuration format

st-pasha:
Travis now has native R support, changing the .yaml file to use the new config format.
See https://github.com/craigcitro/r-travis/wiki/Porting-to-native-R-support-in-Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ dist: trusty
 sudo: required
 cache: packages
 warnings_are_errors: true
-r_check_args: 
+r_check_args: ""
 # empty r_check_args to turn off --as-cran (on by default) as that can be slow
 
 bioc_packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,23 @@
-# Sample .travis.yml for R projects.
 #
-# See README.md for instructions, or for more configuration options,
-# see the wiki:
-#   https://github.com/craigcitro/r-travis/wiki
-# 
-# Hack for 'error on warning' except when the warning is about vignette pdfs/html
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,38 +1,24 @@
...
-  - mv tmp.sh travis-tool.sh
-  - chmod 755 ./travis-tool.sh  
-  - ./travis-tool.sh bootstrap
-install:
-  - ./travis-tool.sh bioc_install IRanges
-  - ./travis-tool.sh bioc_install GenomicRanges
-  - ./travis-tool.sh install_deps
-  
-script:
-  - echo "Commit:" $TRAVIS_COMMIT >> ./DESCRIPTION
```
jangorecki: This info got lost.

.travis.yml
```diff
@@ -1,38 +1,24 @@
...
-  - ./travis-tool.sh run_tests
+bioc_packages:
+  - IRanges
+  - GenomicRanges
 
-after_failure:
-  - ./travis-tool.sh dump_logs
+r_github_packages:
```
jangorecki: We might not need to install them from github, CRAN versions were most likely updated till now.

.travis.yml
mattdowle: @jangorecki  Is it as simple as adding that line back in?  I'm looking in upgrade instructions and I can't see "script:" mentioned.

jangorecki: Putting this into `before_install` is probably a way to go, at least this one is documented in travis native r doc. This metadata is used to check if current installed version commit matches to the one in drat. No need to install or download pkg then. If there is no such metadata it needs to download and reinstall every time.

PR #2150, opened 2017-05-08
retain timezones on CJ()

### LINKED ISSUE #2029: [Bug] CJ() looses timezone of POSIXct vector
I have a problem when using CJ() with POSIXct objects.

If CJ is used to construct a data.table from a POSIXct vector with timezone UTC and a second grouping vector, the resulting POSIXct column in the data.table has local timezone. 
If the POSIXct vector is the only input to CJ, the timezone is preserved. 
Also, plain data.table() preserves the timezone. 

See below for a reproducible example and my sessionInfo(). Maybe related to (https://github.com/Rdatatable/data.table/pull/1778)

```
library(data.table)

## get groups and dates as input for CJ()
input_groups <- letters[1:3]
input_times  <- as.POSIXct(sprintf("2016-%02d-01", 1:12), tz = "UTC")

## input_times timezone is UTC
all(lapply(input_times, attr, "tzone") == "UTC")
# [1] TRUE

## the following shows that CJ changes timezones to NULL (local time).
test <- CJ(group = input_groups, test_time = input_times)
## test$test_time timezone is not UTC but NULL
all(lapply(test$test_time, attr, "tzone") == "UTC")
# [1] FALSE
unique(lapply(test$test_time, attr, "tzone"))
# [[1]]
# NULL

## if only a single POSIXct vector is supplied, timezone is preserved.
test2 <- CJ(test_time = input_times)
all(lapply(test2$test_time, attr, "tzone") == "UTC")
# [1] TRUE

## also, original data.table() behaves as expected.
test2 <- data.table(group = rep(input_groups, length(input_times)), date = rep(input_times, length(input_groups)))
setkey(test2, group, date)
all(lapply(test2$date, attr, "tzone") == "UTC")
# [1] TRUE
```

Here is my sessionInfo()

```
R version 3.3.0 (2016-05-03)

locale:

attached base packages:

other attached packages:
[1] data.table_1.10.4

[1] tools_3.3.0

```

Kind regards,
Markus Bonsch

RoyalTS:
fixes #2029


R/setkey.R
```diff
@@ -332,6 +332,7 @@ CJ <- function(..., sorted = TRUE, unique = FALSE)
     dups = FALSE # fix for #1513
     # using rep.int instead of rep speeds things up considerably (but attributes are dropped).
     j = lapply(l, class)  # changed "vapply" to avoid errors with "ordered" "factor" input
+    tzones = lapply(l, function(col) attr(col, 'tzone'))
```
mattdowle: This can be  tzones = lapply(l, attr, 'tzone')  please and can it be moved down inside the `if()` so that it is only calculated when needed?

R/setkey.R
```diff
@@ -353,6 +354,9 @@ CJ <- function(..., sorted = TRUE, unique = FALSE)
                 l[[i]] = rep.int(rep.int(y, times = rep.int(x[i], n[i])), times = nrow/(x[i]*n[i]))
             if (any(class(l[[i]]) != j[[i]]))
                 setattr(l[[i]], 'class', j[[i]]) # reset "Date" class - rep.int coerces to integer
+            if (any(!sapply(tzones, is.null))) {
```
mattdowle: Are the tzone for all columns being looked at here,  within a loop through columns?   It doesn't seem quite right to me but I haven't debugged or run it.   See next comment on tests.

inst/tests/tests.Rraw
```diff
@@ -9898,6 +9898,10 @@ unlink(f)
 test(1761.1, fread("1\n2\n3", fill=TRUE), data.table(V1=1:3))
 test(1761.2, fread("1\n2\n3", fill=FALSE), data.table(V1=1:3))
 
+# CJ should retain timezone information, #2029
+df <- CJ(week=as.POSIXct('2016-01-01', tz = 'UTC'), id=1:10)
+test(1762, attr(df$week, 'tzone'), 'UTC')
```
mattdowle: More tests please including multi-column data.table with 1 and 2 columns of POSIXct with other non-POSIXct (say plain integer) columns at the beginning, middle and end.  Enough variations to test that the code is actually retaining the tzone on the desired columns and not adding tzones to columns that shouldn't, because I'm not sure it's right currently just by glancing at it.  The test is just a single column.

R/setkey.R
RoyalTS: Makes sense. Was trying to mirror the code that copies over classes. Any reason the `j = ...` line above my edit isn't inside the `else if`?

inst/tests/tests.Rraw
RoyalTS: You mean something like`CJ(dt1=as.POSIXct('2016-01-01', tz = 'UTC'), id=1:10, dt2=as.POSIXct(c('2016-01-01', '2016-01-02'), tz = 'UTC'))` (behaves as expected) and various permutations of the order of these columns?

MarkusBonsch:
Dear RoyalTS,

thanks a lot for taking care of this. One suggestion / proposal:
can't we extend this to all attributes by simply
`attribs <- lapply(l, attributes)`
instead of
`tzones = lapply(l, function(col) attr(col, 'tzone'))`
and 
`attributes(l[[i]]) <- attribs[[i]]`
instead of
`setattr(l[[i]], 'tzone', tzones[[i]])`
? This would be a consistent solution for all attributes that might pop up in the future.

R/setkey.R
MarkusBonsch: I think that this if-statement can be simply removed.
It serves no real purpose. If the `tzones` are all `NULL`, well, let them all be `NULL` and set them anyways.

RoyalTS: I suppose this could make sense if `is.null` is significantly faster than just setting the attributes even if they're `NULL`? In any case, the same would go for the lines that reset class immediately above?

MarkusBonsch: The difference to the line above is, that it tests for each `i`. 
So, if we want to retain the tzone if-statement, it should read:
`if(!is.null(tzones[[i]])){...}`. That would make more sense in my opinion.

MarkusBonsch:
I have posted a CJ implementation that consistently retains all arguments and addresses Matt's concerns below. Additionally, I have proposed a comprehensive set of tests of the new implementation that also addresses Matt's comments.

I conducted a benchmark for a ten-column CJ input to test if the `if`-statement speeds up things 

|     expr   |   min|  lq |               mean       |median       |uq          |max           |neval|
| ---- | ----- | --- | --- | ---- | ------------- | ------------- | ------------- |
| CJ_master  | 0.878273  | 1.209456  | 6.242917  | 1.762042   | 3.674923   | 185.8787   | 10000
 | 
| CJ_attributes_with_if | 0.871674  | 1.215139  | 6.244461  | 1.940190   | 3.676939   | 192.1245   | 10000
   | 
CJ_attributes_no_if  | 0.872408  | 1.209823  | 6.735057  | 1.850016  | 3.686286    | 184.1207   | 10000| 

Both implementations are just as fast as the master. Therefore, I would propose to keep the `if` because it is a little safer since attributes are only updated if necessary.

There is, however, a speed issue when all ten columns have an attribute that needs to be replaced:

 | expr      |min       |lq     |mean   |median       |uq      |max |neval|
| ---- | ----- | --- | --- | ---- | ------------- | ------------- | ------------- |
|CJ_master |0.905031 |1.389070 |4.556453 |1.690564 |3.792404 |185.8065 |10000 |
|CJ_attributes_with_if |1.306778 |2.585514 |8.374277 |4.480982 |4.905639 |187.9216 |10000   |
|CJ_attributes_no_if |1.275621 |2.599443 |8.327741 |4.498394 |4.911871 |184.9495 |10000|

Again, there is no difference between the implementation with `if` and without `if`. And it is not surprising that retaining attributes comes at a cost. The good thing is, that this cost really only arises if attributes are actually present. 

Kind regards, 
Markus

Here is the code:
```
# CJ implementation that retains attributes
CJ <- function(..., sorted = TRUE, unique = FALSE){
    # Pass in a list of unique values, e.g. ids and dates
    # Cross Join will then produce a join table with the combination of all values (cross product).
    # The last vector is varied the quickest in the table, so dates should be last for roll for example
    l = list(...)
    if (unique) l = lapply(l, unique)
    dups = FALSE # fix for #1513
    if (length(l)==1L && sorted && length(o <- forderv(l[[1L]])))
        l[[1L]] = l[[1L]][o]
    else if (length(l) > 1L) {
        # using rep.int instead of rep speeds things up considerably (but attributes are dropped).
        attribs = lapply(l, attributes)  # remember attributes for resetting after rep.int
        n = vapply(l, length, 0L)
        nrow = prod(n)
        x = c(rev(take(cumprod(rev(n)))), 1L)
        for (i in seq_along(x)) {
            y = l[[i]]
            # fix for #1513
            if (sorted) {
                if (length(o <- forderv(y, retGrp=TRUE))) y = y[o]
                if (!dups) dups = attr(o, 'maxgrpn') > 1L 
            }
            if (i == 1L) 
                l[[i]] = rep.int(y, times = rep.int(x[i], n[i]))   # i.e. rep(y, each=x[i])
            else if (i == length(n))
                l[[i]] = rep.int(y, times = nrow/(x[i]*n[i]))
            else
                l[[i]] = rep.int(rep.int(y, times = rep.int(x[i], n[i])), times = nrow/(x[i]*n[i]))
            if (!is.null(attribs[[i]])){
                attributes(l[[i]]) <- attribs[[i]] # reset all attributes that were destroyed by rep.int
            }
        }
    }
    setattr(l, "row.names", .set_row_names(length(l[[1L]])))
    setattr(l, "class", c("data.table", "data.frame"))

    if (is.null(vnames <- names(l))) 
        vnames = vector("character", length(l)) 
    if (any(tt <- vnames == "")) {
        vnames[tt] = paste("V", which(tt), sep="")
        setattr(l, "names", vnames)
    }
    l <- alloc.col(l)  # a tiny bit wasteful to over-allocate a fixed join table (column slots only), doing it anyway for consistency, and it's possible a user may wish to use SJ directly outside a join and would expect consistent over-allocation.
    if (sorted) {
        if (!dups) setattr(l, 'sorted', names(l)) 
        else setkey(l) # fix #1513
    }
    l
}
```

And here are the tests
```
# CJ retains attributes and classes, #2150

l <- list(a = as.POSIXct(c("2016-01-01", "2017-01-01"), tz = "UTC"),
          b = as.POSIXct(c("2016-01-01", "2017-01-01")),
          c = as.Date("2015-01-01"), ## according to comment about CJ loosing date class
          d = factor(c("a", "b", "c"), ordered = TRUE), ## according to comment about bug with ordered factors
          e = factor(c("a", "b", "c"), ordered = FALSE),
          f = c(1,2),
          g = c("a", "b"),
          h = c(TRUE, FALSE))
setattr(l$g, "test", "testval")## add hand-made attribute

test(1762.1, lapply(l, attributes), lapply(do.call(CJ, l), attributes))
test(1762.2, lapply(l, class), lapply(do.call(CJ, l), class))

l <- list(a = factor(c("a", "b", "c"), ordered = TRUE),
          b = as.POSIXct(c("2016-01-01", "2017-01-01")),
          c = as.Date("2015-01-01"),
          d = factor(c("a", "b", "c"), ordered = TRUE),
          e = as.POSIXct(c("2016-01-01", "2017-01-01"), tz = "UTC"),
          f = c(1,2),
          g = c("a", "b"),
          h = c(TRUE, FALSE))


test(1762.3, lapply(l, attributes), lapply(do.call(CJ, l), attributes))
test(1762.4, lapply(l, class), lapply(do.call(CJ, l), class))

l <- list(a = factor(c("a", "b", "c"), ordered = TRUE),
          b = as.POSIXct(c("2016-01-01", "2017-01-01")),
          c = as.Date("2015-01-01"),
          d = factor(c("a", "b", "c"), ordered = TRUE),
          e = c(TRUE, FALSE),
          f = c(1,2),
          g = c("a", "b"),
          h = as.POSIXct(c("2016-01-01", "2017-01-01"), tz = "UTC"))

test(1762.5, lapply(l, attributes), lapply(do.call(CJ, l), attributes))
test(1762.6, lapply(l, class), lapply(do.call(CJ, l), class))

l <- list(a = NA,
          c = c(1,2),
          d = as.POSIXct("2016-01-01", tz = "UTC"))

test(1762.7, lapply(l, attributes), lapply(do.call(CJ, l), attributes))
test(1762.8, lapply(l, class), lapply(do.call(CJ, l), class))
```

RoyalTS:
This is great! How should we do this? Close this PR and you'll create one of your own?

MarkusBonsch:
I am not familiar with github. Can't you give me push rights for your PR? Or you just insert the suggestions yourself, and tell in the news that I also contributed.
as you prefer.

RoyalTS:
Just made the changes, all tests are passing. Can we get another reviewer pass on this?

MarkusBonsch:
Do you know, why this AppVeyor check fails? It leaves this nasty red cross at the project.

RoyalTS:
Just more recent patches creating merge conflicts. All resolved, this is now mergeable again in principle.

PR #2160, opened 2017-05-12
Add configurable parameter nSamplingPoints to fread(), 

### LINKED ISSUE #2157: [Request] Provide an option to override fread's jumpiness
Reading [this file](https://gist.github.com/franknarf1/6c55b0c9c080b9dadf2266537aa28978#file-simulated-data) with fread, I get ...

>Error in fread("bah.csv") : 
> Internal error: Sampling jump point 10 is before the last jump ended

It would be nice to have an option to tell `fread` to traverse the file naively (or whatever), so I don't have to write a wrapper to catch this error and take it to read.csv.

Tested with...

```
data.table 1.10.5 IN DEVELOPMENT built 2017-05-09 05:01:26 UTC; travis

R version 3.3.3 (2017-03-06)
Running under: Windows 10 x64 (build 14393)

locale:

attached base packages:

other attached packages:
[1] data.table_1.10.5

[1] tools_3.3.3

```

st-pasha:
which provides work-around for occasional cases when default number of jump points is too high, causing overlap among sampled regions. Closes #2157


mattdowle:
Fixed the error in https://github.com/Rdatatable/data.table/commit/57c056f77f1cc284d923ee07e2765259b87dba15 so the workaround parameter isn't needed.

PR #2200, opened 2017-06-12
Fix issue #2194 (crash when reading a file with size divisible by 4096)

### LINKED ISSUE #2194: fread segfaults when reading a file with size a multiple of 4k
Happens in line 768: `*_const_cast(eof) = eol;`
The error is:
```
 *** caught bus error ***
address 0x1012b9000, cause 'non-existent physical address'
```
[4k.txt](https://github.com/Rdatatable/data.table/files/1058979/4k.txt)

st-pasha:
Outline of the change:

* The idea of writing `\n` to the last byte of the memory-mapped region sometimes fails to work, and so will be removed. Instead, we'll use the following strategy: search the file from the end for the last newline, and then move the last line (appending a newline) into a separate memory buffer. This splits the entire file into 2 separate parts: "main" part `sof .. eof`, and "hidden" part `soh .. eoh`. For example, if the input was `A,B\n1,2\n3,4`, then the first part will become `A,b\n1,2\n`, and the second `3,4\n`. Both parts will have valid newline endings.

* The iteration over the file will have to be organized more carefully: while previously it was as simple as
    ```
    const char *ch = sof;
    while ((ch < eof) && <condition>) ...;
    ```
    Now it becomes more complicated:
    ```
    const char *ch = sof, *end = eof;
    while (((ch < end) || (eoh && (end != eoh) && (ch=soh) && (end=eoh))
           && <condition>) ...;
    ```
    fortunately, this doesn't add any runtime overhead (assuming compiler does proper short-circuit evaluation).

* All parser functions will no longer require any `ch < eof` tests: the test `*ch == eol` is sufficient to test the end of the field. The `eof` variable is even no longer global (it can't be: different threads might be scanning different portions of the file simultaneously). This change is the original reason why we are going to such trouble to ensure there's always an `eol` at the end file. There is only 1 problem here: quoted string fields with embedded newlines.

    In order to circumvent this difficulty, we change the semantics of the `_Bool Field(const char **ptr, void *target)` function (currently this function parses the next string field and returns true/false indicator of success/failure; on success it also advances the pointer `ptr` to the end of the field just read). In the new semantics this function has `int` return value, 0 meaning success, 1 failure, and 2 "continuation failure". The latter status is returned for fields that are correctly quoted but the parser hit a newline before the field ended (thus, the `Field()` parser will stop parsing at a newline like all other parsers do). When status 2 is returned, the parser will also fill out the `lenOff target` structure with detected offset and length until the newline, and it will advance the pointer `ptr`.

    Additionally, there will be a second field parsing function `int continue_parse_string(const char **ptr, lenOff *target)`, which will (a) assume that it begins inside a quoted field, (b) in case of successful scan increments `target->off` instead of overwriting it. This function also returns status 0 for success (field scanned till the end), 1 for failure (field is invalid under current quoting rule, for example quote is not followed by a field separator), and 2 for "continuation failure" (i.e. the field didn't end on this line: keep scanning). Both statuses 0 and 2 advance pointer `ptr` and increment `target->off`.

* Now the field parser logic will have to become more complicated as well: if previously it was
    ```
    while (!fun[type[j]](&ch, target[j])) {
      type[j]++;
      ch = fieldStart;
    }
   ```
    Now it is more like this:
    ```
    while (ret = fun[type[j]](&ch, target[j])) {
      while (ret == 2) {
        ch += eolLen;
        if (ch == end) {
          if (end == eof) { ch = soh; end = eoh; }
          else { ret = 1; break; }
        }
        ret = continue_parse_string(&ch, target[j]);
      }
      if (ret == 1) {
        type[j]++;
        ch = fieldStart;
      }
      // if `ret` is 0 then don't do anything: field was parsed correctly
    }
    ```
    This adds small amount of overhead, but only for multi-line string fields, which are sufficiently rare. Overall runtime impact should be negligible for the most common cases.

* There are also lots of minor code changes: added `ASSERT(cond)` macro that expands into `if (!cond) STOP("Internal error on line %d, please report", __LINE__)` -- this is intended for checks that should in theory never fail (and thus shouldn't show up red on the coverage report); the `pos` variable was removed, instead the `sof` variable is used to track the beginning of data section (it would be too hard to juggle `pos` / `sof` / `soh` triple).

Closes #2194


mattdowle:
That was fun!!

The goal of all this was to avoid checking `ch<eof` for every single byte of input just because some files end abruptly without a final newline. We don't want to have to ask user to `echo>>filename.txt` and they might not have write permission to do that anyway.

Master had been my last attempt but that created a bus error when `fileSize%4096==0`.
Pasha added tests for this and fixed in this PR#2200 but code complexity in `fread.c` went up.
Now trying a different cleaner approach keeping Pasha's tests verbatim and fixing a few other long-standing related concerns as well.

* For files which abruptly end without final newline, it now writes `\0` to the last byte of cow mmp page which is reliable on all platforms. Otherwise, it finds the last newline and writes `\0` there (again, cow last page).
* Since it writes `\0` now rather than newline, uncertainty of writing `eol` when `eolLen==2` goes away.
* No allocation or copy needed for either final field or final line; `finalByte` stored instead.
* No copy means `eof` doesn't need to be moved which was hard because `eof` is shared across threads.
* Does not read or write `eof+1` so bus error when `fileSize` is multiple of 4096 goes away.
* Isolated code to deal with final field in one place. This does not run other than for files which end abruptly without final newline. So if problems do occur, `echo>>file.txt` has a better chance of fixing the problem because the new code for abrupt-ending won't then run. However, abruptly ended files should just work and they are tested and user should not need to add the final new line.
* Fewer lines of code changed (236: +161 -75) than PR#2200 (847: +567 -280). No new functions. Fewer new variables. No arguments added to existing functions. However, I didn't remove `ch<eof` in processors yet to keep that separate which would have increased the changed line count.
* Now that `*eof=='\0'` can be relied upon, run-away `strstr()` problem goes away if `skip="string"` is not present.
* Newline detection is now scanned from the end backwards to avoid any unclosed quoted fields on the first line (again, thanks to Pasha spotting). Still not perfect but much better and not worse in any cases. Intend to remove `eol` detection altogether anyway to allow for mixed line ending types.

PR #1483, opened 2015-12-31
implements #1482: adds option to print.data.table to exclude lower register of column names when printing more than 20 rows

MichaelChirico:
also updated the R-Forge issue number to the corresponding GitHub issue number. Not sure if that's kosher...

Issue #645, opened 2014-06-08
[R-Forge #1957] Optional Arguments for Printing


arunsrinivasan:
Submitted by: Yike Lu; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=1957&group_id=240&atid=978)

It would be nice to have a few simple optional arguments for displaying/printing data.tables, for example:
1) Turn off row numbers,
2) Turn off smart table wrapping
3) Force print all entries
4) Put spaces between "by" groupings for higher groupings... e.g. DT[, sum(x), by=list(a,b,c)]  prints as

a    b     c   V1
1    1     1
1    1     2

1    0     3

2    1     1
2    1     2
2    1     3

Perhaps a depth variable can be specified here.

5) underline/underscores for column header, horizontal bar to separate key and nonkey columns.

MichaelChirico:
1) & 3) are done, AFAIK.

Not sure anything can be done about 2), if I understand correctly?

4) Could go for some more formalization, not sure exactly what was intended.

First part of 5) seems pretty simple, maybe a `name.sep` or `bezel` argument; second part seems a bit tougher.

As mentioned in #1483 fulfilling this should probably entail creating an `.Rd` for `print.data.table`. And exporting `print.data.table`?

franknarf1:
Personally, 2 and 4 would be very handy. I have a wrapper to do 4 (though the BY vars still show up in my "blank" rows) and just expand my R console across more screens for 2. 

(I do this because I have been reviewing a feed of data periodically, summarized to about 20 groups, each having about four rows for subgroup.)

MichaelChirico:
@franknarf1 could you add some detail about what 4) means for you? I think I'm missing something in the formatting from R-Forge.

From how I read it, it's something like

    DT <- data.table(a = rep(1:2, each = 3), b = rep(1:3, 2))

    DT[ , 1L, by = .(a, b)]
```
DT[ , 1L, by = .(a, b)]
#    a b V1
# 1: 1 1  1
# 2: 1 2  1
# 3: 1 3  1
#                   # <- blank row here as we switch 'a' group from 1 to 2
# 4: 2 1  1
# 5: 2 2  1
# 6: 2 3  1
```

And the `depth` Yike had in mind was saying how frequently we use the spacing for more fine grouping -- if we did `.(a, b, c)` with `depth = 2`, we'd put spaces after every switch of `b`.

Does that sound like what you have in mind?

[...]

PR #2296, opened 2017-08-10
Closes #1482, adds options for controlling how print displays column names

### LINKED ISSUE #1482: FR: Option to exclude the lower register of column names in printing
From this SO question: http://stackoverflow.com/questions/34511495/how-to-disable-data-table-names-at-the-bottom

Some users want the option to skip printing the column names at the bottom (only happens when the whole table is printed and there are more than 20 rows)

This is an easy fix with almost no overhead, so I don't see the harm in adding the option (I suppose the solution could conflict with the other associated bug fix, #97, but that seems unlikely).

Will add a PR for this soon.

MichaelChirico:
More progress towards #1523. I'm not convinced we need the three options anymore... in particular wondering whether `col.names = "none"` would actually be used. See also #1483.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,7 +32,11 @@
 
 8. `as.data.table()` gains new method for `array`s to return a useful data.table, [#1418](https://github.com/Rdatatable/data.table/issues/1418).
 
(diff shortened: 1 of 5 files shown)


franknarf1:
> I'm not convinced we need the three options anymore... in particular wondering whether col.names = "none" would actually be used.

`col.names = "none"` might suit my [use-case mentioned](https://github.com/Rdatatable/data.table/issues/645#issuecomment-321895416) in #645 . If I `split` by group and then loop over `print`, I may want to skip col names after the first table, for example.

PR #2302, opened 2017-08-15
Fix NOTE bc extra files

jangorecki:
fixes --as-cran NOTE
```
* checking top-level files ... NOTE
Non-standard files/directories found at top level:
  ‘cc.R’ ‘Makefile’
```
and (this only affects gitlab builds)
```
Unknown, possibly mis-spelled, fields in DESCRIPTION:
  ‘Commit’
```

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -10,3 +10,5 @@
 ^\.RData$
 ^\.Rhistory$
 ^data\.table_.*\.tar\.gz$
+^cc\.R$
+^Makefile$
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,7 +36,6 @@ build:
   script:
     - Rscript -e 'install.packages("knitr", repos=file.path("file:",normalizePath("bus/mirror-packages/cran")))'
     - rm -r bus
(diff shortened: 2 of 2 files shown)


mattdowle:
Thanks for spotting!
I thought at the time those files were added that it was strange tests passed.  It's only under `--as-cran` that non-standard files/directories are checked then?  I wonder why not all the time.  Not enough time to ask r-devel why that is (and if no good reason, suggest that's changed in R) so moving on.

jangorecki:
@mattdowle yes, check for those files happens only when using `--as-cran` and it [seems to be](https://github.com/wch/r-source/blob/0fbfc3cf1d3c76e663fc2ca863e6fd1330bff5ee/doc/manual/R-ints.texi#L4079) intentional, but can be controlled with env var.
```
@item _R_CHECK_TOPLEVEL_FILES_
Report on top-level files in the package sources that are not described
in `Writing R Extensions' nor are commonly understood (like
@file{ChangeLog}).  Variations on standard names (e.g.@:
@file{COPYRIGHT}) are also reported.
Default: false (but true for CRAN submission checks).
```

PR #2329, opened 2017-09-05
Do not include symbol on lhs of := in othervars

### LINKED ISSUE #2326: .SD mistakenly includes column being set when get() appears in j
When I use `:=` together with `.SD`, `.SDcols` and `get()`, for example, 

```r
dt[, z := ncol(.SD) + get("y"), .SDcols = "x"]
```

The first time is ok, but whenever `z` already exists, the code above will produce incorrect results since `.SD` includes both `x` and `z`, which should be `x` only, as restricted by `.SDcols`.

The following code is a minimal example and demonstrates the problem:

```r
library(data.table)

dt <- data.table(x = seq(1, 10), y = seq(10, 1))
dt[, z := ncol(.SD) + y, .SDcols = "x"]
dt
dt[, z := ncol(.SD) + get("y"), .SDcols = "x"]
dt

dt[, z := {
  str(.SD)
  ncol(.SD) + get("y")
}, .SDcols = "x"]

dt[, z := NULL]
dt[, z := {
  str(.SD)
  ncol(.SD) + get("y", inherits = FALSE)
}, .SDcols = "x"]
```

The output looks like 

```
> library(data.table)
> dt <- data.table(x = seq(1, 10), y = seq(10, 1))
> dt
     x  y
 1:  1 10
 2:  2  9
 3:  3  8
 4:  4  7
 5:  5  6
 6:  6  5
 7:  7  4
 8:  8  3
 9:  9  2
10: 10  1
> dt[, z := ncol(.SD) + y, .SDcols = "x"]
> dt
     x  y  z
 1:  1 10 11
 2:  2  9 10
 3:  3  8  9
 4:  4  7  8
 5:  5  6  7
 6:  6  5  6
 7:  7  4  5
 8:  8  3  4
 9:  9  2  3
10: 10  1  2
> dt[, z := ncol(.SD) + get("y"), .SDcols = "x"]
> dt
     x  y  z
 1:  1 10 12
 2:  2  9 11
 3:  3  8 10
 4:  4  7  9
 5:  5  6  8
 6:  6  5  7
 7:  7  4  6
 8:  8  3  5
 9:  9  2  4
10: 10  1  3
> dt[, z := {
+   str(.SD)
+   ncol(.SD) + get("y")
+ }, .SDcols = "x"]
Classes ‘data.table’ and 'data.frame':	10 obs. of  2 variables:
 $ x: int  1 2 3 4 5 6 7 8 9 10
 $ z: int  12 11 10 9 8 7 6 5 4 3
 - attr(*, ".data.table.locked")= logi TRUE
> dt[, z := NULL]
> dt[, z := {
+   str(.SD)
+   ncol(.SD) + get("y", inherits = FALSE)
+ }, .SDcols = "x"]
Classes ‘data.table’ and 'data.frame':	10 obs. of  1 variable:
 $ x: int  1 2 3 4 5 6 7 8 9 10
 - attr(*, ".data.table.locked")= logi TRUE
```

renkun-ken:
This is an attempt to fix #2326.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -74,6 +74,8 @@
 
 16. `split.data.table` respects `factor` ordering in `by` argument, [#2082](https://github.com/Rdatatable/data.table/issues/2082). Thanks to @MichaelChirico for identifying and fixing the issue.
 
+17. `.SD` would incorrectly include symbol on lhs of `:=` when `.SDcols` is specified and `get()` appears in `j`. Thanks @renkun-ken for reporting and the PR. Closes #2326.
+
 #### NOTES
 
 1. `?data.table` makes explicit the option of using a `logical` vector in `j` to select columns, [#1978](https://github.com/Rdatatable/data.table/issues/1978). Thanks @Henrik-P for the note and @MichaelChirico for filing.
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1052,7 +1052,7 @@ chmatch2 <- function(x, table, nomatch=NA_integer_) {
                 # fix for long standing FR/bug, #495 and #484
                 allcols = c(names(x), xdotprefix, names(i), idotprefix)
 
-                # Do not include z in othervars when j is z := expr
+                # Do not include z in othervars when j is z := expr (#2326)
                 av_allcols <- intersect(av, allcols)
                 if (is.call(jsub) && length(jsub[[1L]]) == 1L && jsub[[1L]] == ":=" && is.symbol(jsub[[2L]])) {
                   av_allcols <- setdiff(av_allcols, as.character(jsub[[2L]]))
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1054,7 +1054,7 @@ chmatch2 <- function(x, table, nomatch=NA_integer_) {
 
                 # Do not include z in othervars when j is z := expr
                 av_allcols <- intersect(av, allcols)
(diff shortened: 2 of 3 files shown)


MichaelChirico:
Hi Kun, thanks!

Please be sure to include tests and make sure the updated package compiles and that `test.data.table()` runs without error -- in particular I think it's likely there's a reason the LHS was included in `othervars`, which should be covered by another test, so this bug fix will have to allow that use case to continue to function.

See here:
https://github.com/Rdatatable/data.table/blob/master/Contributing.md

mattdowle:
Hi Kun,
Thanks from me too.  In case you are on Ubuntu, I recently added my dev cycle in the cc.R file in the project root.  That's how to easily make changes and run all tests, at least how I do it.  Once that is passing you can do the longer R CMD build and R CMD check before pushing the pull request.  If you are on Mac or Windows then cc.R will likely need some tweaking.
Hope that helps,
Matt

renkun-ken:
Thanks for reminding! I'm working on macOS at the moment (or may turn to Ubuntu if necessary) and will need a little while to get familiar with `data.table`'s testing and find out what breaks previous test cases.

mattdowle:
Great PR!!  Thanks!

MichaelChirico:
Nice @renkun-ken! The PR process can be a steep learning curve, glad you could navigate it :nerd_face:

Issue #2331, opened 2017-09-06
[Request] setkeyv can be accelerated if key already exists


MarkusBonsch:
setkeyv() can be accelerated significantly in cases where the key already exists.
This can be useful e.g. if you create a function that takes a data.table as an argument and you need to set a key but don't know if the user has already set the key on the input.
With the new implementation, you can just use setkey without worrying about speed penalties.

setkeyv() does two things:
1. determine the correct sorting calling `forderv`
2. Rearranging the data.table by reference calling `Creorder`

Currently, if the key already exists, the call to `forderv` is still executed and only step 2 is skipped.
The only reason is a sanity check that the data.table is really sorted by the key.
I believe, it is not necessary to perform this sanity check each time, especially since it has been around for quite a while so that potential bugs should have popped up.

mattdowle:
Great! Your PR merged.  See comment there :  https://github.com/Rdatatable/data.table/pull/2332#issuecomment-327575497

PR #2332, opened 2017-09-06
faster setkeyv if keys already exist.

MarkusBonsch:
Adresses issue #2331.

A small benchmark highlights the benefit of the new implementation (code at the end of the comment):

DT is a data.table with 1e7 rows. x1 and x2 are columns with 10 groups each (1:10), x3 is numeric random numbers.
Time is median time in milliseconds of 100 iterations

|   |   | master | pull request |
|--- | --- | --- | --- | 
| 1 | no existing key ; setkey(DT, x1, x2) | 441 | 445 |
| 2 | existing key on x1 ; setkey(DT, x1, x2) | 284 | 286 |
| 3 | existing key on x1 and x2 ; setkey(DT, x1, x2) | 111 | 49 |
| 4 | existing key on x1, x2, and x3 ; setkey(DT, x1, x2) | 110 | 49 |

Cases 3 and 4 see a significant speed improvement since forderv is not called.

The only drawback is that the internal check that the data.table is really sorted by the key is not performed anymore. This has two implications that are absolutely acceptable from my perspective:
1. The internal logic of data.table is not tested in each call of setkeyv. I believe testing the sanity of the internal logic is a task for `test.data.table()`, and should not hamper the speed of a function that is called frequently.
2. If users 'go under the hood' and manipulate the `sorted` attribute manually, it is not guaranteed that the key is still correct. I believe that users that go under the hood should be expected to know what they do. No need to maintain a costly check for this very specific case.

Of course, I had to remove the test that checked this warning.

Here is the code for my benchmark:

```
library(data.table)
library(microbenchmark)

## timing for setkey
set.seed(100)
nrow <- 1e7
dat_nokey <- data.table(x1 = sample(1:10,nrow, replace = TRUE), x2 = sample(1:10, replace = TRUE), x3 = rnorm(nrow))

dat_onekey <- copy(dat_nokey)
setkey(dat_onekey, x1, verbose = TRUE)

dat_twokey <- copy(dat_nokey)
setkey(dat_twokey, x1, x2, verbose = TRUE)

dat_threekey <- copy(dat_nokey)
setkey(dat_threekey, x1, x2, x3, verbose = TRUE)

timing_setkey <- microbenchmark(nokey =    setkey(copy(dat_nokey), x1, x2),
                                onekey =   setkey(copy(dat_onekey), x1, x2),
                                twokey =   setkey(copy(dat_twokey), x1, x2),
                                threekey = setkey(copy(dat_threekey), x1, x2),
                                times = 100)

## test that all results are identical (except for threekey which is expected to be different)
nokey    = setkey(copy(dat_nokey), x1, x2)
onekey   = setkey(copy(dat_onekey), x1, x2, verbose = TRUE)
twokey   = setkey(copy(dat_twokey), x1, x2)
threekey = setkey(copy(dat_threekey), x1, x2)

if(!identical(nokey, onekey)) stop("Error")
if(!identical(nokey, twokey)) stop("Error")
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,6 +37,8 @@
     * gains `print.keys` argument, `FALSE` by default, which displays the keys and/or indices (secondary keys) of a `data.table`. Thanks @MichaelChirico for the PR, Yike Lu for the suggestion and Arun for honing that idea to its present form.
     
     * gains `col.names` argument, `"auto"` by default, which toggles which registers of column names to include in printed output. `"top"` forces `data.frame`-like behavior where column names are only ever included at the top of the output, as opposed to the default behavior which appends the column names below the output as well for longer (>20 rows) tables. `"none"` shuts down column name printing altogether. Thanks @MichaelChirico for the PR, Oleg Bondar for the suggestion, and Arun for guiding commentary.
+    
+10. setkeyv accelerated if key already exists [#2331](https://github.com/Rdatatable/data.table/issues/2331). Thanks to @MarkusBonsch for the PR.
 
 
 #### BUG FIXES
--- a/R/setkey.R
+++ b/R/setkey.R
@@ -45,7 +45,18 @@ setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TR
         miss = !(cols %in% colnames(x))
         if (any(miss)) stop("some columns are not in the data.table: " %+% cols[miss])
     }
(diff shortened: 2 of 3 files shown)


mattdowle:
Well spotted!  Thanks for the benchmarks, explaining why test 347 was removed up front and including the news item.  That all made it easy to accept.

PR #2337, opened 2017-09-07
Fixed .shallow to consistently retain keys and indices.

### LINKED ISSUE #2336: [BUG] .shallow messes up keys and indices
.shallow can retain keys with the argument `retain.key = TRUE`. However, this feature suffers from two problems that need to be solved before the function can be exported as requested in #2323:

1. Key gets lost if more columns than key are chosen with the `cols` argument.
```
library(data.table)
DT <- data.table(x1 = sample(1:10, 100, replace = TRUE),
                 x2 = sample(1:10, 100, replace = TRUE),
                 x3 = sample(1:10, 100, replace = TRUE),
                 y  = rnorm(100), 
                 key = c("x1", "x2", "x3"))
## as expected
identical(key(data.table:::.shallow(DT, retain.key = TRUE)), c("x1", "x2", "x3"))
# TRUE
identical(key(data.table:::.shallow(DT, cols = c("x1", "x2", "x3"), retain.key = TRUE)), c("x1", "x2", "x3"))
# TRUE
## key gets lost for no reason
test <- data.table:::.shallow(DT, cols = c("x1", "x2", "x3", "y"), retain.key = TRUE)
identical(key(test), c("x1", "x2", "x3"))
# FALSE
## although the copy is of course sorted properly
data.table:::is.sorted(test, by = c("x1", "x2", "x3"))
# TRUE
```

2. Indices remain, even if the columns are dropped
```
setindex(DT, x2, y)
setindex(DT, x1, y)

## index just gets copied, no matter if the columns are not present anymore
test <- data.table:::.shallow(DT, cols = c("x3"), retain.key = TRUE)
names(attributes(attr(test, "index")))
# "__x2__y" "__x1__y"
names(test)
# "x3"
```

MarkusBonsch:
Closes #2336.

Keys and indices are retained correctly by shallow if `retain.key = TRUE` and `cols != NULL`.
The largest possible part of an index is retained.
If the original index was on x1, x2, x3, and `cols = c("x1", "x2"), 
the shallow copy gets the same index for c("x1", "x2") (if there was a native index on x1 and x2, this is kept)

Tests have been added.

A benchmark (code at end of this post) shows that there is no negative speed impact:

| master | PR |
| --- | --- |
| 33 ms | 33 ms |

The improved key retainment makes it possible to clean `setattr` calls in `foverlaps.R` while still passing all tests.

Code for benchmark

```
library(data.table)
library(microbenchmark)

nrow <- 1e6
ncol <- 20
nindex <- 20
ncopycol <- 10

DT <- data.table(x1 = rnorm(nrow))
setindex(DT, x1)
index <- c("x1")
for(col in seq_len(ncol-1)){
  DT[, paste0("x", col + 1) := rnorm(nrow)]
  if(col < nindex){
    index <- c(index, paste0("x", col+1))
    setindexv(DT, index)
  }
}

test <- microbenchmark(shallow = data.table:::.shallow(DT, 
                                                       cols = paste0("x", seq_len(ncopycol)), 
                                                       retain.key = TRUE), 
                       times = 100, 
                       unit = "ms")
```


MarkusBonsch:
Gets a problem on foverlaps test in travis that I don't get locally. Will need to investigate.

MarkusBonsch:
Initially, my local tests didn't include fOverlaps since genomicRanges was not installed. Now, everything should be fixed.

PR #2355, opened 2017-09-13
Make fread handle cases when allocnrow is too small

### LINKED ISSUE #2246: Crash in fread in dev when under-estimating the number of rows
```
require(data.table)
DT1 = data.table(A=rep(123, 100), B=456)
DT2 = data.table(A=rep(1, 200), B=2)
DT = rbind(DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT2, DT1, DT1)
fwrite(DT, f<-tempfile())
ff = fread(f)
```
Normally this produces
```
Error in fread(f) : 
  Internal error: DTi(2900) > allocnrow(2208) but nrows=9223372036854775807 (not limited)
```
but occasionally also
```
R(22039,0x7fff7504e000) malloc: *** error for object 0x7f957ec29608: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

st-pasha:
* Behavior of `allocateDT` is modified to handle the case when it is called with a different number of `nrows` than previously
* If too many rows is found, break out of the main reading loop into a single-threaded context; then call `allocateDT` with a new estimated number of columns; and finally restart the main reading loop from exactly the same point where we have left, avoiding the need to rescan the entire file again.
* Added a test
* Added verbose output in case such "nrows bump" is encountered; and also verbose output about the parameters of the main scan: `nJumps`, `chunkBytes` and `lastRowEnd-pos`

Closes #2246


src/fread.c
```diff
@@ -1604,14 +1609,20 @@ int freadMain(freadMainArgs _args) {
...
   nth = imin(nJumps, nth);
 
+
   read:  // we'll return here to reread any columns with out-of-sample type exceptions
+  {
+  if (verbose) {
+    DTPRINT("[11] Read the data\n");
+    DTPRINT("  jumps=[%d..%d), chunk_size=%llu, total_size=%llu\n", jump0, nJumps, (llu)chunkBytes, (llu)(lastRowEnd-pos));
+  }
+  ASSERT(allocnrow <= nrowLimit);
```
mattdowle: Need the values of `allocnrow` and `nrowLimit` included in the error message please if assert fails.

PR #2363, opened 2017-09-15
Add parser for extended floating point numbers

### LINKED ISSUE #1800: Handle 1.#INF, 1.#IND, 1.#QNAN and 1.#SNAN
In a [pull request](https://github.com/Rdatatable/data.table/pull/1788) I was asked to open an issue on this subject, so here it is.

The reason for the proposed patch for `fread` was that C/C++ code compiled with a visual studio compiler will output e.g. `1.#INF` instead of `inf` or `#1.IND` instead of `nan` when writing such floating point values to a text file using a `printf` like function. Since R compiles its packages using gcc, it will not recognize these strings as doubles, causing entire columns to be interpreted as text instread of numbers.

The current patch always checks for this, and therefore causes a slowdown of 2.2%. I've also tried making it optional by specifying an extra boolean argument to `fread`. This then sets a certain function pointer to e.g. `strtod` directly, or to the modified code in `strtod_wrapper` that performs the extra checks. However, even such a straightforward change still causes a slowdown of 1.5% when the extra checks are _not_ used.

st-pasha:
Closes #1800


mattdowle:
NB: fails are due to knitr/R-devel unrelated to this PR.  master has the same problem will have to look at separately.

st-pasha:
Fixed the parser so that it returns `NAND` as before.
Added new issue #2363 to make sure that we fix the `test()` function at some point so that it can catch problems like these.

mattdowle:
After adding the stricter `identical`, one of them (1830.6) fails with a NA/NaN difference to be sorted out.

PR #2367, opened 2017-09-16
Fix warnings/compilation errors in fread.c

st-pasha:
Some trivial modifications:

* Replaced `restrict` with `__restrict__` (the former is not available on all platforms, the latter is a cross-platform macro);
* Replaced variable `this` with `pch` ("this" is a keyword in C++);
* Replaced `_Bool` with `bool` as defined in `<stdbool.h>` (the former is not available in some compilers);
* `<omp.h>` is now included from "fread.h";
* fix few warnings related to typecasts.

Also, following Matt's comment, this PR fixes the parsers for int32/int64/double fields so that internal counters in them would not accidentally overflow when given a number that is slightly longer than the width of `int8_t` type. Tests for those edge cases are also added.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -23,7 +23,7 @@
   #include <stdio.h>     // snprintf
   #include <math.h>      // ceil, sqrt, isfinite
 #endif
-#include <omp.h>
+#include <stdbool.h>
 #include "fread.h"
 #include "freadLookups.h"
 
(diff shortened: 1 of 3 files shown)


src/fread.c
```diff
@@ -653,7 +654,7 @@ static void parse_double_regular(FieldParseContext *ctx)
     acc = 10*acc + digit;
     ch++;
   }
-  sf = ch - ch0;
+  sf = (uint_fast8_t)(ch - ch0);
```
mattdowle: This helped me see that `sf` shouldn't be declared as 8 bits on line 649.  A field longer than 256 digits wouldn't get caught. If the overflow in `sf` left say `sf==5`,  an incorrect numeric would be read.  Perhaps `sf` should be `uint_fast32_t`, and declared here too.  Maybe even `size_t` which is the only guaranteed type to correctly subtract any two pointers.

src/fread.c
```diff
@@ -203,23 +203,23 @@ static void printTypes(int ncol) {
 }
 
 
-static inline void skip_white(const char **this) {
+static inline void skip_white(const char **ptr) {
```
mattdowle: `this` wasn't a good name and `ptr` seems too generic too.   How about `myPos`, `threadPos`, `threadLoc`, `threadCh`, `myCh`, or something like that.   Some name to convey it's used to update the calling thread's position.

st-pasha: Agree, `ptr` isn't very descriptive. However I'd avoid naming the parameter with anything "thread"-related -- this would create an impression that the functions have to be called from multi-threaded contexts only.
How about `inputPtr`, `inputCh`, `inputLoc` or `chPtr` ?

src/fread.c
st-pasha: good catch, we should probably add a test for this too...

PR #2374, opened 2017-09-21
Detect header when empty column names

### LINKED ISSUE #2370: [Regression] fread fails to detect headers when one of the names is empty
```
> fread(",A,B\n1,2,3\n2,5,6\n")
   V1 V2 V3
1: NA  A  B
2:  1  2  3
3:  2  5  6
```

mattdowle:
Closes #2370


mattdowle:
@MichaelChirico  To my surprise, it turns out the fix is affected by the `logical01` and was a good spot by Pasha. The pull is correctly failing currently.  When `logical01=FALSE`, the 2nd field writer function is chosen (type 2 rather than type 1) and that stops the fix from working.  I'll have to think of a better fix!

There are 19 occurrences of `header=` in the tests and I can some with TRUE and FALSE passed to `fread`.   Or did you mean in combination with missing column names too perhaps?

Issue #2388, opened 2017-09-27
[BUG] unloading data.table causes segmentation fault


MarkusBonsch:
The following piece of code causes a hard crash very often (not always):

```
library(data.table)

DT <- data.table(x1 = c(1,1,1,1,1,2,2,2,2,2),
                 x2 = c(1,1,2,2,2,1,1,2,2,2),
                 x3 = c(1,2,1,1,2,1,1,1,1,2),
                 y  = rnorm(10), 
                 key = c("x1", "x2", "x3"))

rm(DT)

detach("package:data.table", unload = TRUE)

```
I could confirm the problem for the current development version, the CRAN version, both under ubuntu 16.04 with R 3.4.1 and Windows with R 3.3.3.

While this might point to a C memory allocation problem, debugging with valgrind didn't produce any errors. 
The following code doesn't cause the crash, so it's definitely the object creation that messes things:

```
library(data.table)

detach("package:data.table", unload = TRUE)
```

Maybe I am just reporting known behaviour?

Cheers,
Markus

helix123:
I can confirm this with
R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1
data.table_1.10.4

Happens every 3rd time or so.

MarkusBonsch:
Has been found before: #990

jangorecki:
closing as dup of #990

Issue #2372, opened 2017-09-20
[Request] assigning with := should retain keys better


MarkusBonsch:
Currently, keys and secondary indices are dropped, if any of the key columns is altered via a `:=` assignment.
This is not necessary, however, since the key is still valid for all key columns before the assigned one.
Given this, we can better retain keys and secondary indices and improve data.table performance.
An example:
```
DT <- data.table(x1 = sample(1:10, 100, replace = TRUE),
                 x2 = sample(1:10, 100, replace = TRUE),
                 x3 = sample(1:10, 100, replace = TRUE),
                 y  = rnorm(100))

setkey(DT, x1, x2, x3)

## DT has a key on x1, x2, x3
print(key(DT))
# [1] "x1" "x2" "x3"

## assigning to x3 drops the key completely:
DT[2, x3 := 1000]
print(key(DT))
# NULL

## even though the sorting of x1 and x2 is of course not affected:
print(length(data.table:::forderv(DT, by = c("x1", "x2"))) == 0)
# TRUE


```

MarkusBonsch:
I am working on a solution. Just need to memCheck for memory leaks before creating a pull request

mattdowle:
This was closed by your PR #2389 - thanks! Awesome!  The text at the top of the PR "closes [#2372]" had brackets around the number so it didn't auto close this issue.  I edited that text after the merge but it still didn't link up.  So I guess it needs to be right at the time of the merge.

MarkusBonsch:
I have discovered a minor fault in my implementation: 

I am assuming that secondary indices behave exactly like the key. 
This is, however, wrong: 
* the key is just a piece of information on the existing ordering. 
* The secondary key is a recipe for reordering.

If there is a key on columns x and y, it is a fully valid key on x as well. If there is an index on columns x and y, it is also a fully valid index on x, but it may cause spurious reordering in y if used as an index for x. The only secondary index that is fully equivalent to a key is a `integer(0)` index. 
Although, there is currently no really clear policy on ordering based on indices (issue #2400), this can cause unexpected behaviour and needs to be fixed.

Therefore, I need to change my implementation as follows:
* if any column in a secondary index has been modified, drop it
* except if it is an `integer(0)` index: then, the index is still valid for all columns until the assigned one (exclusive)

I will fix asap and create a PR.

PR #2389, opened 2017-09-27
Improved key handling when assigning with ':='

MarkusBonsch:
This closes [#2372].

Now, when updating columns with :=, existing keys and indices are retained up to the first modified column (exclusive).
Example: key on x1, x2, x3 --> assignment on x2 --> key on x1.

The tests 1400 ff had to be modified since they were testing for the old behaviour.
Additional tests have been added to make sure, all cases are handled, also on empty data.tables.

I tested for C memory allocation problems with `R -d valgrind` and found no issues.
There is an issue with a segmentataion fault on unloading the package but I discovered that this is a problem of the master (issue [#2388])

A small benchmark (code at the end of this post) shows that there is no speed penalty associated with the new implementation:

| master | PR |
| --- | --- |
|  0.25 	ms | 0.27 ms |

```
library(data.table)
library(microbenchmark)

nrow <- 1
ncol <- 20
nindex <- 20
nassigncol <- 10

DT <- data.table(x1 = rnorm(nrow))
setindex(DT, x1)
index <- c("x1")
for(col in seq_len(ncol-1)){
  DT[, paste0("x", col + 1) := rnorm(nrow)]
  if(col < nindex){
    index <- c(index, paste0("x", col+1))
    setindexv(DT, index)
  }
}

test <- microbenchmark(assign = copy(DT)[, paste0("x", (ncol - nassigncol):ncol) := 1], 
                       times = 100, 
                       unit = "ms")

```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -49,6 +49,8 @@
     
 10. setkeyv accelerated if key already exists [#2331](https://github.com/Rdatatable/data.table/issues/2331). Thanks to @MarkusBonsch for the PR.
 
+11. improved key handling when assigning with ':=' [#2372](https://github.com/Rdatatable/data.table/issues/2372). Keys and secondary indices used to be droopped completely if one of the columns was affected by ':='. Now, as much as possible of the remaining keys and indices is retained for re-use. Thanks to @MarkusBonsch for the PR.
+
 
 #### BUG FIXES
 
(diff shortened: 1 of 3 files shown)


mattdowle:
Sorry - I guess I broke it by removing `shift` alias.  I didn't think that was needed as `shift` is exported. I'll fix.

I've added you as a project member so that you can create branches inside the main repo if you'd like.  This will mean I can pull, test and push to your branch, rather than me using the online editor as I'm doing now.

MarkusBonsch:
Thank you very much for the invitation! I am happy to join.

MarkusBonsch:
Thank you for cleaning up. I am still learning :)

Issue #2411, opened 2017-10-11
Group aggregate performance significantly decreses with more groups


renkun-ken:
With the same number of function being called in `j`, combined evaluation and separate evaluation by group will result in significantly different performance as the number of unique groups increases. 

```r
set.seed(123)
n <- 1e7
x <- data.table(a = rnorm(n), b = sample(1:10, n, replace = TRUE), c = sample(1:10, n, replace = TRUE))

system.time(x[, .(m = mean(a), s = sd(a)), by = .(b, c)][, m/s])
system.time(x[, .(mean(a) / sd(a)), by = .(b, c)])
```

Ideally, the two expressions being timed should not exhibit major difference in performance since the second is a combination of the first, that is, in both `j`, `mean(a)` and `sd(a)` must be calculated. However, the second expression can be extremely slow with more groups.

Here are some benchmarks with different number of groups of `(b, c)`:

```
> set.seed(123)
> n <- 1e7
> x <- data.table(a = rnorm(n), b = sample(1:10, n, replace = TRUE), c = sample(1:10, n, replace = TRUE))
> system.time(x[, .(m = mean(a), s = sd(a)), by = .(b, c)][, m/s])
   user  system elapsed 
  0.408   0.020   0.426 
> system.time(x[, .(mean(a) / sd(a)), by = .(b, c)])
   user  system elapsed 
  0.300   0.012   0.311 
> x <- data.table(a = rnorm(n), b = sample(1:100, n, replace = TRUE), c = sample(1:10, n, replace = TRUE))
> system.time(x[, .(m = mean(a), s = sd(a)), by = .(b, c)][, m/s])
   user  system elapsed 
  0.508   0.028   0.534 
> system.time(x[, .(mean(a) / sd(a)), by = .(b, c)])
   user  system elapsed 
  0.372   0.012   0.386 
> x <- data.table(a = rnorm(n), b = sample(1:1000, n, replace = TRUE), c = sample(1:10, n, replace = TRUE))
> system.time(x[, .(m = mean(a), s = sd(a)), by = .(b, c)][, m/s])
   user  system elapsed 
  0.568   0.020   0.584 
> system.time(x[, .(mean(a) / sd(a)), by = .(b, c)])
   user  system elapsed 
  0.636   0.004   0.640 
> x <- data.table(a = rnorm(n), b = sample(1:10000, n, replace = TRUE), c = sample(1:10, n, replace = TRUE))
[...]

PR #2412, opened 2017-10-12
clarify/correct man page for GForce

MichaelChirico:
Related to #2411, noticed some bad grammar/need for elucidation. Curious, is the bit about `order` -> `forder` still needed, since R uses `data.table`'s sorting by default?

### DIFF
--- a/man/datatable-optimize.Rd
+++ b/man/datatable-optimize.Rd
@@ -52,22 +52,24 @@ At optimisation level \code{>= 2}, i.e., \code{getOption("datatable.optimize")} 
 
 \itemize{
 
-    \item When expressions in \code{j} which contains only these functions
-    \code{min, max, mean, median, var, sd, prod}, for example,
-    \code{dt[, list(mean(x), median(x), min(y), max(y)), by=z]}, they are very
-    effectively optimised using, what we call, \emph{GForce}. These functions
-    are replaced with \code{gmean, gmedian, gmin, gmax} instead.
+    \item Expressions in \code{j} which contain only the functions
+    \code{min, max, mean, median, var, sd, prod} (for example,
+    \code{dt[, list(mean(x), median(x), min(y), max(y)), by=z]}), they are very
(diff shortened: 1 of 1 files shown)


franknarf1:
Reading `?order`, I see that "radix" is not the default method. Also, it looks like radix was added in 3.3.0 while data.table's current dependency is >= 3.0.0.

Btw, there are other cases where GForce goes into effect besides those in the doc now (`first`, `last`, others?). Not sure if you want to exhaustively list or write the doc to say "and others" or something.

MichaelChirico:
@franknarf1 thanks, good call.

I think an exhaustive list of `base` functions is appropriate, but we can use `g*` wildcard instead of listing again.

MichaelChirico:
I'm also seeing `g[` in the list from

    grep('^g', ls(getNamespace('data.table'), all = TRUE), value = TRUE)

That's fast `.SD[i]` subsetting, is that right?

franknarf1:
Yeah, I think so, but only if `i` is length-one. The [code for gnthvalue](https://github.com/Rdatatable/data.table/blob/aac8c32d14df26c311673c81f1e5b705cb1151dd/src/gsumm.c#L786) has an error about length > 1.

mattdowle:
@MichaelChirico  
> Curious, is the bit about `order` -> `forder` still needed, since R uses `data.table`'s sorting by default?

Good point. The documentation needs an update then, and perhaps that optimization can be phased out. We still currently need it for older versions of R and the base R defaults changed some time after data.table's forder was promoted to base, too.  There's a vignette somewhere which has a demo of forder being faster than base::order which is no longer true, too.  It would be nice if the documentation said that I in fact proposed the change to R in my presentation at useR! 2015 in Denmark (slide 2). Since I've heard it said that's its a shame base R is faster now and that I can't claim data.table is faster, which is a shame to hear since I saw it as a good thing to contribute back to R itself.

PR #2420, opened 2017-10-14
2-space indentation

mattdowle:
Long overdue consistent indentation.
Now that all outstanding pull requests are merged, it's a good time.

This merge touches almost every line of all C and all R files. It is strictly indentation only, no logic changes. I've eyeballed every line and am confident no dramatic sweeping mistake has taken place. The diff tools often display as if large blocks of code have changed but in fact they're just not detecting the indentation changes very well. I did lightly touch up formatting in places, in particular inside files which contained both 4-space and 2-space indentation.

I've heard it said that if a person is named by `git blame` on every line, then they own the file and can re-license it!  So just to be clear: that cannot be true, as this merge demonstrates.  I don't own the project, all its contributors do.  The license of data.table can't be changed unless all its contributors agree and this merge doesn't change that.  `git blame` is for quickly establishing the reason for a recent breaking change.

The coverage fail ("88.14% of diff hit (target 91.26%)") seems acceptable since every line is touched including the lines that are yet to be covered.

### DIFF
--- a/CRAN_Release.cmd
+++ b/CRAN_Release.cmd
@@ -21,7 +21,7 @@ grep -RI --exclude-dir=".git" --exclude="*.md" --exclude="*~" --color='auto' -n 
 # Ensure no calls to omp_set_num_threads() [to avoid affecting other packages and base R]
 grep --exclude="data.table/src/openmp-utils.c" omp_set_num_threads data.table/src/*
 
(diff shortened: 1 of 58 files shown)


MichaelChirico:
Not sure if you know about the `?w=1` trick to exclude whitespace differences on GitHub:

https://github.com/Rdatatable/data.table/pull/2420/files?w=1

PR #2424, opened 2017-10-17
Closes #2007 by adding a test to ensure explicit select = NULL works

### LINKED ISSUE #2007: fread error with explicit 'select = NULL'
data.table version: 1.10.0
R version 3.3.2
(full session info at bottom of post)

Default value for `fread` argument `select` is `NULL` according to the function's help file.

I get an error when i explicitly pass `select = NULL` to fread (I use `fread` in a wrapper function and pass `select` to the wrapper).

Example, first two lines work as expected, third line yields an error:
```R
# 1: read some data
fread("a,b\n1,2")
# 2: same but select just the second column
fread("a,b\n1,2", select = 2)
# 3: explicitly make no selection (should yield same result as 1)
fread("a,b\n1,2", select = NULL, verbose = T)
```
> Input contains a \n (or is ""). Taking this to be text input (not a filename)
> Detected eol as \n only (no \r afterwards), the UNIX and Mac standard.
> Positioned on line 1 after skip or autostart
> This line is the autostart and not blank so searching up for the last non-blank ... line 1
> Detecting sep ... ','
> Detected 2 columns. Longest stretch was from line 1 to line 2
> Starting data input on line 1 (either column names or first row of data). First 10 characters: a,b
> All the fields on line 1 are character fields. Treating as the column names.
> Count of eol: 1 (including 0 at the end)
> Count of sep: 1
> nrow = MIN( nsep [1] / (ncol [2] -1), neol [1] - endblanks [0] ) = 1
> Type codes (point  0): 11
> Type codes: 11 (after applying colClasses and integer64)
> Type codes: 11 (after applying drop or select (if supplied)
> Allocating 2 column slots (2 - 0 dropped)
> Read 1 rows. Exactly what was estimated and allocated up front
>    0.000s (  0%) Memory map (rerun may be quicker)
>    0.000s (  0%) sep and header detection
>    0.000s (  0%) Count rows (wc -l)
>    0.000s (  0%) Column type detection (100 rows at 10 points)
>    0.000s (  0%) Allocation of 1x2 result (xMB) in RAM
>    0.000s (  0%) Reading data
>    0.000s (  0%) Allocation for type bumps (if any), including gc time if triggered
>    0.000s (  0%) Coercing data already read in type bumps (if any)
>    0.000s (  0%) Changing na.strings to NA
>    0.000s        Total
> Error in setcolorder(ans, reorder) : 
>   neworder is length 0 but x has 2 columns.

Session info:
```R
sessionInfo()
```

> R version 3.3.2 (2016-10-31)
> Platform: x86_64-redhat-linux-gnu (64-bit)
> Running under: Red Hat Enterprise Linux Client 7.3 (Maipo)
> 
> locale:
>  [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8     LC_MONETARY=en_GB.UTF-8   
>  [6] LC_MESSAGES=en_GB.UTF-8    LC_PAPER=en_GB.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
> [11] LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       
> 
> attached base packages:
> [1] stats     graphics  grDevices utils     datasets  methods   base     
> 
> other attached packages:
> [1] data.table_1.10.0 HYPEtools_0.4-4   hydroGOF_0.3-8    zoo_1.7-14       
> 
> loaded via a namespace (and not attached):
>  [1] Rcpp_0.12.9      lattice_0.20-34  intervals_0.15.1 FNN_1.1          reshape_0.8.6    spacetime_1.2-0  class_7.3-14     gstat_1.1-4     
>  [9] grid_3.3.2       plyr_1.8.4       e1071_1.6-7      pbapply_1.3-1    sp_1.2-4         xts_0.9-7        tools_3.3.2      automap_1.0-14  
> [17] parallel_3.3.2   hydroTSM_0.4-2-1
>

MichaelChirico:
@st-pasha Alternatively, we could just remove the `select = NULL` default and continue with the code as it was before. I think this is preferred as it allows users to have a value to set explicitly in dynamic use cases.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,7 +35,7 @@
 
 5. Improved auto-detection of `character` inputs' formats to `as.ITime` to mirror the logic in `as.POSIXlt.character`, [#1383](https://github.com/Rdatatable/data.table/issues/1383) Thanks @franknarf1 for identifying a discrepancy and @MichaelChirico for investigating.
 
-6. `setcolorder()` now accepts less than `ncol(DT)` columns to be moved to the front, [#592](https://github.com/Rdatatable/data.table/issues/592). Thanks @MichaelChirico for the PR. 
+6. `setcolorder()` now accepts less than `ncol(DT)` columns to be moved to the front, [#592](https://github.com/Rdatatable/data.table/issues/592). Thanks @MichaelChirico for the PR. This also incidentally fixed [#2007](https://github.com/Rdatatable/data.table/issues/2007) whereby explicitly setting `select = NULL` in `fread` errored; thanks to @rcapell for reporting that and @dselivanov and @MichaelChirico for investigating and providing a new test.
 
 7. Three new *Grouping Sets* functions: `rollup`, `cube` and `groupingsets`, [#1377](https://github.com/Rdatatable/data.table/issues/1377). Allows to aggregation on various grouping levels at once producing sub-totals and grand total.
 
--- a/R/fread.R
+++ b/R/fread.R
@@ -107,7 +107,8 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
       cols = which(colClasses=="factor")
   }
   setfactor(ans, cols, verbose)
-  if (!missing(select)) {
+  # 2007: is.missing is not correct since default value of select is NULL
+  if (!is.null(select)) {
     # fix for #1445
     if (is.numeric(select)) {
       reorder = if (length(o <- forderv(select))) o else seq_along(select)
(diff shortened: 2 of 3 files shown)


mattdowle:
@MichaelChirico  The test has a `2` that should be `2L` to pass, iiuc.

mattdowle:
Btw @MichaelChirico, feel free to create branches directly in root now that you're a project member.  That way we can commit to each other's branches if need be.

MichaelChirico:
cool, will do

romanzenka:
@MichaelChirico Please do not switch to more 'missing' tests and arguments without default values.
I am wrapping fread in my own function and it is a pain to pass over "missing" parameters in a function call. Much better provide a default that is NULL and then test for is.null.

PR #2426, opened 2017-10-18
Clean-up of typos/etc in man pages

MichaelChirico:
Used the following code to generate a list of potential typos, then went over the returned list by hand. Can repeat this snippet periodically from now:

```
library(hunspell)
library(magrittr)
library(data.table)

fn = list.files('~/data.table/man', full.names = TRUE)
names(fn) = gsub('.*/', '', fn)
# create a list of words to ignore --
#   all functions in data.table namespace and their arguments
stopwords = ls(envir = getNamespace('data.table'))
stopwords = 
  lapply(stopwords, function(fname) 
    tryCatch(names(formals(fname)), error = function(e) NULL)) %>% 
  unlist %>% unique %>% c(stopwords) %>% unique
# more common words to ignore
stopwords = c(stopwords, 'www', 'http', 'https', 'stackoverflow',
              'com', 'org', 'html', 'htm', 'seealso', 'emph', 
              'href', 'preformatted', 'dontrun', 'datatable')

lapply(fn, function(ff) readLines(ff) %>% paste(collapse = ' ') %>% 
         hunspell_find(dict = 'en_GB') %>% unlist %>% setdiff(y = stopwords))
```

Some extra things besides typos I did:

- Tried to use British spelling consistently
- Tried to eliminate uses of `df` and `dt` in the examples since both of these are functions in `base`
- Added trailing newlines where missing (a side benefit of using `readLines` in the above snippet, this comes out automatically)


MichaelChirico:
Found this package & vignette:

https://ropensci.org/blog/technotes/2017/09/07/spelling-release

Seems this could be integrated with Travis to check every commit for spelling errors. Would have to maintain a dictionary of stopwords. Not sure if worth the setup.

PR #2436, opened 2017-10-23
Better quote rule decision for tied options.

### LINKED ISSUE #2196: fread heuristic predicts wrong column count on a valid file
Minimally reproducible example:
```
fread('1,2,"3,a"\n4,5,"6,b"', verbose=T)
```
produces parsing log
```
Detecting sep ...
  sep==','(ascii 44)  with 2 lines of 3 fields using quote rule 0
  sep==','(ascii 44)  with 2 lines of 4 fields using quote rule 3
Detected 4 columns on line 1. This line is either column names or first data row (first 30 chars): <<1,2,"3,a">>
```
and the output dataset:
```
   V1 V2 V3 V4
1:  1  2 "3 a"
2:  4  5 "6 b"
```

### LINKED ISSUE #2404: fread's jumps causing trouble
This file [copied to a gist](https://gist.github.com/franknarf1/c2ed060c2a8e8a874689751701278d6e) used to read in with fread (a year or so ago when I wrote my script), but now I get...

```
library(data.table)
fread("grr.csv")
```
> Error in fread("grr.csv") : 
>   Line 7 from sampling jump 1 starting <<AAAAAAAAAA>> has more than the expected 17 fields. Separator ',' occurs at position 241 which is character 9 of the last field: <<AAAAAAAA,AAAAAA>>. Consider setting 'comment.char=' if there is a trailing comment to be ignored.

In contrast, `read.csv("grr.csv")` still works, so that's what I will switch to for now. I guess Pasha's solution to my last issue along these lines might have worked for me here as well (#2157 allowing customization of the parameters behind the fread's jumping behavior), though I don't know it well enough to be sure.

mattdowle:
Closes #2404 
Closes #2196
Thanks @franknarf1 for the great file!

It wasn't the jumps per se.  The first 100 lines all had quoted fields which all contained a comma.  The `quoteRule` was then bumped to 3 (meaning ignore quoting) because it found 100 consistent lines of 17 columns rather than 16 lines with that quoteRule (17 beat 16).  I've changed it to only resolve ties (i.e. different ways to parse the same number of consistent lines) with a different `sep` and no longer the same `sep` but different `quoteRule`. The lower `quoteRule` (0=double and 1=escape) are the standard ones so those should (and now do) take preference in these tied cases.

That wrong quote rule was then causing the jump problem later.  I've lessened the impact of that (it now skips sample jumps where there's a format error even after nextGoodLine) but there's more work to do on that.

The test file `grr.csv` is added to the test suite with 4 large columns removed to reduce the size from 860KB down to 283KB.  It generated CRAN's >5MB warning at the original size.  The columns removed are the ones with long strings (>100 chars) without any of the quoting features being tested.

First 16 lines from the test file with quoted fields highlighted : 

image


st-pasha:
Does this also solve https://github.com/Rdatatable/data.table/issues/2196 ?

mattdowle:
@st-pasha  Yes.  I'll add a test to this PR ...

PR #2461, opened 2017-11-03
Non equi issue 2360 fix

arunsrinivasan:


arunsrinivasan:
Not yet complete.

mattdowle:
I always wondered if deleting someone else's comment was possible. It seems it is but GitHub notes the delete it in the log which is good (unlike edits we can make to other people's comments without any apparent log). Arun's comment just said "not yet ready" but I knew it was as we had discussed, in case anyone was wondering what was deleted : - )

PR #2467, opened 2017-11-06
Remove RVersion.h as a dependency

st-pasha:
Due to unclear license status of "Rversion.h", I think we'd be better-off not including it (especially since we barely use it at all).


mattdowle:
Change applied before license change PR #2456

Issue #2275, opened 2017-07-20
Strange results from a non-equi join with multiple conditions


franknarf1:
This was brought up [on SO](https://stackoverflow.com/q/45212781/). 

The goal is to find out if each row in DT1 has a match in DT2 in the sense of `on=.(RANDOM_STRING, DATE >= START_DATE, DATE <= EXPIRY_DATE)`:

```
set.seed(123)
library(data.table)
library(stringi)
# Sorry that it requires stringi; I couldn't find another way forward besides the OP's verbatim example.

n <- 100000

DT1 <- data.table(RANDOM_STRING = stri_rand_strings(n, 5, pattern = "[a-k]"),
                  DATE = sample(seq(as.Date('2016-01-01'), as.Date('2016-12-31'), by="day"), n, replace=T))

DT2 <- data.table(RANDOM_STRING = stri_rand_strings(n, 5, pattern = "[a-k]"),
                  START_DATE = sample(seq(as.Date('2015-01-01'), as.Date('2017-12-31'), by="day"), n, replace=T))

DT2[, EXPIRY_DATE := START_DATE + floor(runif(1000, 200,300))]
```
My usual approach is to do a join, counting matches with `.N` and `by=.EACHI`. However, the OP found that this fails here:

```
# correct result (takes a long time)

    DT1[, m_ok := DT2[.BY, on=.(RANDOM_STRING), inrange(DATE, START_DATE, EXPIRY_DATE)], by=RANDOM_STRING]

# my usual approach

    DT1[, MATCHED := FALSE]

    DT1[!(MATCHED), MATCHED := DT2[.SD, on=.(RANDOM_STRING, START_DATE <= DATE, EXPIRY_DATE >= DATE), .N, by=.EACHI ]$N > 0L ]

# comparison

    DT1[MATCHED != m_ok | (MATCHED & is.na(m_ok))] 

    # shows many cases of failure from the attempt
    # for example...

    DT1[RANDOM_STRING == "egkja"]; DT2[RANDOM_STRING == "egkja"]

    #    RANDOM_STRING       DATE MATCHED  m_ok
    # 1:         egkja 2016-05-19    TRUE FALSE
    # 2:         egkja 2016-06-02   FALSE FALSE
    # 3:         egkja 2016-05-20    TRUE FALSE
    # 4:         egkja 2016-03-16   FALSE  TRUE

    #    RANDOM_STRING START_DATE EXPIRY_DATE
[...]

PR #2471, opened 2017-11-08
Generated two large test files and removed

mattdowle:
Placed R code from issue #2275 into tests.Rraw to auto generate `DT1` and `DT2` rather than `fread` them in.  Removed the need for `stringi`.  The random data isn't exactly the same since `stringi` uses a different RNG I assume, but I don't think the particular seed matters in this case. Please check though.
To pass CRAN check note : 
```
* checking installed package size ... NOTE
  installed size is  9.3Mb
    libs    1.1Mb
    tests   6.6Mb
```
due to these files from PR #2461 : 
```
$ ls -lh inst/tests/issue_2275_dt*.csv
-rw-r--r-- 1 mdowle mdowle 1.7M Nov  8 12:31 inst/tests/issue_2275_dt1.csv
-rw-r--r-- 1 mdowle mdowle 2.7M Nov  8 12:31 inst/tests/issue_2275_dt2.csv
```


arunsrinivasan:
Thanks, Matt. I tried to reduce the size, but couldn't replicate the issue. I didn't realise the file was too big. I tested this data with the commit prior to the commit that fixed it and am able to reproduce the issue. So all good.

PR #2477, opened 2017-11-13
Closes #2475 -- setnames on a table with NA names works as expected

### LINKED ISSUE #2475: setnames(x, new) fails if somehow anyNA(names(x))
The data.table interface to names don't seem to support making one NA, but if it does happen, it cannot be fixed with `setnames(x, new)`:

```
x = setNames(data.frame(a = 1, b = 2), c(NA, "b"))

library(data.table)
setDT(x)
setnames(x, c("a", "b"))
names(x)
# [1] NA  "b"
```

From user3496254 SO: https://stackoverflow.com/questions/47228836/r-data-table-bug-cannot-rename-column-names-which-are-na

MichaelChirico:
Other option would be something like `which(!vapply(seq_along(old), function(ii) nx[ii] == old[ii], logical(1L)))`. Looks ugly and efficiency gain (if any) probably trivial

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -110,6 +110,8 @@
 
 25. `by=.EACHI` works now when `list` columns are being returned and some join values are missing, [#2300](https://github.com/Rdatatable/data.table/issues/2300). Thanks to @jangorecki and @franknarf1 for the reproducible examples which have been added to the test suite.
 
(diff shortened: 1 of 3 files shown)


jangorecki:
This PR broke R 3.0.0 compatibility (which is currently broken by another commit but it is not relevant here) because there is no `anyNA` function in R 3.0.0. @MichaelChirico

MichaelChirico:
oh darn. should we test R version, or just use any(is.na())?

jangorecki:
Testing R version... makes sense for very wide tables, million of columns, which is not very common scenario but it is probably worth extra `if` in the code. There might be a C function in data.table code for anyNA, not sure, this would be the best to use.

PR #2480, opened 2017-11-13
Closes #2046 and closes #2111. Fixes -ve length vectors issue with GForce

### LINKED ISSUE #2046: gmedian failing on simple case
On 1.10.4:

```
dt <- data.table(x=c(1,1,1),y=c(3,0,0))
setkey(dt, x)
dt[, median(y), by=x]
```
Yields the following error:
Error in gmedian(y) : negative length vectors are not allowed

If dt doesn't have a key set, it doesn't throw the error (but I can't tell if gmedian is even being called in that case).  Also calling stats::median instead of median above works because it bypasses gmedian directly.

### LINKED ISSUE #2111: optimized by group gvar() returns error: "negative length vectors are not allowed" when within-group variance is 0
Using latest development version of data.table (1.10.5):

This returns an error:
```
library(data.table)
testDT <- data.table(col1 = c(1,1,1, 2,2,2), col2 = c(2,2,2,1,1,1), ID = c(rep(1,3), rep(2,3)))
setkeyv(testDT, "ID")
testDT[, lapply(.SD, var), by = ID]
```
`Error in gvar(col1) : negative length vectors are not allowed`

This however works fine:
```
testDT[, lapply(.SD, stats::var), by = ID]
   ID col1 col2
1:  1    0    0
2:  2    0    0
```

arunsrinivasan:
Closes #2046 
Closes #2111 
Issues related to GForce as explained by @kmillar under https://github.com/Rdatatable/data.table/issues/2046#issuecomment-283478516. 

Fixed by checking if `maxgrpn` attribute is NULL and if so, generating it from `max(len__)`. Have written in some detail under the fix too.


mattdowle:
Delay merging this one is because it's overlapping with #2474 (attributes of `o`) and then concentrated on #2488. Getting there ...

arunsrinivasan:
Thanks!

Issue #2407, opened 2017-10-10
Wrong calculation of Isoweek for year 2015 (in tz != "UTC")


MoebiusAV:
As posted in [https://stackoverflow.com/questions/43944430/r-why-does-the-default-package-for-isoweek-change-depending-how-i-use-it/46655291#46655291](url) data.table isoweek function sometimes gives a different result of ISO standart.

 I found an explination, running data.table function step by stepThe first step of function is converting Date as.POSIXlt. 
The result will be different if you input a character or a Date 
`as.POSIXlt('2015-01-02')`
`[1] "2015-01-02 COT"`

`as.POSIXlt(as.Date('2015-01-02'))`
`[1] "2015-01-02 UTC"`

As you can see, what change is the time Zone. 
In the third step of function, in both cases, it returns a Date in Local time
`(year_start <- as.POSIXct(paste0(as.POSIXlt(nearest_thurs)$year +1900L, "-01-01"))) `
`[1] "2015-01-01 COT"`
That way, in the forth and last step, the date difference between UTC and Local time gives some decimals 'aditional' to time difference between Local Time dates (in my case 5 hours, or 0.208). That is why the final result of weeks changes.

In conclusion, if you force date as character, the function will calculate everything in Local time and the result will be right
-- | --

MichaelChirico:
Following up on my SO comment -- without a reproducible example, I can't confirm this works, but it seems the answer is to change:

https://github.com/Rdatatable/data.table/blob/master/R/IDateTime.R#L275

    xlt <- as.POSIXlt(x)

To 

    xlt <- as.POSIXlt(x, tz = 'UTC')

(that's what is seems is done by `lubridate::isoweek` -> `lubridate::make_datetime`):

```
# lubridate::isoweek
#  c/f https://github.com/tidyverse/lubridate/blob/e4ad2c3024c4e114c5f046c7fa0d73e2168f0d2b/R/accessors-week.r#L36
#  ** on CRAN, lubridate::isoweek is what now corresponds to
#      lubridate:::.other_week(x, 1)
x <- as.POSIXlt(x)
```

```
# lubridate::make_datetime
#   c/f https://github.com/tidyverse/lubridate/blob/c10c8cfff6a8d86e58c5865c1fb1c93463be647e/R/instants.r#L109
.POSIXct(numeric(), tz = tz) # default tz = 'UTC'
```

(actually, on further look, these look the same. leaving this for my own notes)

[...]

PR #2500, opened 2017-11-23
Simplified isoweek, #2407 #2421

mattdowle:
Follow up on #2407 and #2421 
Tests 1702.1 and 1702.3 failed on win-builder, because of win-builder running in Germany in a different timezone.  They will have failed depending on time of day running the test.
Perplexed as to why `isoweek` used `POSIXlt` in the first place. Further, why it had any time zone usage at all. Probably a hangover from years back. Replaced with simpler integer days from the 1 Jan 1970 epoch (a Thursday, conveniently).

Will create another issue to purge all usage of POSIXlt (still seems much usage in IDateTime.R) and add a grep to CRAN_Release.cmd to catch any usage in future.


MichaelChirico:
Nice. IIRC I was just mirroring `lubridate::isoweek` and used `POSIXlt` as they did. Not sure why I didn't try harder to skip it. Lazy coding 😱  -- we have our own `wday` function.

Timezone consideration was only added recently due to rounding issues, and those are only introduced (IIUC) because of the `POSIXlt` call.

MichaelChirico:
I'm also noticing a new formula on the Wiki page that looks even easier:

https://en.wikipedia.org/wiki/ISO_week_date#Calculating_the_week_number_of_a_given_date

mattdowle:
Yeah that POSIXlt has been there a long time I think; all usage needs to be purged. That last formula from wikipedia looks about the same complexity to me.  More beneficial would be to do it in C as all those vector ops are allocating each time.  At least it'll now be better than (truly awful) 40 byte POSIXlt.

MichaelChirico:
Agreed. Looks more concise, but actually implementing it is a bear & probably much slower.

I ran the following to be more confident we're not missing anything:

```
date_range = seq.Date(as.Date('1850-01-01'), as.Date('2250-01-01'), by = 'day')

sum(lubridate::isoweek(date_range) != data.table::isoweek(date_range))
```

Included a 1% sample of this since I guess adding 140000-row test is overkill.

mattdowle:
NB: was confused as to coverage fail but it's just that test() has been strengthened and its branches that happen on error aren't being tested, because all tests are passing.

mattdowle:
Great test @MichaelChirico!

MichaelChirico:
re:codecov -- is it overkill to include a meta-test testing `test` 🤔

MichaelChirico:
Also ran this to check there's nothing wrong with `POSIXct` either (there shouldn't be, so not adding another test -- `POSIXct` is covered in 1702.3):

```
#12622694400 is the length in seconds of a 400-year cycle
posix_range = as.POSIXct(sample(12622694400, 1e6), origin = '1970-01-01 00:00:00 UTC')

sum(lubridate::isoweek(posix_range) != data.table::isoweek(posix_range))
# [1] 0
```

mattdowle:
A meta-test testing `test()` is a good idea.  Trouble is how to test it, as inside `test()` it logs the test number for the summary inside the final `stop()` at the end.  If there's a way, it would be good, though.

PR #2501, opened 2017-11-23
importing data.table: doc for rev deps developers

jangorecki:
no executable R code chunks, only markdown

### DIFF
--- a/vignettes/datatable-importing.Rmd
+++ b/vignettes/datatable-importing.Rmd
@@ -1,0 +1,152 @@
+---
+title: "Importing data.table"
+date: "`r Sys.Date()`"
+output:
+  rmarkdown::html_vignette:
+    toc: true
+    number_sections: true
+vignette: >
+  %\VignetteIndexEntry{Importing data.table"}
+  %\VignetteEngine{knitr::rmarkdown}
+  \usepackage[utf8]{inputenc}
+---
+
+<style>
+h2 {
+    font-size: 20px;
+}
+</style>
+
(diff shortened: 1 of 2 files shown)


vignettes/datatable-importing.Rmd
```diff
@@ -0,0 +1,152 @@
+---
+title: "Importing data.table"
+date: "`r Sys.Date()`"
+output:
+  rmarkdown::html_vignette:
+    toc: true
+    number_sections: true
```
jangorecki: looks like I could avoid below ` {.unnumbered}` styles by setting this to false...

mattdowle:
I never knew before that using `Depends` was the same as the user typing `library(data.table)`. Discovered that talking to Jan on email.  So, yes, definitely agree `Depends:` is not recommended; packages should use `Imports:` only.  So much so, maybe `data.table:::cedta()` should warn when it detects the reverse dependency has depended on rather than imported data.table?

Are there any plans for `Depends:` to deprecate accepting R packages in that field? We use `Depends:` to state the version of R we depend on and perhaps other packages state system libraries there.  But is there any advantage in an R package being in `Depends:` rather than `Imports:`?

jangorecki:
@mattdowle I don't think it will be deprecated as there are still valid use cases for Depends.

> Field ‘Depends’ should nowadays be used rarely, only for packages which are intended to be put on the search path to make their facilities available to the end user (and not to the package itself): for example it makes sense that a user of package latticeExtra would want the functions of package lattice made available.

For the same reason my [dtq](https://github.com/jangorecki/dtq) package should Depends on data.table.

jangorecki:
Vignette can be viewed on https://rdatatable.gitlab.io/data.table/library/data.table/doc/datatable-importing.html

PR #2517, opened 2017-12-13
Fix reading of a simple 1x1 dt with NA value

### LINKED ISSUE #2516: File containing a single NA value is read as chr instead of logi
```
> str(fread("A\n\n"))
Classes ‘data.table’ and 'data.frame':	0 obs. of  1 variable:
 $ A: chr 
```
Expected: a 1x1 data.table of logical type, instead getting a 0x1 data.table of character type.

st-pasha:
Closes #2516


mattdowle:
Not sure about this.
This PR has this behaviour : 
```
> fread("A\n")
Empty data.table (0 rows) of 1 col: A
> fread("A\n\n")
    A
1: NA
> fread("A\n\n\n")
    A
1: NA
2: NA
> fread("A,B\n")
Empty data.table (0 rows) of 2 cols: A,B
> fread("A,B\n\n")
Empty data.table (0 rows) of 2 cols: A,B
> fread("A,B\n\n\n")
Empty data.table (0 rows) of 2 cols: A,B
>
```

which would be a change from CRAN version which does this : 
```
> fread("A\n")
Empty data.table (0 rows) of 1 col: A
> fread("A\n\n")
Empty data.table (0 rows) of 1 col: A
> fread("A\n\n\n")
Empty data.table (0 rows) of 1 col: A
> fread("A,B\n")
Empty data.table (0 rows) of 2 cols: A,B
> fread("A,B\n\n")
Empty data.table (0 rows) of 2 cols: A,B
> fread("A,B\n\n\n")
Empty data.table (0 rows) of 2 cols: A,B
```

Thoughts?

st-pasha:
Consider this file:
```
Aa,Bb
10,20

30,40
```
Technically, this is not a valid CSV -- there shouldn't be a blank line there, but still, if this is the data, what do you do? You could choose a different coping strategy: (1) set `fill=TRUE` and fill the second line with NAs; (2) set `blank.lines.skip=TRUE` and ignore the second line; (3) the default action is to stop reading at the first blank line:
```
> fread("Aa,Bb\n10,20\n\n30,40\n")
    Aa Bb
1: 10 20
Warning message:
In fread("Aa,Bb\n10,20\n\n30,40\n") :
  Found the last consistent line but text exists afterwards. Consider fill=TRUE and/or blank.lines.skip=TRUE. First 200 characters of discarded line: <<30,40>>
```
Similarly, if the blank line occurs at the end, only this time fread doesn't display any warning:
```
> fread("Aa,Bb\n10,20\n30,40\n\n")
    Aa Bb
1: 10 20
2: 30 40
```
In both cases the CSV is still invalid, and fread deals with it in a way that it thinks is "most reasonable".

----
Now let's go back to the case in question. Consider the following file:
```
Aa
10

30
```
Unlike the first case, this is already a **valid CSV**: there is only one way to read it:
```
> fread("Aa\n10\n\n20\n")
   Aa
1: 10
2: NA
3: 20
```
Similarly, if the empty line occurs at the end of the file, it still remains a valid CSV file, and it _should_ be read as such:
```
> fread("Aa\n10\n20\n\n")  # expected output:
   Aa
1: 10
2: 20
3: NA
```
The fact that it doesn't I view as a bug:
```
> fread("Aa\n10\n20\n\n")  # actual output
   Aa
1: 10
2: 20
```
To recap, in the first case (2+ -column file) blank lines make it an invalid CSV file, and there could be difference in opinion as to what coping strategy is best. At the same time, in the second case (1-column file) blank lines are allowed in a valid CSV file, and therefore they should be treated accordingly.

Issue #1656, opened 2016-04-19
fread could read *base data types* directly for efficiency


jangorecki:
When dealing with data that uses standard POSIXct format - an UNIX Epoch time - counting seconds from 1970-01-01. We could read that data directly POSIXct just by specifying `colClasses` to force coercion column on default arguments. It can also speed up general IO by reducing the need for date time formatting when using `fwrite` and later parsing it again to numeric with `fread`.
The same applies to Date.
```r
library(data.table)
dt = data.table(a=as.Date("2015-01-01"), b=as.POSIXct("2015-01-01 15:30:20"))
fwrite(dt, "fwrite-tests.csv", verbose = TRUE)
fread("fwrite-tests.csv", colClasses = c(a="Date", b="POSIXct"))
# so we could have Date and POSIXct back
```

clarkdk:
It would be awesome if `fwrite` by default wrote `Date` and `POSIXct` objects in standard formats (ie ISO 8601, or `Lubridate`) that `fread` would automatically recognize and read in as `Date` or `POSIXct`, without requiring `colClasses` to be specified for each datetime column. A default automatic approach would be really welcome since I spend a lot of programming time converting datetime character strings back to POSIXct when importing time series data, even when I have written the file myself with `write.csv` or `fread`. It would presumably also require an assumption (default value) for `tz` on import, unless the default format included the time zone offset on every value. Perhaps a `POSIXct.format="format-string"` parameter could be included to specify the format string to use when writing and reading `POSIXct` values, since `colClasses = c(a="Date", b="POSIXct")` doesn't allow a specific format and/or timezone to be specified.

jangorecki:
@clarkdk You discuss quite a different issue, as it requires parsing strings that stores date/datetime. This issue is about date/datetime stored as numerics, and those can be effectively optimized for IO speed if fwrite and fread would be able to write and read those types as their numerics transparently (or using just `colClasses`). So no parsing is required. If you would like to speedup parsing strings then [fasttime](https://cran.r-project.org/web/packages/fasttime) can help.

clarkdk:
@jangorecki OK, now I understand. I will make a separate FR. In implementing datetimes as numerics in `fread`/`fwrite`, please don't rule out a possible future feature involving writing and parsing of datetimes as formatted strings.

jangorecki:
now when have yaml support this feature is even more useful as we can maximise IO without trade-offs

PavoDive:
[...]

Issue #1445, opened 2015-11-26
when 'select' argument is provided, columns should be returned in the order provided to that argument 


arunsrinivasan:
```r
require(data.table) # devel
fread("a,c,b\n1,2,3", select=c("b", "c"))
#    c b
# 1: 2 3
```

I'd have expected the columns to be in the order provided to `select`. Just ran across this bug today.

MichaelChirico:
Any reason to care about the column order? I've noticed this but never particularly minded. Usually the order I specify them in is meaningless.

arunsrinivasan:
@MichaelChirico perhaps because the reason I ran into this issue is due to recently implemented `col.names` argument. You can't use `col.names` argument together with `select` if you can't be sure of the order of columns in the result.

```r
fread(my_file, select=c("b", "c"), col.names = c("B", "C"))
```
names the columns wrongly in this case.. and so one has to do:

```r
cols = c("b", "c")
dt = fread(my_file, select=cols)
setcolorder(dt, cols)
setnames(dt, cols, c("B", "C"))
```

MichaelChirico:
@arunsrinivasan duly noted, I had missed the `col.names` update. Very useful.

mattdowle:
Agreed.  And fread should error if the column requested isn't present.

```R
> fread("a,c,b\n1,2,3", select=c("b", "foo"))
   b
1: 3 
# TODO:  Error: column name "foo" passed to 'select=' not found in data
```

Issue #1426, opened 2015-11-10
FR for `fread`: if `select` is used, `colClasses` need only correspond to the columns in `select`


MichaelChirico:
I never filed a FR for a [question I raised a year ago on SO](http://stackoverflow.com/questions/25691637/using-colclasses-and-select-arguments-of-fread-simultaneously).

The current canon for using `select` and `colClasses` simultaneously is (IMO) unwieldy.

Consider:

```
#file to read
ffile <- paste0(paste(paste0("V", 1:20), collapse = ","),
              "\na,b,c,d,e,1,2,3,4,5,1.1,1.2,1.3,1.4,1.5,",
              "TRUE,FALSE,TRUE,FALSE,TRUE")

#columns to take
sel <- c(2, 10, 13:15, 20)
#types of all columns
tps <- rep(c("character", "integer",
           "numeric", "logical"),
         rep(5, 4))
```

Here's the best I could come up with as a programmatic way to use `fread`:

```
DT <- fread(ffile, select = paste0("V", sel),
            colClasses =
              sapply(unique(tps[sel]),
                     function(x) paste0("V", sel[which(tps[sel] == x)])))
```

(gross; could be spelled out explicitly as the following, but this is generally unsatisfying:)

```
DT <- fread(ffile, select = paste0("V", sel),
            colClasses =
              list(character = "V1", integer = "V10",
                   numeric = c("V13","V14","V15"),
                   logical = c("V20")))
```

To me it would make much more sense to be able to simply write:

```
DT <- fread(ffile, select = paste0("V", sel), colClasses = tps[sel])
```

But this currently produces the error:

> Error in `fread(ffile, select = paste0("V", sel), colClasses = tps[sel])` : 
>   `colClasses` is unnamed and length 6 but there are 20 columns. See `?data.table` for `colClasses` usage.

Parsimonious, and as far as I can tell unambiguous. Any reason why this wouldn't work?

MichaelChirico:
What about just reversing the order of dealing with `select` vs. `colClasses` -- switch [Lines #1141 - 1157](https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L1141-L1157) with [Lines #1059 - 1112](https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L1059-L1112), and reset `ncol` to `length(select)` within the `select` condition?

dselivanov:
+1 for this, also annoying me. Will try to check.

[...]

PR #2545, opened 2018-01-03
Various colClasses enhancements

### LINKED ISSUE #491: [R-Forge #5323] fread: colClasses does not convert to non-builtin types
Submitted by: James Sams; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5323&group_id=240&atid=975)

If I set a column to `numeric`, `character`, etc., the conversion works as expected. However, if I set it to a user created class, it seems like `fread` gets confused. It recognises that the column is supposed to be converted, but doesn't do the conversion.

The message generated by `fread` in this situation (sample code attached) is:

``` S
# Detected eol as \n only (no \r afterwards), the UNIX and Mac standard.
# Using line 2 to detect sep (the last non blank line in the first 'autostart') ... none found (see ?fread). 
# Deducing this is a single column input. Otherwise, please specify 'sep' manually.
# First row with 1 fields occurs on line 1 (either column names or first row of data)
# All the fields on line 1 are character fields. Treating as the column names.
# Count of eol after first data row: 2
# Subtracted 1 for last eol and any trailing empty lines, leaving 1 data rows
# Type codes: 0 (first 5 rows)
# Column 1 ('date') was detected as type 'integer' but bumped to 'numeric' as requested by colClasses
#    0.000s (  0%) Memory map (rerun may be quicker)
#    0.000s (  0%) sep and header detection
#    0.000s (  0%) Count rows (wc -l)
#    0.000s (  0%) Column type detection (first, middle and last 5 rows)
#    0.000s (  0%) Allocation of 1x1 result (xMB) in RAM
#    0.000s (  0%) Reading data
#    0.000s (  0%) Allocation for type bumps (if any), including gc time if triggered
#    0.000s (  0%) Coercing data already read in type bumps (if any)
#    0.000s (  0%) Changing na.strings to NA
#    0.000s        Total
```

### LINKED ISSUE #1634: fread doesn't check colClasses to be valid type
```
> sapply(fread("A,B\n1,2",colClasses="numeric"),class)   # ok
        A         B 
"numeric" "numeric" 
> sapply(fread("A,B\n1,2",colClasses="integer"),class)    # ok
        A         B 
"integer" "integer" 
> sapply(fread("A,B\n1,2",colClasses="string"),class)      # type "string" works?
          A           B 
"character" "character" 
> sapply(fread("A,B\n1,2",colClasses="c"),class)             
          A           B 
"character" "character" 
> sapply(fread("A,B\n1,2",colClasses="foo"),class)      # should be error that foo isn't valid type?
          A           B 
"character" "character" 
> 

```

### LINKED ISSUE #2025: fread stringsAsFactors=0.2
For character columns, if  `uniqueN(column)/nrow < 0.2` then make the column a factor.  Generalizing current binary argument to a scale between FALSE (0.0) and TRUE (1.0).
Request came up for onwards model build which switches based on character / factor type.
No change to default FALSE.

### LINKED ISSUE #2610: fread colClasses support for POSIXct
Currently fread colClasses does not support conversion to POSIXct:
```r
str(data.table::fread("c1,c2\n2018-01-31 03:16:57,b", colClasses=c(c1="POSIXct")))
#Classes 'data.table' and 'data.frame':	1 obs. of  2 variables:
# $ c1: chr "2018-01-31 03:16:57"
# $ c2: chr "b"
# - attr(*, ".internal.selfref")=<externalptr> 
```
related to #1450
once solved update: https://stackoverflow.com/questions/41903259/data-table-colclasses-conversion-to-posixct

HughParsonage:
Closes #491  fread: colClasses does not covert to non-builtin types
Closes #2610  colClasses = POSIXct
Closes #1634  fread doesn't check colClasses to be valid type
Closes #2025 allow `stringsAsFactors` parameter to be a fraction between 0 and 1 
(no issue number) When `colClasses` includes multiple list elements named `factor` (*e.g.* `list(factor = 1, factor = 2)`), all elements are respected, not just the first.
#1445 has been fixed when `select` is non-monotonic.

Not tackled yet:
#1656 fread/fwrite \*base data types\* directly for efficiency
#1426 if `select` is used, `colClasses` need only correspond to the columns in `select`


mattdowle:
Hi @HughParsonage.  I've invited you to be project member; you'll need to accept the invite before the change will apply.  This enables you to create branches in the main repository directly so we can all push to each other's branches.  Welcome!

inst/tests/tests.Rraw
```diff
@@ -481,12 +481,14 @@ test(165, subset(DT,a>2), DT[a>2])
 test(166, suppressWarnings(split(DT,DT$grp)[[2]]), DT[grp==2])
 
 if ("package:ggplot2" %in% search()) {
-    test(167, names(print(ggplot(DT,aes(b,f))+geom_point()))[c(1,3)], c("data","plot"))
+    DF <- data.frame( a=1:5, b=11:50, d=c("A","B","C","D"), f=1:5, grp=1:5 )
+    res167 <- names(print(ggplot(DF,aes(b,f))+geom_point()))[c(1,3)]
+    test(167, names(print(ggplot(DT,aes(b,f))+geom_point()))[c(1,3)], res167)
     # The names() is a stronger test that it has actually plotted, but also because test() sees the invisible result
```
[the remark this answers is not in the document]
mattdowle: Is it possible you hit this unrelated issue:  https://github.com/Rdatatable/data.table/issues/2546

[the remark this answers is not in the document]
HughParsonage: I understood the test to mean that using `data.table` as opposed to `data.frame` does not affect `ggplot`s. Note `res167` uses `data.frame`.

On my computer: with ggplot2 v 2.2.1.9000, 

```
names(print(ggplot(DT,aes(b,f))+geom_point()))[c(1,3)]
# [1] "data" "scales"
```

not `c("data", "plot")`. Indeed, none of the names are `"plot"`. Would it be better to use all the names?

inst/tests/tests.Rraw
```diff
@@ -5994,7 +5996,9 @@ test(1444.4, fread(paste(paste("Hz.BB.GHG.", 1:10, sep = ""), collapse="\n"), se
              data.table(V1="Hz",V2="BB",V3="GHG",V4=1:10))
 
 # doubled quote inside a quoted field followed by an embedded newline
+if (.Platform$OS.type == "unix") {
 test(1445, fread("doublequote_newline.csv")[7:10], data.table(A=c(1L,1L,2L,1L), B=c("a","embedded \"\"field\"\"\nwith some embedded new\nlines as well","not this one","a")))
+}
```
mattdowle: This is passing successfully on AppVeyor Windows tests and on CRAN Windows tests.  Something else must be going wrong on your Windows machine, to fix separately.  Important to keep the test running on Windows.

inst/tests/tests.Rraw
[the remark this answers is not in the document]
mattdowle: Oh I see now.  I didn't spot the difference `DF` vs `DT` in those two lines.   `ggplot2` changed its names then and if you upgrade to CRAN version ggplot2 v2.2.1 (released a year ago) it should work.  Or, is 2.2.1.9000 a more recent development version of ggplot2.  If so, has it changed again?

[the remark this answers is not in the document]
mattdowle: Yep, ggplot2 2.2.1.9000 is latest dev version you have.  Good test.  Ok they changed again and your solution looks good to me.  Maybe just a comment alongside pointing out that `DF` is being used to find out what the names are.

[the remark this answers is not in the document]
HughParsonage: Yes, `ggplot2` is the current dev version. Looking at the code for `ggplot2`, I can't tell what the expected behaviour ought to be:

`ggplot.data.frame` returns the list that I see:   https://github.com/tidyverse/ggplot2/blob/7b5c185eff0dabef50e838701ab805ae09a1170f/R/plot.r#L92

`ggplot_build.ggplot` returns the list that the test originally expected. https://github.com/tidyverse/ggplot2/blob/152910a698121ac4fa771d110507e607d45cfc8f/R/plot-build.r#L99

Note that the test appears to proxy the actual appearance of the plot, and the plot does indeed occur on my machine.

inst/tests/tests.Rraw
HughParsonage: Would you prefer I raise this an issue and revert this change, or relax the test to permit e.g. `embedded ""field""\r\nwith some embedded new\r\nlines as well`  as well as `embedded ""field""\nwith some embedded new\nlines as well`?

mattdowle: Yes : new issue please and revert the change.  Why does it pass on AppVeyor and CRAN WIndows but it doesn't pass on your Windows?

HughParsonage: Ahh, I think it's because when I clone the repo, `\n` is replaced with `\r\n` by **git**. My bad. 

```r
fread("https://raw.githubusercontent.com/Rdatatable/data.table/master/inst/tests/doublequote_newline.csv")[["B"]][8]
#> [1] "embedded \"\"field\"\"\nwith some embedded new\nlines as well"

fread("inst/tests/doublequote_newline.csv")[["B"]][8]
#> [1] "embedded \"\"field\"\"\r\nwith some embedded new\r\nlines as well"
```

HughParsonage:
Do you want me to attempt a simpler version using the same approach (most of the complexity is just to construct the phrasing of the warning message),  like

```r
    if (!all(colClasses %chin% c(NA, "logical", "integer", "numeric", "double", "character", "factor"))) {
      warning("colClasses contains unsupported values.")
    }
```
(with corresponding tests modified).

or do you think listing all the valid classes in `fread.R` is simply the wrong approach at this time? Obviously such an approach would still not be in the spirit of S3 dispatch, but this PR is merely intended as a stop-gap for the issue while `colClasses = "Date"` etc are not yet implemented.

mattdowle:
My current thought is that the dispatch to the class method should be implemented rather than this stop gap.  I'm thinking that would be quicker.
To confirm the thought: `integer64` is a valid type for `colClasses=` currently and should work.  But we must be missing a test for passing `integer64` to `colClasses=` because `integer64` isn't defined in `check_colClasses_validity()`.

mattdowle:
I thought it would be a case of just calling something like `DT[, (col):=as(col, colClasses[i])]` for each column near the end of `fread()` at R level.
This was the R message I had in mind :
```
> as("2018-01-03", "foo")
Error in as("2018-01-03", "foo") : 
  no method or default for coercing “character” to “foo”
```
So that's what `fread` user would see if they used `"foo"` in `colClasses=`.
But then `Date` fails too which I didn't expect : 
```
> as("2018-01-03", "Date")
Error in as("2018-01-03", "Date") : 
  no method or default for coercing “character” to “Date”
It seems you have to use `as.Date` like this : 
> as.Date("2018-01-03")
[1] "2018-01-03"
```
Maybe `as()` is S4 not S3?  I looked at `?as` which didn't help on first glance.

HughParsonage:
I was thinking something like `existsMethod` or `hasMethod`, but I can't seem to work them out:

```r
existsMethod("as.numeric")
#> [1] TRUE
existsMethod("as.Date")
#> [1] FALSE
```

HughParsonage:
Apologies for the enormous diff, but there were a few tricky paths, as evinced by the number of additional tests

  * the `as.complex`, `as.raw`, `as.POSIXct`, `as.Date` have to be implemented slightly differently (*i.e.* not simply through `as()`.
  * `factor` required a slight bugfixes for some corner cases
  * the interaction with `select` and `drop` was complicated
  * `"NULL"` elements weren't being implemented correctly.

HughParsonage:
Thanks for the feedback. And you're right about `as_factor`: it's an excellent function ... but I believe you wrote it! I took it out of `setfactor` so it could be made compatible with the other `colClasses` functions.

I used that part of `utils::read.table` as my starting point too and honestly did set out to make it longer. Let me try to explain why I failed to keep it so compact. I appreciate that it would have been better to have written clearer code so it could speak for itself, but here we go.

### Why are there 3 `switch`s?
Once `set_colClasses_ante` is called and we've established that there are new classes to apply, there are two possible forms of `colClasses`:

  1. A list
  2. An unnamed character vector

(Note that in the case of a named character vector, it is converted to a list by `colClasses = tapply(names(colClasses), colClasses, c, simplify=FALSE)`.)

The outermost `switch` separates the code for these two forms. The other two instances of `switch` do essentially the same thing, the first in the `"list"` case and the second in the `"character"` case. Note that these inner `switch`s are very similar to the implementation of `utils::read.table`: a loop over the columns and applying a different coercion based on `colClasses`. One could write the excerpt from `utils::read.table` as 

```r
for (i in (1L:cols)[do]) {
  data[[i]] <- 
    switch(colClasses[i], 
           "NA" =  type.convert(data[[i]], as.is = as.is[i], dec = dec, 
                                numerals = numerals, na.strings = character(0L)),
           "factor" = as.factor(data[[i]]),
           "Date" = as.Date(data[[i]]),
           "POSIXct" = as.POSIXct(data[[i]]),
           methods::as(data[[i]], colClasses[i]))
}
```

and indeed in the `"character"` branch, this is basically what the code looks like. 

### Why do all the classes have a new function defined with a `try_` prefix?

This was a design decision which you may not agree with but seemed more in the spirit of how `colClasses` is currently implemented. When the application of the user-overridden `colClasses` is not safe, my implementation emits a warning but then always reverts back to `"character"`.

```r
fread(input = "A,B\nMarch 18 2005,2015-01-01", colClasses = "Date")
#> Warning: Column 1 was set by colClasses to be 'Date', but fread encountered the following error:
#>  character string is not in a standard unambiguous format
#> so the column will be left as type character.
#>                A          B
#> 1: March 18 2005 2015-01-01
```

 By contrast, `read.table` might force the coercion, or might error with a somewhat arcane error message. 

```r
read.csv(text = "A,B\nMarch 18 2005,2015-01-01", colClasses = "Date")
#> Error in charToDate(x) : 
#>  character string is not in a standard unambiguous format
```

It may be more appropriate to error in the case of an error, but my view is that it is better to warn and revert as I/O might be time-consuming. In addition, returning an error could break existing code if people had been unknowingly using an unsupported class. 

### What's the `preamble` for?

Since I made the decision to revert failed coercion attempts, I thought it best to provide an informative warning message. The preamble simple states `"Column %s was set by colClasses to be '%s', but fread encountered the following", j, cls)`. This is followed by the error/warning message that was encountered and finally a line mentioning that the column will be kept as type character. This may be taking the idea of functional programming too far, (one could just write out the warning message for each of the five or so classes) however I did actually write inconsistent warning messages on my first attempt, so I wrote this to avoid that.

### Why is there an `is.sorted(select)`? 

At the point the `colClasses` are set, it is not clear whether the order of `select` has been respected (See issue #1445.) While it may be possible to deduce the answer by static code analysis now, you may decided to implement the ordering inside fread.c later. The point of `is.sorted` is to ensure that `colClasses` and `select` are consistent, regardless of when `setcolorder` occurs, or whether each element of `colClasses` is ordered or has the same order as `select`. 

I appreciate that there's more code than I've explained, so happy to answer more questions.

HughParsonage:
I should mention the following things, which might explain why the complexity of the code increased so much:

  * There was a bug for factor columns that I noticed when rewriting the code: `colClasses = list(factor = 1, factor = 2)` would only set column 1 to factor. This is the reason I use a nested `for` loop in the `list` colClasses type.
  * When I checked, `NULL` colClasses were being ignored, not dropped. This bug was not present in the current CRAN version, so might be due to something new in `fread.c`. Nonetheless, `set_colClasses_ante` ensures they are dropped.

mattdowle:
Bug finds and fixes are great. Brief comments to each part : 
**Why are there 3 switchs?**   How about putting the single switch after the line that massages `colClasses` into a common format?  Rather than switching types depending on the different formats for `colClasses`.
**Why do all the classes have a new function defined with a try_ prefix?** I like the design choice and the warning very much.  Instead of creating a new "try_" function for each type, I'm more asking if a single `try()` can be wrapped around the switch to the appropriate coercion method/function.
**What's the preamble for?** I like the warning etc but it was the repetition of the pre/post around functions that I was questioning.  I'll have to look at this bit again, maybe once we've simplified the switch first.
**Why is there an is.sorted(select)?**  I looked at 1445 but I'm not grasping the interaction between `colClasses=` and `select=`.  They independently refer to columns in the original data, don't they? 
Could you move the branch into the main project please.  Then I and others can push to it.

HughParsonage:
Will take a look at the rest, but how do I 'move the branch into the main project'?

mattdowle:
Not sure how to move branch either :)  Looks like this PR's target can be changed to a new branch and then we could merge it to there. But then we'd need a new PR to merge from there to master.
Can you grant write access to me (and Rdatatable team ideally) on your existing branch.  That might be easiest actually if it works.

HughParsonage:
Does the branch I just created `colClasses-valid-type-2545` suffice?

HughParsonage:
@mattdowle I think I've addressed your first three points. What you suggested is better. The only potential downside is that it results in repeated queries to `switch` in cases like `colClasses = list(factor = 1:10e3)` but I doubt the overhead (1µs in that example) is worth the few cases of extremely wide but very short files. 

Investigating the fourth point (`is.sorted`) revealed a bug in the implementation of the original fix for #1445 when `select` is a non-monotonic set of column numbers. I've added a fix and tests for that particular issue.

mattdowle:
Thanks for giving me push access to your branch.  Let's proceed working in this PR from your branch then.  In future, create branch in this main project now you're a member, but on this occasion, please just delete colClasses-valid-type-2545.  IIUC!

MichaelChirico:
this from `TODO` in `tests.Rraw` is covered here, right?

> # TO DO: tests of freading classes like Date and the verbose messages there.

HughParsonage:
@MichaelChirico Yes.

HughParsonage:
Would also close #491.

HughParsonage:
I've edited the top post to better reflect the pull request in its current form. In particular, I found other issues related to this pull request, though there may be more. If there are other issues that are very closely related, it might be worth knocking them over here too, though I'm mindful it's easier to review and accept tightly scoped PRs.

HughParsonage:
Closes #2610

HughParsonage:
Hi @mattdowle , I've noticed that this pull request has gone a bit stale. I understand that the PR is a bit unwieldy, so if you like I can close and reissue the pull request as-is on a proper branch.

I could also split this PR up to reflect the changes a bit more discretely if you'd like. 

The reason I bring this up now is that the PR includes a patch for the ggplot2 tests, which are currently failing on CRAN. This unfortunately means a package of mine is liable to be archived on CRAN (due to that package's unfortunate dependency tree). 

Let me know. My priority is making the PR easy to accept or reject!

st-pasha:
@HughParsonage Small question: what do the ❓marks next to issues #1426 and #1656 in this PR's description mean? Does it mean those issues remain unresolved? Is it because they are too hard to resolve, or you don't know how to resolve, or out-of-scope for this PR, or you don't know whether they should be implemented at all, or ..?

HughParsonage:
Closest meaning would be 'partially resolved':

### if select is used, colClasses need only correspond to the columns in select (with caveat)
Basically resolved: there are combinations of `select` and `colClasses` that don't have a clear definition. But otherwise resolved. 

### Issue #1656 fread/fwrite *base data types* directly for efficiency
Not fully resolved: base data types can now be implemented within `fread`, but not in C, and unlike currently supported types, they are not inferred from the content of the file -- they have to be specified.

R/fread.R
```diff
@@ -164,29 +169,61 @@ autostart=NA)
...
+             "complex" = as.complex(v),
+             "raw" = as_raw(v),  # Internal implementation
+             "Date" = as.Date(v),
+             "POSIXct" = as.POSIXct(v),
+             # finally:
+             methods::as(v, new_class))
+      },
+      warning = fun <- function(e) {
+        etype = if (inherits(e,"error")) "error" else "warning"
+        warning(sprintf("Column '%s' was set by colClasses to be '%s' but fread encountered the following %s:\n\t%s\nso the column has been left as type '%s'",
```
MichaelChirico: Construct without `sprintf` (I don't think I have write access to Hugh's branch...)

R/fread.R
```diff
@@ -164,29 +169,61 @@ autostart=NA)
...
+  for (j in which(colClassesAs!="")) {       # # 1634
+    v <- .subset2(ans, j)
+    new_class = colClassesAs[j]
+    new_v <- tryCatch({    # different to read.csv; i.e. won't error if a column won't coerce (fallback with warning instead)
+      switch(new_class,
+             "factor" = as_factor(v),
+             "complex" = as.complex(v),
+             "raw" = as_raw(v),  # Internal implementation
+             "Date" = as.Date(v),
+             "POSIXct" = as.POSIXct(v),
```
MichaelChirico: Is there any way to build customized timezones into the API? Otherwise we should use `tz = 'UTC'` by default? Since the same string on different machines will parse differently otherwise...

man/fread.Rd
```diff
@@ -42,7 +42,14 @@ autostart=NA
...
+  \item{colClasses}{If \code{NULL}, the default, the classes of the columns are inferred from the input's contents.
+  Otherwise, \code{colClasses} specifies the classes for the columns, as in \code{\link[utils]{read.csv}}, but with the following differences:
+  \itemize{
+    \item{If coercion results in an error or introduces \code{NA}s, the attempt is aborted for that column with warning and the column's type is left unchanged (probably \code{character}).}
+    \item{Named list of vectors of column names or numbers are supported where the list names are the class names. The \code{list} form makes it easier to set a batch of columns to be a particular class; see examples. When column numbers are used in the `list` form, they refer to the column number in the file, not the column number after \code{select} or \code{drop} has been applied.}
+    \item{Columns are not demoted to a lower type if this would risk loss of information. You have to coerce such columns afterwards yourself, if you really require data loss.}
```
MichaelChirico: Points 1 & 3 have a lot of overlap, maybe just combine?

R/fread.R
```diff
@@ -123,8 +124,12 @@ autostart=NA)
   }
   if (!is.null(colClasses) && is.atomic(colClasses)) {
     if (!is.character(colClasses)) stop("colClasses is not type list or character vector")
-    if (!length(colClasses)) stop("colClasses is character vector ok but has 0 length")
-    if (!is.null(names(colClasses))) {   # names are column names; convert to list approach
+    if (!length(colClasses)) {
+      colClasses=NULL;
+    } else if (identical(colClasses, "NULL")) {
```
MichaelChirico: what's the use case for this? It's not mentioned in the manual...

R/fread.R
mattdowle: I don't mind either way, but it's ok with `sprintf` isn't it?

R/fread.R
mattdowle: Good idea:  could add `tz=` to fread I guess.  But shouldn't default be whatever `as.POSIXct()` does by default (which is local time iirc) ?

man/fread.Rd
mattdowle: 1 is more about classes like POSIXct which take as input the character string.
3 is more about numeric data like 3.14 being specified as integer.
Maybe more examples and/or better wording?

R/fread.R
MichaelChirico: I'm basically arguing against R default behavior since I've always found it a bit murky to use inferred timezones --> `fread` behaves differently on different machines.

Adding `tz` to `fread` feels a bit strange, maybe accept `tz` as a "class" within `colClasses` a la

`colClasses = list(POSIXct = 'time', tz = 'UTC')`

Or (requires a bit more parsing within our code but I think feels pretty natural):

`colClasses = list(POSIXct[UTC] = 'time')`

R/fread.R
mattdowle: Not sure. Just saw it done in Hugh's branch and liked it. Maybe just to catch beginner errors?  There is a warning that it is taken to mean NULL.
I can see that it breaks the case-1-not-different rule.  So a 1-column file should have the column dropped (and null data.table returned) when colClasses="NULL"?  For consistency.  On the other hand beginners might well try "NULL".

mattdowle: Maybe user has a loop through types,  and all columns should be coerced to that type. `for (class in c("integer","double","NULL")) fread(....,  colClasses=class)`

R/fread.R
MichaelChirico: I took the cue from [WRE Section 1.7](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Diagnostic-messages):

> Try not to split up messages into small pieces. In C error messages use a single format string containing all English words in the messages.
> In R error messages do not construct a message with `paste` (such messages will not be translated) but via multiple arguments to `stop` or `warning`, or via `gettextf`.

I already purged other instances of this construction elsewhere in the code, I can't find the PR right now...

mattdowle: As it's merged now, follow up PRs are good.  I double-checked Hugh is a project member so he can create branches too.

R/fread.R
MichaelChirico: Seems innocuous enough, just wondering if we should document it if there's a specific use case in mind... @HughParsonage any thoughts?

man/fread.Rd
MichaelChirico: Sounds good. At worst a tiny bit repetitive but maybe that's a good thing as it's pretty crucial.

R/fread.R
MichaelChirico: yep, working on it 👍

R/fread.R
mattdowle: But eventually `colClasses` won't be needed to get a `POSIXct` (it would be built in as native type). Hence `tz=` being argument to `fread`.

MichaelChirico: And I guess `tz` argument could have an API like `colClasses` to allow flexibly specifying multiple time zones:

`tz = list(UTC = c('start', 'end'), 'Australia/South' = c('start_local', 'end_local'))`

and

`tz = c(start = 'UTC', end = 'UTC', start_local = 'Australia/South', end_local = 'Australia/South')`

jangorecki: I think it make sense to use `tz="UTC"` and just document that

MichaelChirico: my only hesitation is I'm not _100%_ sure we can do this before implementing the tz argument... e.g. some times only exist in certain time zones, so conversion may fail if we assume UTC but could read on user's machine otherwise?

jangorecki: if it may fail then `tryCatch`, providing more complex structure to `colClasses` is good idea but should be added with care (ideally another PR), ideally aligning to API we will use in csvy specification.

PR #2562, opened 2018-01-10
trivial (temporary?) fix for confusing list column printing, related,…

MichaelChirico:
… part of #1523

There's some more discussion on the merits of instead using `toString` here which I tend to like, see the discussion on #1523; filing this PR to jumpstart that convo & offer a temporary fix in any case since this was trivial to do.

Also brought `is.formula` into `print.data.table.R` since that's where it belongs. Side note -- shouldn't the function instead be `inherits(x, 'formula')`?

### DIFF
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -111,9 +111,9 @@ format.data.table <- function (x, ..., justify="none") {
   }
   format.item <- function(x) {
     if (is.atomic(x) || is.formula(x)) # FR #2591 - format.data.table issue with columns of class "formula"
-      paste(c(format(head(x,6), justify=justify, ...), if(length(x)>6)""),collapse=",")  # fix for #5435 - format has to be added here...
+      paste(c(format(head(x, 6L), justify=justify, ...), if(length(x) > 6L) "..."), collapse=",")  # fix for #5435 - format has to be added here...
     else
-      paste("<",class(x)[1L],">",sep="")
+      paste("<", class(x)[1L], ">", sep="")
   }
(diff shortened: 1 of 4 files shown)


R/print.data.table.R
```diff
@@ -145,3 +145,6 @@ shouldPrint = function(x) {
 # for removing the head (column names) of matrix output entirely,
 #   as opposed to printing a blank line, for excluding col.names per PR #1483
 cut_top = function(x) cat(capture.output(x)[-1L], sep = '\n')
+
+# FR #2591 - format.data.table issue with columns of class "formula"
+is.formula <- function(x) class(x) == "formula"
```
MichaelChirico: ^shouldn't this be `inherits(x, 'formula')`?

mattdowle: Yes. Good spot. I'm about to merge so I'll do it quickly.

mattdowle:
Agree with adding `...` after comma though.  Better and clearer.  And good spot on lack of `inherits`.

MichaelChirico:
all the same I think it's a holdover until we hash out the right way of incorporating `toString` which seems better for the goal of printing... the first 6 elements is probably accurate most of the time, but I think `toString` offers appropriate flexibility

mattdowle:
Yep - `toString` sounds good with a width for the column I guess, rather than fixed 6 items which can result in varying width.  As long as separator is space-free `","` rather than `", "`!  I can't see a way to change that in `toString`.

MichaelChirico:
Hmm, agree `","` is better. probably just write our own method.

PR #2565, opened 2018-01-13
Fixed quotes showing in tables() output in dev.

mattdowle:
In master before this PR I was seeing quotes and a NULL in the output of `tables()` as follows.
```
> DT=data.table(1L)
> tables()
     NAME NROW   NCOL   MB   COLS KEY   
[1,] "DT" "   1" "   1" " 0" "V1" "NULL"
Total: 0MB
```

Seems to have arose in PR #1804 for issue #1648.
This PR simplifies `tables()` by leaving more of the formatting to standard `print(DT)` rather than creating a character matrix (something I did originally and was always a bit messy.)
Please check @MichaelChirico -- thanks.

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -18,36 +18,24 @@ tables <- function(mb=TRUE, order.col="NAME", width=80,
       data.table(NAME = dt_n,
                  NROW = nrow(DT),
                  NCOL = ncol(DT))
-    if (mb)
-      # mb is an option because object.size() appears to be slow.
-      # **TO DO: revisit**
-      set(info_i, , "MB",
-        #1048576 = 1024^2
-        round(as.numeric(object.size(DT))/1048576))
+    if (mb) set(info_i, , "MB", round(as.numeric(object.size(DT))/1024^2))
+    # mb is an option because object.size() appears to be slow. TO DO: revisit
     set(info_i, , "COLS", list(list(names(DT))))
     set(info_i, , "KEY", list(list(key(DT))))
     if (index) set(info_i, , "INDICES", list(list(indices(DT))))
     info_i
   }))
-  info[ , NROW := format(sprintf("%4s", prettyNum(NROW, big.mark=",")), justify="right")]   # %4s is for minimum width
-  info[ , NCOL := format(sprintf("%4s", prettyNum(NCOL, big.mark=",")), justify="right")]
-  if (mb) {
-    total = sum(info$MB)
-    info[ , MB := format(sprintf("%2s", prettyNum(MB, big.mark=",")), justify="right")]
-  }
   if (!order.col %in% names(info)) stop("order.col='",order.col,"' not a column name of info")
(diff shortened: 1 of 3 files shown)


MichaelChirico:
starting to travel, will try and pull and check offline while so. review
next wk

R/tables.R
```diff
@@ -18,36 +18,24 @@ tables <- function(mb=TRUE, order.col="NAME", width=80,
...
-  colnames(m)[3] = sprintf(paste("%",nchar(m[1,"NCOL"]), "s", sep=""), "NCOL")
-  if (mb) colnames(m)[4] = sprintf(paste("%", nchar(m[1,"MB"]), "s", sep=""), "MB")
-  m[ , "COLS"] = substring(m[,"COLS"], 1L, width)
-  m[ , "KEY"] = substring(m[,"KEY"], 1L, width)
   if (!silent) {
-    print(m, quote=FALSE, right=FALSE)
-    if (mb) cat("Total: ", prettyNum(as.character(total), big.mark=","), "MB\n", sep="")
+    # prettier printing on console
+    tt = copy(info)
+    tt[ , NROW := format(sprintf("%4s", prettyNum(NROW, big.mark=",")), justify="right")]   # %4s is for minimum width
```
MichaelChirico: I think the `width` argument of `format.default` can be used instead of calling `sprintf`:

    format(prettyNum(NROW, big.mark=","), width = 4L, justify="right")

little test I ran to convince myself:

```
all(replicate(1000, {
  ns = prettyNum(sample(20000, 10), big.mark = ',')
  wid = format(sprintf('%4s', ns), justify = 'right')
  spr = format(ns, width = 4L, justify = 'right')
  identical(wid, spr)
}))
# [1] TRUE
```

R/tables.R
```diff
@@ -18,36 +18,24 @@ tables <- function(mb=TRUE, order.col="NAME", width=80,
...
+    tt[ , NCOL := format(sprintf("%4s", prettyNum(NCOL, big.mark=",")), justify="right")]
+    if (mb) tt[ , MB := format(sprintf("%2s", prettyNum(MB, big.mark=",")), justify="right")]
+    print(tt, class=FALSE, nrow=Inf)
+    if (mb) cat("Total: ", prettyNum(as.character(sum(info$MB)), big.mark=","), "MB\n", sep="")
```
MichaelChirico: why `as.character` here? `prettyNum` should handle the type conversion?

R/tables.R
MichaelChirico: Anyway we use the `format(prettyNum(...))` enough times here to warrant a function (`commaNum`? `print_num`?). Maybe defined within the scope of `tables`? Or something in `utils.R`?

R/tables.R
mattdowle: Good spot!

mattdowle:
Thanks @MichaelChirico. Good spots and changes.

Issue #2546, opened 2018-01-03
Fails on CRAN due to setwd() to read-only installation directory


mattdowle:
Current errors on CRAN including R-patched (so it's not R-devel) pasted below.
Thanks to Kurt's investigation, it's due to data.table calling `setwd()` to installation directory to reach test files loaded by `fread`.  But the installation directory is (very reasonably) now mounted read-only on CRAN and a write of the plot in non-interactive mode is failing for the ggplot2 compatibility tests.
To reproduce: install data.table, change install directory to read-only, `library(data.table)`, `library(ggplot2)`, `test.data.table()` (I didn't yet try that, but in theory that should reproduce it).
One solution is to stop using `setwd()` to avoid any accidental or superfluous writes there. But then all `fread` tests loading test files need to be changed somehow, prefixed with the location.

![image](https://user-images.githubusercontent.com/2779998/34536112-0ac553ac-f079-11e7-8467-89249feb7213.png)

```
Test 167 Error in (function (file = if (onefile) "Rplots.pdf" else "Rplot%03d.pdf", :
  cannot open file 'Rplots.pdf'
Test 167.1 Error in (function (file = if (onefile) "Rplots.pdf" else "Rplot%03d.pdf", : 
  cannot open file 'Rplots.pdf'
Test 167.2 Error in (function (file = if (onefile) "Rplots.pdf" else "Rplot%03d.pdf", : 
  cannot open file 'Rplots.pdf'
Test 168 Error in (function (file = if (onefile) "Rplots.pdf" else "Rplot%03d.pdf", : 
  cannot open file 'Rplots.pdf'
Test 168.1 Error in (function (file = if (onefile) "Rplots.pdf" else "Rplot%03d.pdf", : 
  cannot open file 'Rplots.pdf'
   
Running test id 1696 
Error in eval(exprs[i], envir) : 
5 errors out of 5946 (lastID=1751, endian==little, sizeof(long double)==16, sizeof(pointer)==8) in inst/tests/tests.Rraw on Tue Jan 2 07:11:13 2018.
Search tests.Rraw for test numbers: 167, 167.1, 167.2, 168, 168.1.
Calls: test.data.table -> sys.source -> eval -> eval
Execution halted 
```

HughParsonage:
I think this would be a solution

```r
if ("package:ggplot2" %in% search() && requireNamespace("grDevices", quietly = TRUE)) {
[...]

PR #2571, opened 2018-01-17
add test from TODO list insuring dcast plays well with being imported

MichaelChirico:
This has been sitting in the TODO queue of `tests.Rraw` for a while:

> Add test for fixed bug #5519 - dcast returned error when a package imported data.table, but dint happen when "depends" on data.table. This is fixed (commit 1263 v1.9.3), but not sure how to add test.

I couldn't check the original issue (R Forge days), so I'm not 100% sure this captures the goal of the comment. But the infrastructure for building a mini-package within the test is the trickier part anyway and that's covered here.


MichaelChirico:
Can't say I understand the AppVeyor/Travis results... I guess it must be related to permissions from creating/installing the package, then uninstalling and removing? `test.data.table()` works on my machine

HughParsonage:
I think the best approach would be to create a genuinely separate minimal package, then add it to the drat repository and to `Suggests:` then attach it (conditionally) as normal.

MichaelChirico:
Open to ideas; mainly filed the PR as an impetus for conversation. Not really sure how to interact with drat -- one potential problem with the current setup is that `data.table` is already loaded when `testDTimport`'s function is run. Not sure if that would mask potential reversions of the bug. Seems the way around that is to add tests to `testDTimport`; not sure if there'd be a way to trigger `testDTimport` to install & test itself from within `data.table` tests 🤔

HughParsonage:
This is complicated but one solution would be to place the relevant test in the minimal package rather than here and run reverse dependency checks from `data.table`, to include the minimal package.

mattdowle:
Nice try!!  But we can't go anywhere near things like calling `install.packages()` from tests. It just won't fly on CRAN. Very reasonably, no writes to the system allowed. If any writes happen they must happen to `tempfile()` and that's about it.  Imagine if the `install.packages()` worked but then the test failed or crashed unexpectedly before the `remove.packages()`. How would the write to the library get cleaned up in that case?  I recently learnt in #2546 that CRAN's library is now mounted read-only, too, which seems like a good thing, hence the `testDir()` addition to avoid even just `setwd()` to library.

@HughParsonage's thought should fly. There is one test package already (DtNonAsciiTests) in the level up, here: https://github.com/Rdatatable. This test could be added there and that package's name changed to something more general (like DtExtraTests?).  We never got to finishing off that package.  IIRC the idea was it had tests that could not live on CRAN due to non-ASCII characters in test files which are not allowed (just on Mac iirc).  Maybe that could be overcome now, with fresh eyes.

Closing for now to be started again in DtExtraTests package or similar.

MichaelChirico:
@mattdowle cool, I always wondered what the `DtNonAsciiTests` repository was about. Agreed it could serve more general testing purposes as well.

Issue #1416, opened 2015-10-29
fread ignores first 2 lines of file with tabs after data in several lines 


tdhock:
To see the problem please download https://gist.githubusercontent.com/tdhock/bc1d9c0a76e4df386199/raw/fe436fb41e42c9ba89d8b567abfca587d323aa36/iterations.txt

and then run

``` r
> data.table::fread("iterations.txt")
     V1       V2           V3     V4 V5
 1:   2 -3274352 63539392.790   23.6 NA
 2:   3 -2166076  1108276.003   34.5 NA
 3:   4 -2054893   111182.812   45.3 NA
 4:   5 -1992373    62520.297   56.0 NA
 5:   6 -1913422    78950.778   66.8 NA
 6:   7 -1802173   111249.167   77.7 NA
...
```

The problem seems to be that the first two lines do not have a trailing tab but all the rest of the lines do (screenshot from emacs which highlights all the tab characters):

![screenshot-emacs-tab](https://cloud.githubusercontent.com/assets/932850/10827127/a7581c4a-7e43-11e5-9a82-6a7841afea5d.png)

I expected the output of read.table:

``` r
> head(read.table("toby/iterations/iterations.txt", header=TRUE))
  Iteration EstimatedLogLikelihood       Change TotalTimeSec
1         1              -66813745            -         12.3
2         2               -3274352 63539392.790         23.6
3         3               -2166076  1108276.003         34.5
4         4               -2054893   111182.812         45.3
5         5               -1992373    62520.297         56.0
6         6               -1913422    78950.778         66.8
```

Thanks for looking into this,

``` r
> sessionInfo()
R version 3.2.2 (2015-08-14)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu precise (12.04.5 LTS)

locale:
 [1] LC_CTYPE=en_CA.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_CA.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_CA.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

[...]

PR #2580, opened 2018-01-21
Auto fill too few column names

### LINKED ISSUE #763: fread does not handle row.names produced by write.table
Hi,
I am using `data.table` version `1.9.2`
base::write.table has default `row.names = TRUE`. When using fread to this this the file, it doesn't handle the row names correctly. I understand that `fread` want to preserve the information from `input` file. Then it should consequently handle the `header` option correctly. 

The current workaround is to use `write.table(row.names = FALSE)`. But I would suggest to have `row.names` option in `fread` for flexibility.

``` R
DT = data.table(x=rep(c("a","b","c"),each=3), y=c(1,3,6), v=1:9)
write.table(DT, file = "test_writeDT.tsv", sep = "\t")
> fread(input="test_writeDT.tsv") #header is lost
   V1 V2 V3 V4
1:  1  a  1  1
2:  2  a  3  2
3:  3  a  6  3
4:  4  b  1  4
5:  5  b  3  5
6:  6  b  6  6
7:  7  c  1  7
8:  8  c  3  8
9:  9  c  6  9
> fread(input="test_writeDT.tsv", header=TRUE) # Wrong header
   1 a 1 1
1: 2 a 3 2
2: 3 a 6 3
3: 4 b 1 4
4: 5 b 3 5
5: 6 b 6 6
6: 7 c 1 7
7: 8 c 3 8
8: 9 c 6 9
> fread(input="test_writeDT.tsv", drop = 1, header=TRUE) # Looks about right, but still wrong header
   a 1 1
1: a 3 2
2: a 6 3
3: b 1 4
4: b 3 5
5: b 6 6
6: c 1 7
7: c 3 8
8: c 6 9
> read.table("test_writeDT.tsv") # Correct
  x y v
1 a 1 1
2 a 3 2
3 a 6 3
4 b 1 4
5 b 3 5
6 b 6 6
7 c 1 7
8 c 3 8
9 c 6 9
```

### LINKED ISSUE #1625: fread incorrectly skips header on this file
fread("http://www.nzta.govt.nz/resources/crash-analysis-system-data/docs/crash-data-2014.csv") :
  Starting data input on line 2 and discarding line 1 because it has too few or too many items to be column names or data: "TLA NAME","CRASH ROAD","CRASH DIST","CRASH DIRN","INTSN","SIDE ROAD","CRASH ID","CRASH DATE","CRASH DOW","CRASH TIME","MVMT","VEHICLES","CAUSES","OBJECTS STRUCK","ROAD CURVE","ROAD WET","LIGHT","WTHRa","JUNC TYPE","TRAF CTRL","ROAD MARK","SPD LIM","CRASH FATAL CNT","CRASH SEV CNT","CRASH MIN CNT","PERS AGE1","PERS AGE2","EASTING","NORTHING"

From : http://stackoverflow.com/questions/28100868/download-file-inserts-an-extra-cr-after-each-crlf

### LINKED ISSUE #1818: fread row.names option
When reading in a text file generated by write.table function in R with both row names and column names included, fread reports the following warning because the header row has one less element than the remaining data rows. As a result the header will just be omitted. fill=TRUE doesn't work since the fill will be the last element and all positions are incorrect.
The warning reads: Starting data input on line 2 and discarding line 1 because it has too few or too many items to be column names or data
I am wondering whether a parameter similar to row.names in data.frame function can be added so that fread knows the first column is the row names and will expect the first row have one less element.

mattdowle:
Closes #1625
Closes #1818
Closes #763 

#1416 has more complex nuances I'll leave for another time.


mattdowle:
The close and reopen was [this AppVeyor workaround](https://help.appveyor.com/discussions/problems/4830-stuck-on-appveyor-was-unable-to-build-non-mergeable), which worked.

mattdowle:
Probably more common than the example in #1625 are the examples in #1818, #763 and #1416 where the first column is a row name or index with a missing column name.  If there is just one column name missing, changing the fill to be on the left.

PR #2628, opened 2018-02-14
Fixed bug where setattr crashed on character(0) value.

### LINKED ISSUE #2386: setattr crashed when given empty character vector as name or value
Trying to run setattr(numeric(1), "class", character(0)) or setattr(numeric(1), character(0), "value")  crashes because the setattrib c function doesn't check if the value or name string vector actually has length > 0 before accessing it with STRING_ELT.

MarkusBonsch:
Closes #2386 .
`setattr(1, "class", character(0))` used to cause a segmentation violation. Now, it throws a proper error.


mattdowle:
Great find.  I guess this will fail on any object, not just `1`, that `1` was just to illustrate?  I sof, the side-issue then with `1L` and possibly `1` too is that they could be R's internal length-1 globals.  Then `setattr()` could potentially add an attribute to the global value of `1L`.  Similarly to how data.table could change the global value of `TRUE` to `FALSE` several years ago in R-devel before release that was caught early on via tests against R-devel .  Will use this PR as a nudge to look into length-1 globals status in R-devel/altrep.

MarkusBonsch:
@mattdowle Wow, I didn't even know these things existed :). You are completely right, `1` was just to illustrate. I replaced with `data.table(x=1:10)` in the tests in order to avoid side-effects.

PR #2629, opened 2018-02-16
Improve fread parser for floats

### LINKED ISSUE #2625: fread parses some invalid tokens as numeric
```
> fread("A,B,C,D,E\n.,+.,.e,.e+,0e\n")
   A B C D E
1: 0 0 0 0 0
```

st-pasha:
This PR allows fread to recognize floating-point literals with arbitrary many digits, and parse them as doubles. This includes numbers such as `10000000000000000000000000000.0` or `2.23498720349871093847109387e-21`, but not `100000000000000000000000000` (long integer will still be parsed as string).

The reason for this change is that we have already seen multiple datasets where floating-point values are stored with more than the "canonical" 17 significant digits. To an unsuspecting user, such number looks completely innocuous: eg. `2.49102793750273308`, and it comes at a complete surprise that fread would parse it as a string... It is also relatively easy to produce these extra digits of precision "by accident" asking for too many digits in `printf()`, or printing with `%f` format, etc.

This PR also fixes issue #2625 by being more careful about checking whether the literal is valid.

Closes #2625


MichaelChirico:
Just noticed this in `examples(fread)`

```
DT = fread("A\n1.010203040506070809010203040506\n")  # too precise for double, so read as character
# TODO: add numerals=c("allow.loss", "warn.loss", "no.loss") from base::read.table
typeof(DT$A)=="character"   # TRUE
```

Seems related

Issue #2531, opened 2017-12-19
fread spends too much time in is_url/is_secureurl/is_file for long in memory input


javrucebo:
## Summary 

When `fread` is fed with a character string as input the routine spends considerable amount of time detecting that the supplied input is *not* a filename or an url.

This is due to `grepl` not scaling well for large input as used in fread.

Example 
```R
is_url <- function(x) grepl("^(http|ftp)s?://", x)
```
Although the pattern is anchored at the beginning of the string, running `grepl` for large inputs will take a lot of time for large inputs (more detailed benchmarks further down).
This will lead then to the full call to `fread` possibly **spending a third of the time for those supposedly simple checks**. (example also below)

Possible solutions could be one or more of the following
* changing the call to `grepl` to use PERL regexp engine `perl=TRUE` 
  Alternatively use another method to determine whether the input starts with url (see benchmarks below)
* adding an explicit optional argument which denotes string input, e.g. `str=` to denote that the input is to be considered as the data and skip the tests for url or file. This would be  similar in spirit to the `file=` argument.
  As a side-effect this would also allow input to be read which consits only of url's and having no header, e.g.
 ```R
fread("http://hkhfsk\nhttp://fhdkf\nhttp://kjfhskd\nhttp://hfkjf", header=FALSE)
```
* consider an input, which exceeds a certain number of characters in length, not being an URL

## Profiling example

```R
library(data.table)
# create a random character input in 5 columns with 20000*200 lines
randomString <- function() {
  paste(sample(c(LETTERS, letters), sample(1:9,1)), collapse="")
}
input <- paste(rep(paste(replicate(20000,paste(replicate(5, randomString()), 
                                              collapse="\t")),
                         collapse="\n"),200),
               collapse="\n")
# profile the call to fread
Rprof()
invisible(fread(input, header=FALSE))
Rprof(NULL)
summaryRprof()
```
```
$by.self
        self.time self.pct total.time total.pct
".Call"      5.34    64.49       5.34     64.49
"grepl"      2.94    35.51       2.94     35.51
```

## Benchmarking `grepl` and friends

Comparing different functions to verify whether a string starts with any of http(s)/ftp(s) or file shows that `grepl` scales badly and is by far the slowest of the tested variants.
Adding simply `perl=TRUE` already improves by around factor 100 for large inputs (code further below)

[...]

PR #2630, opened 2018-02-16
change the order of `file.exists(input)` and `grep('\\n|\\r', input)` in `fread.R`

javrucebo:
change the order of `file.exists(input)` and `grep('\\n|\\r', input)`
in `fread.R` in order to speed up execution time when `input` is the
actual data. `file.exists` is very slow, so checking first
for new lines speeds up the process.

For benchmark on `file.exists` see also [this comment](https://github.com/Rdatatable/data.table/issues/2531#issuecomment-366179409) in #2531

Not adding additional tests, as no change in functionality, just speed up of execution. (current tests all pass locally).
Not adding NEWS entry, as there is already one for #2531 and this PR is just a trivial addition.

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -33,10 +33,10 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
     if (!is.character(input) || length(input)!=1L) {
       stop("'input' must be a single character string containing a file name, a system command containing at least one space, a URL starting 'http[s]://', 'ftp[s]://' or 'file://', or, the input data itself containing at least one \\n or \\r")
     }
-    if (file.exists(input)) {
-      if (isTRUE(file.info(input)$isdir)) stop("File '",input,"' is a directory. Not yet implemented.")
-    } else if ( input == "" || length(grep('\\n|\\r', input)) ) {
+    if ( input == "" || length(grep('\\n|\\r', input)) ) {
       # input is data itself containing at least one \n or \r
+    } else if (file.exists(input)) {
+      if (isTRUE(file.info(input)$isdir)) stop("File '",input,"' is a directory. Not yet implemented.")
     } else {
       if (substring(input,1L,1L)==" ") {
         stop("Input argument is not a file name and contains no \\n or \\r, but starts with a space. Please remove the leading space.")


MichaelChirico:
it seems like an R bug that length(grep(.)) could be faster than grep(.),
no? isn't that the point of having a function like gepl in the first place??

On Feb 16, 2018 5:02 PM, "javrucebo" <notifications@github.com> wrote:

change the order of file.exists(input) and grep('\\n|\\r', input)
in fread.R in order to speed up execution time when input is the
actual data. file.exists is very slow, so checking first
for new lines speeds up the process.

See also discussion in #2531
<https://github.com/Rdatatable/data.table/issues/2531>

Not adding additional tests, as no change in functionality, just speed up
of execution. (current tests all pass locally).
Not adding NEWS entry, as there is already one for #2531
<https://github.com/Rdatatable/data.table/issues/2531> and this PR is just
a trivial addition.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/Rdatatable/data.table/pull/2630
Commit Summary

   - change the order of `file.exists(input)` and `grep('\\n|\\r', input)`

File Changes

   - *M* R/fread.R
   <https://github.com/Rdatatable/data.table/pull/2630/files#diff-0> (6)

Patch Links:

   - https://github.com/Rdatatable/data.table/pull/2630.patch
   - https://github.com/Rdatatable/data.table/pull/2630.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Rdatatable/data.table/pull/2630>, or mute the thread

javrucebo:
@MichaelChirico 
The original reported issue with `grepl` and the use here are distinct cases.
There is not much difference in testing for newlines between `grep` and `grepl`.
What was slow is the anchored search at the beginning of the string - clearly this should be much faster be fixed in base 'R`

```
input <- paste(rep("1,2,3,4.567,some text field", 1e7),collapse="\n")
system.time(length(grep("\n", input)))
##   user  system elapsed 
##  0.000   0.000   0.001 
system.time((grepl("\n", input)))
##   user  system elapsed 
##      0       0       0 
system.time((grepl("^http://", input)))
##   user  system elapsed 
##  2.683   0.000   2.685 
```

HughParsonage:
True, though checking `file.exists` is faster than checking for newlines then checking `file.exists`.

It would seem `grepl("\n", ., fixed = TRUE) || grepl("\r", ., fixed = TRUE)` is faster still, noting we have already checked that length(.) == 1 a few lines earlier.

javrucebo:
@HughParsonage 
the point here is, if `input` is the actual raw data - and therefore potentially very large - then `file.exists` is really slow and eating up most of execution time of `fread`.

If on the other hand `input` is a file name, then we can safely assume that it is a relatively short string (not megabytes) and having the `length(grep('\\n|\\r', input))` check before `file.exists` does not add a real penalty for small inputs.

then `length(.)==1` check you refer to is checking something else - whether the input is a vector with length > 1, which fread does not accept.

Issue #2593, opened 2018-01-27
inconsistencies between merge.data.frame and merge.data.table


MarkusBonsch:
We currently have several issues open that are concerned with the differences in behavior between `merge.data.table` and `merge.data.frame`. This issue is aimed at collecting all of them in order to facilitate fast and efficient solutions without duplicated work.

* #2592  (severe bug imho)
* #2587 merge.data.table ignores incomparables without warnings
* #2574 sort = FALSE does't generate equivalent results with data.frame.merge 
* #2594 forderv orders NAs differently than base::order by default

Issue #2585, opened 2018-01-23
Installation via "R CMD INSTALL --no-multiarch --with-keep.source data.table" fails on dev


MichaelChirico:
Platform info

```
R version 3.4.1 (2017-06-30)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: OS X El Capitan 10.11.6
```

    R CMD INSTALL --no-multiarch --with-keep.source data.table

fails for me on the current dev. (this is also the default command run by `Install and Restart` in RStudio). The error is:

```
Error: package or namespace load failed for ‘data.table’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/Library/Frameworks/R.framework/Versions/3.4/Resources/library/data.table/libs/datatable.so':
  dlopen(/Library/Frameworks/R.framework/Versions/3.4/Resources/library/data.table/libs/datatable.so, 6): Symbol not found: _ENC2UTF8
  Referenced from: /Library/Frameworks/R.framework/Versions/3.4/Resources/library/data.table/libs/datatable.so
  Expected in: flat namespace
 in /Library/Frameworks/R.framework/Versions/3.4/Resources/library/data.table/libs/datatable.so
```

I assume this is related to the work on #2566/#2462.

If instead we just run

```
R CMD build data.table
R CMD INSTALL data.table_1.10.5.tar.gz --no-multiarch --with-keep.source data.table
```

it runs fine...

HughParsonage:
Permissions issue?

MichaelChirico:
`sudo R CMD INSTALL --no-multiarch --with-keep.source data.table` also fails

mattdowle:
Is this one still a problem in dev as of now?  Back in January, `ENC2UTF8` was a function and that might have been the problem (it is mentioned in the error you pasted).  Now it's a macro, I'm hoping it's gone away?  Perhaps related the the fact that R itself has a function of the same name, and perhaps linking the datatable.so into R was a conflict somehow on Mac.

MichaelChirico:
@mattdowle actually no. I do however get these warnings:

```
fread.c:2206:144: warning: format specifies type 'void *' but the argument has
      type 'const char *' [-Wformat-pedantic]
  ...jump=%d, headPos=%p, thisJumpStart=%p, sof=%p", jump, headPos, thisJumpS...
                      ~~                                   ^~~~~~~
                      %s
[...]

PR #2631, opened 2018-02-17
Fixed duplicate column names in `merge()` when by.x in names(y)

sritchie73:
When joining two `data.tables` using `merge()` the resulting `data.table` will contain duplicate column names if `by.x != by.y` and `by.x` is also in `names(y)`.

An example:

```{r}
# Create example data.frames
parents <- data.table(name=c("Sarah", "Max", "Qin", "Lex"), 
                      sex=c("F", "M", "F", "M"), 
                      age=c(41, 43, 36, 51))
children <- data.table(parent=c("Sarah", "Max", "Qin"), 
                       name=c("Oliver", "Sebastian", "Kai-lee"),
                       sex=c("M", "M", "F"), 
                       age=c(5,8,7))

# Merge() creates a duplicated "name" column:
merge(parents, children, by.x = "name", by.y = "parent")
```
Output:
```
   name sex.x age.x      name sex.y age.y
1   Max     M    43 Sebastian     M     8
2   Qin     F    36   Kai-lee     F     7
3 Sarah     F    41    Oliver     M     5
```

This behaviour is also present in `base:::merge.data.frame()`, but throws an additional warning:

```
Warning message:
In merge.data.frame(parents, children, by.x = "name", by.y = "parent") :
  column name ‘name’ is duplicated in the result
```

This patch fixes this problem by checking for names shared between `by.x` and `names(y)`, and adding the appropriate `suffixes` to those column names.


sritchie73:
Also added an equivalent warning message to base for cases when duplicate column names are returned. With the patch above, the only way this can happen is if the user supplies identical values to the suffixes argument (*e.g.* `suffixes=c("", "")`).

inst/tests/tests.Rraw
```diff
@@ -11671,6 +11671,16 @@ unlink(f)
...
+# PR #2631 - Fix duplicated names arising in merge when by.x in names(y)
+# Test should fail in there are any duplicate names after a join
+parents = data.table(name=c("Sarah", "Max"), sex=c("F", "M"), age=c(41, 43))
+children = data.table(parent=c("Sarah", "Max", "Max"), 
+                       name=c("Oliver", "Sebastian", "Michelle"),
+                       sex=c("M", "M", "F"), age=c(5,8,7))
+joined = merge(parents, children, by.x="name", by.y="parent")                       
+test(1877.1, length(names(joined)), length(unique(names(joined)))) 
+test(1877.2, merge(parents, children, by.x="name", by.y="parent", suffixes=c("",""),
+     warning = "column names 'name', 'sex', 'age' are duplicated in the result"),
```
HughParsonage: Trailing `,` should be removed.

sritchie73:
RE: failed checks: I can't work out how to write the test that checks for the appropriate warning message - would appreciate advice here since there's no documentation for data.table:::test().

I also can't run the tests on my local machine because the package fails to compile (I don't have OpenMP).

mattdowle:
Looks good.  On `test()`, search `tests.Rraw` for `warning=` to see other examples in other tests to follow. You can also pass `output=` and/or `error=` to search for substrings in those messages.
It should compile ok without OpenMP as it should detects if OpenMP is available and switches out.  How are you compiling and what is the message it fails with at what point?  This is my dev cycle: [cc.R]( https://github.com/Rdatatable/data.table/blob/master/cc.R)

sritchie73:
Thanks @mattdowle, I couldn't get the package to build using RStudio package build system - but I probably have something misconfigured in my environment.  I was able to get the tests to work with a bit of trial and error.

MichaelChirico:
I added a little bit to the [Contributing](https://github.com/Rdatatable/data.table/wiki/Contributing) wiki to make it a bit easier to understand how the testing regime works. Thanks for the PR.

MarkusBonsch:
I love it, although it breaks consistency with `merge.data.frame`.  There are quite some other issues with merge.data.table at the moment (#2593). Once these are all fixed (I am working on it), we have to update the documentation to clearly reflect, where we have differences to `merge.data.frame`.

MichaelChirico:
@sritchie73 I assume the Rstudio issue is the same as #2585. I've been using the command line to re-install lately

sritchie73:
Thanks @MarkusBonsch , for what its worth I've also sent a similar patch to the R-devel mailing list for merge.data.frame - but the two people who responded (neither core dev team) were pessimistic about my chances of getting the patch accepted.

@MichaelChirico I get a different error:

```
==> R CMD INSTALL --no-multiarch --with-keep.source data.table

* installing to library ‘/Library/Frameworks/R.framework/Versions/3.4/Resources/library’
** libs
clang -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG   -I/usr/local/include  -fopenmp -fPIC  -Wall -g -O2  -c assign.c -o assign.o
clang: error: unsupported option '-fopenmp'
make: *** [assign.o] Error 1
* removing ‘/Library/Frameworks/R.framework/Versions/3.4/Resources/library/data.table’
* restoring previous ‘/Library/Frameworks/R.framework/Versions/3.4/Resources/library/data.table’

Exited with status 1.
```

In the end I managed to get data.table:::test() working almost the same way in my local R session, by copying the file preamble, then fixed the remaining issues that came up in Travis after pushing the changes out to my branch.

sritchie73:
Thanks @jangorecki - I had used `paste` rather than `paste0` in the first instance to match the code style for the existing suffix handling which immediately precedes my patch. Would you prefer I change it to `paste0`, or keep it consistent with the code preceding it?

jangorecki:
Lets wait for R-devel, paste is not that important

mattdowle:
Given Martin Maechler's reply today, it's looking likely to be accepted for R. Wow! Well navigated! I updated manual page accordingly (including link to thread too as Jan suggested) and will merge.
Will invite you to be project member too.

sritchie73:
Thanks!

I am following up with Martin to clarify the functionality of the proposed `no.dups` argument so we can mirror the behaviour of `merge.data.frame`.

PR #2640, opened 2018-02-21
Correct vignette example

HughParsonage:
* `utils::packageVersion` takes a pkg name, `base::package_version` treats a string as a package version for comparison purposes

### DIFF
--- a/vignettes/datatable-importing.Rmd
+++ b/vignettes/datatable-importing.Rmd
@@ -132,7 +132,7 @@ Slightly more extended way which also ensure that installed data.table is in ver
 ```r
 my.write = function (x) {
   if(requireNamespace("data.table", quietly=TRUE) &&
-    (package_version("data.table") >= "1.9.8")) {
+    utils::packageVersion("data.table") >= package_version("1.9.8")) {
     data.table::fwrite(x, "data.csv")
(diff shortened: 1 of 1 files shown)


MichaelChirico:
nice. i've always wondered what `package_version` is....

vignettes/datatable-importing.Rmd
```diff
@@ -132,7 +132,7 @@ Slightly more extended way which also ensure that installed data.table is in ver
 ```r
 my.write = function (x) {
   if(requireNamespace("data.table", quietly=TRUE) &&
-    (package_version("data.table") >= "1.9.8")) {
+    utils::packageVersion("data.table") >= package_version("1.9.8")) {
```
mattdowle: Good spot!  `package_version` is superfluous on the RHS though because `>=` will already dispatch to the correct method via the LHS class.  Examples in `?packageVersion` confirm that.

MichaelChirico:
Seems `package_version` serves the purpose of (1) validation (is an input string valid as a package version?) and (2) decomposition:

```
dput(package_version('2.0.13'))
# structure(list(c(2L, 0L, 13L)), class = c("package_version", "numeric_version"))
```

PR #2641, opened 2018-02-23
CJ fails with proper error msg

### LINKED ISSUE #2636: CJ could fail with proper message when exceeds limit
```r
library(data.table)
CJ(1:100, 1:100, 1:100, 1:100, 1:100, 1:100, 1:100, 1:100, 1:100, 1:100)
#Error in rep.int(y, times = rep.int(x[i], n[i])) : invalid 'times' value
```

jangorecki:
Closes #2636


MichaelChirico:
Could you add a tag in comment for [this line](https://github.com/Rdatatable/data.table/blob/654b44bb5637af6c4f24fe4a91bdb40d98be0b53/R/setkey.R#L357) to update to use `lengths` if/when we update dependency to R >= 3.2.0?

```
n = vapply(l, length, 0L)
```

MichaelChirico:
I would make the error message more helpful to users, e.g.

```
stop("Input would result in ", nrow, " rows, which exceeds ", .Machine$integer.max, ", the largest storable integer on this machine.")
```

PR #2643, opened 2018-02-24
Out-of-sample quote rule bump with warning, #2265

### LINKED ISSUE #2265: fread Unable to handle mis-quoted field if it is out-of-sample
This example:
```
require(data.table)
DT = data.table(A=rep("abc", 10000), B="def")
DT[110, A:='"a"b']
fwrite(DT, f<-tempfile(), quote=F)
fread(f)
```
produces an error message which is misleading:
```
Expecting 2 cols but row 0 contains only 1 cols (sep=','). Consider fill=true. <<"a"b,def>>
```
At least it doesn't crash (which I thought it would given that `type[0]` gets bumped up from `CT_STRING` into a non-existent type)...

mattdowle:
Closes #2265 
`test()` no longer relies on `option(warn=2)`.  Multiple warnings() are now tested, as needed for the new test for this issue.
`\n` in `output=` is now left in rather than stripped out.  Or `output=` can be a vector of lines, too.


st-pasha:
:+1: But I like how you expressed this in terms of just another re-read pass -- I thought it would be much harder than that...

mattdowle:
Yes exactly. If a quote rule type bump occurs there is a warning on that line always, so at least the user knows something might be improper in the file. Some refinement needed for sure but it's erring on the side of caution for now (always warns if any quote rule bump occurred).  Hm. That's not true if there's a one-row footer auto removed too, currently.  Will improve that ...

PR #2649, opened 2018-03-01
Fixed sep=NULL (i.e. readLines) when quoted fields are present, #2548

### LINKED ISSUE #2548: fread Internal error: first line has field count -1 but expecting 1
```
> fread('"A","B"\n', sep='')
Error in fread("\"A\",\"B\"\n", sep = "") : 
  Internal error: first line has field count -1 but expecting 1
```

mattdowle:
Closes #2548 
When `sep=NULL` it now avoids the `quoteRule` and `ncol` detection loop. Instead it just sets `quoteRule` to 3 (i.e. ignore quoting) and `ncol` to 1 directly, to be cleaner. The end of lines are taken literally just as `wc -l` would, even when quotes are present (which may be why the user desires to use sep=NULL and parse each line themselves.)  It was failing before because countfields() saw the quoted field and then expected to see the separator after the end of the quoted field, which is not the case when `sep=NULL`.


MichaelChirico:
@st-pasha add `?w=1` to the url to suppress whitespace changes:

https://github.com/Rdatatable/data.table/pull/2649/files?w=1

mattdowle:
I'm surprised `?w=1` dealt with the lines moving around as well indentation. But it did. Impressive!

MichaelChirico:
IIUC it's the same as the whitespace flag on command-line git

Issue #2586, opened 2018-01-23
na.strings is too literal when column is quoted on file


HughParsonage:
`#` `Min reprex`

``` r
na_string <- "x y z"
out <- fread('A,B,C\nD,"x y z",E', na.strings = na_string, verbose = TRUE, sep = ",")
anyNA(out$B)
#> [1] FALSE
### yet
na_string %chin% out$B
#> [1] TRUE
```

Using `na_string <- '"x y z"'` will get the right answer, but that was a bit difficult to deduce.

Output of verbose output:

```r
Input contains a \n or is "". Taking this to be text input (not a filename)
[01] Check arguments
  Using 12 threads (omp_get_max_threads()=12, nth=12)
  NAstrings = [<<"x y z">>]
  None of the NAstrings look like numbers.
  show progress = 1
  0/1 column will be read as boolean
[02] Opening the file
  `input` argument is provided rather than a file name, interpreting as raw text to read
[03] Detect and skip BOM
[04] Arrange mmap to be \0 terminated
  \n has been found in the data so any mixture of line endings is allowed other than \r-only line endings. This is common and ideal.
[05] Skipping initial rows if needed
  Positioned on line 1 starting: <<A,B,C>>
[06] Detect separator, quoting rule, and ncolumns
  Using supplied sep ','
  sep=','  with 2 lines of 3 fields using quote rule 0
  Detected 3 columns on line 1. This line is either column names or first data row. Line starts as: <<A,B,C>>
  Quote rule picked = 0
  fill=false and the most number of columns found is 3
[07] Detect column types, good nrow estimate and whether first row is column names
  Number of sampling jump points = 1 because (17 bytes from row 1 to eof) / (2 * 17 jump0size) == 0
  'header' determined to be true due to column 2 containing a string on row 1 and a lower type (bool8) on row 2
  Type codes (jump 000)    : A1A  Quote rule 0
  =====
  Sampled 2 rows (handled \n inside quoted fields) at 1 jump points
  Bytes from first data row on line 1 to the end of last row: 17
  Line length: mean=8.50 sd=3.54 min=6 max=11
  Estimated number of rows: 17 / 8.50 = 2
  Initial alloc = 2 rows (2 + 0%) using bytes/max(mean-2*sd,min) clamped between [1.1*estn, 2.0*estn]
  All rows were sampled since file is small so we know nrow=2 exactly
  =====
[08] Assign column names
[09] Apply user overrides on column types
[...]

PR #2652, opened 2018-03-02
na.strings getOption() added so ,, can be read as NA by default in future

### LINKED ISSUE #1159: fread reads in empty fields as logical NA
The following results in very unpleasant behavior if you read and subsequently write a csv:

``` R
fread('a,b\n1,')
#   a  b
#1: 1 NA
```

Second column is read in as a logical column, which means empty values get converted to NA. Those columns should instead be read in as character.

### LINKED ISSUE #2106: fwrite/fread single column input with NA -vs- empty lines
I create a toy example.

temp <- data.table(a=c(1,NA,2,3,999,NA))

I save it:
fwrite(temp, "temp.csv", quote=FALSE, sep=",", append=F)

and read it again:
my <- fread("temp.csv", stringsAsFactors=F)

As you can see only the first line is read.

I don't know if it's a problem with fread or with fwrite's output file.

### LINKED ISSUE #2214: Irreversible empty string handling by fread() and fwrite()
First off, thank you for this fantastic package. It effortlessly powers many of my data caving adventures.

Sometimes, it's necessary to distinguish between null (`NA`) and empty (`""`) strings, and I'm trying to establish a pipeline that preserves this distinction with minimal markup. This doesn't currently work. To work, `fread()` would need to distinguish between ` ` and `""`, and `fwrite()` ideally would quote empty strings when `quote = "auto"`.

Consider this data.table:

```{r}
dt <- data.table::data.table(chr = c(NA, "", "a"), num = c(NA, NA, 1))
```

Here is the fwrite output with `quote="auto"`:

```{r}
csv <- paste(
  capture.output(
    data.table::fwrite(dt, quote = "auto")
  ),
  collapse = "\n"
))
cat(csv)
```

> ```
> chr,num
> ,
> ,
> a,1
> ```

The empty string is not quoted, and thus indistinguishable from the null string. They are both read back in as empty strings:

```{r}
data.table::fread(csv)
```

> ```
>    chr num
> 1:      NA
> 2:      NA
> 3:   a   1
> ```

If instead we force quotes, the distinction is kept between the null and empty strings:

```{r}
csv_quoted <- paste(
  capture.output(
    data.table::fwrite(dt, quote = TRUE)
  ),
  collapse = "\n"
)
cat(csv_quoted)
```

> ```
> "chr","num"
> ,
> "",
> "a",1
> ```

However, there is no way to read them back in as such. Either they are both empty:

```{r}
data.table::fread(csv_quoted)
```

> ```
>    chr num
> 1:      NA
> 2:      NA
> 3:   a   1
> ```

Or both null:

```{r}
data.table::fread(csv_quoted, na.strings = "")
```

> ```
>    chr num
> 1:  NA  NA
> 2:  NA  NA
> 3:   a   1
> ```

### LINKED ISSUE #2217: fread should distinguish between quoted and unquoted empty strings
See #2214 

For a file like this:
    
    A,B
    1,"foo"
    2,
    3,""
 
the B column should be understood as `c("foo", NA, "")`, not `c("foo", "", "")` or `c("foo", NA, NA)`.

The proposed mechanism is the following: make 2 "string" types. The first type (default) reads empty fields as empty strings. However this type fails to recognize field `""` as valid input and instead type-bumps. The second type treats empty field as NA and `""` as an empty string.

### LINKED ISSUE #2281: fread/fwrite default options for NA mismatch
Hi,

first thank you for this package. :)
I had a problem when saving and reading a csv with fread and fwrite 
```
library(data.table)
dt <- data.table(a = c(NA_integer_, 1L, 2L), b = c("a", "b", NA_character_))
tmp <- tempfile()
fwrite(dt, tmp)
dt2 <- fread(tmp)
dt
#    a  b
#1: NA  a
#2:  1  b
#3:  2 NA
dt2
#    a b
#1: NA a
#2:  1 b
#3:  2  
all.equal(dt,dt2)
#[1] "Column 'b': 'is.NA' value mismatch: 0 in current 1 in target"
```

So I realised fread and fwrite have different default for na(-strings), so I checked read.csv and write.csv and they use the same default option.
In the end it mixes empty strings with NA which is not an expected default behavior imho.
Shouldn't fwrite use the same default option as write.csv i.e. na = "NA" or NA? so that a call to fread after fwrite would recover the same object?

Cheers,
Benoit

### LINKED ISSUE #2524: fread(fill=TRUE) fills character fields with empty strings instead of NAs
Test case:
```
> fread("A,B,C\n1,foo,bar\n2", fill=TRUE)
   A   B   C
1: 1 foo bar
2: 2        
```
Expected output:
```
> data.table(A=c(1L, 2L), B=c("foo", NA_character_), C=c("bar", NA_character_))
   A   B   C
1: 1 foo bar
2: 2  NA  NA
```

mattdowle:
Closes #2106 (again)
Closes #2217 
Closes #2214
Closes #2281 
Closes #1159

#2524 is related and in discussion as to whether filled character columns should have NA always independently of `na.strings`. Can address that separately to this PR.

Standardizing `fread`'s default :  `,,` means `NA` for all types consistently (in particular in string columns).  `,"",` means empty string as written by `fwrite` by default (change made in dev some months back).
See also comment in reopened issue [here](https://github.com/Rdatatable/data.table/issues/2106#issuecomment-370061249) that this PR reverts the `fwrite` change in dev for 1-column DTs back to the same consistent default in v1.10.4 as on CRAN.

For all input data (i.e. all types, NA or "", 1 column or >1 column),  `fread(fwrite(DT)) == DT`  should be true without needing to change any arguments. This is not true before this PR.

TODO:
- [ ] ~allow quoted na.strings as per #2586 and change doc again.~ Left open issue in this milestone to address separately to this PR. This PR is just about `,,` -vs- `,"",` default.
- [x] display NA in character columns as `<NA>` just like base R to distinguish from `""` and `"NA"`


mattdowle:
For booleans both `,,` and `,NA,` are read as `NA`, regardless of the `na.strings=` control.  Since write.csv writes logical as `TRUE` | `FALSE` | `NA`.  Unlike numeric columns where it writes `NA` as `,,`.

Good points on breakage.  Usual `option()` could be provided to return the old behaviour with notice at the top of NEWS about potential breaking change.  The next version is quite a major change to `fread` (e.g. `logical01`too) so if it goes ahead, now seems like as good a time as any, for v1.11.0.

I'll sweep through all fread issues and see if any others are in this area.

More can be done to auto detect files which have used `"NA"` for NA in character columns.  Maybe `na.strings=` could be `__auto__` like `skip=` is now.

Views welcome from others and we'll keep this PR open a while.  So far I've tried to fix the issues linked at the top.

mattdowle:
readr's defaults seem to be NA-loving. It doesn't seem to mind about distinguishing at all.
screenshot from 2018-03-02 19-52-14 resized

HughParsonage:
There's `read_csv(txt, quoted_na = T/F)` though.

mattdowle:
I saw `quoted_na=`.  It's about choice of defaults.  I'm finding the choice of `quoted_na = TRUE` by default to be odd, since quoted strings are primarily how most software distinguish the literal from the NA-meaning string.  Same area as your #2586.

mattdowle:
Ok, and now the new `yaml` tests need wrapping in a branch `if ("yaml" %in% search()) ...`   Have a look at the other tests of any of the Suggests packages for the exact condition.   On AppVeyor we don't run Suggests tests just to speed it up.  It's just on Travis that all Suggests tests run too.   We figure that in dev it's enough to test Suggests on one platform.  It's nice for AppVeyor to return results quickly as it's Windows we're most interesting in after push, since we've dev'd and tested on Linux.  In pre-release checks to CRAN, Suggests tests then run on win-builder.

NEWS.md
```diff
@@ -3,6 +3,18 @@
...
+
+2. `fread` now reads a column of all 0's and 1's as `logical` rather than `integer`, for convenience to avoid needing to change the type afterwards or use `colClasses`. The old behaviour can be restored with `options(datatable.logical01=FALSE)`. We felt this default change was ok to make because in all operations there should be no difference: R treats `logical` and `integer` the same. If this change does cause a problem, the option is provided to restore old behaviour while you update your code. Similarly, `fwrite` now writes `logical` columns as `0/1` by default, controlled by the same option. `0/1` is smaller and faster than `"TRUE"/"FALSE"`, which can make a significant difference to space and time the more `logical` columns there are. Further, a column of `TRUE/FALSE`s is ok, as well as `True/False`s and `true/false`s, but mixing styles (e.g. `TRUE/false`) is not and will be read as type `character`.
```
MichaelChirico: I would be a bit more careful on the wording:

> in all operations there should be no difference: R treats `logical` and `integer` the same.

But that's not true:

```
DT = data.table(l_int = c(0, 0, 1, 0), l_log = c(FALSE, FALSE, TRUE, FALSE), i = 1:4)
DT[(l_int)]
#    l_int l_log i
# 1:     0 FALSE 1
DT[(l_log)]
#    l_int l_log i
# 1:     1  TRUE 3
```

I think (?) more accurate is that all arithmetic expecting `integer` and getting `logical` will go through as expected. Of course any function running an `is.integer` test will fail (and _vice versa_ for `is.logical` on `integer` columns).

NEWS.md
```diff
@@ -157,6 +167,8 @@ the behaviour of `base:::merge.data.frame()`. Thanks to @sritchie73 for reportin
 
 35. `CJ()` now fails with proper error message when results would exceed max integer, [#2636](https://github.com/Rdatatable/data.table/issues/2636).
 
+36. `NA` in character columns now display as `<NA>` just like base R to distinguish from `""` and `"NA"`.
```
MichaelChirico: this is nice, no need for `quote = TRUE` argument by default 👍

NEWS.md
jangorecki: I had the same concern. I think as long as we add option for 1.10.6 and change its default from 1.10.8 will be fine.

jangorecki:
If we do 1/0 instead of TRUE/FALSE we could also make #1656.

HughParsonage:
@MichaelChirico For me it was: from this page, select the `Files changed` tab, click the friendly green button `Review`, then the `Approve` radio button.

MichaelChirico:
@HughParsonage thanks, I thought I remembered it on this (Conversation) tab

PR #2656, opened 2018-03-04
csvy reading capabilities

### LINKED ISSUE #1701: Add CSVY support for fread() and fwrite()
The [rio](https://github.com/leeper/rio) already [allows](https://github.com/leeper/rio/blob/e1365beec61ed79ebdf2e911f4b0375b5a1e48e8/R/import_methods.R#L160) reading of [csvy](http://csvy.org/) formats by relying on the helper package (same author) [csvy](https://github.com/leeper/csvy/blob/master/R/read_csvy.R#L17-L88). It'd be great to leverage its features to read/write.

MichaelChirico:
Closes #1701 

adding @HughParsonage to review in anticipation of interactions with his new paradigm for handling `colClasses`.

- it might be nice to have `yaml = grepl('\\.csvy$', input)` but I'm wary of using `grepl` since #2531. there should be a way to set `has_yaml = 'auto'` and deal with it like that but I haven't put too much thought into it.

- could also consider letting `yaml` potential give the location of a separate file which contains the metadata

- should spend some time ironing this into the standard laid out [here](https://frictionlessdata.io/specs/table-schema/) to the extent reasonable, though some things should be ignored (e.g., standard calls for default column type to be `string`, but `fread` is capable of being smarter than that)

suggestions welcome.


R/fread.R
```diff
@@ -79,8 +73,144 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
...
+  if (skip=="__auto__") {
+    if (has_yaml) {
+      skip=0L
+    } else {
+      skip=-1L   # skip="string" so long as "string" is not "__auto__". Best conveys to user something is automatic there (than -1 or NA).
+    }
+  }
   if (is.double(skip)) skip = as.integer(skip)
+  if (has_yaml) {
+    if (!require('yaml'))
```
HughParsonage: `requireNamespace('yaml', quietly = TRUE)` I think is preferred

HughParsonage:
Thanks. Happy to review, but I'll defer till PR #2545 is resolved.

R/fread.R
MichaelChirico: Yes indeed, travis gave me the same warning

MichaelChirico:
hmm, seems appveyor didn't figure out how to add `yaml` newly added to `Suggests` (though Travis did), any suggestion?

mattdowle:
@MichaelChirico  At the top of `tests.Rraw`, `yaml` needs to be added to line 4 there too.   I can't quite remember the reason the test suite has this extra step, but it's probably to do with allowing the compatibility tests (i.e. tests with Suggests packages) to be optional in dev where typically I skip all those to speed up the dev cycle.

mattdowle:
@MichaelChirico Ok, and now the new yaml tests need wrapping in a branch if ("yaml" %in% search()) ... Have a look at the other tests of any of the Suggests packages for the exact condition. On AppVeyor we don't run Suggests tests just to speed it up. It saves installing the (sometimes large) Suggests packages and all their dependencies in the VM image. It's just on Travis that all Suggests tests run too. We figure that in dev it's enough to test Suggests on one platform. It's nice for AppVeyor to return results quickly as it's Windows we're most interesting in after push, since we've dev'd and tested on Linux. In pre-release checks to CRAN, Suggests tests then run on win-builder.

MichaelChirico:
OK, will do.

I'm on a holding pattern for this for now, as there was a recent change to the csvy standard and now some back-and-forth on perhaps updating it again:

https://github.com/leeper/csvy/issues/13

jangorecki:
You can find CI job which runs all Suggests on Linux. Just push branch here and read only gitlab mirror will run Ci for it: https://gitlab.com/Rdatatable/data.table/pipelines

MichaelChirico:
FWIW, `yaml` is a dependency of `knitr`, which is already in our Suggests, so I don't think adding `yaml` to Suggests increases our total dependency burden

jangorecki:
Does it support to provide factor levels in yaml? If so please add simple tests for that. If not could we get that in scope or as new follow-up issue?

MichaelChirico:
@jangorecki I don't see anything in the official spec re:factors, so I guess we'll be the market leaders there.

@mattdowle With that last commit, `fread` side is done; we can merge whenever (not sure if you want to wait for 1.12.3)

I think I know how I'll do the `fwrite` side, but probably best to save for another PR.

MichaelChirico:
@mattdowle tried with using `missing()` instead of `match.call()`, but by the time we reach the `yaml` section some of the arguments have already been populated (and are no longer `missing()`), whereas the actual `call` returned by `match.call()` is more static.

Could add something at the head of `fread` along the lines of `if (yaml) { missing_sep = missing(sep); missing_na.strings = missing(na.strings); ...}` but I like the `match.call()` approach better... reverting

R/fread.R
```diff
@@ -201,7 +203,7 @@ fread <- function(input="",file=NULL,text=NULL,cmd=NULL,sep="auto",sep2="auto",d
                      brackify(yaml_names), '\n')
     # process header first since it impacts how to handle colClasses
     if ('header' %chin% yaml_names) {
-      if (!missing(header)) warning("User-supplied 'header' will override that found in metadata.")
+      if ('header' %chin% call_args) warning("User-supplied 'header' will override that found in metadata.")
```
jangorecki: I would prefer `message` instead of warning. If it would be other way around, yaml overriding user provided arguments, then warning is good.

MichaelChirico: 👌 makes sense; done

R/fread.R
```diff
@@ -142,6 +148,136 @@ fread <- function(input="",file=NULL,text=NULL,cmd=NULL,sep="auto",sep2="auto",d
...
+           'unskipped line (', 1L+skip, '), which does not ',
+           'constitute the start to a valid YAML header ',
+           '(expecting something matching regex "', yaml_border_re,
+           '"); please check your input and try again.')
+    }
+
+    yaml_comment_re = '^#'
+    yaml_string = character(0L)
+    while (TRUE) {
+      this_line = readLines(f, n=1L)
```
jangorecki: what if there is only yaml header in the file and it is not completed by `----`? this line can thrown confusing warning on end of file with no new line

R/fread.R
```diff
@@ -142,6 +148,136 @@ fread <- function(input="",file=NULL,text=NULL,cmd=NULL,sep="auto",sep2="auto",d
...
+    }
+    if ('schema' %chin% yaml_names) {
+      new_types = sapply(yaml_header$schema$fields, `[[`, 'type')
+      if (any(null_idx <- sapply(new_types, is.null)))
+        new_types = do.call(c, new_types)
+      synonms = rbindlist(list(
+        character = list(syn = c('character', 'string')),
+        integer = list(syn = c('integer', 'int')),
+        numeric = list(syn = c('numeric', 'number', 'double'))
+      ), idcol = 'r_type')
```
jangorecki: what about `factor = list(syn=c("factor","categorical"))` and `integer64`?

R/fread.R
MichaelChirico: See `inst/tests/csvy/test_incomplete_header.csvy` and test `1985.16`

R/fread.R
MichaelChirico: Sure, missed those. Can keep an eye out for #2545 to expand it a bit more too.

MichaelChirico:
So much for trying `git pull origin master --rebase`....

jangorecki:
@MichaelChirico another good option instead of `rebase` is `git reset --soft <last common sha to master>` then `git stash`, `git pull upstream master`, `git stash pop`, resolve conflicts

MichaelChirico:
@jangorecki thanks, seems I still couldn't get around a `--force` though. oh well

appveyor.yml
```diff
@@ -43,6 +43,7 @@ build_script:
   - travis-tool.sh r_install xts
   - travis-tool.sh r_install nanotime
   - travis-tool.sh r_install R.utils
+  - travis-tool.sh r_install yaml
```
MichaelChirico: @jangorecki do we need to edit anything in to the GitLab CI for this? Or it's covered by the `knitr` dependency?

jangorecki: it is enough to defined it in DESCRIPTION file

DESCRIPTION
```diff
@@ -43,6 +43,7 @@ Authors@R: c(
 Depends: R (>= 3.1.0)
 Imports: methods
 Suggests: bit64, curl, R.utils, knitr, xts, nanotime, zoo, yaml
+SystemRequirements: zlib
```
jangorecki: Does it mean that `zlib` has to be present to install data.table, even if one is not going to use compression? I hope not

mattdowle: zlib necessary to compile. Seems like it's standard though. Windows and Mac users installing binaries from CRAN won't need to install anything special, afaik.
If there's a way to make zlib optional, would be ideal of course. But there's plenty of other higher priority things to do first. I'd wait to see if anyone reports difficulties due to missing zlib first.

mattdowle: zlib necessary to compile. Seems like it's standard though. Windows and Mac users installing binaries from CRAN won't need to install anything special, afaik.
If there's a way to make zlib optional, would be ideal of course. But there's plenty of other higher priority things to do first. I'd wait to see if anyone reports difficulties due to missing zlib first.

PR #2659, opened 2018-03-06
Remove internal R v3.1.0 switches

### LINKED ISSUE #2498: Remove internal 3.1.0 switches now dependency is R 3.1.0

mattdowle:
Closes #2498 

There were 3 used, all FALSE since R 3.1.0: 
```
.R.listCopiesNamed
.R.assignNamesCopiesAll
.R.subassignCopiesOthers
```

and 1 unused (TRUE) :  
```
.R.subassignCopiesVecsxp
```

These are all still checked, but now `stop()` in `.onLoad` so we know in future if anything changes in R-devel.

### DIFF
--- a/CRAN_Release.cmd
+++ b/CRAN_Release.cmd
@@ -156,7 +156,7 @@ vi ~/.R/Makevars  # make the -O3 line active again
 ###############################################
 
 cd ~/build
-wget -N ftp://ftp.stat.math.ethz.ch/pub/Software/R/R-devel.tar.gz
+wget -N https://stat.ethz.ch/R/daily/R-devel.tar.gz
 rm -rf R-devel
 tar xvf R-devel.tar.gz
 cd R-devel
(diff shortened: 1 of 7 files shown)


mattdowle:
Grrr. This was supposed to be a quick PR to get something out of the way. It seems AppVeyor is now running on 32bit Windows even though ARCH=x64.  That has revealed an unrelated bug due to using `size_t` in fread.c.  Fixing in this PR at the same time.  Perhaps we can add 32bit AppVeyor test as well as the existing two 64bit ones, @jangorecki (I'm not sure how to do that)?

mattdowle:
Wow. It is not my day.  Now there's a R-devel change today, too? ...

mattdowle:
It passes locally for me using latest R-devel r74362. AppVeyor has now updated exact same R-devel version.  Now I see that AppVeyor is passing fine in tests_x64 and just failing in test_i386.
Looks like an outstanding R-devel 32-bit-only problem.  I'll let them know.  In the meantime, no PRs will pass and if I do merge anyway, master will then be in fail state.

mattdowle:
At last!  @st-pasha please have a look and approve the `fread.c` changes which I did incidentally in this PR : 
- size_t => [u]int64_t to work on 32bit.  Since size_t is smaller there, and we use nrowLimit==INT64_MAX to signal unlimited nrows, for example.  I went for (signed) int64_t for things like `DTi`, `allocnrow` for consistency with `fread.h` which uses signed int64_t,  but I don't mind either way.
- a few memory faults (e.g. in NextGoodLine) picked up by ASAN.

No need to look at the R-side stuff.

PR #2671, opened 2018-03-13
Prune long tests

### LINKED ISSUE #2105: test.data.table() error when optional package reshape is not installed
#Hello.

I've just used test.data.table() to ckeck my data.table  1.10.5 installation and got this errors:

> 
> Running test id 168     Don't know how to automatically pick scale for object of type ITime. Defaulting to continuous.
> Test 168 Error in seq.int(0, to0 - from, by) : 'to' cannot be NA, NaN or infinite
> Running test id 168.1     Test 168.1 Error in seq.int(0, to0 - from, by) : 'to' cannot be NA, NaN or infinite
> Running test id 486     Tests 487 and 488 not run. If required call library(reshape) first.
> Running test id 1383.2     Tests 1383.3 not run. If required call library(reshape) first.
> Error in eval(expr, envir, enclos) : 
>   2 errors out of 5737 (lastID=1759, endian==little, sizeof(long double)==16, sizeof(pointer)==8) in inst/tests/tests.Rraw on Fri Apr 07 14:17:57 2017. Search tests.Rraw for test numbers: 168, 168.1.

I have reshape2 installed instead of reshape, maybe the test should move to the newer packages.

mattdowle:
Closes #2105

Primary driver of this is win-builder.  As is desirable and great, it checks with `--as-cran` and checks all our Suggests. It was taking over 20 mins (similar to Travis) and sometimes halting; i.e. log file stops early with no error message. I guess this is a resource limit being reached or something similar to 'oom killer' on Linux.  Which is reasonable.

Added a report of the 10 longest tests, aggregated to the `integer` test ID level.

Moved heavy packages in Suggests into a section marked for moving to a separate package.  Also, I know some people think data.table needs the packages in Suggests to operate.  So it would be good to simplify Suggests to packages we really do suggest,  such as `bit64`.  The dependency chain is thoroughly checked by `--as-cran`,  even within Suggests, so it would be good to move compatibility tests with heavy packages out.

Wrote to Uwe and he suggested that `--as-cran` checks all URLs are alive, too. If one is slow or unreliable,  `R CMD check --as-cran` can hang on that.  I modified R-devel to time that stage and report the URL which took the longest to contact.  Indeed, this helped me find some unreliable ones (even locally) which I removed in this PR.  I've submitted the R-devel patch to Uwe and offered to refine it.

```
$ diff R-devel/src/library/tools/R/QC.R ../R-devel/src/library/tools/R/QC.R
7050c7050,7051
<         bad <- tryCatch(check_url_db(url_db_from_package_sources(dir),
---
>         started <- proc.time()[3L]
>         bad <- tryCatch(check_url_db(urldb<-url_db_from_package_sources(dir),
7052a7054,7055
>         timetaken <- as.integer(proc.time()[3L] - started)
>         if (timetaken>10) cat("  Contacting",NROW(urldb),"URLs for a response (including all references in .Rd files) took", timetaken, "seconds\n")

$ diff R-devel/src/library/tools/R/urltools.R ../R-devel/src/library/tools/R/urltools.R
80c80
< function(f)    
---
> function(f)
116c116
<     if(is.null(files)) 
---
>     if(is.null(files))
399a400,401
>     slowestURLTime <- 0.0
>     slowestURL <- ""
401a404
>         tt <- proc.time()[3L]
402a406,407
>         took <- proc.time()[3L] - tt
>         if (took > slowestURLTime) { slowestURLTime <<- took; slowestURL <<- u; }
575a581,583
>         } else if (slowestURLTime > 5) {
>             # All were fine but sometimes one pesky one can take a long time to respond
>             cat("\n  Longest time to check URL alive was", as.integer(slowestURLTime), "seconds for", slowestURL, "\n")
```

Locally, from this branch, I now see (using modified R-devel) : 
```Contacting 60 URLs for a response (including all references in .Rd files) took 27 seconds```

`--as-cran` still takes 8min17s for me at the `checking CRAN incoming feasibility ...` stage.  So the 27 seconds for testing URLs is a small part of that.  There isn't any installing happening in this step.  This timing (extra ~8 mins) is consistent with the extra time on win-builder and on Travis. Continuing to investigate as it doesn't make sense ...

### CODE REVIEW

**Comment by MichaelChirico** [comment]
We're free to add our own field to `DESCRIPTION` right? Could add something like `Supplementary` to store those axed from Suggetss

**Comment by mattdowle** [comment]
Yes we could add our field.  But there's only any point in mentioning `ggplot2`, for example, if we're testing compatibility with it (which is why it's in Suggest now).  Otherwise why would `ggplot2` be `Supplementary`?

**Comment by jangorecki** [comment]
We cannot add new field to DESCRIPTION as it will raise a NOTE. It would be nice to simplify suggests. I don't think as-cran will check Rraw so we should be able to add a switch, or some environment variable which makes test.data.table run heavy suggests, which by default will be FALSE. Other option is to move those tests into separate integration.Rraw, test.data.table already support multiple Rraw.

**Comment by mattdowle** [comment]
I got to the bottom of why `--as-cran` takes so long. And no-surprise, it's a surprise.
Luckily,  R-release exhibited the 8min slowdown but my modified R-devel did not.  I debug()'d the difference.
Consider in base R : 
```
> tools:::url_db_from_package_sources
function (dir, add = FALSE) 
{
    meta <- .read_description(file.path(dir, "DESCRIPTION"))
    db <- rbind(url_db_from_package_metadata(meta), url_db_from_package_Rd_db(Rd_db(dir = dir)), 
        url_db_from_package_citation(dir, meta), url_db_from_package_news(dir))
    if (requireNamespace("xml2", quietly = TRUE)) {
        db <- rbind(db, url_db_from_package_HTML_files(dir), 
            url_db_from_package_README_md(dir), url_db_from_package_NEWS_md(dir))
    }
    if (add) 
        db$Parent <- file.path(basename(dir), db$Parent)
    db
}
```

On win-builder and in my R-release locally, xml2 is installed.  This adds 644 URLs from our NEWS.md to the 60 in *.Rd.  At a rough rate of 0.5s per URL, that's over 5 extra minutes.  If some of those 644 URLs are slow or unreliable, then it's easily 8 mins or longer. Certainly it explains all the variance. Especially on win-builder and CRAN where multiple packages are checked in parallel.

On my local R-devel,  I did not install xml2.  So the branch above is skipped and "just" 60 URLs are contacted.

**Comment by jangorecki** [comment]
As NEWS.md is only growing it would be best to be able to disable (not only by removing pkg) check of that file. Maybe you can mention that in your R-devel email.

**Comment by mattdowle** [comment]
Thanks. I suggested that to them.  In the meantime, perhaps we could move older news into NEWS.0.md like base R does.  It would save render time on GitHub, too (we want to encourage people to click it and read live NEWS).  IIRC from looking at the sources, just NEWS.md is checked.  It seems there is a `news()` function in R, so maybe there is even a proper way of doing the postfix ".0".

**Comment by mattdowle** [comment]
I've created old NEWS.0.md now and win-builder is down from 20m to 10m, so that's good.

However, still not passing on win-builder : 
```
... snip
  Running test id 1645.012      
  Running test id 1645.013      
  Running test id 1645.014      Error: memory exhausted (limit reached?)
```
We have to pass win-builder to make it to CRAN.  More to do to prune test 1645 (non-equi).  So

**Comment by mattdowle** [comment]
@jangorecki  You're right. I removed a random package from Suggests and did `remove.packages()` on it too and `R CMD check --as-cran` still passes.  So, as you thought, it doesn't seem to reach inside tests.Rraw.   This is great.  So yes, separate new file (not new package) which is not run by anything other than by us and by users manually; e.g. `test.data.table( with.other.packages=TRUE )`.  Now that we know the super-long times of `--as-cran` are actually because of the URL pinging,  the driver for this is minimizing the package Suggests so it lasts longer on CRAN in OK state without other package changes breaking it for unimportant reasons (like ggplot2 changing the `names` of its results, which should not break our test).

**Comment by jangorecki** [comment]
Yes, better separate Rraw than new package, as long as we don't have to.
Ideally we could keep in Suggests only packages which we use inside data.table functions, not only tests. So at least bit64, xts, zoo, curl, maybe something more. Then we still need to maintain a list of extended Suggests somewhere on side so we can easily automate installation of all extended suggests.

**Comment by HughParsonage** [comment]
Would it be worth requesting R-core to amend the quality control to skip URLs of the form `https://github.com/<repo>/<package>/(issues|pull)/[0-9]+` provided `https://github.com/<repo>/<package>` works? I imagine that there are thousands of such checks on CRAN. While it's possible to flub the issue number in the link, I question whether it's worth the computer time checking whether it works. (And as annoying as broken links are, I think a more common mistake would be copying the wrong issue number than choosing a wrong number, which the QC wouldn't pick up.) 

I've got no skin in the game, but I think lowering the standards a bit to save check time is better than partitioning `NEWS`'s.

**Comment by mattdowle** [comment]
Great idea @HughParsonage.  Yes -- please go ahead and suggest it.  Even better,  you could submit a patch to R-devel.  If you just suggest it, I doubt it will go anywhere just because someone actually needs to do it and it isn't precisely clear what exactly it is you're suggesting.  Would there be a whitelist?  A whitelist prefix?  Would the whitelist be per-package or globally for R?  Would it controlled by an option?  etc etc etc.  However, if you do it, test it and submit an actual patch, it is much more likely to be accepted.  @jangorecki  did this recently, his patch to R is accepted and he's thanked in R's NEWS for it.   The code that checks packages is R code in tools packages, so it's easy to work with and test. No compiling of R-devel needed for example.
In the meantime, release to CRAN is pressing as we're in ERROR state for a long time now and this is impacting others, so in our case right now, partitioning NEWS and moving on, I had to do.

**Comment by jangorecki** [comment]
R 3.1.0 job started to fail after this commit with:
```
  Running ‘main.R’ [105s/103s]
 ERROR
Running the tests in ‘tests/main.R’ failed.
Last 13 lines of output:
   3: 1648  9.006    91
   4: 1509  6.496     1
   5: 1646  5.426    91
   6: 1779  5.231    13
   7: 1644  5.123    91
   8: 1642  4.722    91
   9: 1437  2.935   528
  10: 1223  2.176   728
  
  Error in eval(expr, envir, enclos) : 
    1 error out of 7321 in 00:01:42 on Thu Mar 15 20:32:31 2018. [endian==little, sizeof(long double)==16, sizeof(pointer)==8]. Search inst/tests/tests.Rraw for test number 167.2.
  Calls: test.data.table -> sys.source -> eval -> eval
```

**Comment by HughParsonage** [comment]
Hey @mattdowle  (or anyone), would you mind reviewing this patch before I send it off to R-dev? (Never written a patch before.)

```
From 78fc0eff94e73498077d1c6f6fdfe6583a89deb1 Mon Sep 17 00:00:00 2001
From: HughParsonage <hugh.parsonage@gmail.com>
Date: Sat, 17 Mar 2018 18:45:53 +1100
Subject: [PATCH] Exclude http checks of likely GitHub URLs

---
 src/library/tools/R/QC.R | 31 ++++++++++++++++++++++++++++++-
 1 file changed, 30 insertions(+), 1 deletion(-)

diff --git a/src/library/tools/R/QC.R b/src/library/tools/R/QC.R
index 89b73fa..95efbee 100644
--- a/src/library/tools/R/QC.R
+++ b/src/library/tools/R/QC.R
@@ -7046,7 +7046,36 @@ function(dir, localOnly = FALSE)
     if(!capabilities("libcurl") && !localOnly)
         out$no_url_checks <- TRUE
     else {
-        bad <- tryCatch(check_url_db(url_db_from_package_sources(dir),
+        url_db_to_check <- url_db_from_package_sources(dir)
+        .excl_likely_GitHub <- function(url_db) {
+          # Do not include URLs which are known to be correct, given
+          # the package and GitHub's structure, viz.
+          # https://github.com/<??>/<package>/issues
+          # https://github.com/<??>/<package>/issues/<number>
+          # https://github.com/<??>/<package>/pull/<number>
+          # Approximately 10,000 (Mar 2018) out of 70,000.
+
+          # That is, exclude URLs that:
+          # 1. are from github.com,
+          # 2. contain 'issues' or 'pull', and
+          # 3. have a duplicated repo (need to check each repo once, but only once)
+          urls <- url_db[["URL"]]
+          GitHub_urls <- startsWith(urls, "https://github.com")
+          if (any(GitHub_urls)) {
+            issues_or_pulls <- grepl("/(issues|pull)(/[0-9]+/?)?", urls)
+
+            GitHub_repo <- sub("/(issues|pull)(/[0-9]+/?)?",
+                               "",
+                               urls)
+
+            url_db[!(GitHub_urls & issues_or_pulls & duplicated(GitHub_repo)), ]
+          } else {
+            url_db
+          }
+        }
+        url_db_to_check <- .excl_likely_GitHub(url_db_to_check)
+
+        bad <- tryCatch(check_url_db(url_db_to_check,
                                      remote = !localOnly),
                         error = identity)
         if(inherits(bad, "error")) {
-- 
2.11.0.windows.3
```

**Comment by MichaelChirico** [comment]
maybe more specific I the title that you're targeting issue/PR references

**Comment by jangorecki** [comment]
Can we also remove GenomicRanges from Suggests? It is pretty heavy, some of recursive dep is 17+MB
```
Installing package(s) ‘GenomicRanges’
also installing the dependencies ‘bitops’, ‘RCurl’, ‘GenomeInfoDbData’, ‘zlibbioc’, ‘BiocGenerics’, ‘S4Vectors’, ‘IRanges’, ‘GenomeInfoDb’, ‘XVector’
```

PR #2675, opened 2018-03-15
prevent the utf8 string from being collected by the garbage collector in forder()

### LINKED ISSUE #2674: data.table breaks when there're millions of Chinese characters
related to #2566 

I find a serious issue that when there're millions of non-ASCII characters encoded in non-UTF8 encoding (see the following example), `setkey()`  on that column causes `data.table` becomes extremely slow and throws the error that `'translateCharUTF8' must be called on a CHARSXP` in the end.

**After that error, all the `data.table` function calls end up with another error that `Internal error: savetl_init checks failed`.**

 I will investigate and report more details later. Hopefully, I can file a PR to fix this.

image

# Example

NOTE, you have to execute this on __a windows machine with GB2312 as the default encoding (i.e., a Simplified Chinese Windows Machine)__. Otherwise it won't work. Also, if it won't fail for the first time, try to execute twice. I've tried this on several machines in my office. I'm quite confident it's reproducible.

```r
library(data.table)
dt <- data.table(
  x = sample(
    c("公允价值变动损益", "红利收入", "价差收入", "其他业务支出", "资产减值损失"),
    1e7,
    TRUE
  ),
  z = 1
)
setkey(dt, x)
```

# session info

```
R version 3.4.3 (2017-11-30)

Matrix products: default

locale:
[1] LC_COLLATE=Chinese (Simplified)_People's Republic of China.936 
[2] LC_CTYPE=Chinese (Simplified)_People's Republic of China.936   
[3] LC_MONETARY=Chinese (Simplified)_People's Republic of China.936
[4] LC_NUMERIC=C                                                   
[5] LC_TIME=Chinese (Simplified)_People's Republic of China.936    

attached base packages:

other attached packages:
[1] data.table_1.10.5

 [1] digest_0.6.15         crayon_1.3.4          withr_2.1.1           rprojroot_1.3-2       assertthat_0.2.0     
 [6] R6_2.2.2              backports_1.1.2       magrittr_1.5          rlang_0.2.0           cli_1.0.0            
[11] rstudioapi_0.7.0-9000 testthat_1.0.2.9000   devtools_1.13.3.9000  desc_1.1.1            tools_3.4.3          
[16] pkgload_0.0.0.9000    yaml_2.1.16           compiler_3.4.3        pkgbuild_0.0.0.9000   memoise_1.1.0        
[21] usethis_1.3.0   
```

# UPDATES

- I'm quite confident that `ENC2UTF8` is very slow for millions of chars now but still not sure if the issue is caused by this or not. Moreover, it's hard for me to understand why it's slow because it seems like R itself implements `enc2utf8` [in a similar way](https://github.com/wch/r-source/blob/44d54d6f848468a7353d99cc9be0255105185975/src/main/util.c#L1837)  **EDIT:** `enc2utf8()` takes a long time (17s) to convert 1e7 chars, too. So this is not the issue.
- I doubt it's related to `savetl()`, `savetl_end()`. I'm not familiar with how the global string pool works in R. However, I doubt that the utf-8 char created by `data.table` gets released when `gc()` causes (that's why it only occurs when the number of chars is large). If the char gets released and `savetl_end()` tries to modify a non-existed char's `truelength`...
- Should be related to GC, SEXP, may need PROTEC ... Basically confirmed...

shrektan:
closes #2674 

If somehow the garbage collector was triggered during sorting (like there're millions of non-ASCII characters), it leads to the collapse of `data.table` (see #2674 for details) **because it assumes there're converted UTF-8 chars in the global string pool.** This PR tries to fix this issue. 

In addition, it brings performance enhancement in the case of millions non-ASCII chars, because now it only needs to be converted to UTF-8 once. **Before this PR, the strings need to be converted twice in ` csort_pre()` and `csort()` respectively**, which may be a big issue for a large character vector (for example, on my computer, `enc2utf8()` takes about 20s for a 1e7 length Chinese character).

This PR is unfinished because it needs:

- [ ] a new test
- [ ] entries in NEWS.md

I will add them later if I can be sure that there's no need for further modification on 

https://github.com/Rdatatable/data.table/blob/master/src/forder.c#L1227

or any other places.

### DIFF
--- a/src/forder.c
+++ b/src/forder.c
@@ -866,7 +866,7 @@ static void csort(SEXP *x, int *o, int n)
   /* can't use otmp, since iradix might be called here and that uses otmp (and xtmp).
      alloc_csort_otmp(n) is called from forder for either n=nrow if 1st column,
      or n=maxgrpn if onwards columns */
-  for(i=0; i<n; i++) csort_otmp[i] = (x[i] == NA_STRING) ? NA_INTEGER : -TRUELENGTH(ENC2UTF8(x[i]));
+  for(i=0; i<n; i++) csort_otmp[i] = (x[i] == NA_STRING) ? NA_INTEGER : -TRUELENGTH(x[i]);
   if (nalast == 0 && n == 2) {                        // special case for nalast==0. n==1 is handled inside forder. at least 1 will be NA here
     if (o[0] == -1) for (i=0; i<n; i++) o[i] = i+1;    // else use o from caller directly (not 1st column)
     for (int i=0; i<n; i++) if (csort_otmp[i] == NA_INTEGER) o[i] = 0;
@@ -899,7 +899,7 @@ static void csort_pre(SEXP *x, int n)
   // savetl_init() is called once at the start of forder
   old_un = ustr_n;
   for(i=0; i<n; i++) {
-    s = ENC2UTF8(x[i]);
+    s = x[i];
     if (TRUELENGTH(s)<0) continue;   // this case first as it's the most frequent. Already in ustr, this negative is its ordering.
     if (TRUELENGTH(s)>0) {  // Save any of R's own usage of tl (assumed positive, so we can both count and save in one scan), to restore
       savetl(s);          // afterwards. From R 2.14.0, tl is initialized to 0, prior to that it was random so this step saved too much.
@@ -1087,8 +1087,8 @@ SEXP forder(SEXP DT, SEXP by, SEXP retGrp, SEXP sortStrArg, SEXP orderArg, SEXP 
 {
   int i, j, k, grp, ngrp, tmp, *osub, thisgrpn, n, col;
   Rboolean isSorted = TRUE;
-  SEXP x, class;
-  void *xd;
+  SEXP x, ux, class;
+  void *xd, *uxd;
 #ifdef TIMING_ON
   memset(tblock, 0, NBLOCK*sizeof(clock_t));
   memset(nblock, 0, NBLOCK*sizeof(int));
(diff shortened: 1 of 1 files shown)


mattdowle:
Oh. Wow. Really great find. Did it take you long to find it?!

The question now is what about R itself:  https://github.com/wch/r-source/blob/trunk/src/main/radixsort.c#L1133.   It doesn't use ENC2UTF8 there, so my first guess is that R itself is ok.  Our code was ported to R before ENC2UTF8 was added to data.table in dev, I guess.

Looking at the source tar.gz of data.table 1.10.4-3 on CRAN,  I do see ENC2UTF8 but only used in StrCmp and  StrCmp2,  and its result is not saved.  It's only used in the way you've pointed out, in data.table dev.

So is this a data.table-1.10.5 only problem do you think?  Have you seen it in R itself without data.table loaded,  or in data.table_1.10.4-3 or earlier?

mattdowle:
I was going to add the test and NEWS item, as you've done the hard work.  But it's a branch in your repo so I don't think I can, easily at least.  You're a project member and trusted, so please create branches directly in the main project then we can work on the same branch together.

Is there any slow down when all the strings are ASCII?  The added `allocVector` seems to be populated even when all strings are ASCII.

shrektan:
I forgot the fact that I have the right to open a new branch here. I've created a new PR #2678 based on a branch in the main project so that you can edit it easily. For further comments please see #2678.

MichaelChirico:
just throwing it out there -- I've had a problem I can't reproduce where
some Chinese characters are considered distinct by by=, but recognized as
identical by keyby=. this issue looks like it might solve that (but since I
can't figure out how to reproduce reliably, it'll be hard to tell)

PR #2678, opened 2018-03-16
prevent the utf8 string from being collected by the garbage collector in forder()

### LINKED ISSUE #2674: data.table breaks when there're millions of Chinese characters

shrektan:
Closes #2674
This PR replaced PR #2675 (see comments there). 

If somehow the garbage collector was triggered during sorting (like there're millions of non-ASCII characters), it leads to the collapse of `data.table` (see #2674 for details) **because it assumes there're converted UTF-8 chars in the global string pool.** This PR tries to fix this issue. 

In addition, it brings performance enhancement in the case of millions non-ASCII chars, because now it only needs to be converted to UTF-8 once. **Before this PR, the strings need to be converted twice in ` csort_pre()` and `csort()` respectively**, which may be a big issue for a large character vector (for example, on my computer, `enc2utf8()` takes about 20s for a 1e7 length Chinese character).

# TODO
- [x] a new test
- [x] entries in NEWS.md
- [x] `by = .(x, y)` should return the same row as `by = .(y, x)` (see comment below)
- [x] should not affect the performance if all the strings are ASCII or encoded in UTF-8.


src/forder.c
```diff
@@ -866,7 +866,7 @@ static void csort(SEXP *x, int *o, int n)
   /* can't use otmp, since iradix might be called here and that uses otmp (and xtmp).
      alloc_csort_otmp(n) is called from forder for either n=nrow if 1st column,
      or n=maxgrpn if onwards columns */
-  for(i=0; i<n; i++) csort_otmp[i] = (x[i] == NA_STRING) ? NA_INTEGER : -TRUELENGTH(ENC2UTF8(x[i]));
+  for(i=0; i<n; i++) csort_otmp[i] = (x[i] == NA_STRING) ? NA_INTEGER : -TRUELENGTH(x[i]);
```
MichaelChirico: great catch 👏

shrektan:
@mattdowle It's the first time I realize that the `radixsort` in R is the same as in `data.table`. Thanks for the great work! I don't think I've encountered any issue before so it should be a problem of data.table-1.10.5 only. 

However, regarding to https://github.com/Rdatatable/data.table/blob/master/src/forder.c#L1227 , are you sure that it's ok because the `xd` here will not be converted to `UTF-8`?

About the performance, it gets significantly improved when there're lots on non-ASCII chars but yes, it becomes a little slower if all the strings are ASCII. See the tests on my computers:

```r
library(data.table)
nonascii_string <- function(n, utf8 = TRUE) {
  x <- c("公允价值变动损益", "红利收入", "价差收入", "其他业务支出", "资产减值损失")
  if (isTRUE(utf8)) x <- enc2utf8(x)
  sample(x, n, TRUE)
}

# test1
tmp <- data.table(x = sample(letters, 1e8, TRUE))
system.time(setkey(tmp, x)) 
# test2
tmp <- data.table(x = sample(letters, 1e8, TRUE), y = sample(letters, 1e8, TRUE))
system.time(setkey(tmp, y, x)) 
# test3
tmp <- data.table(x = nonascii_string(1e7))
system.time(setkey(tmp, x)) 
# test4
tmp <- data.table(x = nonascii_string(1e7), y = nonascii_string(1e7))
system.time(setkey(tmp, y, x)) 
# test5
tmp <- data.table(x = nonascii_string(1e5, FALSE))
system.time(setkey(tmp, x)) 
```
version | test1 | test2 | test3 | test4 | test5
---------|------|-------|-------|-------| -----
v1.9.6  | 2.0s | 6.13s | 0.14s | 0.41s | 0.00s
master  | 2.0s | 6.16s | 0.19s | 0.39s | 0.47s (if not fail)
this version  | 2.50s | 6.94s | 0.18s | 0.39s | 0.14s

shrektan:
I confirm that we need to edit here as well.

https://github.com/Rdatatable/data.table/blob/4d8545e8ee65edac22724f9d917bb9e76dbac838/src/forder.c#L1227

```r
library(data.table)
utf8_strings <- enc2utf8(c("红利收入", "价差收入"))
native_strings <- enc2native(utf8_strings)
mixed_strings <- c(utf8_strings, native_strings)
DT <- data.table(x = mixed_strings, y = 1)
DT[, .N, by = .(x, y)]

          x y N
1: 红利收入 1 2
2: 价差收入 1 2

DT[, .N, by = .(y, x)]

   y        x N
1: 1 红利收入 1
2: 1 价差收入 1
3: 1 红利收入 1
4: 1 价差收入 1
```

shrektan:
@mattdowle I've pushed more commits to fix the cases on grouping (a.k.a, `DT[, j, by = .(x, y)]`) and the overheads for all ASCII/UTF8 columns.

mattdowle:
Thanks for all this!  It's looking great to me.  According to codecov, `need2utf8` always returns false. Is there a way to add a test to cover those 4 lines?

shrektan:
I guess adding an example uses the `Latin-1` encoding may work. I'll submit one later.

shrektan:
The error log I downloaded from `AppVeyor` says:

```
Running test id 1896.4      Test 1896.4 ran without errors but failed check that x equals y:
> x = nrow(DT[, .N, by = .(z, x, y)]) 
First 6 of 1 (type 'integer'): [1] 10
> y = 5L 
First 6 of 1 (type 'integer'): [1] 5
```

I can't understand the failure because the following code gives the correct answer `5` on both my Mac and Windows (I even tried to set my language and locale to the United States)...

```r
utf8_strings <- c("\u00e7ile", "fa\u00e7ile", "El. pa\u00c5\u00a1tas", "\u00a1tas", "\u00de")
latin1_strings <- iconv(utf8_strings, from = "UTF-8", to = "latin1")
mixed_strings <- c(utf8_strings, latin1_strings)
DT <- data.table(x = mixed_strings, y = c(latin1_strings, utf8_strings), z = 1)
nrow(DT[, .N, by = .(z, x, y)])
# 5
```

PR #2680, opened 2018-03-16
keep git commit metadata in proper DESCRIPTION field

jangorecki:
Previously we kept Commit field which was not valid in package check, we can use Revision field for that purpose. This PR will allow users to lookup exact git commit they are using.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ environment:
   - R_VERSION: devel
 
 before_build:
-  - cmd: ECHO Revision: %APPVEYOR_REPO_COMMIT%>>DESCRIPTION
+  - cmd: ECHO Revision: $Env:APPVEYOR_REPO_COMMIT>>DESCRIPTION
 
 build_script:
   - set _R_CHECK_FORCE_SUGGESTS_=false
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -29,6 +29,9 @@ environment:
 
   - R_VERSION: devel
 
+before_build:
+  - cmd: ECHO no Revision metadata added to DESCRIPTION
+  #translate from unix: - cmd: ECHO "Revision:" $CI_BUILD_REF >> ./DESCRIPTION
```
jangorecki: some win user please push to this branch corresponding windows command so packages in appveyor will also contain Revision field. You also need to change env var name to appveyor commit env var.

MichaelChirico:
does this affect `.onAttach`?

jangorecki:
@MichaelChirico this PR no, but another one can put commit hash to package startup message.

appveyor.yml
mattdowle: Do we have any Windows developers on the project?   Or do you mean a Windows user.  I'm not really following.   Does this need to happen before merging,  or can I just merge the PR now.

jangorecki: @mattdowle We can merge the first commit from this branch, other ones are failed attempts to append line to DESCRIPTION file on windows. Appveyor does not build drat repo so it is not as much important for this PR.

PR #2693, opened 2018-03-21
helper to keep dev data.table up to date, closes #2681

### LINKED ISSUE #2681: Efficient update data.table to latest devel
Skip downloading package if your git commit is latest one. This is already done in db-benchmarks: https://github.com/h2oai/db-benchmark/blob/0922840dd7bc768aa11f937128bd066fc1246986/helpers.R#L97-L118
It is very useful for scheduling data.table updates or updating from within a workflow.

jangorecki:
Closes #2681 
New function to efficiently update dev data.table to latest version. Before downloading package it checks commit revision of currently installed data.table, and if new commit revision is available in our drat repository then package is downloaded and installed. Such function can be called on startup of R process if it is desired to stay up to date with devel version. It was used before in h2o benchmarks of data.table in https://github.com/h2oai/db-benchmark/blob/0922840dd7bc768aa11f937128bd066fc1246986/helpers.R#L97-L118
Function works for any drat repo as long as drat has commit information in PACKAGES file. Function defaults are set for updating data.table.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -101,6 +101,8 @@ export(dcast.data.table)
 S3method(dcast, data.table)
 
 import(utils)
+S3method(update, dev.pkg)
+export(update.dev.pkg)
 S3method(tail, data.table)
 S3method(head, data.table)
 import(stats)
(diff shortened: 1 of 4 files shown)


jangorecki:
example:
```sh
Rscript -e 'install.packages("data.table", repos="https://rdatatable.gitlab.io/data.table")'
Rscript -e 'data.table::update.dev.pkg()'
#R data.table package is up-to-date at 8c41a5d6a5e7509ca382b52577f7371ef177280b (1.10.5)
```

PR #2698, opened 2018-03-22
test.with.other.packages

### LINKED ISSUE #2696: Add integration tests to CI
to run
```r
test.data.table(with.other.packages = TRUE)
```

jangorecki:
Closes #2696

Just setting env var will invoke second call to `test.data.table` doing integration tests only.
```
TEST_DATA_TABLE_WITH_OTHER_PACKAGES: "TRUE"
```
IMO better would be to have that inside single `test.data.table` call but capability to run multiple Rraw files has just been removed from there.

### DIFF
--- a/inst/tests/other.Rraw
+++ b/inst/tests/other.Rraw
@@ -5,13 +5,13 @@ if (exists("test.data.table",.GlobalEnv,inherits=FALSE)) {
   warning("This is dev where with.other.packages should not be run. Instead, use a fresh R session with data.table installed. ",
           "Not doing so in dev can be the cause of both false errors and false passes.")
 }
-if (!"package:data.table" %in% search()) stop("data.table should be already attached, see usage header in inst/tests/other.Rraw")
+if (!"package:data.table" %in% search()) stop("data.table should be already attached. Usage: require(data.table); test.data.table(with.other.packages=TRUE)")
 
 test = data.table:::test
 INT = data.table:::INT
 
-pkgs = c("ggplot2", "hexbin", "plyr", "caret", "xts", "gdata", "zoo", "nlme", "bit64", "knitr", "plm", "parallel", "dtq")
-if (any(duplicated(pkgs))) stop("Internal error: pkgs has a dup")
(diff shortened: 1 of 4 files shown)


inst/tests/other.Rraw
```diff
@@ -5,26 +5,31 @@ if (exists("test.data.table",.GlobalEnv,inherits=FALSE)) {
...
-  assign(paste0("test_",s), loaded<-suppressWarnings(suppressMessages(require(s, character.only=TRUE))))
-  if (!loaded) cat("\n**** Other package",s,"is not installed. Tests using it will be skipped.\n\n")
-}
+pkgs = c("ggplot2", "hexbin", "plyr", "caret", "xts", "gdata", "zoo", "nlme", "bit64", "knitr", "plm", "parallel")
+if (any(duplicated(pkgs))) stop("Packages defined to be loaded for integration tests in 'inst/tests/other.Rraw' contains duplicates.")
+
+is.require = function(pkg) suppressWarnings(suppressMessages(isTRUE(require(pkg, character.only=TRUE, quietly=TRUE, warn.conflicts=FALSE))))
+loaded = sapply(pkgs, is.require)
+cat("\nPackages to be tested:\n\n")
+print(data.table(pkg=pkgs, loaded)[loaded==TRUE, version:=as.character(sapply(pkg, function(p) format(packageVersion(p))))][])
```
mattdowle: This is nice. If a package isn't installed there used to be a message saying its tests will be skipped. Could that message be brought back whilst still keeping the better new table as well?   I imagine that when called from CI, the CI package would catch/warn about packages not installed at that point.  I've more in mind the (much rarer case) of users running `test.data.table(with.other.packages=TRUE)`.

jangorecki: @mattdowle Done. I adjusted default to the same way that R CMD check would expect.
By default all integration suggested pkgs are required, this can be turned off by `_R_CHECK_FORCE_SUGGESTS_=FALSE` env var.

jangorecki:
https://gitlab.com/Rdatatable/data.table/-/jobs/60530581/artifacts/file/bus/test-r-release/data.table.Rcheck/tests/main.Rout

PR #2701, opened 2018-03-22
Minor grammar changes + microbenchmark example

HughParsonage:
Only substantial change is the example advice to use `setattr` which I noticed in `setDT`. If it's invalid, I'll remove.

### DIFF
--- a/vignettes/datatable-benchmarking.Rmd
+++ b/vignettes/datatable-benchmarking.Rmd
@@ -104,4 +104,4 @@ You can control how much threads you want to use with `setDTthreads`.
 
 ## avoid `data.table()` inside a loop
 
-As of now `data.table()` has an overhead, thus inside loops it is preferred to use `as.data.table()` or `setDT()` or maybe even `setattr(<list>, "class", c("data.table", "data.frame"))` on a valid list.
+As of now `data.table()` has an overhead, thus inside loops it is preferred to use `as.data.table()` or `setDT()` on a valid list.
--- a/vignettes/datatable-benchmarking.Rmd
+++ b/vignettes/datatable-benchmarking.Rmd
@@ -17,7 +17,7 @@ h2 {
 }
 </style>
 
-This document is a guide to accurate performance measurements of data.table, highlighting best practices and traps to avoid.
+This vignette documents the best ways to accurately measure data.table's performance, as well as common traps to avoid.
 
 ## `microbenchmark` is not always appropriate
 Repetitive benchmarking like that used by `microbenchmark` may not always be appropriate. 
--- a/vignettes/datatable-benchmarking.Rmd
+++ b/vignettes/datatable-benchmarking.Rmd
@@ -2,7 +2,7 @@
 title: "Benchmarking data.table"
 date: "`r Sys.Date()`"
 output:
-  rmarkdown::html_document:
+  rmarkdown::html_vignette:
     toc: true
     number_sections: true
     
--- a/vignettes/datatable-benchmarking.Rmd
+++ b/vignettes/datatable-benchmarking.Rmd
@@ -20,7 +20,7 @@ h2 {
 This document is a guide to accurate performance measurements of data.table, highlighting best practices and traps to avoid.
 
 ## `microbenchmark` is not always appropriate
-Repeating benchmarking many times usually does not fit well for data processing tools. 
+Repetitive benchmarking like that used by `microbenchmark` may not always be appropriate. 
 While it makes perfect sense for timing atomic calculations, it may provide misleading results for common data processing tasks, which are typically large than the scale for which `microbenchmark` was designed, and run only once.
 
 For example, the following gives the impression that `data.table` is an order of magnitude slower than `data.frame` for updating columns.
@@ -56,6 +56,10 @@ system.time(DT[, y := .I])
 #>    0.01    0.01    0.03
 ```
 
+Further, when benchmarking `set*` functions it only makes to measure the first run. Those functions update data.table by reference thus after the first run the data.table involved in the measurement will have changed.
+
+To correctly measure the average performance of `set`, protect your data.table from being updated by reference operations by using `copy` or `data.table:::shallow`. Be aware `copy` might be very expensive relative to the operation you want to time as it duplicates the whole table, but this is what other packages usually do. It is unlikely we want to include duplication time in time of the actual task we are benchmarking. It may be prudent to measure the execution time of `copy` alone and subtract it from the operation with both `copy` and the operation being timed.
+
 Matt once said:
 
 > I'm very wary of benchmarks measured in anything under 1 second. Much prefer 10 seconds or more for a single run, achieved by increasing data size. A repetition count of 500 is setting off alarm bells. 3-5 runs should be enough to convince on larger data. Call overhead and time to GC affect inferences at this very small scale.
@@ -89,11 +93,7 @@ options(datatable.use.index=TRUE)
 `options(datatable.optimize=2L)` will turn off optimization of subsets completely, while `options(datatable.optimize=3L)` will switch it back on.
 `use.index=FALSE` will force query not to use index even if it exists, but existing keys are used for optimization. `auto.index=FALSE` only disables building index automatically when doing subset on non-indexed data.
 
-## _by reference_ operations
 
-When benchmarking `set*` functions it make sense to measure only the first run. Those functions update data.table by reference thus in subsequent runs they run on a different data.table.
-
-Hence, to correctly measure the performance of `set`, protect your data.table from being updated by reference operations by using `copy` or `data.table:::shallow`. Be aware `copy` might be very expensive as it needs to duplicate whole object, but this is what other packages usually do. It is unlikely we want to include duplication time in time of the actual task we are benchmarking. It may be prudent to measure the execution time of `copy` alone. 
 
 
 
--- a/vignettes/datatable-benchmarking.Rmd
+++ b/vignettes/datatable-benchmarking.Rmd
@@ -35,9 +35,11 @@ library(data.table)
 Benchmarking any code seems easy only on the surface. In practice, there are many pitfalls to accidentally fall into, and many opportunities to draw wrong conclusions. This vignette documents mistakes we've made or encountered to better guide speed comparisons of data.table with base R and other packages. 
 
 ## Know thy query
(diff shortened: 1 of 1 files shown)


vignettes/datatable-benchmarking.Rmd
```diff
@@ -50,22 +91,17 @@ options(datatable.use.index=TRUE)
...
 ## multithreaded processing
 
-One of the main factor that is likely to impact timings is number of threads in your machine. In recent versions of data.table some of the functions has been parallelized.
+One of the main factors that is likely to impact timings is the number of threads on your machine. In recent versions of data.table some of the functions has been parallelized.
 You can control how much threads you want to use with `setDTthreads`.
 
 ## avoid `data.table()` inside a loop
 
-As of now `data.table()` has an overhead, thus inside loops it is preferred to use `as.data.table()` or `setDT()` on a valid list.
+As of now `data.table()` has an overhead, thus inside loops it is preferred to use `as.data.table()` or `setDT()` or maybe even `setattr(<list>, "class", c("data.table", "data.frame"))` on a valid list.
```
jangorecki: I wouldn't go as far here, it would probably need also `alloc.col`.

HughParsonage: Remove altogether, or recommend `alloc.col` in conjuction with it?

jangorecki: `setDT` is already enough, there won't be significant difference to it.

vignettes/datatable-benchmarking.Rmd
```diff
@@ -17,11 +17,56 @@ h2 {
...
 </style>
 
-This document is meant to guide on measuring performance of data.table. Single place to documents best practices or traps to avoid.
+This vignette documents the best ways to accurately measure data.table's performance, as well as common traps to avoid.
+
+## `microbenchmark` is not always appropriate
+Repetitive benchmarking like that used by `microbenchmark` may not always be appropriate. 
+While it makes perfect sense for timing atomic calculations, it may provide misleading results for common data processing tasks, which are typically larger than the scale for which `microbenchmark` was designed, and run only once.
+
+For example, the following gives the impression that `data.table` is an order of magnitude slower than `data.frame` for updating columns.
```
st-pasha: Again this, together with the example below, creates a bad impression like I'm being gaslighted.
To an untrained eye, the example looks perfectly reasonable. And if you don't explain what exactly is bad with the example, I will probably be thinking in the back of my head: "ok looks like data.table is sometimes orders of magnitudes slower, and in some other example it was only 1.5 times slower (0.02 vs 0.03). So why do they say there's no real difference?"

Btw measuring the second example (n=1e7) with microbenchmark still shows that data.table is more than twice slower on average:
```
> microbenchmark(DF$y <- seq.int(nrow(DF)), DT[, y := .I], times=50L)
Unit: milliseconds
                      expr       min        lq     mean   median       uq
 DF$y <- seq.int(nrow(DF))  8.267302  9.715634 21.92438 19.45919 21.14266
         DT[, `:=`(y, .I)] 17.495159 20.004846 47.44068 45.19089 49.50766
```
And I appreciate the fact that the data is being changed by-reference, so running this code multiple times may overwrite the old data instead of creating a new vector. However I don't see why this would create any perceptible performance difference: allocating a vec is supposed to be extremely fast...
 
Also I just re-run the benchmark for n=1e8, and the speed difference is 3 times, not in data.table's favor. So I don't know, maybe pick a different example? I'm running R 3.4.2 and afair it had some performance improvements to `seq.int`.

vignettes/datatable-benchmarking.Rmd
```diff
@@ -17,11 +17,56 @@ h2 {
...
+DF <- as.data.frame(DT)
+system.time(DF$y <- seq.int(nrow(DF)))
+#>    user  system elapsed 
+#>    0.01    0.00    0.02
+system.time(DT[, y := .I])
+#>    user  system elapsed 
+#>    0.01    0.01    0.03
+```
+
+Further, when benchmarking `set*` functions it only makes sense to measure the first run. Those functions update data.table by reference thus after the first run the data.table involved in the measurement will have changed.
```
st-pasha: I understand this point from theoretical perspective, but is there a tangible evidence that setting something multiple times makes any performance difference? If yes, then this vignette is probably the best place to enunciate such examples. If no, then forcing a copy of data before each measurement creates more statistical noise than it purports to solve.

vignettes/datatable-benchmarking.Rmd
```diff
@@ -17,11 +17,56 @@ h2 {
 }
 </style>
 
-This document is meant to guide on measuring performance of data.table. Single place to documents best practices or traps to avoid.
+This vignette documents the best ways to accurately measure data.table's performance, as well as common traps to avoid.
```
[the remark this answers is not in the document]
st-pasha: We can try something along the following lines:

> Benchmarking any code seems easy only on the surface. In practice, there are many pitfalls to accidentally fall into, and many opportunities to draw wrong conclusions. This vignette attempts to document the best practices on how to assess the performance of common data manipulation tasks. It provides examples on how to compare the speed of `data.table` vs that of `dplyr` and base `R`.

st-pasha:
In broad strokes, the following could be the document outline:

1. First section: **"scale matters"**. Explain that if we view the dataset as a function of `n` (number of rows), then it would be impossible to find any code whose time is proportional to `n`. Which is why when measuring performance the most important thing is to measure it at different `n`s. It is a mortal sin to microbenchmark some code for say `n=10`, and then assume that any inferences you draw will hold for `n=10,000,000`.
2. Second section: **"going wide"**. Another dimension for a data frame is the number of columns. Some operations take the same amount of time no matter how wide the data frame is, while others are not that lucky. This may be especially true when comparing against implementations that like to create many copies of the data, or if they store the data row-wise, etc.
3. Third section: **"disk cache"**. This is about reading a file the first time vs. the second time (and how in most use cases it is the first time that only matters).
4. **"internal optimizations"**. Some implementations may cache / memoize certain results, which makes subsequent runs faster compared to the first. Always try to time the code when running once, and then compare with multiple runs from the microbenchmark. Are the timings approximately same? If there's a big discrepancy, you might need to look into the implementation.
5. **"miscellaneous"**. Other things, like running in `--vanilla` mode, making sure you have a production version of the package and not a debug one, etc. Also about being polite, and if you're about to publish some benchmarking results against another library -- reach out to the authors of that other package to check with them if you're using their library correctly.

mattdowle:
Perhaps the fact that 3 calls to `getOption()` appear in the argument list of `[.data.table` could be used as an example. The default could be added back to `getOption()` and the microbenchmark results presented.  Some inferences then made about the operation (say, adding a column like above).  Then continue with an `Rprof()` and discover, "oh, wait, all the time is in getOption?`.  Then explain call overhead and demonstrate what we mean when we say that the inference made earlier is correct or not depending on whether the user does need to call `DT[...]` in a loop or not.  In that case, actually scale up the data so that the call overhead in `DT[...]` becomes an insignificant percentage, and demonstrate using a reproducible example the inference then swapping around.  It could be said that this is a disadvantage of data.table in that users need to know this and use a different function (`set*`) to avoid this call overhead.

vignettes/datatable-benchmarking.Rmd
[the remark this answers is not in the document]
jangorecki: @st-pasha Good points, but I wouldn't mention any competitors here like dplyr/base R. This will be quite obvious as it appears in every benchmark.

vignettes/datatable-benchmarking.Rmd
jangorecki: There are going to be big improvements to `seq` function in next major R release. They won't be materialized unless needed, so better to chose some function that will stay valid for this example.

HughParsonage:
Thanks Pasha. Agreed. I've added a bit to reflect the first part of this structure and will continue on. 

(N.B. I've added some package to the vignette, without realizing that of course they'd need to be in Suggests. I'll leave as-is (i.e. broken) for now while the content is still under discussion. Once the PR's content is basically accepted, we can trim away those packages or include them in Suggests as appropriate.)

vignettes/datatable-benchmarking.Rmd
```diff
@@ -17,7 +17,102 @@ h2 {
 }
 </style>
 
-This vignette documents the best ways to accurately measure data.table's performance, as well as common traps to avoid.
+```{r setup}
+library(knitr)
+opts_chunk$set(fig.width = 8, fig.height = 5)
+```
+
+```{r loadPackages}
```
jangorecki: No point in adding new dependencies in vignette which doesn't even need to evaluate anything. If you use R code chunks here do not add any new dependencies. We just simplified Suggests a moment ago.

HughParsonage:
I've written as much as my expertise can carry me. In particular, I'm not sure what sort of query would be affected by the width. Happy to make changes, just might not be the best person for it.

vignettes/datatable-benchmarking.Rmd
HughParsonage: Agreed, just making the code runnable for this discussion while we work out whether what I'm saying is worthwhile. Intend to remove these dependencies just before being merged.

mattdowle: `microbenchmark` is one of these dependencies, iiuc, and it's used in this vignette, which seems reasonable.  Don't we need to add `microbenchmark` to DESCRIPTION:Suggests then?  How could that dependency be removed before merging?

HughParsonage: `microbenchmark` probably can't, (unless we want to avoid explicit code examples and talk more generally), but things like `magrittr` and `hutils` definitely can. `ggplot2` could be replaced by base graphics if needed.

jangorecki: We can just paste the code used with microbenchmark and not use it inside R code chunk. We actually don't need to use R code chunks at all in this doc. It meant to describe good practice for benchmarking, and not running benchmarks. Joshua Ulrich is new maintainer of microbenchmark and he manage to strip ggplot2 from imports to suggests so microbenchmark is now really "micro", but still if we don't use it inside data.table I don't see the point of making our Suggests list longer again. I filled #2740 because there are still some packages that could be removed from there.

mattdowle: Yes agreed. Sounds good. So this vignette will display the benchmark code and results, but not actually run the benchmark.  That makes sense, as otherwise, a potentially long running benchmark would need to run when `R CMD check` runs.  Ok. Good plan.

HughParsonage:
Superseded by #2954

mattdowle:
#2954 doesn't really address @st-pasha's good comments above, which are partly based on discussions with me.  Since this is a PR and its branch is deleted, I can't reopen it, so I'll create a new issue.

PR #2711, opened 2018-03-30
Closes #2695 -- setindexv creates multiple indices when passed list

### LINKED ISSUE #2695: setindexv could accept list of characters to create multiple indices
It could be internally optimized for speed but for now it will be enough to run in loop.
API to create index on `a` field and second index on `a, b` fields:
```r
setindexv(DT, list("a",c("a","b"))
```

MichaelChirico:
Closes #2695 

PS I am loving the new `test.data.table()` diagnostics 👍

### DIFF
--- a/R/setkey.R
+++ b/R/setkey.R
@@ -28,8 +28,6 @@ key2 <- function(x) {
   stop("key2() is now deprecated. Please use indices() instead.")
 }
 
-
-
 setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TRUE)
 {
   if (is.null(cols)) {   # this is done on a data.frame when !cedta at top of [.data.table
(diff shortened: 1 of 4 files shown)


R/setkey.R
```diff
@@ -9,7 +9,14 @@ setkey <- function(x, ..., verbose=getOption("datatable.verbose"), physical=TRUE
 
 # FR #1442
 setindex <- function(...) setkey(..., physical=FALSE)
-setindexv <- function(...) setkeyv(..., physical=FALSE)
+setindexv <- function(x, cols, verbose=getOption("datatable.verbose")) {
+  if (is.list(cols)) {
+    sapply(cols, setkeyv, x = x, verbose = verbose, physical = FALSE)
+    return(invisible(x))
+  } else {
+    setkeyv(x, cols, verbose, physical=FALSE)
```
jangorecki: I would use `verbose=verbose` as this argument can easily shift in future if we add another one in front. But as of now works well.

MichaelChirico: Good catch, will fix.

PR #2712, opened 2018-03-30
Closes #2688 -- document some details on download.file in fread

### LINKED ISSUE #2688: Better documentation on use of file://
Examining `?fread` after answering [this question on SO](https://stackoverflow.com/q/49357652/3576984), I notice the information I provided was lacking; here's all there is (emphasis mine):

> Either the file name to read (containing no \n character), a shell command that pre-processes the file (e.g. fread("grep blah filename")) or the input itself as a string (containing at least one \n), see examples. In both cases, a length 1 character string. **A filename input is passed through path.expand for convenience and may be a URL starting http:// or file://**.

There's no mention in the documentation of the internal use of `download.file` (even for `http://` inputs).

MichaelChirico:
Closes #2688 

Also went ahead and forced `"auto"` into `"internal"` for `file://` inputs since `?download.file` says this is the result of using `"auto"` for file:// URLs. Feels more natural/robust to explicitly declare `method = "internal"` as a way to better track if this evolves.

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -53,7 +53,7 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
         curl::curl_download(input, tmpFile, mode="wb", quiet = !showProgress)
       }
       else if (str6=="ftp://" || str7== "http://" || str7=="file://") {
-        method = if (str7=="file://") "auto" else getOption("download.file.method", default="auto")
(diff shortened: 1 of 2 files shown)


mattdowle:
(please put the "Closes #nnnn" line at the top of the first comment in the PR -- I edited this one.  One line per issue closed.)

MichaelChirico:
@mattdowle noted. it's a shame GitHub doesn't hyperlink the issue number in the PR title

PR #2566, opened 2018-01-13
fix the bug when keys contain non UTF8 strings

shrektan:
Closes #2462. #1826 is the older issue...

For strings, `data.table` compare their values in UTF8 encoding. However, due to missing two `ENC2UTF8` in `csort()` and `csort_pre()`, the order that `data.table` creates actually depends on the encoding. On Windows, the fact that the default encoding is not UTF8 leads to some weird output when there're strings in keys.

Please tell me that if there's anything more need to be done, thanks.

### Here's what we get now.

``` r
library(data.table)
dt <- data.table(
  x = c("公允价值变动损益", "红利收入", "价差收入", "其他业务支出", "资产减值损失"),
  y = enc2utf8(c("公允价值变动损益", "红利收入", "价差收入", "其他业务支出", "资产减值损失")),
  z = 1:5
)
setkey(dt, x)
dt[]
##                   x                y z
## 1:         价差收入         价差收入 3
## 2: 公允价值变动损益 公允价值变动损益 1
## 3:     其他业务支出     其他业务支出 4
## 4:         红利收入         红利收入 2
## 5:     资产减值损失     资产减值损失 5


dt[J("公允价值变动损益"), z]
## [1] 1


setkey(dt, y)
dt[]
##                   x                y z
## 1:         价差收入         价差收入 3
## 2: 公允价值变动损益 公允价值变动损益 1
## 3:     其他业务支出     其他业务支出 4
## 4:         红利收入         红利收入 2
## 5:     资产减值损失     资产减值损失 5


dt[J("公允价值变动损益"), z]
## [1] 1


sessionInfo()
## R version 3.4.1 (2017-06-30)
## Platform: x86_64-w64-mingw32/x64 (64-bit)
## Running under: Windows 7 x64 (build 7601) Service Pack 1
## 
## Matrix products: default
## 
## locale:
## [1] LC_COLLATE=Chinese (Simplified)_People's Republic of China.936 
## [2] LC_CTYPE=Chinese (Simplified)_People's Republic of China.936   
## [3] LC_MONETARY=Chinese (Simplified)_People's Republic of China.936
## [4] LC_NUMERIC=C                                                   
## [5] LC_TIME=Chinese (Simplified)_People's Republic of China.936    
## 
## attached base packages:
## [1] stats     graphics  grDevices utils     datasets  methods   base     
## 
## other attached packages:
## [1] data.table_1.10.5
## 
## loaded via a namespace (and not attached):
[...]

PR #2715, opened 2018-03-31
improve test 1864 by using latin1 encoded strings

shrektan:
So it can be tested on all the platforms. It improves the test for PR #2566.


inst/tests/tests.Rraw
```diff
@@ -11146,16 +11146,20 @@ DT = data.table(col1 = c(1,1,1, 2,2,2), col2 = c(2,2,2,1,1,1), ID = c(rep(1,3),
...
+utf8_strings <- c("\u00e7ile", "fa\u00e7ile", "El. pa\u00c5\u00a1tas", "\u00a1tas", "\u00de")
+latin1_strings <- iconv(utf8_strings, from = "UTF-8", to = "latin1")
+mixed_strings <- c(utf8_strings, latin1_strings)
+DT1 <- data.table(x = mixed_strings, y = c(latin1_strings, utf8_strings), z = 1:10)
+DT2 <- copy(DT1)
+setkey(DT1, x)
+setkey(DT2, y)
+test(1864.1, DT1$x, sort(c(utf8_strings, utf8_strings), method = "radix"))
+test(1864.2, DT2$y, sort(c(utf8_strings, utf8_strings), method = "radix"))
+test(1864.3, DT1[J(utf8_strings)], DT1[J(latin1_strings)])
```
mattdowle: Imagine that, for reason of a strange bug, both `DT1[J(utf8_strings)]` and `DT1[J(latin1_strings)]` returned NULL.  The test would still pass. I like this pattern for that reason : 
test(1864.3,  DT1[J(utf8_strings)],  ans<- data.table( .. known answer .. ) )
test(1864.4,  DT1[J(latin1_strings)], ans)

mattdowle: Perhaps use the integers in `z` to construct `ans`.

mattdowle: Otherwise, great!

shrektan: You're right. Using an explicit answer is indeed better.

PR #2719, opened 2018-04-01
fread gains index argument for setting secondary indices

### LINKED ISSUE #2633: fread could set index as well
having a `key` argument for `fread` is quite nice, it helps cut down on code like:

```
DT = fread('path.csv')
setkey(DT, ID)
```

As well as being robust to changing the name of `DT` over time without having to track down every line where `DT` was used. 

Since this wasn't any deep change to how `fread` works, but rather just entails running `setkeyv` just before returning `ans`, it would be easy to offer an `index` argument with similar goals.

```
DT = fread('path.csv', key = 'ID')
setindex(DT, ID2)
# could be streamlined as
DT = fread('path.csv', key = 'ID', index = 'ID2')
```

The only roadblock to implementing this right away is a small API issue -- should we allow multiple indices to be set, and if so what's the proper syntax? Most natural to me is: if `index` is given as atomic with length `1`, `strsplit` it along `,` (as is done for `key`); with length `>1`, this is a single index (again, like `key`); and with `index` a `list`, `*apply` the logic from the first two cases to each element.

MichaelChirico:
Closes #2633 

Also fleshed out some more details in `?fread` about the order of arguments & improved the error for a bad `key` argument, and simplified the test for `setindexv` accepting a list


R/fread.R
```diff
@@ -124,12 +124,19 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
...
     if (length(key) == 1L) {
-      key = strsplit(key, split = ",")[[1L]]
+      key = strsplit(key, split = ",", fixed = TRUE)[[1L]]
     }
     setkeyv(ans, key)
   }
+  if (!is.null(index) && data.table) {
+    if (!all(sapply(index, is.character)))
+      stop("index argument of data.table() must be a character vector naming columns (NB: col.names are applied before this)")
+    to_split = sapply(index, length) == 1L
```
MichaelChirico: need to fix this, won't work correctly for atomic vector

R/fread.R
```diff
@@ -124,12 +124,27 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
...
-      key = strsplit(key, split = ",")[[1L]]
+      key = strsplit(key, split = ",", fixed = TRUE)[[1L]]
     }
     setkeyv(ans, key)
   }
+  if (!is.null(index) && data.table) {
+    if (!all(sapply(index, is.character)))
+      stop("index argument of data.table() must be a character vector naming columns (NB: col.names are applied before this)")
+    if (is.list(index)) {
+      to_split = sapply(index, length) == 1L
```
MichaelChirico: i guess this should be `vapply_1i`

PR #2721, opened 2018-04-01
Closes #1912 -- bquote usage excluded from replace_dot_alias

### LINKED ISSUE #1912: Bug: .() mis-interpreted when being used as plotmath
I just found out that sometimes `.()` is needed to create dynamically-labeled plots in `base` R:

```
png("~/Desktop/dterror.png")
par(mfrow = c(1, 2))
DT[ , {
  plot(1:10)
  text(2, 9, bquote(R^2 == .(summary(lm(y ~ x))$r.squared)))}]

plot(1:10)
text(2, 9, bquote(R^2 == .(summary(lm(y ~ x, data = DT))$r.squared)))
dev.off()
```

dterror

When used outside of `[.data.table`, the text is converted (as expected) to the evaluated value of the regression. But `[.data.table` (I imagine) is detecting this as an alias for `list` and substituting, leading to the unexpected result on the left.

MichaelChirico:
#1912 -- `bquote` itself has a different usage of `.()` which was being undone by `replace_dot_alias`; an exception is made. AFAIK `bquote` is the only function in `base` with this behavior (not that this is a particularly searchable issue), so this should suffice.

h/t @ecoRoland for the code


MichaelChirico:
haha whoops. fails quite catastrophically. CI is on 🔥 ... fixing

PR #2722, opened 2018-04-02
Adapted verbose message when subsetting optimization is switched off.

### LINKED ISSUE #2700: verbose messages does not appear on index and cross product of subset elements > 1e4
```r
library(data.table)
dt=data.table(a=1L,b=2L)
setindexv(dt, c("a","b"))
dt[a%in%1:100 & b%in%1:100, verbose=TRUE]
#Optimized subsetting with index 'a__b'
#on= matches existing index, using index
#Starting bmerge ...done in 0.000sec 
#   a b
#1: 1 2
dt[a%in%1:1000 & b%in%1:1000, verbose=TRUE]
#   a b
#1: 1 2
```
As we have index on those columns it would be useful to have verbose output.
Ideally something like `Optimized subsetting with index 'a__b' disabled due to ...`.

MarkusBonsch:
Closes #2700.
No logic changes, only one verbose message and one test affected.
The new verbose message is:
`"Subsetting optimization disabled because the cross-product of RHS values exceeds 1e4, causing memory problems."`

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -2907,6 +2907,7 @@ isReallyReal <- function(x) {
   ## convert i to data.table with all combinations in rows.
   if(length(i) > 1L && prod(vapply(i, length, integer(1L))) > 1e4){
     ## CJ would result in more than 1e4 rows. This would be inefficient, especially memory-wise #2635
+    if (verbose) {cat("Subsetting optimization disabled because the cross-product of RHS values exceeds 1e4, causing memory problems.\n");flush.console()}
     return(NULL)
   }
   ## Care is needed with names as we construct i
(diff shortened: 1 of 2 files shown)


MarkusBonsch:
No idea, why the R-devel test is not passing. Breaks at unt test 1808.3, which I didn't touch.

jangorecki:
R-devel often produce false positives.

mattdowle:
R-devel test is with latest nightly snapshot of R-devel. A change yesterday in R-devel by R-core might have broken data.table and other packages. These tend to be quickly found and fixed within days.  If it persists for a few days then we'll have a more detailed look and see if it's a change in R that we need to accommodate in data.table.  We have also seen true memory faults in data.table show up in R-devel only, though. The reasons vary but there's a always a reason.

PR #2725, opened 2018-04-03
provide explicit answer for the test 1896.1

### LINKED ISSUE #2724: R 3.1.0 test 1896.1 failure
3.1.0 started to fail after this commit: https://github.com/Rdatatable/data.table/commit/ecf92e9ac3f34fa3e884c6e347a8624ac2e3b314

shrektan:
Closes #2724 

The test 1896.1 used the radix sort which is only available after R 3.3.0 thus leads to the issue.


shrektan:
@jangorecki Would you please check if this patch fixes the issue on R3.1.0? Thanks.

jangorecki:
@shrektan yes it did: https://gitlab.com/Rdatatable/data.table/-/jobs/60888495

shrektan:
I've added a simple comment.

inst/tests/tests.Rraw
```diff
@@ -11550,6 +11550,9 @@ test(1895, getDTthreads(verbose=TRUE), output="omp_get_max_threads.*omp_get_thre
 utf8_strings = c("\u00e7ile", "fa\u00e7ile", "El. pa\u00c5\u00a1tas", "\u00a1tas", "\u00de")
 latin1_strings = iconv(utf8_strings, from = "UTF-8", to = "latin1")
 DT = data.table(x = sample(latin1_strings, 1000, replace=TRUE), key = "x")
+# The ans below is generated by `sort(utf8_strings, method = "radix")` on R (>= 3.3.0).
+# Note that you should generally avoid to call `sort(., method = "radix")` in `data.table`'s test
+# because `data.table` could be used on any R version that is larger than 3.1.0.
```
jangorecki: use `equal or larger than 3.1.0` or `at least in 3.1.0`.

PR #2730, opened 2018-04-09
test use less memory, pass on 2GB ram machine

jangorecki:
Test which was added in https://github.com/Rdatatable/data.table/commit/6c654e9bd2cf00396b6be15dc1a0e850915bbb26 uses quite a lot memory and often fail on machine with 2GB ram.
Using parent commit of the above https://github.com/Rdatatable/data.table/commit/b49ac6be7911e01b8fa10b996d9a82b256a608ef I was able to reproduce issue using smaller size of data, thus we can safely use smaller data in this test.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -2912,7 +2912,7 @@ test(1034, as.data.table(x<-as.character(sample(letters, 5))), data.table(V1=x))
 
   # segfault of unprotected var caught with the help of address sanitizer
   set.seed(1)
-  val = sample(c(1:5, NA), 1e6L, TRUE)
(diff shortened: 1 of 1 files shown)


mattdowle:
As luck would have it, this should fix [this fail](https://ci.appveyor.com/project/Rdatatable/data-table/build/1.0.2250/job/5hily4a96cuyp4la) for PR #2719 just seen on R-devel-only Windows-only 32bit-only : 
```
    Running test id 1642.03      Error: memory exhausted (limit reached?) 
```
My guess is that's nothing to do with R-devel or Windows or 32bit, but it's just that instance.  And nothing to do with that test (1642.03) either but the root cause being this earlier `melt` test reduced in size by this PR.

PR #2734, opened 2018-04-10
Fix bug when joining on columns of different types

### LINKED ISSUE #2592: joins with type coercion: 1.5 == 1 is TRUE
Maybe this is a known issue, but to me it came as a very bad surprise.
During a join, if `i` is coerced to integer to match `x`'s column type, 1.5 is joined to 1:
```
library(data.table)
dt1 <- data.table(x = 1L, y = 1)
dt2 <- data.table(x = 1.5, z = 2)
dt1[dt2, on = "x", verbose = TRUE]
# Calculated ad hoc index in 0 secs
# Coercing double column i.'x' to integer to match type of x.'x'. Please avoid coercion for efficiency.
# Starting bmerge ...done in 0 secs
#     x y z
# 1: 1 1 2
# 1: 1 1 2

merge(dt2, dt1, by = "x")
#    x y z
# 1: 1 1 2
```
The reason is the blind coercion to integer of dt2$x during `bmerge`. 
We should adopt the base R loic of `merge.data.frame` where the join columns get coerced to the 'highest' involved type (https://github.com/wch/r-source/blob/e690b0d6998dfbc360f0fa14492eb8648df20949/src/main/unique.c) Lines 902ff:
```
    /* Coerce to a common type; type == NILSXP is ok here.
     * Note that above we coerce factors and "POSIXlt", only to character.
     * Hence, coerce to character or to `higher' type
     * (given that we have "Vector" or NULL) */
    if(TYPEOF(x) >= STRSXP || TYPEOF(table) >= STRSXP) type = STRSXP;
    else type = TYPEOF(x) < TYPEOF(table) ? TYPEOF(table) : TYPEOF(x);
    PROTECT(x	  = coerceVector(x,	type)); nprot++;
PROTECT(table = coerceVector(table, type)); nprot++;
```

MarkusBonsch:
Closes #2592 

**Main purpose**
The old behaviour of joining could lead to severe bugs when joins with different column types were involved (see #2592). This has been fixed, namely:
* Joining on numeric to integer doesn't lead to weird matches of 1 == 1.5 anymore.
* Joining on factor to integer throws an error now and doesn't confusingly join the integer to the factor level number anymore.

**Changes**
* The whole process of type checking and conversion in `bmerge.R` has been overhauled according to the suggestions in the second comment of #2592. 
* I had to move the calculation of `xo` (the order in the `x` `data.table`) into bmerge after all type conversions are over. Otherwise, `xo` would not reflect the latest state after all modifications and problems similar to #2881 could occur.

**Tests**
* All existing tests are passing without adaptation.
* 106 new tests have been added to check the type conversion routine extensively.

**Others**
* Originally, the idea was to add warnings on type conversions during joins. Unfortunately this is infeasible for now since it would make several hundred tests fail. For example, such conversion happens in `fread` internally. This would need to be cleaned before.
* @st-pasha  suggested that type conversion should be symmetric between `x` and `i` (see his post below). While I agree, this is IMHO infeasible for now, since it would introduce breaking changes. 
* @jangorecki suggested to implement warnings and symmetric type conversion as an option. I am not sure, what the benefit would be since I don't work with `data.table` options myself. So I didn't implement so far.


MarkusBonsch:
The decreased code coverage is due to a section, where I  implement the behaviour for columns where all of the following return FALSE: `is.logical`, `is.integer`, `base::is.double`, `is.character`, `is.factor`. 
I haven't managed to come up with an example that satisfies all of these, so maybe this part should be removed. I tried raw or complex vectors, but these fail earlier in `forder`.

mattdowle:
Looks great and much more robust.  New tests are extensive!
1. How about defining a data matrix in the code, something like this : 
```
i type down rows
x type along cols
                1     2     3     4   ...
1. logical      y     w     w     w
2. integer      e     y     w
3. intAsReal    e     w     w
4. reallyReal   e    cw           y
5. ...
```
where y = types match as expected in diagonal,  e = error,  w = no coerce but warn,   cw = coerce and warn
Then the code would look up the behaviour from the matrix and do that action code.   That would be easier to read, debug and maintain in future; e.g. adding new types in future.  For example, in writing this comment I looked at position [3,3] and changed it from "y" to "w". Since, although the types match in that case we'd want to suggest to the user they change both sides from `double` to `integer` for space and time efficiency.

2. For the "other" type, maybe that should just be error?  What happens now if either or both `typeof()=="complex"` for example?  I'd expect/hope error.  That error could be tested and then the coverage would be 100%.

MarkusBonsch:
Great, I like your suggested behaviour very much!
* Concerning 1: I will update so that the matrix is used for lookup
* Concerning 2: I will implement that any column that `!is.integer & !base::is.double& !is.character & !is.factor & !inherits(col,"integer64")` throws an error. Currently, complex column in x throws an error before arriving at bmerge in `forder` when attempting to create an index. Maybe, complex in i will work to cover the code in unit tests, I will test that.
* Concerning 3: I will update the NEWS.md

MarkusBonsch:
I have refactored the code according to @mattdowle 's proposals with 2 exceptions:
1. I had difficulties defining the warning message for the flag`w` for 'no coercion but warning': what should it warn about? Would it only be for intAsReal columns as suggested by @mattdowle? Then it could only be entered at very specific positions in the matrix. This is not good from my perspective. Therefore, I would propose to treat this warning seperately, outside the matrix. After further thought, I decided not to implement it at all for a very simple reason: I don't believe that `bmerge` is the correct place to warn users about storing integers as integers. If that is intended, it should happen on creation of intAsReal columns in my opinion. I am very open for other suggestions on this topic. The current possible coercion strategies are the following:
```
  ##------------------|-----------------------------------------------------
  ## y (yes):         | no coercion necessary, types match
  ##------------------|-----------------------------------------------------
  ## e (error):       | throw an error because of incompatible types
  ##------------------|-----------------------------------------------------
  ## ci (coercion i): | coerce the column in i to the type of the column in x
  ##------------------|-----------------------------------------------------
  ## ciw:             | same as above, but with warning.
  ##------------------|-----------------------------------------------------
  ## cx (coercion x): | coerce the column in x to the type of the column in i
  ##------------------|-----------------------------------------------------
  ## cxw:             | same as above, but with warning.
  ##------------------|-----------------------------------------------------
```
2. Originally, we wanted to issue warnings to users on joins with non-matching types (currently, there are only verbose message warnings). I did not do that so far since this would affect > 100 existing unit tests, including cases, where internal `data.table` code would cause warnings. So, before doing so, I would propose to clean the code and tests for unnecessary type mismatches and then implementing the warnings, which is as simple as replacing `ci` and `cx` with `ciw` and `cxw` in the type conversion matrix respectively (see table under 1.).

R/bmerge.R
```diff
@@ -10,83 +10,239 @@ bmerge <- function(i, x, leftcols, rightcols, io, xo, roll, rollends, nomatch, m
...
+    if(sourceType %chin% c("logical", "integer", "double") &
+       to         %chin% c("logical", "integer", "double")){
+      ## for these classes, we can do mode coercion to retain other class attributes, e.g. IDate
+      ## identical types have been caught above
+      out       <- x
+      mode(out) <- to
+    }
+    else {
+      ## we need as.'to'() conversion
+      converter <- match.fun(paste0("as.", to))
```
jangorecki: there are just few choices here, why not just few `if (...) as.x else if (...) as.y else if ...`? making such converted make perfect sense if we want to support arbitrary coercion.

st-pasha:
Should the type conversion matrix be symmetrical* (assuming `ci <-> cx` exchange upon transpose)? That is, do we expect that for any two types `A`, `B` the following should hold:
```
  [A, B] = y    <=>  [B, A] = y
  [A, B] = e    <=>  [B, A] = e
  [A, B] = ci   <=>  [B, A] = cx
  [A, B] = ciw  <=>  [B, A] = cxw
```
Currently it doesn't, for example there's `ct["integer", "logical"] = "e"` whereas `ct["logical", "integer"] = "ci"` (there are other mismatched pairs like this too).

MarkusBonsch:
You are right, @st-pasha, I also think that this would be the ideal behavior. However, this would alter existing behavior beyond fixing bugs IMHO. That is why I didn't implement it so far. If there is an agreement to change the behavior to be symmetrical as suggested by you, I am more than willing to implement. Note, however, that very likely existing tests have to be updated.

R/bmerge.R
MarkusBonsch: I wanted to be ready for other `as.xxx` conversion, but if you want, I can change.

jangorecki: Is there any "other" on roadmap? I am not aware of any. Still it will need to be added to matrix anyway.

jangorecki:
My suggestion is to add warnings and make matrix symmetric in this PR but as option. This is big breaking change and it might need to wait for more than 1 release cycle to be changed to TRUE by default. We need to release soon to CRAN, so putting this as an option gives us much more time to check revdeps and inform maintainer. This we can do already after release to CRAN so maintainers can smoothly migrate by using option on recent CRAN build. @mattdowle what you think?

MarkusBonsch:
While I like the idea of using this PR to consolidate type conversion behaviour in joins in general, we should remember that its primary goal is to fix two pretty terrible bugs in master:
```
library(data.table)
dt <- data.table(intCol = 1:2, numCol = c(1.5, 2.5), factCol = factor(c("a","a")))
dt[dt, on="intCol==numCol", nomatch = 0L]
#   intCol numCol factCol i.intCol i.factCol
#  1:      1         1.5          a           1           a
#  2:      2         2.5          a           2           a
## 1.5 is joine dto 1 and 2.5 to 2 !

dt[dt, on="intCol==factCol", nomatch = 0L]
#   intCol numCol factCol i.intCol i.numCol
# 1:       a       1.5       a          1         1.5
# 2:       a       1.5       a          2         2.5
## 1 gets joined to "a" since "a" is factor level 1.
```

So maybe, if the decision on further changes needs more discussion, it can be seperated and this PR merged as a bugfix.

R/bmerge.R
MarkusBonsch: No there isn't and you are completely riht about the matrix. If you want me to change the implementation of coerceClass, please indicate briefly, I will then do it.

MarkusBonsch:
Before this can make it to master, I have to investigate, whether we have to take care about the `xo` argument since x may be modified inside bmerge.R now (similar to problem #2881  and solution).

MarkusBonsch:
I have updated the code so that `xo` is now determined inside `bmerge` after all potential type conversions are over. I have updated the top post to reflect the new features of this PR.

mattdowle:
Looks great on first glance; e.g. streamlining like removing `xo` from bmerge feels right. I'll need to go through it a few times in more detail just to understand it myself but I don't see any hurdles.

Can a news item be added please.  It feels like a bug fix, iiuc.

jangorecki:
@mattd As you are going to look at `bmerge`, could you also briefly evaluate how difficult it would be to split current `bmerge` into `bmergeR` and `bmerge` C in future?

MarkusBonsch:
@jangorecki I am not sure, what you are hinting at with `bmergeR` and `bmergeC`. Currently, the R function  `bmerge` is used to take care about type conversions and factor level consolidations. Then, `Cbmerge` is called to do the real merge.

jangorecki:
@MarkusBonsch I mean separation of code as in fread and fwrite, where `funR` is used for R C API, and `fun` is used to operate in plain C. This allows to use those inner function without going through R C API.

MarkusBonsch:
Thank you so much for taking care of `trimws` @mattdowle. I didn't find the time. If anything else pops up, I will try to fix it.

mattdowle:
`ciw` and `cxw` aren't used anywhere in the ct matrix so there isn't any coverage on the warnings. Maybe the warning code can be dropped then,  or if not, there should be entries using them.  Still looking ...

PR #2738, opened 2018-04-12
Align declaration and definition of bmerge_r

michaelquinn32:
The parameter names in the declaration of bmerge_r (line 38) differs from its definition (line 212). This cleans that up.

### DIFF
--- a/src/bmerge.c
+++ b/src/bmerge.c
@@ -35,7 +35,7 @@ static double roll, rollabs;
 static Rboolean rollToNearest=FALSE;
 #define XIND(i) (xo ? xo[(i)]-1 : i)
 
-void bmerge_r(int xlow, int xupp, int ilow, int iupp, int col, int thisgrp, int lowmax, int uppmax);
+void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisgrp, int lowmax, int uppmax);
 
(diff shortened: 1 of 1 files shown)


mattdowle:
Good spot. Thank you.

michaelquinn32:
Well, the issue was caught by clang-tidy (not me), which we run on everything at Google. I've seen some repos that run it as a part of travis.
https://github.com/ainfosec/ci_helloworld/blob/7e026cae19cb8bdb4362e6cd7468c724d7c3a569/.travis.yml#L133

Would you find that useful?

jangorecki:
@michaelquinn32 which checks were used by you to detect issues on data.table?

michaelquinn32:
That was caught by:
I caught it with: http://clang.llvm.org/extra/clang-tidy/checks/readability-inconsistent-declaration-parameter-name.html

Since data.table is in C, a lot of the different checks don't apply, IMO. But, this could still be useful:

* bugprone: https://clang.llvm.org/extra/clang-tidy/checks/list.html
* the clang analyzer stuff (exclude the Cpp part) https://clang-analyzer.llvm.org/available_checks.html
* the misc checks
* and the readability checks

You might want to poke around in the links I attached. There could be other applicable stuff that I just missed.

Thanks!

PR #2741, opened 2018-04-12
Measuring memory usage in test.data.table, reduce mem usage in some tests

### LINKED ISSUE #2739: Reduce memory usage of test.data.table
We have been discussing with Matt offline and he suggested to call gc() in test() to measure memory usage.
memtest
Attached graph shows there are 3 tests which uses significantly more memory than others. If possible we should reduce memory use of those.

jangorecki:
Closes #2739
More comments inline.
Setting `TEST_DATA_TABLE_MEMTEST=TRUE` and running `test.data.table` will produce `memtest.csv` and following plot. On Linux of course.
memtest

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -127,6 +127,8 @@ test-r-release-vanilla: # check minimal installation, no suggested deps, no vign
   dependencies:
   - mirror-packages
   - build
+  variables:
+    TEST_DATA_TABLE_MEMTEST: "TRUE"
   script:
     - mkdir -p bus/$CI_BUILD_NAME
     - Rscript -e 'source(Sys.getenv("CI_TOOLS")); if (length(pkgs<-packages.dcf("DESCRIPTION"))) install.packages(pkgs, repos=file.path("file:",normalizePath("bus/mirror-packages/cran")))'
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -35,6 +35,7 @@ test.data.table <- function(verbose=FALSE, pkg="pkg", silent=FALSE, with.other.p
   assign("started.at", proc.time(), envir=env)
   assign("lasttime", proc.time()[3L], envir=env)  # used by test() to attribute time inbetween tests to the next test
   assign("timings", data.table( ID = seq_len(3000L), time=0.0, nTest=0L ), envir=env)   # test timings aggregated to integer id
+  assign("memtest", as.logical(Sys.getenv("TEST_DATA_TABLE_MEMTEST", "FALSE")), envir=env)
   # It doesn't matter that 3000L is far larger than needed for other and benchmark.
   if(isTRUE(silent)){
     try(sys.source(fn,envir=env), silent=silent)  # nocov
@@ -69,6 +70,11 @@ compactprint <- function(DT, topn=2L) {
 
 INT = function(...) { as.integer(c(...)) }   # utility used in tests.Rraw
 
+memory_usage = function() {
+  # returns RSS memory occupied by current R process in MB rounded to 2 decimale places
+  round(as.numeric(system(sprintf("ps -o rss %s | tail -1", Sys.getpid()), intern=TRUE)) / (1024^2), 2)
+}
+
 test <- function(num,x,y=TRUE,error=NULL,warning=NULL,output=NULL) {
   # Usage:
   # i) tests that x equals y when both x and y are supplied, the most common usage
@@ -91,6 +97,7 @@ test <- function(num,x,y=TRUE,error=NULL,warning=NULL,output=NULL) {
     assign("ntest", get("ntest", parent.frame()) + 1L, parent.frame(), inherits=TRUE)   # bump number of tests run
     lasttime = get("lasttime", parent.frame())
     timings = get("timings", parent.frame())
+    memtest = get("memtest", parent.frame())
     time = nTest = NULL  # to avoid 'no visible binding' note
     on.exit( {
        now = proc.time()[3]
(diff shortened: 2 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -2912,7 +2912,7 @@ test(1034, as.data.table(x<-as.character(sample(letters, 5))), data.table(V1=x))
 
   # segfault of unprotected var caught with the help of address sanitizer
   set.seed(1)
-  val = sample(c(1:5, NA), 1e5L, TRUE)
+  val = sample(c(1:5, NA), 1e4L, TRUE)
```
jangorecki: confirmed this test is valid also using 1e4L size

inst/tests/tests.Rraw
```diff
@@ -3627,7 +3627,7 @@ test(1149.2, forderv(numeric(0)), integer(0))
 
 # test uniqlengths
 set.seed(45)
-x <- sample(c(NA_integer_, 1:1e5), 1e7, TRUE)
+x <- sample(c(NA_integer_, 1:1e4), 1e6, TRUE)
```
jangorecki: old test, not possible to git reset to test validity of smaller size

inst/tests/tests.Rraw
```diff
@@ -7120,8 +7120,8 @@ test(1548.1, unique(unlist(lapply(ans1, Encoding))), "unknown")
 test(1548.2, unique(unlist(lapply(ans2, Encoding))), "UTF-8")
 
 # #1167 print.data.table row id in non-scientific notation
-DT <- data.table(a = rep(1:5,3*1e6), b = rep(letters[1:3],5*1e6))
-test(1549, capture.output(print(DT)), c("          a b", "       1: 1 a", "       2: 2 b", "       3: 3 c", "       4: 4 a", "       5: 5 b", "      ---    ", "14999996: 1 b", "14999997: 2 c", "14999998: 3 a", "14999999: 4 b", "15000000: 5 c"))
+DT <- data.table(a = rep(1:5,3*1e5), b = rep(letters[1:3],5*1e5))
+test(1549, capture.output(print(DT)), c("         a b", "      1: 1 a", "      2: 2 b", "      3: 3 c", "      4: 4 a", "      5: 5 b", "     ---    ", "1499996: 1 b", "1499997: 2 c", "1499998: 3 a", "1499999: 4 b", "1500000: 5 c"))
```
jangorecki: Reducing data size here is likely to impact this test. Anyway not able to reproduce on parent commit of this fix, maybe related to R 3.2.0 (used at the time of test) vs R 3.4.4 (used now)?

inst/tests/tests.Rraw
```diff
@@ -7639,17 +7639,19 @@ test(1588.1, dt[f %in% 3:4], dt[3:4])
...
-Encoding(x) <- "latin1"
-xx <- iconv(x, "latin1", "UTF-8")
-y = sample(c(x,xx), 10, TRUE)
-oy = if (length(oy <- forderv(y))) oy else seq_along(y)
-test(1590.4, oy, order(y))
-Encoding(xx) = "unknown"
-y = sample(c(x,xx), 10, TRUE)
-oy = if (length(oy <- forderv(y))) oy else seq_along(y)
-test(1590.5, oy, order(y))
+if (!memtest) {
```
jangorecki: any idea why calling `gc()` and `system()` in each `test()` makes this test failing? @mattdowle

inst/tests/tests.Rraw
```diff
@@ -10648,7 +10650,7 @@ test(1779.2, uniqueN(lt)==1L)
 # some random 1e6 timestamps old defaults vs new methods UTC
 intpx = function(x) as.integer(as.POSIXct(x, origin = "1970-01-01", tz = "UTC"))
 set.seed(1)
-i = sample(seq(intpx("2014-10-12"), intpx("2015-10-12")), 1e6, TRUE)
+i = sample(intpx("2015-10-12")-intpx("2014-10-12"), 1e5, TRUE) + intpx("2014-10-12")
```
jangorecki: reduced data size and also removed generating 3M elements `seq` just to sample from it.

.gitlab-ci.yml
```diff
@@ -127,6 +127,8 @@ test-r-release-vanilla: # check minimal installation, no suggested deps, no vign
   dependencies:
   - mirror-packages
   - build
+  variables:
+    TEST_DATA_TABLE_MEMTEST: "TRUE"
```
jangorecki: this activates memory stats collection, it is now active for fastest test job, from 3min it is now 14min.

inst/tests/tests.Rraw
mattdowle: Great. Knowing that makes it easy to approve -- thanks.

R/test.data.table.R
```diff
@@ -69,6 +71,24 @@ compactprint <- function(DT, topn=2L) {
 
 INT = function(...) { as.integer(c(...)) }   # utility used in tests.Rraw
 
+ps_mem = function() {
+  if (!identical(.Platform$OS.type, "unix")) {
+    warning("data.table internal function 'ps_mem' is designed to work only on Linux, please upgrade your OS and re-run.")
```
mattdowle: 'please upgrade your OS [to Linux]' ... LOL.  I know I've written similar in the past, but I vaguely remember a complaint so I changed it. Better just to remove that part. Also, it's not technically correct because this will work on MacOS too, won't it?  And it should work on Windows 10 where native Linux tools are now included.  To be fully correct, it may be worth checking if `ps` system command with the parameters needed is available and functioning as expecting and if not, then error message would be "this OS does not support `ps -o rss <pid>`. It may be possible to install ps".

R/test.data.table.R
```diff
@@ -127,12 +149,19 @@ test <- function(num,x,y=TRUE,error=NULL,warning=NULL,output=NULL) {
...
+  }
   if (is.null(output)) {
     x = tryCatch(withCallingHandlers(x, warning=wHandler), error=eHandler)
     # save the overhead of capture.output() since there are a lot of tests, often called in loops
   } else {
     out = capture.output(print(x <<- tryCatch(withCallingHandlers(x, warning=wHandler), error=eHandler)))
   }
+  if (memtest) {
+    mem = as.data.frame(as.list(c(inittime=inittime, timestamp=timestamp, test=num, ps_mem(), gc_mem())))
+    fwrite(mem, "memtest.csv", append=TRUE)
```
mattdowle: `fwrite` accepts a raw `list()`. If the items are unnamed, no column names are written.
```R
> fwrite(list(1,2,3))
1,2,3
> fwrite(list(foo=1,2,3))
foo,"",""
1,2,3
```
So, iiuc, it can be just : 
```R
mem = list(inittime, timestamp, num, ps_mem(), gc_mem())
fwrite(mem, "memtest.csv", append=TRUE)
```

inst/tests/tests.Rraw
mattdowle: Good. Looks fine to me, too, to be smaller. Now that CRAN_Release.cmd has --strict-barier together with ASAN, and have seen that work and detect problems now that were missed before, I'm much confident we don't need to "go big" in tests like this.

inst/tests/tests.Rraw
mattdowle: Agreed. Looks fine to reduce this one too.

inst/tests/tests.Rraw
mattdowle: Not sure either. What's the error message?  Maybe there's a true memory fault here that the additional `gc()` tickles out.  One option would be to run `test.data.table()` with memtest on in R-devel compiled with `--strict-barrier` and `ASAN`, as per CRAN_Release.cmd commands.  That has a good chance of giving us a line number at the root cause.  If not, then it's valgrind in an instrumented build of R, possibly with gctorture on too (see commands in CRAN_Release.cmd).
Suggest raising a new issue for that though, to look at separately. No need to hold up this PR because of this.

inst/tests/tests.Rraw
mattdowle: Agreed.  Good.

PR #2742, opened 2018-04-12
Various fread fixes

### LINKED ISSUE #2697: in fread, NAs do not interact properly with whitespace
This small file contains numeric values only, as well as some NAs. 
```
> DT = fread('A,   B,    C\n17,  34, 2.3\n3.,  NA,   1\nNA ,  2, NA \n0,0.1,0', verbose=TRUE)
Input contains a \n or is "". Taking this to be text input (not a filename)
[01] Check arguments
  Using 8 threads (omp_get_max_threads()=8, nth=8)
  NAstrings = [<<NA>>]
  show progress = 1
  0/1 column will be read as boolean
[02] Opening the file
  `input` argument is provided rather than a file name, interpreting as raw text to read
[03] Detect and skip BOM
  \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
  Positioned on line 1 starting: <<A,   B,    C>>
  Detecting sep automatically ...
  sep=','  with 5 lines of 3 fields using quote rule 0
  Detected 3 columns on line 1. This line is either column names or first data row. Line starts as: <<A,   B,    C>>
  Quote rule picked = 0
  fill=false and the most number of columns found is 3
[07] Detect column types, good nrow estimate and whether first row is column names
  Number of sampling jump points = 2 because (59 bytes from row 1 to eof) / (2 * 59 jump0size) == 0
  Type codes (jump 000)    : A7A  Quote rule 0
  'header' determined to be true due to column 2 containing a string on row 1 and a lower type (float64) in the rest of the 4 sample rows
  All rows were sampled since file is small so we know nrow=4 exactly
[08] Assign column names
  After 0 type and 0 drop user overrides : A7A
  Allocating 3 column slots (3 - 0 dropped) with 4 rows
[11] Read the data
  jumps=[0..1), chunk_size=1048576, total_size=46
  jumps=[0..1), chunk_size=1048576, total_size=46
Read 4 rows x 3 columns from 59 bytes file in 00:00.000 wall clock time
[12] Finalizing the datatable
  Type counts:
         3 : string    'A'
=============================
   0.000s ( 18%) Memory map 0.000GB file
   0.000s ( 42%) sep=',' ncol=3 and header detection
   0.000s (  4%) Column type detection using 4 sample rows
   0.000s (  4%) Allocation of 4 rows x 3 cols (0.000GB) of which 4 (100%) rows used
   0.000s ( 32%) Reading 1 chunks (0 swept) of 1.000MB (-2147483648 rows) using 1 threads
   +    0.000s (  4%) Parse to row-major thread buffers (grown 0 times)
   +    0.000s (  2%) Transpose
   +    0.000s ( 26%) Waiting
   0.000s ( 12%) Rereading 1 columns due to out-of-sample type exceptions
   0.000s        Total
Column 2 ("B") bumped from 'float64' to 'string' due to <<  NA>> on row 1
```
Few problems can be seen here:
* `Type codes (jump 000)    : A7A`, meaning the first and the last column were recognized as strings, not numeric. NAs were not properly parsed.
* The data was re-read ("Column 2 bumped from 'float64' to 'string' due to <<  NA>> on row 1"). No re-reads should be happening for values that are within the sampling lines.
* `Number of sampling jump points = 2` it's a trivial thing of course, but why 2 sampling points for a tiny file?
* `Reading 1 chunks (0 swept) of 1.000MB (-2147483648 rows)` : negative number of rows reported.

### LINKED ISSUE #2735: logical01 parameter may prevent column name detection
This behavior is correct:
```
> fread(",A,B\n1,0,1\n2,0,1\n3,1,1\n")
      V1     A     B
   <int> <int> <int>
1:     1     0     1
2:     2     0     1
3:     3     1     1
```

However if I use `logical01=TRUE`, suddenly column names stop being recognized:
```
> fread(",A,B\n1,0,1\n2,0,1\n3,1,1\n", logical01=TRUE)
      V1     V2     V3
   <int> <char> <char>
1:    NA      A      B
2:     1      0      1
3:     2      0      1
4:     3      1      1

```

(explicitly setting `header = TRUE` fixes this but shouldn't be necessary)

mattdowle:
Closes #2735 
Closes #2697


R/fread.R
```diff
@@ -81,6 +81,16 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
   stopifnot(length(skip)==1L, !is.na(skip), is.character(skip) || is.numeric(skip))
   if (skip=="__auto__") skip=-1L   # skip="string" so long as "string" is not "__auto__". Best conveys to user something is automatic there (than -1 or NA).
   if (is.double(skip)) skip = as.integer(skip)
+  stopifnot(is.null(na.strings) || is.character(na.strings))
+  tt = grep("^ +$", na.strings)
+  if (length(tt)) {
+    warning('na.strings contains some blank strings. They have been removed. strip.white=TRUE (default) together with na.strings="" turns any number of spaces into NA in string columns')
+    na.strings = na.strings[-tt]
```
HughParsonage: Should this be 

```r
tt = grep("^\\s+$", na.strings)
if (strip.white && length(tt)) {
  warning('na.strings contains strings consisting only of whitespace, yet strip.white=TRUE. These will be ignored: strip.white=TRUE (default) together with na.strings="" turns any number of spaces into NA in string columns'.)
}
```

*i.e.* not just repeated spaces and not 'blank' strings but 'whitespace-only' strings.

R/fread.R
```diff
@@ -81,6 +81,16 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
...
+  }
+  tt = which(substring(na.strings,1L,1L)==" " || substring(na.strings, nn<-nchar(na.strings), nn)==" ")
+  if (length(tt))  {
```
HughParsonage: `|` not `||` I think. Or alternatively,

```r
if (any(substring(na.strings,1L,1L)==" ") || any(substring(na.strings, nn<-nchar(na.strings), nn)==" ")) {
```

R/fread.R
mattdowle: Well spotted. Yes, will do.

R/fread.R
mattdowle: Well spotted. Thanks, will do.

Issue #586, opened 2014-06-08
fread could accept multiple files


arunsrinivasan:
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2589&group_id=240&atid=978)

As highlighted here :
http://stackoverflow.com/a/15102639/403310

MichaelChirico:
This, like #717, is something that seems to come up all the time.

Perhaps this could be a new (essentially wrapper) function `freadl` -- depends on how heavily we want to overload the `file` argument in `fread`.

Also related to #989.

mattdowle:
Also raised here:  https://stackoverflow.com/questions/41918974/rbindlist-a-small-number-of-large-data-table-s

MichaelChirico:
I guess this is really just #2582 in disguise

PR #2753, opened 2018-04-16
Add `text` argument to fread

### LINKED ISSUE #1423: [Request] add text= argument to fread, to read from multi-line character vectors
`fread()` cannot read from a multi-line text object, but it can read a length 1 character vector. Multi-line vectors are typically created using `readLines()`. My feature request is to support input from multi-line character vectors, as `read.table()` does.

`read.table()` accepts character vector input in two ways: 
1) as a text connection on the file= argument -- `read.table(file=textConnection(textvar,...),...)`, or
2) using the `text=` argument, which wraps the input with a `textConnection()`. 
In both cases, the character vector can be of any length, and is read as text lines.

A work-around for `fread` is to collapse the character vector to length 1, using `paste0()` or `stri_flatten()`, as in

``` R
  paste0(textvar, collapse = "\n")
  stri_flatten(textvar, collapse = "\n")
```

and use the result as input to the fread's `input=` argument. Unless a faster way is immediately obvious, the `text=` argument could be initially implemented using one of these collapse functions. This would be quite convenient when text files need pre-processing before reading with fread..

When connections are implemented in fread ([#561](https://github.com/Rdatatable/data.table/issues/561)) then the text= argument could be implemented as a `textConnection`, as `read.table(text=...)` now does.

HughParsonage:
In light of #2531, it may be useful to add a `text` argument for specialized cases where `input` is known in advance to be large and not a file/system command.

This may have the side-effect of reducing the timings of the test suite, if the text argument is used for small examples in tests.

Closes #1423: I've provided a way for `fread` to accept a multi-length `text` argument, as in `read.table`, as described in that issue page.


R/fread.R
```diff
@@ -1,5 +1,5 @@
...
+fread <- function(input="",file,text=NULL,sep="auto",sep2="auto",dec=".",quote="\"",nrows=Inf,header="auto",na.strings=getOption("datatable.na.strings","NA"),stringsAsFactors=FALSE,verbose=getOption("datatable.verbose",FALSE),skip="__auto__",select=NULL,drop=NULL,colClasses=NULL,integer64=getOption("datatable.integer64","integer64"), col.names, check.names=FALSE, encoding="unknown", strip.white=TRUE, fill=FALSE, blank.lines.skip=FALSE, key=NULL, index=NULL, showProgress=interactive(), data.table=getOption("datatable.fread.datatable",TRUE), nThread=getDTthreads(), logical01=getOption("datatable.logical01", FALSE), autostart=NA)
```
mattdowle: On first glance, I'm thinking `text=` should have no default (consistent with `file=`) and then tested with `missing(text)` rather than `=NULL`?

mattdowle:
Agree `text=` should be added, but for a different reason.  Since #2531 is fixed, I doubt there's a speed issue and I'd be surprised if it speeds up `test.data.table()`.  Also a bit hesitant over `paste0()` of length>1 text= character vector (as per #1423)  ... where and why is the root use-case there? If it's connections, they could be supported directly rather than opening the door to a workaround we will have to support in future.  If it's large `text=` input then that's a potentially very large single string being added to the global character cache.
The compelling reason for `text=` is security, so that folks can provide text input directly but also state that it should never be passed to the system's shell to be run there.  Adding an option to turn off fread sending `input=` to the shell might also be welcome perhaps.

R/fread.R
HughParsonage: OK: I vaguely remember a preference for `=NULL` over missing for new arguments from another PR, but will change.

HughParsonage:
Can't specify a root use-case myself -- simply noticed it as an open issue. Though my hunch (and that's all it is) is that for a lot of "production" or "non-interactive" settings, there may be a need to rely on `fread` correctly interpreting `input` as `text`. Indeed, it may be useful to add a `system_input` argument when `input` must be interpreted as a system command so that a greater range of system commands could be used. 

But it's a minor enhancement-- and maybe the `paste0` hack would attract more pain than gain, so perhaps close for now.

st-pasha:
My 2¢ regarding #1423:

Clearly not features that are suggested ought to be implemented: some are too hard, some are too much of a corner case, some would unnecessary encumber the API, some cannot be implemented without sacrificing performance of the core use-case.

However I feel that a proper place for making such an argument is the issue itself, giving the original author a better ability to defend the utility of their request. Preferably this should be done before the FR is implemented, so as not to cause any grievance to the author of the PR.

So, is #1423 a reasonable feature? I can foresee some uses for it: for example if I have a text file that I need to pre-process somehow. Say, there are comments on some lines that need to be stripped. Or it's a SQL table dump and I need to strip away "insert into mytbl values(". Or some lines contain junk and I want to remove them. In fact, if I have a problematic file, why not fread it with sep='' first, correct the problems, and then fread the corrected DT once again?

@HughParsonage  The only thing I'd suggest is to allow character vector as an explicit `text=` argument only. This is because another possible interpretation of a character vector is that it is a list of files that we need to read (see #586).

HughParsonage:
@st-pasha Sorry I don't understand this:

> The only thing I'd suggest is to allow character vector as an explicit text= argument only. This is because another possible interpretation of a character vector is that it is a list of files that we need to read (see #586).

HughParsonage:
I've added a test to ensure that `fread(c("foo", "bar", "baz"))` errors as expected.

NEWS.md
```diff
@@ -49,6 +49,7 @@ These options are meant for temporary use to aid your migration, [#2652](https:/
     * `skip=` and `nrow=` are more reliable and are no longer affected by invalid lines outside the range specified. Thanks to Ziyad Saeed and Kyle Chung for reporting, [#1267](https://github.com/Rdatatable/data.table/issues/1267).
     * Ram disk (`/dev/shm`) is no longer used for the output of system command input. Although faster when it worked, it was causing too many device full errors; e.g., [#1139](https://github.com/Rdatatable/data.table/issues/1139) and [zUMIs/19](https://github.com/sdparekh/zUMIs/issues/19). Thanks to Kyle Chung for reporting. Standard `tempdir()` is now used. If you wish to use ram disk, set TEMPDIR to `/dev/shm`; see `?tempdir`.
     * Detecting whether a very long input string is a file name or data is now much faster, [#2531](https://github.com/Rdatatable/data.table/issues/2531). Many thanks to @javrucebo for the detailed report, benchmarks and suggestions.
+    * New argument `text` for an inputs which is known to be a literal string of the data.
```
st-pasha: New argument `text` for an input which is known to be a literal string. This can also accept a character vector representing a multi-line input.

R/fread.R
mattdowle: Was this what you were remembering : https://github.com/Rdatatable/data.table/pull/2424#issuecomment-373095530.  That's a good point about wrapping fread.  Maybe I should think again.

HughParsonage: Yes, that's the one. I think we could change `file` to `file=NULL` and test `is.null(file)` without harm.

PR #2755, opened 2018-04-17
`fread` fill=true and sep= provided could still read as 1-column

### LINKED ISSUE #2666: `fread` fill=true and sep= provided could still read as 1-column
Here's a minimal example of some small files that I was trying to read with `fread`:

```
library(data.table)

V1 <- c("A;B;C", "D", "E;F")
V2 <- c("A;B;C", "D", "E")

fread(paste(V1, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE)
#    V1 V2 V3
# 1:  A  B  C
# 2:  D      
# 3:  E  F   

fread(paste(V2, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE)
#       V1
# 1: A;B;C
# 2:     D
# 3:     E
```

Notice that the second file only returns 1 column while I was expecting 3. It seems to be ignoring the `sep` provided and guessing based on the remaining rows. Here are some other "files", one of which also does not work, and one that does:

```
V3 <- c("A;B;C", ";D", "E")
V4 <- c("A;B;C", "D", ";E")

fread(paste(V3, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE)
#       V1
# 1: A;B;C
# 2:    ;D
# 3:     E

fread(paste(V4, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE)
#    V1 V2 V3
# 1:  A  B  C
# 2:  D      
# 3:     E   
```

I tried specifying `colClasses`:

```
fread(paste(V2, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE, colClasses = list(character = 1:3))
# Error in fread(paste(V2, collapse = "\n"), sep = ";", header = FALSE,  : 
#   Column number 2 (colClasses[[1]][2]) is out of range [1,ncol=1]
```

And then tried setting `skip = 0` (which works):

```
fread(paste(V2, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE, skip = 0)
#    V1 V2 V3
# 1:  A  B  C
# 2:  D      
# 3:  E      
```

However, I don't want to set `skip = 0` because then it doesn't seem to work if a `sep` value is not found in the first row:

```
V5 <- c("D", "E", "A;B;C")

fread(paste(V5, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE, skip = 0)
#       V1
# 1:     D
# 2:     E
# 3: A;B;C

fread(paste(V5, collapse = "\n"), sep = ";", header = FALSE, fill = TRUE)
#    V1 V2 V3
# 1:  D      
# 2:  E      
# 3:  A  B  C
```

Two questions:

1. Should `fread` be ignoring a manually specified `sep` value?
1. The documentation says that `skip` defaults to 0, but `formals(fread)$skip` returns `[1] "__auto__"`. Should the documentation be updated to explain what `"__auto__"` represents?

----------------------------

```
sessionInfo()
# R version 3.4.2 (2017-09-28)
# Platform: x86_64-pc-linux-gnu (64-bit)
# Running under: Ubuntu 17.10
# 
# Matrix products: default
# BLAS: /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
# LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1
# 
# locale:
#  [1] LC_CTYPE=en_IN.UTF-8       LC_NUMERIC=C               LC_TIME=en_IN.UTF-8       
#  [4] LC_COLLATE=en_IN.UTF-8     LC_MONETARY=en_IN.UTF-8    LC_MESSAGES=en_IN.UTF-8   
#  [7] LC_PAPER=en_IN.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
# [10] LC_TELEPHONE=C             LC_MEASUREMENT=en_IN.UTF-8 LC_IDENTIFICATION=C       
# 
# attached base packages:
# [1] stats     graphics  grDevices utils     datasets  methods   base     
# 
# other attached packages:
# [1] data.table_1.10.5
# 
# [1] compiler_3.4.2 tools_3.4.2    yaml_2.1.17 
```

mattdowle:
Closes #2666 

The root cause was the fallback to single-column input that happened in all cases.

* reorganized sep/quoteRule detection into different sections depending on whether fill=true/false and whether ncol==1 or >1. Different rules in different situations should be easier now. The numLines and numFields VLAs are no longer needed and there isn't a second step to find the first line again (to avoid inconsistencies there).
* added warning when healing quoteRule is selected in jump 0 sample. (There was already a warning when healing quoteRule is chosen out-of-sample.)  Any invalid files that are auto-healed should produce warning.
* fill=true now uses the longest line from the jump 0 sample (which will most often be the column names on row 1), so long as the first row has more than one field with that sep/quoteRule.  As before, no auto-skip when fill=true.


inst/tests/tests.Rraw
```diff
@@ -3959,11 +3959,14 @@ test(1215,
    data.table(N_ID=175931L, VISIT_DATE="2013-03-08T23:40:30", REQ_URL='http://aaa.com/rest/api2.do?api=getSetMobileSession&data={"imei":"60893ZTE-CN13cd","appkey":"android_client","content":"Z0JiRA0qPFtWM3BYVltmcx5MWF9ZS0YLdW1ydXoqPycuJS8idXdlY3R0TGBtU', REQType=2L)
 )
 test(1216.1, fread('A,B,C\n1.2,Foo"Bar,"a"b\"c"d"\nfo"o,bar,"b,az""\n'),
```
st-pasha: This looks confusing: do you mean to have 
`'A,B,C\n1.2,Foo"Bar,"a"b"c"d"\nfo"o,bar,"b,az""\n'`
or
`'A,B,C\n1.2,Foo"Bar,"a"b\\"c"d"\nfo"o,bar,"b,az""\n'` ?

mattdowle:
> It seems that now quote rules 2 and 3 are considered "improper" , and a warning will be issued whenever one of these rules is employed.

Yes. Other csv readers halt with 'file invalid' message. We should at least have a warning.

> This is something we should be more upfront with in PR description.

It is mentioned in this PR's description, point 2.  Feel free to edit better wording in there as I'm not sure what you mean.

> and in the Changelog

Good point.  Will add something to NEWS.

> Also I couldn't understand what does "self healing quote rule" mean, which was mentioned several times in comments.

"Self-healing" is meant to convey "automatic-fixing".  The idea is that quote rules 2 and 3 deal with invalid files.  Without them, the file would fail to read due to the otherwise invalid quote.

> Note that this could be a breaking change for people who run fread under "warnings are errors" mode.

Yes and that's better behaviour. Such users will want to know that `fread` thinks their file is invalid and an attempt to fix it has been made by data.table. They will likely want to only use fully correct and fully valid formats, so they want fread to warn about any little thing that might be wrong.

Note there is no way to control quote rule currently. Due in part not being sure what its parameter values should take, since the quote rule is still bedding down. It could be "auto.warn" (default),  "auto.silent", or a specific rule by number or name.  Currently user can turn off the new warning with `suppressWarnings()` but that also turns off all warnings from that fread call too. I suggest I create a new feature request for `quoteRule=` to be implemented in future, otherwise we'll never get anything out the door to CRAN.

>  After all, if a user chooses an uncommon sep (e.g. \x01), then it may be perfectly normal not to quote any fields regardless of whether they contain any quote characters or not.

Yes and that's why quote rule 3 exists.  But that is unusual and should be warned about because it's more likely something is wrong in a regular comma-separated-file.  If file is using unusual sep for that reason without any quoted fields, but where the field does contain quotes, user could then specify `quoteRule="noquote"` or perhaps use existing `quote=""`, to turn off the warning, as per FR suggested above. Folks reading that call would then see immediately that an unquoted file was being read.  I suggest we leave that to a later date since it's less common.

inst/tests/tests.Rraw
mattdowle: Good point.  I'm not sure what is meant by that extra \ in that test before the `\"c`.  It's superfluous since opening the string `'` automatically escapes any `"` inside the string.  I checked that as follows : 
```
> identical('A,B,C\n1.2,Foo"Bar,"a"b"c"d"\nfo"o,bar,"b,az""\n',
+           'A,B,C\n1.2,Foo"Bar,"a"b\"c"d"\nfo"o,bar,"b,az""\n')
[1] TRUE
```
So out of the two choices you present, the first one is what it does.  Maybe the second one is what was intended.

I'll add them all in, with a comment alongside.

mattdowle:
@st-pasha  Have addressed your comments and this PR is complete I hope. Please re-review.

PR #2757, opened 2018-04-17
branch CI jobs turned off for travis and appveyor, closes #2743

### LINKED ISSUE #2743: Remove branch CI and keep just pr CI
When PR is checked, there are 4 jobs (2 AppVeyor for /branch and /pr,  and 2 for Travis for /branch and /push).
Branch means the branch is tested, I assume.  It might be possible that the branch is behind master which is why this might be useful.
/pr and /push means master is tested as if PR were merged into master, right?

I've never really seen branch result different to pr/push result, though.  But we have to wait for both to run which doubles the time to see if a PR passes after a push.

When non-project members do PR from a fork in their repo, I noticed (iirc) the PR just does pr/push check and not branch. Which makes sense because there is no branch in the data.table project in that case.  And that works fine and great.

@jangorecki If that's all correct, could we remove the CI for branch?  Keeping 2 out of the 4 CI jobs for /pr (AppVeyor) and /push (Travis) .  And if so, how is that change made?

jangorecki:
Closes #2743
travis changed in setting with travis webui, appveyor in yml file.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,8 @@ init:
 install:
   ps: Bootstrap
 
+skip_branch_with_pr: true
+
 environment:
   global:
     CRAN: http://cloud.r-project.org/


jangorecki:
appveyor still builds two branches here, maybe this needs to be pulled in master to work. Re-check after merge.

PR #2774, opened 2018-04-19
Fsort fix

### LINKED ISSUE #2760: fsort error when decreasing=TRUE
Error occurs when using fsort(x, decreasing=TRUE) 
It seems that internally  'order' cannot be converted to -1 when using: forderv(x, order = !decreasing, na.last = na.last)  

```
# use default examples, error occurs both version 1.10.4.3 or 1.10.5
x = runif(1e6)
system.time(ans1 <- sort(x, method="quick"))
system.time(ans2 <- fsort(x))
fsort(x, decreasing=T)
Error in forderv(x, order = !decreasing, na.last = na.last) : 
  x is a single vector, length(order) must be =1 and it's value should be 1 (ascending) or -1 (descending).
此外: Warning message:
In fsort(x, decreasing = T) :
  Input is not a vector of type double. New parallel sort has only been done for double vectors so far. Invoking relatively inefficient sort using order first.
```
`#` `Output of sessionInfo()`
R version 3.4.1 (2017-06-30)
Running under: Ubuntu 16.04.4 LTS

Matrix products: default
BLAS/LAPACK: /opt/openblas/lib/libopenblasp-r0.2.20.dev.so

locale:
 [1] LC_CTYPE=zh_TW.UTF-8    LC_NUMERIC=C            LC_TIME=lzh_TW          LC_COLLATE=zh_TW.UTF-8 
 [5] LC_MONETARY=lzh_TW      LC_MESSAGES=zh_TW.UTF-8 LC_PAPER=lzh_TW         LC_NAME=C              
 [9] LC_ADDRESS=C            LC_TELEPHONE=C          LC_MEASUREMENT=lzh_TW   LC_IDENTIFICATION=C    

attached base packages:

other attached packages:
[1] data.table_1.10.5

[1] compiler_3.4.1 tools_3.4.1    yaml_2.1.18

### LINKED ISSUE #2772: fsort segfault when vector contains NA
On my machine, a simple fsort of a vector with NAs crashes R (not always but sometimes. There is definitely a problem with memory allocation / access on C level:
```
library(data.table)
x <- c(1, NA)
fsort(x, na.last = FALSE)
# crashes R

MarkusBonsch:
Closes #2772 and closes #2760 

Following changes:
1. `fsort` now tests if `x` contains NAs. If so, it falls back to `forderv` since `Cfsort` can't handle NAs so far.
2. Fixed bug where `fsort` failed with `decreasing = TRUE` argument.
3. More informative warnings. Before, `fsort` would always warn about x being integer if it fell back to `forderv`. Now it issues proper warnings if it falls back because of `decreasing = TRUE` or NA values in x.
4. Added tests for the above mentioned changes.

### DIFF
--- a/R/setkey.R
+++ b/R/setkey.R
@@ -249,8 +249,13 @@ fsort <- function(x, decreasing = FALSE, na.last = FALSE, internal=FALSE, verbos
     # fsort is now exported for testing. Trying to head off complaints "it's slow on integer"
     # The only places internally we use fsort internally (3 calls, all on integer) have had internal=TRUE added for now.
     # TODO: implement integer and character in Cfsort and remove this branch and warning
-    if (!internal) warning("Input is not a vector of type double. New parallel sort has only been done for double vectors so far. Invoking relatively inefficient sort using order first.")
-    o = forderv(x, order=!decreasing, na.last=na.last)
+    if (!internal){
+      if(typeof(x) != "double") warning("Input is not a vector of type double. New parallel sort has only been done for double vectors so far. Invoking relatively inefficient sort using order first.")
+      if(decreasing) warning("New parallel sort has not been implemented for decreasing=TRUE so far. Invoking relatively inefficient sort using order first.")
+      if(na.last)    warning("New parallel sort has not been implemented for na.last=TRUE so far. Invoking relatively inefficient sort using order first.")
+    }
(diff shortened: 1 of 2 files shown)


MarkusBonsch:
No NEWS entry since fsor thas never been anounced anyways

Issue #2070, opened 2017-03-21
[BUG] fread error on particular csv in latest development build


yaakovfeldman:
I have several large csv files. Using the latest development build (windows binary from AppVeyor), fread fails on one of them with error:

`Error in fread("myfile.csv") : 
  A row outside the large sample was i) either too short and fill=FALSE ii) too long or iii) a jump batch didn't end properly. This message will be made specific before release to CRAN.`

This is after the commit 'fread parallel' https://github.com/Rdatatable/data.table/commit/0b3cd65381656330b66c5670c2c044f76e2ac03f

The previous commit and earlier still work fine, so something must have changed in this commit.

The csv has 89205 observations of 9 variables. Unfortunately, it is sensitive. What can I try to help make a reproducible example?

EDIT: anonymised csv in comment below

MichaelChirico:
is it possible to replace all of the sensitive fields with ### or some
other anonymizer?

yaakovfeldman:
I now have a reproducible example! (Anonymising didn't work earlier)
[csv.zip](https://github.com/Rdatatable/data.table/files/858334/csv.zip)

I get the error above with
`fread(myfile.csv)`

mattdowle:
Reproduced - thanks.

mattdowle:
Many thanks for providing this file.  Works for me now.  Please open a new issue if you find anything else.

PR #2785, opened 2018-04-23
Provide backwards-compatibility for easycsv

HughParsonage:


jangorecki:
Why not
```r
getOption("datatable.showProgress", interactive ())
```
I used that option sometimes, only to FALSE :)

HughParsonage:
@jangorecki  Yes, that might be better. I just wasn't sure how deliberate the decision to remove the option was, or whether `interactive()` should trump the option. I seem to remember this change caused quite a lot of difficulty, so I wanted to keep the change minimal with respect to the current dev.

MichaelChirico:
are you able to find the commit/issue in which it was changed? for citation
purposes

HughParsonage:
6342551187394b21f965b040511917387a9c35a2

MichaelChirico:
Hmm, I don't see anything there or in linked #2070 suggesting the need for the argument change

bogind:
I just uploaded a fix to the easycsv functions that use `fread`.
It should work even after the update.
if it won't you may feel free to contribute there, easycsv is relatively small and simple and as far as i'm concerned any help is welcome.

mattdowle:
Thanks all for efforts here. I don't recall why I dropped that option. And as you found too, I can't see anything in the history either. Did I change `showProgress` default around the same time that I (erroneously) thought that the fread multi-threaded crash in RStudio was to do with the progress bar?  Anyway, @jangorecki's suggestion looks best to me.  I may have been motivated at the time to remove options because of the overhead of `getOption()` when called a lot in a loop.  But that's not a concern for fread/fwrite and also, more recent efforts to improve getOption speed in R itself have been largely successful.  So the default argument being a call to getOption including default=interactive() directly there is clearest and most helpful to users.  I don't anticipate any disagreement given the comments above so I'll change now and merge.

mattdowle:
With my change just now,  easycsv will break again.  Because it calls `getOption("datatable.showProgress")` which returns NULL and I've taken away the branch Hugh added to accept NULL.  I've done this because @bogind has kindly written above that they have changed `easycsv`. Changing easycsv in this case is the best thing to do because data.table is unusual for setting options to a default value when data.table loads, so packages should not assume that by using `getOption()` without passing `default=`.   This brings us into line with correct and common usage of `getOption()`.  The main thing is that `args(fread)` and `?fread` are now very clear how `showProgress` is set and what the default is. The argument and its default are checked by `R CMD check` to be consistent in `fread.R` and `fread.Rd`, too.

mattdowle:
Same applies for package SIRItoGTFS, if that's ok with you @bogind?

bogind:
@mattdowle  fine with me. 
it will get a fix too in the next couple of days.

Issue #2778, opened 2018-04-21
rolling functions, rolling aggregates, sliding window, moving average


jangorecki:
To gather requirements in single place and refresh ~4 years old discussions creating this issue to cover _rolling functions_ feature (also known as _rolling aggregates_, _sliding window_ or _moving average_/_moving aggregates_).

## rolling functions

- [x] rollmean
- [x] rollsum
- [x] rollmin (#7299)
- [x] rollmax (#7264)
- [x] rollmedian (#7315)
- [x] rollprod (#7299)
- [ ] rollsd (#7361)
- [ ] rollvar (#7361)
- [x] rollapply (user provided FUN) (rewritten for more features in #7272)

## features

- [x] multiple columns at once
- [x] multiple windows at once
- [x] multiple columns and multiple windows at once
- [x] atomic vectors input and single window returns atomic vectors
- [x] various length list of vectors
- [x] align: left/center/right
- [x] handling NAs
- [x] handling Inf (#7264)
- [x] fill constant
- [x] long vector support
- [x] partial window support (#4968, PR #7264)
- [x] adaptive window support
- [x] use openmp to parallelize calculation of multiple columns/windows
- [x] rounding error correction
- [x] timing in verbose mode from parallel region (blocked by ~#3422~, #3423)
- [x] partial & adaptive support (#7264)
- [x] `give.names` argument, same as `shift` has (#7264)
- [x] frollapply `by.column=FALSE` (issue #4887, PR #7272)
- [x] unevenly spaced time series (#3241, PR #7291)

jangorecki:
Proposed `rollmean` implementation, simplified.

```r
x = data.table(v1=1:5, v2=1:5)
k = c(2, 3)
```
```
i - single column
j - single window
m - int referring to single row
w - current row's sum of rolling window
r - answer for each i, j
```
```
for i in x
  for j in k
  r = NA_real_
  w = 0
    for m in 1:length(i)
      w = w + i[m]
      w = w - i[m-j]
      r[m] = w / j
```

MichaelChirico:
yes, and many more rolled functions follow the same basic idea (including
rolling standard deviation/any expectation-based moment, and any function
like rollproduct that uses invertible * instead of + to aggregate within
the window

On Apr 21, 2018 12:03 PM, "Jan Gorecki" <notifications@github.com> wrote:

Proposed rollmean implementation, simplified.

x = data.table(v1=1:5, v2=1:5)k = c(2, 3)

[...]

PR #2795, opened 2018-04-24
[WIP] Rolling functions: rollmean

jangorecki:
#2778
Opening PR so I can put in-line comments.
`roll` branch can be installed from:
```r
install.packages("data.table", repos="https://jangorecki.gitlab.io/data.table")
```


.gitlab-ci.yml
```diff
@@ -190,7 +190,7 @@ integration: # merging all artifacts so multiple deploy jobs can build same repo
     - linux
   only:
     - gl-ci-upgrade
-    - master
+    - roll
```
jangorecki: this was only added for development to render manual into html https://jangorecki.gitlab.io/data.table/library/data.table/html/roll.html
to be removed before merge

src/roll.c
```diff
@@ -0,0 +1,143 @@
...
+    error("na.rm must be TRUE or FALSE");
+  
+  ans = PROTECT(allocVector(VECSXP, nk * nx)); protecti++;                // allocate list to keep results
+  
+  thisfill = PROTECT(coerceVector(fill, REALSXP)); protecti++;
+  
+  if (!LOGICAL(adaptive)[0]) {                                          // adaptive=FALSE
+    if (salign == RIGHT) {                                              // align right scenario
+      for (i=0; i<nx; i++) {                                            // loop over columns
+        this = AS_NUMERIC(VECTOR_ELT(x, i));                            // extract column/vector from data.table/list
```
jangorecki: I haven't seen `AS_NUMERIC` used in other places in data.table? Is there any better way to handle integers here? Most commonly double will be as input, so if `AS_NUMERIC` on double is expensive then better to make check and apply only for integers (if any).

src/roll.c
```diff
@@ -0,0 +1,143 @@
...
+  if (!isLogical(narm) || length(narm)!=1 || LOGICAL(narm)[0]==NA_LOGICAL)
+    error("na.rm must be TRUE or FALSE");
+  
+  ans = PROTECT(allocVector(VECSXP, nk * nx)); protecti++;                // allocate list to keep results
+  
+  thisfill = PROTECT(coerceVector(fill, REALSXP)); protecti++;
+  
+  if (!LOGICAL(adaptive)[0]) {                                          // adaptive=FALSE
+    if (salign == RIGHT) {                                              // align right scenario
+      for (i=0; i<nx; i++) {                                            // loop over columns
```
jangorecki: We could merge two for loops into one, so we can parallelize on columns and on window sizes, not just on one of them.

R/roll.R
```diff
@@ -0,0 +1,5 @@
+rollmean <- function(x, n, fill=NA, align=c("right", "left", "center"), na.rm=FALSE, adaptive=is.list(n)) {
+  align = match.arg(align)
+  ans = .Call(Crollmean, x, as.integer(n), fill, align, na.rm, adaptive)
```
jangorecki: this `as.integer(n)` needs to be pushed to C to handle `is.list(n)`

man/roll.Rd
```diff
@@ -0,0 +1,70 @@
+\name{roll}
```
jangorecki: html preview of latest manual in roll branch (in my gitlab repo)  on: https://jangorecki.gitlab.io/data.table/library/data.table/html/roll.html

src/roll.c
```diff
@@ -0,0 +1,143 @@
+#include "data.table.h"
+#include <Rdefines.h>
+
+static SEXP rollmeanVectorRaw(SEXP tmp, SEXP this, R_len_t xrows, R_len_t thisk, SEXP thisfill, SEXP narm) {
```
jangorecki: This function is wrapped into openmp from `rollmean` function. Is it fine to pass arguments as is or should be passed as pointers?

jangorecki:
To document performance before changing to less safe implementation as default. Currently NAs are carefully handled, consistent to `mean`. Most of other implementations of rolling functions uses less safe implementation asssuming that users knows there are no NAs in input.
Using 37a2c6481f6f16eb49799817ad1c174d32544290, running R 3.4.4 on low-spec laptop.

```r
set.seed(100)
nx = 1e2
n = 1e1
x = rnorm(nx)
  times=10, check=function(x) all(sapply(x[-1L], function(xx) isTRUE(all.equal(x[[1L]], xx)))),
  TTR = TTR::runMean(x, n),
  caTools = caTools::runmean(x, n, alg="fast", endrule="NA", align="right"),
  RollingWindow = RollingWindow::RollingMean(x, n)[,1L],
  data.table = data.table::rollmean(x, n)
  , RcppRoll = RcppRoll::roll_mean(x, n, na.rm=FALSE, fill=NA, align="right")
  , zoo = zoo::rollmean(x, n, fill=NA, align="right")
)
```

x 1e2, n 1e1
```
Unit: microseconds
          expr     min       lq      mean    median       uq      max neval
           TTR 560.091  684.040  690.2129  711.1665  734.827  764.367    10
       caTools 160.954  169.320  236.4268  223.8325  275.615  380.126    10
 RollingWindow 133.164  141.343  214.2981  230.5745  249.513  302.420    10
    data.table  44.888   76.030   85.4691   81.8265   99.339  117.204    10
      RcppRoll  67.867   70.629  120.9170   99.7220  163.390  268.299    10
           zoo 986.585 1011.368 1119.9627 1124.7945 1188.202 1302.189    10
```

x 1e3, n 1e1
```
Unit: microseconds
          expr     min      lq     mean   median      uq     max neval
           TTR 637.999 730.146 760.2156 785.1325 800.233 834.724    10
       caTools 193.506 206.794 229.7091 219.9095 245.167 304.726    10
 RollingWindow 172.316 174.936 235.4080 259.0965 274.026 312.051    10
    data.table  58.126  72.665 100.2140  95.3750 120.688 148.950    10
      RcppRoll  83.899  85.076 116.5263 111.0480 152.396 161.810    10
```

x 1e4, n 1e2
```
Unit: microseconds
          expr      min       lq      mean    median       uq       max neval
           TTR 1412.708 1429.343 9173.8234 1752.2830 1878.349 76848.760    10
       caTools  619.508  667.867 1106.3960  716.9155 1011.104  3979.661    10
 RollingWindow  666.840  706.603  755.6150  725.5880  753.821  1082.872    10
    data.table  182.739  196.613  258.8993  267.6015  283.145   341.743    10
      RcppRoll 1421.062 1483.873 1525.9808 1500.8180 1590.647  1678.830    10
```

x 1e5, n 1e2
```
Unit: milliseconds
          expr       min        lq      mean    median        uq      max neval
           TTR 10.116153 10.191691 19.241091 13.213800 14.249784 81.34873    10
       caTools  6.248023  6.694888 16.165964  8.171910 10.804314 84.87915    10
 RollingWindow  6.508730  7.526900  8.293678  8.140531  8.750700 10.39594    10
    data.table  1.602543  1.630726  9.303160  1.656925  2.247274 76.48370    10
      RcppRoll 13.838378 13.850348 21.316412 13.872491 14.426226 87.00335    10
```

x 1e6, 1e2
```
Unit: milliseconds
          expr       min        lq      mean    median        uq      max neval
           TTR  72.03572  80.68157 102.51816  85.53985 143.06142 157.4600    10
       caTools  45.33968  50.17756  72.79733  66.24561  77.62001 126.9556    10
 RollingWindow  49.47656  56.08493  77.89106  61.52251  65.78007 154.7952    10
    data.table  14.50545  15.06085  25.46016  16.49857  20.04283 101.1363    10
      RcppRoll 137.26536 137.59214 143.40246 142.54625 147.71187 152.4923    10
```

x 1e7, 1e3
```
Unit: milliseconds
           TTR 1096.4466 1140.0330 1167.0187 1186.5519 1200.1805 1211.9262    10
       caTools  732.2755  771.1645  776.4271  772.8063  778.3894  855.3379    10
 RollingWindow  813.2730  848.2222  884.8904  860.8890  932.4350  973.1801    10
    data.table  197.3938  198.0064  204.3920  198.3398  205.1418  230.5858    10
```

x 1e6, 1e2 (another machine, 3.5.0)
```
Unit: milliseconds
          expr      min       lq      mean    median        uq       max neval
           TTR 41.03658 47.13120  49.28153  49.84807  50.44624  60.16660    10
       caTools 20.27890 25.56096  25.54016  25.91907  27.04546  27.80501    10
 RollingWindow 24.86482 25.59458  26.60596  26.42391  26.94626  29.06679    10
    data.table 20.02211 23.40855  28.60857  24.81573  25.68302  71.60708    10
      RcppRoll 89.60089 98.71843 100.99648 103.08588 105.02559 107.74189    10
```

x 1e7, n 1e3 (another machine, 3.5.0)
```
Unit: milliseconds
           TTR 356.0238 356.7269 373.9795 381.2254 382.1224 383.2438    10
       caTools 183.3161 184.2282 194.6519 185.3704 208.7500 235.8043    10
 RollingWindow 196.2354 196.8223 200.6811 197.6962 198.9685 226.3088    10
    data.table 196.2237 196.3263 197.4707 196.5529 199.3752 200.1665    10
```

x 1e8, n 1e3 (another machine, 3.5.0)
```
Unit: seconds
           TTR 3.508480 3.520491 3.535235 3.535334 3.549046 3.558227    10
       caTools 1.806459 1.823235 1.831513 1.828431 1.846696 1.850970    10
 RollingWindow 1.933523 1.943957 1.958477 1.953425 1.973955 1.984235    10
    data.table 1.956160 1.956815 1.967360 1.957155 1.973101 2.004486    10
```

x 1e9, n 1e4 (another machine, 3.5.0)
```
Unit: seconds
          expr      min       lq      mean   median        uq       max neval
           TTR 41.26109 74.19614 100.16993 87.32295 122.88115 183.34267    10
       caTools 26.39530 42.57174  67.62508 66.50060  77.90169 129.17666    10
 RollingWindow 22.48919 46.55308  72.76441 58.64682 111.95625 127.41458    10
    data.table 21.24190 22.30101  25.61906 23.50045  23.73178  49.19603    10
```

jangorecki:
Amended implementation for less NA care:
```r
library(data.table)
set.seed(108)
nx = 1e9
n = 1e4
nas = 1e5
x1 = rnorm(nx)
x2 = rnorm(nx)
x1n = copy(x1)
x1n[sample(nx, nas)] = NA
x2n = copy(x2)
x2n[sample(nx, nas)] = NA

system.time(frollmean(x1, n, hasNA=TRUE))
#   user  system elapsed 
# 18.812   1.140  19.951 
system.time(frollmean(x2, n))
#   user  system elapsed 
#  7.852   1.080   8.929 
system.time(frollmean(x1n, n, hasNA=TRUE))
#   user  system elapsed 
# 18.204   1.984  20.182 
system.time(frollmean(x2n, n))
#   user  system elapsed 
# 27.300   1.872  29.172 
```
Optimistic scenario gets nice speed up, pessimistic scenario slow down but still returns correct answer (as opposed to most of other implementations).
`rollmean` renamed to `frollmean`, these commits not yet pushed to GH.

src/roll.c
mattdowle: Correct: I don't recall ever using AS_NUMERIC. If a coerce is needed then `coerceVector()` IIRC and then it's clear a copy of a whole vector is being done.  It's strange to see the AS_NUMERIC()/coerceVector() inside a loop like that in the snippet above.  And VECTOR_ELT() is used for `list()` not vectors,  so this is coercing each item of a list vector it seems.  Oh, this is a loop through each column I guess then,  in which case it makes sense.  Anyway,  that's all that sprang to mind on quick glance.  It's great to be experimenting like this at C level.

src/roll.c
```diff
@@ -0,0 +1,143 @@
...
+  if (!LOGICAL(adaptive)[0]) {                                          // adaptive=FALSE
+    if (salign == RIGHT) {                                              // align right scenario
+      for (i=0; i<nx; i++) {                                            // loop over columns
+        this = AS_NUMERIC(VECTOR_ELT(x, i));                            // extract column/vector from data.table/list
+        xrows = length(this);                                           // for list input each vector can have different length
+        #pragma omp parallel num_threads(MIN(getDTthreads(), nk))
+        {
+          #pragma omp for schedule(dynamic)
+          for (j=0; j<nk; j++) {                                          // loop over multiple windows
+            SET_VECTOR_ELT(ans, i*nk+j, tmp=allocVector(REALSXP, xrows)); // allocate answer vector for this column-window
```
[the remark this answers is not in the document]
mattdowle: Correct: SET_VECTOR_ELT is definitely not thread safe and should never be used inside parallel code. What you observe is expected and correct behaviour; i.e. it might seem fine on small data or even big data with several threads. But as you increase the complexity, it will start to break in random ways.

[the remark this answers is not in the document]
mattdowle: SET_VECTOR_ELT doesn't just assign,  but it updates the reference count on the object being assigned, and reduces the reference count on the object no longer pointed to by that slot.  When a `gc()` comes along, it's easiest to see that problems can occur then, but they can occur before then too. In addition to knowing this and doing things correctly,  as we get more multi-threaded, the test suite surviving `torture()` is getting more important.  This is why pruning the large tests was important,  to have a chance of torture < 24hrs.

src/roll.c
mattdowle: Good thought. But this leaves too much up to OpenMP, I've found.  My current thought.  To date, I prefer parcelling up the work myself and having more loops, in fact.  In the 2-loop version above, we can look at the j=0..nk loop and reason with that. We know one thread will do that exclusively.  Automatic variables inside that loop definition will be private to that thread, with no sharing.   You can't do that sort of reasoning when there's just one big fat loop handed over to OpenMP.  There's no way to stop the programmer doing that, but it isn't the best way to use OpenMP, I believe, currently.

R/roll.R
mattdowle: I'd have to look another time, but your statement sounds correct.

src/roll.c
mattdowle: You're doing well!  Good question.  There has to be an entry point taking SEXP if the function is callable from R; that's clear.  But then when we want to call that from other C code, it is inconvenience to have to construct the SEXP objects just to call that function.   So we have sometimes turned the entry point into a thin wrapper which then passes the regular C variables on to the internal C function.   Then again,  if the function is highly iterated, it may be either inline function or the algo changed to not use a function call at all.   So the answer is you can pass arguments, or C pointers or SEXP pointers,  each situation is different.

mattdowle:
I've commented inline above on the coding questions. Great progress!
Benchmarks:  I didn't like the timings in micro and millseconds because of the scale (as folks here well know) but the more recent timings in seconds I can work with.  It sounds like the roll branch can already handle NAs in a better way than exists already in other packages. Features come first, so that's good, but those features could be contributed to those packages. Then there's speed, second. Speed could also be contributed to other existing packages. Why are we doing this inside data.table? Why are we integrating it instead of contributing to existing packages and using them from data.table? That needs to be answered and my guess is it comes down to syntax (features only possible or convenient if built into data.table; e.g. inside [...] and optimized) and building data.table internals into the rolling function at C level; e.g. froll* should be aware and use data.table indices and key.  If so, more specifics on that are needed; e.g. a simple short example.  If there is no convincing argument for integrating, then we should contribute to the other packages instead.
I've edited a link to the roll branch into the top of #2778, and closed this WIP PR. I've got to focus on getting release to CRAN for now. Let's keep discussion on this feature in one place on that issue. Is there no way to comment on code in a branch before creating a PR?

src/roll.c
st-pasha: I feel having an R function called from multi-threaded code is too dangerous. Constructing/destroying SEXPs will probably lead to memory corruption; but even simple access to SEXP api may have unpredictable consequences if R does something non-trivial under the hood...

mattdowle: @st-pasha This comment feels a little out of place. I don't know why you're writing this here. I've written quite a bit on multi-threaded issues elsewhere in comments in this PR. So I'm not sure your point exactly here.

jangorecki: Solved by using `void` instead of `SEXP`. Function saves results into `tmp` so don't have to return anything.

src/roll.c
```diff
@@ -0,0 +1,143 @@
...
+
+SEXP rollmean(SEXP obj, SEXP k, SEXP fill, SEXP align, SEXP narm, SEXP adaptive) {
+  
+  R_len_t i=0, j, nx, nk, xrows, protecti=0;
+  SEXP x, kl, tmp=R_NilValue, this, ans, thisfill;
+  enum {RIGHT, CENTER, LEFT} salign = RIGHT;
+  
+  if (!length(obj)) return(obj);                                          // empty input: NULL, list()
+  if (isVectorAtomic(obj)) {                                              // wrap atomic vector into list()
+    x = allocVector(VECSXP, 1);
```
jangorecki: there should be PROTECT here and for same check for `k`.

src/roll.c
jangorecki: I believe most of the time one of `nx` / `nk` will be equal to 1. So to calculate multiple windows for same column, or same rolling window for multiple columns. In such cases we might easily lose threaded processing at all.
Merging those two loops won't result into "big fat" loop, as `nx` are `nk` are unlikely to be big numbers (of course there might some use cases to calculate 4000 columns on 4000 window sizes).
Before trying to merge I will look around, maybe there is some canonical way to handle that in openmp. If there isn't then best will be to see timings of two approaches.

src/roll.c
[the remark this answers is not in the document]
jangorecki: solved by separate loop to allocate storage for answer, then in pragma loop only `VECTOR_ELT` is used.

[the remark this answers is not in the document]
jangorecki: solved by separate loop to allocate storage for answer, then in pragma loop only `VECTOR_ELT` is used.

src/roll.c
jangorecki: @mattdowle What if just one column from DT has to be converted
```
SET_VECTOR_ELT(dt, col-1, coerceVector(VECTOR_ELT(dt, col-1)))
```
are this going to update in place?

mattdowle: `coerceVector` allocates a new column.  Then `SET_VECTOR_ELT` assigns item `col-1` of the VECSXP dt to point to the new column.  If there is a type change of this column, then there is no safe way (no way in the R API) to avoid allocating the new column.  If the type size was the same (so, not `int` (4) to `double` (8)), then you could in theory change the type of the vector by reference and then manipulate the bits in this vector. But not really worth it.  The coercion is a contiguous scan so I wouldn't worry about churning through one columns worth of RAM for coercion (which is fairly rare anyway).

jangorecki: @mattdowle there is type size change, from int to double. But how to change that for just one/few columns without affecting the original data? If I will follow the way it is done in `shallow` I need to create new VECSXP fill it will pointers to doubles from old DT, and convert integer to doubles allocating new columns. Is there better way?

mattdowle: Without affecting the original data, yes like in shallow. I'm not aware of a better way, other than coercing int to double on the fly when you need to use those values in that column.

PR #2804, opened 2018-04-28
Print topn rows properly when class=TRUE (#2803)

### LINKED ISSUE #2803: print(dt, class=TRUE) shows only "topn - 1" rows
When `class=TRUE` is provided to `print.data.table()`, only `topn - 1` rows are printed. Is this a bug, or a feature? If the former is the case, as I suppose, the flaw seems to be in the line 87 of `print.data.table.R`.

Minimal example:
```
library(data.table)
print(as.data.table(iris), topn=3L, class=TRUE)
#      Sepal.Length Sepal.Width Petal.Length Petal.Width   Species
#             <num>       <num>        <num>       <num>    <fctr>
#   1:          5.1         3.5          1.4         0.2    setosa
#   2:          4.9         3.0          1.4         0.2    setosa
#  ---
# 148:          6.5         3.0          5.2         2.0 virginica
# 149:          6.2         3.4          5.4         2.3 virginica
# 150:          5.9         3.0          5.1         1.8 virginica
```

`sessionInfo()`

```
R version 3.5.0 (2018-04-23)
Running under: macOS High Sierra 10.13.4

Matrix products: default
BLAS: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRblas.0.dylib

locale:
[1] C/en_US.UTF-8/C/C/C/C

attached base packages:

other attached packages:
[1] data.table_1.10.4-3

[1] compiler_3.5.0
```

Thanks.

heavywatal:
Closes #2803.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -225,6 +225,8 @@ Thanks to @sritchie73 for reporting and fixing [PR#2631](https://github.com/Rdat
 
 40. Attempt to allocate a wildly large amount of RAM (16EB) when grouping by key and there are close to 2 billion 1-row groups, [#2777](https://github.com/Rdatatable/data.table/issues/2777). Thanks to @jsams for the detailed report.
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Would you mind adding a test to be sure this behavior doesn't inadvertently revert in the future? There's some pointers on the [Contributing](https://github.com/Rdatatable/data.table/wiki/Contributing) page

PR #2805, opened 2018-04-28
If column '..x' exists (including prefix in name), it takes precedence for now for backwards compatibility

mattdowle:
To avoid breaking `socialmixr` and several other Bioc packages.
See v1.11.0 rev dep status [here](https://github.com/Rdatatable/data.table/issues/2779#issuecomment-383403641).

Breakage was : 
```
Error: processing vignette 'introduction.Rmd' failed with diagnostics:
..original.lower.age.limit in j is looking for original.lower.age.limit in calling scope,
but a column '..original.lower.age.limit' exists. Column names should not start with ..
```

### DIFF
--- a/CRAN_Release.cmd
+++ b/CRAN_Release.cmd
@@ -359,7 +359,7 @@ R
 .libPaths()   # should be just 2 items: revdeplib and the base R package library
 update.packages(ask=FALSE)
 # if package not found on mirror, try manually a different one:
-install.packages("MIAmaxent", repos="http://cloud.r-project.org/")
+install.packages("<pkg>", repos="http://cloud.r-project.org/")
 update.packages(ask=FALSE)   # a repeat sometimes does more, keep repeating until none
 
 # Follow: https://bioconductor.org/install/#troubleshoot-biocinstaller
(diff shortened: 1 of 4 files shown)


mattdowle:
Merging now to make progress in rev dep checking.  @MichaelChirico (and anyone else available) please still review, post merge, and we'll revert if need be.

CRAN_Release.cmd
```diff
@@ -359,7 +359,7 @@ R
 .libPaths()   # should be just 2 items: revdeplib and the base R package library
 update.packages(ask=FALSE)
 # if package not found on mirror, try manually a different one:
-install.packages("MIAmaxent", repos="http://cloud.r-project.org/")
+install.packages("<pkg>", repos="http://cloud.r-project.org/")
```
MichaelChirico: @mattdowle is this intentional?

MichaelChirico:
LGTM

CRAN_Release.cmd
mattdowle: Yes. MIAmaxent was just an old example. This .cmd file is semi-automatic: I press F5 on each line to send to the currently active R console.

PR #2806, opened 2018-04-28
fread coerce-down error downgraded to ignore & verbose

mattdowle:
To avoid breakage to `tstools`. But ignore & verbose seems better anyway. 
See v1.11.0 rev dep status [here](https://github.com/Rdatatable/data.table/issues/2779#issuecomment-383403641).

Breakage was as follows (due to colClasses="numeric" applying to all columns) :  
```
  adding: tmp/Rtmp1qxRZX/test_ts_2018_04_28.json (deflated 82%)
── 1. Error: imports work (@testWriteTimeseries.R#252)  ────────────────────────
Attempt to override column 1 <<date>> of inherent type 'string' down to 'float64' which will lose accuracy. If this was intended, please coerce to the lower type afterwards. Only overrides to a higher type are permitted.
1: expect_equal(ts, read_ts(csv_wide_name)) at testthat/testWriteTimeseries.R:252
2: quasi_label(enquo(expected), expected.label)
3: eval_bare(get_expr(quo), get_env(quo))
4: read_ts(csv_wide_name)
5: read_ts.csv(file)
6: fread(file, sep = sep, stringsAsFactors = FALSE, colClasses = "numeric")
```

That error is now a verbose message and the request to coerce down is ignored.


mattdowle:
Merging now just to make progress in rev dep checking. @st-pasha please still review, post merge, and we'll revert if need be. Hope ok.

PR #2807, opened 2018-04-29
Let print.data.table return invisible(x)

heavywatal:
`print.data.table` should invisibly return its first argument instead of `NULL`. It is compatible with the standard `print.data.frame` and [tibble's `print.tbl_df`](https://github.com/tidyverse/tibble/blob/master/R/tbl-df.r), and supports pipe-style coding:

```
head_iris = iris %>%
  as.data.table() %>%
  print() %>%
  head(3L) %>%
  print()
```

I would be happy to add a bullet on the NEWS if this PR is accepted. Maybe NOTES section? Thanks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -225,7 +225,7 @@ Thanks to @sritchie73 for reporting and fixing [PR#2631](https://github.com/Rdat
 
 40. Attempt to allocate a wildly large amount of RAM (16EB) when grouping by key and there are close to 2 billion 1-row groups, [#2777](https://github.com/Rdatatable/data.table/issues/2777). Thanks to @jsams for the detailed report.
 
-41. Fix a bug that `print(dt, class=TRUE)` shows only `topn - 1` rows. Thanks to @heavywatal for reporting [#2803]((https://github.com/Rdatatable/data.table/issues/2803) and filing [PR#2804]((https://github.com/Rdatatable/data.table/pull/2804).
+41. Fix a bug that `print(dt, class=TRUE)` shows only `topn - 1` rows. Thanks to @heavywatal for reporting [#2803](https://github.com/Rdatatable/data.table/issues/2803) and filing [PR#2804](https://github.com/Rdatatable/data.table/pull/2804).
 
 
 #### NOTES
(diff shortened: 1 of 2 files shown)


MichaelChirico:
if i'm not mistaken (i may well be) the thinking behind not doing this in the first place is that always returning (e.g.) a 7GB `data.table` will be costly.

heavywatal:
Good point. But an object seems not to be copied until it is modified (in a normal way). Maybe only an pointer/reference to the original object is returned by `return(invisible(x))`.

```
library(data.table)                                         
x = as.data.table(iris)                                     
tracemem(x)                                                 
#> [1] "<0x7f9d00dcc400>"
y = print(x, topn=1L)                                       
#>      Sepal.Length Sepal.Width Petal.Length Petal.Width   Species
#>   1:          5.1         3.5          1.4         0.2    setosa
#>  ---                                                            
#> 150:          5.9         3.0          5.1         1.8 virginica
tracemem(y)                                                 
#> [1] "<0x7f9d00dcc400>"
y[, SPECIES := toupper(Species)]   # in-place; affects x too
print(x, topn=1L)                                           
#>      Sepal.Length Sepal.Width Petal.Length Petal.Width   Species   SPECIES
#>   1:          5.1         3.5          1.4         0.2    setosa    SETOSA
#>  ---                                                                      
#> 150:          5.9         3.0          5.1         1.8 virginica VIRGINICA
y$species = tolower(y$SPECIES)     # copied for the first time!
#> tracemem[0x7f9d00dcc400 -> 0x7f9d02f78148]: eval eval withVisible withCallingHandlers doTryCatch tryCatchOne tryCatchList tryCatch try handle timing_fn evaluate_call <Anonymous> evaluate in_dir block_exec call_block process_group.block process_group withCallingHandlers process_file <Anonymous> <Anonymous> <Anonymous> <Anonymous> do.call saveRDS withCallingHandlers 
#> tracemem[0x7f9d02f78148 -> 0x7f9d02f780d8]: copy $<-.data.table $<- eval eval withVisible withCallingHandlers doTryCatch tryCatchOne tryCatchList tryCatch try handle timing_fn evaluate_call <Anonymous> evaluate in_dir block_exec call_block process_group.block process_group withCallingHandlers process_file <Anonymous> <Anonymous> <Anonymous> <Anonymous> do.call saveRDS withCallingHandlers
```

mattdowle:
There's no speed or memory penalty to returning `invisible(x)`.  Isn't it related to auto-printing and `:=`? I just looked up our FAQ for the first time in a long time (where have all the FAQ numbers gone!!) and there's one there about auto-printing and invisible() [ that was a nightmare at the time]. But that FAQ doesn't help really.

Well, it passes tests, which includes auto-printing (`/tests/autoprint.R`) so on that basis I can't think why not.

Does it pass `test.data.table(with.other.packages=TRUE)` ?
Does it pass (stated dependency) R 3.1.0 ?

MichaelChirico:
^ as matt was writing i was running a quick benchmark to convince myself there's no speed penalty; confirmed on a 15GB `data.table` there's no speed difference.

Can't see why it would fail on `3.1.0`...

PR #2808, opened 2018-04-29
skip= now consistent as if file started on that line, again

### LINKED ISSUE #2786: sep ignored (possible caused by skip-ing to a non-header)
Potential regression identified from reverse dependency checks (#2779):

```
f = paste0('https://raw.githubusercontent.com/HannahVMeyer/',
           'PhenotypeSimulator/master/inst/extdata/genotypes/',
           'genome/genotypes_genome.txt')

data.table::fread(f, skip="Samples:", data.table=FALSE, 
                  sep=" ", colClasses="character", verbose = TRUE)
# Input contains no \n. Taking this to be a filename to open
# [01] Check arguments
#   Using 8 threads (omp_get_max_threads()=8, nth=8)
#   NAstrings = [<<NA>>]
#   None of the NAstrings look like numbers.
#   skip to string = <<Samples:>>
#   show progress = 1
#   0/1 column will be read as integer
# [02] Opening the file
#   Opening file /var/folders/w8/z_8_41zd3f31mlfxlstp44hr0000gn/T//Rtmp6JFSdN/file577b5717007a
#   File opened, size = 351.7KB (360156 bytes).
#   Memory mapped ok
# [03] Detect and skip BOM
# [04] Arrange mmap to be \0 terminated
#   \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
# [05] Skipping initial rows if needed
# Found skip='Samples:' on line 43. Taking this to be header row or first row of data.
#   Positioned on line 43 starting: <<Samples:>>
# [06] Detect separator, quoting rule, and ncolumns
#   Using supplied sep ' '
#   No sep and quote rule found a block of 2x2 or greater. Single column input.
#   Detected 1 columns on line 43. This line is either column names or first data row. Line starts as: <<Samples:>>
#   Quote rule picked = 0
#   fill=false and the most number of columns found is 1
# [07] Detect column types, good nrow estimate and whether first row is column names
#   Number of sampling jump points = 1 because (329408 bytes from row 1 to eof) / (2 * 322821 jump0size) == 0
#   Type codes (jump 000)    : A  Quote rule 0
#   'header' determined to be true because all columns are type string and a better guess is not possible
#   All rows were sampled since file is small so we know nrow=100 exactly
# [08] Assign column names
# [09] Apply user overrides on column types
#   After 0 type and 0 drop user overrides : A
# [10] Allocate memory for the datatable
#   Allocating 1 column slots (1 - 0 dropped) with 100 rows
# [11] Read the data
#   jumps=[0..1), chunk_size=1048576, total_size=329399
# Read 100 rows x 1 columns from 351.7KB (360156 bytes) file in 00:00.004 wall clock time
# [12] Finalizing the datatable
#   Type counts:
#          1 : string    'A'
# =============================
#    0.002s ( 52%) Memory map 0.000GB file
#    0.001s ( 27%) sep='' ncol=1 and header detection
#    0.000s (  0%) Column type detection using 100 sample rows
#    0.000s (  1%) Allocation of 100 rows x 1 cols (0.000GB) of which 100 (100%) rows used
#    0.001s ( 20%) Reading 1 chunks (0 swept) of 1.000MB (each chunk 100 rows) using 1 threads
#    +    0.000s (  6%) Parse to row-major thread buffers (grown 0 times)
#    +    0.001s ( 14%) Transpose
#    +    0.000s (  1%) Waiting
#    0.000s (  0%) Rereading 0 columns due to out-of-sample type exceptions
#    0.004s        Total
```

This part in particular is odd (because ` `) is indeed present on the first line of _data_ (and in fact `header = TRUE` was auto-detected):

> Using supplied sep ' '

> No sep and quote rule found a block of 2x2 or greater. Single column input.

mattdowle:
Closes #2786 as part of v1.11.0 rev dep checks #2779 

As @MichaelChirico found (thanks for help as it speed me up) it was indeed caused by `skip=` to a non-header.  Specifically a row with 1 column name, preceding the data of 2 columns.

It is still a breaking change for `PhenotypeSimulator`, but before this PR, the fix was not clear.  Both possible fixes are now added as test 1909.* using a reduced size version of `genotypes_genome.txt`.


mattdowle:
@st-pasha As before, merging before review, as rev dep checks are pressing. Hope ok. I don't forsee any risk with this one, otherwise I'd wait.  We can always revert later.
For files where `skip=` lands on the column names, there is no change and it works as before. But now if the `skip=` lands on a line which has one too few, or more than one too few column names, then it now works in the regular new way (auto-fill of column names) just as if the file had started on that line in the first place ... easier to reason with and explain, and saves an internal variable (`autoSkip`).
No other tests change other than an esoteric one (1578.5).

Issue #2767, opened 2018-04-18
Test suite hits 2GB limit on 32bit R-devel


mattdowle:
I've just seen this occur, here : 
https://ci.appveyor.com/project/Rdatatable/data-table/build/1.0.2337/job/3rppwcw31ent8g1r

Jan saw it too today here : 
https://github.com/Rdatatable/data.table/pull/2761#issuecomment-382277970

So, despite all our efforts, there's still more to do.

The error I saw is similar to Jan's : 
```
  Running test id 1667.1      Error: memory exhausted (limit reached?)
  Error: memory exhausted (limit reached?)
  Error: memory exhausted (limit reached?)
  Execution halted
```

That message is R's, here : 
https://github.com/wch/r-source/blob/trunk/src/main/memory.c#L2059
and looking at the context there (inside an err_malloc function) suggests AppVeyor instance is declining R's malloc request.  (i.e. it doesn't seem like some kind of stack limit inside R itself).

@jangorecki 
* What is the memory limit of an AppVeyor instance?  I had a quick look and 4GB came up quickly but I'm not sure if that's correct.  If it is 4GB I don't see how we are reaching that.
* We never knew before, but now with your new memtest charts, it seems the max memory used is towards the end of the test suite and reaches 160MB.  If that's right, it seems very low.  I can't imagine AppVeyor's limit being that low.  Certainly a lot less than 4GB and we know it now passes on your 2GB instance.
* I already reduced the memory usage of the non-equi-join tests in PR #2733.  I'm fairly confident that helped a lot and that the memory usage is no longer in that area.
* According to the memtest charts there are 3 large step-ups in memory usage very early in the test suite. If the mem usage table could be sorted by change and the top 10 printed at the end of the test suite then we could see which tests those step-ups are for.  If the data used in those 3 tests is `rm()`'d after those tests, then perhaps more memory will be available for the tests at the end.   However, as I said above, it's strange that the memtest charts show only 160MB being used!  Reducing those 3 tests will be wasted effort if, somehow, the memtest results are not capturing the actual mem usage on AppVeyor.
* A long shot ... might it be that a Suggests package test is running on AppVeyor which eats more RAM, which does not run in the environment used to create the memtest charts.

HughParsonage:
Isn't i386 limited to 2 GB?

mattdowle:
[...]

PR #2810, opened 2018-04-29
Avoided GenomicRanges by writing known results to .Rdata, #2726

### LINKED ISSUE #2726: move GenomicRanges to integration tests
Currently installing suggested dependency GenomicRanges requires in total 11 packages to be installed, including one 17MB package. To keep testing process tiny we need to handle foverlaps tests without GenomicRanges.
Related: https://github.com/Rdatatable/data.table/issues/2740

mattdowle:
Closes #2726 `
Current problems in #2767 point towards GenomicRanges or its dependencies. Removing it, as we wanted to do anyway, might help.


jangorecki:
I would avoid keeping binary data in git but use csv instead but we can amend that after release.

mattdowle:
Why avoid .Rdata?  With .Rdata the object types and attributes are retained; it's easier.

jangorecki:
diff, git diff, head, etc. are not useful for binary data. Binary files are products, while in git (and other version control) it is best to keep sources, like code to produce binary files, or substitute like csv.

mattdowle:
Saw this again on 32bit. Confirms it's not to do with GenomicRanges then : 
```
Running test id 1894.3      
  Running test id 1894.4      Error: memory exhausted (limit reached?)
  Error: memory exhausted (limit reached?)
  Execution halted
```

MichaelChirico:
made it 200 tests further at least 😅

mattdowle:
Merging despite `memory exhausted` error, as that's a different issue : #2767

PR #2847, opened 2018-05-07
improve integration tests, closes #2846

### LINKED ISSUE #2846: integration tests improvements
Integration tests could/should:
- raise failure for job when one of the tests has failed
- print nice summary as tests.Rraw
- be able to collect memory statistics
- do not raise warning in `parallel` package tests when run on single threaded machine

jangorecki:
closes #2846

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -56,6 +56,7 @@ test-r-release: # R-release most comprehensive tests, force all suggests, also i
     _R_CHECK_FORCE_SUGGESTS_: "TRUE"
     OPENBLAS_MAIN_FREE: "1"
     TEST_DATA_TABLE_WITH_OTHER_PACKAGES: "TRUE"
+    TEST_DATA_TABLE_MEMTEST: "TRUE"
   image: registry.gitlab.com/jangorecki/dockerfiles/r-builder
   dependencies:
   - mirror-packages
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -56,6 +56,7 @@ test-r-release: # R-release most comprehensive tests, force all suggests, also i
     _R_CHECK_FORCE_SUGGESTS_: "TRUE"
     OPENBLAS_MAIN_FREE: "1"
     TEST_DATA_TABLE_WITH_OTHER_PACKAGES: "TRUE"
+    TEST_DATA_TABLE_MEMTEST: "TRUE"
```
jangorecki: This will be removed later on to speed up CI as `test-r-release` job already runs integration tests and it is the slowest from all jobs, no point to make it slower with memtest. Now active to check memory usage of integration tests.

jangorecki:
I will put link to summary of all jobs, including memtest.
Memtest on integration tests (other.Rraw) only.
memtest-other

.gitlab-ci.yml
jangorecki: from 17 minutes it is now 32 minutes, will disable that in next PR.

jangorecki: for windows builds it increase time from 6 min to 50 min

PR #2868, opened 2018-05-12
Closes #2791; adds faster second/minute/hour calculations for UTC object

### LINKED ISSUE #2791: Integer arithmetic can speed up date/time functions on objects in UTC seconds
As implemented on the [`fast_hour`](https://github.com/Rdatatable/data.table/tree/fast_hour) branch.

If someone knows a way to retrieve a timezone's offset without resorting to `as.POSIXlt`, we can make this approach fully* flexible as well.

*Well, not _fully_ flexible... there's always weird cases like [this](http://www.nas.gov.sg/blogs/archivistpick/time-zone-adjustment/).

Perhaps we can offer a `use_exact` argument (`TRUE` by default?) which will induce `as.POSIXlt` to do all the edge-case handling.

MichaelChirico:
Closes #2791 

Here's a comparison:

```
fsecond  <- function(x) {
  if (inherits(x, 'POSIXct') && attr(x, 'tzone') == 'UTC') {
    as.integer(unclass(x) %% 60L)
  } else {
    as.integer(as.POSIXlt(x)$sec)
  }
}
fminute  <- function(x) {
  if (inherits(x, 'POSIXct') && attr(x, 'tzone') == 'UTC') {
    #ever-so-slightly faster than x %% 3600 %/% 60
    as.integer(unclass(x) %/% 60L %% 60L)
  } else {
    as.POSIXlt(x)$min
  }
}
fhour <- function(x) {
  if (inherits(x, 'POSIXct') && attr(x, 'tzone') == 'UTC') {
    #ever-so-slightly faster than x %% 86400 %/% 3600
    as.integer(unclass(x) %/% 3600L %% 24L)
  } else {
    as.POSIXlt(x)$hour
  }
}

NN = 1e8

# 2018-05-01 01:23:45 UTC
test = structure(sample(1525137825, NN, TRUE),
                 class = c('POSIXct', 'POSIXt'), tzone = 'UTC')

library(microbenchmark)
# current master as of 2018-05-12 05:40:00 UTC
library(data.table)

mb = microbenchmark(
  times = 10L,
  old_h = hour(test),
  new_h = fhour(test),
  old_m = minute(test),
  new_m = fminute(test),
  old_s = second(test),
  new_s = fsecond(test)
)
print(mb, unit = 's', signif = 2L)
# Unit: seconds
#   expr  min   lq mean median   uq max neval
#  old_h 7.40 8.40 8.30   8.40 8.50 8.8    10
#  new_h 1.50 1.60 1.70   1.60 1.80 2.0    10
#  old_m 7.40 8.30 8.30   8.40 8.50 8.8    10
#  new_m 1.60 1.60 1.70   1.60 1.80 2.3    10
#  old_s 7.80 8.90 8.90   9.00 9.10 9.1    10
#  new_s 0.71 0.74 0.91   0.89 0.98 1.4    10
mbDF = summary(mb)
mbDF[grep('old', mbDF$expr), 'median'] / 
  mbDF[grep('new', mbDF$expr), 'median']
# [1]  5.260573  5.213114 10.169740
identical(hour(test), fhour(test))
# [1] TRUE
identical(minute(test), fminute(test))
# [1] TRUE
identical(second(test), fsecond(test))
# [1] TRUE
```

So 5x, 5x, and 10x speed-up by using the updated function. Also tried on `NN = 1e7` and got similar results (4x, 4x, 10x)

No new tests needed, I believe


R/IDateTime.R
```diff
@@ -252,9 +252,30 @@ as.ITime.times <- function(x, ...) {
 #   lubridate routines do not return integer values.
 ###################################################################
 
-second  <- function(x) as.integer(as.POSIXlt(x)$sec)
-minute  <- function(x) as.POSIXlt(x)$min
-hour    <- function(x) as.POSIXlt(x)$hour
+# if we know the object is in UTC, can calculate the hour much faster
+second  <- function(x) {
+  if (inherits(x, 'POSIXct') && attr(x, 'tzone') == 'UTC') {
```
jangorecki: `attr` can easily return NULL, safer to use `identical(attr(.), "UTC")`

R/IDateTime.R
```diff
@@ -252,9 +252,30 @@ as.ITime.times <- function(x, ...) {
...
+    as.integer(x) %% 60L
+  } else {
+    as.integer(as.POSIXlt(x)$sec)
+  }
+}
+minute  <- function(x) {
+  if (inherits(x, 'POSIXct') && attr(x, 'tzone') == 'UTC') {
```
jangorecki: same as above

R/IDateTime.R
```diff
@@ -252,9 +252,30 @@ as.ITime.times <- function(x, ...) {
...
+    #ever-so-slightly faster than x %% 3600 %/% 60
+    as.integer(x) %/% 60L %% 60L
+  } else {
+    as.POSIXlt(x)$min
+  }
+}
+hour <- function(x) {
+  if (inherits(x, 'POSIXct') && attr(x, 'tzone') == 'UTC') {
```
jangorecki: and of course here

MichaelChirico:
ah yes. and tests

PR #2872, opened 2018-05-13
Update chmatch timings

HughParsonage:


man/chmatch.Rd
```diff
@@ -41,27 +41,30 @@ chgroup(x)
...
 # any problems early as R itself changes and other packages run.
-# The comments here apply when N has been changed to 1e7.
+# The comments here apply when N has been changed to 1e8
+# and were run on 2018-05-13 with R 3.5.0 and data.table 1.11.2.
 N = 1e5
 
 u = as.character(as.hexmode(1:10000))
 y = sample(u,N,replace=TRUE)
 x = sample(u)
                                            #  With N=1e7 ...
```
jangorecki: this line still mention 1e7

HughParsonage:
@jangorecki Yes. I was surprised too -- hence the change! But I'm a bit less surprised now that I've realized those timings come from some four years ago.

mattdowle:
Great update, thanks.  One day, when we don't have such a long list of other things to do, it would be nice to see a graph of performance of ops like this over the years.

PR #2873, opened 2018-05-13
Closes #2870 -- new options for handling milliseconds for as.ITime

### LINKED ISSUE #2870: `as.ITime` should round sub-second values instead of `floor`/truncate
(referencing closed issue #2156 as requested by Matt)

`as.ITime` converts time strings to integers, so it makes sense that any inputs with sub-second precision would be rounded to the nearest second. Unfortunately, it always rounds *down* (probably just truncating the input to discard decimal values), which doesn't seem like the best option to me. 

##
## The issue
For example:
```
> as.ITime( "00:00:10.999" )
[1] "00:00:10"
```

This isn't just being printed that way, the integer behind the scenes agrees with the printed value:
```
> as.numeric( as.ITime( "00:00:10.999" ) )
[1] 10
```

##
## Comparison with `chron::times`
The `chron` package uses numeric instead of integer values, so this isn't an issue there:
```
> chron::times( "00:00:10.999" )
[1] 00:00:11

# multiply here to convert to number of seconds since midnight for comparison
> as.numeric( chron::times( "00:00:10.999" ) ) * 24*60*60
[1] 10.999
```

##
## Comparison with `as.POSIXct`
`as.POSIXct` prints the "truncated" value, but retains the sub-second precision in the numeric value behind the scenes:
```
> as.POSIXct( "1970-01-01 00:00:10.999", tz = "UTC" )
[1] "1970-01-01 00:00:10 UTC"
> as.numeric( as.POSIXct( "1970-01-01 00:00:10.999", tz = "UTC" ) )
[1] 10.999
```

### Something odd I noticed with `as.POSIXct`

Oddly, `POSIXct` seems to truncate sometimes, possibly depending on timezone, and possibly something to do with negative numbers. Not sure why this is, and it's not related to `data.table` at all, just something I noticed when looking at the above, so you might notice it too:
```
> as.POSIXct( "1970-01-01 00:00:10.999", tz = "UTC" )
[1] "1970-01-01 00:00:10 UTC"
> as.POSIXct( "1970-01-01 00:00:10.999", tz = "Australia/Melbourne" )
[1] "1970-01-01 00:00:11 AEST"
```

MichaelChirico:
#2870 Not sure I'm completely satisfied with the API as submitted here, open to changing.

Questions/possibilities:

 - Is `ms = 'ceil'` really useful? will anybody use this (✅ per jan)? if not, can make the argument yes/no (and rename, maybe `round_ms = TRUE/FALSE`?)  🚫 
 - Not sure I used `...` right here, is that the right way to pass `...` down to separate methods? I assume passing `...` to methods is free due to lazyeval.
 - on my machine, test 1102.12 is failing, I couldn't figure out why.  [EDIT: I assume the source of the issue was my having an object called `c` in `.GlobalEnv` prior to testing starting. probably that test could be improved to be robust to this]
 - not fed through `getOption` / didn't include any option for controlling this


MichaelChirico:
@jangorecki my Q is: I included use case for using `ceiling` instead of `floor` (= truncation), do we think this adds value? Or is the question _only_ of truncation vs. rounding?

And then, do we provide this option through `options` or just as a function argument

jangorecki:
For me rounding is less useful actually, I would need myself ceiling and floor, so best to have options for all three.

MichaelChirico:
ok, for certain at least three then. what about this weird issue where the
truncation direction depends on whether the date is before or after Jan 1
1970 UTC?

MichaelChirico:
Just realized I forgot to add anything to the manual. Surprised the checks passed

rossholmberg:
Could the `truncate` option be replaced with `floor`? That would tidy up the inconsistency for negative numbers, and having `floor` and `ceil` as options seems logical.

It's a little slower, but I don't think the difference would be very noticeable on the whole.

MichaelChirico:
floor and as.integer have the same result. the negative problem is a result
of the arbitrary choice of Jan 1 1970 as 0.

rossholmberg:
Are you sure? I believe the behaviour of `floor` and `as.integer` are different for negative numbers:

```
as.integer( -1.1 )
[1] -1

as.integer( floor( -1.1 ) )
[1] -2
```

MichaelChirico:
😅 true. but i still maintain that the behavior is an accident of the arbitrary choice of 0.

Perhaps there are 4 options, then (including `floor` in addition to `as.integer`). `'truncate'` at least maintains consistency with printing of `POSIXct` objects.

rossholmberg:
Agreed re the (semi) arbitrary zero point. Of the options proposed, only `truncate` is affected by that zero point though, since `round`, `floor`, and `ceil` all behave consistently with positive and negative numbers. Given that behaviour, I would think that any user likely to use the `truncate` option would be wise to choose either `floor` or `ceil` instead.

PR #2882, opened 2018-05-15
gc() race in data.table with R-devel

### LINKED ISSUE #2767: Test suite hits 2GB limit on 32bit R-devel

### LINKED ISSUE #2866: memory leak
Only when using the current 3.6.0 nightly build, I found what appears to be a memory leak in data.table

```r
require(data.table)
n <- 2e6
df <- data.frame(a=rnorm(n),
                 b=factor(rbinom(n,5,prob=0.5),1:5,letters[1:5]),
                 c=factor(rbinom(n,5,prob=0.5),1:5,letters[1:5]))
dt <- setDT(df)
print(pryr::mem_used())
fff <- function(aref) {
  ff <- lapply(1:5, function(i) {
    dt2 <- dt[,list(sumA=sum(get(aref))),by=b][,c:=letters[i]]
    dt2
  })
  return(rbindlist(ff))
}
for(i in 1:10) {
  f <- fff("a")
  rm("f")
  gc()
  print(pryr::mem_used())
}
```
I would expect the printed memory usage to stay flat (since `f` is removed and `gc()` is called) but it goes up. Notice that it always takes some time. adding `[,c:=letters[i]]` makes it cause problems faster but is not necessary (you may have to increase 10 to 200 or something)

example output
```r
66.6 MB
66.6 MB
66.6 MB
66.6 MB
87.3 MB
190 MB
```

`#` `Output of sessionInfo()`
```r
R Under development (unstable) (2018-05-10 r74708)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          

attached base packages:

other attached packages:
[1] data.table_1.11.3

[1] compiler_3.6.0   pryr_0.1.4       magrittr_1.5     tools_3.6.0     
[5] Rcpp_0.12.16     stringi_1.1.7    codetools_0.2-15 stringr_1.3.0  
```

[on SO](https://stackoverflow.com/questions/50298822/is-there-a-memory-leak-in-the-data-table-package-in-r-3-6-0)

This is a problem on the CRAN version and the Git Hub version of data.table.

mattdowle:
Closes #2866 
Closes #2767 

The approach of this PR is to ensure ALTREP vectors are not allowed as columns in a DT.  We like and benefit from ALTREP very much in R code such as `[.data.table` where sequence vectors are used a lot.  But as columns in a data.table, ALTREPs are not so appropriate.  Internals like `:=` assign by reference are rewritten on the basis of columns already being materialized (expanded).

`setDT()` now expands ALTREP columns.  The reproducible examples used `setDT()` to create the test data.table because `data.table()` already expanded ALTREP columns (by happy accident) in `CcopyNamedInList`.  That function now checks for ALTREP just in case (as well as MAYBE_REFERENCED as before) just to be safe.

Luke said that ALTREPs may in future be more than just sequence vectors; e.g. distributed arrays that cannot be expanded.  But in that case, data.table will need code changes anyway to deal with such arrays.  If and when that happens, the expansion will fail on such ALTREPs which is reasonable, graceful behaviour; much better than a subsequent gc race at least.

The above is long-term approach with no plans to change; i.e. data.table is unlikely to ever support ALTREP columns in a data.table.

What is short term though, is in all the parallel regions this PR will add checks that no SEXP being used inside the parallel region are ALTREPs (and fail if so). In R-devel, there's only a problem with INTEGER(), REAL() etc on ALTREP vectors.  Those functions are still thread-safe on regular vectors, currently.  This is a short term solution in the interests of getting an update to CRAN which is intermittently in error state on R-devel due to the gc race.  In future we will still take all API use out of parallel regions as Luke suggested.  That involves a new approach around the parallel regions which will take time to work through.

- [x] `setDT()` expands ALTREPs  (`data.table()` already did)
- [x] tests added
- [x] news item
Add checks before all parallel regions that no ALTREPs are present.
All files with parallel regions : 
- [x] subset.c

moved to follow up PR : 
- [ ] between.c
- [ ] freadR.c
- [ ] fwrite.c
- [ ] fsort.c
- [ ] reorder.c


jangorecki:
memtest after this build from R devel Windows
memtest_i386
memtest_x64
looks it is fixed :1st_place_medal:

PR #2901, opened 2018-05-24
Closes #2898, fread error on empty input downgraded to warning

### LINKED ISSUE #2898: Don't error on empty files -- warn & skip
I'm trying to read a file that was output from parquet to csv(s) from Spark.

In its infinite wisdom, Spark created some empty files. So this work flow failed:

```
read_f = list.files('path/to/csvs', pattern = 'csv$', full.names = TRUE)
DT = rbindlist(lapply(read_f, fread))
```

It's kind of a pain to have to single out empty files (basically add the line `read_f = read_f[file.info(read_f)$size > 0]`) when the vast majority of the time this operation works as intended (since it's rare for spark to output empty files) -- is there any reason `fread` can't just warn for such a file and skip?

MichaelChirico:
#2898.

This change will break a few tests around `885`:

 - 885
 - 885.3

As near as I can tell, these tests are identical. I can easily remove these tests once we agree to what now amounts to a change in behavior (admittedly minor). Is there a reason we errored on empty input before?

Similar tests that are _not_ broken are:

 - 885.1: `fread("")` fails. I think this should continue to fail
 - 885.2 `fread()` fails. I think this should continue to fail as well (this test is essentially a test of the default `input = ""`)


R/fread.R
```diff
@@ -26,44 +26,60 @@ fread <- function(input="",file,sep="auto",sep2="auto",dec=".",quote="\"",nrows=
   stopifnot(nThread>=1L)
   if (!missing(file)) {
     if (!identical(input, "")) stop("You can provide 'input=' or 'file=', not both.")
-    if (!file.exists(file)) stop("File '",file,"' does not exist.")
-    if (isTRUE(file.info(file)$isdir)) stop("File '",file,"' is a directory. Not yet implemented.") # dir.exists() requires R v3.2+, #989
+    file_info = file.info(file)
+    if (is.na(file_info$size)) stop("File '",file,"' does not exist or is non-readable.")
```
MichaelChirico: `is.na(file.info(file)$size)` should be equivalent to `!file.exists(file)`, as per my reading of `?file.info`:

> Entries for non-existent or non-readable files will be NA.

Switching to this since there are now three checks, all of which can be run through the single call to `file.info`: (1) file exists? (2) file is directory? (3) file non-empty?

mattdowle:
The title would be better:  "fread error on empty file downgraded to warning.  Is that more accurate, if I understand correctly?
It should return empty data.table (data.table(NULL)) rather than raw NULL, shouldn't it?

MichaelChirico:
yes & yes

MichaelChirico:
ok, it seems we agree the old behavior can be axed, so I'll complete this
PR next time I'm around a laptop

mattdowle:
Based on my experience, I'd personally prefer a warning about empty files.  Those are more often an error than real. For example a process opens a file and then fails later to write data to it, leaving the file utterly empty.  Where I work with data, I normally expect to see some data.
fwrite(data.table(NULL)) error is a bug.

MichaelChirico:
A bug how? meaning correct behavior is to create.file and warn for this
case of output?

Shall I roll into this PR or address separately?

mattdowle:
ok not a bug then.  yes I'd agree correct behaviour would be write empty file with warning.  0-row data is understandable to occur (column names only in a file), but utterly empty data.table() being written is something I'd prefer to be warned about in practice.

MichaelChirico:
@jangorecki should be ready to go, LMK

NEWS.md
```diff
@@ -5,6 +5,10 @@
 
 #### NEW FEATURES
 
+#### NEW FEATURES
```
jangorecki: duplicated header of "new features"

R/fwrite.R
```diff
@@ -51,6 +51,17 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
     showProgress = FALSE
     eol = "\n"  # Rprintf() is used at C level which knows inside it to output \r\n on Windows. Otherwise extra \r is output.
   }
+  if (ncol(x) == 0L && file != "") {
+    # should cover append = TRUE case as well
+    if (file.exists(file)) {
+      warning("Input has no columns; doing nothing.")
```
jangorecki: what if file exists and is not empty and we used `append=FALSE`, file should be overwritten then, isn't it? otherwise there is small incosistency

man/fread.Rd
```diff
@@ -97,7 +97,7 @@ When \code{input} begins with http://, https://, ftp://, ftps://, or file://, \c
 
 }
 \value{
-    A \code{data.table} by default. A \code{data.frame} when argument \code{data.table=FALSE}; e.g. \code{options(datatable.fread.datatable=FALSE)}.
+    A \code{data.table} by default. A \code{data.frame} when argument \code{data.table=FALSE}; e.g. \code{options(datatable.fread.datatable=FALSE)}. \code{NULL} if the target is empty.
```
jangorecki: not `NULL` but `data.table(NULL)`, correct? or for `data.table=FALSE` `data.frame(NULL)`.

NEWS.md
```diff
@@ -5,6 +5,10 @@
...
+
+1. `fread` no longer errors on empty files, and instead returns `NULL`, [#2898](https://github.com/Rdatatable/data.table/issues/2898). This improves compatibility with, e.g., using `fread` + `rbindlist` on a directory in which some files may be empty, since `rbindlist` already skips `NULL` elements.
```
jangorecki: why not `data.table(NULL)`? consistency in returned data type is quite important. This is the reason why `drop=TRUE` arg in `[` is so bad.

R/fwrite.R
jangorecki: Agree it can be painful, so if we sacrifice consistency, could we mention that in warning? something like
```r
"Input has no columns"
if(!append & filesize>0) "empty file has not been materialized to not overwrite existing non-empty file"
"doing nothing."

```

PR #2913, opened 2018-05-29
Avoid tryCatch

### LINKED ISSUE #2910: avoid `tryCatch()` when checking `.datatable.aware`
This line 

https://github.com/Rdatatable/data.table/blob/70db3e48738026388c6487e83ee902ad37256ca5/R/cedta.R#L37

could be re-written to avoid using `tryCatch()`, e.g., 

```
isTRUE(getNamespace(nmspc)$.datatable.aware)
```
This would help with C-level debugging approaches that set a break point on `Rf_error()` and are then faced with handling the spurious errors generated by the `tryCatch()` implementation. It is also cleaner and much faster 

```
f0a <- function(nmspc)
    identical(
        TRUE,
        tryCatch(
            get(".unknown", getNamespace(nmspc), inherits = FALSE),
            error = function(e) NULL
        )
    )

f0b <- function(nmspc)
    identical(
        TRUE,
        tryCatch(
            get(".datatable.aware", getNamespace(nmspc), inherits = FALSE),
            error = function(e) NULL
        )
    )

f2a <- function(nmspc)
    isTRUE(getNamespace(nmspc)$.unknown)
f2b <- function(nmspc)
    isTRUE(getNamespace(nmspc)$.datatable.aware)
```
with
```
> identical(f0a(nmspc), f2a(nmspc))
[1] TRUE
> identical(f0b(nmspc), f2b(nmspc))
[1] TRUE
> microbenchmark(f0a(nmspc), f0b(nmspc), f2a(nmspc), f2b(nmspc))    
Unit: microseconds
       expr    min      lq     mean  median      uq    max neval
 f0a(nmspc) 33.487 34.8515 36.00464 35.3975 35.8820 70.852   100
 f0b(nmspc) 14.297 14.9975 16.15132 15.7670 16.6780 38.982   100
 f2a(nmspc)  2.143  2.5455  2.86360  2.8270  3.1695  5.153   100
 f2b(nmspc)  2.909  3.4030  3.75350  3.6625  3.9715 12.302   100
```

mtmorgan:
Closes #2910

### DIFF
--- a/R/cedta.R
+++ b/R/cedta.R
@@ -38,7 +38,7 @@ cedta <- function(n=2L) {
     (nsname == "base"  && all(c("FUN", "X") %in% ls(parent.frame(n)))  ) || # lapply
     (nsname %chin% cedta.pkgEvalsUserCode && any(sapply(sys.calls(), function(x) is.name(x[[1L]]) && (x[[1L]]=="eval" || x[[1L]]=="evalq")))) ||
     nsname %chin% cedta.override ||
-    isTRUE(asNamespace(nsname)$.datatable.aware)
+    isTRUE(getNamespace(nsname)$.datatable.aware)
   if (!ans && getOption("datatable.verbose")) {
     cat("cedta decided '",nsname,"' wasn't data.table aware. Call stack with [[1L]] applied:\n",sep="")
     print(sapply(sys.calls(), "[[", 1L))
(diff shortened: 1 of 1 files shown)


mattdowle:
Very nice! Thank you.

mattdowle:
No rationale really! `cedta()` was written when R packages didn't have to have a namespace. It probably just hung around since then. Then R 2.14.0 added a default namespace to packages without one which helped a lot.  But we still supported R < 2.14.0 for a few years after that. Time passed and now we've dropped support for R < 3.1.0 so can definitely tidy up things like this.

Issue #1086, opened 2015-03-18
[Request] Allow locking or protecting a data.table from any modifications


ClaytonJY:
This issue arose from Stack Overflow here: https://stackoverflow.com/questions/29085334/lock-or-protect-a-data-table-in-r

I would like to be able to make a data.table unmodifiable. In the comments below my question, two changes to the data.table package are proposed by other users:
1) make modify-by-reference functions respect the `lockBinding()` function
2) make modify-by-reference functions respect a "read.only" attribute. Jan Gorecki has already implemented this for `[.data.table` here: https://github.com/jangorecki/data.table/tree/read.only

Either solution would work for me.

jangorecki:
maybe we can use `drop` param?
```r
# in-place assign by reference
dt[, a := b]
dt[, a := b, drop=TRUE]
# shallow copy of `a`
dt[, a := b, drop=FALSE]
```
This way we can not modify our initial data.table, and have the results materialized as copy. And there is `drop` finally useful.

eantonya:
I think OP wants to prevent semi-accidental modifies, not intentional ones. To achieve what the above `drop` example does, I'd just use `copy` now and imo that's easier/more understandable.

jangorecki:
Good point.
I'm working on package which *injects* `read.only` branch functionality into data.table, so user can stick to data.table and prevent not intentional modification. Still it can be improved for `[, := ]` detection as it now copies the object even if you are not using `:=`. But one extra copy is a cheap trade-off for *locking* DT.

jangorecki:
@ClaytonJY I just pushed that functionality into separate package. It allows you to stick to data.table master. [jangorecki/dtq](https://github.com/jangorecki/dtq) will inject same functionality as mentioned data.table branch.
It is heavy in dev but `read.only` feature is stable. If you want to use just `read.only` feature you might want to turn off the main feature of package by `options("dtq.log" = FALSE)`.

aryoda:
Any news about the state of locking data.tables? I ran into the same requirement, see
https://stackoverflow.com/questions/44995065/r-how-to-make-a-data-table-read-only-without-copying-it-e-g-for-data-validat

aryoda:
This is a (partial?) duplicate of the older https://github.com/Rdatatable/data.table/issues/778

MichaelChirico:
[...]

Issue #778, opened 2014-08-21
Any modification on a locked binding should fail, even if operated by reference.


arunsrinivasan:
At the moment, even if a binding is locked, it can still be modified by reference:

``` R
ee = new.env();
ee$dt = data.table(x=1:5)
lockBinding("dt", ee)
ee$dt$y <- 6:10 ## error
# Error in ee$dt$y <- 6:10 : cannot change value of locked binding for 'dt'
ee$dt[, y := 6:10] ## works, but shouldn't
#    x  y
#1: 1  6
#2: 2  7
#3: 3  8
#4: 4  9
#5: 5 10
```

[Relevant SO post](http://stackoverflow.com/q/25426810/559784).

aryoda:
Is there a chance to implement this feature soon?

It is more R'ish (intuitive) than the reference FR #1086 which also proposes a read-only attribute as a possible alternative...

I have gathered all possible references of Stack Overflow questions here: https://stackoverflow.com/questions/44995065/r-how-to-make-a-data-table-read-only-without-copying-it-e-g-for-data-validat

aryoda:
Another unwanted side effect of the current implementation is, that assigning the `data.table` variable to another new variable (= same memory address) does also modifies the "lockBound" first variable:

```
ee$dt2 <- ee$dt
ee$dt2[, z := 11:15]  # does also modify dt. That shouldn't happen!
ee$dt
# x  y  z
# 1: 1  6 11
# 2: 2  7 12
# 3: 3  8 13
# 4: 4  9 14
# 5: 5 10 15
```

The implementation of this FR should also prevent this side effect.

MichaelChirico:
I'm not sure it's possible to prevent this. I don't see a way to check whether `dt` is locked in the first example (`ee$dt`).

We need to know `bindingIsLocked("dt", ee)`, but from within `[.data.table`, I don't think it's possible to recover, in general, the environment of input `x`.

e.g.

```r
trace(data.table:::`[.data.table`, at=1, quote(message("x's environment: ", format(environment(x)))))
ee$dt[, y := 6:10]
# Tracing `[.data.table`(ee$dt, , `:=`(y, 6:10)) step 1 
# x's environment: NULL
```

In this _particular_ case, we can probably back out `ee` from looking at `substitute(x)`, but it becomes ~impossible very quickly:

```r
foo = function() ee$dt
trace(data.table:::`[.data.table`, at=1, quote(message("substitute(x): ", substitute(x))))
[...]

Issue #2323, opened 2017-08-28
[Request] export shallow or similar


franknarf1:
I have a function, `toopt(th, dat)`. The goal is to minimize the function in terms of its first argument. The second argument is a data.table. I guess this is a pretty common use case.

```
library(data.table)

f = function(dat, th){
    dat[, prob := dpois(x, th) ][]
}

toopt = function(th, dat){
    mydat = data.table:::shallow(dat)
    f(mydat, th)
    mydat[, - sum(log(prob))]
}

nid = 10
nobs = 10
par = 20

DT = data.table(id = seq_len(nid))[, .(x = rpois(nobs, par)), by=id]
optimize(toopt, dat = DT, c(0,100))
mean(DT$x) # the closed-form solution
```

Inside `optim` or `optimize`, the function is called many times as it searches for the optimum.

I'm posting in response to [Matt's SO post and comments below](https://stackoverflow.com/a/45925735/). Sorry if this example is too trivial or abstract. In my actual example, I have a sequence of functions like `f` that add columns (stored in separate functions so I can isolate them for easier testing and to reduce duplication, since I have multiple `toopt` functions that are similar but not identical); and my `DT` has some columns that I don't need for the calculations in `toopt`.

Besides functions passed to `optim`, I also use / want to use `shallow` in other functions, so I can benefit from the efficiency of not making a copy without worrying about whether my input will be messed with (if I didn't use `copy`) or about whether I need to to prune my input table's columns down to the minimum required for the function (if I did use `copy`).

franknarf1:
Besides dtplyr, another recent example [from SO](https://stackoverflow.com/q/57208594):

> So you want to create a new column in a data.table? Any reason this doesn't work: `mtcarsDT[, newcol := myFunc()]`? – Andrew 18 hours ago
> Yes and No. I don't really want to create a column, I just want to view the result of the expression next to the existing columns without having to save the column and then delete it later – Dean MacGregor 18 hours ago

PR #2918, opened 2018-05-31
Locking data.table tables

roberthinch:
There are many open issues (e.g. #778 and #1086 ) on Github and questions on StackOverflow asking for the ability to lock the contents of a data.table from all updates. One suggestion has been to use the attribute .data.table.locked which blocks some update (e.g. := ) to block all updates. Below is an example how setnames could be altered to to do this. I'm new to the data.table code, so I have a few questions:
1. Can the attribute .data.table.locked be used or does it currently have a specific other use?
2. Is there a deeper place in the code where we can block updates as opposed to updating all the R functions?
3. Should lockBinding(dt,env) be extended to set .data.table.locked?

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -2495,6 +2495,7 @@ setnames <- function(x,old,new) {
   # But also more convenient than names(DT)[i]="newname"  because we can also do setnames(DT,"oldname","newname")
   # without an onerous match() ourselves. old can be positions, too, but we encourage by name for robustness.
   if (!is.data.frame(x)) stop("x is not a data.table or data.frame")
+  if (!is.null(attr(x,".data.table.locked")) && attr(x,".data.table.locked")) stop("data.table is locked so names can't be updated")
   if (length(names(x)) != length(x)) stop("dt is length ",length(dt)," but its names are length ",length(names(x)))
   if (missing(new)) {
     # for setnames(DT,new); e.g., setnames(DT,c("A","B")) where ncol(DT)==2


jangorecki:
1. it does have specific use, inside `[.data.table` when using `.SD` object, this attributes is set on `.SD` to prevent from performing operations that should not be performed there/are not yet defined/implemented.

And because of the above, your PR will fail on test 1516
```r
DT <- SRC[, src_cols, with=FALSE][, setnames(.SD, tgt_cols)]
```

roberthinch:
Thanks Jan, so looks like we should introduce a new attribute specifically for this purpose. 

On point 3 I agree, lockBinding() does not completely lockdown members of referenced objects in general  (e.g. in R6 class objects). Having a new function(e.g.  locktable( dt )), is probably best way to turn on locking.

jangorecki:
@roberthinch I used attribute called `read.only` but it was only handled in `[.data.table` to prevent `:=` from editing the source data.table. It is not very efficient, and `set*` function are not handled. This work is in [dtq package](https://github.com/jangorecki/dtq).
If you don't want to alter your source object then `copy` is best way to go. It directly translates to a requirement of retaining source data intact in design of data processing workflows. Whenever you have such requirement you should use `copy`, otherwise it makes more sense to update data in place, for speed and for memory efficiency.
The case of `setnames` is a bit unfortunate because it doesn't modify data but only metadata, so `copy` is relatively expensive here. This hopefully will be handled in https://github.com/Rdatatable/data.table/issues/2323.

roberthinch:
@jangorecki  The reason for wanting to lock the table is to avoid the expensive copy() on large reference tables. 

How come your read.only work has not been integrated in to the main data.table package? Seems like this is the start of the solution to the issues.

jangorecki:
There are many solutions that could be included but haven't been. Most of the time it is because we are aware they could be implemented more efficiently and we just want to make it once, properly. Still PRs are very welcome, even if they are not so perfect. They can push our work forward towards some features, as it was for example with `fwrite`.

Issue #2880, opened 2018-05-15
as.IDate ignores origin


DavidArenburg:
Searched for this in the issues but couldn't find it. It seem like `as.IDate` ignores `origin`
```r
library(data.table) #v1.11.2
as.IDate(1000, origin = "1930-01-01")
## [1] "1972-09-27"
as.Date(1000, origin = "1930-01-01")
## [1] "1932-09-27"
```

DavidArenburg:
I've took a look in the source code and I don't see any attention to `origin` in `as.IDate.numeric`. I think we could make a small change using something like

```r
as.IDate.numeric <- function(x, origin = "1970-01-01", ...) as.IDate(origin, ...) + x
as.IDate(1000)
## [1] "1972-09-27"
as.IDate(1000, origin = "1930-01-01")
## [1] "1932-09-27"
```

This also won't brake compatibility with the old behavior

tdeenes:
Found the same bug today. @DavidArenburg , have you considered submitting a PR for this?

PR #2923, opened 2018-06-03
Fix idate

DavidArenburg:
Closes [#2880](https://github.com/Rdatatable/data.table/issues/2880)


DavidArenburg:
Hmmm... It seems I messed something up...

mattdowle:
I've invited you to be project member David.  Welcome!  Once you accept you can create branches directly in Rdatatable.  Then I could have pushed to this branch right here and sorted it out.
I'll delete src-i386 directory from this branch locally and commit to master.  Think that src-i386 is the source of the large diff.  I'll try and squash to avoid the large diff being added to history.

mattdowle:
Squashed and merged in https://github.com/Rdatatable/data.table/commit/c3562cadcef3a45e62427722b8d80b42370ec7e7.  Fix-up was needed outside this PR so it didn't merge this PR per se.

PR #2935, opened 2018-06-14
Fix documentation of as.matrix to reflect current behaviour

### LINKED ISSUE #2930: as.matrix.data.table incorrect results when nrows == 1

``` r
library(data.table)
d<-data.table(name=c("bob","joe"),age=10:11,IQ=c(130,140))
as.matrix(d, 1) 
#>     age  IQ
#> bob  10 130
#> joe  11 140

as.matrix(d[1, ], 1)
#>   name  age  IQ   
#> 1 "bob" "10" "130"
```

<sup>Created on 2018-06-13 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).</sup>

The problem is that the rownames argument to as.matrix.data.table is not being respected as documented:

> `rownames`: optional, a single column name or column index to use as the
          'rownames' in the returned 'matrix'. If 'TRUE' the 'key' of
          the 'data.table' will be used if it is a single column,
          otherwise the first column in the 'data.table' will be used.
          Alternative a vector of length 'nrow(x)' to assign as the row
          names of the returned 'matrix'.

We should expect  the 2nd call above  to yield a 1x2 matrix of numeric having the single rowname of "bob".  Instead, it yields a 1x3 character matrix.

```
R version 3.4.3 (2017-11-30)
Running under: CentOS Linux 7 (Core)

Matrix products: default
BLAS: /n/apps/CentOS7/install/r-3.4.3/lib64/R/lib/libRblas.so
LAPACK: /n/apps/CentOS7/install/r-3.4.3/lib64/R/lib/libRlapack.so

locale:
[1] C

attached base packages:

other attached packages:
[1] data.table_1.11.2 devtools_1.13.5  

[1] compiler_3.4.3 withr_2.1.2    memoise_1.1.0  digest_0.6.15
```

HughParsonage:
Closes #2930 from the documentation aspect. There may be a case for `nrow(x) == 1` prioritizing column indices, however I chose a documentation change rather than a code change as (a) it's already released, (b) it's consistent with base R.

### DIFF
--- a/man/as.matrix.Rd
+++ b/man/as.matrix.Rd
@@ -12,18 +12,21 @@ of the columns in the \code{data.table} as the \code{matrix} \code{rownames}.
 \arguments{
 \item{x}{a \code{data.table}}
 \item{rownames}{optional, one of the following specifications of the row names:
-\enumerate{
-\item If \code{length(rownames) == length(nrow(x))} then \code{rownames} itself
-will be assigned as the row names of the returned matrix.
-\item If \code{rownames} is a length-1 character or numeric vector,
-this specifies the column name or column index to use as
-the row names in the returned \code{matrix}.
-\item If \code{rownames=TRUE}, then the first column of \code{x}
-is used, unless \code{x} has a key and that key is a single column in which
-case \code{key(x)} will be assigned as the row names of the returned matrix.
-}
-}
-Note that if \code{length(rownames) == nrow(x) == 1} then the first specification above takes precedence, so \code{as.matrix(data.table("a", "b", "c"), rownames = 2)} is a 3-column matrix with row name \code{2}, not a 2-column matrix with row name \code{b}.}
+  \enumerate{
+    \item If \code{length(rownames) == length(nrow(x))} then \code{rownames} itself
+    will be assigned as the row names of the returned matrix.
+    \item If \code{rownames} is a length-1 character or numeric vector,
+    this specifies the column name or column index to use as
+    the row names in the returned \code{matrix}.
+    \item If \code{rownames=TRUE}, then the first column of \code{x}
+    is used, unless \code{x} has a key and that key is a single column in which
+    case \code{key(x)} will be assigned as the row names of the returned matrix.
+  }
+  N.B. if \code{length(rownames) == nrow(x) == 1} then the first specification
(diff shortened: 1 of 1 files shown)


man/as.matrix.Rd
```diff
@@ -11,30 +11,37 @@ of the columns in the \code{data.table} as the \code{matrix} \code{rownames}.
...
 \item{x}{a \code{data.table}}
-\item{rownames}{optional, a single column name or column index to use as 
-the \code{rownames} in the returned \code{matrix}. If \code{TRUE} the
-\code{\link{key}} of the \code{data.table} will be used if it is a 
-single column, otherwise the first column in the \code{data.table} will
-be used. Alternative a vector of length \code{nrow(x)} to assign as the
-row names of the returned \code{matrix}.}
+\item{rownames}{optional, one of the following specifications of the row names:
+\enumerate{
+\item If \code{length(rownames) == length(nrow(x))} then \code{rownames} itself
```
jangorecki: not sure but it might not be possible to use `\item` inside `\item{}{... \enumerate{.}}`. Probably best move to lower more descriptive sections.

HughParsonage: They can be definitely be nested. I separated them so that I could emphasize the first point later. Too OTT?

jangorecki: so maybe incorrect number/placement of `{` somewhere

HughParsonage: Ah thanks!

malcook:
Out of continued curiosity, what aspect of base R is this consistent with?  Thanks

jangorecki:
@malcook data.frame instead of data.table

sritchie73:
Just looking at this now, I actually agree with @malcook and hadn't considered the edge case where the data.table had only 1 row.

sritchie73:
#2938 now fixes this from a functionality point of view, i.e. making the functionality work the same when the data.table has only 1 row.

HughParsonage:
@sritchie73 for the record, would you mind rejecting this pull request seeing as yours supersedes?

sritchie73:
Sure, didn't realise I had the power to do so!

Closing this PR as it is superseded by PR #2938

PR #2944, opened 2018-06-20
Closes #1947, exception for gmin/gmax not to fail on ordered factors

### LINKED ISSUE #1947: min/max fails on ordered factors when using by
It seems as this is a regression in the latest version (data.table_1.9.8), as it worked before 

```
library(data.table)
test <- data.table(V1=factor(rep(c("a","b"), 10), levels=c("a", "b"), ordered=TRUE), 
V2=rep(c("c","d", "e", "f"), 5))
test[,min(V1)]           # (1)
test[,min(V1),by=V2]     # (2)
```

(1) correctly works but (2) returns an error:
`Error in gmin(V1) : min is not meaningful for factors.`

MichaelChirico:
#1947 I'm kind of surprised how easy this fix was.


NEWS.md
```diff
@@ -7,6 +7,8 @@
 
 #### BUG FIXES
 
+1. `gmin` and `gmax` no longer fail on _ordered_ factors, for which these operations are indeed meaningful, [#1947](https://github.com/Rdatatable/data.table/issues/1947#issuecomment-398212372). Thanks to @mcieslik-mctp for identifying and @mbacou for the nudge.
```
jangorecki: link to issue is enough, don't need to link particular comment :)

inst/tests/tests.Rraw
```diff
@@ -11983,6 +11983,36 @@ test(1914.8, identical(dtGet(dt.comp, 1), dt[[1]]))
...
+test(1915.3, DT[ , min(V1)],
+     structure(1L, .Label = c("a", "b"), class = c("ordered", "factor")))
+test(1914.4, DT[ , max(V1)],
+     structure(2L, .Label = c("a", "b"), class = c("ordered", "factor")))
+
+## make sure GForce is activated
+old_optim = options(datatable.optimize = Inf)
+test(1914.5, DT[ , min(V1), by = V2],
+     data.table(
+       V2 = c("c", "d", "e", "f"),
```
jangorecki: In both GForce tests you used distinct values to group by in V2 field, which is somehow an edge case when grouping, in case of min/max then it is simply "as is". Would be a little bit safer to have GForce tests with 3 rows matching same group, so min/max on that is actually tested.

arunsrinivasan: @MichaelChirico nice catch/fix. If you can ping me when you've added the test you'd like to, I'd be happy to merge.

PR #2954, opened 2018-06-24
Vignettes update, closes #2952

### LINKED ISSUE #2952: Update _keys_ and _indices_ vignette for subset with index on multiple fields
Starting from this place and further:
https://github.com/Rdatatable/data.table/blob/master/vignettes/datatable-keys-fast-subset.Rmd#4-binary-search-vs-vector-scans
There are examples of subset `flights[origin == "JFK" & dest == "MIA"]` which assume binary search does not take place. Since we recently got indices to work on multiple fields combined with `&` this should be updated.
This can be clearly observe in conclusion there:
> The speed-up is **~1x!**

and in _indices_ vignette

> n the future, we plan to extend auto indexing to expressions involving more than one column. Also we are working on extending binary search to work with more binary operators like <, <=, > and >=. Once done, it would be straightforward to extend it to these operators as well.
> We will extend fast subsets using keys and secondary indices to joins in the next vignette, “Joins and rolling joins”.

jangorecki:
closes #2952


vignettes/datatable-benchmarking.Rmd
```diff
@@ -30,42 +24,104 @@ sudo lshw -class disk
...
+#Subsetting optimization disabled because the cross-product of RHS values exceeds 1e4, causing memory problems.
+#...
+```
 
-## subset: index aware benchmarking
+# subset: index aware benchmarking
 
-For convinience data.table automatically builds index on fields you are doing subset data. It will add some overhead to first subset on particular fields but greatly reduce time to query those columns in subsequent runs. When measuring speed best way is to measure index creation and query using index separately. Having such timings it is easy to decide what is the optimal strategy for your use case.
-To control usage of index use following options (see `?datatable.optimize` for more details):
+For convinience `data.table` automatically builds index on fields you are doing subset data. It will add some overhead to first subset on particular fields but greatly reduce time to query those columns in subsequent runs. When measuring speed best way is to measure index creation and query using index separately. Having such timings it is easy to decide what is the optimal strategy for your use case.
```
MichaelChirico: convenience

jangorecki: Push to branch please :)

jaapwalhout: Some other possible small grammar improvements:

~~builds index~~ => builds an index
~~greatly reduce time~~ => greatly reduces time
~~speed best way~~ => speed, the best way
~~using index separately~~ => using an index separately

mattdowle:
This is great detail w.r.t. data.table but high level general benchmarking principles need introducing first like the pitfalls of `microbenchmark()` on small data, call overhead etc.  Follow up issue: #3028.

PR #2961, opened 2018-07-03
rolling mean

jangorecki:
Address #2778
Proposed `frollmean` implementation, all feature to be supported are implemented, also docs. When it will pass review I will re-use it for rest of rolling functions, and then prepare for 1.12.0.


arunsrinivasan:
Very nice! Will take a look at this ASAP and write back.

jangorecki:
just pushed update
- rebased to recent master
- uses memmove instead of memcpy
- fixed compiler warning
- proper way of nested parallelism
- now parallel will work with verbose=T when there is no print inside parallel region
- reorg code for less repetition and C level wrappers
- some segfault fixes

src/frollR.c
```diff
@@ -0,0 +1,209 @@
...
+  else if (!strcmp(CHAR(STRING_ELT(align, 0)), "left")) ialign = -1;
+  else error("Internal error: invalid align argument in rolling function, should have been caught before. please report to data.table issue tracker.");
+
+  if (badaptive && ialign!=1)
+    error("using adaptive TRUE and align argument different than 'right' is not implemented");
+  
+  SEXP ans;
+  ans = PROTECT(allocVector(VECSXP, nk * nx)); protecti++;      // allocate list to keep results
+  double* dans[nx*nk];                                          // pointers to answer columns
+  double* dx[nx];                                               // pointers to source columns
```
jangorecki: @mattdowle are those double pointers safe? in other place I got segfault with such initialization. Should these be changed to `R_alloc`?

mattdowle:
Starting to look through this.
Can't see a NEWS item.
Suggest adding "*experimental function names and arguments; new in v1.12.0*" to the top of the manual pages so that it's clear that it's new and might change.
I saw some `malloc` that don't have return value checked. Those need `if (==NULL) error("alloc failed...)` adding, or change to `R_alloc()` is preferable where possible.

jangorecki:
@mattdowle NEWS entry added, also checks for `malloc`. I did not use `R_alloc` because those functions are attempting to be R agnostic.

As of now, it might happened that `error("could not allocate...")` will be raise from parallel region. The one defined in `frollfunR` when user provides 2+ columns to roll on AND/OR 2+ window sizes for rolling. There is nested parallel region in `fadaptiverollmeanFast` function where allocation happens but allocation is made before parallel region is entered there. Still it might be in parallel region of parent function `frollfunR`. Do I understand correctly this is not safe? If so I had to either
- move those allocations to frollfunR before entering topmost parallel region
- implement error handling from parallel region by returning some kind of `ans_t` storing status, error message, and results (when no exception occurred).

mattdowle:
Yes, the malloc itself is thread-safe (thread lock inside it). It's just the R API `error()` that isn't. Allocating up front before parallel region is usually best. Even a large malloc() will be almost instant so there's no time lost if it's not used; just higher memory usage and the possibility of failing to allocate something that isn't needed. Only at the point of touching allocated memory does anything happen (e.g. calloc vs malloc).

jangorecki:
Added exception handling so fails of `malloc` are handled using this exception handling mechanism. Additionally one type of console output (printed with `verbose`) is now warning. It should be warning but I simplified it before to not need to handle exceptions (unit test for that had comment `# this could be warning but no warns from plain C code`). Because now we have exception handling mechanism I turned those console outputs into warnings. There is a limit on how long error/warning string can be, set to 255, which is safe because that string does not depend on user input.
Build and install using gcc-7 going perfectly well. When using `cc(F)` with same gcc I am getting following warning:
```
frollR.c:124:23: warning: missing braces around initializer [-Wmissing-braces]
       dans[i*nk+j] = ((double_ans_t) { .ans=REAL(VECTOR_ELT(ans, i*nk+j)), .status=0, .message="\0" });
                       ^
                                                                                                {    }
```

src/types.h
```diff
@@ -0,0 +1,7 @@
+#include<stdint.h>
+
+typedef struct double_ans_t {
+  double *ans;
+  uint8_t status;   // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
+  char message[4][256]; // STDOUT: output, STDERR: message, warning, error
```
jangorecki: @mattdowle Currently this field is only used for storing errors and warnings, which won't exceed 255 chars. For storing console output for `verbose=T` (so verbose can work from parallel region) we will need to extend size of that char. Ideally we need to make it dynamic. For this PR this is enough as `verbose=T` disables topmost parallelism, and nested parallelism (where applicable) is not using `verbose` from parallel region. The cost of that is that verbose does not print in real-time, as opposed to `fread` which uses sophisticated method for printing safely and still do parallel processing.

Issue #1880, opened 2016-10-14
Add keepby= to do what by= does now.


mattdowle:
For discussion ...

Currently `by=` returns the groups in the order each group first appeared. I think most often most people (including myself now) actually want the groups returned in group order by default; i.e. what `keyby=` does. Worse, some people think the groups are returned in _random_ order and that that is for speed reasons; i.e. speed comes first. In fact, it actually takes data.table longer to return the groups in original order than it does to order the groups (it finds the order of the groups to find the groups in the first place rather than using hash tables); it has to do an extra step to get back to original order.  Of the users that have been left with the impression that data.table returns groups in random order for speed reasons, some of those then start to fear that the row order within the groups are not retained for speed reasons either.  Nothing could be further from the truth.

NB1: We're not talking about the order of the rows _within_ each group here at all.   The order of the rows within each group is always retained. Always has been always will be. Set in stone.  We're only talking about the order that the groups appear in the result returned.   This FR doesn't apply when a `:=` is present either; e.g. recall that the groups don't even have to be contiguous with a `:=` by group.

NB2: When we say the 'first appeared in the data' or 'order of the rows within each group' we mean after `i` has been applied, if present.  Since `DT[i, j, by]` is the same as `DT[i][,j,by]`.
 
Keeping the groups in first appearance order comes up for some/many users and is really important for them. In fact that's what I needed when I first created `by=` which is why I made that the default. But it can appear strange to others that the order that the groups first appear in the data should be relied on. Therefore, this FR is to create `keepby=` to do what `by=` does now.  That way readers of the code in future will know that this query is expecting to keep first appearance order of groups retained.  Retain the important ability, but make it clearer that the query uses it. I haven't seen this ability in other software so another but smaller motivating factor is to more easily explain that "data.table has `keepby=`".

Alternative keyword:  `batchby=`.  But to me that conveys batches defined by contiguous groups, like [this question](http://stackoverflow.com/a/27826941/403310) where the same group values occur later but the user wishes them to be a separate group.   For that, the new `rleid()` on the right hand side of `by=` makes most sense.  That way it can be applied on a column basis. 

For example, if this goes ahead, I'll change [this answer](http://stackoverflow.com/a/8833899/403310) to use `keepby=` to make it clearer to readers of that solution.  That question requires something to be done depending on the _previous_ group.  See also #606. 

Then over 2 years in the usual way, two options : 
[...]

PR #2973, opened 2018-07-11
Various updates to intro vignette, closes #2966

### LINKED ISSUE #2966: `forder` not faster than `base::order`?
I was reading the [Introduction to data.table](https://cloud.r-project.org/web/packages/data.table/vignettes/datatable-intro.html) documentation and found this part:

image

It seems like for some reason, `base::order` is getting the same performance as `data.table::forder`.

MichaelChirico:
@mattdowle copying the comment from the issue thread here:

Noticing this in the FAQ as well, maybe I don't understand but doesn't this contradict #1880?

> * All we did was to change `by` to `keyby`. This automatically orders the result by the grouping variables in increasing order. Note that `keyby()` is applied after performing the operation, i.e., on the computed result.

### DIFF
--- a/vignettes/datatable-intro.Rmd
+++ b/vignettes/datatable-intro.Rmd
@@ -449,11 +449,11 @@ ans
 
 #### {.bs-callout .bs-callout-info}
 
-* All we did was to change `by` to `keyby`. This automatically orders the result by the grouping variables in increasing order. Note that `keyby()` is applied after performing the operation, i.e., on the computed result.
+* All we did was to change `by` to `keyby`. This automatically orders the result by the grouping variables in increasing order. In fact, due to the internal implementation of `by` first requiring a sort before recovering the original table's order, `keyby` is typically faster than `by` because it doesn't require this second step.
 
-**Keys:** Actually `keyby` does a little more than *just ordering*. It also *sets a key* after ordering by setting an `attribute` called `sorted`. But we'll learn more about `keys` in the *Keys and fast binary search based subset* vignette.
+**Keys:** Actually `keyby` does a little more than *just ordering*. It also *sets a key* after ordering by setting an `attribute` called `sorted`. 
 
-For now, all you have to know is that you can use `keyby` to automatically order the result by the columns specified in `by`.
+We'll learn more about `keys` in the *Keys and fast binary search based subset* vignette; for now, all you have to know is that you can use `keyby` to automatically order the result by the columns specified in `by`.
 
 ### c) Chaining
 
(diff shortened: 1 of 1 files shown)


vignettes/datatable-intro.Rmd
```diff
@@ -16,30 +16,31 @@ knitr::opts_chunk$set(
...
 
-Briefly, if you are interested in reducing *programming* and *compute* time tremendously, then this package is for you. The philosophy that *data.table* adheres to makes this possible. Our goal is to illustrate it through this series of vignettes.
+* *automatically* optimising operations internally, and very effectively, by knowing precisely the data required for each operation, leading to very fast and memory efficient code.
 
+Briefly, if you are interested in reducing *programming* and *compute* time tremendously, then this package is for you. The philosophy that `data.table` adheres to makes this possible. Our goal is to illustrate it through this series of vignettes.
 
 ## Data {#data}
 
 In this vignette, we will use [NYC-flights14](https://github.com/arunsrinivasan/flights/wiki/NYC-Flights-2014-data) data. It contains On-Time flights data from the [Bureau of Transporation Statistics](http://www.transtats.bts.gov/DL_SelectFields.asp?Table_ID=236) for all the flights that departed from New York City airports in 2014 (inspired by [nycflights13](https://github.com/hadley/nycflights13)). The data is available only for Jan-Oct'14.
```
MichaelChirico: > The data is available only for Jan-Oct'14.

By chance, is the reason for this that this vignette was released before the rest of 2014 data was released?

vignettes/datatable-intro.Rmd
```diff
@@ -16,30 +16,31 @@ knitr::opts_chunk$set(
...
-
-* *automatically* optimising operations internally, and very effectively, by knowing precisely the data required for each operation and therefore very fast and memory efficient.
+* performing analysis *fluidly* without the cognitive burden of having to map each operation to a particular function from a potentially huge set of functions available before performing the analysis.
 
-Briefly, if you are interested in reducing *programming* and *compute* time tremendously, then this package is for you. The philosophy that *data.table* adheres to makes this possible. Our goal is to illustrate it through this series of vignettes.
+* *automatically* optimising operations internally, and very effectively, by knowing precisely the data required for each operation, leading to very fast and memory efficient code.
 
+Briefly, if you are interested in reducing *programming* and *compute* time tremendously, then this package is for you. The philosophy that `data.table` adheres to makes this possible. Our goal is to illustrate it through this series of vignettes.
```
franknarf1: Maybe add something like...

> (See `vignette(package = "data.table")` for the full list.) 

I think most people reading the vignette don't realize it's bundled with the package (and that an online version they're reading might be out of date or otherwise out of sync with the version they have installed...).

Along the same lines, the final line of this vignette refers to the "next" vignette, but I don't know how the reader would know which vignette that is if they're not on the wiki (which has a numbered list).

vignettes/datatable-intro.Rmd
mattdowle: Very likely, yes.

PR #2976, opened 2018-07-16
Closes #2975 -- correct error message in transpose.c

### LINKED ISSUE #2975: transpose(l, fill=NULL)
Hello, 

 In `transpose.c`:
```
  if (length(fill) != 1)
    error("fill must be NULL or length=1 vector.");
```
Hence `transpose(l, fill=NULL)` returns this error message, because `length(NULL)` is `0`. So it's not true that `fill` can be `NULL`. Perhaps you meant `NA` ?

MichaelChirico:
#2975


arunsrinivasan:
I'm not entirely sure what I did to merge it to patch-1 branch instead of master. Will take a look later. Apologies.

Issue #2081, opened 2017-03-27
[Request] Don't inspect i=CJ(...) names in DT[CJ(...), on=] joins


franknarf1:
Looking again at #1596 , one point against it is that some folks may be taking advantage of the default V1, V2, ... names, so auto-assigning different names would break their code.

One alternative, that would address my primary use-case, would be for `CJ` to not need names during a join:

```
library(data.table)
DT = unique(data.table(datasets::CO2)[, .(Plant, Type, Treatment)])
setkey(DT, Plant, Type)

# good -- no names needed using list
DT[.("Mc1", "Mississippi"), .N, by=.EACHI]

# good -- no names needed using CJ and (implicitly) on = key
DT[CJ(Plant, Type, unique=TRUE), .N, by = .EACHI]

# bad -- breaks for explicit on = key 
DT[CJ(Plant, Type, unique=TRUE), on=key(DT)]

# bad -- breaks for on=some-non-key
DT[CJ(Plant, Treatment, unique = TRUE), on=.(Plant, Treatment)]
```

So I guess I'm asking for an exception in `[.data.table` that ignores names in `i` when `i=CJ(...)`, similar to how names are ignored with list inputs or when `on=` is implicitly the key.

Of course, my desired syntax would also work if FR #1596 went through.

MichaelChirico:
I think I'm fine with breaking some code relying on `V1`. Just filed #2977.

PR #2977, opened 2018-07-17
CJ now auto-names its input

### LINKED ISSUE #1596: FR: assign names to CJ() like data.table() does
I frequently do

```
DT[CJ(colA = colA, colB = colB, unique=TRUE), on=c("colA","colB")]
# to complete missing levels

# or 
DT[, CJ(colA = colA, colB = colB, unique=TRUE)][!DT, on=c("colA","colB")]
# to identify missing levels
# http://stackoverflow.com/a/36065607/1191259
```

It would be nice if I could get away with writing `colA` and `colB` fewer times. The FR here is for

```
CJ(colA, colB, unique=TRUE, names=TRUE) 
```

to infer the names `colA` and `colB`, perhaps using whatever method is used by data.frame() and data.table() (`make.names`?). 

(The name repetition could be reduced further if `on=.Icols` were a thing, I suppose, but I'll leave that for a separate FR.)

SO posts to update...

* https://stackoverflow.com/a/46243102/

### LINKED ISSUE #2192: CJ(sorted = TRUE) document wording
CJ documentation has the line regarding sorted paramter

> sorted  logical. Should the input order be retained?

The explanation is the opposite of the actual setting. If you want the input order to be retained, you set sorted = FALSE.

`CJ(c('c','b','a'), c('xx','zz','yy'), sorted = TRUE)`

will have 'a' before 'b'

MichaelChirico:
Closes #1596,  
Closes #2192
Related #2081

- technically a breaking change since I'm sure there's code out there relying on `CJ(a, b)` being named `V1`,`V2` (there was in our own testing, after all).
- no _new_ tests since this breaks a few old tests, I think the behavior is covered by these existing tests.
- minor unrelated change: `if (n < 1L) return( null.data.table() )` moved earlier

Hope is that this `name_dots` function can be used for fixing this weird autonaming issue as well (but that fix is a bit messier):

```
data.table(a = 1:5, b = 6:10)[ , if (.BY$a > 2) .(b), by = a]
# vs
data.table(a = 1:5, b = 6:10)[ , .(b), by = a]
```


jangorecki:
I would postpone this to 1.12

mattdowle:
Or it could be put in now with an option to turn on new behaviour (default off for now).  We're getting used to this option shenanigans now, so it's not really much more effort.  Then change default to on in 1.12.0 and the 0.5% of folk that break on 1.12.0 can revert until they update.

MichaelChirico:
thanks for following up, haven't found time to sit down for this these few weeks +1

Issue #2879, opened 2018-05-15
collect more statistics about the data


jangorecki:
data.table could collect more statistics about the data while processing. This allows potential optimizations, not limited to internal data.table code. Users can use them to speed up their code and design more data-driven functions.
List of measures to collect:
- [x] is sorted
- [x] has index
- [ ] has NA
- [ ] number of groups (uniqueN)
- [ ] count of biggest group
- [ ] is unique (uniqueN == .N)
- [ ] range (min, max)

optionally, as I don't see obvious optimizations coming from those:
- [ ] NA count
- [ ] sd, var

franknarf1:
I'd also be interested in collection of statistics related to a join (eg, does each row of `i` have exactly one matching row in `x` in an `x[i]` join?), though I don't know if that's relevant to optimization nor where such statistics would be collected (since they pertain to multiple tables, so embedding them in one table doesn't make sense). Anyway, just a though, mentioned earlier in https://github.com/Rdatatable/data.table/issues/2022

jangorecki:
@franknarf1 _is unique_ on a field would guarantee that joining to that table on this field won't produce multiple matches.

jangorecki:
We need to plug collections of those statistics into existing functions, so explicitly calling `analyze` won't be needed to get benefits. Posting as a reference:
```r
analyze <- function(x, ...) {
  stopifnot(is.data.table(x))
  ## by data.table statistics
  d <- dim(x)
  nr <- d[1L]
  nc <- d[2L]
  cols <- names(x)

  ans = list(is.sorted=NA, is.indexed=NA, is.unique=NA, is.ascii=NA,
             unique.n=NA_integer_, maxgrp.n=NA_integer_, na.count=NA_integer_,
             has.na=NA, all.na=NA, has.nan=NA,
             min=NA_real_, max=NA_real_)
  ans = rbindlist(sapply(names(x), function(x) ans, simplify=FALSE), idcol="column")
  ## by columns statistics
  ## indices
  o <- sapply(simplify=FALSE, cols, data.table:::forderv, x=x, sort=TRUE, retGrp=TRUE)
  if (is.null(attr(x, "index", exact = TRUE))) setattr(x, "index", integer())
  idx <- attr(x, "index", TRUE)
  is.ascii <- function(x) {
    if (!is.character(x)) return(NA)
    asc <- iconv(x, "latin1", "ASCII")
    !any(is.na(asc) | asc != x)
  }
[...]

PR #2989, opened 2018-08-01
Improve helpfulness of warning message during on-assignment type coer…

MichaelChirico:
…cion.

Original tweet & some discussion here:

https://twitter.com/sarahbeeysian/status/1021359529789775872


franknarf1:
Maybe consider changing tone instead of removing the sentence? Like "Setting the column type as intended up front is the best way to get around this if your RHS has the correct type." or something.

Fwiw, for me...

> set the column type correctly up front

... was the most useful part of the message as a new user and probably something I wouldn't think of from reading the new message.

> inspect your initial LHS types and be sure they're as expected

... gets at the point too indirectly and no longer contains a recommended fix.

I don't find the twitter complaint ("it's rude because I know better already") convincing as an argument to remove it when other users might know less than they do.

----

Also, the first pitfall you highlight 

> A common source of this error is mismatch of return types when assigning by group, especially in edge cases. 

...has its own (error) message already:

```
library(data.table)
DT = data.table(a = 1:2)
DT[, v := if (.GRP==1L) "a" else 1L, by=a]
> Error in `[.data.table`(DT, , `:=`(v, if (.GRP == 1L) "a" else 1L), by = a) : 
>   Type of RHS ('integer') must match LHS ('character'). To check and coerce would impact performance too much for the fastest cases. Either change the type of the target column, or coerce the RHS of := yourself (e.g. by using 1L instead of 1)
```

Also, you refer to it as "this error", though it is a warning, which may be confusing.

mattdowle:
Please @franknarf1 and @MichaelChirico review before merge. (The drop down under reviewers won't let me pick either of you.  Michael because you submitted the PR I guess and Frank because you're not a project member it seems?!  But you can both add your reviews I believe.)

src/assign.c
```diff
@@ -593,21 +592,34 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values, SEXP v
...
+                warning("Coerced %s RHS to %s to match the type of the target column (column %d named '%s'). If the target column's type %s is correct, it's best for efficiency to avoid the coercion and create the RHS as type %s. To achieve that consider R's type postfix: typeof(0L) vs typeof(0), and typeof(NA) vs typeof(NA_integer_) vs typeof(NA_real_). You can wrap the RHS with as.%s() to avoid this warning, but that will still perform the coercion. If the target column's type is not correct, it's best to revisit where the DT was created and fix the column type there; e.g., by using colClasses= in fread(). Otherwise, you can change the column type now by plonking a new column (of the desired type) over the top of it; e.g. DT[, `%s`:=as.%s(`%s`)]. If the RHS of := has nrow(DT) elements, then the assignment is called a column plonk and is the way to change a column's type. Column types can be observed with print(x,class=TRUE) and sapply(x,class).",
```
franknarf1: > Column types can be observed with print(x,class=TRUE) and sapply(x,class).

Some users won't know types for non-atomic classes (integer = IDate, factor; double = Date). Could change `sapply(x,class)` to `sapply(x,typeof)`, but I'm not sure how to change the reference to `print`. Maybe some rewording or an extra sentence could work. Alternately, maybe `print` could have another option, like verbose = TRUE that prints maximal info (somewhat substituting for str and similar to `tables()`):

```
library(data.table)
DT = data.table(id = 1:2, d = as.IDate(Sys.Date()) + 0:1)
setkey(DT, id)
setindex(DT, d, id)

print(DT, verbose = TRUE) # fake code
#       id          d
#    <int>     <IDat>
#               <int>
# 1:     1 2018-08-10
# 2:     2 2018-08-11
# 
# key: id
# indices: 
# - d, id
```

(... Also displaying new statistics from #2879)

Anyway, I guess the type/class distinction will only matter in rare cases, like...

```
library(data.table)
DT = data.table(id = 1:2, d = as.IDate(Sys.Date()) + 0:1)

DT[1, d := "1999-01-01"] 
# gets coerced to NA_integer_

DT[1, d := 1999-01-01] 
# worse, if forgetting quotes and class/type, it gets silently handled as an int
```

In contrast, sub-assigning a character to a factor works as expected.

MichaelChirico:
Awesome! Much more thorough overhaul 👍

PR #3010, opened 2018-08-24
Add cmd= to fread()

mattdowle:
Rationale explained in news item added.


mattdowle:
Please thumbs up/down anybody watching.

NEWS.md
```diff
@@ -11,7 +11,19 @@
...
+```
+   fread(aFileName)                     # if aFileName is not a file name but a command, fread no longer runs the command.
+   fread(someText)                      # if someText is not some text but a command, fread no longer runs the command.
+   fread("grep blah file.txt")          # command is a literal string constant appearing in your code and is safe; still works and is common usage.
+   x = "grep blah file.txt"
+   fread(x)                             # no longer runs command. fread(cmd=x) is now needed.
+   y = "blah"
+   fread(paste("grep",y,"file.txt"))    # no longer runs command. fread(cmd=paste("grep",y,"file.txt")) is now needed.
+```
+Owing to the security nature of this change, there is no option provided to restore old behaviour.
```
jangorecki: I recommend to keep old behavior for 1.11.6 and change default to new behavior in 1.12.0. Assuming 1.11.6 is a patch release to 1.11.0 we should avoid breaking changes.

sritchie73:
I think this may break many people's scripts (some of my own included), due to the common use case of needing a system command to read in gzipped files (e.g. `fread("zcat file.txt.gz")`) as reported by many in #717

NEWS.md
HughParsonage: I agree and would go further: offer `cmd` argument as a safer option but keep existing behaviour in `input` for ever. `f` for friendly and all that. We're not even sure that this security concern has *ever* been exploited.

mattdowle:
@sritchie73 That won't break. See the last sentence in news item 5 starting "Other than" and see the examples in the news item. Perhaps the part "`cmd=` must now be used in order to run a command" which (is actually untrue) can be rephrased. That sentence was supposed to be read in context together with the sentence following it (starting "Other than").

HughParsonage:
I think he mean things like `lapply(paste("zcat", dir(path = ".gz$")), fread)`

mattdowle:
That's a good example. It could be allowed since it contains no variables. But if it did contain variables, then I see it would be particularly inconvenient if this PR went ahead. Since `lapply` works via the first argument.

sritchie73:
I had indeed missed the "other than" part. That being said, I do have some scripts that use paste to build from a variable, but I'm happy to concede that's bad practice and should be rewritten to use `cmd=` if its safer.

mattdowle:
Great, I'll improve the news item.
It boils down to packages and systems we have no control over. What can we do to close the security concern. Asking them to add `file=` is not bad but needs them to do it. I'm thinking it's best to close the back door firmly and quickly, with the cost of convenience. Then bring back the convenience in other and better ways, like implementing `file=` to accept a vector of ".gz" files directly so the `lapply` and `rbindlist` wouldn't be needed either.

mattdowle:
Another round of thumbs up/down please.  Just to the extent of merging to master and seeing how it goes in dev.  We can always revert/tweak in dev before release.
I downgraded the error to warning.  The warning only happens if variables are used.  News item pared down.

st-pasha:
It appears to me that if security is the concern (and it's a very valid concern!), then issuing a mere warning is not enough: warnings in production systems often go unnoticed, or even if they are noticed the hacker has already done the damage.

So I would vote for command without cmd= being the error. It doesn't have to be in this PR, but at least eventually. Then it would be good for the warning to mention that "the input is recognized as a command, but this behavior is deprecated and will be removed in version X". Or don't wait until version X, and make it an error straight away. Yes, this will be a breaking change, but if the fix is easy, and the error tells you what the fix is, then maybe it's ok. Especially since it is done in the name of security, and most people would appreciate added security.

I appreciate Hugh's remark that this has probably never been exploited before, but do we really want to wait until it is?

mattdowle:
@MichaelChirico Good point, will do.
@st-pasha Yes the news items contains "In the near future the warning will be upgraded to error."

HughParsonage:
> It appears to me that if security is the concern (and it's a very valid concern!), then issuing a mere warning is not enough: 

I completely disagree. From a security perspective, this cure could be worse than the disease. After all, this breaks previous documented behaviour and indeed was *recommended* for many tasks. If it breaks a workflow that was previously working and that had no other security concerns (i.e. it accepted no user input) then that is in some sense a denial of service. 

If the security problems are really a concern that we feel it is the responsibility of `fread` to fix, in my view it would be better to whitelist commands than to direct developers to use `cmd=` blindly. After all, if a developer is just sending unsanitized user input to `input=`, it does not fortify the script if they just use `cmd=` instead. By emitting a warning, we risk a kneejerk reaction to restore previous behaviour without any improvement to security. 

I'd recommend a message, not a warning (since warnings can be elevated to errors). Or a whitelist. Plus documentation urging care with system commands.

I'm reminded of this email by Linus:

st-pasha:
> After all, this breaks previously documented behavior

Here's the problem with documented behaviors: nobody reads the documentation. Ok, that's obviously an exaggeration, but if I were to guess the percent of people who ever used `fread` without reading its documentation, I'd say 90% or more.

What most people do, is they search for a tutorial (or SO answer), they find something that looks like what they need
```
DT = fread("datafile.csv")
``` 
they adopt it for their use case
```
DT = fread(datafile)
```
if it works -- they are happy, everything's great. The user may probably check to see what happens if `datafile` refers to an invalid file -- they find it raises a reasonable exception, and that's it.

What they missed in this process is that there are conditions under which the contents of `datafile` string will be executed. This is not at all obvious. There is no way to guess or even begin to suspect that perhaps one needs to consult the documentation. The command roughly says "read file", and an unsophisticated user will think that this is exactly what it does, no more nor less.

It is these users that we're trying to protect, perhaps at the annoyance of the more sophisticated minority.

mattdowle:
@st-pasha Yes true, but I now realize thanks to @hughparsonage that all those people we have in mind there have something else in common: they're not creating apps/apis at all. Let alone apps/apis which are public facing where input from user is passed straight to fread input=. So the vast majority have zero risk.

What Hugh said chimes with me. For example, I depend on xdotool to send commands from gedit to the target bash window. I love this ability and you can prise it from my cold dead hands.  But Ubuntu took it away in Wayland as a deliberate decision for security reasons. Really annoying and I don't appreciate it. When I upgraded to 18.04, Wayland was the default and my world broke. A few hours of internet trawling layer and I revert back to X where I still am today.

Hugh suggested downgrade to message() which should satisfy everyone as per Linus advice: "do no harm, just report."  Once folk are informed and we gather feedback and new info, we can then remove message, or make it warning/error.  Sound ok?

PR #3015, opened 2018-08-29
New warning for %between%

### LINKED ISSUE #3014: potential issue with the between operator
The between operator seems to not operate as intended:
```
d <- data.table(A=1:10)
d[,B:=3]
d[A %between% c(B,B+1)]  ##  incorrect result since default setting includes bounds 
#    A B
# 1: 3 3
d[A %between% c(3,3+1)]  ## correct result when using same notation with hard-coded bounds
#    A B
# 1: 3 3
# 2: 4 3
d[between(A,B,B+1)]      ## correct result when using different notation
#    A B
# 1: 3 3
# 2: 4 3
```

My R session and data.table version:
```
R version 3.4.3 (2017-11-30) -- "Kite-Eating Tree"
> library(data.table)
data.table 1.11.2
```

Also tried:
```
 R version 3.5.1 (2018-07-02) -- "Feather Spray"
> library(data.table)
data.table 1.11.5 IN DEVELOPMENT built 2018-08-16 23:02:55 UTC; root  Latest news: r-datatable.com
```

MichaelChirico:
Closes #3014

### DIFF
--- a/R/between.R
+++ b/R/between.R
@@ -12,7 +12,18 @@ between <- function(x,lower,upper,incbounds=TRUE) {
 }
 
 # %between% is vectorised, #534.
-"%between%" <- function(x, y) between(x, y[[1L]], y[[2L]], incbounds=TRUE)
+"%between%" <- function(x, y) {
+  if (length(y) > 2L) {
+    ysub = substitute(y)
+    warning("RHS has length() greater than 2. ",
+            if (is.call(ysub) && ysub[[1L]] == 'c')
+              sprintf("Perhaps you meant %s? ",
+                      capture.output(print(`[[<-`(ysub, 1L, quote(list))))),
+            "The first element should be the lower bound(s); ",
+            "the second element should be the upper bound(s).")
+  }
+  between(x, y[[1L]], y[[2L]], incbounds=TRUE)
+}
 # If we want non inclusive bounds with %between%, just +1 to the left, and -1 to the right (assuming integers)
 
 # issue FR #707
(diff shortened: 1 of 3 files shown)


R/between.R
```diff
@@ -12,7 +12,18 @@ between <- function(x,lower,upper,incbounds=TRUE) {
 }
 
 # %between% is vectorised, #534.
-"%between%" <- function(x, y) between(x, y[[1L]], y[[2L]], incbounds=TRUE)
+"%between%" <- function(x, y) {
+  if (length(y) > 2L) {
+    ysub = substitute(y)
+    warning("RHS has length() greater than 2. ",
```
HughParsonage: à la `if (1:5 > 1) "a"` we should indicate that part of the input is being ignored: "RHS has length > 2 and only the first and second elements will be used".

MichaelChirico: I thought to include that, but didn't want an overlong message; you don't think that's strongly implied by the warning here?

Maybe:?

> RHS has length() greater than 2 and the rest will be ignored. [...]

HughParsonage: I think a problem can arise when the user says `y = c(B, B + 1)`, 'knowing' that `length(y) > 2` -- duh! It's vectorized; the warning is nothing to worry about. 

On second thoughts, maybe this should be an error. Documentation is pretty clear that `y` must be length-2, it was just never enforced. I'm trying to think of a case where only a warning is appropriate.

The suggestion (`Perhaps...`) should obviously stay: that's imo the most valuable part of this PR.

MichaelChirico: I also was struggling to think of a use case that's not subsumed by just running `[1:2]` on whatever your input is (and even the `[1:2]` cases i couldn't think of a _real_ use case)

MichaelChirico: @HughParsonage updated

PR #3017, opened 2018-08-29
do not run ncol on list

### LINKED ISSUE #3003: fwrite unit tests fail when run in clean env
when only data.table was installed on clean R, at the current moment cannot reproduce locally but issue [still occurs](https://gitlab.com/Rdatatable/data.table/pipelines).
```
Running the tests in ‘tests/main.R’ failed.
Last 13 lines of output:
  Running /builds/Rdatatable/data.table/bus/test-r-release-vanilla/data.table.Rcheck/data.table/tests/tests.Rraw 
  
  **** Suggested package bit64 is not installed. Tests using it will be skipped.
  
  
  **** Suggested package xts is not installed. Tests using it will be skipped.
  
  
  **** Suggested package nanotime is not installed. Tests using it will be skipped.
  
  
  Running test id 1.1      Error in if (ncol(x) == 0L && file != "") { : 
    missing value where TRUE/FALSE needed
  Calls: test.data.table -> sys.source -> eval -> eval -> test -> fwrite
  Execution halted
```
issue introduced by https://github.com/Rdatatable/data.table/commit/c0f8647a65a718da47b7f0b615d31e0a7791ffd6

I don't see any reason why this fails only for clean R session. FYI @MichaelChirico

jangorecki:
As `x` might be a list, we should conditionally use `ncol` / `length`. Might eventually resolve #3003.

### DIFF
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -51,8 +51,7 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
     showProgress = FALSE
     eol = "\n"  # Rprintf() is used at C level which knows inside it to output \r\n on Windows. Otherwise extra \r is output.
   }
-  nc = if (is.data.frame(x)) ncol(x) else length(x) # is.data.frame also TRUE for data.table
-  if (nc == 0L && file != "") {
+  if (NCOL(x)==0L && file!="") {
     if (file.exists(file)) {
       warning("Input has no columns; doing nothing.",
               if (!append)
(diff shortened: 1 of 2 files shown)


MichaelChirico:
isn't this what `NCOL` is for?

jangorecki:
`NCOL` does not address that...
```r
> df=data.frame(a=1,b=2)
> l=as.list(df)
> NCOL(df)
[1] 2
> NCOL(l)
[1] 1
```

MichaelChirico:
oh i see... `NCOL` defaults to `1L`, not to `length(x)` 👍

PR #3019, opened 2018-08-29
Follow up to cmd= fread

mattdowle:
Follow up to PR #3010 as discussed there.
This PR actions [this comment](https://github.com/Rdatatable/data.table/pull/3010#issuecomment-416824029) in that PR.


inst/tests/tests.Rraw
```diff
@@ -9598,7 +9598,7 @@ test(1702.4, isoweek(isotest$input_date), isotest$expected_output)
 if (.Platform$OS.type=="unix") {
   cat("a,b\n4,2", file=f<-tempfile())
   cmd <- sprintf("cat %s", f)
-  test(1703.0, fread(cmd), data.table(a=4L, b=2L), warning="please use fread(cmd=...)")
+  test(1703.0, grep("Please use fread.cmd=.*security concern.*Please read item 5 in the NEWS file for v1.11.6", capture.output(fread(cmd), type="message")), 1L)
```
jangorecki: On 3.1.0
```
Running test id 1703      Taking input= as a system command ('cat /tmp/RtmpR14die/filee8a1f81dd46') and a variable has been used in the expression passed to `input=`. Please use fread(cmd=...). There is a security concern if you are creating an app, and that app could have a malicious user, and the app is not running in a secure envionment (e.g. the app is running as root). Please read item 5 in the NEWS file for v1.11.6 for more information.
Test 1703  should not fail but failed with error:  bad argument 
```

```r
args(capture.output)
#function (..., file = NULL, append = FALSE)
```

MichaelChirico: `tryCatch(fread(cmd), message = identity)$message`?

PR #3031, opened 2018-09-05
RHS NULL in prepareFastSubset

mattdowle:
In rev dep checking I found that data.table broke CRAN package `eplusr`.  Not bad really to have just one breakage in an edge case.
The one-line addition in this PR makes it pass.
Please check ok @MarkusBonsch and is there a way to create a test for this?
Since `mode(NULL)=="NULL"` it might be to do with the `mode()!=mode()` that Hugh and I recently removed in PR #3026.

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -2934,6 +2934,7 @@ isReallyReal <- function(x) {
     if(col %chin% names(i)) return(NULL) ## repeated appearance of the same column not suported (e.g. DT[x < 3 & x < 5])
     ## now check the RHS of stub
     RHS = eval(stub[[3L]], x, enclos)
+    if (is.null(RHS)) return(NULL)
     # fix for #961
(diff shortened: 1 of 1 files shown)


MarkusBonsch:
Thanks, @mattdowle  for fixing my bugs :) It is strange, however, that this bug is new, since the piece of code that you changed was just copy-pasted by me when implementing `.prepareFastSubset`. So maybe your guess is right and it has to do with PR #3026.

mattdowle:
Thanks @MarkusBonsch.  The NULL example came from eplusr package.  It was a bit tortuous to debug it as the NULL occurred on the 18th iteration of a loop and I wasn't able to establish the root cause.  I think the way it is now with this PR is better, especially as it's more like your original.  Passes CRAN packages.  Would be ideal to add a test but bigger problems to deal with.

PR #3035, opened 2018-09-06
[typo] it's -> its when possessive

HughParsonage:


MichaelChirico:
Went ahead and expunged all remaining (grammatically correct) instances of `it's` in favor of being a bit more formal: `it is`. Thinking is "no ambiguity left if it's `it is` vs `its`"; feel free to roll back if this is overly-pedantic.

Also got a bit trigger-happy with correcting formatting problems I came across in `frank.Rd`; feel free to revert these as well if needed and I'll file separately. 

For the record I used this to find the changes:

    grep -E "\b[Ii]t'?s\b" man/*

Issue #2962, opened 2018-07-04
R 3.4.3 -> 3.5 performance drop


s-Nick-s:
I've noticed a significant drop in data.table performance after upgrading to R 3.5.
For example:
```R
set.seed(1)
types <- c("A", "B", "C", "D", "E", "F")
obs <- 4e7
dt1 <- data.table(percent = round(runif(obs, min = 0, max = 1), digits = 2),
                  type = as.factor(sample(types, obs, replace = TRUE)))
microbenchmark::microbenchmark(
  test1 <- one[, list(percent_total = sum(percent)), by = type]
, times = 30)
```
On 3.5 gives me:
```R
Unit: milliseconds
                                                          expr    min      lq     mean   median       uq     max neval
 test1 <- dt1[, list(percent_total = sum(percent)), by = type] 522.53 539.668 576.2352 572.8301 602.0069 677.892    30
>sessionInfo()
R version 3.5.0 (2018-04-23)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 9 (stretch)

Matrix products: default
BLAS: /usr/lib/openblas-base/libblas.so.3
LAPACK: /usr/lib/libopenblasp-r0.2.19.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C             
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.11.4

loaded via a namespace (and not attached):
[1] microbenchmark_1.4-4 compiler_3.5.0       tools_3.5.0  
```
But on 3.4.3 it's ~20% faster:
```R
Unit: milliseconds
                                                          expr      min       lq     mean   median       uq      max neval
 test1 <- dt1[, list(percent_total = sum(percent)), by = type] 415.5938 445.8137 477.6562 470.0536 499.8974 576.9203    30
> sessionInfo()
R version 3.4.3 (2017-11-30)
Platform: x86_64-pc-linux-gnu (64-bit)
[...]

PR #3040, opened 2018-09-08
Faster by=key(DT)[1] and keyby= index

mattdowle:
```
N = 5e8
set.seed(1)
DT = data.table(
  id = sample(LETTERS[1:6], N, replace=TRUE),
  val = sample(10, N, replace=TRUE)
)
setkey(DT,id)
```

Before this PR : 
```
> system.time(DT[,sum(val),by=id,verbose=TRUE]) 
Detected that j uses these columns: val 
Finding groups using uniqlist ... 6.924sec 
Finding group sizes from the positions (can be avoided to save RAM) ... 0.002sec 
lapply optimization is on, j unchanged as 'sum(val)'
GForce optimized j to 'gsum(val)'
Making each group and running j (GForce TRUE) ... 3.156sec 
   user  system elapsed 
 10.567   0.601  11.168 
```

With this PR : 
```
> system.time(DT[,sum(val),by=id,verbose=TRUE]) 
Detected that j uses these columns: val 
Finding groups using uniqlist ... 0.373sec 
Finding group sizes from the positions (can be avoided to save RAM) ... 0.002sec 
lapply optimization is on, j unchanged as 'sum(val)'
GForce optimized j to 'gsum(val)'
Making each group and running j (GForce TRUE) ... 3.190sec 
   user  system elapsed 
  4.046   0.615   4.662 
```

The `uniqlist` part reduces from 6.9s to 0.4s.

ToDo:

- [x] ~add missing test of by=key(DT)[1] where that is integer64~  wasn't missing.
- [x] numeric rounding to pass test 1196.2

Found as a result of investigating and thanks to #2962.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@
 
 6. New `options(datatable.CJ.names=TRUE)` changes `CJ()` to auto-name its inputs exactly as `data.table()` does, [#1596](https://github.com/Rdatatable/data.table/issues/1596). Thanks @franknarf1 for the suggestion. Current default is `FALSE`; i.e. no change. The option's default will be changed to `TRUE` in v1.12.0 and then eventually the option will be removed. Any code that depends on `CJ(x,y)$V1` will need to be changed to `CJ(x,y)$x` and is more akin to a bug fix due to the inconsistency with `data.table()`.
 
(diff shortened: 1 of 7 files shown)


jangorecki:
do we want this in patch release? I think there are potentially other improvements we could do here, thus might make sense to postpone to 1.12.0

jangorecki:
The other improvements I had in mind was #3042, and it will probably takes some time. Lets get this one merged for 1.11.6, they don't seems to be that much related.

PR #3043, opened 2018-09-12
Fix AIX compilation issue with NAN and INFINITY

ayappanec:
We have recently ported R to AIX .
https://www.ibm.com/developerworks/aix/library/aix-toolbox/alpha.html#R
When trying to install data.table, there is a compilation issue with NAN and INFINITY. 

fread.c:79:28: error: initializer element is not constant
 static const double NAND = (double)NAN;
                            ^
fread.c:80:28: error: initializer element is not constant
 static const double INFD = (double)INFINITY;
                            ^
With gcc -E option , we see this

static const unsigned int _SINFINITY = 0x7f800000;
static const unsigned int _SQNAN = 0x7fc00000;

static const double NAND = (double)
 79 "fread.c" 3 4
                                  (*((float *)(&_SQNAN)))
 79 "fread.c"
                                     ;
static const double INFD = (double)
 80 "fread.c" 3 4
                                  (*((float *)(&_SINFINITY)))
 80 "fread.c"

So they don't seem to qualify as constant literals hence the compilation issue. 
I have tested the fix in Linux as well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,8 @@
 
 10. `fintersect` failed on tables with a column called `y`, [#3034](https://github.com/Rdatatable/data.table/issues/3034). Thanks to Maxim Nazarov for reporting.
 
+11. Compilation fails in AIX because NAN and INFINITY macros definition in AIX make them not constant literals, [#3043](https://github.com/Rdatatable/data.table/pull/3043). Thanks to Ayappan for reporting and fixing.
+
 #### NOTES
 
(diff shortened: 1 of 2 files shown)


jangorecki:
Hi, thanks for catching that, appreciate your PR.
Would be even better if you could:
- add entry in NEWS
- put comment next to `init()` declaration referring issue/PR number
Also note that we have [Contributing wiki](https://github.com/Rdatatable/data.table/wiki/Contributing) which describes how to make PR.

ayappanec:
Thanks. I have updated the PR as per your info.

mattdowle:
@ayappanec Could you contact me please (via email if you prefer) with your first and last name?  So I can add you as contributor in DESCRIPTION file for #3144.

PR #3048, opened 2018-09-14
update integration tests for impact from ggplot2 and xts updates

### LINKED ISSUE #3047: integration tests failing due to breaking changes in external pkgs
this passes
```
         pkg loaded version
 1:  ggplot2   TRUE   2.2.1
 5:      xts   TRUE  0.10.2
```
this fails
```
         pkg loaded version
 1:  ggplot2   TRUE   3.0.0
 5:      xts   TRUE  0.11.1
```
with
```
Error in eval(exprs[i], envir) : 
  4 errors out of 23 in 7.404sec on Fri Sep 14 11:06:59 2018. [endian==little, sizeof(long double)==16, sizeof(pointer)==8, TZ=Etc/UTC, locale='LC_CTYPE=en_US.UTF-8;LC_NUMERIC=C;LC_TIME=en_US.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONETARY=en_US.UTF-8;LC_MESSAGES=en_US.UTF-8;LC_PAPER=en_US.UTF-8;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=en_US.UTF-8;LC_IDENTIFICATION=C']. Search inst/tests/other.Rraw for test numbers: 1.1, 1.3, 1.4, 6.4.
```
https://github.com/Rdatatable/data.table/blob/master/inst/tests/other.Rraw

```
 test(1.1, names(print(ggplot(DT,aes(b,f))+geom_point()))[c(1,3)], c("data","plot"))
    test(1.3, names(print(ggplot(DT) + geom_hex(aes(b, f)) + facet_wrap(~grp)))[c(1,3)], c("data","plot"))
  # Test plotting ITime with ggplot2 which seems to require an as.data.frame method for ITime, #1713
  datetimes = c("2011 NOV18 09:29:16", "2011 NOV18 10:42:40", "2011 NOV18 23:47:12",
                "2011 NOV19 01:06:01", "2011 NOV19 11:35:34", "2011 NOV19 11:51:09")
  DT = IDateTime(strptime(datetimes,"%Y %b%d %H:%M:%S"))
  test(1.4, print(DT[,qplot(idate,itime)])$ranges, error="geom_point requires the following missing aesthetics: y")  # y=print(qplot(DT$idate,DT$itime))$ranges
```

```
test(6.4, xts::last(DT), 9L)        # would prefer last row, as data.table::last does
```

I am able to reproduce ggplot2 issue locally but not xts.

I see entry in xts news file
```
o  first() and last() now keep dims when they would otherwise be dropped by a
   regular row subset. This is consistent with head() and tail(). Thanks to
   Davis Vaughan for the report (#226).
```

jangorecki:
Closes #3047

### DIFF
--- a/inst/tests/other.Rraw
+++ b/inst/tests/other.Rraw
@@ -46,7 +46,7 @@ if (loaded[["ggplot2"]]) {
   datetimes = c("2011 NOV18 09:29:16", "2011 NOV18 10:42:40", "2011 NOV18 23:47:12",
                 "2011 NOV19 01:06:01", "2011 NOV19 11:35:34", "2011 NOV19 11:51:09")
   DT = IDateTime(strptime(datetimes,"%Y %b%d %H:%M:%S"))
-  test(1.4, tryCatch(print(DT[,qplot(idate,itime)])$ranges, message=function(m) m[["message"]]), "Don't know how to automatically pick scale for object of type ITime. Defaulting to continuous.\n") # updated as breaking change introduced to ggplot2, see #3047
+  test(1.4, print(DT[,qplot(idate,itime)])$ranges, message="Don't know how to automatically pick scale for object of type ITime. Defaulting to continuous") # ggplot changed; see #3047
   test(1.5, print(DT[,qplot(idate,as.POSIXct(itime,tzone=""))])$ranges, print(qplot(idate,as.POSIXct(itime,tzone=""),data=DT))$ranges)
   try(graphics.off(),silent=TRUE)
 }
@@ -215,7 +215,7 @@ if (memtest) {
   } else {
     warning("test.data.table runs with memory testing but did not collect any memory statistics.", call.=FALSE)
   }
-} 
+}
 if (nfail > 0) {
   if (nfail>1) {s1="s";s2="s: "} else {s1="";s2=" "}
   cat("\r")
--- a/inst/tests/other.Rraw
+++ b/inst/tests/other.Rraw
@@ -36,17 +36,17 @@ if (all(c("package:reshape","package:reshape2") %in% search())) {
 
 if (loaded[["ggplot2"]]) {
   DT = data.table( a=1:5, b=11:50, d=c("A","B","C","D"), f=1:5, grp=1:5 )
-  test(1.1, names(print(ggplot(DT,aes(b,f))+geom_point()))[c(1,3)], c("data","plot"))
+  test(1.1, names(print(ggplot(DT,aes(b,f))+geom_point()))[c(1,3)], c("data","scales")) # update as described in #3047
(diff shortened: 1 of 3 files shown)


inst/tests/other.Rraw
```diff
@@ -36,17 +36,17 @@ if (all(c("package:reshape","package:reshape2") %in% search())) {
...
-    test(1.3, names(print(ggplot(DT) + geom_hex(aes(b, f)) + facet_wrap(~grp)))[c(1,3)], c("data","plot"))
+    test(1.3, names(print(ggplot(DT) + geom_hex(aes(b, f)) + facet_wrap(~grp)))[c(1,3)], c("data","scales"))
   }
   # Test plotting ITime with ggplot2 which seems to require an as.data.frame method for ITime, #1713
   datetimes = c("2011 NOV18 09:29:16", "2011 NOV18 10:42:40", "2011 NOV18 23:47:12",
                 "2011 NOV19 01:06:01", "2011 NOV19 11:35:34", "2011 NOV19 11:51:09")
   DT = IDateTime(strptime(datetimes,"%Y %b%d %H:%M:%S"))
-  test(1.4, print(DT[,qplot(idate,itime)])$ranges, error="geom_point requires the following missing aesthetics: y")  # y=print(qplot(DT$idate,DT$itime))$ranges
+  test(1.4, tryCatch(print(DT[,qplot(idate,itime)])$ranges, message=function(m) m[["message"]]), "Don't know how to automatically pick scale for object of type ITime. Defaulting to continuous.\n") # updated as breaking change introduced to ggplot2, see #3047
   test(1.5, print(DT[,qplot(idate,as.POSIXct(itime,tzone=""))])$ranges, print(qplot(idate,as.POSIXct(itime,tzone=""),data=DT))$ranges)
```
MichaelChirico: we don't have a message argument to test?

mattdowle: Never really used `message()` until recently.  Good idea to add `message=` to `test()`.  `capture.output()` can capture output or message stream, but not both at the same time I discovered recently.  What springs to mind first is that `test()` could accept `output=` or `message=` and it could do it using `capture.output()` rather than `tryCatch`.

But why are we testing that ggplot returns this message anyway?  If they improve ggplot and the message goes away,  that'll be a breaking change for others.Rraw again.  But is that something we want to break on?  Same goes for why we were testing for the error.  I did that I know,  and I'm asking myself.  Maybe Jan thought I must have had a reason and followed my (bad) lead.

jangorecki: I agree it is better to remove those tests are the risk of breaking in future in high

Issue #2341, opened 2017-09-07
`[.data.table` requires import of other function to work when data.table is imported by package


sz-cgt:
When writing packages, I generally declare an Imports dependency in the DESCRIPTION then use fully qualified names when invoking the package functions, e.g. data.table::fread().

However, this method fails to import the S3 methods into a namespace that my package can use.  Specifically, `[` fails and falls back to the `[.data.frame` unless I explicitly import at least one function from the data.table package in my NAMESPACE.  While I can add importFrom(data.table, ...), it seems a bit of a hack to me.

Is this a data.table specific issue, or a general failing in R that requires with regard to S3 methods in package namespaces?

franknarf1:
Not sure if this addresses it, but the usual reference is the final question of the FAQ: https://stackoverflow.com/questions/10527072/using-data-table-package-inside-my-own-package/10529888#10529888

sz-cgt:
The final FAQ covers it but is an ugly hack in my opinion.  I should not have to pollute my package's namespace.

As I said in the original post, , importing a single symbol from the data.table package fixes the problem.  I guess that's because the import() and importFrom() in the NAMESPACE generally do the same thing, with importFrom() having an additional final operation (filtering symbols).

For now I'll live with importFrom(data.table,data.table)

close #2341

sz-cgt:
One last remark: apparently the directions in the data.table::cedata comments (lines 17-21) also work.  Adding

`.datatable.aware = TRUE`

to my package also fixes the issue.  Should this be part of the FAQ?

nutle:
> One last remark: apparently the directions in the data.table::cedata comments (lines 17-21) also work. Adding
> 
> `.datatable.aware = TRUE`
> 
> to my package also fixes the issue. Should this be part of the FAQ?

Where and how exactly do you add this line? Adding this straightforwardly to NAMESPACE returns me 

    Error: unknown namespace directive: TRUE

Adding this anywhere else does not work, only the ``importFrom(data.table,data.table)`` worked for me.

MichaelChirico:
@jangorecki is this part of the importing DT vignette?

sz-cgt:
You need to include that line in a .R file, not NAMESPACE.  I usually put it in the file that has my package documentation, `@importFrom` statements etc.  For example:

```
# Import package operators
#' @importFrom magrittr "%>%" "%<>%"
#' @importFrom data.table ":=" "%like%" "%between%"
#' @export
[...]

PR #3069, opened 2018-09-25
Add some info about cedta to importing vignette

MichaelChirico:
@jangorecki I forget if we're compiling the vignettes ourselves? Or if compilation is happening in CI?

Inspired by #2341

### DIFF
--- a/vignettes/datatable-importing.Rmd
+++ b/vignettes/datatable-importing.Rmd
@@ -29,7 +29,7 @@ It is very easy to use `data.table` as a dependency due to the fact that `data.t
 
 ## `DESCRIPTION` file {DESCRIPTION}
 
-The first place to define a dependency in a package is the `DESCRIPTION` file. Most commonly, you will need to add `data.table` under the `Imports:` field. Doing so will necessitate an installation of `data.table` before your package can compile/install. As mentioned above, no other packages will be installed because `data.table` does not have any dependencies of its own. You can also specify the minimal required version of a dependency; for example, if your package is using the `fwrite` function, which was introduced in `data.table` in version 1.9.8, you should incorporate this as `Imports: data.table (>= 1.9.8)`. This way you can ensure that the version of `data.table` installed is 1.9.8 or later before your users will be able to install your package. Besides the `Imports:` field, you can also use `Depends: data.table`; we discourage this approach because it forces the end users of your package to use `data.table`.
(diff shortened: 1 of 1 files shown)


jangorecki:
vignette is being compiled on R CMD build, but we do not push it to repo

vignettes/datatable-importing.Rmd
```diff
@@ -148,12 +160,23 @@ my.write = function (x) {
   }
 }
 ```
-When using package as a suggested dependency you should not import it in NAMESPACE file, just mention it in DESCRIPTION file. You also have to use `data.table::` prefix when calling data.table functions because none of them are imported.
+
+When using a package as a suggested dependency, you should not `import` it in the `NAMESPACE` file. Just mention it in the `DESCRIPTION` file. You also _must_ use the `data.table::` prefix when calling `data.table` functions because none of them are imported.
+
+## `data.table` in `Imports:` but nothing exported
```
jangorecki: "nothing imported" seems more appropriate, and I would remove `:` from `Imports:`

vignettes/datatable-importing.Rmd
```diff
@@ -148,12 +160,23 @@ my.write = function (x) {
...
+
+Some users may prefer ([e.g.](https://github.com/Rdatatable/data.table/issues/2341)) may prefer to eschew using `importFrom` or `import` in their `NAMESPACE` file and instead use `data.table::` qualification on all internal code (of course keeping `data.table` under their `Imports:` in `DESCRIPTION`).
+
+In this case, the un-exported function `[.data.table` will revert to calling `[.data.frame` since `data.table` has no way of knowing that the parent package is aware it's attempting to make calls against the syntax of `data.table`'s query API (which could lead to unexpected errors).
```
jangorecki: "unexpected errors" could be clarified, it means that API for data.frame and dt is different. Reader might got impression there are some non-handled, non-recognized (unexpected) errors occurring.

MichaelChirico: Will do. Admittedly I'm flying a _bit_ by the seat of my pants here as I don't have _deep_ understanding of what's going on, just the basic idea.

vignettes/datatable-importing.Rmd
MichaelChirico: Yes, good

PR #3074, opened 2018-09-26
Added delayed S3 for R-devel

### LINKED ISSUE #3070: S3 method request from CRAN
I see you have S3 methods which you export but do not register:
```
     From         Package    Generic   Method             
[1,] "data.table" "xts"      "as.xts"  "as.xts.data.table"
[2,] "data.table" "zoo"      "as.Date" "as.Date.IDate"    
[3,] "data.table" "reshape2" "melt"    "melt.data.table"  
```
For the current development version of R, we have recently added a
simple mechanism for delayed S3 registration: using 
```
   if(getRversion() >= "3.6.0") {
       S3method(pkg::gen, cls)
   }
```
will register 'gen.cls' as S3 method for class 'cls' and generic 'gen'
from package 'pkg' only when the namespace of 'pkg' is loaded.
(Similarly, one can use
```
   if(getRversion() >= "3.6.0") {
       S3method(pkg::gen, cls, fun)
   }
```
in case the method is named 'fun' and not 'gen.cls'.)

Can you please add the necessary delayed S3 method registrations to your
NAMESPACE, and submit a new version of your package as quickly as 
possible?

I guess that applies to 1 and 3: not sure about 2?

mattdowle:
Closes #3070

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -29,7 +29,7 @@ export(rleid)
 export(rleidv)
 export(rowid)
 export(rowidv)
-export(as.xts.data.table)
+
 export(uniqueN)
 export(setDTthreads, getDTthreads)
 # set operators
@@ -89,17 +89,18 @@ S3method(within, data.table)
 S3method(is.na, data.table)
 S3method(format, data.table)
 S3method(Ops, data.table)
-
 S3method(anyDuplicated, data.table)
 
-export(split.data.table)
 S3method(split, data.table)
 export(melt)
-export(melt.data.table)
-S3method(melt, data.table)
+if(getRversion() >= "3.6.0") S3method(reshape2::melt, data.table) else S3method(melt, data.table)
 export(dcast)
-export(dcast.data.table)
 S3method(dcast, data.table)
+if(getRversion() >= "3.6.0") {
+  S3method(xts::as.xts, data.table)
+} else {
+  export(as.xts.data.table)
+}
 
 import(utils)
 S3method(update, dev.pkg)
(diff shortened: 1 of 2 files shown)


NAMESPACE
```diff
@@ -89,17 +89,18 @@ S3method(within, data.table)
...
 S3method(format, data.table)
 S3method(Ops, data.table)
-
 S3method(anyDuplicated, data.table)
 
-export(split.data.table)
 S3method(split, data.table)
 export(melt)
-export(melt.data.table)
-S3method(melt, data.table)
```
MichaelChirico: still need `S3method(melt, data.table)`, no?

mattdowle: still there, just moved

NAMESPACE
```diff
@@ -30,6 +30,7 @@ export(rleidv)
 export(rowid)
 export(rowidv)
 export(as.xts.data.table)
+if(getRversion() >= "3.6.0") S3method(xts::as.xts, data.table)
```
jangorecki: and how that refers to the fact we do not import/depend xts? maybe we can/should add `requireNamespace` there too?

mattdowle: but then xts wouldn't be optional? wouldn't it have to move from suggests to imports?

mattdowle: I'm wondering what as.xts.data.table even does and why leaving it to as.xts.data.frame won't do. Maybe we can remove it.

jangorecki: I don't know how R parse NAMESPACE in regards to `::` and dependencies defined in `DESCRIPTION`. Doing `if(getRversion() >= "3.6.0" && requireNamespace("xts", silently=TRUE))` wouldn't make `xts` mandatory, `requireNamespace` returns FALSE when no package found. Maybe `requireNamespace` is not allowed in NAMESPACE, don't know that. I don't see any warnings on any build. So seems legit.

jangorecki: The fact is that every use of Suggested dependencies was always escaped with `requireNamespace` and here is the first place AFAIK where it isn't escaped.

jangorecki: I escape I mean not try to access xts namespace if it is not available. I haven't heard about that new mechanism, if so the check for R 3.6 handles it already.

Issue #3056, opened 2018-09-20
Bind package version into .dll/.so


mattdowle:
So that the R package still using an old .dll on Windows can't happen.  Almost every time we upgrade, someone somewhere on Windows hits this problem; e.g. #3055.
The worry is that it happens more than we know, but users get wrong results silently.  It's quite fortunate when a mismatch is severe enough to cause an error, so at least the user knows something is wrong.
 
But how to achieve this?  It might mean checking the version matches between R and .dll on each and every call, which might be expensive.

Maybe it's better to fix the issue upstream in R itself (install.packages) or RStudio if it's an RStudio-only problem.  Since it must affect other packages too.  We need someone with i) the will and ii) Windows, to solve it once and for all.

shrektan:
I have some experience on this and it's true I have to be very careful when updating the R package that uses compiled codes on Windows. 

Whenever the R package with complied code is  being loaded, the dll file will be locked on Windows and can't be unlinked. If you are trying to upgrade the package via `install.package()` at the same time, R will complain that the dll file can't be unlinked but the package itself will be installed anyway, which causes a lot of pain for R users, because we normally open multiple sessions... So whenever I want to upgrade my package, I have to close all of the R scripts that being running in advance. 

It's **not** an RStudio only issue, I can confirm it will happen even outside of RStudio.

![installpackage](https://user-images.githubusercontent.com/8368933/45862595-0111d700-bda5-11e8-9ce7-d64ac23e0204.JPG)

shrektan:
I think we have two options on this:

1. Report it to R Core and hope they will fix this in `install.packages()`
1. As you said, bind the version info in the dll file. However, I think we only need to check it whenever the data.table package is being loaded (e.g. check the version in `.onLoad(libname, pkgname)`)

jangorecki:
@shrektan thanks for confirming in R session. Your proposed solution looks interesting!
Alternatively, probably not very optimal way would be to copy dll on package load, and use a fresh copy, so original `dll` is never locked. It could eventually re-used already copied dll. Copy would append pkg version like  `data.table_1.11.6.dll` then any session that attempts to load that version of pkg will not need to copy dll but only re-use already copied. So locks would be made only on `pkg_ver.dll` files, and `install.packages` would always update `pkg.dll`. This seems to be safe, but as said might not be optimal.

MichaelChirico:
[...]

PR #3088, opened 2018-10-01
Check DLL MD5

mattdowle:
Attempt at #3056


MichaelChirico:
Very cool, nice idea using `MD5sums` and glad there's something in `base` for it.

PR #3099, opened 2018-10-05
new error when on= provided but not i=

mattdowle:


franknarf1:
Fyi, this also triggers without `on=`, when I'm interactively typing a command and decide that I want to see the table before going further, like...

```
library(data.table)
# data.table 1.11.9 IN DEVELOPMENT built 2018-10-06 03:32:28 UTC; root  Latest news: r-datatable.com
DT = data.table()
DT[,]
# Error in `[.data.table`(DT, , ) : 
#   When i and j are both missing, no other argument should be used. Empty [] is useful after := to have the result printed.
```

PR #3105, opened 2018-10-12
Replace %in% with %chin% where appropriate

MichaelChirico:
This installs & builds but `test.data.table()` segfaults:

```
test.data.table()
Running /Users/michael.chirico/Library/R/3.5/library/data.table/tests/tests.Rraw 

 *** caught segfault ***
address 0x0, cause 'memory not mapped'

Traceback:
 1: data.table(ID = seq_len(3000L), time = 0, nTest = 0L)
 2: assign("timings", data.table(ID = seq_len(3000L), time = 0, nTest = 0L),     envir = env)
 3: test.data.table()
```

Within `data.table`, the bug comes almost immediately at:

```
.Call(CcopyNamedInList,x)
```

### DIFF
--- a/CRAN_Release.cmd
+++ b/CRAN_Release.cmd
@@ -191,7 +191,7 @@ cd ~/GitHub/data.table
 Rdevel-strict CMD INSTALL data.table_1.11.8.tar.gz
 # Check UBSAN and ASAN flags appear in compiler output above. Rdevel was compiled with them so should be passed through to here
 Rdevel-strict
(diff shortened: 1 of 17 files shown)


mattdowle:
After that fix for the all-NA case I ran it through Rdevel-strict with ASAN and strict-barrier: passed ok.

MichaelChirico:
OK. there was some other error on RStudio-based development again... switched to terminal and things were working again but hadn't come back to here & haven't had time to investigate the RStudio issue.

LGTM & ready to merge!

PR #3107, opened 2018-10-12
Better error message for unmatched patterns() in melt()

### LINKED ISSUE #3106: melt fails when pattern not found
```
DT = data.table(id = 1:3, a1 = rnorm(10), a2 = rnorm(10))
melt(DT, id.vars = 'id', measure.vars = patterns(a = 'a', b = 'b'))
```

> Error in melt.data.table(DT, id.vars = "id", measure.vars = patterns(a = "a",  : 
>  Unsupported type 'NULL'

Quite a confusing error

MichaelChirico:
Closes #3106


MichaelChirico:
poking around coverage a bit, i notice a bunch of misses appear to be around "internal error"s like:

https://codecov.io/gh/Rdatatable/data.table/src/cd93a0642f51044adccaf36fea84acac4c85d3fa/src/ijoin.c

Should we `#nocov` these errors? Seeing as how the whole point of them is that we don't know how any use case could end up there..

PR #3108, opened 2018-10-13
nocov for internal errors & unimplemented foverlaps feature

MichaelChirico:
@arunsrinivasan is it correct that `foverlaps` `type = 'equal'` is un-implemented? I see a bunch of code in `ijoin.c` related to `case EQUAL`, I guess it's just incomplete?

### DIFF
--- a/R/IDateTime.R
+++ b/R/IDateTime.R
@@ -84,7 +84,7 @@ round.IDate <- function (x, digits=c("weeks", "months", "quarters", "years"), ..
   if (!inherits(e1, "IDate"))
     stop("can only subtract from \"IDate\" objects")
   if (storage.mode(e1) != "integer")
-    stop("Internal error: storage mode of IDate is somehow no longer integer")
+    stop("Internal error: storage mode of IDate is somehow no longer integer") # nocov
   if (nargs() == 1L)
     stop("unary - is not defined for \"IDate\" objects")
   if (inherits(e2, "difftime"))
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -322,7 +322,7 @@ chmatch2 <- function(x, table, nomatch=NA_integer_) {
         ..syms = av
       }
     } else if (is.name(jsub)) {
-      if (substring(jsub, 1L, 2L) == "..") stop("Internal error:  DT[, ..var] should be dealt with by the branch above now.")
+      if (substring(jsub, 1L, 2L) == "..") stop("Internal error:  DT[, ..var] should be dealt with by the branch above now.") # nocov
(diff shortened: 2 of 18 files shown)


mattdowle:
Wait, so `// # nocov` works on C code now in covr? When did that become possible?

MichaelChirico:
Hmm I saw it in master already, didn't think twice about it:

https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L149

Though it doesn't appear to work perfectly as this PR has also reduced the number of hit lines apparently?

mattdowle:
> Though it doesn't appear to work perfectly as this PR has also reduced the number of hit lines apparently?
Yeah I've noticed this before in code cov results. Not sure but it feels like when we touch a file, that file's misses get counted as new misses even though they were missed before.  What I've sometimes done is add new tests to cover those cases on the basis that the coverage needs adding anyway when I've been in that part of the code anyway.  Other times, when it would mean too much work right now (like now) I just leave it to later.  It's not as though those misses will be forgotten because we always have the overall 9% missed to tackle at any time.

MichaelChirico:
now 8% 😎

mattdowle:
> Hmm I saw it in master already, didn't think twice about it:
> https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L149

But that one doesn't work : 
image

Maybe it must be `// # nocov` with the `#` like that, as you did.

mattdowle:
Wow - look at that.  Indeed added the hash in 25224ab made the difference.  I compared codecov results before and after and the coverage went up by that single line!  You're saying you discovered this by accident @MichaelChirico ?!

mattdowle:
Yep : master coverage (independent of diff) confirms that fread.c line is now covered. The `#` must be present in `\\ # nocov` then.
image
Yay!

MichaelChirico:
seems so 😅

I googled around aimlessly a little before... seems like a bug that // nocov wouldn't work but // # nocov does 🤔

MichaelChirico:
@mattdowle FWIW according to `covr`'s [`README`](https://github.com/r-lib/covr/blob/master/README.md),

> you can exclude lines from the coverage by putting special comments in your source code.

and

> The patterns used can be specified by setting the global options `covr.exclude_pattern`, `covr.exclude_start`, `covr.exclude_end`.

So, I `covr` just scans every file in every package directory and ignores based on these options.

I guess we could also define our own `covr.exclude_pattern` to be `'#|//'` to make the C `nocov` sections look more natural, but that's probably overkill.

mattdowle:
Defining `covr.exclude_pattern` sounds good to me. If it works, you could raise PR to covr to change the default.

PR #3110, opened 2018-10-17
Export is_na()/any_na()

### LINKED ISSUE #3016: Export is_na()
Is there a reason why [`is_na()` and `any_na()`](https://github.com/Rdatatable/data.table/blob/90d3736ffc5e172a4e33f4f00bfbc84b898b3b37/R/data.table.R#L2253-L2258) are not exported (apart from the mssing argument checks as mentioned)? I need `base::complete.cases()` functionality for a `data.table` and the only efficient way I found was using `data.tables:::is_na()`. Or am I missing something? Happy to submit a PR.

nbenn:
Closes #3016.
I'm happy to add a sentence or two to the man pages. Any suggestions where this would fit in best? To have an entry dedicated to this seems excessive, no? 

---
Edit from Matt
In this PR I deleted setrev too. An unrelated and unused internal function.  The PR morphed a bit (I'll change the title) because is_na and any_na usage internally needed bringing up to date, leading to as_list not being needed anymore too, etc.  See comments below.


MichaelChirico:
Looks good! Can you add some tests of expected failures of these functions as well?

MichaelChirico:
About the manual page, you might consider adding it to `na.omit.data.table.Rd`, though I think it does make sense for it to have its own page (one might imagine other similar utilities being added to this page eventually)

nbenn:
Fixes #3016.

nbenn:
@mattdowle: are these questions directed at me? In case they are:

> Are these the best names to export?

I don't think so tbh. `is_na()` is not very telling. Something like `any_na_per_row()` would be more informative (but also a bit unwieldy). I think the 'base' R terminology is `stats::complete.cases()`. `any_na()` is ok imo. But this really is your decision. I just went with the status quo.

> Should they be generic?

What should be generic here? `any_na_per_row()`, so that `dplyr` could implement this for a `tibble` as well? Is that likely to happen? Where should this generic function live? In `data.table`? This makes little sense to me.

> Why isn't base R's anyNA generic and could we submit patch to R to make it generic?

Base R's anyNA *is* generic. Are you suggesting we do something like

```r
anyNA.data.table <- function(x, mode = c("all", "row_wise", "col_wise"),
                             columns = seq_along(x), ...) {

  mode <- match.arg(mode)

  if (mode == "all") {
    any_na(x, columns)
  } else if (mode == "row_wise") {
    is_na(x, columns)
  } else {
    vapply(x[, columns], is.na, logical(1L))
  }
}
```

> What's the name used in dplyr / tibble?

I don't know, I'm no `dplyr` expert. I don't think there is a function for row-wise `anyNA()` in `dplyr`. All I can come up with is 

```r
df %>% complete.cases
df %>% is.na %>% rowSums == 0
```

man/na.omit.data.table.Rd
```diff
@@ -1,41 +1,59 @@
...
 \alias{na.omit}
-\title{ Remove rows with missing values on columns specified }
+\alias{is_na}
+\alias{any_na}
+\title{Find missing values per row and remove rows with missing values on columns specified}
 \description{
-    This is a \code{data.table} method for the S3 generic \code{stats::na.omit}. The internals are written in C for speed. See examples for benchmark timings.
-
-    \code{bit64::integer64} type is also supported.
+  \code{is_na()} checks a \code{data.table} for \code{NA} values in row-wise fashion, returning \code{TRUE} if a given row contains a \code{NA} entry and \code{FALSE} otherwise. This corresponds to \code{rowSums(is.na(.))}, but implemented in \code{C} for speed. \code{any_na()} returns the equivalent of \code{any(is_na(.))}, but again is implemented in \code{C} for speed. In order to remove rows containing \code{NA} entries, an S3 generic function to \code{stats::na.omit} if provided for \code{data.table} objects. This internally calls \code{is_na()}. See examples for benchmark timings. \code{bit64::integer64} type is also supported.
```
st-pasha:
```suggestion
  \code{is_na(DT)} checks the \code{data.table} for \code{NA} values in row-wise fashion, returning a logical vector of length `nrow(DT)`, with each element being \code{TRUE} for rows contains an \code{NA} entry and \code{FALSE} otherwise. This is equivalent to \code{as.logical(rowSums(is.na(DT)))}, but implemented in \code{C} for speed. \code{any_na(DT)} returns the equivalent of \code{any(is_na(DT))}, but again is implemented in \code{C} for speed. In order to remove rows containing \code{NA} entries, an S3 generic function to \code{stats::na.omit} if provided for \code{data.table} objects. This internally calls \code{is_na()}. See examples for benchmark timings. \code{bit64::integer64} type is also supported.
```

st-pasha:
Thanks, @nbenn, for the PR. Any new functionality is always great.
I posted above a suggestion on how the documentation can be clarified with respect to the return value of `is_na()`.

As for the name of the function itself, I too agree that it is confusing. When the function was private it didn't really matter that it was like that. However, the best time to think about the API is at the point of releasing it for the first time: any subsequent changes would be much more difficult.

Specifically, I find `is_na()` confusing because it looks and sounds like `is.na()`, and therefore people will assume that this is basically the same function. People don't read the documentation if they think they can guess what a function does. 

As an alternative, may I suggest `ranyna()` or `rhasnas()`. This stands for **R**owwise **any na**, and is inspired by base functions `rbind / cbind`. At some point we might also add `canyna()`, which would be the equivalent of `as.logical(colSums(is.na(DT)))`.

nbenn:
How do people feel about extending the `anyNA()` generic? Dumb idea? I don't know.. I kinda like it. But this is ultimately not my decision. I'll just go with what you guys agree upon.

@st-pasha How do you feel about some underscores in there? E.g. `r_any_na()`? I find `ranyna()` or `canyna()` a bit hard to read. I do appreciate the brevity though.

MichaelChirico:
I don't like `ranyna`.... the parallel to `rbind`/`cbind` doesn't jump out to me.

If the output is the same as `complete.cases` shouldn't we just copy that?

Maybe `incomplete_row`? `is_incomplete_row`?

HughParsonage:
On the name: I think `complete_cases` and `all_complete_cases` is probably best (obviously this would be `!is_na` and `!any_na`).

That being said, I'm conscious of @st-pasha's good principle of debating the worth of an issue on the issue page, but as the PR is already here...

I'm not seeing much of speed benefit from these functions.  For example `is_na` appears slower than `complete.cases`. Doesn't make the function worthless -- but might make it not worth exporting.

``` r
library(data.table)
# ?fread
# Demo speed-up
n = 1e8
DT = data.table( a=sample(1:1000,n,replace=TRUE),
                 b=sample(1:1000,n,replace=TRUE),
                 c=rnorm(n),
                 d=sample(c("foo","bar","baz","qux","quux"),n,replace=TRUE),
                 e=rnorm(n),
                 f=sample(1:1000,n,replace=TRUE) )
DT[2,b:=NA_integer_]
DT[4,c:=NA_real_]
DT[3,d:=NA_character_]
DT[5,d:=""]
DT[2,e:=+Inf]
DT[3,e:=-Inf]

is_na <- data.table:::is_na
cc <- stats::complete.cases

bench::mark(!cc(DT),  # ! means slight disadvantage to cc
            is_na(DT))
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 x 10
#>   expression   min  mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 !cc(DT)    2.37s 2.37s  2.37s 2.37s     0.423     763MB     1     1
#> 2 is_na(DT)  2.51s 2.51s  2.51s 2.51s     0.398     381MB     0     1
#> # ... with 1 more variable: total_time <bch:tm>
do_coalesce <- function(x) hutils::coalesce(x, first(x[complete.cases(x)]))
DT_no_NA <- DT[, lapply(.SD, do_coalesce)]

bench::mark(!cc(DT_no_NA), 
            is_na(DT_no_NA))
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 x 10
#>   expression   min  mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 !cc(DT_no~  2.5s  2.5s   2.5s  2.5s     0.400     763MB     1     1
#> 2 is_na(DT_~ 2.52s 2.52s  2.52s 2.52s     0.396     381MB     0     1
#> # ... with 1 more variable: total_time <bch:tm>

DT_all_NA <- data.table(x = rep(NA, 1e8))
for (i in 1:50) {
  set(DT_all_NA, j = paste0("V", i), value = rep(NA_character_, 1e8))
}
bench::mark(!cc(DT_all_NA), 
            is_na(DT_all_NA))
#> # A tibble: 2 x 10
#>   expression   min  mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 !cc(DT_al~ 22.7s 22.7s  22.7s 22.7s    0.0440     763MB     0     1
#> 2 is_na(DT_~ 22.6s 22.6s  22.6s 22.6s    0.0442     381MB     0     1
#> # ... with 1 more variable: total_time <bch:tm>
```

<sup>Created on 2018-11-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

``` r
devtools::session_info()
#>  setting  value                       
#>  version  R version 3.5.1 (2018-07-02)
#>  os       Windows 10 x64              
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_Australia.1252      
#>  ctype    English_Australia.1252      
#>  tz       Australia/Sydney            
#>  date     2018-11-01                  
#> 
#>  package     * version date       lib source        
#>  assertthat    0.2.0   2017-04-11 [1] CRAN (R 3.5.1)
#>  backports     1.1.2   2017-12-13 [1] CRAN (R 3.5.0)
#>  base64enc     0.1-3   2015-07-28 [1] CRAN (R 3.5.0)
#>  bench         1.0.1   2018-06-06 [1] CRAN (R 3.5.1)
#>  callr         3.0.0   2018-08-24 [1] CRAN (R 3.5.1)
#>  cli           1.0.1   2018-09-25 [1] CRAN (R 3.5.1)
#>  crayon        1.3.4   2017-09-16 [1] CRAN (R 3.5.1)
#>  data.table  * 1.11.8  2018-09-30 [1] CRAN (R 3.5.1)
#>  debugme       1.1.0   2017-10-22 [1] CRAN (R 3.5.1)
#>  desc          1.2.0   2018-05-01 [1] CRAN (R 3.5.1)
#>  devtools      2.0.0   2018-10-19 [1] CRAN (R 3.5.1)
#>  digest        0.6.18  2018-10-10 [1] CRAN (R 3.5.1)
#>  evaluate      0.12    2018-10-09 [1] CRAN (R 3.5.1)
#>  fansi         0.4.0   2018-10-05 [1] CRAN (R 3.5.1)
#>  fastmatch     1.1-0   2017-01-28 [1] CRAN (R 3.5.0)
#>  fs            1.2.6   2018-08-23 [1] CRAN (R 3.5.1)
#>  glue          1.3.0   2018-07-17 [1] CRAN (R 3.5.1)
#>  htmltools     0.3.6   2017-04-28 [1] CRAN (R 3.5.1)
#>  hutils        1.2.0   2018-10-31 [1] local         
#>  knitr         1.20    2018-02-20 [1] CRAN (R 3.5.1)
#>  magrittr      1.5     2014-11-22 [1] CRAN (R 3.5.1)
#>  memoise       1.1.0   2017-04-21 [1] CRAN (R 3.5.1)
#>  pillar        1.3.0   2018-07-14 [1] CRAN (R 3.5.1)
#>  pkgbuild      1.0.2   2018-10-16 [1] CRAN (R 3.5.1)
#>  pkgload       1.0.1   2018-10-11 [1] CRAN (R 3.5.1)
#>  prettyunits   1.0.2   2015-07-13 [1] CRAN (R 3.5.1)
#>  processx      3.2.0   2018-08-16 [1] CRAN (R 3.5.1)
#>  profmem       0.5.0   2018-01-30 [1] CRAN (R 3.5.1)
#>  ps            1.2.0   2018-10-16 [1] CRAN (R 3.5.1)
#>  R6            2.3.0   2018-10-04 [1] CRAN (R 3.5.1)
#>  Rcpp          0.12.19 2018-10-01 [1] CRAN (R 3.5.1)
#>  remotes       2.0.1   2018-10-19 [1] CRAN (R 3.5.1)
#>  rlang         0.3.0   2018-10-22 [1] CRAN (R 3.5.1)
#>  rmarkdown     1.10    2018-06-11 [1] CRAN (R 3.5.1)
#>  rprojroot     1.3-2   2018-01-03 [1] CRAN (R 3.5.1)
#>  sessioninfo   1.1.0   2018-09-25 [1] CRAN (R 3.5.1)
#>  stringi       1.2.4   2018-07-20 [1] CRAN (R 3.5.1)
#>  stringr       1.3.1   2018-05-10 [1] CRAN (R 3.5.1)
#>  testthat      2.0.1   2018-10-13 [1] CRAN (R 3.5.1)
#>  tibble        1.4.2   2018-01-22 [1] CRAN (R 3.5.1)
#>  usethis       1.4.0   2018-08-14 [1] CRAN (R 3.5.1)
#>  utf8          1.1.4   2018-05-24 [1] CRAN (R 3.5.1)
#>  withr         2.1.2   2018-03-15 [1] CRAN (R 3.5.1)
#>  yaml          2.2.0   2018-07-25 [1] CRAN (R 3.5.1)
#> 
#> [1] C:/R/R-3.5.1/library
```

</details>

mattdowle:
@nbenn Yes the questions were directed at anyone watching, including you. I didn't know that `is.na` and `anyNA` were generic in R.  I had looked at their definitions, saw the `.Primitive()` call and no `UseMethod()` and assumed they weren't generic, forgetting about internally-generic functions like these. Thanks for setting me straight.

Given this, yes I very much like your suggestion to export `anyNA.data.table` and `is.na.data.table` methods. Especially, the `mode=` and `columns=` controls : 
```
anyNA.data.table <- function(x, mode = c("all", "row_wise", "col_wise"),
                             columns = seq_along(x), ...) {
```
Assuming we all agree exporting these methods is the right approach, that resolves the function name question then. If so, now we move on to the `mode=` and `columns=` argument names.  In other `data.table` functions we've used `by=` rather than `columns=` so should it be consistent?  I like `columns=` though in this case which is probably clearer than `by=`.  Also `mode=` in R is a function that returns the type of the object, so could there be some confusion. Is there any other precedent for such an argument name? I do like the name `mode=` though.

I don't think I 'decide' as such. I just raise points to reach a point that everyone agrees anyway.

The internal usage is minimal and can easily be changed to use the method : 
```
$ grep is_na *.R
data.table.R:is_na <- function(x, by=seq_along(x)) .Call(Cdt_na, x, by)
frank.R:    set(x, j = "..na_prefix..", value = is_na(x, cols))
$ grep any_na *.R
data.table.R:          if (any_na(as_list(xo))) xo = xo[!is.na(xo)]
data.table.R:any_na <- function(x, by=seq_along(x)) .Call(CanyNA, x, by)
data.table.R:      if (any_na(as_list(RHS))) {
setops.R:    if (any_na(as_list(ans))) {
setops.R:    if (any_na(as_list(ans))) {
```

Which raises a question about what `as_list` is internally and if `isNA` and `anyNA` methods should be exported for `list` too:  `isNA.list` and `anyNA.list`.   I don't know if a primitive generic would dispatch to data.table's method for `list` or whether its own method for list would take precedence. I guess the latter.

`complete.cases` is interesting.  It isn't generic though so is that why underscore was suggested (complete_cases).
```
> complete.cases
function (...) 
.External(C_compcases, ...)
<bytecode: 0x5587f0a98fa0>
<environment: namespace:stats>
>
```

I think I still prefer is.na and anyNA methods for data.table with those new controls (`mode` and `columns`) which provides more built-in functionality than complete.cases. Perhaps an `invert=` too to avoid needing the `!` afterwards, if needed.

Although, @HughParsonage found that `complete.cases` is faster than `is_na` but not as memory efficient. The plot thickens!

mattdowle:
Taking a small manageable bite first,  `as_list` seems to be at least 4 years old.  That was from a time when R's `list()` copied named inputs; i.e. `list(x)` copied `x`.  That was changed in R 3.1.0 which we now depend on,  so all use of `as_list()` can be replaced by just `list()` now.

mattdowle:
So it seems I can push to remote forks after all. I cloned the fork separately and then the push (above: https://github.com/Rdatatable/data.table/pull/3110/commits/e83cd587b3bf0395f209190b8e2482a22b8685d7) authenticated easily.

jangorecki:
According to benchmarks above there doesn't seems to be an advantage in exporting those function.

man/na.omit.data.table.Rd
mattdowle: I don't have permission to accept this suggestion on this fork. Seem like a GitHub GUI permission issue, because I can commit to the fork using command line. So I'll commit it separately. Thanks.
image

R/utils.R
```diff
@@ -82,6 +82,32 @@ name_dots = function(...) {
   vnames
 }
 
+col_as_int = function(tbl, cols) {
```
jangorecki: I think we could re-use C function from there: https://github.com/Rdatatable/data.table/blob/master/src/nafill.c#L3
Ideally moving it to new `utils.c` file.
I can do it but best if we merge this foreign-remote based PR so we can push follow ups into Rdatatable remote.

MichaelChirico:
Looking at this PR again after quite long:

 1. `any_na()` is just `anyNA()`; `anyNA.data.frame()` runs `any(vapply(x, anyNA, NA, USE.NAMES = FALSE))` which should be pretty fast. I guess the advantage is the `by=` argument allowing a subset of columns
 2. `is_na()` is just `pany()` which is implemented in #4448 (draft). Interesting to know we already have an internal version of that, not sure I recognized that before. It might be interesting to compare their performance.

Given how far behind master this PR wound up (our fault, sorry about that), I lean towards just closing in light of the above as well as the inconclusive discussion here otherwise.

PR #3125, opened 2018-10-25
fwrite coerces input to data.table when given a matrix

### LINKED ISSUE #2613: fwrite could as.data.table if given a matrix
Not sure how many people would prefer to have `fwrite` automatically know to apply `as.data.table` to a `matrix` (or, better, just accept a `matrix` and handle the "conversion" internally instead of doing so in R, which can be costly).

The former is easy to implement, the latter, a bit harder (since matrices are stored column-wise in memory, we'll have to do some index arithmetic to write row-wise)

fparages:
Closes #2613 

When given a matrix as an input, fwrite now coerces it to a class data.table, to avoid an error in the stopifnot(is.list(x))


jangorecki:
I am not sure if warning is really necessary. Coercing can be slow but I think there is no other faster alternative, so it is still proper way to go.
I would extend tests a little bit more, `matrix("foo")` is kind of edge case of matrix. We should also test matrices of bigger dimensions, like
```r
matrix(1:3, nrow=1, ncol=3)
matrix(1:3, nrow=3, ncol=1)
matrix(1:9, nrow=3, ncol=3)
```
and matrices having `dimnames`, both named columns and rows.

R/fwrite.R
```diff
@@ -26,6 +26,10 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
   nThread = as.integer(nThread)
   # write.csv default is 'double' so fwrite follows suit. write.table's default is 'escape'
   # validate arguments
+  if(is.matrix(x)){ # coerce to data.table if input object is matrix
```
jangorecki: we usually put white space between `if` and `(`.

fparages:
Hi @MichaelChirico @jangorecki
Any update about this PR? Is it good to get merged?
Thanks!!

PR #3153, opened 2018-11-19
Fix scoping bug in split.data.table

### LINKED ISSUE #3151: Bug: split.data.table can not split on a factor column if it is named 'x'
Just found the following bug:
```r
library(data.table)
temp1 <- data.table(x = factor("a"))
split(temp1, by = "x")
# Error: is.data.table(x) is not TRUE

# this works
temp2 <- data.table(.x = factor("a"))
split(temp2, by = ".x")

# this works as well
temp3 <- data.table(x = "a")
split(temp3, by = "x")
```

So the problem only emerges if the splitting variable is called 'x' and it is a factor. The problem can be rooted back to [this](https://github.com/Rdatatable/data.table/blob/55168edde9bd08bb11d6f75545535ea3165a4119/R/data.table.R#L2366) `temp = eval(dtq)` call in `split.data.table`. In the `dtq` unevaluated call, `make.levels(x, cols=.cols, sorted=.sorted)` finds the 'x' variable instead of the 'x' data.table. A quick fix is to write `make.levels(.___x, cols=.cols, sorted=.sorted)` instead, and do a temporary assignment `.___x <- x` before `temp = eval(dtq)`.  

SessionInfo (the bug is also present in 1.11.8):
```r
R version 3.5.1 (2018-07-02)
Running under: Linux Mint 18.2

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] data.table_1.11.9 switchr_0.13.0   

[1] compiler_3.5.1  tools_3.5.1     RCurl_1.95-4.11 remotes_2.0.2  
[5] bitops_1.0-6
```

tdeenes:
Closes #3151
- a cleaner solution compared to the quickfix suggested in the issue
- all previous tests pass
- two new tests added

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,7 +18,7 @@
 
 3. Unmatched `patterns` in `measure.vars` fail early and with feedback, [#3106](https://github.com/Rdatatable/data.table/issues/3092).
 
-4. `split.data.table` failed if `x` had an "x" factor column, closes [#3151](https://github.com/Rdatatable/data.table/issues/3151). Thanks to @tdeenes for reporting and fixing. Tests added.
+4. `split.data.table()` failed if `DT` had a factor column named `"x"`, [#3151](https://github.com/Rdatatable/data.table/issues/3151). Thanks to @tdeenes for reporting and fixing.
 
 #### NOTES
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
LGTM 👍 thanks!

mattdowle:
Thanks @tdeenes. I invited you to be project member so you can create branches directly in the main project now.

PR #3159, opened 2018-11-25
Fix file size display bug in fread

st-pasha:
The bug only manifests for files of size > 1TB. Which is why it was not spotted earlier, and why it's impossible to write tests for it.

The problem was that the expression `(1 << shift)` (when `shift = 40`) was interpreted as `int(1) << 40`, causing an overflow, and as a result incorrect file size was displayed in the output. On the other hand `(1LL << shift)` is `long long int(1) << 40`, which no longer overflows.


mattdowle:
Great thanks. Nice find!

PR #3162, opened 2018-11-28
forder3

mattdowle:
Slow down in dev ordering all-unique input fixed; thanks to @arunsrinivasan's testing.
```
N = 1e8
set.seed(1)
x = sample(N)                       # v1.11.8  master     this PR
system.time(o <- forderv(x))        # 4.5s     5.9s-7.0s  1.2s
!base::is.unsorted(x[o])  # TRUE

x = sample(N/10, N, replace=TRUE)   
system.time(o <- forderv(x))        # 2.6s     1.3s       0.8s
```

Long-standing flip-flop group sizes removed; now recursive and writes directly to final group size stack via thread buffers.
Nested parallelism recently added, removed. skew still handled.
timetaken() in verbose mode now displays both elapsed and cpu time

Looks like AppVeyor timings on this PR are back to normal at about 5 mins (were over 10 min on master) so that's a good sign.

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1463,9 +1463,7 @@ chmatch2 <- function(x, table, nomatch=NA_integer_) {
     if (length(byval) && length(byval[[1L]])) {
       if (!bysameorder && identical(byindex,FALSE)) {
         if (verbose) {last.started.at=proc.time();cat("Finding groups using forderv ... ");flush.console()}
-        # tt = .Internal(Sys.time())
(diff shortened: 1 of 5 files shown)


jangorecki:
those are incredible speed ups! 3-4 times reduced vs. 1.11.8

mattdowle:
@jangorecki fingers crossed. needs battle testing. could you set off a db-bench rerun please to see how it does on server as I missed the 11pm start time.   I'm hoping this will fix the 2nd run being slower in those odd cases, too.

PR #3166, opened 2018-11-30
Support negative values of n in shift

### LINKED ISSUE #1708: shift should accept negative n
I think using "type = lag/lead" is a lot of unnecessary typing, and using a negative shift would be much simpler and more intuitive. It would also allow for more nifty multiple shifts going from lead to lag by typing e.g. `shift(vec, -5:5)`.

The "type" argument can still stay and negative shift would simply reverse the meaning of "type".

`n = 0` should also be accepted and simply return the original vector.

MichaelChirico:
Closes #1708 

My first non-trivial (but also basically trivial) foray into C code so some careful eyes are warranted. I don't think I changed much but still..

Note well the two comments to this PR


src/shift.c
```diff
@@ -23,194 +23,146 @@ SEXP shift(SEXP obj, SEXP k, SEXP fill, SEXP type) {
...
-          for (m=0; m<xrows; m++)
-            SET_VECTOR_ELT(tmp, m, (m < INTEGER(k)[j]) ? VECTOR_ELT(thisfill, 0) : VECTOR_ELT(elem, m - INTEGER(k)[j]));
-          copyMostAttrib(elem, tmp);
-        }
-        break;
-
-      default :
-        error("Unsupported type '%s'", type2char(TYPEOF(elem)));
-      }
-      copyMostAttrib(elem, tmp);
```
MichaelChirico: I honestly thing these lines are vestigial and I think they may be slowing down `shift` unnecessarily... there's no analogue in the `'lead'` branch and I think what it's accomplishing is already done in the `INTSXP` branch.

inst/tests/tests.Rraw
```diff
@@ -12416,6 +12415,28 @@ for (i in 1:4) {
...
+DT = data.table(x = 1:10, y = 10:1)
+
+test(1960.1, shift(DT$x, -1), c(2L, 3L, 4L, 5L, 6L, 7L, 8L, 9L, 10L, NA))
+test(1960.2, shift(DT$x, -1, type = 'lead'),
+     c(NA, 1L, 2L, 3L, 4L, 5L, 6L, 7L, 8L, 9L))
+test(1960.3, shift(DT$x, -1, fill = 0L),
+     c(2L, 3L, 4L, 5L, 6L, 7L, 8L, 9L, 10L, 0L))
+test(1960.4, shift(DT$x, -1, give.names = TRUE),
+     structure(c(2L, 3L, 4L, 5L, 6L, 7L, 8L, 9L, 10L, NA),
+     .Names = c("V1_lag_-1", NA, NA, NA, NA, NA, NA, NA, NA, NA)))
```
MichaelChirico: This is certainly awkwardly named. The naming is done at the R level, so it would be very easy for me to change this to `V1_lead_1`. Any thoughts?

MichaelChirico:
Codecov doesn't appear to have re-triggered with the second commit; can do manually?

inst/tests/tests.Rraw
jangorecki: better than `lag`/`lead` in this case to use `shift_1` or `shift_-1`

src/shift.c
jangorecki: which "these" precisely? what is it is accomplishing that is already done in INTSXP?

man/shift.Rd
```diff
@@ -14,9 +14,9 @@ shift(x, n=1L, fill=NA, type=c("lag", "lead"), give.names=FALSE)
 }
 \arguments{
   \item{x}{ A vector, list, data.frame or data.table. }
-  \item{n}{ Non-negative integer vector denoting the offset to lead or lag the input by. To create multiple lead/lag vectors, provide multiple values to \code{n}. }
-  \item{fill}{ Value to pad by. }
-  \item{type}{ default is \code{"lag"}. The other possible value is \code{"lead"}. }
+  \item{n}{ integer vector denoting the offset by which to lead or lag the input. To create multiple lead/lag vectors, provide multiple values to \code{n}; negative values of \code{n} will "flip" the value of \code{type}, i.e., \code{n=-1} and \code{type='lead'} is the same as \code{n=1} and \code{type='lag'}. }
+  \item{fill}{ Value to use for padding when the window goes beyond the input length. }
+  \item{type}{ default is \code{"lag"} (look "backwards"). The other possible value is \code{"lead"} (look "forwards"). }
```
jangorecki: I would more strongly suggest to use negative `n` instead of `lead`.

MichaelChirico:
@jangorecki oh my! i didn't even bother to check TBH, my bad.

I knew this would have some overlap with your work on rolling operations in #2961, I should have read around a bit more carefully.

Agree there's room for more improvement w parallelism but I saw that the switch to allowing +/- `n` in the current code base would be an easy switch and went for it 🔪

PR #3168, opened 2018-11-30
 Bunch of coverage tests

### LINKED ISSUE #3167: frankv + ties.method = 'random' + na.last = NA gives a warning
```
frankv(list(c(1, 1, 2, 5, 4, 3, 4, NA, 6)), ties.method = 'random', na.last = NA)
```

> Warning message:
> In set(x, i = if (is.na(na.last)) which_(nas, FALSE) else NULL,  :
>   Supplied 9 items to be assigned to 8 items of column '..stats_runif..' (1 unused)

If I understand the documentation correctly, this is wrong behavior.

This came up in the midst of a coverage improvement PR I'm working on, so this will be finished soon.

### LINKED ISSUE #3169: Schrodinger's cols in setorderv (internally several !length(cols) behaviors)
[These two](https://github.com/Rdatatable/data.table/blob/master/R/setkey.R#L318-L323) lines suggest different behavior for what to do with

```
DT = data.table(a = 1:3)
setorderv(DT, character(0L))
```

Obviously the first branch is hit first and simply `return`s the input. But the other branch suggests the behavior should be more like `setkey` and use _all_ columns. 

The first branch's behavior is consistent with `setorder` where `setorder(DT)` would just return `DT` unaltered.

Perhaps what's intended is for the latter branch to handle `missing(cols)` and the former branch to handle `!length(cols)`?

MichaelChirico:
Closes #3167 
Closes #3169

Ran out of steam to do more tests for now.

I'm not sure if we have any way of doing the `!cedta()` tests so I just `nocov`'d them.


R/setkey.R
```diff
@@ -219,7 +219,7 @@ forder <- function(x, ..., na.last=TRUE, decreasing=FALSE)
   if (ncol(x) == 0L) stop("Attempting to order a 0-column data.table.")
   if (is.na(decreasing) || !is.logical(decreasing)) stop("'decreasing' must be logical TRUE or FALSE")
   cols = substitute(list(...))[-1L]
-  if (identical(as.character(cols),"NULL") || !length(cols)) return(NULL) # to provide the same output as base::order
+  if (identical(as.character(cols),"NULL") || !length(cols) || (length(cols) == 1L && !nzchar(cols))) return(NULL) # to provide the same output as base::order
```
MichaelChirico: this covers `forder(DT, )` case, which errored otherwise later on and with a weird message

R/setkey.R
```diff
@@ -310,6 +310,7 @@ setorder <- function(x, ..., na.last=FALSE)
 
 setorderv <- function(x, cols, order=1L, na.last=FALSE)
 {
+  if (missing(cols)) cols = colnames(x)
```
MichaelChirico: As discussed in #3169; not setting as default argument for now as that may be a breaking change.

MichaelChirico:
@jangorecki `1960.92` gave this output locally:

```
data.table(grouping = c(NA_integer_, NA_integer_), V1 = c(5, 5))
```

but all the CI seems to give the output 

```
data.table(grouping = c(NA_integer_, 0L), V1 = c(5, 5))
```

Can you confirm which is right/why the difference?

R/setkey.R
jangorecki: what breaking change it would be precisely?

jangorecki:
I get locally
```
   grouping V1
1:       NA  5
2:        0  5
```

MichaelChirico:
Doesn't close #3172... I don't know what's going on here, some where primitive-base-S4 thing going on I can't seem to fix:

```
x = as.IDate('2018-08-01')
# should be triggering the inherits(e2, 'difftime') branch of -.IDate but..
x - difftime(x, x - 1)
```

MichaelChirico:
@jangorecki yea it took me a while to cook this one up... `R --vanilla` I get same output.

Rstudio `sessionInfo()`:

<details> 

```
R version 3.5.1 (2018-07-02)
Running under: macOS High Sierra 10.13.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib

locale:

attached base packages:

other attached packages:
[1] data.table_1.11.9

[1] compiler_3.5.1 tools_3.5.1    yaml_2.2.0  
```

</details>

`R --vanilla` `sessionInfo()`:

<details>

```
R version 3.5.1 (2018-07-02)
Running under: macOS High Sierra 10.13.6

Matrix products: default
BLAS: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRblas.0.dylib

locale:

attached base packages:

other attached packages:
[1] data.table_1.11.9

[1] compiler_3.5.1 tools_3.5.1  
```

</details>

R/setkey.R
MichaelChirico: I dunno. I guess _adding_ a default argument is different from _removing_ a default argument. Will fix

MichaelChirico:
> Any idea what was the profile config that was messing up output from NA, 0 to NA, NA?

not a clue...

mattdowle:
Looks great.  Nearly +1% coverage!
I moved the `:::` to the right place.
3 test are failing: 1962.86, 1962.92 and 1962.108.  I had a quick look but didn't immediately see what was up.

MichaelChirico:
Fixed the failing tests. One was giving me whack-a-mole problems so I just dropped it. If there's still some coverage missed we'll get back to it eventually.

MichaelChirico:
@mattdowle not sure how `patch` could possibly be showing what it is on the diff target given the number of tests added... guess i still don't understand the inner workings of codecov well enough

MichaelChirico:
@mattdowle I see... a couple of those were branches that I tried to take a swing at and couldn't hit for the life of me...

So I gave up on adding tests there for now (resisted the urge to `nocov`) them in fact...

mattdowle:
@MichaelChirico Ok I see. Agree. Good approach. Will merge.

inst/tests/tests.Rraw
```diff
@@ -12483,7 +12489,266 @@ gs = groupingsets(d, j = sum(val), by = c("a", "b", "c"),
...
+     error = 'not used (yet)')
+test(1962.08, unique(DT, fromLast = TRUE),
+     data.table(x = c(1, 2, 3), key = 'x'))
+
+test(1962.09, anyDuplicated(DT, by = NULL), 2L)
+
+## uniqlist.R
+test(1962.10, uniqlist(1:5),
+     error = 'not type list')
+test(1962.11, uniqlist(list()), list(0L))
```
jangorecki, line 12520: @MichaelChirico This is an example where blind aiming for higher code cov can cause an issue. AFAIU `uniqlist` was used more widely before, as of now it is being used in very few places. None of those, AFAICT, requires this case to be handled in the way as it is defined by 1962.11 test. So now in #4372 I am changing this behaviour and because of this test (that has been added in a batch, for code cov, and not really because it was required), my PR might look like a breaking change.
I am just elaborating, so that we can be aware it might be better not to increase codecov like this, because we adapt use cases (tests) to the code, and it should be the opposite, code should be adapted to a uses cases (tests).

MichaelChirico: i think it's fine anyway. the change is in fact breaking right? just that we don't care if internal behavior changes. easy enough to note on the new PR that the altered tests are of internal behavior.

Issue #3165, opened 2018-11-30
R-devel change reveals DATAPTR inside parallel regions


mattdowle:
Luke Tierney contacted me about a memory issue in R-devel.  Tests are passing both on CRAN and Travis/Appveyor, but memory usage is higher in some cases.  As before the garbage collector turns off.
Not sure when exactly the R-devel change was made, but some time in the last month or so.

Luke wrote : 
```
I distilled the issue in 'constellation' down to the attached file
from the examples. If I run this the memory usage is much bigger than
previously and the gc() output is garbled. It's a
multi-threading issue again; everything looks fine with
OMP_NUM_THREADS=1. With mutlipe threads you are
calling DATAPTR from threads other than the main one and that creates
a race on setting the R_GCEnabled flag, so eventually it is getting
stuck on off. I instrumented the places where the GC is disabled and
tracked this as the first one from a thread other than the main one:

#4  0x00007ffff78ba9d5 in DATAPTR (x=x@entry=0x1167458)
     at ../../../R/src/include/Rinlinedfuns.h:106
#5  0x00007fffea02c4c8 in subsetVectorRaw (target=0x4529590, source=0x1167458,
     idx=0x4175210, any0orNA=FALSE) at subset.c:44
#6  0x00007fffea02c7a6 in subsetDT (x=<optimized out>, rows=<optimized out>,
     cols=<optimized out>) at subset.c:272
```

Given Luke's detail, it's easy to see the problem in the code without needing to reproduce. All R API usage needs taking outside all parallel regions. I delayed doing that last time due to time pressure, and now it needs tackling.

There's a long shot that this is why the 2nd run is slower on db-bench.

This warrants accelerating release of 1.12.0, not least because it impacts Luke.

MichaelChirico:
awesome of Luke to do all the hard work! 💯

mattdowle:
I'm confirming with Luke that it's fixed before closing ..

jangorecki:
@mattdowle yes I compile R-devel daily, will add this flag there. Added to not forget in #3147

PR #3170, opened 2018-11-30
Reworked subset.c 

mattdowle:
`subset.c` complete for #3165.  Other files still to review.
Removed all R API use in parallel regions.
Removing the catch-and-expand ALTREP. That's left to R which will do that inside INTEGER(), REAL() etc when they see ALTREP vecs (which is now ok as those R API calls are now outside parallel regions).
Parallel within column rather than across columns:  A very big subset of 1 column should benefit from 8 cores now, whereas before that was single threaded.
No `critical` or anything complicated. Simpler.
Tighter branch-free loops by dealing with zeros and negatives up-front.
Also makes progress in being able to stop defining `USE_RINTERNALS` : should not be needed for this file now either to work or for performance.

### DIFF
--- a/src/subset.c
+++ b/src/subset.c
@@ -81,7 +81,7 @@ static void subsetVectorRaw(SEXP ans, SEXP source, SEXP idx, const bool anyNA)
     PARLOOP(0)
   } break;
   default :
-    error("Internal error: column type '%s' not supported by data.table subset. All known types are supported so please report as bug.", type2char(TYPEOF(source)));
+    error("Internal error: column type '%s' not supported by data.table subset. All known types are supported so please report as bug.", type2char(TYPEOF(source)));  // # nocov
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/subset.c
```diff
@@ -1,283 +1,265 @@
 #include "data.table.h"
 
-static SEXP subsetVectorRaw(SEXP target, SEXP source, SEXP idx, Rboolean any0orNA)
+static void subsetVectorRaw(SEXP ans, SEXP source, SEXP idx, const bool anyNA)
 // Only for use by subsetDT() or subsetVector() below, hence static
 {
-  if (!length(target)) return target;
+  const int n = length(idx);
+  if (length(ans)!=n) error("Internal error: subsetVectorRaw length(ans)==%d n=%d", length(ans), n);
```
jangorecki: could be `nocov` here

PR #3191, opened 2018-12-06
fixed IDateTime.R use of base::.Date

mattdowle:
Follow up to PR #3168
CI Pipeline revealed can't use `base::.Date`
Changed back to how it was before #3168 (calling `base::'-.Date'` when `e2` is `reallyReal`) but without `as.Date(e2)` so that the bug fix in #3168 and its new tests still work.
Since CI Pipeline is in error status now and I have all day before Michael and Jan wake up, and pipeline takes over an hour to run,  merging now to see if it passes and make progress.  @MichaelChirico please check post-merge and revert if necessary. If so a new test will need to be added since no tests changed in this follow up.


jangorecki:
PR solved CI fail

PR #3202, opened 2018-12-11
Parallel gsum and gmean

mattdowle:
First step in parallelizing gforce (gsumm.c).
`gsum` done. Since `gmean` calls `gsum`, mean is done too other than the extra steps for `na.rm=TRUE`.
No changes to tests at all,  they all pass.
Code needs tidying, will come back to that.
Merging early to see how it does on server (db-bench) as I've only tested on laptop so far.


jangorecki:
https://h2oai.github.io/db-benchmark/
well done

PR #3221, opened 2018-12-14
foverlaps now also handle nomatch=NULL same as [.data.table

### LINKED ISSUE #3096: foverlaps nomatch argument
`foverlaps` manual entry for `nomatch` argument states the following.

> Same as `nomatch` in `match`. When a row (with interval say, [a,b]) in x has no match in y, nomatch=NA (default) means NA is returned for y's non-by.y columns for that row of x. nomatch=0 means no rows will be returned for that row of x. The default value (used when nomatch is not supplied) can be changed from NA to 0 using options(datatable.nomatch=0).

IMO the first sentence is misleading, because it is not that much similar to `match` function when it comes to `nomatch=0`. Ideally will be to make this argument consistent to changes requested in #857.
Note that `match(1:2, 2:3, nomatch=NULL)` is the same as `match(1:2, 2:3, nomatch=NA)`, so we don't want that.

jangorecki:
Closes #3096

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 
 1. `fread()` can now read a remote compressed file in one step; `fread("https://domain.org/file.csv.bz2")`. The `file=` argument now supports `.gz` and `.bz2` too; i.e. `fread(file="file.csv.gz")` works now where only `fread("file.csv.gz")` worked in 1.11.8.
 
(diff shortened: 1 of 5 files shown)


mattdowle:
Coverage fail looks real:  new branch needs a test please

jangorecki:
@mattdowle done

Issue #3211, opened 2018-12-13
Release checks


mattdowle:
Hi @renkun-ken.
Could you run latest data.table dev through your systems please.
There are some significant internal code changes that it would be good to run past your checks early.
Please check that more cores are being used more of the time, and whether your test suite total time is any faster or not.
Thanks!!
Matt

renkun-ken:
I notice that significant changes are made recently so I keep up with those on a daily basis. So far the system runs perfectly with 4c9a1f0.

I'll test against the latest commit very soon.

mattdowle:
Thanks @renkun-ken!  
We could maybe have a look at fork again. See past problems here : 
https://github.com/Rdatatable/data.table/blob/master/src/openmp-utils.c#L73
Are you on Linux/Mac/WIndows?  ( I didn't think Windows had fork, but somebody told me recently that it does.)

renkun-ken:
I'm on Linux.

I test a small part of system against 0daefad2dcba8dd15f907969eab60a58b7f28ef1 and the computing time reduced 30%. I'll do a more thorough test soon.

I filed #2885 on auto-restore multi-threading after fork. fst adopts this behavior for quite a while and it works for me perfectly so far.

renkun-ken:
Latest commit 0daefad2dcba8dd15f907969eab60a58b7f28ef1 is broken here. I'll take a closer look and file an issue about it soon.

renkun-ken:
The problem lies on a assumed behavior that `dt[TRUE]` always creates a shallow copy of `dt` so that adding columns to the shallow copy does not influence the original data.table. See #3214.

mattdowle:
Thanks @renkun-ken.  Now fixed in #3213 and merged.

mattdowle:
@renkun-ken Many thanks for this invaluable testing!!
#3216 fixed, and #2885 (restore multi threading after fork) done too.
Please rerun whenever convenient.

renkun-ken:
A full run with 76d4a622e0ee3c5ed8e3ecf693026bf0ed7f4fe3 is in progress...

renkun-ken:
Just found another issue which seems to be breaking change. The updated subsetting seems to under-allocate. See #3228.

renkun-ken:
Except empty `.SDcols` now produces an error (which I think is by design in recent changes?), as #3229 has just been merged, no other issues are found so far.

mattdowle:
Thanks again for the quick turnaround!

[...]

Issue #3185, opened 2018-12-05
.SDcols gives a strange message if length() 0


MichaelChirico:
Also identified in a comment by @HughParsonage here:

https://github.com/Rdatatable/data.table/issues/1789#issuecomment-322655733

For now I'll solve this by just short-circuiting to `null.data.table()` if `!length(.SDcols)`... the broader issue raised in #1789 of handling length-`0` `.SDcols` in the case of `:=` will require a bit more work.

PR #3238, opened 2018-12-20
Empty .SDcols back to 1.11.8 behaviour, plus fix for integer()

### LINKED ISSUE #1789: zero length .SDcols
It would be nice if .SDcols could be zero length.   In #1786 we discussed this code which works but a special case had to be made for the situation where there were no numeric columns.  

```
iris.dt <- as.data.table(iris)
nums <- which(sapply(iris.dt, is.numeric))
if (length(nums)) iris.dt[, (nums) := .SD/rowMeans(.SD), .SDcols = nums]
```

It would be nice if .SDcols could allow the specification of no columns so that this would work even in the case that iris has no numeric columns.  

```
iris.dt <- as.data.table(iris)
nums <- which(sapply(iris.dt, is.numeric))
iris.dt[, (nums) := .SD/rowMeans(.SD), .SDcols = nums]
```

(Incidentally it would also be nice if nums could be logical so that we could have `nums <- sapply(iris.dt, is.numeric)` .  That works in .SDcols= but not on the LHS of := .)

mattdowle:
Closes #1789
Follow up to #3185 
Fixes regression in dev before release reported by Kun Ren in comments in #3211

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -72,7 +72,7 @@
 
 11. `as.data.table.default` method will now always copy its input, closes [#3230](https://github.com/Rdatatable/data.table/issues/3230). Thanks to @NikdAK for reporting.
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
always prefer to have more robustness!! thanks @kunren!!

PR #3248, opened 2018-12-30
froll manual improvements

### LINKED ISSUE #3246: [Documentation] Minor comments on roll() help text
While trying to learn about the new `frollmean` function (dev built 2018-12-15), I noted some possible edits of the help text which you may consider.

Thanks a lot for your great work with implementing rolling functions in `data.table`!

Happy new `data.table` year all of you!

-----

**Arguments**
>`x`: [...] data.frame or data.table of numeric fields

In neither `?data.frame` nor `?data.table` the term 'field' is found. Perhaps use 'columns'?

>`fill`	numeric, value to pad by, default NA.

Suggested edit:
numeric, value to pad by. Defaults to NA.

>`algo`: When set to `"exact"` then slower algorithm

Suggested edit:
When set to `"exact"`, then slower algorithm

>`algo` [...] **It** suffers [...] perform [...] handle
 
Shouldn't it be 'performs' and 'handles'?

>`align`: character, define if window frame covers preceding rows `"right"`, following rows `"left"` or centered `"center"`, default `"right"`.

Suggested edit:
character, define if window covers preceding rows (`"right"`), following rows (`"left"`) or is centered (`"center"`). Defaults to `"right"`.

>`na.rm`: logical, should missing values be removed when calculating window, default `FALSE`. For details on handling other non finite values see details below.

Suggested edit:
logical. Should missing values be removed when calculating window? Defaults to `FALSE`. For details on handling other non-finite values, see details below.

>`hasNA` logical, if it is known that `x` contains `NA` then setting to `TRUE` will speed up, default `NA`.

Suggested edit:
logical. If it is known that `x` contains `NA`, then `hasNA = TRUE` will speed up. Defaults to `NA`.

**Details**
>They always returns

Suggested edit:
They always return

>This is so that it can be used conveniently within data.table's syntax

...reads a bit awkward to me, and it's not entirely clear what 'this' and 'it' refer to here. Maybe either be more explicit:
The accepted input and return objects ensure a convenient use of `roll` functions within `data.table` syntax.

...or simply: 
Thus, `roll` functions can be used conveniently within `data.table` syntax.

>Argument `n` allows multiple values to calculate multiple rolling windows or if `adaptive=TRUE` then it expects a list, each list element must be integer vector of window size corresponding to every `column[row]` from `x`.

Suggested edit:
Argument `n` allows multiple values to apply rolling functions [or: perform rolling calculations] on multiple window size. If `adaptive=TRUE`, then it expects a list. Each list element must be integer vector of window size corresponding to every `column[row]` from `x`.

The "window size corresponding to every `column[row]`" is unclear to me.

>For precise handling of non-finite values use `algo="exact"`. Argument `algo="exact"` will make rolling functions to perform extra computation for floating point rounding error correction. 

Suggested edit:
For precise handling of non-finite values use `"algo="exact"`, which will perform extra computation for floating point rounding error correction.

>This is useful mostly when when input data has distant outlier.

Suggested edit (given previous suggestion)
`algo="exact"` is useful mostly when input data have distant outlier. 

>Adaptive rolling functions are special cases where for each single observation has own corresponding rolling window width

This sentence needs some tuning. The "where for each single observation has own corresponding rolling window width" is unclear.

>Due to the logic of that function following

Here "that function" refers to "Adaptive rolling function**s**" in the previous sentence. Check so the numerus match. Add comma after function (function, following)

 >if list of integer vectors is passed to `x` then all list vectors

Suggested edit:
if list of integer vectors is passed to `x`, then all list vectors

>When multiple columns or multiple windows width are provided then they are

Suggested edit:
When multiple columns or multiple windows width are provided, then they are

>**Note** Users coming from most popular package for rolling functions zoo might expect following difference in data.table implementation.

Suggested edit:
Below is a summary of the main differences between rolling functions in `data.table` and corresponding functions in the `zoo` package:

>rolling function will always return same length of results as provided input.

Suggested edit:
rolling function will always return result of the same length as the input. 

>`fill` by default `NA`.

Suggested edit:
`fill` defaults to `NA`.

>`fill` accept only constant values, no support for _na.locf_ or other functions.

Suggested edit:
`fill` accepts only constant values. It does not support `na.locf` or other functions.

>`align` is by default `"right"`.

Suggested edit:
`align` defaults to `"right"`.

>`na.rm` is respected, no need to use other function when having `NA` values.

Suggested edit:
`na.rm` is respected, and other functions are not needed when input contains `NA`.

>when `adaptive=FALSE` (default) then `n` must be a numeric vector, list is not accepted.

Suggested edit:
when `adaptive=FALSE` (default), then `n` must be a numeric vector. List is not accepted.

>when `adaptive=TRUE` then `n` must be vector 

Suggested edit:
when `adaptive=TRUE`, then `n` must be vector 

>there is no `partial` window support.

Suggested edit:
`partial` window is not supported.

**Examples**

>three above are embarrassingly parallel using openmp

Feels like something is missing in this comment.

jangorecki:
Closes #3246

### DIFF
--- a/man/froll.Rd
+++ b/man/froll.Rd
@@ -101,7 +101,7 @@ frollmean(x, n, fill=NA, algo=c("fast", "exact"), align=c("right",
   
   Be aware that rolling functions operates on the physical order of input.
   If the intent is to roll values in a vector by a logical window, for
-  example an hour, or a day, one has to ensure that there are not gaps in
+  example an hour, or a day, one has to ensure that there are no gaps in
   input. For details see \href{issue Rdatatable/data.table#3241}{https://github.com/Rdatatable/data.table/issues/3241}.
 }
 \examples{
@@ -116,7 +116,7 @@ frollmean(d[, .(V1)], c(3, 4))
 frollmean(d, c(3, 4))
 ## three calls above will use multiple cores when available
 
-# partial window of size 3
+# partial window using adaptive rolling function
 an = function(n, len) c(seq.int(n), rep(n, len-n))
 n = an(3, nrow(d))
 frollmean(d, n, adaptive=TRUE)
--- a/man/froll.Rd
+++ b/man/froll.Rd
@@ -20,7 +20,7 @@ frollmean(x, n, fill=NA, algo=c("fast", "exact"), align=c("right",
   \item{x}{ vector, list, data.frame or data.table of numeric columns. }
   \item{n}{ integer vector, for adaptive rolling function also list of
     integer vectors, rolling window size. }
-  \item{fill}{ numeric, value to pad by. Default to \code{NA}. }
+  \item{fill}{ numeric, value to pad by. Defaults to \code{NA}. }
   \item{algo}{ character, default \code{"fast"}. When set to \code{"exact"},
     then slower algorithm is used. It suffers less from floating point
     rounding error, performs extra pass to adjust rounding error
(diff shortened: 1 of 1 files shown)


Henrik-P:
Hi again!

Just a tiny comment. In some places you write:
>Default to

Shouldn't it be "Default**s**, because I assume the verb refers to "The argument value" / "It" (singular)?

jangorecki:
of course it should, thanks for spotting that

Issue #2016, opened 2017-02-05
Support for writing to gzip (.gz) with fwrite


MichaelChirico:
`write.csv` can handle the following:

```{r}
library(data.table)
DT = data.table(a = 1:10, b = 11:20)

tmp.gz = tempfile(fileext = '.csv.gz')
gzf = gzfile(tmp.gz, 'w')
write.csv(DT, gzf)
close(gzf)
```

But `fwrite` cannot:

```
tmp.gz = tempfile(fileext = '.csv.gz')
gzf = gzfile(tmp.gz, 'w')
fwrite(DT, gzf)
close(gzf)
```

> Error: `is.character(file) && length(file) == 1 && !is.na(file)` is not `TRUE`

(because the `gzf` object is more complicated than just being a file path).

`fwrite(DT, tmp.gz)`, as expected, does not perform compression -- the output is the same size as `fwrite(DT, tempfile())` (which is different from `write.csv`).

I'm using `fwrite` to pipe some output to [Vowpal Wabbit](https://github.com/JohnLangford/vowpal_wabbit/wiki/Command-line-arguments), trying to use the `--compressed` argument to save on some disk space. Workaround is to `fwrite` then `gzip`, but this seems sub-optimal.

JhossePaul:
+1

rargelaguet:
+1

msgoussi:
Dear Mr.
When I export data.table using fwrite, the size of the file is 31.3GB 
then when I use gzip in order to compress the file, I get: 
Error in writeBin(bfr, con = out, size = 1L) : 
  'Calloc' could not allocate memory (10000000 of 1 bytes)

How can I solve this problem?
Thanks

ncravino:
+1

Griefberg:
+1

wookietreiber:
Please don't post comments with simply "+1". This adds absolutely no value. Either **write something of value** that adds something to the discussion **or use the thumbs up emoticon** / reaction on the original post to express interest in the issue and a solution to it.

I'm in favor of deleting such comments, including mine BTW, because it is of no consequence to the technical discussion of this issue as writing "+1". Please, people, please ;)

PR #3278, opened 2019-01-12
Add gzip support to fwrite

philippechataignon:
This is a first attempt to implement gzipped csv output in fwrite (issue #2016).

It uses zlib library and replaces open/write/close used for csv file by gzopen/gzwrite/gzclose for gzipped csv. A second buffer, which size is around 10% bigger than main buffer, is allocated for each thread. zlib is thread-safe and the gzip compression use all the available threads.
  
Option `compress="gzip"` is added to fwrite and is automatically set when file ends with `.gz`. Default is `compress = "none"`.

On my system (Debian Linux), r-base-dev install zlib1g-dev, which is needed to compile. I don't need to add a -lz for gcc to compile but that may be not true for others systems and on Windows or Mac. I guess that file `cc.R` must be modified to indicate the zlib dependence but it's too hard for me.

I've added 2 tests but it seems to be difficult to test a binary output like a gzipped csv. Test uses command `zcat` and runs only on unix platforms.

Please feel free to test.

### DIFF
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -7,7 +7,7 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
            dateTimeAs = c("ISO","squash","epoch","write.csv"),
            buffMB=8, nThread=getDTthreads(verbose),
            showProgress=getOption("datatable.showProgress", interactive()),
-           compress = c("none", "gzip"), 
+           compress = c("default", "none", "gzip"), 
            verbose=getOption("datatable.verbose", FALSE)
            ) {
   isLOGICAL = function(x) isTRUE(x) || identical(FALSE, x)  # it seems there is no isFALSE in R?
@@ -41,7 +41,7 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
     dec != sep,  # sep2!=dec and sep2!=sep checked at C level when we know if list columns are present
     is.character(eol) && length(eol)==1L,
     length(qmethod) == 1L && qmethod %chin% c("double", "escape"),
-    length(compress) == 1L && compress %chin% c("none", "gzip"),
+    length(compress) == 1L && compress %chin% c("default", "none", "gzip"),
     isLOGICAL(col.names), isLOGICAL(append), isLOGICAL(row.names),
     isLOGICAL(verbose), isLOGICAL(showProgress), isLOGICAL(logical01),
     length(na) == 1L, #1725, handles NULL or character(0) input
@@ -50,7 +50,7 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
     length(nThread)==1L && !is.na(nThread) && nThread>=1L
     )
   
-  is_gzip <- compress == "gzip" || grepl("\\.gz$", file)
+  is_gzip <- compress == "gzip" || (compress == "default" && grepl("\\.gz$", file))
   
   file <- path.expand(file)  # "~/foo/bar"
   if (append && missing(col.names) && (file=="" || file.exists(file)))
@@ -81,4 +81,3 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
           showProgress, is_gzip, verbose)
   invisible()
 }
-
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -7,10 +7,13 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
            dateTimeAs = c("ISO","squash","epoch","write.csv"),
            buffMB=8, nThread=getDTthreads(verbose),
            showProgress=getOption("datatable.showProgress", interactive()),
-           verbose=getOption("datatable.verbose", FALSE)) {
+           compress = c("none", "gzip"), 
+           verbose=getOption("datatable.verbose", FALSE)
+           ) {
   isLOGICAL = function(x) isTRUE(x) || identical(FALSE, x)  # it seems there is no isFALSE in R?
   na = as.character(na[1L]) # fix for #1725
   if (missing(qmethod)) qmethod = qmethod[1L]
+  if (missing(compress)) compress = compress[1L]
   if (missing(dateTimeAs)) { dateTimeAs = dateTimeAs[1L] }
   else if (length(dateTimeAs)>1L) stop("dateTimeAs must be a single string")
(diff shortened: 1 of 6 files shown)


R/fwrite.R
```diff
@@ -38,13 +41,17 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
...
+    length(compress) == 1L && compress %chin% c("none", "gzip"),
     isLOGICAL(col.names), isLOGICAL(append), isLOGICAL(row.names),
     isLOGICAL(verbose), isLOGICAL(showProgress), isLOGICAL(logical01),
     length(na) == 1L, #1725, handles NULL or character(0) input
     is.character(file) && length(file)==1L && !is.na(file),
     length(buffMB)==1L && !is.na(buffMB) && 1L<=buffMB && buffMB<=1024,
     length(nThread)==1L && !is.na(nThread) && nThread>=1L
     )
+  
+  is_gzip <- compress == "gzip" || grepl("\\.gz$", file)
```
jangorecki: Nice to recognise filename but we could warn if user explicitly set compress to none and uses gz filename

inst/tests/tests.Rraw
```diff
@@ -9266,13 +9266,17 @@ test(1658.8, fwrite(data.table(a=c(1:5), b=c(1:5)), quote=TRUE),
 test(1658.9, fwrite(data.table(a=c(1:3), b=c(1:3)), quote=TRUE),
             output='"a","b"\n1,1\n2,2\n3,3')
 
-# block size one bigger than number of rows
-test(1658.11, fwrite(data.table(a=c(1:3), b=c(1:3)), quote=TRUE),
-            output='"a","b"\n1,1\n2,2\n3,3')
+# fwrite output to console ignore compress
+test(1658.11, fwrite(data.table(a=c(1:3), b=c(1:3)), compress="gzip"),
```
jangorecki: 1. Why the old unit test was removed?
2. We could warn in such case

inst/tests/tests.Rraw
```diff
@@ -9266,13 +9266,17 @@ test(1658.8, fwrite(data.table(a=c(1:5), b=c(1:5)), quote=TRUE),
...
+            output='a,b\n1,1\n2,2\n3,3')
 
-# block size one less than number of rows
-test(1658.12, fwrite(data.table(a=c(1:3), b=c(1:3)), quote=TRUE),
-            output='"a","b"\n1,1\n2,2\n3,3')
+# fwrite gzipped output
```
jangorecki: Good test, but again, the old one got removed

man/fwrite.Rd
```diff
@@ -52,6 +53,7 @@ fwrite(x, file = "", append = FALSE, quote = "auto",
   \item{buffMB}{The buffer size (MB) per thread in the range 1 to 1024, default 8MB. Experiment to see what works best for your data on your hardware.}
   \item{nThread}{The number of threads to use. Experiment to see what works best for your data on your hardware.}
   \item{showProgress}{ Display a progress meter on the console? Ignored when \code{file==""}. }
+  \item{compress}{If compress = \code{"gzip"} or if \code{file} ends in \code{.gz}, even if compress = \code{"none"}, then the output format is gzipped csv. Output to console is never gzipped even if compress = \code{"gzip"}. By default, compress = \code{"none"}.}
```
jangorecki: Well documented behaviour. I would put "compress" into "code" tag as it refers to argument variable, together with "=" sign.

philippechataignon: OK. I reformat documentation like this and describe new `default` option. See https://github.com/Rdatatable/data.table/pull/3278/commits/b3c2ae9fe8d0e8fd5d382bc44f580630e1b3cbb3

R/fwrite.R
jangorecki: Might be better "auto" instead of default. Anyway new argument is not required as we might do:
```r
is_gzip <- compress == "gzip" || (missing(compress) && grepl("\\.gz$", file))
```

inst/tests/tests.Rraw
jangorecki: If so, that should be mentioned somewhere, at least in PR comment. Otherwise it looks like PR can introduce breaking changes.
Giving new numbers might be still better way to go, as removed unit test numbers might be referenced somewhere else in the code or in github issues, so if you re-use those, it might result in confusion which will lead to `git blame` and potentially waste time tracking that. I would keep old tests and create new ones. As you are adding tests for new feature it is perfectly justified to not just add `1658.max+1` be append new tests at the end of tests.Rraw unit tests using new number.

inst/tests/tests.Rraw
```diff
@@ -9353,6 +9349,27 @@ test(1658.34, fwrite(matrix(1:4, nrow=2, ncol=2), quote = TRUE), output = '"V1",
...
 
+# fwrite output to console ignore compress
+test(1658.37, fwrite(data.table(a=c(1:3), b=c(1:3)), compress="gzip"),
+            output='a,b\n1,1\n2,2\n3,3')
+
+# fwrite force gzipped output
+if (.Platform$OS.type=="unix") {
+  f <- tempfile()
+  fwrite(data.table(a=c(1:3), b=c(1:3)), file=f, compress="gzip")
+  test(1658.38, system(paste("zcat", f), intern=T), output='[1] "a,b" "1,1" "2,2" "3,3"')
```
jangorecki: there is manual check that will make those line (and next test too) fail, related to `T` instead of `TRUE`, see https://github.com/Rdatatable/data.table/blob/7d2acb58f0063211a0ede54e3727b3cce14065e6/CRAN_Release.cmd#L61

philippechataignon:
Thanks for reviews. I will close this PR for now to improve this enhancement. I saw 2 problems :

* actually append is not implemented for gzip : I don't know if it has a sense to append a gzip.
* the approach with gzwrite create a bootleneck : only one thread compress. Gzip is thread would be a better solution but it requires to modify the actual buffer gestion.

st-pasha:
There is a program called `pigz` which implements parallel writing of gzip archives. It works on top of `zlib`: each thread creates compresses its own block of data, and then sends ZLIB_SYNC_FLUSH command, which allows compressed blocks be concatenated.  Same strategy can be used here.

PR #3288, opened 2019-01-16
Fwrite gzip

### LINKED ISSUE #2016: Support for writing to gzip (.gz) with fwrite

### LINKED ISSUE #2612: [request] print informative error if fwrite uses more memory than allowed
This request follows my post on [StackOverflow](https://stackoverflow.com/questions/48591456/make-r-foreach-and-domc-dopar-work-with-single-threaded-data-table-fwrite/48593514#48593514). I've found that fwrite will do strange things in memory constrained environments. If write.table uses more memory than it is allowed, I've found that my high performance computing cluster will kill the job. But if I use fwrite, the job will not be killed--instead, fwrite will either write nothing, write an empty file, write part of the file, or write the file with errors--but the R code will appear to have finished correctly. It's taken me quite some time to figure this all out--it would be nice if there were a way to alert users in the future that the write operation failed because of memory constraints (i.e., so they know they just need to boost the RAM allocation instead of hunting down errors in their code that don't exist). I don't imagine that's an easy task (I know this is a rather vague request), but I thought I would bring this to your attention because I've spent the last few days wrangling with the issue.

### LINKED ISSUE #2974: fwrite crashes for tables with large columns
Bit of an edge case here, but when trying to `fwrite` a table with a very large column R crashes every time.

Potentially related to #2612.

**Minimal reproducible example:**
```
huge_vector <- paste(rep('x', 30e6), collapse='')

foo <- data.table::data.table(x=huge_vector)

tf <- tempfile()

data.table::tables()
# NAME NROW NCOL MB COLS KEY
# 1:  foo    1    1 29    x    
# Total: 29MB

# this works fine
write.table(foo, file=tf)

# this reliably crashes my R session
data.table::fwrite(foo, file=tf)
```

**Output of sessionInfo():**
```
R version 3.4.3 (2017-11-30)

Matrix products: default

locale:
[1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252    LC_MONETARY=English_Australia.1252
[4] LC_NUMERIC=C                       LC_TIME=English_Australia.1252    

attached base packages:

[1] compiler_3.4.3    tools_3.4.3       yaml_2.1.19       data.table_1.11.0
```

For what it's worth, I have 16GB RAM on the system with ~8GB free before running the above example.

philippechataignon:
Closes #2016 
Closes #2974 
Closes #2612 

I reopen a PR to discuss about its contents. Its a complete rewrite of https://github.com/Rdatatable/data.table/pull/3278 : it doesn't use gzopen/gzwrite/gzclose because only one thread compress and that's slow.

In fwrite, output file is always open in binary mode. The idea is to take data from uncompressed buffer in normal case and to compress it and take data from compressed buffer when gzip is active. The code for managing file output is quite the same in normal and gzip mode.

The main difference concern the buffers :
* there are no realloc in this version : if the buffer is too small, fwrite stops and suggests to increase the buffMB parameter. Its too difficult to realloc the compressed and the uncompressed buffer. But, in fact, actual buffer policy doesn't work very good despite realloc. See https://github.com/Rdatatable/data.table/issues/2974 https://github.com/Rdatatable/data.table/issues/2612.
* at the beginning, fwrite compute the max line length. Actually it tries to write in the buffer and add the ending position. But if the buffer is too small, its too late : buffer overflow occurs. Thats what happens in https://github.com/Rdatatable/data.table/issues/2974. In this PR, I use strlen for string/factor and an maximum length integer by type (9 for 32 bits, 19 for int 64 for example). It's not a problem to overestimate this line.

Thanks for any advice on this PR.


mattdowle:
Fantastic PR -- thanks @philippechataignon.

PR #3294, opened 2019-01-19
Remove const to pass Oracle compiler on Solaris

### LINKED ISSUE #3285: Error with Oracle compiler (likely Solaris only)
Email received from Prof Ripley : 

---

As you will see from the CRAN results page, this newly fails to install 
with the Oracle compiler.  It does install with gcc but many of the 
packages depending on segfault in their checks, in function gforce.

(This seems not to be showing currently: the nub is
```
/home/ripley/developerstudio12.6/bin/cc -xc99 
-I"/home/ripley/R/cc/include" -DNDEBUG   -I/usr/local/include  -xopenmp 
-KPIC  -O -xlibmieee -xlibmil -xtarget=native -xcache=generic -nofstore 
-c froll.c -o froll.o
"froll.c", line 114: left operand must be modifiable lvalue: op "="
"froll.c", line 149: left operand must be modifiable lvalue: op "="
```
which are inside OpenMP sections. Using a parallel install shows similar 
issues in frollR, frolladaptive at least.)

mattdowle:
Closes #3285 

I went a bit further than the patch from Prof Ripley which removed the `const`. With the following thought process.
1. The `MIN(getDTthreads(), nx)` can be left to OpenMP.  Either OpenMP will only start the number of threads it needs for very small loops, or the overhead of selecting threads from the pool is so small it's not worth worrying about.  Either way it's not worth the small extra code complexity of MIN() so just leave that to OpenMP.  I don't actually know if this is true, but of all the things we don't know, I'm willing to trust OpenMP on this and move on to something higher value.
2. Since the MIN() was probably why it was held in a variable up front (to split up the lines),  I removed the variable and put the getDTthreads() back inline the #pragma line.   Several reasons.  First,  saving a variable saves accidentally using them (e.g. `nth` vs `threads`).  It's not a hard and fast rule to avoid creating new variables but in this case I think it makes sense.  Second, there are some `grep`'s in CRAN_Release.cmd that look for the pragma lines and check that num_thread() is specified.  Because missing that directive would then be sensitive to the global OpenMP setting shared by other packages.  It's slightly easier for me when running CRAN_Release.cmd to see the consistent pattern num_threads(getDTthreads()) without a variable being there.  Now that we know a `const var` being passed to num_threads() is the problem on Solaris,  the CRAN_Release.cmd greps could be strengthened to ensure no variable passed to num_threads() just because otherwise we'd have to find a way that such variable wasn't declared const.
3. I collapsed the two pragma into one to save lines and indentation and because that's what I've done elsewhere.  So far I've only split up into two pragma when it's necessary for some per-thread initialization.  In these cases where it can be one pragma then I do use one pragma, as a general rule.
4. I removed the schedule(static) because that's the default schedule.  If we explicit that then the thought is that for consistency it should be explicit everywhere (and all defaults?) in all parallel regions. This would distract the eye from non-default code which is more important.
5. I haven't been using the shared() directive, rather leaving that to OpenMP.  I can see whether the variable is shared or private from where it is declared. Needing to look at the shared() directive and what it names is an extra step when looking and understanding the code.  If using shared() is supposed to make the code clearer, then I find myself looking to check it is really shared and whether there are any other shared variables that are not named in it. So I prefer just to leave out the shared() directive.
6. When assigning to variables of type bool,  use `true` not `1`.  To signal to the reader's eye that it's a bool that can only take true/false.  This is the very reason we are ok with assigning naked to a shared bool.  Naked in the sense of no atomic protection.  In this special case, the threads can only set the bool true so it doesn't matter if they overlap.   Even if the bool is actually an int it doesn't matter if a thread race happens because any bits written anywhere in the variable will be true (non-zero).  However it's safe to assume the leading bit would only be written and a naked one-way write is safe.
7. Our usage of bool in this way is not what `volatile` is for, iiuc. So I removed the volatile keyword. See this article:  https://barrgroup.com/Embedded-Systems/How-To/C-Volatile-Keyword.   What we really want for this idiom is a qualifier "naked one-value write" so the compiler would check for us that a single value (true) was only ever written to that variable from inside the parallel region.  But I don't think such a qualifier exists in C.  `volatile` on the other hand, iiuc, is for reading a value that might have been changed by an external process (and therefore has to be read and re-read a lot even when the optimizer might be sure no update could have happened to it).  However, when we write to that bool we don't care what its value is now, we just want that write to get there eventually and stick. The important thing is that no thread could ever write `false`.

That's the way I think about these items currently.


jangorecki:
Regarding comment to 1: after investigation I can confirm that nested parallelism utilize all available cores, so the outer omp for loop does not block access to remaining cores for a inner omp for loop.

One thing I spotted, not related to this branch, is that running the following code on machine having 4 cpus
```r
library(data.table)
n = 5e6
set.seed(108)
dt = setDT(lapply(1:4, function(i) rnorm(n)))
system.time(ans<-frollmean(dt, c(1e3, 2e3), algo="exact"))
```
It spawns 15 threads, each using around 23-26% cpu. Not sure how that compares to running 4 threads using 99-100% each.

mattdowle:
Let's tackle the nested parallelism first.  Why do we have any?  I experimented with it once and managed to not need it. I think we should avoid nested parallelism wherever possible. I can see just one use of `omp_set_nested()` and it's in `frollR.c`.  Each thread gets a column. But then it's nested within each column too. Why does that section need to be nested?

> schedule(static)

The default for `def-sched-var` in all implementations is static I believe.
We can go through all 45 parallel regions and write `schedule(static)` for clarity,  and have a grep in CRAN_Release.cmd to ensure that `schedule(static)` is present.  Or we can say that all implementations do have default static now and if they change in future then they probably have very good reason to and those implementations know better than us. def-sched-var is not affected by any environment variable so it's not like users can change it.  Although, it is implementation defined, so an implementation might provide a way for a user to change it.

> volatile

I understand you're trying to document. But my point was that's not what volatile is for, iiuc. Using it in that context can give the wrong impression to someone who does know what the true use of volatile is for. How about defining our own qualifier : _`onlyWriteTrue_ flag = false;`  or something like that:  `#define _onlyWriteTrue_ bool.`

> shared()

Do you know if when you use shared(), can you still use a shared variable that isn't named in shared()? So the compiler catches that for us?  If so, always using shared() might be a good standard (although extra steps).   Personally,  I think it just gets in the way.  You know that variables are shared unless declared inside the thread's scope -- and you can see that.  I don't want to have to remember to look at the parallel directive as well to see if a variable has been declared shared or private there. In general. So far.

jangorecki:
> Let's tackle the nested parallelism first. Why do we have any?

by default (algo="fast") we use only one level, but for algo="exact" we calculate mean of a window for every single observation, which can be made safely in parallel (at least in theory, as #3346 exposed).

> I experimented with it once and managed to not need it.

I remember, in parallel subset.

> I think we should avoid nested parallelism wherever possible.

Agree. According to [this SO](https://stackoverflow.com/questions/4317551/openmp-what-is-the-benefit-of-nesting-parallelizations) it is not easy to control max threads in nested parallelism. We would need to calculate value to pass to inner `num_threads`.
I will re-work that part - as it was initially - when algo="exact" the outer loop will not goes in parallel, only inner loop will be parallel.
Anyway it might be better after merging this PR to avoid conflicts.

> I can see just one use of `omp_set_nested()` and it's in `frollR.c`. Each thread gets a column. But then it's nested within each column too. Why does that section need to be nested?

Because the nested part is time consuming, and outer loop might have use just few cores. If (number of columns * number of windows) to calculate would be lesser than numbers of threads then algo will not use all existing threads, which is important as that inner loop is slow.

> schedule(static)
> ...
> is not affected by any environment variable so it's not like users can change it.

There are [`OMP_SCHEDULE=sheduling-type` and `omp_set_schedule`](http://jakascorner.com/blog/2016/06/omp-for-scheduling.html) but changing default doesn't make sense in our case. Different loops should have different preference. IIUC `static` best fit when iterations are similar in time complexity, while `dynamic` fits when iterations time complexity varies. That is why in `frollmeanExact` (inner parallel loop) when we iterate on each rows and calculating window mean the `schedule(static)` is used. At the same time in `frollfunR` (outer parallel loop) the `schedule(auto)` is used because one iteration might be calculating window size 30 and another 3000, so their time will be very different. We can generalize that iterations over rows should use `static` while iterations over columns `auto`/`dynamic`.

> volatile

agree.

> shared()

I understand your point. On the other hand this clause nicely documents that there is interaction between threads happening in the loop. Looking at variables it can be spotted of course but it is not that verbose.

mattdowle:
> There are [`OMP_SCHEDULE=sheduling-type` and `omp_set_schedule`](http://jakascorner.com/blog/2016/06/omp-for-scheduling.html)

But those control the schedule when `schudule(runtime)` is used.  They don't control the schedule when the schedule directive is missing.

jangorecki:
I found good reference about omp schedule options: https://www.dartmouth.edu/~rc/classes/intro_openmp/schedule_loops.html

PR #3304, opened 2019-01-22
setkey reorder indices rather than drop

### LINKED ISSUE #1158: setkeyv capable to reorder index
While making `setkeyv` we might want to keep indexes created on the reordered dataset. Instead of recreating each one of then it would be more friendly and I believe (much?) more efficient to manage it in one step.

``` r
library(data.table)
dt <- data.table(a=1:5,b=sample(letters[1:3],5,TRUE))[sample(1:5)]
dt
#    a b
#1: 2 c
#2: 5 c
#3: 1 c
#4: 4 c
#5: 3 a
set2keyv(dt,"b")
key2(dt)
# [1] "b"
setkeyv(dt,"a") # FR: new arg to that call
key2(dt)
# NULL
```

`setkeyv` could accept new argument e.g. `update.index`, then maybe use single `forder` call to apply order both on dataset and on index?

jangorecki:
closes #1158 
has to be well reviewed, especially the part that we are unable(?) to retain identical index to the one that would be freshly build because of within groups order - test 1981.9
It would be interesting to change default to `update.index=TRUE` and see if any revdeps are failing, if there is no bug in the logic in this PR they shouldn't.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 #### NEW FEATURES
 
-1. `setkey` and `setkeyv` gains new argument `update.index`. When `TRUE` provided it will retain existing indices when sorting data.table by reordering them so they are valid for newly keyed data.table. By default argument is set to `FALSE` because of overhead (relatively small but still) related to reordering indices. Closes [#1158](https://github.com/Rdatatable/data.table/issues/1158).
+1. `setkey()` and `setkeyv()` gain `keep.indices=`, [#1158](https://github.com/Rdatatable/data.table/issues/1158). This option retains any existing indices (i.e. secondary keys) by reordering them appropriately given the new row order. This is more efficient and more convenient than creating indices anew after the `setkey()`. This option is experimental, subject to change and currently `FALSE` by default. The default may be changed to `TRUE` in future via a global option such as `datatable.keep.indices`.
 
 #### BUG FIXES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 #### NEW FEATURES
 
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 #### NEW FEATURES
 
+1. `setkey` and `setkeyv` gains new argument `update.index`. When `TRUE` provided it will retain existing indices when sorting data.table by reordering them so they are valid for newly keyed data.table. By default argument is set to `FALSE` because of overhead (relatively small but still) related to reordering indices. Closes [#1158](https://github.com/Rdatatable/data.table/issues/1158).
```
MichaelChirico: I would expand on "relatively small but still" a bit / emphasize/expand on "reordering indices", on first read it sounds like `setindex` will be applied all over again, but the actual implementation is much smarter 👍

jangorecki: @MichaelChirico feel free to push directly. You will probably find some others phrasing improvements. It would be interesting to change default to `update.index=TRUE` and see if any revdeps are failing, if there is no bug in the logic in this PR they shouldn't.

mattdowle:
Very nice. Since the default is FALSE I figured there was no harm in accelerating its release so that folk can experiment with it and test.

mattdowle:
Looks good. Yeah, as you pointed out, we'll have to find a solution to make 1981.9 pass. The reordered index has to be identical to the rebuilt index. I made all other changes leaving just this to resolve.  I changed 1981.9 to fail to remind us what needs to be done before merging this PR.

jangorecki:
This PR is nearly complete, yet the test that is failing has to be resolved, before even trying to merge it to current master. Which may be another challenge after almost 5 years.
As explained in https://github.com/Rdatatable/data.table/issues/1158#issuecomment-1849041104 this FR is no longer relevant to me and personally I prefer to focus my time in other issues. If anyone is willing to take over, then sure, go ahead. Closing for now as it is dangling long enough.

PR #3309, opened 2019-01-23
froll unit tests fails check on failure

### LINKED ISSUE #3308: froll unit tests will not throw exception on failure
froll unit tests runs on every package check, but they will currently not throw exception on failures. Any failed unit test should throw exception so package check will fail also. For that we need to either merge tests/froll.R into inst/tests/tests.Rraw or slightly adjust tests/froll.R. The latter one is preferred because there is going to be more changes to that file until #2778 is resolved. Once it will be finished it will be best to merge.

jangorecki:
Closes #3308
script is no longer `cc` friendly, but plays nice for `Rscript tests/froll.R` and `R CMD check`

### DIFF
--- a/src/frolladaptive.c
+++ b/src/frolladaptive.c
@@ -24,7 +24,7 @@ void fadaptiverollmean(unsigned int algo, double *x, uint_fast64_t nx, double_an
  */
 
 void fadaptiverollmeanFast(double *x, uint_fast64_t nx, double_ans_t *ans, int *k, double fill, bool narm, int hasna, bool verbose) {
-  if (verbose) Rprintf("%s: running for input length %"PRIu64", hasna %d, narm %d\n", __func__, nx, hasna, (int) narm);
+  if (verbose) Rprintf("%s: running for input length %llu, hasna %d, narm %d\n", __func__, nx, hasna, (int) narm);
   bool truehasna = hasna>0;                                     // flag to re-run if NAs detected
   long double w = 0.0;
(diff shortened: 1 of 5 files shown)


jangorecki:
PR detected there are failures on 32bit Windows machine.

jangorecki:
Windows builds on GL CI which we can observed recently might be related to fix in this PR. Running CI for 1.11.8 passed windows builds.
@mattdowle PRIu64 has been changed to llu to avoid new header which might not be available on all platforms. I think we go merge it, and see if windows build issues will be resolved.

src/frolladaptive.c
```diff
@@ -24,7 +24,7 @@ void fadaptiverollmean(unsigned int algo, double *x, uint_fast64_t nx, double_an
  */
 
 void fadaptiverollmeanFast(double *x, uint_fast64_t nx, double_ans_t *ans, int *k, double fill, bool narm, int hasna, bool verbose) {
-  if (verbose) Rprintf("%s: running for input length %lu, hasna %d, narm %d\n", __func__, nx, hasna, (int) narm);
+  if (verbose) Rprintf("%s: running for input length %llu, hasna %d, narm %d\n", __func__, nx, hasna, (int) narm);
```
jangorecki: just for future readers, `%llu` format works fine for `uint_fast64_t` in `Rprintf` but when using C `sprintf` we need to cast `uint_fast64_t` to `long long unsigned`, otherwise compiler warning will be raised.

PR #3310, opened 2019-01-24
Remove USE_RINTERNALS and DATAPTR

mattdowle:
NB: This branch also makes `:=` recycling an error : new feature 1. The memrecycle internals needed attention for this branch so I decided to do it at the same time. All the changes to tests look much better without recycling I think.

Passing R 3.5.2 since DATAPTR is now exposed in R API but still we're not supposed to use it.
Not passing R 3.1.0 until DATAPTR is removed (72 calls) : 
- [ ] assign.c
- [ ] bmerge.c
- [ ] dogroups.c
- [ ] fmelt.c
- [ ] freadR.c
- [ ] fwriteR.c
- [ ] gsumm.c
- [ ] quickselect.c
- [ ] rbindlist.c
- [ ] reorder.c
- [ ] shift.c
- [ ] uniqlist.c

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 #### NEW FEATURES
 
+1. `:=` no longer recycles too-short RHS vectors to match the length on the LHS. Like base R, there was a warning when recycling left a remainder but no warning when the LHS length was an exact multiple of the RHS length. All cases are now an error unless `length(RHS)==1` or `length(RHS)==length(LHS)`. Consistent feedback was that recycling is rarely useful and more often a bug. In rare cases where you need to recycle a length>1 vector, use `rep()` explicitly. Single values are still recycled silently as before.
+
 #### BUG FIXES
 
 #### NOTES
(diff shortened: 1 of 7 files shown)


NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 #### NEW FEATURES
 
+1. `:=` no longer recycles too-short RHS vectors to match the LHS length. There was a warning when recycling left a remainder but no warning when the LHS length was an exact multiple of the RHS length (the same behaviour as base R). All recycling is now an error unless `length(RHS)==1` or `length(RHS)==length(LHS)`. Consistent feedback was that recycling is rarely useful and more often a bug. In rare cases where you need to recycle a length>1 vector, use `rep()` explicitly. Single values are still recycled silently as before.
```
MichaelChirico: might be worth doing a revdep run with this commit, i'm curious how many dependencies assume this behavior

mattdowle: Running. First 80 of 750 complete. Just 1 affected so far: antaresViz

mattdowle: Complete. 16 affected out of 748.  I've updated the top comment.

Issue #2876, opened 2018-05-14
Possible inconsistency for empty j


MichaelChirico:
My fix to #2142 led to the following difference in behavior:

```
DT = data.table(a = c(1, 1, 2))
DT[ , , by = a] # returns full data.table
DT[ , {}, by = a] # returns 0-row data.table

DT[1L, , by = a] # errors
DT[1L, {}, by = a] # returns 0-row data.table
```

Any opinions on what, if anything, should be changed about the above?

mattdowle:
With latest version (1.12.0) we have this inconsistency : 
```
> DT[, , by = a] 
       a
   <num>
1:     1
2:     1
3:     2
Warning message:
In `[.data.table`(DT, , , by = a) :
  i and j are both missing so ignoring the other arguments
> DT[1L, , by = a] 
Error in `[.data.table`(DT, 1L, , by = a) : 
  'by' or 'keyby' is supplied but not j
> 
```

I'd say both should be error.  And that the `by= or keyby= is supplied but not j` error should occur in the first case too.

mattdowle:
After #3322 behavior is now :
```
 > DT[ , , by = a]
Error in `[.data.table`(DT, , , by = a) : 
  by= or keyby= is supplied but not j=
> DT[1L, , by = a] 
Error in `[.data.table`(DT, 1L, , by = a) : 
  by= or keyby= is supplied but not j=
> 
```

PR #3318, opened 2019-01-26
Closes #3313 -- DT[,{}] returns DT, consistent with DT[,]

### LINKED ISSUE #3313: DT[ , {}] should return DT, consistent with DT[ , ]
Follow-up to #2142 and related to #2876... in the original issue I said:

> I guess I would expect this to just return DT since that's what DT[ , ] does

But the fix I filed returns `NULL`, which I don't think makes sense

smingerson:
Closes #3313. I have reviewed [Contributing](https://github.com/Rdatatable/data.table/wiki/Contributing) several times, but this is my first PR. 

I've modified an existing test (1762) that checked `DT[,{}]` was `NULL`, rather than adding a new one. I left the original issue number next to the test and added the new one, I couldn't tell whether that was the convention.

There are some similar questions about `j={}` in #2876, but it is not clear to me what the proper behavior is for all the variations.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 1. `rbindlist()` of a malformed factor missing levels attribute is now a helpful error rather than a cryptic error about `STRING_ELT`, [#3315](https://github.com/Rdatatable/data.table/issues/3315). Thanks to Michael Chirico for reporting.
 
+2. `DT[,{}]` returns `DT` instead of `NULL`, which is consistent with `DT[,]`, [#3313](https://github.com/Rdatatable/data.table/issues/3313). Thanks to @MichaelChirico for reporting and @smingerson for fixing.
+
 #### NOTES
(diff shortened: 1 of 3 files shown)


MichaelChirico:
LGTM, thanks!

mattdowle:
Thanks for the PR!  Great first PR -- nothing wrong with it.  It's just that Jan's https://github.com/Rdatatable/data.table/issues/3313#issuecomment-457886964 and my follow up there https://github.com/Rdatatable/data.table/issues/3313#issuecomment-458359128 agree that current behavior is justified; i.e. `DT[,{}]` returning `NULL` is consistent to other things (e.g. `is.null({})==TRUE` and `DT[,NULL]==NULL`) and doesn't need to be consistent with `DT[,]`.  Whether `j` is supplied or not is the significant difference.

PR #3331, opened 2019-01-29
Add OSX to Travis build matrix

### LINKED ISSUE #3326: macOS / OSX builds in GLCI
We already have linux and windows builds in CI. While linux covers most of all macOS potential issue, there exists some that are macOS specific, as the one reported in https://github.com/Rdatatable/data.table/issues/3300#issuecomment-458032675.
It would be then valuable to have macOS runners.
Whenever there will be some macOS machine we could put our runner into, please @abal5 let us know. It would run only master branch so the overall load made by the runner would be pretty low. C compiles fast, and our tests, while there are many of them, they are quite compact. Linux builds takes from 6min to 15min, depending on the config.

MarcusKlik:
Closes #3326.

In this PR the Travis configuration file has been updated to expand the Travis build matrix to OSX. In order to compile `data.table` on Travis OSX, an  OpenMP enabled compiler is required, so the `llvm` homebrew package is installed before running R CMD build. Also, global variables _LDFLAGS_ and _CFLAGS_ were added (for OSX) as described in the `data.table` [wiki documentation](https://github.com/Rdatatable/data.table/wiki/Installation#step-3).

In addition to adding OSX, also the _previous major release_ and _R-devel_ builds of R were selected in the Travis build matrix. The OSX _R-devel_ build was explicitly excluded from the matrix because Travis is currently not installing _R-devel_ correctly. The total number of Travis builds is now 5 for each commit on the master branch.

As can be seen from the Travis OSX build output [here](https://travis-ci.org/MarcusKlik/data.table/builds/486159786), the OSX build is executed successfully. The total build time on OSX is usually a little longer than for Linux (in this case 14-16 minutes as compared to 10 for Linux)

If the _previous major release_ and _R-devel_ builds are not required, the `r:` field can be removed from _travis.yml_.

Note that code coverage is now calculated on all 5 builds, we could adapt the _after success:_ field to run the coverage only on a linux/release build.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,29 @@ warnings_are_errors: true
 r_check_args: ""
 # empty r_check_args to turn off --as-cran (on by default) as that can be slow
 
-branches:
-  only:
-    - "master"
+# branches:
+#   only:
+#     - "master"
+
+r:
+  - oldrel
+  - release
+  - devel
+
+os:
+  - linux
+  - osx
+
+matrix:
+  exclude:
+    - r: devel  # exclude osx r-devel build as it is currently unstable
+      os: osx
+
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install llvm &&
+    export PATH="/usr/local/opt/llvm/bin:$PATH" &&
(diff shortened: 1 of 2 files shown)


mattdowle:
Fantastic! Thanks, Mark.

PR #3338, opened 2019-01-31
Allow to customize format of objects in 'print.data.table'

mllg:
This PR introduces a S3 generic (`fmt()`, name is up for discussion) which is called on non-atomic items in `print.data.table()`. This allows to customize how more complex objects are formatted.

Here is a simple example illustrating a use case:

```r
obj1 = structure(list(id = "foo"), class = "myclass")
obj2 = structure(list(id = "bar"), class = "myclass")
DT = data.table(row = 1:2, objs = list(obj1,  obj2))

print(DT)

     row      objs
   <int>    <list>
1:     1 <myclass>
2:     2 <myclass>

fmt.myclass <- function(x, ...) paste0("<", class(x)[1L], ":", x$id, ">")
registerS3method("fmt", "myclass", fmt.myclass)

print(DT)

     row          objs
   <int>        <list>
1:     1 <myclass:foo>
2:     2 <myclass:bar>
```

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -172,3 +172,6 @@ S3method(unique, ITime)
 # which.first
 # which.last
 
+# fmt generic to support custom printers
+export(fmt)
+S3method(fmt, default)
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -115,7 +115,7 @@ format.data.table <- function (x, ..., justify="none") {
     else if (is.atomic(x) || inherits(x,"formula")) # FR #2591 - format.data.table issue with columns of class "formula"
       paste(c(format(head(x, 6L), justify=justify, ...), if (length(x) > 6L) "..."), collapse=",")  # fix for #5435 - format has to be added here...
     else
-      paste0("<", class(x)[1L], ">")
+      fmt(x)
   }
   # FR #1091 for pretty printing of character
   # TODO: maybe instead of doing "this is...", we could do "this ... test"?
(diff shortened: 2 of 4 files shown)


jangorecki:
Thanks. Great idea! For future please create issue describing problem to be solved by PR.

MichaelChirico:
actually I like the approach of this PR. format.data.table already uses format.item. we could replace that with a generic and allow end user this flexibility

mllg:
NB: I doubt that we can use `base::format()` as generic. For printing we would need a single string representation, but `format()` has a rather undefined (or lets call it "more complex"?) return value.

MichaelChirico:
Agreed, which is why `format.item` was devised in the first place IIUC.

But something like `col_format` or `item_format` can be our own & customized, very similar to the current PR (if we're going to export I think `fmt` is not a good production name)

mattdowle:
I like the approach of this PR too.  But how about calling `base::toString` generic? Then user can define `toString` methods.

MichaelChirico:
I seem to recall trying out `toString` for another PR and decided against it, I think because of the 3.1.0 dependency? Though the function itself is just a wrapper, could easy do an `ifdef`-type approach for older R versions.

---

Found the PR/discussion, see

#2562 / #1523

MichaelChirico:
Actually realizing this covers exactly Frank's comment here:

https://github.com/Rdatatable/data.table/issues/1523#issuecomment-413937502

MichaelChirico:
Succeeded in #3414

PR #3341, opened 2019-01-31
na.fill LOCF, NOCB, #854

### LINKED ISSUE #854: Fill in missing data
http://stackoverflow.com/questions/26171958/fill-in-missing-values-by-group-in-data-table

IMO, the best syntax  would be  a function `fill_na`

``` R
DT[, value_filled := fill_na(value, roll , rollends), by = id]
```

but it would require some work - I'm not even sure it's feasible. 

An easier solution would be a function `setna` that works depending on the keys of DT

``` R
setna <- function(.data, cols,roll = TRUE ,  rollends = if (roll=="nearest") c(TRUE,TRUE)
  else if (roll>=0) c(FALSE,TRUE)
  else c(TRUE,FALSE)){
  keys <- key(.data)    
  if (length(keys)<1){
    stop(".data must be keyed by at least one variable")
  }   
  for (col in cols){
    eval(substitute(.data[, (col) := .data[!is.na(x)][.data, value, roll = roll, rollends = rollends]], list(x = as.name(col))))
  }
}
setkey(DT, id, date)
setna(DT, "value")
```

jangorecki:
Closes #854
It has much bigger scope but this PR I think address most of the user use cases.
For integer and double should be fully functional. There is some redundancy there in the process, declaring pointers to integer and double, even if only one of those has been used. That allows to use more generic struct (`types.h` file), which can point to int and double as well.
Looking forward for feedback.


jangorecki:
Below optimistic scenario for data.table where 20 columns has to be adjusted with NOCF on a 20 cores machine, vs single threaded R implemented zoo.
```r
set.seed(108)
library(data.table)
library(zoo)
packageVersion("data.table")
#[1] ‘1.12.1’
packageVersion("zoo")
#[1] ‘1.8.4’
N=1e8
x=rnorm(N)
random_na = function(x, k) {
  nx = length(x)
  if (k > nx) k = nx/2 # dont need more than half of NAs
  i = sample(nx, k)
  x[i] = NA
  x
}
d = lapply(1:20, function(i) random_na(x=x+i, k=i*1e5))
sapply(d[1:3], function(x) sum(is.na(x)))
#[1] 100000 200000 300000
setDT(d)
d[1:3, 1:3]
#          V1       V2       V3
#       <num>    <num>    <num>
#1: 0.8871079 1.887108 2.887108
#2: 0.6181806 1.618181 2.618181
#3: 0.9083082 1.908308 2.908308
prettyNum(dim(d), big.mark="'")
#[1] "100'000'000" "20"

system.time(setDF(ans1<-fna.fill(d, "locf")))
#   user  system elapsed
#  8.834   7.456   0.890

setDF(d)
system.time(ans2<-na.locf(d, na.rm=FALSE))
#   user  system elapsed 
# 56.552  36.852  93.403

all.equal(ans1, ans2)
#[1] TRUE
```

jangorecki:
also provides simple API to one the very popular problems: https://stackoverflow.com/a/54504456/2490497

mattdowle:
This PR seems to export two new functions: `fnafill` and `setnafill`.  What's the difference between them?  My first thought is that there are two operations:  i) replacing NAs with another value (like 0) and ii) filling NAs forwards with the last observation.  Then there's `nomatch=` (and the pending issue to move LOCF() into nomatch=) and `roll=`. And similar function names in other package like na.locf.

The point of data.table is to keep data irregular.  Then join a regular series to it using roll= (or in future nomatch=) to fill forward or backwards or limit staleness.  I wonder if the people wanting to fill NAs realize that it's not supposed to be done that way.

jangorecki:
Both functions can replace with constant value and locf/nocb, the difference that `setnafill` makes that in-place, which is actually much more common scenario than creating new columns. `nomatch` will be designed to be used during joins, and will likely carry some overhead because of that.
Do not merge this PR yet, I am planing to add `cols` argument to `setnafill` functions so it will be more convenient to pass whole dt to it, and fill only particular columns, like `setnafill(bigdt, "locf", cols=c("v3","v5"))` rather than `dt[, setnafill(list(v3, v5), "locf")]` (counter intuitive to update in place without `:=`) or `setnafill(list(bigdt$v3, bigdt$v5), "locf")`.

jangorecki:
I added mentioned functionality so it is easy to fill only selected columns from data.table. New argument `cols` accept int, double or character.
```r
setnafill(dt, fill=0, cols=c("V1","V2"))
setnafill(dt, fill=0, cols=1:2)
setnafill(dt, fill=0, cols=c(1,2)
```
Processing of that argument has been isolated into own function `SEXP colnamesInt(SEXP x, SEXP cols);` where `x` is a data.table to match columns against to, and return value is `SEXP integer vector` of column positions. We should be able to re-use it in few places, but lets keep scope of this PR narrow.

jangorecki:
Added `verbose` argument to print to console time spent on each column, and all of them. Processing time of  of each column is measured in parallel region, but printing is deferred so it is thread safe. Example on 40 thread machine and 40 columns data.table.
```r
library(data.table)
N = 1e8
dt = setDT(c(
  replicate(20, sample(c(seq_len(N*0.9),rep(NA_integer_, N*0.1))), simplify=FALSE),
  replicate(20, sample(c(seq_len(N*0.9)/2,rep(NA_real_, N*0.1))), simplify=FALSE)
))
system.time(ans<-nafill(dt, "locf", verbose=TRUE))
```
```
nafillInteger: took 2.095s
nafillInteger: took 2.017s
nafillInteger: took 2.051s
nafillInteger: took 2.028s
nafillInteger: took 1.861s
nafillInteger: took 2.075s
nafillInteger: took 1.907s
nafillInteger: took 1.872s
nafillInteger: took 1.974s
nafillInteger: took 1.798s
nafillInteger: took 2.107s
nafillInteger: took 1.864s
nafillInteger: took 2.023s
nafillInteger: took 1.730s
nafillInteger: took 2.056s
nafillInteger: took 1.866s
nafillInteger: took 2.024s
nafillInteger: took 1.892s
nafillInteger: took 1.978s
nafillInteger: took 1.956s
nafillDouble: took 3.140s
nafillDouble: took 2.987s
nafillDouble: took 3.148s
nafillDouble: took 3.003s
nafillDouble: took 3.828s
nafillDouble: took 3.060s
nafillDouble: took 3.135s
nafillDouble: took 3.209s
nafillDouble: took 3.090s
nafillDouble: took 3.069s
nafillDouble: took 3.052s
nafillDouble: took 3.200s
nafillDouble: took 3.163s
nafillDouble: took 3.113s
nafillDouble: took 3.220s
nafillDouble: took 3.167s
nafillDouble: took 3.129s
nafillDouble: took 3.010s
nafillDouble: took 3.008s
nafillDouble: took 2.497s
nafillR: parallel processing of 40 column(s) took 3.979s
   user  system elapsed 
 23.375  75.110   4.138
```

MichaelChirico:
```
nafillInteger: took 2.095s
nafillInteger: took 2.017s
nafillInteger: took 2.051s
nafillInteger: took 2.028s
nafillInteger: took 1.861s
nafillInteger: took 2.075s
nafillInteger: took 1.907s
nafillInteger: took 1.872s
```

Such output doesn't look all that helpful. Everything is subsumed by `nafillR: parallel processing of 40 column(s) took 3.979s`, no?

Is it possible to tag the output with the column name, e.g.

```
nafillInteger: V1 took 2.095s
nafillInteger: V2 took 2.017s
nafillInteger: V3 took 2.051s
nafillInteger: V4 took 2.028s
...
```

jangorecki:
> Is it possible to tag the output with the column name, e.g.

Inner functions that populate that message are not aware of column name. It is like passing a list to `lapply`, you cannot access the name of each list.
This is actually good suggestion to extend `ans_t` struct to carry a tag. This raise problem how to define such tag, column name is obvious, but why not column index? for froll it wouldn't be just column name but concatenated column name and window size. And because it is a low level structure it would be best to have it well and strictly defined.Is it possible to tag the output with the column name, e.g.

> What about a case of 100s of columns, do we still want to print everything?

Yes, there is `verbose=FALSE` to relief that pain. The only reasonable solution would be to allow `verbose` to be int, decremented with each inner function call. So for more details you call `verbose=2`, for seeing just messages from top called function `verbose=1`.

MichaelChirico:
> there is `verbose=FALSE` to relief that pain

True (not `TRUE` 😉), OK.

> why not column index?

Also would be fine. Something to make the output there actionable... "Oh, Column 45 took 10x as long as all the other columns, I should investigate..."

jangorecki:
@MichaelChirico added index in printed message, it is indexing from 0 btw.

MichaelChirico:
Switched to one-based unless there's something I'm missing?

mattdowle:
I added a brief news item. Can someone add a good brief example to the news item please as follow up PR.
I edited Jan's original comment to say that this PR closes #854.  If that wasn't right thing to do,  could anything not addressed be added as new follow up issues please.

src/nafill.c
```diff
@@ -0,0 +1,194 @@
...
+    case INTSXP :
+      nafillInteger(ix[i], inx[i], itype, ifill, &vans[i], bverbose);
+      break;
+    }
+  }
+  if (bverbose) toc = omp_get_wtime();
+
+  for (R_len_t i=0; i<nx; i++) {
+    if (bverbose && (vans[i].message[0][0] != '\0')) Rprintf("%s: %d: %s", __func__, i+1, vans[i].message[0]);
+    if (vans[i].message[1][0] != '\0') REprintf("%s: %d: %s", __func__, i+1, vans[i].message[1]); // # nocov start
```
jangorecki: this cannot be suppressed with `suppressMessages`

src/nafill.c
```diff
@@ -0,0 +1,194 @@
...
+    error("fill must be a vector of length 1");
+
+  double dfill=NA_REAL;
+  int32_t ifill=NA_INTEGER;
+  if (itype==0) {
+    if (isInteger(fill)) {
+      ifill = INTEGER(fill)[0];
+      dfill = INTEGER(fill)[0]==NA_INTEGER ? NA_REAL : (double)INTEGER(fill)[0];
+    } else if (isReal(fill)) {
+      ifill = ISNA(REAL(fill)[0]) ? NA_INTEGER : (int32_t)REAL(fill)[0];
```
jangorecki: @mattdowle is this type of casting safe? and similar in line 152. Should I always use `coerceVector` in such cases?

jangorecki: it looks that `coerceVector` is not necessary: https://github.com/Rdatatable/data.table/pull/3627/commits/638382355a08bb9fe2d4b7927702548695e74e2a#diff-74714f9d6425434f378360350d1a4c78R41

PR #3344, opened 2019-02-02
Test arbitrary script

jangorecki:
more flexibility in running non-standard scripts
prepare for moving froll test scripts to use `test.data.table`
ensure all scripts runs even if one failed

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -95,7 +95,7 @@ oldOptions = options(
 ##########################
 
 test(1.1, tables(env=new.env()), null.data.table(), output = "No objects of class")
-test(1.2, tables(silent=TRUE), data.table(NAME="timings", NROW=3000L, NCOL=3L, MB=0, COLS=list(c("ID","time","nTest")), KEY=list(NULL)))
+test(1.2, tables(silent=TRUE), data.table(NAME="timings", NROW=9999L, NCOL=3L, MB=0, COLS=list(c("ID","time","nTest")), KEY=list(NULL)))
 
 TESTDT = data.table(a=as.integer(c(1,3,4,4,4,4,7)), b=as.integer(c(5,5,6,6,9,9,2)), v=1:7)
 setkey(TESTDT,a,b)
@@ -13315,4 +13315,3 @@ test(1980, names(data.table(x)), "x")
 ###################################
 
 options(oldOptions)  # set at top of this file
-setDTthreads(0) # this has not been altered directly before so not sure if we should really reset it here
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,7 @@
 variables:
   CRAN_MIRROR: "https://cloud.r-project.org"
-  _R_CHECK_FORCE_SUGGESTS_: "FALSE"
+  _R_CHECK_FORCE_SUGGESTS_: "false"
+  _R_CHECK_NO_STOP_ON_TEST_ERROR_: "true"
 
 stages:
   - dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,5 +58,6 @@ notifications:
 env:
   global:
     - PKG_CFLAGS="-O3 -Wall -pedantic"
+    - _R_CHECK_NO_STOP_ON_TEST_ERROR_=true
     # drat using @jangorecki token
     - secure: "CxDW++rsQApQWos+h1z/F76odysyD6AtXJrDwlCHlgqXeKJNRATR4wZDDR18SK+85jUqjoqOvpyrq+5kKuyg6AnA/zduaX2uYE5mcntEUiyzlG/jJUKbcJqt22nyAvFXP3VS60T2u4H6IIhVmr7dArdxLkv8W+pJvf2Tg6kx8Ws="
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,6 +23,8 @@ environment:
 # The default for R_ARCH is i386, though, for which multiarch would just compile and test 32bit, hence setting R_ARCH to x64
     GCC_PATH: mingw_64
 # Default GCC_PATH appears to be gcc-4.6.3 which is now unsupported as from Rtools.exe v3.4.
+    _R_CHECK_NO_STOP_ON_TEST_ERROR_: true
+# continue tests even if some script failed
 
   matrix:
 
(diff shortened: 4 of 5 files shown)


R/test.data.table.R
```diff
@@ -1,27 +1,34 @@
-test.data.table <- function(verbose=FALSE, pkg="pkg", silent=FALSE, with.other.packages=FALSE, benchmark=FALSE, Rraw=NULL) {
-  if (exists("test.data.table",.GlobalEnv,inherits=FALSE)) {
+test.data.table <- function(verbose=FALSE, pkg="pkg", silent=FALSE, with.other.packages=FALSE, benchmark=FALSE, script=NULL) {
```
jangorecki: Renamed as any file extension will work. Also file doesn't have to be in `inst/tests`, so anyone can do own test script using our `test()` calls and run it `test.data.table(script="../myproj/my-dt-test-script.R")`. It works also from `cc("~/Desktop/my_dev_script.R")`.

R/test.data.table.R
```diff
@@ -1,27 +1,34 @@
...
+  fulldir = file.path(rootdir, subdir)
 
-  if (!is.null(Rraw)) {
-    stopifnot(is.character(Rraw), length(Rraw)==1L, !is.na(Rraw), nzchar(Rraw))
-    fn = Rraw
+  if (!is.null(script)) {
+    stopifnot(is.character(script), length(script)==1L, !is.na(script), nzchar(script))
+    if (!identical(basename(script), script)) {
+      subdir = dirname(script)
+      fulldir = normalizePath(subdir, mustWork=FALSE)
```
jangorecki: `mustWork=FALSE` as there is check below on `fn` object

R/test.data.table.R
```diff
@@ -1,27 +1,34 @@
...
+      fn = script
+    }
   } else {
     stopifnot( !(with.other.packages && benchmark) )
     fn = if (with.other.packages) "other.Rraw"
          else if (benchmark) "benchmark.Rraw"
          else "tests.Rraw"
   }
-  fn = file.path(d, fn)
+  fn = setNames(file.path(fulldir, fn), file.path(subdir, fn))
```
jangorecki: `fulldir` is full path, while when printing summary we might want to retain current better looking `inst/test/file.Rraw` rather than full path. Thus this nice path is now stored as name of `fn`.

R/test.data.table.R
```diff
@@ -36,24 +43,24 @@ test.data.table <- function(verbose=FALSE, pkg="pkg", silent=FALSE, with.other.p
...
-  assign("testDir", function(x)file.path(d,x), envir=env)
+  assign("testDir", function(x) file.path(fulldir, x), envir=env)
   assign("nfail", 0L, envir=env)
   assign("ntest", 0L, envir=env)
   assign("whichfail", NULL, envir=env)
   setDTthreads(2) # explicitly limit to 2 so as not to breach CRAN policy (but tests are small so should not use more than 2 anyway)
   assign("started.at", proc.time(), envir=env)
   assign("lasttime", proc.time()[3L], envir=env)  # used by test() to attribute time inbetween tests to the next test
-  assign("timings", data.table( ID = seq_len(3000L), time=0.0, nTest=0L ), envir=env)   # test timings aggregated to integer id
+  assign("timings", data.table( ID = seq_len(9999L), time=0.0, nTest=0L ), envir=env)   # test timings aggregated to integer id
```
jangorecki: so tests can have up to 9999 num

cc.R
```diff
@@ -82,7 +82,7 @@ cc = function(test=TRUE, clean=FALSE, debug=FALSE, omp=!debug, cc_dir=Sys.getenv
   sourceDir(paste0(cc_dir,"/R"))
   assign("testDir", function(x)paste0(cc_dir,"/inst/tests/",x), envir=.GlobalEnv)
   .onLoad()
-  if (test) test.data.table()
+  if (is.logical(test) && isTRUE(test)) test.data.table() else if (is.character(test)) test.data.table(script=test)
```
jangorecki: `cc(T)` `cc(F)` same as before, but now also `cc("tests.Rraw")`, `cc("/home/user/git/data.table/inst/tests/tests.Rraw"`) works

inst/tests/tests.Rraw
```diff
@@ -13315,4 +13315,3 @@ test(1980, names(data.table(x)), "x")
 ###################################
 
 options(oldOptions)  # set at top of this file
-setDTthreads(0) # this has not been altered directly before so not sure if we should really reset it here
```
jangorecki: this has not been set anywhere in the script so there is no need to restore. Number of threads should be controlled here by env var, so CRAN can be handled nicely. Assuming `getDTthreads` will work properly.

inst/tests/froll.Rraw
```diff
@@ -0,0 +1,838 @@
...
+ans3 = frollmean(x, n, algo="exact")
+ans4 = frollmean(x, n)
+anserr = list(
+  froll_exact_f = ans4-ans1,
+  froll_exact_t = ans3-ans1,
+  fastma = ans2-ans1
+)
+errs = sapply(lapply(anserr, abs), sum, na.rm=TRUE)
+if (.Platform$OS.type=="unix") {
+  test(9999.024, errs[["froll_exact_t"]]==0)
```
jangorecki: @mattdowle is it safe to assume exactness on linux? It failed on windows. If we are not sure we could change to `all.equal`, for windows it was not enough.

jangorecki: Actually fix didn't resolve windows issue, `exact=T` happens to be less exact than `exact=F`. Will need to investigate that on Windows. We are likely to hit that problem on CRAN as froll tests will kick in there on windows.

tests/froll.R
```diff
@@ -1,832 +1,2 @@
...
-x = makeNA(rnorm(1e3)); n = sample(51, length(x), TRUE)
-afun_compare(x, n)
-x = makeNA(rnorm(1e3+1)); n = sample(51, length(x), TRUE)
-afun_compare(x, n)
-rm(num)
-
-setDTthreads(oldDTthreads)
-if (length(.test.froll.failed)) stop(sprintf("froll unit tests failed: %s.", paste(.test.froll.failed, collapse=", "))) else cat("froll unit tests successfully passed\n")
+require(data.table)
+test.data.table(script="froll.Rraw")
```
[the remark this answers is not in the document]
jangorecki: this will go in another PR after fixing windows rounding issue.

inst/tests/froll.Rraw
jangorecki: fixing rounding issue here is not really related to this PR so will be submitted separately to not block this PR

Issue #3301, opened 2019-01-22
Remove #define USE_RINTERNALS


mattdowle:
With ALTREP in R 3.5+ it's no longer true that defining USE_RINTERNALS before including Rinternals.h makes REAL(), INTEGER() etc into macros.   That's why I did it originally and it's no longer true, so there's no point in doing it anymore. It was good advice because it was a lot faster. The new advice with R 3.5+ is to not use the R API inside loops.
See also: https://github.com/Rdatatable/data.table/issues/2866#issuecomment-389643625 and https://github.com/Rdatatable/data.table/issues/3188#issuecomment-444757748

Section 5.14 of R-exts manual contains : 

> If efficiency is essential, the internal versions of the accessors can be obtained by defining ‘USE_RINTERNALS’ before including Rinternals.h. If you find it necessary to do so, please do test that your code compiles without ‘USE_RINTERNALS’ defined, as this provides a stricter test that the accessors have been used correctly. Also be prepared to adjust your code should R internals change. 
> The accessor functions, and other functions in the R API, are also subject to change to support the ‘ALTREP’ project (https://svn.r-project.org/R/branches/ALTREP/ALTREP.html). Code that does not define ‘USE_RINTERNALS’ should not be affected by these changes, but code that does define ‘USE_RINTERNALS’ may need to be adjusted.

I've tried turning off USE_RINTERNALS and data.table compiles and passes all tests with latest versions of R because DATAPTR is exposed in the R API. But still we're not supposed to use it. Not passing R 3.1.0 until DATAPTR is removed (~72~ 22 calls) : 
- [x] assign.c (#3610 #4095; v1.12.8)
- [x] bmerge.c (#3610; v1.12.4)
- [x] dogroups.c (#3610 #4164; v1.12.10)
- [x] fmelt.c (#3350; v1.12.2)
- [x] freadR.c [#4164; v1.12.10]
- [x] fwriteR.c [#4164; v1.12.10]
- [x] gsumm.c (#3350; v1.12.2)
- [x] quickselect.c (#3377; v1.12.2)
- [x] rbindlist.c (#3350 #3455; v1.12.2)
- [x] reorder.c (#4164; v1.12.10)
- [x] shift.c (#3348; v1.12.2)
- [x] uniqlist.c (#3350; v1.12.2)

jangorecki:
[...]

PR #3350, opened 2019-02-04
remove some DATAPTR

jangorecki:
#3301, some obvious DATAPTR to remove
@mattdowle check one inline comment

### DIFF
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -441,7 +441,7 @@ SEXP getvaluecols(SEXP DT, SEXP dtnames, Rboolean valfactor, Rboolean verbose, s
           for (k=0; k<thislen; k++)
             REAL(target)[counter + k] = REAL(thiscol)[INTEGER(thisidx)[k]-1];
         } else {
-          memcpy((char *)DATAPTR(target)+j*data->nrow*size, (char *)DATAPTR(thiscol), data->nrow*size);
+          memcpy((char *)REAL(target)+j*data->nrow*size, (char *)REAL(thiscol), data->nrow*size);
         }
         break;
         case INTSXP :
@@ -449,7 +449,7 @@ SEXP getvaluecols(SEXP DT, SEXP dtnames, Rboolean valfactor, Rboolean verbose, s
           for (k=0; k<thislen; k++)
             INTEGER(target)[counter + k] = INTEGER(thiscol)[INTEGER(thisidx)[k]-1];
         } else {
-          memcpy((char *)DATAPTR(target)+j*data->nrow*size, (char *)DATAPTR(thiscol), data->nrow*size);
+          memcpy((char *)INTEGER(target)+j*data->nrow*size, (char *)INTEGER(thiscol), data->nrow*size);
         }
         break;
         case LGLSXP :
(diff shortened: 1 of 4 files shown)


src/uniqlist.c
```diff
@@ -65,7 +65,7 @@ SEXP uniqlist(SEXP l, SEXP order)
       }
     } break;
     case STRSXP : {
-      SEXP *vd=(SEXP *)DATAPTR(v), prev, elem;   // TODO: tried to replace DATAPTR here but (SEXP *)&STRING_ELT(v,0) results in lvalue required as unary ‘&’ operand
+      SEXP *vd=(SEXP *)CHAR(v), prev, elem;
```
jangorecki: `CHAR` I found in http://biostat.mc.vanderbilt.edu/wiki/Main/RInternals please take a look before merging @mattdowle

src/fmelt.c
```diff
@@ -618,7 +618,7 @@ SEXP getidcols(SEXP DT, SEXP dtnames, Rboolean verbose, struct processData *data
         // From assign.c's memcrecycle - only one SET_STRING_ELT per RHS item is needed to set generations (overhead)
         for (k=0; k<data->nrow; k++) SET_STRING_ELT(target, k, STRING_ELT(thiscol, k));
         for (j=1; j<data->lmax; j++)
-          memcpy((char *)DATAPTR(target)+j*data->nrow*size, (char *)DATAPTR(target), data->nrow*size);
+          memcpy((char *)CHAR(target)+j*data->nrow*size, (char *)CHAR(target), data->nrow*size);
```
jangorecki: and here `CHAR` macro

jangorecki: this should stress this change in `getidcols` fun, tested on master vs this branch and looks good
```r
d = data.table(a=as.character(letters[1:5]), v1=1:5, v2=1:5)
d[, "b" := paste0(a,rev(a),a)]
d[, "a" := paste0(a,a,a)]
melt.data.table(d, id=c("a","b"), measure.vars=c("v1","v2"))
```

src/rbindlist.c
```diff
@@ -736,8 +736,8 @@ SEXP rbindlist(SEXP l, SEXP sexp_usenames, SEXP sexp_fill, SEXP idcol) {
       case INTSXP:
       case LGLSXP:
         if (TYPEOF(thiscol) != TYPEOF(target)) error("Internal logical error in rbindlist.c, type of 'thiscol' should have already been coerced to 'target'. please report to data.table issue tracker.");
-        memcpy((char *)DATAPTR(target) + ansloc * SIZEOF(thiscol),
-             (char *)DATAPTR(thiscol),
+        memcpy((char *)LOGICAL(target) + ansloc * SIZEOF(thiscol),
+             (char *)LOGICAL(thiscol),
```
mattdowle: We can't use LOGICAL on a REALSXP.  That fails with a check that R does when USE_RINTERNALS is not defined.

src/uniqlist.c
mattdowle: See https://github.com/Rdatatable/data.table/pull/3350#pullrequestreview-199731517

src/rbindlist.c
jangorecki: oops, I missed it refers to REAL, INT and LGL. Thanks

PR #3359, opened 2019-02-05
Update setDT.Rd

DavidArenburg:
Keep `?setDT` docs consistent with `?as.data.table` regarding the fact that `keep.rownames` accepts not only a Boolean rather you can pass it a new column name.

Mentioned on SO under a high volume Q/A by @snoram [here](https://stackoverflow.com/questions/29511215/convert-row-names-into-first-column/29511387?noredirect=1#comment95868434_29511387)

### DIFF
--- a/man/setDT.Rd
+++ b/man/setDT.Rd
@@ -12,7 +12,7 @@ setDT(x, keep.rownames=FALSE, key=NULL, check.names=FALSE)
 }
 \arguments{
   \item{x}{ A named or unnamed \code{list}, \code{data.frame} or \code{data.table}. }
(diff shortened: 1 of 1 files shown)


jangorecki:
Thanks for PR, it would be ideally to support that manual change with one simple unit test.

DavidArenburg:
@jangorecki A unit test for an .Rd file?

jangorecki:
Unit test that tests feature mentioned in manual manual change, if we don't have unit test for that and it was not documented before, then it could eventually silently stopped working making this new documentation invalid. I found test which covers that, so there is no need for adding another one: 1508.4.

PR #3360, opened 2019-02-05
travis do only 2 builds

### LINKED ISSUE #3357: speed up travis builds
https://github.com/Rdatatable/data.table/pull/3331 PR submitted by @MarcusKlik is very useful. Downside of it is that it increase travis check time from 8 minutes to 50 minutes. As we have extended CI on GitLab it make sense to keep travis fast. We still want to have OSX there, but 1 Linux build and 1 OSX build will be enough.

jangorecki:
Closes #3357 
it takes now 20min. Osx build is slow, takes 15min, probably because of installing llvm each time. @MarcusKlik any idea if we can cache that somehow?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,19 @@
-#
-# Travis configuration guide for R language:
-#   https://docs.travis-ci.com/user/languages/r/
-#   https://github.com/craigcitro/r-travis/wiki/Porting-to-native-R-support-in-Travis
-#
 language: r
 dist: trusty
 sudo: required
 cache: packages
 warnings_are_errors: true
-r_check_args: ""
-# empty r_check_args to turn off --as-cran (on by default) as that can be slow
 
 branches:
   only:
     - "master"
 
 r:
-  - oldrel
   - release
-  - devel
 
 os:
   - linux
   - osx
-
-matrix:
-  exclude:
-    - r: devel  # exclude osx r-devel build as it is currently unstable
-      os: osx
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install llvm &&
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,50 +1,36 @@
...
-#   https://docs.travis-ci.com/user/languages/r/
-#   https://github.com/craigcitro/r-travis/wiki/Porting-to-native-R-support-in-Travis
-#
 language: r
 dist: trusty
 sudo: required
 cache: packages
 warnings_are_errors: true
-r_check_args: ""
-# empty r_check_args to turn off --as-cran (on by default) as that can be slow
```
jangorecki: `_R_CHECK_CRAN_INCOMING_REMOTE_=false` defined below should resolve problem of slow `--as-cran`

MarcusKlik:
Hi @jangorecki ,

yes, I see that the `llvm` installation takes almost 5 minutes, that's quite long!
The same is true for the [brew update](https://travis-ci.org/MarcusKlik/data.table/jobs/489252678#L55) command at the start of the build, that one also takes a bite out of the total running time.

We could cache the `homebrew` directory to cache the `llvm` download:

```bash
cache:
  directories:
    - $HOME/Library/Caches/Homebrew

before_cache:
  - brew cleanup
```

But as Travis stores the cache at a storage provider (see [this doc](https://docs.travis-ci.com/user/caching/#how-does-caching-work)), I think that downloading from the cache will probably not be much faster than a direct download.

Also, following the Travis documentation on [caching](https://docs.travis-ci.com/user/caching/) it seems that we should be able to cache the `llvm` installation directory itself, but somehow `homebrew` [fails to install llvm correctly](https://travis-ci.org/MarcusKlik/data.table/jobs/489252678#L174-L176) on the cached (initially empty) directory.

Some more testing to be done I think...

jangorecki:
Thanks @MarcusKlik for detailed answer on that. I think we can also remove test coverage report from OSX build. It takes 330s and linux test coverage should be enough.

Issue #2960, opened 2018-07-02
`test` function is insensitive for NaN vs NA difference


jangorecki:
For differentiating `NA` from `NaN` currently behavior of `test` function is consistent to `all.equal` while more appropriate IMO would be to expect `identical` behavior.
```r
all.equal(NaN, NA_real_)
#[1] TRUE
identical(NaN, NA_real_)
#[1] FALSE
```
Unfortunately it is not that simple, we would need to overwrite `all.equal.numeric` to retain current features that comes from `all.equal` and be able to differentiate NA from NaN.

This ideally should be fixed upstream in R giving an option to differentiate both.

jangorecki:
bump...
just refreshing memory as I got hit by this again

PR #3364, opened 2019-02-06
escape tests if NaN-NA behaviour does not match

### LINKED ISSUE #3353: froll compiled using -O0 fails 2 tests
We've lost coverage stats due to this : 

```
Still running (2 of 20): Rscript -e library(covr);codecov()
The command Rscript -e library(covr);codecov() exited with 1.
Log:
Error: Failure in `/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/Rtmp1nDGQA/R_LIBS154b3111ce992/data.table/data.table-tests/froll.Rout.fail`
UE)
> afun_compare(x, n)
> x = c(rep(NA, 60), rnorm(1e3), rep(NA, 60)); n = sample(51, length(x), TRUE)
> afun_compare(x, n)
> x = c(rep(NA, 60), rnorm(1e3+1), rep(NA, 60)); n = sample(51, length(x), TRUE)
> afun_compare(x, n)
> #### random NA
> x = makeNA(rnorm(1e3)); n = sample(50, length(x), TRUE)
> afun_compare(x, n)
> x = makeNA(rnorm(1e3+1)); n = sample(50, length(x), TRUE)
> afun_compare(x, n)
> x = makeNA(rnorm(1e3)); n = sample(51, length(x), TRUE)
> afun_compare(x, n)
> x = makeNA(rnorm(1e3+1)); n = sample(51, length(x), TRUE)
> afun_compare(x, n)
> rm(num)
> 
> setDTthreads(oldDTthreads)
> if (length(.test.froll.failed)) stop(sprintf("froll unit tests failed: %s.", paste(.test.froll.failed, collapse=", "))) else cat("froll unit tests successfully passed\n")
Error: froll unit tests failed: 9999.123, 9999.162.
Execution halted
Execution halted
/Users/travis/.travis/job_stages: line 523: 87220 Terminated: 15          travis_jigger "${!}" "${timeout}" "${cmd[@]}"
```

jangorecki:
Closes #3353.

### DIFF
--- a/inst/tests/froll.Rraw
+++ b/inst/tests/froll.Rraw
@@ -433,32 +433,43 @@ ma = function(x, n, na.rm=FALSE, nf.rm=FALSE) {
   if (!is.double(x)) x = as.double(x)
   if (!is.integer(n)) n = as.integer(n)
   ans = rep(NA_real_, nx<-length(x))
-  if (nf.rm) x[!is.finite(x)] = NA_real_ # exact=F consistency due to https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17441
+  if (nf.rm) x[!is.finite(x)] = NA_real_ # algo=fast consistency due to https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17441
   for (i in n:nx) ans[i]=mean(x[(i-n+1):i], na.rm=na.rm)
   ans
 }
+compiler_o_flag_match = function() {
+  ## escape tests for NA/NaN exactness if compiled with -O0 flag so NaN behaviour changes: #3353 needed only for .123 and .162 but few other tests might be affected on different platforms also
+  ## related to https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17441
+  x = c(NA, NaN)
+  y = c(NaN, NA)
+  `&&`(
+    identical(mean(x), frollmean(x, 2L, algo="exact")[2L]),
+    identical(mean(y), frollmean(y, 2L, algo="exact")[2L])
+  )
+}
+do_NaN_NA_tests = compiler_o_flag_match()
 n = 4
 x = 1:16
 x[5] = NaN
 test(9999.120, identical(frollmean(x, n), ma(x, n, nf.rm=TRUE)))
-test(9999.121, identical(frollmean(x, n, algo="exact"), ma(x, n)))
+if (do_NaN_NA_tests) test(9999.121, identical(frollmean(x, n, algo="exact"), ma(x, n)))
 x[6] = NA
 test(9999.122, identical(frollmean(x, n), ma(x, n, nf.rm=TRUE)))
-test(9999.123, identical(frollmean(x, n, algo="exact"), ma(x, n)))
+if (do_NaN_NA_tests) test(9999.123, identical(frollmean(x, n, algo="exact"), ma(x, n)))
 #### test inconsistency of NaN-NA order is consistent to https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17441
 x[5] = NA
 x[6] = NaN
 test(9999.124, identical(frollmean(x, n), ma(x, n, nf.rm=TRUE)))
-test(9999.125, identical(frollmean(x, n, algo="exact"), ma(x, n)))
+if (do_NaN_NA_tests) test(9999.125, identical(frollmean(x, n, algo="exact"), ma(x, n)))
 x[5] = Inf
 test(9999.126, identical(frollmean(x, n), ma(x, n, nf.rm=TRUE)))
-test(9999.127, identical(frollmean(x, n, algo="exact"), ma(x, n)))
+if (do_NaN_NA_tests) test(9999.127, identical(frollmean(x, n, algo="exact"), ma(x, n)))
 x[6] = -Inf
 test(9999.128, identical(frollmean(x, n), ma(x, n, nf.rm=TRUE)))
-test(9999.129, identical(frollmean(x, n, algo="exact"), ma(x, n)))
+if (do_NaN_NA_tests) test(9999.129, identical(frollmean(x, n, algo="exact"), ma(x, n)))
 x[5:7] = c(NA, Inf, -Inf)
 test(9999.130, identical(frollmean(x, n), ma(x, n, nf.rm=TRUE)))
-test(9999.131, identical(frollmean(x, n, algo="exact"), ma(x, n)))
+if (do_NaN_NA_tests) test(9999.131, identical(frollmean(x, n, algo="exact"), ma(x, n)))
 
 #### adaptive window
 ama = function(x, n, na.rm=FALSE, fill=NA, nf.rm=FALSE) {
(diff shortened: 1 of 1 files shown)


inst/tests/froll.Rraw
```diff
@@ -433,32 +433,43 @@ ma = function(x, n, na.rm=FALSE, nf.rm=FALSE) {
...
   for (i in n:nx) ans[i]=mean(x[(i-n+1):i], na.rm=na.rm)
   ans
 }
+compiler_o_flag_match = function() {
+  ## escape tests for NA/NaN exactness if compiled with -O0 flag so NaN behaviour changes: #3353 needed only for .123 and .162 but few other tests might be affected on different platforms also
+  ## related to https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17441
+  x = c(NA, NaN)
+  y = c(NaN, NA)
+  `&&`(
+    identical(mean(x), frollmean(x, 2L, algo="exact")[2L]),
```
mattdowle: Better to test the `x` and `y` independently against fixed known results (`NA` or `NaN` in this case). Then I for example as reader of the test can more easily see what is happening here.  For example, I'm currently wondering if base R changes result depending on whether R itself has been compiled with -O0 or not, or whether it's just frollmean that changes result depending on whether data.table has been compiled with -O0 or not.
Also the wording that this behavior depends on -O0 seems odd.  It's really undefined or variable behaviour that -O0 picks up currently.  But equally it could happen on other platforms:  Solaris and PowerPC spring to mind.
How about changing the tests to test if either NA or NaN are returned?  The PR currently turns off the tests, so in the event that `Inf` was returned (a definite error), the test wouldn't run because it was turned off.  Better to always test that one of the accepted values is always returned.  `is.na(NA)` and `is.na(NaN)` both return `TRUE`, so `is.na()` could be used by the test.

jangorecki: > Equally, couldn't it happen on other platforms? Solaris and Power spring to mind.

NaN - NA interaction is not undefined by any standards, thus is platform specific, and what we can observe here, also `-O` flag specific.

----

> How about changing the tests to test if _either_ NA or NaN are returned?

Those particular tests are there to test `algo="exact"` which has to be identical to base R result. [`froll` manual](https://rdatatable.gitlab.io/data.table/library/data.table/html/froll.html) states:

> When set to `"exact"` ... carefully handles all non-finite values.

and

> It also handles NaN, +Inf, -Inf consistently to base R.

There are other tests for `algo="fast"` where NaN as treated as NA.
Agree on that it is better to ignore NaN-NA difference than just disabling those tests. I will amend.
BTW. slightly related #2960

----

> Better to always test that one of the accepted values is always returned.

Best to not "always" but only when issue that issue occurs.

jangorecki:
Changes amended, R mean is compared vs constant values, same frollmean is compared against constant values. Based on that comparison we either replace NaN to NA or leave as is.

mattdowle:
As Martin wrote in #17441 : 
"See ?NA, which says  
  Numerical computations using NA will normally result in NA: a possible exception is where NaN is also involved, in which case either might result (which may depend on the R platform). "

Given this, why not make frollmean return a consistent result which does not depend on the platform? In other areas of data.table we have departed from base R where it makes sense to. I don't see why you are trying to match identically the platform-specific behavior. Or, relax these tests to accept NA or NaN when mixed NA and NaN inputs are present.  I don't think anyone needs to match precisely the results when mixed NA, NaN are present.  Especially when R doesn't guarantee a particular result in this case anyway.

Put another way ... `maybe_NaN_to_NA()` is used in tests which test `frollmean()`.  But `maybe_NaN_to_NA()` itself contains calls to `frollmean()` and switches depending on the behavior observed.  That's a circle that's too complicated in a test environment.  It should be cleaner and simpler.

jangorecki:
> Given this, why not make frollmean return a consistent result which does not depend on the platform?

It will cost performance. We outsource that part to code generated by compiler, same as R.
Quoting Tomas Kalibera on why R cannot have consistent behaviour on that:
> Yes, the performance overhead of fixing this at R level would be too 
large and it would complicate the code significantly. The result of 
binary operations involving NA and NaN is hardware dependent (the 
propagation of NaN payload) - on some hardware, it actually works the 
way we would like - NA is returned - but on some hardware you get NaN or 
sometimes NA and sometimes NaN. Also there are C compiler optimizations 
re-ordering code

----

> `maybe_NaN_to_NA()` itself contains calls to `frollmean()` and switches depending on the behavior observed. That's a circle that's too complicated in a test environment. It should be cleaner and simpler.

`maybe_NaN_to_NA` actually doesn't call `frollmean` but depends on data.table handling of NaN-NA. We have to detect somehow how data.table behaves on that matter. Introducing new function or using `gmean` would only complicate more, that is why I used `frollmean`. It can be cleaner and simpler, as you proposed, but then we are switching off those exactness test for every environment, not only for those where compilation flags or platform specifics differs.

----

> failing on CRAN Solaris

Failed froll tests are not failing CRAN status in 1.12.0. There was no `stop` in the script if some tests failed. This is now changed and on 1.12.2 submission we will be able to see how Solaris and other platforms there handles that.

----

> This would be a 2 line change (plus comment to each) rather than the +29-14 change of the PR currently.

Yes, those two are failing now, but on other platforms we could get failures on others amended. Using `maybe_NaN_to_NA` was a safe way to test this wherever possible, otherwise treat NaN as NA.

mattdowle:
> `maybe_NaN_to_NA` actually doesn't call `frollmean`

Here's the line that does: https://github.com/Rdatatable/data.table/pull/3364/commits/38c947e0e900b2aa7d78a0cbe9543518fe24a12d#diff-83559ac036a3b9d070fc219734cf6d39L447

> Yes, those two are failing now, but on other platforms we could get failures on others amended. Using `maybe_NaN_to_NA` was a safe way to test this wherever possible, otherwise treat NaN as NA.

But the root cause is that `identical()` is being passed to `test()` in many of the froll tests. That's too strict. `test()` is designed to compare x to y allowing for tolerance using `all.equal()`. `all.equal()` considers NA and NaN to be the same.  When there's a difference between x and y, test() prints the outputs that differ and the output of all.equal (which can be a helpful message sometimes) and we see that in the logs.  When you call identical() yourself and pass that to test(), that's overriding the intention of test(): it's too strict and we lose the diagnostics.

$1.6 of R-exts contains this bullet point at the end : 

> Only test the accuracy of results if you have done a formal error analysis. Things such as checking that probabilities numerically sum to one are silly: numerical tests should always have a tolerance. That the tests on your platform achieve a particular tolerance says little about other platforms. R is configured by default to make use of long doubles where available, but they may not be available or be too slow for routine use. Most R platforms use ‘ix86’ or ‘x86_64’ CPUs: these may use extended precision registers on some but not all of their FPU instructions. Thus the achieved precision can depend on the compiler version and optimization flags—our experience is that 32-bit builds tend to be less precise than 64-bit ones. But not all platforms use those CPUs, and not all83 which use them configure them to allow the use of extended precision. In particular, ARM CPUs do not (currently) have extended precision nor long doubles, and long double was 64-bit on HP/PA Linux.
If you must try to establish a tolerance empirically, configure and build R with --disable-long-double and use appropriate compiler flags (such as -ffloat-store and -fexcess-precision=standard for gcc, depending on the CPU type84) to mitigate the effects of extended-precision calculations.

We are not supposed to use identical(), see what breaks on CRAN and then change it according to platform differences. That's a lot of work for us and impacts CRAN maintainers too. We're supposed to use `all.equal()` in the first place, as `test()` does built-in.

Thanks for relaxing these tests. Much simpler and clearer now. Will merge. Let's come back to the other tests that use identical().

jangorecki:
> Here's the line that does: [38c947e#diff-83559ac036a3b9d070fc219734cf6d39L447](https://github.com/Rdatatable/data.table/commit/38c947e0e900b2aa7d78a0cbe9543518fe24a12d#diff-83559ac036a3b9d070fc219734cf6d39L447)

I explicitly put that outside of function call to use that switch as constant defined based on base R compilation flag and data.table compilation flag, and/or platform.

> `identical()`

I explain in new issue: https://github.com/Rdatatable/data.table/issues/3371

> see what breaks on CRAN

As mentioned above. Nothing as those tests were not raising exception on fails, none of current CRAN failures is because of that. Do we have access to `tests/froll.Rout` from CRAN checks? then I could see in output if any of tests failed.

PR #3370, opened 2019-02-07
Minor fix for the `.checkTypos` error message

mnazarov:
The (non-found) object name appears twice in the `.checkTypos` error message:
```
Error in .checkTypos(e, names(x)) : 
  Object 'exclude2' not found. Perhaps you intended exclude2exclude
```
According to git blame, this was introduced in https://github.com/Rdatatable/data.table/commit/bb28b25916c16986802ab2780e4da4bdc21c9856 after removing `sprintf`


mattdowle:
Thanks, @mnazarov!

PR #3376, opened 2019-02-08
fix relative-URLs for Rd-to-vignette

russHyde:
In the `data.table.pdf` that is presently on CRAN (v1.12.0) all links from the
pdf to the vignettes incite a 404 error (eg, at bottom of p14)

The `href`s that point to the vignettes from some `.Rd` files were incorrect.

After the fix, 
`https://cran.r-project.org/web/packages/data.table/data.table.pdf`
should now point to
`..../packages/data.table/vignettes/datatable-faq.html`
for example, where it previously pointed to
`..../packages/doc/datatable-faq.html`.

### DIFF
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -31,7 +31,7 @@ set(x, i = NULL, j, value)
 \item{value}{ A list of replacement values to assign by reference to \code{x[i, j]}. }
 }
 \details{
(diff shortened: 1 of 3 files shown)


mattdowle:
@russHyde Thanks!

PR #3281, opened 2019-01-13
Use setattr() instead of structure()

MichaelChirico:
Closes #3280

PR #3387, opened 2019-02-11
Force copy in as.IDate

### LINKED ISSUE #3385: #3281 appears to cause as.IDate to update by reference
#3281 

Old behaviour

```r
x <- 500L
as.IDate(x)
#> [1] "1971-05-16"
x
#> 500
```

New behaviour:

```r
x <- 500L
as.IDate(x)
#> [1] "1971-05-16"
x
#> [1] "1971-05-16"  # oops
```

I think this should be reverted.

MichaelChirico:
Closes #3385 

Also caught that `setattr` was returning invisibly; wrapping in `()` fixes this.


mattdowle:
As per @HughParsonage's comment (https://github.com/Rdatatable/data.table/issues/3385#issuecomment-462216420) shouldn't we just use `attr<-`.  R is better now. For example if R knows that the input is not referenced, newer R versions might even set the attribute directly with no copy.  Which is better than forcing a copy always, as this PR currently does.

MichaelChirico:
OK. Do we want to keep something in `.onLoad` to try and monitor R behavior in case something changes? Or we're happy devolving that to upstream testing in `r-core`?

mattdowle:
I would lean towards double-checking that `attr<-` behaves as Hugh said, and then go with that. A check on R copy behaviour of `attr<-` would be good too and that could go in test.test.datable( .. other.Rraw .. ) if you want to. So that CRAN tests don't fail if it changes, but so that we know.

MichaelChirico:
> double-checking that `attr<-` behaves as Hugh said

As for this, I'll refer to https://github.com/Rdatatable/data.table/pull/3281#issuecomment-461079626

The benchmark/memory trace is on `class<-`, which is the appropriate `base` function to use here.

MichaelChirico:
> I wonder if `attr<-` was improved before or after R 3.1.0.

This is a good point, I'm glad I caught your edit... I don't have `docker` on my machine at the moment so I can't use @jangorecki's image to test... Jan, if you see this before me, could you try and confirm? Running the code linked above should suffice.

jangorecki:
> Jan, if you see this before me, could you try and confirm?

don't have docker too, you can fork data.table on gitlab and push there to branch and tests should kick in.

MichaelChirico:
OK, just went ahead and installed docker, I'll need it sooner or later anyway.

Two problems with replicating:

> `microbenchmark` has an (unstated) dependency on R 3.2.0 - https://github.com/joshuaulrich/microbenchmark/issues/11

> R was not compiled with support for memory profiling

Is the latter something that we can/should fix in the Docker image @jangorecki ?

jangorecki:
> > R was not compiled with support for memory profiling
> 
> Is the latter something that we can/should fix in the Docker image @jangorecki ?

Could be but isn't that better just to download R 3.1.0 sources and build it so you don't have to use docker? use the steps from docker r-3.1.0 and should goes smoothly. Unless you are debugging CI pipeline which uses docker it will be easier on locally build and installed R.

R/IDateTime.R
```diff
@@ -12,30 +12,30 @@ as.IDate.default <- function(x, ..., tz = attr(x, "tzone")) {
 
 as.IDate.numeric <- function(x, origin = "1970-01-01", ...) {
   if (origin=="1970-01-01") {
-    # standard epoch
-    setattr(as.integer(x), "class", c("IDate", "Date"))
+    # standard epoch; need copy() since as.integer(integer()) doesn't copy
+    (setattr(copy(as.integer(x)), "class", c("IDate", "Date")))
```
mattdowle: If as.integer(x) does convert (normally it will) then it will return a fresh object only to be always copied a 2nd time by this copy().  Just a `class(x)<` should be fine now that R 3.1.0 much improved the copies.

mattdowle:
I removed c.factor.R and tidied things up. Ready to merge.
@MichaelChirico please check. I'll wait for your confirm before merging. GH won't let me request your review because it's your PR :)

Issue #2572, opened 2018-01-18
TODO items from tests.Rraw


MichaelChirico:
Items cut out from the end of `tests.Rraw` when this issue was first created:

- [x] Tests involving GForce functions needs to be run with optimisation level 1 and 2, so that both functions are tested all the time.
- [x] Add test for fixed bug #5519 - dcast returned error when a package imported data.table, but dint happen when "depends" on data.table. This is fixed (commit 1263 v1.9.3), but not sure how to add test. [#2571]
- [x] test and highlight in docs that negatives are fine and fast in forderv (ref R wish #15644)
- [x] bring ?setkey description up to date; #3399
- [x] tests of freading classes like Date and the verbose messages there. [#2545]
- [x] Test mid read bump of logical to character, rereads original True and False character strings faithfully.
- [x] add examples of multiple LHS (name and position) and multiple RHS to example(":=")
- [x] tests on double in add hoc by
- [x] test on -i that retain key e.g.  DT[-4] and DT[-4,sum(v),by=b] should both retain key
- [x] test on out of bound i subsets e.g. 6:10 when DT has 7 rows, and mixed negative and positive i integer is error.
- [x] test that ordered subsets when i is unkeyed now retain x's key (using is.sorted(f__))
- [x] add FAQ that  eval() is evaled in calling frame so don't need a, then update SO question of 14 March. See the test using variable name same as column name. Actually, is that true?  Need "..J".
- [x] revisit [SO answer](https://stackoverflow.com/a/10230625/3576984). Can there be a better way than the repeated `lapply(f,eval,.SD)`?
- [x] change all 1 to 1L internally (done in data.table.R, other .R to do)
- [x] check the "j is named list could be inefficient" message from verbose than Chris N showed recently to 15 May
- [x] !make sure explicitly that unnamed lists are being executed by dogroups!
- [x] Add to warning about a previous copy that class<-, levels<- can also copy whole vector.  *Any* fun<- form basically.
- [x] use looped := vs set test in example(":=") or example(setnames) to test overhead in [.data.table is tested to stay low in future.
- [x] add tests on smaller examples with NAs for 'frankv', even though can't compare to base::rank.

---

`TODO` items sprinkled throughout the rest of the file:

[...]

PR #3389, opened 2019-02-11
updating setorder.Rd

MichaelChirico:
Related to #2572 (point 3)

I may have gone overboard with the radix sort references :man_shrugging:


mattdowle:
Nice edit. Much clearer.
?setkey looks a bit out-of-date I noticed (I added a new point 4 to #2572 for that).

PR #3394, opened 2019-02-12
[WIP] froll more robust, no nested parallelism

jangorecki:
- no longer has nested parallelism - for algos that uses parallelism to compute rollmean the outer parallelism over columns/window sizes is turned off
- verbose does not turn off parallel processing anymore, all verbose messages are deferred to print outside of parallel region
- some code refactor to use #pragma omp if
- timing
- test.data.table is more flexible when passing `output` as vector of mixed constant strings and regexps


jangorecki:
looks like `%llu` format bites on windows

jangorecki:
this PR will be split into two

PR #3399, opened 2019-02-13
Modernize ?setkey; and tackle long-deprecated functions

MichaelChirico:
Towards #2572 

`git blame` on `setkey.R` shows `set2key` hasn't been changed in 3 years, and `key<-` in 4 years.

I think we can safely delete these finally, and strip all but the historical reference from the documentation; that's done in this PR


jangorecki:
why not keep it for 2 more years?

HughParsonage:
> why not keep it for 2 more years?

Indeed, why not keep it indefinitely? One thing I notice helping out people with R is that there is a very large tail of users who might use R once or twice a year and there are many users who use it only once. All they know is 'this is an R script which does X, run it if you want X'. They have no hope of recovering from errors like these. 

I think intervening with an new error is good if it avoids unexpected behaviour (i.e. a bug), but with everything else let sleeping dogs lie.

This is especially so with `key(x)<-` because the warning message says 'You can safely ignore this warning...'. We could remove that but I think it's pretty costless to just leave it.

jangorecki:
Also base R is known for its backward compatibility. Think about blog posts from where people will just try to copy paste code. SO we could easily find and update, but there is not really need for that. Warnings are OK because they will still produce final outcome.

MichaelChirico:
I think we should move `key<-` to an error first, yes; will redact and put a deprecation error.

But for `set2key` family, from `?Deprecated`:

> When an object is **about to be removed from R** it is first deprecated and should include a call to `.Deprecated`.

(emphasis mine)

Deprecation means eventually that functionality will be removed completely. I think 3 years is plenty "eventual".

jangorecki:
good to put `.Deprecated` call there, but there is no real reason to remove it after 3 years.

mattdowle:
The past communication/guidance in NEWS is best to drive this rather than dates of commits.

From v1.11.0 on CRAN May 2018 (less than a year ago) : 
> 11. As `set2key()` and `key2()` have been warning since v1.9.8 (Nov 2016), their warnings have now been upgraded to errors. Note that when they were introduced in version 1.9.4 (Oct 2014) they were marked as 'experimental' in NEWS item 4. They will be removed in one year.
    Was warning : 
      set2key() will be deprecated in the next relase. Please use setindex() instead.
    Now error:
      set2key() is now deprecated. Please use setindex() instead.

One year from then would be May 2019.

mattdowle:
@MichaelChirico  Just been going through `?setkey`.  Yep : this was sorely needed!  Good call.

PR #3409, opened 2019-02-17
Add tests of forder NA/NaN behavior

MichaelChirico:
Also, all the `TODO`s here are now pulled into #2572, so removing them from the tests file (where they're less visible)

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -2391,7 +2391,7 @@ test(863, after < before+0.5)
 # rbindlist should look for the first non-empty data.table - New changes (from Arun). Explanation below:
 # Even if data.table is empty, as long as there are column names, they should be considered.
 # Ex: What if all data.tables are empty? What'll be the column name then?
-# If there are no names, then the first non-empty set of names will be allocated. I think this is the way to do it.. TODO: Should write to Matt about it.
+# If there are no names, then the first non-empty set of names will be allocated.
 test(864.1, rbindlist(list(data.table(foo=logical(0),bar=logical(0)), DT<-data.table(baz=letters[1:3],qux=4:6))), setnames(DT, c("foo", "bar")))
 test(864.2, rbindlist(list(list(logical(0),logical(0)), DT<-data.table(baz=letters[1:3],qux=4:6))), DT)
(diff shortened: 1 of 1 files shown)


MichaelChirico:
> they weren't just `dput(forder())` right?

They were :grimacing:

jangorecki:
So just please examine if returned values are expected ones. Such methodology of adding tests does tests only that returned values will not change over time but not if they are correct. It happens few times to me that I pushed incorrect tests like that.

MichaelChirico:
Thanks @jangorecki. I actually think current behavior is inconsistent:

```
x = data.table(r = c(6, 4, 2, NA, 1, NaN, 5, NaN, 9, 10, NA))

x[order(r, na.last = TRUE)]
#         r
#     <num>
#  1:     1
#  2:     2
#  3:     4
#  4:     5
#  5:     6
#  6:     9
#  7:    10
#  8:   NaN
#  9:   NaN
# 10:    NA
# 11:    NA
x[order(r, na.last = FALSE)]
#         r
#     <num>
#  1:    NA
#  2:    NA
#  3:   NaN
#  4:   NaN
#  5:     1
#  6:     2
#  7:     4
#  8:     5
#  9:     6
# 10:     9
# 11:    10
x[order(-r, na.last = TRUE)]
#         r
#     <num>
#  1:    10
#  2:     9
#  3:     6
#  4:     5
#  5:     4
#  6:     2
#  7:     1
#  8:   NaN
#  9:   NaN
# 10:    NA
# 11:    NA
x[order(-r, na.last = FALSE)]
#         r
#     <num>
#  1:    NA
#  2:    NA
#  3:   NaN
#  4:   NaN
#  5:    10
#  6:     9
#  7:     6
#  8:     5
#  9:     4
# 10:     2
# 11:     1
```

jangorecki:
I renamed to WIP so it won't be yet merged, lets first agree on expected output.

mattdowle:
Great discussion.  I think of the ordering of NA and NaN as different to the ordering of the finite numbers: if `na.last` is provided then `NA` and `NaN` are put at the beginning or the end in a fixed order such that `NA` is always at the end with `NaN` before the NA.
I installed both v1.9.8 and v1.11.8 and checked the results checked in this PR match the results then.  All good.  So for the reason of consistency over time, these tests look good to me.

Issue #605, opened 2014-06-08
[R-Forge #2392] Add length postfix to list column output


arunsrinivasan:
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2392&group_id=240&atid=978)

Trailing comma is easily missed.  Something like :

  A                 B      C
foo         1,2,3,4     42
bar  3,4,5,6,[12]     43

or put the postfix on all items for consistency.

jangorecki:
It does work already but for a list elements inside a list column
```r
DT = data.table(A=c("foo","bar"), B=list(c(1,2,3,4), c(3,4,5,6,7,8,9,10,11,12)), C=c(42,43))
DT
#        A               B     C
#   <char>          <list> <num>
#1:    foo         1,2,3,4    42
#2:    bar 3,4,5,6,7,8,...    43
DT[, B := lapply(B, as.list)]
DT
#        A          B     C
#   <char>     <list> <num>
#1:    foo  <list[4]>    42
#2:    bar <list[10]>    43
```
Do we want to have length printed in the first case as well?
Should the print be changed as well?
```r
#        B
#    <list>
#  <num[4]>
# <num[10]>

PR #3414, opened 2019-02-17
format_col/format_list_item printing generics for customization

MichaelChirico:
Successor to @mllg's original PR (I don't have push access there so just continuing here) #3338.

I've built upon @mllg's initial idea to allow two dimensions of print output customization -- one at the column level and one at the row level (for `list` columns). The latter was covered in #3338.

Currently test 1293 fails:

```
# Bug #5435 - print.data.table and digits option:
DT <- structure(list(fisyr = 1995:1996, er = list(c(1, 3), c(1, 3)),
    eg = c(0.0197315833926059, 0.0197315833926059), esal = list(
        c(2329.89763779528, 2423.6811023622), c(2263.07456978967,
        2354.16826003824)), fr = list(c(4, 4), c(4, 4)), fg =
c(0.039310363070415,
    0.039310363070415), fsal = list(c(2520.85433070866, 2520.85433070866
    ), c(2448.55449330784, 2448.55449330784)), mr = list(c(5,
    30), c(5, 30)), mg = c(0.0197779376457164, 0.0197779376457164
    ), msal = list(c(2571.70078740157, 4215.73622047244),
c(2497.94263862333,
    4094.82600382409))), .Names = c("fisyr", "er", "eg", "esal",
"fr", "fg", "fsal", "mr", "mg", "msal"), class = c("data.table",
"data.frame"), row.names = c(NA, -2L))

ans1 = capture.output(print(DT, digits=4, row.names=FALSE))
ans2 = c(" fisyr  er      eg      esal  fr      fg      fsal    mr      mg      msal",
         "  1995 1,3 0.01973 2330,2424 4,4 0.03931 2521,2521  5,30 0.01978 2572,4216",
         "  1996 1,3 0.01973 2263,2354 4,4 0.03931 2449,2449  5,30 0.01978 2498,4095")
test(1293, ans1, ans2)
```

IIUC the problem is that `digits` is passed on through `...` `format_list_item` and this messes with the output since numerics in lists have `digits` applied as well.

In a sense this is a more consistent interpretation, so I didn't bother keeping the status quo behavior, but don't want to overwrite the old test behavior without consulting here first.


mattdowle:
Can you explain why not `toString` please?  I don't follow the comments that followed https://github.com/Rdatatable/data.table/pull/3338#issuecomment-459891351.  R 3.1.0 has `toString` so no backporting is needed.

MichaelChirico:
Its just shy of being flexible enough; see your comment here:

https://github.com/Rdatatable/data.table/pull/2562#issuecomment-356795666

mattdowle:
I sort of see. Just shy meaning the space in `collapse=", "` in toString.default? This PR is defining our own methods though.  Can't they be `toString` methods rather than our own new class?  We will rarely be falling back to `toString.default` and if we do, we could always mask `toString.default` with our own that uses `collapse=","`.

MichaelChirico:
OK. For some reason I thought `toString` was somehow recent but apparently it's quite old:

https://github.com/wch/r-source/blame/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/library/base/R/toString.R

I'm still not quite convinced though, as I'm still hung up on the following:

This PR introduces two new generics: `format_col` and `format_list_item`. Are we aiming to replace both of these by `toString`? Seems like it reduces flexibility. Otherwise, we keep just one and then have e.g. `format_col` and `toString` for column/list item formatting?

MichaelChirico:
@mattdowle merged this to master and overwrote changes built in https://github.com/Rdatatable/data.table/pull/3500 with what was originally proposed here. I think the approach here is cleaner. Also subsumed the NEWS note there into the item for this PR.

mllg:
Just checking back ... is this PR dead? 

FWIW, here is the approach of custom printers for tibbles: https://pillar.r-lib.org/reference/pillar_shaft.html

randomgambit:
happy to do some testing when ready. This is a must have

MichaelChirico:
Related: https://github.com/Rdatatable/data.table/issues/605

MichaelChirico:
OK finally merged & updated this PR against current master. We fail one test:

```
DT2 = data.table(
  Dcol = as.Date('2016-01-01') + 0:2,
  Pcol = as.POSIXct('2016-01-01 01:00:00', tz = 'UTC') + 86400L*(0:2),
  gcol = TRUE, Icol = as.IDate(16801) + 0:2,
  ucol = `class<-`(1:3, 'asdf')
)
test(1610.2, capture.output(print(DT2, class=TRUE)),
     c("         Dcol                Pcol   gcol       Icol   ucol",
       "       <Date>              <POSc> <lgcl>     <IDat> <asdf>",
       "1: 2016-01-01 2016-01-01 01:00:00   TRUE 2016-01-01      1",
       "2: 2016-01-02 2016-01-02 01:00:00   TRUE 2016-01-02      2",
       "3: 2016-01-03 2016-01-03 01:00:00   TRUE 2016-01-03      3"))
```

That's because in this PR I removed the `timezone` argument... I originally did so before a release.

Now I guess we'd have to do a deprecation cycle. Or should we continue to support `timezone` argument?

Default behavior in this PR is to always print the time zone if it's there.

.ci/ci.R
```diff
@@ -5,10 +5,10 @@
 ## added ver argument to produce R version independent urls
 ## https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17420
 contrib.url <-
-function (repos, type = getOption("pkgType"), ver) 
+function (repos, type = getOption("pkgType"), ver)
```
jangorecki: Could we revert those changes? It doesn't really fit this PR and it is polluting history of those files.

MichaelChirico: am i the only one with the pre-commit hook blocking trailing whitespace / multiple trailing newlines?

i added these since i was being blocked from committing.  maybe it's something i set up on this machine only...

MichaelChirico:
TODO: address `expression` column printing:

https://github.com/Rdatatable/data.table/pull/4196#issuecomment-630094998

mattdowle:
@MichaelChirico This looks good to me to merge now. You?

R/print.data.table.R
```diff
@@ -205,6 +172,55 @@ paste_dims = function(x) {
...
+  else
+    paste0("<", class(x)[1L], paste_dims(x), ">")
+}
+
+# FR #1091 for pretty printing of character
+# TODO: maybe instead of doing "this is...", we could do "this ... test"?
+char.trunc <- function(x, trunc.char = getOption("datatable.prettyprint.char")) {
+  trunc.char = max(0L, suppressWarnings(as.integer(trunc.char[1L])), na.rm=TRUE)
+  if (!is.character(x) || trunc.char <= 0L) return(x)
+  idx = which(nchar(x) > trunc.char)
```
MichaelChirico, line 219: not crucial for this PR in particular, but wouldn't `nchar(x, 'width')` be more appropriate?

MichaelChirico:
Made a few tweaks, ready for merge now

R/print.data.table.R
mattdowle: Good idea. Would need a test too in a new PR: your Chinese multi-byte data?

PR #3415, opened 2019-02-17
minor fix to remove partial matching which warns in R 3.1.0

jangorecki:
https://github.com/Rdatatable/data.table/commit/ca43a475fdcb4abf40e5915c5611270e89898160


arunsrinivasan:
Thanks for catching that @jangorecki

PR #3421, opened 2019-02-19
foverlaps R 3.1.0 compatible again

jangorecki:


R/foverlaps.R
```diff
@@ -54,7 +54,8 @@ foverlaps <- function(x, y, by.x=if (!is.null(key(x))) key(x) else key(y), by.y=
   yval1 = y[[yintervals[1L]]]; yval2 = y[[yintervals[2L]]]
   if (!storage.mode(xval1) %chin% c("double", "integer") || !storage.mode(xval2) %chin% c("double", "integer") || is.factor(xval1) || is.factor(xval2)) # adding factors to the bunch, #2645
     stop("The last two columns in by.x should correspond to the 'start' and 'end' intervals in data.table 'x' and must be integer/numeric type.")
-  isTRUEorNA <- function(x) isTRUE(x) || is.na(x) # x is assumed to  be logical
+  isTRUEorNA <- function(x) identical(x,TRUE) || identical(x,NA)  # also ensures x must be logical length 1
+  # base::isTRUE in 3.5.2 is implemented as `is.logical(x) && length(x)==1L && !is.na(x) && x` which seems inefficient vs just identical()
```
jangorecki: minor difference is that `isTRUE` still returns true for named TRUE, while `identical` will return FALSE in such case.

R/utils.R
```diff
@@ -1,5 +1,17 @@
 # all non-exported / unused internal (utility) functions
 
+# R 3.5.0 made isTRUE inefficient :
+#   `is.logical(x) && length(x)==1L && !is.na(x) && x`
+# Before R 3.5.0, isTRUE was defined as simply:
+#   identical(TRUE,x)
+# which was simpler. The new definition uses 8 allocations of the length-1 intermediaries. Probably doesn't matter but might sometimes.
+# It was changed in R so that isTRUE(c(a=TRUE)) returned TRUE: https://github.com/wch/r-source/commit/828997ac6ecfb73aaa0aae9d1d0584a4ffc50881#diff-b41e3f9f1d389bb6f7a842cd5a3308b8
+# Which we don't need. We prefer simplicity and efficiency of identical().
+isTRUE        <- function(x) identical(x,TRUE)  # just as base::isTRUE before R 3.5.0. Masks base::isTRUE for internal use.
```
jangorecki: There are many place where `isTRUE` has been used in before. @mattdowle did you review that none of them should handle named TRUE?
Using that expose us to contribution that will not be aware of this difference vs base R behaviour so we should always keep that in mind when reviewing code that uses `isTRUE`.
I find those functions very convenient but safer would be to prefix them like `.isTRUE` or `isTRUE_`.

MichaelChirico: > The new definition uses 8 allocations of the length-1 intermediaries

Is that true? I thought lazy-eval meant that each component of the `&&` will only evaluate sequentially, so in the `!is.logical(x)` case this will abort immediately and the rest will be unevaluated/allocated, which should capture the vast majority of mistaken argument declarations. Certainly `is.logical(x) && length(x) == 1L` capture almost everything...

If we really prefer the baremetal `identical(x, TRUE)`, I agree Jan's `.isTRUE` is preferred.

jangorecki: I think Matt meant worst case scenario here

mattdowle: Worst case is most common case, right? i.e. passing TRUE or FALSE to isTRUE and the whole chain is evaluated.

jangorecki: I saw named TRUE discussion once on R-devel mailing list. I doubt we would have a test for such an edge case like named TRUE instead of a TRUE, so lack of failure on R 3.1.0 is not best to rely on. Lets see if revdeps will complain.

MichaelChirico: For named case, can just use `as.logical` internally?

```
x = c(a = TRUE)
microbenchmark::microbenchmark(times = 10000L,
                               as.logical(TRUE),
                               as.logical(x))
Unit: nanoseconds
             expr min  lq     mean median  uq   max neval
 as.logical(TRUE)  48  63  70.7291     68  70  9280 10000
    as.logical(x) 135 150 167.8907    152 158 10635 10000
```

mattdowle: If named TRUE ever does come up, perhaps we can revisit this at that stage?
Or if anyone can provide a case where named TRUE arises or is remotely useful for us in data.table where we're checking valid and switching on arguments.

MichaelChirico: My most common case for using named objects is probably lookup tables... could imagine keeping a table like `append = c(grp1 = TRUE, grp2 = TRUE, grp3 = FALSE)` then `DT[ , fwrite(.SD, append = append[.BY$grp], by = grp]`.

A glance over the arguments to which this applies suggests this is not particularly likely TBH. 

OTOH, I also don't think it's particularly common to use any of these arguments iteratively 10000s of times where the `identical` vs. `R{3.4.4}::base::isTRUE` difference is particularly noticeable.

Leaning towards kicking the can down the road and seeing if it bites for anyone...

mattdowle: If we support named TRUE and start relying on it with tests, then we should backport it to R 3.1.0

mattdowle: Interestingly (4 runs to see max change) : 
```
> microbenchmark(base::isTRUE(x), isTRUE(x), identical(x,TRUE), times=10000)
Unit: nanoseconds
               expr  min   lq      mean median     uq     max neval cld
    base::isTRUE(x) 3646 3793 4142.9255   3869 3953.5 2154817 10000   c
          isTRUE(x)  399  425  459.2389    437  457.0   10711 10000 a  
 identical(x, TRUE)  742  807  877.6774    837  914.0   12432 10000  b 
> microbenchmark(base::isTRUE(x), isTRUE(x), identical(x,TRUE), times=10000)
Unit: nanoseconds
               expr  min   lq      mean median   uq     max neval cld
    base::isTRUE(x) 3622 3761 4414.0061   3896 4858 2121984 10000   c
          isTRUE(x)  401  424  485.8224    435  455   10853 10000 a  
 identical(x, TRUE)  730  799  973.3035    828  899   11012 10000  b 
> microbenchmark(base::isTRUE(x), isTRUE(x), identical(x,TRUE), times=10000)
Unit: nanoseconds
               expr  min   lq      mean median   uq     max neval cld
    base::isTRUE(x) 3647 3789 3906.9718   3865 3946   20719 10000   c
          isTRUE(x)  402  425  458.4353    438  456   41668 10000 a  
 identical(x, TRUE)  740  810 1078.2153    842  911 2011141 10000  b 
> microbenchmark(base::isTRUE(x), isTRUE(x), identical(x,TRUE), times=10000)
Unit: nanoseconds
               expr  min   lq      mean median   uq     max neval cld
    base::isTRUE(x) 3627 3765 4427.1266   3891 4859 2044037 10000   c
          isTRUE(x)  399  424  489.0931    436  456   11610 10000 a  
 identical(x, TRUE)  739  798  979.1621    826  904   11365 10000  b 
```

Note how the max can be high for 1 and 3.  Seems like the `base::` prefix.   So `identical()` isn't fastest after all.  Ok so `isTRUE` in R 3.5.0 is fine then.  Maybe it's to do with the `.Internal()` call that `identical()` has and it itself has quite a few arguments. Compared to the primitives of `&&` and `!` in `isTRUE` of R 3.5.0.

Now that `utils.R` has our `isTRUE` etc, it will be easy to change back...

jangorecki:
```
git checkout 426b2f6680b85843e1e4d622992df8a48667e609
cc(F)
is.factor(data.table(a=c("a","b"), stringsAsFactors=TRUE)$a)
#[1] TRUE
is.factor(data.table(a=c("a","b"), stringsAsFactors=c("a"=TRUE))$a)
[1] FALSE
```
master branch is fine

mattdowle: stringsAsFactors=c("a"=TRUE) is weird. Who or why would do that?! That didn't work before R 3.5.0 anyway.
But yes I gave up and changed master to accept named TRUE and backported it to R 3.1.0 too, commit be8c7f6 links to this PR. So tests can be added now if you like.

PR #3425, opened 2019-02-21
eval on quoted expression of := will handle dot alias

jangorecki:
solves https://stackoverflow.com/questions/41228076/using-data-tables-shortcut-in-quoted-expressions
@franknarf1


jangorecki:
@mattdowle done

Issue #3400, opened 2019-02-13
[bug] fread handling embedded NUL characters


mdavy86:
This issue is similar to several previous issues;

* #2435
* #2485
* #1895
* [stack overflow](https://stackoverflow.com/search?q=%5Bdata.table%5D+NUL)

loading a file containing `NUL` ASCII character (in bytes `as.raw(0)`), except that I have a minimal reproducible example which appears to cause a segfault on line [fread.R@146](https://github.com/Rdatatable/data.table/blob/079968bf9e498f1f2efbdd1239a254850b042663/R/fread.R#L146)

This example is based on simulation software output where very rarely there can be `NUL` characters in the body of the file (issue #2485 has already resolved `NUL` characters at the end of a file). It appears `NUL` characters at the beginning of a file are acceptable as well.

The header field is key=value pairs, and the data field is to be __read__ into a data.table. In the example, `NUL` characters has been inserted into the body, you only need one to cause an error which cannot be caught with error handling.

```R
library(data.table)
## environment
sessionInfo()
## example #1
n <- 1
bytes <- c(charToRaw("a=b\nA  B  C\n1  2  3\n"), rep(as.raw(0), n), charToRaw("4  5  6\n"))
writeBin(bytes, "test.txt")
## fread hangs
try(fread("test.txt", skip=1, header=TRUE, verbose=TRUE))
```

Verbose tracelog is provided using a file `test1.R`;

```bash
## bash here doc...
cat - > test1.R << 'EOF'
library(data.table)
## environment
sessionInfo()
## example #1
n <- 1
bytes <- c(charToRaw("a=b\nA  B  C\n1  2  3\n"), rep(as.raw(0), n), charToRaw("4  5  6\n"))
writeBin(bytes, "test.txt")
## fread hangs
try(fread("test.txt", skip=1, header=TRUE, verbose=TRUE))
EOF
```

Running `test1.R`;

```bash
$ Rscript test1.R 
R version 3.5.2 (2018-12-20)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: CentOS Linux 7 (Core)

Matrix products: default
BLAS: /software/statistical/R-3.5.2/lib64/R/lib/libRblas.so
LAPACK: /software/statistical/R-3.5.2/lib64/R/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_NZ.UTF-8       LC_NUMERIC=C
[...]

PR #3433, opened 2019-02-27
Improve handling NULL characters

kiwiroy:
AFAICT the following is making for an *"infinite"* loop (nrowLimit is very large by default).

https://github.com/Rdatatable/data.table/blob/0dfb91d91e0b37f8ea235985b5724f1805b51b53/src/fread.c#L2151

Adding `nrowLimit = myNrow` to the branch appears to be a fix to force it to be the last line.

## References

#3400


mattdowle:
Thanks @kiwiroy! I've invited you to be project member.

PR #3435, opened 2019-02-28
default reduced to half of all logical CPUs

### LINKED ISSUE #3298: Better use number of physical cores instead of logical cores by default
It is great that the latest release (v1.12.0) add more support for multithreading in subsetting and restore-after-fork behavior by default. The current default threads is `omp_get_max_threads()` which tries to achieve maximal performance on a *completely task-free server*. In practice, however, this default number of threads makes data.table multithreading cost much much more time in computing on a slightly busier server.

In my case, I'm using a server of 40 physical cores (`RhpcBLASctl::get_num_cores()`) and 80 threads (`RhpcBLASctl::get_num_procs()`). In status of no other running tasks, the default behavior of  data.table is to occupy all 80 threads which works well. But if any other running tasks occupies a small number of threads (e.g. 5-10), data.table multithreading would cost much longer time than no-multithreading at all since those CPUs are blocked.

I'm not sure if it makes sense not to occupy all threads by default. I'd recommend occupying all physical cores by default which seems more practical and may achieve better performance when other tasks are using CPUs or hyperthreading is enabled.

### LINKED ISSUE #3395: Performance drop with 1.12.0 (Selection + assignment)
I'm not sure if it's the same issue as #3330 because I haven't set the DTthreads parameters to 1.
But I could imagine that the default value is 1. 
This list of operations take less than 1 second  to be executed with data.table  v1.11.8.
With data.table v1.12.0 it's take more than 4 seconds !

```
>   myDataTable[colJ != "production", colK:=na.omit(colK), by=.(colA)]
>   myDataTable[colJ != "production", colH:=na.omit(colH), by=.(colA)]
>   myDataTable[colJ != "production", colL:=na.omit(colL), by=.(colA)]
>   myDataTable[colJ == "code",colM:=colE, by=.(colA)]
>   myDataTable[,colM:=na.omit(colM), by=.(colA)]
>   myDataTable[is.na(colM), colM:=""]
>   myDataTable[colJ == "test",colN:=colE, by=.(colA)]
>   myDataTable[,colN:=na.omit(colN), by=.(colA)]
```

And there is only 4 rows in my data table : 

```
print(myDataTable)

     colA colB colC colD colE colF colG colH colI colJ colK colL colM colN
1 	 text text text text text text text text text text text text text text
2 	 text text text text text text text text text text text text text text
3 	 text text text text text text text text text text text text text text
4  	 text text text text text text text text text text text text text text
```                   
(You can replace "text" by whatever) 

We see this problem only in our production environment and not in other environment. This production server is really busy so it can explain why it's so long. But the test with  v1.11.8 was done in the same environment without performance issue (We done this test many times with both version).

We run R in a Docker container with the system Ubuntu 16.04.

```
> sessionInfo()
R version 3.4.4 (2018-03-15)
Running under: Ubuntu 16.04.5 LTS

Matrix products: default

locale:

attached base packages:

[1] compiler_3.4.4
```

I hope this will be fixed in a future data.table release because we will be blocked in v1.11.8 until this issue is not resolved.
Thank you for your support.

mattdowle:
Closes #3298 
Closes #3395 

I investigated various methods for determining number of cores.  The one referred to in #3395 (`RhpcBLASctl::get_num_cores()`) seems like the only one that actually works; `parallel::detectCores()` doesn't work for me.  But `RhpcBLASctl::get_num_cores()` reads the system files on linux to do it so could possibly be slow, and would incur a new dependency.
So I figured why not just keep it very simple and default to 50% of logical CPUs. This leaves plenty of room for other processes. If the user wants to use more or less they can set OMP_NUM_THREADS up to the number of logical CPUs, or call `setDTthreads()`.


NEWS.md
```diff
@@ -44,15 +44,11 @@
...
+7. `getDTthreads()` respected the `OMP_NUM_THREADS` environment variable but not `OMP_THREAD_LIMIT`, [#3300](https://github.com/Rdatatable/data.table/issues/3300). Note that these environment variables should be set before the R session starts; using the R command `Sys.setenv()` to set them is too late because the OpenMP runtime is already running by then (use `setDTthreads()` instead). Further, the default number of threads is reduced to half the number of logical CPUs. The previous default of using all logical CPUs was reported to cause significant slow downs when other non-trivial processes are also running, [#3395](https://github.com/Rdatatable/data.table/issues/3395), [#3298](https://github.com/Rdatatable/data.table/issues/3298). The number of logical CPUs is determined by `omp_get_num_procs()`. This value has been added to the output of `getDTthreads(verbose=TRUE)`.
```
jangorecki: We need more verbose message that all resources might not be utilised, ideally on top of NEWS. AFAIU there are much more users having no problems using all logical cores. Maybe message even in package startup message? Something like
> data.table is using X cores while Y logical cores are available, for details see `?setDTthreads`.

man/openmp-utils.Rd
```diff
@@ -3,24 +3,26 @@
...
 }
 \usage{
-setDTthreads(threads = 0, restore_after_fork = NULL)
-getDTthreads(verbose = getOption("datatable.verbose", FALSE))
+  setDTthreads(threads = 0, restore_after_fork = NULL)
+  getDTthreads(verbose = getOption("datatable.verbose", FALSE))
 }
 \arguments{
-  \item{threads}{ An integer >= 0. Default 0 means use all CPU available and leave the operating system to multi task. }
+  \item{threads}{ An integer >= 0. Default and recommended 0 means to use half of the logical CPUs available so as to leave plenty of resources for other processes. }
```
jangorecki: Would be useful to provide user option to use all cores without forcing them to figure out how many logical cores they have. Maybe `threads=Inf`?

mattdowle:
I haven't noticed much difference between 4 and 8 threads on my laptop either (I have 4 cores). Our algos are mostly cache bound.
I see Jan's point there's an inconvenience aspect to needing to use setDTthreads() and know the number to give it. Also the package has to be loaded first to use it, unlike options() which can be set in .Rprofile on a server without user needing to change any code to add the setDTthreads() call.
How about a new option: options(datatable.logicalcpu.percent=0.50) to control the default. Once the option is there, Jan could see the impact it has on db-bench.

HughParsonage:
I personally think the default should be single-core since, as @renkun-ken remarked, more cores currently offers only small performance benefit on commodity hardware (with the possible exception of fread/fwrite), but can be much slower if there is other parallelism going on. In the latter case, the parallelism can't really be managed a priori; the operator would need to set the cores according to the particular apparatus being used.

This is not really a problem with data.table -- I see it in lots of places, even outside R. One thing that strikes me about the present is the high flux of packages moving from single thread to parallel operations. In isolation this can provides performance benefits, but their interaction is a bit tricky. One can imagine a scenario where a user is using data.table with its implicit parallelism together with another package X. Package X then moves from single-core to multi-core and there's a performance regression. This can be pretty tricky to track down *especially* since the NEWS in package X will say the functions should be faster.

I think fread and fwrite are special though: the performance benefit appears to be large and they are typically less likely to be used in combination with other packages. 

Thanks for inviting me to review. I've approved because I think the PR represents an improvement.

man/openmp-utils.Rd
HughParsonage: Perhaps it would be better to have `threads = NULL` be the default, keep `threads = 0` to retain the previous behaviour, and special-case `if (is.null(threads)) threads = <half of CPU> else threads`. That way, people relying on `threads = 0` to work as documented can continue to, but people just using 'the data.table default' are moved to the new default.

jangorecki:
I think it is strongly depends on data and query you are running. Single core vs multi core makes significant difference, at least for queries we are benchmarking
hist_groupby_plot-1
Machine we are using for that is (at the moment) 20 cpu, 125 GB mem. Once this PR will be merged I will run benchmark so we can see how halving cpus impacts speed.

mattdowle:
Glad we're all on the same page.  I'm surprised by Hugh's comments that more threads don't often help other than for fread/fwrite.  I suspect what's going on is a combination of recent improvements that really show on large data (db-bench) plus some aspects that now cause significant slow downs on small / iterated tasks which otherwise would are fast anyway without needing parallelism. For example, still need to look at uniqueN by group as raised here:  https://github.com/Rdatatable/data.table/issues/3395#issuecomment-463938243.  So this PR wouldn't fully close #3395, there'll be a follow up issue for that.

I was going to add the `option(datatable.logicalcpu.percent)` before merging so Jan didn't need to retest.  But the tricky thing is we don't have to read and re-read this option every time we call getDTthread() which happens quite a lot internally for each and every parallel region.  To convey that it's a default option which is read once when data.table is loaded, I'm now thinking to create 2 new environment variables instead:  `R_DATATABLE_NCPU_PERCENT` and `R_DATATABLE_NCPU`.   Such variables can be set on command line or set in R using Sys.setenv() before loading data.table.  This gives the maximum flexibility for production usage.   If both are provided then the minimum.  Startup message could print the result (as Jan suggested).

jangorecki:
what about recent @HughParsonage comment?

> Perhaps it would be better to have `threads = NULL` be the default, keep `threads = 0` to retain the previous behaviour, and special-case `if (is.null(threads)) threads = <half of CPU> else threads`. That way, people relying on `threads = 0` to work as documented can continue to, but people just using 'the data.table default' are moved to the new default.

`data.table::setDTthreads` can be added to .Rprofile so the change to existing programs can be avoided.

jangorecki:
https://gitlab.com/Rdatatable/data.table/-/jobs/169970092
```
 getDTthreads()='omp_get_num_procs()==1; R_DATATABLE_NUM_PROCS_PERCENT=="" (default 50); R_DATATABLE_NUM_THREADS==""; omp_get_thread_limit()==2147483647; omp_get_max_threads()==1; OMP_THREAD_LIMIT==""; OMP_NUM_THREADS==""; data.table is using 1 threads. This is set on startup, and by setDTthreads(). See ?setDTthreads.; RestoreAfterFork==true']. Search tests/tests.Rraw for test numbers: 1997.09, 1997.11.
```

jangorecki:
In our benchmarking environment it is definitely better to use all threads. Results below, negative `time_change` means speed up.

---

top 5 speed up:
```
data           |question          | be8c7f6| 1638c76| time_change|
|:--------------|:-----------------|-------:|-------:|-----------:|
|G1_1e8_1e2_0_0 |mean v1:v3 by id4 |   1.578|   1.367|       -0.13|
|G1_1e7_1e1_0_0 |mean v1:v3 by id4 |   0.230|   0.212|       -0.08|
|G1_1e9_2e0_0_0 |sum v1 by id1     |  12.136|  11.300|       -0.07|
|G1_1e9_1e2_0_1 |sum v1:v3 by id6  |  17.929|  16.720|       -0.07|
|G1_1e9_1e2_0_0 |sum v1:v3 by id6  |  17.811|  16.778|       -0.06|
```
top 5 slow down:
```
|data           |question              | be8c7f6| 1638c76| time_change|
|:--------------|:---------------------|-------:|-------:|-----------:|
|G1_1e8_1e2_0_1 |sum v1 by id1:id2     |   0.375|   0.483|        0.29|
|G1_1e9_1e2_0_1 |sum v1 by id1         |   2.835|   3.521|        0.24|
|G1_1e9_1e2_0_1 |sum v1 by id1:id2     |   3.493|   4.233|        0.21|
|G1_1e7_1e2_0_0 |sum v1 by id1:id2     |   0.088|   0.104|        0.18|
|G1_1e7_1e2_0_1 |sum v1 mean v3 by id3 |   0.119|   0.140|        0.18|
```

mean change by `question`
```
|question              | mean_time_change|
|:---------------------|----------------:|
|sum v1 by id1         |        0.0516667|
|sum v1 by id1:id2     |        0.1258333|
|sum v1 mean v3 by id3 |        0.0750000|
|mean v1:v3 by id4     |        0.0109091|
|sum v1:v3 by id6      |        0.0400000|
```

mean change by `data`
```
|data           | mean_time_change|
|:--------------|----------------:|
|G1_1e7_1e2_0_0 |            0.048|
|G1_1e7_1e1_0_0 |            0.018|
|G1_1e7_2e0_0_0 |            0.016|
|G1_1e7_1e2_0_1 |            0.132|
|G1_1e8_1e2_0_0 |            0.014|
|G1_1e8_1e1_0_0 |            0.060|
|G1_1e8_2e0_0_0 |            0.058|
|G1_1e8_1e2_0_1 |            0.142|
|G1_1e9_1e2_0_0 |            0.042|
|G1_1e9_1e1_0_0 |            0.072|
|G1_1e9_2e0_0_0 |            0.030|
|G1_1e9_1e2_0_1 |            0.098|
```

mattdowle:
@jangorecki Very interesting thanks!  Maybe something like 80% is optimal on the server?

MichaelChirico:
certainly for full disclosure insights we learn about tuning here should be
included in the benchmarks page

actually this is one of the toughest parts of using e.g. spark -- picking
levels of any of 30-40 hyperparameters for tuning your job without much in
the way of actionable guidance -- so anything we can offer there would also
be a big social benefit.

also, one could argue that if we have done this tuning for data.table and
not the other packages, were giving ourselves an unfair leg up...

jangorecki:
@MichaelChirico we will use all cores in benchmark, where available, same as for spark and others. I doubt using less than 100% did ever result in better performance in that environment. There are generally no other processes running in the background there.

HughParsonage:
@jangorecki Those numbers don't look significant to me. Even with 10 billion groups the differences are only fractions of a second. And this is under optimal setting for parallelism: no other processes, no nested parallelism.

I get no significant differences with a 'normal' setup (i.e. some other processes running) even with the aggregations most suited to parallelization. That is, when testing execution times of the same operation, sometimes the parallel version is faster, sometimes it's slower. Never more than 1% or so.

My argument for single-core default is that the small gains under favourable conditions are outweighed by the risk of enormous performance regressions under unforeseen unfavourable conditions.

jangorecki:
Those numbers are up to 1 billion. Differences are up to 29% - using 50% logical cores, not huge difference, for ordinary production 50% logical cores will good enough, but for benchmark we should aim for fastest possible option. Will run benchmarks using single core.

mattdowle:
@hughparsonage the columns that look like fractions of seconds are actually percentages, I think.
@jangorecki - can you scale by 100 and use a % in that output please. And where it's seconds use "s" or "secs" so there's no doubt.

jangorecki:
`be8c7f6` 20 cores (100% of logical cores) vs `1638c76` 1 core.
We can see that in best case single core is 43% slower, in worst case it is slower more than 5 times.

---

top 5 speed up - or actually top 5 smallest slow down
```
|data           |question          | be8c7f6| 1638c76|time_change_pct |
|:--------------|:-----------------|-------:|-------:|:---------------|
|G1_1e7_1e1_0_0 |mean v1:v3 by id4 |   0.230|   0.331|43.9%           |
|G1_1e7_2e0_0_0 |mean v1:v3 by id4 |   0.266|   0.393|47.7%           |
|G1_1e8_1e1_0_0 |mean v1:v3 by id4 |   2.037|   3.371|65.5%           |
|G1_1e9_1e1_0_0 |mean v1:v3 by id4 |  20.152|  33.797|67.7%           |
|G1_1e7_1e1_0_0 |sum v1 by id1     |   0.143|   0.267|86.7%           |
```

top 5 slow down
```
|data           |question              | be8c7f6| 1638c76|time_change_pct |
|:--------------|:---------------------|-------:|-------:|:---------------|
|G1_1e9_1e2_0_1 |sum v1 by id1:id2     |   3.493|  22.344|539.7%          |
|G1_1e9_1e2_0_1 |sum v1 by id1         |   2.835|  17.907|531.6%          |
|G1_1e7_1e2_0_1 |sum v1 mean v3 by id3 |   0.119|   0.749|529.4%          |
|G1_1e7_1e2_0_1 |sum v1:v3 by id6      |   0.135|   0.814|503.0%          |
|G1_1e8_1e2_0_1 |sum v1 by id1:id2     |   0.375|   2.258|502.1%          |
```

by `question`
```
|question              |mean_time_change_pct |
|:---------------------|:--------------------|
|sum v1 by id1         |221.8%               |
|sum v1 by id1:id2     |289.9%               |
|sum v1 mean v3 by id3 |366.4%               |
|mean v1:v3 by id4     |94.7%                |
|sum v1:v3 by id6      |382.0%               |
```

by `data`
```
|data           |mean_time_change_pct |
|:--------------|:--------------------|
|G1_1e7_1e2_0_0 |315.9%               |
|G1_1e7_1e1_0_0 |220.9%               |
|G1_1e7_2e0_0_0 |108.9%               |
|G1_1e7_1e2_0_1 |388.8%               |
|G1_1e8_1e2_0_0 |317.6%               |
|G1_1e8_1e1_0_0 |207.9%               |
|G1_1e8_2e0_0_0 |141.3%               |
|G1_1e8_1e2_0_1 |395.9%               |
|G1_1e9_1e2_0_0 |307.2%               |
|G1_1e9_1e1_0_0 |243.9%               |
|G1_1e9_2e0_0_0 |160.4%               |
|G1_1e9_1e2_0_1 |411.6%               |
```

PR #3451, opened 2019-03-10
Need to recalculate ustr_maxlen after strings being converted to UTF8

### LINKED ISSUE #3397: Non-ASCII key sometimes fails to work in version 1.12.0 and dev
Previously, my production environment is using data.table 1.11.4. After upgrading to version 1.12.0 (CRAN version), I find sometimes the non-ASCII strings cannot be matched... It's very very difficult to reproduce... However, finally get managed to this reproducible example...

Note, again, this only happens on Windows, only when the column being keyed is encoded in native encoding. What's strange is that I cannot reproduce it on Windows 10. It can be only reproduced on Windows 7 (succeeded on 2 computers so it should not be an issue related to my computer).

Moreover, at first, it only occurs when `options(stringsAsFactors = FALSE)` being set... However, it's unrelated in the below example code.

I'll try to debug and fix it...

```r
library(data.table)

x <- '借:Cash|借:损益类-交易费用|借:损益类-价差收入|借:损益类-公允价值变动损益|贷:资产类-公允价值变动|贷:资产类-成本'
v <- c(
  x, 
  "借:Cash|借:损益类-交易费用|借:损益类-价差收入|借:损益类-公允价值变动损益|贷:资产类-公允价值变动|贷:资产类-应计利息|贷:资产类-成本", 
  "借:Cash|借:损益类-利息收入|贷:资产类-成本", "借:Cash|借:损益类-利息收入|贷:资产类-成本|贷:资产类-应计利息", 
  "借:Cash|借:损益类-利息收入|贷:资产类-成本|贷:资产类-应计利息|贷:资产类-折溢价", 
  "借:Cash|借:损益类-利息收入|贷:资产类-成本|贷:资产类-应计利息|借:资产类-折溢价", 
  "借:Cash|借:损益类-利息收入|贷:资产类-应计利息", "借:Cash|借:损益类-利息收入|贷:资产类-应计利息|贷:资产类-成本", 
  "借:Cash|借:损益类-利息收入|借:权益类-资本公积|贷:资产类-公允价值变动|贷:资产类-应计利息|贷:资产类-成本"
)

tmp <- data.table(a = v, b = 1, key = 'a')
print(tmp[J(x), b])  # returns NA on 1.12.0 and dev; returns 1 on 1.11.4
print(tmp[, b[v == x]]) # always return 1
```

shrektan:
Closes #3397. 

`ustr_maxlen` should be recalculated if there's any strings that need to be converted to UTF-8. Otherwise, it leads to wrong orders in certain cases.  This PR fixes the example provided in #3397. 

### TODO

- [x] ~~find a test case that based on `latin-1 `- so that it can be checked on all platforms~~
     No it's not possible (at least according to my experiments) to write a working case based purely on latin1 encoding. 
    Anyway, I added a test based on my original example, which should work on both Windows 7 and Windows 10 (tried on 3 machines - win7, win10, default locale English and Chinese - so it should be correct)
- [x] update NEWS


inst/tests/tests.Rraw
```diff
@@ -13626,6 +13626,23 @@ dx = data.table(id = 1L, key = "id")
...
+# test the non-ASCII strings are correctly sorted on Windows, #3397
+if (.Platform$OS.type == 'windows') local({
+  lc_collate <- Sys.getlocale(c('LC_COLLATE'))
+  lc_ctype <- Sys.getlocale(c('LC_CTYPE'))
+  Sys.setlocale('LC_COLLATE', "Chinese (Simplified)_China.936")
+  Sys.setlocale('LC_CTYPE', "Chinese (Simplified)_China.936")
+  on.exit({
+    Sys.setlocale('LC_COLLATE', lc_collate)
+    Sys.setlocale('LC_CTYPE', lc_ctype)
+  }, add = TRUE)
```
jangorecki: Not sure if `add=TRUE` is appropriate here. Locale setting is properly reverted to old when `local()` finishes. `add=TRUE` make sense if there would be any other preceding `on.exit` calls inside `local`.

shrektan: It's the same with or without `add=TRUE` . It's just a good practice (in  my opinion)  to always add this to avoid overriding the previous `on.exist()` logic.

Issue #3427, opened 2019-02-23
rbind with data.table having more than 254 columns reorders column names


wsrfad:
```
dt <- data.table(
     	sample = 1
     )    
    i = 1
    while(i <= 254) {
    	col <- paste("x", i, sep = "_")
    	dt[[col]] = i
    	i = (i + 1)
    }
    
    combined_dt <- rbind(dt, dt)
    print(names(combined_dt)) # Columns get reordered
    
    combined_dt <- rbindlist(list(dt, dt))
    print(names(combined_dt)) # Column names remain same
```

Henrik-P:
Related to [rbindlist use.names=TRUE returns random column order when ncol>255](https://github.com/Rdatatable/data.table/issues/3373)?

MichaelChirico:
yes, duplicate

PR #3455, opened 2019-03-13
rbindlist reworked

### LINKED ISSUE #524: [R-Forge #5690] rbindlist/rbind should be able to recycle wherever possible
Submitted by: Arun ; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5690&group_id=240&atid=978)

I'm quite satisfied with the implementation of use.names and fill arguments and moving everything to C, with the recent commit (1266 v1.9.3). However, as written here: http://lists.r-forge.r-project.org/pipermail/datatable-help/2014-May/002552.html I'd love for the recycling feature to be present as well, as this'd allow for replacing the "binding" inside `[.data.table` to use `rbindlist` with `use.names = TRUE`. 

This'll also solve http://lists.r-forge.r-project.org/pipermail/datatable-help/2014-April/002480.html issue (if we use `rbindlist` internally in `[.data.table` with use.names=TRUE (or keeping it as an additional argument).

### LINKED ISSUE #1302: rbindlist supports for non-atomic columns
The same way that `rbindlist` makes `c()`-like process on the atomic columns (column _"a"_ in below example) it would be nice to have non-atomic columns (column _"z"_ in below example) combined using `list()`-like process.

``` r
library(data.table)
l = list(
    list(a=1, z=list()),
    list(a=2, z=list("m"))
)
rbindlist(l)
# Error in rbindlist(l) : 
#   Column 2 of item 1 is length 0, inconsistent with first column of that item which is length 1. rbind/rbindlist doesn't recycle as it already expects each item to be a uniform list, data.frame or data.table

# expected
str(data.table(a=c(1,2), z=list(list(),list("m"))))
# Classes ‘data.table’ and 'data.frame':    2 obs. of  2 variables:
#  $ a: num  1 2
#  $ z:List of 2
#   ..$ : list()
#   ..$ :List of 1
#   .. ..$ : chr "m"
```

It appears currently it can handle list but not in a generic manner:

``` r
# this works - but *z* losses list class in each row
l = list(
    list(a=1, z=list("z")),
    list(a=2, z=list(c("a","b")))
)
str(rbindlist(l))
# Classes ‘data.table’ and 'data.frame':    2 obs. of  2 variables:
#  $ a: num  1 2
#  $ z:List of 2
#   ..$ : chr "z"
#   ..$ : chr  "a" "b"

# this not works
l = list(
    list(a=1, z=list("z",1)),
    list(a=2, z=list(c("a","b")))
)
str(rbindlist(l))
# Error in rbindlist(l) : 
#   Column 2 of item 1 is length 2, inconsistent with first column of that item which is length 1. rbind/rbindlist doesn't recycle as it already expects each item to be a uniform list, data.frame or data.table

# expected
str(data.table(a=c(1,2), z=list(list("z",1),list(c("a","b")))))
# Classes ‘data.table’ and 'data.frame':    2 obs. of  2 variables:
#  $ a: num  1 2
#  $ z:List of 2
#   ..$ :List of 2
#   .. ..$ : chr "z"
#   .. ..$ : num 1
#   ..$ :List of 1
#   .. ..$ : chr  "a" "b"
```

it could handle any non-atomic types, not just lists.

### LINKED ISSUE #1349: rbind/rbindlist don't combine integer64 columns correctly with other types
``` r
dt1 = data.table(a = as.integer64(1), b = 1)
dt2 = data.table(a = 'a', b = 2)
dt3 = data.table(a = 1, b = 3)

# this works correctly
rbind(dt2, dt3)
#   a b
#1: a 2
#2: 1 3

# this doesn't work (and has an error message that's not true as per the above example)
rbind(dt1, dt2)
#Error in rbindlist(l, use.names, fill, idcol) : 
#  Class attributes at column 1 of input list at position 2 does not match with column 1 of input list at position 1. Coercion of objects of class 'factor' alone is handled internally by rbind/rbindlist at the moment.

# this doesn't work either
rbind(dt1, dt3)
```

### LINKED ISSUE #1754: obscure error in melt
I have [some more super-wide data](https://gist.github.com/franknarf1/c962a8abe9f34f054ba7970c2cadf2c9) and wanted to melt it:

```
melt(DT, 1:2)
# Error in rbindlist(l, use.names, fill, idcol) : 
#   Value of SET_STRING_ELT() must be a 'CHARSXP' not a 'character'
```

if I had instead run with `melt(DT, names(DT)[1:2])`, I'd see the same message. Interestingly, when I repeat the command, I get a different error: 

```
# Error in rbindlist(l, use.names, fill, idcol) : 
#   Value of SET_STRING_ELT() must be a 'CHARSXP' not a 'integer'
```

After that, I never see the 'character' error again. I ran this a few times on my single-row example and managed to crash R. This error isn't a big problem, though; just means I need to approach the problem differently.

### LINKED ISSUE #1871: rbindlist should deal with NULL values
I'm working with a big JSON string from web containing a list of lists, each has the same set of fields of scalar values. However, some items have `NULL` values, which makes `rbindlist` refuse to work.

A simple example is 

``` r
p <- list(list(a = 1, b = 2, x = NULL), list(a = 2, b = 3, x = 10))
rbindlist(p)
```

This produces the following error

```
Error in rbindlist(p) : 
  Column 3 of item 1 is length 0, inconsistent with first column of that item which is length 1. rbind/rbindlist doesn't recycle as it already expects each item to be a uniform list, data.frame or data.table
```

which is quite understandable since `NULL` is zero-length like `numeric(0)`. For me, it would be nice to have an argument or something to regard such zero-length entries as `NA` so that I don't have to manually iterate over all items and set them to missing values.

### LINKED ISSUE #2113: [Doc request] ?rbindlist shouldn't refer to rbind.data.table, which doesn't exist
The doc says

> `rbind.data.table` gained a `fill` argument to fill missing columns with `NA` in `v1.9.2`, which allowed for `rbind(...)` binding unequal number of columns.

But there is no `rbind.data.table`. Maybe this instead:

> `rbind(...)` gained a `fill` argument to fill missing columns with `NA` in `v1.9.2`, which allowed for binding unequal number of columns.

I guess this part of the docs will go away in a few version anyway (since it's just explaining history), so no big deal. I did find it confusing, though.

### LINKED ISSUE #2303: segfault when comparing a `NULL` column
I experience a segfault when running the below code on Windows, Mac, and Linux. While I've used the unexported `data.table:::null.data.table` for the minimal reprex, I first encountered the crash when [testing a package which Imports `data.table`](https://travis-ci.org/HughParsonage/grattan/builds/265137409). 

```r
library(data.table)
# data.table 1.10.5 IN DEVELOPMENT built 2017-08-15 16:41:46 UTC; appveyor
# The fastest way to learn (by data.table authors): https://www.datacamp.com/courses/data-analysis-the-data-table-way
# Documentation: ?data.table, example(data.table) and browseVignettes("data.table")
# Release notes, videos and slides: http://r-datatable.com
# Warning message:
#   package ‘data.table’ was built under R version 3.4.1 
sessionInfo()
# R version 3.4.0 (2017-04-21)
# Platform: x86_64-w64-mingw32/x64 (64-bit)
# Running under: Windows >= 8 x64 (build 9200)
# 
# Matrix products: default
# 
# locale:
#   [1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252    LC_MONETARY=English_Australia.1252
# [4] LC_NUMERIC=C                       LC_TIME=English_Australia.1252    
# 
# attached base packages:
#   [1] stats     graphics  grDevices utils     datasets  methods   base     
# 
# other attached packages:
#   [1] data.table_1.10.5    RevoUtilsMath_10.0.0
# 
#   [1] compiler_3.4.0   RevoUtils_10.0.4 tools_3.4.0 

x <- NULL
DT <- data.table:::null.data.table()
DT2 <- DT[, .(x)]
DT2[x > 1]
```

>    *** caught segfault ***
  address (nil), cause 'memory not mapped'
  Segmentation fault (core dumped)

### LINKED ISSUE #2305: [Request] Allow NULL columns in data.table()
This could be either a question or a request. The question would be: "Are NULL columns allowed in a data.table?" The request:

I was looking at #2303 and noticed this error:

```
library(data.table)
data.table(character(0), NULL)
# Error in data.table(character(0), NULL) : column or argument 2 is NULL
```

It would arguably be nice for `data.table()` to allow the construction of a table like this.

`as.data.table.list` allows it with no error:

```
DT = as.data.table(list(y = character(0), x = NULL))
```
`NULL` is idiomatic as a column value for grouped operations and (mostly) `rbindlist`: 

```
data.table(g = 1:2)[, if (.GRP == 1L) Sys.Date(), by=g]
rbind(DT, list(y = "a", x = 12))

# also works, albeit with a warning
data.table(g = 1:2)[, .(v = c("yo", "zo"), z = if (.GRP == 1L) Sys.Date()), by=g] 
# fails, with an informative error message
rbind(DT, list(y = "a", x = Sys.Date())) 
```

And finally, the "wildcard" behavior of `NULL` in grouped operations and `rbindlist` is built upon in packages like [vetr](https://cran.r-project.org/package=vetr) (see the "alike" vignette):

```
library(vetr)
alike(DT, data.table(y = "a", x = "b")) # TRUE
alike(DT, data.table(y = "a", x = 12)) # TRUE
```

I may be overlooking some arguments against it. The main ones that come to mind are: 

* There is no way to add such a `NULL` column by reference.
* Does this have any meaning or use for tables with rows?
* There are bugs (like the one I linked at the top) that go away if `NULL` columns aren't a thing (?).

If it's decided that data.tables _shouldn't_ have `NULL` columns (which would also be fine), then I guess `as.data.table.list(x)` may need to change to throw an error; or to drop the `NULL` elements of `x` even when `max(lengths(x)) == 0`.

### LINKED ISSUE #2819: rbindlist support for raw types

```r
library(data.table)
rbindlist(list(data.table(x = raw(1)), data.table(x = raw(1))))
#> Error in rbindlist(list(data.table(x = raw(1)), data.table(x = raw(1)))) : 
#>  Unsupported column type 'raw'
```

`#` `Output of sessionInfo()`
```
R version 3.5.0 (2018-04-23)

Matrix products: default

locale:
[1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252    LC_MONETARY=English_Australia.1252
[4] LC_NUMERIC=C                       LC_TIME=English_Australia.1252    

attached base packages:

other attached packages:
[1] data.table_1.11.0

[1] compiler_3.5.0  magrittr_1.5    tools_3.5.0     yaml_2.1.18     fastmatch_1.1-0 hutils_1.1.0  
```

### LINKED ISSUE #3032: SET_STRING_ELT error when `rbindlist` contains both tables with ordered factors of different levels and tables with ordered factors of common levels
Related issue https://github.com/Rdatatable/data.table/issues/899 was closed in https://github.com/Rdatatable/data.table/commit/67ace19ab19680e2056f02bf08d0b3a9b5db43df

``` r
library(data.table)
# data.table 1.11.5 IN DEVELOPMENT built 2018-09-05 02:11:00 UTC; appveyor  Latest news: r-datatable.com

dt1 <- data.table(x = as.ordered(letters[c(1, 3, 5:8)]))
dt2 <- data.table(x = as.ordered(letters[c(2, 5:8)]))
dt3 <- data.table(x = as.ordered(letters[c(2, 5:8)]))

rbindlist(list(dt1, dt2, dt3), use.names = TRUE, fill = TRUE)
#> Error in rbindlist(list(dt1, dt2, dt3), use.names = TRUE, fill = TRUE):
#>   attempt to set index 7/7 in SET_STRING_ELT
```

<sup>Created on 2018-09-06 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).</sup>

`#` `Output of sessionInfo()`
```
R version 3.5.1 (2018-07-02)

Matrix products: default

locale:
[1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252   
[3] LC_MONETARY=English_Australia.1252 LC_NUMERIC=C                      
[5] LC_TIME=English_Australia.1252    

attached base packages:

other attached packages:
[1] data.table_1.11.5

 [1] Rcpp_0.12.18     ps_1.1.0         rprojroot_1.3-2  digest_0.6.16   
 [5] crayon_1.3.4     withr_2.1.2      assertthat_0.2.0 R6_2.2.2        
 [9] backports_1.1.2  magrittr_1.5     reprex_0.2.0     evaluate_0.11   
[13] stringi_1.1.7    rlang_0.2.2      rstudioapi_0.7   callr_3.0.0     
[17] whisker_0.3-2    rmarkdown_1.10   devtools_1.13.6  tools_3.5.1     
[21] stringr_1.3.1    yaml_2.2.0       compiler_3.5.1   processx_3.2.0  
[25] base64enc_0.1-3  clipr_0.4.1      htmltools_0.3.6  memoise_1.1.0   
[29] knitr_1.20 
```

### LINKED ISSUE #3343: rbind/rbindlist failing on empty input when source is NULL list
```
DT1 = list(a = NULL)
DT2 = list(a = NULL)
setDT(DT1)
setDT(DT2)

rbind(A = DT1, B = DT2, idcol = 'id')
```

> Error in rbindlist(l, use.names, fill, idcol) : 
>  attempt to set an attribute on NULL

It's because `class(DT1$a)` is `NULL`. I'm not sure what attribute is being set.

Or maybe this is a problem with `setDT`?

Fix for me now is to correct my upstream inputs for the empty cases, but this error seems wrong.

### LINKED ISSUE #3373: rbindlist use.names=TRUE returns random column order when ncol>255
Follow up to #3306.  The printing issue there is fixed by not using use.names=TRUE in that case.  But rbindlist in general still has the issue. Also see:  https://github.com/Rdatatable/data.table/pull/3352#issuecomment-461718583

minimal example:

```
ncol = 30
DT = setDT(replicate(ncol, rnorm(3L), simplify = FALSE))
identical(colnames(rbind(DT[1], DT[3])), colnames(DT))
# [1] TRUE


ncol = 300
DT = setDT(replicate(ncol, rnorm(3L), simplify = FALSE))
identical(colnames(rbind(DT[1], DT[3])), colnames(DT))
# [1] FALSE
```

mattdowle:
Closes #3373  (and dup #3427) 
Closes #1871
Closes #1302 
Closes #3343
Closes #524 
Closes #3032 
Closes #2819 
Closes #2305 
Closes #2303
Closes #2113 
Closes #1754 
Closes #1349 
Removed hashing code in rbindlist.c: 960 lines down to 360. Coverage up from 88.5% to 100%
Faster combining of factor columns.
Position of dups is now retained rather than grouping together.
`chin` created at C level for readability and `in` argument removed from C level `chmatch`.
`chmatch2` moved into chmatch.c and simplified; benchmark reran and confirmed no slow down
- [x] reinstate ordered factor tests: 1120-1125 & 1168.2, 1393.1, 1393.2
- [x] order of dups now retained:  1288.02
- [x] add tests for closed issues above. Prefix with "closes" when done including news item.
- [x] move all issues above to "closes"
- [x] diff coverage 100%
- [x] 2 memory todo in rbindlist.c
- [x] reduce test 2000's size by 10x
- [x] rbindlist use.names="check" default


src/chmatch.c
```diff
@@ -37,45 +40,111 @@ SEXP chmatch(SEXP x, SEXP table, R_len_t nomatch, Rboolean in) {
...
+    SET_TRUELENGTH(td[i], 0);  // reinstate 0 rather than leave the -i-1
   savetl_end();
   UNPROTECT(1);
   return(ans);
 }
 
-SEXP chmatchwrapper(SEXP x, SEXP table, SEXP nomatch, SEXP in) {
-  return(chmatch(x,table,INTEGER(nomatch)[0],LOGICAL(in)[0]));
+// for internal use from C :
+SEXP chmatch(SEXP x, SEXP table, int nomatch) {  // chin=  chmatchdup=
```
jangorecki: if it is for internal use from C why not return C object type instead of SEXP?

inst/tests/tests.Rraw
```diff
@@ -13626,6 +13633,54 @@ dx = data.table(id = 1L, key = "id")
 di = list(z=c(2L, 1L))
 test(1999.2, key(dx[di]), NULL)
 
+# chmatchdup test from benchmark at the bottom of chmatch.c
+set.seed(45L)
+x = sample(letters, 1e6, TRUE)
+y = sample(letters, 1e7, TRUE)
```
jangorecki: 76.3 MB mem resource, probably the most mem consuming unit test, still not much

src/assign.c
```diff
@@ -942,37 +942,51 @@ void memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
     // list columns already have each item initialized to NULL
     break;
   default :
-    error("Unsupported type '%s'", type2char(type));
+    error("Internal error: writeNA pass a vector of type '%s'", type2char(TYPEOF(v)));
   }
+}
+
+
+SEXP allocNAVector(SEXPTYPE type, R_len_t n)
```
jangorecki: unless it complicates a lot, maybe we could switch to `R_xlen_t` already?

src/assign.c
```diff
@@ -942,37 +942,51 @@ void memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
-    for (int i=0; i<n; i++) SET_STRING_ELT(v, i, NA_STRING);
+    // If there's ever a way added to R API to pass NA_STRING to allocVector() to tell it to initialize with NA not "", would be great
+    for (int i=from; i<=to; ++i) SET_STRING_ELT(v, i, NA_STRING);
     break;
   case VECSXP :
     // list columns already have each item initialized to NULL
     break;
   default :
-    error("Unsupported type '%s'", type2char(type));
+    error("Internal error: writeNA pass a vector of type '%s'", type2char(TYPEOF(v)));
```
jangorecki: `nocov could be here

src/chmatch.c
```diff
@@ -14,15 +14,18 @@ SEXP match_logical(SEXP table, SEXP x) {
...
 }
 
-SEXP chmatch(SEXP x, SEXP table, R_len_t nomatch, Rboolean in) {
-  R_len_t i, m;
-  SEXP ans, s;
+static SEXP chmatchMain(SEXP x, SEXP table, int nomatch, bool chin, bool chmatchdup) {
   if (!isString(x) && !isNull(x)) error("x is type '%s' (must be 'character' or NULL)", type2char(TYPEOF(x)));
   if (!isString(table) && !isNull(table)) error("table is type '%s' (must be 'character' or NULL)", type2char(TYPEOF(table)));
-  PROTECT(ans = allocVector(in ? LGLSXP : INTSXP,length(x))); // if fails, it fails before savetl
+  if (chin && chmatchdup) error("Internal error: either chin or chmatchdup should be true not both");
```
jangorecki: another `nocov`

src/fmelt.c
```diff
@@ -350,6 +350,71 @@ static void preprocess(SEXP DT, SEXP id, SEXP measure, SEXP varnames, SEXP valna
...
+// TODO keep the original factor columns as factor and use new technique in rbindlist.c. The calling
+// environments are a little difference hence postponed for now (e.g. rbindlist calls writeNA which
+// a general purpose combiner would need to know how many to write)
+// factorType is 1 for factor and 2 for ordered
+// will simply unique normal factors and attempt to find global order for ordered ones
+{
+  int maxlevels=0, nitem=length(factorLevels);
+  for (int i=0; i<nitem; ++i) {
+    SEXP this = VECTOR_ELT(factorLevels, i);
+    if (!isString(this)) error("Internal error: combineFactorLevels in fmelt.c expects all-character input");
```
jangorecki: `nocov`

src/fmelt.c
```diff
@@ -350,6 +350,71 @@ static void preprocess(SEXP DT, SEXP id, SEXP measure, SEXP varnames, SEXP valna
...
+    maxlevels+=length(this);
+  }
+  if (!isString(target)) error("Internal error: combineFactorLevels in fmel.c expects a character target to factorize");
```
jangorecki: `nocov`

src/chmatch.c
mattdowle: Just because that SEXP is most often a vector. SEXP is convenient even internally so we can leave R to do the memory management for us.  Otherwise for single values, yes C type internally.

inst/tests/tests.Rraw
mattdowle: Good point. I was only thinking about the time of the test (0.09s) and didn't think about the time or space for the sample. 10x smaller would be better. Will do.

src/assign.c
mattdowle: A bit too much to juggle right now. It probably won't be more complicated, but there's a chance it will, and that unkown adds slightly to the cognitive burden. Thinking we'll just do a global replace in the next release and have to sweep through everything (including all `int`) then.

mattdowle:
On #1302 did you see my comment there: https://github.com/Rdatatable/data.table/issues/1302#issuecomment-472653934.  I'm looking at it and thinking it's correct behaviour.  A list column isn't variable size:  there should be the same number of slots as rows in the data.table.  Some or all of those slots may be NULL.  Then each slot can contain anything (that's where the variable length comes in : within the cell).
In other words, if this is desired output : 
```
list(list("a"),list("b"))  # list of lists
```
then this should be the input : 
```
l = list(
  list(a=1, z=list(list("a"))),
  list(a=2, z=list(list("b")))
)
identical(rbindlist(l)$z, list(list("a"),list("b")))  # TRUE
```
and the current and old behaviour looks correct : 
```
l = list(
  list(a=1, z=list("a")),
  list(a=2, z=list("b"))
)
rbindlist(l)$z, c(list("a"),list("b"))  # TRUE
identical(c(list("a"),list("b")), list("a", "b"))  # TRUE
```

jangorecki:
@mattdowle that make sense. As long as we can achieve what we need it is fine. I pushed some extra unit tests for behaviour requested in #1302 having modified input.

mattdowle:
I'm thinking this behaviour could be better : 
```
> DT1 = data.table(a=1:2, b=5:6)
> DT2 = data.table(b=7:8, a=3:4)
> rbindlist(list(DT1,DT2))
       a     b
   <int> <int>
1:     1     5
2:     2     6
3:     7     3
4:     8     4
> 
```

could instead be : 
```
> DT1 = data.table(a=1:2, b=5:6)
> DT2 = data.table(b=7:8, a=3:4)
> rbindlist(list(DT1,DT2))
Warning message:
  Columns with different names have been stacked together. Set use.names=TRUE to
  match column names (like rbind() does), or use.names=FALSE to turn off this warning.
       a     b
   <int> <int>
1:     1     5
2:     2     6
3:     7     3
4:     8     4
> rbindlist(list(DT1,DT2), use.names=TRUE)
       a     b
   <int> <int>
1:     1     5
2:     2     6
3:     3     7
4:     4     8
> rbindlist(list(DT1,DT2), use.names=FALSE)
       a     b
   <int> <int>
1:     1     5
2:     2     6
3:     7     3
4:     8     4
> 
```
We should have to explicitly write `use.names=FALSE` to get the current default of stacking ignoring names.   Default for use.names could be NA to convey this change.
Only if all the names matched would rbindlist() silently stack them as happens now.

src/rbindlist.c
```diff
@@ -225,19 +225,21 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
       if (firsti==-1) { firsti=i; firstw=w; firstCol=thisCol; }
-      else if (!factor && !R_compute_identical(thisClass, getAttrib(firstCol,R_ClassSymbol), 0)) {
+      else if (!factor && !int64 && !R_compute_identical(getAttrib(thisCol, R_ClassSymbol), getAttrib(firstCol,R_ClassSymbol), 0)) {
         error("Class attribute on column %d of item %d does not match with column %d of item %d.", w+1, i+1, firstw+1, firsti+1);
       }
     }
     // in future warn ... if (factor && anyNotStringOrFactor) warning("Column %d contains a factor but not all items for the column are character or factor", idcol+j+1);
     if (!foundName) { char buff[12]; sprintf(buff,"V%d",j+1), SET_STRING_ELT(ansNames, idcol+j, mkChar(buff)); }
     if (factor) maxType=INTSXP;  // any items are factors then a factor is created (could be an option)
+    if (int64 && maxType!=REALSXP) error("Internal error: column %d of result is determined to be integer64 but maxType==%d != REALSXP", j+1, maxType); // nocov
```
jangorecki: Missing `#` near nocov

mattdowle: Good sport - thanks. Looks like C coverage doesn't distinguish parts of lines, unlike R coverage.

renkun-ken:
@mattdowle, I'm curious if there's any possibility that `rbindlist` benefits from multi-threading?

src/subset.c
```diff
@@ -270,6 +270,7 @@ SEXP subsetDT(SEXP x, SEXP rows, SEXP cols) {
     ansn = LENGTH(rows);  // has been checked not to contain zeros or negatives, so this length is the length of result
     for (int i=0; i<LENGTH(cols); i++) {
       SEXP source = VECTOR_ELT(x, INTEGER(cols)[i]-1);
+      if (isNull(source)) error("Internal error: column %d of data.table is NULL; malformed", i+1);
```
jangorecki: missing nocov

mattdowle:
@renkun-ken It should benefit for all types other than character. Stacking factor columns should be faster now with this PR.

src/subset.c
mattdowle: It's covered though because C coverage doesn't inspect within line.

MichaelChirico:
quite the PR!! 💥

renkun-ken:
I run my systems with this PR merged and everything looks good :)

PR #3458, opened 2019-03-14
setorder can keep keys in more cases

### LINKED ISSUE #3456: setorder shouldn't wipe key unnecessarily
```
DT = data.table(a = rep(4:1, 1:4), v = rnorm(10))
setkey(DT, a)
setorder(DT, a, v)
key(DT)
# NULL
```

Obviously `DT` is still sorted by `a` after `setorder`; that's why the second `setkey` here only runs `setattr` (btw, we should report this is happening if `verbose = TRUE`)

```
DT = data.table(a = rep(4:1, 1:4), v = rnorm(10))
setkey(DT, a, v)
setkey(DT, a, verbose = TRUE)
```

The note in the code is:

> if `forderv` is not 0-length, it means order has changed. So, set key to `NULL`, else retain key.

So I guess it could be a conceptual question -- do we reset the key if the order has changed _but the table is still sorted by the stated key_?

This is particularly odd if I want to sort the last column in decreasing order. Is there any have `DT` keyed by `a` but guaranteed `v` is in decreasing order within `a`?

MichaelChirico:
Closes #3456


MichaelChirico:
I'm afk but just struck me that all(order >=0) is slightly better

Issue #2837, opened 2018-05-05
FR: warn/error when updating and i has duplicates


MichaelChirico:
```
DT = data.table(a = 1:5)
DT[c(1, 1, 2), a := 3:5][]
#    a
# 1: 4
# 2: 5
# 3: 3
# 4: 4
# 5: 5
```

It seems the final element to be assigned (here 4 corresponds to the second instance of 1 in `i`). Not clear what the right behavior is in this case; my guess is most often it's a user mistake, hence a warning. But also possibly an error since "correct" behavior is ambiguous.

MichaelChirico:
not possible to use a hash table to say "seen"/"not seen" at each element?

what about an option to green light the speed hit?

also in the case of a keyed subset, the cost is trivial

franknarf1:
Somewhat related: the same problem during a join https://github.com/Rdatatable/data.table/issues/2022

Fwiw, I find that I use both joins and Michael's approach, since often

```
# from the OP
mDT = data.table(row = c(1,1,2), new = 3:5)
DT[mDT$row, a := mDT$new]
```

is more expedient than adding a row-number column...

```
# adding a column and doing a join
mDT = data.table(row = c(1,1,2), new = 3:5)
DT[, row := .I]
DT[mDT, on=.(row), a := i.new]
```

so my code is vulnerable to both idioms until/unless I add a dupe check, `anyDuplicated(mDT, by=on_cols)` or something.

jangorecki:
if we want to add so many checks we should also collect more attributes. We already keep info if object is sorted, we can also put info if any is NA, or if there are duplicates. So at least we can reduce overhead related to extra checks.

MichaelChirico:
Related: #2879. Agree with Pasha that potential speed hit should be the primary consideration re: whether to implement this. At a minimum, we should make sure there's a quick blurb in `?data.table` highlighting the existing behavior.

PR #3460, opened 2019-03-17
.Last.updated value after :=, closes #1885

### LINKED ISSUE #1885: Save .Last.updated after each :=
Similar to base R's `.Last.value`, data.table could save `.Last.updated` holding the number of rows updated by the last `:=`.  This could be used afterwards in production code to check that if row updates are expected then they occur and for reporting and logging.

jangorecki:
Closes #1885 
Now resolved: ~~but adds extra overhead of `sum(!is.na(irows))`, so should not be merged. Most likely changes needs to be pushed to C code where the actual updated number of rows is established. Not doing it now to avoid merge conflicts with rbindlist rework branch~~


mattdowle:
Good idea. Would be nice to have this.
The only thing is that packages aren't allowed to write to `.GlobalEnv` without user's consent (CRAN policy).  It could be optional, default off, turned on by user. But then what if a package wished to use it inside that package?  data.table could have an object which it exports though: `.Last.updated`, a length-1 integer vector. Packages are allowed to change variables they export (like assignInNamespace does).  Then users can access the value via search() or explicitly using `data.table::.Last.updated`.
Is `.Last.updated` the best name?  Maybe it should have `nrows` or similar in the name somehow to convey it's a number of rows.

jangorecki:
I pushed new version, using C and writing variable in data.table namespace.
The issue is that I am getting failure of one of the tests when running package check, but I am not able to reproduce it interactively. I have no clue about the reason for that.
```
Running test id 2028.14      Test 2028.14 ran without errors but failed check th
at x equals y:
> x = .Last.nrow 
First 6 of 1 (type 'integer'): [1] 1
> y = 4L 
First 6 of 1 (type 'integer'): [1] 4
Mean relative difference: 3
```

MichaelChirico:
No manual entry?

mattdowle:
Good catch. Could be added to `?:=` with alias at the top of that `.Rd` so `?.Last.updated` works.

jangorecki:
I remember I made manual, but it seems I was just using `git add -u` so new files has not been added. Strangely, I must have deleted that branch locally already.

mattdowle:
I guess this means that `R CMD check` requires exported functions to be documented but not exported variables. That's odd if so. I can't think why it shouldn't check exported variables too. Maybe we have other undocumented exported variables then.

R/onLoad.R
```diff
@@ -1,5 +1,7 @@
 # nocov start
 
+.Last.updated <- vector("integer", 1L) # exported variable; number of rows updated by the last := or set(), #1885
```
MichaelChirico: This means the initial value (even if `data.table` is not loaded) is `0L`, is that preferable to `NA_integer_`?

```
library(data.table)
.Last.updated
# [1] 0
DT = data.table(a = 1)
DT[FALSE, a := 2]
.Last.updated
# [1] 0
```

Conceptually, I think `.Last.updated` is different in the two cases?

MichaelChirico:
Is there any benefit to having a `reset_last_value` function as well, or it should be taken as immutable? I note it already seems like we can't overwrite its value manually...

MichaelChirico:
We need to add the same note as #2837 alongside the `.Last.updated` man page, wherever it ends up:

```
DT = data.table(a = 1L)
DT[c(1, 1), a := 2:3]
DT[]
#    a
# 1: 3
.Last.updated
# [1] 2
```

jangorecki:
I don't think we need to worry about first use, and yes, good point about duplicates on assignment, will add together with manual

PR #3461, opened 2019-03-17
R dataptr out of the loops in gmean

jangorecki:


src/gsumm.c
```diff
@@ -476,35 +476,40 @@ SEXP gmean(SEXP x, SEXP narm)
...
+  case LGLSXP: case INTSXP: {
+    int *restrict ix = INTEGER(x);
     for (int i=0; i<n; i++) {
       int thisgrp = grp[i];
-      int ix = (irowslen == -1) ? i : irows[i]-1;
-      if(INTEGER(x)[ix] == NA_INTEGER) continue;
-      s[thisgrp] += INTEGER(x)[ix];  // no under/overflow here, s is long double
+      int xirows = (irowslen == -1) ? i : irows[i]-1;
+      if (ix[xirows] == NA_INTEGER) continue;
+      s[thisgrp] += ix[xirows];  // no under/overflow here, s is long double
```
mattdowle: It's a bit difficult to review this change, since the same variable name now means something else.  Also the name `xirows` conveys (to me anyway) a maximal constant value, not something that is changing on each iteration.
When I've been taking R API outside loops like this I've tried to use a `d` suffix (to represent the dataptr) on the end of the existing SEXP name.  So in this case the dataptr for `x` would be called `xd`.

mattdowle: @jangorecki `restrict` is only needed for when we're writing to that memory.  In this case, we're just reading from `x` so it should be `const` which is even stronger and more helpful to the compiler than `restrict`.

src/gsumm.c
```diff
@@ -476,35 +476,40 @@ SEXP gmean(SEXP x, SEXP narm)
...
       c[thisgrp]++;
     }
+  }
     break;
   default:
     free(s); free(c);
     error("Type '%s' not supported by GForce mean (gmean) na.rm=TRUE. Either add the prefix base::mean(.) or turn off GForce optimization using options(datatable.optimize=1)", type2char(TYPEOF(x)));
   }
   ans = PROTECT(allocVector(REALSXP, ngrp));
+  double *restrict dans = REAL(ans);
```
mattdowle: Although, there is a write to `ans` here, there's no parallel region.

src/gsumm.c
jangorecki: for variable names I used `i` to refer integer (when using `INTEGER()`) and `d` when using double (`REAL`).

mattdowle: I know. But I've been using a d suffix elsewhere to mean 'dataptr'. I'm happy to choose a different better standard if you'd like.
In this case, consider : 
image
vs what I changed it to : 
image
I'm just trying to say that I found the latter easier and faster to review. Because the change is limited to the `INTEGER(x)` being taken outside, and because the naming convention is the same as elsewhere.

Issue #3462, opened 2019-03-17
ghead argument "n" is missing, with no default


jangorecki:
```r
data.table(a=1:2, b=2:1)[, head(.SD), a]
Error in ghead(.SD) : argument "n" is missing, with no default
```
this should use default `n` of `head.data.table` which is 6L
same `tail`

mattdowle:
Closed by #3463.

PR #3463, opened 2019-03-17
Gforce first last, head tail, coverage

### LINKED ISSUE #2030: gfirst(.SD) throws an error about not using head(.SD, n), but the latter works
I tried this and got an error:

    library(data.table)
    DT = data.table(id = c(1L,1L,2L), v = 1:3)
    DT[, first(.SD), by=id, .SDcols="v", verbose = TRUE]
    # Finding groups using forderv ... 0 sec
    # Finding group sizes from the positions (can be avoided to save RAM) ... 0 sec
    # lapply optimization is on, j unchanged as 'first(.SD)'
    # GForce optimized j to 'gfirst(.SD)'

> Error in gfirst(.SD) : 
>   GForce head can only be applied to columns, not .SD or similar. To get head of all items in a list such as .SD, either add the prefix utils::head(.SD) or turn off GForce optimization using options(datatable.optimize=1).

I tried head and it did work, contrary to the error message:

    DT[, head(.SD, 1L), by=id, .SDcols="v", verbose=TRUE]
    # Finding groups using forderv ... 0.02 sec
    # Finding group sizes from the positions (can be avoided to save RAM) ... 0 sec
    # lapply optimization changed j from 'head(.SD, 1L)' to 'list(head(v, 1L))'
    # GForce optimized j to 'list(ghead(v, 1L))'
    # Making each group and running j (GForce TRUE) ... 0 secs
    #    id v
    # 1:  1 1
    # 2:  2 3

Not a big deal and maybe it'll be handled in the process of of doing #735

jangorecki:
closes #2030, #3462


jangorecki:
Coverage must be wrong here as f0a0a94 did not introduce any new untested lines and none of unit tests were removed

Issue #3473, opened 2019-03-26
Release checks


mattdowle:
- [x] CRAN_Release.cmd (e.g. UBSAN, ASAN, rchk, gctorture)
- [x] GitLab Pipeline (e.g. R 3.1.0, --enable-strict-barrier)
- [x] News item review and tidy
- [x] win-builder 3*OK
- [x] Full rerun latest revdeps refreshed (774)
- [x] @renkun-ken final go-first run please (last successful run with da0d81dde6946b5d8aa4cdf476684af781422f6a)
- [x] Submitted to CRAN  (several opportunities to halt after this)
- [x] CRAN pre-test success email received and OK
- [x] Bump dev version
- [x] Copy and paste `revdep.R::status()` output as new comment to this issue 
- [x] CRAN revdep status impact results email received
- [x] Matt replied to confirm and explain differences to submission note
- [x] CRAN email received to confirm accepted
- [x] Add publish date to v1.12.2 title in NEWS.md.  Update version and date in wiki/Installation heading
- [x] Tweet to #rdatatable:   https://twitter.com/MattDowle/status/1115721380723429376

mattdowle:
Final revdep status : 
```
> status()

Installed data.table to be tested against: 1.12.1 

CRAN:
 ERROR   :   1 : expss 
 WARNING :   0 :  
 NOTE    : 188 
 OK      : 432 
 TOTAL   : 621 / 621 
 
BIOC:
 ERROR   :   6 : BiocParallel dada2 GENESIS ImmuneSpaceR maftools metagene 
 WARNING :  21 : CONFESS crossmeta DeepBlueR ELMER eQTL flowWorkspace GenoGAM GOTHiC
                 gpart iCNV methylPipe MinimumDistance netSmooth RiboProfiling rTANDEM
                 S4Vectors SISPA TENxBrainData TitanCNA Ularcirc Uniquorn 
 NOTE    : 102 
 OK      :  24 
 TOTAL   : 153 / 153 
 
TOTAL          : 774 
```

`expss` is marked `--no-tests` on CRAN. I see similar issues locally with data.table v1.12.0. So `expss` is not caused by data.table 1.12.2.
Bioconductor issues are long-standing. I'll continue to send emails; see CRAN_Release.cmd [here](https://github.com/Rdatatable/data.table/blob/master/CRAN_Release.cmd#L432).

PR #3483, opened 2019-03-28
rbindlist check warning downgraded to message

mattdowle:
To pass package `tatoo` for #3473.
It's a print not message() for the reason written in news item.
Extensive control option added with tests.


jangorecki:
`REprintf` is to make R's message from C

mattdowle:
@jangorecki `REprintf` prints on `stderr`. It doesn't do what R level `message()` does; e.g. it can't be suppressed with `suppressMessages()`.
However, `rbindlist()` could construct a call to `message()` and `eval()` that at C level though. Will try that.

jangorecki:
@mattdowle good to know, it took me a while to find `REprintf` and now I hear it is not what I wanted.
You think we could turn that into a function? so we can use it whenever `message` is needed.
```c
eval(PROTECT(lang2(install("message"),PROTECT(ScalarString(mkChar(buff))))), R_GlobalEnv); UNPROTECT(2)
```

mattdowle:
Making it a function would be good.  I was going to.  The tricky thing was that `Rprintf`, `REprintf`, `warning`, and `error` all take `...` variable argument list.  We'd want the new `message` to follow the same parameter signature so that we could use it the same way as we use warning and error; e.g.  `message("Column %d ['%s'] has a problem", int, char *)`.  I couldn't remember how to do that in C using `va_arg` so I just went with this way for now. Or we could create `void message(const char *)` that takes a fixed string like that. In this example in `rbindlist`, because the new option allows control of the message/warning/error,  that section is using `sprintf` first anyway.  But in general it would be nice to have `message()` as function.
Btw, being able to fetch an `option()` value from C worked out nicely. Saves having to pass the option value down from R and is much more convenient and robust.  I didn't know `GetOption()` was available in the R API at C level until recently.  Maybe we can use it more in other places.  For example,  there are some function where I wanted to add some verbose output but `verbose` hadn't been passed through from R level. We could wrap GetOption("datatable.verbose") into a C function `bool verbose()` and just be careful not to call that inside loops.

jangorecki:
@mattdowle agree. We have one such case in `fastmean`.

PR #3486, opened 2019-03-30
Widen TOC for the FAQ vignette.

vlulla:


mattdowle:
@vlulla Great - thank you.

PR #3500, opened 2019-04-10
printtz

### LINKED ISSUE #2842: printing of POSIXct should (could?) include tz info if it's there
```
DT = data.table(
  t1 = as.POSIXct('2018-05-01 12:34:56', tz = 'UTC'),
  t2 = as.POSIXct('2018-05-01 12:34:56', tz = 'Asia/Singapore')
)
DT
#                     t1                  t2
# 1: 2018-05-01 12:34:56 2018-05-01 12:34:56
DT$t1
# [1] "2018-05-01 12:34:56 UTC"
DT$t2
# [1] "2018-05-01 12:34:56 +08"
```

The printed output makes `t1` and `t2` seem identical, but they're not.

The same problem applies to `data.frame`s, to be fair:

```
setDF(copy(DT))[]
#                    t1                  t2
# 1 2018-05-01 12:34:56 2018-05-01 12:34:56
```

Probably timezone printing should be on by default due to the above, but at least there should be an option.

fparages:
Allows printing of timezones when printing data.tables with POSIX columns (if tzone attribute is defined).

Closes #2842.  An item in #1523 list.

Defined as optional via a parameter timezone with default value = FALSE, for backwards compatibility.


fparages:
In line with @franknarf1 's comment in #1523 :
What's the consensus: would you prefer the new feature to print out time offset with respect to UTC, or the full tz name (as it is PRed now)?

MichaelChirico:
https://en.wikipedia.org/wiki/ISO_8601

ISO 8601 timestamps would be ideal IMO:

> Time zones in ISO 8601 are represented as local time (with the location unspecified), as UTC, or as an offset from UTC.

Many other languages also understand the `T` delimiter between YMD and HMS... I also think base R's `strptime` should be improved to handle ISO 8601 timestamps natively but that's another issue...

It would be amazing if `fwrite` could write ISO timestamps & `fread` could read them but that feels a bit like scope creep -- I think the improvement to upstream in base R would be better...

mattdowle:
Thanks @fparages! I've invited you to be project member. (You'll need to accept the invite before GitHub adds you.)

PR #3510, opened 2019-04-16
empty argument in j output caught early & stopped

### LINKED ISSUE #3507: [Feature Request] Improve error message in case of omitted aggreation
```r
library(data.table)
dt <- data.table(x = runif(100), y = sample(LETTERS[1:5], 100, TRUE))
dt[, .(avg=mean(x), ), by=y]   ## dangling comma past mean()
```

yields

```
Error in dotN(this_jsub) : 
  argument "this_jsub" is missing, with no default
```

whereas 1.11.8 and 1.12.0 produced the somewhat clearer:

```
Error in list(.External(Cfastmean, x, FALSE), ) : argument 2 is empty
```

It would be helpful to make the message a little plainer.  Took me longer than I care to admit
to realize I left a dangling comma to yield on empty 'operator'.

```r
#  Output of sessionInfo()
R version 3.5.3 (2019-03-11)
Running under: Ubuntu 18.10

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.3.3.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8
 [8] LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C

attached base packages:

other attached packages:
[1] data.table_1.12.2

[1] compiler_3.5.3
```

MichaelChirico:
Closes #3507 Given that some other languages support trailing commas, e.g. Python:

```
python3
$ x = [1, 2, ]
$ x
# [1, 2]
```

It might be worth considering whether to support that here by simply dropping that element of `jsub` (which would break from base R).

Probably coders should be held to picking their commas to match the language -- in R, it's better to use _leading_ commas rather than trailing (for the record, this practice greatly facilitates including/dropping fields in exploratory work, since lines can be commented/uncommented without breaking syntax)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -41,6 +41,8 @@
 
 2. `keyby=colName` could use the wrong index and return incorrect results if both `colName` and `colNameExtra` (where `colName` is a leading subset of characters of `colNameExtra`) are column names and an index exists on `colNameExtra`, [#3498](https://github.com/Rdatatable/data.table/issues/3498). Thanks to Xianying Tan for the detailed report and pinpointing the source line at fault.
 
+3. `j = list(x, )` would sometimes correctly error but with a confusing diagnosis, [#3507](https://github.com/Rdatatable/data.table/issues/3507). This is now caught and stopped much earlier with a helpful admonition; thanks @eddelbuettel for the report.
+
 #### NOTES
 
 1. `rbindlist`'s `use.names="check"` now emits its message for automatic column names (`"V[0-9]+"`) too, [#3484](https://github.com/Rdatatable/data.table/pull/3484). See news item 5 of v1.12.2 below.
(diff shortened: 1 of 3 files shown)


mattdowle:
Great.  For completeness, done :  https://github.com/Rdatatable/data.table/commit/8eb3bf4fd04124033a4856fd5772f80afc514525

Issue #3492, opened 2019-04-08
Two test failures on armv7hl


QuLogic:
Two tests fail on [armv7hl](https://koji.fedoraproject.org/koji/taskinfo?taskID=34042977) with 1.12.2, when they didn't on 1.12.0, tests 1018 and 1737.5:

```
Running the tests in 'tests/main.R' failed.
Last 13 lines of output:
   3: 1848  6.522     1
   4: 1835  6.181     1
   5: 1912  5.300     2
   6: 1223  4.036   728
   7:  861  3.794     1
   8: 1639  3.770   143
   9: 1253  3.723   485
  10: 1779  3.067    13
  
  Error in test.data.table() : 
    2 errors out of 6578 in 00:02:16 elapsed (00:02:02 cpu) on Mon Apr  8 03:08:59 2019. [endian==little, sizeof(long double)==8, sizeof(pointer)==4, TZ=NA, locale='C', l10n_info()='MBCS=FALSE; UTF-8=FALSE; Latin-1=FALSE', getDTthreads()='omp_get_num_procs()==4; R_DATATABLE_NUM_PROCS_PERCENT=="" (default 50); R_DATATABLE_NUM_THREADS==""; omp_get_thread_limit()==2147483647; omp_get_max_threads()==4; OMP_THREAD_LIMIT==""; OMP_NUM_THREADS==""; data.table is using 2 threads. This is set on startup, and by setDTthreads(). See ?setDTthreads.; RestoreAfterFork==true']. Search tests/tests.Rraw for test numbers: 1018, 1737.5.
  In addition: Warning message:
  In parse(n = -1, file = file, srcfile = NULL, keep.source = FALSE) :
    invalid input found on input connection '/builddir/build/BUILD/data.table/data.table.Rcheck/data.table/tests/tests.Rraw'
```

# Minimal reproducible example
```
$ _R_CHECK_FORCE_SUGGESTS_=0 R CMD check data.table
```
(xts and nanotime are not packaged yet, so I need to skip the Suggests check.)

# Output of sessionInfo()
I can't get this exactly because it's on a builder. The closest I can do is the output from a local build (though platform will be different of course):
```
R version 3.5.3 (2019-03-11)
Platform: x86_64-redhat-linux-gnu (64-bit)
Running under: Fedora 31 (Rawhide)

Matrix products: default
BLAS/LAPACK: /usr/lib64/R/lib/libRblas.so

locale:
[...]

PR #3515, opened 2019-04-22
fix CRAN noLD note

jangorecki:
should fix "noLD" note listed in Additional issues on CRAN check https://cran.r-project.org/web/checks/check_results_data.table.html

### DIFF
--- a/inst/tests/froll.Rraw
+++ b/inst/tests/froll.Rraw
@@ -125,7 +125,7 @@ anserr = list(
 )
 errs = sapply(lapply(anserr, abs), sum, na.rm=TRUE)
 if (!(.Platform$OS.type=="windows" && getDTthreads()>1L)) { # windows 2+ threads rounding issue: #3346
-  test(9999.024, errs[["froll_exact"]]==0)
+  if (.Machine$sizeof.longdouble == 16L) test(9999.024, errs[["froll_exact"]]==0) # only where long double available, otherwise we get noLD CRAN note
   test(9999.025, errs[["froll_fast"]]>errs[["froll_exact"]])
 }
 test(9999.026, errs[["fastma"]]>errs[["froll_exact"]])


QuLogic:
BTW, as a note, the system at #3492 has `sizeof(long double) == 8` and still passes this test, so it seems to only be a problem for systems without long double entirely.

jangorecki:
Yes, there is noLD is an extra check on cran reported separately in additional issues section. And this PR was only meant to fix this.

QuLogic:
Yes, but the check in this PR disabled the test for sizeof==8 as well, not just noLD.

jangorecki:
This is not an issue, as per WRE those tests shouldn't be here in the first place. It make sense to keep them only to spot regression, so single common use case is enough.

mattdowle:
@jangorecki  But something doesn't quite add up here.  @QuLogic has a point:  they have a machine on which .Machine$sizeof.longdouble is 8 and yet test 9999.024 passes for them.  That's strange isn't it?

jangorecki:
> the system at #3492 (armv7hl) has sizeof(long double) == 8 and still passes this test

@QuLogic by "pass this test" you mean it is being evaluated? or simply does not raise error (thus is probably skipped).
Could you please check on latest devel? #3492 was about 1.12.2 and that test got changed later on.
Below code block so you don't need to run all tests:
```r
library(data.table)
test = data.table:::test
set.seed(108)
x = sample(c(rnorm(1e3, 1e6, 5e5), 5e9, 5e-9))
n = 15
ma = function(x, n, na.rm=FALSE) {
  ans = rep(NA_real_, nx<-length(x))
  for (i in n:nx) ans[i] = mean(x[(i-n+1):i], na.rm=na.rm)
  ans
}
fastma = function(x, n, na.rm) {
  if (!missing(na.rm)) stop("NAs are unsupported, wrongly propagated by cumsum")
  cs = cumsum(x)
  scs = shift(cs, n)
  scs[n] = 0
  as.double((cs-scs)/n)
}
ans1 = ma(x, n)
ans2 = fastma(x, n)
ans3 = frollmean(x, n)
ans4 = frollmean(x, n, algo="exact")
anserr = list(
  fastma = ans2-ans1,
  froll_fast = ans3-ans1,
  froll_exact = ans4-ans1
)
errs = sapply(lapply(anserr, abs), sum, na.rm=TRUE)
if (!(.Platform$OS.type=="windows" && getDTthreads()>1L)) { # windows 2+ threads rounding issue: #3346
  if (.Machine$sizeof.longdouble == 16L) test(9999.024, errs[["froll_exact"]]==0) # only where long double available, otherwise we get noLD CRAN note
  if (Sys.info()["machine"] == "x86_64") test(9999.025, errs[["froll_fast"]]>errs[["froll_exact"]]) # floating point arithmetic issue on various machines #3491
}
test(9999.026, errs[["fastma"]]>errs[["froll_exact"]])
test(9999.027, errs[["fastma"]]>errs[["froll_fast"]])
```

mattdowle:
@jangorecki In #3492 they tested 1.12.2 and reported a total of 2 tests fail. Which means 9999.024 is running and passing.  This PR switches it off in 1.12.4 which they haven't tried yet.

QuLogic:
Yes, that's correct; I've tested 1.12.2. I did not test 1.12.4 because that doesn't exist yet.

PR #3518, opened 2019-04-22
`between` verbose, POSIXct support, RHS dot-alias, int64 support

### LINKED ISSUE #2315: dot list alias doesn't work with `between`
From `?between`:
    
    X = data.table(a = 1:5, b = 6:10, c = c(5:1)) 
    X[c %between% list(a, b)]

`list` can't be replaced by `.( )` here: 
    
    X[c %between% .(a, b)]
    # Error in .(a, b) : could not find function "."

### LINKED ISSUE #3516: `between` could produce verbose messages

### LINKED ISSUE #3517: `between` could coerce bounds num to int if possible
Not sure but this might be regression, we did not have `verbose` before so cannot easily confirm.
If a bound arg is numeric, then input is coerced to numeric
```r
between(1:5, 2L, 4, verbose=TRUE)
#between parallel processing of double using closed bounds with recycling took    0.000s
#[1] FALSE  TRUE  TRUE  TRUE FALSE
```
only if all three arguments are integers then processing continue on integers
```r
between(1:5, 2L, 4L, verbose=TRUE)
#between parallel processing of integer with recycling took    0.000s
#[1] FALSE  TRUE  TRUE  TRUE FALSE
```
I think we should handle that and coerce `4` into `4L`. The only  problem I think about is that `lower` and `upper` can be vectors, so checking for `isReallyReal` might be expensive?

### LINKED ISSUE #3519: `between` on POSIXct fallback to slow character field processing
```r
require(data.table)
x = as.POSIXct("2016-09-18 07:00:00") + 0:10*60*15
dn = as.POSIXct('2016-09-18 08:00:00')
up = as.POSIXct('2016-09-18 09:00:00')
between(x, dn, up, incbounds=FALSE, verbose=TRUE)
#between on character field, fallback to slow R routine
```

jangorecki:
closes #3516
closes #3519
closes #2315
closes #3517


R/between.R
```diff
@@ -4,13 +4,18 @@ between <- function(x,lower,upper,incbounds=TRUE,verbose=getOption("datatable.ve
   if (is.logical(lower)) lower = as.integer(lower)   # typically NA (which is logical type)
   if (is.logical(upper)) upper = as.integer(upper)   # typically NA (which is logical type)
   is_strictly_numeric <- function(x) is.numeric(x) && !inherits(x, "integer64")
```
arunsrinivasan: Couldn't you just change `is_strictly_numeric` to `inherits(x, "POSIXct") || (is.numeric(x) && inherits(x, "integer64"))` instead of the `unclass` stuff?

R/between.R
```diff
@@ -1,25 +1,35 @@
 # is x[i] in between lower[i] and upper[i] ?
-between <- function(x,lower,upper,incbounds=TRUE) {
+between <- function(x,lower,upper,incbounds=TRUE,verbose=getOption("datatable.verbose")) {
   if (is.logical(x)) stop("between has been x of type logical")
   if (is.logical(lower)) lower = as.integer(lower)   # typically NA (which is logical type)
   if (is.logical(upper)) upper = as.integer(upper)   # typically NA (which is logical type)
   is_strictly_numeric <- function(x) is.numeric(x) && !inherits(x, "integer64")
+  if (inherits(x, "POSIXct") && inherits(lower, "POSIXct") && inherits(upper, "POSIXct")) { #3519
```
MichaelChirico: can we add logic for `!is.na(as.POSIXct(lower))` as well?

I do worry about timezones silently mismatching...

R/between.R
jangorecki: @arunsrinivasan probably I could, but I got confused why int64 has been explicitly excluded from that, shouldn't it just work same as numeric? or unhandled int64 NA is a problem here?

R/between.R
jangorecki: Only if `x` is already POSIXct, then yes, make sense. But as you noted the tz might be an issue.

MichaelChirico: Yep. How about `if (inherits(x, 'POSIXct') && !is.null(attr(x, 'tzone')))`, we pass the existing `tzone` to `lower`/`upper`.

jangorecki: @MichaelChirico sounds fine, you can push to this branch :)

MichaelChirico: Pushed a commit

inst/tests/tests.Rraw
```diff
@@ -14071,9 +14071,20 @@ dn = as.POSIXct('2016-09-18 08:00:00')
 up = as.POSIXct('2016-09-18 09:00:00')
 test(2028.11, between(x, dn, up, verbose=TRUE), output="between parallel processing of double using closed bounds with recycling took")
 test(2028.12, between(x, dn, up, incbounds=FALSE, verbose=TRUE), output="between parallel processing of double using open bounds with recycling took")
+
+## also handling of string lower/upper bounds _only when x has a time zone_
```
jangorecki: Actually this should work for arbitrary class, not just character.

MichaelChirico: My first concern is whether this will produce confusing warnings/errors or some classes, but I guess that's just a matter of getting an error at `x>=lower & x<=upper` in the non-parallel branch or getting it at `as.POSIXct`.

Second concern is for `inherits(lower 'POSIXct')` with mismatched timezones. I've always found it odd that `as.POSIXct(.POSIXct(0, tz = 'UTC'), tz = 'US/Arizona')` will ignore the new timezone. But, I guess that doesn't matter since it doesn't change the numeric representation anyway.

R/between.R
arunsrinivasan: Because the code I wrote didn't take care of int64s.

R/between.R
```diff
@@ -4,22 +4,45 @@ between <- function(x,lower,upper,incbounds=TRUE) {
   if (is.logical(lower)) lower = as.integer(lower)   # typically NA (which is logical type)
   if (is.logical(upper)) upper = as.integer(upper)   # typically NA (which is logical type)
   is_strictly_numeric <- function(x) is.numeric(x) && !inherits(x, "integer64")
+  try_posix_cast <- function(x, tz) {
+    x_time = as.POSIXct(x, tz = tz)
+    if (is.na(x_time)) return(x)
```
mattdowle: This line is missing coverage according to codecov.  Needs to be covered before merge please.

R/between.R
```diff
@@ -4,22 +4,45 @@ between <- function(x,lower,upper,incbounds=TRUE) {
...
+    return(x_time)
+  }
+  if (inherits(x, "POSIXct")) {
+    # allow fast between on POSIX vs non-explicit-POSIX in some circumstances
+    #   (biggest worry is hard-to-predict timezone mismatch)
+    if (!is.null(tz <- attr(x, 'tzone'))) {
+      lower = try_posix_cast(lower, tz)
+      upper = try_posix_cast(upper, tz)
```
mattdowle: Rather than casting at R level, couldn't between.c handle both integer64 and POSIXct at C level.  My first thought is that if x, upper and lower aren't all POSIXct and all in the same timezone, there should at least be a warning.

jangorecki: Could be, but handling character on input of `Cbetween` will significantly increase its code complexity. I think it make sense to leave coercion (at least from character type) in R. We need to wrap it with `tryCatch`, as you noticed in another comment, there are no tests for failed coercion of character to POSIXct.

R/between.R
jangorecki: `unclass` removed, instead of `is_strictly_numeric` we now have
```r
is.supported = function(x) is.integer(x) || (storage.mode(x)=="double" && !inherits(x, "integer64"))
```
will try to add int64 support as AFAIK it should be only a matter of handling NAs.

inst/tests/tests.Rraw
jangorecki: Errors would be definitely confused, there was no tests for that so you haven't noticed. I added more robust coercion, and error handling. Timezone mismatch is also a little bit more robust now. Tests will follow, want to check coverage report first.

R/between.R
jangorecki: I added error, rather than warning. If we want to have warning instead I would prefer to drop timezone from all arguments, warn, and proceed. But error should be OK, tz mismatch should be aligned on user side.

jangorecki: int64 support is still TODO

R/between.R
jangorecki: tests for exceptions added

MichaelChirico: thanks @jangorecki. anything else needed for the code i added?

jangorecki:
still adding int64 support, and better coercion for numeric https://github.com/Rdatatable/data.table/issues/3517

src/forder.c
```diff
@@ -1275,9 +1275,10 @@ SEXP isReallyReal(SEXP x) {
   // which needed a repeat of the argument. Hence simpler and more robust to return 0 when not type double.
   if (isReal(x)) {
     int n=length(x), i=0;
+    double *dx = REAL(x);
```
jangorecki: This is likely to speed up code in few other places, sometimes speed up might be significant.

src/between.c
```diff
@@ -1,8 +1,24 @@
 #include "data.table.h"
 
+#define NA_INTEGER64 LLONG_MIN
+#define MAX_INTEGER64 LLONG_MAX
+
+bool isReallyRealC(SEXP x) {
```
jangorecki: I wrote new `isReallyRealC` because I don't need allocating SEXP and I couldn't rewrite existing one because it returns location of first really real value, which is not what I need.
Ideally we could rename existing `isReallyReal` to `whichFirstReallyReal` and new `isReallyRealC` to `isReallyReal`. It will touch few other files so might be better to do it outside of this PR.

PR #3525, opened 2019-04-26
froll uses deferred verbose messages, closes #3423

### LINKED ISSUE #3423: print uint64 using inttypes.h `PRIu64` instead of `llu`
`%llu` format works fine when using from R C API functions like `Rprintf`, but it fails on Windows when using in plain C like `sprintf`. We could use `PRIu64` instead, defined in `inttypes.h`.
According to http://r.789695.n4.nabble.com/package-config-file-for-windows-td921218.html `inttypes.h` is available on CRAN's Windows machines.
According to https://github.com/pocoproject/poco/pull/1426 `inttypes.h` is available on Solaris.

We need this to be able to construct a message from plain C code, where message body contains length of a vector, thus can be int64. This is used for deferred error/warnings/verbose messages from code that runs in parallel regions.

jangorecki:
closes #3423
unfortunately starting from gcc-8 there are compilation warnings (related [SO](https://stackoverflow.com/a/50509459/2490497))
```
froll.c:16:38: warning: passing argument 1 to restrict-qualified parameter aliases with argument 3 [-Wrestrict]
     if (verbose) sprintf(ans->message[0], "%s%s: window width longer than input vector, returning all NA vector\n", ans->message[0], __func__);
                          ~~~~~~~~~~~~^~~                                                                            ~~~~~~~~~~~~~~~
```
those field are defined as
```c
char message[4][4096];
```
and initialised as `{"\0","\0","\0","\0"}`:
```c
((double_ans_t) { .ans=REAL(VECTOR_ELT(ans, i*nk+j)), .status=0, .message={"\0","\0","\0","\0"} });
```

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -1,14 +1,14 @@
 #include<stdint.h>
 
-typedef struct double_ans_t {
+typedef struct double_ans_t { // used in froll
   double *ans;
-  uint8_t status;   // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
-  char message[4][256]; // STDOUT: output, STDERR: message, warning, error
+  uint8_t status;             // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
+  char message[4][4096];      // STDOUT: output, STDERR: message, warning, error
 } double_ans_t;
 
-typedef struct ans_t {
+typedef struct ans_t {        // used in nafill
   int32_t *int_v;
   double *dbl_v;
-  uint8_t status;   // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
-  char message[4][256]; // STDOUT: output, STDERR: message, warning, error
+  uint8_t status;             // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
+  char message[4][256];       // STDOUT: output, STDERR: message, warning, error
 } ans_t;
--- a/src/types.h
+++ b/src/types.h
@@ -5,6 +5,7 @@ typedef struct double_ans_t { // used in froll
   uint8_t status;             // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
   char message[4][4096];      // STDOUT: output, STDERR: message, warning, error
 } double_ans_t;
+// implicit n_message limit discussed here: https://github.com/Rdatatable/data.table/issues/3423#issuecomment-487722586
 
 typedef struct ans_t {        // used in nafill
   int32_t *int_v;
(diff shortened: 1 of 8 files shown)


mattdowle:
I resolved the problems.  Explanation here:  https://github.com/Rdatatable/data.table/issues/3423#issuecomment-487204046

src/types.h
```diff
@@ -1,14 +1,16 @@
...
+// implicit n_message limit discussed here: https://github.com/Rdatatable/data.table/issues/3423#issuecomment-487722586
 
-typedef struct ans_t {
+typedef struct ans_t {        // used in nafill
   int32_t *int_v;
   double *dbl_v;
-  uint8_t status;   // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
-  char message[4][256]; // STDOUT: output, STDERR: message, warning, error
+  uint8_t status;             // 0:ok, 1:message, 2:warning, 3:error; unix return signal: {0,1,2}=0, {3}=1
+  char message[4][256];       // STDOUT: output, STDERR: message, warning, error
```
mattdowle: Should this 256 in `ans_t` be 4096 too, consistent with `double_ans_t`?  The discussion in #3423 is based on `message` allocation of 4096.  The `snprintf` limit is 500 and could feasibly overwrite 256.

jangorecki: `ans_t` is currently used only in `nafill.c`, and only single messages are being written there, thus 256 is fine there: https://github.com/Rdatatable/data.table/blob/master/src/nafill.c#L61 (todo `sprintf->snprintf`)
but
we can/**should** drop `double_ans_t` and use `ans_t` instead if **we agree** on the overhead of creating extra integer pointer - `ans_t` creates both double and int pointers. I now think it is very safe, `ans_t` refers to column-rolling_window combination, so unlikely to scale more than few hundreds.

I can follow up in this PR. Change `double_ans_t` to `ans_t`.

jangorecki:
Looks fine
```sh
Rscript -e 'library(data.table); d=data.table(a=1:10); out=capture.output(ans<-frollmean(d, rep(2:6, 500), verbose=TRUE)); cat(c(head(out, 50), tail(out, 50)), sep="\n")'
```
```
frollfunR: allocating memory for results 1x2500
frollfunR: 1 column(s) and 2500 window(s), if product > 1 then entering parallel execution
frollfunR: 1:
frollmeanFast: running for input length 10, window 2, hasna 0, narm 0
frollmean: processing algo 0 took 0.000s
frollfunR: 2:
frollmeanFast: running for input length 10, window 3, hasna 0, narm 0
...
frollfunR: 2499:
frollmeanFast: running for input length 10, window 5, hasna 0, narm 0
frollmean: processing algo 0 took 0.000s
frollfunR: 2500:
frollmeanFast: running for input length 10, window 6, hasna 0, narm 0
frollmean: processing algo 0 took 0.000s
frollfunR: processing of 1 column(s) and 2500 window(s) took 0.191s
```
calculate 2500 iterations, (parallel using 2 threads) is under 0.2s, overhead is verbose is pretty low.

PR #3536, opened 2019-05-03
Fwrite csvy

### LINKED ISSUE #3534: Add CSVY support for fwrite
#1701 is done, but if we really want to be fully supportive of CSV+metadata we'll need to support it in `fwrite` as well.

Basic idea will be to use `yaml::write_yaml` + `fwrite(append = TRUE)`. Will make `yaml && append` an error.

MichaelChirico:
Closes #3534

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,8 @@
     fwrite(DT, "data.csv.gz")   #   2MB; 1.6s
     identical(fread("data.csv.gz"), DT)
     ```
+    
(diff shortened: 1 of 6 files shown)


MichaelChirico:
Hmm not sure why the failure on AppVeyor... must be something to do with `eol`...

MichaelChirico:
ok fixed it... I'll add a \r\n eol test when I'm back to my machine

MichaelChirico:
OK, ready to merge from my side, pending approval

PR #3539, opened 2019-05-04
de-duplicate test numbers (including numerically equal)

### LINKED ISSUE #3538: duplicate test numbers
There are some duplicate test numbers in `tests.Rraw`:

```
tests = readLines('inst/tests/tests.Rraw')
# not all tests start at line beginning,
#  some are commented,
#  ignore those in loops by requiring explicit numbering;
#  use , matching for semi-manual numbering like 123.4+i/1000 in a loop
tests = grep('^[^#]*test\\(\\s*[0-9.]+,', tests, value = TRUE)
tests = gsub('.*test\\(([0-9.]+),.*', '\\1', tests)
# test numbers are interpreted as numbers
#   so 17.1 and 17.10 and 17.100 should be considered the same
tests[duplicated(as.numeric(tests))]
```

These are duplicated:

```
 [1] "658.3"   "677.2"   "1039"    "1102.14" "1105"    "1134.3"  "1144.1" 
 [8] "1144.2"  "1144.3"  "1144.4"  "1355"    "1463.33" "1463.34" "1463.27"
[15] "1543.7"  "1545.2"  "1557.1"  "1557.2"  "1557.3"  "1557.4"  "1571.1" 
[22] "1571.2"  "1571.3"  "1571.5"  "1573"    "1574.1"  "1574.2"  "1574.3" 
[29] "1574.4"  "1579.6"  "1579.7"  "1579.8"  "1579.9"  "1579.10" "1579.11"
[36] "1579.12" "1579.13" "1626.63" "1655.4"  "1658.5"  "1658.23" "1658.24"
[43] "1658.41" "1658.42" "1658.43" "1686.1"  "1701.6"  "1714.1"  "1739.1" 
[50] "1839.6"  "1871.2"  "1962.41" "1962.41" "1992.1" 
```

investigating & de-duplicating...

^ there were about 2x as many when looking for numerical duplicates; those are all fixed in the PR.

MichaelChirico:
Closes #3538

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -1869,7 +1869,7 @@ test(654, ans1, ans2)
 options(datatable.optimize = 0L)
 test(656.1, DT[ , mean(x), by=grp1, verbose=TRUE], output='(GForce FALSE)')
 test(657.1, DT[ , list(mean(x)), by=grp1, verbose=TRUE], output="(GForce FALSE)")
-test(658.3, DT[ , list(mean(x), mean(y)), by=grp1, verbose=TRUE], output="(GForce FALSE)")
+test(658.1, DT[ , list(mean(x), mean(y)), by=grp1, verbose=TRUE], output="(GForce FALSE)")
 options(datatable.optimize = 1L)
 test(656.2, DT[ , mean(x), by=grp1, verbose=TRUE], output='(GForce FALSE)')
 test(657.2, DT[ , list(mean(x)), by=grp1, verbose=TRUE], output="(GForce FALSE)")
@@ -1934,7 +1934,7 @@ l = list(data.table(a=1:2, b=7:8),
 test(676.1, rbindlist(l[1:3]), ans<-data.table(a=1:6,b=7:12), message="Column 2 [[]'baz'[]] of item 2 is missing in item 1.*Use fill=TRUE.*or use.names=FALSE")
 test(676.2, rbindlist(l[1:3], use.names=FALSE), ans)
 test(677.2, rbindlist(l[c(10,1,10,2,10)]), ans<-data.table(a=1:4,b=7:10), message="Column 2 [[]'baz'[]] of item 4 is missing in item 2")  # NULL items ignored
-test(677.2, rbindlist(l[c(10,1,10,2,10)], use.names=FALSE), ans)
+test(677.3, rbindlist(l[c(10,1,10,2,10)], use.names=FALSE), ans)
 test(678, rbindlist(l[c(1,4)]), error="Item 2 has 1 columns, inconsistent with item 1 which has 2")
 test(679.1, rbindlist(l[c(1:2,5)]), ans<-data.table(a=c(1:4,15:16), b=c(7:10,17L,17L)), message="Column 2 [[]'baz'[]] of item 2 is missing in item 1")
 test(679.2, rbindlist(l[c(1:2,5)], use.names=FALSE), ans)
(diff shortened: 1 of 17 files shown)


inst/tests/tests.Rraw
```diff
@@ -1934,7 +1934,7 @@ l = list(data.table(a=1:2, b=7:8),
 test(676.1, rbindlist(l[1:3]), ans<-data.table(a=1:6,b=7:12), message="Column 2 [[]'baz'[]] of item 2 is missing in item 1.*Use fill=TRUE.*or use.names=FALSE")
 test(676.2, rbindlist(l[1:3], use.names=FALSE), ans)
 test(677.2, rbindlist(l[c(10,1,10,2,10)]), ans<-data.table(a=1:4,b=7:10), message="Column 2 [[]'baz'[]] of item 4 is missing in item 2")  # NULL items ignored
-test(677.2, rbindlist(l[c(10,1,10,2,10)], use.names=FALSE), ans)
+test(677.3, rbindlist(l[c(10,1,10,2,10)], use.names=FALSE), ans)
```
QuLogic: Based on the above few lines, shouldn't this still be 677.2 and the one above 677.1 (i.e., these look like variations of 676.*)?

MichaelChirico: Indeed! Thanks for the extra eyes.

Main goal of this PR is that if test `x` fails, you can <kbd>Ctrl</kbd> + <kbd>F</kbd> for `x` and find the right test, rather than to clean up all the tests (which would be a mammoth undertaking)

mattdowle:
@MichaelChirico  I see the duplicate fixed and formatting of numbers ok, but what are the blocks of new/moved tests for?  Seeing as it's over 700 lines changed.

MichaelChirico:
those got "bumped" to the bottom of the stack, e.g. if there's tests

1470
1471
1471
1472
1473
...

the duplicate 1471 is pulled out and added to the bottom

IIUC this happened because of (1) some slips on merging concurrent PRs or
(2) some melt tests were moved around and as a result their numbers got
repeated later?

inst/tests/tests.Rraw
```diff
@@ -3737,23 +3737,6 @@ test(1144.4, DT[, list(lx=x[.N]), by=x], data.table(x=DT$x, lx=DT$x))
 ans = copy(DT)
 test(1144.3, DT[,`:=`(lx=tail(x,1L)), by=y], ans[, lx := x])
 
-# FR #2356 - retain names of named vector as column with keep.rownames=TRUE
-x <- 1:5
-setattr(x, 'names', letters[1:5])
-test(1144.1, as.data.table(x, keep=TRUE), data.table(rn=names(x), x=unname(x)))
```
MichaelChirico: @mattdowle e.g. this one, see directly above is `test(1144` for a completely different issue and directly below is `test(1145`.

Bumping 1145 to the end would have less of a diff I guess?

mattdowle: Nice find. Seems like the original merge was half-right in that there's a gap for 1143 between 1142 and the 1st 1144. So it'd be better to bump the first 1144.* back to 1143 and then leave the 2nd 1144.* where it is.

inst/tests/tests.Rraw
```diff
@@ -9752,11 +9673,6 @@ test(1685, grep("dtu", capture.output(foo())), 7L)
 test(1686.1, uniqueN(1L), 1L)
```
mattdowle: iiuc, if this 1686.1 was changed to 1685.1, the 4 lines just below (1681.1 and 1686.2) wouldn't need to change or move?

inst/tests/tests.Rraw
```diff
@@ -7617,39 +7600,7 @@ d = as.IDate(seq(as.Date("2015-01-01"), as.Date("2015-01-15"), by='1 day'))
...
-i = 2L
-test(1557.1, dt[, paste0("index", 1:i), with=FALSE], dt[, index1:index2, with=FALSE])
-test(1557.2, dt[, paste0("index", 1:i), with=FALSE], dt[, 1:2, with=FALSE])
-test(1557.3, dt[, 5:4, with=FALSE], dt[, i:s, with=FALSE])
-test(1557.4, dt[, .SD, .SDcols=paste0("index", 1:i)], dt[, .SD, .SDcols=index1:index2])
-
-# fix for #1354
-test(1558, as.ITime(NA), setattr(NA_integer_, 'class', 'ITime'))
-
-# fix for #1352
```
mattdowle: aside:  note how less-useful it is when just a number is used in the comment, rather than writing a few words about what's being tested too.

inst/tests/tests.Rraw
MichaelChirico: Oh, good catch, I didn't plumb around for gaps too much. Will do & squash.

inst/tests/tests.Rraw
MichaelChirico: seems right. can re-run the dup check to be sure.

MichaelChirico:
@mattdowle updated, though the diff didn't shrink much. Shall I try and squash all the commits?

mattdowle:
@MichaelChirico thanks. It's ok - I'll squash on merge. I'd prefer to not have the tests moved to the end if possible. Can they be left in-place but fitted into the id numbering, somehow. Maybe the fmelt section could be renumbered.  It's not the number of lines that's troubling - that's looking good with lots of small sections where we can see that just the number changed.

MichaelChirico:
> it's harder to check nothing else changed than just the test number.

Not sure I follow / don't quite get the end goal... adding a few more comments, could help to clarify

mattdowle:
When a block of 10 lines moves to the end I see 10 lines deleted and 10 lines added. I can't be sure that nothing else inadvertently changed within those 10 lines.  But when the old and new lines are together in the diff, I look at the highlighted differences within the lines, and also I compare line by line the old and new blocks to make sure no lines added/deleted.  I can't do that when the block moves a long distance in the file.

MichaelChirico:
OK, gotcha. A sort of future-proofing for historical purpose of revisiting this PR/commit later 👍

MichaelChirico:
working on a fix now

MichaelChirico:
@mattdowle done

mattdowle:
> A sort of future-proofing for historical purpose of revisiting this PR/commit later
Exactly. If in future we ever find ourselves looking back and wondering if this PR removed/changed a test by accident, then we can rule out this PR more easily.

Issue #2799, opened 2018-04-27
fastmean optimization is not used


jangorecki:
Optimization of `mean` to `fastmean` is not mentioned when using `verbose=TRUE`. AFAIU following call should be optimized
```r
 data.table(x=c(1:5,NA,6:9))[, mean(x, na.rm=TRUE), verbose=TRUE]
#Detected that j uses these columns: x 
#[1] 5
```
In the above example `[.data.table` processing is terminated at this point, so `.optmean` function doesn't run at all. https://github.com/Rdatatable/data.table/blob/65fa5c0adbab3bb900c07581a1333e7b46ca5b9c/R/data.table.R#L1403
Minor note on `.optmean`,.Detection of `na.rm` argument is made as follows:
```r
if (length(expr)==3L && identical("na",substring(names(expr)[3L], 1L, 2L)))
```
First two first letter of `na.rm` are required, while base `mean` call will work just with first letter:
```r
mean(c(1,NA,3), n=T)
#[1] 2
```

mattdowle:
Also fastmean.c has R API usage inside loops which could be taken outside.
We're only dispatching to fastmean in the no-grouping case.  Ideally we could remove `fastmean`, but base R still does the `x[!is.na(x)]` step first.  Maybe a short-circuit could be added to gforce for the no-grouping case.

Issue #1160, opened 2015-05-23
Inconsistently dropping class/attributes of column vector


renkun-ken:
I'm creating [a package](https://github.com/renkun-ken/formattable) to make formattable data. It works nice with `data.table` in most cases, but in the following cases, `data.table` somehow shows inconsistent behavior:

```r
library(data.table)
library(formattable)
p <- data.table(
  id = c(1, 2, 3, 4, 5), 
  name = c("A1", "A1", "B1", "B1", "C1"),
  balance = accounting(c(52500, 36150, 25000, 18300, 7600), format = "d"),
  growth = percent(c(0.3, 0.3, 0.1, 0.15, 0.15), format = "d"),
  ready = formattable(c(TRUE, TRUE, FALSE, FALSE, TRUE), "yes", "no"))
```

Then `p` looks like

```
   id name balance growth ready
1:  1   A1  52,500    30%   yes
2:  2   A1  36,150    30%   yes
3:  3   B1  25,000    10%    no
4:  4   B1  18,300    15%    no
5:  5   C1   7,600    15%   yes
```

Then I do some aggregation:

```rconsole
> p[, .(balance = mean(balance)), by = .(name)]
   name balance
1:   A1  44,325
2:   B1  21,650
3:   C1   7,600
> p[, .(balance = mean(balance), growth = min(growth)), by = .(name)]
   name balance growth
1:   A1  44,325    30%
2:   B1  21,650    10%
3:   C1   7,600    15%
> p[, .(balance = mean(balance), growth = min(growth), ready = all(ready)), by = .(name)]
   name balance growth ready
1:   A1   44325    30%   yes
2:   B1   21650    10%    no
3:   C1    7600    15%   yes
> p[, .(balance = mean(balance), growth = min(growth)), by = .(name)]
   name balance growth
1:   A1  44,325    30%
2:   B1  21,650    10%
3:   C1   7,600    15%
> p[, .(balance = mean(balance), growth = min(growth), ready = all(ready)), by = .(name)]
   name balance growth ready
1:   A1   44325    30%   yes
2:   B1   21650    10%    no
3:   C1    7600    15%   yes
> p[, .(balance = last(balance), growth = min(growth), ready = all(ready)), by = .(name)]
   name balance growth ready
1:   A1  36,150    30%   yes
2:   B1  18,300    10%    no
3:   C1   7,600    15%   yes
[...]

PR #3546, opened 2019-05-06
GForce shutoff for non-standard classes

### LINKED ISSUE #1876: combining `mean` with reformatted `Date` class and `by` throwing an error
I've come across an issue using a formatted date column with `by`. Take the following data:

```
library( data.table )
data <- data.table( session = c( 1,1,1,1,2,2,2,2,2,2,3,3,3,3 ),
                    date = as.Date( c( "2016-01-01", "2016-01-02", "2016-01-03", "2016-01-03",
                                       "2016-04-30", "2016-04-30", "2016-05-03", "2016-05-03", "2016-05-03", "2016-05-03",
                                       "2016-08-28", "2016-08-28", "2016-08-28", "2016-08-28" ) )
)
```

I want to mark each session with a label, formatted `%b-%Y`, based on the mean date for that session.

I can find the mean date of each session, using the `by` parameter:

```
output <- copy( data )[ , Month := mean( date ), by = session ]
```

I can also reformat a mean date the way I want within `data.table`:

```
output <- copy( data )[ , Month := format( mean( date ), "%b-%Y" ) ]
```

But I can't do both:

```
output <- copy( data )[ , Month := format( mean( date ), "%b-%Y" ), by = session ]
```

The above returns an error:
`
    Error in prettyNum(.Internal(format(x, trim, digits, nsmall, width, 3L,  : 
    invalid 'trim' argument
    In addition: Warning message:
    In mean(date) : argument is not numeric or logical: returning NA
`

Note I can do what I need in two steps (below), and it works OK, but there does seem to be something going wrong with the above :

```
output <- copy( data )[ , Month := mean( date ), by = session 
                        ][ , Month := format( Month, "%b-%Y" ) ]
```

It also works fine with `mean.Date`:

```
output <- copy( data )[ , Month := format( mean.Date( date ), "%b-%Y" ), by = session ]
```

### LINKED ISSUE #3079: Date and POSIXct coerced to numeric when calculating median by group
I have data with class `Date` ('date') and `POSIXct` ('time'), and a grouping variable 'g'

    d <- data.table(
       date = as.Date(c("2018-01-01", "2018-01-03", "2018-01-08",
                     "2018-01-10", "2018-01-25", "2018-01-30")),
       g = rep(letters[1:2], each = 3))
    d[ , time := as.POSIXct(date)]

When calculating median of 'date' and 'time' by group, the result is coerced to numeric:

    d[ , median(date), by = g]
    #    g    V1
    # 1: a 17534
    # 2: b 17556

    d[ , median(time), by = g]
    #    g         V1
    # 1: a 1514937600
    # 2: b 1516838400

------

However, 'date' and 'time' is _not_ coerced when calculating median _without_ grouping:

    d[ , median(date)]
    # [1] "2018-01-09"

    d[ , median(time)]
    # [1] "2018-01-09 01:00:00 CET"

------
Other things I've tried which don't coerce:

Mean 'date' and 'time' by group:

    d[ , mean(date), by = g]
    #    g         V1
    # 1: a 2018-01-04
    # 2: b 2018-01-21

    d[ , mean(time), by = g]
    #    g                  V1
    # 1: a 2018-01-04 01:00:00
    # 2: b 2018-01-21 17:00:00

Median 'date' and 'time' by group using `aggregate`:

    aggregate(date ~ g, data = d, median)
    #   g       date
    # 1 a 2018-01-03
    # 2 b 2018-01-25

    aggregate(time ~ g, data = d, median)
    #   g                time
    # 1 a 2018-01-03 01:00:00
    # 2 b 2018-01-25 01:00:00

-----

R version 3.5.1 (2018-07-02)
Running under: Windows Server >= 2012 x64 (build 9200)
data.table_1.11.6

### LINKED ISSUE #3533: Operations with "by" do not dispatch to S3 methods
When using `by`, it appears that data.table is not able to dispatch S3 methods. The closest similar issue I could find was with this old SO one which was fixed in v1.9.6: https://stackoverflow.com/questions/33177009/s3-method-dispatch-in-data-table-when-using-by-clause. In this case, however, although it DOES maintain the class of the column being operated on, it does not dispatch the correct S3 method unless that is explicitly provided. In the reprex below, `direction` is of class `circular` and `mean(c(20, 350))` should give 5, not 185. This works without `by` or when using `by` and explicitly calling `circular::mean.circular`, but not when using `by` and plain old `mean`:

``` r
library(data.table)

dt <- data.table(
  index = c("A", "A"),
  direction = circular::circular(
    c(20, 350),
    template = 'geographics',
    modulo = '2pi',
    units = 'degrees'
  )
)
dt
#>    index direction
#> 1:     A        20
#> 2:     A       350
dt[, mean(direction)]  # correct
#> Circular Data: 
#> Type = angles 
#> Units = degrees 
#> Template = geographics 
#> Modulo = 2pi 
#> Zero = 1.570796 
#> Rotation = clock 
#> [1] 5
dt[, mean(direction), by = index]$V1 # wrong!
#> Circular Data: 
#> Type = angles 
#> Units = degrees 
#> Template = geographics 
#> Modulo = 2pi 
#> Zero = 1.570796 
#> Rotation = clock 
#> [1] 185
dt[, circular::mean.circular(direction), by = index]$V1 # correct
#> Circular Data: 
#> Type = angles 
#> Units = degrees 
#> Template = geographics 
#> Modulo = 2pi 
#> Zero = 1.570796 
#> Rotation = clock 
#> [1] 5

sessionInfo()
#> R version 3.5.2 (2018-12-20)
#> Running under: Ubuntu 16.04.5 LTS
#> 
#> Matrix products: default
#> BLAS: /usr/lib/libblas/libblas.so.3.6.0
#> LAPACK: /usr/lib/lapack/liblapack.so.3.6.0
#> 
#> locale:
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.12.2
#> 
#>  [1] Rcpp_1.0.1      mvtnorm_1.0-10  digest_0.6.18   magrittr_1.5   
#>  [5] evaluate_0.13   highr_0.8       stringi_1.4.3   boot_1.3-20    
#>  [9] rmarkdown_1.12  tools_3.5.2     stringr_1.4.0   xfun_0.6       
#> [13] yaml_2.2.0      compiler_3.5.2  circular_0.4-93 htmltools_0.3.6
#> [17] knitr_1.22

### PR DESCRIPTION
**Pull request by MichaelChirico**
Closes #3079
Closes #1876 
Closes #3533

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -14370,6 +14370,17 @@ if (test_yaml) {  # csvy; #1701
        DT_yaml, warning = 'Combining a search.*YAML.*')
 }
 
+# #3533, #1876, #3079 -- ensure GForce/fastmean are not applied incorrectly under method dispatch
+DT = data.table(ID = 1:2, v = 3:4, d = .Date(1:2))
+class(DT$v) = 'mymean'
+mean.mymean = function(x) min(x) + 1
+test(2033.1, DT[ , mean(v)], 4)
+options(datatable.optimize = 0)
+trueDT = data.table(ID = 1:2, V1 = c(4, 5))
+test(2033.2, DT[ , mean(v), by = ID], trueDT)
+options(datatable.optimize = 2)
+test(2033.3, DT[ , mean(v), by = ID, verbose = TRUE], trueDT, output = 'GForce FALSE')
+
 
 ###################################
 #  Add new tests above this line  #
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -4105,8 +4105,8 @@ test(1196.2, DT[,.N,by=val]$N, INT(1,1,2))
 test(1197.2, DT[.(x),.N], 2L)
 setNumericRounding(old_rounding)
 
-DT = data.table(id=1:2, val1=6:1, val2=6:1)   # 5380
-test(1199, DT[, sum(.SD), by=id], error="GForce sum can only be applied to columns, not .SD or similar.*looking for.*lapply\\(.SD")
+#DT = data.table(id=1:2, val1=6:1, val2=6:1)   # 5380
+#test(1199, DT[, sum(.SD), by=id], error="GForce sum can only be applied to columns, not .SD or similar.*looking for.*lapply\\(.SD")
 
 # Selection of columns, copy column to maintain the same as R <= 3.0.2, in Rdevel, for now
 # Otherwise e.g. setkey changes the original columns too.  TO DO: could allow shallow copy, perhaps.
@@ -7039,6 +7039,11 @@ test(1521, x[, b := 5], data.table(a=c(1,2), b=5))
 
 # Fix for #1160, fastmean retaining attributes
 x = data.table(a = c(2,2,1,1,2), b=setattr(1:5, 'class', c('bla', 'integer')))
+mean.bla = function(x) {
+  mu = mean.default(x)
+  class(mu) = class(x)
+  mu
+}
 test(1522, class(x[, .(mean(b), all(b)), by=a]$V1), c('bla', 'integer'))
 
 # Fix for #1145, .N lock handled properly
@@ -9617,6 +9622,7 @@ dt = data.table(x=1, y=2)
 test(1671, dt[, z := sd, by=x], error="invalid type/length (closure/1)")
 
 # 1683
+options(datatable.optimize = Inf)
 DT <- data.table(V1 = rep(1:2, 3), V2 = 1:6)
 test(1672.1, DT[ , .(.I[1L], V2[1L]), by = V1],
      output = "   V1 V1 V2\n1:  1  1  1\n2:  2  2  2")
--- a/NEWS.md
+++ b/NEWS.md
@@ -77,6 +77,8 @@
 
 5. `fread()` could crash if `quote=""` (i.e. ignore quotes), the last line is too short, and `fill=TRUE`, [#3524](https://github.com/Rdatatable/data.table/pull/3524). Thanks to Jiucang Hao for the report and reproducible example.
 
+6. `data.table` grouping optimization (`GForce` & `fastmean`) ignored method dispatch and could give incorrect results for non-simple classes, [#3533](https://github.com/Rdatatable/data.table/issues/3533), [#1876](https://github.com/Rdatatable/data.table/issues/1876), and [#3079](https://github.com/Rdatatable/data.table/issues/3079). Thanks to @d-sci, @rossholmberg, and @Henrik-P, respectively, for reporting.
+
 #### NOTES
 
 1. `rbindlist`'s `use.names="check"` now emits its message for automatic column names (`"V[0-9]+"`) too, [#3484](https://github.com/Rdatatable/data.table/pull/3484). See news item 5 of v1.12.2 below.
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1725,7 +1725,7 @@ replace_dot_alias <- function(e) {
         .gforce_ok <- function(q) {
           if (dotN(q)) return(TRUE) # For #5760
           # Need is.symbol() check. See #1369, #1974 or #2949 issues and explanation below by searching for one of these issues.
-          cond = is.call(q) && is.symbol(q[[1]]) && (q1c <- as.character(q[[1]])) %chin% gfuns && !is.call(q[[2L]]) && is_strictly_numeric(eval(q[[2L]], x, parent.frame()))
+          cond = is.call(q) && is.symbol(q[[1]]) && (q1c <- as.character(q[[1]])) %chin% gfuns && !is.call(q[[2L]]) && is_vector_extended(eval(q[[2L]], x, parent.frame()))
           # run GForce for simple f(x) calls and f(x, na.rm = TRUE)-like calls
           ans = cond && (length(q)==2L || identical("na",substring(names(q)[3L], 1L, 2L))) && (!q1c %chin% c("head","tail")) # head-tail uses default value n=6 which as of now should not go gforce
           if (identical(ans, TRUE)) return(ans)
@@ -1761,7 +1761,7 @@ replace_dot_alias <- function(e) {
       nomeanopt=FALSE  # to be set by .optmean() using <<- inside it
       oldjsub = jsub
       .fastmean_ok = function(q) {
-        is.call(q) && is.symbol(q[[1L]]) && q[[1L]] == "mean" && is_strictly_numeric(eval(q[[2L]], x, parent.frame()))
+        is.call(q) && is.symbol(q[[1L]]) && q[[1L]] == "mean" && is_vector_extended(eval(q[[2L]], x, parent.frame()))
       }
       if (jsub[[1L]]=="list") {
         fastmean_assigned = FALSE
@@ -1901,7 +1901,7 @@ replace_dot_alias <- function(e) {
 
 # variable is numeric & won't attempt to dispatch to non-standard gfuns element
 #   prevents errors like #3533, #1876, #3079
-is_strictly_numeric = function(x) class(x)[1L] %chin% c('integer', 'integer64', 'numeric')
+is_vector_extended = function(x) is.vector(x) || inherits(x, 'integer64')
 
 .optmean <- function(expr) {   # called by optimization of j inside [.data.table only. Outside for a small speed advantage.
   if (length(expr)==2L)  # no parameters passed to mean, so defaults of trim=0 and na.rm=FALSE
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1764,13 +1764,24 @@ replace_dot_alias <- function(e) {
         is.call(q) && is.symbol(q[[1L]]) && q[[1L]] == "mean" && is_strictly_numeric(eval(q[[2L]], x, parent.frame()))
       }
       if (jsub[[1L]]=="list") {
+        fastmean_assigned = FALSE
         for (ii in seq_along(jsub)[-1L]) {
           this_jsub = jsub[[ii]]
           if (dotN(this_jsub)) next; # For #5760
           # Addressing #1369, #2949 and #1974. Added is.symbol() check to handle cases where expanded function definition is used insead of function names. #1369 results in (function(x) sum(x)) as jsub[[.]] from dcast.data.table.
-          if (.fastmean_ok(jsub)) jsub[[ii]] = .optmean(this_jsub)
+          if (.fastmean_ok(jsub)) {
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by MichaelChirico** [comment]
Two old tests were altered/eliminated here; see comments on each

File: inst/tests/tests.Rraw [hunk 1]
```diff
@@ -4105,8 +4105,8 @@ test(1196.2, DT[,.N,by=val]$N, INT(1,1,2))
 test(1197.2, DT[.(x),.N], 2L)
 setNumericRounding(old_rounding)
 
-DT = data.table(id=1:2, val1=6:1, val2=6:1)   # 5380
-test(1199, DT[, sum(.SD), by=id], error="GForce sum can only be applied to columns, not .SD or similar.*looking for.*lapply\\(.SD")
+#DT = data.table(id=1:2, val1=6:1, val2=6:1)   # 5380
+#test(1199, DT[, sum(.SD), by=id], error="GForce sum can only be applied to columns, not .SD or similar.*looking for.*lapply\\(.SD")
```
MichaelChirico: I'm not sure this ever should have been an `error`... Behavior in this PR is to degrade from trying to `GForce` `sum(.SD)` and erroring to just never trying to `GForce` in the first place.

The error is nice in that it provides feedback that the user may be doing it wrong...

Potentially we could add some more checks for this specific case and continue to pass this test?

inst/tests/tests.Rraw
```diff
@@ -7039,6 +7039,11 @@ test(1521, x[, b := 5], data.table(a=c(1,2), b=5))
 
 # Fix for #1160, fastmean retaining attributes
 x = data.table(a = c(2,2,1,1,2), b=setattr(1:5, 'class', c('bla', 'integer')))
```
MichaelChirico: This test is building explicitly wrong (IMO) behavior by expecting `fastmean` to run on `bla`-class objects.

Essentially it's forcing `mean.default` to be run on all classes & then returning the object with the old class... As we've seen in #3533 etc, this implicit bypassing of the method dispatch can be ill-advised.

IIUC that's why `mean.default` will coerce to `numeric` instead of retaining `c('bla', 'integer')` -- especially the `integer` part feels explicitly wrong. 

```
x = data.table(a = c(2,2,1,1,2), b=1:5)
setattr(x$b, 'class', c('bla', 'integer'))
class(mean(x$b))
# [1] numeric
```

The original issue of #1160 would have been fixed by this PR as well (just not using `fastmean`), so leaving the test in and adjusting by providing a `mean.bla` method. @arunsrinivasan made the original PR, any thoughts?

jangorecki: only that there is one pending issue related to fast mean https://github.com/Rdatatable/data.table/issues/2799

mattdowle:
> As we've seen in #3533 etc, this implicit bypassing of the method dispatch can be ill-advised.

But that issue is very unusual in that mean(c(20,350)) is not 185 but 5 (in the `circular` package). It's long-standing behavior that data.table calculates summary stats (mean, sum, sd, min, max) on the underlying type and then copies the class attributes on to the result. The class attributes are most often used for class-specific formatting. For unusual packages like `circular`, needing to write out the method name `mean.circular(c(20,350))` seems reasonable to me as it reminds the reader of the code that it is not a standard mean.  Either that or turn off optimization when used with `circular`.

The cause of #3079 is we're missing a copyMostAttrib in gmedian which is inconsistent with all the other gforce functions which already do that copy.

MichaelChirico:
I think I disagree pretty strongly here, this leads to bad user experience, and also there's no guarantee for `mean` methods to be exported (we often don't export methods, for example). So

 1. New users have to be thoroughly aware that we're bypassing method dispatch (adversarial coding), and the only reason we're doing so is for speed, accuracy be damned.
 2. We're consigning people to use `:::` which is bad practice and forces users to use undocumented objects in their code, the alternative being that again, they have to know `data.table` front-to-back and be aware to shut off `datatable.optimize`.

I believe the approach here is much more user-friendly -- expert users can `as.numeric` their exotic classes if they really want `GForce` efficiency. And the fact that we have three outstanding related issues on this means that users are getting caught off-guard by the default behavior already.

Is there a compromise approach? Perhaps using this PR but also issuing a `warning` or `verbose` message that efficiency is lost due to class decoration?

mattdowle:
This PR turns off gforce optimization for any object with a class, doesn't it? Just because of one case (`circular`) where the mean is not the mean.

> And the fact that we have three outstanding related issues on this means that users are getting caught off-guard by the default behavior already.

I wrote that the cause of #3079 is we're missing a copyMostAttrib in gmedian which is inconsistent with almost all the other gforce functions which already do that copy.

> accuracy be damned

I think this is a bit unfair to write.

MichaelChirico:
> This PR turns off gforce optimization for any object with a class, doesn't it? Just because of one case (circular) where the mean is not the mean.

Yes, but we also don't really know that `circular` is the only "culprit" (culprit used very loosely -- the point of S3 dispatch is you are allowed to define any mean you'd like and users don't really have to know the implementation details). I see a few `mean` methods exported in my (probably relatively small) `.libPaths()[1L]`:

```
find . -type f -print | grep NAMESPACE | xargs grep "S3method[(]mean,"
./data.table/NAMESPACE:S3method(mean, IDate)
./rlang/NAMESPACE:S3method(mean,quosure)
./units/NAMESPACE:S3method(mean,units)
./circular/NAMESPACE:S3method(mean, circular)
```

`base` also has a few more attached:

```
methods(mean)
# [1] mean.Date     mean.default  mean.difftime mean.POSIXct  mean.POSIXlt 
```

"the mean is not the mean" is not for us to decide. It's very heavy-handed of us to bypass method dispatch and say "your class wanted to do a mean this way, but trust us, the _arithmetic_ mean is what you want". It's easy to imagine creating a class that would default to using geometric or harmonic means, for example, which are perfectly valid "mean"s, statistically speaking.

> I think this is a bit unfair to write.

It was strongly worded & a bit flip, but the kernel holds true. Bypassing method dispatch for speed _automatically_ is saying that accuracy is not paramount; speed is. We spent a lot of effort in #3056 trying to prevent users from potentially getting silently wrong results; does that diligence not apply here?

jangorecki:
Agree with Michael that we should not automatically ignore method dispatch. It will be more safe and still user friendly if we would export `gmean`/`fmean` so user who want to ignore method dispatch can call those directly.

mattdowle:
@jangorecki 

> It will be more safe and still user friendly if we would export gmean/fmean so user who want to ignore method dispatch can call those directly.

You realize this is big change in ethos and will require a lot of people to change. Do I understand your suggestion correctly? How is it still user friendly to require (a lot of) users to start to use new `g` functions explicitly otherwise there could be significant slow downs.

@MichaelChirico 

> Bypassing method dispatch for speed automatically is saying that accuracy is not paramount; speed is. We spent a lot of effort in #3056 trying to prevent users from potentially getting silently wrong results; does that diligence not apply here?

Because nobody has proposed a solution yet that doesn't have a greater cost.  The title of this PR refers to "non-standard" classes, but really it shuts off gforce for any class, for lots of standard classes.

In general, data.table does break some eggs to make an omelette. And it leaves some eggs broken. For example, `:=` is not perfect.  By your argument, we should shut down `:=` too.

jangorecki:
> How is it still user friendly to require (a lot of) users to start to use new g functions explicitly otherwise there could be significant slow downs.

What I meant is that `gmean` would be required only when user want to skip S3 dispatch, normally `mean` will optimise to `gmean` unless there is a "non-standard" class having defined mean method.
Black/white listing is another option as it doesn't come out often.

MichaelChirico:
> lots of standard classes

I'm not seeing the cost that much. We also only have one test in the suite that relies on `GForce` applying to non-`numeric` class. So as much as we don't have ample evidence users are getting bit by current practice, we also don't have much evidence that people are relying on the current behavior. We could potentially run rev-dep on this branch? That would also give empirical weight to whether this change imposes "great cost".

Getting `GForce` to work with this PR is as simple as applying `mean` to the `as.numeric` version of your column, if artihmetic mean is what you intend. If your class is really just decorated-numeric (e.g. something that has nice `print` methods but storage mode is simple `numeric` and `mean.default` applies), `as.numeric` should be almost instant even on large data. 

Another alternative is to use `mean.default` as an entry point for `GForce`, since it is definitely exported -- users could use that to signal it's safe for us to redirect to `GForce` regardless of class.

> How about adding a blacklist of classes for which gforce should be turned off

The worry is still that unaware users can end up with silently wrong results. Whereas with `cedta`, the worst that can happen is some extra printing.

mattdowle:
> We also only have one test in the suite that relies on GForce applying to non-numeric class.

I'm not following what you mean by "non-numeric class". This issue and PR isn't about character data per se, for example.  is.vector() is used in this PR for its feature of considering a vector with a class attribute not be a vector.  Maybe you mean non-vector class rather than non-numeric class.  The fact we're missing tests just shows that we thought one test was sufficient to make sure class attribute was retained by gforce functions. But over time the gforce abilities have been extended and one (gmedian) was missed. So we need more tests.

> We could potentially run rev-dep on this branch? That would also give empirical weight to whether this change imposes "great cost".

revdep wouldn't show anything. revdep is for correctness (so much as revdep package tests cover). But this cost would be a slow-down.

> Getting GForce to work with this PR is as simple as applying mean to the as.numeric version of your column, if artihmetic mean is what you intend.

I'm not following what you mean here. That sounds to me like they would have to know and then do something new manually. Currently gforce is an automatic optimization.  It's not about the compute speed of having to do `as.numeric`,  it's the inconvenience of users having to change their code to write `as.numeric`.

> Another alternative is to use mean.default as an entry point for GForce, since it is definitely exported -- users could use that to signal it's safe for us to redirect to GForce regardless of class.

Again - I don't know what this means. What is an entry point for GForce ?

> The worry is still that unaware users can end up with silently wrong results. 

Not if `circular` is added to a list.  The first known example in 10 years.  If more are discovered, we can think again.

> Whereas with cedta, the worst that can happen is some extra printing.

That's not exactly true.  It's much better now that `DT[,j]` does do what `DF[,j]` does when `j` is a numeric or character vector, but it used to be that `j` was returned straight back and it could be a silent wrong-result.  There are still some wrinkles related to that, but it's much better but not perfect.

MichaelChirico:
> revdep wouldn't show anything

we could edit this branch to error instead of setting `GForce = FALSE` and try then

> they would have to know and then do something new manually

Yes. I'm fine with this personally (perhaps pending some numeric evidence via the revdep check)

> What is an entry point for GForce

Another case of users having to know and then doing something manually. This time, instead of `as.numeric` (which could have a memory hit of storing two columns), users fine with using arithmetic mean on non-strictly-numeric columns do so by having `mean.default(col)` in `j`.

PR #3547, opened 2019-05-07
colClasses correspond to select

### LINKED ISSUE #1426: FR for `fread`: if `select` is used, `colClasses` need only correspond to the columns in `select`

mattdowle:
Closes #1426 
Follow up to #2545 

- [x] avoid needing new option (`datatable.colClassesSelectOrder` was attempted in this PR)
- [x] retain backwards compatibility with no need to migrate
- [x] avoid any ambiguities in syntax / intention
- [x] allow `colClasses` to be specified inside `select` directly in one place


MichaelChirico:
neat idea! lgtm

MichaelChirico:
on closer read I'm a bit confused. can't we just allow select to be used exactly as colClasses?

just pass colClasses to select instead of colClasses and it will also subset to only the mentioned columns?

mattdowle:
That is what it does.  It's exactly as colClasses when colClasses is a vector.  When select is used like this,  colClasses can't be used at the same time.

MichaelChirico:
Am I misreading this error?

```
STOP("When select= is a list it must be two items: select=list(<which columns>, <their types>). But it has %d items.", LENGTH(selectArg));
```

MichaelChirico:
As `colClasses`, I'd write this:

```
fread(file, select=list(c("ID","userClass"), c("character","userClass")))
```

as

```
fread(file, select=list(character = 'ID', userClass = 'userClass'))
```

mattdowle:
Why not : 
```
fread(file, select=c(ID="character", someCol='userClass'))
```
I expanded my comment above a bit regarding the selection order.

MichaelChirico:
> Why not

for the same reason as `colClasses`, which accepts both `c(column = 'type')` and `list(type = 'column')` format

mattdowle:
Or,  instead of : 
```
fread(file, select=list(c("ID","someCol"), c("character","integer64")))
```
how about : 
```
fread(file, select=c("ID","someCol"), selectClasses=c("character","integer64"))
```
When selectClasses is used, select must be provided and colClasses should not be provided.
This would free up select=list(...) to work like list form of colClasses,  The order of columns returned would be the appearance order in the file,  and select= could be used as well as colClasses to change the returned order (i.e. `fread(..., select=c(1,6,2,3,4), colClasses=list(numeric=c(1,4,6), character=2:3))` would still work).

mattdowle:
@MichaelChirico  Better?
(The uniqlist coverage seems spurious -- I'll look at that while you consider.)

MichaelChirico:
I think what I was missing is the part where `select` is also used to specify column order, which creates ambiguity in the use-as-`colClasses` case, is that right?

But if order matters, you're just consigned to using select-as-vector now?

(haven't gotten a chance to read carefully)

man/fread.Rd
```diff
@@ -40,16 +40,12 @@ yaml=FALSE, autostart=NA
   \item{stringsAsFactors}{ Convert all character columns to factors? }
   \item{verbose}{ Be chatty and report timings? }
   \item{skip}{ If 0 (default) start on the first line and from there finds the first row with a consistent number of columns. This automatically avoids irregular header information before the column names row. \code{skip>0} means ignore the first \code{skip} rows manually. \code{skip="string"} searches for \code{"string"} in the file (e.g. a substring of the column names row) and starts on that line (inspired by read.xls in package gdata). }
-  \item{select}{ Vector of column names or numbers to keep, drop the rest. }
+  \item{select}{ A vector of column names or numbers to keep, drop the rest. From v1.12.4, \code{select} may specify types too in the same way as \code{colClasses}; i.e., a vector of \code{colname=type} pairs, or a \code{list} of \code{type=col(s)} pairs. The order that the columns are specified determines the order of the columns in the result. }
```
MichaelChirico: >  The order that the columns are specified determines the order of the columns in the result.

This part has been true, maybe move it before "From v1.12.4"?

R/fread.R
```diff
@@ -319,20 +320,9 @@ yaml=FALSE, autostart=NA)
...
-        setcolorder(ans, rank)
-      }
-    } else {
-      if (!identical(names(ans), select)) {
-        reorder = select[select %chin% names(ans)] # any missing columns are warned about in freadR.c and skipped
-        setcolorder(ans, reorder)
-      }
-    }
-  }
+  selectOrder = attr(ans, "selectOrder")  # only present if select= was specified
```
MichaelChirico: this can't be done at the C level directly? rather than passing from C/fread to R back to C/setcolorder?

`setcolorder` ensures:

 - `!anyDuplicated(neworder)` [should be true of `selectOrder`]
 - `length(neworder) == length(x)` [ditto]
- `if (is.character(neworder))` check duplicates [`selectOrder` is `INTSXP` so this branch irrelevant]

So it seems OK to do from C?

src/freadR.c
```diff
@@ -161,11 +163,34 @@ SEXP freadR(
...
   selectSxp = selectArg;
   dropSxp = dropArg;
+  selectColClasses = false;
+  if (!isNull(selectSxp)) {
+    if (!isNull(dropSxp)) STOP("Use either select= or drop= but not both.");
+    if (isNewList(selectArg)) {
+      if (!isNull(colClassesSxp))
+        STOP("select= is type list for specifying types in select=, but colClasses= has been provided as well. Please remove colClasses=.");
+      if (!length(getAttrib(selectArg, R_NamesSymbol)))
+        STOP("select= is type list so expecting list(type1=cols1, type2=cols2, ...) but has no names");
```
MichaelChirico: "but the input has no names"

MichaelChirico:
Added comments; front-end API looks good 👍

man/fread.Rd
mattdowle: Good idea, will do.

R/fread.R
mattdowle: Good idea. Yes would be less moving parts done at C level. Ok will do

src/freadR.c
mattdowle: Good spot. I'll reword that.  (The "no names" there refers to the names of the select= argument (the type names) rather than the input data.)

PR #3550, opened 2019-05-07
fcast coverage (mainly removing uncoverable lines)

MichaelChirico:


R/fcast.R
```diff
@@ -165,11 +165,7 @@ dcast.data.table <- function(data, formula, fun.aggregate = NULL, sep = "_", ...
       # tryCatch(fill.default <- dat[0L][, eval(fun.call)], error = function(x) stop(errmsg, call.=FALSE))
       if (nrow(fill.default) != 1L) stop(errmsg, call.=FALSE)
     }
-    if (!any(valnames %chin% varnames)) {
-      dat = dat[, eval(fun.call), by=c(varnames)]
-    } else {
-      dat = dat[, { .SD; eval(fun.call) }, by=c(varnames), .SDcols = valnames]
```
MichaelChirico: when this branch was written, the branch at 131:

```
if (length(dupidx)) {
```

didn't exist. Now that it's there, the names are already unique by the time they reach this branch, so only `!any()` is possible.

R/fcast.R
```diff
@@ -220,26 +216,6 @@ dcast.data.table <- function(data, formula, fun.aggregate = NULL, sep = "_", ...
       # removed 'setcolorder()' here, #1153
     setattr(ans, 'names', c(lhsnames, allcols))
     setDT(ans); setattr(ans, 'sorted', lhsnames)
-  } else {
-    # formula is of the form x + y ~ . (rare case)
```
MichaelChirico: Added a comment for Arun on the original PR here:

https://github.com/Rdatatable/data.table/commit/25a74dfd8b738e5e5556144d38713b6b003a56f2

I don't think this branch is possible specifically due to L116:

```
lvars = lapply(lvars, function(x) if (length(x)) x else quote(`.`))
```

PR #3558, opened 2019-05-12
.Last.updated manual entry

jangorecki:
follow up of comments in #3460


mattdowle:
Sentence and link (+see also section) from `?:=`?

PR #3560, opened 2019-05-14
leave opt compiler flag default, follow up of 7810b22

jangorecki:
https://github.com/Rdatatable/data.table/commit/7810b224ed59d9ade5375ffd7945cf4e0fa3d785#r33513563
https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html#Debugging-Options

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -138,8 +138,8 @@ test-rel-cran-lin: # currently released R on Linux
     _R_CHECK_CRAN_INCOMING_REMOTE_: "FALSE"   # Other than no URL checking (takes many minutes) or 'Days since last update 0' NOTEs needed, #3284
   before_script:
     - mkdir -p ~/.R
-    - echo 'CFLAGS=-O3'> ~/.R/Makevars   # No -g because -g increases datatable.so size from 0.5MB to 1.5MB and breaches 'installed package size <= 5MB' note
-    - echo 'CXXFLAGS=-O3' >> ~/.R/Makevars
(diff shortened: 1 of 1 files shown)


mattdowle:
By default, do you mean R's default (which is -O2 with R on Linux from Ubuntu repo but I don't know how it works from docker) or do you mean the compiler's default?
By defining `CFLAGS=-O3` I was turning off R's default which was "-g -O2".
By defining `CFALGS=-g0` I think you're achieving the same as not defining `-g`,  but also now using gcc's default (which is -O0?) rather than R's default (-O2).  I thought you wanted exactly the same optimization level that R uses by default.

I'm (now) a little frustrated that we're spending time on this given nothing is broken here, and given we've plenty of more pressing issues.  I'll just merge anyway.

jangorecki:
You are correct. What I was trying to achieve was to append `-g0` to R's flag, but it doesn't work like this.
Now it compiles like
```
gcc -std=gnu99 -I"/usr/share/R/include" -DNDEBUG    -fopenmp -fpic  -g0 -c assign.c -o assign.o
```
https://gitlab.com/Rdatatable/data.table/-/jobs/212431091/artifacts/file/bus/test-rel-cran-lin/data.table.Rcheck/00install.out
Are you aware of a way to just disable `-g` and leave other flags as R's default?
if not then one way is to just use
```
CFLAGS=-g0 -O2 -fopenmp -Wall -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2
```
which should be same or close to R's defaults.

mattdowle:
No I'm not aware of how to do that. How I did it originally was fine I thought. The elephant in the room is why is R's default -O2 anyway and not -O3.

PR #3564, opened 2019-05-15
gmedian retain class

### LINKED ISSUE #3079: Date and POSIXct coerced to numeric when calculating median by group

mattdowle:
Closes #3079 
Follow up to #3546 
gmedian now retains the class, other than for integer64 since it always returns double.

Also reduced lines of code in gmedian by adding an integer64 case to quickselect.c, moving the 2nd-half minimum loops into quickselect.c, and finding location of median inside quickselect rather than passing that in.  Existing test 1579 covered int, double, integer64 and na.rm cases nicely.

R API raised outside loops in gmedian (since altrep R 3.5 greater overhead).

A few trailing spaces removed (my editor does that on save).

A few unrelated tests tweaked to pass in dev mode when data.table is not installed.

Checked all gforce function retain attributes. Added comment to the only one that doesn't (`gvarsd1`) since class unlikely applicable for sd/var.

### DIFF
--- a/src/gsumm.c
+++ b/src/gsumm.c
@@ -1181,7 +1181,7 @@ SEXP gvarsd1(SEXP x, SEXP narm, Rboolean isSD)
         error("Type '%s' not supported by GForce sd (gsd). Either add the prefix stats::sd(.) or turn off GForce optimization using options(datatable.optimize=1)", type2char(TYPEOF(x)));
       }
   }
-  copyMostAttrib(x, ans);
+  // no copyMostAttrib(x, ans) since class (e.g. Date) unlikely applicable to sd/var
   UNPROTECT(2);
   return (ans);
 }
--- a/src/gsumm.c
+++ b/src/gsumm.c
@@ -20,10 +20,6 @@ static int maxgrpn = 0;
 static int *oo = NULL;
 static int *ff = NULL;
 static int isunsorted = 0;
-static union {
-  double d;
-  long long ll;
-} u;
 
 // from R's src/cov.c (for variance / sd)
 #ifdef HAVE_LONG_DOUBLE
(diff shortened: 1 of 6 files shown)


src/gsumm.c
```diff
@@ -776,177 +772,56 @@ SEXP gmax(SEXP x, SEXP narm)
...
+    double *subd = REAL(PROTECT(allocVector(REALSXP, maxgrpn))); // allocate once upfront and reuse
+    int64_t *xi64 = (int64_t *)REAL(x);
+    double  *xd = REAL(x);
+    for (int i=0; i<ngrp; ++i) {
+      int thisgrpsize = grpsize[i], nacount=0;
+      for (int j=0; j<thisgrpsize; ++j) {
+        int k = ff[i]+j-1;
+        if (isunsorted) k = oo[k]-1;
+        k = (irowslen == -1) ? k : irows[k]-1;
+        if (isInt64 ? xi64[k]==NA_INTEGER64 : ISNAN(xd[k])) nacount++;
```
jangorecki: Isn't that better to move int64 check outside of the loop and duplicate loop to avoid extra branching in each iteration?

mattdowle:
I declared `isInt64` as `const bool` so that branch prediction should be very good. Interesting to test but, thanks to branch prediction, I doubt there will be a measurable difference in this case.
It is probably worth creating totally branch free cases; i.e. taking isunsorted and irowslen outside too. But that would complicate/expand the code.  The move of the R API outside the loops should have already had a greater benefit.
However, I knew that when gather() is applied this serial gather will go away. Then gmedian can apply the quickselect in parallel. That's what should be done next.

src/gsumm.c
mattdowle: Further comment here: https://github.com/Rdatatable/data.table/pull/3564#issuecomment-492925130.  Added here otherwise it doesn't look like I replied when viewing the diff.

PR #3572, opened 2019-05-18
New vignette -- Usages of .SD 

### LINKED ISSUE #3412: .SD vignette based on SO answer
[This SO answer](https://stackoverflow.com/a/47406952/3576984)

Encouraged to include as vignette:

 - https://stackoverflow.com/questions/8508482/what-does-sd-stand-for-in-data-table-in-r/47406952#comment81784007_47406952
 - https://twitter.com/El_Grund/status/1097068970203668480

I want to be sure to add a bit about chaining since I mentioned but forgot to include that in the original answer.

Another hold-up is the use of `Lahman` data. I guess I can try and find a URL instead of adding a Suggests for that package...

MichaelChirico:
Closes #3412

Not sure if we should track the png on GH or not


jangorecki:
I haven't gone through yet but looks like a comprehensive guide on using `.SD`.
- I would only avoid such title, as some people might see it as "cryptic symbols".
- Also as discussed in linked issue, we can have scope of that vignette extended for other tricks in `j`. Would be very useful if you could leave placeholders for that in the document, so it can be filled by others.
- png looks unnecessarily big, not sure if 12KB will make difference but recently Matt with dealing with compiler flags to reduce the size of package.

mattdowle:
Haven't look either yet but just on the png size, removing -g compiler flag saved 1MB recently (.so reduced from 1.5MB to 0.5MB) so the package size is now apx 4MB of 5MB limit.  12KB not an issue (1.2% of remaining).

mattdowle:
For future reference ...
It seems the Travis error when building vignettes and it suggests that rmarkdown is not available and to install it, is spurious (seen that before, iirc). It just means that something is wrong with the vignette somewhere and you have to reproduce it locally to find what's wrong.
In addition to changing the RData version from 3 to 2, I needed to change cache= from TRUE to FALSE. Otherwise it produced a similar message about version 3 format meaning that the package then depends on R 3.5+.

jangorecki:
why not use csv.gz instead of RData? there is no risk that vignette can be build on newer R only due to format incompatibility?

vignettes/datatable-sd-usage.Rmd
```diff
@@ -0,0 +1,258 @@
...
+  collapse = TRUE,
+  fig.path = 'plots/',
+  out.width = '100%',
+  dpi = 144
+)
+```
+
+This vignette will explain the most common ways to use the `.SD` variable in your `data.table` analyses. It is an adaptation of [this answer](https://stackoverflow.com/a/47406952/3576984) given on StackOverflow.
+
+# What is `.SD`?
```
jangorecki: https://rdatatable.gitlab.io/data.table/library/data.table/doc/datatable-sd-usage.html
rendered from R it result into `What is <code>.SD</code>` tab name in browser, maybe better remove code and leave .SD as plaintext

vignettes/datatable-sd-usage.Rmd
```diff
@@ -0,0 +1,258 @@
...
+par(oma = c(2, 0, 0, 0))
+lm_coef = sapply(models, function(rhs) {
+  # using ERA ~ . and data = .SD, then varying which
+  #   columns are included in .SD allows us to perform this
+  #   iteration over 16 models succinctly.
+  #   coef(.)['W'] extracts the W coefficient from each model fit
+  Pitching[ , coef(lm(ERA ~ ., data = .SD))['W'], .SDcols = c('W', rhs)]
+})
+barplot(lm_coef, names.arg = sapply(models, paste, collapse = '/'),
+        main = 'Wins Coefficient\nWiith Various Covariates',
```
jangorecki: `Wiith` double `i`

vignettes/datatable-sd-usage.Rmd
```diff
@@ -0,0 +1,258 @@
...
+        col = col16, las = 2L, cex.names = .8)
+```
+
+The coefficient always has the expected sign (better pitchers tend to have more wins and fewer runs allowed), but the magnitude can vary substantially depending on what else we control for.
+
+## Conditional Joins
+
+`data.table` syntax is beautiful for its simplicity and robustness. The syntax `x[i]` flexibly handles two common approaches to subsetting -- when `i` is a `logical` vector, `x[i]` will return those rows of `x` corresponding to where `i` is `TRUE`; when `i` is _another `data.table`_, a (right) `join` is performed (in the plain form, using the `key`s of `x` and `i`, otherwise, when `on = ` is specified, using matches of those columns).
```
jangorecki: there is also a case of `DT["someid"]`

vignettes/datatable-sd-usage.Rmd
```diff
@@ -0,0 +1,258 @@
...
+
+```{r sd_team_best_year}
+Teams[ , .SD[which.max(R)], by = teamID]
+```
+
+Note that this approach can of course be combined with `.SDcols` to return only portions of the `data.table` for each `.SD` (with the caveat that `.SDcols` should be fixed across the various subsets)
+
+_NB_: `.SD[1L]` is currently optimized by [_`GForce`_](https://jangorecki.gitlab.io/data.table/library/data.table/html/datatable-optimize.html) ([see also](https://stackoverflow.com/questions/22137591/about-gforce-in-data-table-1-9-2)), `data.table` internals which massively speed up the most common grouped operations like `sum` or `mean` -- see `?GForce` for more details and keep an eye on/voice support for feature improvement requests for updates on this front: [1](https://github.com/Rdatatable/data.table/issues/735), [2](https://github.com/Rdatatable/data.table/issues/2778), [3](https://github.com/Rdatatable/data.table/issues/523), [4](https://github.com/Rdatatable/data.table/issues/971), [5](https://github.com/Rdatatable/data.table/issues/1197), [6](https://github.com/Rdatatable/data.table/issues/1414)
```
jangorecki: use Rdatatable namespace instead of jangorecki: https://Rdatatable.gitlab.io/data.table/library/data.table/html/datatable-optimize.html

PR #3573, opened 2019-05-20
key argument for as.data.table methods

### LINKED ISSUE #890: Add an optional argument (key= ...) to as.data.table
This is for convenience, a bit tired of writing:

``` R
dt = as.data.table(x)
setkeyv(dt, c("s1", "s2"))
dt
```

I would be much nicer to have

``` R
dt = as.data.table(x, key = c("s1", "s2"))
```

Thank you,
You are doing amazing work with `data.table`.

MichaelChirico:
Closes #890

@jangorecki thanks for the heads up that this never ended up merged (3 years ago!)

Requesting review since there's a potentially breaking change to `as.data.table.array` -- `key` argument has to be brought to the front (S3 method consistency), so anyone that's relying on the ordering of the arguments to `as.data.table.array` will be affected by this PR.

PS Jan, could you please change the manual entry for the `sorted` argument to `as.data.table.array`? It gives no information about what it's used for, just that it's passed to the method.


R/as.data.table.R
```diff
@@ -64,20 +63,21 @@ as.data.table.matrix <- function(x, keep.rownames=FALSE, ...) {
...
-  setattr(value,"class",c("data.table","data.frame"))
-  alloc.col(value)
+  # setkey now to allow matrix column names as key
+  setkeyv(value, key)
+  value
 }
 
 # as.data.table.array - #1418
-as.data.table.array <- function(x, keep.rownames=FALSE, sorted=TRUE, value.name="value", na.rm=TRUE, ...) {
+as.data.table.array <- function(x, keep.rownames=FALSE, key=NULL, sorted=TRUE, value.name="value", na.rm=TRUE, ...) {
```
jangorecki: This change is fine, breaking but minor, needs to be mentioned in NEWS. It is quite unlikely that someone was using here argument matching by position on 3+ position.

MichaelChirico: @jangorecki looking again I see now what `sorted` argument does. It's a bit in conflict with potential `key` argument since it's `TRUE` by default.

As of now, user would have to do `as.data.table(ARRAY, key = key_col, sorted = FALSE)`...

Any ideas the right way to proceed? Most natural is to issue an error if `sorted && !is.null(key)`, but it's a bit awkward from user standpoint.

jangorecki: maybe just raise error when both are provided? `sorted` is just shortcut for `key=rev(head(names(ans), -1L))`

jangorecki: it is unlikely that user will need other key, then the one created by `sorted=TRUE`, on a such structure

jangorecki: maybe just mention in doc that `sorted` is ignored if `key` is provided

jangorecki: best would be to add some unit tests for those cases too

MichaelChirico: > best would be to add some unit tests for those cases too

was working on this when i thought of the conflict.

> it is unlikely that user will need other key, then the one created by sorted=TRUE, on a such structure

True. Won't bother over-engineering then.

MichaelChirico:
@jangorecki ok done

jangorecki:
Looks like https://codecov.io/gh/Rdatatable/data.table/compare/c68e95ec5860a947092be8c3fbfadd8f657ebf72...36e137b0b95175464746ad252a6800f40f8c5fe9/src/R/as.data.table.R#L155 cannot be covered. Unnamed list doesn't hit that branch, maybe safe to remove?

MichaelChirico:
Not quite, I missed it. Adding a test now.

MichaelChirico:
NVM, @jangorecki was right (though my test indeed was not working as intended), that line is un-hittable. `mn > 0` case will cover naming for all normal inputs; even if `mn == 0`, `setDT` handles the naming.

PR #3575, opened 2019-05-21
Edit help text of `J` and `CJ`

### LINKED ISSUE #3511: Help text of J and CJ
Some suggestions for the `J` and `CJ` help page:

-----

**Header**
Creates A Join Data Table -> Creates a join `data.table`

-----

**Usage**
`# DT[J(\dots)]`
Why is the `J` example commented away?

-----

`DT[J(\dots)]` -> `DT[J(...)]`
`DT[\dots]` -> `DT[...]`
`DT[SJ(\dots)]` -> `DT[SJ(...)]`

I assume it's the `#` which messes with the rendering of `\dots`

-----

**Arguments**
**sorted** Should the input be sorted (ascending order)?

I think it should be described that `sorted` is not "only" about sorting, but also whether keys are set. Now the key part is only mentioned in the first `CJ` example. Perhaps borrow the wording in **Value** of `SJ`:

>Should `setkey()` be called on all the columns in the order they were passed to `CJ`?

And similar for "If `FALSE`, the input order is retained":

>If `FALSE`, the result is unkeyed and input order is retained

------

**Value**

(S)orted (J)oin -> **S**orted **J**oin
(C)ross (J)oin -> **C**ross **J**oin

It looks a bit cluttered with the parentheses. I suggest a bold first letter instead. This also makes it consistent with the style used in the help text of `.SD`.

Again, I think it should be mentioned that the result of `CJ` is keyed (see above).

-----

**Examples**

`DT = data.table(A=5:1,B=letters[5:1])` -> `DT <- data.table(A = 5:1, B = letters[5:1])`
`setkey(DT,B) # re-orders` -> `setkey(DT, B) # reorders`

Insert spaces for readability (here and in other places), no dash in reorder.

`DT[.("b")]      # same. Style of package plyr.`

The dot alias for `list` now feels like an integral part of `data.table`, so no need to refer to another package, especially not the semi-deprecated `plyr` package. If anything, refer to relevant `data.table` docs on `.( )`.

Henrik-P:
Closes #3511
Edits `?J` and `?CJ`. Also added pointer to use of `on=` argument when joining and some other minor formatting.


Henrik-P:
Thanks for your rapid feedback @jangorecki. Sorry about the `<-`, I keep that in mind for any future edits. Cheers.

jangorecki:
```
* checking for code/documentation mismatches ... WARNING
Functions or methods with usage in documentation object 'J' but not in code:
  '['
* checking Rd \usage sections ... WARNING
Undocumented arguments in documentation object 'J'
  'DT' 'J(...)'
Objects in \usage without \alias in documentation object 'J':
  '['
Functions with \usage entries need to have the appropriate \alias
entries, and all their arguments documented.
The \usage entries must correspond to syntactically valid R code.
See chapter 'Writing R documentation files' in the 'Writing R
Extensions' manual.
```
before opening PR always best to do
```sh
make build && make check
```

man/J.Rd
```diff
@@ -2,57 +2,58 @@
...
+CJ(c(5, NA, 1), c(1, 3, 2)) # sorted and keyed data.table
+do.call(CJ, list(c(5, NA, 1), c(1, 3, 2)))  # same as above
+CJ(c(5, NA, 1), c(1, 3, 2), sorted = FALSE) # same order as input, unkeyed
+# use for 'unique = ' argument
+x <- c(1, 1, 2)
+y <- c(4, 6, 4)
 CJ(x, y) # output columns are automatically named 'x' and 'y'
-CJ(x, y, unique=TRUE) # unique(x) and unique(y) are computed automatically
-
+CJ(x, y, unique = TRUE) # unique(x) and unique(y) are computed automatically
```
MichaelChirico: style note: unique=TRUE

man/J.Rd
```diff
@@ -2,57 +2,58 @@
...
 \description{
-  Creates a \code{data.table} to be passed in as the \code{i} to a \code{[.data.table} join.
+  Creates a \code{data.table} for use in \code{i} in a \code{[.data.table} join.
 }
 
 \usage{
-# DT[J(\dots)]                           # J() only for use inside DT[\dots].
-SJ(\dots)                                # DT[SJ(\dots)]
-CJ(\dots, sorted = TRUE, unique = FALSE)  # DT[CJ(\dots)]
+DT[J(\dots)]                              # J() only for use inside DT[...].
```
MichaelChirico: now you see why this was commented before 🤪

R CMD check looks at this as a call to a function J, so it's throwing a block because there is no J function

(^this isn't quite right... error posted by Jan is best)

man/J.Rd
```diff
@@ -2,57 +2,58 @@
...
+SJ(\dots)                                 # DT[SJ(...)]
+CJ(\dots, sorted = TRUE, unique = FALSE)  # DT[CJ(...)]
 }
 
 \arguments{
   \item{\dots}{ Each argument is a vector. Generally each vector is the
-  same length but if they are not then the usual silent repetition is applied. }
-  \item{sorted}{ logical. Should the input be sorted (ascending order)? If \code{FALSE}, the input order is retained. }
+  same length, but if they are not then the usual silent recycling is applied. }
```
MichaelChirico: actually this description of dots doesn't really apply to the CJ case...

man/J.Rd
```diff
@@ -2,57 +2,58 @@
...
+  \item{sorted}{ logical. Should \code{setkey()} be called on all the columns in the order they were passed to \code{CJ}? }
```
MichaelChirico: phrasing here is a bit off... currently could imply setkey is applied repeatedly in the order of `...`

man/J.Rd
```diff
@@ -2,57 +2,58 @@
...
 
     \code{x[data.table(id)]} is the same as \code{x[J(id)]} but the latter is more readable. Identical alternatives are \code{x[list(id)]} and \code{x[.(id)]}.
 
-    \code{x} must have a key when passing in a join table as the \code{i}. See \code{\link{[.data.table}}
+    When using a join table in \code{i}, \code{x} must either be keyed or the \code{on} argument be used to indicate the columns in \code{x} and \code{i} which should be joined. See \code{\link{[.data.table}}.
 }
 \value{
   \itemize{
-    \code{J} : the same result as calling list. \code{J} is a direct alias for list but results in clearer more readable code.
+    \code{J} : the same result as calling \code{list}, for which \code{J} is a direct alias.
```
MichaelChirico: I like the note about readability, if not necessarily the exact phrasing. See the recently closed question about why J exists -- here is the place to emphasize why we don't just get rid of in favor of . or list

Henrik-P: Thanks a lot Jan, Michael and Matt for your additional edits. Obviously I need to learn how to properly make checks also for a seemingly simple edit like this. Sorry for the inconvenience and thanks for your patience and help with my very first PR to data.table.

MichaelChirico: more than happy to help. thank you!

jangorecki: no worries @Henrik-P. Most important tricks to learn is `make build && make check` before PR and use of `cc(F)` (it is like build & reload in rstudio). Thanks for investigating your time in data.table!

PR #3577, opened 2019-05-22
Formally remove relationship to reshape2

### LINKED ISSUE #3549: Completely axe reshape2 from code
As mentioned in #3542, [`reshape2` is dead since a year ago](https://github.com/hadley/reshape).

We removed it from suggests but still have some snippets like:

```
    ns = tryCatch(getNamespace("reshape2"), error=function(e)
         stop("The dcast generic in data.table has been passed a ",class(data)[1L]," (not a data.table) but the reshape2 package is not installed to process this type. Please either install reshape2 and try again, or pass a data.table to dcast instead."))
    ns$dcast(data, formula, fun.aggregate = fun.aggregate, ..., margins = margins,
             subset = subset, fill = fill, value.var = value.var)
```

I don't think it makes sense to support such functionality anymore.

Question is, are we safe to just get rid of it right away, or do we need to lifecycle it out?

MichaelChirico:
Closes #3549
A first pass at removing `reshape2` where possible.

The `else` branches in `melt` and `dcast` may or may not still be relevant... I checked the referenced package (`edarf`) and there's no mention of `reshape2` in its current `DESCRIPTION`; anyway I imagine revdep checks for other packages will come back problematic.

Still we should at least start the deprecation process of support in the next release... doesn't make much sense to me to maintain this code complexity for a deprecated package.


jangorecki:
Instead of removing tests I would move them into integration tests, which are not being run on CRAN, but we do run them in CI.
https://github.com/Rdatatable/data.table/blob/master/inst/tests/tests-DESCRIPTION
https://github.com/Rdatatable/data.table/blob/master/inst/tests/other.Rraw

MichaelChirico:
@jangorecki no tests removed here. One test was edited since the error message had `reshape2` in it (this was changed in `fmelt.c`)

Issue #1770, opened 2016-07-13
fwrite UTF8


skanskan:
Hello

Could you please add an option to fwrite in order to choose the character encoding, please? 
Such as UTF-8.
I know already that fread has that option.

JhossePaul:
Having encoding issues in Windows. Windows encoding is a real pain

+1

kzmlbyrk:
I have the same issue in windows. It would be great having an encoding option in fwrite.

msgoussi:
I have the same problem when I export data.table 
if I use fwrite: a string Côte d'Ivoire becomes Cأ´te d'Ivoire.
Thanks

amjiuzi:
any result of this issue?  i think it necessary for window platform
+1

kongdd:
I also have the same encoding problem in  windows. Hope it could be solved.

amjiuzi:
hi, @mattdowle, any result on this issue?

EspenRosenquist:
Yes, please! fread can encode, but if the file is saved with fwrite, latin characters in colnames are effed up and the one solution I have found is to reencode colnames with iconv, which is a rather brittle solution.

Lhemamou:
Any news?

igorstorm:
I am also having this issue.

y41u42002:
I also have this issue for encoding problem, hope for adding option.

nesscx:
Any progress on this?

MichaelChirico:
@skanskan @JhossePaul @kzmlbyrk @msgoussi @amjiuzi @EspenRosenquist @igorstorm @y41u42002 @nesscx @pachamaltese @elisendavila @lz1nwm @kongdd @byapparov @dpprdan @AndrewsOR @alexeyknorre @bendae19 @BastienFR @bobSpacewalk @szugat @alexiaaslan @yaakovfeldman @MathieuMarauri @franknarf1 @themeo @rsaporta @kuzmenkov111 @lucasmation @sindribaldur

Could anyone please offer a reproducible example? It's easier to work through a solution with one in hand. Thanks

skanskan:
I'm not using it for some time but wirting just UTF-8 would be good for me.
Though maybe other old-fashioned prefer ASCII.

shrektan:
Yes. UTF8 by default is better for everyone.

pachamaltese:
[...]

PR #3580, opened 2019-05-22
Fwrite bom

### LINKED ISSUE #3488: fwrite: Support writing UTF-8 files with BOM
Could we get fwrite to support writing UTF8-files with BOM like `readr::write_excel_csv()`? This would be of great help to us who have to deal with excel user :/

philippechataignon:
Closes #3488 
Related to #1770 too. It adds a parameter 'with_bom' in fwrite to create a UTF-8 file with BOM. By default, with_bom is FALSE and fwrite creates a UTF-8 file without BOM.
When with_bom is TRUE, BOM sequence (EF BB BF) is added at the beginning of the file.
with_bom is compatible with options `yaml = T` and `compress="gzip"`


MichaelChirico:
Awesome PR! Tackling `fwrite`'s most wanted again :)

For argument name, maybe `use_bom` or even `use_byte_order_mark`?

R/fwrite.R
```diff
@@ -54,6 +56,7 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
   file <- path.expand(file)  # "~/foo/bar"
   if (append && missing(col.names) && (file=="" || file.exists(file)))
     col.names = FALSE  # test 1658.16 checks this
+  if (with_bom && !col.names) stop("with_bom can be TRUE only if col.names is TRUE")
```
MichaelChirico: Maybe just ignore `with_bom` here (possibly with `warning`)?

R/fwrite.R
```diff
@@ -101,14 +104,22 @@ fwrite <- function(x, file="", append=FALSE, quote="auto",
         header = col.names, sep = sep, sep2 = sep2, eol = eol, na.strings = na,
         dec = dec, qmethod = qmethod, logical01 = logical01
       )
+      if (with_bom) {
+          bom <- raw(3L)
+          bom[1] <- as.raw(0xEF)
+          bom[2] <- as.raw(0xBB)
+          bom[3] <- as.raw(0xBF)
+          writeBin(bom, file)
```
MichaelChirico: Does this interact properly with `append`? I guess `with_bom && append` should be another error or ignore case

MichaelChirico: I see, this is nested within `if(yaml)` so already `append` case is handled

jangorecki:
My vote for arg name would be just `bom` or eventually `useBom`

R/fwrite.R
MichaelChirico: Shouldn't `with_bom && append` be a warning/error/ignore case as well?

jangorecki: What BOM couldn't be used without col names? I recall using some excel data without col names.

philippechataignon: Because of the organization of fwrite : rows writing is threaded and every thread has its own buffer. It's difficult to write BOM in this section of fwrite. The header is written with its own buffer : with_bom uses this buffer and so col.names must be TRUE.

philippechataignon:
BOM is often written in uppercase. What do you think of `addBOM` ?

MichaelChirico:
> BOM is often written in uppercase

so is YAML... `bom` would work for me

philippechataignon:
Let's go for `bom` 4e28d619

R/fwrite.R
philippechataignon: > Shouldn't `with_bom && append` be a warning/error/ignore case as well?

You're right. I propose that when appending, `bom` is silently set to FALSE, like `col.names`.
See f8775587. So stop is only when creating and col.names = FALSE.

MichaelChirico:
Could you clarify the relation to #1770?

philippechataignon:
#1770 is not very clear and his title is fwrite UTF8, so it seems to be related. My understanding is that, in Windows world (I use Linux), utf-8 files often need to have a BOM to be recognized as UTF-8. If not, they may be considered as "Latin-1" and there are "encoding" problems. I'm french, so I meet often "Latin-1" files. I'm happy that `fread` can handle them. But I never need to write Latin-1 files. And I never had encoding problems with utf8 files written with `fwrite `.

MichaelChirico:
OK, that's what I was thinking as well. We can leave it open and post in that thread seeing if it solves peoples' problems after merging.

R/fwrite.R
jangorecki: I would add internal `stopifnot` just in case if the code will be moved later on, to ensure what Michael mentioned.

mattdowle:
Hi @philippechataignon. Looks great. I'll try and get it passing now and merge.
You're a project member already: please next time create the branch directly in the main project. Just makes it easier for us to push changes to your branch. I can still push to your branch ok but I have to clone separately and my local diff tool comparing branch to master doesn't work unless it's a branch in main project.

mattdowle:
Good - passing now.  I needed to change yaml tests 2033.06 and 2033.07 to pass. I just replaced with the new values to get them to pass, without understanding why or if those are correct to change:   @philippechataignon  @MichaelChirico can you check please.
Also that one line needs covering in fwrite.R.

MichaelChirico:
I had a glance at the logs yesterday and saw that, seems familiar to an error I was getting on the original PR that was overcome by explicitly setting eol across platforms.

Dunno why this PR affected it. maybe strip.white is needed? will check

MichaelChirico:
Investigating, not sure I'll have time to finish. I notice that with the interaction with `yaml` seems broken...

I see:

```
tmp = tempfile()
DT = data.table(a = 1:5, b = c(pi, 1:4), c = letters[1:5])
fwrite(DT, tmp, yaml = TRUE, eol = '\r\n'); length(readLines(tmp))
# [1] 90
fwrite(DT, tmp, yaml = TRUE, eol = '\r\n', bom = TRUE); length(readLines(tmp))
# [2] 30
fwrite(DT, tmp, yaml = TRUE, eol = '\r\n', bom = FALSE); length(readLines(tmp))
# [3] 60
```

MichaelChirico:
@philippechataignon i believe the problem is that `writeBin` always appends to the connection, so it won't work if the file already exists.

MichaelChirico:
Really have no idea why test 2033.11 is failing on Windows but not Linux

MichaelChirico:
Latest commit tried using `close(file())` to reset any existing file before proceeding; same thing -- success on linux, failure on windows

philippechataignon:
What do you think of removing compatibility between `bom` and `yaml` for release 1.12.4 ? This will let more time for analysing. For me, even before this PR, I have `eol: |2+` in output. Except `yaml` header, all `fwrite`writing is made in C. I think this can be an aim to write yaml header in C and, as side effet, `yaml` will be compatible with `compress=gzip`.

MichaelChirico:
the `eol: |2+` part was due to the `append` bug. `eol: |2+` is the expected output when `eol = '\n'`; this was left in the file for the `eol = '\r\n'` tests because the file wasn't being overwritten.

I'm not sure 1.12.4 release is imminent so no need to write it off yet. that said, I'd be fine merging this and filing a follow-up to fix the windows bug.

mattdowle:
> I think this can be an aim to write yaml header in C and, as side effet, yaml will be compatible with compress=gzip.
Done. Great idea. It also writes the bom when no column names too.
Just two test fails left, just on Windows ...

mattdowle:
That's odd. I don't see why the two errors happen on Windows : 
```
Running test id 1658.5      Test 1658.5 produced 1 warnings but expected 0
Expected: 
Observed: invalid input found on input connection 'C:/Users/appveyor/AppData/Local/Temp/1\Rtmpa2ZQeO\file95075ee329b'
...
Running test id 2033.12      Test 2033.12 produced 1 warnings but expected 0
Expected: 
Observed: invalid input found on input connection 'C:/Users/appveyor/AppData/Local/Temp/1\Rtmpa2ZQeO\file95054067179'
```

MichaelChirico:
nice Matt!

do I understand right that your debugging for Windows here was to put a debugging tracer line then push & wait for appveyor & repeat?

mattdowle:
Yep.

mattdowle:
I'm expecting you and Philippe very likely will want to review and revise. I just wanted to merge now to see if it passed the stronger tests in GL pipelines, and get it into master too so further work (follow up PRs) can be in a branch.

jangorecki:
gcc-8 reports
```
fwrite.c: In function ‘fwriteMain’:
fwrite.c:692:26: warning: overflow in implicit constant conversion [-Woverflow]
     if (args.bom) {*ch++=0xEF; *ch++=0xBB; *ch++=0xBF; }  // 3 appears above (search for "bom")
                          ^~~~
fwrite.c:692:38: warning: overflow in implicit constant conversion [-Woverflow]
     if (args.bom) {*ch++=0xEF; *ch++=0xBB; *ch++=0xBF; }  // 3 appears above (search for "bom")
                                      ^~~~
fwrite.c:692:50: warning: overflow in implicit constant conversion [-Woverflow]
     if (args.bom) {*ch++=0xEF; *ch++=0xBB; *ch++=0xBF; }  // 3 appears above (search for "bom")
                                                  ^~~~
```

mattdowle:
Thanks @jangorecki. Now fixed in c21ac65.
We should add `-Wall -pedantic` to one of the GLCI jobs.

jangorecki:
@mattdowle we have `r-rel-lin` using those, and it produces those warnings in `00install.out`, but because those are compiler warnings R package check is fine about them. We would need to add extra grep on that file.

Issue #2406, opened 2017-10-10
data.table does not compile on OSX 10.13


hannesmuehleisen:
It seems the `-fopenmp` option to clang now causes an error in Apple LLVM 9.

```
> devtools::install_github("Rdatatable/data.table")
...
clang: error: unsupported option '-fopenmp'
```

```
> sessionInfo()
R version 3.4.2 (2017-09-28)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13

Matrix products: default
BLAS: /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] crayon_1.3.2         digest_0.6.12        withr_2.0.0         
 [4] assertthat_0.2.0     R6_2.2.2             git2r_0.19.0        
 [7] httr_1.3.1           debugme_1.0.2        rlang_0.1.2.9000    
[10] curl_3.0             callr_1.0.0.9000     devtools_1.13.3.9000
[13] tools_3.4.2          pkgload_0.0.0.9000   compiler_3.4.2      
[16] processx_2.0.0.1     pkgbuild_0.0.0.9000  memoise_1.1.0       
[19] knitr_1.17           usethis_0.0.0.9000  
```

```
clang -v
Apple LLVM version 9.0.0 (clang-900.0.37)
Target: x86_64-apple-darwin17.0.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

renkun-ken:
See https://github.com/coatless/r-macos-clang

mattdowle:
Also, CRAN provides a binary for MacOS and that now supports OpenMP, so you shouldn't need to compile.   If you do need to compile, then our [Mac Instructions](https://github.com/Rdatatable/data.table/wiki/Installation#openmp-enabled-compiler-for-mac) have been reported to work.

smcinerney:
a) CRAN only seems to have MacOS binaries up to R 3.2, not 3.3 or 3.4?

> > options(install.packages.check.source = "no")
[...]

PR #3582, opened 2019-05-22
check for index in setkey

### LINKED ISSUE #2889: Setting key to an existing index can skip forder step
Consider

```
DT = data.table(
  ID1 = rep(10:1, 10:1),
  ID2 = rep(1:10, 1:10)
)
setindex(DT, ID1, verbose = TRUE)
# forder took 0 sec
setkey(DT, ID1, verbose = TRUE)
# forder took 0 sec
# reorder took 0 sec
```

I believe the second `forder` is superfluous -- only the physical reordering is necessary in the `setkey`.

saraswatmks:
Closes #2889 

The idea the following:
1. Check in `setkeyv` if the data.table already has an index
2. If yes, use that index and don't run `setindex` inside it again.


mattdowle:
@saraswatmks Excellent! Have invited you to be project member (you need to accept the notification) so next time you can create a branch in the main project.

mattdowle:
I don't see `cols` used in the branch when `x` has some indices existing.  Doesn't it need to see if the requested key (i.e. `cols`) exists as an index? If it does it can use that ordering to change the physical order, otherwise (i.e. if no index exists for `cols`) it still needs to call `forder`.

saraswatmks:
@mattdowle thanks for your quick feedback. I've made the change. Now we check:
1. If the given `cols` exists as an index. If it does, we use that existing order.
2. If the given `cols` doesn't exist as an index, we run `forder` as usual.

I am on macos. No matter what I do I can't run test locally. I need to push everytime and wait for pipeline to throw some error. Could you guide me to some link so that I can test it locally before pushing ? I have already [tried this](https://github.com/Rdatatable/data.table/issues/2406) and it hasn't helped.

MichaelChirico:
hi @saraswatmks could you share more details about what's not working? especially your locale info, that seems to be the most common source of headaches

mattdowle:
@saraswatmks 
https://github.com/Rdatatable/data.table/wiki/Installation#openmp-enabled-compiler-for-mac
https://github.com/Rdatatable/data.table/wiki/Contributing#minimal-first-time-pr
If those don't work please post full output here showing error messages.

R/setkey.R
```diff
@@ -86,12 +86,25 @@ setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TR
...
   }
   if (!is.character(cols) || length(cols)<1L) stop("Internal error. 'cols' should be character at this point in setkey; please report.") # nocov
-  if (verbose) {
-    tt = suppressMessages(system.time(o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)))  # system.time does a gc, so we don't want this always on, until refcnt is on by default in R
-    # suppress needed for tests 644 and 645 in verbose mode
-    cat("forder took", tt["user.self"]+tt["sys.self"], "sec\n")
+  
+  # get existing index name if any
+  found_index <- NULL
+  if(is.null(indices(x))){
```
mattdowle: Need to be a `!is.null` here?  iiuc that's why line 107 isn't covered.

R/setkey.R
```diff
@@ -86,12 +86,25 @@ setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TR
...
+      if (verbose) {
+          tt = suppressMessages(system.time(o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)))  # system.time does a gc, so we don't want this always on, until refcnt is on by default in R
+          # suppress needed for tests 644 and 645 in verbose mode
+          cat("forder took", tt["user.self"]+tt["sys.self"], "sec\n")
+      } else {
+          o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)
+      }
   } else {
-    o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)
+      o <- attr(attributes(x)$index, which=found_index, exact = TRUE)
```
mattdowle: There could be a verbose message here when it finds an existing index and uses it,  with a new test that checks that message using `output=`.

jangorecki:
@saraswatmks at which point `R CMD build data.table --no-build-vignettes` and `R CMD check data.table_1.12.3.tar.gz --ignore-vignettes` are not working?

saraswatmks:
@MichaelChirico I get this error when I run `R CMD build .`

`* checking for file ‘./DESCRIPTION’ ... OK
* preparing ‘data.table’:
* checking DESCRIPTION meta-information ... OK
* cleaning src
* installing the package to build vignettes
      -----------------------------------
** libs
/usr/local/opt/llvm/bin/clang -fopenmp -I/anaconda3/lib/R/include -DNDEBUG   -I/usr/local/opt/gettext/include -I/usr/local/opt/llvm/include   -fPIC  -g -O3 -Wall -pedantic -std=gnu99 -mtune=native -pipe -c assign.c -o assign.o
In file included from assign.c:1:
In file included from ./data.table.h:1:
/anaconda3/lib/R/include/R.h:55:11: fatal error: 'stdlib.h' file not found
# include <stdlib.h> /* Not used by R itself, but widely assumed in packages */
          ^~~~~~~~~~
1 error generated.
make: *** [assign.o] Error 1
* removing ‘/private/var/folders/6l/7tvz3rz510n4gp2gsrpq6mxr0000gp/T/RtmpMitKB2/Rinstfc6676e4c658/data.table’
      -----------------------------------
ERROR: package installation failed`

saraswatmks:
@jangorecki Thanks for helping. When I run `R CMD check data.table_1.12.3.tar.gz --ignore-vignettes` I get

```
* using log directory ‘/Users/manish/Documents/open-source-projects/data.table.Rcheck’
* using R version 3.4.2 (2017-09-28)
* using platform: x86_64-apple-darwin13.4.0 (64-bit)
* using session charset: UTF-8
* using option ‘--ignore-vignettes’
* checking for file ‘data.table/DESCRIPTION’ ... OK
* this is package ‘data.table’ version ‘1.12.3’
* checking package namespace information ... OK
* checking package dependencies ...Warning: unable to access index for repository https://CRAN.R-project.org/src/contrib:
  internet routines cannot be loaded
Warning: unable to access index for repository https://bioconductor.org/packages/3.5/bioc/src/contrib:
  internet routines cannot be loaded
Warning: unable to access index for repository https://bioconductor.org/packages/3.5/data/annotation/src/contrib:
  internet routines cannot be loaded
Warning: unable to access index for repository https://bioconductor.org/packages/3.5/data/experiment/src/contrib:
  internet routines cannot be loaded
 ERROR
Packages suggested but not available: ‘R.utils’ ‘nanotime’
```

saraswatmks:
Here's my sessionInfo

```
> sessionInfo()
R version 3.5.1 (2018-07-02)
Running under: macOS  10.14.5

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib

locale:

attached base packages:

 [1] Rcpp_1.0.0        rstudioapi_0.8    magrittr_1.5      usethis_1.4.0     devtools_2.0.1   
 [6] pkgload_1.0.2     R6_2.3.0          rlang_0.3.0.1     tools_3.5.1       pkgbuild_1.0.2   
[11] data.table_1.11.8 sessioninfo_1.1.1 cli_1.0.1         withr_2.1.2       remotes_2.0.2    
[16] yaml_2.2.0        assertthat_0.2.0  digest_0.6.18     rprojroot_1.3-2   crayon_1.3.4     
[21] processx_3.2.0    callr_3.0.0       base64enc_0.1-3   fs_1.2.6          ps_1.2.1         
[26] testthat_2.0.1    glue_1.3.0        memoise_1.1.0     compiler_3.5.1    desc_1.2.0       
[31] backports_1.1.2   prettyunits_1.0.2
```

jangorecki:
@saraswatmks install those two missing packages or use `_R_CHECK_FORCE_SUGGESTS_=false` env var before running package check.

R/setkey.R
saraswatmks: sorry, my bad

R/setkey.R
saraswatmks: made the changes, please check.

saraswatmks:
> @saraswatmks install those two missing packages or use `_R_CHECK_FORCE_SUGGESTS_=false` env var before running package check.

@jangorecki Just to make sure I am doing the right way, what is the common way to do this ? For ex: I am doing the following:
1. Go to terminal. Create a conda env and install all these dependencies using `conda install`
2. Run commands from terminal.

Am I following the right way? Or there is something in `RStudio` which can help me skip these steps.

jangorecki:
1. skip conda
2. use plain R
3. use existing [Makefile](https://github.com/Rdatatable/data.table/blob/master/Makefile), it makes life easier, then in console
```sh
cd data.table
make build && make check
```

MichaelChirico:
to each his own but I cringe at the word `conda`. it has only caused me headaches. RStuido has build&reload and clean&reload functions which will help a lot.

I've been thinking it might be nice to have a short vignette or blog post or wiki page on "how to fix a data.table issue for a first time contributor" giving  some tips and tricks on getting over the hump.

you might want to check the file cc.R which has a _lot_ of (densely packed) accumulated knowledge on debugging & package development

MichaelChirico:
PS so it doesn't get lost in the trees -- thanks for the PR! we really appreciate all community efforts at fixing&improving the package & hope the experience is more illuminating than frustrating 😁

R/setkey.R
```diff
@@ -86,12 +86,26 @@ setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TR
...
+  
+  # get existing index name if any
+  found_index <- NULL
+  if(!is.null(indices(x))){
+      found_index <- names(attributes(attributes(x)$index))
+      found_index <- gsub("^__","", found_index)
+  }
+  
+  # forder only if index is not present
+  if(!identical(found_index, cols)){
```
mattdowle: This works if there is one index,  as the test tests. But there can be multiple indexes (each index is a set of columns). It needs to find if any of the indexes match `cols` and then use that one.  The test needs expanding for cases of multiple indexes and `cols` both existing and not existing when there are multiple (so it tests it picks up the correct one).

R/setkey.R
```diff
@@ -86,12 +86,26 @@ setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TR
...
+      if (verbose) {
+          tt = suppressMessages(system.time(o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)))  # system.time does a gc, so we don't want this always on, until refcnt is on by default in R
+          # suppress needed for tests 644 and 645 in verbose mode
+          cat("forder took", tt["user.self"]+tt["sys.self"], "sec\n")
+      } else {
+          o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)
+      }
   } else {
-    o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)
+      cat("using existing index for", found_index, "\n")
```
mattdowle: Needs an `if (verbose)` please otherwise this is always being printed.

R/setkey.R
```diff
@@ -86,12 +86,26 @@ setkeyv <- function(x, cols, verbose=getOption("datatable.verbose"), physical=TR
     if (!typeof(.xi) %chin% c("integer","logical","character","double")) stop("Column '",i,"' is type '",typeof(.xi),"' which is not supported as a key column type, currently.")
   }
   if (!is.character(cols) || length(cols)<1L) stop("Internal error. 'cols' should be character at this point in setkey; please report.") # nocov
-  if (verbose) {
-    tt = suppressMessages(system.time(o <- forderv(x, cols, sort=TRUE, retGrp=FALSE)))  # system.time does a gc, so we don't want this always on, until refcnt is on by default in R
-    # suppress needed for tests 644 and 645 in verbose mode
-    cat("forder took", tt["user.self"]+tt["sys.self"], "sec\n")
+  
+  # get existing index name if any
+  found_index <- NULL
```
mattdowle: We prefer `=` in data.table.  I don't agree with the common and widely marketed advice to use `<-`.  Single `=` is the same as C and many other languages.  I sometimes hear, e.g. from Python folk,  that R code using `<-` looks "old"/"not modern" and I see their point.  I've always used `=`.   I save `<-` for use when passing function arguments to do an assign and pass in one go;   e.g.   `write(DT, file=f<-tempfile());   ... do something with f ...` .   I'm often swapping between C and R (which I think more people should do too since C is not as hard as some people want you to believe).  When doing this, using `=` to mean assign and `==` to mean equals in two languages consistently (R and C) is nice.   And by the way,  the people who laugh at R for using `L` postfix to mean integer ... it comes from C (it's the same as C) and it makes a lot of sense.

saraswatmks:
@MichaelChirico Thanks to you guys for your kindness and patience. I've been wanting to contribute since a long time.
What else needs to be done in this PR ?

mattdowle:
> What else needs to be done in this PR ?

@saraswatmks I made some comments above to be addressed please.

R/setkey.R
saraswatmks: thanks for detailed insight. made the change.

R/setkey.R
saraswatmks: fixed it.

R/setkey.R
saraswatmks: fixed it.

inst/tests/tests.Rraw
```diff
@@ -6031,6 +6031,19 @@ thisDT <- copy(DT)[, c("aaa", "b") := 2]
...
+setindex(DT, a)
+test(1419.60, allIndicesValid(DT), TRUE)
+test(1419.61, setkey(DT, a, verbose=TRUE), output="using existing index for a")
+DT <- data.table(a = c(1,1,1,1,1,2,2,2,2,2),
+                 aaa = c(1,1,2,2,2,1,1,2,2,2),
+                 bbb = c(1,1,2,0,1,1,1,0,1,1))
+setindex(DT, a)
+setindex(DT, aaa)
+test(1419.62, allIndicesValid(DT), TRUE)
+test(1419.63, setkey(DT, aaa, verbose=TRUE), output="using existing index for aaa")
```
mattdowle: Same request here as above for 1419.61.  This test needs to check the result of this setkey() command has done the correct thing (e.g. changed the physical row order properly) as well as doing that for the right reason (`output=`).
Since setkey() returns DT invisibly, the easiest way is just to pass y=data.table(...) into this test().  See other tests() in his file for examples.

R/setkey.R
```diff
@@ -86,12 +86,30 @@ setkeyv = function(x, cols, verbose=getOption("datatable.verbose"), physical=TRU
...
+      }
   } else {
-    o = forderv(x, cols, sort=TRUE, retGrp=FALSE)
+      # find the matching index
+      ix =  found_index[which(found_index == new_possible_index)]
+      if (verbose){
+          cat("using existing index for", gsub("^__","", ix), "\n")
+          o <- attr(attributes(x)$index, which=ix, exact = TRUE)
+      } else {
+          o <- attr(attributes(x)$index, which=ix, exact = TRUE)
```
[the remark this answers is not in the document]
saraswatmks: @mattdowle I've made the changes. Correct me if I am wrong, I think we need `ix` here because, let's take a scenario where a user set multiple indexes. So when we do `indices(DT)`, we get `a, b, __a__b`, if the user does `setkey(DT, a), setkey(DT, b), setkey(DT, a, b)` for a table `DT = data.table(a = c(...), b = c(...), c = c(...))` . In this case, we need to find the `ix` index of which index is matched.

DESCRIPTION
```diff
@@ -39,7 +39,8 @@ Authors@R: c(
   person("Michael","Quinn",        role="ctb"),
   person("@javrucebo","",          role="ctb"),
   person("@marc-outins","",        role="ctb"),
-  person("Roy","Storey",           role="ctb"))
+  person("Roy","Storey",           role="ctb"),
+  person("Manish","Saraswat",      role="ctb"))
```
jangorecki: those entries should be managed by Matt and Arun only

MichaelChirico: @jangorecki from Matt above:

> Also, please add a news item, and please add yourself to the bottom of contributors list in DESCRIPTION.

jangorecki: resolved them

R/setkey.R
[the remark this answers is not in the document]
mattdowle: Look at line 104 though.  I commented before that line 104 could be removed.  Line 104 is, in the end, just renaming `index` as `idx`, iiuc.  This `which=` is looking up the item by name anyway.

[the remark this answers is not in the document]
saraswatmks: Done. Thanks for your patience.

inst/tests/tests.Rraw
```diff
@@ -6031,6 +6031,31 @@ thisDT <- copy(DT)[, c("aaa", "b") := 2]
...
+test(1419.60, allIndicesValid(DT), TRUE)
+setindex(DT, NULL)
+setkey(DT, a)
+test(1419.61, DT$a, c(1,1,2,3,4,5,6,7,8,9))
+setkey(DT, NULL)
+setindex(DT, a)
+test(1419.62, setkey(DT, a, verbose=TRUE), output="using existing index for a")
+
+# check setkey incase of existing multiple indexes
+DT <- data.table(a = c(3,3,4,4,5,6,1,1,7,2,2),
```
mattdowle: `a` has 11 items here but `aaa` and `bbb` have 10 so there's a warning when running test.data.table() runs in dev and when users would run test.data.table().  It's a lot easier and faster to run locally first before pushing. See the `cc.R` script in the root of the project : that's what I use in dev.

inst/tests/tests.Rraw
```diff
@@ -6031,6 +6031,31 @@ thisDT <- copy(DT)[, c("aaa", "b") := 2]
...
+                 bbb = c(1,1,2,0,1,1,1,0,1,1))
+setkey(DT, a)
+test(1419.63, DT$a, c(1,1,2,2,3,3,4,4,5,6,7))
+setkey(DT, NULL)
+test(1419.64, setkey(DT, a, verbose=TRUE), output="forder took")
+setkey(DT, NULL)
+setindex(DT, a)
+setindex(DT, aaa)
+test(1419.65, allIndicesValid(DT), TRUE)
+test(1419.66, setkey(DT, aaa, verbose=TRUE), output="using existing index for aaa")
```
mattdowle: This doesn't check the result of setkey is correct.  Please see previous comments where that was requested :  https://github.com/Rdatatable/data.table/pull/3582#discussion_r287548138 and https://github.com/Rdatatable/data.table/pull/3582#discussion_r287548675

mattdowle:
I added the result checks to the tests.  That actually finds a problem in setindex() where it is incorrectly using an existing index.  That commit will fail on test 1419.66 and then I'll fix that next.

inst/tests/tests.Rraw
saraswatmks: thanks for chipping in. I wasn't fully aware of `test` function parameters. Now I understand it better. Also, I will try to fix my local environment (I still don't know how am I gonna do that) to avoid stupid errors in the PR.

inst/tests/tests.Rraw
```diff
@@ -5668,13 +5668,13 @@ test(1376.06, indices(DT), c("b","a"))  # 2 secondary keys of single columns
...
-test(1376.09, list(DT[a==2L], indices(DT)), list(DT[9L],"a"))  # create indices for next test
+test(1376.09, list(DT[a==2L], indices(DT)), list(DT[2L],"a"))  # create indices for next test
 setindex(DT,NULL)
 test(1376.10, list(key(DT), indices(DT)), list("b", NULL))
 options(datatable.auto.index = FALSE)
-test(1376.11, list(DT[a==2L], indices(DT)), list(DT[9L],NULL))
+test(1376.11, list(DT[a==2L], indices(DT)), list(DT[2L],NULL))
 options(datatable.auto.index = TRUE)
-test(1376.12, list(DT[a==2L], indices(DT)), list(DT[9L],"a"))
+test(1376.12, list(DT[a==2L], indices(DT)), list(DT[2L],"a"))
```
mattdowle: These 3 tests changes from 9L to 2L were strange that they were needed.  Anyway, the latest commit leaves these tests unchanged which seems more correct.

mattdowle: Yep: those 3 tests should not have been changed to 9L in this PR.  The setkey(DT,b) inbetween tests 1376.07 and 1376.08 had broken under this PR and wasn't changing the row order properly.

jangorecki:
This change will be superseded (possibly taking it all away during git merge) by https://github.com/Rdatatable/data.table/pull/4386 which do same optimization (among others) but on C level rather than R.

PR #3587, opened 2019-05-23
frollmean exact consistent to base R about Inf

jangorecki:
exact more exact

### DIFF
--- a/inst/tests/froll.Rraw
+++ b/inst/tests/froll.Rraw
@@ -875,7 +875,6 @@ if (requireNamespace("zoo", quietly=TRUE)) {
         for (na.rm in c(FALSE, TRUE)) {
           for (fill in c(NA_real_, 0)) {
             for (algo in algos) {
-              if (fun=="sum" & algo!="fast") next
               num <<- num + num.step
               eval(substitute( # so we can have values displayed in output/log rather than variables
                 test(.num,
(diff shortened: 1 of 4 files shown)


inst/tests/froll.Rraw
```diff
@@ -801,23 +801,81 @@ test(9999.191, frollmean(c(1:2,NA,4:10), rep(2L,10), adaptive=TRUE, hasNA=FALSE)
...
     num.step = 0.0001
     #### fun, align, na.rm, fill, exact
-    for (fun in c("mean")) { # ,"sum"
+    for (fun in funs) {
       for (align in c("right","center","left")) {
         for (na.rm in c(FALSE, TRUE)) {
           for (fill in c(NA_real_, 0)) {
-            for (algo in c("fast", "exact")) {
+            for (algo in algos) {
+              if (fun=="sum" & algo!="fast") next
```
jangorecki: this line shouldn't be here, will amend that when submitting frollsum. It does not have any impact as we don't test for `fun="sum"` yet

src/froll.c
```diff
@@ -154,26 +154,32 @@ void frollmeanExact(double *x, uint_fast64_t nx, ans_t *ans, int k, double fill,
...
-      if (nc == 0) {                                          // no NAs in current window
-        for (int j=-k+1; j<=0; j++) {                         // sub-loop on window width
-          err += x[i+j] - res;                                // measure roundoff for each obs in window
-        }
-        ans->dbl_v[i] = (double) (res + (err / k));           // adjust calculated fun with roundoff correction
-      } else if (nc < k) {
-        for (int j=-k+1; j<=0; j++) {                         // sub-loop on window width
-          if (!ISNAN(x[i+j])) err += x[i+j] - res;            // measure roundoff for each non-NA obs in window
+      if (w > DBL_MAX) tmp_res = R_PosInf;                    // handle Inf for na.rm=TRUE consistently to base R
+      else if (w < -DBL_MAX) tmp_res = R_NegInf;
```
jangorecki: those two lines is the main change in this PR

PR #3593, opened 2019-05-24
Follow-up to #3572

MichaelChirico:
Jan had some extra feedback on #3572.

Regarding the `<code>` block in the header, I think that's just a GitLab issue. `title` YAML field is supposed to go into the browser tab, I thought... I guess GitLab is not [using MultiMarkdown](https://stackoverflow.com/questions/17097894/markdown-tag-for-document-title), or something like that? The `datatable-intro` vignette has the [same issue on Gitlab](https://rdatatable.gitlab.io/data.table/library/data.table/doc/datatable-intro.html) but it renders correctly [on CRAN](https://cran.r-project.org/web/packages/data.table/vignettes/datatable-intro.html), so ignoring that part.


jangorecki:
It is not strictly related to GitLab, I am getting same `<code>` after rendering vignette locally on my laptop. I think CRAN uses some pandoc configuration that handles that better.

PR #3595, opened 2019-05-25
warning -> error for datatable.old.key.by, as promised

MichaelChirico:


jangorecki:
I would postpone that to 1.13.0 to not deprecate in minor releases

NEWS.md
```diff
@@ -146,6 +146,8 @@
 
 9. `DT[col]` where `col` is a column containing row numbers of itself to select, now suggests the correct syntax (`DT[(col)]` or `DT[DT$col]`), [#697](https://github.com/Rdatatable/data.table/issues/697). This expands the message introduced in [#1884](https://github.com/Rdatatable/data.table/issues/1884) for the case where `col` is type `logical` and `DT[col==TRUE]` is suggested.
 
+10. `datatable.old.unique.by.key` is now fully deprecated and its usage will result in an error, one year after was advised in the release notes for v1.11.0.
+
```
mattdowle: More clarity here would be nice, and reminding folks how careful we've been. How about:

The `datatable.old.unique.by.key` option has been warning for 1 year that it is deprecated: `Please stop using it and pass 'by=key(DT) instead for clarity`. This warning is now upgraded to error as per note 10 in v1.11.0 (May 2018) warned, and note 1 in v1.9.8 (Nov 2016) before that. In May 2020 the option will be removed.

MichaelChirico: Sounds good to me

mattdowle:
@jangorecki It's already deprecated really.  It's already been warning for a year whenever it is used. So it's pretty minor to upgrade to error.

PR #3596, opened 2019-05-25
port CJ to C

### LINKED ISSUE #3597: CJ can handle list() input if sorted=FALSE but fails confusingly
```
CJ(1:2, list(1:3, 1))
```

> Error in forderv(y, retGrp = TRUE) : 
  Column 2 is length 1 which differs from length of column 1 (3)

```
CJ_old(1:2, list(1:3, 1), sorted = FALSE)
   V1    V2
1:  1 1,2,3
2:  1     1
3:  2 1,2,3
4:  2     1
```

The logic for building the `CJ` output from `list` input is the same, but `sorted` has to be `FALSE` because there's no ordering of lists.

We can either `stop` here or force `sorted = FALSE`... either way the current behavior is confusing.

MichaelChirico:
Also closes #3597

Putting on my C training wheels... guidance on ways to improve appreciated 🙌

Follow-up to https://github.com/orgs/Rdatatable/teams/project-members/discussions/17

Benchmarking with `setDTthreads(2L)`

```
setDTthreads(2L)
system.time(DT <- CJ(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11))
  #  user  system elapsed 
  # 2.823   0.578   1.707 
nrow(DT)
rm(DT); gc()
system.time(CJ_newR(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11))
 #   user  system elapsed 
 # 16.663   2.202  15.514
gc()
system.time(CJ_old(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11))
  #  user  system elapsed 
  # 5.959   0.637   1.953 
```

Paralellization working well here:

```
setDTthreads(8L)
system.time(DT <- CJ(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11))
  #  user  system elapsed 
  # 5.506   1.215   0.911 
```

For the record the speed of the pure-R `CJ` is really quite impressive! it's basically the same speed as declaring an empty list with that number of rows:

```
system.time(replicate(10L, numeric(39916800L), simplify = FALSE))
#    user  system elapsed 
#   0.899   0.912   1.891 
system.time(CJ_old(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11))
#    user  system elapsed 
#   5.994   0.648   1.920 
```


src/cj.c
```diff
@@ -0,0 +1,53 @@
...
+      break;
+    case STRSXP: {
+      SEXP this_col = PROTECT(allocVector(STRSXP, NN)); nprotect++;
+      for (int i = 0; i < NN; i++) {
+        SET_STRING_ELT(this_col, i, STRING_ELT(this_v, (i % modulo) / div));
+      }
+      SET_VECTOR_ELT(out, j, this_col);
+    } break;
+    default:
+      error("Type '%s' not supported by CJ.", type2char(TYPEOF(this_v)));
```
MichaelChirico: do I need to `UNPROTECT` in an error branch?

R/CJ.R
```diff
@@ -0,0 +1,69 @@
...
+  } else if (ncol > 1L && !emptyList) {
+    # using rep.int instead of rep speeds things up considerably (but attributes are dropped).
+    n = vapply_1i(l, length) #lengths(l) will work from R 3.2.0 (also above)
+    nrow = prod(n)
+    if (nrow > .Machine$integer.max) {
+      stop("Cross product of elements provided to CJ() would result in ",nrow," rows which exceeds .Machine$integer.max == ",.Machine$integer.max)
+    }
+    
+    # apply sorting
+    if (sorted && any(idx <- vapply_1b(l, is.list))) stop("'sorted' is TRUE but element ", which(idx), " is a list, which can't be sorted; try setting sorted = FALSE")
```
MichaelChirico: so it's not missed because of the diff from moving CJ to its own file -- this is the part that closes #3597

src/cj.c
mattdowle: That's one nice thing: R will clean up all objects on error so you don't need UNPROTECT.
Yay -- you're into C!  Party!
We've been taking R API usage outside loops recently since R 3.5 added overhead. So take the REAL() calls outside (see other C code in data.table for examples but look at files that have been more recently revised).

mattdowle:
quick suggestion: use `Rprof()` to confirm the timing you've shown is all in the C. It's feasible there's some surprising overhead at the R level before it gets to C.  Best to double-check.

src/cj.c
```diff
@@ -0,0 +1,64 @@
...
+        INTEGER(this_col)[i] = this_v_int[(i % modulo) / div];
+      }
+      SET_VECTOR_ELT(out, j, this_col);
+    }
+      break;
+    case REALSXP: {
+      SEXP this_col = PROTECT(allocVector(REALSXP, NN)); nprotect++;
+      double *this_v_dbl = REAL(this_v);
+      for (int i = 0; i < NN; i++) {
+        REAL(this_col)[i] = this_v_dbl[(i % modulo) / div];
```
MichaelChirico: @mattdowle is this how to do it? Or is there a way to get the LHS out of the loop too?

mattdowle: Yep that's it.  Add a `double *this_col_dbl = REAL(col)` too and use on LHS.  Note our naming conventions are still in flux (_dbl or D or d suffix for double or data, etc) - don't worry for now.

src/cj.c
```diff
@@ -0,0 +1,66 @@
...
+      for (int i = 0; i < NN; i++) {
+        this_col_ptr[i] = this_v_ptr[(i % modulo) / div];
+      }
+      SET_VECTOR_ELT(out, j, this_col);
+    }
+      break;
+    case INTSXP: {
+      SEXP this_col = PROTECT(allocVector(INTSXP, NN)); nprotect++;
+      int *this_v_ptr = INTEGER(this_v);
+      int *this_col_ptr = INTEGER(this_col);
```
MichaelChirico: @mattdowle is this the proper way? (also for `VECSXP`/`STRSXP` cases it seems there's no analog?)

mattdowle: Correct.  You can't do it for VECSXP/STRSXP because that would be what's known as "behind the write barrier".  You can partially do it though by reading (RHS) but not writing (LHS =).

MichaelChirico:
Updated timing after moving R API calls out of the loop as suggested. Getting close to _matching_ the pure R version on current master.

src/cj.c
mattdowle: I'm off to bed now an it's a long weekend here -- so have fun in C!   Where did you get the sense that CJ() was slow and that it could be sped up -- is the benchmark you put in this PR what motivated you or is there another one?
The idea from way-back was that CJ() would not actually materialize the full table.  It would just mark itself as "CJ" object and then bmerge would know how to recycle from the irregular non-expanded structure appropriately without needing to expand it.

MichaelChirico: No benchmark, just from staring at the code now and then and thinking "there's gotta be a faster way", and then happening on the modulo/div version here in my scratch work on an unrelated issue but alas. I'm still surprised how fast it is... `rep.int` is a beast. I guess `rep.int` is doing most of the work, and the recursion is only touching scalars instead of the full data (so it's not slowing down).

The alt-rep-y form does sound promising (certainly way more memory efficient), though I do use `CJ` often outside of `i` for populating the crossjoin tables for other reasons.

MichaelChirico: enjoy the weekend!

MichaelChirico:
I can't make heads or tails of the `Rprof` output, though it does look like the new version is (slightly) more memory efficient:

```
Rprof(interval = .005, memory.profiling = TRUE)
CJ(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11)
Rprof(NULL)
summaryRprof(memory = 'both')
# $by.self
#                 self.time self.pct total.time total.pct mem.total
# "CJ"                3.195    99.22      3.200     99.38    1370.5
# "print.default"     0.010     0.31      0.010      0.31       0.2
# "[.data.table"      0.005     0.16      0.005      0.16       0.3
# "options"           0.005     0.16      0.005      0.16       0.3
# "pmatch"            0.005     0.16      0.005      0.16       0.0
# 
# $by.total
#                     total.time total.pct mem.total self.time self.pct
# "CJ"                     3.200     99.38    1370.5     3.195    99.22
# "<Anonymous>"            0.020      0.62       0.7     0.000     0.00
# "print.data.table"       0.020      0.62       0.7     0.000     0.00
# "print.default"          0.010      0.31       0.2     0.010     0.31
# "print"                  0.010      0.31       0.2     0.000     0.00
# "[.data.table"           0.005      0.16       0.3     0.005     0.16
# "options"                0.005      0.16       0.3     0.005     0.16
# "pmatch"                 0.005      0.16       0.0     0.005     0.16
# ".deparseOpts"           0.005      0.16       0.0     0.000     0.00
# "["                      0.005      0.16       0.3     0.000     0.00
# "char.trunc"             0.005      0.16       0.3     0.000     0.00
# "deparse"                0.005      0.16       0.0     0.000     0.00
# "do.call"                0.005      0.16       0.3     0.000     0.00
# "format.data.table"      0.005      0.16       0.3     0.000     0.00
# "format"                 0.005      0.16       0.3     0.000     0.00
# "FUN"                    0.005      0.16       0.3     0.000     0.00
# "lapply"                 0.005      0.16       0.3     0.000     0.00
# "name_dots"              0.005      0.16       0.0     0.000     0.00
# "rbindlist"              0.005      0.16       0.3     0.000     0.00
# "suppressWarnings"       0.005      0.16       0.3     0.000     0.00
# "tail.data.table"        0.005      0.16       0.3     0.000     0.00
# "tail"                   0.005      0.16       0.3     0.000     0.00
# 
# $sample.interval
# [1] 0.005
# 
# $sampling.time
# [1] 3.22

Rprof(interval = .005, memory.profiling = TRUE)
CJ_old(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11)
Rprof(NULL)
summaryRprof(memory = 'both')
# $by.self
#                 self.time self.pct total.time total.pct mem.total
# "CJ_old"            6.295    99.76      6.300     99.84    1472.0
# "[.data.table"      0.005     0.08      0.005      0.08       0.2
# "%in%"              0.005     0.08      0.005      0.08       0.1
# "print.default"     0.005     0.08      0.005      0.08       0.5
# 
# $by.total
#                    total.time total.pct mem.total self.time self.pct
# "CJ_old"                6.300     99.84    1472.0     6.295    99.76
# "<Anonymous>"           0.010      0.16       0.7     0.000     0.00
# "print.data.table"      0.010      0.16       0.7     0.000     0.00
# "[.data.table"          0.005      0.08       0.2     0.005     0.08
# "%in%"                  0.005      0.08       0.1     0.005     0.08
# "print.default"         0.005      0.08       0.5     0.005     0.08
# "["                     0.005      0.08       0.2     0.000     0.00
# "deparse"               0.005      0.08       0.1     0.000     0.00
# "head.data.table"       0.005      0.08       0.2     0.000     0.00
# "head"                  0.005      0.08       0.2     0.000     0.00
# "mode"                  0.005      0.08       0.1     0.000     0.00
# "name_dots"             0.005      0.08       0.1     0.000     0.00
# "print"                 0.005      0.08       0.5     0.000     0.00
# "rbindlist"             0.005      0.08       0.2     0.000     0.00
# 
# $sample.interval
# [1] 0.005
# 
# $sampling.time
# [1] 6.31
```

src/cj.c
mattdowle: yep: CJ uses rep.int and its implemented in C and a .Internal() R function too (which makes it fast to call in a loop).  So really your're trying to beat C with C.  `Rprof` on the current CJ might be misleading actually as I think someone said `Rprof` doesn't count .Internal() R functions.
Are definitely getting the same result in the same order?  Doing "(i % modulo) / div" on every single iteration is inefficient because it's really just batches of increasing i.  It's possible that all those % and / add up in this case (it's contiguous assign so cache effects shouldn't come into play).

mattdowle:
It's a good case to try parallelizing.  Follow the #omp pragma examples in reorder.c or subset.c maybe.   Each column of CJ() result can be done independently in parallel.  There isn't any random cache cache (all sequential) so you should get good speedups; e.g. 2 threads should be close to twice as fast in this case. Can't go parallel for strings in this case,  but logical, integer and real should be good (and factor is integer).

src/cj.c
MichaelChirico: > it's really just batches of increasing i

Not sure I follow this.

Latest commit splits out the first and last column as separate cases to skip one of the `%`/`/` operations, gave about 10% boost.

MichaelChirico:
> Can't go parallel for strings in this case

I think I'm not following how to parallelize across columns then... first iterate over the columns to see which can be done in parallel, then batch it out?

Maybe parallelizing rows is the easier route?

MichaelChirico:
OK I tried parallelizing across rows on the `INTEGER`/`REAL`/`LOGICAL` branches, I think it worked well. With 2 threads, we equal the speed of the current implementation and still get the 10% memory savings. More threads and we do better than currently. 

Since really what's been parallelized is the _row_ index, are we safe to include the `VECSXP` and `STRSXP` cases in parallel as well?

MichaelChirico:
Now that the basic idea is working & improves the current implementation, should we try and bring more of the logic into C? Not sure there's much benefit...

src/cj.c
jangorecki: generally better to parallelise upper layers if possible, thus parallelising columns should give better improvement

src/cj.c
```diff
@@ -0,0 +1,126 @@
+#include "data.table.h"
+
+SEXP cj(SEXP base_list) {
+  int JJ = LENGTH(base_list), nprotect = 0;
+  SEXP out = PROTECT(allocVector(VECSXP, JJ)); nprotect++;
+  int NN = 1;
+  for (int j = 0; j < JJ; j++) NN *= LENGTH(VECTOR_ELT(base_list, j));
+  int div = NN, modulo;
+  
+  //#pragma omp parallel for num_threads(getDTthreads())
```
jangorecki: generally better to parallelise upper layers if possible, thus parallelising columns should give better improvement

src/cj.c
jangorecki: REAL, INTEGER and LOGICAL

jangorecki:
very nice PR

src/cj.c
MichaelChirico: See comment above... any suggestions for how to do that? I couldn't see how to make the parallelization work without substantially complicating the logic when there's string input (or list for that matter, but we could block lists if that's a real impediment)

Or maybe I'm just missing something...

MichaelChirico: Also, I guess most of the time the input to `CJ` is one, two, or three columns, but output could still be millions of rows... doesn't it make more sense to parallelize across rows then?

jangorecki: make sense!

src/cj.c
```diff
@@ -0,0 +1,126 @@
...
+    div = modulo/LENGTH(VECTOR_ELT(base_list, j));
+    switch(TYPEOF(this_v)) {
+    case LGLSXP: {
+      SEXP this_col = PROTECT(allocVector(LGLSXP, NN)); nprotect++;
+      int *restrict this_v_ptr = LOGICAL(this_v);
+      int *this_col_ptr = LOGICAL(this_col);
+      if (j == 0) {
+        #pragma omp parallel for num_threads(getDTthreads())
+        for (int i = 0; i < NN; i++) {
+          this_col_ptr[i] = this_v_ptr[i / div];
```
jangorecki: you iterate over all rows of the output, maybe we could do `memcpy` of blocks of rows? then iterate on blocks, this can be in parallel also. One block per thread.

src/cj.c
```diff
@@ -4,20 +4,26 @@ SEXP cj(SEXP base_list) {
   int JJ = LENGTH(base_list), nprotect = 0;
   SEXP out = PROTECT(allocVector(VECSXP, JJ)); nprotect++;
   int NN = 1;
+  // already confirmed to be less than .Machine$integer.max in R
```
jangorecki: so we compute it twice? maybe handle this here only

MichaelChirico: Was looking at that, but then we'd have to either (a) move the sorting step to C or (b) do the sorting only to `stop` right afterwards. `NN` should be almost instant to compute of course...

jangorecki: we can also pass that value from R to C as extra attribute, but not sure if it is worth as it will be instant computation anyway.

MichaelChirico: exactly... complicating the interface for micro gain, unless `...` is hundreds of columns, in which case almost surely the `integer.max` limit will be hit anyway unless most of `...` has length 1... seems unlikely.

MichaelChirico: For porting the sort to C... fine I guess, but there's the funky bit about generating `dups` there as well, would have to return `dups` as an attribute on the C output, sounds clunky & again probably unnecessary...

src/cj.c
MichaelChirico: Not quite sure I follow you here, sorry

jangorecki: here is R code for that
```r
library(data.table)
x = 1:3; nx = length(x)
y = 1:5; ny = length(y)
ans_y = vector("integer", nx*ny)
# pragma here
for (i in seq.int(nx)) ans_y[seq.int(ny)+(i-1L)*length(y)] = y

all.equal(CJ(x, y)$y, ans_y)
```

MichaelChirico: ohh gotcha. just loop over y once, figure out where each element goes, and copy all at once, makes sense

jangorecki: Looks trivial for 2 columns, for 3+ columns gets slightly more complicated, but I believe this is the way to go if we want to have that in C optimally
```r
CJ(1:2, 1:3, 1:4)
```
```
      V1    V2    V3
    <int> <int> <int>
 1:     1     1     1
 2:     1     1     2
 3:     1     1     3
 4:     1     1     4
 5:     1     2     1
 6:     1     2     2
 7:     1     2     3
 8:     1     2     4
```

mattdowle:
Please don't move CJ() at R level into its own file. This breaks the commit history: CJ.R looks like new code and we can't see what changed inside it. If we did this for all functions then we'd have a ton of files. It's not perfect currently, but I don't struggle with the way things are organized currently:  fairly similar functions being together in one file.

mattdowle:
With c41c1a4 I see : 
```
> system.time(DT <- CJ(1:2, 1:3, 1:4, 1:5, 1:6, 1:7, 1:8, 1:9, 1:10, 1:11))
   user  system elapsed   # v1.12.2
  0.966   0.288   1.216 

   user  system elapsed   # now
  0.212   0.434   0.205
```

Would be good to compare character type too.

MichaelChirico:
How many threads for that speedup? Awesome stuff!

mattdowle:
With character ids : 
```R
ids = as.vector(outer(LETTERS, LETTERS, paste0))
system.time(DT <- CJ(ids, 1:72, 1:73, 1:74))   # 5GB; 250m rows
#   user  system elapsed   
#  3.030   0.950   3.963  # v1.12.2
#  1.563   1.298   1.910  # now

ids = as.factor(ids)
system.time(DT <- CJ(ids, 1:72, 1:73, 1:74))
#   user  system elapsed
#  2.334   0.833   3.152  # v1.12.2
#  0.513   1.122   0.409  # now
```

MichaelChirico:
It just struck me to worry about how this interacts with ALTREP, are we covered there? Particularly since `CJ` will often be used like `CJ(1:10, 1:20)`...

mattdowle:
Yes should be fine. INTEGER(), REAL() etc auto-expand ALTREPs. To work with ALTREPs without expanding them, you have to switch to use different R API calls. In this case of `CJ()` definitely not worth it because the work happens as the product of the lengths.

PR #3598, opened 2019-05-25
cedta() catches another case where utils could be top-level namespace

### LINKED ISSUE #2972: Strange errors when running examples with local=TRUE argument
I generally run examples listed in help documentation using the `example` command with the argument `local=TRUE` because I don't like polluting my workspace with variables/functions defined from running examples!  So, I tried calling this: `example(":=", package="data.table", local=TRUE)` which yields an incomprenehsible error!  However, calling `example(":=", package="data.table")` works correctly but then creates three variables (`DT`, `sq_col_idx`, and `sq_cols`) in my workspace.  Thinking that this was just a unique case with `:=` I tried `example("CJ", package="data.table", local=TRUE)` which tells me that it cannot find function`J`.

It appears that these issues are because of the NSE evaluation that data.table uses.  If it is too much work to make `example` work with `local=TRUE` argument can you please ensure that `example(":=", package="data.table")` removes all the variables it created from the workspace?

MichaelChirico:
Closes #2972

Running `example(":=", package="data.table", local=TRUE)` winds up with `ns` as `utils`, a new case to `cedta`. `local = FALSE` has `ns` `NULL`.

Is it safe to run `example` in the tests? feels a bit off.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -120,6 +120,8 @@
 
 15. `rbind` and `rbindlist` of zero-row items now retain (again) the unused levels of any (zero-length) factor columns, [#3508](https://github.com/Rdatatable/data.table/issues/3508). This was a regression in v1.12.2 just for zero-row items. Unused factor levels were already retained for items having `nrow>=1`. Thanks to Gregory Demin for reporting.
 
+16. Running the `:=` examples with `example` and `local=TRUE` works as expected, [#2972](https://github.com/Rdatatable/data.table/issues/2972). Thanks @vlulla for the report.
+
 #### NOTES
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -120,6 +120,8 @@
 
 15. `rbind` and `rbindlist` of zero-row items now retain (again) the unused levels of any (zero-length) factor columns, [#3508](https://github.com/Rdatatable/data.table/issues/3508). This was a regression in v1.12.2 just for zero-row items. Unused factor levels were already retained for items having `nrow>=1`. Thanks to Gregory Demin for reporting.
 
+16. Running the `:=` examples with `example` and `local=TRUE` works as expected, [#2972](https://github.com/Rdatatable/data.table/issues/2972). Thanks @vlulla for the report.
```
mattdowle: 'as expected' doesn't mean much to a reader of news. 'works rather than error'.  The item can be shortened by just writing `example(`:=`, local=TRUE)` at the beginning, because 'with' isn't quite as clear.

PR #3600, opened 2019-05-26
frollapply

jangorecki:
#2778


src/frollR.c
```diff
@@ -206,3 +206,34 @@ SEXP frollfunR(SEXP fun, SEXP obj, SEXP k, SEXP fill, SEXP algo, SEXP align, SEX
...
+  SEXP w = PROTECT(allocVector(REALSXP, ik)); protecti++;
+  double *dw = REAL(w);
+  SEXP ans = PROTECT(allocVector(REALSXP, nx)); protecti++;
+  double *dans = REAL(ans);
+
+  SEXP call = PROTECT(lang2(fun, R_NilValue)); protecti++;
+  SETCADR(call, w);
+
+  for (int i=0; i<ik-1; i++) dans[i] = NA_REAL;
+  for (R_xlen_t i=ik-1; i<nx; i++) {
```
jangorecki: @mattdowle could you please take a look at this loop and suggest if there is better way than `memcpy` to achieve so called "online" rolling window. I was looking at FIFO queues but we need double array ultimately so linked-list to double array conversion on each iteration would probably be bad idea. Best would be not to copy but to shift pointer location.
`w` is all values in current window, must be SEXP so we can call `eval` on that.

jangorecki: comments under this Q are quite relevant: https://stackoverflow.com/questions/57334741/subsetting-a-large-vector-uses-unnecessarily-large-amounts-of-memory

jangorecki:
It is now vectorized over `x` and `n`, same as other rolling functions. Also support of `...` has been added recently. `align` argument and `fill`.
It is pretty ready to review. As of now only unit tests left.

jangorecki:
@mattdowle I put some comments so it is easier to understand why some loops exists

MichaelChirico:
whoops... not sure why the diff exploded. I:

```
git checkout master && git pull
git checkout -b frollapply && git pull origin frollapply
# make edits
git commit -am 'tidy up NEWS'
git push origin frollapply
```

jangorecki:
@MichaelChirico For me Rdatatable is defined as `upstream` rather than `origin`
instead of pulling to a branch I merge to it
```sh
git checkout master && git pull upstream master
git checkout -b frollapply
git merge master
# fix conflicts
git add
git commit
git push upstream frollapply
```

jangorecki:
Added WIP for now. Current code treats output of a `FUN` as double, so if one uses for example `uniqueN` then the result is quite unexpected `1.482197e-323`, due to use of `REAL(INTSXP)`. We should at least warn about that.

MichaelChirico:
I'm usually using `git merge` if that branch already exists locally & had been doing `git pull` for a newly created branch... actually thinking now I may already have had `frollapply` branch locally from earlier in this PR, so maybe that's the issue.

jangorecki:
https://github.com/Rdatatable/data.table/pull/3600/commits/621753401bc59bd066ea198e37ab47bd9c2be080 address the issue of `FUN` being not type-stable etc., in a minimal-overhead way. Only the first iteration checks if results are length 1 and of type double. If type is int/logical, it runs loop using extra coercion to double, which seems to be minimally faster than wrapping `as.numeric(FUN(x))` in R.

PR #3602, opened 2019-05-26
Added [<-.IDate so that rbind.data.frame retains integer IDate

### LINKED ISSUE #2008: IDate storage mode not preserved in data frames
Hi,
To save space, I've been using IDate within my data frames (not data tables) to save space. Maybe this is a user bug right here, if IDate isn't meant to be used outside of data.tables, but I'll continue.

Seems that rbind() does not preserve the integer storage mode of IDate inside of data.frames, which causes problems when using IDate arithmetic. Instead, rbind coerces IDate into storage.mode double. My code has worked perfectly for a while. I only started seeing errors after updating to the latest data.table 1.10.0 and R 3.3.2. However, I'm not sure if the coercion to double didn't happen in the earlier versions, or just that it wasn't checked before.

Here's an example:

```
x <- data.frame(date=as.IDate(Sys.Date()))
y <- rbind(x, x)
x$date -1
# [1] "2017-01-29"
y$date -1
```

> Error in `-.IDate`(y$date, 1) : 
  Internal error: storage mode of IDate is somehow no longer integer
```
storage.mode(x$date)
# [1] "integer"
storage.mode(y$date)
# [1] "double"
```

MichaelChirico:
Closes #2008 

I still don't _really_ understand why these changes are necessary (and both are in fact necessary).

I posted this on r-devel, might be some insight there: https://stat.ethz.ch/pipermail/r-devel/2019-May/077866.html

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -122,6 +122,8 @@
 
 16. `rbind` and `rbindlist` of an item containing an ordered factor with levels containing an `NA` (as opposed to an NA integer) could segfault, [#3601](https://github.com/Rdatatable/data.table/issues/3601). This was a a regression in v1.12.2. Thanks to Damian Betebenner for reporting.
 
+17. `rbind.data.frame` on `IDate` columns corrupted the `integer` storage mode, [#2008](https://github.com/Rdatatable/data.table/issues/2008). Thanks to @rmcgehee for reporting.
+
 #### NOTES
 
 1. `rbindlist`'s `use.names="check"` now emits its message for automatic column names (`"V[0-9]+"`) too, [#3484](https://github.com/Rdatatable/data.table/pull/3484). See news item 5 of v1.12.2 below.
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -993,16 +993,17 @@ test(343, DT[,(mynewcol):=21L], data.table(b=1:3,newname=21L))
 mycols = 1:2
 test(344, DT[,(mycols):=NULL], data.table(NULL))
 
-
-# It seems that the .Internal rbind of two data.frame coerces IDate to numeric. Tried defining
-# "[<-.IDate" as per Tom's suggestion, and c.IDate to no avail (maybe because the .Internal code
-# in bind.c doesn't look up package methods?). Anyway, as from 1.8.1, double are allowed in keys, so
-# these still work but for a different reason than before 1.8.1: the results are IDate stored as double,
-# rather than before when is worked because by and setkey coerced double to integer.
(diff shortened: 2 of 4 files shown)


MichaelChirico:
Josh Ulrich replied on r-devel, misunderstanding was pretty simple -- `[<-` is called by `rbind.data.frame`. Meanwhile `[<-.Date` does a funny thing where it retains its class while also calling `as.Date`, which converts to `numeric`.

So the approach here seems correct.

inst/tests/tests.Rraw
```diff
@@ -993,16 +993,17 @@ test(343, DT[,(mynewcol):=21L], data.table(b=1:3,newname=21L))
 mycols = 1:2
 test(344, DT[,(mycols):=NULL], data.table(NULL))
 
-
-# It seems that the .Internal rbind of two data.frame coerces IDate to numeric. Tried defining
```
MichaelChirico: Wow, the level of _deja vu_ on reading this:

> It seems that the .Internal rbind of two data.frame coerces IDate to numeric. Tried defining "[<-.IDate" as per Tom's suggestion, and c.IDate to no avail (maybe because the .Internal code in bind.c doesn't look up package methods?)

In fact I went through the same existential crisis for a while this morning! I'm not sure what was tried for `[<-.IDate` back then, maybe the `NextMethod` approach? I tried that as well but it seems fundamentally doomed because `...` in `NextMethod` can't be passed as `origin` to the `as.Date` call in `[<-.Date`, which is needed if we pass the underlying `integer` to `[<-.Date` as `value`. If we pass an `IDate`, the #2008 problem happens -- `double` with `IDate` class. I didn't try running `as.Date(value)` but that would mean coercing numeric just to get `[<-.Date` and then back to integer aftewards... seems silly.

Anyway I think we can change this test that was testing that a bug was still a bug. Now the bug is fixed. I'm not sure if the subsequent test is still needed, but it can't hurt.

NEWS.md
```diff
@@ -122,6 +122,8 @@
 
 16. `rbind` and `rbindlist` of an item containing an ordered factor with levels containing an `NA` (as opposed to an NA integer) could segfault, [#3601](https://github.com/Rdatatable/data.table/issues/3601). This was a a regression in v1.12.2. Thanks to Damian Betebenner for reporting.
 
+17. `rbind.data.frame` on `IDate` columns corrupted the `integer` storage mode, [#2008](https://github.com/Rdatatable/data.table/issues/2008). Thanks to @rmcgehee for reporting.
```
mattdowle: 'corrupted' implies something broken (seriously!) but it's just a type change, still worked.  I'll soften it when merging.

PR #3605, opened 2019-05-26
Change zbuffsize formula

### LINKED ISSUE #3599: gzip error in fwrite with bom and no col.names
In current master, this code ends with error:

```
DT <- data.table(l=letters, n=1:26)
fwrite(DT, f1 <- "f1.csv.gz", bom=T, col.names=F)
```

Verbose output
```
> omp_get_num_procs()==4
> R_DATATABLE_NUM_PROCS_PERCENT=="" (default 50)
> R_DATATABLE_NUM_THREADS==""
> omp_get_thread_limit()==2147483647
> omp_get_max_threads()==4
> OMP_THREAD_LIMIT==""
> OMP_NUM_THREADS==""
> data.table is using 2 threads. This is set on startup, and by setDTthreads(). See ?setDTthreads.
> RestoreAfterFork==true
> No list columns are present. Setting sep2='' otherwise quote='auto' would quote fields containing sep2.
> Column writers: 11 3 
> args.doRowNames=0 args.rowNames=0 doQuote=-128 args.nrow=26 args.ncol=2 eolLen=1
> maxLineLen=31. Found in 0.000s
> Writing bom (true), yaml (0 characters) and column names (false) ... Buffersize: 3, ZBuffersize: 19
> Error in fwrite(DT, f1 <- "f1.csv.gz", bom = T, col.names = F) : 
>   Compress gzip error: -5
```

-5 is a *buffer error*. `Buffersize: 3, ZBuffersize: 19` is an added debug : 19 comes from formula size + 10% * size + 16. In zlib, there is a function which plays that role : [compressBound(sourceLen)](https://docs.huihoo.com/symbian/s60-5th-edition-cpp-developers-library-v2.1/GUID-FE27AB35-C6FD-4F11-802D-0D5FCFFC2976/html/mrt/zlib-compressbound-1.html). In this case, compressBound(3) returns 16 and there is still a -5 error code.

I propose the use the formula : compressBound(headerLen) + 512 for the first buffer (bom + yaml + col.names) and to use compressBound for the threaded buffers.

PR follows if agreed.

philippechataignon:
Closes #3599 

With original buffer size = 3, previous formula gives 19 and compress ends with *buffer error*.

In zlib, there is a *deflateBound* function which computes the right size but needs a zlib stream structure. This PR uses this function and adapts gzip compression.

So we split *compressbuff* function in 2 functions :

* *init_stream* which initializes a zlib stream structure and can be  passed to *deflateBound*

* compressbuff which makes the compression

*compressbuff* has been simplified because we are sure that buffer doesn't exceeds maxint (4G) because max buffer size is 1G. With Z_FINISH, we are sure that all the buffer is compressed in one pass (or this is an unrecoverable error), so the do loop can be removed.

*deflateEnd* call is not more in *compressbuff* and is called when stream has been used (in fwriteMain).

### DIFF
--- a/src/fwrite.c
+++ b/src/fwrite.c
@@ -703,16 +703,15 @@ void fwriteMain(fwriteMainArgs args)
       int ret1=0, ret2=0;
       if (args.is_gzip) {
         z_stream stream;
-        if(init_stream(&stream)) {
-          free(buff);
-          STOP("Can't allocate gzip stream structure");
-        }
+        if(init_stream(&stream)) { free(buff); STOP("Can't allocate gzip stream structure"); } // # nocov
         size_t zbuffSize = deflateBound(&stream, headerLen);
         char *zbuff = malloc(zbuffSize);
+        // # nocov start
         if (!zbuff) {
             free(buff);
             STOP("Unable to allocate %d MiB for zbuffer: %s", zbuffSize / 1024 / 1024, strerror(errno));
         }
+        // # nocov end
         size_t zbuffUsed = zbuffSize;
         ret1 = compressbuff(&stream, zbuff, &zbuffUsed, buff, (int)(ch-buff));
         if (ret1 == Z_OK)
@@ -768,7 +767,7 @@ void fwriteMain(fwriteMainArgs args)
   if(args.is_gzip){
     z_stream stream;
     if(init_stream(&stream))
-      STOP("Can't allocate gzip stream structure");
+      STOP("Can't allocate gzip stream structure"); // # nocov
     zbuffSize = deflateBound(&stream, buffSize);
     deflateEnd(&stream);
   }
@@ -832,7 +831,7 @@ void fwriteMain(fwriteMainArgs args)
       if (args.is_gzip && !failed) {
         z_stream mystream;
         if(init_stream(&mystream)) {
-          failed = -998;
+          failed = -998; // # nocov
         }
         if (!failed) {
             myzbuffUsed = zbuffSize;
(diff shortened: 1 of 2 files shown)


src/fwrite.c
```diff
@@ -817,8 +829,15 @@ void fwriteMain(fwriteMainArgs args)
       }
       // compress buffer if gzip
       if (args.is_gzip && !failed) {
-        myzbuffUsed = myzbuffSize;
-        failed = compressbuff(myzBuff, &myzbuffUsed, myBuff, (int)(ch-myBuff));
+        z_stream mystream;
+        if(init_stream(&mystream)) {
```
mattdowle: @philippechataignon Can this `init_stream` be raised up? It's being redone by each thread for every chunk currently, iiuc.

philippechataignon: Actually the logic is the same than in the the previous version with no overhead : for each chunk, a zstream is initialized, deflate is called with a Z_FINISH flush option, and deflateEnd is called. Each chunk is compressed independently (good for speed but not for the compression ratio) and written in order. With Z_FINISH, zlib manual says " After deflate has returned Z_STREAM_END, the only possible operations on the stream are deflateReset or deflateEnd.". I don't try deflateReset : perhaps it's possible to use it and to have only one allocation by thread, I will have a look. I'm quite sure that each thread must have its own ztream structure : it can't be before parallel region.

philippechataignon:
```
err = deflate(stream, Z_FINISH);
return err == Z_STREAM_END ? Z_OK : err;
```
If err != Z_STREAM_END, we must return a non zero value to catch the error. Actually, if `err==Z_OK`, which is possible in partial compressing, test is *false* but return err = Z_OK which is false too. But we must consider it as an error because we need only one pass compress.

 Commit 7c0a1202 fix this.

PR #3609, opened 2019-05-28
parallel any_na

jangorecki:


src/utils.c
```diff
@@ -0,0 +1,54 @@
...
+  uint64_t inx[nx];
+  for (R_len_t i=0; i<nx; i++) {
+    inx[i] = xlength(VECTOR_ELT(x, i));
+    ix[i] = INTEGER(VECTOR_ELT(x, i));
+  }
+
+  uint64_t first[nx]; // unused here
+  #pragma omp parallel for num_threads(getDTthreads())
+  for (R_len_t i=0; i<nx; i++) {
+    any_na_integer(ix[i], inx[i], &bans[i], &first[i], false);
```
jangorecki: last argument `false` disables parallelism over rows, as we parallelise over columns in this loop

src/utils.c
```diff
@@ -0,0 +1,54 @@
...
+#include <Rdefines.h>
+
+void any_na_integer(int *x, uint64_t nx, bool *ans, uint64_t *first, bool parallel) {
+  bool found=false;
+  #pragma omp parallel for if (parallel) num_threads(getDTthreads())
+  for (uint64_t i=0; i<nx; i++) {
+    if (found) continue;
+    if (x[i]==NA_INTEGER) {
+      found = true;
+      first[0] = i;
```
st-pasha: If multiple threads find an NA simultaneously and try to write into `*first`, an Undefined Behavior will occur (possibly with the result that `*first` will end up containing some garbage value).

A bigger problem though if some thread finds an NA value, but it is not the FIRST value in the vector. Say if each thread gets to check 1000 rows, and there are NA values in rows 999 and 1000, then the second thread will find the NA in row 1000 immediately and signal all threads to stop, long before the first thread gets to row 999.

A better strategy would be something like this:
```
found_index = INTMAX;
#omp for dynamic(1) reduce(min:found_index)
for( i = 1:NBATCHES ) {
  if (found) continue;
  for (row = i*BATCH_SIZE:(i+1)*BATCH_SIZE) {
    if (x[row] is na) {
      found = true;
      found_index = row; 
      break;
    }
  }
}
```

jangorecki: I removed `first` related logic for now. Originally added it so extra computation is not lost, but it is not yet used. Your code is put there in comment so when needed can be added.

jangorecki:
Some initial benchmarks shows there is no much to gain for atomic input - `any_na_integer`. It can be up to 42% of the `anyNA` timing, but the numbers are so tiny that it doesn't change much, and there is always an overhead which can be easily seen on best case scenario. And for `list_any_na_integer` speed up of course better but still numbers are relatively small.

```r
library(data.table)

any_na_integer = data.table:::any_na_integer

N = 1e1 # to measure overhead

x = sample(N)            ## worst case
system.time(for(i in 1:100) any_na_integer(x))
#   user  system elapsed 
#  0.082   0.000   0.022 
system.time(for(i in 1:100) anyNA(x))
#   user  system elapsed 
#  0.001   0.000   0.002 
x = c(NA,sample(N-1))    ## best case
system.time(for(i in 1:100) any_na_integer(x))
#   user  system elapsed 
#  0.032   0.000   0.005 
system.time(for(i in 1:100) anyNA(x))
#   user  system elapsed 
#  0.002   0.000   0.002 

N = 1e9

setDTthreads(20L) # 50%
x = sample(N)            ## worst case
system.time(any_na_integer(x))
#   user  system elapsed 
#  1.619   2.743   0.301 
system.time(anyNA(x))
#   user  system elapsed 
#  0.588   0.000   0.588 
x = c(NA,sample(N-1))    ## best case
system.time(any_na_integer(x))
#   user  system elapsed 
#  1.838   0.000   0.105 
system.time(anyNA(x))
#   user  system elapsed 
#      0       0       0 

setDTthreads(40L) # 100%
x = sample(N)            ## worst case
system.time(any_na_integer(x))
#   user  system elapsed 
#  3.621   0.511   0.249 
system.time(anyNA(x))
#   user  system elapsed 
#  0.593   0.000   0.594 
x = c(NA,sample(N-1))    ## best case
system.time(any_na_integer(x))
#   user  system elapsed 
#  3.758   0.000   0.111 
system.time(anyNA(x))
#   user  system elapsed 
#      0       0       0 

list_any_na_integer = data.table:::list_any_na_integer

N = 1e8

setDTthreads(20L) # 50%
x = replicate(40, sample(N), simplify=F)            ## worst case
system.time(list_any_na_integer(x))
#   user  system elapsed 
#  6.763   6.859   1.648 
system.time(sapply(x, anyNA))
#   user  system elapsed 
#  2.399   0.404   2.803 
x = replicate(40, c(NA,sample(N-1)), simplify=F)    ## best case
system.time(list_any_na_integer(x))
#   user  system elapsed 
#  4.486   0.000   0.267 
system.time(sapply(x, anyNA))
#   user  system elapsed 
#      0       0       0 

setDTthreads(40L) # 100%
x = replicate(40, sample(N), simplify=F)            ## worst case
system.time(list_any_na_integer(x))
#   user  system elapsed 
#  8.246  21.610   0.931 
system.time(sapply(x, anyNA))
#   user  system elapsed 
#  2.433   0.144   2.577 
x = replicate(40, c(NA,sample(N-1)), simplify=F)    ## best case
system.time(list_any_na_integer(x))
#   user  system elapsed 
#  5.622   0.002   0.144 
system.time(sapply(x, anyNA))
#   user  system elapsed 
#      0       0       0 
```

PR #3610, opened 2019-05-28
take out macros from loops

jangorecki:
avoid overhead of R macros
no logic change, only take out macros out of the loops, and few DATAPTR removal

### DIFF
--- a/src/bmerge.c
+++ b/src/bmerge.c
@@ -344,7 +344,7 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
     double *dxc = REAL(xc);
     isInt64 = INHERITS(xc, char_integer64);
     uint64_t (*twiddle)(void *, int) = isInt64 ? &i64twiddle : &dtwiddle;
-    // TODO: remove this last remaining use of i64twiddle. remove DATAPTR too.
+    // TODO: remove this last remaining use of i64twiddle.
     ival.ull = twiddle(dic, ir);
     while(xlow < xupp-1) {
       mid = xlow + (xupp-xlow)/2;
--- a/src/bmerge.c
+++ b/src/bmerge.c
@@ -298,7 +298,7 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
     // ilow and iupp now surround the group in ic, too
   }
     break;
-  case STRSXP : {
+  case STRSXP :
     if (op[col] != EQ) error("Only '==' operator is supported for columns of type %s.", type2char(TYPEOF(xc)));
     ival.s = ENC2UTF8(STRING_ELT(ic,ir));
     while(xlow < xupp-1) {
@@ -337,7 +337,6 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
       xval.s = ENC2UTF8(STRING_ELT(ic, o ? o[mid]-1 : mid));
       if (xval.s == ival.s) tmpupp=mid; else ilow=mid;   // see above re ==
     }
-  }
     break;
   case REALSXP : {
     double *dic = REAL(ic);
--- a/src/bmerge.c
+++ b/src/bmerge.c
@@ -231,11 +231,13 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
     // it was checked in bmerge() that the types are equal
   } else xc = nqgrp;
   switch (TYPEOF(xc)) {
-  case LGLSXP : case INTSXP :   // including factors
-    ival.i = (col>-1) ? INTEGER(ic)[ir] : thisgrp;
+  case LGLSXP : case INTSXP : {  // including factors
+    const int *iic = INTEGER(ic);
+    const int *ixc = INTEGER(xc);
+    ival.i = (col>-1) ? iic[ir] : thisgrp;
     while(xlow < xupp-1) {
       mid = xlow + (xupp-xlow)/2;   // Same as (xlow+xupp)/2 but without risk of overflow
-      xval.i = INTEGER(xc)[XIND(mid)];
+      xval.i = ixc[XIND(mid)];
       if (xval.i<ival.i) {          // relies on NA_INTEGER == INT_MIN, tested in init.c
         xlow=mid;
       } else if (xval.i>ival.i) {   // TO DO: is *(&xlow, &xupp)[0|1]=mid more efficient than branch?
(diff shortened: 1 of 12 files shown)


jangorecki:
I removed some `DATAPTR` in `bmerge` also. They were used in `twiddle`, as a result they couldn't be `const double *` but just `double *`, address https://github.com/Rdatatable/data.table/issues/3301
some macros in bmerge were not removed as I want this PR to be very simple, and logic there would need to be rewritten more heavily.

jangorecki:
not sure how much noise is in this measurement, but `test.data.table` timings on
master: 58.5s
c-loops: 49.5s

src/dogroups.c
```diff
@@ -177,45 +180,45 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
         if (LOGICAL(verbose)[0]) { tblock[0] += clock()-tstart; nblock[0]++; }
       } else {
         // Fairly happy with this block. No need for SET_* here. See comment above.
-        for (k=0; k<grpn; k++) INTEGER(I)[k] = INTEGER(order)[ INTEGER(starts)[i]-1 + k ];
+        for (k=0; k<grpn; k++) iI[k] = iorder[ istarts[i]-1 + k ];
         for (j=0; j<length(SDall); j++) {
           size_t size = SIZEOF(VECTOR_ELT(SDall,j));
           target = VECTOR_ELT(SDall,j);
           source = VECTOR_ELT(dt,INTEGER(dtcols)[j]-1);
-          int *Id = INTEGER(I);
```
jangorecki: this was initialized sooner, and already used sooner

src/dogroups.c
```diff
@@ -177,45 +180,45 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
-            memcpy((char *)DATAPTR(VECTOR_ELT(SDall,j)),  // direct memcpy best here, for usually large size groups. by= each row is slow and not recommended anyway, so we don't mind there's no switch here for grpn==1
+            memcpy((char *)DATAPTR(VECTOR_ELT(SDall,j)),  // direct memcpy best here, for usually large size groups. by= each row is slow and not recommended anyway, so we don't mind there's no switch here for grpn==1 // TODO remove DATAPTR
                    (char *)DATAPTR(VECTOR_ELT(dt,INTEGER(dtcols)[j]-1))+rownum*size,
                    grpn*size);
             // SD is our own alloc'd memory, and the source (DT) is protected throughout, so no need for SET_* overhead
           }
           for (j=0; j<length(xSD); j++) {
             size_t size = SIZEOF(VECTOR_ELT(xSD,j));
-            memcpy((char *)DATAPTR(VECTOR_ELT(xSD,j)),  // ok use of memcpy. Loop'd through columns not rows
+            memcpy((char *)DATAPTR(VECTOR_ELT(xSD,j)),  // ok use of memcpy. Loop'd through columns not rows // TODO remove DATAPTR
```
jangorecki: in cases like this we have to `switch` and case by case? or is there any better way?

mattdowle: maybe call `memrecycle`

mattdowle:
I added a check to setcolorder at C level that it has a strict permutation. In future the (expensive) `anyDuplicated` at R level setcolcorder could be removed and left to C level, with the C level errors embellished and tested with coverage of course.
As I was going through checking I started to move variables down to local scope while I was there (particularly loop counters) and changed i++ to ++i.  We might as well give all compilers on all platforms, even old ones, the best chance:  https://www.quora.com/Do-you-use-i++-or-++i-in-a-for-loop-And-why/answer/Matt-Parsons-12
I removed curly braces around case statements where no local variables are created for that case.  When I see a curly after a case, I think "ok, local variable is here, I'm expecting to see one".  That's the only reason a case needs curlies otherwise the local variable declaration doesn't compile.

src/fmelt.c
```diff
@@ -478,29 +478,19 @@ SEXP getvaluecols(SEXP DT, SEXP dtnames, Rboolean valfactor, Rboolean verbose, s
...
+          for (int k=0; k<thislen; ++k)
             itarget[counter + k] = ithiscol[ithisidx[k]-1];
         } else {
           memcpy(itarget + j*data->nrow, ithiscol, data->nrow*size);
         }
-      }
-        break;
-      default : error("Unknown column type '%s' for column '%s'.", type2char(TYPEOF(thiscol)), CHAR(STRING_ELT(dtnames, INTEGER(thisvaluecols)[i]-1)));
+      } break;
+      default :
```
jangorecki: I would keep it in same line, so if you collapse `{` of switch then every case, including default, takes only single line

src/bmerge.c
```diff
@@ -298,7 +298,7 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
     // ilow and iupp now surround the group in ic, too
   }
     break;
-  case STRSXP : {
```
jangorecki: this `{` was here for consistency to other cases, and nice collapsing code

src/dogroups.c
```diff
@@ -48,8 +48,8 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
   iSD = PROTECT(findVar(install(".iSD"), env)); protecti++; // 1-row and possibly no cols (if no i variables are used via JIS)
   xSD = PROTECT(findVar(install(".xSD"), env)); protecti++;
   R_len_t maxGrpSize = 0;
-  const int *ilens = INTEGER(lens);
-  for (R_len_t i=0; i<LENGTH(lens); i++) {
+  const int *ilens = INTEGER(lens), n=LENGTH(lens);
+  for (R_len_t i=0; i<n; ++i) {
```
jangorecki: @mattdowle in this commit many changes from `i++` to `++i` or reverse, but at the same time no logic change, what is the reason for that?

src/bmerge.c
mattdowle: by collapsing do mean in editor?  If we add `{` to all `case:` I don't think that's very standard.  I wrote my thinking here:  https://github.com/Rdatatable/data.table/pull/3610#issuecomment-497099447

src/dogroups.c
mattdowle: Did you see https://github.com/Rdatatable/data.table/pull/3610#issuecomment-497099447?

jangorecki: found answer in PR comment

src/bmerge.c
jangorecki: yes, collapsing code in IDE, makes it very convenient

src/fmelt.c
mattdowle: I like long lines as you know, but this seems a step too far to put the error on the same line as the default:  For the reason that the other cases in this switch aren't like that due to being bigger.  When all the cases are one short line, then yes but not in this case. Consistency within the switch.

jangorecki: then what can help here is
```
default : {
  error("a");
}
```
collapse-able to single line

PR #3612, opened 2019-05-29
add message that datatable.nomatch will be deprecated

### LINKED ISSUE #3585: deprecate `datatable.nomatch`; message first step
_Originally posted by @mattdowle in https://github.com/Rdatatable/data.table/pull/3578#issuecomment-494912023_

I fear it's not sufficient to just document this. There should be something more robust. I've been aware of this issue for some time and have not created any new options that affect behavior, only to manage migration temporarily and stated in news that those options will eventually be removed (e.g. `datatable.old.unique.by.key`).   We only have 1 long-standing option which changes the results: `datatable.nomatch`.  If a user changes that option for their own usage (because they prefer the nomatch=0L default), it could affect packages behavior. **Even packages that didn't set any datatable options and were completely compliant with the text in this PR.**

`datatable.naturaljoin` is now the 2nd one. Can we find an acceptable way for natural join to be explicit so we don't need the new option?

Here are all the options as of now in dev : https://github.com/Rdatatable/data.table/blob/master/R/onLoad.R#L44.  Other than the those two (`datatable.nomatch` and `datatable.naturaljoin`) they all affect printing, optimization, or are temporary for migration.

This does mean we should try to remove `datatable.nomatch` too,  or find a way for code inside packages to ignore its value.

jangorecki:
Closes #3585 (revised)
First step is to inform and communicate.


mattdowle:
Not sure if you saw my earlier comment here : https://github.com/Rdatatable/data.table/pull/3578#issuecomment-495317059.  This would allow us to not introduce the new `datatable.natural` option.  And we'd deprecate `datatable.nomatch` too.  There would be no need to document unsafe options, because there wouldn't be any unsafe options.

jangorecki:
Yes, but it is less trivial, and need message about `nomatch` option anyway. Deprecating `nomatch` will takes ages, this PR is first step. I can remove naturaljoin option altogether in this PR.

mattdowle:
Yes deprecating nomatch option would need time and communication, and a message would be first step.

>  I can remove naturaljoin option altogether in this PR.

Do you mean you agree with my suggestion to use `on=list()` to trigger natural join,  or some other way to avoid the option?

jangorecki:
I agree but I don't want to include that in this PR. It would be good to have it in separate PR.

PR #3630, opened 2019-06-05
Add c, seq, mean methods so ITime retains class

### LINKED ISSUE #3628: c(ITime) and seq( ITime ) lose ITime attribute and coerce to integer
```
t1 <- seq(from = as.ITime("0:00"), to = as.ITime("0:40"), by = as.ITime("0:06"))
str(t1)
# int [1:7] 0 360 720 1080 1440 1800 2160

# workaround:
t2 <- as.ITime(seq(from = as.ITime("0:00"), to = as.ITime("0:40"), by = as.ITime("0:06")))
str(t2)
# 'ITime' int [1:7] 00:00:00 00:06:00 00:12:00 00:18:00 00:24:00 00:30:00 00:36:00

```

Ran into it while answering https://stackoverflow.com/a/56454903/3817004

Related issue: [unique( ITime ) loses ITime attribute and coerces to integer](https://github.com/Rdatatable/data.table/issues/249) ?

Just recognized that the same happens with `c()`:

```
c(as.ITime("0:00"), as.ITime("1:00:00"))
# [1]    0 3600
```

Output of sessionInfo():

```
R version 3.6.0 (2019-04-26)

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] magrittr_1.5      data.table_1.12.3

[1] compiler_3.6.0  hms_0.4.2       tools_3.6.0     pkgconfig_2.0.2 rlang_0.3.4  
```

MichaelChirico:
Closes #3628 

Also deleted `split.IDate` and `cut.IDate` methods. They are untested with no examples, and I really don't know what the expected behavior is. `split` returns a `list` and really the important part of the method is how to handle the splitting factor `f`, not the first argument. `split.default` already retains the `IDate` class:

```
dput(split.default(as.IDate(0L), 'foo'))
# list(foo = structure(0L, class = c("IDate", "Date")))
```

And `cut` returns a `factor` so I don't know how `as.IDate` is supposed to operate on that.


MichaelChirico:
appveyor fail appears spurious

jangorecki:
Ideally to raise warning about functions to be deprecated, then in next version remove. Also mention in NEWS.

MichaelChirico:
In this case, I don't think those functions work at all, so I'm not sure how anybody could have used them... a Note in NEWS maybe

jangorecki:
I understand but it is unwise just to drop exported functions without any warning.

PR #1236, opened 2015-07-21
fixes #504 fread now handles all kind of NAs without coercion to char

dselivanov:
Tests added.  Test 882 adjusted to match correct behaviour. 
I tested this a lot (also `test.data.table()` helps! ), but may be we need to add more tests for this. I believe I need some help with writing good tests.

Now `na.strings = NULL` means no coercion at all. Old hardcoded handling for "NA" string removed from everywhere.
@arunsrinivasan, please review code before merging. This PR touches a lot of `fread.c` internals. It affects `Strto*()` functions which are workhorses for data reading and conversion.
Also It would be great if @mattdowle also will take a look, because as I see he wrote most of the `fread` code.

PR #3648, opened 2019-06-18
Fix for Intel compiler bug

### LINKED ISSUE #3647: forder with icc 2019/openmp 4 skips radix sort on final descent
Intel's compiler 2019 (Version 19.0.4.243 Build 20190416) seems to be allergic to setting the value of skip (forder.c:905) by calculation for its side-effect use in forder.c:922.  It's probably optimized out because the second loop contains the same body without the `skip` condition.

Clearly a compiler bug and I reported this issue to the Intel guys.

However, to be more explicit I propose to calculate the value of skip afterwards and separate from the key counting.

Any dataset with group size >256 should do.

`#` `Output of sessionInfo()`
    > sessionInfo()
    R version 3.6.0 (2019-04-26)
    Running under: openSUSE Leap 42.3
    
    Matrix products: default
    BLAS/LAPACK: /opt/intel/compilers_and_libraries_2019.4.243/linux/mkl/lib/intel64_lin/libmkl_intel_lp64.so
    
    locale:
     [1] LC_CTYPE=C.UTF-8     LC_NUMERIC=C         LC_TIME=C           
     [4] LC_COLLATE=C         LC_MONETARY=C        LC_MESSAGES=C       
     [7] LC_PAPER=C           LC_NAME=C            LC_ADDRESS=C        
    [10] LC_TELEPHONE=C       LC_MEASUREMENT=C     LC_IDENTIFICATION=C 
    
    attached base packages:
    
    other attached packages:
    [1] data.table_1.12.3
    
    [1] compiler_3.6.0

hroptatyr:
Closes #3647 
Calculate `skip` condition after and separate from key counting.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -129,7 +129,6 @@
     # TRUE
     ```
 
-
 #### BUG FIXES
 
 1. `first`, `last`, `head` and `tail` by group no longer error in some cases, [#2030](https://github.com/Rdatatable/data.table/issues/2030) [#3462](https://github.com/Rdatatable/data.table/issues/3462). Thanks to @franknarf1 for reporting.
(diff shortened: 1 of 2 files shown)


mattdowle:
Great, thanks for the report and fix. I'll just add a minor tweak, a news item, and merge.

mattdowle:
This was a code contribution, so I'm sorry @hroptatyr, I should have added you to DESCRIPTION as a contributor at the time. I'll correct my mistake now in a commit to master linking to this PR.
Found as a result of following up on discussion at the end of #1236.
(This contribution was after the license change from GPL to MPL so it doesn't need to be included in the forthcoming retrospective-permission PR.)

Issue #2209, opened 2017-06-19
[Doc request] add quote/eval/substitute guidance back to FAQ


franknarf1:
Elevating [my comment](https://github.com/Rdatatable/data.table/issues/1936#issuecomment-281465037) to a request. I just saw this brought up [on SO again today](https://stackoverflow.com/questions/44632867/passing-columns-in-a-variable-dynamically-when-aggregating-in-data-table) and it seems to come up frequently.

Here is [the old version](https://github.com/Rdatatable/data.table/blob/c682d7e55dcc4b430d69a4e68b5cbbeb44889734/vignettes/datatable-faq.Rmd), for reference:

> 1.6 OK, but I don't know the expressions in advance. How do I programatically pass them in?
> 
> To create expressions use the `quote()` function. We refer to these as `quote()`-ed expressions to
> save confusion with the double quotes used to create a character vector such as `c("x")`. The simplest `quote()`-ed expression is just one column name:
> 
> ```{r}
> DT = data.table(x = 1:3)
> q = quote(x)
> 
> #return a vector
> DT[ , eval(q)]
> 
> q = quote(list(x))
> 
> #return a one-column 'data.table'
> DT[ , eval(q)]
> ```
> 
> Since these are _expressions_, we are not restricted to column names only:
> 
> ```{r}
> q = quote(mean(x))
> DT[ , eval(q)]  # identical to 'DT[ , mean(x)]'
> 
> DT = data.table(x = 1:3, y = 4:6, z = 7:9)
> q = quote(list(x, sd(y), mean(y*z)))
> DT[ , eval(q)]  # identical to 'DT[ , list(x, sd(y), mean(y*z))]'
> ```
> 
> However, if it's just simply a vector of column names you need, it may be simpler to pass a character vector to `j` and use `with = FALSE`.
> 
> To pass an expression into your own function, one idiom is as follows:
> 
> ```{r}
> DT = as.data.table(iris)
> setkey(DT, Species)
> myfunction = function(dt, expr) {
>     e = substitute(expr)
>     dt[ , eval(e), by = Species]
> }
> myfunction(DT, sum(Sepal.Width))
> ```
> 
[...]

PR #3649, opened 2019-06-18
Link to FAQ 1.11

hadley:
I'm pretty sure this is supposed to link to 1.11 ("What is the difference between X[Y] and merge(X, Y)?") not ("Anything else about X[Y, sum(foo*bar)]?")

I'd also be happy to add a link directly to the FAQ (https://cran.r-project.org/web/packages/data.table/vignettes/datatable-faq.html#MergeDiff) if you thought that was useful.

### DIFF
--- a/man/merge.Rd
+++ b/man/merge.Rd
@@ -67,7 +67,7 @@ control for providing the columns to merge on with the help of the newly impleme
 \code{by.x} and \code{by.y} arguments.
 
 For a more \code{data.table}-centric way of merging two \code{data.table}s, see
-\code{\link{[.data.table}}; e.g., \code{x[y, \dots]}. See FAQ 1.12 for a detailed
+\code{\link{[.data.table}}; e.g., \code{x[y, \dots]}. See FAQ 1.11 for a detailed
 comparison of \code{merge} and \code{x[y, \dots]}.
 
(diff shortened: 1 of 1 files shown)


franknarf1:
It might be a side effect of the removal of the old 1.6 https://github.com/Rdatatable/data.table/issues/2209 (Fortunately, it looks like your fix the only one needed. I don't see any other references to 1.*)

MichaelChirico:
https://github.com/Rdatatable/data.table/commit/0575dcc0856066be43601bf12176d1d866274553

I think this is the commit that broke the ordering. Moving to a hard link to the tag would indeed probably be best but requires internet connection. Is there a way to link to the installed vignette instead?

franknarf1:
@MichaelChirico I usually just refer to 

> FAQ question "yada yada?"

[on SO](https://stackoverflow.com/a/56569680) anyways. You can't reliably link to the local vignette if it's not built (or whatever it's called), I guess.

hadley:
Unfortunately we haven't found a good way to local vignettes. In the tidyverse, we usually get around this by linking to the website instead (although this is a little tricky if we have websites for both CRAN and github versions of the package).

mattdowle:
Great. Thanks, @hadley.

PR #3661, opened 2019-06-26
coalesce works also on dt, closes #3660

### LINKED ISSUE #3660: coalesce on a list: data.frame vs data.table
I was playing around with the new `coalesce` function (thanks a lot for your work with this!), especially the option that "`x` may be a `list` of atomic vectors". This works when the "list" is a `data.frame`, but fails when it is a `data.table`. Am I overinterpreting the meaning of "a `list` of atomic vectors", or may we expect `coalesce` to work also on a `data.table` when it happily accepts a `data.frame`? 

Using data from `?coalesce`:

    x = c(11L, NA, 13L, NA, 15L, NA)
    y = c(NA, 12L, 5L, NA, NA, NA)
    z = c(11L, NA, 1L, 14L, NA, NA) 

    # works
    # a "real" list of vectors
    coalesce(list(x, y, z))
    # [1] 11 12 13 14 15 NA

    # works
    # vectors collected in a data.frame
    coalesce(data.frame(x, y, z))
    # [1] 11 12 13 14 15 NA

    # fails
    # vectors collected in a data.table
    coalesce(data.table(x, y, z))
    # Error in coalesce(data.table(x, y, z)) : 
    #  Item 2 is length 5 but the first item is length 6. Only singletons are recycled.

The error message is generated [here](https://github.com/Rdatatable/data.table/blob/dbb0d0b196530f52c6d54d6f2195bda48a955e78/src/coalesce.c#L28), but being a C illiterate, I can unfortunately not suggest any cure. 

-----

My use case was something like:
    
    # works 
    df <- data.frame(x, y, z)
    df$coal <- coalesce(df)
    
    # fails
    dt <- data.table(x, y, z)
    dt[ , coal := coalesce(dt)]
    # Error in coalesce(dt) : 
    # Item 2 is length 5 but the first item is length 6. Only singletons are recycled.

---------

`data.table` 1.12.3 IN DEVELOPMENT built 2019-06-19 19:07:48 UTC

jangorecki:
closes #3660

### DIFF
--- a/src/coalesce.c
+++ b/src/coalesce.c
@@ -1,7 +1,8 @@
 #include "data.table.h"
 
 SEXP coalesce(SEXP x, SEXP inplaceArg) {
-  if (TYPEOF(x)!=VECSXP) error("argument 'x' must be a list");
+  if (TYPEOF(x)!=VECSXP)
+    error("Internal error in coalesce.c: input is list(...) at R level"); // # nocov
   if (!IS_TRUE_OR_FALSE(inplaceArg))
     error("Internal error in coalesce.c: argument 'inplaceArg' must be TRUE or FALSE"); // # nocov
   const bool inplace = LOGICAL(inplaceArg)[0];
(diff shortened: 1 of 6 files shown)


src/coalesce.c
```diff
@@ -1,42 +1,56 @@
 #include "data.table.h"
 
-SEXP coalesce(SEXP x, SEXP values, SEXP inplaceArg) {
-
-  if (!isVectorAtomic(x)) error("argument 'x' must be an atomic vector");
+SEXP coalesce(SEXP x, SEXP inplaceArg) {
+  if (TYPEOF(x)!=VECSXP) error("argument 'x' must be a list");
```
MichaelChirico: is this a `nocov`/internal error? since `list(...)` is in R?

mattdowle: Oh yes, it is now. Good spot. Will change.

PR #3676, opened 2019-07-03
Closes #3664 -- diligently avoid partial argument matching

### LINKED ISSUE #3664: Disable partial match warnings when running test.data.table
Since R is such a dynamic and permissive language I often find it hard to spot bugs that are result of my poor typing skills.  So, to help myself I set the following options, along with others,  in my .Rprofile file: `options(warnPartialMatchArgs=TRUE, warnPartialMatchAttr=TRUE,warnPartialMatchDollar=TRUE)`.  Recently, I have gotten interested in writing tests for R code and I was aware that `data.table` has extensive testing so I started learning it to get a little feel for it.  So, I tried running the `test.data.table()` function today.  After my amazement at the thoroughness of the testing used by data.table subsided I realized that it reported that there were some errors.  However, after a little bit of digging about I realized that it was from my options settings!  So, is it possible to modify the `test.data.table()` function to check if these options are set before running the tests?  And, it would be awesome if the function disabled these options while the tests are running and then reset the options to original ones.  Just fyi, I use [Patrick Burns's withOptions](https://www.burns-stat.com/the-options-mechanism-in-r/) function to temporarily turn off these functions for running the test.data.table function. 

Currently, I do (using Patrick Burns's withOptions functions)
```
withOptions(list(warnPartialMatchArgs=F,warnPartialMatchAttr=F,warnPartialMatchDollar=F), test.data.table())
```
But it would be awesome if I could just do 
```
test.data.table()
```

MichaelChirico:
#3664 

Confirmed we're doing OK for `attr` with:

```
grep -r "\battr[(]" R | grep -v "exact=TRUE"
```

There's one call to `attr` that's not using `exact = TRUE` inside of `.shallow`, since it's actually being called as `attr<-` (didn't look so at first glance but it threw "too many arguments" error when I first tried).

A bunch of partial matches were being used in `test.data.table()`. With the commit here, `test.data.table()` runs fine using the options suggested by @vlulla.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -222,7 +222,7 @@
 
 11. We intend to deprecate the `datatable.nomatch` option, [more info](https://github.com/Rdatatable/data.table/pull/3578/files). A message is now printed upon use of the option (once per session) as a first step. It asks you to please stop using the option and to pass `nomatch=NULL` explicitly if you require inner join. Outer join (`nomatch=NA`) has always been the default because it is safer; it does not drop missing data silently. The problem is that the option is global; i.e., if a user changes the default using this option for their own use, that can change the behavior of joins inside packages that use `data.table` too. This is the only `data.table` option with this concern.
 
-12. Cleaned up internal code to only use exact argument matching (especially using `exact=TRUE` for calls to `attr`). This is for both safety and efficiency, [#3664](https://github.com/Rdatatable/data.table/issues/3664). Thanks to @vlulla who ferreted this out by using the `warnPartialMatchArgs` and `warnPartialMatchAttr` options.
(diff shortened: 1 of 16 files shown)


MichaelChirico:
i thought of that and convinced myself we didn't want such tests but maybe you're right. If so, though, we should create tests explicitly for that purpose. (and skip them if the `warnPartialMatch` options are set, I guess)

vlulla:
@MichaelChirico you did a thorough job of fixing lots of files!  Great work!  And, thank you!

mattdowle:
Very nice PR - thanks @vlulla and @MichaelChirico!  I just did a minor tweak to the news item and turned these options on so the tests always run with them on, to avoid any future partial argument matching creeping back in.  One reveal on nanotime:  https://github.com/eddelbuettel/nanotime/pull/49

MichaelChirico:
great! I meant to suggest the same for turning on the option. onwards and upwards!

Issue #3657, opened 2019-06-25
data.table::if_else() or data.table::iif()


shrektan:
`base::ifelse()` is notorious slow and will change the attributes somehow. 

```r
ifelse(TRUE, as.Date('2019-06-25'), as.Date('2019-06-25'))
# [1] 18072
```

dplyr implements `dplyr::if_else()` which I think is good. However, I always reluctant to use dplyr because of it's heavy dependencies. 

So I'm proposing a feature request here... I think both `data.table::if_else()` and `data.table::iif()` are good name candidates (though the former conflicts with `dplyr::if_else()`)

jangorecki:
alternatively `fifelse`

HughParsonage:
For comparison, [`hutils::if_else`](https://github.com/HughParsonage/hutils/blob/master/R/if_else.R) 

from the description: 

> Lightweight `dplyr::if_else` with the virtues and vices that come from such an approach.

2005m:
I have shared with Matt a prototype implemented in C .
It has good performance without using openmp. (Up to x10 faster than `base::ifelse` and `dplyr::if_else`)
I also have a version that uses openmp which is useful when inputs are large.
I have tried to stick to the philosophy of the data.table package as much as I am aware of it.
The prototype is compatible with R 3.1.0 and has no dependency.

Now, `base::ifelse` and `dplyr::if_else` have different behaviours.
I am trying to list all the use cases of a "fifelse" function and check which behaviours is wanted and which is not.Below you will find some examples.

**Example 1:** (`base::ifelse` does not keep dates format)

```
dates <- as.Date(c("2011-01-01","2011-01-02","2011-01-03","2011-01-04","2011-01-05"))

dplyr::if_else(dates == "2011-01-01", dates - 1, dates)
#[1] "2010-12-31" "2011-01-02" "2011-01-03" "2011-01-04" "2011-01-05"

base::ifelse(dates == "2011-01-01", dates - 1, dates)
#[1] 14974 14976 14977 14978 14979
```

Which behaviour do we want? My prototype currently outputs the same as dplyr which in
my view makes more sense but I can add a flag to have the behaviour of base::ifelse as well?
  
**Example 2:** (`base::ifelse` does not keep factor class)

```
v <- factor(sample(letters[1:5], 10, replace = TRUE))

dplyr::if_else(v %in% c("a","b","c"), v, factor(NA))
#[1] <NA> c    b    <NA> <NA> b    c    c    c    <NA>
[...]

PR #3678, opened 2019-07-04
fifelse - Issue #3657

2005m:
#3657


R/fifelse.R
```diff
@@ -0,0 +1,10 @@
+fifelse <- function(test, yes, no){
+  switch(class(yes),
```
jangorecki: class might return 2+ length char vector, is it handled by switch?

src/fifelse.c
```diff
@@ -0,0 +1,338 @@
...
+  
+  // Check if same class
+  if(!R_compute_identical(class_type,getAttrib(b,R_ClassSymbol),0))
+    error("Item 'yes' has different class than item 'no'. Please make sure that they have same class type.");
+  
+  // Check if factor
+  if(isFactor(a))
+  {
+    a = PROTECT(asCharacterFactor(a));
+    b = PROTECT(asCharacterFactor(b));
```
jangorecki: Afair this will make factor class always slower than character, would be nice to have it optimised where possible

src/fifelse.c
```diff
@@ -0,0 +1,338 @@
...
+      break;
+    }
+    break;
+    
+  default: error("Type %s is not supported.",type2char(ta)); break;
+  }
+  
+  // Check if class type is Date and adjust
+  if(!isNull(class_type))
+    if( CHAR(STRING_ELT(class_type,0)) == CHAR(STRING_ELT(Rf_mkString("Date"),0)) )
```
jangorecki: We have such character strings predefined in init.c, those should be preferred over mkString

src/fifelse.c
```diff
@@ -0,0 +1,338 @@
...
+      setAttrib(result, R_ClassSymbol, mkString("Date"));
```
jangorecki: It might be better to use copyMostAttributes to handle more classes, not just Date. See nafill.c

man/fifelse.Rd
```diff
@@ -0,0 +1,45 @@
...
+### Example 1
+dates <- as.Date(c("2011-01-01","2011-01-02","2011-01-03","2011-01-04","2011-01-05"))
+ifelse(dates == "2011-01-01", dates - 1, dates) # does not display the desired output
+# [1] 14974 14976 14977 14978 14979
+data.table::fifelse(dates == "2011-01-01", dates - 1, dates)
+# [1] "2010-12-31" "2011-01-02" "2011-01-03" "2011-01-04" "2011-01-05"
+
+### Example 2
+x   <- c(-5e4L:5e4L) < 0
+microbenchmark::microbenchmark(
```
jangorecki: Those would be ideally to put in NEWS.md entry, not the package examples as user running those might not have those packages installed

R/fifelse.R
```diff
@@ -0,0 +1,10 @@
+fifelse <- function(test, yes, no){
+  switch(class(yes)[1L],
```
MichaelChirico: maybe best to use `is.factor`.

What's the expected behavior for something like `ifelse(TRUE, data.table(1:5), data.table(5:1))`? Same as for any other `list`? If so, should use `is.list` instead as well

2005m: This is a tricky one! It is also related to the examples I mentioned in my post of [#3657 ](https://github.com/Rdatatable/data.table/issues/3657). 
I was asking for your thoughts given that it was not clear to me what should be returned for `list`?
  
You said:
_"I'm not sure how much attention we should devote to handling list output just yet, in any case._"

And I hence limited myself to implement something very basic.
  
As I mentioned `ifelse` does not always make sense to me.
`dplyr::if_else` and `htulis::if_else` are more strict, so is `fifelse`, but works sometimes differently.

**1/** Right now, here is what would be returned by fifelse:
  
```
fifelse(TRUE, data.table(1:5), data.table(5:1))
# Error in fifelse(test, yes, no) : Type list is not supported.

fifelse(TRUE, list(1:5), list(5:1))
# Error in fifelse_impl(test, unlist(yes), unlist(no)) : 
# Length of 'yes' and 'no' must be 1 or equal to length of 'test'. 

```
I believe the error message should be changed so that `data.table` is consistent with a `list`. the length of `yes` and `no` should be either equal to `1` or the length of `test`.
Anyway, `ifelse` would work for the above examples and `dplyr::if_else` would only work for 
the second example.

**2/** Note that the following would work:
  
```
fifelse(-2L:2L < 0, list(1:5), list(5:1))
# [[1]]
# [1] 1
# 
# [[2]]
# [1] 2
# 
# [[3]]
# [1] 3
# 
# [[4]]
# [1] 2
# 
# [[5]]
# [1] 1
```
but would return something different than `ifelse` and `dyplr::if_else`.
I was thinking about adding support to data.frame and data.table and treat them like a `list`,
what do you think?

If you refer to my post on Issue [#3657 ](https://github.com/Rdatatable/data.table/issues/3657) you will find more examples with `list`, for which
I am not really sure what to return. As shown ifelse returns (in some cases) outputs that I do not
understand or outputs that loose the class type.

In any cases, I am happy to implement what the members of the Rdatatable team believe make sense. Please do let me know what you think the fifelse function should return in these examples (including [#3657 ](https://github.com/Rdatatable/data.table/issues/3657) ). It is quite difficult to say what is right or wrong here.

As to your suggestion to use `is.factor` and `is.list`, I believe that would involve more test and `if` functions and would slow down the code compare to a single `switch` test that does a dispatch to the  corresponding value. I may be wrong, if so, please feel free to suggest an alternative implementation and I would not mind at all amending the code.

MichaelChirico: indeed... this rabbit hole is why I originally suggested to keep away. Actually I'd expect `fifelse(TRUE, list(1:5), list(5:1))` to return `list(1:5)` -- this is how `ifelse` would be used to replace elements of a list column, for example.

I would stick to using `if (is.factor)` and `if (is.list)` branches, with the `list` branch requiring `(length(yes) == length(test) || length(yes) == 1L) && (length(no) == length(test) || length(no) == 1L)`. The efficiency difference (inasmuch as there is one, which I'm not sure of) will be on the order of nanoseconds, and only happens once per function call.

MichaelChirico: i.e. I would expect the following normal usage of `fifelse`:

```
DT = data.table(a = 0:5, b = list(1, list(2:3), list(4:6), list(6:4), list(3:2), 1))
DT[ , b := fifelse(a > 2, b, list(0))]
#now, DT$b == list(0, 0, 0, list(6:4), list(3:2), 1), i.e. rows 4,5,6 unaffected
```

List columns are a bit funny... if I'm not mistaken, it would also be consistent (and convenient) if singleton replacements could be en-listed (wrapped correctly in `list()`), i.e.:

```
DT = data.table(a = 0:5, b = list(1, list(2:3), list(4:6), list(6:4), list(3:2), 1))
DT[ , b := fifelse(a > 2, b, 0)]
```

Could also work (a bit of an anti-pattern but I think the intent is clear). This is consistent with `list` assignment directly:

```
DT = data.table(a = list(1))
DT[1, a := 0]
# now DT$a == list(0)
```

2005m: I have pushed a new version. Let's see if it passes tests. 
Your examples should all work now.
`is.list` is not needed anymore as list are now handle directly by the C code.
This test `(length(yes) == length(test) || length(yes) == 1L) && (length(no) == length(test) || length(no) == 1L)` is not needed either as something similar is done in the C code.
More tests have been added to `test.Rraw` file.
When I have time, I will be looking at moving the rest of R code to C (if possible) so `is.factor` or `class` are not needed. I will be also looking at the factor/character topic mentioned by @jangorecki .

Thanks for the feedback.

src/init.c
```diff
@@ -170,6 +171,7 @@ R_CallMethodDef callMethods[] = {
 {"CinitLastUpdated", (DL_FUNC) &initLastUpdated, -1},
 {"Ccj", (DL_FUNC) &cj, -1},  
 {"Ccoalesce", (DL_FUNC) &coalesce, -1},
+{"CfifelseR", (DL_FUNC) &fifelseR, 3},
```
MichaelChirico: Huh, TIL what that third argument is for (number of arguments)...

@mattdowle is there any reason we don't specify the number of arguments for any other routines?

> For routines with a variable number of arguments..., one specifies -1 for the number of arguments which tells R not to check the actual number passed.

Maybe an efficiency thing (# args is not checked)?

MichaelChirico:
Minor point that if we're going to call it `fifelse`, should we also change `coalesce` to `fcoalesce`? I guess the major difference is there's no `base::coalesce`...

So far in our namespace we have the following `^f` functions:

| Function | Conflict in `base`? |
| --------- | ------------------ |
| `fintersect` | ✅ |
| `foverlaps` | ❌  |
| `frank{v}` | ✅ |
| `f{read,write}` | ❌ |
| `froll{mean,sum}` | ❌ |
| `fsetdiff` | ✅ |
| `fsetequal` | ✅ |
| `fsort` | ✅ |
| `funion` | ✅ |

inst/tests/tests.Rraw
```diff
@@ -15194,6 +15194,78 @@ l = replicate(ceiling(log10(.Machine$integer.max)), rep(1L, 10L), simplify = FAL
...
+test(2062.035, fifelse(test_vec, as.Date("2011-01-01"), FALSE), error = "Item 'yes' is of type double but item 'no' is of type logical. Please make sure they are of the same type.")
+test(2062.036, fifelse(test_vec_na, 1+0i, 0+0i), as.complex(out_vec_na))
+test(2062.037, fifelse(test_vec_na, rep(1+0i,12L), 0+0i), as.complex(out_vec_na))
+test(2062.038, fifelse(test_vec_na, rep(1+0i,12L), rep(0+0i,12L)), as.complex(out_vec_na))
+test(2062.039, fifelse(test_vec_na, 1+0i, rep(0+0i,12L)), as.complex(out_vec_na))
+test(2062.040, fifelse(test_vec, as.raw(0), as.raw(1)), error = "Type raw is not supported.")
+test(2062.041, fifelse(TRUE,1,as.Date("2019-07-07")), error = "Item 'yes' has different class than item 'no'. Please make sure that they have same class type.")
+test(2062.042, fifelse(TRUE,1L,factor(letters[1])), error = "Item 'yes' has different class than item 'no'. Please make sure that they have same class type.")
+test(2062.043, fifelse(TRUE, list(1:5), list(5:1)), list(1:5))
+test(2062.044, fifelse(as.logical(NA), list(1:5), list(5:1)), list(as.integer(NA)))
```
MichaelChirico: the `all(is.na(test))` case is interesting, can we add some more tests? I guess the behavior here is right, namely that the output should be `NA` & match the type of `yes`/`no`

MichaelChirico: e.g. I notice:

```
dput(fifelse(NA, list(0), list(1)))
# list(NA_integer_)
```

Here `yes` is `list(numeric())` but the output is `list(integer())`.

I think this is fine (it would I think be overly complicated to match the recursive structure completely), but we should add a test & document.

Would `list(logical())` be preferred? Based on type heirarchy...

inst/tests/tests.Rraw
```diff
@@ -15194,6 +15194,78 @@ l = replicate(ceiling(log10(.Machine$integer.max)), rep(1L, 10L), simplify = FAL
...
+test(2062.045, fifelse(FALSE, list(1:5), list(5:1)), list(5:1))
+test(2062.046, fifelse(TRUE, data.table(1:5), data.table(5:1)), data.table(1:5))
+test(2062.047, fifelse(FALSE, data.table(1:5), data.table(5:1)), data.table(5:1))
+test(2062.048, fifelse(TRUE, data.frame(1:5), data.frame(5:1)), data.frame(1:5))
+test(2062.049, fifelse(FALSE, data.frame(1:5), data.frame(5:1)), data.frame(5:1))
+test(2062.050, fifelse(c(TRUE,FALSE), list(1:5,6:10), list(10:6,5:1)), list(1:5,5:1))
+test(2062.051, fifelse(c(as.logical(NA),TRUE), list(1:5,6:10), list(10:6,5:1)), list(as.integer(NA),6:10))
+test(2062.052, fifelse(c(FALSE,TRUE), list(1:5,6:10), list(10:6,5:1)), list(10:6,6:10))
+test(2062.053, fifelse(c(as.logical(NA),TRUE), list(1:5), list(10:6,5:1)), list(as.integer(NA),1:5))
```
MichaelChirico: This doesn't error because `yes` is treated as a singleton, right? I believe this is confirmed by test 55 below

MichaelChirico:
Just finished a pass-through of reading the tests. Behavior looks good overall, made some minor changes to error messages.

Thanks!

2005m:
Something is not right with the changes you made. Appveyor and travis are failling.
I will look into it when I find some time.

R/test.data.table.R
```diff
@@ -366,12 +366,20 @@ test = function(num,x,y=TRUE,error=NULL,warning=NULL,output=NULL,message=NULL) {
     # For test 617 on r-prerel-solaris-sparc on 7 Mar 2013
     # nocov start
     if (!fail) {
-      cat("Test",numStr,"ran without errors but failed check that x equals y:\n")
-      cat("> x =",deparse(xsub),"\n")
-      if (is.data.table(x)) compactprint(x) else {cat("First 6 of ", length(x)," (type '", typeof(x), "'): ", sep=""); print(head(x))}
```
MichaelChirico: was seeing this in the output

Screenshot 2019-07-08 at 5 13 59 PM

so originally came here to add the '\n' & stayed to clean things up a bit.

man/fifelse.Rd
```diff
@@ -0,0 +1,35 @@
...
+}
+\value{
+A vector of the same length as \code{test} and attributes as \code{yes}. Data values are taken from the values of \code{yes} and \code{no}.
+}
+\examples{
+### Example 1
+dates <- as.Date(c("2011-01-01","2011-01-02","2011-01-03","2011-01-04","2011-01-05"))
+ifelse(dates == "2011-01-01", dates - 1, dates)
+# [1] 14974 14976 14977 14978 14979
+data.table::fifelse(dates == "2011-01-01", dates - 1, dates)
```
MichaelChirico: i wouldn't use `data.table::` in the examples for the package

2005m: That is fine by me. I'll change it. Thanks.

inst/tests/tests.Rraw
2005m: I added test 61 for the singleton

man/fifelse.Rd
2005m: I did not finish the openmp version (will commit when I can) but as discussed it is faster for large vectors but slower for small one ( -1e2:1e2 < 0 ). It makes fifelse slower than all the other functions (i.e. dplyr::fi_else, hutils::if_else and base::ifelse ). fifelse works for huge vectors like -1e9L:1e9L < 0 and the others run out of memory.

2005m: @MichaelChirico , please see branch **fifelse-openmp** in my repo. You will find a function called `fifelse2` which is using multi-threading.  I couldn't push the test/benchmark higher as my laptop is really old.

```
# fifelse2 uses 2 threads.

x   <- -1e2L:1e2L < 0
   data.table::fifelse2(x, 1L, 0L),
   data.table::fifelse(x, 1L, 0L),
   times = 100L
 )
# Unit: microseconds
#                             expr    min      lq     mean median     uq     max neval
#  data.table::fifelse2(x, 1L, 0L) 67.138 71.2010 77.01662 72.698 75.691 418.221   100
#   data.table::fifelse(x, 1L, 0L) 14.112 14.7535 17.41797 18.816 19.671  26.514   100


x   <- -1e8L:1e8L < 0 # 763Mb
   data.table::fifelse2(x, 1L, 0L),
   data.table::fifelse(x, 1L, 0L),
   times = 100L
 )
# Unit: milliseconds
#                             expr      min       lq     mean   median       uq      max neval
#  data.table::fifelse2(x, 1L, 0L) 472.9533 489.7613 497.1877 495.4631 500.5052 607.3494   100
#   data.table::fifelse(x, 1L, 0L) 639.5606 658.2875 663.7909 661.8020 666.5692 739.6911   100

```

Please let me know what you think. Thank you.

man/fifelse.Rd
```diff
@@ -37,6 +37,39 @@ data.table::fifelse(v \%in\% c("a","b","c"), v, factor(NA))
...
+# Error: `false` must be an integer vector, not a list
+# Call `rlang::last_error()` to see a backtrace
+
+hutils::if_else(c(TRUE, FALSE), 1L, list(0L))
+# Error in hutils::if_else(c(TRUE, FALSE), 1L, list(0L)) : 
+#   typeof(false) == list, yet typeof(true) == integer. Both true and false must have the same type.
+
+# The above is useful given that data.table allows list columns as shown below:
+DT <- data.table::data.table(a = 0:5, b = list(1, list(2:3), list(4:6), list(6:4), list(3:2), 1))
+all.equal(DT[ , b := data.table::fifelse(a > 2, b, 0)],DT[ , b := base::ifelse(a > 2, b, 0)])
```
franknarf1: I guess this all.equal test does not work. The code for the b column could give very different results, but we would still see TRUE because b is overwritten in-place. Eg, `DT = data.table(x = 1); all.equal(DT[, b := 1], DT[, b := "yo"])`.

2005m: Thank you. I corrected it. Hope it is better now!

R/test.data.table.R
jangorecki: could we isolate changes related to `test` into new PR?

man/fifelse.Rd
```diff
@@ -64,8 +64,9 @@ hutils::if_else(c(TRUE, FALSE), 1L, list(0L))
 #   typeof(false) == list, yet typeof(true) == integer. Both true and false must have the same type.
 
 # The above is useful given that data.table allows list columns as shown below:
-DT <- data.table::data.table(a = 0:5, b = list(1, list(2:3), list(4:6), list(6:4), list(3:2), 1))
-all.equal(DT[ , b := data.table::fifelse(a > 2, b, 0)],DT[ , b := base::ifelse(a > 2, b, 0)])
+DT1 <- data.table::data.table(a = 0:5, b = list(1, list(2:3), list(4:6), list(6:4), list(3:2), 1))
+DT2 <- data.table::data.table(a = 0:5, b = list(1, list(2:3), list(4:6), list(6:4), list(3:2), 1))
+all.equal(DT1[ , b := data.table::fifelse(a > 2, b, 0)],DT2[ , b := base::ifelse(a > 2, b, 0)])
```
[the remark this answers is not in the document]
2005m: So one does not have to load data.table to run the example.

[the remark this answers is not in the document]
jangorecki: It is not a practical, we can safely assume that user did load pkg when playing with its examples

NEWS.md
```diff
@@ -131,6 +131,23 @@
 
 19. New function `fifelse(test,yes, no)`, also written in C, has been implemented by Morgan Jacob following feature request [#3657](https://github.com/Rdatatable/data.table/issues/3657). It is comparable to `base::ifelse`, `dplyr::if_else}` and  `hutils::if_else`. It returns a vector of the same length as logical vector `test` but unlike \code{base::ifelse} the output type is consistent with those of \code{yes} and \code{no}. Please see `?data.table::fifelse` for more details.
 
+    ```R
```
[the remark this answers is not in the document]
2005m: Sorry not sure to understand that comment.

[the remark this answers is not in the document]
jangorecki: The input provided in this benchmark is always increasing sequence, first half will be TRUE, then second half FALSE. Not sure if it matters here but in general It is possible for a cpu to guess what will be further `if` branches inside the loop and optimise this. You can read more about that https://en.wikipedia.org/wiki/Branch_predictor. Best to include some randomness, as some of algorithms could benefit more from branch prediction than others.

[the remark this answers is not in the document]
2005m: Thank you Jan

2005m:
What do you think of allowing fifelse to coerce the yes and no arguments to the appropriate data type in case they are of different types? For example if we have fifelse(c(TRUE,FALSE), 1L, 2) ,  1L would be coerced to double. That is what ifelse would do. if_else from dplyr and hutils do not support this. I work on a prototype ...

NEWS.md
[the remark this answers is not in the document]
jangorecki: @2005m if you run new code in exactly same environment then you can observe how branch prediction affected the timings. All `mean` and `median` timings significantly larger.

src/fifelse.c
```diff
@@ -0,0 +1,499 @@
...
+  {
+    if(atest)
+    {
+      if(!R_ExternalPtrAddr(alevels))
+      {
+        setAttrib(result, R_LevelsSymbol, blevels);
+      } else {
+        setAttrib(result, R_LevelsSymbol, alevels);
+      }
+      setAttrib(result, R_ClassSymbol, mkString("factor"));
```
jangorecki: `mkString("factor")` -> `char_factor`
this is also defined in init.c

2005m:
I thought I did add set.seed...will double check. Thanks

jangorecki:
@2005m I think error would be good for now in case of different data types.

NEWS.md
```diff
@@ -135,6 +135,26 @@
...
+      base::ifelse(x, 1L, 0L),
+      dplyr::if_else(x, 1L, 0L),
+      times = 100L
+    )
+    # Unit: microseconds
+    #                            expr      min        lq      mean   median        uq       max neval
+    #  data.table::fifelse(x, 1L, 0L)  600.819  649.1415  716.5613  669.026  771.4435  1078.481   100
+    #      hutils::if_else(x, 1L, 0L) 1053.678 1090.6680 1512.2376 1143.053 1346.1765 13827.810   100
+    #         base::ifelse(x, 1L, 0L) 3225.178 3302.3650 4229.8110 3504.419 4230.9615 12712.126   100
+    #       dplyr::if_else(x, 1L, 0L) 3306.855 3390.2430 4496.2153 3557.660 4466.7985 16077.566   100
```
MichaelChirico: FWIW I just ran the timing on `5e7` instead of `5e4`, the timings were roughly the same for all candidates

mattdowle:
The restrict isn't needed but it might help the compiler optimize.  Good point, will add.

I tried to benchmark `vctrs::ifelse` but I couldn't find it in the `vctrs` package. I checked both CRAN version and latest dev. I did `objects(pos=, patt="if")` and I searched GitHub repo for 'ifelse' and 'if_else'.  So I went for prefixing it with "(forthcoming) vctrs::ifelse" in the news item with a link to the nice page by Hadley that he mentioned.  @hadley

MichaelChirico:
oh! from just reading the code I thought restrict was an openMP thing. this Q&A was helpful for 
any future readers

https://stackoverflow.com/q/745870/3576984

2005m:
Maybe a stupid question. I was wondering why is openMP not used for STRSXP and VECSXP ?

MichaelChirico:
my understanding (probably very superficial) is that STRSXP uses R's global string caching to optimize, and there's no architecture in base R for making that cache work with parallel code.

I'm not 100% sure about VECSXP, but they are similar in that both require SET_VECTOR_ELT/SET_STRING_ELT to update and VECTOR_ELT/STRING_ELT to retrieve, so it seems something similar is going on.

not a stupid question at all!

mattdowle:
@2005m STRSXP and VECSXP are what's known as 'behind the write barrier'. You have to write to them using R's SET_ functions which, for example, increment the reference counter of the value being assigned, and decrement the old value. That's not thread safe. You can read from them in parallel, just not write.

MichaelChirico:
From the R internals manual:

https://cran.r-project.org/doc/manuals/r-release/R-ints.html#The-write-barrier

```
1.7 The write barrier and the garbage collector
R has long had a generational garbage collector, and bit gcgen in the sxpinfo header is used in the implementation of this. This is used in conjunction with the mark bit to identify two previous generations.

There are three levels of collections. Level 0 collects only the youngest generation, level 1 collects the two youngest generations and level 2 collects all generations. After 20 level-0 collections the next collection is at level 1, and after 5 level-1 collections at level 2. Further, if a level-n collection fails to provide 20% free space (for each of nodes and the vector heap), the next collection will be at level n+1. (The R-level function gc() performs a level-2 collection.)

A generational collector needs to efficiently ‘age’ the objects, especially list-like objects (including STRSXPs). This is done by ensuring that the elements of a list are regarded as at least as old as the list when they are assigned. This is handled by the functions SET_VECTOR_ELT and SET_STRING_ELT, which is why they are functions and not macros. Ensuring the integrity of such operations is termed the write barrier and is done by making the SEXP opaque and only providing access via functions (which cannot be used as lvalues in assignments in C).

All code in R extensions is by default behind the write barrier. The only way to obtain direct access to the internals of the SEXPRECs is to define ‘USE_RINTERNALS’ before including header file Rinternals.h, which is normally defined in Defn.h. To enable a check on the way that the access is used, R can be compiled with flag --enable-strict-barrier which ensures that header Defn.h does not define ‘USE_RINTERNALS’ and hence that SEXP is opaque in most of R itself. (There are some necessary exceptions: foremost in file memory.c where the accessor functions are defined and also in file size.c which needs access to the sizes of the internal structures.)

For background papers see http://homepage.stat.uiowa.edu/~luke/R/barrier.html and http://homepage.stat.uiowa.edu/~luke/R/gengcnotes.html.
```

2005m:
Thank you @MichaelChirico  and @mattdowle

mattdowle:
@2005m is now a project member - welcome! Great first PR!

man/fifelse.Rd
```diff
@@ -29,12 +29,12 @@ fifelse(dates == "2011-01-01", dates - 1, dates)
...
-# [1] a    b    c    <NA> <NA>
-# Levels: a b c d e
+v = factor(letters[1:3])
+base::ifelse(c(TRUE,FALSE,TRUE), v, factor("a",levels=levels(v)))
+# [1]  1  1  3
+     fifelse(c(TRUE,FALSE,TRUE), v, factor("a",levels=levels(v)))
+# [1] a    a    c
+# Levels: a b c
 
 \dontrun{
```
jangorecki: This section will fail for package check using `--run-dontrun`, we are not using non-suggested packages anywhere in the code, including dontrun sections.
I would stick with putting `hutils`, `dplyr` chunks into NEWS.md, and keeping fifelse examples focused on fifelse rather than its alternatives (+base).

2005m: Maybe we shall remove the `hutils` and `dplyr` call in the coalesce.Rd `dontrun` section as well? If yes, can it be included in your PR?

jangorecki: @2005m yes, I will do new PR for that.

PR #3680, opened 2019-07-05
should keep attributes when coercing dbl to int

### LINKED ISSUE #3679: joining integer Date with double Date results in integer
The current dev version has a bug when joining Date stored in integer with Date stored in double. More specifically speaking, when joining `x` (integer Date) with `i` (double Date), the Date class will disappear results in a naked integer vector.

You may ask why to use an integer Date? `data.table::IDate()` is an example. The package `fst` will store the Date as integer as well. 

I believe the desired behavior is to keep the Date class even if the date is stored in an integer vector. A date vector is what data.table v1.11.4 (at least this version) would return in the this case.

``` r
library(data.table)
dbl_date <- structure(17896.0, class = "Date")
int_date <- structure(17896L, class = "Date")
storage.mode(dbl_date)
#> [1] "double"
storage.mode(int_date)
#> [1] "integer"
x <- data.table(date = int_date, value = 10, key = 'date')
i <- data.table(date = dbl_date, key = 'date')
str(x[i])
#> Classes 'data.table' and 'data.frame':   1 obs. of  2 variables:
#>  $ date : int 17896 ############# see, it's an integer###############
#>  $ value: num 10
#>  - attr(*, "sorted")= chr "date"
str(i[x])
#> Classes 'data.table' and 'data.frame':   1 obs. of  2 variables:
#>  $ date : Date, format: "2018-12-31"  ############# but this is normal Date r###############
#>  $ value: num 10
#>  - attr(*, "sorted")= chr "date"
```

<sup>Created on 2019-07-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

``` r
devtools::session_info()
#>  setting  value                                              
#>  version  R version 3.4.4 (2018-03-15)                       
#>  os       Windows 7 x64 SP 1                                 
#>  system   x86_64, mingw32                                    
#>  ui       RTerm                                              
#>  language (EN)                                               
#>  collate  Chinese (Simplified)_People's Republic of China.936
#>  ctype    Chinese (Simplified)_People's Republic of China.936
#>  tz       Asia/Taipei                                        
#>  date     2019-07-05                                         
#> 
#>  package     * version    date       lib source        
#>  assertthat    0.2.1      2019-03-21 [2] CRAN (R 3.4.4)
#>  backports     1.1.3      2018-12-14 [2] CRAN (R 3.4.4)
#>  callr         3.2.0      2019-03-15 [2] CRAN (R 3.4.4)
#>  cli           1.1.0      2019-03-19 [2] CRAN (R 3.4.4)
#>  crayon        1.3.4      2017-09-16 [2] CRAN (R 3.4.4)
#>  data.table  * 1.12.3     2019-07-05 [1] local         
#>  desc          1.2.0      2018-05-01 [2] CRAN (R 3.4.4)
#>  devtools      2.0.1      2018-10-26 [2] CRAN (R 3.4.4)
#>  digest        0.6.18     2018-10-10 [2] CRAN (R 3.4.4)
#>  evaluate      0.13       2019-02-12 [2] CRAN (R 3.4.4)
#>  fs            1.2.7      2019-03-19 [2] CRAN (R 3.4.4)
#>  glue          1.3.1      2019-03-12 [2] CRAN (R 3.4.4)
#>  highr         0.8        2019-03-20 [2] CRAN (R 3.4.4)
#>  htmltools     0.3.6.9000 2018-05-07 [2] local         
#>  knitr         1.22       2019-03-08 [2] CRAN (R 3.4.4)
#>  magrittr      1.5        2014-11-22 [2] CRAN (R 3.4.4)
#>  memoise       1.1.0      2017-04-21 [2] CRAN (R 3.4.4)
#>  pkgbuild      1.0.3      2019-03-20 [2] CRAN (R 3.4.4)
#>  pkgload       1.0.2      2018-10-29 [2] CRAN (R 3.4.4)
#>  prettyunits   1.0.2      2015-07-13 [2] CRAN (R 3.4.1)
#>  processx      3.3.0      2019-03-10 [2] CRAN (R 3.4.4)
#>  ps            1.3.0      2018-12-21 [2] CRAN (R 3.4.4)
#>  R6            2.4.0      2019-02-14 [2] CRAN (R 3.4.4)
#>  Rcpp          1.0.1      2019-03-17 [2] CRAN (R 3.4.4)
#>  remotes       2.0.2      2018-10-30 [2] CRAN (R 3.4.4)
#>  rlang         0.3.3      2019-03-29 [2] CRAN (R 3.4.4)
#>  rmarkdown     1.12       2019-03-14 [2] CRAN (R 3.4.4)
#>  rprojroot     1.3-2      2018-01-03 [2] CRAN (R 3.4.3)
#>  sessioninfo   1.1.1      2018-11-05 [2] CRAN (R 3.4.4)
#>  stringi       1.4.3      2019-03-12 [2] CRAN (R 3.4.4)
#>  stringr       1.4.0      2019-02-10 [2] CRAN (R 3.4.4)
#>  testthat      2.1.1.9000 2019-06-22 [2] local         
#>  usethis       1.4.0      2018-08-14 [2] CRAN (R 3.4.4)
#>  withr         2.1.2      2018-03-15 [2] CRAN (R 3.4.4)
#>  xfun          0.6        2019-04-02 [2] CRAN (R 3.4.4)
#>  yaml          2.2.0      2018-07-25 [2] CRAN (R 3.4.4)
#> 
#> [1] D:/R/R-dev
#> [2] D:/app/R_lib/3.1
#> [3] D:/app/R/library
```

</details>

shrektan:
Closes #3679

### DIFF
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -90,6 +90,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
           val = as.integer(i[[ic]])
+          attributes(val) <- attributes(i[[ic]])
           set(i, j=ic, value=val)
           set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
         }
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -90,7 +90,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
           val = as.integer(i[[ic]])
-          attributes(val) <- attributes(i[[ic]])
+          attributes(val) = attributes(i[[ic]])
           set(i, j=ic, value=val)
           set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
         }
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -89,8 +89,9 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # common case of ad hoc user-typed integers missing L postfix joining to correct integer keys
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
-          val = as.integer(i[[ic]])
-          attributes(val) = attributes(i[[ic]])
+          val = i[[ic]]
+          # according to ?as.integer, it's faster than `as.integer()` and will preserve the attributes
+          storage.mode(val) = "integer"
           set(i, j=ic, value=val)
           set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
         }
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -90,7 +90,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
           val = as.integer(i[[ic]])
-          if (!is.null(attributes(i[[ic]]))) attributes(val) = attributes(i[[ic]])
+          if (!is.null(attributes(i[[ic]]))) attributes(val) = attributes(i[[ic]])  # to retain Date for example; 3679
           set(i, j=ic, value=val)
           set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
         }
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -89,9 +89,8 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # common case of ad hoc user-typed integers missing L postfix joining to correct integer keys
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
-          val = i[[ic]]
-          # according to ?as.integer, it's faster than `as.integer()` and will preserve the attributes
-          storage.mode(val) = "integer"
+          val = as.integer(i[[ic]])
+          if (!is.null(attributes(i[[ic]]))) attributes(val) = attributes(i[[ic]])
           set(i, j=ic, value=val)
           set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
         }
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15194,6 +15194,14 @@ l = replicate(ceiling(log10(.Machine$integer.max)), rep(1L, 10L), simplify = FAL
 l$unique = TRUE
 test(2061, do.call(CJ, l), setkey(unique(as.data.table(head(l, -1L)))))
 
+# #3679 -- integer / double Date merge should retain attributes
+dbl_date <- structure(17896.0, class = "Date")
+int_date <- structure(17896L, class = "Date")
+x <- data.table(date = int_date, value = 10, key = 'date')
+i <- data.table(date = dbl_date, key = 'date')
+test(2062.1, x[i, class(date)], 'Date')
+test(2062.2, i[x, class(date)], 'Date')
+
 ###################################
 #  Add new tests above this line  #
 ###################################
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15253,8 +15253,10 @@ dbl_date = structure(17896.0, class = "Date")
 int_date = structure(17896L, class = "Date")
 x = data.table(date = int_date, value = 10, key = 'date')
 i = data.table(date = dbl_date, key = 'date')
-test(2064.1, x[i, class(date)], 'Date')
-test(2064.2, i[x, class(date)], 'Date')
+test(2064.1, x[i, class(date), verbose=TRUE], 'Date',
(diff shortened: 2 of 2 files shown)


R/bmerge.R
```diff
@@ -90,6 +90,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
           val = as.integer(i[[ic]])
+          attributes(val) = attributes(i[[ic]])
```
MichaelChirico: is this something that should/could be left to `copyMostAttrib` at the C level?

At the R level, I think we usually have something like `if (!is.null(attributes(i[[ic]]))) attributes...`

shrektan: I think it should be very very trivial compared to the above use of as.integer().

shrektan: But if other codes add the if in front, I will adapt the convention. :D

MichaelChirico: yep... only thought of it since i passed it recently in `CJ`

shrektan:
@MichaelChirico I believe use ` storage.mode<-` is even better. According to `?as.integer`, 

> Like as.vector it strips attributes including names. (To ensure that an object x is of integer type without stripping attributes, use storage.mode(x) <- "integer".)


## benchmark

``` r

x1 <- as.double(1:100)
x2 <- structure(as.double(17000:17100), class = 'Date')

f1 <- function(x) {
  out <- as.integer(x)
  attributes(out) <- attributes(x)
  out
}

f2 <- function(x) {
  out <- as.integer(x)
  if (!is.null(attributes(x))) attributes(out) <- attributes(x)
  out
}

f3 <- function(x) {
  storage.mode(x) <- 'integer'
  x
}

bench::mark(
  f1(x1),
  f2(x1),
  f3(x1)
)
#> # A tibble: 3 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 f1(x1)       1.24µs   1.46µs   619709.      448B     124.
#> 2 f2(x1)       1.21µs   1.31µs   509117.      448B       0 
#> 3 f3(x1)       1.05µs   1.26µs   426692.    1.27KB       0

bench::mark(
  f1(x2),
  f2(x2),
  f3(x2)
)
#> # A tibble: 3 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 f1(x2)       2.15µs   3.19µs   264622.      912B      0  
#> 2 f2(x2)       2.33µs   3.11µs   295207.      912B     29.5
#> 3 f3(x2)       1.05µs   1.47µs   547739.    1.28KB      0
```

<sup>Created on 2019-07-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/bmerge.R
```diff
@@ -89,8 +89,9 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # common case of ad hoc user-typed integers missing L postfix joining to correct integer keys
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
-          val = as.integer(i[[ic]])
-          attributes(val) = attributes(i[[ic]])
+          val = i[[ic]]
```
MichaelChirico: Does this affect `i` in user's environment? I think it shouldn't but check to be sure?

R/bmerge.R
```diff
@@ -89,7 +89,9 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           # common case of ad hoc user-typed integers missing L postfix joining to correct integer keys
           # we've always coerced to int and returned int, for convenience.
           if (verbose) cat("Coercing double column i.",names(i)[ic]," (which contains no fractions) to type integer to match type of x.",names(x)[xc],".\n",sep="")
-          val = as.integer(i[[ic]])
+          val = i[[ic]]
+          # according to ?as.integer, it's faster than `as.integer()` and will preserve the attributes
+          storage.mode(val) = "integer"
```
MichaelChirico: I think it's faster in the case `val` has attributes but slower when not -- same benchmark with much larger vector:

```
N = 1e7
x1 = as.double(sample(1:2e4, N, TRUE))
x2 <- .Date(x1)

f1 <- function(x) {
  out <- as.integer(x)
  attributes(out) <- attributes(x)
  out
}

f2 <- function(x) {
  out <- as.integer(x)
  if (!is.null(attributes(x))) attributes(out) <- attributes(x)
  out
}

f3 <- function(x) {
  storage.mode(x) <- 'integer'
  x
}

bench::mark(
  f1(x1),
  f2(x1),
  f3(x1)
)

bench::mark(
  f1(x2),
  f2(x2),
  f3(x2)
)
```

```
# A tibble: 3 x 13
  expression    min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr> <bch:> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 f1(x1)       16ms 18.2ms      48.4    38.1MB     4.84    20     2      413ms
2 f2(x1)     16.4ms 17.6ms      56.1    38.1MB    11.8     19     4      339ms
3 f3(x1)     32.5ms 34.5ms      28.6    38.1MB     4.77    12     2      419ms
# … with 4 more variables: result <list>, memory <list>, time <list>,
#   gc <list>
```

```
# A tibble: 3 x 13
  expression    min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr> <bch:> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 f1(x2)       32ms 32.5ms      30.6    38.1MB     5.09    12     2      393ms
2 f2(x2)     31.6ms 32.4ms      30.9    38.1MB     8.42    11     3      356ms
3 f3(x2)     33.5ms   35ms      28.5    38.1MB     2.20    13     1      455ms
# … with 4 more variables: result <list>, memory <list>, time <list>,
#   gc <list>
```

R/bmerge.R
jangorecki: Best to unit test to be sure

R/bmerge.R
shrektan: It's interesting that if I run the code interactively, `f3()` is always the fastest. This is what I expected because `storage.mode<-` calls a primitive R function. Moreover, since the only difference is about the attributes, it shouldn't have such huge (2x) impact on the performance....

But if I use reprex to run, it's always the slowest. Here's what I get if run interactively.

image

shrektan: I believe you are right. After some experiments, I realize that use your version (check attributes if NULL and then copy the attributes) is the best choice. `storage.mode<-` is much slower if the vector is lengthy (I have no idea why it's the case). Moreover, use `storage.mode<-` is kind of confusing.

PR #3688, opened 2019-07-10
setkey works on tables with complex columns

MichaelChirico:
Closes #1444 
Closes #3704

Issue #1444, opened 2015-11-26
Support for grouping/ordering/computing on columns of type complex


atalikami:
Hello,

The only question or information I can find on this is one I asked myself on stack overflow (http://stackoverflow.com/questions/32570742/r-data-table-package-and-complex-values) but it wasn't ever resolved.

The data.table package does not appear to support complex numbers, even though there was a specific bug fix addressing the problem, issue #982. I have updated to the latest development version, 1.9.7.

Now, I can have a data table which contains complex numbers, but nothing can then be done to it. For example:

```
stacktest<-data.table(x=complex(real=1:60, imaginary=-1:-60),y=LETTERS[1:10],z=rep(1:2,5))
```

This works, the data.table exists. However, if I try and use the setkey command:

```
setkey(stacktest, z)

Error in setkeyv(x, cols, verbose = verbose, physical = physical) : 
  Item 1 of list is type 'complex' which isn't yet supported

```

Or, if I try and run a function (any function as far as I can tell, I first discovered this trying to use dcast) on a subset of the data:

```
stacktest2<-stacktest[,.(abs(.SD)),by=.(y)]

Error in `[.data.table`(stacktest, , .(abs(.SD)), by = .(y)) : 
  Type 15 in .SD column 0
```

Being as issue #982 was specific to allowing complex numbers to work, I feel as though this is a bug, not a feature request. 

Thanks.

MichaelChirico:
Merge this with #602?

arunsrinivasan:
This issue is more detailed.. will close the other one.

MichaelChirico:
Came across this again in working on #3639. Strange that you can't order any table with complex values, considering `forder` works just fine:

```
# works fine
stacktest_keyed = copy(stacktest[order(z)])
setkey(stacktest_keyed, z)

# is essentially the same thing? but fails
setkey(stacktest, z)
```

Issue #982, opened 2014-12-10
[.data.table fails with complex/raw columns


richierocks:
To reproduce, try:

```
DT <- data.table(x = as.complex(1:5))
DT[1]
## Error in `[.data.table`(DT, 1) : Unknown column type 'complex'
```

Similarly:

```
DT <- data.table(x = as.raw(1:5))
DT[1]
## Error in `[.data.table`(DT, 1) : Unknown column type 'raw'
```

The problem appears to be originating from the line:

```
ans[[target]] = .Call(CsubsetVector,x[[source]],irows)
```

Loosely related issue: https://github.com/Rdatatable/data.table/issues/602

Issue #602, opened 2014-06-08
[R-Forge #2422] Handle columns with complex numbers


arunsrinivasan:
Submitted by: Paul Metzner; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2422&group_id=240&atid=978)

Necessary to use data.table for signal processing, especially for the analysis of power spectra.

MichaelChirico:
I suppose this issue is open because `complex` columns are not feature-complete?

E.g, 

```
set.seed(102439)
DT <- data.table(C = c(1+2i, 1+2i, 1-2i, 1-2i), N = runif(4))
DT[ , .N, by = C]
```

> Error in`` `[.data.table`(DT, , .N, by = C)``: 
  column or expression 1 of `by` or `keyby` is type `complex`. Do not quote column names. Usage: `DT[,sum(colC),by=list(colA,month(colB))]`

jangorecki:
The last one `DT[ , base::sum(C), by = I]` looks like a bug. This `sapply(split(DT$C, DT$I), sum)` works.

arunsrinivasan:
dup of #1444.

PR #3689, opened 2019-07-10
Closes #3639 -- proper handling of CPLXSXP in dogroups

### LINKED ISSUE #3639: Wrong values in j with complex values if "by" has more than 1 group.
I understand that complex number support is not fully implemented (see [this](https://github.com/Rdatatable/data.table/issues/1444) issue), but this is weird and, very serious, IMHO, because it can be a big source of silent errors.  

``` r
library(data.table)
set.seed(42)
data <- setDT(expand.grid(x = 1:10,
                    a = c("a", "b"),
                    b = 1:2))
data[, z := rnorm(1:.N) + rnorm(1:.N)*1i]


data2 <- data[, .(x = x, z = z), by = .(a, b)]
```

Now both data.tables should have the same data, but…

``` r
rbind(data[a == "b" & b == 2 & x == 1],
      data2[a == "b" & b == 2 & x == 1], use.names = TRUE)
#>    x a b                    z
#> 1: 1 b 2 0.4554501-1.0431189i
#> 2: 1 b 2 0.6359504+0.2765507i
```

The problems are not evently distributed:

``` r
setnames(data2, "z", "z2")
merged <- data2[data, on = c("x", "a", "b")]

library(ggplot2)
#> Registered S3 methods overwritten by 'ggplot2':
#>   method         from 
#>   [.quosures     rlang
#>   c.quosures     rlang
#>   print.quosures rlang
ggplot(merged, aes(x, abs(z - z2))) +
    geom_line() +
    facet_grid(a ~ b)
```

And it doesn’t happen if “by” has only one group

``` r
data2 <- data[, .(x = x, z = z), by = .(interaction(a, b, sep = "."))]


data[a == "b" & b == 2 & x == 1]$z == data2[interaction == "b.2" & x == 1]$z
#> [1] TRUE
```

<sup>Created on 2019-06-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

MichaelChirico:
#3639

Haven't quite figured out the ins and outs of `CPLXSXP` and `COMPLEX()` usage yet but I think this gets at the kernel of the issue -- `dogroups` assumes `SIZEOF` is either 4 or 8, not so for `CPLXSXP` (16). Will come back to this.

FWIW `grep -Er "==\s?4" src` only turns up `reorder` and `dogroups`, so I think this and #3688 should cover this class of issue.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15200,6 +15200,15 @@ d2 = data.table(r = 1:5, s = seq(0L, 20L, 5L))
 test(2062.1, d1[d2, on = .(a <= s, b >= s), j = .SD],   ans<-data.table(a=INT(0,5,10,10,15,20), b=INT(0,5,10,10,15,20)))
 test(2062.2, d1[d2, on = .(a <= s, b >= s)][, .(a, b)], ans)
 
+# #3639 -- complex values in grouping
+set.seed(42)
+DT = CJ(x = 1:10, a = c("a", "b"), b = 1:2)
+DT[ , z := complex(rnorm(1:.N), rnorm(1:.N))]
+## can simplify this test after #1444
+test(2063.1, all.equal(setkey(copy(DT), NULL), DT[, .(x = x, z = z), by = .(a, b)][order(x, a, b)], ignore.col.order = TRUE))
+test(2063.2, DT[ , base::sum(z), by = a], data.table(a = c('a', 'b'), V1 = c(5.0582228485073+0i, -1.8644229822705+0i)))
+test(2063.3, DT[ , sum(Mod(z)), by = b], data.table(b = 1:2, V2 = c(16.031422657932, 13.533483145656)))
+
 
 ###################################
 #  Add new tests above this line  #
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15207,7 +15207,7 @@ DT[ , z := complex(rnorm(1:.N), rnorm(1:.N))]
 ## can simplify this test after #1444
 test(2063.1, all.equal(setkey(copy(DT), NULL), DT[, .(x = x, z = z), by = .(a, b)][order(x, a, b)], ignore.col.order = TRUE))
 test(2063.2, DT[ , base::sum(z), by = a], data.table(a = c('a', 'b'), V1 = c(5.0582228485073+0i, -1.8644229822705+0i)))
-test(2063.3, DT[ , sum(Mod(z)), by = b], data.table(b = 1:2, V2 = c(16.031422657932, 13.533483145656)))
+test(2063.3, DT[ , sum(Mod(z)), by = b], data.table(b = 1:2, V1 = c(16.031422657932, 13.533483145656)))
 
 
 ###################################
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15211,8 +15211,8 @@ test(2063.3, DT[ , sum(Mod(z)), by = b], data.table(b = 1:2, V1 = c(16.031422657
 # also works for assignment, as noted in #3690
 DT[ , z_sum := base::sum(z), by = .(a, b)]
 test(2063.4, DT[ , z_sum := base::sum(z), by = .(a, b)][1:3, z_sum],
-     c(3.05896822455656+0i, -2.81616342436407+0i, -3.04864360704606+0i))
-test(2063.4, DT[1L, z_sum := 1i][1L, z_sum], 1i)
+     c(1.8791864549242+0i, 3.17903639358309+0i, -4.18868631527035+0i))
+test(2063.5, DT[1L, z_sum := 1i][1L, z_sum], 1i)
 
 ###################################
 #  Add new tests above this line  #
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -7335,7 +7335,8 @@ DT3 = rbindlist(list(DT2, list(x="d", y=7L, mul=100L)))
 DT3 = DT3[sample(nrow(DT3))]
 DT1[ , x_num := match(x, letters) + .1]
 DT3[ , x_num := match(x, letters) + .1]
-test(1540.36, DT1[DT3, .(x_num), by=.EACHI, on=c(x_num="x_num")])
+test(1540.36, DT1[DT3[1:3], .(y = x_num), by=.EACHI, on=c(x_num="x_num")],
+     data.table(x_num = c(3.1, 4.1, 3.1), y = c(2.1, NA, NA)))
 
 # to do: add tests for :=
 
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15208,7 +15208,11 @@ DT[ , z := complex(rnorm(1:.N), rnorm(1:.N))]
 test(2063.1, all.equal(setkey(copy(DT), NULL), DT[, .(x = x, z = z), by = .(a, b)][order(x, a, b)], ignore.col.order = TRUE))
 test(2063.2, DT[ , base::sum(z), by = a], data.table(a = c('a', 'b'), V1 = c(5.0582228485073+0i, -1.8644229822705+0i)))
 test(2063.3, DT[ , sum(Mod(z)), by = b], data.table(b = 1:2, V1 = c(16.031422657932, 13.533483145656)))
-
+# also works for assignment, as noted in #3690
+DT[ , z_sum := base::sum(z), by = .(a, b)]
+test(2063.4, DT[ , z_sum := base::sum(z), by = .(a, b)][1:3, z_sum],
+     c(3.05896822455656+0i, -2.81616342436407+0i, -3.04864360704606+0i))
+test(2063.4, DT[1L, z_sum := 1i][1L, z_sum], 1i)
 
 ###################################
 #  Add new tests above this line  #
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15208,11 +15208,20 @@ DT[ , z := complex(rnorm(1:.N), rnorm(1:.N))]
 test(2063.1, all.equal(setkey(copy(DT), NULL), DT[, .(x = x, z = z), by = .(a, b)][order(x, a, b)], ignore.col.order = TRUE))
 test(2063.2, DT[ , base::sum(z), by = a], data.table(a = c('a', 'b'), V1 = c(5.0582228485073+0i, -1.8644229822705+0i)))
 test(2063.3, DT[ , sum(Mod(z)), by = b], data.table(b = 1:2, V1 = c(16.031422657932, 13.533483145656)))
+## mimicking test 171.3 for coverage
+DT = data.table(A=c(25L,85L,25L,25L,85L), B=c("a","a","b","c","c"), z=0:4 + (4:0)*1i)
+test(2063.4, DT[ , data.table(A, z)[A==25, z] + data.table(A, z)[A==85, z], by=B],
+     data.table(B = c('a', 'c'), V1 = c(1, 7) + (c(7, 1))*1i))
+## mimicking test 771 for coverage
+a = data.table(first=1:6, third=c(1i,1,1i,3,3i,4), key="first")
+b = data.table(first=c(3,4,4,5,6,7,8), second=1:7, key="first")
+test(2063.5, b[ , third:=a[b, third, by=.EACHI]], error="Supplied 2 items to be assigned to 7 items of")
+
 # also works for assignment, as noted in #3690
 DT[ , z_sum := base::sum(z), by = .(a, b)]
-test(2063.4, DT[ , z_sum := base::sum(z), by = .(a, b)][1:3, z_sum],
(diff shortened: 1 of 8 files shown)


src/dogroups.c
```diff
@@ -382,14 +392,20 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
-      } else {
+      } else if (tsize==8) {
         double *td = REAL(target);
         double *sd = REAL(source);
         for (int r=0; r<maxn; ++r) td[ansloc+r] = sd[igrp];
+      } else {
+        // #3634 -- CPLXSXP columns have size 16
+        double complex *td = (double complex *)COMPLEX(target);
+        double complex *sd = (double complex *)COMPLEX(source);
+        for (int r=0; r<maxn; ++r) td[ansloc+r] = sd[igrp];
```
MichaelChirico: TBH I don't really know what this branch is doing so I don't know how to put in a test for it.

src/dogroups.c
```diff
@@ -143,6 +145,11 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
         case REALSXP :
           REAL(VECTOR_ELT(SDall,j))[0] = NA_REAL;
           break;
+        case CPLXSXP :
```
MichaelChirico: I don't know what this `switch` is doing, but here's how to initialize `CPLXSXP` -- set the `.r` and `.i` components to `NA_REAL` (taken from `r-source/src/main/complex.c`)

src/dogroups.c
```diff
@@ -382,14 +401,20 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
         double *sd = REAL(source);
         for (int r=0; r<maxn; ++r) td[ansloc+r] = sd[igrp];
+      } else {
+        // #3634 -- CPLXSXP columns have size 16
+        double complex *td = (double complex *)COMPLEX(target);
+        double complex *sd = (double complex *)COMPLEX(source);
+        for (int r=0; r<maxn; ++r) td[ansloc+r] = sd[igrp];
       }
       // Shouldn't need SET_* to age objects here since groups, TO DO revisit.
     }
```
MichaelChirico: There's one more `switch (TYPEOF(target))` just below here (also I don't know what it's doing), I couldn't figure out how to get that one to work though. Attempts:

```
case CPLXSXP : {
  double complex *td = (double complex *)(COMPLEX(target) + thisansloc);
  // attempt 1; fails because double complex doesn't have .r/.i, only CPLXSXP does
  for (int r=0; r<maxn; ++r) { td[r].r = NA_REAL; td[r].i = NA_REAL; }
  // attempt 2; fails because i'm not allowed to assign to RHS of creal()/cimag()
  for (int r=0; r<,maxn; ++r) { creal(td[r]) = NA_REAL; cimag(td[r]) = NA_REAL; }
} break;
```

MichaelChirico:
The code is working, just need to improve coverage (any ideas?)

src/dogroups.c
```diff
@@ -415,6 +446,11 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
           double *td = REAL(target)+thisansloc;
           for (int r=0; r<maxn; ++r) td[r] = NA_REAL;
         } break;
+        case CPLXSXP : {
+        double complex *td = (double complex *)(COMPLEX(target) + thisansloc);
+        for (int r=0; r<maxn; ++r) { td[r] = NA_REAL + NA_REAL*I; }
```
MichaelChirico: again, no idea what this branch is really doing so I haven't confirmed it works, but at least it compiles

src/dogroups.c
```diff
@@ -382,14 +401,20 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
+      if (tsize==4) {
         int *td = INTEGER(target);
         int *sd = INTEGER(source);
         for (int r=0; r<maxn; ++r) td[ansloc+r] = sd[igrp];
-      } else {
+      } else if (tsize==8) {
         double *td = REAL(target);
         double *sd = REAL(source);
         for (int r=0; r<maxn; ++r) td[ansloc+r] = sd[igrp];
+      } else {
```
MichaelChirico: this one is in the `ngrpcols` loop, so covering it requires being able to group `by` complex, which we can't (yet)

src/dogroups.c
```diff
@@ -167,6 +174,10 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
         case REALSXP :
           REAL(VECTOR_ELT(xSD,j))[0] = NA_REAL;
           break;
+        case CPLXSXP :
+          COMPLEX(VECTOR_ELT(xSD, j))[0].r = NA_REAL;
+          COMPLEX(VECTOR_ELT(xSD, j))[0].i = NA_REAL;
+          break;
```
MichaelChirico: IIUC this is coming from a join, so we won't be able to cover it yet.

Minimal example to trigger the `INTSXP` branch of this `switch` (adapted test 1283.3):

```
dt1 <- data.table(a=rep(1:2, each=2), c=c(7L, 2L, 4L, 8L))
dt2 <- data.table(b=rep(2:3), c=c(2L, 6L), d=c(20L, 8L))
dt1[dt2, on = c(a = 'b'), .(a=a), by=.EACHI]
```

I believe it's because `j` uses a column from the LHS of `on`.

But joins aren't available for complex yet...

MichaelChirico:
OK. The remaining codecov problems I think are outside the scope of this PR (see comments). does it make sense to just delete that part of the patch and save it for when it's actually testable?

Will also file a coverage commit for `dogroups`

inst/tests/tests.Rraw
```diff
@@ -7326,6 +7326,16 @@ test(1540.35, DT1[DT3, lapply(.SD, function(x) x * mul),
...
+## more coverage tests for by = .EACHI, on = c(LHS = 'RHS') for numeric type
+set.seed(45L)
+DT1 = data.table(x=sample(letters[1:3], 15, TRUE), y=sample(6:10, 15, TRUE),
+                 a=sample(100, 15), b=runif(15))
+DT2 = CJ(x=letters[1:3], y=6:10)[, mul := sample(20, 15)][sample(15L, 5L)]
+DT3 = rbindlist(list(DT2, list(x="d", y=7L, mul=100L)))
+DT3 = DT3[sample(nrow(DT3))]
+DT1[ , x_num := match(x, letters) + .1]
+DT3[ , x_num := match(x, letters) + .1]
+test(1540.36, DT1[DT3, .(x_num), by=.EACHI, on=c(x_num="x_num")])
```
MichaelChirico: strangely enough, I couldn't get this test to hit the `LGLSXP` branch:

```
DT1[ , x_lgl := x > = 'a']
DT3[ , x_lgl := x > 'a']
DT1[DT3, .(x_lgl), by=.EACHI, on=c(x_lgl="x_lgl")]
```

doesn't do it...

src/dogroups.c
```diff
@@ -167,20 +174,24 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
         case STRSXP :
           SET_STRING_ELT(VECTOR_ELT(xSD,j),0,NA_STRING);
           break;
         case VECSXP :
           SET_VECTOR_ELT(VECTOR_ELT(xSD,j),0,R_NilValue);
           break;
```
MichaelChirico: Should we nocov this branch (or set it to error even?)

IIUC hitting this branch would require first joining on `VECSXP` column in `on`, I don't think we have any plan to support list-to-list joins?

src/assign.c
```diff
@@ -1002,6 +1011,11 @@ void writeNA(SEXP v, const int from, const int n)
       for (int i=from; i<=to; ++i) vd[i] = NA_REAL;
     }
   } break;
+  case CPLXSXP: {
+    Rcomplex *vd = COMPLEX(v);
+    Rcomplex NA_CPLX = {NA_REAL, NA_REAL};
```
MichaelChirico: a bit strange why we have to define this ourselves but I see it done like this in a few places in `r-source` (e.g. `subset.c`)

mattdowle:
Have we heard from many users requesting complex support?
I'll try and get through these PRs asap, and hopefully they are quick, but it's adding complexity (pun intended) when another approach could be just to not support complex yet until other higher value things are done first.
If the idea is to pick off something where you can learn the code base, that's a good idea, and I'm ok to merge them in on that basis.  Just hoping these 6 PRs are quick and they don't cause unintended effects.

Some links to users and fields requesting complex support please, would keep me happy that it's worth doing.

src/dogroups.c
mattdowle: don't no cov it.  I'd say don't error either because something somewhere might be using it.  I'd say just leave it as-is for now until it can be covered at a later date.

src/assign.c
mattdowle: Yes I see 3 in src/main/subset.c too.  Those are the only occurence of "NA_C" in the entire source tree of R.  So since it's so rarely used, I guess it just hasn't been looked at.  Which seems reasonable.

eliocamp:
I'm one happy data.table user that uses complex numbers frequently and would be even happier if complex support is enhanced :smile:. Perusing github I see there are some issues addressing it:

* https://github.com/Rdatatable/data.table/issues/33

* https://github.com/Rdatatable/data.table/issues/256

* https://github.com/Rdatatable/data.table/issues/602

 * https://github.com/Rdatatable/data.table/issues/982

* https://github.com/Rdatatable/data.table/issues/1444

* https://github.com/Rdatatable/data.table/issues/3639

Not a lot (understandably), but non-zero.

mattdowle:
@eliocamp Great, thanks. I looked through them. It looks to me as if it's programmers checking that all types are supported for completeness reasons, rather than users in the real-world actually needing complex number support.   I did find one field mentioned in those issues : 
* signal processing, especially for the analysis of power spectra.

What's your field that you use complex numbers in?

eliocamp:
Meteorology, but I use some signal processing methods to analyse fields. Complex Principal Component Analysis, for example.

mattdowle:
Good to know! Are there any articles that show how meteorology, signal processing and/or complex-PCA need complex numbers in the context of data.table-like operations like grouping? I did complex number theory in undergraduate applied math and much enjoyed it, but I've never used complex numbers since then.

mattdowle:
All looks good. I realize the diff coverage is under target but that's ok, because as Michael wrote, other PRs will cover those in future.

MichaelChirico:
> pick off something where you can learn the code base

in fact that's the main thinking. I imagine there are some use cases around Fourier series running around but yea, I haven't come across it myself in a while either

src/subset.c
```diff
@@ -77,7 +77,6 @@ static void subsetVectorRaw(SEXP ans, SEXP source, SEXP idx, const bool anyNA)
   case CPLXSXP : {
     Rcomplex *sp = COMPLEX(source);
     Rcomplex *ap = COMPLEX(ans);
-    Rcomplex NA_CPLX = { NA_REAL, NA_REAL };
```
MichaelChirico: looking at it again, would just inlining `= (Rcomplex){NA_REAL, NA_REAL}` do the trick?

mattdowle: I don't think `(Rcomplex){NA_REAL, NA_REAL}` will work. Because `Rcomplex` is a `struct`. The {NA_REAL, NA_REAL} is special syntax just for initialization of a variable/struct.

eliocamp:
> Good to know! Are there any articles that show how meteorology, signal processing and/or complex-PCA need complex numbers in the context of data.table-like operations like grouping? I did complex number theory in undergraduate applied math and much enjoyed it, but I've never used complex numbers since then.

I've never read anything specific about data.table. Here's a paper about the method: https://journals.ametsoc.org/doi/abs/10.1175/1520-0450%281984%29023%3C1660%3ACPCATA%3E2.0.CO%3B2

This is the code I'm using now. It gets the Hilbert transform of the data and then does an EOF (Empirical Orthogonal Functions --what we meteorologists call PCA) for each season, dataset and height level. Because of #3639 I had to use the workaround of separating the real an imaginary parts and then joining them together after the grouping was done. 

```r
ceof <-  data %>%
  .[, c("R", "I") := ReIm(spectral::analyticFunction(hgt.minus)),
    by = .(lat, season, year, dataset, lev)] %>% 
  .[, {
    d <- copy(.SD)[, hgt := R + I*1i]
    list(ceof = list(
      EOF(hgt ~ year | lon + lat, n = 1:2, 
          B = 5000, suffix = "cPC", data = d)
    ))
  }, by = .(dataset, season, lev)]
```
`EOF()` (a function I created and available in the metR package) returns a list of data.tables that are the matrices of the Singular Value Decomposition in "long" format. So `ceof` is a data.table with a column names "ceof" (yeah, not very smart naming convention) that that holds said list of data.tables, one for each combination of dataset, season and height level.

mattdowle:
Thanks @eliocamp, helps a lot!  Continuing here: https://github.com/Rdatatable/data.table/pull/3701#issuecomment-512605868

Issue #3690, opened 2019-07-10
Master list: Feature completion for complex-valued columns


MichaelChirico:
As a follow-up to #1444 and #3689, I think it's useful to keep a list of all the features missing for users who want to do work involving complex-valued columns `z`:

 - [x] 1. Sorting by `z` (#1703); by extension `frank(z)`, `rowid(z)`, `dcast`, `setkey`, `uniqueN`
 - [x] 2. Grouping `by` z
 - [x] 3. `z` I/O (`fread` (without `colClasses`) & `fwrite`; overlap with #2247)
 - [x] 4. `CJ(z)`
 - [x] 5. Assign `z` `by` group
 - [x] 6. `GForce` on `z`
 - [ ] 7. ~Joins on `z`~    wait until a user requests it with justification (that part of bmerge.c is complex) 
 - [x] 8. `coalesce(z)`
 - [x] 9. `shift(z)`
 - [x] 10. `unique(z)`
- [x] 11. `:=` #3639

v1.12.4 NEWS contains : 

![image](https://user-images.githubusercontent.com/2779998/61567164-771db380-aa33-11e9-90bf-5c6064544c6f.png)

---
# Examples

```
DT = data.table(a = 1:5, z = 6:10 + (5:1)*1i)
```

### 1. 

```
DT[order(z)]
```

> Error in forderv(ans, cols, sort = TRUE, retGrp = FALSE, order = if (decreasing) -order else order,  : 
  Column 1 of by= (1) is type 'complex', not yet supported. Please use the by= argument to specify columns with types that are supported. See NEWS item in v1.12.2 for more information.

### 2. 

```
DT[ , .N, by = z]
```

> Error in `[.data.table`(DT, , .N, by = z) : 
  column or expression 1 of 'by' or 'keyby' is type complex. Do not quote column names. Usage: DT[,sum(colC),by=list(colA,month(colB))]

### 3. 

```
fwrite(DT)
```

> Error in fwrite(DT) : 
  Column 2's type is 'complex' - not yet implemented in fwrite.

```
fread('a\n1+i')
#         a
#    <char>
# 1:    1+i
```

### 4. 

```
DT[CJ(a, z)]
```

> Error: Column 1 of by= (1) is type 'complex', not yet supported. Please use the by= argument to specify columns with types that are supported. See NEWS item in v1.12.2 for more information.

Coming from `forderv` so first this is a repeat of 1

```
DT[CJ(a, z, sorted = FALSE)]
```

[...]

PR #3692, opened 2019-07-10
gsum, gmean, gfirst, glast, g[ for complex vectors

MichaelChirico:
part of #3690

Haven't added any tests yet but the basic examples work. Will return to this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -94,7 +94,7 @@
 
 16. `as.data.table` now unpacks columns in a `data.frame` which are themselves a `data.frame`. This need arises when parsing JSON, a corollary in [#3369](https://github.com/Rdatatable/data.table/issues/3369#issuecomment-462662752). `data.table` does not allow columns to be objects which themselves have columns (such as `matrix` and `data.frame`), unlike `data.frame` which does. Bug fix 19 in v1.12.2 (see below) added a helpful error (rather than segfault) to detect such invalid `data.table`, and promised that `as.data.table()` would unpack these columns in the next release (i.e. this release) so that the invalid `data.table` is not created in the first place.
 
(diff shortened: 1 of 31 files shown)


MichaelChirico:
Hmm I'm segfaulting `gsum`, will have to take a closer look

mattdowle:
For future reference, "torture by Rstrict yielded" means using R-devel built with UBSAN, ASAN and strict-barrier (see section in CRAN_Release.cmd) and this dev cycle : 
```
R CMD build .
Rdevel-strict CMD INSTALL data.table_1.12.3.tar.gz  # with -O0 -g
Rdevel-strict
require(data.table)
DT = data.table(A=1:5, B=-3i, C=2147483647L)
gctorture2(step=100)
DT[, sum(B), by=A%%2L, verbose=TRUE]
```
It was an overwrite into R's heap which is hard to debug since all you get is a crash in the next gc which could be a long time after the root cause corruption occurred.  So the first thing to do is to torture so that you get closer to where the root cause corruption happens. This also enabled me to reproduce locally on 64bit Linux (it was nothing to do with 32bit Windows, it's just 32bit Windows is often most sensitive to this type of corruption as it seems to gc more, perhaps). Anyway, in this case, fortunately, it was this new test and the new functionality (it could have been an old lurking fault which was only now eked out by the new code/test and not due to the new code per se).  Anyway, in this case it was the gx = R_alloc() that had not been allocated big enough (sizeof(double) was the largest type and now it is sizeof(Rcomplex)). When it's an overwrite to memory in R's own heap (as R_alloc returns) ASAN doesn't track that. ASAN tracks malloc and calloc.  R's heap (R_alloc) is doing its own memory management so if our C code overwrites to that, those overwrites aren't detected by ASAN.  But by using torture and Rprintf I was able to hone in on the area of code where the overwrite was happening and then by looking at the code and comments I spotted the R_alloc sizeof too small since I knew to look for things like that.  In other words, by "yielded" I didn't mean that the tool just took me straight there; I mean it helped. Fresh eyes after a night's sleep also helped.

PR #3697, opened 2019-07-11
fwrite supports complex-valued columns, part of #3690

MichaelChirico:
#3690 

Not 100% confident I did the new writer & adapting `writeFloat64` correctly (particularly I wasn't sure why I needed to `*pch = ch` both in `writeFloat64` and `writeFloat64Scalar`. Extra eyes appreciated


MichaelChirico:
need to add a test of fwrite()ing am unsupported column type

Issue #3189, opened 2018-12-06
Master list of most-requested issues


MichaelChirico:
Redirecting from post at `/orgs/.../discussions`. Inspired by happening to notice an issue getting more support (through GitHub reactions) and borrowing the idea to scrape the GH API from [this](https://www.tidyverse.org/articles/2018/04/upvoting-issues/) nice article over at tidyverse. Here are all open issues with at least 5 positive reactions:

 - [ ] #635 - `[`
 - [ ] #3992 |`nafill`
 - [ ] #1639 |`on`
 - [ ] #2146 | `NEW`
 - [ ] #3957 | `NEW`
 - [ ] #1063 |`NEW`
 - [ ] #657 | `i`
 - [ ] #2778 | `froll`
 - [ ] #4970 | `NEW`
 - [ ] #3467 | `NEW`
 - [ ] #788 | `[`
 - [ ] #5621 | `DT()`
 - [ ] #4804 | `NEW`
 - [ ] #3241 | `froll`
 - [ ] #3739 |`uniqueN`
 - [ ] #1717 | `JOINS`
 - [ ] #599 | `JOINS`
 - [ ] #3721 | `JOINS`
 - [ ] #2879 | `NEW`
 - [ ] #2219 | `NEW`
 - [ ] #1700 | `JOINS`
 - [ ] #1543 | `:=`
 - [ ] #935 | `JOINS`
 - [ ] #857 | `JOINS`
 - [ ] #4697 | `ALTREP`
 - [ ] #4677 | `JOINS`
 - [ ] #4100 | `INTERNAL`
 - [ ] #4066 | `nafill`
 - [ ] #3815 | `GForce`
 - [ ] #3009 | `JOINS`
 - [ ] #3002 | `JOINS`
 - [ ] #1120 | `GForce`
 - [ ] #1086 | `NEW`
 - [ ] #5020 | `.SD`
 - [ ] #4572 | `NEW`
 - [ ] #3946 | `JOINS`
 - [ ] #3104 | `NEW`

-----

#### Fulfilled already

 - [x] #4348 | `INSTALLATION`
 - [x] #3060 | `NEW`
 - [x] #2181 | `JOINS`
 - [x] #2311 | `GForce`
 - [x] #4258 | `NEW`
 - [x] #3911 | `rbindlist` - [x] #2020 |`fwrite`| - fwrite ignores options(scipen = 999)
[...]

PR #3701, opened 2019-07-12
forderv handles complex input

### LINKED ISSUE #1703: Extend forder support to 'complex' type.

MichaelChirico:
Closes #1703 
and other parts of #3690.

I took a look into adding support "properly" (directly on the complex columns themselves) at the C level but the effort level there seems prohibitive -- would have to modularize the code a bit to allow `CPLXSXP` to delegate to `REALSXP` engine on `.r` and `.i` components. #1703 was filed 3 years ago without much follow-up, so I assume the simple approach here will do just fine. A quick benchmark suggested the current implementation beats `base::order` and an approach that explicitly delegates like `order(Re(z), Im(z))` in terms of speed & memory efficiency.

Sorting itself is done I think.

Now working on the stuff downstream of `forderv` that this should unlock -- `key`s, joins, etc.

TO DO:

 - [x] `frank`
 - [x] `rowid`
 - [x] `dcast`
 - [x] `by`
 - [x] `rleid`
 - [x] `setorder`/`setkey`
 - [x] `unique`
 - [x] add follow-up tests to #3689 (coverage in dogroups.c)


MichaelChirico:
Have taken this pretty far but could go for some help to go further.

@mattdowle could you weigh in on how to proceed with `bmerge` & `uniqlist`?

 - Joins break at https://github.com/Rdatatable/data.table/blob/master/src/bmerge.c#L234 this `SWITCH(TYPEOF())`. I'm not exactly sure what it's doing, but IIUC there's a pretty big logic shift around the [use of `dtwiddle`](https://github.com/Rdatatable/data.table/blob/master/src/bmerge.c#L346): `uint64_t (*twiddle)(void *, int) = isInt64 ? &i64twiddle : &dtwiddle;`, where `CPLXSXP` columns are _two_ 64-bit `REALSXP`. It feels like we'll have to internally mash `on = c(z1 = 'z2')` into `on = c(Re(z1) = Re(z2), Im(z1) = Im(z2))`, but I'm weary of going down that path too hastily since it'll be messy to clean up afterwards.
 - `setkey` fails using `uniqlist` at https://github.com/Rdatatable/data.table/blob/master/src/uniqlist.c#L56 this `SWITCH(TYPEOF())`. Again, not sure exactly what it's doing, but again there's use of `dtwiddle`. Same concerns.

Any thoughts?

R/setkey.R
```diff
@@ -51,14 +51,9 @@ setkeyv = function(x, cols, verbose=getOption("datatable.verbose"), physical=TRU
   }
   if (identical(cols,"")) stop("cols is the empty string. Use NULL to remove the key.")
   if (!all(nzchar(cols))) stop("cols contains some blanks.")
-  if (!length(cols)) {
```
MichaelChirico: this `if` is redundant & contradictory -- just a few lines earlier (L47) has `if (!length(cols))` as a `warning`. I guess this branch was some leftover or copy/paste from `setkey`... revealed by Codecov

mattdowle:
Ok so this is a lot of progress! But it's risky. There's a lot of core files changes here, and as we've seen from https://github.com/Rdatatable/data.table/pull/3692#issuecomment-512527727, adding support for the new type can have unintended and hard to track memory faults because it's a new type that's bigger too (16 bytes).

The new `cplxTwiddled` is 30 lines (out of +300 in this PR). I had been thinking we could remove dtwiddle actually, as I've been reducing the use of dtwiddle over time. This new twiddle would need to be tested and debugged separately for `complex`. The simpler way to handle complex in ordering would be to treat it as two columns of double and reuse the existing tested twiddle on real and imaginary parts. Either the complex column could be unpacked so it really was 2 columns in memory (i.e. unpack the rowwise two-doubles), or, the algos could be extended so that when they saw a `complex` column they could iterate twice over it (the first time over the real part and the second time over the imaginary part) as if it were two columns of double. The latter would save the allocation and unpacking step and would work better in joins too.

I haven't seen any requests from actual users who want to order, group and join on `complex` columns.  I think until we've got actual use-cases and requests from users, we should draw a line between i) supporting `complex` in value columns (those PRs merged, like ops on complex in `j` should work when grouping), and ii) grouping/joining on columns which are `complex`.  For now at least.

The other aspect is rather than adding `complex` support to grouping/joining as a new type with all the cases in this PR, a generic any-bytes approach should be possible which would also handle grouping by `list` columns. That's probably the way to go, but I see much higher value tasks on the list; e.g. https://github.com/Rdatatable/data.table/issues/3189

Can we close this PR? We can support complex in value columns now, but not join or group by until a user provides a use case that needs grouping or joining on complex columns.  The issue can be left open with this distinction clear.

mattdowle:
I hadn't seen this from @eliocamp before I wrote the comment above.
https://github.com/Rdatatable/data.table/pull/3689#issuecomment-512061333
Reading now ...

mattdowle:
@eliocamp The info in your comment linked above was very helpful, thanks!  In that `by = .(dataset, season, lev)` those columns (dataset, season, lev) are not `complex` iiuc.  So in data.table-dev as of now,  `complex` should work in value columns when grouping and joining non-complex columns.  But joining-on-complex and grouping-by-complex columns won't be supported.  Is that going to be ok for you?

MichaelChirico:
agree this is the heaviest new piece of code--> should think about potential maintenance burden.

perhaps this PR could be scaled back to do forder only? that part doesn't require the by/joins/unique to work, it simply creates a new sorting object based on the real&complex parts.

inefficient, but convenient.

mattdowle:
Major scaling back would be good. The reorder.c change here would be good to merge for example.  If `cplxTwiddled` can be removed then maybe forder.c, but then I'm looking at the 50 lines inside the n_cplx branch and wondering why that needs so many lines.

MichaelChirico:
the reorder.c change was originally this PR but I needed part of it for the bmerge/uniqlist stuff. I'll reopen that one to merge separately.

https://github.com/Rdatatable/data.table/pull/3688

MichaelChirico:
for the n_cplx branch, 12 lines or so are comments.

a lot of the rest is duplication -- apply to real part, apply to complex part.

but the idea is to reconstruct a new DT (in order of the original `by` vector) where the complex columns are replaced by two columns (real part, complex part)

MichaelChirico:
OK, have scaled back by removing the `cplxTwiddled` bit & anything using it.

The rest of the diff is pretty boilerplate & unlocks some decent functionality:

 - `by = z`
 - `dcast(z ~`
 - `frank(z)`
 - `rowid(z)`
 - `rleid(z)`
 - `set{key,order}(z)`

Will file (and close) a follow-up PR with the rest as a starting point if needed in the future

src/dogroups.c
```diff
@@ -153,34 +153,36 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
       SETLENGTH(I, grpn);
       INTEGER(I)[0] = 0;
       for (int j=0; j<length(xSD); ++j) {
         switch (TYPEOF(VECTOR_ELT(xSD, j))) {
-        case LGLSXP :
-          LOGICAL(VECTOR_ELT(xSD,j))[0] = NA_LOGICAL;
-          break;
+        case LGLSXP : // #nocov
+          LOGICAL(VECTOR_ELT(xSD,j))[0] = NA_LOGICAL; // #nocov
+          break; // #nocov
```
MichaelChirico: This is a case like in test 1540.36 (`on` column from `x` used in `j` during `by=.EACHI`)

IIUC `bmerge` will already have coerced `LGLSXP` to `INTSXP` by the time `dogroups` is working on it in this line, so it's not actually possible in the current code to reach this line.

I left as `nocov` but it could also be an `Internal error` as something to monitor?

eliocamp:
> @eliocamp The info in your comment linked above was very helpful, thanks! In that `by = .(dataset, season, lev)` those columns (dataset, season, lev) are not `complex` iiuc. So in data.table-dev as of now, `complex` should work in value columns when grouping and joining non-complex columns. But joining-on-complex and grouping-by-complex columns won't be supported. Is that going to be ok for you?

Yes, its eminently reasonable. Complex values are probably exclusively used as measured values and not category-like so I don't imagine there are many reasonable workflows that would include a `by = complex` and in that case using `by = ReIm(complex)` would be a relatively easy workaround.

src/forder.c
```diff
@@ -440,11 +440,62 @@ SEXP forder(SEXP DT, SEXP by, SEXP retGrpArg, SEXP sortGroupsArg, SEXP ascArg, S
...
+        SET_VECTOR_ELT(new_dt, j, VECTOR_ELT(DT, by_idx));
+        INTEGER(new_asc)[j] = INTEGER(ascArg)[i];
+        INTEGER(new_by)[j] = j+1;
+        j += 1;
+      }
+    }
+    DT = new_dt;
+    ascArg = new_asc;
+    by = new_by;
   }
```
mattdowle: I'm a bit nervous about this branch.  I looks constrained to only when n_cplx>0 (which is good that it's isolated), and I know _what_ it is doing (splitting complex into real and imaginary columns) but I don't see _how_ it's doing it. I'm going to need more time to review it and make sure it doesn't impact anything unexpected.

mattdowle: How about dealing with complex columns further down in this switch.  (GitHub wouldn't let me add a comment on that line as it's outside the diff perhaps? so I pasted an image)
image

src/uniqlist.c
```diff
@@ -200,6 +200,11 @@ SEXP rleid(SEXP l, SEXP cols) {
           // 8 bytes of bits are identical. For real (no rounding currently) and integer64
           // long long == 8 bytes checked in init.c
           break;
+        case CPLXSXP: {
+          // tried to make long long complex * but got a warning that it's a GNU extension
+          Rcomplex *pz = COMPLEX(jcol);
+          same = (long long)pz[i].r == (long long)pz[i-1].r && (long long)pz[i].i == (long long)pz[i-1].i;
```
mattdowle: This `(long long)` is a _cast_ and won't achieve the desired op.  This cast will trunc 3.4 to 3 for example.  What's needed is a type _pun_, which is what the REALSXP case just above does.

src/uniqlist.c
```diff
@@ -232,6 +237,14 @@ SEXP rleid(SEXP l, SEXP cols) {
         }
       }
       break;
+    case CPLXSXP: {
+      Rcomplex *pzjcol = COMPLEX(jcol);
+      for (R_xlen_t i=1; i<nrow; i++) {
+        bool same = (long long)pzjcol[i].r == (long long)pzjcol[i-1].r &&
+                    (long long)pzjcol[i].i == (long long)pzjcol[i-1].i;
```
mattdowle: same here (pun needed not cast)

src/forder.c
MichaelChirico: when i first took a look i decided against that (I think my initial reaction was `range` doesn't make sense for complex, but of course it does because they're well-ordered using the lexicographic rule). will take another look.

The `n_cplx` branch is a bit messy, and took me several iterations to get there as I kept failing a few more tests each time. I think it can be cleaned up a little by snazzier use of the `j` index.

Idea is: receive `by` which is say length `n`, `m` of which are complex columns.

Instead of sorting using the input list `DT`, do so using a new list (constructed as `new_dt`).

If there are no complex columns, `new_dt[[1]]` is `DT[[ by[1] ]]`, `new_dt[[2]]` is `DT[[ by[2] ]]`, and so on.

If there _are_ complex columns, say the first is at the third index of `by`, then `new_dt[[3]]` is `Re( DT[[ by[3] ]] )` and `new_dt[[4]]` is `Im( DT[[ by[3] ]] )`. Then continuing on as usual. That "bump" of the index by a complex column is what made things complicated.

mattdowle: I wasn't thinking of doing a single range for `complex`.  The thought was to trick the loop into doing a complex column twice. The first time doing range for the real part only. The 2nd time calling range for the imaginary part only.  There might need to be new range_complex_r and range_complex_i perhaps.

mattdowle: The other option is to postpone this PR to a later date. Elio's response above confirms that setkey-on-complex and grouping-on-complex isn't high priority. It was `complex` value columns that was the main thing;  which you did and merged.

src/uniqlist.c
MichaelChirico: So to clarify in the `REALSXP` branch:

```
long long *ll = (long long *)REAL(jcol);
same = ll[i]==ll[i-1];
```

What's happening here is `REAL` is _actually_ a pointer to `double`, but `ll` is a `long long` pointer to the same address as where `jcol` is.

So then `ll[i] == ll[i-1]` treats the elements of `jcol` (which are actually real) as 8-byte integers & compares those 64 bits.

Is that the gist of it?

src/forder.c
MichaelChirico: Sure. The main thing done in this PR is to close #1703 (grouping/keying/others mentioned here come almost-for-free after that). And this branch is the workhorse for that. Could `wontfix`&close that issue I guess.

src/uniqlist.c
mattdowle: Yes: exactly.

MichaelChirico: Is there a native 16-byte type to pun to? I'm not seeing one. I guess we could just use pure equality-on-doubles & leave it at that as it's low enough priority.

mattdowle: Another way, since all the cases have the same code other than the comparison, is to take away all the cases and just do one `memcmp` for all types (just pass the size to memcmp).  Would need benchmarking on large data to see if it was any slow.   (I didn't look closely to see if there are any other difference between all those cases in this switch in this file.)

src/forder.c
MichaelChirico: > I wasn't thinking of doing a single range for complex. The thought was to trick the loop into doing a complex column twice. The first time doing range for the real part only. The 2nd time calling range for the imaginary part only. There might need to be new range_complex_r and range_complex_i perhaps.

Oh, I think I see now. The idea is, instead of bracketing `[re_min, re_max]` and `[im_min, im_max]` separately, to just bracket both components with `[min(re_min, im_min), max(re_max, im_max)]`

mattdowle: Not sure what you mean by bracketing.  But I meant treating the complex column just as if it were two separate columns of double.

mattdowle: It would nice for completeness to get this merged and closed #1703, yes. But I'm also an eye that we've both been working on complex number support for 3 days now (elapsed) and continuing. And Elio said going this far (sorting complex) isn't needed. There isn't any demand for this from users as far as I can see.

MichaelChirico: I think the biggest impediment to that is I don't see a way to get a `double *` object pointing to the real/imaginary _vectors_ (like `REAL(COMPLEX(x))` or `COMPLEX(x).r`, hope that makes sense), which is what all of the functions for `REAL` columns take as input.

MichaelChirico: > It would nice for completeness to get this merged and closed #1703, yes. But I'm also an eye that we've both been working on complex number support for 3 days now (elapsed) and continuing. And Elio said going this far (sorting complex) isn't needed. There isn't any demand for this from users as far as I can see.

True... I'm just loath to close the PR given that it's passing tests. Also understand the tradeoff that is the maintenance burden.

mattdowle: > I think the biggest impediment to that is I don't see a way to get a double * object pointing to the real/imaginary vectors (like REAL(COMPLEX(x)) or COMPLEX(x).r, hope that makes sense), which is what all of the functions for REAL columns take as input.

Yes makes sense. That's why I wrote above: "There might need to be new range_complex_r and range_complex_i perhaps."   But I didn't look at it closely.

> maintenance burden

It's not a burden in the sense that more code is just always generally bad.  It's that this particular change could cause hard to trace bugs because of the type of change it is to such a core function.  It's risk.  And it's time in reviewing it.

> it's passing tests

But this shows the tests aren't sufficient: the cast isn't correct but it's still passing tests.

MichaelChirico: > But this shows the tests aren't sufficient: the cast isn't correct but it's still passing tests.

Very true, I meant more that it's passing _existing_ tests (so "no" existing functionality is being affected).

Would adding a lot more tests to the suite for the new code be a good direction?

mattdowle: A good direction would be to focus on the most requested issues: #3189

PR #3715, opened 2019-07-20
keep.rownames argument for transpose

### LINKED ISSUE #1886: feature request : transpose(keep.rownames = T)?
transposing up to now required this:

```
dt = data.table(x=1:5, y=6:10)
t(dt) %>% data.table(keep.rownames = T)
```

, so the `transform` function has potential to cut this short. 
It has however no `keep.rownames = T` argument (or something similar), so the variable names disappear:

```
transpose(dt)
#    V1 V2 V3 V4 V5
#1:  1  2  3  4  5
#2:  6  7  8  9 10
```

what I hoped to see was something like this:

```
#    rn  V1 V2 V3 V4 V5
#1:  x  1  2  3  4  5
#2:  y  6  7  8  9 10
```

MichaelChirico:
Closes #1886
via #3189 

I think the logic is basically done. Just want to do some more robustness checks & minor details, e.g.

As of now, the new column is named `V1`, should we name it `rn` & bump the others to `1-n`? `2-(n+1)`?


mattdowle:
The confusing thing for me on this one is that `t(dt)` already does keep the row names :

```
> dt = data.table(x=1:5, y=6:10)
> dt
       x     y
   <int> <int>
1:     1     6
2:     2     7
3:     3     8
4:     4     9
5:     5    10
> t(dt)
  [,1] [,2] [,3] [,4] [,5]
x    1    2    3    4    5    # there are the row names kept (x and y)
y    6    7    8    9   10
> class(t(dt))
[1] "matrix"
>
```

So the new `keep.rownames` argument feels more like it should be named `return a data.table` perhaps.
Also I can see a new request being created in future that  DT == t(t(DT)) should be true.  That's not true currently because t() returns a matrix.  Maybe a new function name like `twist`, `mirror`, `rotate`, `invert`, `spin` to convey that a data.table would be returned.
If one were to rotate a DT 90 degrees clockwise, then the user specifying which column should become the column names would make sense (and that argument would be `col.names=` or similar).
If one were to rotate a DT 90 degrees anticlockwise, then the column names becoming the first column called `rn` by default makes perfect sense (but setting `col.names=NULL` would drop them, perhaps).
I briefly considered `screw` and `unscrew`. But I guess a lot of people don't know that screw=clockwise and unscrew=anticlockwise.

MichaelChirico:
Also, `screw(my_data)` sounds like a frustrated Friday evening more than a successful manipulation 😉

MichaelChirico:
BTW @mattdowle this PR / issue is for `transpose`, not `t`

mattdowle:
Oh. We have our own `transpose`?  There doesn't seem to be a `data.table` method for `t()`.

MichaelChirico:
`transpose` is the workhorse of `tstrsplit` & it's also been exported since then yep

mattdowle:
> As of now, the new column is named V1, should we name it rn 

Yes `rn` feels right to me.

> bump the others to 1-n? 2-(n+1)?

Not sure what you mean here,  but   `rn`, `V1`, `V2`, ...  seems right to me just like the original issue showed.

I'm not sure about the argument name `keep.rownames`.   To me is means take the row names that exist and keep them.  But that's not what this argument does at all. In fact it is to do with keeping column names, not row names.  Now that I realize that `transpose` takes a data.table input and returns a data.table, then rownames don't come into it really.   Plus there needs to be another argument to specify which column becomes the column names (if any) and what should that be called?   How about changed keep.rownames to `names.to.column`, and adding `column.to.names`.

Hm, I'm thinking there is no argument name that easily conveys what it does.  How about changing `keep.rownames` to just `rn`.  In other places in data.table we have an `rn` column at the beginning so users might easily guess what that means?    So  `rn=TRUE` would mean yes-create-rn  (i.e. use column names to create the rn column).  Then `cn="rn"` (a 2nd new argument) could be used to do the reverse  (i.e. use the "rn" column to create the column names in the result).

NEWS.md
```diff
@@ -135,7 +135,29 @@
...
 
-21. `transpose` gains `keep.rownames` argument which stores the input's names at the beginning of the output (the first list element or the first column), [#1886](https://github.com/Rdatatable/data.table/issues/1886). Thanks to @ghost for the request.
+21. New function `fifelse(test, yes, no)` has been implemented in C by Morgan Jacob, [#3657](https://github.com/Rdatatable/data.table/issues/3657). It is comparable to `base::ifelse`, `dplyr::if_else`, `hutils::if_else`, and (forthcoming) [`vctrs::if_else()`](https://vctrs.r-lib.org/articles/stability.html#ifelse). It returns a vector of the same length as `test` but unlike `base::ifelse` the output type is consistent with those of `yes` and `no`. Please see `?data.table::fifelse` for more details.
+
+    ```R
+    # default 4 threads on a laptop with 16GB RAM and 8 logical CPU
+    x = sample(c(TRUE,FALSE), 3e8, replace=TRUE)  # 1GB
```
jangorecki: why you went 5e8 to 3e8 for ifelse benchmark?

NEWS.md
```diff
@@ -135,7 +135,29 @@
...
+      times = 5L, unit="s"
+    )
+    # Unit: seconds
+    #                            expr  min  med  max neval
+    #        base::ifelse(x, 7L, 11L)  8.5  8.6  8.8     5
+    #      dplyr::if_else(x, 7L, 11L)  9.4  9.5  9.7     5
+    #     hutils::if_else(x, 7L, 11L)  2.6  2.6  2.7     5
+    # data.table::fifelse(x, 7L, 11L)  1.5  1.5  1.6     5  # setDTthreads(1)
+    # data.table::fifelse(x, 7L, 11L)  0.8  0.8  0.9     5  # setDTthreads(2)
+    # data.table::fifelse(x, 7L, 11L)  0.4  0.4  0.5     5  # setDTthreads(4)
```
jangorecki: why not 8 threads too? IMO only make sense 1, default, max

NEWS.md
mattdowle: Just to make it 1GB.  I don't remember changing it from 5e8 to 3e8 though; can't remember -- didn't I change something else at the same time too? Main change was getting away from 100 runs on small size. It can be 5e8 too.

NEWS.md
mattdowle: I have 4 cores,  hyper-threading gives 8 but then that leaves nothing for the OS and other apps/services.  It was on my laptop and I didn't want to include a fully-maxed-out number of threads. On a dedicated server that's ok,  but this was just a quick benchmark on my laptop for purpose of news item.

mattdowle: The reason for including 2 was to show the halving in time from 1 to 2. Then going to 4 halving again.  Which is pretty efficient parallelism.  As long as the single-thread run is not slower than a non-multithreaded-enabled algo, which is why I showed 1 thread just after the next fastest (hutils).

mattdowle:
Follow-up PRs expected to be needed.
I can see coverage failing on Travis but I don't see why.  Will merge anyway and see if master checks throw any light on it.

mattdowle:
Coverage issue seems to be related to `covr` upgrade yesterday. I filed issue: https://github.com/r-lib/covr/issues/391

MichaelChirico:
`covr` 3.3.1 is on CRAN now btw

PR #3720, opened 2019-07-21
add default to all switch statements in src; coverage

MichaelChirico:
Inspired by: https://github.com/Rdatatable/data.table/pull/3692#commitcomment-34330836

Ran:

```
for (f in list.files('src', pattern = '\\.c', full.names = TRUE)) {
  l = readLines(f)
  nswitch = sum(grepl('switch\\s*\\(', l))
  ndefault = sum(grepl('default\\s*:', l))
  if (nswitch != ndefault)
    cat(sprintf('%s switches: %d, defaults: %d\n', basename(f), nswitch, ndefault))
}
```

And found the files here with `default`-less switches. Added internal errors for them+`nocov`'d, except in the one case where it was actually reasonable to reach there. Had to make a small edit to `writeNA` to make it even easier to do so (even without that `fill=expression(1)` would get there).

Mostly, the `default`-less switches are `switch(enum)` so it should be impossible in the current code to reach. These errors guard against editing the `enum`s' valid values in the future & neglecting to cover all the bases in testing new functionality.

### DIFF
--- a/src/fcast.c
+++ b/src/fcast.c
@@ -59,7 +59,7 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
         }
       }
     } break;
-    case STRSXP: {
+    case STRSXP:
       for (int j=0; j<ncols; ++j) {
         SET_VECTOR_ELT(ans, nlhs+j+i*ncols, target=allocVector(TYPEOF(thiscol), nrows) );
         copyMostAttrib(thiscol, target);
@@ -68,8 +68,8 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
           SET_STRING_ELT(target, k, (thisidx == NA_INTEGER) ? STRING_ELT(thisfill, 0) : STRING_ELT(thiscol, thisidx-1));
         }
       }
-    }  break;
-    case VECSXP: {
+      break;
+    case VECSXP:
       for (int j=0; j<ncols; ++j) {
         SET_VECTOR_ELT(ans, nlhs+j+i*ncols, target=allocVector(TYPEOF(thiscol), nrows) );
         copyMostAttrib(thiscol, target);
@@ -78,7 +78,7 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
           SET_VECTOR_ELT(target, k, (thisidx == NA_INTEGER) ? VECTOR_ELT(thisfill, 0) : VECTOR_ELT(thiscol, thisidx-1));
         }
       }
-    }  break;
+      break;
     default: error("Unsupported column type in fcast val: '%s'", type2char(TYPEOF(thiscol))); // #nocov
     }
     if (count) UNPROTECT(1);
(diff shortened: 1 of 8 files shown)


src/fcast.c
```diff
@@ -69,8 +68,8 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
           SET_STRING_ELT(target, k, (thisidx == NA_INTEGER) ? STRING_ELT(thisfill, 0) : STRING_ELT(thiscol, thisidx-1));
         }
       }
-      break;
-    case VECSXP:
+    }  break;
```
mattdowle: case statements only need `{` when there's a local variable defined within that case. Otherwise it won't compile.  So I've rounded on only putting `{` when they are needed to remind my eyes that the case has local variables.  It doesn't matter and I'll merge anyway.  Just explaining the background.

src/frollR.c
```diff
@@ -113,7 +113,7 @@ SEXP frollfunR(SEXP fun, SEXP obj, SEXP k, SEXP fill, SEXP algo, SEXP align, SEX
   if (!dans) error("%s: Unable to allocate memory answer", __func__); // # nocov
   double* dx[nx];                                               // pointers to source columns
   uint_fast64_t inx[nx];                                        // to not recalculate `length(x[[i]])` we store it in extra array
-  if (bverbose) Rprintf("%s: allocating memory for results %dx%d\n", __func__, nx, nk);
+  if (verbose) Rprintf("%s: allocating memory for results %dx%d\n", __func__, nx, nk);
```
mattdowle: `bverbose` was correct here.  `verbose` is the SEXP passed to the function.  I've often appended "Arg" to argument names at C level to avoid confusion like this, since we often use the SEXP args at the top of the function to check and extract the raw C values (like bool verbose, called `bverbose` in this case). The change to `verbose` is passing checks because it's always true;  a pointer value is always true (the argument is never NULL).  The extra verbosity is passing checks because our `test()` doesn't check for the lack of verbosity; that's usually caught when we or users run test.data.table ourselves.   But that doesn't catch this one either because this one is over in the separate froll tests script.

src/ijoin.c
```diff
@@ -34,15 +34,15 @@ SEXP lookup(SEXP ux, SEXP xlen, SEXP indices, SEXP gaps, SEXP overlaps, SEXP mul
   count = (int *)INTEGER(VECTOR_ELT(ux, uxcols-2));
   type_count = (int *)INTEGER(VECTOR_ELT(ux, uxcols-1));
   switch (mult) {
-  case FIRST:
+  case FIRST: {
     switch(type) {
-    case EQUAL:
+    case EQUAL: {
```
mattdowle: I commented earlier about the `{` in case statements and was going to merge anyway as it wasn't a big deal.  But then I carried on through the PR and saw the many more changes (168) in this file (ijoin.c).  Best to undo these `{` additions then please. General rule: only use curlies please when they're needed to declare variables local to that case.

src/fcast.c
MichaelChirico: gotcha, the rule makes sense, good to know

MichaelChirico:
@mattdowle done with `{`, there are still a bunch of whitespace changes, shall I bother?

`-w` diff in UI:

https://github.com/Rdatatable/data.table/pull/3720/files?w=1

mattdowle:
great. no worries - I did the whitespace as I was in diff anyway

mattdowle:
Wow - that's an increase of 0.34% coverage! That's a pretty big bang for that buck of relatively few lines.

PR #3724, opened 2019-07-23
Use proper NA for integer64 on assign

### LINKED ISSUE #1459: Integer64 columns not assigned NA when doing a left merge

### LINKED ISSUE #3723: Partially create an integer64 column results in garbage values
``` r
library(data.table)
dt <- data.table(
  x = c(1L, 2L)
)
dt[x == 1L, y:=bit64::as.integer64(5)][]
#>    x                   y
#> 1: 1                   5
#> 2: 2 9218868437227407266
str(dt)
#> Classes 'data.table' and 'data.frame':   2 obs. of  2 variables:
#>  $ x: int  1 2
#>  $ y:integer64 5 9218868437227407266 
#>  - attr(*, "index")= atomic  
#>   ..- attr(*, "__x")= int
```

<sup>Created on 2019-07-23 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

``` r
devtools::session_info()
#>  setting  value                                              
#>  version  R version 3.4.4 (2018-03-15)                       
#>  os       Windows 7 x64 SP 1                                 
#>  system   x86_64, mingw32                                    
#>  ui       RTerm                                              
#>  language (EN)                                               
#>  collate  Chinese (Simplified)_People's Republic of China.936
#>  ctype    Chinese (Simplified)_People's Republic of China.936
#>  tz       Asia/Taipei                                        
#>  date     2019-07-23                                         
#> 
#>  package     * version    date       lib source        
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 3.4.4)
#>  backports     1.1.3      2018-12-14 [1] CRAN (R 3.4.4)
#>  bit           1.1-14     2018-05-29 [1] CRAN (R 3.4.4)
#>  bit64         0.9-7      2017-05-08 [1] CRAN (R 3.4.0)
#>  callr         3.2.0      2019-03-15 [1] CRAN (R 3.4.4)
#>  cli           1.1.0      2019-03-19 [1] CRAN (R 3.4.4)
#>  crayon        1.3.4      2017-09-16 [1] CRAN (R 3.4.4)
#>  data.table  * 1.12.3     2019-07-08 [1] local         
#>  desc          1.2.0      2018-05-01 [1] CRAN (R 3.4.4)
#>  devtools      2.0.1      2018-10-26 [1] CRAN (R 3.4.4)
#>  digest        0.6.18     2018-10-10 [1] CRAN (R 3.4.4)
#>  evaluate      0.13       2019-02-12 [1] CRAN (R 3.4.4)
#>  fs            1.2.7      2019-03-19 [1] CRAN (R 3.4.4)
#>  glue          1.3.1      2019-03-12 [1] CRAN (R 3.4.4)
#>  highr         0.8        2019-03-20 [1] CRAN (R 3.4.4)
#>  htmltools     0.3.6.9000 2018-05-07 [1] local         
#>  knitr         1.22       2019-03-08 [1] CRAN (R 3.4.4)
#>  magrittr      1.5        2014-11-22 [1] CRAN (R 3.4.4)
#>  memoise       1.1.0      2017-04-21 [1] CRAN (R 3.4.4)
#>  pkgbuild      1.0.3      2019-03-20 [1] CRAN (R 3.4.4)
#>  pkgload       1.0.2      2018-10-29 [1] CRAN (R 3.4.4)
#>  prettyunits   1.0.2      2015-07-13 [1] CRAN (R 3.4.1)
#>  processx      3.3.0      2019-03-10 [1] CRAN (R 3.4.4)
#>  ps            1.3.0      2018-12-21 [1] CRAN (R 3.4.4)
#>  R6            2.4.0      2019-02-14 [1] CRAN (R 3.4.4)
#>  Rcpp          1.0.1      2019-03-17 [1] CRAN (R 3.4.4)
#>  remotes       2.0.2      2018-10-30 [1] CRAN (R 3.4.4)
#>  rlang         0.3.3      2019-03-29 [1] CRAN (R 3.4.4)
#>  rmarkdown     1.12       2019-03-14 [1] CRAN (R 3.4.4)
#>  rprojroot     1.3-2      2018-01-03 [1] CRAN (R 3.4.3)
#>  sessioninfo   1.1.1      2018-11-05 [1] CRAN (R 3.4.4)
#>  stringi       1.4.3      2019-03-12 [1] CRAN (R 3.4.4)
#>  stringr       1.4.0      2019-02-10 [1] CRAN (R 3.4.4)
#>  testthat      2.1.1.9000 2019-06-22 [1] local         
#>  usethis       1.4.0      2018-08-14 [1] CRAN (R 3.4.4)
#>  withr         2.1.2      2018-03-15 [1] CRAN (R 3.4.4)
#>  xfun          0.6        2019-04-02 [1] CRAN (R 3.4.4)
#>  yaml          2.2.0      2018-07-25 [1] CRAN (R 3.4.4)
#> 
#> [1] D:/app/R_lib/3.1
#> [2] D:/app/R/library
```

</details>

MichaelChirico:
Closes #3723 
Closes #1459

There's probably a better way to do this, but I couldn't think of one.

The problem is `writeNA` is `integer64`-aware, but `allocNAVector` is not because it operates on `SEXPTYPE` so it forgets about the source vector & its attributes.

Landed on defining `allocNAVectorLike` to allocate an empty vector akin to `x`. Played with two other versions before scrapping:

 - Completely re-define `allocNAVector` to operate as `allocNAVectorLike` does. This didn't work because of the usages in `fmelt`, [e.g.](https://github.com/Rdatatable/data.table/blob/master/src/fmelt.c#L401). There, `allocNAVector` isn't matching a specific column, but rather the bumped-to type in reshaping (using the type hierarchy on the `measure.vars`).
 - Add a parameter `is_i64` to `allocNAVector` and `writeNA`. First this feels pretty hackish (what other attributes will need a new/different parameter?), and second I got stuck because some usages of `writeNA` are sending an `SEXP` and I'm not confident we can always `INHERITS(x, char_integer64)` in those cases in `rbindlist`.


inst/tests/tests.Rraw
```diff
@@ -15428,6 +15428,14 @@ test(2071.10, dcast(data.table(a=1, b=1, l=list(list(1))), a ~ b, value.var='l')
 test(2071.11, dcast(data.table(a = 1, b = 2, c = 3), a ~ b, value.var = 'c', fill = '2'),
      data.table(a=1, `2`=3, key='a'))
 
+# partial instantiation of integer64 column was creating NA_REAL, not INT64_MIN
+if (test_bit64) {
+  # sub-assign from #3723
+  test(2072.1, data.table(x=1:2)[1, y := bit64::as.integer64(0L)]$y, as.integer64(c(0L, NA)))
```
jangorecki: we don't need `bit64::` prefix as bit64 is attached

src/assign.c
```diff
@@ -1040,6 +1040,14 @@ SEXP allocNAVector(SEXPTYPE type, R_len_t n)
   UNPROTECT(1);
   return(v);
 }
+// #3723 -- build NA vector like x -- mainly used to copy attributes pass this to writeNA
+SEXP allocNAVectorLike(SEXP x, R_len_t n) {
```
jangorecki: why not put that in `assign.c`?
```c
SEXP v = PROTECT(allocNAVector(TYPEOF(thisvalue), nrow));
copyMostAttrib(thisvalue, v);
SET_VECTOR_ELT(dt, coln, newcol=v)
UNPROTECT(1);
```

jangorecki: ok, I read `writeNA` and now I understand

jangorecki: Although I am not sure if it make sense to copy attrs every time, some of them might be heavy.
Maybe check for class and redirect for int64, or copy class only?

MichaelChirico: Feels weird to me to be so `int64`-specific in a fix... users with heavy `attr` can use `as.vector` to strip attributes?

jangorecki:
There is a more tricky case that is being handled by this PR
```r
data.table(a=1:2, b=bit64::as.integer64(1:2))[2L, c("a","b") := NA][]
#       a                   b
#   <int>               <i64>
#1:     1                   1
#2:    NA 9218868437227407266
```
You might not try to address it in this PR. I rewrote https://github.com/Rdatatable/data.table/blob/a2e7f5e82f9ccba448eb4cfc08e65d5a01931dbf/src/utils.c#L125 to be more generic and it should be well applicable in this case.

PR #3731, opened 2019-07-30
between NA bounds, nanotime support

### LINKED ISSUE #3522: between changed behaviour from NA bounds 'unknown' to 'missing'
Dear all,

I have discovered recently the change introduced in version 1.12.0 regarding `between` and how `NA` are now treated.

Given this change may break routines relying on the previous behavior, I was wondering whether in the future a parameter could be introduced so that `NA` are not treated as missing values but as `NA` (i.e. would return the same as `x >= lower & x <= upper`).

Thank you very much

Best,
Cyrille

### LINKED ISSUE #3667: Different `%between%` behavior on character vectors undocumented
At the moment, the help documentation for `%between%` states that 

> between is equivalent to x >= lower & x <= upper when incbounds=TRUE, or x > lower & y < upper when FALSE. With a caveat that NA in lower or upper are taken as a missing bound and return TRUE not NA.

While this appears to be true for numeric vectors, for character vectors, `NA` is not treated as a missing bound and yields `NA`, as seen below, and this difference is not noted in the documentation. Since (theoretically) the lexicographic order is a total order, one might expect the same or similar behavior unless indicated otherwise, and the documentation appears to imply this, given it does not mention character vectors are treated differently. The different behavior on character vectors should probably be either (a) documented if this is intended or (b) altered to treat `NA` as a missing bound (as with numeric vectors).

`#` `Minimal reproducible example`]

```
library(data.table)

numbers <- 1:26
numbers %between% c(13, NA)
#  [1] FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE  TRUE
# [14]  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE
letters %between% c("m", NA_character_)
#  [1] FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE    NA
# [14]    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA
```

`#` `Output of sessionInfo()`

```
R version 3.6.0 (2019-04-26)
Running under: macOS Mojave 10.14.5

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] data.table_1.12.3

[1] compiler_3.6.0 tools_3.6.0   
```
`data.table` package is the most recent stable development version pulled today using `install.packages`.

jangorecki:
closes #3667
For upper/lower argument of vector type (non-scalar) could be better optimised. IMO it is not a priority as we ultimately need to push down character handling into C. Which was also mentioned in comment

> TODO: support character between in Cbetween and remove this branch

This PR at least put expected functionality and unit tests in place so when we proceed to C development we have a properly working R code to match to.

----

also closes #3522 by fallback to R code

----

also added nanotime support


jangorecki:
failure of CI is quite interesting.
github is automerging this branch into fifelse, and CI is testing some artificial merge commit: https://github.com/Rdatatable/data.table/commit/ea7d608bec892917b4940dfc63eba54f150f4472
at first glance it looks like a bug

mattdowle:
This one was tricky! Changes I made were 
* NAbounds implemented at C level
* Supported character at C level
* Changed new NAbounds= from TRUE/FALSE to TRUE/NA. The idea being that TRUE/FALSE is hard to remember which way around it means. But if you remember it as what is returned for that bound when there's an NA, then TRUE/NA hopefully is easier to remember.  Explained in the news item and man page too, using the adjective 'unlimited' bound. 
* Rinherits() internal C function created which observes S4 inheritance from integer64 (nanotime).
* Moved INHERITS to utils.c to be next to Rinherits(). Internal functions so we can change their names to be clearer in future.

Issue #3270, opened 2019-01-11
Clarify error message on by = list()


MichaelChirico:
This line has an error message:

https://github.com/Rdatatable/data.table/blob/master/R/data.table.R#L954

```
if (length(bysubl)<2L) stop("When 'by' or 'keyby' is list() we expect something inside the brackets")
```

However I don't think it's possible to reach that branch in the current code flow because:

```
DT = data.table(a = 1:10)
DT[ , sum(a), by = .()]
#    V1
# 1: 55
```

This output looks OK to me... should we just axe this error, or should we move the branch logic so that it actually gets triggered (e.g. put the `length(bysubl) < 2L` condition a few lines earlier)?

jangorecki:
Mentioned error has been already removed, I filled PR adding unit tests for current behaviour. It will also address my issue reported in #648.

PR #3738, opened 2019-08-01
fix autonaming issue in groupings sets for special symbols

### LINKED ISSUE #3653: groupingsets with autonamed j creates extra column for ungrouped set
I stumbled over this when I was rolling up sample size (`.N`) for different sets of grouping variables, including no grouping.

If using `groupingsets` with an unnamed expression in `j` which is "autonamed" (e.g. `.N`) and including an "ungrouped" grouping set, then an additional variable is created for the ungrouped set. 

Some data:

    d <- data.table(x = c("a", "a", "b"))
    d
    #    x
    # 1: a
    # 2: a
    # 3: b
With a "grouped" grouping set only, an unnamed expression with `.N`  in `j` works: the result is autonamed `N`:

    groupingsets(d, j = .N, by = "x", sets = list("x"))
    #    x N
    # 1: a 2
    # 2: b 1

However, when also including an "ungrouped" grouping set (`character()`), an additional column is generated for the ungrouped result:

    groupingsets(d, j = .N, by = "x", sets = list("x", character()))
    #       x  N V1
    # 1:    a  2 NA
    # 2:    b  1 NA
    # 3: <NA> NA  3

With an ungrouped grouping set only, the `N` column is included despite it was only used for the grouped result in the previous example:

    groupingsets(d, j = .N, by = "x", sets = list(character()))
    #       x  N V1
    # 1: <NA> NA  3

Similar examples with named expressions in `j`, like `j = .(N = .N)`, works.
 
Additional (admittedly contrived) attempts with `.GRP` and `.I` in `j` give similar results. Thus, it seems like this behaviour may be related to autonaming.

jangorecki:
closes #3653


jangorecki:
Code simplified, no need for workarounds because `by=.()` already works. Related https://github.com/Rdatatable/data.table/issues/3270.

PR #3741, opened 2019-08-03
fifelse followup

### LINKED ISSUE #3740: fifelse followup
- names of `test` could be retained
```r
fifelse(c(a=T,b=F), 1:2, 2:1)
#[1] 1 1
ifelse(c(a=T,b=F), 1:2, 2:1)
#a b 
#1 1 
```
- recycling here needs to be fixed, AFAIU might result into segfault
```r
fifelse(c(TRUE, FALSE), 1:2, list(0L))
#[1]         1 544979712
ifelse(c(TRUE, FALSE), 1:2, list(0L))
#[[1]]
#[1] 1
#
#[[2]]
#[1] 0
```
- remove hutils, dplyr code from examples
- remove console output in examples

jangorecki:
closes #3740

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15509,6 +15509,17 @@ test(2072.066, fifelse(c(TRUE, TRUE, TRUE, FALSE, NA, FALSE), factor(letters[1:6
 test(2072.067, fifelse(c(TRUE, NA, TRUE, FALSE, FALSE, FALSE), factor(NA), factor(NA)),
                factor(c(NA,NA,NA,NA,NA,NA)))
 
+# fifelse #3740
+if (test_bit64) {
+  i = as.integer64(1:4)+3e9
+  test(2073.01, fifelse(c(TRUE,FALSE,NA,TRUE), i, i+100), c(i[1L], i[2L]+100, as.integer64(NA), i[4]))
+}
+if (test_nanotime) {
+  n = nanotime(1:4)
+  test(2073.02, fifelse(c(TRUE,FALSE,NA,TRUE), n, n+100), c(n[1L], n[2L]+100, as.integer64(NA), n[4]))
+}
+
+
 ###################################
 #  Add new tests above this line  #
 ###################################
--- a/man/fifelse.Rd
+++ b/man/fifelse.Rd
@@ -1,88 +1,41 @@
 \name{fifelse}
 \alias{fifelse}
+\alias{ifelse}
 \title{ Fast ifelse }
 \description{
-\code{data.table::fifelse} is comparable to \code{base::ifelse}, \code{dplyr::if_else} and  \code{hutils::if_else}.
-Like these functions, it returns a value with the same length as \code{test} and filled with value from \code{yes} or \code{no}.
-This function is mostly written in C for speed and unlike \code{base::ifelse} the output type is consistent with those of \code{yes} and \code{no}.
+\code{fifelse} is a faster and more robust replacement of \code{\link[base]{ifelse}}. It returns a value with the same length as \code{test} filled with corresponding values from \code{yes}, \code{no} or eventually \code{NA}, depending on \code{test}.
 }
 \usage{
   fifelse(test, yes, no)
 }
 \arguments{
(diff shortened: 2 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -15482,10 +15482,10 @@ test(2072.042, fifelse(TRUE,1L,factor(letters[1])),     error="'yes' has differe
 test(2072.043, fifelse(TRUE, list(1:5), list(5:1)), list(1:5))
 test(2072.044, fifelse(as.logical(NA), list(1:5), list(5:1)), list(NULL))
 test(2072.045, fifelse(FALSE, list(1:5), list(5:1)), list(5:1))
-test(2072.046, fifelse(TRUE, data.table(1:5), data.table(5:1)), data.table(1:5))
-test(2072.047, fifelse(FALSE, data.table(1:5), data.table(5:1)), data.table(5:1))
-test(2072.048, fifelse(TRUE, data.frame(1:5), data.frame(5:1)), data.frame(1:5))
-test(2072.049, fifelse(FALSE, data.frame(1:5), data.frame(5:1)), data.frame(5:1))
+test(2072.046, fifelse(TRUE, list(data.table(1:5)), list(data.table(5:1))), list(data.table(1:5)))
```
jangorecki: I think it is in-proper use. data.frame should be handled as ordinary list here, thus wrapping into `list` to achieve equality. Consider
```r
fifelse(TRUE, data.table(1:5, 2:6), data.table(5:1, 6:2))
```

src/fifelse.c
```diff
@@ -50,10 +34,9 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b)
     UNPROTECT(2);
   }
 
-  // Check here the length of the different input variables.
   if (len1!=1 && len1!=len0) error("Length of 'yes' is %lld but must be 1 or length of 'test' (%lld).", len1, len0);
   if (len2!=1 && len2!=len0) error("Length of 'no' is %lld but must be 1 or length of 'test' (%lld).", len2, len0);
-  const int64_t amask = len1>1 ? INT64_MAX : 0;
+  const int64_t amask = len1>1 ? INT64_MAX : 0; // for scalar 'a' bitwise AND will reset iterator to first element: pa[i & amask] -> pa[0]
```
MichaelChirico: oh i see now. helpful comment indeed!

and the thinking is the compiler will optimize away all the repeated `&` calls?

jangorecki: bitwise AND of `any_number & 0` will result `0`, while of `any_number & max_number` will result `any_number`, due to the fact that `max_number` is represented by all `1`, i.e. `1 & 15` -> `0001 & 1111` results `0001`.

2005m:
Thank you Jan, I will have a look at the changes. What about coercing logical, complex and string (like ifelse) instead of just int to double? I can look at it and suggest something if you want?
I can try to suggest something for en-listing as well.

jangorecki:
@2005m I prefer to have strict type checks and require user to provide correct ones, int-double is just an exception because it is so widely used, and make sense, while coercing numeric to character is rather useless.

man/fifelse.Rd
```diff
@@ -1,88 +1,34 @@
 \name{fifelse}
 \alias{fifelse}
+\alias{ifelse}
 \title{ Fast ifelse }
 \description{
-\code{data.table::fifelse} is comparable to \code{base::ifelse}, \code{dplyr::if_else} and  \code{hutils::if_else}.
-Like these functions, it returns a value with the same length as \code{test} and filled with value from \code{yes} or \code{no}.
-This function is mostly written in C for speed and unlike \code{base::ifelse} the output type is consistent with those of \code{yes} and \code{no}.
+\code{fifelse} is a faster and more robust replacement of \code{\link[base]{ifelse}}. It returns a value with the same length as \code{test} filled with corresponding values from \code{yes}, \code{no} or eventually \code{NA}, depending on \code{test}. Supports \code{bit64}'s \code{integer64} and \code{nanotime} classes.
```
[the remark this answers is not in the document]
jangorecki: will do

PR #3742, opened 2019-08-03
manual update, fixes to pass `--run-dontrun`

jangorecki:
I would generally avoid posting timings in examples, despite putting into `dontrun`. As we can see, timings changes heavily over time, a 10 folds in below examples. It then make sense to keep them in NEWS.md where entries are bounds to a point in time. Moreover timings are highly affected by configuration, like number of threads, disk speed. It is not very practice to flesh out all such details in examples.

### DIFF
--- a/man/fread.Rd
+++ b/man/fread.Rd
@@ -248,7 +248,7 @@ table(sapply(DT,class))
 fread("http://www.stats.ox.ac.uk/pub/datasets/csb/ch11b.dat")
 
 # Decompresses .gz and .bz2 automatically :
-fread("http://stat-computing.org/dataexpo/2009/2008.csv.bz2")
+fread("http://stat-computing.org/dataexpo/2009/1987.csv.bz2")
 
 }
 
--- a/man/fread.Rd
+++ b/man/fread.Rd
@@ -229,36 +229,22 @@ cat("File size (MB):", round(file.info("test.csv")$size/1024^2),"\n")
 # 50 MB (1e6 rows x 6 columns)
 
 system.time(DF1 <-read.csv("test.csv",stringsAsFactors=FALSE))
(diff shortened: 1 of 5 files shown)


man/test.data.table.Rd
```diff
@@ -24,11 +24,4 @@ test.data.table(verbose=FALSE, pkg="pkg", silent=FALSE,
 When \code{silent} equals to \code{TRUE} it will return \code{TRUE} if all tests were successful. \code{FALSE} otherwise. If \code{silent} equals to \code{FALSE} it will return \code{TRUE} if all tests were successful. Error otherwise.
 }
 \seealso{ \code{\link{data.table}} }
-\examples{
-\dontrun{
-  library(data.table)
-  test.data.table()
```
jangorecki: this was causing http://r.789695.n4.nabble.com/problematic-dev-off-after-running-examples-without-plots-td4753880.html

man/fread.Rd
```diff
@@ -163,53 +229,39 @@ cat("File size (MB):", round(file.info("test.csv")$size/1024^2),"\n")
 # 50 MB (1e6 rows x 6 columns)
 
 system.time(DF1 <-read.csv("test.csv",stringsAsFactors=FALSE))
-# 60 sec (first time in fresh R session)
+# 5.4 sec (first time in fresh R session)
```
jangorecki: I would generally avoid posting timings in examples, despite putting into `dontrun`. As we can see, timings changes heavily over time, a 10 folds in below examples. It then make sense to keep them in NEWS.md where entries are bounds to a point in time. Moreover timings are highly affected by configuration, like number of threads, disk speed. It is not very practice to fleshout all such details in examples.

Issue #3739, opened 2019-08-02
uniqueN() is very slow compared to length(unique())


shrektan:
I'm using the lastest dev version of `data.table`. Still `uniqueN()` is an order of magnitude slower than `length(unique())` . So slow I think it should be tagged as a bug ... See the reprex example below.

``` r
library(data.table)
set.seed(1000)
mk_rd_words <- function(min = 4, max = 20) {
  n <- floor(runif(1, min, max))
  paste0(sample(c(letters, LETTERS), size = n, replace = TRUE), collapse = '')
}
words <- vapply(1:1000, function(x) mk_rd_words(4, 50), FUN.VALUE = 'a')

n <- 1e4
tbl <- data.table(
  a = sample(words, size = n, replace = TRUE),
  b = sample(words, size = n, replace = TRUE)
)

bench::mark(
  tbl[, .(N = uniqueN(b)), keyby = a],
  tbl[, .(N = length(unique(b))), keyby = a]
)
#> # A tibble: 2 x 6
#>   expression                                      min   median `itr/sec`
#>   <bch:expr>                                 <bch:tm> <bch:tm>     <dbl>
#> 1 tbl[, .(N = uniqueN(b)), keyby = a]        169.49ms  169.7ms      5.90
#> 2 tbl[, .(N = length(unique(b))), keyby = a]   8.57ms   9.38ms    105.  
#> # ... with 2 more variables: mem_alloc <bch:byt>, `gc/sec` <dbl>
```

<sup>Created on 2019-08-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

shrektan:
Well, I see. For very large vectors, `uniqueN()` is faster. The problem is if `uniqueN()` is supposed to be called millions of times (e.g. there're many subgroups), the performance's slowing down is significant. 

Maybe for small atomic vector, we just use length(unique()) instead?


``` r
library(data.table)
set.seed(1000)
# small vector, uniqueN() is slower
x <- sample(1:1e5, size = 1e2, replace = TRUE)
microbenchmark::microbenchmark(
  uniqueN(x),
  length(unique(x))
)
#> Unit: microseconds
#>               expr    min      lq     mean  median     uq      max neval
#>         uniqueN(x) 20.744 21.3460 32.05373 21.7050 22.226 1010.481   100
#>  length(unique(x))  4.757  5.3565  6.18967  5.9215  6.698   18.199   100
# big vector, uniqueN() is faster
[...]

Issue #3725, opened 2019-07-24
uniqueN could be GForce optimised + GForce could be optimised for := too.


phileas-condemine:
This issue follows a discussion during useR!2019 after the presentation of data.table by Arun @arunsrinivasan 

Hello,
Thanks for the amazing job, I love data.table !
I am using uniqueN to verify l-diversity for anonymization purposes.
The data I am working with is around 30M rows, easily ingested by data.table.
Unfortunately uniqueN is not as fast as other functions.
I tried to parallelize the grouping using setDTthreads as I can go up to 16 on my rstudio server instance. 
First I get a benchmark using a simple sum over numeric.
![simple_sum_DT](https://github.com/phileas-condemine/repex_slow_uniqueN/raw/master/perf_DT_basic_sum_grouping_parallelized.png)

Then I do basically the same thing but apply uniqueN over character (factor would give the same results).

![uniqueN_DT](https://github.com/phileas-condemine/repex_slow_uniqueN/raw/master/perf_DT_uniqueN_grouping_parallelized.png)

Here is the code for a repex https://github.com/phileas-condemine/repex_slow_uniqueN/blob/master/repex_slow_uniqueN.R

Additional info : 
```
> library(data.table)
data.table 1.12.2 using 8 threads (see ?getDTthreads).
```

Here is my session_info()

```
> sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-pc-linux-gnu (64-bit)

Matrix products: default
BLAS: /opt/microsoft/ropen/3.5.1/lib64/R/lib/libRblas.so
LAPACK: /opt/microsoft/ropen/3.5.1/lib64/R/lib/libRlapack.so

locale:
 [1] LC_CTYPE=fr_FR.UTF-8       LC_NUMERIC=C               LC_TIME=fr_FR.UTF-8       
 [4] LC_COLLATE=fr_FR.UTF-8     LC_MONETARY=fr_FR.UTF-8    LC_MESSAGES=fr_FR.UTF-8   
 [7] LC_PAPER=fr_FR.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=fr_FR.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[...]

Issue #3438, opened 2019-03-01
Rerun repeated uniqueN test


mattdowle:
Was reported here : https://github.com/Rdatatable/data.table/issues/3395#issuecomment-464436815
I said I'd follow up here: https://github.com/Rdatatable/data.table/pull/3435#issuecomment-468415545
Double check all those results under the new default.

jangorecki:
Up to date timings below. There is still a problem in uniqueN by group.
We basically need what @mattdowle described in https://github.com/Rdatatable/data.table/pull/3743#issuecomment-525023671 and #1120 would also be nice.
```r
library(data.table)
N_X = 1e6
n_day = 60
n_clientid = 1e5
n_Platform = 7
X = data.table(
  day = sample(1:n_day, N_X, TRUE),
  clientid = as.character(sample(1:n_clientid, N_X, TRUE)),
  Platform = as.character(sample(1:n_Platform, N_X, TRUE))
)

setDTthreads(NULL) # default
getDTthreads()
#[1] 10
system.time(
  X[, .(x = uniqueN(day) - 1L,
        first_active_day = min(day),
        last_active_day = max(day)),
    by = .(Platform, clientid)]
)
#   user  system elapsed 
#172.084   1.640  18.179 

setDTthreads(1)
getDTthreads()
#[1] 1
system.time(
  X[, .(x = uniqueN(day) - 1L,
        first_active_day = min(day),
        last_active_day = max(day)),
    by = .(Platform, clientid)]
)
#   user  system elapsed 
# 10.608   0.596  11.206 

setDTthreads(0)
getDTthreads()
#[1] 20
system.time(
  X[, .(x = uniqueN(day) - 1L,
        first_active_day = min(day),
        last_active_day = max(day)),
    by = .(Platform, clientid)]
)
#still computing
```

mattdowle:
Closed by #4484. See benchmark near the top of the top comment here: https://github.com/Rdatatable/data.table/pull/4484#issue-422469250

Issue #1120, opened 2015-04-16
uniqueN slower than length(unique())


jangorecki:
Most recent data.table, R 3.1.3.
Described in: http://stackoverflow.com/a/29684533/2490497

ben519:
Came looking for this.  I run into this issue a lot - my recent case being unbearably slow.  My case looks more like this

```
dt <- data.table(
  A=sample(100000, 1000000, replace=TRUE), 
  B=sample(100000, 1000000, replace=TRUE), 
  C=sample(1000000, 1000000, replace=TRUE)
)

# slow
> system.time(result <- dt[, list(UniqueCs=uniqueN(C)), keyby=list(A, B)])
   user  system elapsed 
 11.409   0.031  11.442 

# fast
> system.time(result <- dt[, list(UniqueCs=.N), keyby=list(A, B, C)])
   user  system elapsed 
  0.081   0.005   0.084
```
I'd think `uniqueN` should take about as long as aggregating with its argument.

MichaelChirico:
Confirming timings of @ben519... 

Ran on 1.9.6:

```
system.time(result <- dt[, list(UniqueCs=uniqueN(C)), keyby=list(A, B)])
#    user  system elapsed 
#   8.032   0.004   8.029 
system.time(result <- dt[, list(UniqueCs=.N), keyby=list(A, B, C)])
#    user  system elapsed 
#   0.496   0.004   0.498 
```

Ran on 1.9.7:

```
system.time(result <- dt[, list(UniqueCs=uniqueN(C)), keyby=list(A, B)])
#    user  system elapsed 
#  11.764   0.488   9.706 
system.time(result <- dt[, list(UniqueCs=.N), keyby=list(A, B, C)])
#    user  system elapsed 
#   0.100   0.008   0.109 
```

(I missed his edit, but the difference is marginal)

MichaelChirico:
Update if improved:

https://github.com/ben519/mltools/pull/10

MichaelChirico:
Ideal case where `uniqueN` is much faster than alternatives (list i.e. non-scalar input here)

https://stackoverflow.com/a/53890905/3576984

sindribaldur:
When used with with an `by` argument and many groups `uniqueN()` slowness very bad:

```
irisdt <- setDT(iris[sample(1:150, size = 10000, replace = TRUE), ])
irisdt[, Sepal.Width := Sepal.Width+ sample(0:50, size = 10000, replace = TRUE)]
[...]

PR #3743, opened 2019-08-03
uniqueN escape forder to base R for small atomic vectors, #1120

jangorecki:
this address some of the slowest uses cases of uniqueN by make groups
related #3739, #1120, #3725, #3438

### DIFF
--- a/R/duplicated.R
+++ b/R/duplicated.R
@@ -140,14 +140,25 @@ uniqueN = function(x, by = if (is.list(x)) seq_along(x) else NULL, na.rm=FALSE) 
     by = key(x)
     stop(error_oldUniqueByKey)
   }
+  #verbose = getOption("datatable.verbose")
   if (is.null(x)) return(0L)
   if (!is.atomic(x) && !is.data.frame(x))
     stop("x must be an atomic vector or data.frames/data.tables")
   if (is.atomic(x)) {
-    if (is.logical(x)) return(.Call(CuniqueNlogical, x, na.rm=na.rm))
+    if (is.logical(x)) {
+    #  if (verbose) cat("uniqueN detected logical type, using optimised C uniqueNlogical\n")
+      return(.Call(CuniqueNlogical, x, na.rm=na.rm))
+    }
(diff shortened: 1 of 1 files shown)


R/duplicated.R
```diff
@@ -140,14 +140,25 @@ uniqueN = function(x, by = if (is.list(x)) seq_along(x) else NULL, na.rm=FALSE)
...
-    if (is.logical(x)) return(.Call(CuniqueNlogical, x, na.rm=na.rm))
+    if (is.logical(x)) {
+    #  if (verbose) cat("uniqueN detected logical type, using optimised C uniqueNlogical\n")
+      return(.Call(CuniqueNlogical, x, na.rm=na.rm))
+    }
+    escape_n = 1e3L #getOption("datatable.escapeUniqueN",1e7L)
+    if (length(x) <= escape_n) {
+      #if (verbose) cat("uniqueN detected atomic type of length ", length(x), " which is small enough to use base R ", if (na.rm) "na.omit+" else "", "duplicated\n",sep="")
+      if (na.rm) x = na.omit(x)
+      return(length(x)-sum(duplicated(x)))
```
HughParsonage: Probably slightly quicker to not `na.omit` but just do:

```r
return(length(x) - sum(duplicated(x)) - {na.rm && anyNA(x)})
```

jangorecki: I am not getting how that would work

HughParsonage: If `na.rm = FALSE` then my version is just `length(x) - sum(duplicated(x)) - 0`. 

If `na.rm = TRUE` then it will be `length(x) - sum(duplicated(x)) - 1 * anyNA(x)`.

MichaelChirico: (key insight being that `uniqueN` can differ by at most 1 for `na.rm` vs `!na.rm`)

jangorecki: nice one!
it seems that it might eventually differ by 2
```
> x=c(1,2,NA,2,3,NaN,NaN,NA)
> uniqueN(x)
[1] 5
> uniqueN(x, na.rm=T)
[1] 3
> uniqueN(na.omit(x))
[1] 3
```

Issue #3654, opened 2019-06-20
column types not inferred for rbind. 


ocschwar:
Here's a simple test case: 
```
library(data.table)
library(circular)
X <- data.table(Time='2013-01-01 05:00:00', speed=5.0, direction = circular::circular(22))
Y <- data.table(Time='2013-01-0117:00:00', speed=4.5, direction = 172)
rbind(X,Y)
```

The manual page for data.table::rbindlist continues to say this: 

     If column ‘i’ does not have the same type in each of the list
     items; e.g, the column is ‘integer’ in item 1 while others are
     ‘numeric’, they are coerced to the highest type.

But that is no longer the behavior I see.  Could type coercion be done at least optionally? 

`> sessionInfo()
R version 3.4.4 (2018-03-15)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.2 LTS

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.2.20.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] graphics  grDevices datasets  utils     stats     methods   base     

other attached packages:
 [1] REhedge_0.2.42    REdata_0.0.6      REveal_1.10       REviz_0.1        
 [5] circular_0.4-93   bindrcpp_0.2.2    data.table_1.12.2 maps_3.3.0       
 [9] odbc_1.1.6        devtools_1.13.6   testthat_2.1.1   

loaded via a namespace (and not attached):
  [1] readxl_1.2.0        backports_1.1.2     spam_2.1-4         
  [4] sm_2.2-5.6          plyr_1.8.4          lazyeval_0.2.1     
  [7] sp_1.3-1            rclipboard_0.1      crosstalk_1.0.0    
[...]

PR #3765, opened 2019-08-13
coerceClass

### LINKED ISSUE #3913: class-aware C coercion function, no-alloc coerce where possible
We should have a function for coercing an object to another type and class, in a efficient way, and class aware (for double type, different classes uses different NA representation so we need to handle that).
It is important because coercion is a very common operation, especially `num-int` or `int-num` for user convenience so they don't have to type `L` suffix, but there are tons of other applications.

Initial work on that can be found in https://github.com/Rdatatable/data.table/pull/3765 (generic coerceClass)

No-alloc coerce written for C assign function https://github.com/Rdatatable/data.table/pull/3909 is good foundation, that potentially could be extracted out to an standalone function.

jangorecki:
closes #3913

Internally used in `nafill` function `coerceFill` has been heavily modified, and renamed to `coerceClass`. Main purpose of the function is to handle coercion between types/classes with care on int64.
Comparing to previous `coerceFill` version, new `coerceClass`:
- handle coerce of vector, not only scalar
- handle coercion from any class to any class
- int64 coercion is taken care for integer and double only
- coercion between int / double / int64 is handled in the function, otherwise R's `coerceVector` is used - as a result, consider char to int64 (`coerceClass("1", list(integer(0), bit64::as.integer64(0)))`). This is going to be addressed in this PR, placeholders are in place. Also there might be some missing warnings like (NA introduced by coercion, Int overflow), which are currently not checked/issued in our int / double / int64 handling.

utils.c tests moved into own test script

----

motivation:
- extend `nafill` for types other than integer, numeric and int64
- to be re-used as efficient way to handle conversions like
```r
data.table(a=1:2, b=bit64::as.integer64(1:2))[2L, c("a","b") := NA][]
#       a                   b
#   <int>               <i64>
#1:     1                   1
#2:    NA 9218868437227407266
data.table(a=1:2, b=bit64::as.integer64(1:2))[2L, c("a","b") := coerceClass(NA, .SD)][]
#   <int> <i64>
#1:     1     1
#2:    NA  <NA>
```
in `assign`, `rbindlist` and potentially other places where we need to take care of int64 NAs handling or other classes like #3654

----

As of current moment I don't understand coverage issue, it reports few closing brackets `}` that are not being covered. Maybe a matter of waiting for update.

### DIFF
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -8,4 +8,5 @@ setcoalesce = function(...) .Call(Ccoalesce, list(...), TRUE)
 fifelse = function(test, yes, no) .Call(CfifelseR,test, yes, no)
 
 colnamesInt = function(x, cols, check_dups=FALSE) .Call(CcolnamesInt, x, cols, check_dups)
-coerceFill = function(x) .Call(CcoerceFillR, x)
+coerceClass = function(x, out) .Call(CcoerceClassR, x, out)
+coerceFill = function(x) coerceClass(x, list(0L, 0, structure(0, class="integer64"))) # temporarily till new tests in place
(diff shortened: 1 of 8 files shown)


jangorecki:
This PR should also take care of https://github.com/Rdatatable/data.table/issues/3913

jangorecki:
superseded by #4491

Issue #3766, opened 2019-08-13
Incorrect setkey result


kirillmayantsev:
Hello!

**Minimal reproducible example**:
```
library(data.table)
x = data.table(a = 1, b = c(1, 4, 2, 3), c = c(4, 3, 2, 1))
y = x[a == 1, .(d = paste(b, c, sep = "_"), b, b2 = b, c)]
y
setkey(y, d)
y

> y
     d b b2 c
1: 1_4 1  1 4
2: 4_3 4  4 3
3: 2_2 2  2 2
4: 3_1 3  3 1
> setkey(y, d)
> y
     d b b2 c
1: 1_4 1  1 4
2: 2_2 3  3 2
3: 3_1 4  4 1
4: 4_3 2  2 3
```

After setting a key the order of "b" and "b2" columns is incorrect.
There are 3 "workarounds" which help to keep "y" data.table consistent:
1. Remove filtration ("a == 1")
2. Remove column duplication ("b2 = b")
3. Assign to "b2" a copy of "b" (as in example "b2" has the same reference as "b" I guess)

But still the described behavior seems to be a bug.
This issue could be related to https://github.com/Rdatatable/data.table/issues/3496

**Output of sessionInfo()**:
```
R version 3.5.1 (2018-07-02)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7600)

Matrix products: default

locale:
[1] LC_COLLATE=Russian_Russia.1251  LC_CTYPE=Russian_Russia.1251    LC_MONETARY=Russian_Russia.1251
[4] LC_NUMERIC=C                    LC_TIME=Russian_Russia.1251    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.12.3

loaded via a namespace (and not attached):
[1] compiler_3.5.1 tools_3.5.1    yaml_2.2.0    
```

MichaelChirico:
I think the problem is `b` and `b2` are the same memory:

```
x = data.table(a = 1, b = c(1, 4, 2, 3), c = c(4, 3, 2, 1))
y = x[a == 1, .(d = paste(b, c, sep = "_"), b, b2 = b, c)]

identical(address(y$b), address(y$b))
# [1] TRUE
```

So when sorting happens, it sorts `b` twice.

I'm not sure if this is expected behavior, or a bug in `[`, or a bug in `setkey`.

[...]

Issue #2162, opened 2017-05-13
setkey on cross-referenced data.tables produces inconsistent results


renkun-ken:
The following code produces inconsistent results:

```r
library(data.table)

d1 <- data.table(x = 1:6, y = 6:1)
d2 <- list(x = d1$x, y = d1$y, x2 = d1$x)
setDT(d2)

str(d1)
str(d2)

str(lapply(d1, address))
str(lapply(d2, address))

setkey(d2, y)
str(d1)
str(d2)
```

The output is as follows:

```
> library(data.table)
> d1 <- data.table(x = 1:6, y = 6:1)
> d2 <- list(x = d1$x, y = d1$y, x2 = d1$x)
> setDT(d2)
> str(d1)
Classes ‘data.table’ and 'data.frame':	6 obs. of  2 variables:
 $ x: int  1 2 3 4 5 6
 $ y: int  6 5 4 3 2 1
 - attr(*, ".internal.selfref")=<externalptr> 
> str(d2)
Classes ‘data.table’ and 'data.frame':	6 obs. of  3 variables:
 $ x : int  1 2 3 4 5 6
 $ y : int  6 5 4 3 2 1
 $ x2: int  1 2 3 4 5 6
 - attr(*, ".internal.selfref")=<externalptr> 
> str(lapply(d1, address))
List of 2
 $ x: chr "0x6eaa310"
 $ y: chr "0x6eaa358"
> str(lapply(d2, address))
List of 3
 $ x : chr "0x6eaa310"
 $ y : chr "0x6eaa358"
 $ x2: chr "0x6eaa310"
> setkey(d2, y)
> str(d1)
Classes ‘data.table’ and 'data.frame':	6 obs. of  2 variables:
 $ x: int  1 2 3 4 5 6
 $ y: int  1 2 3 4 5 6
 - attr(*, ".internal.selfref")=<externalptr> 
> str(d2)
Classes ‘data.table’ and 'data.frame':	6 obs. of  3 variables:
 $ x : int  1 2 3 4 5 6
 $ y : int  1 2 3 4 5 6
 $ x2: int  1 2 3 4 5 6
 - attr(*, ".internal.selfref")=<externalptr> 
 - attr(*, "sorted")= chr "y"
```

[...]

PR #3768, opened 2019-08-15
Closes #3496 and #3766 -- detect & block setkey from double-reordering identical columns

### LINKED ISSUE #3496: error using function setkey
I know how to avoid this error, but I donot know why this error happen if I run it in this way.
datain<-data.frame(chrom=c("chr17","chr4","chr5","chr13"),map=c(81061047,106061533,40102442,73791553),rs=c("rs75954926","rs7679673","rs7708610","rs78341008"),start=c(79061048,104061534,38102443,71791554),end=c(83061048,108061534,42102443,75791554))
datain
 datain$chr<-datain$chrom
 setDT(datain)
setkey(datain, chr, start, end)
datain

Thanks!

Alex

MichaelChirico:
#3496 and #3766 

I'm not sure the change to `data.table.R` is necessary -- seems the exact-copies coming out of `j` haven't caused any other issues (unless there's something else in the issue tracker waiting for us?), and the change to `setkeyv` alone fixes both issues. But it does feel odd to me to allow this. Thoughts @mattdowle / @jangorecki ?


MichaelChirico:
Wasn't sure about the changes to `data.table.R`, but incidentally, they fix #2245. So I'll leave those changes & add a test to close that.

MichaelChirico:
OK. Would have to sit back down with this to figure which part exactly is needed to solve #2245...

renkun-ken:
I notice that I raised #2162 long time ago which is a similar but different case. I guess it's impossible to also address that at the moment?

PR #3770, opened 2019-08-16
Closes #3760 -- prevent setDT for data.frame w matrix columns

### LINKED ISSUE #3760: setDT fails to detect invalid columns -> obscure error downstream
```
DF = data.frame(a = 1:5)
DF$m = matrix(6:15, ncol = 2L)
setDT(DF)
print(DF)
```

> Error in dimnames(x) <- dn : 
  length of 'dimnames' [1] not equal to array extent
Error during wrapup: 'length(x) = 2 > 1' in coercion to 'logical(1)'

MichaelChirico:
#3760 

Not sure this is an ideal solution, but it works & indeed `as.data.table` is the appropriate function for such objects.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -333,6 +333,8 @@
 
 19. Subsetting does a better job of catching a malformed `data.table` with error rather than segfault. A column may not be NULL, nor may a column be an object which has columns (such as a `data.frame` or `matrix`). Thanks to a comment and reproducible example in [#3369](https://github.com/Rdatatable/data.table/issues/3369) from Drew Abbot which demonstrated the issue which arose from parsing JSON. The next release will enable `as.data.table` to unpack columns which are `data.frame` to support this use case.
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
@jangorecki can you think of any other types of validations we'd want to do upfront here? if so maybe best to have a `validate_dt` function we can build up starting from here.

MichaelChirico:
looking for `NULL` `dim` failed test 1980, which has an array with `length(dim(x)) == 1L` that ends up being `setkey`'d (confirmed that in this case, `setDT` wipes out the `dim`). Played around with changing `as.data.table.list` to route this case through `as.data.table` but we'd have to route through `as.data.table.array`, which fails because there's a check that `length(dim(x)) >= 3` there.

If we change this to route this case to `as.data.table.matrix` instead of erroring, the solution I had wound up with the wrong name: 

```
y = as.array(1:5)
names(data.table(y))
# [1] 'x'
```

So I gave up on that route; the one here does fine. Just recording my thoughts for posterity.

PR #3774, opened 2019-08-18
Closes #1926 -- unlock shallow copy of .SD when needed in joins in j

### LINKED ISSUE #1926: .SD is locked for DT[, DT2[.SD]] joins
Based on [an SO question](http://stackoverflow.com/q/40772423/), I was trying something like...

    library(data.table)
    DT = data.table(id = 1:2, v = 3:4)
    DT2 = data.table(id = 1, x = 5)
    DT[id == 1, DT2[.SD, on="id"]]

> Error in set(i, j = lc, value = newval) : 
>   .SD is locked. Updating .SD by reference using := or set are reserved for future use. Use := in j directly. Or use copy(.SD) as a (slow) last resort, until shallow() is exported.

I wasn't expecting to see this error since I'm not using `set`.

MichaelChirico:
#1926 

@franknarf1 I went through a few of the examples you posted and the fix here seems to work, it would be great if you could try and stress test it a bit anyway to make sure all the bases are covered 🙇

@jangorecki / @mattdowle do we need to re-set the locked attr `on.exit()` for the shallow copies as well?

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15671,6 +15671,15 @@ if (test_bit64) {
   test(2077.06, int64_int32_match(d[, sum(i32, na.rm=TRUE), g], d[, sum(i64, na.rm=TRUE), g]))
 }
 
+# #1926 -- unlock .SD for bmerge in j
+DT = data.table(id=1:2, v=3:4)
+DT2 = data.table(id=1, x=5)
+## DT2.i is numeric so bmerge does coercion with set()
+test(2078.1, DT[id == 1, DT2[.SD, on="id"]], data.table(id=1L, x=5, v=3L))
+df1 = data.table(a=1:5, b=c(0, 0, 1, 0, 2))
+df2 = data.table(c=c(1, 1, 2, 2, 3), d=c(3, 4, 3, 5, 4))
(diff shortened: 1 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -15671,6 +15671,15 @@ if (test_bit64) {
   test(2077.06, int64_int32_match(d[, sum(i32, na.rm=TRUE), g], d[, sum(i64, na.rm=TRUE), g]))
 }
 
+# #1926 -- unlock .SD for bmerge in j
+DT = data.table(id=1:2, v=3:4)
+DT2 = data.table(id=1, x=5)
+## DT2.i is numeric so bmerge does coercion with set()
+test(2078.1, DT[id == 1, DT2[.SD, on="id"]], data.table(id=1L, x=5, v=3L))
```
jangorecki: is there a tests that verify that `i` data.table was not altered after join?

MichaelChirico: I would assume so but it's a good idea to add one explicitly here.

jangorecki: @MichaelChirico still missing this test, isn't it?

jangorecki:
We should definitely get this in 1.12.4 to clean up new vignette. @MichaelChirico

PR #3782, opened 2019-08-20
simplify hour() etc for ITime

### LINKED ISSUE #3158: Faster second, minute and hour for objects of class ITime
The functions `second`, `minute` and `hour` currently have a faster method, by bypassing `as.POSIXlt()`, for POSIXct objectss under `UTC`. See [here](https://github.com/Rdatatable/data.table/blob/master/R/IDateTime.R#L275#L298).

The faster logic needs to be also included for `ITime` objects. The fix is quite simple. The if statement in all these functions need an additional `|| inherits(x, "ITime")`

Here's a quick benchmark:

```r
# Current version
second  <- function(x) {
  if (inherits(x,'POSIXct') && identical(attr(x,'tzone'),'UTC')) {
    # if we know the object is in UTC, can calculate the hour much faster
    as.integer(x) %% 60L
  } else {
    as.integer(as.POSIXlt(x)$sec)
  }
}

# new version
fSecond<- function(x) {
  if (inherits(x, "ITime") || (inherits(x,'POSIXct') && identical(attr(x,'tzone'),'UTC'))) {
    # if we know the object is in UTC, can calculate the hour much faster
    as.integer(x) %% 60L
  } else {
    as.integer(as.POSIXlt(x)$sec)
  }
}

require(data.table)
x <- setattr(0:86399, "class", "ITime")
y <- sample(x, 50e7, TRUE)

system.time(a1 <- second(y))
#    user  system elapsed 
#   29.00   10.72   39.78 
system.time(a2 <- fSecond(y))
#    user  system elapsed 
#    3.51    0.56    4.06 
 
identical(a1, a2)
# [1] TRUE
```

MichaelChirico:
Closes #3158

Will need to be re-worked after #3781


R/IDateTime.R
```diff
@@ -302,24 +302,27 @@ as.POSIXlt.ITime = function(x, ...) {
 #   lubridate routines do not return integer values.
 ###################################################################
 
+do_fastroutine = function(x)
```
jangorecki: please another name, `do_fastroutine` can be anything. `tz_insensitive`, or `utc_compatible`, or anything that refer to function of it

R/IDateTime.R
```diff
@@ -302,24 +302,27 @@ as.POSIXlt.ITime = function(x, ...) {
...
+  (inherits(x, 'POSIXct') && identical(attr(x, 'tzone', exact=TRUE), 'UTC')) ||
+  inherits(x, 'ITime')
 second  = function(x) {
-  if (inherits(x, 'POSIXct') && identical(attr(x, 'tzone', exact=TRUE), 'UTC')) {
+  if (do_fastroutine(x)) {
     # if we know the object is in UTC, can calculate the hour much faster
     as.integer(x) %% 60L
```
jangorecki: does it make sense to change this like into `as.integer(x %% 60)` to prevent overflow?
```
as.integer(5e9) %% 2L
[1] NA
Warning message:
NAs introduced by coercion to integer range 
```
or maybe not worth...
if `%%` is much faster on integer than double then I don't think it makes much sense as issues like such overflow is quite unlikely

R/IDateTime.R
MichaelChirico: Waiting for #3781 (hence WIP label)... will probably drop this helper after that's merged (since we can definitely `as.integer` on `ITime` but not `POSIXct` case)

R/IDateTime.R
MichaelChirico: #3781

PR #3796, opened 2019-08-28
new arg to gsum to allow gmean to ignore overflow warning

### LINKED ISSUE #986: Unnecessary warning from gmean about sum greater than integer can hold
See at 3:01 into this video :
https://www.youtube.com/watch?v=AmE4LXPQErM

MichaelChirico:
Closes #986 

Seems to me this is the way to go, the other ways (e.g. rewrite `gmean` without using `gsum` are pretty prohibitive for kind of edge case)


HughParsonage:
Recommend removing the warning altogether. Can't think of circumstances where it would be necessary or useful.

PR #3800, opened 2019-08-29
detect and redirect ifelse usage in j to fifelse

### LINKED ISSUE #2677: warn or optimize usage of `ifelse()`
Not sure the status of `ifelse()` speed in R these days.  Years ago it used to be really slow and I avoided it.  Is that still the case?
I've seen this idiom quite a bit in the wild : 

```R
DT[, newCol := ifelse(existingCol == 42,  1L, 2L)]
```

under `verbose=TRUE` there's a message about the plonk.  It's slow and RAM inefficient I assume (haven't tested).

Better would be using i, as per #2676 :  

```R
DT[existingCol==42, newCol:=1L, fill=2L]
```

Nested `ifelse(ifelse(ifelse()))` also quite common.  Either detect and warn about `ifelse()` usage and point to a document about how to avoid, or optimize it automatically.

MichaelChirico:
Closes #2677 

Simply copying the approach used for `strptime` to accomplish this.

I think the `ifelse` --> `if else` redirection is more generally useful and considered filing a patch to `base` but it won't be a simple drop-in because `ifelse` does the "wrong" thing with attributes (one of the reasons we like `fifelse` in the first place) so it would need to be done more carefully to maintain backwards compatibility.

I don't think we have to worry about that for usage in `j`.


jangorecki:
Why not compute both `ifelse` and `fifelse` and warn when both does not match(asking for report to our issues)? then for future version we can change that to compute only `fifelse`.
Making change like as it is now is pretty risky, maybe using an option, that would be disabled by default now.

MichaelChirico:
well, we do know a lot of cases when they won't be `identical` -- e.g. for `yes/no` `Date` objects, and I think we want to assert `fifelse` is superior in such case.

Is there a way to compare the objects and only raise `ifelse` result if the objects are _unexpectedly_ different?

jangorecki:
OK, so if we want to have it in 1.12.4 then best to use this with option by default disabled.

MichaelChirico:
I'm not so sure... escape valve to `base::ifelse` is still there. We never had such option about `strptime`... OTOH you're right that this will almost certainly break a bunch of revdeps. It might be good to leave this in and see just how widespread `ifelse`-in-`j` usage is "in the wild", would be a nice thing to note and proselytize on...

MichaelChirico:
I'll essentially be MIA until mid-late next week. Good point re: potential issues w/ large cardinality grouping x openMP start-up, we should benchmark.

MichaelChirico:
OK found some time this morning.

The reason to use `warning` is that even though we do have a bunch of tests & `ifelse` is conceptually simple, I worry about edge cases with factor/non-atomic/S4 outputs where the switch to `fifelse` might break unanticipated. I think a `warning` popping up in those cases is best... optimization level sounds good, but I would use level 3 -- shutting off `GForce` because of an `ifelse` conflict seems harsh.

Will make that change & then investigate the grouping bit.

MichaelChirico:
Have run this benchmark:

```
library(microbenchmark)
library(data.table)
set.seed(2384)
NN = 1e7
DT = data.table(u = runif(NN), yes = rnorm(NN), no = rnorm(NN))

runs = CJ(
  n_grps = 10^(0:6),
  bool_pct = seq(0, 1, length.out = 5L)
)
runs[ , run_id := .I]

performance = runs[ , {
  print(.BY$run_id)
  DT[ , c('grp', 'bool') := .(sample(n_grps, .N, TRUE), u > bool_pct)]
  summary(microbenchmark(
    times = 10L,
    base = DT[ , base::ifelse(bool, yes, no), by = grp],
    DT = DT[ , fifelse(bool, yes, no), by = grp]
  ), unit='s')
}, by = .(run_id, n_grps, bool_pct)]

performance[ , expr := as.character(expr)]

performance[ , {
  par(mar = c(10.1, 5.1, 4.1, 2.1))
  print(.SD[ , .(median, paste(n_grps, bool_pct, expr, sep = ' / '))])
  barplot(median, names.arg = paste(n_grps, bool_pct, expr, sep = ' / '),
          col = c('red', 'blue'),
         xlab = '', las = 2L, ylab = 'Median evaluation time (s)', log = 'y')
  mtext(side = 1L, line = 8, '# Groups / % False / Package')
  NULL
}]
```

I _really_ don't know what's going on with base plotting (the labels appear to be all-wrong) but `red` is `base` and `blue` is `data.table`:

Screenshot 2019-09-01 at 12 54 06 AM

Note the log scale.

While there may or may not be a per-group cost for `fifelse`, it seems to be all the worse for `base::ifelse`

MichaelChirico:
@mattdowle OK I think this is good to go

jangorecki:
Log scale is very non-intuitive to compare those timings.
It doesn't look safe to merge.
```r
library(data.table) # default 20 threads
set.seed(108)
sample_dups = function(N, ratio) {
  unq = sample(N, max(N*(1-ratio), 1L))
  sample(c(unq, sample(unq, N*ratio)))
}
NN = 1e7
DT = data.table(test=sample(c(TRUE,FALSE), NN, TRUE), yes=rnorm(NN), no=rnorm(NN), grp=sample_dups(NN, 0.1))

system.time(DT[, base::ifelse(test, yes, no), by=grp])
#   user  system elapsed 
# 66.619   0.687  63.239
system.time(DT[, data.table::fifelse(test, yes, no), by=grp])
#    user   system  elapsed 
#3054.675    8.496  154.353 
```

MichaelChirico:
I don't see those timings:

```
system.time(DT[, base::ifelse(test, yes, no), by=grp])
#   user  system elapsed 
# 55.233   0.358  53.143 
system.time(DT[, data.table::fifelse(test, yes, no), by=grp])
#    user  system elapsed 
# 270.953   0.743  68.057 
```

and repeated:

```
microbenchmark::microbenchmark(times=10L, base=DT[, base::ifelse(test, yes, no), by=grp], DT=DT[, data.table::fifelse(test, yes, no), by=grp])
R Unit: seconds
 base 52.63954 55.03329 56.32826 56.08346 56.43899 60.67210    10
   DT 68.92007 72.52211 75.28474 76.14368 77.77457 80.74988    10
```

It is slower, though I'm not quite sure what's wrong w my benchmark 🤔

mattdowle:
Have postponed to next release.  Looks good but a bit too risky for this stage of cycle.  At least `fifelse` is exported and available and folk can start playing with it. Then this PR will turn optimization on by default to use it in the next release, assuming no difficulties arise in the meantime.

MichaelChirico:
Sounds good, as was expected I think. The reason it kept the 1.12.4 Milestone was with the aim to run revdep with this change to get an idea of the potential impact.

But that can also easily be done after release.

MichaelChirico:
Would it be possible to automatically re-route certain parallel loops to serial ones for a certain length condition? A macro is the first thing that comes to mind, something like (?)

https://github.com/Rdatatable/data.table/blob/d99b8ae4aa85a2426abc307a94c70905bd050416/src/fifelse.c#L75-L78

becomes

```
    AUTOLOOP(len0) {
      pans[i] = pl[i]==0 ? pb[i & bmask] : (pl[i]==1 ? pa[i & amask] : pna);
    }
```

where

```
#define PARALLEL_THRESHOLD 10000
#define AUTOLOOP(len0) \
  if(len0 < PARALLEL_THRESHOLD) \
    for (int i=0; i<len0; ++i) \
  else \
    #pragma omp parallel for num_threads(getDTthreads()) \
    for (int64_t i=0; i<len0; ++i) \
```

I don't think that's proper syntax but is something like that possible?

jangorecki:
AFAIU looping in macro will be portable only using `do ... while(0)` construct. It should be easier to just tweak getDTthreads, if returns 1, then there should(?) be no omp overhead. There is another way to escape omp using `if` in `pragma` as made in nafill https://github.com/Rdatatable/data.table/blob/d99b8ae4aa85a2426abc307a94c70905bd050416/src/nafill.c#L147-L148
this would be much cleaner way too

MichaelChirico:
Yes that looks great. Then we can focus on how to set the threshold -- do we want to be more sophisticated than to pick a fixed threshold? Wouldn't be too hard to write e.g. a Bayesian Optimization script to pick the threshold for a given loop targeting performance improvements at various input sizes, e.g.

jangorecki:
It sounds like best would be to have a new function `useDTthreads` having signature
```c
int useDTthreads(DTthreads, Niter, Pratio)
```
that could be used in a way
```c
#pragma omp parallel for num_threads(useDTthreads(getDTthreads(), nx, p))
```
where `Niter` is number of iterations to perform, and `Pratio` is a ratio used to make adjusting more dynamic according to an algorithm.

yet to confirm if `num_threads(1)` disables omp for most compilers, ideally looking at assembly of compiled code.

MichaelChirico:
As mentioned by @ColeMiller1 probably we won't want to do this unless we'll fix #4549

MichaelChirico:
I'm not so sure about this change. I would wait for a user request.

PR #3801, opened 2019-08-29
some clarifications in ?data.table

### LINKED ISSUE #3145: [Documentation] list/.() can be omitted in j on single column/expression
From the vignette [Introduction to data.table](https://cloud.r-project.org/web/packages/data.table/vignettes/datatable-intro.html):

>As long as `j`-expression returns a `list`, each element of the list will be converted to a column in the resulting `data.table`. 

>When there’s only one column or expression to refer to in `j` and `by`, we can drop the `.()` notation. This is purely for convenience. 

In `?data.table`, the first sentence appears in the **Arguments** section on `j`. However, the second piece of information does not appear here. Furthermore, in **Details**, the drop of `.()` is only explicitly described for `by`: "`.()` can be omitted in `by` on single expression for convenience"

I suggest that the part "When there’s only one column or expression to refer to in `j` [...], we can drop the `.()` notation. This is purely for convenience." is added also to the help text in **Arguments** section about `j`.

MichaelChirico:
Closes #3145

Do we put anything in the `NEWS` for a documentation-only change?


jangorecki:
No

Issue #609, opened 2014-06-08
[R-Forge #2360] Do list names clobber when list is final value of j={...}, too


arunsrinivasan:
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=2360&group_id=240&atid=978)

Should be fairly straightforward, with a good performance gain.

Highlighted by Steve L offlist on 7 Nov 2012,  and with related fix in 1.8.3.

See test 865.

MichaelChirico:
IINM this is also covered by #3802

mattdowle:
Closed by #3802

PR #3802, opened 2019-08-29
Autoname j

### LINKED ISSUE #2478: Auto-naming in j inconsistent (multi-statement j)
I find it odd that the following don't give the same output:

```
DT = data.table(a = 1:3, b = 4:6)

DT[ , .(a, b = b + 1)]
#    a b
# 1: 1 5
# 2: 2 6
# 3: 3 7


DT[ , {
  b = b
  .(a, b = b + 1)
}]
#    V1 b <- Why is a auto-inferred above, but not here?
# 1:  1 5
# 2:  2 6
# 3:  3 7
```

MichaelChirico:
Closes #2478 and #609

The meat is in the `do_j_names` function internal to `[.data.table`  so it inherits the right parent & can edit `jvnames` with `<<-`. Also realized this would clean up the fix done in #3604 so did that. The rest of the diff (in https://github.com/Rdatatable/data.table/pull/3802/commits/0b2e058e0f05e4d7fb51223b8e912266a2d95260) comes from using our `vapply_1{i,b,c}` wrappers more consistently

I imagine this one will have some revdep breaking (e.g. we broke two of our own tests)... maybe save for the next release then?


MichaelChirico:
Also FWIW I think we can make a small change to allow `chmatch` to accept `is.name` objects so we can use `jsub[[1L]] %chin% c('list', '.')`, etc, I copied this from `relop.c` in `r-source`, but it was breaking things, any ideas?

```
    // adapted from R source src/main/relop.c for how symbol == character is handled
    if (isSymbol(x)) {
      PROTECT_INDEX xp;
      PROTECT_WITH_INDEX(x, &xp);

      SEXP tmp = allocVector(STRSXP, 1);
      PROTECT(tmp);
      SET_STRING_ELT(tmp, 0, PRINTNAME(x));
      REPROTECT(x = tmp, xp);
      UNPROTECT(1);
    }
```

jangorecki:
I would say no to breaking changing in 1.12.4

PR #3806, opened 2019-08-31
Add na= to fifelse(test, yes, no, na)

### LINKED ISSUE #3753: make fifelse(test, yes, no, na)
(I've comment in https://github.com/Rdatatable/data.table/issues/3740#issuecomment-519751021. After a second though, I think it's better to open a new issue ticker but feel free to close it if it regards as duplicated)

Just tried the new `fifelse()`, fantastic!

However, I like the choice that `dplyr::if_else()` made - has an additional argument named `missing`.
It's because in R, the logic scalar may have three different values, `TRUE`, `FALSE` and `NA`.

Currently, `fifelse()` returns `NA` for test `NA`, I think it very good. But it may be better if we can have a function with an additional argument `na` like this:

```r
fifelse(test, yes, no, na)
```

Is it a good idea?

2005m:
Closes #3753


src/fifelse.c
```diff
@@ -42,57 +42,78 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b) {
   const int *restrict pl = LOGICAL(l);
   SEXP ans = PROTECT(allocVector(TYPEOF(a), len0)); nprotect++;
   copyMostAttrib(a, ans);
+  
+  if(!isNull(na))
+  {
```
jangorecki: please try to align to existing coding style of that file, so the following
```c
  if(!isNull(na))
  {
```
turns into
```c
  if (!isNull(na)) {
```

inst/tests/tests.Rraw
```diff
@@ -15864,6 +15864,22 @@ test(2095, any(grepl('override', capture.output(dcast(DT, a~b, fun.aggregate=len
 # gmean intermediate can overflow integers without warning, #986
 test(2096, data.table(a=c(1L,1L), v=c(2e9L, 2e9L))[, mean(v), a], data.table(a=1L, V1=2e9))
 
+# fifelse, #3753, add argument na for missing values
```
MichaelChirico: since it's quite recent fix, I would include these tests in the old suite

R/wrappers.R
```diff
@@ -5,7 +5,7 @@
 coalesce    = function(...) .Call(Ccoalesce, list(...), FALSE)
 setcoalesce = function(...) .Call(Ccoalesce, list(...), TRUE)
 
-fifelse = function(test, yes, no) .Call(CfifelseR, test, yes, no)
+fifelse = function(test, yes, no, na=NULL) .Call(CfifelseR, test, yes, no, na)
```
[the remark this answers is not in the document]
MichaelChirico: NA I think complicates things in terms of checking whether it has the same type as yes/no

src/fifelse.c
```diff
@@ -42,57 +42,78 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b) {
...
     #pragma omp parallel for num_threads(getDTthreads())
     for (int64_t i=0; i<len0; ++i) {
-      pans[i] = pl[i]==0 ? pb[i & bmask] : (pl[i]==1 ? pa[i & amask] : NA_CPLX);
+      pans[i] = pl[i]==0 ? pb[i & bmask] : (pl[i]==1 ? pa[i & amask] : pna);
     }
   } break;
   case VECSXP : {
     const SEXP *restrict pa = VECTOR_PTR(a);
     const SEXP *restrict pb = VECTOR_PTR(b);
+    const SEXP *restrict pna = VECTOR_PTR(na);
```
jangorecki: I think this line might be related to this 32bit windows failure.

R/wrappers.R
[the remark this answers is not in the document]
jangorecki: agree it complicates but now the `args(fifelse)` returns misleading feedback to user. We could eventually pass extra arg to C `missing(na)` which will make it easier.

inst/tests/tests.Rraw
2005m: where exactly do you want me to move them to?

R/wrappers.R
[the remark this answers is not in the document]
2005m: yes exactly it does, that is why I used `NULL`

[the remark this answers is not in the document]
jangorecki: Best if we can not use default value and just check `missing(na)` from C. And if missing then use NA of appropriate type, otherwise confirm types are matching.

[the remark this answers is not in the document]
jangorecki: there is [R_MissingArg](https://github.com/wch/r-source/blob/09b5b3f820648c20b088ba922f617848c0c00ab9/src/include/Rinternals.h#L878)

src/fifelse.c
```diff
@@ -43,8 +43,7 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
   SEXP ans = PROTECT(allocVector(TYPEOF(a), len0)); nprotect++;
   copyMostAttrib(a, ans);
   
-  if(!isNull(na))
-  {
+  if(!isNull(na)) {
```
jangorecki: also extra space between `if` and `(`

inst/tests/tests.Rraw
jangorecki: `fifelse` tests are already scattered in non-continuous chunks so I think it is OK as is.

man/fifelse.Rd
```diff
@@ -11,7 +11,7 @@
 \arguments{
   \item{test}{ A logical vector }
   \item{yes, no}{  Values to return depending on \code{TRUE}/\code{FALSE} element of \code{test}. They must be the same type and be either length \code{1} or the same length of \code{test}.}
-  \item{na}{ Value to return if an element of \code{test} is \code{NA}. It must be the same type as \code{yes} and \code{no} and lenght \code{1}.}
+  \item{na}{ Value to return if an element of \code{test} is \code{NA}. It must be the same type as \code{yes} and \code{no} and lenght \code{1}, by default it is \code{NA}.}
```
HughParsonage: spelling: 'lenght'

R/wrappers.R
```diff
@@ -5,7 +5,7 @@
 coalesce    = function(...) .Call(Ccoalesce, list(...), FALSE)
 setcoalesce = function(...) .Call(Ccoalesce, list(...), TRUE)
 
-fifelse = function(test, yes, no, na=NULL) .Call(CfifelseR, test, yes, no, na)
+fifelse = function(test, yes, no, na) .Call(CfifelseR, test, yes, no, if (missing(na)) substitute() else na)
```
HughParsonage: I don't agree with this change: missingness is difficult to work with. Why not just have `na=NULL` as the default then handle `NULL` as documented. Also clarifies its status as an optional argument in the `usage` section of the documentation.

jangorecki: why it would be difficult to work with? could you provide an example that would impose any difficulty on user, based on current implementation

HughParsonage: Suppose I want to use `data.table::fifelse` in my package. Then I too have to check for missingness every time I call it. 

```r
bad_fizzbuzz <- function(number, na) {
  fifelse(number %% 15, "fizzbuzz", "", na)
}
bad_fizzbuzz(1:100) # error

bad_fizzbuzz <- function(number, na) {
  if (missing(na)) {
    fifelse(number %% 15, "fizzbuzz", "") 
  } else {
    fifelse(number %% 15, "fizzbuzz", "", na) 
  }
}


```

jangorecki: I am not able to reproduce the problem. I am getting an error related to `test` argument, not `na`, which I can fix by:
```r
bad_fizzbuzz <- function(number, na) {
  fifelse(as.logical(number %% 15), "fizzbuzz", "", na)
}
bad_fizzbuzz(1:100)
```

HughParsonage: OK that one doesn't work. But the point is I have to be careful about `missing(na)` whenever I use a package that relies on missingness. 

Maybe this?

```r
lapply(list(list(yes = 1, no = 2, na = 3), list(yes = 2, no = 4)), function(el) fifelse(c(TRUE, FALSE, NA), el$yes, el$no, el$na))
```

jangorecki: Now it is better, although it still not convincing me.
Providing 0-length object to `na` should mean that we want to skip those elements from the answer. So be equivalent to
```r
ina = which(is.na(test))
fifelse(test[!ina], yes[!ina], no[!ina])
```
Which I don't think make sense to implement due to complexity of dynamic length answer. Because of that I will try to address your use case.

HughParsonage: The point is not that a user really would intend to provide zero-length NA, but there are circumstances where there is no way to dynamically convey missingness in a nested function. 

Using `NULL` is clearer and more powerful. Reserve `missing()` for when an argument is truly 'missing' -- *i.e.* should have been provided but wasn't and that fact needs to be caught early -- not for defaults.

HughParsonage: The other advantage to using `NULL` is that it's consistent with `dplyr::if_else` and `hutils::if_else` so users can use it as a drop-in replacement.

jangorecki: I amended code to support NULL. Note that setting default value to NULL would contradict manual about "must be length 1". I think it should satisfy the use case now.

mattdowle: @jangorecki the manual could contain 'Must be NULL or length 1". Then it could be fully consistent with `dplyr::if_else` and `hutils::if_else`.

mattdowle: I see the case for the default being`na=NA`. I think I find that clearer than `na=NULL` because when NA occurs in `test`, it is NA that is returned by default.   The manual could state that `na=NULL` is interpreted as if `na=NA`, for consistency with `dplyr::if_else` and `hutils::if_else`.

jangorecki: The point of not using `na=NA` default is that it adds (very tiny) overhead. Because `is.na` have to check for each type separately. Even if there is an R C function for that it still some extra overhead comparing to `na == R_MissingArg` or `isNull(na)`. It would be handy to have type agnostic NA (julia has such NA). 
I tend to agree now that default `na=NULL` is fine as long as we document that.

jangorecki: fix pushed, consistency to dplyr and hutils was already mentioned in this manual.

jangorecki:
yet it can be improved to handle
```r
fifelse(c(T,F,NA),c(1,1,1),c(2,2,2),NA)
#Error in fifelse(c(T, F, NA), c(1, 1, 1), c(2, 2, 2), NA) : 
#  'yes' is of type double but 'na' is of type logical. Please make sure that both arguments have the same type.
```

HughParsonage:
Not wanting to block the pull request, but should we support `length(na) == length(cond)`? Technically both `dplyr::if_else` and `hutils::if_else` do, although on reflection that may be a mistake.

```r
dplyr::if_else(c(NA, NA), 1:2, 1:2, 3:4)
```

Low priority as `coalesce` can already be used (and probably should be used) for such a purpose. And we can always support it in the future if it turns out there's a use case that necessitates it.

jangorecki:
yes, coalesce comes to my mind
```r
fifelse = function(test, yes, no, na) {
  if (!is.null(na)) coalesce(.Call(test, yes, no), na)
  else .Call(test, yes, no)
}
```
not sure if it make sense to duplicate that C functionality

shrektan:
Using `coalesce()` to implement has a drawback as `fifelse()` should only handle the NA in the `test` and leave the `NA` in `yes` or `no` as it is. For example, `fifelse(TRUE, NA, NA, 2)` should return `NA` instead of `2`. 

BTW, `fcase()` is a great idea. I do really like to have `fcase()` in data.table...

2005m:
@shrektan , if it is ok with the other member, could you please open a feature request for `fcase` and describe how you would expect it to work? It will other members to give their views as well. I might have a look at it in the future and not in the next 2 months as I am too busy. Thank you.

MichaelChirico:
yes please open an issue but I don't think it'll make this release so no
rush

PR #3811, opened 2019-09-02
rbindlist works for expression column

### LINKED ISSUE #546: [R-Forge #5252] rbindlist should support expression columns
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5252&group_id=240&atid=978)

From Jan Górecki on email :

``` r
library(data.table)
# desired solution
a <- data.table(c1 = 1, c2 = 'asd', c3 = expression(as.character(Sys.time())))
b <- data.table(c1 = 3, c2 = 'qwe', c3 = expression(as.character(Sys.time()+5)))
dt <- rbind(a,b)
#Error in rbindlist(allargs) : Unsupported column type 'expression'
dt <- rbindlist(list(a,b))
#Error in rbindlist(list(a, b)) : Unsupported column type 'expression'

# current workaround
a <- data.table(c1 = 1, c2 = 'asd', c3 =
list(expression(as.character(Sys.time()))))
b <- data.table(c1 = 3, c2 = 'qwe', c3 =
list(expression(as.character(Sys.time()+5))))
dt <- rbind(a,b)
dt <- rbindlist(list(a,b))
# both ok, but the desired solution would be:
dt[,list(c1,c2,eval(c3))]
# and it will not produce expected results because of previous workaround
# another workaround must be done:
dt[,list(c1,c2,lapply(c3,eval))]

```

MichaelChirico:
Closes #546 

Seems we shouldn't go too far down the path of allowing `expression` columns but the fix here is quite simple/easy to follow.


inst/tests/tests.Rraw
```diff
@@ -15870,6 +15870,10 @@ test(2097.1, DT[.(NA_character_)], data.table(fac=NA_character_, v=2L, key="fac"
 test(2097.2, DT[J(NA)],          error="Factor columns must join to factor or character columns")
 test(2097.3, DT[J(NA_integer_)], error="Factor columns must join to factor or character columns")
 
+# expression columns in rbindlist #546
+a <- data.table(c1 = 1, c2 = 'asd', c3 = expression(as.character(Sys.time())))
+b <- data.table(c1 = 3, c2 = 'qwe', c3 = expression(as.character(Sys.time()+5)))
+test(2098, rbind(a,b)$c3, expression(as.character(Sys.time()), as.character(Sys.time()+5)))
```
jangorecki: I don't find this useful at all. The new expression doesn't even evaluate. I am against combining language objects in rbindlist. The useful scenario I had in the past was to have them in a list column, which at that time I was not aware.

jangorecki:
maybe just add unit tests modifying original example by wrapping expressions into lists? then PR will cover some existing use case.

mattdowle:
Marked as 1.13.0 as a way to exclude from 1.12.4, but still come back to it.

mattdowle:
codecov results aren't updating. log states that upload succeeded but then there's a timeout after that line. a rerun didn't fix it.  proceeding to merge.

jangorecki:
@mattdowle any idea how this change is going to be useful?

PR #3812, opened 2019-09-02
Clean up & improve fastmean

### LINKED ISSUE #2799: fastmean optimization is not used

MichaelChirico:
Closes #2799

Only noticed @jangorecki had self-assigned (and more importantly labeled for next release) after doing this part (switched to match `mean(x, n=TRUE)` and pulled out R API usage from `for` loops).

Remains is to activate this outside grouping context -- wasn't sure whether to (1) duplicate code in non-grouping case around here:

https://github.com/Rdatatable/data.table/blob/3d0dc92e811fb15d01292566c82c136502c50155/R/data.table.R#L1273

(2) move the `fastmean` code somewhere earlier so that it applies before the code splits grouping v non-grouping evaluation:

https://github.com/Rdatatable/data.table/blob/3d0dc92e811fb15d01292566c82c136502c50155/R/data.table.R#L1664-L1697

(3) write a function `do_fastmean` which can be called in both cases (necessitates creating a copy of `jsub`, though we also do that a few other places now, cost seems to be ok even for "large" `jsub`

I also noticed `CPLXSXP` code is commented out, maybe worth supporting complex now too as an addendum to #3690

### DIFF
--- a/src/fastmean.c
+++ b/src/fastmean.c
@@ -51,7 +51,8 @@ SEXP fastmean(SEXP args)
     case INTSXP: {
       int *pi = INTEGER(x);
       for (i = 0; i<l; i++) {
-        if(pi[i] == NA_INTEGER) continue;
+        if (pi[i] == NA_INTEGER)
+          continue;
         s += pi[i];   // no under/overflow here, s is long double not integer
         n++;
       }
@@ -64,7 +65,8 @@ SEXP fastmean(SEXP args)
     case REALSXP: {
       double *pr = REAL(x);
       for (i = 0; i<l; i++) {
-        if(ISNAN(pr[i])) continue;  // TO DO: could drop this line and let NA propogate?
+        if (ISNAN(pr[i]))
+          continue;
         s += pr[i];
         n++;
       }
@@ -75,7 +77,7 @@ SEXP fastmean(SEXP args)
       s /= n;
       if(R_FINITE((double)s)) {
         for (i = 0; i<l; i++) {
-          if(ISNAN(pr[i])) continue;
+          if (ISNAN(pr[i])) continue;
           t += (pr[i] - s);
         }
         s += t/n;
@@ -92,7 +94,10 @@ SEXP fastmean(SEXP args)
     case INTSXP: {
       int *pi = INTEGER(x);
       for (i = 0; i<l; i++) {
-        if(pi[i] == NA_INTEGER) {UNPROTECT(1); return(ans);}
+        if (pi[i] == NA_INTEGER) {
+          UNPROTECT(1);
+          return(ans);
+        }
         s += pi[i];
       }
       REAL(ans)[0] = (double) (s/l);
@@ -101,7 +106,10 @@ SEXP fastmean(SEXP args)
     case REALSXP: {
       double *pr = REAL(x);
       for (i = 0; i<l; i++) {
-        if(ISNAN(pr[i])) {UNPROTECT(1); return(ans);}
+        if (ISNAN(pr[i])) {
+          UNPROTECT(1);
+          return(ans);
+        }
         s += pr[i];
       }
       s /= l;
--- a/src/fastmean.c
+++ b/src/fastmean.c
@@ -77,7 +77,8 @@ SEXP fastmean(SEXP args)
       s /= n;
       if(R_FINITE((double)s)) {
         for (i = 0; i<l; i++) {
-          if (ISNAN(pr[i])) continue;
+          if (ISNAN(pr[i]))
+            continue;
           t += (pr[i] - s);
         }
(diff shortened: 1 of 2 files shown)


src/fastmean.c
```diff
@@ -48,21 +48,24 @@ SEXP fastmean(SEXP args)
   if (narm) {
     switch(TYPEOF(x)) {
     case LGLSXP:
-    case INTSXP:
+    case INTSXP: {
+      int *pi = INTEGER(x);
       for (i = 0; i<l; i++) {
-        if(INTEGER(x)[i] == NA_INTEGER) continue;
-        s += INTEGER(x)[i];   // no under/overflow here, s is long double not integer
+        if(pi[i] == NA_INTEGER) continue;
```
jangorecki: - minor thing but you could also adjust spacing here `if (...`
- `continue` could go into new line so code coverage can test that

src/fastmean.c
```diff
@@ -48,21 +48,24 @@ SEXP fastmean(SEXP args)
...
         REAL(ans)[0] = R_NaN;  // consistent with base: mean(NA,na.rm=TRUE)==NaN==mean(numeric(),na.rm=TRUE)
       break;
-    case REALSXP:
+    }
+    case REALSXP: {
+      double *pr = REAL(x);
       for (i = 0; i<l; i++) {
-        if(ISNAN(REAL(x)[i])) continue;  // TO DO: could drop this line and let NA propogate?
-        s += REAL(x)[i];
+        if(ISNAN(pr[i])) continue;  // TO DO: could drop this line and let NA propogate?
```
jangorecki: `if` spacing, `continue`

comment could be removed, NAs should not propagate because it is `narm` branch

src/fastmean.c
```diff
@@ -72,41 +75,46 @@ SEXP fastmean(SEXP args)
...
   } else {  // narm==FALSE
     switch(TYPEOF(x)) {
     case LGLSXP:
-    case INTSXP:
+    case INTSXP: {
+      int *pi = INTEGER(x);
       for (i = 0; i<l; i++) {
-        if(INTEGER(x)[i] == NA_INTEGER) {UNPROTECT(1); return(ans);}
-        s += INTEGER(x)[i];
+        if(pi[i] == NA_INTEGER) {UNPROTECT(1); return(ans);}
```
jangorecki: `if` spacing
expand `{` into new lines for test coverage

MichaelChirico:
> re 1) and 2) fastmean should be used only in non-grouping context according to Matt comment in issue to be addressed.

Oh, I read that differently (currently we're only deploying `fastmean` in grouping but we could do so in non-grouping), @mattdowle can clarify?

src/fastmean.c
```diff
@@ -75,7 +77,7 @@ SEXP fastmean(SEXP args)
       s /= n;
       if(R_FINITE((double)s)) {
         for (i = 0; i<l; i++) {
-          if(ISNAN(pr[i])) continue;
+          if (ISNAN(pr[i])) continue;
```
jangorecki: why new newline here also for `continue` to be in code coverage?

jangorecki:
once we will add verbose print it will be clear where `fastmean` is being used, or if at all.

MichaelChirico:
> once we will add verbose print it will be clear where fastmean is being used, or if at all.

https://codecov.io/gh/Rdatatable/data.table/src/nogroup_fastmean/src/fastmean.c

(not that we shouldn't add verbose print but I haven't had time to figure out the API we're using for that)

We're clearly using `fastmean` based on the coverage tests? Or am I missing something

jangorecki:
2799 described some cases where it is not being used, so worth to track that better
```c
const bool verbose = GetVerbose();
if (verbose)
  Rprintf("%s: something",__func__);
```

MichaelChirico:
Not clear how important this is...

Issue #3581, opened 2019-05-22
revdep rerun


mattdowle:
Rerun with latest master to see the impact of #2734.   @MarkusBonsch was very careful to make no breaking changes.  Then I continued in the PR and did change a few things; e.g. now retaining the i column type when mismatch types are joined.

```
CRAN:
 ERROR   :   7 : antaresProcessing corpustools CoSMoS grattan KRIG tcpl VIM 
 WARNING :   1 : REIDS 
 NOTE    : 209 
 OK      : 449 
 TOTAL   : 666 / 666 
```

- [ ] antaresProcessing
- [ ] corpustools 
- [ ] CoSMoS
- [ ] grattan
- [ ] KRIG
- [ ] tcpl
- [ ] VIM
- [ ] REIDS

[fail.log](https://github.com/Rdatatable/data.table/files/3208442/fail.log)

MichaelChirico:
` genderizeR` gives the same errors on CRAN `data.table`. I see what looks like a bug in the code; have filed: https://github.com/kalimu/genderizeR/issues/9

MichaelChirico:
`musica` appears related to `rbindlist` combining `IDate` column (underlying `integer` with attribute) and `factor`:

```
DT1 = data.table(
  prse = structure(5478L, class = c("IDate", "Date"))
)
DT2 = data.table(
  prse = structure(1L, .Label = c("1970-01-01"), class = "factor")
)
rbind(DT1, DT2)
#          prse
# 1:       5478
# 2: 1970-01-01
```

MichaelChirico:
`optiSel` appears spurious. `WARNING` from its own C++ code and `NOTE` about the package size

MichaelChirico:
Oh, actually, `musica` is due to this:

https://github.com/Rdatatable/data.table/pull/3630#issue-285319167

Apparently `musica` was relying on untested behavior of `cut.IDate` to return an `IDate`; now it returns a factor but `mean.IDate` still returns `IDate`, hence the mismatch across groups for `rbindlist`.

Will restore `cut.IDate` and add some tests...

MichaelChirico:
I don't `parallelMap` has anything to do with us. Can reproduce the test error by cloning their GH & running `source('run-all.R')` from their `tests` directory, but the stack trace is:

```
Browse[1]> f
── 1. Error: batchtools mode (@test_batchtools.R#11)  ──────────────────────────
unused argument (V1 = 1)

[...]

Issue #3447, opened 2019-03-08
export forder


jangorecki:
so it can be legally used outside of `[.data.table`

mattdowle:
Discussion here: https://github.com/Rdatatable/data.table/pull/3817#discussion_r320659515

akersting:
I would also like to have `forder` and `forderv` exported. These are very useful functions also for other packages.

mattfidler:
I think you upvote on the top to show your interest

jangorecki:
Before exporting, we should resolve #4467, also `lazy` argument may need some rework because its current API is more internal-like.

MichaelChirico:
I'm curious the advantage of exporting `forder()`, doesn't base R's `order()` already do basically the same thing?

mattfidler:
Hi @MichaelChirico it is parallelized.  R's forder is not.

MichaelChirico:
Thanks. What about `DT[order(...), which=TRUE]`?

mattfidler:
> Thanks. What about `DT[order(...), which=TRUE]`?

For me, this requires access to the `dsl` inside of data.table, which is a bit difficult when I call it from `C`/`C++`

mattfidler:
Other use-cases may be more difficult.  Techincally from within C, I can load the data.table namespace and call the function directly, though it violates CRAN policies.  I suppose it does not if I have agreement with the `data.table` maintainers that this is legal.

MichaelChirico:
#4015 seems better suited to your use case

mattfidler:
That is why I opened the issue 😄;  This would be a work-around in the mean time.

mattfidler:
To really do this well you would have to have a non-linking API so that there is no binary dependencies.  (ie. don't simply register the C callable).

I have done something similar with `rxode2` recently.  It took me some time to figure out what needed to be done and I am planning to post a blog post about it.

MichaelChirico:
Thanks! Indeed it makes me lean to keeping `forder()` private on the R side. Then we could play around a little more with the API without exposing anything to the user. Context: #5167.

mattfidler:
In case you are interested in my solution to the API problem I wrote a blog post about it here.

There is an incomplete sentence right now which I intend to fix, but it does describe an approach.

https://blog.nlmixr2.org/blog/2024-09-18-nlmixr2-3.0.0-release/#api

PR #3817, opened 2019-09-04
order -> forder optimization follow up

mattdowle:
Follow up to #3604
Fixes revdeps `neatRanges` and `rbi.helpers`, #3581 

`forder(x,...)` changed to `forder(...)`  so can be a drop-in replacement for `base::order` in the cases where `base::order` returns a useful result. This makes the optimization to forder easier and works now when applied by `do.call` as `neatRanges` and `rbi.helpers` do.
Minimized the manipulation in the body of `forder()`. It now uses an `eval` approach instead of each column using `point()`.


R/setkey.R
```diff
@@ -204,49 +204,41 @@ forderv = function(x, by=seq_along(x), retGrp=FALSE, sort=TRUE, order=1L, na.las
...
+  asc = rep.int(1L, length(sub)-1L)  # ascending (1) or descending (-1) per column
+  # the idea here is to intercept - before vectors so that :
+  # i) we can pass the decreasing argument for that column through to C to avoid what normally happens in R (allocate a new vector and apply - to every element first)
+  # ii) - on character vector works; ordinarily in R that fails with type error
+  # We intercept the unevaluated expressions and massage them before evaluating in with(DT) scope or not depending on the first item.
+  for (i in seq.int(2L, length(sub))) {
+    v = sub[[i]]
+    while (is.call(v) && length(v)==2L && ((s<-v[[1L]])=="-" || s=="+")) {
+      if (s=="-") asc[i-1L] = -asc[i-1L]
+      sub[[i]] = v = v[[-1L]]  # remove the leading - (most commonly, or possibly +, --, +-- etc) to avoid the R level allocation upon eval
```
jangorecki: any particular reason why `v[[-1L]]` and not `v[[2L]]` instead?

R/setkey.R
```diff
@@ -204,49 +204,41 @@ forderv = function(x, by=seq_along(x), retGrp=FALSE, sort=TRUE, order=1L, na.las
...
     }
   }
-  cols = seq_along(ans)
-  # Supported column types are checked at C level
-  o = forderv(ans, cols, sort=TRUE, retGrp=FALSE, order= if (decreasing) -order else order, na.last)
-  if (!length(o)) o = seq_along(ans[[1L]]) else o
+  x = eval(sub[[2L]], parent.frame(), parent.frame())
+  if (is.list(x)) {
+    if (length(x)==0L && is.data.frame(x)) stop("Attempting to order a 0-column data.table or data.frame.")
+    sub[2L] = NULL  # change list(DT, ...) to list(...)
```
jangorecki: Ultimately it make sense to redirect users to use `forderv` directly when they want to pass DT into forder. If you agree, then maybe a verbose print for now informing about that? because any change here would be significant (and unnecessary) breaking change. Better to provide a verbose (or eventually non-verbose) print to user, so no one will be harmed.

R/setkey.R
```diff
@@ -204,49 +204,41 @@ forderv = function(x, by=seq_along(x), retGrp=FALSE, sort=TRUE, order=1L, na.las
...
+  if (length(sub)<2L) return(NULL)  # forder() with no arguments returns NULL consistent with base::order
+  asc = rep.int(1L, length(sub)-1L)  # ascending (1) or descending (-1) per column
+  # the idea here is to intercept - before vectors so that :
+  # i) we can pass the decreasing argument for that column through to C to avoid what normally happens in R (allocate a new vector and apply - to every element first)
+  # ii) - on character vector works; ordinarily in R that fails with type error
+  # We intercept the unevaluated expressions and massage them before evaluating in with(DT) scope or not depending on the first item.
+  for (i in seq.int(2L, length(sub))) {
+    v = sub[[i]]
+    while (is.call(v) && length(v)==2L && ((s<-v[[1L]])=="-" || s=="+")) {
+      if (s=="-") asc[i-1L] = -asc[i-1L]
```
MichaelChirico: why not just `-1L` on the RHS

src/forder.c
```diff
@@ -430,13 +430,15 @@ SEXP forder(SEXP DT, SEXP by, SEXP retGrpArg, SEXP sortGroupsArg, SEXP ascArg, S
     if (!isVectorAtomic(DT)) error("Input is not either a list of columns, or an atomic vector.");
     if (!isNull(by)) error("Input is an atomic vector (not a list of columns) but by= is not NULL");
     if (!isInteger(ascArg) || LENGTH(ascArg)!=1) error("Input is an atomic vector (not a list of columns) but ascArg= is not a length 1 integer");
```
jangorecki: in those few lines above you could move `error()` calls into new line so code coverage can test them

R/setkey.R
MichaelChirico: ah nvm. this handles `-+-+-+-` weird case

mattdowle: yeah it seemed to be doing that already so I left it doing that

R/setkey.R
mattdowle: No reason.  I guess `v[[2L]]` would read a little better.  Ok will change.

R/setkey.R
mattdowle: I don't quite follow. There is no breaking change here so far, at least, afaik. This line that changes (`list(DT,...)` to `list(...)`) is basically just maintaining the old usage of forder(DT, -colA, colB) so that still works. That form isn't what forderv is for, iiuc.

src/forder.c
mattdowle: Thanks, will do.  I sometimes forget this difference between R and C coverage (R coverage does detect uncovered branches within a line, while C coverage doesn't).

R/setkey.R
jangorecki: yes, forder has different API but IMO it worth to make distinction, even if not a breaking change, then just suggests to users. So `forder` is a `order` replacement and `forderv` is meant to be used by passing data.table to it.

mattdowle: We can suggest `forderv` in verbose mode I guess. But although `forderv` is meant for passing `DT` to it, a second aspect of `forderv` is that it doesn't accept expressions of columns. I was hoping that users can just use `DT[order(...), ]` as they have been and not need to know that `forder` and `forderv` exist. The optimization should hide these things from them and the news item (15) is consistent with that (e.g. it doesn't mention forder vs forderv). If verbose mode for `DT[order(...), ]` starts to suggest they use `forderv` it might be a bit confusing because i) they didn't even know they were using forder, and ii) neither `forder` or `forderv` are exported or documented.  Hopefully we never need to export them since optimization can take care of it.

src/forder.c
```diff
@@ -445,11 +445,11 @@ SEXP forder(SEXP DT, SEXP by, SEXP retGrpArg, SEXP sortGroupsArg, SEXP ascArg, S
       Rprintf("forder.c received %d rows and %d columns\n", length(VECTOR_ELT(DT,0)), length(DT));
   }
   if (!length(DT))
-    error("DT is an empty list() of 0 columns");
+    error("Internal error: DT is an empty list() of 0 columns");  // # nocov  should have been caught be colnamesInt, test 2099.1
```
jangorecki: does it make sense to push down call to `colnamesInt` from R to C? more related logic next to each other, and (probably very tiny) speed improvement.

R/setkey.R
jangorecki: is `DT[order(...), ]` optimised to `forder(DT, ...)`? if not, then the verbose message wouldn't be printed then. But agree that would be confusing till forder is not exported. Ultimately we want to export it https://github.com/Rdatatable/data.table/issues/3447 "so it can be legally used outside of `[.data.table`"

src/forder.c
mattdowle: colnamesInt is already a .Call to C. You added it quite recently iirc and it's nice. There are 11 calls currently from R level which seems right and nice to me.

jangorecki: I mean to call it from forder.c not from forder R

R/setkey.R
mattdowle: I see now, so verbose message would just be when first argument is `DT`.  `DT[order(...),]` is optimized to `DT[forder(...),]` so wouldn't print even in verbose mode.  I added link to #3447 back to this discussion so we can come back to it.

src/forder.c
mattdowle: Could do I suppose.  I've just made forderv even smaller at R level, so forderv could be a single .Call() eventually.  Moving `colnamesInt` call down to C for that reason might be nice. I feel like I've spent way too long on this already so let's pencil that in for future, if that's ok.

Issue #2053, opened 2017-03-08
testthat does not recognize data.tables ?


fruce-ki:
I started developing a new package using newer dat.table version and got stopped fairly early on but what seems to be an issue between 'testthat' and 'data.table'. I did not encounter that issue working with both packages for my previous package.

I have a function like this in my package:
```
#' Title
#' 
#' @import data.table
#' @export
#' 
DT_test <- function()
{
	dt <- data.table(acol= c("aA","aB", "aC"), 
				   bcol= c("bA", "bB", "bC"))
	dt[, ccol := c(1,2,3) ]
	return(dt)
}
```

And this is the corresponding unittest:
```
test_that("The data.table assignment works", {
	dt <- DT_test()
       # test stuff
})
```

I get this error:
```
Failed -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1. Error: The data.table assignment works (@test-1-devel.R#17) -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Check that is.data.table(DT) == TRUE. Otherwise, := and `:=`(...) are defined for use in j, once only and in particular ways. See help(":=").
1: DT_test() at /Users/cymon/workspace/Treat/tests/testthat/test-1-devel.R:17
2: dt[, `:=`(ccol, c(1, 2, 3))] at /Users/cymon/workspace/Treat/R/func.R:53
3: `[.data.table`(dt, , `:=`(ccol, c(1, 2, 3))) at /Users/cymon/workspace/Treat/R/func.R:53
4: `[.data.frame`(x, i, j)
5: `:=`(ccol, c(1, 2, 3))
6: stop("Check that is.data.table(DT) == TRUE. Otherwise, := and `:=`(...) are defined for use in j, once only and in particular ways. See help(\":=\").")
```

But if I call the function from a prompt or run it line for line it runs fine.

```
Session info -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.3.2 (2016-10-31)
 system   x86_64, darwin13.4.0        
 ui       RStudio (1.0.136)           
 language (EN)                        
 collate  en_GB.UTF-8                 
 tz       Europe/London               
 date     2017-03-08                  

[...]

PR #3818, opened 2019-09-04
use .datatable.aware when data.table in Suggests

tdhock:
Hi this PR was requested by @jangorecki after a short discussion on the linked issue (#2053). The change to the vignette adds some discussion about what to do when data.table is in Suggests (not imported) and used only in tests.

### DIFF
--- a/vignettes/datatable-importing.Rmd
+++ b/vignettes/datatable-importing.Rmd
@@ -82,6 +82,8 @@ test_that("generate dt", { expect_true(nrow(gen()) == 100) })
 test_that("aggregate dt", { expect_true(nrow(aggr(gen())) < 100) })
 ```
 
+If `data.table` is in Suggests (but not Imports) then you need to declare `.datatable.aware=TRUE` in one of the R/* files to avoid "object not found" errors when testing via `testthat::test_package` or `testthat::test_check`.
+
 ## Dealing with "undefined global functions or variables"
 
 `data.table`'s use of R's deferred evaluation (especially on the left-hand side of `:=`) is not well-recognised by `R CMD check`. This results in `NOTE`s like the following during package check:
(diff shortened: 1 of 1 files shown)


vignettes/datatable-importing.Rmd
```diff
@@ -82,6 +82,8 @@ test_that("generate dt", { expect_true(nrow(gen()) == 100) })
 test_that("aggregate dt", { expect_true(nrow(aggr(gen())) < 100) })
 ```
 
+If `data.table` is in Suggests (but not Imports) then you need to declare `.datatable.aware=TRUE` in one of the R/* files to avoid "object not found" errors when testing via `testthat::test_package` or `testthat::test_check`.
```
jangorecki: could you confirm this statement is also valid when NAMESPACE is present? you can import anything from base R, or export dummy function there

vignettes/datatable-importing.Rmd
```diff
@@ -167,7 +169,9 @@ my.write = function (x) {
 }
 ```
 
-When using a package as a suggested dependency, you should not `import` it in the `NAMESPACE` file. Just mention it in the `DESCRIPTION` file. You also _must_ use the `data.table::` prefix when calling `data.table` functions because none of them are imported.
+When using a package as a suggested dependency, you should not `import` it in the `NAMESPACE` file. Just mention it in the `DESCRIPTION` file. 
+When using `data.table` functions in package code (R/* files) you need to use the `data.table::` prefix because none of them are imported. 
+When using `data.table` in package tests (e.g. tests/testthat/test* files), you need to declare `.datatable.aware=TRUE` in one of the R/* files.
```
jangorecki: same here, good to know if presence of NAMESPACE file matters

vignettes/datatable-importing.Rmd
tdhock: works with blank NAMESPACE, and when I do `importFrom("stats", "sd")`

vignettes/datatable-importing.Rmd
tdhock: I don't think NAMESPACE matters. presence of `.datatable.aware=TRUE` means no error on `testthat::test_package` (as expected); absence means unexpected "object not found" error.

MichaelChirico: I think all this can be answered with `cedta` code:

```
isTRUE(ns$.datatable.aware)
```

where `ns` is the namespace of the call, so `.datatable.aware` covers all bases.

Earlier we see:

```
"data.table" %chin% names(getNamespaceImports(ns))
# getNamespaceImports main code:
.getNamespaceInfo(ns, "imports")
```

AFAICT these `names` correspond to `importFrom` or `import` statements in `NAMESPACE`

tdhock: not sure I understand. can you please clarify? are you saying that you may be able to modify cedta to fix the issue, without having me add `.datatable.aware=TRUE` in my package? that would be ideal.

MichaelChirico: no, I think without being in imports, `.datatable.aware` is probably the canonical way to do things. `.datatable.aware=TRUE` seems pretty minimal after all! Just a quick line of code to signal to `data.table` that you know we're here.

My comment was hoping to shed light on the discussion on whether `NAMESPACE` matters

tdhock: ok

PR #3819, opened 2019-09-04
INHERITS=>Rinherits for nanotime

mattdowle:
Follow up to #3806


jangorecki:
isn't it better to use extra `||` condition and avoid R call?

mattdowle:
Why better? The time of an R call for this should be immeasurable. Whereas hard coding nanotime is just supporting nanotime and not supporting any other S4 class that inherits from integer64 in the same way.

PR #3820, opened 2019-09-05
slightly relaxed .set_ops_arg_check

mattdowle:
To pass `corpustools` in #3581

### DIFF
--- a/R/setops.R
+++ b/R/setops.R
@@ -46,17 +46,13 @@ funique = function(x) {
   if (any(found)) stop("unsupported column type", if (sum(found) > 1L) "s" else "",
                        " found in x or y: ", brackify(bad_types[found]))
   super = function(x) {
-    # allow character/factor and integer/numeric through because from v1.12.4 i's type is retained by joins
+    # allow character->factor and integer->numeric because from v1.12.4 i's type is retained by joins, #3820
     ans = class(x)[1L]
-    if (ans=="factor") ans = "character"
-    else if (ans=="integer") ans = "numeric"
-    ans
+    switch(ans, factor="character", integer="numeric", ans)
   }
-  sx = sapply(x, super)
-  sy = sapply(y, super)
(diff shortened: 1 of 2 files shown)


R/setops.R
```diff
@@ -45,7 +45,19 @@ funique = function(x) {
                              vapply(y, typeof, FUN.VALUE = ""))
   if (any(found)) stop("unsupported column type", if (sum(found) > 1L) "s" else "",
                        " found in x or y: ", brackify(bad_types[found]))
-  if (!identical(lapply(x, class), lapply(y, class))) stop("x and y must have the same column classes")
+  super = function(x) {
```
MichaelChirico: Seems like a `switch` statement would be good:

```
ans = class(x)[1L]
switch(ans, factor = 'character', integer = 'numeric', ans)
```

R/setops.R
```diff
@@ -45,7 +45,19 @@ funique = function(x) {
...
+    # allow character/factor and integer/numeric through because from v1.12.4 i's type is retained by joins
```
MichaelChirico: nit pick: I would write `character->factor` and `integer->numeric` for clarity

R/setops.R
```diff
@@ -45,7 +45,19 @@ funique = function(x) {
...
+    ans = class(x)[1L]
+    if (ans=="factor") ans = "character"
+    else if (ans=="integer") ans = "numeric"
+    ans
+  }
+  sx = sapply(x, super)
+  sy = sapply(y, super)
+  if (!identical(sx, sy)) {
+    w = which.first(sx!=sy)
+    stop("Item ",w," of x is '",sx[w],"' but the corresponding item of y is '", sy[w], "'.")
```
MichaelChirico: wouldn't it be better to return the original class? e.g. if you gave `factor` and this error tells you `character` it'll tilt some heads

PR #3822, opened 2019-09-05
NaN considered NA_REAL_ in coalesce

### LINKED ISSUE #3790: coalesce() can't handle NaN
Currently, NaN is treated as if a non-NA value but I expect it's the same as NA_real_, because `is.na(NaN)` returns `TRUE`.

``` r
library(data.table)
coalesce(NaN, 1)
#> [1] NaN
is.na(NaN)
#> [1] TRUE
```

<sup>Created on 2019-08-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

MichaelChirico:
Closes #3790 

As mentioned there, (1) we're already treating them the same for the `CPLXSXP` case and (2) both `dplyr` and `hutils` consider `NaN` as `NA_REAL_` for their versions of `coalesce`.


shrektan:
I would appreciate if this can be merged in the next CRAN release. Thanks.

PR #3828, opened 2019-09-06
between tzone to pass CoSMoS

mattdowle:
To resolve `CoSMoS` in #3581 
```
require(CoSMoS)
example(analyzeTS)
```
WIP because although CoSMoS now gets past its `between()` call with the new message, comparing as UTC means nothing passes the subset and a later line then fails, as follows :
```
> ## Load data included in the package
> ## (to find out more about the data use ?precip)
> data('precip')
> ## Don't show: 
> ## test for one month to make it fast
> precip <- precip[between(date, as.POSIXct('1990-1-01', format('%Y-%m-%d'), tz = 'America/Regina'), as.POSIXct('1990-1-5', format('%Y-%m-%d'), tz = 'America/Regina'))]
'between' arguments are all POSIXct but have mismatched tzone attributes: ['', 'America/Regina', 'America/Regina']. The UTC times will be compared.
> a <- analyzeTS(precip)
Error in cor(x, y, use = "p") : 'x' is empty
Calls: analyzeTS ... seasonalACF -> lapply -> FUN -> sapply -> lapply -> FUN -> cor
Execution halted
```

Resolved. Need to run revdeps with TZ=UTC as CRAN does.

### DIFF
--- a/R/between.R
+++ b/R/between.R
@@ -21,8 +21,10 @@ between = function(x, lower, upper, incbounds=TRUE, NAbounds=TRUE) {
       tt = attr(x,"tzone",exact=TRUE)
       if (is.null(tt)) "" else tt
     })
+    # lower/upper should be more tightly linked than x/lower, so error
+    #   if the former don't match but only inform if they latter don't
     if (tzs[2L]!=tzs[3L]) {
-      stop("'between' lower= and upper= are both POSIXct but have different tzone attributes: ", brackify(tzs[2:3],quote=TRUE), ". Please align their tzone.")
+      stop("'between' lower= and upper= are both POSIXct but have different tzone attributes: ", brackify(tzs[2:3],quote=TRUE), ". Please align their time zones.")
       # otherwise the check in between.c that lower<=upper can (correctly) fail for this reason
     }
     if (tzs[1L]!=tzs[2L]) {
--- a/R/between.R
+++ b/R/between.R
@@ -17,12 +17,17 @@ between = function(x, lower, upper, incbounds=TRUE, NAbounds=TRUE) {
   }
   # POSIX check timezone match
   if (is.px(x) && is.px(lower) && is.px(upper)) {
-    tz_match = function(x, y, z) { # NULL match "", else all identical
-      ((is.null(x) || !nzchar(x)) && (is.null(y) || !nzchar(y)) && (is.null(z) || !nzchar(z))) ||
-        (identical(x, y) && identical(x, z))
+    tzs = sapply(list(x,lower,upper), function(x) {
+      tt = attr(x,"tzone",exact=TRUE)
+      if (is.null(tt)) "" else tt
+    })
+    if (tzs[2L]!=tzs[3L]) {
+      stop("'between' lower= and upper= are both POSIXct but have different tzone attributes: ", brackify(tzs[2:3],quote=TRUE), ". Please align their tzone.")
+      # otherwise the check in between.c that lower<=upper can (correctly) fail for this reason
     }
-    if (!tz_match(attr(x, "tzone", exact=TRUE), attr(lower, "tzone", exact=TRUE), attr(upper, "tzone", exact=TRUE))) {
-      stop("'between' function arguments have mismatched timezone attribute, align all arguments to same timezone")
+    if (tzs[1L]!=tzs[2L]) {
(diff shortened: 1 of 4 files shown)


R/between.R
```diff
@@ -17,12 +17,17 @@ between = function(x, lower, upper, incbounds=TRUE, NAbounds=TRUE) {
...
   if (is.px(x) && is.px(lower) && is.px(upper)) {
-    tz_match = function(x, y, z) { # NULL match "", else all identical
-      ((is.null(x) || !nzchar(x)) && (is.null(y) || !nzchar(y)) && (is.null(z) || !nzchar(z))) ||
-        (identical(x, y) && identical(x, z))
+    tzs = sapply(list(x,lower,upper), function(x) {
+      tt = attr(x,"tzone",exact=TRUE)
+      if (is.null(tt)) "" else tt
+    })
+    if (tzs[2L]!=tzs[3L]) {
+      stop("'between' lower= and upper= are both POSIXct but have different tzone attributes: ", brackify(tzs[2:3],quote=TRUE), ". Please align their tzone.")
```
MichaelChirico: feels odd to `stop` here but `message` in the `tzone(x)!=tzone(lower)` case

MichaelChirico:
`a <- analyzeTS(precip)` is not erroring for me with our `between_tz` and CRAN `CoSMoS`, are you on `CoSMoS` dev?

R/utils.R
```diff
@@ -84,8 +84,9 @@ name_dots = function(...) {
 
 # convert a vector like c(1, 4, 3, 2) into a string like [1, 4, 3, 2]
 #   (common aggregation method for error messages)
-brackify = function(x) {
+brackify = function(x, quote=FALSE) {
   # arbitrary cutoff
+  if (quote && is.character(x)) x = paste0("'",head(x,11),"'")
```
MichaelChirico: `head(x, 10L)` better, no?

R/between.R
mattdowle: It's settled with me after a few hours.  lower and upper are twins really, it would be hard to imagine them originating from different places in the code since they're created together.  Hence leave that as error; do the minimum necessary to pass CoSMoS.    But it's normal for `x` to originate from elsewhere, so treating x vs the twins differently to one twin vs the other twin seems reasonable.

R/utils.R
mattdowle: But then the next line doesn't add the `...`   The head(x,11) was to avoid doing the quoting on a huge 1m vector potentially.  11 just enough for the next line > 10 to trigger.

mattdowle:
Passes for me with TZ=UTC.  Now with this PR fix,  the second problem depends on your local timezone.  I had been running under `revdepsh` without TZ=UTC set.  `run()` sets `TZ=`.

R/utils.R
MichaelChirico: ah of course. it's early... was thinking `>=` not `>`.

revdep.R
```diff
@@ -1,6 +1,6 @@
 # Run by package maintainer via these entries in ~/.bash_aliases :
 #   alias revdepr='cd ~/build/revdeplib/ && R_LIBS_SITE=none R_LIBS=~/build/revdeplib/ _R_CHECK_FORCE_SUGGESTS_=false R_PROFILE_USER=~/GitHub/data.table/revdep.R R'
-#   alias revdepsh='cd ~/build/revdeplib/ && export R_LIBS_SITE=none && export R_LIBS=~/build/revdeplib/ && export _R_CHECK_FORCE_SUGGESTS_=false'
+#   alias revdepsh='cd ~/build/revdeplib/ && export TZ=UTC && export R_LIBS_SITE=none && export R_LIBS=~/build/revdeplib/ && export _R_CHECK_FORCE_SUGGESTS_=false'
```
MichaelChirico: i feel like this is setting us up for trouble as it's common not to have `TZ` set and we may want this to help us see related errors, no?

MichaelChirico: (I don't have `TZ` set btw)

mattdowle: CRAN sets `TZ=UTC`.  The trouble in testing under local timezone is that there's no cross-platform way to do it.

mattdowle: Although you don't have TZ set, your timezone is different to mine. I guess yours overlaps with the fixed timezone used by `CoSMoS` in a different way than mine, and that causes it to pass.

PR #3843, opened 2019-09-09
add test against regression on #2041

### LINKED ISSUE #2041: join or select row by col with Chinese in
library(data.table)
library(dplyr)

# case1
a1 <- data.table(
  a = c("a", "b", "c", "d"),
  b = c(1:12)
)

b1 <- data.table(
  a = c("a", "b", "e", "d")
)

a1[b1, on = "a"]

# case2
a1 <- data.table(
  a = c("中文1", "中文2", "中文3", "中文4"),
  b = c(1:12)
)

b1 <- data.table(
  a = c("中文1", "中文2", "中文5", "中文4")
)

a1[b1, on = "a"]

# case3
a1 <- data.table(
  a = c("中文一", "中文二", "中文三", "中文四"),
  b = c(1:12)
)

b1 <- data.table(
  a = c("中文一", "中文二", "中文五", "中文四")
)

a1[b1, on = "a"]
# something wrong !!!!!!!!!!!!!!
a1[a == "中文一"]
# something wrong !!!!!!!!!!!!!!
b1 %>%
  left_join(
    a1,
    by = "a"
  )

# case4
a1 <- data.table(
  a = c("中文一", "中文二", "中文三", "中文四"),
  b = c(1:12)
)

b1 <- data.table(
  a = c("中文一", "中文二", "中文三", "中文五")
)

a1[b1, on = "a"]

MichaelChirico:
Closes #2041 

(also hoping Appveyor will let us know if the issue persists on Windows)


shrektan:
I think the test is not needed because it should have been addressed by my previous PRs long time ago, see https://github.com/Rdatatable/data.table/pull/2566, https://github.com/Rdatatable/data.table/pull/2678 and https://github.com/Rdatatable/data.table/pull/3451 ... In fact, except my recent two issues, I'm pretty sure everything works fine because I'm using data.table to handle large numbers of non-ASCII related problems everyday.

jangorecki:
It is not needed as of now but we are adding it to be future proof

shrektan:
I mean the tests are already there. 

https://github.com/Rdatatable/data.table/blob/3bb130d51ffcb3a8ac2541ab8157c8df6f353bfa/inst/tests/tests.Rraw#L11948-L11964

MichaelChirico:
@shrektan thanks for tracking that down. I think that helps pin down why the test fails to compile on Windows.

To your point, this test is a little different though, right? That test is about keys on non-UTF8, this test has no keys & is just about subsetting.

shrektan:
I think it's about using the indexes or not and the previous tests should have used indexes as well. Anyway, I'm fine to make this test more specific. 

However, I suggest to use latin1 strings instead of the Chinese strings. The reason is if the native encoding of the Windows machine that runs the code is not GBK based, the test may fail. In contrast, the latin1 strings are supposed to be valid under any R environment. At least, converting latin1 to UTF-8 (or reverse) is always possible.

So if the Chinese strings really need to be used, at least before running the test, we should check `enc2utf8(enc2native(the_utf8_strings)) == the_utf8_strings` so that we can be sure the UTF-8 strings can be re-encoded correctly under the certain Windows machine.

PR #3846, opened 2019-09-09
i.realDouble joins to x.int again by coercing x.int to double

mattdowle:
Resolves `grattan` in #3581 

In v1.12.2 when joining i.realDouble to x.int,  fractions in i.realDouble would be truncated and would succeed in joining to those integers in x. That was widely agreed as wrong behavior. Then in dev we changed that to a type error in just the case when fractions are detected to be present in the i column.  This PR finally gets the balance right, I hope, by coercing x.int to double instead so the numbers with fractions will not return a match, and the type and contents of the i column is still preserved in the result. There is a verbose message consistent with the other similar coercions too.

In future we could introduce an option to turn all of the verbose coercion messages in bmerge (not just this one) into warnings or errors.  That way users can turn on the strict type-match-required option for production (to avoid wasteful coercions) without getting in the way of revdeps or new users.

The "Coerced" to "Coercing" change in verbose messages is just to make the verbose messages consistent. Chose "Coercing" to convey that the message is printed before the action and isn't past tense yet. If any of the coercions fails with an error for any reason then the verbose message will have just printed in current tense.

### DIFF
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -76,12 +76,12 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
         xclass=="logical" || iclass=="logical" ||
         xclass=="factor" || iclass=="factor") {
       if (anyNA(i[[ic]]) && all(is.na(i[[ic]]))) { # TODO: allNA function in C
-        if (verbose) cat("Coerced all-NA i.",names(i)[ic]," (",iclass,") to type ",xclass," to match type of x.",names(x)[xc],".\n",sep="")
+        if (verbose) cat("Coercing all-NA i.",names(i)[ic]," (",iclass,") to type ",xclass," to match type of x.",names(x)[xc],".\n",sep="")
         set(i, j=ic, value=match.fun(paste0("as.", xclass))(i[[ic]]))
         next
       }
       else if (anyNA(x[[xc]]) && all(is.na(x[[xc]]))) {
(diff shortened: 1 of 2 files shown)


MichaelChirico:
cc @HughParsonage as `grattan` co-author too

R/bmerge.R
```diff
@@ -105,10 +105,12 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           if (!is.null(attributes(i[[ic]]))) attributes(val) = attributes(i[[ic]])  # to retain Date for example; 3679
           set(i, j=ic, value=val)
           set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
+        } else {
+          if (verbose) cat("Coercing integer column x.",names(x)[xc]," to type double to match type of i.",names(i)[ic]," which contains fractions.\n",sep="")
```
MichaelChirico: Does `setNumericRounding` come into play here? If so we should note such (in the `verbose` message and/or in dox)

mattdowle: Yes it comes into play.  But it does anyway for regular double to double type-matching joins too.  `setNumericRounding` is really just to resolve numeric accuracy; more like very-slight-inaccuracy.  I'm not sure the wording to use to best convey this. I just got swamped by another 11 revdep fails in the full-rerun so I'll try and get through those as priority.  If you have something in mind for the message here, feel free to add in a follow-up PR, or raise an issue to follow-up so we don't forget. Agree.

PR #3853, opened 2019-09-10
setnames match to dup in colnames downgraded from error to warning

mattdowle:
Resolves `tsbox` in #3581


mattdowle:
Upload of codecov results failed with timeout error. Happens quite a lot. Low risk to proceed anyway and coverage will catch up on master anyway.

PR #3859, opened 2019-09-11
first and last not load xts namespace when not needed

### LINKED ISSUE #3857: first / last POSIXct should not load xts namespace
```r
library(data.table)
x = as.POSIXct("2019-01-01")
last(x)
#Registered S3 method overwritten by 'xts':
#  method     from
#  as.zoo.xts zoo 
#[1] "2019-01-01 CET"
```

### LINKED ISSUE #3858: first on empty DT should return empty DT
```r
dt = data.table(a=1,b=2)[0,]
first(dt)
#       a     b
#1:    NA    NA
head(dt, 1)
#Empty data.table (0 rows and 2 cols): a,b
```

jangorecki:
first on empty dt returns empty dt
closes #3857
closes #3858

we need a revdep re-run after merging this


mattdowle:
Merged to include in tonight's full revdep rerun #3581

PR #3869, opened 2019-09-13
Shut off warnPartialMatchArgs for certain R versions

### LINKED ISSUE #3865: Fail on win-builder for R 3.5.3
Passes OK on R 3.6.1

But fails with this on R 3.5.3 :
```
Running test id 1658.07      Test 1658.07 produced 1 warnings but expected 0
  Expected: 
  Observed: partial argument match of 'along' to 'along.with'
  
  Running test id 1658.08      
  Running test id 1658.09      Test 1658.09 produced 1 warnings but expected 0
  Expected: 
  Observed: partial argument match of 'along' to 'along.with'
  ```

MichaelChirico:
Closes #3865 

Confirmed the fix and the message on `rocker/r-ver:3.5.3`.

@jangorecki it strikes me that these `warn.*` options are not being turned on for the other test suites (`froll`, `nafill`, etc). Should we move this option setting to the `data.table:::test` level? Or just manually do this on a per-file basis?

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -60,11 +60,27 @@ test.data.table = function(verbose=FALSE, pkg="pkg", silent=FALSE, with.other.pa
   # sample method changed in R 3.6 to remove bias; see #3431 for links and notes
   # This can be removed (and over 120 tests updated) if and when the oldest R version we test and support is moved to R 3.6
 
-  oldverbose = options(datatable.verbose=verbose)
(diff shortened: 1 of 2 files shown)


jangorecki:
why not set it in `test.data.table`?

MichaelChirico:
ah, that's what i meant, yes

jangorecki:
definitely a way to go, test scripts should have as few config as possible, all should go to `test.data.table`, maybe even isolated into helper function having config only for easier maintenance.

Issue #3076, opened 2018-09-26
Disallow `Depends:`; require `Imports:` only.


mattdowle:
I've only recently realized how bad `Depends:` is,  thanks to Jan's importing vignette.  I just made its discouragement stronger :  https://github.com/Rdatatable/data.table/pull/3069/commits/51590bcfec6e9cccc9271800bcf01749a42c7e80

We could disallow Depends. This would also be beneficial to `cedta()`'s awkward implementation on its last line where it needs to do the tryCatch just for packages which Depend; that line could be removed.

But before we disallow Depends, we'd need to ask 69 CRAN packages to change from Depends to Imports.  (Most revdeps already Import.)   The longer we leave it, the greater the potential for new packages using Depends to be added to CRAN and the harder it will be to change.

- [ ] Ac3net
- [ ] AF
- [ ] batchtools
- [ ] bea.R
- [ ] behavr
- [ ] BuyseTest
- [ ] cdparcoord
- [ ] cffdrs
- [ ] classifierplots
- [ ] clickstream
- [ ] corpustools
- [ ] cvAUC
- [ ] damr
- [ ] dfmeta
- [ ] drgee
- [ ] easycsv
- [ ] edgeRun
- [ ] EurosarcBayes
- [ ] gbp
- [ ] GenomicTools
- [ ] GenomicTools.fileHandler
- [ ] greport
- [ ] haploReconstruct
- [ ] heatwaveR
- [ ] heims
- [ ] ie2misc
- [ ] iemisc
- [ ] JWileymisc
- [ ] koRpus
- [ ] LabourMarketAreas
- [ ] lookupTable
- [ ] LSPFP
- [ ] miLineage
- [ ] mrMLM
- [ ] mrMLM.GUI
- [ ] multicastR
- [ ] musica
- [ ] networkR
- [ ] NNS
- [ ] openSTARS
- [ ] orgR
- [ ] panelaggregation
- [ ] partools
- [ ] penaltyLearning
- [ ] pkggraph
- [ ] QuantTools
- [ ] Rbitcoin        @jangorecki please could you change from Depends to Imports :-)
- [ ] reinsureR
- [ ] riskRegression
- [ ] Rnets
- [ ] robCompositions
- [ ] RSauceLabs
- [ ] RWildbook
- [ ] sentometrics
- [ ] simPop
- [ ] simstudy
- [ ] sitree
- [ ] skm
- [ ] slim
- [ ] sparseFLMM
- [ ] stranger
[...]

PR #3899, opened 2019-09-21
prevent growth rate in depend and nudge them to import

mattdowle:
We're seeing new packages and revisions to existing packages that depend on data.table. We need to get them to import instead:  #3076.
In particular new packages that depend won't have received my previous emails.

So the idea is to prevent the growth rate, as a first step.

No change to any existing package,  not even a message yet.

But whenever a package next publishes,  they'll be required to import not depend.

This PR won't even catch most of the 73 packages; they'll still pass on their next update if they continue to depend and continue to work without even message. However, here are the 16 mentioned in the comments that will be caught whenever they next update.  Better than nothing.

batchtools
corpustools
CoSMoS
dfmeta
GenomicTools.fileHandler
haploReconstruct
JWileymisc
reinsureR
riskRegression
sentometrics
somspace
TreeLS
VIM 
musica
RWildbook
slim


jangorecki:
> So the idea is to prevent the growth rate, as a first step.

AFAIR when you submit CRAN package that Depends on another, CRAN maintainer asks why it Depends and not Imports. I have no idea if it is a rule for all CRAN maintainers, but IMO it is the best place to prevent growth rate.

Maybe there is a space to coordinate that with CRAN? AFAIU Imports rather Depends is ideal to-be state for all packages, so data.table could be a pilot package for works towards such ideal state.

inst/tests/tests.Rraw
```diff
@@ -1596,7 +1596,7 @@ test(550, DT[,sum(ID),by=key(DT)]$V1, INT(6,1,29,2,17))
...
 DT = as.data.table(DF)
-test(551, nrow(reshape(DT, v.names = c("all", "di.recip", "irr"),
+test(551, nrow(stats::reshape(DT, v.names = c("all", "di.recip", "irr"),
```
jangorecki: how was that spotted? env var for default packages set to empty?

R/cedta.R
```diff
@@ -37,7 +37,7 @@ cedta = function(n=2L) {
     (nsname %chin% cedta.pkgEvalsUserCode && any(sapply(sys.calls(), function(x) is.name(x[[1L]]) && (x[[1L]]=="eval" || x[[1L]]=="evalq")))) ||
     nsname %chin% cedta.override ||
     isTRUE(ns$.datatable.aware) ||  # As of Sep 2018: RCAS, caretEnsemble, dtplyr, rstanarm, rbokeh, CEMiTool, rqdatatable, RImmPort, BPRMeth, rlist
-    tryCatch("data.table" %chin% get(".Depends",paste("package",nsname,sep=":"),inherits=FALSE),error=function(e)FALSE)  # both ns$.Depends and get(.Depends,ns) are not sufficient
+    dependsNotImports(nsname)
```
jangorecki: I would say it is risky due to possible overhead that would occur in many places. `.onLoad` could still provide a way for some validation on that without putting an overhead in a critical place.

MichaelChirico: Won't this line of code only be run if the preceding are all `FALSE`? My first reaction was the same but I thought it is rare to get there

mattdowle: Yes it only reaches here via the tests that call `base`, `utils` or `stats`. I've seen covr/codecov tell us the number of hits per line but I haven't seen it tell us where those hits came from.

inst/tests/tests.Rraw
mattdowle: Locally in test I added a `stop()` to the start of `dependsNotImports` to create an error if it was ever reached, and this was of just 4 tests (iirc) that were affected.  I just added the `stats::` prefix for clarity, not because it was needed.

mattdowle:
> such a check is something that CRAN could/should coordinate/provide API for

But i) CRAN is just two people, and ii) data.table's requirement is somewhat unique because of our data.table-aware concept.
It would be good to turn this into something actionable; e.g. we could create a patch to `R CMD check --as-cran` to warn if data.table appears in Depends. If we tested this, it's something concrete to propose to CRAN to accept.

mattdowle:
I emailed this CRAN today.  This PR is closed then pending that outcome.  As Jan said, it's much better to place this check in CRAN incoming.

---

Dear CRAN,

Please can you apply the patch attached to `R CMD check --as-cran` with immediate effect. I've tested locally and the output, to pick a recent example, for sentometrics would be :

```
* checking package dependencies ... ERROR

data.table should be in Imports not Depends. Please contact its
maintainer for more information.

* DONE
```

Without `--as-cran`, sentometics and other existing packages on CRAN using Depend for data.table will continue to pass OK.

There are 73 CRAN packages which use Depends for data.table.  We have communicated with them for over 6 months but progress has been slow in getting them to use Imports instead. There are no known reasons why any package cannot use Imports instead.

Best, Matt

Two small changes to check.R and QC.R :
```
diff -r ./R-devel/src/library/tools/R/check.R ./R-devel-strict/src/library/tools/R/check.R
5373c5373
<                 if(length(res$suggested_but_not_installed))
---
>                 if(length(res$suggested_but_not_installed)) {
5380c5380,5385
<                 wrapLog(msg_DESCRIPTION)
---
>                    wrapLog(msg_DESCRIPTION)
>                 }
>                 if(length(res$depends_should_be_imports)) {
>                    wrapLog("data.table should be in Imports not Depends. Please",
>                            "contact its maintainer for more information.\n\n")
>                 }
diff -r ./R-devel/src/library/tools/R/QC.R ./R-devel-strict/src/library/tools/R/QC.R
3105a3106,3110
>     ## data.table should be import not depend; requested by data.table maintainer
>     if (check_incoming && "data.table" %in% depends) {
>       bad_depends$depends_should_be_imports <- "data.table"
>     }
> 
```

PR #3900, opened 2019-09-21
document internal test function

### LINKED ISSUE #2960: `test` function is insensitive for NaN vs NA difference

jangorecki:
- moves documentation of a `test` function from https://github.com/Rdatatable/data.table/wiki/Contributing into `man/test.Rd` (the proper place for function documentation)
- change order of `test`'s `message` arg to be just after the `warning`, to stack exceptions together
- removes deprecated arguments in `test.data.table` (`with.other.packages`, `benchmark`)
- change a default value of "dev-mode" argument `pkg="pkg"` to a nicer default `pkg="."` - maybe we could remove that argument at all?
- closes #2960 by documenting behaviour

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -1,9 +1,9 @@
-test.data.table = function(verbose=FALSE, pkg="pkg", silent=FALSE, with.other.packages=FALSE, benchmark=FALSE, script="tests.Rraw") {
+test.data.table = function(verbose=FALSE, pkg=".", silent=FALSE, script="tests.Rraw") {
   if (exists("test.data.table", .GlobalEnv,inherits=FALSE)) {
     # package developer
     # nocov start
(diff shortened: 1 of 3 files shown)


mattdowle:
Decrease in coverage just because we didn't merge master into this PR. Fine to skip in this case.

PR #3902, opened 2019-09-22
use typeof in print(class=TRUE)

### LINKED ISSUE #2998: print(DT,class=TRUE) could show type too
As Frank highlighted in https://github.com/Rdatatable/data.table/pull/2989#pullrequestreview-145215814,  maybe `print.data.table` could show both `class` and `typeof`.
(I just did a minimal change to that PR to make progress, and created this issue to follow up in future.)

MichaelChirico:
Closes #2998

I have a feeling `⊆` won't show up on windows. Open to suggestions for a better separator of `class(x)[1L]` and `typeof(x)`. `⊆` looks clear to me but maybe I've just been staring at it too long. Other things I considered: `#`, `->`, `/`, `~`, `-`.

Given this and breaking-ish nature ("front-end"-only change), marking for 1.13.0.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -224,6 +224,8 @@
     # [1] "A" "b" "c"
     ```
 
+29. The `class` argument to `data.table`'s `print` method now includes the `typeof` the column as well, [#2998](https://github.com/Rdatatable/data.table/issues/2998). `typeof` is quite important in R since it reveals the fundamental storage type of a given object, which is immutable without making copies, whereas the `class` of an object is closer to a decoration -- `setattr` can change the `class` without copies. As such, objects in R with the same `typeof` but different class are easier to combine than when `typeof` differs.
+
 ## BUG FIXES
 
 1. `first`, `last`, `head` and `tail` by group no longer error in some cases, [#2030](https://github.com/Rdatatable/data.table/issues/2030) [#3462](https://github.com/Rdatatable/data.table/issues/3462). Thanks to @franknarf1 for reporting.
@@ -1118,4 +1120,3 @@ When `j` is a symbol (as in the quanteda and xgboost examples above) it will con
 
 
 # data.table v1.9.8 (Nov 2016) back to v1.2 (Aug 2008) has been moved to [NEWS.0.md](https://github.com/Rdatatable/data.table/blob/master/NEWS.0.md)
-
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -76,14 +76,21 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     colnames(toprint)=rep("", ncol(toprint))
   if (isTRUE(class) && col.names != "none") {
     #Matching table for most common types & their abbreviations
-    class_abb = c(list = "<list>", integer = "<int>", numeric = "<num>",
-      character = "<char>", Date = "<Date>", complex = "<cplx>",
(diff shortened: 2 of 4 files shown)


TysonStanley:
Yeah, I think relying on special characters is also risky across platforms. I do love the idea of having both class and type of when they differ. I few brainstorming options:

1. Use `|` as in `<fctr|int>`
2. Use `+` as in `<fctr+int>`
3. Use `-` as in `<fctr-int>`

Personally, I like the idea of using `+` as it implies that both are present (i.e., this variable is a factor and an integer).

MichaelChirico:
My issue with `+` is it evokes for me multiple classes, e.g. `POSIXct+POSIXt`. Ditto basically for `|`. Maybe `\` or `/` would be a good cross-platform substitute? `POSc/int` or `POSc\int` or `int/POSc` or `int\POSc`

TysonStanley:
Yeah, I can see that. With that in mind, I think `/` makes a lot of sense and looks good visually to me (I'm not a fan of `\` for some reason...). Regardless of what you go with, having both `class` and `typeof` is a fantastic idea.

MichaelChirico:
Here's the full list of ASCII char we could use:

```
! " # $ % & ' (
) * + , - . / :
; < = > ? @ [ \
] ^ _ ` { | } ~
```

maybe `$`?

`int$fctr` for R hints at `fctr` being "part of" `int` since it's "extracted from" `int`. Similarly for `int@fctr` but I guess S4 is less familiar in general.

`>` or `<` would be great but are taken by the boundary delimiters already (we could change those pretty easily I think). 

Some other palatable options: `<fctr^int>`, `<fctr:int>`, `<fctr:int>`

jangorecki:
`<int:fctr>`? And what about extra option?

TysonStanley:
I really like the `<fctr:int>` (at least the way in looks). Dollar sign makes sense to me but doesn’t look great, in my opinion (it’s too crowded). Still think `<fctr/int>` also makes sense and looks good too.

jangorecki:
As int is a parent type for factor I would keep it on the left side of `:`

MichaelChirico:
Forgot to reference the Twitter poll...

Seems like `<factor:integer>` and `<integer@factor>` both do OK.

I also think we should shoehorn this into the `class` argument, which should be relaxed to have 2 levels:

 - `datatable.print.class = 0` (equiv. `FALSE`): current default
 - `datatable.print.class = 1` (equiv. `TRUE`): current `TRUE`, print `class(x)[1L]` only
 - `datatable.print.class = 2` (new): as in current PR

https://twitter.com/michael_chirico/status/1207559515043848193

jangorecki:
poll did not include such a nice way as `integer:factor`

MichaelChirico:
I think I left it out because for `:`, it's not clear to me the parent-child relationship. `integer::factor` I guess gets closer to what R users are in terms of `integer` being a "package" of which `factor` is a member of its "namespace".

But `integer:factor` vs `factor:integer`, I'm not so sure. I noticed recently I think the `tbl` (or it's `pillar`? really hard to trace) print method is using `:` to collapse column types, so it might be extra confusing for users switching between them.

jangorecki:
Actually if a typeof is a function of a class then, we may skip this extra info as well. It won't bring anything that user don't know. Maybe then print(type as a new option

PR #3903, opened 2019-09-23
add recommendation to importFrom data.table special symbols

MichaelChirico:
As suggested by Hadley here:

https://github.com/r-lib/roxygen2/issues/461#issuecomment-534077447

It's a good idea for exactly the reason that there may be some change down the line where we set e.g. `.N != NULL` and this would facilitate cascading this to downstreams.

### DIFF
--- a/vignettes/datatable-importing.Rmd
+++ b/vignettes/datatable-importing.Rmd
@@ -37,7 +37,7 @@ The next thing is to define what content of `data.table` your package is using. 
 
 You may also want to use just a subset of `data.table` functions; for example, some packages may simply make use of `data.table`'s high-performance CSV reader and writer, for which you can add `importFrom(data.table, fread, fwrite)` in your `NAMESPACE` file. It is also possible to import all functions from a package _excluding_ particular ones using `import(data.table, except=c(fread, fwrite))`.
 
-Be sure to read also the note about non-standard evaluation in `data.table` [below](#globals)
+Be sure to read also the note about non-standard evaluation in `data.table` in [the section on "undefined globals"](#globals)
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Should we also recommend `importFrom(data.table, ':=')`? Rather than setting `:=` to be `function(...) NULL` downstream... I guess it shouldn't make much of a difference (unless it's re-exported) but the messaging would be a but more consistent / easier to convey.

vignettes/datatable-importing.Rmd
```diff
@@ -33,10 +33,12 @@ The first place to define a dependency in a package is the `DESCRIPTION` file. M
 
 ## `NAMESPACE` file {NAMESPACE}
 
-The next thing is to define what content of `data.table` your package is using. This needs to be done in the `NAMESPACE` file. Most commonly, package authors will want to use `import(data.table)` which will import all (exported, i.e., listed in `data.table`'s own `NAMESPACE` file) functions from `data.table`.
+The next thing is to define what content of `data.table` your package is using. This needs to be done in the `NAMESPACE` file. Most commonly, package authors will want to use `import(data.table)` which will import all exported (i.e., listed in `data.table`'s own `NAMESPACE` file) functions from `data.table`.
 
 You may also want to use just a subset of `data.table` functions; for example, some packages may simply make use of `data.table`'s high-performance CSV reader and writer, for which you can add `importFrom(data.table, fread, fwrite)` in your `NAMESPACE` file. It is also possible to import all functions from a package _excluding_ particular ones using `import(data.table, except=c(fread, fwrite))`.
 
+Be sure to read also the note about non-standard evaluation in `data.table` [below](#globals)
```
jangorecki: rename `below` link to something meaningful, for example

> ... non-standard evaluation in `data.table` in [globals](#globals) section below

vignettes/datatable-importing.Rmd
```diff
@@ -122,6 +124,12 @@ The case for `:=` is slightly different, because `:=` is interpreted as a functi
 `:=` = function(...) NULL
 ```
 
+The case for `data.table`'s special symbols (`.SD`, `.BY`, `.N`, `.I`, `.GRP`, and `.EACHI`; see `?.N`) is slightly different. We recommend you import whichever of these values you use from `data.table`'s namespace to protect against any issues arising from the unlikely scenario that we change the exported value of these in the future, e.g. if you want to use both `.N` and `.I`, a minimal `NAMESPACE` would have:
```
jangorecki: mention it is only relevant when not `import(data.table)`

MichaelChirico: updated, please have a look

PR #3907, opened 2019-09-24
refine scipen NEWS item

MichaelChirico:
Feel free to refine further. The idea is to caution against just setting `scipen=999` or `scipen=.Machine$integer.max` which should have inefficient behavior w.r.t. the buffer.

Speaking of `scipen=.Machine$integer.max`, just realized I never stress tested this to see if anything breaks.

Just did so and seems OK (at least, it didn't crash):

```
# write one huge file
fwrite(rbindlist(replicate(1000000L, iris, simplify=FALSE)), tempfile(), scipen = .Machine$integer.max)
# Written 81.0% of 150000000 rows in 20 secs using 4 threads. maxBuffUsed=1%. ETA 4 secs.     

# write many tiny files
invisible(replicate(10000L, fwrite(iris, tempfile(), scipen = .Machine$integer.max)))
```

I guess the `maxBuffUsed=1%` number is where we see the effects (is 1% rounding up as well?)


NEWS.md
```diff
@@ -43,7 +43,7 @@
...
+    * Gains `scipen` [#2020](https://github.com/Rdatatable/data.table/issues/2020), the number 1 most-requested feature [#3189](https://github.com/Rdatatable/data.table/issues/3189). The default is `getOption("scipen")` so that `fwrite` will now respect this R option in the same way as `base::write.csv` and `base::format`, as expected. The parameter and option name have been kept the same as base R's `scipen` for consistency and to aid online search. It stands for 'scientific penalty'; the number of characters to add to the width within which non-scientific number format is used if it will fit. A high penalty essentially turns off scientific format. The efficiency-conscious should note that there will be a memory as well as a speed penalty to using large `scipen` values since we use this to allocate space for numeric fields; users with efficiency concerns but a need for high precision should set a generous but not arbitrarily large `scipen` according to their use case.
```
jangorecki: Nice, but yet could be more precise by giving an example.
I can imagine a user want to have a guarantee that scientific notation won't be stored in a csv, so it will be more portable. They can now use `format(x, sci=F)`.
Does `round(x, 10)` and then `fwrite(scipen=10)` resolves that?

mattdowle:
It didn't crash because I put in the clamp,  and you commented on that. Don't you remember?
I'm not sure it's true there's a speed penalty. The buffers will be bigger but they won't be used, that's all. So in the worst case, there's a possibility that the buffers will fail to be allocated.   The wording could be, in the allocation-fail message too, that if working memory fails to allocate then try reducing scipen.

mattdowle:
Coverge decrease spurious; comparing old branch to latest master.  Just a news item so skipping.

Issue #3626, opened 2019-06-04
Adding list column to one row data table


jakob-r:
I can not add a list column to a data.table with one row:
```r
library(data.table)
dt = data.table(a = 1)
list_column = list(list(a = 1, b = 2))
# does not work:
dt$b = list_column
```
```
Error in set(x, j = name, value = value) : 
  Supplied 2 items to be assigned to 1 items of column 'b'. If you wish to 'recycle' the RHS please use rep() to make this intent clear to readers of your code.
```
```r
dt[, b := list_column]
```
```
Error in `[.data.table`(dt, , `:=`(b, list_column)) : 
  Supplied 2 items to be assigned to 1 items of column 'b'. If you wish to 'recycle' the RHS please use rep() to make this intent clear to readers of your code.
```

The following workaround works and prodcues the desired result.
```r
dtb = data.table(b = list_column)
cbind(dt, dtb)
```
```
       a      b
   <num> <list>
1:     1 <list>
```

MichaelChirico:
could you confirm you are on the most recent master (share sessionInfo())?

a related bug was fixed recently

jangorecki:
if you are using devel version from our devel drat repos please provide
```r
data.table:::.git()
```

jakob-r:
Sorry. I always tend to forget. But yes. I just installed the latest version from github.

```
> sessionInfo()
R version 3.6.0 (2019-04-26)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Arch Linux

Matrix products: default
BLAS:   /usr/lib/libblas.so.3.8.0
LAPACK: /usr/lib/liblapack.so.3.8.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] colorout_1.2-1    usethis_1.5.0     devtools_2.0.2    data.table_1.12.3

[...]

PR #3909, opened 2019-09-25
Coerce 01

### LINKED ISSUE #996: Allow 0 and 1 singletons to coerce to logical without warning in `set()` and `:=` ?
Highlighted via http://stackoverflow.com/questions/7235657/fastest-way-to-replace-nas-in-a-large-data-table/7249454?noredirect=1#comment43831879_7249454

``` R
DT = data.table(a=1:4, b=c(FALSE, TRUE, NA, FALSE))
DT
#    a     b
#1: 1 FALSE
#2: 2  TRUE
#3: 3    NA
#4: 4 FALSE
set(DT,3L,1L,0)  # ok, no warning
set(DT,3L,2L,0)  # should be ok too without warning
# Warning message:
# In set(DT, 3L, 2L, 0) :
#   Coerced 'double' RHS to 'logical' to match the column's type; may have truncated precision. Either 
# change the target column to 'double' first (by creating a new 'double' vector length 4 (nrows of entire 
# table) and assign that; i.e. 'replace' column), or coerce RHS to 'logical' (e.g. 1L, NA_[real|integer]_, as.*, 
# etc) to make your intent clear and for speed. Or, set the column type correctly up front when you create 
# the table and stick to it, please.
DT
#    a     b
#1: 1 FALSE
#2: 2  TRUE
#3: 0 FALSE
#4: 4 FALSE
```

Apply to 0, 1, 0L and 1L.  Just length-1 vectors (i.e. singletons), not longer,  as already in place for 0 and 1 to integer.

### LINKED ISSUE #3915: NA as a factor level malformed by rbindlist
```
data.table(V1 = factor(as.character(c(NA, 1:100, NA)), exclude = NULL))
# Error in as.character.factor(x) : malformed factor
data.table(V1 = factor(as.character(c(NA, 1:3, NA)), exclude = NULL))
#      V1
# 1: <NA>
# 2:    1
# 3:    2
# 4:    3
# 5: <NA>
```

Is this the expected behaviour? 

Another example here: https://stackoverflow.com/q/58103098/4552295

I'm getting this result with `data.table_1.12.2` and `R version 3.6.1`.

mattdowle:
Closes #996 
(Closes #3915 too; slightly related regarding NA factor level)
Motivation in the news item.
The complex code removed needed work anyway as it was allocating within a truelength-clobber region. Clearly it's not ideal to making a change as big as this at this stage. On the other hand, it clears up a lot of wrinkles, hopefully.
Several potential coerces removed; e.g. when assigning factor to factor it was coercing the source to character which has gone away now. Plus some other similar todo done.Still some new todo's inline but it's passing locally.  Will run it past revdeps to see the impact.
Interested in first thoughts on this approach.


src/assign.c
```diff
@@ -819,89 +702,166 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+  } else if (TYPEOF(target)!=TYPEOF(source)) {
+    // checks up front, otherwise we'd need checks twice in the two branches that cater for 'where' or not
+    bool pun = false;
+    switch(TYPEOF(target)) {
+    case LGLSXP:
+      if (isInteger(source)) {
+        const int *sD = INTEGER(source);
+        for (int i=0; i<slen; ++i) {
+          const int val = sD[i];
+          if (val!=0 && val!=1 && val!=NA_INTEGER) error("Cannot assign %d to a logical column.", val);
```
MichaelChirico: could report `i` here... I'm not sure it will always be easy to understand what `i` is referring to though, e.g. assigning subsets

src/assign.c
```diff
@@ -819,89 +702,166 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+        }
+        pun = true;
+      } else if (isReal(source)) {   // TODO: cater for integer64 here
+        const double *sD = REAL(source);
+        for (int i=0; i<slen; ++i) {
+          const double d = sD[i];
+          if (!ISNAN(d) && d!=0.0 && d!=1.0) error("Cannot assign %f to a logical column.", d);
```
MichaelChirico: in the earlier branch you did the NA check last, I guess that will be more efficient in general, do that here as well?

src/assign.c
```diff
@@ -819,89 +702,166 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+        int w = INTEGER(isReallyReal(source))[0];  // first fraction present (1-based), 0 if none
+        if (w>0) {
+          warning("Coerced double RHS to integer to match the type of the target column (column %d named '%s'). One or more RHS values contain fractions which have been lost; e.g. item %d with value %f has been truncated to %d.", colnum, colname, w, REAL(source)[w-1], (int)REAL(source)[w-1]);
+        }
+        pun = true;
+      } else if (isLogical(source)) {
+        pun = true;
+      }
+      break;
+    case REALSXP:
```
MichaelChirico: Worth adding punning for `CPLXSXP`? should be similar to `REALSXP`. i can have a go if so.

src/assign.c
```diff
@@ -911,16 +871,41 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
       } else {
-        memcpy(INTEGER(target)+start, INTEGER(source), slen*SIZEOF(target));
+        switch (TYPEOF(source)) {
+        case LGLSXP:
+        case INTSXP:
+          memcpy(td, INTEGER(source), slen*SIZEOF(target));  // correct type ideal length>1 case
+          break;
+        case REALSXP: {
+          const double *sourceD = REAL(source);
+          for (int i=0; i<len; ++i) {
```
MichaelChirico: the `memcpy` branch is more efficient, is that right? should we report something under `verbose` then?

src/assign.c
```diff
@@ -982,13 +967,27 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+        } else {
+          for (int i=0; i<len; ++i) {
+            const int val = sd[i];
+            SET_STRING_ELT(target, start+i, val==NA_INTEGER ? NA_STRING : ld[val-1]);
+          }
+        }
       } else {
-        const SEXP *val = STRING_PTR(source);
-        for (int i=0; i<len; i++) SET_STRING_ELT(target, start+i, val[i]);
+        if (TYPEOF(source)!=STRSXP) error("Internal error. Not coerced earlier.");
```
MichaelChirico: nocov

src/assign.c
```diff
@@ -1003,27 +1002,81 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+          for (int i=0; i<len; ++i) {
+            const int w = wd[i];
+            if (w<1) continue;
+            double d = sd[i];
+            td[w-1] = ISNAN(d) ? NA_INTEGER : (int)d;
+          }
+        }
+      } break;
+      default:
+        error("Internal error");
```
MichaelChirico: nocov

src/assign.c
```diff
@@ -1003,27 +1002,81 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+      case REALSXP: {
+        const double *sd = REAL(source);
+        for (int i=0; i<len; i++) {
+          const int w = wd[i];
+          if (w<1) continue;
+          td[w-1] = sd[i&mask];
+        }
+      } break;
+      default:
+        error("Internal error");
```
MichaelChirico: nocov

src/assign.c
jangorecki: then we can also test that branch is reached exactly when we expect that by testing verbose output.

NEWS.md
```diff
@@ -224,6 +224,8 @@
     # [1] "A" "b" "c"
```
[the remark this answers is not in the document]
MichaelChirico: later read from C what's happening, confirmed by Q

mattdowle:
Thanks for the great reviews and comments. Seems like a cautious and tentative thumbs up then, depending on revdep results and addressing comments.
5 of 740 packages affected under this PR: `bigreadr` `hutils` `restatapi` `VIM` `wyz.code.testthat`
At least one seems correct: "Cannot assign 0.571429 to a logical column.".  Working through them...

src/assign.c
```diff
@@ -819,89 +702,166 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+      if (!isString(target) && !isNewList(target))
+        error("Cannot assign 'factor' to '%s'. Factors can only be assigned to factor, character or list columns.", type2char(TYPEOF(target)));
+      // else assigning factor to character is left to later below, avoiding wasteful asCharacterFactor
+    } else if (!sourceIsFactor && !isString(source)) {
+      // source is not character or factor.  If it's all-NA in another type, let it fall through
+      // to assign the NA later as-is without any coerce.  Otherwise, error.
+      bool ok = true;
+      switch(TYPEOF(source)) {
+      case LGLSXP: case INTSXP: {
+        const int *sd = INTEGER(source);
```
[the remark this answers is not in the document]
mattdowle: you mean the `allNA` I wrote in this PR?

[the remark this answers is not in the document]
MichaelChirico: yes, seems allNA was added but not used?

inst/tests/tests.Rraw
```diff
@@ -1218,16 +1216,14 @@ test(423, truelength(DT), 1028L)
 DT <- data.table(a = factor(c("A", "Z")), b = 1:4)
 DT[1,1] <- "Z"
 test(424, DT, data.table(a=factor(c("Z","Z","A","Z")),b=1:4))
-test(425, DT[1,1] <- 1, 1, warning="Coerced 'double' RHS to 'integer'")
-test(426, DT, data.table(a=factor(c("A","Z")),b=1:4))
-DT[1,1] <- 2L
+test(425, DT[1,1] <- 1, error="Cannot assign 'double' to 'factor'")
+test(426, DT[1,1] <- 2L, error="Cannot assign 'integer' to 'factor'")
```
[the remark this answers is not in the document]
mattdowle: turns out that no revdep fails because of it.  But yes, will reinstate anyway.

mattdowle:
Now passes the 2 revdeps that were affected (hutils and VIM). Will do another full rerun after getting full coverage.
Condensed logic using macro.  This gets the salient coerce parts next to each other for easier study.  Overall it's now less lines: diff was +368 - 276 yesterday and now with the macro is +366 - 398.

src/assign.c
```diff
@@ -811,322 +797,201 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source,
...
       break;
     case REALSXP:
-      pun = isLogical(source) || isInteger(source) || TYPEOF(source)==RAWSXP;
-      break;
-    case VECSXP:
-      pun = true;
+      if (targetIsI64 && isReal(source) && !sourceIsI64) {
+        int firstReal=0;
+        if ((firstReal=INTEGER(isReallyReal(source))[0])) {
+          sprintf(memrecycle_message, "coerced to integer64 but contains a non-integer value (%f at position %d); precision lost.", REAL(source)[firstReal-1], firstReal);
```
jangorecki: shouldn't we use `snprintf` here?

src/assign.c
```diff
@@ -811,322 +797,201 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source,
...
+    } else {                                   \
+      for (int i=0; i<len; i++) {              \
+        const STYPE val = sd[i];               \
+        const CTYPE cval = CAST;               \
+        ASSIGN;                                \
+      }                                        \
+    }                                          \
+  }}
+
+#define COERCE_ERROR(targetType) error("type '%s' cannot be coerced to '%s'", type2char(TYPEOF(source)), targetType); // 'targetType' for integer64 vs double
```
jangorecki: shouldn't be `source` also an argument to `COERCE_ERROR`, it iss being used inside

jangorecki: OK, I understand it is taken from parent scope in `switch` and I assume it never changes

src/assign.c
```diff
@@ -811,322 +797,201 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source,
...
-      } else {
-        const SEXP val = TYPEOF(source)==VECSXP ? VECTOR_ELT(source, 0) : source;
-        for (int i=0; i<len; i++) SET_VECTOR_ELT(target, start+i, val);
-      }
-      break;
-    default :
-      error("Unsupported type in assign.c:memrecycle '%s' (no where)", type2char(TYPEOF(target)));  // # nocov
+
+#undef BODY
+#define BODY(STYPE, RFUN, CTYPE, CAST, ASSIGN) \
```
jangorecki: this is really cool!

R/bmerge.R
```diff
@@ -75,12 +75,12 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
     if (xclass=="character" || iclass=="character" ||
         xclass=="logical" || iclass=="logical" ||
         xclass=="factor" || iclass=="factor") {
-      if (anyNA(i[[ic]]) && all(is.na(i[[ic]]))) { # TODO: allNA function in C
+      if (anyNA(i[[ic]]) && allNA(i[[ic]])) {
```
jangorecki: this change is nice but I would hold that for now, we support only few classes in `allNA`, are we sure none of the others can reach here.

src/assign.c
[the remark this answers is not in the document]
mattdowle: Yes PR was WIP then; now it uses. I've often wondered why `allNA` didn't seem to exist. So I thought you might have meant we did already have `all_na` or something already (like the request to export `any_na`), or that base/utils already had `all.na` or something somewhere that I didn't know about.

mattdowle:
- [ ] full coverage
- [ ] visit all comment
- [ ] the memcpy todo

src/assign.c
```diff
@@ -489,138 +488,22 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
...
-              }
-            }
-          }
-        }
-      }
     }
-    memrecycle(targetcol, rows, 0, targetlen, RHS);  // also called from dogroups where these arguments are used more
-    UNPROTECT(thisprotecti); // unprotect inside loop through columns to save protection stack
+    memrecycle(targetcol, rows, 0, targetlen, thisvalue, coln+1, CHAR(STRING_ELT(names, coln)));
+    // memrecycle is also called from dogroups;          ^^ last 2 arguments just for messages
```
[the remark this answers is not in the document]
mattdowle: thanks. I removed that part of the comment now.

src/assign.c
mattdowle: yep. good to do in future.

src/assign.c
mattdowle: what I had in mind was that if d is nan, then !=0.0 returns nan not true.  But good point. Might be worth coming back to.   ISNAN in R is defined to be C isnan so it's a very fast compare.  I doubt we'd be able to get a measurable difference here -- it's all register ops.

src/assign.c
mattdowle: done.

src/assign.c
mattdowle: yep there's a todo comment in the place to add the verbose/strict option.  Can do it in the next release.

NEWS.md
```diff
@@ -193,16 +193,16 @@
     set.seed(108)
     x = rnorm(1e6); n = 1e3
     rollfun = function(x, n, FUN) {
-      ans = rep(NA_real_, nx<-length(x))
+      nx = nx = length(x)
```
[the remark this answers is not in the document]
mattdowle: oops.  I just fixed the extra `nx=`.  My motivaton on this was that my editor was displaying several pages of highlighted green after this section. Something to do with the R syntax highlighting.  Could you make the other changes to that news item as you see fit please.
Why do 2 rows in the timings table have NA in the median column?  Perhaps explain that alongside as a comment to those rows.

NEWS.md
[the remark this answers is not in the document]
mattdowle: I don't understand this one. If something still needs to be done, please follow up as new comment.

src/assign.c
mattdowle: good catch on `snprintf`. done.

> ideally, for 1.13.0 we could try to take out coercion from assign and make it re-usable

that's what I did in this PR iiuc; i.e.  the assign code is taken out of assign() and placed into memrecycle().  memrecycle() is callable from elsewhere.

src/assign.c
mattdowle: yep this is a little inelegant.  it was because the type of REALSXP doesn't distinguish between "double" and "integer64", which is what I was trying to say by the comment alongside.  We probably do have a slightly higher level function superType() or something which does this.  If not we should create it and replace all type2char(TYPEOF(...)) with superType() instead.

R/bmerge.R
mattdowle: good idea.  would agree but I'm not sure what we'd wait for: all tests and all revdeps pass.  So might as well make progress.  Plus is quite a simple function:  hard to get wrong (touch wood!)

src/assign.c
```diff
@@ -819,89 +702,166 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source)
...
+        const SEXP *targetLevelsD=STRING_PTR(targetLevels), *sourceLevelsD=STRING_PTR(sourceLevels);
+        SEXP newSource = PROTECT(allocVector(INTSXP, length(source))); protecti++;
+        savetl_init();
+        for (int k=0; k<nTargetLevels; ++k) {
+          const SEXP s = targetLevelsD[k];
+          const int tl = TRUELENGTH(s);
+          if (tl>0) {
+            savetl(s);
+          } else if (tl<0) {
+            // # nocov start
```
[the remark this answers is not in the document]
mattdowle: Yes.  Just internal errors.  If somehow factor levels contain duplicates (an error because levels must be unique in R),  the 2nd dup would already have our negative truelength assigned to it from the first one (so that's one way to get tl<0). If there are no duplicates then what I had in mind was if core R ever changes their own truelength clobber to be negative (we assume and check that R's internal usage is always positive).   This is what I was trying to get at in the error message: "internal error: levels of target are either not unique or have truelength<0".

mattdowle:
Good: the formatting for coverage is doing its job:
image

NEWS.md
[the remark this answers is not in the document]
jangorecki: I pushed proposed changes.
@mattdowle please re-pull before committing

NEWS.md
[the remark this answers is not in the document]
jangorecki: fixed in a415095ca382a1b5520027739541ef19ef0dd1a3

R/bmerge.R
jangorecki: @mattdowle I pushed that in extra commit now, could you please re-run revdeps? it was previously returning `false` by default, so the potential issues could have been missed. Now it is more explicit and will raise error.

src/utils.c
```diff
@@ -36,29 +36,50 @@ bool allNA(SEXP x) {
...
-  return false;
+  }
+  case VECSXP: {
+    return false;
+  }
+  case RAWSXP: {
+    return false;
+  }
+  default:
+    error("Internal error: unsupported type '%s' passed to allNA()", type2char(TYPEOF(x))); // # nocov
```
jangorecki: We have `raw` and `list` in tests, instead of returning `false` by default we return `false` only for those two, and raise error otherwise. I still see a potential for issues here, for example:
If user defines own class using a list type. Then user can have `is.na.my_class` method, which can work on list, while our `allNA` will not work returning FALSE always.
```
my_cls = structure(list(a=1), class="my_class")
is.na.my_class = function(x) TRUE
```

src/utils.c
```diff
@@ -36,29 +36,50 @@ bool allNA(SEXP x) {
...
   // whereas this allNA can often return early on testing the first value without reading the rest
   const int n = length(x);
-  if (n==0) return true; // for empty raw, and NULL, to match R's all(is.na(raw())) true result, test 2116.13-14
+  if (n==0) // for empty raw, and NULL, to match R's all(is.na(raw())) true result, test 2116.13-14
+    return true;
   switch (TYPEOF(x)) {
   case LGLSXP:
   case INTSXP: {
     const int *xd = INTEGER(x);
-    for (int i=0; i<n; ++i)   if (xd[i]!=NA_INTEGER)   return false;
```
jangorecki: @mattdowle sorry for removing your code alignment. I made it more friendly for codecov now.

NEWS.md
```diff
@@ -192,24 +192,24 @@
...
 
     ### fun             mean     sum  median
-    # base rollfun     8.815   5.151  60.175
+    # base_rollapply   8.815   5.151  60.175
     # zoo::rollapply  34.373  27.837  88.552
-    # zoo::roll[fun]   0.215   0.185      NA
+    # zoo::roll[fun]   0.215   0.185      NA   ## median not fully supported
     # frollapply       5.404   1.419  56.475
-    # froll[fun]       0.003   0.002      NA
+    # froll[fun]       0.003   0.002      NA   ## median not yet supported
```
mattdowle: @jangorecki Ahah: I had not guessed that was the reason for the NA (i.e. median not supported).  When I was reading quickly like most people,  I thought it was mean, sum and median of a set of timings,  a bit like microbenchmark output.  Now it makes sense!

NEWS.md
```diff
@@ -221,7 +221,7 @@
     # [1] "A" "B" "C"
     setnames(DT, 2:3, tolower)
     names(DT)
-    # [1] "A" "b" "c"
+    # [1] "a" "b" "c"
```
mattdowle: @jangorecki Why this change?   The `setnames` call sets `2:3` so `c("A","b","c")` was correct.

mattdowle: I guess the 2:3 wasn't clear enough when reading quickly.  I changed it to c(1,3) to make it clearer in 7cad29b

jangorecki: Oh, yes, I haven't notice 2:3! Thanks

src/utils.c
mattdowle: Good points. `raw` doesn't support NA so that's why `raw`, other than `allNA(raw())` is true and I included test for.  All empty vectors return true.
I wasn't planning on exporting `allNA` in this release.  Just using it internally to start with.
`list`: yes there's some wrinkles there.  `allNA(list())` is correctly true but I don't think I covered a test for that.  But `allNA(list(NA))` is false whereas `all(is.na(list(NA)))` is true in base R. So that needs fixing.  I don't think we pass `list` to `allNA` internally in this PR,  but it should be error at least until it supports it,  otherwise we're just relying on test coverage too much.  Will do.

mattdowle: base R behaviour:
```R
> all(is.na(list(NA, NA)))
[1] TRUE
> all(is.na(list(NA, c(NA))))
[1] TRUE
> all(is.na(list(NA, c(NA,NA))))
[1] FALSE
```

MichaelChirico: there's an explicit

```
	if (!isVector(s) || length(s) != 1)
	    pa[i] = 0;
```

in `do_isna` in `src/main/coerce.c`. No comment as to why... not changed in 19 years

https://github.com/wch/r-source/commit/74bb7a0bdc5e4b21464ec87343f35bf834a6620c

mattdowle: Interesting.  Actually, turning off list support in allNA (so that it errors with unsupported) does cause 3 tests to fail:  1294.05, 1294.12 and 1294.19.   Just in this PR because this PR is now using allNA in bmerge.

mattdowle: And those 3 tests all warn with `Coerced list RHS to logical|integer|double` with on RHS `list(list(1))`.  Comforting that the circle leads back to this `coerce_01` PR.

mattdowle: Since those 3 were warnings already, and with #3626 in mind, I turned them to error and moved on.  It's strange we were allowing these `list` singletons to be coerced to atomic, albeit with warning.  I can see it causing problems with #3626 so those 3 being error make sense for now I hope.

R/utils.R
```diff
@@ -12,6 +12,7 @@ if (base::getRversion() < "3.5.0") {
 }
 isTRUEorNA    = function(x) is.logical(x) && length(x)==1L && (is.na(x) || x)
 isTRUEorFALSE = function(x) is.logical(x) && length(x)==1L && !is.na(x)
+allNA = function(x) .Call(C_allNAR, x)
```
MichaelChirico: just ran a quick benchmark and we blow `all(is.na())` out of the water (as expected when `FALSE` but we're also faster when `TRUE` and for moderate vector size).

Worth patching to `base`? It seems not as popular to use as `anyNA` in `base` internal code:

```
grep -Er "all\s*\(is\.na\s*\(" ~/github/r-source --exclude=*tests* | wc -l
# 28, 63 if we include tests code
grep -Er "anyNA\s*\(" ~/github/r-source --exclude=*tests* | wc -l
# 190, 197 if we include tests code
```

MichaelChirico: (as a TODO after release of course)

src/utils.c
```diff
@@ -33,7 +33,7 @@ SEXP isReallyReal(SEXP x) {
 }
 
 bool allNA(SEXP x, bool errorForBadType) {
-  // less space and time than any(is.na(x)) at R level because that creates full size is.na(x) first before any()
+  // less space and time than all(is.na(x)) at R level because that creates full size is.na(x) first before all()
```
mattdowle: oops. thanks @MichaelChirico good spot

mattdowle:
More review needed please. Merging to run it through stronger tests in GLCI and so I can move on to the remaining issues in this milestone. Can still add comments in the closed PR and/or open follow-up PRs. Will do another full revdep rerun too.

jangorecki:
As of now I see multiple places that could be extended with newlines for C codecov.

jangorecki:
we have some strange error not related to this PR
```
ERROR: Preparation failed: Error response from daemon: received unexpected HTTP status: 500 Internal Server Error (executor_docker.go:188:0s)
```
only for R-devel linux image

jangorecki:
Seems to be working fine now.

PR #3920, opened 2019-09-28
optimized Rinherits

### LINKED ISSUE #3919: Rinherits short-circuit
It should only call R if it's an S4 object, otherwise it can use INHERITS.
It can also have an if for nanotime too to save calling R for nanotime.

mattdowle:
Closes #3919

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -209,17 +209,22 @@ inline bool INHERITS(SEXP x, SEXP char_) {
 }
 
 bool Rinherits(SEXP x, SEXP char_) {
- // motivation was nanotime which is S4 and inherits from integer64 via S3 extends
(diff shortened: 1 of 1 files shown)


mattdowle:
As expected and desired, the `eval()` branch is no longer covered.  Will leave it uncovered for this release and come back to it in future when we cover the remaining C uncovered lines. I put a TODO commment there that describes how to cover it.

jangorecki:
looks good, but I think it make sense to fill a request for a better interface to S4 classes

mattdowle:
By 'fill a request' you mean ask R core to do something?

jangorecki:
yes

mattdowle:
I'll leave that to you then.

PR #3923, opened 2019-09-30
non-character-allNA done directly rather than tricking BODY

mattdowle:
Follow up to #3909 
On balance I figured it was clearer to do the two cases directly (hopping via 'where' or not) to assign the NA_STRING rather than tricking the BODY macro.
Plus some formatting.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -683,6 +683,8 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source,
     error("Internal error: recycle length error not caught earlier. slen=%d len=%d", slen, len); // # nocov
   // Internal error because the column has already been added to the DT, so length mismatch should have been caught before adding the column.
   // for 5647 this used to limit slen to len, but no longer
+  if (colname==NULL)
+    error("Internal error: memrecycle has received NULL colname"); // # nocov
   *memrecycle_message = '\0';
   int protecti=0;
   if (isNewList(source)) {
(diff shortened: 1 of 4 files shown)


src/assign.c
```diff
@@ -930,48 +929,55 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source,
...
   switch (TYPEOF(target)) {
   case RAWSXP: {
     Rbyte *td = RAW(target) + off;
     switch (TYPEOF(source)) {
-    case RAWSXP:  if (mem) {
+    case RAWSXP:
+      if (mc) {
                     memcpy(td, RAW(source), slen*sizeof(Rbyte)); break;
-          } else    BODY(Rbyte, RAW,    Rbyte, val,                                     td[i]=cval)
+      } else        BODY(Rbyte, RAW,    Rbyte, val,                                     td[i]=cval)
```
jangorecki: Newline for BODY expression would be useful for coverage

src/assign.c
```diff
@@ -930,48 +929,55 @@ const char *memrecycle(SEXP target, SEXP where, int start, int len, SEXP source,
 
 #define COERCE_ERROR(targetType) error("type '%s' cannot be coerced to '%s'", type2char(TYPEOF(source)), targetType); // 'targetType' for integer64 vs double
 
-  const int off = (length(where)?0:start); // off = target offset; e.g. called from rbindlist with where=R_NilValue and start!=0
-  const bool mem = length(where)==0 && slen==len;  // only used if types match too
+  const int off = length(where) ? 0 : start;  // off = target offset; e.g. called from rbindlist with where=R_NilValue and start!=0
+  const bool mc = length(where)==0 && slen==len;  // mc=memcpy; only used if types match too
+  const int *wd = length(where) ? INTEGER(where)+start : NULL;
```
jangorecki: wd seems to be used only in string branch, why not to move that line there?

src/assign.c
mattdowle: Please see image I pasted here before which shows the newline isn't needed for `else`.
https://github.com/Rdatatable/data.table/pull/3909#issuecomment-535728419

src/assign.c
mattdowle: wd is used in BODY macro

PR #3926, opened 2019-10-01
i symbol incorrect not found in calling scope

### LINKED ISSUE #3669: character column of x no longer available as ii in x[ y[ii] ]
I used code like this for a long time:

```
# example
library(data.table)
# data.table 1.12.3 IN DEVELOPMENT built 2019-06-19 19:02:40 UTC; travis using 4 threads (see ?getDTthreads).  Latest news: r-datatable.com
iDT = data.table(key = "i_id",
    i_id = c("A", "B", "C", "D"), 
    g = state.name[c(1,1,2, 3)], 
    e_date = as.IDate(c("2019-01-20", "2019-01-20", "2019-01-01", "2019-01-01")),
    e_time = as.ITime(c("14:00", "20:00", "20:00", "20:00"))
)
set.seed(1)
vDT = data.table(i_id = unique(iDT$i_id))[, .(v = runif(5,0,10), p = sample(c(5,5,10,10,10))), by=i_id]

# old code that now errors due to bare i_id in the i= clause
vDT[c("B", "C", "A"), on=.(i_id), .N, by=.EACHI][iDT[i_id, order(e_date, e_time)]]

# longer/clearer version of the same
tmp = vDT[c("B", "C", "A"), on=.(i_id), .N, by=.EACHI]
tmp[iDT[i_id, order(e_date, e_time)]]
```

but the last line now yields 

> Error in `[.data.table`(iDT, i_id, forder(x, e_date, e_time)) : 
  i_id is not found in calling scope but it is a column of type character. If you wish to select rows where that column contains TRUE, or perhaps that column contains row numbers of itself to select, try DT[(col)], DT[DT$col], or DT[col==TRUE] is particularly clear and is optimized. When the first argument inside DT[...] is a single symbol (e.g. DT[var]), data.table looks for var in calling scope.

Contrary to the error message, `i_id` is in calling scope from `tmp`, right? Here's working code illustrating expected behavior:

```
tmp[iDT[tmp$i_id, order(e_date, e_time)]]
#    i_id N
# 1:    C 5
# 2:    A 5
# 3:    B 5
```

My code was confusing (took me a minute to figure out what I was trying to do before I could start debugging) so I don't mind if the idiom I used no longer works. I'm just bringing it up in case the scoping is intended to work there.

mattdowle:
Closes #3669 
Working now in dev so just added tests.

Fixed by #3817 on Sep 4th. I tested with the commit before that which failed with the same error reported in dev Jun 27th.  I'm not quite sure precisely why #3817 fixes it, or what created the problem in dev before that, but it had changes in the area of evaluating `i`, so we'll call it a win.

Thanks @franknarf1 for testing dev!

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16314,6 +16314,30 @@ test(2119.15, data.table(a=1L)[,  newcol := list(2L, 3L)],       error="Supplied
 test(2119.16, data.table(a=1:2)[, newcol := list(list(2L, 3L))], ans<-data.table(a=1:2, newcol=list(2L,3L)))
 test(2119.17, data.table(a=1:2)[, newcol := list(2L, 3L)],       ans)
 
+# i symbol fetch from calling scope; #3669
+iDT = data.table(key = "i_id",
+    i_id = c("A", "B", "C", "D"),
+    g = state.name[c(1,1,2,3)],
+    e_date = as.IDate(c("2019-01-20", "2019-01-20", "2019-01-01", "2019-01-01")),
+    e_time = as.ITime(c("14:00", "20:00", "20:00", "20:00"))
+)
+set.seed(1)
+vDT = data.table(i_id = unique(iDT$i_id))[, .(v = runif(5,0,10), p = sample(c(5,5,10,10,10))), by=i_id]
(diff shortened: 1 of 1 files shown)


MichaelChirico:
gotta love it when the last issue solves itself!

PR #3933, opened 2019-10-03
Installed size note

### LINKED ISSUE #3932: installed size 5.1MB note
On fedora-clang but not fedora-gcc, interestingly. A little odd because they both include `-g` which increases `.so` size by 3 times. But 5.1MB has only just breached the 5.0MB threshold, so easiest for now is just do some more compression on the data files used in tests. In future we could compress the test files as well; `tests.Rraw` is the biggest at 908KB.

```
Version: 1.12.4
Check: installed package size
Result: NOTE
     installed size is 5.1Mb
     libs 1.5Mb
     tests 1.7Mb
Flavors: r-devel-linux-x86_64-fedora-clang, r-patched-solaris-x86
```

image

mattdowle:
Closes #3932 
By compressing a further 27 test data files, `inst/test` reduces from 1.696MB to 1.420MB,  a saving of 276KB.  Should cover the 100KB needed.   I tried both `gz` and `bz2` for each file and used the smallest for each (which was `gz` in a few cases).
To go further we could compress `*.Rraw` files. `bz2` would reduce `tests.Rraw` from 908K to 214K.  Maybe we could add a step to release procedures to compress `*.Rraw`.  Then decompress again for dev to continue conveniently. The compressed versions of `*.Rraw` need not be committed to `git`.  Or, better still, `R CMD build` could be made to compress the `*.Rraw` if that's possible.
Anyway, for now the 276KB saving should be enough for the CRAN check machines which check size with the `.so` including `-g`.

New status:
image


jangorecki:
This PR is nice, we compress data that we use for test. We don't need to have them versioned (or able to preview) in git.
I would strongly disagree on compressing Rraw because we will lose those capabilities for test script. Repo will be growing in size very fast because each single commit of zipped Rraw will be a brand new file, removed on next commit.
IMO the state to-be is to get more flexibility on package size from core R / CRAN. Something that allows package author to set within the package that he/she is aware about the size and it should not be a concern. Of course not just removing a limit, that would be dangerous, but something like `R_PKG_EXPECTED_SIZE_MB=8` env var.

MichaelChirico:
Agree CRAN could help us out here. I think we're being good communitarians by including tests in `inst` so end users have access to our full test suite to double-confirm on their machines the build is working as intended.

The alternative to remove the tests from the release (which would save us 1/3 of our space immediately) should be recognized as bad by all parties, but the marginal cost of shrinking the package each release continues to go up...

Recently I tried to do something like: run `test.data.table()` and save every `data.table()` object produced, to try and identify how much duplication there is in simple `data.table` objects being used for one-off tests (eventually goal being to turn this into a small set of `testing_datatables` which can be re-used on new tests to save the characters of creating the same object in different regions of tests). Quite labor intensive & the upshot would have saved only a little bit of space.

mattdowle:
@jangorecki
I'm frustrated that you don't appear to have read a whole paragraph I wrote up front:
> Maybe we could add a step to release procedures to compress *.Rraw. Then decompress again for dev to continue conveniently. The compressed versions of *.Rraw need not be committed to git. Or, better still, R CMD build could be made to compress the *.Rraw if that's possible.

@MichaelChirico 
> I think we're being good communitarians by including tests in inst so end users have access to our full test suite to double-confirm on their machines the build is working as intended.

I do too.  But some in the community disagree. They think we are circumventing CRAN policy which is that tests should not be installed by default. The official mechanism is `--install-tests` which is off by default. This has caused conflict to the extent I now avoid interaction on it. If there are simple things we can do that have no downside, then I take that approach. This saves communication, requests to CRAN maintainers, and conflict with other people in the community.

@jangorecki 
> Of course not just removing a limit, that would be dangerous, but something like R_PKG_EXPECTED_SIZE_MB=8 env var. @mattdowle
Maybe. But keep in mind the history I have described above. They don't agree that tests should be installed by default. How are you going to solve those politics?  We have a simple option that we can take now : compress *.Rraw in release procedures without committing to git.

PR #3942, opened 2019-10-05
fix NA filled by shift(nanotime, fill=NA)

### LINKED ISSUE #3945: shift(nanotime, fill=NA) does not fill nanotime NA
Calling `shift` on a `nanotime` does not identify the `nanotime` as inheriting from `bit64::integer64`, and thus fills the numeric `NA` instead of the `integer64`:

```
data.table::shift(nanotime::nanotime(1:4))
```
```
[1] "2262-02-18T20:47:17.227407266+00:00" "1970-01-01T00:00:00.000000001+00:00"
[3] "1970-01-01T00:00:00.000000002+00:00" "1970-01-01T00:00:00.000000003+00:00"
```

Fixed by https://github.com/Rdatatable/data.table/pull/3942:

```
[1] NA                                    "1970-01-01T00:00:00.000000001+00:00"
[3] "1970-01-01T00:00:00.000000002+00:00" "1970-01-01T00:00:00.000000003+00:00"
```

mschubmehl:
Closes #3945 

Currently `shift(nanotime(1:4))` incorrectly fills a `numeric` `NA` rather than an `integer64` `NA` into the result vector. This is because we incorrectly check that the class is exactly `integer64`, rather than inherited from it, as we do in most other contexts. This PR fixes that, and adds test cases for shift called on nanotimes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 ## BUG FIXES
 
+1. `shift()` on a `nanotime` with the default `fill=NA` now fills a `nanotime` missing value correctly, [#3942](https://github.com/Rdatatable/data.table/pull/3942). Thanks to @mschubmehl for the PR.
+
 ## NOTES
 
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 
 ## BUG FIXES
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
LGTM. Could file an issue as well that we could close with this? Easier to find it there if anyone else notices the issue before next release

mschubmehl:
Closes https://github.com/Rdatatable/data.table/issues/3945, and `NEWS.md` updated to reflect that. Thanks, all!

PR #3943, opened 2019-10-06
assign iris from scratch in tests

MichaelChirico:
Originally planned to remove `with=FALSE` usage from `subset.data.table` (see commits) but Jan points out this is not a good idea.

In the process I found this fix for `tests.Rraw` though so may as well merge that.


MichaelChirico:
Yes agreed. let me have another look.

MichaelChirico:
@jangorecki thanks for the review, on second thought I agree & reverted. But keeping the bit about `iris` in tests so not closing.

PR #3967, opened 2019-10-14
internal isRealReallyInt, closes #3966

### LINKED ISSUE #3966: isRealReallyInt could be callable from R
C function `isRealReallyInt` could be accessible from R, no need to export. Would be useful for some data validation.

jangorecki:
closes #3966


HughParsonage:
Should `isRealReallyInt(INT_MAX + 1)` be FALSE (or at least defined)?

jangorecki:
Thanks @HughParsonage, very good comment. That seems to be tricky problem, we either need to align to R's integer definition or extend it for integers bigger than 2.14e9. According to @mattdowle recent comments R's numeric can loselessly hold integers up to 2^52. I think it make sense for function in this PR so support those.

jangorecki:
Changing `(int)dx[i]` into `(int64_t)dx[i]` allows to check up to
```r
isRealReallyInt(9223372036854775295)
#[1] TRUE
isRealReallyInt(9223372036854775296)
#[1] FALSE
```
but such things are quite heavily platform specific

MichaelChirico:
> such things are quite heavily platform specific.

It's not great to have results depend on platform but I'm not sure it's avoidable, so I think the `int64_t` cast is fine. On a given platform, it answers the function's name -- is this double-stored value representable as integer?

jangorecki:
AFAIU platform specific is the range, so the maximum value that still returns TRUE.

HughParsonage:
Looking at its use in `between.c` and `frollR.c` the function seems to be used to determine whether the input is type `integer` not merely a whole number. So `isRealReallyInt((dbl)INT_MAX + 1)` should be `FALSE`.

mattdowle:
> such things are quite heavily platform specific

Perhaps true for `long double` but regular `double` in C is almost-surely IEEE754. For example it's 64bit even on 32bit.

PR #3975, opened 2019-10-17
update GLCI to use more recent win r-devel, which is 4.0 already, clo…

### LINKED ISSUE #3969: upgrade GLCI windows machine: R-devel to 4.0
Appveyor runs pretty recent windows R-devel checks.
On a windows machine running in GLCI we need to manually upgrade windows R-devel. This time because of R 3 to R 4 upgrade we also need to take care of references to a R-devel numbering.
At the moment in https://github.com/Rdatatable/data.table/blob/46cd3d38213527b8829854ff8cb83cb5e9a6ae8b/.gitlab-ci.yml there are three places that defines `3.7`, those need to be updated to `4.0` because a R-devel is a 4.0 numbering already. This is also important for publishing R-devel binaries that are located under `bin/windows/contrib/4.0/PACKAGES`.

jangorecki:
closes #3969
on our windows CI runner R-devel got upgraded to r77294 (2019-10-15).
"R-devel" directory now stores the new 4.0 rather than 3.7 version, thus any pipelines running now, before merging this PR might get in trouble.
This PR is already running GLCI here: https://gitlab.com/jangorecki/data.table/pipelines/89602324 once completed should be safe to merge.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -207,7 +207,7 @@ test-dev-win: # R-devel on windows
     TEST_DATA_TABLE_MEMTEST: "FALSE" # disabled as described in #3147
   allow_failure: false
   script:
-    - Rscript -e "source('.ci/ci.R'); cu<-contrib.url(getOption('repos'), ver=Sys.getenv('R_BIN_VERSION')); print(cu); print(rownames(available.packages(cu))); install.packages(dcf.dependencies('DESCRIPTION', which='all'), quiet=TRUE, contriburl=cu)"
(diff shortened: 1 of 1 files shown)


jangorecki:
actually I think we can run into troubles because cran seems to still use 3.7 as R-devel
https://cloud.r-project.org/bin/windows/contrib/3.7/PACKAGES
https://cloud.r-project.org/bin/windows/contrib/4.0/PACKAGES (404)

jangorecki:
Solution was to keep 3.7 env vars to match CRAN's urls for binaries.
At the same time recent Windows R-devel when install.packages has to be forced to point to 3.7 explicitly because by default it point to 4.0. This is global issue that affects all packages.

jangorecki:
This PR only enables support for recent R-devel, which tries to install from 4.0 url, but CRAN is still on 3.7. We keep our CRAN mirrors on 3.7 for consistency till CRAN will migrate to 4.0 url.
Should be passing now: https://gitlab.com/jangorecki/data.table/pipelines/89622870

PR #3978, opened 2019-10-18
minor documentation fixes (change \dots to ...)

mnazarov:
Just cleaning up remaining `\dots` artifacts in documentation.
As mentioned in https://github.com/Rdatatable/data.table/pull/3091#issuecomment-426681344 the `\dots` macros is not parsed when inside a comment inside R code.

### DIFF
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -10,13 +10,13 @@
 }
 \usage{
 # 1. LHS := RHS form
-# DT[i, LHS := RHS, by = \dots]
-# DT[i, c("LHS1", "LHS2") := list(RHS1, RHS2), by = \dots]
+# DT[i, LHS := RHS, by = ...]
+# DT[i, c("LHS1", "LHS2") := list(RHS1, RHS2), by = ...]
 
 # 2. Functional form
 # DT[i, `:=`(LHS1 = RHS1,
 #            LHS2 = RHS2,
-#            \dots), by = \dots]
+#            ...), by = ...]
 
 set(x, i = NULL, j, value)
 }
--- a/man/rbindlist.Rd
+++ b/man/rbindlist.Rd
@@ -8,7 +8,7 @@
 }
 \usage{
 rbindlist(l, use.names="check", fill=FALSE, idcol=NULL)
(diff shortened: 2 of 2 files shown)


mnazarov:
I _do_ see `\dots` on my system...

On further exploration, it seems to be an issue only if `help_type="html"`. That's what I use, and also can be seen e.g. here https://www.rdocumentation.org/packages/data.table/versions/1.12.4/topics/%3A%3D
However, both `help(":=", package = "data.table", help_type = "text")` and `help(":=", package = "data.table", help_type = "pdf")` render fine.

MichaelChirico:
The PR is good as there is indeed a problem but I'm a bit confused. Doesn't seem to be anything on this in WRE or in Parsing Rd files:

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Writing-R-documentation-files
https://developer.r-project.org/parseRd.pdf

The only thing there suggests that there should be no difference between the `\example{}` and `\usage{}` macros (both parsed as R-like).

The difference I see here is `\dots` comes in a comment, which must not be parsed as R...

MichaelChirico:
OK I have found something we can add to CRAN_release. Will add a commit before merging.

MichaelChirico:
@jangorecki I also see dots in `?rbindlist` btw, since `#rbind` is commented out.

MichaelChirico:
@mnazarov can you please put this work on a non-`master` branch? it's complicating things a bit.

mnazarov:
> @mnazarov can you please put this work on a non-`master` branch? it's complicating things a bit.

@MichaelChirico I see you already did this yourself, so feel free to close this PR and proceed with the other one (#3979).

> The PR is good as there is indeed a problem but I'm a bit confused. Doesn't seem to be anything on this in WRE or in Parsing Rd files:

I also found it puzzling especially the difference between html and other formats, could be a potential bug in R's `tools`, but I didn't dig that deep.

MichaelChirico:
I think maybe leave this one open and then merge sequentially. I'm worried if we squash&merge my follow-up your contribution might get buried (as all commits are combined into one) and I'd hate for you to get anything less than full credit for the PR!

mattdowle:
Great.  Thanks @mnazarov and all.

PR #3979, opened 2019-10-18
Rd dots

MichaelChirico:
Follow up to #3978 since I can't push to the fork's master branch


mattdowle:
Very nice check, @MichaelChirico!

PR #3984, opened 2019-10-21
Detect OpenMP support

### LINKED ISSUE #2161: [Request] Test for compiler compatibility with `-fopenmp` before using `SHLIB_OPENMP_CFLAGS`
This will allow installation of **data.table** on systems / compilers that do not support openmp, even if the site Makeconf file defines `SHLIB_OPENMP_CFLAGS`. This is the case on macOS as of R 3.4.0 with the default clang compiler.

Something like the following should be sufficient, note that `ifeq` is a GNU make extension, so it will need to be noted in your `DESCRIPTION`

```shell
CXX11=$(shell "${R_HOME}"/bin/R CMD config CXX11)
ifeq ($(shell $(CXX11) -fopenmp -E -xc++ - 2>&1 >/dev/null && echo 'true'), true)
  PKG_CFLAGS=$(SHLIB_OPENMP_CFLAGS)
  PKG_LIBS=$(SHLIB_OPENMP_CFLAGS)
endif

### PR DESCRIPTION
**Pull request by jimhester**
This detects support for OpenMP and only includes the compiler flags for
it if OpenMP is supported.

This solves the common case on macOS, because R on macOS ships with
`-fopenmp` in SHLIB_OPENMP_CFLAGS, as the CRAN compiler toolchain _does_
support OpenMP, however the XCode toolchain distributed by Apple _does
 not_ support OpenMP, which causes data.table to fail to compile from
 source.

Fixes #2161

### CODE REVIEW

File: .Rbuildignore [hunk 1]
```diff
@@ -15,6 +15,7 @@
 ^NEWS\.0\.md$
 ^README\.md$
 ^_pkgdown\.yml$
+^src/Makevars$
```
jimhester: We need to ignore the generated Makevars in both git and the built package, we generate it from `src/Makevars.in` in the `configure` script.

cleanup
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+rm -f src/Makevars
```
jimhester: We need to cleanup the generated file with the cleanup script

configure
```diff
@@ -50,5 +50,22 @@ fi
...
 echo "zlib ${version} is available ok"
-exit 0
 
+# Find R compilers
+CC=`${R_HOME}/bin/R CMD config CC`
+CFLAGS=`${R_HOME}/bin/R CMD config CFLAGS`
+CPPFLAGS=`${R_HOME}/bin/R CMD config CPPFLAGS`
+
+# Test if we have a OPENMP compatible compiler
+echo "#include <omp.h>\nint main () { return omp_get_num_threads (); }" | ${CC} ${CPPFLAGS} ${PKG_CFLAGS} ${CFLAGS} ${SHLIB_OPENMP_CFLAGS} -E -xc - >/dev/null 2>&1 || R_NO_OPENMP=1;
```
jimhester: Here we test a simple OpenMP program with what is in `SHLIB_OPENMP_CFLAGS` to verify that our current compiler supports OpenMP.

src/Makevars.win
```diff
@@ -0,0 +1,5 @@
+PKG_CFLAGS = $(SHLIB_OPENMP_CFLAGS)
```
jimhester: `configure` will only run on non-Windows, which means that `src/Makevars` will not be generated there, so we need to supply `Makevars.win` for windows. But we don't really need to check if OpenMP is supported, because the Rtools compilers all support it.

configure
```diff
@@ -50,5 +50,22 @@ fi
...
+
+# Write to Makevars
+if [ $R_NO_OPENMP ]; then
+  sed -e "s|@openmp_cflags@||" src/Makevars.in > src/Makevars
+else
+  echo "OpenMP supported"
+  sed -e "s|@openmp_cflags@|\$(SHLIB_OPENMP_CFLAGS)|" src/Makevars.in > src/Makevars
```
jimhester: Then we replace the placeholder `@openmp_cflags@` with either nothing if the compiler doesn't support it or `SHLIB_OPENMP_CFLAGS` if it does.

configure
```diff
@@ -50,5 +50,22 @@ fi
...
+CC=`${R_HOME}/bin/R CMD config CC`
+CFLAGS=`${R_HOME}/bin/R CMD config CFLAGS`
+CPPFLAGS=`${R_HOME}/bin/R CMD config CPPFLAGS`
+
+# Test if we have a OPENMP compatible compiler
+echo "#include <omp.h>\nint main () { return omp_get_num_threads (); }" | ${CC} ${CPPFLAGS} ${PKG_CFLAGS} ${CFLAGS} ${SHLIB_OPENMP_CFLAGS} -E -xc - >/dev/null 2>&1 || R_NO_OPENMP=1;
+
+# Write to Makevars
+if [ $R_NO_OPENMP ]; then
+  sed -e "s|@openmp_cflags@||" src/Makevars.in > src/Makevars
```
MichaelChirico: Can we add a message here indicating openMP wasn't detected? I think we want to be quite loud to the user that they should try and install openMP, perhaps linking to the installation wiki for how to do so.

In that sense the current behavior was nice, since we prefer users to be running with openMP support & the failure would force that (just that the error was not helpful/instructive)

jimhester: Sure, do you have a suggestion on what specifically it should say?

jimhester:
@MichaelChirico I added a message linking to the wiki in 576c852, let me know if there is anything else I can do.

configure
MichaelChirico: thanks Jim, have adjusted, will give Matt final editorial cut

mattdowle:
Thanks, @jimhester! This looks great. The message is great to see on installation. There is a package startup message too if it detects it wasn't compiled at that point.
I've invited you to be project member. You'll need to accept the invite that shows in your GitHub profile or GitHub repository which I'm sure you know but for completeness just in case. I'll move the news item up and add you in the contributors list in a follow up commit to master. (I could commit to the fork but since the PR is passing and ready and green to merge, it's faster for me this way.)

PR #3986, opened 2019-10-22
Finish translation to zh_CN assigned to group 2 (Hongyuan Jia)

hongyuanjia:
## Pull request overview

* This pull request contains translation assigned to group 2(@hongyuanjia). Specifically:
    - L89-169 in `po/R-zh_CN/po`. 25 messages in total.
    - L170-311 in `po/zh_CN.po`. 24 messages in total.


MichaelChirico:
Awesome! Please review @dracodoc

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 msgid "x."
 msgstr ""
 
 msgid "is type"
-msgstr ""
+msgstr "的类型为"
 
 msgid "which is not supported by data.table join"
-msgstr ""
+msgstr "，该类型无法用于 data.table 的联结"
```
dracodoc: It's difficult to translate such a simple and common word `join`. I found the most common translation is 连接 which is not really best. 联结 is better, another option is 联接. Many Chinese documentations actually just use `join` since it's a very simple word.

In any case, I think we should include the original word with (join). As for which word to use, I vote for 联接, but 连接 is acceptable if it's already established (but it's the worst in 3 options).

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 
 msgid "i."
 msgstr ""
 
 msgid "Attempting roll join on factor column when joining x."
-msgstr ""
+msgstr "试图针对数据类型为因子（factor）的列使用滚动联结（roll join），发生在将 x."
```
dracodoc: 试图滚动联接（roll join）因子类型(factor)的列，这发生于将 x.

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 
 msgid "to i."
-msgstr ""
+msgstr "与 i."
 
 msgid ". Only integer, double or character columns may be roll joined."
-msgstr ""
+msgstr "联结时。仅整（integer）型、数值（double）型或字符串（character）型可用于滚动联结（roll join）。"
```
dracodoc: 联接时。但只有整数（integer）、双精度（double）或字符（character）类型的列可以使用滚动联接（roll join）。

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 
 msgid "Incompatible join types: x."
 msgstr ""
```
dracodoc: This one is missed?

联接（join）类型不兼容：x.

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 msgid "RHS has length()"
-msgstr ""
+msgstr "右手侧（RHS）的长度为"
 
 msgid "; expecting length 2."
-msgstr ""
+msgstr "；其长度应为 2。"
 
 msgid "c"
 msgstr ""
```
dracodoc: if there is no translation needed, do we need to repeat msg.id in msg.str? Like

msgid "c"
msgstr "c"

po/R-zh_CN.po
MichaelChirico: I'm quite surprised `join` nor `merg` really show up anywhere in the base translation files 😳 so we don't really have any canon to refer to within R...

FWIW, Chinese Wikipedia is actually using 连接 for SQL join:

https://zh.wikipedia.org/wiki/%E8%BF%9E%E6%8E%A5

po/R-zh_CN.po
MichaelChirico: no, if `msgstr` is empty, simply no translation is done. Some people have been including the duplicate translation, others not... I don't have strong feelings one way or the other

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
-msgstr ""
+msgstr "class 的长度必须为 1"
 
 msgid "between has been passed an argument x of type logical"
-msgstr ""
+msgstr "传入 between 的参数 x 为逻辑（logical）型"
 
 msgid ""
 "'between' function the 'x' argument is a POSIX class while 'lower' was not, "
 "coercion to POSIX failed with:"
```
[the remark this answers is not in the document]
hongyuanjia: Great resource. Will change them to `自动转换` accordingly.

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
+msgstr "。将采用 UTC 时间进行比较。"
 
 msgid "trying to use integer64 class when 'bit64' package is not installed"
-msgstr ""
+msgstr "试图使用 intger64 类型但 'bit64' 包尚未安装"
 
 msgid ""
 "Not yet implemented NAbounds=TRUE for this non-numeric and non-character type"
 msgstr ""
+"将 NAbounds=TRUE 应用于非数值（numeric）型和非字符串（character）型的功能尚未开发"
```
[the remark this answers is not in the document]
hongyuanjia: Nice catch. Agree.

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 
 msgid "Some lower>upper for this non-numeric and non-character type"
-msgstr ""
+msgstr "对于该非数值（numeric）型和非字符串（character）型的输入，存在一部分下界（lower）> 上界（upper）的情况"
```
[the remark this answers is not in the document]
hongyuanjia: Agree.

po/R-zh_CN.po
hongyuanjia: Nice review.

po/R-zh_CN.po
hongyuanjia: Oh my original assignment ended at line 169. Probably better to leave line 172 untouched to avoid merge conflict.

po/R-zh_CN.po
dracodoc: Yes 连接 is a more common usage. However it only have the meaning of `linking`, I would think it just means you `cbind` two things together and don't have the part of matching values, which both `联结` and `联接` provided.

Between `联结` and `联接`, `联结` is hinting a strong binding that hard to break, more like the ending state, `联接` have the process of connecting two things together so I think it's better.

po/R-zh_CN.po
```diff
@@ -87,87 +87,95 @@ msgid "; filled with NA"
...
 msgid "; expecting length 2."
-msgstr ""
+msgstr "；其长度应为 2。"
 
 msgid "c"
 msgstr ""
 
 msgid "Perhaps you meant %s?"
-msgstr ""
+msgstr "或许你指的是 %s？"
```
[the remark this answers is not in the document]
hongyuanjia: Better.

po/R-zh_CN.po
hongyuanjia: Agree.

po/R-zh_CN.po
MichaelChirico: In that sense, maybe 连接 is closer to `join`, which also doesn't have any "intrinsic" sense of matching to me; `merge` gets closer to that.

That said, probably 联接 is preferred since it's more important for users to understand the meaning rather than to match the English. In English, I know we are implicitly bootstrapping on users' understanding of `join`s in the relational database sense which is well known. I leave it up to you to know how well-known the different usages are :)

po/zh_CN.po
```diff
@@ -172,67 +172,81 @@ msgid ""
...
 msgstr ""
+"j 为 '%s' 型。j 必须为整（integer）型、字符串（character）型、或数值（numeric）型（将被强制转换成整型"
+"并发出警告）。"
 
 #: assign.c:377
 msgid ""
 "Can't assign to the same column twice in the same query (duplicates "
 "detected)."
 msgstr ""
+"在一次查询语句中无法对同一列赋值两次（检测出
```
dracodoc: 在一次查询中无法对同一列赋值两次（检测出
重复项）。

po/zh_CN.po
```diff
@@ -172,67 +172,81 @@ msgid ""
...
+msgstr "RHS_list_of_columns == %s\n"
 
 #: assign.c:385
 #, c-format
 msgid ""
 "RHS_list_of_columns revised to true because RHS list has 1 item which is "
 "NULL, or whose length %d is either 1 or targetlen (%d). Please unwrap RHS.\n"
 msgstr ""
+"修改 RHS_list_of_columns 为真因右手侧（RHS）列表存在 1 "
+"为空值（NULL），或长度 %d 为 1 或 targetlen（%d）的项。请拆开右手侧。\n"
```
dracodoc: RHS_list_of_columns改为真（True），因为右手侧列表（RHS list）有一子项为空（NULL）或其长度 %d 为1或 targetlen（%d）。请拆开右手侧。\n

po/zh_CN.po
```diff
@@ -241,17 +255,21 @@ msgid ""
 "table or data.frame since they are lists too). To delete multiple columns "
 "use NULL instead. To add multiple empty list columns, use list(list())."
 msgstr ""
+"给定的 %d 列将被赋值为空列表（可能为空数据、"
```
dracodoc: 提供的%d 列

po/zh_CN.po
```diff
@@ -241,17 +255,21 @@ msgid ""
...
+"表（table）或数据框（data.frame），因后两者也是列表的一种）。删除多个列时"
+"请使用空值（NULL）。添加多个空列表列（list columns）时，请使用 list(list())。"
 
 #: assign.c:395
 #, c-format
 msgid "Recycling single RHS list item across %d columns. Please unwrap RHS.\n"
-msgstr ""
+msgstr "右手侧（RHS）列表项循环应用于 %d 列。请拆开右手侧。\n"
```
dracodoc: 回收重用单个右手侧（RHS）列表子项于 %d 列。 请拆开右手侧。\n

po/zh_CN.po
```diff
@@ -241,17 +255,21 @@ msgid ""
...
 
 #: assign.c:397
 #, c-format
 msgid ""
 "Supplied %d columns to be assigned %d items. Please see NEWS for v1.12.2."
 msgstr ""
+"指定 %d 列赋值给 %d 项。请阅读 v1.12.2 的更新信息（NEWS）。"
```
dracodoc: I think the assignment direction here need to be reversed.

提供的 %d 列将被赋值为  %d 项。请阅读 v1.12.2 的更新信息（NEWS）。

po/zh_CN.po
```diff
@@ -260,17 +278,22 @@ msgid ""
 "set() on a data.frame is for changing existing columns, not adding new ones. "
 "Please use a data.table for that."
 msgstr ""
+"j 中第 %d 项的列编号为 %d，超出了有效范围 [1,ncol=%d]。"
```
dracodoc: j中的列编号里第 %d 项是 %d

po/zh_CN.po
```diff
@@ -282,29 +305,38 @@ msgid ""
 "full length RHS vector such as vector('list',nrow(DT)); i.e., 'plonk' in the "
 "new column."
 msgstr ""
+"赋值语句的右手侧（RHS）指向列 '%s'，其长度为零但并非空值（NULL）。如果"
+"想要删除列请使用空值（NULL）。否则，右手侧必须"
+"长度 > 0，如 NA_integer_。如果想将该列类型修改为"
+"空列表列（list column），正如所有修改列类型的操作一样，请在右手侧提供一个"
+"与该列原数据等长的向量，如 vector('list',nrow(DT))，即，用新数据替换（plonk）重新生成"
+"该列。"
```
dracodoc: 对已有列 '%s' 的赋值其右手侧（RHS）长度为零但并非空值（NULL）
...
如 vector('list',nrow(DT)); 也即用替代品生成新列。

plonk means cheap wine? I think it means generic and fake here but we cannot use too negative words.

po/zh_CN.po
```diff
@@ -282,29 +305,38 @@ msgid ""
...
+msgstr "':=' 右手侧（RHS）  %d 列矩阵将被视为一维向量"
 
 #: assign.c:434
 #, c-format
 msgid ""
 "Can't assign to column '%s' (type 'factor') a value of type '%s' (not "
 "character, factor, integer or numeric)"
 msgstr ""
+"无法赋值给列 '%s'（类型 '因子'（'factor'））一个类型为 '%s' 的值（非
+字符串（character）型、因子（factor）型、整（integer）型或者数值（numeric）型）"
```
dracodoc: 无法给因子（factor）类型列 '%s' 赋类型为 '%s' 的值（不是字符（character），因子（factor），整数（integer），数值（numeric）类中的一种）

po/zh_CN.po
```diff
@@ -313,6 +345,9 @@ msgid ""
 "'recycle' the RHS please use rep() to make this intent clear to readers of "
 "your code."
 msgstr ""
+"试图将 %d 项赋值给 %d 列 '%s'。如果想"
+"'循环'（'recycle'）右手侧，请使用 rep() 以将该意图清晰地表述给阅读"
+"代码的人。"
```
dracodoc: I added a () here to keep the order correct and make the meaning clear.
试图将 %d 项赋值给 %d 项（列 '%s'）

回收重用（recycle）

po/zh_CN.po
```diff
@@ -3070,48 +3109,48 @@ msgstr ""
...
 "batchSize=%d  lastBatchSize=%d\n"
-msgstr "�ڲ�����nrow=%d ngrp=%d  nbit=%d  shift=%d  highSize=%d  nBatch=%d  "
+msgstr "内部错误：nrow=%d ngrp=%d  nbit=%d  shift=%d  highSize=%d  nBatch=%d  "
 "batchSize=%d  lastBatchSize=%d\n"
 
 #: gsumm.c:98
 #, c-format
 msgid "gforce initial population of grp took %.3f\n"
-msgstr "grp �� gforce ��ʼȺ����ռ�� %.3f\n "
+msgstr "grp 的 gforce 初始群体数占了 %.3f\n "
```
dracodoc: grp 的 gforce 初始群体占了 %.3f\n

po/zh_CN.po
```diff
@@ -3120,133 +3159,133 @@ msgstr ""
 
 #: gsumm.c:345
 msgid "sum is not meaningful for factors."
-msgstr "���ӵĺ�û�����塣"
+msgstr "因子的和没有意义。"
```
dracodoc: 对因子求和没有意义

po/zh_CN.po
```diff
@@ -3120,133 +3159,133 @@ msgstr ""
...
+"设置 options(datatable.optimize=1) 关闭 GForce 优化"
 
 #: gsumm.c:578
 msgid ""
 "GForce mean can only be applied to columns, not .SD or similar. Likely "
 "you're looking for 'DT[,lapply(.SD,mean),by=,.SDcols=]'. See ?data.table."
-msgstr "GForce ��ƽ��ֵ���������У��������� .SD ������"
-"�������ڲ��� 'DT[,lapply(.SD,mean),by=,.SDcols=]'���μ� ?data.table ��"
+msgstr "GForce 求平均值仅适用于列，而不能用于 .SD 或其它。"
+"或许你在查找 'DT[,lapply(.SD,mean),by=,.SDcols=]'。参见 ?data.table 。"
```
dracodoc: 或许你想用的是

po/zh_CN.po
```diff
@@ -3120,133 +3159,133 @@ msgstr ""
...
 #: gsumm.c:650
 #, c-format
 msgid ""
 "Type '%s' not supported by GForce mean (gmean) na.rm=TRUE. Either add the "
 "prefix base::mean(.) or turn off GForce optimization using options(datatable."
 "optimize=1)"
-msgstr "GForce ���ֵ��gmean����֧������ '%s'����ʹ�� base::mean(.) ����"
-"���� options(datatable.optimize=1) �ر� GForce �Ż�"
+msgstr "GForce 求均值（gmean）不支持类型 '%s'。请使用 base::mean(.) 或者"
+"设置 options(datatable.optimize=1) 关闭 GForce 优化"
```
dracodoc: Maybe we need to split the sentence at the exactly place here?
"GForce 求均值（gmean）不支持类型 '%s'。请加上"
" base::mean(.) 前缀或者设置 options(datatable.optimize=1) 关闭 GForce 优化"

po/zh_CN.po
```diff
@@ -3120,133 +3159,133 @@ msgstr ""
...
-msgstr "gmin �� nrow [%d] != length(x) [%d]"
+msgstr "gmin 中 nrow [%d] != length(x) [%d]"
 
 #: gsumm.c:717 gsumm.c:850
 msgid ""
 "No non-missing values found in at least one group. Coercing to numeric type "
 "and returning 'Inf' for such groups to be consistent with base"
-msgstr "������һ�������û�з���ȱʧֵ��"
-"Ϊ���� base ����һ�£�����Щ��ǿ��ת��Ϊ��ֵ���Ͳ����� ��Inf����"
+msgstr "在至少一个分组中没有发现缺失值。"
```
dracodoc: 在至少一个分组中所有值都是缺失的

po/zh_CN.po
```diff
@@ -3120,133 +3159,133 @@ msgstr ""
...
+"为了与 base 保持一致，将这些组强制转换为数值类型并返回 ‘Inf’。"
 
 #: gsumm.c:756 gsumm.c:891
 msgid ""
 "No non-missing values found in at least one group. Returning 'NA' for such "
 "groups to be consistent with base"
-msgstr "������һ�������û�з���ȱʧֵ��"
-"Ϊ���� base ����һ�£���Щ�齫���� ��NA����"
+msgstr "在至少一个分组中没有发现缺失值。"
```
dracodoc: Same with line 3259, meaning reversed here.

po/zh_CN.po
```diff
@@ -3120,133 +3159,133 @@ msgstr ""
...
-msgstr "������һ�������û�з���ȱʧֵ��"
-"Ϊ���� base ����һ�£���Щ�齫���� ��Inf����"
+msgstr "在至少一个分组中没有发现缺失值。"
+"为了与 base 保持一致，这些组将返回 ‘Inf’。"
 
 
 #: gsumm.c:791
 msgid "Type 'complex' has no well-defined min"
-msgstr "�����ܱȽϴ�С����û����Сֵ"
+msgstr "复数不能比较大小，故没有最小值"
```
dracodoc: 复数没有精确意义上的最小值

po/zh_CN.po
MichaelChirico: Have added an FAQ about the possibility to reverse the order of the `%d` if you think it would be more natural; this is especially easy for C messages which are already in "printf"-style format : 

https://github.com/wch/r-source/blob/trunk/src/library/base/po/zh_CN.po#L1259-L1262

TL;DR is to use `%1$d` to refer to the `%d columns` input and `%2$d` to refer to the `%d items` input

po/zh_CN.po
```diff
@@ -172,67 +172,81 @@ msgid ""
 "a logical subset, simply wrap with which(), and take the which() outside the "
 "loop if possible for efficiency."
 msgstr ""
+"i 为 '%s' 类型。i 必须为整（integer）型，或数值（numeric）型（将被强制转换成整型并发出警告）。如果 i 为"
+"一逻辑向量（logical）的子集，请直接将它传给 which()，且如果可能的话将 which() 放置于"
```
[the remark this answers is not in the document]
MichaelChirico: I think this would mainly come up for something like

```
iris = as.data.table(iris)
set(iris, rep(TRUE, nrow(iris)), 'new_col', rep(5L, nrow(iris)))
```

This should be replaced by

```
set(iris, which(rep(TRUE, nrow(iris))), 'new_col', rep(5L, nrow(iris)))
```

The latter part of the message refers to if `set` is being used in a loop:

```
for (jj in 1:100) set(iris, which(rep(TRUE, nrow(iris))), sprintf("V%d", jj), rnorm(nrow(iris)))
```

Will be more efficient with:

```
idx <- which(rep(TRUE, nrow(iris)))
for (jj in 1:100) set(iris, idx, sprintf("V%d", jj), rnorm(nrow(iris)))
```

Since we don't have to run `which` every iteration.

po/zh_CN.po
MichaelChirico: Different definition of `plonk` 😅 

> set down heavily or carelessly.
> "she plonked her glass on the table"
> sit down heavily and without ceremony.
> "he plonked himself down on the sofa"

po/zh_CN.po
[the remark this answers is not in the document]
hongyuanjia: Previously I thought this message came from cases when input of `i` is not the same length of `nrow(dt)` but shorter, and `which(i)` will change it into a row index to make `set` work. So `a logical subet` would mean `a subset of a longer logical vector `.

Now I know the full context. Thanks.

`一个用于筛选的逻辑（logical）向量 ` is definitely a better and more precise one.

po/zh_CN.po
```diff
@@ -172,67 +172,81 @@ msgid ""
...
 #: assign.c:342
 #, c-format
 msgid "Added %d new column%s initialized with all-NA\n"
-msgstr ""
+msgstr "添加了 %d 个新列 %s 并全部初始化为 NA \n"
 
 #: assign.c:347
 msgid "length(LHS)==0; no columns to delete or assign RHS to."
-msgstr ""
+msgstr "length(LHS)==0；未删除任何列或将右手侧（RHS）赋值给任何列。"
```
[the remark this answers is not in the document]
hongyuanjia: This one bothered me a lot. Did not find a concise way to translate `no columns to delete`. Very beautiful translation!

po/zh_CN.po
```diff
@@ -172,67 +172,81 @@ msgid ""
...
 
 #: assign.c:361
 msgid ""
 "set() on a data.frame is for changing existing columns, not adding new ones. "
 "Please use a data.table for that. data.table's are over-allocated and don't "
 "shallow copy."
 msgstr ""
+"set() 用于数据框（data.frame）是为了修改现有列，而非添加新列。"
+"请使用 data.table 来完成上述操作。data.table 采用超额分配（over-allocated）并且不"
+"进行浅拷贝（shallow copy）。"
```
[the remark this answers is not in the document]
hongyuanjia: Better!

po/zh_CN.po
hongyuanjia: I also considered simply using`查询` not `查询语句`, but thought maybe the latter would provide more context to the user. Could you explain what `query` means here? Does it mean the action to be conducted using `set`? If this is the case, is there any better word to describe this?

po/zh_CN.po
hongyuanjia: Nice and better!

po/zh_CN.po
hongyuanjia: Agree.

po/zh_CN.po
dracodoc: I think data.table design borrowed a lot from database queries. Thus a `query` is the step happen in `i` which filter the data, then `j` do the calculation (assignment). Since there is no explicit query sentence in the data.table syntax (like a SQL query), I think it's better to use 查询 not 查询语句.

po/zh_CN.po
hongyuanjia: I searched the web and tried to find an existing translation of `recycle` in programming but failed. `回收重用` is definitely way better than `循环`. There are also other places I need to change accordingly.

po/zh_CN.po
hongyuanjia: @MichaelChirico, being able to change the order of `%d` is great!

This error message comes from the case when `j` Does not have the same length as `values` in `set`. `提供的 %d 列将被赋值为 %d 项 ` sounds like the assignment is to be performed which it won’t.

@dracodoc , do you think `试图将 %2$d 项赋值给 %1$d 列` is an acceptable one? I add `试图` here to indicate this is a failure.

po/zh_CN.po
hongyuanjia: Looking at `query` as the step in `i` indeed make more sense. I agree `查询` is a better one here.

po/zh_CN.po
dracodoc: In the beginning I though this might be a warning but still proceeded. If the NEWS reference is about No.1 in [NEWS](https://github.com/Rdatatable/data.table/blob/master/NEWS.md#new-features-2), then I agree 试图 is a better translation as the attempt failed.

po/zh_CN.po
hongyuanjia: Nice and better! Do you think it is better to use `、` instead of `，`, since they are equal important here.

po/zh_CN.po
dracodoc: I need to correct myself. The whole command is a query, some part do the filtering in i, some part do some manipulation in j. This is the same in SQL as you can select some and create new with functions. 查询 is still better than 查询语句 because for a database query we usually don't need to translate it as 查询语句.

po/zh_CN.po
hongyuanjia: Yes, this refers to the NEWS you referenced. It changed from warning to error in [this](https://github.com/Rdatatable/data.table/pull/3310) pull request.

po/zh_CN.po
dracodoc: I think the difference between `、` and `，` is really minor so either one is OK. I just used the same one in the English sentence.

po/zh_CN.po
hongyuanjia: Thanks for explaining this!

po/zh_CN.po
hongyuanjia: After the discussion below, I think this translation should also be revised to reflect it is a failed attempt.

So this one can be:

`试图将 %d 列赋值为空列表...`

@dracodoc, any suggestions?

po/zh_CN.po
hongyuanjia: Agree.

po/zh_CN.po
dracodoc: I agree the revision is better.

po/zh_CN.po
hongyuanjia: Line 311 and below are outside of my assignment. Did not know why there are diffs here because I did not touch them at all. Will these comments be responded by other  collaborators and I can safely ignore review comments below?

dracodoc: Maybe the difference of encoding or different CR/LF between windows and linux. Please ignore them. Though we do need some better approach to avoid this kind of situation, otherwise if you merge directly that will change their parts.
@MichaelChirico This may need your attention.

po/zh_CN.po
hongyuanjia: Nice and better!

po/zh_CN.po
hongyuanjia: I get a little bit confused here. Despite the final translation of word `plonk`, `plonk in the new column` refers to the action of recreating the same column, let’s say `col`, using `set(DT, NULL, 'col', value)`, or creating a new column with different name than `col` using `set(DT, NULL, 'newcol', value)`. I think it refers to the first one. If so, the `new column` is **new** because the contents in that column is new. That’s why I use `用新数据替换重新生成该列`. Because from the user’s view, no new column is created but the existing column is replaced by a list of empty lists.

Any insights? @dracodoc

dracodoc: I'm not sure about `plonk` meaning now. Previously I thought that means using an empty but similar structured copy to replace, now I'm not sure. For the intention I agree with you it's using new data to replace existing ones.

po/zh_CN.po
MichaelChirico: @dracodoc is right, we're using it a bit allusively here.

To me, in `data.table`, `query` == any call to `[.data.table`

po/zh_CN.po
MichaelChirico: in R, recycling is happening here:

```
data.frame(a = 1, b = 1:10) # 1 is recycled 10 times to make column a
matrix(1:3, nrow = 4, ncol = 12) # <- (1:3) is recycled 16 times to make the matrix
```

You can trigger a warning about recycling in base like:

```
sprintf("%s %d", letters, 1:3)
```

```
Error in sprintf("%s %d", letters, 1:3) : 
  arguments cannot be recycled to the same length
```

Chinese version here skirts the issue by mentioning mismatched length; elsewhere, base is also using `回收`

po/zh_CN.po
MichaelChirico: You can trigger this error like so:

```
iris = as.data.table(iris)
set(iris, , c('x', 'y'), list(1, 2, 3))
```
```
Error in set(iris, , c("x", "y"), list(1, 2, 3)) : 
  Supplied 2 columns to be assigned 3 items. Please see NEWS for v1.12.2.
```

po/zh_CN.po
MichaelChirico: @mattdowle maybe we should add an FAQ about the meaning of `plonk` 😅 

maybe these videos can be a good definition:

https://www.youtube.com/watch?v=zoaGwyYH3as

https://www.youtube.com/watch?v=anJugOiu1jU

po/zh_CN.po
MichaelChirico: Don't worry about this. If we think it's important, I can to a sweep-through at the very end using `sed` to replace punctuation.

po/zh_CN.po
MichaelChirico: Yes this is from Chun-Hui's section. There must be some encoding issue... hard to tell from the web GUI.

I don't really know a good solution. Will keep an eye on what happens upon merging this...

hongyuanjia:
@MichaelChirico @dracodoc I have resolved all review comments expect the `plonk` part which is left as my first commit. Please review and comments if there are still issues. Thanks.

MichaelChirico:
LGTM! @dracodoc any last comments?

dracodoc:
LGTM too! Thanks!

hongyuanjia:
Super excited! Thanks @MichaelChirico @dracodoc

PR #3989, opened 2019-10-23
Translation yilei

Zachary-Wu:


MichaelChirico:
Excellent, thanks Yilei! Please help to review @shawnchen1996

Zachary-Wu:
OK, no problem. but it may cost another week. I am a little busy now, cause
you know I have to report my lab-work next Friday.

MichaelChirico:
@Zachary-Wu oh, sorry, I meant that @ShawnChen1996 should review this PR. Your review task will be for Jianda Shi (not yet filed).

In any case, there will be at least 3 more weeks before I hope to wrap up all the PRs. Fret not!

Zachary-Wu:
O, it is my fault. I am a little confused by @.    O(∩_∩)O~

po/R-zh_CN.po
```diff
@@ -733,79 +733,82 @@ msgid "'subset' must evaluate to logical"
...
-msgstr ""
+msgstr "x 不是 data.table 或 data.frame。"
 
 msgid "x has"
-msgstr ""
+msgstr "x 有"
 
 msgid "columns but its names are length"
-msgstr ""
+msgstr "很多纵列，但纵列的名字是"
```
ShawnChen1996: May need context to decide what it really means

po/R-zh_CN.po
```diff
@@ -733,79 +733,82 @@ msgid "'subset' must evaluate to logical"
...
 
 msgid "Passed a vector of type '"
-msgstr ""
+msgstr "传递了类型向量的长度值；"
```
ShawnChen1996: should be "传递了一个向量，元素类型是"

po/R-zh_CN.po
```diff
@@ -733,79 +733,82 @@ msgid "'subset' must evaluate to logical"
...
 
 msgid "'. Needs to be type 'character'."
-msgstr ""
+msgstr "需要按 'character'类型"
```
ShawnChen1996: . 需要是‘character’ 类型

po/R-zh_CN.po
MichaelChirico: The code generating this is from `setnames`:

```
if (length(names(x)) != ncol) stop("x has ",ncol," columns but its names are length ",length(names(x)))
```

This error can be triggered like

```
DT = list(1, 2)
setattr(DT, 'class', class(data.table()))
setnames(DT, c('a', 'b'))
```

```
Error in setnames(DT, c("a", "b")) : 
  x has 2 columns but its names are length 0
```

ShawnChen1996:
```
msgstr "列，但其列名的长度为"
```

po/zh_CN.po
```diff
@@ -1269,201 +1269,209 @@ msgid ""
...
 #: fmelt.c:662
 #, c-format
 msgid "Unknown column type '%s' for column '%s' in 'data'"
-msgstr ""
+msgstr "'data' 中的'%s'列是未知列类型"
 
 #: fmelt.c:673
 msgid "Input is not of type VECSXP, expected a data.table, data.frame or list"
-msgstr ""
+msgstr "输入类型不能是 VECSXP，输入类型应该是 data.table," "data.frame 或 list"
```
ShawnChen1996: I think it should be "输入类型不是 VECSXP，输入类型应该是 data.table," "data.frame 或 list"
I suppose type VECSXP is a base class of data.table, data.frame or list

po/zh_CN.po
```diff
@@ -1269,201 +1269,209 @@ msgid ""
...
 #: fmelt.c:687
 msgid "names(data) is NULL. Please report to data.table-help"
-msgstr ""
+msgstr "names(data)为NULL，请向 data.table-help 报告"
 
 #: forder.c:106
 #, c-format
 msgid "Failed to realloc thread private group size buffer to %d*4bytes"
-msgstr ""
+msgstr "无法重新将线程私有的组大小缓冲区分割为%d*4字节"
```
ShawnChen1996: I am not sure if the following is better
 "无法将线程私有的组大小缓冲区重新分配为%d*4字节"

po/zh_CN.po
```diff
@@ -1269,201 +1269,209 @@ msgid ""
...
 
 #: forder.c:120
 #, c-format
 msgid "Failed to realloc group size result to %d*4bytes"
-msgstr ""
+msgstr "由于%d*4字节，重新分配组大小失败"
```
ShawnChen1996: I am a bit confused by this msg. is it that the `result` should be %d *4 bytes?

po/zh_CN.po
MichaelChirico: The suggested replacement is the same as the original? Sorry I'm not seeing a difference.

FWIW `VECSXP` is the C-level storage type of `list` objects in R (and recall in terms of data storage, `data.frame` and `data.table` are both just special types of `list`s)

po/zh_CN.po
MichaelChirico: When grouping is started, we take a guess at how much storage will be needed. Sometimes, we'll need more than this and have to re-allocate memory to continue; any time memory allocation is done, there's a chance the computer has no memory left and the allocation request will fail.

This error is emitted in one such case -- we're reporting to the user that we tried to `realloc` `%d*4` bytes & failed. `%d*4` is the _total_ target amount to be allocated.

po/zh_CN.po
ShawnChen1996: One character difference between my suggestion and the original. 
"不能是VECSXP" and "不是VECSXP". 
The former one means "the input can' t be VECSXP". My replacement is "The input is not a VECSXP".

po/zh_CN.po
ShawnChen1996: Then, my suggestion is "分配%d*4字节内存时失败"

MichaelChirico:
Hey @Zachary-Wu please help to incorporate the suggested improvements when you get a chance. Thanks again!

Zachary-Wu:
I think @ShawnChen1996's translation is more accurate.

MichaelChirico:
Thanks @Zachary-Wu and @ShawnChen1996 !!

PR #3991, opened 2019-10-24
.SDcols accepts a function to filter on values

### LINKED ISSUE #3950: .SDcols=function shorthand
Much like `.SDcols = patterns` is useful as a shorthand for filtering columns and much like `setnames` now accepts functions, `.SDcols=function` is (1) currently an error and (2) should translate pretty naturally as a shorthand for filtering in `.SDcols`.

e.g.

```
DT[ , lapply(.SD, sum), .SDcols = is.numeric]
```

Should be easy to apply `!` as well.

MichaelChirico:
Closes #3950


NEWS.md
```diff
@@ -65,6 +65,9 @@ unit = "s")
 # data.table::fcase   1.49  1.55  1.67     1.71  1.73   1.86     5
 ```
 
+7. `.SDcols=is.numeric` now works; i.e., `SDcols=` accepts a function which is used to select the columns of `.SD`, [#3950](https://github.com/Rdatatable/data.table/issues/3950). Any function (even _ad hoc_) that returns scalar `TRUE`/`FALSE` for each column will do; e.g., `.SDcols=!is.character` will return _non_-character columns (_a la_ `Negate()`). Note that `patterns=` can still be used for filtering based on the column names.
```
renkun-ken, line 68: `patterns=` looks a bit misleading to me, looks like an argument of `[`, e.g. `dt[i, lapply(.SD, mean), patterns="^x"]`.

mattdowle: Yep, thanks. I filed to #4642 to fix that.

PR #3994, opened 2019-10-25
Avoid `frollapply()` stack overflows and protect limits

### LINKED ISSUE #3993: frollapply() issues
Hi all, I was reading the source for `frollapply()`, and saw two things that I thought might be problematic.

The first is a protection stack overflow. I think that looping here and constructing a SEXP for both the window and the call for each `k` is causing this issue. You could probably restructure this loop to be inside the double loop below it (which you might also have to reverse the order of) to only construct them one at a time, and unprotect them after each usage.

https://github.com/Rdatatable/data.table/blob/b8dd0db71e686d2557df6237e6c0f00c7e2ace03/src/frollR.c#L316

```r
> frollapply(1, rep(1L, 1e5), identity)
Error: protect(): protection stack overflow
```

The other is a C stack overflow. I imagine this is related, and comes from constructing this double pointer array on the stack (and maybe the SEXP arrays as well, but im not sure). `nk = 1e6` is pretty big.

https://github.com/Rdatatable/data.table/blob/b8dd0db71e686d2557df6237e6c0f00c7e2ace03/src/frollR.c#L310

```r
> frollapply(1, rep(1L, 1e6), identity)
Error: segfault from C stack overflow
```

DavisVaughan:
Closes #3993 

The following three examples from #3993 now run without issue, no segfaults!

``` r
library(data.table)

xx <- frollapply(1, rep(1L, 1e5), identity)

yy <- frollapply(1, rep(1L, 1e6), identity)

zz <- frollapply(as.list(rep(1, 1e6)), 1, identity)
```

If someone else wants to take over adding tests (if you want them), please go ahead. I'm not familiar with how you all test the package.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,8 @@ Authors@R: c(
   person("Roy","Storey",           role="ctb"),
   person("Manish","Saraswat",      role="ctb"),
   person("Morgan","Jacob",         role="ctb"),
-  person("Michael","Schubmehl",    role="ctb"))
+  person("Michael","Schubmehl",    role="ctb"),
+  person("Davis","Vaughan",        role="ctb"))
 Depends: R (>= 3.1.0)
 Imports: methods
 Suggests: bit64, curl, R.utils, knitr, xts, nanotime, zoo, yaml
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 ## BUG FIXES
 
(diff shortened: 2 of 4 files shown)


jangorecki:
Thanks for fixes. The second commit is a nice change. I kept all objects pre-allocated to align to the code of other rolling functions, but because _roll apply_ will not run in parallel we can safely override those variables inside the loop. Will add tests.

jangorecki:
Tests were added but disabled by default as they require 1.6 GB, 16 GB and 16GB memory. There is no way to tests those failures in a lightweight way.

PR #4006, opened 2019-10-28
Homepage additions

mattdowle:
Added Presentation and Articles as menu items at the top.
Moved manual to the bottom of vignettes menu. So as to point to the vignettes first before the (potentially off-putting) reference manual.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -9,7 +9,7 @@ development:
 
 navbar:
   structure:
-    left:  [home, introduction, reference, articles, news, benchmarks]
+    left:  [home, introduction, vignettes, news, benchmarks, presentations, articles]
     right: [github]
   components:
     home:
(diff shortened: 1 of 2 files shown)


_pkgdown.yml
```diff
@@ -9,7 +9,7 @@ development:
 
 navbar:
   structure:
-    left:  [home, introduction, reference, articles, news, benchmarks]
```
jangorecki: I think we should not rename articles to vignettes, pkgdown is sensitive to those names, they has to be hardcoded somewhere. You can observe now that on website we still have `references` tab despite you have removed it. It seems to be added anyway. I tried to rename it to "manual", then it was duplicated, one from "manual" and another added automatically "reference". The proper place for translation of labels seems to be in the "components" section, and here in "structure" we should align more to pkgdown terminology.

_pkgdown.yml
```diff
@@ -18,10 +18,7 @@ navbar:
     introduction:
       text: Introduction
       href: articles/datatable-intro.html
-    reference:
```
jangorecki: This tab is still there, added automatically by pkgdown. and when you click on it the hover stays on Vignettes, which is confusing. Will move it back again.

PR #4017, opened 2019-10-31
Allow C-level calls to forder

### LINKED ISSUE #4015: Export forder for use in other packages
Hi,

I was wondering if you would consider exporting forder for use in other packages?  

I can create a pull request if you think this is reasonable.

Best,

Matt.

mattfidler:
Closes #4015 

I don't actually think this is a duplicate;  It allows calling the underlying C code with evaluating in R, which could be faster.


mattdowle:
@lsilvest this PR by @mattfidler exports one more C function. It didn't have the `&` prefix on the function so I added that to be consistent with your #4084 on the assumption you had tested subsetDT and it's working from your package. Then it seemed cleaner to drop the C prefix in the name (since it's already clear from having to call `R_GetCCallable` that it's a C function that data.table is exporting. So I dropped the C prefix from subsetDT in this PR too. Hope ok.

src/init.c
```diff
@@ -238,8 +238,9 @@ void attribute_visible R_init_datatable(DllInfo *info)
 // relies on pkg/src/Makevars to mv data.table.so to datatable.so
 {
   // C exported routines, see ?cdt for details
-  R_RegisterCCallable("data.table", "CsubsetDT", (DL_FUNC) &subsetDT);
-
+  R_RegisterCCallable("data.table", "subsetDT", (DL_FUNC) &subsetDT);
+  R_RegisterCCallable("data.table", "forder", (DL_FUNC) &forder);
```
MichaelChirico: need to add to `?cdt` as well right

mattdowle:
Needs addition to the news item about subsetDT, and addition to ?cdt.

mattfidler:
Sorry Missed your comment @mattdowle.  There is alot of traffic on `data.table`.  

Thank you for the excellent package.  I don't need to be a contributor for this (in my opinion).  However CRAN may differ from my opinion (they have in the past).  I am fine adding myself as a contributor if I need to.

mattfidler:
I also couldn't see a project member email or any buttons; Perhaps the invitation has expired.

jangorecki:
@mattfidler are you planning to incorporate the feedback (cdt manual, news file, description file, comment in forder.c)? or you are not interested anymore in completing this PR?

mattfidler:
I realized this is still open.  I had forgot about this

Issue #940, opened 2014-11-10
[Request] data.table join nomatch=quote(stop())


jangorecki:
Feature Request.
Related to #857.

We have 2 dts:

``` r
library(data.table)
dict <- data.table(symbol = c("a","b","c"), ratio = c(5,10,15), key = "symbol")
dt <- data.table(symbol = c("a","b","c","d"), value = rnorm(4))
```

Which we want join:

``` r
dict[dt][,list(symbol, new_value = ratio * value)]
```

Consider the case when we expect to have match on all data in the join.
In case of nomatch we want to raise error.
Currently it is possible using:

``` r
new_dt <- dict[dt]
if(any(is.na(new_dt$ratio))) stop("missing data in the reference dictionary")
```

Which makes us unable to make chaining in such case.
If we could use something like this:

``` r
dict[dt,nomatch=quote(stop("missing data in the reference dictionary"))
     ][,list(symbol, new_value = ratio * value)]
```

just to stop process and raise the error in case of any nomatch.

PR #4021, opened 2019-11-01
fcase / case_when function for data.table

### LINKED ISSUE #3823: fcase / case_when function for data.table
Related/follow-up to #3657

`case_when` is a common tool in SQL for e.g. building labels based on conditions of flags, e.g. cutting age groups:

```
select
  case when age < 18 then '0-18'
       when age < 35 then '18-35'
       when age < 65 then '35-65'
       else '65+'
  end as age_label
```

Our comrades at `dplyr` have implemented this as `case_when` with an interface like

```
dplyr::case_when(
  age < 18 ~ '0-18',
  age < 35 ~ '18-35',
  age < 65 ~ '35-65',
  TRUE ~ '65+'
)
```

Using & interpeting formulas seems pretty natural for R -- the only other thing I can thing of would be like `on` syntax (`case_when('age<18' = '0-18', ...)`).

As for the back end, `dplyr` is doing a two-pass `for` loop at the R level which e.g. requires evaluating `age < 65` for _all_ rows (whereas this is unnecessary for anything with labels `'0-18'` or `'18-35'`).

I guess we can do much better with a C implementation. Will be interesting to contrast a proper lazy implementation with a parallel version (since IINM doing it in parallel will require first evaluating all the "LHS" values, e.g. as Jan met recently in `frollapply`).

Note that normally I'm doing `case_when` stuff with a look-up join, but it's not straightforward to implement a `case_when` as a join _in general_ -- though I think the two are isomorphic, backing out what the corresponding join should be I think is a hard problem. e.g. in the example here, the order of truth values matters since implicit in the later conditions are `(x & !any(y))` for each condition `x` that was preceded by conditions `y`. This case is straightforward to cast as a join on `cut(age)`, perhaps just using `roll`, but things can be much more complicated when several variables are involved in the truth conditions. So I don't think this route is likely to be fruitful.

2005m:
Closes #3823
I just wanted to share my code.
If you think it is worth it, I can move the code to a branch so you can work on it?


NEWS.md
```diff
@@ -6,6 +6,19 @@
 
 ## NEW FEATURES
 
+1. New function `fcase(.,default)` has been implemented in C, [#3823](https://github.com/Rdatatable/data.table/issues/3823). It is comparable to `dplyr::case_when`. Please see `?data.table::fcase` for more details.
```
MichaelChirico: I would point to SQL first, SQL `CASE WHEN` is the logical ancestor of `dplyr` and `data.table` versions I think. Especially since we don't use the formula interface.

NEWS.md
```diff
@@ -6,6 +6,19 @@
...
+
+```R
+# default 4 threads on a laptop with 16GB RAM and 8 logical CPU
+x = sample(1L:10L, 3e8, replace=TRUE) # 1GB
+microbenchmark::microbenchmark(
+	dplyr::case_when(x < 5L ~ 1L, x > 5L ~ 3L),
+	data.table::fcase((x < 5L, 1L, x > 5L, 3L)),
+	times = 5L,
+	unit = "s"
+)
```
MichaelChirico: should include the timing output as a sample right?

man/fcase.Rd
```diff
@@ -0,0 +1,57 @@
...
+}
+\usage{
+  fcase(..., default=NA)
+}
+\arguments{
+\item{...}{ A sequence consisting of logical condition (\code{when})-resulting value (\code{value}) \emph{pairs} in the following order \code{when1, value1, when2, value2, ..., whenN, valueN}. Logical conditions \code{when1, when2, ..., whenN} must all have the same length, type and attributes. Each \code{value} may either share length with \code{when} or be length 1. Please see Examples section for further details.}
+\item{default}{ Default return value, \code{NA} by default, for when all of the logical conditions \code{when1, when2, ..., whenN} are \code{FALSE} for some entries. }
+}
+\value{
+  Vector with the same length as the logical conditions (\code{when}) in \code{...}, filled with corresponding resulting value from \code{...}, or eventually \code{default}. Attributes of logical conditions in \code{...} are preserved.
```
MichaelChirico: > Attributes of logical conditions in `...` are preserved.

Don't we want to preserve the attributes of the `value`s, not the `when`s?

src/init.c
```diff
@@ -309,6 +311,9 @@ void attribute_visible R_init_datatable(DllInfo *info)
 
   initDTthreads();
   avoid_openmp_hang_within_fork();
+
+  R_RegisterCCallable("data.table", "CfifelseR", (DL_FUNC) &fifelseR);
```
MichaelChirico: This is about opening these functions for packages `LinkingTo` `data.table`, right?

src/fifelse.c
```diff
@@ -140,3 +140,182 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
   UNPROTECT(nprotect);
   return ans;
 }
+
+SEXP fcaseR(SEXP args) {
+  int n=length(args)-2;
+  if (n % 2) {
```
MichaelChirico: I guess the first element of `args` is a throwaway? Otherwise I'd expect `n` to be odd if including the `default`

src/init.c
jangorecki: such exports on C level should be documented somewhere, and in first place shouldn't happen unless we are sure that API is stable enough.

jangorecki:
to not leave it for too late I would encourage to think how to make evaluation of `when` lazy, so `fcase` will actually bring something new to the table. In base R there is `switch` to do that but it is not vectorized.
```r
switch(1, "a", stop("a"))
#[1] "a"
```

NEWS.md
2005m: Yes you are wright. I was wondering if someone with a decent laptop can post timings of a more relevant test? My computer is old and would die before giving me the output of something like : 

```
x = sample(1:100, 3e8, replace = TRUE) # 1 GB

dplyr::case_when(
  x < 10L ~ 0L,
  x < 20L ~ 10L,
  x < 30L ~ 20L,
  x < 40L ~ 30L,
  x < 50L ~ 40L,
  x < 60L ~ 50L,
  x > 60L ~ 60L
)
data.table::fcase(
  x < 10L, 0L,
  x < 20L, 10L,
  x < 30L, 20L,
  x < 40L, 30L,
  x < 50L, 40L,
  x < 60L, 50L,
  x > 60L, 60L
),
times = 5L,
unit ="s")
```

man/fcase.Rd
2005m: You are absolutely right! And it actually does. I need to fix this.

src/init.c
2005m: Yes it is to make these functions available at C level to other packages. I can remove these lines, or can add a couple of lines in NEWS. I was just testing what was documented in the R extension manual.

src/fifelse.c
2005m: The first element of `args` is `default`. Look how the function is defined in the wrappers.R file. The last element of `args` is `R_NilValue`. This is my understanding after reading this :
[https://marlin-na.github.io/r-api/pairlists/](https://marlin-na.github.io/r-api/pairlists/)

MichaelChirico: oh it's the last arg that's the difference. I thought the first was throwaway because of CDR right away. also didn't expect the R_NilValue to count towards length but I guess it must. thanks for clarifying

2005m:
Thinking out loud. Another way could be to split the `when` vector in  `n` equal part, (`n` being the number of core)  and process each part in parallel using the current `fcaseR` c function. That should speed up the process in case of large vectors. I think I will try this approach but first I will need to write a c function to split a vector in equal parts.

2005m:
@jangorecki , I think I will probably have to change the code to make fcase lazy. The only way I see to do that would be to use `list(...)` and `.Call`. I am not sure to understand how `switch` manage to achieve lazy evaluation in C. Do you know? Is it because it is a primitive?  Do you see any other option?

jangorecki:
If there will be no other option, there is always possibility to substitute dots and pass that together with the evaluation frame.

2005m:
@jangorecki , please see if you are happy with the lazy evaluation. See test 2124.72. If you need anything else please let me know. Thank you.

man/fcase.Rd
```diff
@@ -36,7 +36,9 @@ data.table::fcase(
 	x < 5L, as.Date("2019-10-11"),
 	x > 5L, as.Date("2019-10-14")
 )
-# [1] "2019-10-11" "2019-10-11" "2019-10-11" "2019-10-11" NA  "2019-10-14" "2019-10-14" "2019-10-14" "2019-10-14" "2019-10-14"
+# [1] "2019-10-11" "2019-10-11" "2019-10-11"
```
jangorecki: why do we need output in examples? it doesn't seems to be practical because when eventually a print/format/locale changes the print might be not-matching to the current output.

src/fifelse.c
```diff
@@ -314,6 +315,7 @@ SEXP fcaseR(SEXP args) {
     default:
       error("Type %s is not supported.", type2char(TYPEOF(outs)));
     }
+    if(0 == l) break;
```
jangorecki: extra whitespace here please `if (...`

src/init.c
jangorecki: could we have those two new C exports removed from this PR?
lets have R fcase completed first, and also this PR merged first https://github.com/Rdatatable/data.table/pull/4084

R/wrappers.R
```diff
@@ -6,6 +6,7 @@ fcoalesce   = function(...) .Call(Ccoalesce, list(...), FALSE)
 setcoalesce = function(...) .Call(Ccoalesce, list(...), TRUE)
 
 fifelse = function(test, yes, no, na=NA) .Call(CfifelseR, test, yes, no, na)
+fcase   = function(..., default=NA) .Call(CfcaseR, default, parent.frame(), as.list(substitute(...())))
```
jangorecki: what is the purpose of `...()` rather than `...`?

man/fcase.Rd
2005m: I usually put the output commnet because they appear in the documentation and ine can see the output without having to run the code.

R/wrappers.R
2005m: The goal here is that i want to pass a list to the C function, if I don't use as.list what is passed is a pair list and if I don't use parentheses it will just be a ling expression ig i remember well. You can create a simple function with the different option and see the output. I am adding links that might be helpful:
https://stat.ethz.ch/R-manual/R-devel/library/base/html/Paren.html

2005m:
To answer your question about speed, lazy evaluation produce the same speed. You will note that I have added a break at the end of the first loop which prevent the evaluation further expression in case all when condition have already been match with a value. This part might speed up a little bit the compuation in case the user has passed too many unnecessary when condtions.

man/fcase.Rd
jangorecki: I understand but I think it is much safer to not add them. Also pkgdown when render example it includes the output, so we would ended up with double output. I would ideally follow the base R convention about that.

2005m: Fine to remove them.

R/wrappers.R
jangorecki: First reference you provide Paren.html does not seems to be related to this use case. It would be related if you would use `(...)` and not `...()`.

jangorecki:
```r
is.call(quote(...))
#[1] FALSE
is.call(quote(...()))
#[1] TRUE
```
```r
as.list(quote(...()))
#[[1]]
#...

as.list(quote(...))
#[[1]]
#...
```

jangorecki: did you tried `substitute(list(...))` instead?

2005m: I have tried many combinations. I probably tried it yes. I might need to retry it and come back to you. But yes I agree if it behave the same we should use it because it is easier to read.

NEWS.md
```diff
@@ -6,6 +6,37 @@
...
+  x < 10L, 0L,
+  x < 20L, 10L,
+  x < 30L, 20L,
+  x < 40L, 30L,
+  x < 50L, 40L,
+  x < 60L, 50L,
+  x > 60L, 60L
+),
+times = 5L,
+unit ="s")
```
jangorecki: add whitespace

NEWS.md
```diff
@@ -6,6 +6,37 @@
 
 ## NEW FEATURES
 
+1. New function `fcase(...,default)` implemented in C by Morgan Jacob, [#3823](https://github.com/Rdatatable/data.table/issues/3823), is inspired by SQL `CASE WHEN` which is a common tool in SQL for e.g. building labels or cutting age groups based on conditions. `fcase` is comparable to R function `dplyr::case_when`. Please see `?data.table::fcase` for more details.
```
jangorecki: mention lazyness, giving some basic example can be nice here
```r
x = 1:3
fcase(x < 2, "it must be one",
      x < 3, "it must be two",
      x < 4, "it must be three" ,
      stop("provided value is an unexpected one!"))
```

src/fifelse.c
```diff
@@ -140,3 +140,184 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+    case LGLSXP: {
+      const int *restrict pouts = LOGICAL(outs);
+      int *restrict pans = LOGICAL(ans);
+      const int pna = nonna ? LOGICAL(na)[0] : NA_LOGICAL;
+      for (int64_t j=0; j<len2; ++j) {
+        idx = imask ? j : p[j];
+        if (pcons[idx]==1) {
+          pans[idx] = pouts[idx & amask];
+        } else {
+          if(imask) pans[j] = pna;
```
jangorecki: whitespace

src/fifelse.c
```diff
@@ -140,3 +140,184 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+    case INTSXP: {
+      const int *restrict pouts = INTEGER(outs);
+      int *restrict pans = INTEGER(ans);
+      const int pna = nonna ? INTEGER(na)[0] : NA_INTEGER;
+      for (int64_t j=0; j<len2; ++j) {
+        idx = imask ? j : p[j];
+        if (pcons[idx]==1) {
+          pans[idx] = pouts[idx & amask];
+        } else {
+          if(imask) pans[j] = pna;
```
jangorecki: ws

src/fifelse.c
```diff
@@ -140,3 +140,184 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+      const double *restrict pouts = REAL(outs);
+      double *restrict pans = REAL(ans);
+      const double na_double = Rinherits(outs, char_integer64) ? NA_INT64_D : NA_REAL;
+      const double pna = nonna ? REAL(na)[0] : na_double;
+      for (int64_t j=0; j<len2; ++j) {
+        idx = imask ? j : p[j];
+        if (pcons[idx]==1) {
+          pans[idx] = pouts[idx & amask];
+        } else {
+          if(imask) pans[j] = pna;
```
jangorecki: ws, also few other similar cases further

src/fifelse.c
```diff
@@ -140,3 +140,184 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+        } else {
+          if(imask && nonna) SET_VECTOR_ELT(ans, idx, pna);
+          p[l++] = idx;
+        }
+      }
+    } break;
+    default:
+      error("Type %s is not supported.", type2char(TYPEOF(outs)));
+    }
+    if(0 == l) break;
```
jangorecki: ws

src/init.c
2005m: I think i did  all changes you requested. Please let me know if I missed some. Thank you.

NEWS.md
```diff
@@ -6,9 +6,27 @@
 
 ## NEW FEATURES
 
-1. New function `fcase(...,default)` implemented in C by Morgan Jacob, [#3823](https://github.com/Rdatatable/data.table/issues/3823), is inspired by SQL `CASE WHEN` which is a common tool in SQL for e.g. building labels or cutting age groups based on conditions. `fcase` is comparable to R function `dplyr::case_when`. Please see `?data.table::fcase` for more details.
+1. New function `fcase(...,default)` implemented in C by Morgan Jacob, [#3823](https://github.com/Rdatatable/data.table/issues/3823), is inspired by SQL `CASE WHEN` which is a common tool in SQL for e.g. building labels or cutting age groups based on conditions. `fcase` is comparable to R function `dplyr::case_when` however it evaluates its arguments in a lazy way (i.e. only when needed) as shown below. Please see `?data.table::fcase` for more details.
```
jangorecki: > Please see `?data.table::fcase` for more details.

can be

> Please see `?fcase` for more details.

no need for namespace prefix here, it make sense when we compare multiple packages, but here we just point to documentation

man/fcase.Rd
```diff
@@ -18,42 +18,41 @@
...
 \examples{
-x = 1:10 
+x = 1:10
 data.table::fcase(
 	x < 5L, 1L,
 	x > 5L, 3L
 )
-# [1]  1  1  1  1 NA  3  3  3  3  3
 
 data.table::fcase(
```
jangorecki: no need `data.table::` prefix in examples, AFAIR we don't use it anywhere, and when running data.table example we can safely assume that pkg is attached.

R/wrappers.R
jangorecki: R uses `.()` inside `bquote`, and recently (still in R-devel) it also has `..()` use cases there. Thus using `...()` might be confusing, or eventually even break if R parses catches that somehow different, if ever R's introduce `...()` use too.

src/fifelse.c
```diff
@@ -140,3 +140,184 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+        } else {
+          if (imask && nonna) SET_VECTOR_ELT(ans, idx, pna);
+          p[l++] = idx;
+        }
+      }
+    } break;
+    default:
+      error("Type %s is not supported.", type2char(TYPEOF(outs)));
+    }
+    if (0 == l) break;
```
jangorecki: 1. I would reorder and write `l == 0` to make it easier to read
2. also please put `break` into new line so codecov can detect if that part of code was checked during tests

R/wrappers.R
2005m: `list(substitute(...))` does not work it is not the same  as the one current used. It seems to be treated differently at C level. Give it a try if you can. If you have another alternative I open to suggestions.

jangorecki:
Thanks for all adjustments. The more difficult part of the review left, to investigate if tests cover all necessary cases (and eventually confirm that all their output is expected). Adding to coming milestone, as the PR is mature enough to resolve it soon.

jangorecki:
Any idea if it is do-able to make `default` argument also lazy? probably not something easy because we early rely on its value in C code. I can imagine user may want to raise error message when none of the conditions have been satisfied (as requested before in https://github.com/Rdatatable/data.table/issues/940). If you don't have any good idea on doing `default` lazy then probably best to skip this now.

2005m:
We can do it but as you said we use `default` quite early in the C function, so making it lazy would not add much value in my view as opposed to `...` for which we skip unnecessary when/value cases.

2005m:
Some further benchmark:
```
x = sample(1:60,3e8, replace= TRUE) # 1.1GB

# Unit: seconds
#           expr    min     lq     mean   median      uq     max    neval
# case_when       53.05  55.03    57.93    55.72   59.71   66.11        5
# nested_fifelse   9.60  10.77    13.26    10.99   11.12   23.79        5 # setDTthread(1L)
# nested_fifelse   6.76   7.61     7.63     7.69    7.83    8.26        5 # setDTthread(2L)
# nested_fifelse   6.10   6.20     6.46     6.48    6.72    6.81        5 # setDTthread(4L)
# fcase            6.94   7.06     7.43     7.55    7.59    8.02        5

```

inst/tests/tests.Rraw
```diff
@@ -16561,6 +16561,95 @@ DT = data.table(A="a", key="A")
...
+test(2127.63, fcase(test_vec1, TRUE), as.logical(out_vec_oc))
+test(2127.64, fcase(test_vec1, 1+0i), as.complex(out_vec_oc))
+test(2127.65, fcase(test_vec1, list(1)), list(1,1,1,1,1, NULL, NULL, NULL, NULL, NULL, NULL))
+test(2127.66, fcase(test_vec1, as.Date("2019-10-11")), c(rep(as.Date("2019-10-11"),5),rep(NA,6)))
+test(2127.67, fcase(test_vec1, factor("a", levels=letters[1:3])), factor(c(rep("a",5),rep("NA",6)), levels=letters[1:3]))
+test(2127.68, fcase(test_vec1, 1L, default = 1:2), error = "Length of 'default' must be 1.")
+test(2127.69, fcase(test_vec1, 1L, test_vec_na1, 2L), error = "Argument #3 has a different length than argument #1. Please make sure all logical conditions have the same length.")
+test(2127.70, fcase(test_vec1, as.Date("2019-10-11"), test_vec2, 2), error = "Argument #4 has different class than argument #2, Please make sure all output values have the same class.")
+test(2127.71, fcase(test_vec1, 1L, test_vec2, 2:3), error = "Length of output value #4 must either be 1 or length of logical condition.")
+test(2127.72, fcase(TRUE, 1L, FALSE, stop("bang!")), 1L)
```
MichaelChirico: Excellent thoroughness of the test suite!

I may have missed it, but I didn't see any tests _mixing_ length-1 and length-n `value`s in a single `fcase` call. Just a few quick tests should be enough.

man/fcase.Rd
```diff
@@ -0,0 +1,58 @@
...
+\title{fcase}
+\description{
+\code{fcase} is a fast implementation of SQL \code{CASE WHEN} statement for R. Conceptually, \code{fcase} is a nested version of \code{\link{fifelse}} (with smarter implementation than manual nesting). It is comparable to \code{dplyr::case_when} and supports \code{bit64}'s \code{integer64} and \code{nanotime} classes.
+}
+\usage{
+  fcase(..., default=NA)
+}
+\arguments{
+\item{...}{ A sequence consisting of logical condition (\code{when})-resulting value (\code{value}) \emph{pairs} in the following order \code{when1, value1, when2, value2, ..., whenN, valueN}. Logical conditions \code{when1, when2, ..., whenN} must all have the same length, type and attributes. Each \code{value} may either share length with \code{when} or be length 1. Please see Examples section for further details.}
+\item{default}{ Default return value, \code{NA} by default, for when all of the logical conditions \code{when1, when2, ..., whenN} are \code{FALSE} for some entries. }
```
MichaelChirico: `FALSE` or missing, right?

jangorecki:
you can also put a optimistic scenario benchmark where lazyness feature can be observed (in timings rather than just error).

man/fcase.Rd
```diff
@@ -0,0 +1,58 @@
...
+}
+\value{
+  Vector with the same length as the logical conditions (\code{when}) in \code{...}, filled with the corresponding values (\code{value}) from \code{...}, or eventually \code{default}. Attributes of output values \code{value1, value2, ...valueN} in \code{...} are preserved.
+}
+\seealso{
+  \code{\link{fifelse}}
+}
+\examples{
```
MichaelChirico: it would be helpful to have an example where the order of conditions matters

src/fifelse.c
```diff
@@ -140,3 +140,198 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+        } else {
+          if (tn != type0) {
+            error("Resulting value is of type %s but 'default' is of type %s. "
+                     "Please make sure that both arguments have the same type.", type2char(type0), type2char(tn));
+          }
+          if (!R_compute_identical(PROTECT(getAttrib(outs,R_ClassSymbol)),  PROTECT(getAttrib(na,R_ClassSymbol)), 0)) {
+            error("Resulting value has different class than 'default'. "
+                     "Please make sure that both arguments have the same class.");
+          }
+          UNPROTECT(2);
```
MichaelChirico: do we need to `UNPROTECT(2)` inside the error branch as well?

MichaelChirico: no, we don't

MichaelChirico:
Looks great.

I guess a "cheating" way to show utility of laziness is to have one of the conditions/values set to `function() {Sys.sleep(5); rnorm(n)}`. Sleep should not happen under laziness

src/fifelse.c
2005m: I thought we do. This is how we did it in `fifelse`. I will look at the rest of your comments tonight and add what is missing. Thank you

man/fcase.Rd
2005m: Right

inst/tests/tests.Rraw
2005m: Will add

man/fcase.Rd
2005m: Not sure to understand

src/fifelse.c
jangorecki: we don't need in error branch, but this comment reffers to a aline that is non-error branch, so we do need unprotect here.

2005m:
@jangorecki , please see below benchmark with lazyness in action:

```
setDTthreads(4L)
x = sample(1:30, 3e7, replace = TRUE) # 114 MB

  nested_fifelse = fifelse(x < 10L, 0L,
  fifelse(x < 20L, 10L,
  fifelse(x < 30L, 20L,
  fifelse(x < 40L, 30L,
  fifelse(x < 50L, 40L,
  fifelse(x < 60L, 50L,
  fifelse(x > 60L, 60L, NA_integer_
  ))))))),
  fcase = fcase(
    x < 10L, 0L,
    x < 20L, 10L,
    x < 30L, 20L,
    x < 40L, 30L,
    x < 50L, 40L,
    x < 60L, 50L,
    x > 60L, 60L
  ),
  fcase2 = fcase(
    x < 10L, 0L,
    x < 20L, 10L,
    x < 30L, 20L,
    x < 40L, 30L
  ),
  times = 5L,
  unit = "s")
# Unit: seconds
#           expr       min        lq      mean   median       uq      max neval
# nested_fifelse 1.4431070 1.5588002 1.5712342 1.581573 1.595763 1.676928     5
# fcase          0.9279330 0.9678166 0.9846871 0.978591 1.006174 1.042921     5
# fcase2         0.9164689 1.1123252 1.1671696 1.263177 1.265469 1.278408     5

```

mattdowle:
Awesome work everyone!

2005m:
Thank you. Please let me know your view with regards to issue 4114. I also sent a message to the team with regards to issue 1336. Not sure if you received it. I am interested to look into this file backed features but I Ieed someone to tell me exactly what is required. Happy to work with anyone here! In the meantime I wish you all a happy festive end of year.

mattdowle:
Seasonal greetings to you too! I just replied on 4114.  On 1336, yes that's probably the highest priority after >2bn rows. I think we're all interested in working on it and somebody just needs to make a start. My approach was to do >2bn first as a way to prepare for 1336.  Since there's not much point in having 1336 if it can't do >2bn rows. We haven't tried to work together from the start on new features before; it probably needs one person to take the lead.

R/wrappers.R
jangorecki: @2005m 
https://stat.ethz.ch/pipermail/r-devel/2020-May/079413.html
interesting insight into this subject by an R core member

PR #4025, opened 2019-11-03
nafill gains nan argument

### LINKED ISSUE #4020: Should nafill replace NaN values?
This is more of a question than a feature request. The majority of answers to the SO question ["Fastest way to replace NAs in a large data.table"](https://stackoverflow.com/questions/7235657/fastest-way-to-replace-nas-in-a-large-data-table) use approaches that replace both `NA` and `NaN`, since `is.na(NaN)` is true. However, `nafill` and `setnafill` at the moment do not replace `NaN` values. Should these new functions be expected to fill `NaN` values as well?

```
library(data.table)

x <- c(NA, NaN, 0, 1, 2)
y <- x

# using is.na to replace
y[is.na(y)] <- 0
print(y)
# [1] 0 0 0 1 2

# using nafill to replace
nafill(x, fill = 0)
# [1]   0 NaN   0   1   2
```

`#` `Output of sessionInfo()`
```
R version 3.6.0 (2019-04-26)
Running under: macOS Mojave 10.14.6

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] data.table_1.12.6

[1] compiler_3.6.0 tools_3.6.0    knitr_1.25     xfun_0.10  
```

MichaelChirico:
Closes #4020 

I imagine this will need a few small changes but here's 90% of it.

Would be nice to do the C equivalent of `testF = if (nan_is_na) ISNAN else ISNA` outside the loop but I couldn't figure out the proper way to do that.


R/shift.R
```diff
@@ -24,16 +24,16 @@ shift = function(x, n=1L, fill=NA, type=c("lag", "lead", "shift"), give.names=FA
   ans
 }
 
-nafill = function(x, type=c("const","locf","nocb"), fill=NA, verbose=getOption("datatable.verbose")) {
+nafill = function(x, type=c("const","locf","nocb"), fill=NA, nan_is_na=TRUE, verbose=getOption("datatable.verbose")) {
```
jangorecki: argument name is too long IMO
Maybe we could align to API in https://github.com/Rdatatable/data.table/compare/fast-isna function?

src/nafill.c
```diff
@@ -1,22 +1,22 @@
...
 
-void nafillDouble(double *x, uint_fast64_t nx, unsigned int type, double fill, ans_t *ans, bool verbose) {
+void nafillDouble(double *x, uint_fast64_t nx, unsigned int type, double fill, bool nan_is_na, ans_t *ans, bool verbose) {
   double tic=0.0;
   if (verbose)
     tic = omp_get_wtime();
   if (type==0) { // const
     for (uint_fast64_t i=0; i<nx; i++) {
-      ans->dbl_v[i] = ISNA(x[i]) ? fill : x[i];
+      ans->dbl_v[i] = nan_is_na ? (ISNAN(x[i]) ? fill : x[i]) : (ISNA(x[i]) ? fill : x[i]);
```
jangorecki: duplicate whole loop and use `if (nan_is_na) ... else ...` instead, to reduce number of `if` checks inside the loop

man/nafill.Rd
```diff
@@ -10,15 +10,16 @@
...
-setnafill(x, type=c("const","locf","nocb"), fill=NA, cols=seq_along(x),
-          verbose=getOption("datatable.verbose"))
+setnafill(x, type=c("const","locf","nocb"), fill=NA, nan_is_na=TRUE,
+          cols=seq_along(x), verbose=getOption("datatable.verbose"))
 }
 \arguments{
   \item{x}{ vector, list, data.frame or data.table of numeric columns. }
   \item{type}{ character, one of \emph{"const"}, \emph{"locf"} or \emph{"nocb"}. Defaults to \code{"const"}. }
   \item{fill}{ numeric or integer, value to be used to fill when \code{type=="const"}. }
+  \item{nan_is_na}{ \code{logical}; should \code{NaN} be treated as \code{NA} during replacement? }
```
jangorecki: worth to mention it is ignored for non-double types

jangorecki:
could you also remove `verbose` arg from function API? and fetch it at the beginning of C function, passing to lower level functions as argument already.

R/shift.R
MichaelChirico: I like it, will also add a `util` function to handle such arg.

MichaelChirico: I think it'll be easier to use a `bool` internally. How about `nan_is_na` for internal use & `nan` for the function API only?

MichaelChirico: Check latest commit for sample approach

src/nafill.c
MichaelChirico: done

MichaelChirico:
About `verbose`, will come back to that

MichaelChirico:
@jangorecki these two tests are broken, shall we set `nan=NaN` by default for backwards compatibility? or happy to break this?

```
test(1.21, nafill(z, "locf"), c(NA,Inf,3,4,NaN,NaN,7,8,-Inf,-Inf)/2)
test(1.22, nafill(z, "nocb"), c(Inf,Inf,3,4,NaN,7,7,8,-Inf,NA)/2)
```

R/utils.R
```diff
@@ -13,6 +13,13 @@ if (base::getRversion() < "3.5.0") {
 isTRUEorNA    = function(x) is.logical(x) && length(x)==1L && (is.na(x) || x)
 isTRUEorFALSE = function(x) is.logical(x) && length(x)==1L && !is.na(x)
 allNA = function(x) .Call(C_allNAR, x)
+# helper for nan argument (e.g. nafill): TRUE -> treat NaN as NA
+nan_is_na = function(x) {
+  if (length(x) > 1L) stop("Argument 'nan' must be length 1")
+  if (identical(x, NA)) return(TRUE)
+  if (identical(x, NaN)) return(FALSE)
+  stop("Argument 'nan' must be NA (_not_ NA_real_) or NaN")
```
jangorecki: We should allow NA_real_ too

MichaelChirico:
Hmm @mattdowle seems there was an API failure on the `covr` step, rest of job is OK. Re-trigger Travis should be enough unless there's ongoing outage (seems all is fine now)

R/utils.R
```diff
@@ -13,6 +13,13 @@ if (base::getRversion() < "3.5.0") {
 isTRUEorNA    = function(x) is.logical(x) && length(x)==1L && (is.na(x) || x)
 isTRUEorFALSE = function(x) is.logical(x) && length(x)==1L && !is.na(x)
 allNA = function(x) .Call(C_allNAR, x)
+# helper for nan argument (e.g. nafill): TRUE -> treat NaN as NA
+nan_is_na = function(x) {
+  if (length(x) > 1L) stop("Argument 'nan' must be length 1")
```
jangorecki: What if it is zero length, could be handled here too

inst/tests/nafill.Rraw
```diff
@@ -186,3 +186,17 @@ if (test_bit64) {
...
 }
+
+# nan argument to treat NaN as NA in nafill, #4020
+x = c(NA, NaN, 0, 1, 2)
+ans1 = c(0, 0, 0:2)
+ans2 = c(0, NaN, 0:2)
+DT = data.table(a=x, b=x)
+test(7.1, nafill(x, fill=0), ans1)
+test(7.2, nafill(x, fill=0, nan=NaN), ans2)
+test(7.3, nafill(x, 'locf'), c(NA, NA, 0, 1:2))
```
MichaelChirico: @jangorecki note this output -- `NaN` is converted to `NA` -- `nan=NA` means convert `NaN` to `NA` even if not replacing it. Makes sense?

MichaelChirico:
> And verbose arg? Could we remove it on R level here?
Do we have any rule about when `verbose` should/shouldn't be provided as a function argument? Seems pretty harmless

jangorecki:
@MichaelChirico no rules about that, here it looks really unneeded, and could be easily migrated to use option.

MichaelChirico:
OK done the move, please have a look

PR #4034, opened 2019-11-10
Translation dracodoc

dracodoc:
@MichaelChirico ,I just completed my assignment and double checked once. Can you notify the reviewer to look at it? Thanks!


MichaelChirico:
Thanks!! Could you take a look @shrektan?

src/fread.c
```diff
@@ -1495,7 +1495,7 @@ int freadMain(freadMainArgs _args) {
           const char *prevLineStart=ch, *lineStart=ch;
           int lastncol = countfields(&ch);
           if (lastncol<0) continue;  //   invalid file with this sep and quote rule, skip
-          ASSERT(lastncol>0, "Internal error: first non-empty row should always be at least one field; %c %d", sep, quoteRule); // # nocov
+          ASSERT(lastncol>0, _("Internal error: first non-empty row should always be at least one field; %c %d"), sep, quoteRule); // # nocov
```
MichaelChirico: OK, so I figured out what's going wrong here. And now I remember that I had found something wrong here earlier but gave up before understanding what's the issue...

`ASSERT` is doing `char` array literal concatenation:

```
"Internal error in line %d of fread.c, please report on data.table GitHub:  " msg
```

However, this breaks when we do the translation -- note (from `po.h`) that `_(String)` is expanding to `dgettext("data.table", String)`, so if we wrap the `ASSERT` fixed literal with whatever the message is, after macro expansion we'll have:

```
dgettext("data.table", "Internal error in line %d of fread.c, please report on data.table GitHub:  ") dgettext("data.table", msg)
```

The `f() g()` is a syntax error.

I suppose we _could_ complicate things by doing some array copying, etc., to handle this more properly, BUT seeing as how these are all _internal_ errors in the first place (meaning users should never see them, in principle), I don't think it's worth the effort.

So I'll roll back the use of translation with `ASSERT`. Sorry about that!

shrektan:
> Thanks!! Could you take a look @shrektan?

Sure. But I’m only available in this weekend. Will try to take a closer look then. @MichaelChirico

MichaelChirico:
> But I’m only available in this weekend.

No rush @shrektan, for translation reviews I'm looking to finish by mid-December :)

po/R-zh_CN.po
```diff
@@ -930,116 +930,135 @@ msgid "All elements in argument 'x' to 'setDF' must be of same length"
...
 msgstr ""
+"'setDT' 的参数 'x' 中所有的元素都必须是同一长度，但"
+"输入长度的概况（长度（length）：频率（frequency））是："
 
 msgid "%s:%d"
 msgstr ""
 
 msgid "The first entry with fewer than"
-msgstr ""
+msgstr "第一个少于"
```
shrektan: 建议改为"第一个长度少于"

po/R-zh_CN.po
```diff
@@ -930,116 +930,135 @@ msgid "All elements in argument 'x' to 'setDF' must be of same length"
...
 
 msgid "which is not 1 or nrow ("
-msgstr ""
+msgstr "其非 1 或 总行数 nrow ("
 
 msgid ""
 "). For robustness, no recycling is allowed (other than of length 1 RHS). "
 "Consider %in% instead."
 msgstr ""
+")。为了鲁棒性考虑，不进行回收重用（recycling）（除非右手侧长度为 1 ）。"
```
shrektan: 建议改为：

")。考虑到程序的稳健性，只有在右侧元素长度为 1的情况下，我们才会对之进行循环。"

xhdong-umd:
Totally agree the two modifications are better. Thanks!

po/R-zh_CN.po
dracodoc: Totally agree the two modifications are better. Thanks!

MichaelChirico:
Great! Please push an update when you get a chance @dracodoc 👍

po/R-zh_CN.po
dracodoc: Agree, somehow I missed this part. The sentence is a little bit vague "entry fewer than entries" though.

dracodoc:
I have updated and pushed changes. The commit was made with my work github id because I often need to switch between 2 accounts...

MichaelChirico:
Great! Thanks @shrektan and @dracodoc!!

PR #4038, opened 2019-11-11
use _R_CHECK_TESTS_NLINES=0 in CI

MichaelChirico:
Came across this today:

https://yihui.name/en/2017/12/last-13-lines-of-output/

Can make reaction to test fails much quicker if we can see the full output...

Could also just choose something larger like 50, not sure if this will blow up the logs needlessly in successful cases. In any case 13 is often not enough.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,6 +2,8 @@ variables:
   CRAN_MIRROR: "https://cloud.r-project.org"
   _R_CHECK_FORCE_SUGGESTS_: "false"
   _R_CHECK_NO_STOP_ON_TEST_ERROR_: "true"
+  # Block truncation of any error messages in R CMD check
+  _R_CHECK_TESTS_NLINES_: 0
 
 stages:
   - dependencies
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -25,6 +25,8 @@ environment:
 # Default GCC_PATH appears to be gcc-4.6.3 which is now unsupported as from Rtools.exe v3.4.
     _R_CHECK_NO_STOP_ON_TEST_ERROR_: true
 # continue tests even if some script failed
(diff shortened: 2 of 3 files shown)


MichaelChirico:
related setting: `_R_CHECK_VIGNETTES_NLINES_`

> Maximum number of lines to show from the bottom of the output when reporting errors in running or re-building vignettes. ( Value 0 means all lines will be shown.) Default: 10 for running, 25 for re-building.

.gitlab-ci.yml
```diff
@@ -2,6 +2,8 @@ variables:
   CRAN_MIRROR: "https://cloud.r-project.org"
   _R_CHECK_FORCE_SUGGESTS_: "false"
   _R_CHECK_NO_STOP_ON_TEST_ERROR_: "true"
+  # Block truncation of any error messages in R CMD check
+  _R_CHECK_TESTS_NLINES_: 0
```
jangorecki: @MichaelChirico maybe we should add it to the biggest build only? so the cran-like builds will remain consistent about log output, and eventually will help to detect that we are not displaying all informations that we might want to see

MichaelChirico: that makes sense. CRAN already has this set to 0 though right?

anyway I'm not as knowledgeable on which would count as "biggest" so I leave that to you to add

jangorecki: according to documentation CRAN does not alter that env variable

MichaelChirico: hmm, I was going off Matt's comment here

https://github.com/Rdatatable/data.table/pull/4106/files#diff-02e6dee04136eb470c5f4a0f9f40dacbR129

jangorecki: It is possible that output is logged fully till it reaches the exception, and there the 13 lines limit is applied

PR #4043, opened 2019-11-13
setnames(DT,new=) now works

### LINKED ISSUE #4041: Odd parameter names of `setnames`
If you call `setnames` with just a vector of new names, then it looks like `setnames(df, c("date", "place", "value"))`, which is fine. But if you name the argument supplying the new names, then the call looks like `setnames(old = c("date", "place", "value"), df)`, which is confusing because you're supplying new names, not old ones. Consider allowing `old` to be missing, so `new` can be set by name instead.

MichaelChirico:
Closes #4041 

@Kodiologist I re-read your issue more carefully and now understand. Good suggestion! And easy to implement.

Have a look!


Kodiologist:
Excellent, thank you.

PR #4065, opened 2019-11-20
fix first-last xts dispatch #4053

### LINKED ISSUE #4053: data.table  breaks xts first(), last() distpatching
upgrading to data.table 12.6 breaks existing xts first(), last() dispatching. Im not sure what the last working verison was, but the broke somewhere between 12.2 and 12.6. reverting to 12.2 restores expected behavior

```
library(quantmod)
r <- getSymbols("SPY", auto.assign=F)
> last(r)
           SPY.Open SPY.High SPY.Low SPY.Close SPY.Volume SPY.Adjusted
2019-11-15   311.02   311.84  310.26    311.79   62023600       311.79
> library(data.table)
> last(r)
[1] 311.79
```

jangorecki:
closes #4053


inst/tests/tests.Rraw
```diff
@@ -16163,23 +16163,106 @@ test(2107.3, names(DT), c('A','b','c'))
...
   xt = xts(matrix(1:4, 2L, 2L), x)
-  test(2108.05, last(xt, 2L), xt, output="last: using xts::last")
-  test(2108.06, first(xt, 2L), xt, output="first: using xts::first")
+  test(2108.05, last(xt, 2L), xt, output="using xts::last: is.xts(x)")
+  test(2108.06, first(xt, 2L), xt, output="using xts::first: is.xts(x)")
 }
 # first on empty df now match head(df, n=1L), #3858
 df = data.frame(a=integer(), b=integer())
-test(2108.11, first(df), df, notOutput="xts")
-test(2108.12, tail(df), df, notOutput="xts")
```
jangorecki: this `tail(df)` was actually a mistake in old unit test, it was now amended to `last(df)` as it should be in the first place

mattdowle:
Is there a smaller solution than +165 -35 ?

MichaelChirico:
net 101 of that is in tests though right?

jangorecki:
I don't think we can rework the logic to save some lines. As @MichaelChirico noticed, a lot of those lines are tests. Verbosity takes some extra lines too, but it is useful for testing.
Another thing I got in mind was an extra argument `last = function(x, n, ..., fast=FALSE)` when setting `fast=TRUE` we would ignore xts checks and remove all overheads that are here for compatibility. For speed, it is now better to use `x[length(x)]` on a vector, and `x[nrow(x),]` on DT. So `first` and `last` will always suffer from an overhead, unless we give user control to ignore compatibility.
Also there could be extra argument, lets say `fill=FALSE`, when true then result would be always length 1, so for a 0 length that would be an `NA`. But those basically are enchancement, so wanted to keep this PR only on fixing `xts`.

mattdowle:
But why are so many new tests needed?  It's just `first` and `last`! Out of all the complexity we have, these are really simple functions. How has it become so complicated that so many tests are needed? We were aware of the conflict and differences of our first/last to xts' first/last before, and had tests for that. What's gone so wrong that needs all this change for something so simple?!

jangorecki:
These tests are newly added, they are likely to overlap to existing test, but they test for verbose output, so they are more accurate and will detect more changes in future. Functions are simple but integration to xts complicates a lot.

jangorecki:
If we want to have simple functions then as I proposed above we could provide an option to bypass checks for xts compatibility. IMO it make sense to make a new PR for that and in this one only fix regression in method dispatch.

mattdowle:
@jangorecki seems to be just failing one GLCI job (dev-cran-lin) with:
```2 errors out of 9684. Search tests/tests.Rraw for test numbers: 2108.82, 2108.92```
Will leave to you for follow-up PR.

PR #4067, opened 2019-11-20
should check LENGTH(i) before using LENGTH(VECTOR_ELT(i,0))

### LINKED ISSUE #4060: segfaults on R3.4.4 when i is NULL
The following code causes segfauls on R3.4.4.

1. Current it's reproducible on Windows. ~~I haven't tested it on other platforms yet.~~ **UPDATE:** I've reproduced this on Linux.
1. Both the CRAN version and the dev version have this issue.
1. Note, this happens only for R3.4.4 (or the previous version of R). **On R3.5 or R3.6 it works fine** ~~So I doubt it relates to the code handles the ALTREP, which is introduced from R3.5. However, I'm not sure since I haven't digged into this.~~

```r
library(data.table)
tbl <- data.table(A = "a", key = "A")
tbl[J(NULL)]
```

**Update:** the segfault is triggered by `Cbmerge` and I confirm data.table v1.12.3 also has this issue.

shrektan:
Closes #4060 

The below code would trigger segfaults in R3.4.4 or previous version. Now it won't.

```r
library(data.table)
tbl <- data.table(A = "a", key = "A")
tbl[J(NULL)]
```


src/bmerge.c
```diff
@@ -50,8 +50,10 @@ SEXP bmerge(SEXP iArg, SEXP xArg, SEXP icolsArg, SEXP xcolsArg, SEXP isorted, SE
   if (LENGTH(icolsArg) > LENGTH(xcolsArg)) error("Internal error: length(icols) [%d] > length(xcols) [%d]", LENGTH(icolsArg), LENGTH(xcolsArg)); // # nocov
   icols = INTEGER(icolsArg);
   xcols = INTEGER(xcolsArg);
-  xN = LENGTH(VECTOR_ELT(x,0));
-  iN = ilen = anslen = LENGTH(VECTOR_ELT(i,0));
+  xN = LENGTH(x) ? LENGTH(VECTOR_ELT(x,0)) : 0;
```
shrektan: Despite the fact that I don't really think a zero-column x will call this function, I add a defense here in case it happens somehow in the future.

PR #4072, opened 2019-11-21
all.equal no longer error when not a DT

### LINKED ISSUE #4042: all.equal shouldn't ERROR when comparing a data.table to a non-data.table
Currently `all.equal.data.table` gives a hard error when the second argument is not a data.table:

``` r
library(data.table)
df <- data.frame(a=1)
dt <- as.data.table(df)
all.equal(df, dt)
#> [1] "Attributes: < Names: 1 string mismatch >"                                              
#> [2] "Attributes: < Length mismatch: comparison on first 2 components >"                     
#> [3] "Attributes: < Component \"class\": Lengths (1, 2) differ (string compare on first 1) >"
#> [4] "Attributes: < Component \"class\": 1 string mismatch >"                                
#> [5] "Attributes: < Component 2: Modes: numeric, externalptr >"                              
#> [6] "Attributes: < Component 2: target is numeric, current is externalptr >"
all.equal(dt, df)
#> Error in all.equal.data.table(dt, df): 'target' and 'current' must both be data.tables
```

(the error is also thrown if the first argument is not a data.table, but that would only happen if you bypassed method dispatch, in which case it's your own fault)

This makes it impossible to use common paradigms such as `if (isTRUE(all.equal(x, y)) {...}` : I would have to throw a `try` in there just to handle the case that `x` (or one of its list elements or attributes) is a data.table and `y` (or its corresponding element/attribute) isn't. More generally, it seems bad that `all.equal(x, y)` could have such substantially different behaviour than `all.equal(y, x)`. 

This error is thrown near the very top of  `all.equal.data.table`:
```r
if (!is.data.table(target) || !is.data.table(current)) 
        stop("'target' and 'current' must both be data.tables")
```
 
I see two possible and very simple remedies:
1. Literally just replace `stop` with `return` (possibly tailoring the message to tell you which one isn't a data.table)
2. Replace the whole `stop` clause with `NextMethod()` or explicit `all.equal.default` (such that `all.equal(dt, df)` would give approximately the same message as `all.equal(df, dt)`.

Tests 1613.59 and 1613.60 explicitly check for this error, so maybe this is intentional, but consider this my request to change that. If I get a approval on either remedy, I could do the PR myself (it would be my first for an open source project, but seems simple enough!)

```r
sessionInfo()
#> R version 3.5.2 (2018-12-20)
#> Running under: Ubuntu 16.04.5 LTS
#> 
#> Matrix products: default
#> BLAS: /usr/lib/libblas/libblas.so.3.6.0
#> LAPACK: /usr/lib/lapack/liblapack.so.3.6.0
#> 
#> locale:
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.12.2
#> 
#>  [1] compiler_3.5.2  magrittr_1.5    tools_3.5.2     htmltools_0.4.0
#>  [5] yaml_2.2.0      Rcpp_1.0.3      stringi_1.4.3   rmarkdown_1.16 
#>  [9] highr_0.8       knitr_1.25      stringr_1.4.0   xfun_0.10      
#> [13] digest_0.6.22   rlang_0.4.1     evaluate_0.14
```

d-sci:
Closes #4042 
Fix so that `all.equal(DT, not_a_DT)` returns the typical messages rather than error.
I had to modify old tests 1613.59 and 1613.60, which were checking for this error.


R/setops.R
```diff
@@ -103,8 +103,8 @@ fsetequal = function(x, y, all=TRUE) {
 # all.equal ----
 
 all.equal.data.table = function(target, current, trim.levels=TRUE, check.attributes=TRUE, ignore.col.order=FALSE, ignore.row.order=FALSE, tolerance=sqrt(.Machine$double.eps), ...) {
-  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance))
-  if (!is.data.table(target) || !is.data.table(current)) stop("'target' and 'current' must both be data.tables")
+  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance), is.data.table(target))
+  if (!is.data.table(current)) return(NextMethod())
```
MichaelChirico: This will return `all.equal(data.table(a=1), data.frame(a=1))` to `TRUE` right? Shouldn't that be `FALSE` with reason "`current` is not a `data.table`"?

jangorecki: why dispatch to next method instead of returning scalar character about class mismatch? if `target` is a data.table then it make sense to produce an answer in data.table method rather than redirecting to another.

d-sci: It will return
```
[1] "Attributes: < Names: 2 string mismatches >"                              
[2] "Attributes: < Length mismatch: comparison on first 2 components >"       
[3] "Attributes: < Component 1: target is externalptr, current is character >"
[4] "Attributes: < Component 2: Modes: character, numeric >"                  
[5] "Attributes: < Component 2: Lengths: 2, 1 >"                              
[6] "Attributes: < Component 2: target is character, current is numeric >"    
```

which is the same (just with some terms swapped) as the result of `all.equal(data.frame(a=1), data.table(a=1))`

d-sci: I guess I thought that `all.equal(DT, DF)` should return roughly the same info as `all.equal(DF, DT)`. 

But happy to do the just do that if you prefer.

jangorecki: I would rather prefer to return scalar early rather than investigating all differences, `all.equal` can be expensive, especially when you use `ignore.row.order`. Of course `check.attributes=FALSE` should continue to work, the same way as `all.equal(DF, DT, check.attributes=FALSE)` do.

d-sci:
@jangorecki raised an interesting point with regards to `check.attributes = F`. And what the result of `all.equal` should be on other-wise identical `data.table` vs `data.frame`. Base R can actually be oddly inconsistent here on `all.equal(x, y)` vs `all.equal(y,x)`:

```r
all.equal(data.frame(a=1), c(a=1), check.attributes = F)
#> [1] TRUE
all.equal(c(a=1), data.frame(a=1), check.attributes = F)
#> [1] "target is numeric, current is data.frame"
```

Current data.table always errors when the second argument isn't a data.table (regardless of `check.attributes`), and most of `all.equal.data.table` seems written based on the assumption that both args are data.tables. With new commit, this PR takes an easy middle ground, relaxing the error to a message, but still not returning TRUE for `all.equal(data.table(a=1), data.frame(a=1), check.attributes = F`), which would require more work:

```r
all.equal(data.table(a=1), c(a=1), check.attributes = F) # would ERROR before
#> [1] "target is data.table, current is numeric"
all.equal(c(a=1), data.table(a=1), check.attributes = F)
#> [1] "target is numeric, current is data.table"

all.equal(data.table(a=1), data.frame(a=1), check.attributes = F) # would ERROR before
#> [1] "target is data.table, current is data.frame"
all.equal(data.frame(a=1), data.table(a=1), check.attributes = F)
#> [1] TRUE
```

R/setops.R
```diff
@@ -103,8 +103,8 @@ fsetequal = function(x, y, all=TRUE) {
 # all.equal ----
 
 all.equal.data.table = function(target, current, trim.levels=TRUE, check.attributes=TRUE, ignore.col.order=FALSE, ignore.row.order=FALSE, tolerance=sqrt(.Machine$double.eps), ...) {
-  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance))
-  if (!is.data.table(target) || !is.data.table(current)) stop("'target' and 'current' must both be data.tables")
+  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance), is.data.table(target))
+  if (!is.data.table(current)) return(paste0('target is data.table, current is ', data.class(current)))
```
jangorecki: what about `check.attributes=FALSE`?

d-sci: `all.equal(data.frame(a=1), c(a=1), check.attributes = F)` returns TRUE -- do you think it should for `data.table` as well? 

The problem with continuing on despite `current` not being a data.table (could be as simple as a numeric, in that example) is that you have to a lot more careful about calling `setcolorder`, `nrow`, etc, on it which happens in the next few lines. Hence my original idea to just call `NextMethod`. I saw this as an easy middle ground. Maybe if `check.attributes = FALSE` just do `current <- as.data.table(current)` then keep going??

jangorecki: class is an attribute. `NextMethod` could eventually be an option but for a `check.attributes=FALSE` then. Also we need to take care and check if any extra arguments haven't been used, because they will be ignored in a non-data.table all.equal method.

d-sci: Ok, I'll try to think about this some more. As is, I do believe it is still an improvement, though clearly not perfect.

jangorecki: It is improvement, but I don't think we should merge it like this without support for `check.attributes` argument.

d-sci:
Ok, new attempt to handle `check.attributes` appropriately. Basically relying on the fact that `setDT(x)` does a good job of checking whether `x`is at all data.table-like and if so, coercing to a data.table without actually changing any non-attribute data. So, if `current` is not a data.table:
- if `check.attributes = TRUE` return quickly with class mismatch info
- if `check.attributes = FALSE` but`current` is nothing like a data.table, return quickly with message to that effect
- if `check.attributes = FALSE` but`current` can be coerced via `setDT`, do so and continue on (thus still honouring `ignore.col.order`, `ignore.row.order` etc.)

R/setops.R
```diff
@@ -103,8 +103,13 @@ fsetequal = function(x, y, all=TRUE) {
...
 
 all.equal.data.table = function(target, current, trim.levels=TRUE, check.attributes=TRUE, ignore.col.order=FALSE, ignore.row.order=FALSE, tolerance=sqrt(.Machine$double.eps), ...) {
-  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance))
-  if (!is.data.table(target) || !is.data.table(current)) stop("'target' and 'current' must both be data.tables")
+  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance), is.data.table(target))
+
+  if (!is.data.table(current)) {
+    if (check.attributes) return(paste0('target is data.table, current is ', data.class(current)))
+    try({current = setDT(copy(current))}, silent = TRUE)
+    if (!is.data.table(current)) return('target is data.table and current is not of a comparable form')
```
jangorecki: Instead of `comparable form` I would use working that `current was not a data.table and it failed to be coerced to it`.
Also `as.data.table` seems to be more appropriate than `setDT(copy(.))`, as it may handle any class than will provide `as.data.table` method.

inst/tests/tests.Rraw
```diff
@@ -8466,8 +8466,17 @@ test(1613.573, all.equal(cbind(x, factor(1)), cbind(x, factor(1)), ignore.row.or
 test(1613.581, all(all.equal(x, y, ignore.row.order = FALSE, tolerance = 1), all.equal(x, y, ignore.row.order = TRUE, tolerance = 1)))
 test(1613.582, all(all.equal(x, y, ignore.row.order = FALSE, tolerance = sqrt(.Machine$double.eps)/2), all.equal(x, y, ignore.row.order = TRUE, tolerance = sqrt(.Machine$double.eps)/2)), warning = "Argument 'tolerance' was forced")
 
-test(1613.59, all.equal.data.table(1L, 2L), error = "'target' and 'current' must both be data.tables")
-test(1613.60, all.equal(data.table(1L), 2L), error = "'target' and 'current' must both be data.tables")
+# fix for #4042
+test(1613.59, all.equal.data.table(1L, 2L), error = "is.data.table(target) is not TRUE")
+test(1613.601, all.equal(data.table(a=1), data.frame(a=1)), "target is data.table, current is data.frame")
+test(1613.602, all.equal(data.table(a=1), data.frame(a=1), check.attributes = F))
```
jangorecki: those `F` needs to be `FALSE`

inst/tests/tests.Rraw
```diff
@@ -8466,8 +8466,17 @@ test(1613.573, all.equal(cbind(x, factor(1)), cbind(x, factor(1)), ignore.row.or
...
+test(1613.603, all.equal(data.table(a=1), list(a=1), check.attributes = F))
+test(1613.604, all.equal(data.table(a=c(1,1), b=c(2,2)), list(a=1,b=c(2,2)), check.attributes = F), "target is data.table and current is not of a comparable form")
+test(1613.605, all.equal(data.table(a=1), 1L, check.attributes = F), "target is data.table and current is not of a comparable form")
+L1 <- list(a = data.table(1L), b = setattr("foo", "tbl", data.table(1L)))
+L2 <- list(a = list(1L), b = setattr("foo", "tbl", 1L))
+test(1613.606, is.character(all.equal(L1, L2)))
```
jangorecki: instead of just `is.character` maybe testing for a string, then tests will be also self describing itself.

d-sci:
Addressed most of your comments as suggested. I didn't change `copy(setDT(x))` to `as.data.table(x)` because `as.data.table` seems like it could do quite a lot of coercion that really changes underlying values. In particular, it means that both `list(a = 1, b = c(2,2))` and `list(a = c(1,1), b = c(2,2))` would be "all equal" with `data.table(a = c(1,1), b = c(2,2))` even though they wouldn't be with each other (since `as.data.table` can expand columns but `setDT` will error if any list elements are unequal length). `setDT` seems to be really good at telling "what could possibly be a data.table as it currently is, even if it' class isn't data.table", and I can't see a situation where it would fail on `X` but you would still expect `all.equal(DT, X)`

jangorecki:
@d-sci thanks, I am getting your point. The thing is that if we supply a list into `current` one might actually expect it to be expanded, because this is how the list behaves when being coerced to data.table. If one wants to have behavior you described then probably `all.equal(as.list(target), current)` is more appropriate (then of course we dispatch to `all.equal.list`).
If `as.data.table` way handles your use case then I would go with it. Otherwise probably best to hear what others think about `try()` coercion and the consistency of list/data.table equality.

d-sci:
Ok, fair enough, switched over to `as.data.table`. This will certainly work for my use case -- as long as I get a string and not an error for `all.equal(DT, not_a_DT)` I'm happy!

jangorecki:
We need documentation update for `current` argument, to explicitly mention `as.data.table` method used for coercion when `check.attributes=F`. Probably also an example of `all.equal(dt(a=1), 1, check.attributes=FALSE))` which might be surprising.

jangorecki:
Thanks for fixing missing S3 registration. Unless you have something more in mind I think we can remove WIP label already.

d-sci:
Yep, this is ready in my mind! Thanks for all your comments and suggestions.

mattdowle:
Thanks, @d-sci. I added you to DESCRIPTION in this PR and invited you to be project member. Please accept the invite displayed in your GitHub repositories page and then you can create branches in the main project so we can more easily commit to each other's branches. Welcome!

PR #4074, opened 2019-11-23
Print Option: Print columns that fit in single console without wrapping

### LINKED ISSUE #1497: [Request] head.data.table print columns fit to screen width
Would it be useful for head.data.table to have an option to print only the head of columns that fit the screen width, and summarise the rest?  I was imagining something like the printed output from the head of a `tbl_df` in dplyr.  I think it is nice for tables with many columns.

TysonStanley:
Closes #1497

This PR addresses @MichaelChirico's number 11  and, somewhat, number 16 on #1523 by providing a `trunc.cols` argument (with corresponding default `option("datatable.print.trunc.cols")`). For example:

```r
options("width" = 40)
DT <- data.table(thing_11 = vector("integer", 3),
                 thing_21 = vector("complex", 3),
                 thing_31 = as.IDate(paste0("2016-02-0", 1:3)),
                 thing_41 = "aasdfasdfasdfasdfasdfasdfasdfasdfasdfasdf",
                 thing_51 = vector("integer", 3),
                 thing_61 = vector("complex", 3))
print(DT, trunc.cols=TRUE)
#>    thing_11 thing_21   thing_31
#> 1:        0     0+0i 2016-02-01
#> 2:        0     0+0i 2016-02-02
#> 3:        0     0+0i 2016-02-03
#> Variables not shown: thing_41, thing_51, thing_61.
```

Some possible ideas for adjustments:

- Include variable type in the "Variables shown..." message
- Instead of going in the order of the columns, we could go with the `data.table` usual of showing the first and last columns.


man/print.data.table.Rd
```diff
@@ -58,5 +60,18 @@
   setindexv(DT, c("a", "b"))
   setindexv(DT, "a")
   print(DT, print.keys=TRUE)
+
+  # `trunc.cols` will make it so only columns that fit in console will be printed
+  #    with a message that states the variables not shown
+  old_width = getOption("width")
+  options("width" = 40)
```
jangorecki:
```
old_width = options(width = 40)
```
should do, then `options(old_width)` at the end

TysonStanley:
Added unit tests, fixed an issue that appeared with one row data tables when `trunc.cols = TRUE`, and updated the example with the suggestions from @jangorecki. 

Should I add info to `NEWS.md`? 

Also, regarding the classes, I was thinking we could add the classes to the "Variables not shown" message when `class = TRUE` so it shows up in the table and the message. Would it be worth going forward with this?

TysonStanley:
The failed test looks like it just needs to be re-run (it looks like a LaTeX error). 

I added the classes to the message about not printed columns when `class=TRUE`, added a message to `NEWS` about the changes, and updated the tests. @MichaelChirico do you have anything else you’d like me to do for this PR?

MichaelChirico:
@TysonStanley please see the inline comments I added, still a few things to handle. Thanks!

TysonStanley:
@MichaelChirico  I’m not seeing the inline comments showing up. Am I missing it? Also, I like the updated `NEWS`.

MichaelChirico:
@TysonStanley do you see this and several others on the `Files Changed` tab?

https://github.com/Rdatatable/data.table/pull/4074/files

Screenshot 2019-11-28 at 7 50 10 AM

TysonStanley:
@MichaelChirico no, I looked there and couldn’t see any... All I see are the changes to the files (no comments at all).

MichaelChirico:
@TysonStanley aaahhh sorry. I forgot to hit submit on my review. Hence the `Pending` in my screenshot.

R/print.data.table.R
```diff
@@ -88,6 +90,17 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     rownames(toprint)[1L] = ""
   }
   if (quote) colnames(toprint) <- paste0('"', old <- colnames(toprint), '"')
+  if (trunc.cols) {
+    # allow truncation of columns to print only what will fit in console #XXXX
+    widths = dt_width(x, class, row.names, names(x))
```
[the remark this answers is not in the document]
TysonStanley: Yes, this currently accounted for in the `dt_width()` function (which will change slightly with your earlier comment but will retain this feature). It is the `rownum_width` object that accounts for this padding.

R/print.data.table.R
```diff
@@ -88,6 +90,17 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+    cons_width = getOption("width")
+    cols_to_print = widths < cons_width
```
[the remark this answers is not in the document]
TysonStanley: Yes, I'll add the `=` to it.

R/print.data.table.R
```diff
@@ -164,3 +185,29 @@ shouldPrint = function(x) {
...
+  #   and compares it to the console width
+  widths = sapply(x, nchar_width)
+  if (class) widths = ifelse(widths < 6, 6, widths)
+  names = sapply(names, nchar_width)
+  dt_widths = ifelse(widths > names, widths, names)
+  rownum_width = if (row.names) nchar_width(nrow(x)) else 0
+  cumsum(dt_widths + 1) + rownum_width + 2
+}
+toprint_subset <- function(x, cols_to_print) {
+  if (nrow(x) == 1){
```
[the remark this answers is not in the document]
TysonStanley: I added more info in the code but I'll repeat it here :) 

When there was only one row in the data table, the `dim` and `dimnames` attributes were dropped when the subsetting is done to only keep the columns that will be printed. This function saves the attributes and then adds them back after the subset to `toprint` is done.

R/print.data.table.R
```diff
@@ -164,3 +185,29 @@ shouldPrint = function(x) {
...
+}
+dt_width = function(x, class, row.names, names) {
+  # gets the width of the data.table at each column
+  #   and compares it to the console width
+  widths = sapply(x, nchar_width)
+  if (class) widths = ifelse(widths < 6, 6, widths)
+  names = sapply(names, nchar_width)
+  dt_widths = ifelse(widths > names, widths, names)
+  rownum_width = if (row.names) nchar_width(nrow(x)) else 0
+  cumsum(dt_widths + 1) + rownum_width + 2
```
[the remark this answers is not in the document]
TysonStanley: Yes, you are correct. My personal preference is the way I've done it but that, by no means, is a good reason. What do you suggest? Would having another option be the fix here?

R/print.data.table.R
```diff
@@ -164,3 +185,29 @@ shouldPrint = function(x) {
 #   as opposed to printing a blank line, for excluding col.names per PR #1483
 cut_top = function(x) cat(capture.output(x)[-1L], sep = '\n')
 
+# to calculate widths of data.table and console
+nchar_width = function(x) {
+  max(nchar(as.character(x), type = "width"))
```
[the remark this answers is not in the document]
TysonStanley: Yes, this is in the `dt_width()` function in the `cumsum(dt_widths + 1)`.

R/print.data.table.R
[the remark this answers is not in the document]
TysonStanley: I'm now avoiding the coercion. I'm using the `toprint` object (so it is usually going to be quite small). Because `nchar()` works differently on matrices, I've adjusted the `nchar_width()` function, shown below.

```
nchar_width = function(x) {
  each_width = nchar(x, type = "width")
  widths = vector("numeric", ncol(x))
  for (i in 1L:ncol(x)){
    widths[i] = max(each_width[, i])
  }
  widths
}
```

Is it a problem to use a loop here? I figured the number of columns would likely be reasonable for the loop not to be excessive.

In what situation is `toprint` a `data.table` instead of a matrix? Treating it like a matrix is working for all of the tests I've done (large and small tables, `class=TRUE`, `col.names='none'`, etc.) and it is working consistently.

R/print.data.table.R
```diff
@@ -96,6 +109,10 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     } else {
       print(toprint, right=TRUE, quote=quote)
     }
+    if (trunc.cols && length(not_printed) > 0)
```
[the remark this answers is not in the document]
TysonStanley: This has been added to any integer I use (need to get in the habit of this...)

R/print.data.table.R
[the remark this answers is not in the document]
TysonStanley: Yes, I see your point. Honestly, not sure if that is worth pursuing an alternative to `nchar()` but wouldn't be opposed to it if it were necessary. Would `grepl()` and family help here? At least for the char counting?

R/print.data.table.R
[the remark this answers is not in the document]
MichaelChirico: effect is admittedly small case-by-case but once you're in the habit there are some aggregate effects. easy win for efficiency once you're used to it!

R/print.data.table.R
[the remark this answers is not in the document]
MichaelChirico: I don't think it's necessary. We'll leave it until it comes up. I suspect if anything it's a base R issue, not for us.

R/print.data.table.R
```diff
@@ -164,3 +185,29 @@ shouldPrint = function(x) {
 #   as opposed to printing a blank line, for excluding col.names per PR #1483
 cut_top = function(x) cat(capture.output(x)[-1L], sep = '\n')
 
+# to calculate widths of data.table and console
+nchar_width = function(x) {
+  max(nchar(as.character(x), type = "width"))
+}
+dt_width = function(x, class, row.names, names) {
```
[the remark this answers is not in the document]
TysonStanley: You are right. This has been removed.

R/print.data.table.R
```diff
@@ -164,3 +186,47 @@ shouldPrint = function(x) {
...
+  widths
+}
+# gets the width of the data.table at each column
+#   and compares it to the console width
+dt_width = function(x, class, row.names, col.names, names) {
+  widths = nchar_width(x)
+  if (class) widths = ifelse(widths < 6L, 6L, widths)
+  if (col.names == "none") names = sapply(names, nchar, type = "width") else names = 0L
+  dt_widths = ifelse(widths > names, widths, names)
+  rownum_width = if (row.names) max(nchar(as.character(rownames(x)), type = "width")) else 0L
```
MichaelChirico: This one is also an expensive call we can avoid... if `x` has e.g. 1e9 rows, we would convert 1e9 integers to character, then run `nchar`. Whereas from `data.table(1:1e9)` we can see it's always printed as all digits (no scientific notation). So `if(row.names)`, `rownum_width` is `as.integer(ceiling(log10(nrow(x)))+1)`

R/print.data.table.R
```diff
@@ -164,3 +186,47 @@ shouldPrint = function(x) {
...
+  widths
+}
+# gets the width of the data.table at each column
+#   and compares it to the console width
+dt_width = function(x, class, row.names, col.names) {
+  widths = nchar_width(x)
+  if (class) widths = ifelse(widths < 6L, 6L, widths)
+  if (col.names != "none") names = sapply(colnames(x), nchar, type = "width") else names = 0L
+  dt_widths = ifelse(widths > names, widths, names)
+  rownum_width = if (row.names) max(nchar(rownames(x), type = "width")) else 0L
```
MichaelChirico: This one is also an expensive call we can avoid... if `x` has e.g. 1e9 rows, we would convert 1e9 integers to character, then run `nchar`. Whereas from `data.table(1:1e9)` we can see it's always printed as all digits (no scientific notation). So `if(row.names)`, `rownum_width` is `as.integer(ceiling(log10(nrow(x)))+1)`

R/print.data.table.R
TysonStanley: Could we grab the last row name assuming it will be the largest? e.g. `rownames(x)[length(x)]`? We would only need that one to add the widths.

R/print.data.table.R
TysonStanley: Could we grab the last row name assuming it will be the largest? e.g. `rownames(x)[length(x)]`? We would only need that one to add the widths.

R/print.data.table.R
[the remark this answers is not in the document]
MichaelChirico: I think we should pick one and stick with it rather than add more options. We can play around with it in dev for a while & maybe change later.

Just some more thoughts: I guess most common is to have the first few columns with keys/"ID"-style rows. Then have "data" in the "rightmost" columns. So the tradeoff is -- showing all the key columns, vs showing some keys, some data. Still don't have a feel yet for which is more helpful to see in a preview.

R/print.data.table.R
TysonStanley: Sorry, not `length(x)` but `nrow(x)`? Is that too expensive of a call on tons of rows?

R/print.data.table.R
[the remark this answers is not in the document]
TysonStanley: That makes sense. I like the idea of differentiating between the keys and "data" in the preview. Definitely could get to some weird edge cases with this, depending on the console width and the number of keys...

R/print.data.table.R
TysonStanley: Wow. Ignore this question. Just ran some benchmarks with your code.

```
as.integer(ceiling(log10(nrow(DT)))+1)
#> 4
nchar(rownames(DT)[nrow(DT)], type = "width")
#> 3
bench::system_time(as.integer(ceiling(log10(nrow(DT)))+1))
#> process 
#>    19µs 
#>    real 
#>  28.9µs 
bench::system_time(nchar(rownames(DT)[nrow(DT)], type = "width"))
#> process 
#>   203µs 
#>    real 
#>   204µs 
```

I'll put your chunk in there.

R/print.data.table.R
```diff
@@ -164,3 +186,47 @@ shouldPrint = function(x) {
 #   as opposed to printing a blank line, for excluding col.names per PR #1483
 cut_top = function(x) cat(capture.output(x)[-1L], sep = '\n')
 
+# to calculate widths of data.table for PR #4074
+nchar_width = function(x) {
+  each_width = nchar(x, type = "width")
```
MichaelChirico: `apply(nchar(x, type='width'), 2L, max)` right? If so I'd drop this helper and just use

```
widths = apply(nchar(x, type='width'), 2L, max)
```

R/print.data.table.R
TysonStanley: It's even slightly faster than the simplified version using `nchar()`:

```
nchar(nrow(DT), type = "width")+1
#> 4
bench::system_time(nchar(nrow(DT), type = "width"))
#> process 
#>    29µs 
#>    real 
#>  29.4µs 
```

R/print.data.table.R
TysonStanley: Yep, that would do the same. Will replace.

R/print.data.table.R
MichaelChirico: `data.table` only keeps row names for consistency to `data.frame`; they're always `1:nrow(x)`, so `rownames(x)[length(x)]` is always `nrow(x)`. `nrow` will run almost instantaneously since it's an attribute of the column vectors, rather than needing to iterate over the vector to count rows, we just read the attribute. In-memory data is great!

`length(x)` is the same as `ncol(x)` -- this comes from the `list` inheritance. `data.table(a = 1, b = 2)` is "decorated" `list(a=1, b=2)`, the length of which is `2L`.

TysonStanley: Still can't believe how quick your code was. Would have never thought of it.

R/print.data.table.R
```diff
@@ -164,3 +186,39 @@ shouldPrint = function(x) {
 #   as opposed to printing a blank line, for excluding col.names per PR #1483
 cut_top = function(x) cat(capture.output(x)[-1L], sep = '\n')
 
+# to calculate widths of data.table for PR #4074
+# gets the width of the data.table at each column
+#   and compares it to the console width
+dt_width = function(x, class, row.names, col.names) {
+  widths = apply(nchar(x, type='width'), 2L, max)
+  if (class) widths = ifelse(widths < 6L, 6L, widths)
```
MichaelChirico: I really hate `base::ifelse` 😅 I would do `widths = pmax(widths, 6L)`

R/print.data.table.R
```diff
@@ -164,3 +186,39 @@ shouldPrint = function(x) {
...
+  if (col.names != "none") names = sapply(colnames(x), nchar, type = "width") else names = 0L
+  dt_widths = ifelse(widths > names, widths, names)
```
MichaelChirico: ditto here with `pmax`

R/print.data.table.R
TysonStanley: Is some of that hate from its unsafe behavior around coercion? Will update.

R/print.data.table.R
TysonStanley: Replaced.

R/print.data.table.R
```diff
@@ -164,3 +186,39 @@ shouldPrint = function(x) {
...
+  rownum_width = if (row.names) as.integer(ceiling(log10(nrow(x)))+1) else 0L
+  cumsum(dt_widths + 1L) + rownum_width + 1L
```
MichaelChirico: another micro-optimization -- we can do `+2 else 0L`. I think `if(!row.names)` the `+1L` in the next step is wrong. However I also just noticed this:

https://github.com/Rdatatable/data.table/issues/4083

So even if `row.names=FALSE`, it could still occupy some characters. I think you can safely ignore that in this PR, rather than complicating logic to detect whether `---` will be present.

R/print.data.table.R
MichaelChirico: Can see here... it's "slow" though probably not noticeable at the size of `ncol`:

https://github.com/Rdatatable/data.table/issues/2677

R/print.data.table.R
TysonStanley: I wondered about the `---` but hadn't done anything about it yet. Imagine we'll want to wait for that to be resolved before deciding any padding.

I believe the `+1L` in `cumsum(dt_widths + 1L)` should account for the space between the row names and the first column instead of needing to add another to the `rownum_width`.

TysonStanley: But I'll take off the `+1L` after the `rownum_width` bc I think you are right that it is unnecessary.

R/print.data.table.R
```diff
@@ -87,7 +89,19 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
   }
+  if (isFALSE(class) || (isTRUE(class) && col.names == "none")) abbs = ""
   if (quote) colnames(toprint) <- paste0('"', old <- colnames(toprint), '"')
+  if (isTRUE(trunc.cols)) {
+    # allow truncation of columns to print only what will fit in console PR #4074
+    widths = dt_width(toprint, class, row.names, col.names)
+    cons_width = getOption("width")
+    cols_to_print = widths <= cons_width
+    not_printed = colnames(toprint)[!cols_to_print]
+    if (sum(cols_to_print) == 0L) stop("Width of console too small to print a single column when `trunc.cols=TRUE`. Consider increasing the width of the console or use `trunc.cols=FALSE`.", call. = FALSE)
```
MichaelChirico: this should be `!any(cols_to_print)`.

I also think we shouldn't `stop` here -- rather unfortunate for the program to interrupt from printing side effects. I would simply print nothing (besides the "variable not shown..." message).

TysonStanley: Nice. That makes a lot of sense to let it run with only the message itself.

R/print.data.table.R
MichaelChirico: I think the `+1L` is right when `row.names=TRUE`

TysonStanley:
The "only" is in the wrong place in that commit message. It should say: "prints only the message when console too small to fit the first column".

R/print.data.table.R
```diff
@@ -164,3 +189,39 @@ shouldPrint = function(x) {
...
+  }
+}
+# message for when trunc.cols=TRUE and some columns are not printed
+trunc_cols_message = function(not_printed, abbs, class){
+  n = length(not_printed)
+  if (class) classes = paste0(" ", tail(abbs, n)) else classes = ""
+  not_printed_paste = paste0(not_printed, classes, collapse = ", ")
+  cat(sprintf(ngettext(n,
+                       paste0("1 variable not shown: %s"),
+                       paste0(n, " variables not shown: %s")),
```
MichaelChirico: Last thing: for #3935, `ngettext` will work best if we just pass plain formatting strings:

```
ngettext(n, "%d variable not shown: %s",
                       "%d variables not shown: %s"),
              n, not_printed_paste)
```

TysonStanley: Yes, I like that. And if it makes it easier for the translation, I'm all for it. It was the reason I used `ngettext()` in the first place but had rarely used it before.

TysonStanley:
Alright, as long as the tests look like they are passing with this last commit, I'll leave this alone.

TysonStanley:
This one adds the brackets (which I really like).

TysonStanley:
@MichaelChirico Alright, checks look good so I won't make any other changes unless you ask me to.

MichaelChirico:
Looks good to me! Thanks for your diligence!!

mattdowle:
Excellent.  @TysonStanley I've invited you to be project member. You'll need to accept the invitation displayed in your GitHub profile or GitHub repositories page. Then you can create branches directly in the main project.  Welcome!
I'll merge as-is. Then I'll move the news item up and bump the test numbers in a follow-up commit. (Quicker for me that way than bringing the fork up to date.)

jangorecki:
There should be better validation of argument values provided, for example following error message is very confusing.
```r
print(as.data.table(iris), trunc.cols=5L)
     Sepal.Length Sepal.Width Petal.Length Petal.Width   Species
            <num>       <num>        <num>       <num>    <fctr>
  1:          5.1         3.5          1.4         0.2    setosa
  2:          4.9         3.0          1.4         0.2    setosa
  3:          4.7         3.2          1.3         0.2    setosa
  4:          4.6         3.1          1.5         0.2    setosa
  5:          5.0         3.6          1.4         0.2    setosa
 ---                                                            
146:          6.7         3.0          5.2         2.3 virginica
147:          6.3         2.5          5.0         1.9 virginica
148:          6.5         3.0          5.2         2.0 virginica
149:          6.2         3.4          5.4         2.3 virginica
150:          5.9         3.0          5.1         1.8 virginica
Error in print.data.table(as.data.table(iris), trunc.cols = 5L) : 
  object 'not_printed' not found
```

TysonStanley:
Just made a PR for this #4766 that adds a quick check for a logical `trunc.cols` argument at the top of `print.data.table()`.

PR #4087, opened 2019-12-04
upgrade DT helpers works for custom library location

jangorecki:
By default when `lib.loc` or (`loc` in another places in base R) is NULL then `find.package` first looks for loaded namespaces paths, then .libPaths. When calling `update.dev.pkg` we must have had DT namespace loaded, this won't make the difference. This PR allows to call `update.dev.pkg` from DT pkg from library1 but check and upgrade DT in library2.

### DIFF
--- a/R/devel.R
+++ b/R/devel.R
@@ -1,9 +1,9 @@
 # nocov start
 
-dcf.lib = function(pkg, field, lib.loc){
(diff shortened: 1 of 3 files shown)


jangorecki:
moved those devel helpers from onAttach, they don't fit there.

PR #4090, opened 2019-12-06
test.data.table gains showProgress=interactive()

mattdowle:
With the errors and warnings on CRAN recently, the many lines of `Running test <num> ...` have been getting in the way. Added `showProgress=interactive()`.  Thereforefore they won't print in `R CMD check` and won't show on CRAN checks page.

I tried this in the past but the roadblock was that if the script failed in between two `test()` we wanted to know where.  So the idea was that `test()` printed its number on every call so we knew where it was. When a user runs `test.data.table()` the `\r` is honored and it shows rudimentary progress.  Whereas in the log file show by CRAN the `\r` is treated as `\n` and there's many lines of output.

* `test.data.table()` caller now catches the `prevtest` and puts that in a new error message, so we know where it got to should it fail in between `test()` calls.  I tested locally with a dummy error and that works; e.g. the output omits all the 'Running test <num> ...` lines and just shows this now:

```
> test.data.table()  # runs the main test suite of 5,000+ tests in /inst/tests/tests.Rraw
getDTthreads(verbose=TRUE):
  omp_get_num_procs()            8
  omp_get_thread_limit()         2147483647
  omp_get_max_threads()          8
  data.table is using 4 threads. See ?setDTthreads.
test.data.table() running: /home/mdowle/GitHub/data.table/data.table.Rcheck/data.table/tests/tests.Rraw
Error in eval(exprs[i], envir) : dummy stop in between test() calls
Error in test.data.table() :
  Failed after test 453 before the next test() call in /home/mdowle/GitHub/data.table/data.table.Rcheck/data.table/tests/tests.Rraw
Execution halted
```

* an aside fix is that if `test.data.table()` failed in that circumstance (i.e. in between two `test()` calls) it would not reset the options back to the user's values because the `options(oldOptions)` line wouldn't have been reached.  I didn't test that.

* there was another benefit of test() always outputting its test number. That was if the error was a segfault which caused the entire R process to halt.  In that case it was useful to see the last test number written to the log file.  We lose that with this PR.  However, I'm not sure how reliable that was and it's a very long time since we had a problem like that.  If that crops up again, see the new comment in the code in this PR for a more robust possibility to solve that.

This PR was probably easier now because of the work over the last year or so to move header and footer code up from inside the `.Rraw` script itself, up to the `test.data.table()` caller. Might explain why we didn't do this before because it was harder then to capture the last test number before the error.

* Another benefit is that any output from failed tests (i.e. within a test() call) should now be present in the last 13 lines of output displayed by CRAN 00check.log, too.   Rather than lots of 'Running test id <n> ...' lines after the error/warning filling up the last 13 lines.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,8 @@
 
 3. C internals have been standardized to use `PRI[u|d]64` to print `[u]int64_t`. This solves new warnings from `gcc-8` on Windows with `%lld`, [#4062](https://github.com/Rdatatable/data.table/issues/4062), in many cases already working around `snprintf` on Windows not supporting `%zu`. Release procedures have been augmented to prevent any internal use of `llu`, `lld`, `zu` or `zd`.
 
+4. `test.data.table()` gains `showProgress=interactive()` to suppress the thousands of `Running test id <num> ...` lines displayed by CRAN checks when there are warnings or errors.
+
 
 # data.table [v1.12.6](https://github.com/Rdatatable/data.table/milestone/18?closed=1)  (18 Oct 2019)
 
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -282,11 +282,14 @@ test = function(num,x,y=TRUE,error=NULL,warning=NULL,message=NULL,output=NULL,no
     # prevtest to a temp file so we know where it got to from this R process. That should be more reliable
     # than what we were doing before which was for test() to always write its test number to output (which might
     # not be flushed to the output upon segfault, depending on OS).
-  } else {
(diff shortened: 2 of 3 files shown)


jangorecki:
minor change added to be more future proof, in case we will use `showProgress` down the code later on but we will forget to define it for an interactive users of `test()`.

mattdowle:
good idea.  I did minor edit of your comment above in-place just to change 'interactive' to 'developer' ... think that's what you meant iiuc. Since the interactive part of this PR is more about users running test.data.table() vs R CMD check.

PR #4106, opened 2019-12-13
output platform details when tests fail

mattdowle:
When tests failed with an error, the platform details line was missed from output. Thus we couldn't answer the question in #4099's SO question's comment for sure of whether the Solaris flavor with the error was big endian.
This PR ensures that platform line is always present.


MichaelChirico:
@mattdowle possibly related: #4038

mattdowle:
@MichaelChirico Very nice. I didn't know about those env variables. Merged.

On this one, there seems to be a Linux CI connectivity problem on Travis currently.  AppVeyor and OSX are passing though which is surely enough to merge in this case.

PR #4109, opened 2019-12-13
fix compiler warning

tdhock:
Hi this small fix is to avoid the compiler warning "ignoring return value of 'strtod', declared with attribute warn_unused_result [-Wunused-result]" which I get when compiling data.table with gcc on CentOS.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -43,7 +43,8 @@ Authors@R: c(
   person("Manish","Saraswat",      role="ctb"),
   person("Morgan","Jacob",         role="ctb"),
   person("Michael","Schubmehl",    role="ctb"),
-  person("Davis","Vaughan",        role="ctb"))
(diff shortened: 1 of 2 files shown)


tdhock:
that is gcc-6.2.0

mattdowle:
Thanks, @tdhock!  I tried that flag locally (gcc 7.4.0) but it didn't spot it, oddly. Maybe @jangorecki can add `-Wunused-result` to CI.
I added you to contributors in description, and invited you to be project member so you can create branches in the main project now. (You'll need to accept the invitation which appears in your GitHub profile or GitHub repository page).

mattdowle:
Seem to be getting more and more connection / timeouts on Travis these days.  Proceeding to merge without Travis Linux.  Travis MacOS passed fine, and AppVeyor passed.

tdhock:
wow thanks

PR #4117, opened 2019-12-16
extend checks for UTC inputs for robustness

### LINKED ISSUE #4116: fast IDate/Time methods for UTC can be extended to when we can infer tz is UTC
Currently we test:

```
if (inherits(x, "POSIXct") && identical(attr(x, "tzone", exact = TRUE), "UTC")) 
```

However, if `'tzone'` attribute is missing and `Sys.timezone()` is UTC (actually, `%chin% c('GMT', 'UTC', 'Etc/UTC', 'Etc/GMT', 'GMT-0', 'GMT+0', 'GMT0')` is fine), we can do the same shortcut

MichaelChirico:
Closes #4116 

You can confirm all these timezones are the same like:

```
utc_tz = c("Etc/GMT", "Etc/UTC", "GMT", "GMT-0", "GMT+0", "GMT0", "UTC")
T0 = Sys.time() # ditto for .POSIXct(0) and .POSIXct(-1e6)
length(unique(sapply(utc_tz, function(tz) format(T0, tz=tz))))
# [1] 1
```

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -117,6 +117,14 @@ do_patterns = function(pat_sub, all_cols) {
   return(matched)
 }
 
+# check UTC status
+is_utc = function(tz) {
+  # via grep('UTC|GMT', OlsonNames(), value = TRUE)
+  utc_tz = c("Etc/GMT", "Etc/UTC", "GMT", "GMT-0", "GMT+0", "GMT0", "UTC")
+  if (is.null(tz)) tz = Sys.timezone()
+  return(tz %chin% utc_tz)
+}
+
 # nocov start #593 always return a data.table
 edit.data.table = function(name, ...) {
   setDT(NextMethod('edit', name))[]
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,8 @@
 
 ## NOTES
 
+1. `as.IDate`, `as.ITime`, `second`, `minute`, and `hour` now recognize UTC equivalents for speed: GMT, GMT-0, GMT+0, GMT0, Etc/GMT, and Etc/UTC, [#4116](https://github.com/Rdatatable/data.table/issues/4116).
+
 
 # data.table [v1.12.8](https://github.com/Rdatatable/data.table/milestone/15?closed=1)  (09 Dec 2019)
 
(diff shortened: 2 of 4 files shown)


MichaelChirico:
Really dunno why we're getting this on Travis, not able to reproduce:

```
<simpleError in if (tz == "UTC") {    z <- floor(unclass(x)/86400)    attr(z, "tzone") <- NULL    .Date(z)} else as.Date(as.POSIXlt(x, tz = tz)): argument is of length zero>
```

That snippet comes from `base::as.Date.POSIXct` and it should be from `if (NULL == 'UTC')` but the first tracing commit shows there was never a case of getting `tz=NULL` passed to `as.Date.POSIXct`

MichaelChirico:
I find this PR very strange in that the initial commit didn't work on Travis and the current commit does. any ideas?

MichaelChirico:
difference in code is

```
if (is.null(tz)) tz=''
as.IDate(as.Date(x, tz = tz, ...))
```

vs

```
as.IDate(as.Date(x, tz =  if (is.null(tz)) tz='' else tz, ...))
```

I can't see how in the first one `tz` was passed as `NULL` to `as.Date` 🤔

R/utils.R
```diff
@@ -117,6 +117,14 @@ do_patterns = function(pat_sub, all_cols) {
   return(matched)
 }
 
+# check UTC status
+is_utc = function(tz) {
+  # via grep('UTC|GMT', OlsonNames(), value = TRUE)
+  utc_tz = c("Etc/GMT", "Etc/UTC", "GMT", "GMT-0", "GMT+0", "GMT0", "UTC")
```
MichaelChirico: I sorted this alphabetically... perhaps we should sort it by expected frequency instead? Does `%chin%` have a short-out to return once the first match is found?

mattdowle: No. Good idea. If its LHS is length 1, then it should just loop through the RHS and stop early if and when it is found.  (Btw, sorting the input was good thought but doesn't help as it's an order-n two-pass approach using the now-well- established truelength clobber technique.)

mattdowle: When `%chin%` does short-circuit as you suggested, then yes this input should be sorted by expected frequency.

mattdowle: Done: #4121

PR #4119, opened 2019-12-17
(void) rather than ignored=

mattdowle:
Follow up to #4109 
`(void)` rather than `ignored=` to avoid the unused-variable warning:
```
fread.c:1133:12: warning: unused variable ‘ignored’ [-Wunused-variable]
     double ignored = strtod(ch, &end);
```
@tdhock Does this still pass your compiler without warning?


mattdowle:
OSX fail seems to be resource issue / stalled.  Merging without it (Linux and Windows ok).

tdhock:
yes this is fine. (no warnings)

PR #4121, opened 2019-12-17
chin and chmatch length 1 speedup

mattdowle:
Implemented Michael's suggestion here: https://github.com/Rdatatable/data.table/pull/4117#discussion_r358378409

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 ## NEW FEATURES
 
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Great! Let's sort the table in `is_utc` by frequency then (UTC>GMT>Etc/UTC>...). Here, or a separate PR?

mattdowle:
Ok. In this PR might as well. I'll wait for your commit before merging.

Issue #3540, opened 2019-05-04
CSVY wishlist


MichaelChirico:
Filing a follow-up to #1701 & #3534 which handled basic implementation of CSVY support to centralize a wish list for more advanced functionality for CSVY:

### Both fread and fwrite

- [ ] Support for writing a factor's levels & then reading them
- [ ] `yaml = '/path/to/file'` -- store/read metadata stored outside the file itself
- [ ] `yaml = 'auto'` by default -- read/write csvy files automatically by detecting the `csvy` extension

### `fread`

- [ ] Fully sync `fread` arguments with what's output from `fwrite` by default
- [ ] Expand recognized `colClasses` following #2545 
- [ ] auto yaml detection using file contents even if extension is not `.csvy`

### `fwrite`

- [ ] `yaml = list()` -- supply custom key/value pairs to be included in the metadata

rossholmberg:
I'd love to see more user-oriented metadata. Description/note options would be fantastic for storing something along the lines of a brief data readme. I played with something like this myself a while ago (it's all on GitHub), and used an object attribute to store these things in the R object. Here's a section of my readme showing what I mean:
https://github.com/rossholmberg/rcsv/blob/master/README.md#notes-an-added-detail-to-your-datasets

I only had "notes" for the table as a whole, but per-column would be even better.

MichaelChirico:
as in, how to provide customized metadata right? if so yes!! (if not please correct me 😁) something I've thought about but I'm not sure the best API for enabling this. could be that we force user to provide an `attribute` but that feels inconvenient.

jangorecki:
I would lean towards generalising to store DT/cols attributes in yaml metadata. Then we don't need to worry if it is a column attribute, or a table attribute, what is its purpose. This way we handle factor levels, datetime origin/timezone etc. with column attributes, and sorted attribute of a data.table.
Handling indexes is possible (adding technical `__index_cols` integer columns) but probably should be disabled by default it costs extra space, read/write time, and at the same time it is something that can be easily and pretty fast computed.

MichaelChirico:
Makes sense. Agree we shouldn't be writing the index vector, but it would be useful to have a `sortedBy` attribute (or similar) that can indicate the data is already sorted by some columns.

rossholmberg:
I agree Jan, that would be great to be able to store attributes at both DT and col levels.
[...]

PR #4123, opened 2019-12-17
Adds user-created parameters in fread's yaml

eliocamp:
Part of #3540

### DIFF
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -82,33 +82,36 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
       return(invisible())
     }
   }
-  write_yaml <- yaml == TRUE || (is.list(yaml) && length(yaml) != 0)
+  write_yaml = isTRUE(yaml) || (is.list(yaml) && length(yaml) != 0L)
   yaml = if (!write_yaml) "" else {
     if (!requireNamespace('yaml', quietly=TRUE))
       stop("'data.table' relies on the package 'yaml' to write the file header; please add this to your library with install.packages('yaml') and try again.") # nocov
-    schema_vec = sapply(x, class)
-    # multi-class objects reduced to first class
-    if (is.list(schema_vec)) schema_vec = sapply(schema_vec, `[`, 1L)
-    # as.vector strips names
-    schema_vec = list(name=names(schema_vec), type=as.vector(schema_vec))
-    yaml_header = list(
-      source = sprintf('R[v%s.%s]::data.table[v%s]::fwrite',
-                       R.version$major, R.version$minor, format(tryCatch(utils::packageVersion('data.table'), error=function(e) 'DEV'))),
-      creation_time_utc = format(Sys.time(), tz='UTC'),
-      schema = list(
-        fields = lapply(
-          seq_along(x),
-          function(i) list(name=schema_vec$name[i], type=schema_vec$type[i])
-        )
-      ),
-      header=col.names, sep=sep, sep2=sep2, eol=eol, na.strings=na,
-      dec=dec, qmethod=qmethod, logical01=logical01
-    )
-    if (is.list(yaml)) {
-      yaml_header <- c(yaml_header, yaml)  
+    
+    if (isTRUE(yaml)) {
+      yaml = generate_yaml(x = x, col.names = col.names, sep = sep, 
+                            sep2 = sep2, eol = eol, na = na, dec = dec, 
+                            qmethod = qmethod, logical01 = logical01)  
+    } else {
+      names = names(yaml)
+      if (is.null(names)) {
+        names = rep("", length(yaml))
+      }
+      unnamed = names == ""
+      true_yaml = vapply_1b(yaml, isTRUE)
+      
+      generated_yaml = unnamed & true_yaml
+      if (any(unnamed & !true_yaml)) {
+        stop("`yaml` contains unnamed elements that are not `TRUE`")
+      }
+      if (any(generated_yaml)) {
+        yaml_header = generate_yaml(x = x, col.names = col.names, sep = sep, 
+                                    sep2 = sep2, eol = eol, na = na, dec = dec, 
+                                    qmethod = qmethod, logical01 = logical01)  
+        yaml = c(yaml_header, yaml[!generated_yaml])
+      }
     }
     
-    paste0('---', eol, yaml::as.yaml(yaml_header, line.sep=eol), '---', eol) # NB: as.yaml adds trailing newline
+    paste0('---', eol, yaml::as.yaml(yaml, line.sep=eol), '---', eol) # NB: as.yaml adds trailing newline
   }
   file = enc2native(file) # CfwriteR cannot handle UTF-8 if that is not the native encoding, see #3078.
   .Call(CfwriteR, x, file, sep, sep2, eol, na, dec, quote, qmethod=="escape", append,
@@ -116,4 +119,3 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
         showProgress, is_gzip, bom, yaml, verbose)
   invisible()
 }
-
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -83,12 +83,13 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
     }
   }
   write_yaml = isTRUE(yaml) || (is.list(yaml) && length(yaml) != 0L)
-  yaml = if (!write_yaml) "" else {
+  yaml_text = ""
+  if (write_yaml) {
     if (!requireNamespace('yaml', quietly=TRUE))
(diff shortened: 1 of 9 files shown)


R/fwrite.R
```diff
@@ -82,7 +82,8 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
       return(invisible())
     }
   }
-  yaml = if (!yaml) "" else {
+  write_yaml <- yaml == TRUE || (is.list(yaml) && length(yaml) != 0)
```
MichaelChirico: I guess this won't work in general:

```
list(1, 2, 3, 4, list(1, 4, 3, 2)) == TRUE
```

I think the LHS should be `isTRUE(yaml)` instead.

Please also use `0L` for `length` which is always `integer` 😬

R/fwrite.R
```diff
@@ -103,6 +104,10 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
       header=col.names, sep=sep, sep2=sep2, eol=eol, na.strings=na,
       dec=dec, qmethod=qmethod, logical01=logical01
     )
+    if (is.list(yaml)) {
+      yaml_header <- c(yaml_header, yaml)
```
MichaelChirico: (open-ended) is there any extra checks we should do on the `data.table` side here of the user input? or will it be fine to leave it up to `yaml::as.yaml` to do all validation?

inst/tests/tests.Rraw
```diff
@@ -16400,6 +16400,12 @@ test(2124.1, format(as.ITime(t0)), '00:00:00')
 test(2124.2, format(as.IDate(t0)), '2019-10-01')
 Sys.setenv(TZ=oldtz)
 
+# Extra arguments in yaml
+data = data.table(x = 1, y = 100)
+file = tempfile()
```
MichaelChirico: `unlink(file)` afterwards

R/fwrite.R
eliocamp: That was my thought, yes.

inst/tests/tests.Rraw
```diff
@@ -16400,6 +16400,12 @@ test(2124.1, format(as.ITime(t0)), '00:00:00')
...
+fwrite(data, file, yaml =  list(something_else = "o hai!"))
+x = fread(file, yaml = TRUE)
+test(2125, attr(x, "yaml_metadata")$something_else, "o hai!")
```
MichaelChirico: another test that the rest of the `yaml_metadata` object is "as expected" would be nice, to be sure we're doing as advertised -- both writing the "standard" metadata & the user-supplied extras.

Just a test of a single key/value should be enough

eliocamp: Something like `test(2125.02), attr(x, "yaml_metadata")$schema$fields[[1]]$name,  "x")`?

MichaelChirico:
Thanks! Looks good overall.

What do you think of an API like this?

```
# @export
generate_yaml = function(x) # standard YAML we generate now for object x

my_yaml = c(generate_yaml(x), list(something_else = 'o hai!'))
```

A bit more cumbersome, but offers the flexibility to fully customize the YAML header (potentially excluding/editing some key-values we put in there, e.g.).

PS please edit the `man` page accordingly and add a NEWS item

inst/tests/tests.Rraw
MichaelChirico: yep that should work

mattdowle:
LGTMT. Great. Can see some outstanding (e.g. news item and man page) so I added the WIP tag. When you remove the WIP tag, and @MichaelChirico has approved with green tick, I'll merge.
You're already a project member so you can create branches directly in the main project. Makes it easier to commit to each other's PRs. I don't see you in DESCRIPTION though as a contributor; please add yourself there in this PR. Your previous PR (#3685) did touch a .R file but only a warning message so at the time I won't have considered that a code change.  Whereas this new PR does contain code/logic changes.

jangorecki:
those new tests has to be escaped, they uses `yaml` packages which is a suggested dependency

eliocamp:
> What do you think of an API like this?
> 
> ```
> # @export
> generate_yaml = function(x) # standard YAML we generate now for object x
> 
> my_yaml = c(generate_yaml(x), list(something_else = 'o hai!'))
> ```
> 
> A bit more cumbersome, but offers the flexibility to fully customize the YAML header (potentially excluding/editing some key-values we put in there, e.g.).

I like the flexibility with this, but, yeah, it's very verbose. Also I don't like the idea of having to repeat the name of the object inside `frwite`. It opens up the very real possibility of annoying bugs if the user  makes a typo or something like that (I wouldn't like to have to debug something like this `fwrite(some_data, yaml = generate_yaml(som_data))`). 

I like `TRUE` as a shortcut for "just create the standard yaml that data.table creates", so how about this:

```r 
fwrite(x, yaml = list(TRUE, and_now = "something completely different"))
```

I'm also toying around the possibility of making it so that the `yaml` parameter accepts a function that takes a data.table and returns a list (that then will be parsed with as.yaml).

Then, you could write:

```r
fwrite(x, yaml = my_yaml_function)
```

for maximum flexibility and portability.

eliocamp:
I've added some logic and tests for the proposed interface.

R/fwrite.R
```diff
@@ -82,33 +82,40 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
       return(invisible())
     }
   }
-  yaml = if (!yaml) "" else {
+  write_yaml = isTRUE(yaml) || (is.list(yaml) && length(yaml) != 0L)
+  yaml = if (!write_yaml) "" else {
```
MichaelChirico: minor stylistic point -- i always prefer `if (a) X then Y` to `if (!a) Y then X`. i think what you're going for here is being able to see `""` inline but I think code collapse (to `{...}`) is just as good

R/utils.R
```diff
@@ -117,6 +117,37 @@ do_patterns = function(pat_sub, all_cols) {
...
+
+generate_yaml = function(x, col.names = TRUE, sep = ",", sep2 = "", 
+                          eol = if (.Platform$OS.type=="windows") "\r\n" else "\n",
+                          na = "", dec = "", qmethod = "double", 
+                          logical01 = getOption("datatable.logical01", FALSE)) {
+  schema_vec = sapply(x, class)
+  # multi-class objects reduced to first class
+  if (is.list(schema_vec)) schema_vec = sapply(schema_vec, `[`, 1L)
+  # as.vector strips names
+  schema_vec = list(name=names(schema_vec), type=as.vector(schema_vec))
```
MichaelChirico: I'm looking at the code for `schema_vec` again and it's a bit of a mess.

What about `schema_vec = sapply(x, class, simplify = FALSE)`

Then in the `seq_along` loop we use the names of this object and the first values:

```
list(
  name = names(schema_vec)[i],
  type = schema_vec[[ii]][1L]
)
```

R/fwrite.R
eliocamp: That's just the old code :laughing: 
What I sometimes do is to set the default value unconditionally and then change the default if necessary. In this case it would be something like 
```r
yaml_text = ""
if (yaml) {
 yaml_text = blahblahblah
}
```

Would that work?

MichaelChirico: that's fine w me

R/utils.R
eliocamp: I hadn't really look at it (just copy-pasted it into the new function). However, I see now that we are reducing multiclass objects to their first class. Why is that? Can't we store class as a character vector? It is supported by yaml, isn't it?

MichaelChirico: I was thinking the same thing on my first re-read. I'm not sure if `fread` would support that though since `colClasses` needs length-1 classes;

OTOH maybe it's best to make `fread` to the `[1]` and leave the full class in there from `fwrite`

eliocamp: Maybe it's a discussion for another PR. For the time being, I left it as it is.

jangorecki:
@eliocamp are you still planning to address the feedback provided here?

eliocamp:
Sorry, had forgotten about this PR. Thanks for the reminder. 

I think I addressed all the feeback. What's left is maybe expanding the functionality as detailed on my previous comment: https://github.com/Rdatatable/data.table/pull/4123#issuecomment-569441444

Although maybe that's best left for another PR? Let me know if you think anything's missing.

jangorecki:
Yes, functionality is there, but we should also have entry in news, description, etc. @mattdowle commented pointed out that well https://github.com/Rdatatable/data.table/pull/4123#issuecomment-566758186

eliocamp:
Ok, there it is. I had to change Test numbers too.

MichaelChirico:
@eliocamp is this worth reviving? I don't have a sense of how much uptake there has been of `fwrite(yaml=TRUE)`. WDYT?

eliocamp:
Woah, all the memories! 

Yeah, I don't think lots of people are clamoring for this to be implemented. I would be fine with dropping this PR.

PR #4132, opened 2019-12-19
Update melt.data.table.Rd

jwijffels:
I always forget what melt/dcast does. As the information is already in dcast.data.table, I thought that adding this in the docs of melt.data.table would not hurt


MichaelChirico:
Great! Love this kind of PR! Great addition to the documentation.

Issue #2720, opened 2018-04-01
fread parse standard data.table print output


jangorecki:
I would find it convenient to just copy-paste data.table as printed to output.
```r
      ID  time nTest
 1: 1650 6.714    91
 2: 1652 6.376    91
 3: 1648 6.338    91
 4: 1509 5.446     1
 5: 1779 4.541    13
 6: 1646 4.060    91
 7: 1644 3.572    91
 8: 1642 3.416    91
 9: 1437 2.261   528
10: 1151 1.828     2
```
This text copy-pasted from output could be just put in `fread("[ctrl+v]")`. Something like checking `1:` in first col. Makes easier to get some small data.table copy-pasted from Rout log.

MichaelChirico:
How sophisticated do you want it to be?

An extremely rinky-dink version is:

```
fread_rout = function(x) {
  x = strsplit(x, '\n', fixed = TRUE)[[1L]]
  x = sub('^[ \t]+', '', sub('[ \t]+$', '', x))
  x = grep('---', x, fixed = TRUE, invert = TRUE, value = TRUE)
  coln = strsplit(x[1L], '\\s+')[[1L]]
  x = transpose(strsplit(x[-1L], '\\s+'))[-1L]
  setDT(x)
  setnames(x, coln)
  x[ , (coln) := lapply(.SD, function(x) {
    num = suppressWarnings(as.numeric(x))
    if (any(is.na(num[x != 'NA']))) return(x)
    return(num)
  })]
  x[]
}
```

It won't be able to handle anything weird (embedded commas, etc). But I guess that's not the use case here.

MichaelChirico:
we could just add this into the `text` argument's functionality perhaps?

jangorecki:
I would close this issue, because the idea was not that smart, printing formats depends on my factors. It is easy to do function like this for the most common examples, but difficult to make it work more widely, potentially bringing many reports about something being not well parsed.

PR #4142, opened 2019-12-28
Added fread_output() to read data.table/data.frame objects from Rout log. Fix to issue #2720

ZeroDawn0D:


jangorecki:
I think this should be plugged into fread, we don't need a separate function for that. Is it portable across different platforms? I seen split by `\n` which AFAIK may not be portable.

R/fread_output.R
```diff
@@ -0,0 +1,23 @@
...
+                         stringsAsFactors = FALSE,
+                         showProgress=getOption("datatable.showProgress", interactive()),
+                         data.table=getOption("datatable.fread.datatable", TRUE),
+                         logical01=getOption("datatable.logical01", FALSE)
+)
+{
+        s = strsplit(str, '\n', fixed = TRUE)[[1L]]
+        s <- sapply(s, trimws);
+        s[1] <- paste("fread_output_TEMP_COLUMN_NAME", s[1]);
+        str <- s[1];
```
jangorecki: What is the point of this assignment if `str` is not being used later

PR #4148, opened 2019-12-31
test rtools40, #4073

jangorecki:
Follow up of #4073.
This PR is not going to be merged, it is for testing rtools40. Conclusions from those tests will be left here.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -82,7 +82,7 @@ build: # build data.table sources as tar.gz archive
   tags:
     - windows
   before_script:
-    - export PATH="/c/$R_DIR/bin:/c/Rtools/bin:$PATH"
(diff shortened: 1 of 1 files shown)


jangorecki:
setting up path to rtools40 is ignored
```
export PATH="/c/$R_DIR/bin:/c/rtools40/bin:$PATH"
```
build log still shows
```
c:/Rtools/mingw_32/bin/gcc
```
which is Rtools35

jangorecki:
Rtools40 seems to be a new tools, made from scratch, not really an upgrade of the current Rtools. It seems to advice to use R-testing binaries installation. We use R stable and R-devel which seems to not the same. Setup procedure is also different. I think it make sense to wait for Rtools40 to be in a non-experimental state before proceeding.

PR #4153, opened 2019-12-31
error message for S4 objects

### LINKED ISSUE #4135: Notify users that S4 object are not supported for fifelse and fcase
Following up on issue #4131 I think we should notify users that `fifelse` and `fcase` do not support S4 Object. I hence suggest that we add a line of C code like `if (isS4(x))`.

2005m:
closes #4135


2005m:
I have used a nested `if` at C level to avoid stack imbalance.

PR #4154, opened 2019-12-31
Print list-column dims

### LINKED ISSUE #3671: show dimensions of list columns with DT
Hello everyone,

I am all IN for a better integration of DT in the `tidyverse` ecosystem. DT is really great and I think some features of the `tidyverse` could be very useful in a DT setting.

In particular, I am a huge fan of `purrr` and `list columns`. Consider this.

```
> tibble(group = c(1,1,1,2,2,2),
+            val = list(list(1,2,3),list(1,2,3),list(1,2,3),
+                       list(1,2,3),list(1,2,3),list(1,2,3))) 
# A tibble: 6 x 2
  group val       
  <dbl> <list>    
1     1 <list [3]>
2     1 <list [3]>
3     1 <list [3]>
4     2 <list [3]>
5     2 <list [3]>
6     2 <list [3]>

```
As you can see, I can clearly see the dimension (or length) of each element of the list-column.
As far as I know, this is not possible with DT

```
>  data.table(group = c(1,1,1,2,2,2),
+            val = list(list(1,2,3),list(1,2,3),list(1,2,3),
+                       list(1,2,3),list(1,2,3),list(1,2,3))) 
   group    val
1:     1 <list>
2:     1 <list>
3:     1 <list>
4:     2 <list>
5:     2 <list>
6:     2 <list>

```
This is a bit unfortunate. Here the idea is to use `furrr` (which leverages `purrr`) to easily run computations in parallel across the list columns.
```

> mydf[,  newval := future_map(val, ~length(.x))]
> mydf
   group    val newval
1:     1 <list>      3
2:     1 <list>      3
3:     1 <list>      3
4:     2 <list>      3
5:     2 <list>      3
6:     2 <list>      3
```

Being able to see the dimension of the embedded list (or DT) is very important when one creates list columns with whole data.frames and run regressions on them. What do you think? Could the printing of DT show an output similar to the printing of the tibble?

Thanks!

TysonStanley:
Closes #3671

Prints dims with list-columns. For example:

```r
DT = data.table(
  x = 1:2,
  y = list(data.table(x = 1,
                      y = 1),
           data.table(x = 2,
                      y = 2))
)
print(DT, class=TRUE)
#>        x                 y
#>    <int>            <list>
#> 1:     1 <data.table[1x2]>
#> 2:     2 <data.table[1x2]>

DT = data.table(
  x = 1:2,
  y = list(list(x = 1,
                y = c("yes", "no")),
           list(x = 2,
                y = 2))
)
print(DT, class=TRUE)
#>        x         y
#>    <int>    <list>
#> 1:     1 <list[2]>
#> 2:     2 <list[2]>
```

This provides some additional information without printing too much and costs relatively little since it only applies to what is being printed, not the entire data table.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -1631,7 +1631,7 @@ test(557, DT[, f(.SD), by=x], error="[.]SD is locked.*reserved for possible futu
 # Test printing on nested data.table, bug #1803
 DT = data.table(x=letters[1:3], y=list(1:10, letters[1:4], data.table(a=1:3,b=4:6)))
 test(558, capture.output(print(DT)),
-          c("   x               y", "1: a 1,2,3,4,5,6,...", "2: b         a,b,c,d", "3: c    <data.table>"))
+     c("   x                 y", "1: a   1,2,3,4,5,6,...", "2: b           a,b,c,d", "3: c <data.table[3x2]>"))
 test(559, setkey(DT,x)["a",y][[1]], 1:10)   # y is symbol representing list column, specially detected in dogroups
 
 # Test renaming of .N to N
@@ -8280,11 +8280,11 @@ DT1  = data.table(lcol = list(list(1:3), list(1:3), list(1:3)),
                   xcol = as.complex(icol), ocol = factor(icol, ordered = TRUE),
                   fcol = factor(icol))
 test(1610.1, capture.output(print(DT1, class=TRUE)),
-     c("     lcol  icol  ncol   ccol   xcol  ocol   fcol",
-       "   <list> <int> <num> <char> <cplx> <ord> <fctr>",
-       "1: <list>     1     1      a   1+0i     1      1",
-       "2: <list>     2     2      b   2+0i     2      2",
-       "3: <list>     3     3      c   3+0i     3      3"))
+     c("        lcol  icol  ncol   ccol   xcol  ocol   fcol",
+       "      <list> <int> <num> <char> <cplx> <ord> <fctr>",
+       "1: <list[1]>     1     1      a   1+0i     1      1",
+       "2: <list[1]>     2     2      b   2+0i     2      2",
+       "3: <list[1]>     3     3      c   3+0i     3      3"))
 DT2 = data.table(
   Dcol = as.Date('2016-01-01') + 0:2,
   Pcol = as.POSIXct('2016-01-01 01:00:00', tz = 'UTC') + 86400L*(0:2),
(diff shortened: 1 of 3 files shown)


TysonStanley:
If I understand correctly, the current way for `data.table` to print named vectors and arrays is to present the first 6 elements and then `...` so we wouldn't be printing the dimensions for those. As it is, the print method uses `is.atomic()` to make the decision about the printing of the class name (and with this PR the dimensions). Would we want to change that and have the class and dimensions printed for named vectors and arrays?

As for the s4 classes, I've added functionality to handle those and have tests for it.

TysonStanley:
To clarify:

```r
s4class <- setClass("ex_class", slots = list(x = "integer", y = "character", z = "numeric"))
array1 = array(1:4, 2)
array2 = array(1:8, c(2,2))
array3 = array(1:16, c(2,2,2))

DT = data.table::data.table(
  int = 1L:2L,
  list = list(list(x = 1,
                y = c("yes", "no")),
           list(x = 2,
                y = 2)),
  s4 = list(s4class(x = 1L, y = "thing1", z = 3.4),
            s4class(x = 2L, y = "thing2", z = 2.5)),
  array1 = list(array1[1, drop = FALSE],
                array1[2, drop = FALSE]),
  array2 = list(array2[,1, drop = FALSE],
                array2[,2, drop = FALSE]),
  array3 = list(array3[,,1, drop = FALSE],
                array3[,,2, drop = FALSE]),
  named_vec = list(c("var1" = 1, "var2" = 2),
                   c("var1" = 3, "var2" = 4))
  )
print(DT, class=TRUE)
#>      int      list            s4 array1 array2  array3 named_vec
#>    <int>    <list>        <list> <list> <list>  <list>    <list>
#> 1:     1 <list[2]> <ex_class[3]>      1    1,2 1,2,3,4       1,2
#> 2:     2 <list[2]> <ex_class[3]>      2    3,4 5,6,7,8       3,4
```

inst/tests/tests.Rraw
```diff
@@ -8872,13 +8872,13 @@ test(1635.2, fread(text, fill=TRUE), setnames(ans1[, 1:7], c(letters[1:5], paste
 # testing function type in dt, #518
 dt = data.table(x=1, y=sum)
 test(1636.1, class(dt$y), "list")
-test(1636.2, any(grepl("1: 1 <function>", capture.output(print(dt)))), TRUE)
+test(1636.2, any(grepl("1: 1 <function\\[1\\]>", capture.output(print(dt)))), TRUE)
```
MichaelChirico: i would just use `fixed=TRUE`

MichaelChirico:
I'm happy if codecov is happy re:tests :)

Looks good overall; I just wonder whether we should patch up #3414 first and add this on to that

TysonStanley:
That makes sense to me. Seems like #3414 is definitely broader in scope. 

Question: Does #3414 mostly have to do with printing list columns with custom classes and arrays/named vectors/matrices? That’s what it seemed to me from the discussion about using `toString()`. Or does it also include individuals being able to customize printing of, say, data tables in a list column?

MichaelChirico:
Technically that's enabled, yep, through the `format_list_item` generic. So users could define `format_list_item.data.table`.

TysonStanley:
I like the flexibility. As for this PR, I’ll wait to hear back from you once things are more clear on #3414

mattdowle:
Btw, why is your id @TysonStanley but your name Tyson Barrett?  It doesn't matter really other than it catches me out w.r.t. to checking your name is in DESCRIPTION, and when thanking you in news.

inst/tests/tests.Rraw
mattdowle: +1 I updated these old tests to use `output=` too.

TysonStanley:
Great question. My middle name is Stanley and I thought I was being creative but, in hindsight, would have gone with something like "@TysonBarrett" instead... Thanks for working around it.

jangorecki:
I really like the idea that Michael said about using format_list_item to achieve what we need in this PR. Do you think you could follow-up later on with required changes when we have generic method in place?

mattdowle:
@jangorecki anyone can follow up. Tyson has achieved what he wanted, which closes an issue. We can't ask for more than that.

Issue #4159, opened 2020-01-05
returning .SD by group doesn't unlock .SD; and GForce [[ non-atomic type causes trouble


MichaelChirico:
Edit: Playing with one example uncovered two pretty unrelated bugs; not sure if I should open a separate issue. For now, I've edited the title to reflect both bugs. Here's the original description:

Related to #4156, I was playing around with `tidyfast::dt_nest`'s output from their example:

```
dt <- data.table(
  x = rnorm(1e5),
  y = runif(1e5),
  grp = sample(1L:3L, 1e5, replace = TRUE)
)

nested <- dt_nest(dt, grp)
```

We could do this without the external lib like so:

```
nested <- dt[ , list(data = list(.SD)), keyby = grp]
```

And noticed the following error:

```
# works
nested$data
# gives (ugly) output, no error
nested[ , data[[1]], by = grp]
# now erros
nested$data
# [[1]]
# Error in `[.data.table`(x, i, , ) : 
#   Internal error: column type 'NULL' not supported by data.table subset. All known types are supported so please report as bug.
```

A bit hard to debug since most operations involving `nested` give an error, but there's this:

```
str(nested)
# Classes ‘data.table’ and 'data.frame':	3 obs. of  2 variables:
#  $ grp : int  1 2 3
#  $ data:List of 3
#   ..$ :Classes ‘data.table’ and 'data.frame':	33159 obs. of  2 variables:
#   .. ..$ : num  0.0534 -0.7286 0.823 -0.4527 -0.228 ...
#   .. ..$ : num  0.379 0.399 0.282 0.965 0.308 ...
#   .. ..- attr(*, ".internal.selfref")=<externalptr> 
#   .. ..- attr(*, ".data.table.locked")= logi TRUE
#   ..$ :Classes ‘data.table’ and 'data.frame':	33421 obs. of  2 variables:
#   .. ..$ : num  0.844 1.009 1.541 -0.174 0.353 ...
#   .. ..$ : num  0.716 0.798 0.83 0.306 0.279 ...
#   .. ..- attr(*, ".internal.selfref")=<externalptr> 
#   .. ..- attr(*, ".data.table.locked")= logi TRUE
#   ..$ :Classes ‘data.table’ and 'data.frame':	33420 obs. of  2 variables:
#   .. ..$ : num  1.061 -1.909 0.251 1.01 -1.053 ...
#   .. ..$ : num  0.366 0.55 0.62 0.936 0.219 ...
[...]

PR #4160, opened 2020-01-05
be sure to unlock objects returned during grouping

MichaelChirico:
Part of #4159 

I'm not sure this is the best way to do this, open to alternatives. But I did it recursively since we can really control how "deep down" any usages of `.SD` to be returned would be.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16726,6 +16726,10 @@ DT = data.table(
            s4class(x=2L, y="yes", z=1)))
 test(2130.03, print(DT), output=c("   x             y", "1: 1 <ex_class[3]>",     "2: 2 <ex_class[3]>"))
 
+# .SD from grouping should be unlocked, part of #4159
+x = data.table(a=1:3, b=4:6)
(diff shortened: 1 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -16726,6 +16726,10 @@ DT = data.table(
            s4class(x=2L, y="yes", z=1)))
 test(2130.03, print(DT), output=c("   x             y", "1: 1 <ex_class[3]>",     "2: 2 <ex_class[3]>"))
 
+# .SD from grouping should be unlocked, part of #4159
+x = data.table(a=1:3, b=4:6)
+test(2131, lapply(x[ , list(dt = list(.SD)), by = a]$dt, attr, '.data.table.locked'),
```
jangorecki: I think we handle that already in `split.data.table` so this PR could probably simplifies logic there https://github.com/Rdatatable/data.table/blob/d9b7ad5bd583662585032c87afb6ec6aaa5a2b21/R/data.table.R#L2310-L2314

MichaelChirico: I think the use case here is _slightly_ different in general -- I had in mind something like returning `list(list(list(list(.SD))))`. A bit contrived, but I think it's conceivable users can return `.SD` accidentally in less straightforward ways.

Issue #3291, opened 2019-01-16
1.12.0 compilation error on Red Hat Linux / Trisquel


iembry:
Hi, I tried unsuccessfully to install the v1.12.0 (13 Jan 2019) `data.table` package in Trisquel GNU/Linux system. I have the sessionInfo information at the bottom of this issue.

Please see the installation log below:

```
> install.packages("data.table")
Installing package into ‘/x86_64-pc-linux-gnu-library/library’
(as ‘lib’ is unspecified)
trying URL 'https://cloud.r-project.org/src/contrib/data.table_1.12.0.tar.gz'
Content type 'application/x-gzip' length 3311096 bytes (3.2 MB)
==================================================
downloaded 3.2 MB

* installing *source* package ‘data.table’ ...
** package ‘data.table’ successfully unpacked and MD5 sums checked
** libs
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c assign.c -o assign.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c between.c -o between.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c bmerge.c -o bmerge.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c chmatch.c -o chmatch.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c dogroups.c -o dogroups.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c fastmean.c -o fastmean.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c fcast.c -o fcast.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c fmelt.c -o fmelt.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c forder.c -o forder.o
ccache gcc -I"/usr/share/R/include" -DNDEBUG     -fopenmp -fpic  -std=gnu99 -c frank.c -o frank.o
[...]

PR #4162, opened 2020-01-07
print gcc version during install, towards #3291

jangorecki:
towards #3291

### DIFF
--- a/configure
+++ b/configure
@@ -77,5 +77,10 @@ else
   sed -e "s|@openmp_cflags@|\$(SHLIB_OPENMP_CFLAGS)|" src/Makevars.in > src/Makevars
 fi
 
+# compiler info to output #3291
+if [ "$CC"=~"gcc" ]; then
+  GCCV=`${CC} -dumpfullversion -dumpversion`
+  echo "$CC $GCCV"
+fi
 
 exit 0


configure
```diff
@@ -77,5 +77,10 @@ else
   sed -e "s|@openmp_cflags@|\$(SHLIB_OPENMP_CFLAGS)|" src/Makevars.in > src/Makevars
 fi
 
+# compiler info to output #3291
+if [ "$CC"=~"gcc" ]; then
+  GCCV=`${CC} -dumpfullversion -dumpversion`
+  echo "$CC $GCCV"
+fi
```
jangorecki, line 84: @MichaelChirico any idea if it works in macosx bash?
there is a report https://github.com/Rdatatable/data.table/issues/4410#issuecomment-627073357
which seems to not print gcc version (see in link). I double checked that `/usr/local/gfortran/bin/gcc` satisfies `"$CC"=~"gcc"` condition on my ubuntu.

MichaelChirico: What am I looking for? here is my output on OSX before the `clang ... -c -o` commands

```
R CMD INSTALL . --preclean
* installing to library ‘/Library/Frameworks/R.framework/Versions/4.0/Resources/library’
** using staged installation
zlib 1.2.11 is available ok
OpenMP supported
/usr/local/opt/llvm/bin/clang -fopenmp 4.2.1
** libs
```

I guess that second-last line is the `gcc` printout

jangorecki: we don't detect clang, user is using gcc but for some reason it is still not being printed to output

MichaelChirico: I also tried using `/usr/local/gfortran/bin/gcc` instead, here's my `~/.R/Makevars` after:

```
# The following is added per data.table/wiki/Installation
#LLVM_LOC = /usr/local/opt/llvm
GCC_LOC=/usr/local/gfortran
#CC=$(LLVM_LOC)/bin/clang -fopenmp
#CXX=$(LLVM_LOC)/bin/clang++ -fopenmp
CC=$(GCC_LOC)/bin/gcc -fopenmp
CXX=$(GCC_LOC)/bin/g++ -fopenmp
CFLAGS=-g -O3 -Wall -pedantic -std=gnu99 -mtune=native -pipe
CXXFLAGS=-g -O3 -Wall -pedantic -std=c++11 -mtune=native -pipe
#LDFLAGS=-L/usr/local/opt/gettext/lib -L$(LLVM_LOC)/lib -Wl,-rpath,$(LLVM_LOC)/lib
#CPPFLAGS=-I/usr/local/opt/gettext/include -I$(LLVM_LOC)/include
LDFLAGS=-L/usr/local/opt/gettext/lib -L$(GCC_LOC)/lib -Wl,-rpath,$(GCC_LOC)/lib
CPPFLAGS=-I/usr/local/opt/gettext/include -I$(GCC_LOC)/include
```

and then the same preamble output from `R CMD INSTALL . --preclean`:

```
* installing to library ‘/Library/Frameworks/R.framework/Versions/4.0/Resources/library’
** using staged installation
zlib 1.2.11 is available ok
OpenMP supported
/usr/local/gfortran/bin/gcc -fopenmp 6.3.0
** libs
```

jangorecki: OK, it is casued by early exit
```
if [ $msg -ne 0 ]; then
  ...
  exit 0
```

PR #4164, opened 2020-01-08
Final DATAPTR and USE_RINTERNALS removal

### LINKED ISSUE #3301: Remove #define USE_RINTERNALS

mattdowle:
Closes #3301

### DIFF
--- a/src/dogroups.c
+++ b/src/dogroups.c
@@ -405,21 +405,11 @@ SEXP growVector(SEXP x, const R_len_t newlen)
   PROTECT(newx = allocVector(TYPEOF(x), newlen));   // TO DO: R_realloc(?) here?
   if (newlen < len) len=newlen;   // i.e. shrink
   switch (TYPEOF(x)) {
-  case RAWSXP:
-    memcpy(RAW(newx), RAW(x), len*SIZEOF(x));
-    break;
-  case LGLSXP:
-    memcpy(LOGICAL(newx), LOGICAL(x), len*SIZEOF(x));
-    break;
-  case INTSXP:
-    memcpy(INTEGER(newx), INTEGER(x), len*SIZEOF(x));
-    break;
-  case REALSXP:
-    memcpy(REAL(newx), REAL(x), len*SIZEOF(x));
-    break;
-  case CPLXSXP:
-    memcpy(COMPLEX(newx), COMPLEX(x), len*SIZEOF(x));
-    break;
+  case RAWSXP:  memcpy(RAW(newx),     RAW(x),     len*SIZEOF(x)); break;
+  case LGLSXP:  memcpy(LOGICAL(newx), LOGICAL(x), len*SIZEOF(x)); break;
+  case INTSXP:  memcpy(INTEGER(newx), INTEGER(x), len*SIZEOF(x)); break;
+  case REALSXP: memcpy(REAL(newx),    REAL(x),    len*SIZEOF(x)); break;
+  case CPLXSXP: memcpy(COMPLEX(newx), COMPLEX(x), len*SIZEOF(x)); break;
   case STRSXP : {
     const SEXP *xd = SEXPPTR_RO(x);
     for (int i=0; i<len; ++i)
--- a/src/dogroups.c
+++ b/src/dogroups.c
@@ -113,76 +113,44 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
     INTEGER(N)[0] = istarts[i] == NA_INTEGER ? 0 : grpn;
     // .N is number of rows matched to ( 0 even when nomatch is NA)
     INTEGER(GRP)[0] = i+1;  // group counter exposed as .GRP
+    INTEGER(rownames)[1] = -grpn;  // the .set_row_names() of .SD. Not .N when nomatch=NA and this is a nomatch
+    for (int j=0; j<length(SDall); ++j) {
+      SETLENGTH(VECTOR_ELT(SDall,j), grpn);  // before copying data in otherwise assigning after the end could error R API checks
+      defineVar(nameSyms[j], VECTOR_ELT(SDall, j), env);
+      // Redo this defineVar for each group in case user's j assigned to the column names (env is static) (tests 387 and 388)
+      // nameSyms pre-stored to save repeated install() for efficiency, though.
+    }
+    for (int j=0; j<length(xSD); ++j) {
+      defineVar(xknameSyms[j], VECTOR_ELT(xSD, j), env);
+    }
 
     for (int j=0; j<length(iSD); ++j) {   // either this or the next for() will run, not both
-      size_t size = SIZEOF(VECTOR_ELT(iSD,j));
-      memcpy((char *)DATAPTR(VECTOR_ELT(iSD,j)),  // ok use of memcpy. Loop'd through columns not rows // TODO remove DATAPTR
-             (char *)DATAPTR(VECTOR_ELT(groups,INTEGER(jiscols)[j]-1))+i*size,
-             size);
+      memrecycle(VECTOR_ELT(iSD,j), R_NilValue, 0, 1, VECTOR_ELT(groups, INTEGER(jiscols)[j]-1), i, 1, j+1, "Internal error assigning to iSD");
+      // we're just use memrecycle here to assign a single value
     }
     // igrp determines the start of the current group in rows of dt (0 based).
     // if jiscols is not null, we have a by = .EACHI, so the start is exactly i.
     // Otherwise, igrp needs to be determined from starts, potentially taking care about the order if present.
     igrp = !isNull(jiscols) ? i : (length(grporder) ? INTEGER(grporder)[istarts[i]-1]-1 : istarts[i]-1);
     if (igrp>=0 && nrowgroups) for (int j=0; j<length(BY); ++j) {    // igrp can be -1 so 'if' is important, otherwise memcpy crash
-      size_t size = SIZEOF(VECTOR_ELT(BY,j));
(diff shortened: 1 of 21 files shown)


src/utils.c
```diff
@@ -355,3 +355,15 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
   UNPROTECT(1);
   return(ans);
 }
+
+//const void *DATAPTR_RO(SEXP x) {
```
jangorecki: good catch about backporting that...
```
/builds/jangorecki/data.table/bus/test-310-cran-lin/data.table.Rcheck/data.table/libs/datatable.so: undefined symbol: DATAPTR_RO
```

MichaelChirico: IIUC was added in roughly Sep '17 ~ 3.4

https://github.com/wch/r-source/commit/447cdb613a16c6419b6b9cc6bcfa9d9554f97db2

mattdowle:
The travis fail appears to be due to `stringi` updated on CRAN today.  Where does `stringi` come into the CI dependency chain?  It's not in the data.table repo afaics.

jangorecki:
It is required for knitr via stringr

MichaelChirico:
What's devtools used for? github install?

`remotes` used for the same and has lighter footprint

src/reorder.c
```diff
@@ -37,66 +37,61 @@ SEXP reorder(SEXP x, SEXP order)
...
+    // when we're sure we're passing in a strict permutation, and ii) this operation is fundamental to setkey and data.table.
   }
-  // Creorder is for internal use (so we should get the input right!), but the check above seems sensible. The for loop above should run
-  // in neglible time (sequential with prefetch). It will catch NAs anywhere but won't catch duplicates. But doing so would be going too
-  // far given this is for internal use only and we only use this function internally when we're sure it's a permutation.
 
-  char *TMP = malloc(nrow * maxSize);
-  // enough RAM for a copy of one column (of largest type). Writes into the [start,end] subset. Outside [start,end] is wasted in that rarer case
-  // to save a "-start" in the deep loop below in all cases.
+  char *TMP = malloc((end-start+1)*maxSize);
```
jangorecki: Any reason why we don't use R_alloc here?

mattdowle: Nope. Good spot.

src/utils.c
mattdowle: To double-check, I grabbed the released tar.gz for R 3.4.4 and 3.5.0.  DATAPTR_RO was new in 3.5.0.  It's not always clear from the R-devel commits what is 'backported' to the next minor release and what is intended for the next major release.  So although it was committed during the R 3.4 series,  it wasn't released until R 3.5.0.

mattdowle:
We can do review via follow up PRs.  Merging now to make progress, see if it passes stronger GLCI, and to push to anyone testing dev.

mattdowle:
Good. Passing GLCI (e.g. R 3.1.0).
I submitted to win-builder both R-devel and R-devel-ATC and both passed fully ok, too.
@jangorecki Since R 3.5.0 was a significant change point for data.table, how time consuming would it be to add two more tests to GLCI: R 3.4.4 and R 3.5.0.  To test just before and just after ALTREP & DATAPTR_RO. That way we test that we've got the version switch branches correct.

src/frank.c
```diff
@@ -81,42 +78,43 @@ SEXP frank(SEXP xorderArg, SEXP xstartArg, SEXP xlenArg, SEXP ties_method) {
...
   // else if (!strcmp(CHAR(STRING_ELT(ties_method, 0)), "runlength")) ties = RUNLENGTH;
   else error(_("Internal error: invalid ties.method for frankv(), should have been caught before. please report to data.table issue tracker")); // # nocov
-  n = length(xorderArg);
-  SEXP ans = (ties == MEAN) ? PROTECT(allocVector(REALSXP, n)) : PROTECT(allocVector(INTSXP, n));
-  int *ians = INTEGER(ans);
-  double *dans = REAL(ans);
-  if (n > 0) {
+  const int n = length(xorderArg);
+  SEXP ans = PROTECT(allocVector(ties==MEAN ? REALSXP : INTSXP, n));
+  int *ians=NULL;
```
MichaelChirico: why initialize to `NULL`, can't we just do `int *ians;` and `double *dans;`?

src/nafill.c
```diff
@@ -121,16 +121,26 @@ SEXP nafillR(SEXP obj, SEXP type, SEXP fill, SEXP nan_is_na_arg, SEXP inplace, S
...
+    }
   }
   if (!binplace) {
     ans = PROTECT(allocVector(VECSXP, nx)); protecti++;
     for (R_len_t i=0; i<nx; i++) {
       SET_VECTOR_ELT(ans, i, allocVector(TYPEOF(VECTOR_ELT(x, i)), inx[i]));
-      vans[i] = ((ans_t) { .dbl_v=REAL(VECTOR_ELT(ans, i)), .int_v=INTEGER(VECTOR_ELT(ans, i)), .int64_v=(int64_t *)REAL(VECTOR_ELT(ans, i)), .status=0, .message={"\0","\0","\0","\0"} });
+      const SEXP ansi = VECTOR_ELT(ans, i);
+      const void *p = isReal(ansi) ? (void *)REAL(ansi) : (void *)INTEGER(ansi);
+      vans[i] = ((ans_t) { .dbl_v=(double *)p, .int_v=(int *)p, .int64_v=(int64_t *)p, .status=0, .message={"\0","\0","\0","\0"} });
```
MichaelChirico: @jangorecki doesn't `""` null-terminate itself? Or are we double-null-terminating these?

MichaelChirico:
Gave a read and it looks good to me

src/frank.c
jangorecki: AFAIU to avoid compiler warnings about uninitialized variables

src/nafill.c
jangorecki: No idea

mattdowle: Looks like those are double-null terminated. I don't know why either.

src/nafill.c
```diff
@@ -121,16 +121,26 @@ SEXP nafillR(SEXP obj, SEXP type, SEXP fill, SEXP nan_is_na_arg, SEXP inplace, S
   SEXP ans = R_NilValue;
   ans_t *vans = (ans_t *)R_alloc(nx, sizeof(ans_t));
   for (R_len_t i=0; i<nx; i++) {
-    inx[i] = xlength(VECTOR_ELT(x, i));
-    dx[i] = REAL(VECTOR_ELT(x, i));
-    ix[i] = INTEGER(VECTOR_ELT(x, i));
-    i64x[i] = (int64_t *)REAL(VECTOR_ELT(x, i));
+    const SEXP xi = VECTOR_ELT(x, i);
+    inx[i] = xlength(xi);
+    // not sure why these pointers are being constructed like this; TODO: simplify structure
```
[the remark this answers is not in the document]
mattdowle: I'll try and come back and suggest something. There must be a simpler way.

src/dogroups.c
```diff
@@ -113,150 +114,74 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
-            // #3634 -- CPLXSXP columns have size 16
-            Rcomplex *td = COMPLEX(target);
-            const Rcomplex *sd = COMPLEX(source);
-            for (int k=0; k<grpn; ++k) {
-              rownum = iI[k]-1;
-              td[k] = sd[rownum];
-            }
-          }
+          // this is the main non-contiguous gather, and is parallel (within-column) for non-SEXP
+          subsetVectorRaw(VECTOR_ELT(SDall,j), VECTOR_ELT(dt,INTEGER(dtcols)[j]-1), I, /*anyNA=*/false);
```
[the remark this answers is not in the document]
jangorecki: seems that `_Pragma("...")` that we use inside `subsetVectorRaw` was giving an overhead also for 1 threaded computation, fix for that is in https://github.com/Rdatatable/data.table/pull/4558

PR #4188, opened 2020-01-20
New function psum/pprod

### LINKED ISSUE #3467: Add psum?
Could add to data.table and submit to R-core too at the same time. This way psum is available to data.table users back to R 3.1.0 and we can remove from data.table when R dependency reaches if and when it's added to R.

https://stackoverflow.com/questions/13123638/there-is-pmin-and-pmax-each-taking-na-rm-why-no-psum

Note that it hit resistance when I raised it on r-devel (link in the s.o. question).

2005m:
closes #3467. Towards #3189


src/fifelse.c
```diff
@@ -343,3 +343,161 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
   UNPROTECT(nprotect);
   return ans;
 }
+
+SEXP psumR(SEXP na, SEXP args) {
+	if (!IS_TRUE_OR_FALSE(na)) {
```
jangorecki: We use 2 whitespace, you used 8. It is so much horizontal scrolling now.

2005m: I will change all my c files. Thank you for letting me know.

2005m:
Please note that we can had lines for openmp. Happy to do it if you want. These functions are already fast from an absolute point of view and are close to `+` and `*` in timing (even sometimes faster which surprised me)

src/fifelse.c
```diff
@@ -429,28 +429,28 @@ SEXP psumR(SEXP na, SEXP args) {
...
   const int n=length(args);
   if (n <= 1) {
-    error("Please supply at least 2 arguments. %d argument supplied.", n);
+    error(_("Please supply at least 2 arguments. %d argument supplied."), n);
   }
   SEXPTYPE anstype = TYPEOF(SEXPPTR_RO(args)[0]);
   const int64_t len0 = xlength(SEXPPTR_RO(args)[0]);
   if (anstype != INTSXP && anstype != REALSXP) {
-    error("Argument 1 is of type %s. Only integer and double types are supported.", type2char(anstype));
+    error(_("Argument 1 is of type %s. Only integer and double types are supported."), type2char(anstype));
```
MichaelChirico: for translation, better to use %d here as well, then only one message to translate instead of 2

2005m: Sorry Michael, I did not understand what you mean.

src/fifelse.c
```diff
@@ -344,3 +344,169 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
...
   return ans;
 }
+
+SEXP psumR(SEXP na, SEXP args) {
+  if (!IS_TRUE_OR_FALSE(na)) {
+    error(_("Argument 'na.rm' must be TRUE or FALSE and length 1."));
+  }
+  const int n=length(args);
+  if (n <= 1) {
+    error(_("Please supply at least 2 arguments. (%d argument supplied)"), n);
```
MichaelChirico: why error, instead of just returning x itself?

src/fifelse.c
```diff
@@ -344,3 +344,169 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
...
+    if (type != INTSXP && type != REALSXP) {
+      error(_("Argument %d is of type %s. Only integer and double types are supported."),
+      i+1, type2char(type));
+    }
+    if (len1 != len0) {
+      error(_("Argument %d is of length %"PRId64" but argument 1 is of length %"PRId64". "
+      "If you wish to 'recycle' your argument, please use rep() to make this intent "
+      "clear to the readers of your code."), i+1, len1, len0);
+    }
+    if (type > anstype) {
```
MichaelChirico: I think we should use typeorder for this, the ordering of the SEXPTYPEs doesn't necessarily reflect the coercion order. I guess it's fine for just int/double but later we may add other types

src/fifelse.c
MichaelChirico: each _unique_ item wrapped in _() will generate a message to be translated

one with Argument 1, another with Argument %d -- twice as much work for our translation team

vs using Argument %d for both, and hard-coding 1 to fill that in the former case; looks same to user, very similar in code, and easier for translators

src/fifelse.c
2005m: I find it odd to provide 1 argument. A sum operation needs at least 2 arguments. Does it make sense to you to only provide 1 argument?

MichaelChirico: yes, it's odd, but edge case like this can come up in dynamic scripting, where e.g. user is running

```
DT[ , do.call(psum, .SD), .SDcols = is.numeric]
```

and `DT` has a variable number of numeric columns depending on some other conditions in the script.

This way, the user can use the same code to cover the normal case when there are several normal columns, as well as the case when there only ends up being one such column.

2005m: Ok will change that later. Tks

PR #4191, opened 2020-01-22
added test for fixed issue

### LINKED ISSUE #2491: Grouped mean of difftime fails due to Cfastmean
When trying to round an aggregated set of difftime values whilst creating a new data.table from an already existing one, unexpected warnings are thrown and the resulting values are NA. However, creating the data.table first with aggregation, _then_ rounding works correctly.

Related [stackoverflow question](https://stackoverflow.com/q/47270913/698590) includes the example below.

```
library(data.table)
dt<-data.table(
  Date1 = 
    sample(seq(as.Date('2017/10/01'), 
               as.Date('2017/10/31'), 
               by="days"), 24, replace = FALSE) +
    abs(rnorm(24))/10,
  Date2 = 
    sample(seq(as.Date('2017/10/01'), 
               as.Date('2017/10/31'), 
               by="days"), 24, replace = FALSE) +
    abs(rnorm(24))/10,
  Num1 =
    abs(rnorm(24))*10,
  Group = 
    rep(LETTERS[1:4],each=6)
)
dt[,DiffTime:=abs(difftime(Date1,Date2,units = 'days'))]

# Warnings/NA:
class(dt$DiffTime) # "difftime"
dt2<-dt[,.(AvgTime = round(mean(DiffTime),2)), by = .(Group)]

# Works when numeric/not difftime:
class(dt$Num1) # "numeric"
dt2<-dt[,.(AvgNum = round(mean(Num1),2)), by = .(Group)]

# Works, but takes an additional step:
dt2<-dt[,.(AvgTime = mean(DiffTime)), by = .(Group)]
dt2[,AvgTime := round(AvgTime,2)]
```

`#` `Output of sessionInfo()`
```
> sessionInfo()
R version 3.2.5 (2016-04-14)

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          

attached base packages:

other attached packages:
[1] data.table_1.10.4

[1] tools_3.2.5
```

`#` `Output of sessionInfo()`
```
> sessionInfo()
R version 3.4.2 (2017-09-28)
Running under: Windows Server 2008 R2 x64 (build 7601) Service Pack 1

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          

attached base packages:

other attached packages:
[1] data.table_1.10.4-2

[1] tools_3.4.2
```

MichaelChirico:
Closes #2491 

Thanks @Gaffi for pointing out it's fixed. Used `git bisect` to identify the issue was fixed by #3567 (corresponding issue #1876)


inst/tests/tests.Rraw
```diff
@@ -14959,12 +14959,24 @@ test(2041.1, DT[, median(date), by=g], data.table(g=c("a","b"), V1=as.Date(c("20
 test(2041.2, DT[, median(time), by=g], DT[c(2,5),.(g=g, V1=time)])
 
 # 'invalid trim argument' with optimization level 1; #1876
```
MichaelChirico: @mattdowle can you take a look here? I added `datatable.optimize=1` to fit the comment and the original issue, which say the fix is related to level-1 optimization, however I didn't see level 1 being applied for the test. And the output says `GForce is on` which doesn't sound right to me.

mattdowle: @MichaelChirico  Great investigation. I think probably that because optimization level 2 includes all the optimizations in level 1, that the test didn't reduce the optimization level from 2 to 1. Instead it checked via the `output=` that the optimization that had the issue is on at this point (i.e. level 1 or higher is on).

PR #4195, opened 2020-01-22
Fixed typeorder coercion rules for RAWSXP and LGLSXP to match base R

### LINKED ISSUE #4172: typeorder for LGLSXP and RAWSXP in src/init.c is in conflict with R's coercion rules
Copying from a discussion chain with @MichaelChirico in #4144 :

In src/init.c the type order is laid out as follows:

```
static void setSizes() {
  for (int i=0; i<100; ++i) { sizes[i]=0; typeorder[i]=0; }
  // only these types are currently allowed as column types :
  sizes[LGLSXP] =  sizeof(int);       typeorder[LGLSXP] =  0;
  sizes[RAWSXP] =  sizeof(Rbyte);     typeorder[RAWSXP] =  1;
  sizes[INTSXP] =  sizeof(int);       typeorder[INTSXP] =  2;   // integer and factor
  sizes[REALSXP] = sizeof(double);    typeorder[REALSXP] = 3;   // numeric and integer64
  sizes[CPLXSXP] = sizeof(Rcomplex);  typeorder[CPLXSXP] = 4;
  sizes[STRSXP] =  sizeof(SEXP *);    typeorder[STRSXP] =  5;
  sizes[VECSXP] =  sizeof(SEXP *);    typeorder[VECSXP] =  6;   // list column
  if (sizeof(char *)>8) error(_("Pointers are %d bytes, greater than 8. We have not tested on any architecture greater than 64bit yet."), sizeof(char *));
  // One place we need the largest sizeof is the working memory malloc in reorder.c
}
```

The order of LGLSXP and RAWSXP are in conflict with R's coercion rules:

```
c(as.raw(0), TRUE)
# [1] FALSE  TRUE
```

As you can see, the raw is coerced to logical, not the other way round as listed in src/init.c

I believe the coercion rules work this way because a logical vector may contain `NA`, whereas raw vectors cannot:

```
as.raw(NA)
# [1] 00
# Warning message:
# out-of-range values treated as 0 in coercion to raw
```

Changing the order of LGLSXP and RAWSXP in typeorder has minimal impact, breaking only the following tests: 2006.1, 2006.2, and 2129. 

These tests all related to the rbind/rbindlist functionality. Tests 2006.1 and 2006.2 are a simple fix: they simply expect the wrong output: when the typeorder is changed, `rbindlist(..., fill = TRUE)` with missing raw colums are now filled with `NA` in line with the functionality for all other column types, rather than `00` as is in the expected test output. Its not clear to me why 2129 fails, as the input data.tables do not contain raw columns, but the test is for rbind(), so this would take a little debugging.

sritchie73:
Closes #4172

Base R coercion rules mean that raw are coerced to logical, not the other way round:

```
c(as.raw(0), TRUE)
# [1] FALSE  TRUE
```

This PR fixes the typeorder in src/init.c to match the base R coercion rules. This impacts rbindlist, which will now coerce a raw column to a logical column when binding these two together.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16780,7 +16780,7 @@ DT1 = data.table(a=as.raw(1), b=as.raw(2), c=as.raw(3), d=as.raw(4), e=as.raw(5)
 DT2 = data.table(a=as.raw(1), b=TRUE, c=1L, d=1.5, e=complex(real=3, imaginary=1), f="a", g=list(3:5), h=expression(1+1))
 DT3 = setDT(lapply(names(DT1), function(j) c(DT1[[j]], DT2[[j]])))
 setnames(DT3, names(DT1))
-test(2133, rbind(DT1, DT2), DT3)
+test(2133, rbind(DT1, DT2), DT3, warning="Column 2 of item 1: 2 (type 'raw') at RHS position 1 taken as TRUE when assigning to type 'logical' (column 2 named 'b')") # warning expected due to loss of precision when coercing raw to logical
 
 
 
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16774,3 +16774,13 @@ rm(s1, s2, class2132)
 ########################
 #  Add new tests here  #
 ########################
+
+# Check rbindlist coercion rules for raw match base R (e.g. using c()) #4172
+DT1 = data.table(a=as.raw(1), b=as.raw(2), c=as.raw(3), d=as.raw(4), e=as.raw(5), f=as.raw(6), g=as.raw(7), h=as.raw(8))
+DT2 = data.table(a=as.raw(1), b=TRUE, c=1L, d=1.5, e=complex(real=3, imaginary=1), f="a", g=list(3:5), h=expression(1+1))
+DT3 = setDT(lapply(names(DT1), function(j) c(DT1[[j]], DT2[[j]])))
+setnames(DT3, names(DT1))
+test(2133, rbind(DT1, DT2), DT3)
+
+
+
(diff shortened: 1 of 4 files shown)


inst/tests/tests.Rraw
```diff
@@ -16774,3 +16774,13 @@ rm(s1, s2, class2132)
 ########################
 #  Add new tests here  #
 ########################
+
```
MichaelChirico: minor: i think new tests go before the `Add new tests here` bit

inst/tests/tests.Rraw
```diff
@@ -16770,7 +16770,13 @@ test(2132.2, fifelse(TRUE, 1, s2),       error = "S4 class objects (except nanot
 test(2132.3, fcase(TRUE, s1, FALSE, s2), error = "S4 class objects (except nanotime) are not supported. Please see https://github.com/Rdatatable/data.table/issues/4131.")
 rm(s1, s2, class2132)
 
+# Check rbindlist coercion rules for raw match base R (e.g. using c()) #4172
+DT1 = data.table(a=as.raw(1), b=as.raw(2), c=as.raw(3), d=as.raw(4), e=as.raw(5), f=as.raw(6), g=as.raw(7), h=as.raw(8))
+DT2 = data.table(a=as.raw(1), b=TRUE, c=1L, d=1.5, e=complex(real=3, imaginary=1), f="a", g=list(3:5), h=expression(1+1))
+DT3 = setDT(lapply(names(DT1), function(j) c(DT1[[j]], DT2[[j]])))
```
sritchie73: Noting merge conflict for this test with #4196.

To resolve, replace this line with:

```
DT3 = data.table(a=c(DT1$a, DT2$a), b=c(DT1$b, DT2$b), c=c(DT1$c, DT2$c),
                 d=c(DT1$d, DT2$d), e=c(DT1$e, DT2$e), f=c(DT1$f, DT2$f),
                 g=c(DT1$g, DT2$g), h=c(DT1$h, list(DT2$h)))
```

(i.e. because in this current PR coercion of any type with expression leads to an expression vector, while PR #4196 expects that column to be coerced to a list)

Issue #4040, opened 2019-11-12
data.table conversion turns the class of <expr> column into <list> column


JodyStats:
```r
library(data.table)
library(dplyr)
#> 
#> Attaching package: "dplyr"
#> The following objects are masked from "package:data.table":
#> 
#>     between, first, last
#> The following objects are masked from "package:stats":
#> 
#>     filter, lag
#> The following objects are masked from "package:base":
#> 
#>     intersect, setdiff, setequal, union
library(tibble)
library(drake)

plan <- drake::drake_plan(
  expr = dat[, col := val]
)
```
The `drake_plan` returns a tibble to be manipulated. The problem is that one of the variable being "expr" class and when I convert it to data.table the "expr" class turns into "list" class. I do not intend to manipulate this "expr" column and it"s required for downstream processes, its "expr" class must be unchanged after being converted into data.table from tibble. Is this a bug that can be fixed? Otherwise, what is the solution and best practice? Thank you in advance.

```r
# command column class of <expr>
plan
#> # A tibble: 1 x 2
#>   target command              
#>   <chr>  <expr>               
#> 1 expr   dat[, `:=`(col, val)]

# command column class of <expr>
plan %>% mutate(col = "val")
#> # A tibble: 1 x 3
#>   target command               col  
#>   <chr>  <expr>                <chr>
#> 1 expr   dat[, `:=`(col, val)] val

# command column class of <list>
plan <- data.table(plan)[, col := "val"]
plan
#>    target command    col
#>    <char>  <list> <char>
#> 1:   expr  <call>    val

# command column class of <list>
as_tibble(plan)
#> # A tibble: 1 x 3
#>   target command    col  
#>   <chr>  <list>     <chr>
#> 1 expr   <language> val
```

<sup>Created on 2019-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

TysonStanley:
This looks like expected behavior. @MichaelChirico just made a PR regarding `rbindlist()` being able to handle columns of type `<expr>` but it sounds like it isn't going to be something that changes in `data.table()`. 

However, with that in mind, consider the following information:

```r
typeof(plan$command)
#> [1] "list"
typeof(plan$command[[1]])
#> [1] "language"
typeof(data.table(plan)[, command])
#> [1] "list"
typeof(data.table(plan)[, command][[1]])
#> [1] "language"
```

[...]

PR #4196, opened 2020-01-24
All non-atomic types are now converted to lists in rbindlist

### LINKED ISSUE #542: [R-Forge #5260] basic operations like row-subsets on data.table could be faster
Submitted by: Arun ; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5260&group_id=240&atid=978)

One example is shown here:

http://stackoverflow.com/questions/21033196/how-to-access-data-table-rows-by-index

It should be possible to figure out that 'i' is the only argument inside and speed-up this operation?

In general, `[.data.table` needs quite a bit of tweaking/tidying up. This is a reminder to up the priority on this.

### LINKED ISSUE #546: [R-Forge #5252] rbindlist should support expression columns

sritchie73:
Closes #546 and fixes #3811

Example:

```
> A = data.table(c1 = 1, c2 = 'asd', c3 = expression(as.character(Sys.time())))
> B = data.table(c1 = 3, c2 = 'qwe', c3 = expression(as.character(Sys.time()+5)))
> rbind(A,B)
      c1     c2              c3
   <num> <char>          <list>
1:     1    asd <expression[1]>
2:     3    qwe <expression[1]>
> rbind(A,B)$c3
[[1]]
expression(as.character(Sys.time()))

[[2]]
expression(as.character(Sys.time() + 5))
```

A consequence of this PR is that where a data.table has > 1 row and an expression column, each expression is separated out into its own list element:

```
> A = data.table(c1 = 1, c2 = 'asd', c3 = expression(as.character(Sys.time())))
> B = data.table(c1 = 3:5, c2 = c('qwe', "foo", "bar"), 
+                c3 = expression(1+1, print("test"), as.character(Sys.time()+5)))
> rbind(A,B)
      c1     c2              c3
   <num> <char>          <list>
1:     1    asd <expression[1]>
2:     3    qwe <expression[1]>
3:     4    foo <expression[1]>
4:     5    bar <expression[1]>
> rbind(A,B)$c3
[[1]]
expression(as.character(Sys.time()))

[[2]]
expression(1 + 1)

[[3]]
expression(print("test"))

[[4]]
expression(as.character(Sys.time() + 5))
```

I don't know if this is desired behaviour, or whether we should default back to base R's coercion rules (i.e. returning an expression vector) - presumably if the user has constructed an expression vector column of length > 1 they would desire an expression vector back? E.g. #4040 is requesting expression columns to be preserved.

### DIFF
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -528,6 +528,6 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       }
     }
   }
-  UNPROTECT(nprotect);  // ans, coercedForFactor, thisCol, thisElement
+  UNPROTECT(nprotect);  // ans, coercedForFactor, thisCol
   return(ans);
 }
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -322,7 +322,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
 
     if (!foundName) { static char buff[12]; sprintf(buff,"V%d",j+1), SET_STRING_ELT(ansNames, idcol+j, mkChar(buff)); foundName=buff; }
     if (factor) maxType=INTSXP;  // if any items are factors then a factor is created (could be an option)
-    if (int64 && maxType!=REALSXP)
+    if (int64 && maxType!=REALSXP && maxType!=VECSXP)
       error(_("Internal error: column %d of result is determined to be integer64 but maxType=='%s' != REALSXP"), j+1, type2char(maxType)); // # nocov
     SEXP target;
     SET_VECTOR_ELT(ans, idcol+j, target=allocVector(maxType, nrow));  // does not initialize logical & numerics, but does initialize character and list
@@ -517,6 +517,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
           if (TYPEOF(target)==VECSXP) {
             thisCol = PROTECT(coerceAsList(thisCol, thisnrow)); nprotect++;
           }
+
           // else coerces if needed within memrecycle; with a no-alloc direct coerce from 1.12.4 (PR #3909)
           const char *ret = memrecycle(target, R_NilValue, ansloc, thisnrow, thisCol, 0, -1, idcol+j+1, foundName);
           if (ret) warning(_("Column %d of item %d: %s"), w+1, i+1, ret);
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -519,7 +519,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
             SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
             for(int r=0; r<length(thisCol); ++r) {
               SEXP thisElement = VECTOR_ELT(thisCol, r);
-              if (TYPEOF(thisCol) == EXPRSXP) thisElement = coerceVector(thisElement, EXPRSXP); // otherwise LANGSXP
+              if (TYPEOF(thisCol) == EXPRSXP) thisElement = PROTECT(coerceVector(thisElement, EXPRSXP)); nprotect++; // otherwise LANGSXP
               SET_VECTOR_ELT(thisColList, r, thisElement);
             }
             thisCol = thisColList;
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -552,6 +552,6 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       }
     }
   }
-  UNPROTECT(nprotect);  // ans, coercedForFactor, thisCol
+  UNPROTECT(nprotect);  // ans, coercedForFactor, thisCol, thisElement
   return(ans);
 }
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -309,7 +309,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       }
       if (firsti==-1) { firsti=i; firstw=w; firstCol=thisCol; }
       else {
-        if (!factor && !int64) {
+        if (!factor && !int64 && maxType != VECSXP) {
           if (!R_compute_identical(PROTECT(getAttrib(thisCol, R_ClassSymbol)),
                                    PROTECT(getAttrib(firstCol, R_ClassSymbol)),
                                    0)) {
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -514,8 +514,9 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
+          int cnprotect = 0;
           if (maxType==VECSXP) {
-            thisCol = PROTECT(coerceAsList(thisCol, thisnrow)); nprotect++;
+            thisCol = coerceAsList(thisCol, thisnrow, &cnprotect); // cnprotect incremented by 0 or 1
           }
 
           // else coerces if needed within memrecycle; with a no-alloc direct coerce from 1.12.4 (PR #3909)
@@ -523,11 +524,12 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
           if (ret) warning(_("Column %d of item %d: %s"), w+1, i+1, ret);
           // e.g. when precision is lost like assigning 3.4 to integer64; test 2007.2
           // TODO: but maxType should handle that and this should never warn
+          UNPROTECT(cnprotect); // pop coerced 'thisCol' from PROTECT stack - not need now copied into 'target'.
         }
         ansloc += thisnrow;
       }
     }
   }
-  UNPROTECT(nprotect);  // ans, coercedForFactor, thisCol
+  UNPROTECT(nprotect);  // ans, coercedForFactor
   return(ans);
 }
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -514,18 +514,33 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
-          if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)>TYPEOF(target)) {
-            // Exotic non-atomic types need each element to be wrapped in a list, e.g. expression vectors #546
+          if ((TYPEOF(target)==VECSXP) && (isVectorAtomic(thisCol) || TYPEOF(thisCol)==LISTSXP)) {
+            // do an as.list() on the atomic column; #3528
+            // pairlists (LISTSXP) can also be coerced to lists using coerceVector
+            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); nprotect++;
+          } else if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)==EXPRSXP) {
+            // For EXPRSXP each element to be wrapped in a list, e.g. expression vectors #546
             SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
             for(int r=0; r<length(thisCol); ++r) {
               SEXP thisElement = VECTOR_ELT(thisCol, r);
-              if (TYPEOF(thisCol) == EXPRSXP) thisElement = PROTECT(coerceVector(thisElement, EXPRSXP)); nprotect++; // otherwise LANGSXP
+              thisElement = PROTECT(coerceVector(thisElement, EXPRSXP)); nprotect++; // otherwise LANGSXP
               SET_VECTOR_ELT(thisColList, r, thisElement);
             }
             thisCol = thisColList;
-          } else if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)<TYPEOF(target)) {
-            // do an as.list() on the atomic column; #3528
-            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); nprotect++;
+          } else if ((TYPEOF(target)==VECSXP) && !isVector(thisCol) && TYPEOF(thisCol)!=TYPEOF(target)) { 
+            // Anything not a vector we can assign directly through SET_VECTOR_ELT
+            // Although tecnically there should only be one list element for any type met here,
+            // the length of the type may be > 1, in which case the other columns in data.table
+            // will have been recycled. We therefore in turn have to recycle the list elements
+            // to match the number of rows.
+            SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
+            for(int r=0; r<length(thisCol); ++r) {
+              SET_VECTOR_ELT(thisColList, r, thisCol);
+            }
+            thisCol = thisColList;
+          } else if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)!=TYPEOF(target)) {
+            // should be unreachable
+            error("Internal error: rbindlist cannot handle type %s\n", type2char(TYPEOF(thisCol))); // # nocov
           }
           // else coerces if needed within memrecycle; with a no-alloc direct coerce from 1.12.4 (PR #3909)
           const char *ret = memrecycle(target, R_NilValue, ansloc, thisnrow, thisCol, 0, -1, idcol+j+1, foundName);
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -514,33 +514,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
-          if ((TYPEOF(target)==VECSXP) && (isVectorAtomic(thisCol) || TYPEOF(thisCol)==LISTSXP)) {
-            // do an as.list() on the atomic column; #3528
-            // pairlists (LISTSXP) can also be coerced to lists using coerceVector
-            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); nprotect++;
-          } else if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)==EXPRSXP) {
-            // For EXPRSXP each element to be wrapped in a list, e.g. expression vectors #546
-            SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
-            for(int r=0; r<length(thisCol); ++r) {
-              SEXP thisElement = VECTOR_ELT(thisCol, r);
-              thisElement = PROTECT(coerceVector(thisElement, EXPRSXP)); nprotect++; // otherwise LANGSXP
-              SET_VECTOR_ELT(thisColList, r, thisElement);
-            }
-            thisCol = thisColList;
-          } else if ((TYPEOF(target)==VECSXP) && !isVector(thisCol) && TYPEOF(thisCol)!=TYPEOF(target)) { 
-            // Anything not a vector we can assign directly through SET_VECTOR_ELT
-            // Although tecnically there should only be one list element for any type met here,
-            // the length of the type may be > 1, in which case the other columns in data.table
-            // will have been recycled. We therefore in turn have to recycle the list elements
-            // to match the number of rows.
-            SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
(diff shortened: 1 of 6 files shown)


src/rbindlist.c
```diff
@@ -514,7 +514,16 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
-          if ((TYPEOF(target)==VECSXP || TYPEOF(target)==EXPRSXP) && TYPEOF(thisCol)!=TYPEOF(target)) {
+          if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)>TYPEOF(target)) {
+            // Exotic non-atomic types need each element to be wrapped in a list, e.g. expression vectors #546
+            SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
+            for(int r=0; r<length(thisCol); ++r) {
+              SEXP thisElement = VECTOR_ELT(thisCol, r);
+              if (TYPEOF(thisCol) == EXPRSXP) thisElement = coerceVector(thisElement, EXPRSXP); // otherwise LANGSXP
```
sritchie73: Are there other more exotic types that people use in columns? It would be good to test whether we need to always do this additional coercion step, or whether this problem is unique to EXPRSXP vectors.

sritchie73: Do we need another PROTECT call here or on the line before?

jangorecki: We need. Coerced object is new memory location not referenced anywhere this can be gc.

jangorecki: We need to protect here as well. Test for both language object created with `quote` and expressions would be nice. Regarding consistency to base R, it is best to evaluate that taking use cases in account. Language objects IMO has quite the same use case as expressions, not sure what base R behaviour on expressions is meant to archive.

sritchie73: So a PROTECT for both the `coerceVector` and the `VECTOR_ELT`? Or just the `coerceVector` (this line is the same as one referred to in my review comment below).

jangorecki: VECTOR_ELT is just a pointer to an object already referenced, thus protected. Just result of coerceVector.

sritchie73: I'm not sure we need to handle `quote`. Constructing a quote column on its own does strange things because its length is not necessarily one, it depends on the parse tree I think. E.g. 

```
> length(quote(as.character(system.time())))
[1] 2
> length(quote(as.character(system.time()+5)))
[1] 2
> length(quote(c(1, 2, 3)))
[1] 4
> length(quote(1+1))
[1] 3
```

If you try to create a vector of `quote` it turns into a list already (conveniently the desired behaviour in data.table)

```
> c(quote(1+1), quote(as.character(system.time())))
[[1]]
1 + 1

[[2]]
as.character(system.time())
```

In the event of a nrow 1 data.table, all your other columns get recycled to the length of the quote object, which I think is probably not desirable behaviour:

```
> data.table(A=0, B=quote(1+1))
       A      B
   <num> <call>
1:     0  1 + 1
2:     0  1 + 1
3:     0  1 + 1
```

sritchie73: RE: base R. I'm referring to #4040 where the person has asked for data.table NOT to turn each element into a list.

jangorecki: IMO expressions should be wrapped into list as they are not atomic, as as language objects. Quote is length 1 only for symbols, otherwise it is a call, still language object, which length depends on the number of arguments.

sritchie73: I believe the latest commit solves this, but will need to add tests

src/rbindlist.c
```diff
@@ -514,9 +514,33 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
-          if ((TYPEOF(target)==VECSXP || TYPEOF(target)==EXPRSXP) && TYPEOF(thisCol)!=TYPEOF(target)) {
+          if ((TYPEOF(target)==VECSXP) && (isVectorAtomic(thisCol) || TYPEOF(thisCol)==LISTSXP)) {
```
sritchie73: Assuming the desired behaviour here is for a `pairlist` to be converted to a `list`, rather than being converted to a list where each element is a 1-element pairlist.

src/rbindlist.c
```diff
@@ -514,9 +514,33 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+          } else if ((TYPEOF(target)==VECSXP) && !isVector(thisCol) && TYPEOF(thisCol)!=TYPEOF(target)) { 
+            // Anything not a vector we can assign directly through SET_VECTOR_ELT
+            // Although tecnically there should only be one list element for any type met here,
+            // the length of the type may be > 1, in which case the other columns in data.table
+            // will have been recycled. We therefore in turn have to recycle the list elements
+            // to match the number of rows.
+            SEXP thisColList = PROTECT(allocVector(VECSXP, length(thisCol))); nprotect++;
+            for(int r=0; r<length(thisCol); ++r) {
+              SET_VECTOR_ELT(thisColList, r, thisCol);
+            }
```
sritchie73: Demonstration of the recycling issue and why we need this for loop:

```
> A = data.table(a=0L, B=quote(1+1))
> A
       a      B
   <int> <call>
1:     0  1 + 1
2:     0  1 + 1
3:     0  1 + 1
```

The for loop means we get the right result with rbind:

```
> rbind(A)
       a         B
   <int>    <list>
1:     0 <call[3]>
2:     0 <call[3]>
3:     0 <call[3]>
```

Otherwise we would have

```
> rbind(A)
       a         B
   <int>    <list>
1:     0 <call[3]>
2:     0 
3:     0
```

It would be good to fix this in data.table()/as.data.table() so that these non-vector types are wrapped in list at the time of data.table construction

sritchie73:
I've also come across this unexpected behaviour:

```
> A = data.table(c1=0, c2=list(1:3))
> B = data.table(c1=1)
> rbind(A,B,fill=TRUE)
      c1     c2
   <num> <list>
1:     0  1,2,3
2:     1       
```

When `fill=TRUE` and the missing column is `list` type, then the corresponding entries are filled with `NULL` elements. Should these be filled with `NA_logical_` instead?.

src/rbindlist.c
jangorecki: example of both would help

src/rbindlist.c
```diff
@@ -514,18 +514,33 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
-          if ((TYPEOF(target)==VECSXP) && TYPEOF(thisCol)>TYPEOF(target)) {
-            // Exotic non-atomic types need each element to be wrapped in a list, e.g. expression vectors #546
+          if ((TYPEOF(target)==VECSXP) && (isVectorAtomic(thisCol) || TYPEOF(thisCol)==LISTSXP)) {
+            // do an as.list() on the atomic column; #3528
+            // pairlists (LISTSXP) can also be coerced to lists using coerceVector
```
jangorecki: does it coerce recursively unwrapping nested parilists into single list of many elements?

src/rbindlist.c
jangorecki: agree, very nice demo

jangorecki:
Filling with NULL in such a case looks seems fine for me.

src/rbindlist.c
sritchie73: Only the top level pairlist is coerced to a list:

```
> A = data.table(c1 = 1:3, c2 = pairlist(A=1, B=list(1, 2:3), C=pairlist(C1=1, C2=3:4)))
> B = data.table(c1 = 2, c2 = 'asd')
> rbind(A,B)
      c1            c2
   <num>        <list>
1:     1             1
2:     2     <list[2]>
3:     3 <pairlist[2]>
4:     2           asd
```

src/rbindlist.c
sritchie73: I think this makes sense - my understanding is a pairlist is a linked-list, while a regular list is essentially a vector of pointers to the list elements. If we treat a LISTSXP the same way as an EXPRSXP, then you end up with a list, where each element is a pointer to each individual node in the linked list. You end up lose any benefits of having the linked list, because each list element is just a pointer to a 1-node linked list. 

E.g the result is something like:

```
# Linked list, A -> B -> C
ll = pairlist(A=1, B=pairlist(1, 2:3), C='foo')

# Current rbindlist behaviour turns top level pairlist into list (losing the names)
cb = list(1, pairlist(1, 2:3), 'foo')

# Alternate rbindlist behaviour if we treat a pairlist in the same way as expression:
ab = list(pairlist(A=1), pairlist(B=pairlist(1, 2:3)), pairlist(C='foo'))
```

sritchie73:
It departs from the expected behaviour of fill=TRUE for all other column types though.

sritchie73:
TODO: list wrapper of atomic vectors is not class aware.

sritchie73:
Wrapping atomic vectors in list now preserves class. The integer64 case errors because memrecycle copies the integer64 class to the list 

TODO: fix or handle

src/rbindlist.c
```diff
@@ -514,20 +514,24 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+
           // else coerces if needed within memrecycle; with a no-alloc direct coerce from 1.12.4 (PR #3909)
           const char *ret = memrecycle(target, R_NilValue, ansloc, thisnrow, thisCol, 0, -1, idcol+j+1, foundName);
           if (ret) warning(_("Column %d of item %d: %s"), w+1, i+1, ret);
           // e.g. when precision is lost like assigning 3.4 to integer64; test 2007.2
           // TODO: but maxType should handle that and this should never warn
+       
+          // memrecycle incorrectly adds integer64 class to whole list if any element has class integer64
+          if (maxType == VECSXP && INHERITS(target, char_integer64)) 
+            setAttrib(target, R_ClassSymbol, R_NilValue);
```
sritchie73: Ideally would patch in memrecycle

sritchie73: I have so far not been able to track down where or why in memrecycle this happens (both target and thisCol do not inherit char_integer64), and this class assignment to the full list does not seem to happen if I add use other classes (e.g. assigning class "foo" to a character vector column, then rbinding to a list column).

src/rbindlist.c
```diff
@@ -514,20 +514,24 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
         if (w==-1 || !length(thisCol=VECTOR_ELT(li, w))) {  // !length for zeroCol warning above; #1871
           writeNA(target, ansloc, thisnrow);  // writeNA is integer64 aware and writes INT64_MIN
         } else {
-          if ((TYPEOF(target)==VECSXP || TYPEOF(target)==EXPRSXP) && TYPEOF(thisCol)!=TYPEOF(target)) {
-            // do an as.list() on the atomic column; #3528
-            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); nprotect++;
+          if (maxType==VECSXP) {
+            thisCol = PROTECT(coerceAsList(thisCol, thisnrow)); nprotect++;
           }
```
sritchie73: Logic abstracted into a separate function in utils.c because it is otherwise duplicated in asmatrix in #4144. I also envision making an R exposable wrapper so that non-atomic vector columns may be caught and coerced to list in `as.data.table` and `setDT` (future PR once this is merged)

src/rbindlist.c
sritchie73: early copyMostAttrib statement was the culprit. now skipped when maxType=VECSXP

sritchie73:
Wrapping atomic vectors in list now correctly preserves class.

src/rbindlist.c
sritchie73: Not sure if a single PROTECT is sufficient here - the PROTECT stack in coerceAsList may be `thisnrow + 1` if `TYPEOF(thisCol)==EXPRSXP` because each element in the vector has to be coerced to an EXPRSXP and passed to SET_VECTOR_ELT in the list wrapper, otherwise the list elements are `LANGSXP` if you just do `coerceVector(thisCol, VECSXP)`. There is a corresponding UNPROTECT call in `coerceAsList` that handles this, but I don't know if that means wrapping the result here in a single PROTECT risks this memory being freed before rbindlist returns.

src/rbindlist.c
```diff
@@ -292,8 +292,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       }
       SEXP thisCol = VECTOR_ELT(li, w);
       int thisType = TYPEOF(thisCol);
-      // Use >= for #546 -- TYPEORDER=0 for both LGLSXP and EXPRSXP (but also NULL)
-      if (TYPEORDER(thisType)>=TYPEORDER(maxType) && !isNull(thisCol)) maxType=thisType;
+      if (TYPEORDER(thisType)>TYPEORDER(VECSXP) && !isNull(thisCol)) maxType=VECSXP; 
+      else if (TYPEORDER(thisType)>TYPEORDER(maxType) && !isNull(thisCol)) maxType=thisType;
```
sritchie73: Noting here that this line has a merge conflict with #4195.

To resolve this conflict, replace this line with:

```
else if (!isNull(thisCol) && (maxType == -1 || TYPEORDER(thisType)>TYPEORDER(maxType))) maxType=thisType; 
```

src/rbindlist.c
sritchie73: The method i've settled on after hitting the protect stack limit in testing #4144 is to have the C function called from R manage the PROTECT stack, and for functions it calls to simply increment a protect counter. 

I've modified `coerceAsList` to this effect in the latest commit. Inside `coerceAsList`, each new allocated list element PROTECT is popped from the stack once it's been assigned into the new list using SET_VECTOR_ELT. In the coercion loop in rbindlist.c I make a separate nprotect counter for `coerceAsList`, as it may either increment the PROTECT stack by one, or not at all (e.g. when simply returning the input list). This allows us to pop the coerced list from the PROTECT stack after its contents have been copied into the target vector. This means these changes will no longer hit the PROTECT stack limit when handling >50,000 columns that need to be converted to lists.

jangorecki:
does this PR needs update after https://github.com/Rdatatable/data.table/issues/546#issuecomment-604926109 ?

sritchie73:
@jangorecki I'm not sure: I think the issue would be that if you do not wrap each expression in a list, then the length of the expression column will not match the length of the other non-expression columns.

jangorecki:
unless length match, that would happen when each row is meant to store a language object, rather than expression.

jangorecki:
I was going to provide simple example but I noticed that expression looks to be recycled, making the results not what I really wanted
```r
l=list(x=1:2, y=2:3) 
l$expr = expression(1+2,2+3)
sapply(l, length)
#   x    y expr 
#   2    2    2 
as.data.table(l)
#       x     y                     expr
#   <int> <int>                   <expr>
#1:     1     2 expression(1 + 2, 2 + 3)
#2:     2     3 expression(1 + 2, 2 + 3)
```
but after closer look, it seems that the issue is in print data.table, because `expr` column not recycled
```r
as.data.table(l)$expr
expression(1 + 2, 2 + 3)
```
@MichaelChirico any idea if there is an open issue for printing expression column behavior?

jangorecki:
If `expression` would always store the language objects then using it a column would probably make more sense.
But the following behavior makes me think it is just easier to convert that to list
```r
lapply(expression(1+2,3), class)
#[[1]]
#[1] "call"
#
#[[2]]
#[1] "numeric"
```

sritchie73:
That's essentially the conclusion I came to for any object that's not considered internally to be a vector by R (i.e. !isVector() == true in the R's C interface) because they have a length property that does not correspond to the number of elements. This causes problems when trying to print, and also errors when these types are in the first column of the data.table (because the first column is used to determine .N)

jangorecki:
Agree, could you please ensure PR is ready to review? otherwise better to mark it as WIP.

sritchie73:
I've just reviewed the unit tests and run them; the current behaviour of this PR is to wrap each expression element in a list, even if the input expression column is a vector of length > 1:

```
A = data.table(c1 = 1:2, c2 = c('asd', 'xzy'), c3 = expression(as.character(Sys.time()), 1+2))
B = data.table(c1 = 3:5, c2 = c('qwe', "foo", "bar"), 
               c3 = expression(1+1, print("test"), as.character(Sys.time()+5)))
dt = rbind(A,B)
dt[, c3]
[[1]]
expression(as.character(Sys.time()))

[[2]]
expression(1 + 2)

[[3]]
expression(1 + 1)

[[4]]
expression(print("test"))

[[5]]
expression(as.character(Sys.time() + 5))
```

I've added a unit test to cover this case.

MichaelChirico:
> any idea if there is an open issue for printing expression column behavior?

Sounds somewhat familiar but I didn't see anything in `debugonce`.

What's happening is `format(l$expr)` returns length-one, then `cbind` recycles this.

This will be better addressed by #3414 -- in fact I already see a `format_col.expression` method there.

jangorecki:
@sritchie73 one line seems to be not covered yet: https://codecov.io/gh/Rdatatable/data.table/compare/c0052964694a4c618ab182aa474f924d40576d94...00c6d859b158737f130d3f5a1da416693358e14d/diff

sritchie73:
I'll need to do some digging later. I don't see how that line isn't covered given all the other lines in that for loop are covered...

sritchie73:
I injected a few Rprintf statements and can confirm that the tests are running code within that loop before and after the line that's supposedly not covered, so I'm not sure what to make of this:

Running this test:

```
# Test non-atomic columns in rbindlist, e.g. expressions, #546
A = data.table(c1 = 1, c2 = 'asd', c3 = expression(as.character(Sys.time())))
B = data.table(c1 = 3, c2 = 'qwe', c3 = expression(as.character(Sys.time()+5)))
DT = data.table(c1 = c(1, 3), c2 = c("asd", "qwe"), 
                c3 = list(expression(as.character(Sys.time())), expression(as.character(Sys.time() + 5))))
test(2129.1, rbind(A,B), DT)
```

I get both messages below output to the R console:

```
      Rprintf("Inside loop\n");
      // Allocate length-1 vector of appropriate type and set it in the right place in the VECSXP container
      SEXP thisElement = PROTECT(allocVector(TYPEOF(x), 1)); (*nprotect)++;
      SET_VECTOR_ELT(coerced, i, thisElement); UNPROTECT(1); (*nprotect)--; // line supposedly not covered 
      Rprintf("After SET_VECTOR_ELT\n");
```

MichaelChirico:
@sritchie73 just checking if this is still needed? If so we can merge to `master` and resume the review process (nearly 4 years later 😅).

Also if you'd like someone else to take over the PR, do feel free.

cc @ben-schwen since I know you've been working a lot on `rbindlist` lately.

jangorecki:
It may be better to merge other rbindlist PR(s) (Ben had something recently from where we took only regression fix), and once the other one is merged master, then merge master to this one.

MichaelChirico:
#5453 is the other pending `rbindlist()` PR. Let's revisit this after that's addressed.

MichaelChirico:
OK, #5453 is resolved. Both of the issues linked in this PR are already closed/merged.

I don't really see the issue TBH -- here's the current behavior of the examples from OP:

```r
x = rbind(
  data.table(c1 = 1, c2 = 'asd', c3 = expression(as.character(Sys.time()))),
  data.table(c1 = 3, c2 = 'qwe', c3 = expression(as.character(Sys.time()+5))))
x
#       c1     c2                           c3
#    <num> <char>                       <expr>
# 1:     1    asd     as.character(Sys.time())
# 2:     3    qwe as.character(Sys.time() + 5)
x$c3
# expression(as.character(Sys.time()), as.character(Sys.time() + 
#     5))

x = rbind(
  data.table(c1 = 1:2, c2 = c('asd', 'xzy'), c3 = expression(as.character(Sys.time()), 1+2)),
  data.table(c1 = 3:5, c2 = c('qwe', "foo", "bar"),  c3 = expression(1+1, print("test"), as.character(Sys.time()+5))))
x
#       c1     c2                           c3
#    <int> <char>                       <expr>
# 1:     1    asd     as.character(Sys.time())
# 2:     2    xzy                        1 + 2
# 3:     3    qwe                        1 + 1
# 4:     4    foo                print("test")
# 5:     5    bar as.character(Sys.time() + 5)
x$c3
# expression(as.character(Sys.time()), 1 + 2, 1 + 1, print("test"), 
#     as.character(Sys.time() + 5))
```

So I'm closing this given the drift from current `master` & lack of apparent use case (maybe it's just been resolved elsewise in the meantime?).

Please open a new issue if there is still something to be done.

PR #4207, opened 2020-01-29
Added round.ITime and trunc.ITime

JenspederM:
Wanted to add convenience funtions, round and trunc to ITime


JenspederM:
Wish to add round and trunc methods to ITime

R/IDateTime.R
```diff
@@ -240,6 +240,22 @@ rep.ITime = function (x, ...)
   class(y) = "ITime"   # unlass and rep could feasibly not copy, hence use class<- not setattr()
   y
 }
+                           
+round.ITime <- function(x, digits = c("hours", "minutes")) 
+{
+  (setattr(switch(match.arg(digits),
```
MichaelChirico: needs to be added in `NAMESPACE`, needs tests, and please add something to NEWS; please see:

https://github.com/Rdatatable/data.table/wiki/Contributing

JenspederM:
Added 'L' and ensured that vectors are of class "integer"

R/IDateTime.R
```diff
@@ -240,6 +240,30 @@ rep.ITime = function (x, ...)
   class(y) = "ITime"   # unlass and rep could feasibly not copy, hence use class<- not setattr()
   y
 }
+                           
+round.ITime <- function(x, digits = c("hours", "minutes"))
```
MichaelChirico: `round.POSIXt` is using `units` argument, which is probably more intuitive than `digits`

R/IDateTime.R
```diff
@@ -240,6 +240,30 @@ rep.ITime = function (x, ...)
   class(y) = "ITime"   # unlass and rep could feasibly not copy, hence use class<- not setattr()
   y
 }
+                           
+round.ITime <- function(x, digits = c("hours", "minutes")) 
+{
+  (setattr(switch(match.arg(digits),
+                  hours = as.integer(round(unclass(x)/3600L)*3600L),
```
MichaelChirico: `3600` here is fine; `round` output will be numeric already, I imagine it will be slightly faster to multiply `double*double` than `double*(double)int` (i.e. casting `int`->`double` at C level)

JenspederM:
removed 'L' from functions to avoid integer to numeric conversion
aligned 'round()' with 'round.POSIXct' by exchaging formal 'digits' with 'units'

R/IDateTime.R
```diff
@@ -240,6 +240,30 @@ rep.ITime = function (x, ...)
...
+  (setattr(switch(match.arg(digits),
+                  hours = as.integer(round(unclass(x)/3600)*3600),
+                  minutes = as.integer(round(unclass(x)/60)*60)), 
+           "class", "ITime"))
+} 
+
+trunc.ITime <- function(x, units = c("hours", "minutes")) 
+{
+  (setattr(switch(match.arg(units),
+                  hours = as.integer(trunc(unclass(x)/3600)*3600),
```
MichaelChirico: my guess is `x %/% 3600L` will be faster than `trunc(unclass(x)/3600)`. `ITime` value is restricted to `0:86399` and the two are the same in that range.

R/IDateTime.R
```diff
@@ -240,6 +240,30 @@ rep.ITime = function (x, ...)
...
+                  minutes = as.integer(trunc(unclass(x)/60)*60)), 
+           "class", "ITime"))
+}
+
+floor.ITime <- function(x) {
```
MichaelChirico: why should we expect `floor(ITime)` to round to the nearest hour? do you have a use case for this function?

MichaelChirico: At least, why shouldn't the signature here have `units` as well?

JenspederM: `floor` does not allow more than one argument. I considered making two functions: `floor_hour` and `floor_minute`, but decided against it.

MichaelChirico:
Thanks for the PR.

New S3 methods require registration in `NAMESPACE`, see that file for examples.

Your PR also requires tests -- see the section here: https://github.com/Rdatatable/data.table/wiki/Contributing

You can see which lines are untested by checking the codecov bot's comments & looking at the commit diff, e.g. here:

https://codecov.io/gh/Rdatatable/data.table/pull/4207/diff?src=pr&el=tree#diff-Ui9JRGF0ZVRpbWUuUg==

R/IDateTime.R
JenspederM: I just did a benchmark, and trunc(unclass(x)/3600) is slightly faster.

image

MichaelChirico: I think it depends on your input 🙃 

```
> x = structure(sample(86399L, 1e7, TRUE), class = 'ITime')
> microbenchmark::microbenchmark(
+   times = 1000L,
+   trunc(unclass(x)/3600),
+   x %/% 3600L
+ )
Unit: milliseconds
                   expr      min       lq     mean   median       uq       max neval cld
 trunc(unclass(x)/3600) 37.54874 39.88270 48.46855 50.45054 53.37909 246.11452  1000   b
              x%/%3600L 40.50842 42.40509 47.17708 43.23612 54.86358  71.76187  1000  a 
```

R/IDateTime.R
MichaelChirico: OK, understood. Is there any reason to believe `hour` is the more common use case?

JenspederM:
> Thanks for the PR.
> 
> New S3 methods require registration in `NAMESPACE`, see that file for examples.
> 
> Your PR also requires tests -- see the section here: https://github.com/Rdatatable/data.table/wiki/Contributing
> 
> You can see which lines are untested by checking the codecov bot's comments & looking at the commit diff, e.g. here:
> 
> https://codecov.io/gh/Rdatatable/data.table/pull/4207/diff?src=pr&el=tree#diff-Ui9JRGF0ZVRpbWUuUg==

I have added the S3methods to NAMESPACE, but they have yet to be accepted.

R/IDateTime.R
JenspederM: Yeah okay. So you suggest that I exchange `trunc()` with `%/%`.

My reasoning was merely to be consistent in the use of `base` functions :-)

R/IDateTime.R
JenspederM: No not really.

Perhaps these should just be removed so that the decision of rounding will be up to the user.

JenspederM:
I have merged all changes in this branches and closed all other pull requests

R/IDateTime.R
JenspederM: I will change from `trunc` to `%/%`

MichaelChirico:
Just saw your comment abut the 30-second case. Please add a note about this to the documentation -- end users can decide for themselves how they want to resolve this issue for their specific use case. Workarounds are available, we should keep consistent to `base` as possible.

Thanks for merging the PRs here

R/IDateTime.R
MichaelChirico: Are you sure you're not allowed to have a `units` argument?

JenspederM: > floor.ITime <- function(x, units = c("hours", "minutes")) {
+     (setattr(switch(match.arg(units),
+                     hours = as.integer(floor(unclass(x)/3600L)*3600L),
+                     minutes = as.integer(floor(unclass(x)/60L)*60L)), 
+              "class", "ITime"))
+ }
> 
> floor(DT$time, units = "min")

Error in floor(DT$time, units = "min") : 
  2 arguments passed to 'floor' which requires 1

JenspederM: Yep :)

image

inst/tests/tests.Rraw
```diff
@@ -16774,3 +16774,18 @@ rm(s1, s2, class2132)
...
+DT[, ':='(
+  round_hour = round(time, "hour"),
+  round_min = round(time, "min"),
+  trunc_hour = trunc(time, "hour"),
+  trunc_min = trunc(time, "min"),
+  floor_hour = floor(time),
+  ceiling_hour = ceiling(time)
+)]
+
+test(2133.1, TRUE, all(sapply(DT, function(x) class(x) == "ITime")))
```
MichaelChirico: you should be testing for the _values_ as well, not just the output type.

please also include a test that touches on the rounding-at-30 issue you mentioned as well so that we can detect any changes to behavior in the future

NEWS.md
```diff
@@ -73,6 +73,8 @@ unit = "s")
 
 10. The dimensions of objects in a `list` column are now displayed, [#3671](https://github.com/Rdatatable/data.table/issues/3671). Thanks to @randomgambit for the request, and Tyson Barrett for the PR.
 
+11. Added support for `round()`, `trunc()`, `floor()`, `ceiling()` to extend functionality of `.ITime`. `round()` and `trunc()` can be used with arguments digits/units: "hours" or "minutes", whereas `floor()` and `ceiling()` only works for "hour". Thanks to @JensPederM for adding these functions.
```
jangorecki: `ITime` rather than `.ITime`

JenspederM:
> Just saw your comment abut the 30-second case. Please add a note about this to the documentation -- end users can decide for themselves how they want to resolve this issue for their specific use case. Workarounds are available, we should keep consistent to `base` as possible.
> 
> Thanks for merging the PRs here

Don't quite know where to add this....

MichaelChirico:
you'll need to add documentation, I believe the file is `man/IDateTime.R`

MichaelChirico:
Hey @JenspederM looking good! Still failing, you can poke around in the Travis logs, it can be a bit hard to know what to look for, but I see this:

```
Undocumented arguments in documentation object 'IDateTime'
  ‘units’
```

Meaning the tests are passing, but `R CMD check` is not because of the above

R/IDateTime.R
```diff
@@ -240,6 +240,22 @@ rep.ITime = function (x, ...)
   class(y) = "ITime"   # unlass and rep could feasibly not copy, hence use class<- not setattr()
   y
 }
+                           
+round.ITime <- function(x, digits = c("hours", "minutes"), ...)
```
MichaelChirico: I still think the signature here should use `units=` as the argument name if possible. That should also solve the Travis issue we're seeing IINM

MichaelChirico: Ahh, I see. Actually `round.IDate` is already using `digits=` as an argument, and `units=` is the undocumented object.

For consistency with `round.IDate`, I'm OK with `round.ITime` using `digits=`. Then we just have to document the `units` argument & should be OK

JenspederM: exactly. I noticed it when doing the documentation :)

JenspederM:
Found my mistake... Forgot to add units to \arguments{} in the documentation.

JenspederM:
Thank you for your patience and guidance :-D

mattdowle:
Great, thank you! I invited you to be project member; please accept the button that appears in your github profile or projects page. I also added you to contributors list. I couldn't find your full name so I used your id. Full name is better for attribution, if possible, since ids can change over time.

PR #4213, opened 2020-01-30
Use consistent types with fwriteMainArgs.nrow.

QuLogic:
On armv7hl, test 1737.5 fails due to garbage in the column length mismatch error message. This is because the message tries to format `args.nrow` (an `int64_t`) using `%d` (i.e., `int`). Strangely, this does not fail on any other architectures, but this is likely a fluke.

In all the other formatting calls, remove the unnecessary `(int64_t)` type cast, since `fwriteMainArgs.nargs` already is one.

See #3492.

### DIFF
--- a/src/fwriteR.c
+++ b/src/fwriteR.c
@@ -26,7 +26,7 @@ int getStringLen(SEXP *col, int64_t row) {
 int getMaxStringLen(const SEXP *col, const int64_t n) {
   int max=0;
   SEXP last=NULL;
-  for (int i=0; i<n; ++i) {
+  for (int64_t i=0; i<n; ++i) {
     SEXP this = *col++;
     if (this==last) continue; // no point calling LENGTH() again on the same string; LENGTH is unlikely as fast as single pointer compare
     int thisnchar = LENGTH(this);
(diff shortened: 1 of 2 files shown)


src/fwriteR.c
```diff
@@ -26,7 +26,7 @@ int getStringLen(SEXP *col, int64_t row) {
 int getMaxStringLen(const SEXP *col, const int64_t n) {
   int max=0;
   SEXP last=NULL;
-  for (int i=0; i<n; ++i) {
+  for (int64_t i=0; i<n; ++i) {
```
mattdowle: Great spot! Thanks,

src/fwriteR.c
```diff
@@ -228,8 +228,9 @@ SEXP fwriteR(
   int firstListColumn = 0;
   for (int j=0; j<args.ncol; j++) {
     SEXP column = VECTOR_ELT(DFcoerced, j);
-    if (args.nrow != length(column))
-      error(_("Column %d's length (%d) is not the same as column 1's length (%d)"), j+1, length(column), args.nrow);
```
mattdowle: Great spot too. Yes this could have been tripping up tests with a hard to track memory fault like you were seeing.

mattdowle:
Thanks for debugging these issues! I've invited you to be project member. Please accept the invitation button in your GitHub profile or projects page. I'll add you to DESCRIPTION too.

Issue #3060, opened 2018-09-21
progress indicator/bar for "by" operations


eantonya:
I frequently run `by` operations that take minutes and sometimes hours to complete. To deal with the uncertainty of what's going on, I often resort to printing `.BY` in the i-expression:

```
dt = data.table(a = 1:5)
dt[, {print(.BY)
      Sys.sleep(10)
      5}
   , by = a]
```

Would be great to have an automatic progress bar for these, similar to `fread`, if the expected run time is greater than e.g. 30s. Expected run time can be estimated from actual average time spent running the i-expression for each `by` item so far * number of remaining `by` items.

franknarf1:
Dupe of closed https://github.com/Rdatatable/data.table/issues/1409 

I manually add progress-printers when I see an operation going slowly, but it'd be nice if it were automatic whenever time > 30 s as you suggest.

mattdowle:
Related: #3050

jangorecki:
@franknarf1 agree with Frank, but lets print it only for `interactive()`.

eantonya:
I've been thinking about this and ideally I'd like to see the following information:

- number of groups processed
- total number of groups or groups remaining
- time elapsed
- total predicted time or time remaining
- maybe current group (but I'm not sure how you'd do this without creating a mess)

Kodiologist:
I too would like this as an opt-in feature of data.table. In the meantime, here's a code snippet to manually add a progress bar with a percentage:

```
bar = txtProgressBar(style = 3, min = 0,
    max = nrow(unique(d[, GROUP_EXPRESSION))))
d[, by = GROUP_EXPRESSION,
   {setTxtProgressBar(bar, .GRP)
    J_EXPRESSION}]
close(bar)
```

jangorecki:
> if the expected run time is greater than e.g. 30s. Expected run time can be estimated from actual average time spent running the i-expression for each by item so far * number of remaining by items.

`i` is not run for each item in `by`, even if it would, then still such estimation would be quite inaccurate because it doesn't say anything about `j`, to make that automatically we would have to measure time of iterations over progressing groups evaluated in `j`, and then decide to print at some threshold. This unfortunatelly would impose an overhead, it may be better to enable this feature explicitly rather than automatically.

MichaelChirico:
the same can be said of the progress bar for fread/fwrite right? at least w.r.t. the overhead concern

Kodiologist:
[...]

Issue #1206, opened 2015-07-03
Feature Request: local and global versions of `.I`, `.N`


MichaelChirico:
It's always been a bit confusing to me that `.I` is "global" in the sense that it doesn't change with `by`, while `.N` is "local" in the sense that it _does_.

I understand (some of) the advantages of this arrangement, but I think there are ample situations for using a local `.I` (see, e.g. [1](http://stackoverflow.com/questions/31187364/r-data-table-two-questions-about-by-1-i-and-2-order-by), [2](http://stackoverflow.com/questions/30314356/dcast-with-empty-rhs), [3](http://stackoverflow.com/questions/28072052/r-data-table-get-index-for-the-by-blocks/28072335#280723350), etc.) or a global `.N` (nothing off the top of my head, but I'm sure I've had to use `nrow(dt)` in a `j` argument before).

I'm not sure how easy this is to build into the source code, but having `.i` and `.n` be "local" while `.I` and `.N` are "global" seems like an intuitive alternative.

franknarf1:
I agree. It'd be quite a break from backward compatibility, but that notation would be useful and a lot more intuitive.

jangorecki:
Quite a big chance. Not sure if the performance gains are big enough to use `.i` instead of current `1:.N`, anybody measured it? 2.0.0 release is going to have some braking changes so this could be the place to release such change.

arunsrinivasan:
Like the idea very much, but not sure if it's possible at this stage, as it'd break *a lot* of code.. Where were you when this was implemented first :P?

Marked as FR for now.

MichaelChirico:
In my `R` swaddling blankets, I suppose ;)

I understand `.N`->`.n` is a big push, but I rarely need that.

`.i`, however, shouldn't break any code and I would use it all the time!

arunsrinivasan:
Right. But `seq_len(.N)` is `.i` what you're looking for.. Is that not okay? I ask because I find the intent quite clear and understandable. `.i` and `.I` can get quite confusing quickly. If it's really necessary, then maybe `.seqN`? Not sure. I'm always on alert when we've to add more symbols :-).

MichaelChirico:
[...]

PR #4215, opened 2020-01-30
Add .NGRP; the number of groups

MichaelChirico:
Part of #1206 and helps with _ad hoc_ versions of #3060 

Open to suggestions for the name, `.GRPN` is what I used at first, but `.NGRP` feels more natural to me.

The issue suggested `.BY.N` but that doesn't feel very natural to me.

If we have a few options I could run a Twitter poll.


NEWS.md
```diff
@@ -73,6 +73,8 @@ unit = "s")
 
 10. The dimensions of objects in a `list` column are now displayed, [#3671](https://github.com/Rdatatable/data.table/issues/3671). Thanks to @randomgambit for the request, and Tyson Barrett for the PR.
 
+11. New symbol in `j`, `.NGRP`, which counts the overall number of groups in the current grouping, part of [#1206](https://github.com/Rdatatable/data.table/issues/1206).
```
jangorecki: "which counts the overall" - for me it sounds a little bit like something that is not a constant, but _counts_ over current groups, maybe a wording improvement?

eddelbuettel: I don't understand what 'groups in current grouping' is supposed to say.

I read `.NGRP` as "cardinality" of group in current process step aka current subset?  Which would be nice to have indeed!

MichaelChirico: I'll try and come up with a quick example to add.

not sure what you mean by cardinality, but that sounds like just `.N`...

in base terms, do `y=split(DT, by='grp')`, `.N` is `nrow(y[[i]])` (it changes by group `i` which also has synonym `.GRP`) and `.NGRP` is `length(y)`

eddelbuettel: I think that is how I understood it at first, and it is very nice to have / a good add.

In my pedestrian understanding of SQL, set theory and alike 'cardinality' usually describe the count you get for 'select distinct' ie your ,NGRP here.  But if it confuses even you then it also not a good term to use here.   Maybe above just say "which counts the overall number of groups in the currently active data set" (or  "... selection") to avoid 'group' and 'grouping' ?

MichaelChirico: Ah, understood. From my background, cardinality means "the size of a set", just not clear which set is intended 😄

MichaelChirico: Thanks both by the way, I definitely felt the NEWS item was being half-assed but couldn't come up with anything better at the time, I'm glad you both called me out on it 😂

PR #4216, opened 2020-01-30
minor change to fcase error message receiving odd # args

### LINKED ISSUE #4211: Improve error message for fcase
```
y = runif(10)
fcase(y<.1, 1, y<.2, 2, 3)
```

Errors:

```
Please supply an even number of arguments in ..., consisting of logical condition, resulting value pairs (in that order); received 5 inputs.
```

The issue is that we need to explicitly state `default = 3`, it would be helpful to mention this possibility in the error message

MichaelChirico:
Closes #4211


2005m:
We can maybe add a small comment in the Rd file as well?

PR #4218, opened 2020-01-31
Tests for .ITime: round, trunc, floor, and ceiling

JenspederM:
Tests should be extended for correctness. 
One known mistake is in round() when second == 30 since this will make division == x.5, which means that it will round up or down. There is no logic for when it does what. But this is a general problem with the round() function, don't know how to address this.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16774,3 +16774,18 @@ rm(s1, s2, class2132)
 ########################
 #  Add new tests here  #
 ########################
+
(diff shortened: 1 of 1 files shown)


jaapwalhout:
> There is no logic for when it does what.

Yes there is and it is documented in the IEC 60559 standard as mentioned in `?round`.

[For more explanation, see my answer on StackOverflow](https://stackoverflow.com/a/39913257/2204410).

JenspederM:
I’ve added a link to `?round` in the documentation. See The open PR, please.

PR #4223, opened 2020-02-01
convert r-forge issue numbers to github issue numbers where easy to do

### LINKED ISSUE #4222: Make some effort to re-number old issue tracker #s to avoid collisions
There are still a lot of issue numbers from the old bug tracker (R-forge) in our code. For now, many are easy to spot since they have issue numbers >5,000, however, the GH issue tracker is approaching it's 5,000th issue/PR.

If possible it would be nice to edit any issue numbers we can identify as belonging to the old tracker, hopefully in a single commit to minimize the confusion in the git history.

Here's what I see at a glance:

```
grep -Enr "#[5-9][0-9]{3}" {R,src}/*.{R,c}
R/as.data.table.R:36:  # Fix for bug #5408 - order of columns are different when doing as.data.table(with(DT, table(x, y)))
R/data.table.R:60:  if (length(x)==1L && (is.null(x[[1L]]) || (is.list(x[[1L]]) && length(x[[1L]])==0L))) return( null.data.table() ) #5377
R/data.table.R:134:    # Fix for #5070 (to do)
R/data.table.R:139:    # added is.data.table(ans) check to fix bug #5069
R/data.table.R:921:          # fix for #5190. colsub[[1L]] gave error when it's a symbol.
R/data.table.R:1007:        ansvars = setdiff(allcols, bynames) # fix for bug #5443
R/data.table.R:1246:        setattr(ans, "class", class(x)) # fix for #5296
R/data.table.R:1287:    # Temp fix for #921 - check address and copy *after* evaluating 'jval'.  #5114 also related.
R/data.table.R:1326:      setattr(jval, 'class', class(x)) # fix for #5296
R/data.table.R:1374:    # if 'get' is in 'av' use all cols in 'i', fix for bug #5443
R/data.table.R:1487:        # replaced SDenv$.SD to SDenv to deal with Bug #5007 reported by Ricardo (Nice catch!)
R/data.table.R:1631:    dotN = function(x) is.name(x) && x==".N" # For #5760. TODO: Rprof() showed dotN() may be the culprit if iterated (#1470)?; avoid the == which converts each x to character?
R/data.table.R:1644:          if (dotN(q)) return(TRUE) # For #5760
R/data.table.R:1665:              if (dotN(jsub[[ii]])) next; # For #5760
R/data.table.R:1746:    assign(".N", len__, thisEnv) # For #5760
R/data.table.R:1799:  setattr(ans,"class",class(x)) # fix for #5296
R/data.table.R:2826:          "median", "min", "max", "var", "sd", ".N") # added .N for #5760
R/duplicated.R:14:  if (nrow(x) == 0L || ncol(x) == 0L) return(logical(0L)) # fix for bug #5582
R/duplicated.R:17:  # fix for bug #5405 - unique on null data table returns error (because of 'forderv')
R/duplicated.R:96:# FR #5172 anyDuplicated.data.table
R/merge.R:78:      # takes care of #5672 without having to save names. This is how it should be, IMHO.
R/print.data.table.R:74:  # FR #5020 - add row.names = logical argument to print.data.table
R/print.data.table.R:77:    # fixes bug #97 (RF#4934) and #545 (RF#5253)
R/print.data.table.R:146:      paste(c(format(head(x, 6L), justify=justify, ...), if (length(x) > 6L) "..."), collapse=",")  # fix for #5435 - format has to be added here...
R/print.data.table.R:173:    else col = format(char.trunc(col), justify=justify, ...) # added an else here to fix #5435
grep: R/*.c: No such file or directory
grep: src/*.R: No such file or directory
src/dogroups.c:35:    // Fix for #5437, by cols with attributes when also used in `j` lost the attribute.
src/dogroups.c:42:  setAttrib(BY, R_NamesSymbol, bynames); // Fix for #5415 - BY doesn't retain names anymore
src/dogroups.c:104:    // Previously had replaced (i>0 || !isNull(lhs)) with i>0 to fix #5376
src/reorder.c:63:    const size_t size = SIZEOF(v);    // size_t, otherwise #5305 (integer overflow in memcpy)
src/subset.c:268:  // class is also copied here which retains superclass name in class vector as has been the case for many years; e.g. tests 1228.* for #5296
```

We can easily match any of these to those that Arun brought into the GH tracker, and replace them with the GH issue #. Any others, we should mark off as `RF#` to indicate they're from R-forge.

From the GH tracker:

```
library(gh)
library(data.table)

repo = 'Rdatatable/data.table'
issue_query_fmt = "/repos/%s/issues?state=all&per_page=100&page=%d"
issues = list()
page = 1L
while (all(nzchar(resp <- gh(sprintf(issue_query_fmt, repo, page))))) {
  print(page)
  issues = c(issues, resp)
  page = page + 1L
}

# filter pull requests
issues = issues[sapply(issues, function(issue) is.null(issue$pull_request))]

titles = sapply(issues, `[[`, 'title')
numbers = sapply(issues, `[[`, 'number')
rf_idx = grep('r.*forge.*#', titles, ignore.case = TRUE)
rf_numbers = as.integer(gsub('.*#([0-9]+).*', '\\1', titles[rf_idx]))

mapping = data.table(
  gh_number = numbers[rf_idx],
  rf_number = rf_numbers
)
setorder(mapping, rf_number)
fwrite(mapping, 'mapping.csv')
```

Attached `mapping` produced

[mapping.txt](https://github.com/Rdatatable/data.table/files/4142355/mapping.txt)

MichaelChirico:
Closes #4222

### DIFF
--- a/R/as.data.table.R
+++ b/R/as.data.table.R
@@ -31,7 +31,7 @@ as.data.table.Date = as.data.table.ITime = function(x, keep.rownames=FALSE, key=
   as.data.table.list(x, FALSE, key)
 }
 
-# as.data.table.table - FR #4848
+# as.data.table.table - FR #361
 as.data.table.table = function(x, keep.rownames=FALSE, key=NULL, ...) {
   # Fix for bug #43 - order of columns are different when doing as.data.table(with(DT, table(x, y)))
   val = rev(dimnames(provideDimnames(x)))
--- a/R/as.data.table.R
+++ b/R/as.data.table.R
@@ -33,7 +33,7 @@ as.data.table.Date = as.data.table.ITime = function(x, keep.rownames=FALSE, key=
 
 # as.data.table.table - FR #4848
 as.data.table.table = function(x, keep.rownames=FALSE, key=NULL, ...) {
-  # Fix for bug #5408 - order of columns are different when doing as.data.table(with(DT, table(x, y)))
+  # Fix for bug #43 - order of columns are different when doing as.data.table(with(DT, table(x, y)))
   val = rev(dimnames(provideDimnames(x)))
   if (is.null(names(val)) || !any(nzchar(names(val))))
     setattr(val, 'names', paste0("V", rev(seq_along(val))))
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -919,7 +919,7 @@ replace_dot_alias = function(e) {
         } else {
           # FR #355 - negative numeric and character indices for SDcols
           colsub = substitute(.SDcols)
-          # fix for RF#5190. colsub[[1L]] gave error when it's a symbol.
+          # fix for R-Forge #5190. colsub[[1L]] gave error when it's a symbol.
           if (is.call(colsub) && deparse(colsub[[1L]], 500L, backtick=FALSE) %chin% c("!", "-")) {
             negate_sdcols = TRUE
             colsub = colsub[[2L]]
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -655,7 +655,7 @@ replace_dot_alias = function(e) {
       }
       if (is.logical(j)) j <- which(j)
       if (!length(j) && !notj) return( null.data.table() )
-      if (is.factor(j)) j = as.character(j)  # fix for FR: #4867
+      if (is.factor(j)) j = as.character(j)  # fix for FR: #358
       if (is.character(j)) {
         if (notj) {
           if (anyNA(idx <- chmatch(j, names_x))) warning("column(s) not removed because not found: ", brackify(j[is.na(idx)]))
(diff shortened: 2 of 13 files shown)


MichaelChirico:
There's still a bunch of references in `NEWS.0.md`, I think we can regard that as a historical document and not update. Or maybe add a small note about it at the top...

MichaelChirico:
I had a go at using this to convert the rest of the R-Forge numbers that are already duplicates of GH numbers (and for which we know the RF->GH mapping), but there are a lot of them (roughly 400) and the diff is pretty large, it would require a lot of manual work to track down which cases are actually RF and which are actually GH numbers.

It _may_ be possible to use `git blame` here (cross-referencing when the issue number in the comment was written vs when the corresponding GH issue was created), but that sounds like overkill.

for reference:

```
all_paths = list.files('.', pattern = '*\\.(R|Rraw|c)$', recursive = TRUE, full.names = TRUE)
all_files = lapply(all_paths, readLines)

for (ii in mapping[rf_number < 4223, which = TRUE]) {
  all_files = lapply(all_files, function(f) {
    with(mapping[ii], gsub(sprintf('#%d\\b', rf_number), sprintf('#%d', gh_number), f))
  })
}

names(all_files) = all_paths

for (path in all_paths) writeLines(all_files[[path]], path)
```

PR #4238, opened 2020-02-11
fread length-1 colClasses=NA_character_

### LINKED ISSUE #4237: fread colClasses=NA_character_ incorrect
Follow up to #1910. Not sure how this lasted so long.
```R
> fread("A,B\n1,3\n2,4\n", colClasses=c(NA_character_, NA_character_))
       A     B      # correct
   <int> <int>
1:     1     3
2:     2     4
> fread("A,B\n1,3\n2,4\n", colClasses=NA_character_)
        A      B    # incorrect
   <char> <char>
1:      1      3
2:      2      4
> 
```

There's a length-1 specific case at C level which doesn't deal with the NA properly when colClasses is length 1.

mattdowle:
Closes #4237 
Removed the branch for length 1 colClasses at C level so that logic is in one place.

### DIFF
--- a/src/freadR.c
+++ b/src/freadR.c
@@ -241,7 +241,7 @@ static void applyDrop(SEXP items, int8_t *type, int ncol, int dropSource) {
   UNPROTECT(1);
 }
 
-bool userOverride(int8_t *type, lenOff *colNames, const char *anchor, int ncol)
+bool userOverride(int8_t *type, lenOff *colNames, const char *anchor, const int ncol)
 {
   // use typeSize superfluously to avoid not-used warning; otherwise could move typeSize from fread.h into fread.c
   if (typeSize[CT_BOOL8_N]!=1) STOP(_("Internal error: typeSize[CT_BOOL8_N] != 1")); // # nocov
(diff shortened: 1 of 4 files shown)


src/freadR.c
```diff
@@ -241,7 +241,7 @@ static void applyDrop(SEXP items, int8_t *type, int ncol, int dropSource) {
   UNPROTECT(1);
 }
 
-bool userOverride(int8_t *type, lenOff *colNames, const char *anchor, int ncol)
+bool userOverride(int8_t *type, lenOff *colNames, const char *anchor, const int ncol)
```
MichaelChirico: there's a compiler warning for declaring const incorrectly right? what about variables that _could_ be const, is that something that could be detected?

jangorecki: AFAIU more of the variables we are using should be constant as we are more commonly reading values and not writing.

MichaelChirico: yep... any tool we can use to help us increase usage?

jangorecki: I don't think it is that much important to hunt for such cases.

PR #4242, opened 2020-02-13
Add ties.method="last" to frank

MichaelChirico:
Closes #1689

PR #4243, opened 2020-02-14
fix tests for ties.method='last' to work on R 3.1

MichaelChirico:
Follow up to #4242

`ties.method='last'` only introduced on R 3.3 < 3.1, so the tests comparing `base::rank` to `frank` fail there.

This approach shuts of `ties.method='last'` testing for R 3.1 (easiest solution).

cc @jangorecki 

PS anyone remember why we apparently shut off `ties.method='random'` testing? It was failing for me here, not sure why.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -109,6 +109,8 @@ unit = "s")
 
 4. `fifelse` and `fcase` notify users that S4 objects (except `nanotime`) are not supported [#4135](https://github.com/Rdatatable/data.table/issues/4135). Thanks to @torema-ed for bringing it to our attention and Morgan Jacob for the PR.
 
+5. `frank(..., ties.method="random", na.last=NA)` now returns the same random ordering that `base::rank` does, [#4243](https://github.com/Rdatatable/data.table/pull/4243).
+
 
 # data.table [v1.12.8](https://github.com/Rdatatable/data.table/milestone/15?closed=1)  (09 Dec 2019)
 
--- a/R/frank.R
+++ b/R/frank.R
@@ -32,11 +32,14 @@ frankv = function(x, cols=seq_along(x), order=1L, na.last=TRUE, ties.method=c("a
     nas  = x[[ncol(x)]]
   }
   if (ties.method == "random") {
-    v = stats::runif(nrow(x))
     if (is.na(na.last)) {
       idx = which_(nas, FALSE)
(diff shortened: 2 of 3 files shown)


mattdowle:
I loooked at history and I couldn't find `ties.method='random'` test being turned off. Rather it just seems it wasn't tested from the start: https://github.com/Rdatatable/data.table/commit/1ff5e1e53d3495678251f68b4b02ebb398b6ae2f#diff-d322b46c285376a06f948ecef50b6763
Then you caught it and added coverage in test 1962.027.
So the question remains why `random` can't be added in this loop ...

mattdowle:
Ok I get a test fail too. Just in the second test (1369) with `na.last=NA`. So I left `"random"` on in 1368 and added a comment to 1369 so this branch should pass for now.
Will have another look later to see why there's a difference with na.last=NA ...

MichaelChirico:
I did look at it briefly and yes I think NA is the issue

```
# base::rank
random = sort.list(order(x, stats::runif(sum(!nas))))
# frankv
if (ties.method == "random") {
    v = stats::runif(nrow(x))
    if (is.na(na.last)) {
      idx = which_(nas, FALSE)
      set(x, idx, '..stats_runif..', v[idx])
    } else set(x, NULL, '..stats_runif..', v)
    order = if (length(order) == 1L) c(rep(order, length(cols)), 1L) else c(order, 1L)
    cols = c(cols, ncol(x))
  }
```

`runif` is called potentially a different number of times in the two functions which would throw off the seeds

mattdowle:
But the test sets the seed before the calls. Isn't it `runif(sum(!nas)))` vs `runif(nrow(x))` i.e. frank including the NA positions to remove them afterwards could result in a different ordering than the base way.  Both are right I guess, just different.

MichaelChirico:
say the first element of the input is NA

then frank will use the first random draw on an NA, while base::rank will not take any draws, so the random ordering will end up different IINM.

I don't think we need to force exact consistency between rank and frank here -- random ordering should be random after all

mattdowle:
Yes both are random. But we may as well match base unless there's a reason not to. In this case it was easier to match base and turn the test on with 'random' included so it could all be put to bed. Otherwise we'd have had to find another way to test random with na.last=NA.

MichaelChirico:
nice! I see the difference was only a few lines so yes that's preferred

PR #4246, opened 2020-02-17
use only private runner for windows

jangorecki:
Recent failures of dev-win CI job were caused by the fact, that when release-win job was running on our private windows runner, then dev-win job was waiting and it was being picked up by gitlab public windows runners (finally it is there!). Those public runners has different configuration, thus `export PATH` was already failing. This PR should solve the problem. Extra `private` tag has been added to our private windows runner.


jangorecki:
for reference news from gitlab about that https://about.gitlab.com/releases/2020/01/22/gitlab-12-7-released/

Issue #852, opened 2014-10-03
`[.data.table` needs to be more modular.


arunsrinivasan:

jangorecki:
My suggestion regarding this issue would be to isolate `[.data.table` and `bmerge` by adding separate layer-function for preparing `i` data (key manipulation, attributes, etc.) from `i` argument (and possibly some others).
At the current moment the `auto.index` feature can fits in that new intermediate function but there are multiple FR which can be implemented there to not increase complexity of `[`.

jangorecki:
Any have any ideas on this?  

Currently `[.data.table` has a huge chunks of code like `if(missing(i)) ...`, `if(missing(j))`, `# Grouping ...` which has quite a lot nested conditions.  
I think the separation of evaluation of *control flow* conditions from the computation on the dataset can bring more readability of the code. It also allows better isolation of *scenarios* to compute, so results in a more modular code.  
High level view on the current `[.data.table` implementation base on few samples of code:  

```{r}
...
if (!missing(i)) {
        ...
        if (is.call(isub) && isub[[1L]] == as.name("order") && getOption("datatable.optimize") >= 1) {...}
        ...
        if (is.call(isub) && isub[[1L]] == quote(forder)) {...} else if (is.call(isub) && getOption("datatable.auto.index") &&
                 as.character(isub[[1L]]) %chin% c("==","%in%") &&
                 is.name(isub[[2L]]) &&
                 (isub2<-as.character(isub[[2L]])) %chin% names(x) && 
                 is.null(attr(x, '.data.table.locked'))) {...}
} else {...}
...
if (missing(j)) {...} else {
        ...
        if (is.call(jsub) && jsub[[1L]]==as.name(":=")) {...}
        ...
}
...
```

Above code in a proposed structure:    

```{r}
# control flow metadata
do_i = !missing(i)
do_j = !missing(j)
optimize_i = do_i && getOption("datatable.optimize") >= 1 && is.call(isub) && isub[[1L]] == as.name("order")
x_unlocked = is.null(attr(x, '.data.table.locked'))
create_index = do_i && getOption("datatable.auto.index") && x_unlocked && is.call(isub) && as.character(isub[[1L]]) %chin% c("==","%in%") && is.name(isub[[2L]]) && as.character(isub[[2L]]) %chin% names(x)
update_by_ref = do_j && is.call(jsub) && jsub[[1L]]==as.name(":=")

# computation
if(do_i){
        if(optimize_i) {...}
[...]

PR #4248, opened 2020-02-18
Column Select Helper

### LINKED ISSUE #4115: .SDcols with logical vector
In `?data.table` about `.SDcols` argument:

>May be character column names or numeric positions. 

However, it seems like also a `logical` vector is possible, e.g.

    d <- data.table(x = 1, y = "a")
    d[ , .SD, .SDcols = sapply(d, is.numeric)]
    #    x
    # 1: 1

If selecting columns with a logical vector in `.SDcols` is intended to work, should this possibility be added to the help text?
 
Please also note two differences with `j` in this regard:

(1) Using a logical vector longer than the number of columns errors in `j`, but not in `.SDcols`

    d[ , c(TRUE, TRUE, TRUE)]
    # Error in `[.data.table`(d, , c(TRUE, TRUE, TRUE)) : 
    #  Item 3 of j is 3 which is outside the column number range [1,ncol=2]

    d[ , .SD, .SDcols = c(TRUE, TRUE, TRUE)]
    #    x y
    # 1: 1 a

(2) If a logical vector with length `k < ncol(x)` is passed to `j`, it will be filled to length `ncol(x)` with `FALSE`. In `.SDcols`, the logical vector is recycled.

    d[ , TRUE]
    #    x
    # 1: 1

    d[ , .SD, .SDcols = TRUE]
    #    x y
    # 1: 1 a

----

`data.table` 1.12.9 IN DEVELOPMENT built 2019-12-13 23:32:57 UTC

### LINKED ISSUE #4231: internal error when specifying .SDcols
There is an issue in scope in which `.SDcols` is evaluated.
AFAIU there is no need to evaluate it in [`x`'s scope](https://github.com/Rdatatable/data.table/commit/b0ecc0e56850a543cef874106fcda11f7fafc6ac#diff-650e4a11ed4384f8e560a6ebfff4ff53R897), and that is likely caused an issue here.
```r
data.table(V1=-1L, V2=-2L, V3=-3L)[,.SD,.SDcols=-V2:-V1]
#Error in `[.data.table`(data.table(V1 = -1L, V2 = -2L, V3 = -3L), , .SD,  : 
#  Internal error: xcolAns does not pass checks: 1121
```
```r
as.data.table(as.list(1:2))[, .SD,.SDcols=(-1L)]
#Error in colnamesInt(x, cols, check_dups = FALSE) : 
#  argument specifying columns specify non existing column(s): cols[1]=-1
as.data.table(as.list(1:3))[, .SD,.SDcols=(-1L)]
#Error in `[.data.table`(as.data.table(as.list(1:3)), , .SD, .SDcols = (-1L)) : 
#  Internal error: xcolAns does not pass checks: 2211,2
```

ColeMiller1:
Closes #4115 
Closes #4231. 
Towards #852. 

33 errors until ```by``` is included...

See also @jangorecki very detailed approach in C in #4174. I see this as a stepping stone to getting towards Jan's C solution as we solve some of the API issues.

Mostly internal replacing the ```.SDcols``` evaluation with a new helper function with slight refactoring. 

## Functionality differences:

1. If .SDcols evaluates to ```(-1L)``` and the integer is within the column range of dt, the set difference is returned (e.g. someone did ```.SDcols = (-3L)``` on a 5 column data.table which would return ```c(1L, 2L, 4L, 5L)```.
2. If .SDcols is a logical greater than length 1 but less than the length of the dt, a warning message is displayed. The long term solution will be for it to throw an error and later on silently fail. 

## Internal Differences

1. The way ```1:3``` or ```V1:V3``` is evaluated was changed to be more performant. 
```
x = data.table(V1 = 1, V2 = 2, V3 = 1, V4= 5, V5 = 5)
colsub = quote(V1:V3)
  new_way = {
    rnge = data.table:::chmatch(c(as.character(colsub[[2L]]), as.character(colsub[[3L]])), names(x))
    cols = rnge[1L]:rnge[2L]
  },
  old_way =  eval(colsub, data.table:::setattr(as.list(seq_along(x)), 'names', names(x)), parent.frame())
)
##Unit: microseconds
##    expr  min    lq   mean median    uq   max neval
## new_way 20.8 22.05 31.324   23.0 28.85 145.4   100
## old_way 27.8 28.90 35.059   29.7 30.70 137.2   100

colsub = quote(1:3)
  new_way =  eval(colsub),
  old_way =  eval(colsub, data.table:::setattr(as.list(seq_along(x)), 'names', names(x)), parent.frame())
)
##Unit: microseconds
##    expr  min   lq   mean median    uq  max neval
## new_way 10.6 11.4 12.689   11.8 12.20 67.4   100
## old_way 27.2 28.0 30.421   28.5 29.15 70.9   100
```
2. The cool new ```%iscall%``` is used less frequently because the parsing is evaluated differently.
3. The evaluation of ```patterns``` does not actually include ```do_patterns```. In the context of ```.SDcols```, ```j```, or ```by```, the argument of ```patterns(..., cols = SOMETHING)``` does not make sense. Instead, ```grep``` is used directly in the context of ```names(x)```.

## Random note

1. At the end of the tests, there is a new variable ```DT``` in the global environment. Both 2036.1 and  2036.2 produce the new variable. The tests are below:
```
setup = c('DT = data.table(a = 1)')
writeLines(c(setup, 'DT[ , a := 1]'), tmp<-tempfile())
test(2036.1, !any(grepl("1:     1", capture.output(source(tmp, echo = TRUE)), fixed = TRUE)))
## test force-printing still works
writeLines(c(setup, 'DT[ , a := 1][]'), tmp)
test(2036.2, source(tmp, echo = TRUE), output = "1:\\s+1")
```
To Do:
- [x] See if any more tests are needed

Long Term:

- Include by vars and j vars
- Consolidate other column select which would likely include duplicated, unique, setcolorder, setnames, and probably more.
- See what vignettes could be updated.


inst/tests/tests.Rraw
```diff
@@ -6968,8 +6968,8 @@ test(1497, DT[, .SD, .SDcols = !c("a", "c")], DT[, !c("a", "c"), with=FALSE])
 
 # Fix for #1060
 DT = data.table(x=1, y=2, z=3, a=4, b=5, c=6)
-test(1498.1, DT[, .SD, .SDcols=c(TRUE,FALSE)], DT[, c("x", "z", "b"), with=FALSE])
-test(1498.2, DT[, .SD, .SDcols=!c(TRUE,FALSE)], DT[, !c("x", "z", "b"), with=FALSE])
+test(1498.1, DT[, .SD, .SDcols=c(TRUE,FALSE)], DT[, c("x", "z", "b"), with=FALSE], warning = "When .SDcols is a logical vector, each column should have a TRUE or FALSE entry. Vector will be repeated to length of data.table. Warning will change to error in year 2022.")
```
MichaelChirico: we are usually doing a deprecation cycle rather than a fixed date, e.g. it becomes warning on this release, then error on next release

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+  ## or anything that looks for columns
+  
+  #' @param x required; a data.frame, data.table, tibble, or list.
+  #' @param colsub takes the same arguments as .SDcols such as
+  #' `patterns()`, `var1:var2`, a variable, or a vector of characters.
+  #' When TRUE, will return seq_len(length(x))
+  #' @param origin_call takes character vector of length 1. In the short-term
+  #' there will be different modes based on the call. In the long-term, it's just
+  #' to help the end-user what caused the error
+  #' @return Integer vector with an attribute of whether negation was in the call
```
MichaelChirico: great to see internal documentation like this!

I would be more explicit here "a boolean attribute, `"negate"`"

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+  
+  if (length(colsub) == 1L && isTRUE(colsub)){
+    ansvals = seq_len(x_len)
+    attr(ansvals, "negate") = FALSE
+    return(ansvals)
+  }
+  
+  # FR #4979 - negative numeric and character indices for SDcols
+  # fix for #5190. colsub[[1L]] gave error when it's a symbol.
+  negate_cols = is.call(colsub) && deparse(colsub[[1L]], 500L, backtick=FALSE) %in% c("!", "-")
```
MichaelChirico: we could use `%iscall%` here?

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+  if (negate_cols) colsub = colsub[[2L]]
```
MichaelChirico: `!` is unary operator so `[[2L]]` works without qualification, but `-` might have `[[3L]]`, I remember this came up in an issue involving `with=FALSE` somewhat recently

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+  
+  if (is.call(colsub)){
+    # fix for #1216, make sure the parentheses are peeled from expr of the form (((1:4)))
+    while(colsub %iscall% "(") colsub = as.list(colsub)[[-1L]]
```
MichaelChirico: should include `{` here too maybe?

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+    if (length(colsub) == 3L && colsub[[1L]] == ":") {
```
MichaelChirico: Maybe switch to `%chin%`? Since Matt added support to avoid coercion of `SYMSXP` to `STRSXP` IIUC. Worth benchmarking maybe.

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+      if (is.name(colsub[[2L]])){
+        # cols is of the format a:c
+        rnge = chmatch(c(as.character(colsub[[2L]]), as.character(colsub[[3L]])), names(x))
```
MichaelChirico: what about `!is.name(colsub[[3L]])`?

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+        cols = rnge[1L]:rnge[2L]
+      } else {
+        # cols is of the format 1:3
+        cols = eval(colsub)
```
MichaelChirico: similarly `is.name(colsub[[3L]])`

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+      } else { #e.g. patterns("one", "or", "more", "patterns") 
+        x_names = names(x)
+        cols = Reduce(intersect, lapply(colsub, function(col) grep(eval(col, parent.frame(2L)), x_names)))
+      }
+    } else {
+      cols = eval(colsub, parent.frame(2L))
+    }
+  } else {
+    # if colsub is also a name in the list, we prioritize it over anything in the environment
+    if (exists(as.character(colsub), where = x))
```
MichaelChirico: I see `colsub` coerced twice, here's a less readable version:

```
if (!exists(cols <- as.character(colsub), where = x)) cols = eval(colsub, parent.frame(2L), parent.frame(2L))
```

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+    cols = unlist(cols, use.names = FALSE)
+  }
+  
+  if (origin_call == "by") {
+    #check for weird comma delimited "V1,V2,V3" and give warning message
+    #in the by eval we'll check if by = list(...) otherwise we can come here.
+  }
+  
+  if (anyNA(cols))
+    stop(origin_call, " missing at the following indices: ", brackify(which(is.na(cols))))
```
MichaelChirico: please use `gettextf` 😃

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+  
+  if (is.logical(cols)) {
+    if (length(cols) != x_len) {
+      ## TODO change to error in 2022
+      warning("When ", origin_call, " is a logical vector, each column should have a TRUE or FALSE entry. Vector will be repeated to length of data.table. Warning will change to error in year 2022.")
```
MichaelChirico: ditto `gettextf`

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+      cols = rep(cols, length.out = x_len)
+    }
+    cols = which_(cols)
+  }
+                                        
+  if (is.numeric(cols)) {
+    cols = as.integer(cols)
+    if (length(cols) == 1L && cols < 0L && abs(cols) <= x_len) cols = setdiff(seq_len(x), abs(cols))
```
MichaelChirico: no need for `abs` call here, we already checked `cols < 0L` so just do `-cols <= -x_len`

R/utils.R
```diff
@@ -140,3 +140,107 @@ edit.data.table = function(name, ...) {
...
+  #' @param origin_call takes character vector of length 1. In the short-term
+  #' there will be different modes based on the call. In the long-term, it's just
+  #' to help the end-user what caused the error
+  #' @return Integer vector with an attribute of whether negation was in the call
+  
+  # see @jangorecki #4174 for how to get the vast majority of this into C
+  
+  x_len = length(x)
+  
+  if (length(colsub) == 1L && isTRUE(colsub)){
```
MichaelChirico: this branch & L231 need tests

MichaelChirico:
Looks great overall! A much needed refactor

R/utils.R
ColeMiller1: Keeping as is - ```colsub[[1L]] == ":"``` is an order of magnitude faster. Still, I don't see benchmarks using nanoseconds that frequently.
```r
e = quote(1:3)

Unit: nanoseconds
                                expr  min   lq mean median   uq   max neval
                      e[[1L]] == ":"  400  500  647    600  700  2600   100
 data.table:::"%chin%"(e[[1L]], ":") 5200 5500 6126   5700 5900 39500   100

File: R/utils.R (hunk 3 above)
Comment (ColeMiller1) [reply R/utils.R]: Yes! I kept the original blindly in there - who knows what's up with the ```500L``` and ```backtick=FALSE)```.  

Do we care about microsecond performance? ```%iscall%``` is generally slower than ```is.call(e) && e[[1L]] == "something"```. But on the bright side, it looks great.

```r
## looks great
  colsub %iscall% c("!", "-")
## is faster
  is.call(colsub) && (colsub[[1L]] == "!" || colsub[[1L]] == "-")

e = quote(1:3)
  is.call(e) && (e[[1L]] == ":" || e[[1L]] == "!"),
  data.table:::'%iscall%'(e, c(":", "!")),
  is.call(e) && as.character(e[[1L]]) %in% c(":", "!")
)
Unit: nanoseconds
                                                 expr  min   lq  mean median  
     is.call(e) && (e[[1L]] == ":" || e[[1L]] == "!")  900 1100  1681   1400  
              data.table:::"%iscall%"(e, c(":", "!")) 6400 6900 10446   7650 
 is.call(e) && as.character(e[[1L]]) %in% c(":", "!") 2000 2200  3269   2400

File: inst/tests/tests.Rraw (hunk 1 above)
Comment (ColeMiller1) [reply inst/tests/tests.Rraw]: I like dates because it's more tangible to the end-user. But I assume this would be warning until 1.14 and then make it as an error?

**Comment by jangorecki** [comment]
My general advice would be to split `j` handling from `.SDcols` handling into separte function, that eventually share some internal functions.

**Comment by jangorecki** [comment]
> .SDcols now accepts a single name as an argument such as data.table(V1 = 1)[, .SD, .SDcols = V1]

I would say this is unexpected.

**Comment by jangorecki** [comment]
> If .SDcols is TRUE, all columns will be returned which is a step towards group_by_all or just a generic way to select all columns.

IMO .SDcols when being logical should always be equal length to ncol, otherwise raise exception.

File: R/utils.R (hunk 3 above)
Comment (jangorecki) [reply R/utils.R]: in this benchmark you also include `:::` call, which IMO should not be included

File: R/utils.R (hunk 7 above)
Comment (jangorecki) [reply R/utils.R]: exactly, some recent commit in my branch was addressing cases like `var1:5` or `1:var1`

File: R/utils.R (hunk 3 above)
Comment (MichaelChirico) [reply R/utils.R]: after defining `%iscall%` in scope:

```
Unit: nanoseconds
                                                 expr  min   lq     mean median     uq     max neval cld
     is.call(e) && (e[[1L]] == ":" || e[[1L]] == "!")  724  788  1037.61  817.5  867.0   12307   100   a
                               e %iscall% c(":", "!") 1661 1738 23163.82 1793.0 1852.5 2123262   100   a
 is.call(e) && as.character(e[[1L]]) %in% c(":", "!") 1319 1424  1636.51 1500.5 1607.0   10766   100   a
```

Comment (jangorecki) [reply R/utils.R]: these are nonoseconds so it I thing those numbers may vary a lot on different machines

Comment (MichaelChirico) [reply R/utils.R]: increasing `neval`

```
Unit: nanoseconds
                                                 expr  min   lq      mean median   uq       max neval cld
     is.call(e) && (e[[1L]] == ":" || e[[1L]] == "!")  461  545  823.7597    580  626 169334249 1e+06  a 
                               e %iscall% c(":", "!") 1050 1206 1449.7929   1289 1380  23513043 1e+06   b
 is.call(e) && as.character(e[[1L]]) %in% c(":", "!")  819  989 1178.1340   1060 1158   2651089 1e+06   b
```

Here, it's not fair to use `e = quote(1:3)` since we _always_ stop out at the first part of `||`, here's the timing switching the order:

```
Unit: nanoseconds
                                                 expr  min   lq     mean median   uq       max neval cld
     is.call(e) && (e[[1L]] == "!" || e[[1L]] == ":")  760  854 1049.941    888  941  23475615 1e+06   a
                               e %iscall% c(":", "!") 1039 1184 1480.185   1230 1312  10060629 1e+06   a
 is.call(e) && as.character(e[[1L]]) %in% c(":", "!")  822  967 1498.495   1024 1100 277516160 1e+06   a
```

Lastly, `%in%` is not optimized for `character` input, which is why we made `%chin%`:

```
Unit: nanoseconds
                                                   expr  min   lq     mean median   uq      max neval cld
       is.call(e) && (e[[1L]] == "!" || e[[1L]] == ":")  759  853 1037.578    896  968 19875883 1e+06 a  
                                 e %iscall% c(":", "!") 1024 1178 1370.742   1233 1329 10193437 1e+06   c
 is.call(e) && as.character(e[[1L]]) %chin% c(":", "!")  805  945 1143.439    998 1077 24712648 1e+06  b 
```

Most of the gap appears to be due to the overhead of using `%chin%` instead of its code directly:

```
Unit: nanoseconds
                                             expr  min   lq     mean median   uq       max neval cld
 is.call(e) && (e[[1L]] == "!" || e[[1L]] == ":")  762  861 1186.890    899  966 195036437 1e+06   a
                           e %iscall% c(":", "!") 1030 1176 1383.657   1230 1316  10084976 1e+06   a
         is.call(e) && e[[1L]] %chin% c(":", "!")  724  842 1025.460    889  959  26607578 1e+06   a
```

Comment (jangorecki) [reply R/utils.R]: are you getting similar results when trying on a different machine?

File: R/utils.R (hunk 4 above)
Comment (jangorecki) [reply R/utils.R]: yes, there might be `quote(v1-v2)` which is still `-` call but length 3

File: R/utils.R (hunk 3 above)
Comment (ColeMiller1) [reply R/utils.R]: I stand corrected, thanks! So when you do ```cc()```, you can just do ```e %iscall% ":"``` without any additional hurdles? I only used ```:::``` because ```%iscall%``` is not exported.

File: R/utils.R (hunk 6 above)
Comment (ColeMiller1) [reply R/utils.R]: Based on other conversation, my benchmark is likely faulty. Still, I doubt ```%chin%``` would improve performance over the current 600 nanoseconds. I am re-resolving changing ```colsub[[1L]] == ":"``` to ```colsub[[1L]] %chin% ":"```

File: R/utils.R (hunk 3 above)
Comment (MichaelChirico) [reply R/utils.R]: actually i just defined it by hand, but yes, `cc()` should accomplish the same IINM

File: R/utils.R (hunk 7 above)
Comment (ColeMiller1) [reply R/utils.R]: Now both need to be names. If either is a character, a new error is raised.  Otherwise, it evaluates in the parent frame. I largely followed Jan's work so now ```V2 -V1``` errors as well.

File: R/utils.R [hunk 14]
```diff
@@ -140,3 +140,101 @@ edit.data.table = function(name, ...) {
...
+    stop(gettextf("%s missing at the following indices: %s", mode, brackify(which(is.na(cols)))))
+  
+  if (is.logical(cols)) {
+    if ((col_len <- length(cols)) == 1L) {
+      cols = rep(cols, length.out = x_len)
+    } else if (col_len != x_len) {
+      ## TODO change to error in 2022
+      warning(gettextf("When %s is a logical vector, each column should have a TRUE or FALSE entry. The current logical vector of length %d will be repeated to length of data.table. Warning will change to error in the next verion.", mode, col_len, domain = "R-data.table"))
+      cols = rep(cols, length.out = x_len)
+    }
```
ColeMiller1: RE: recycling logical vectors. Base recycles vectors to the length of the data.frame. See ```iris[, TRUE]``` or ```iris[, c(TRUE, FALSE)]```. I guess I'm leaning towards allowing vectors repeat.

But more broadly, I am interested in some way to select all the variables. ```colsub = TRUE``` seemed like a quick and easy way. ```names(.SD)``` could work but if this function was ever applied to ```duplicated``` or ```melt``` or any other functions that allow column selection, I am not sure ```names(.SD)``` would make sense. Maybe a made up call ```all_cols()``` that we could evaluate?

R/utils.R
```diff
@@ -140,3 +140,101 @@ edit.data.table = function(name, ...) {
...
+      # each pattern gives a new filter condition, intersect the end result
+      x_names = names(x)
+      cols = Reduce(intersect, lapply(as.list(colsub)[-1L], function(col) grep(eval(col, parent.frame(2L)), x_names)))
+    } else {
+      cols = eval(colsub, parent.frame(2L), parent.frame(2L))
+    }
+  } else {
+    # if colsub is also a name in the list, we prioritize it over anything in the environment
+    if (!exists(cols <- as.character(colsub), where = x)) 
+      cols = eval(colsub, parent.frame(2L), parent.frame(2L))
```
ColeMiller1: This is where ```dt[, .SD, .SDcols = V1]``` would allow a variable a variable within the ```dt``` to be picked ahead of anything in the global environment. Would it address your concern if there's a new warning to indicate that there's also a variable with the same name in both the ```dt``` and global environment? To me, if there's a variable within the ```dt```, it should be expected behavior that the ```dt``` variable is prioritized.

jangorecki: Using unquoted column name in SDcols is not really necessary there. As well it is not really required for consistency as we do not allow list (v1, v2) syntax, as we do in j arg. Only unquoted form of v1:v2 should be there IMO.

R/utils.R
jangorecki: Your use of colsub=TRUE can be achieved by not passing SDcols, or passing names, or rep(TRUE, ncol(DT)) or seq_along. Most of which requires you to refer to DT, which is not chaining friendly. You can always use a function (...) TRUE to workaround this limitation.
In the issue related to logical vector in SDcols we discussed recycling and so far conclusion whas to not recycle.

ColeMiller1:
I think everything has been addressed. Rolled back the two semi-big changes where ```.SDcols = TRUE``` would return all columns and ```.SDcols = V1``` where V1 was a variable name within ```dt``` would return the ```V1``` column. 

I did not include ```{``` brackets. ```(({(1:3)}))``` will currently work and ```((((V1:V3)))``` will currently work. I am unsure of the use case of ```({V1:V3})``` but if there is still desire, it would be easy to add. 

For additional scrutiny, please take a second look at the warnings / stops. 

Thanks for all the comments and time you two spent reviewing.

ColeMiller1:
Also, I am more than happy to change this to C. Based on my initial benchmarks, Jan's C method is generally faster - I have a ```"negate"``` attribute while my timings for Jan's include ```:::```. I think the R could be refactored to improve performance (e.g., if ```V1:V3``` matches, there's no need to do additional checks) but then I would have to repeat similar code multiple times.

Similarly, I could start work on the ```by``` or ```j``` aspects but that would increase the reach of this PR further than ideal. After merging, I'd start work on ```by``` as I think it's a better candidate than ```j```.

```r
# remotes::install_github("Rdatatable/data.table", ref = "colselect")
library(data.table)

x = as.data.table(lapply(1:5, c))

e = quote(1:3)
col_helper(x, e, ".SDcols"),
data.table:::exprCols(x, 1:3, ".SDcols", TRUE, environment())
)
Unit: microseconds
                                                          expr  min    lq   mean median    uq   max neval
                                   col_helper(x, e, ".SDcols") 17.1 18.10 20.966  19.80 20.40 104.9   100
 data.table:::exprCols(x, 1:3, ".SDcols", TRUE, environment()) 16.5 17.45 19.264  19.05 19.85  56.6   100

e = quote(V1:V3)
  col_helper(x, e, ".SDcols"),
  data.table:::exprCols(x, V1:V3, ".SDcols", TRUE, environment())
)
Unit: microseconds
                                                            expr  min   lq   mean median   uq   max neval
                                     col_helper(x, e, ".SDcols") 21.4 22.4 25.375   23.7 24.3 104.9   100
 data.table:::exprCols(x, V1:V3, ".SDcols", TRUE, environment()) 17.2 18.1 20.082   20.4 20.9  56.2   100

cols = c("V1", "V2", "V3")
e = quote(cols) 
  col_helper(x, e, ".SDcols"),
  data.table:::exprCols(x, cols, ".SDcols", TRUE, environment())
)
Unit: microseconds
                                                           expr  min   lq   mean median   uq  max neval
                                    col_helper(x, e, ".SDcols") 11.1 12.0 13.740  13.20 13.9 44.7   100
 data.table:::exprCols(x, cols, ".SDcols", TRUE, environment()) 17.5 18.7 21.313  19.35 20.2 92.0   100
```

jangorecki:
We need to run revdeps check against this branch. What could also be useful is to give an option to escape it, so users who might got affected by this change can easily turn it off.

ColeMiller1:
Just saw the edit. I will work on including ```by```. The only way I would start work on ```j``` is if you would be cool with a ```select.j(...)``` helper function in ```j```. Otherwise, keeping a global option for ```j``` wouldn't really get us towards a modular data.table and would only increase the complexity of the code.

jangorecki:
Of course the option would be only for a while, to ensure users are not affected.

ColeMiller1:
The strangest part of ```by``` is that it allows vectors in the ```parent.frame``` to be used for grouping. Even stranger is that we also allow these vectors to be names or even arguments in lists. 

While I don't mind having a slow deprecation of names being in the parent.frame, it would make this PR easier if we could break the use of names in list referring to variables out of frame. Otherwise, it is a lot of checks for a use case that should be discouraged - we should have never let out-of-frame variables be in the by!

```
library(data.table)
n = 5L
dt = data.table(V1 = rnorm(n))
set.seed(1L)
out_var = sample(n, n, replace = TRUE)

dt[, sum(V1), by = out_var] ##slow deprecation
dt[, sum(V1), by = list(out_var)] ##break

**Comment by ColeMiller1** [comment]
The silence is deafening for breaking ```by = list(parent_frame_variable)``` :). I have fixed it but I still plan on trying to start a slow deprecation process. 

I assume changing errors / warnings is OK. I also assume new behavior is OK (e.g., ```by = is.factor```). Is new naming convention OK assuming reverse dependency is fine?

```
DT = data.table(a = 1:10)
DT[ , b := 10:1]
## test 1984.04
## current:
data.table(expression = c(1, 0), V1 = c(6, 5))
   expression    V1
        <num> <num>
1:          1     6
2:          0     5

## proposed:
DT[ , mean(b), by = eval(expression(a %% 2))]
##       a    V1
##   <num> <num>
##1:     1     6
##2:     0     5
```

jangorecki:
New behaviour like providing function to by better to be avoided. If no one requested that then there is not really a need (yet) to have it, and maintain it. Changed messages are fine. Changed behaviour is fine as long as there is an issue for that, where there is an agreement on the change. Changed behaviour that is not really a fix, but change to API has to be optional, like list(parent_scope_var). Then affected users can migrate the code more easily.

ColeMiller1:
This PR is complete for ```.SDcols```. My goal towards a modular [data.table was to introduce consistency for column selection and reuse code where possible. I am closing as that goal does not seem possible.

PR #4252, opened 2020-02-20
Writes and reads attributes to and from yaml

eliocamp:
This is to address #3540. With this, factor levels are correctly saved and retrieved, as well as other column attributes.

The only issue I'm coming up upon is that using `fread(..., yaml = FALSE)` now produces an error. The issue seems to be in the internal C++ code because the malformed table comes straight from the ` ans <- .Call(CfreadR,...)` part of the code. 

``` r
library(data.table)
file <- tempfile()
data = data.table(x = 1, y = 100, 
                  c = factor(letters[1:3], levels = letters[1:4]))

fwrite(data, file, yaml = TRUE)
x = fread(file, yaml = FALSE)
#> Warning in fread(file, yaml = FALSE): Detected 1 column names but the data has
#> 2 columns (i.e. invalid file). Added 1 extra default column name for the first
#> column which is guessed to be row names or an index. Use setnames() afterwards
#> if this guess is not correct, or fix the file write command that created the
#> file to create a valid file.
#> Warning in fread(file, yaml = FALSE): Stopped early on line 20. Expected 2
#> fields but found 1. Consider fill=TRUE and comment.char=. First discarded non-
#> empty line: <<sep2:>>
```

It does work correctly with `yaml = TRUE`, though

``` r
library(data.table)
file <- tempfile()
data = data.table(x = 1, y = 100, 
                  c = factor(letters[1:3], levels = letters[1:4]))
attr(data$x, "attr") <- "attribute"

fwrite(data, file, yaml = TRUE)
x = fread(file, yaml = TRUE)

levels(x$c)
#> [1] "a" "b" "c" "d"
attr(x$x, "attr")
#> [1] "attribute"
```

<sup>Created on 2020-02-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/.dev/.Rprofile
+++ b/.dev/.Rprofile
@@ -1,0 +1,14 @@
+# Matt's ~/.Rprofile is a link to this file at ~/GitHub/data.table/.dev/.Rprofile
+
+# options(repos = c(CRAN="http://cran.stat.ucla.edu"))
+# options(repos = c(CRAN=c("http://cran.stat.ucla.edu", "http://cloud.r-project.org")))  # both needed for revdep checks sometimes
+options(repos = c(CRAN="http://cloud.r-project.org"))
+
+options(help_type="html")
+options(error=quote(dump.frames()))
+options(width=200)
+options(digits.secs=3)  # for POSIXct to print milliseconds
+suppressWarnings(RNGversion("3.5.0"))  # so when I create tests in dev there isn't a mismatch when run by cc()
+
+Sys.setenv(PROJ_PATH=path.expand("~/GitHub/data.table"))
+source(paste0(Sys.getenv("PROJ_PATH"),"/.dev/cc.R"))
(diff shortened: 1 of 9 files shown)


MichaelChirico:
Converting to draft instead of `label:WIP`

MichaelChirico:
We haven't seen much interest in csvy TBH, I will leave it as KTLO and possibly eventually deprecate it. My sense is that users that actually require having schema in their files are just using a binary format that builds in this support more naturally.

Here's the references to `csvy` on CRAN as of now:

https://github.com/search?q=org%3Acran+lang%3AR+%2Ffread%5C%28%2F+%2Fyaml%5Cs*%3D%2F+-repo%3Acran%2Fdata.table&type=code

NB {mindicador} was removed from CRAN 4 years ago https://cran.r-project.org/web/packages/mindicador/index.html

PR #4256, opened 2020-02-23
small readability fix to GL CI

MichaelChirico:
IINM `toString` shouldn't be necessary since `l` comes from `readLines` and `shQuote` accepts `NULL` and `character()`


jangorecki:
superseeded by #4431

Issue #2590, opened 2018-01-26
j expression on an empty data.table sometimes produce output


ethanbsmith:
My understanding is that the j expression is evaluated for each row in the data table. Therefore, I expect an empty data.table to produce no output. In the 3 cases below, I don't understand why case 2 produces any output or why it produces warnings. I also don't think that test 2 and 3 producing different output could be the desired/intended result

    a <- data.table(foo = 0, bar = 1)
    b <- a[1==2] #empty table
    print("Test 1")
    b[, .(foo, bar)] # expected output
    print("Test 2")
    b[, .(foo, bar, baz = "Hello")] #Unexpected output
    print("Test 3")
    b[1==2, .(foo, bar, baz = "Hello")] #expected output

Output:

    [1] "Test 1"
    Empty data.table (0 rows) of 2 cols: foo,bar
    [1] "Test 2"
         baz
    1: Hello
    Warning messages:
    1: In as.data.table.list(jval) :
      Item 1 is of size 0 but maximum size is 1, therefore recycled with 'NA'
    2: In as.data.table.list(jval) :
      Item 2 is of size 0 but maximum size is 1, therefore recycled with 'NA'
    [1] "Test 3"
    Empty data.table (0 rows) of 3 cols: foo,bar,baz

```
> sessionInfo()
R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 16299)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:
[1] parallel  stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] XML_3.98-1.9         doParallel_1.0.10    iterators_1.0.8      foreach_1.4.4        jsonlite_1.4         curl_2.6             data.table_1.10.5    quantmod_0.4-12      TTR_0.23-2           xts_0.10-0          
[11] zoo_1.8-0            RODBC_1.3-15         plotrix_3.6-6        RevoUtilsMath_10.0.0 RevoUtils_10.0.5     RevoMods_11.0.0      MicrosoftML_1.5.0    mrsdeploy_1.1.2      RevoScaleR_9.2.1     lattice_0.20-35     
[21] rpart_4.1-11         checkpoint_0.4.0    

[...]

PR #4262, opened 2020-02-27
Should never recycle the zero-length vector

### LINKED ISSUE #3727: data.table(double(0), 1) returns a zero-row table
One of pattern of using data.table is to return a data.table at the end of a function, like:

```r
data.table(
  A = xxx,
  B = yyy,
  Z = zzz,
)
```

I often need to include some constant columns, like the datetime, the id, etc. Most of time it returns what I want except when all other inputs are zero length vector. For example:

``` r
data.table::data.table(
  a = double(),
  b = 1
)
#> Warning in as.data.table.list(x, keep.rownames = keep.rownames, check.names
#> = check.names): Item 1 has 0 rows but longest item has 1; filled with NA
#>     a b
#> 1: NA 1
```

So I have to explicitly use `rep()` everytime in order to ensure it returns the desired result. 

``` r
a <- double()
n <- length(a)
data.table::data.table(
  a = a,
  b = rep(a, n)
)
#> Empty data.table (0 rows and 2 cols): a,b
```

It's verbose, inconvenient and often being forgotten in the first place because in my mind, the length 1 column should always be recycled... so it's actually a little bit inconsistent, IMO:

```r
### gets a data.table with row 2
data.table::data.table(
  a = double(2),
  b = 1
)
### gets a data.table with row 3
data.table::data.table(
  a = double(3),
  b = 1
)
### Should get a zero-row data.table but gets a 1-row table with a warning
data.table::data.table(
  a = double(0),
  b = 1
)
```

while tibble gets what I expect:

``` r
tibble::tibble(
  a = double(0),
  b = 1
)
#> # A tibble: 0 x 2
#> # ... with 2 variables: a <dbl>, b <dbl>
```

shrektan:
Closes #3727


jangorecki:
why only scalar and not `data.table(A=double(),B=c(1.0,2.0))`?

shrektan:
> why only scalar and not `data.table(A=double(),B=c(1.0,2.0))`?

Good question. The answer is that it should be "undefined behavior" since we don't know what the user is expecting. In contrast, recycling the scalar is almost what users really want.

Both `tibble()` and `data.frame()` throw errors for such cases while at present `data.table` fills NA with a warning. As it's "undefined behavior", I just let it be for now.

shrektan:
@mattdowle It would be great if this patch can get your approval first, as I'm developping an internal package where DT[J('a', x), ...] is used a lot while `x` could possible be a zero-length date... The current behavior forces me to handle the length-0 `x` case specially, literally everywhere... If you agree with this patch's behavior, I can ensure I'm good to depend on this patch, which will save my lots of efforts...

Thanks a lot.

shrektan:
> why only scalar and not `data.table(A=double(),B=c(1.0,2.0))`?

@jangorecki Now I agree with you that, we should make `data.table(A=double(),B=c(1.0,2.0))` to be zero output as well. Will update the PR soon.

jangorecki:
1967.35 and 2139.1 are currently failing in this PR, it is quite likely that some revdeps might be failing as well, blocking release process. **IF** the change is going to be incorporated, it needs to be made in a less breaking manner. Maybe hold on with farther changes till we all agree on behavior.
I am having the very same problem in #4370. I need to recycle there as well, but my preference is to recycle 0 and 1 rows data.table to bigger one, quite opposite what you are trying to push here.
Good references for you can be `data.frame` function, and for me `cbind`, they both seems to be consistent on recycling, 0 rows throws error.
```r
d1 = data.frame(f=1L, g=1L)
d2 = data.frame(f=1:2, g=1:2)

data.frame(d2, d1)
#  f g f.1 g.1
#1 1 1   1   1
#2 2 2   1   1

cbind(d2, d1)
#  f g f g
#1 1 1 1 1
#2 2 2 1 1
```

shrektan:
The problem is: 

Is recyling the zero-length element meaningful, since there's no value at all? Reclying it with `NA` is not necessarily a reasonable one. 

That's why data.frame, cbind throws error while tibble() gives the zero-row table. 

In addition, as I've mentioned in #3727, recycling the zero-length element will trigger a warning of data.table. Someone may accidentally depend this behavior, but I doubt anyone would do it intentionally... Depsite I will take an investigation of the failed tests.

jangorecki:
Agree, you have my vote for recycling to 0 rows. It is easy to escape that by `DT[J('a', x[1L]), ...]`, the other way around is not that easy.

jangorecki:
@shrektan please fix broken tests and remove WIP label. Maybe we don't need a "less breaking manner" because that was a warning already.

R/as.data.table.R
```diff
@@ -152,6 +152,10 @@ as.data.table.list = function(x,
   ncol = sum(eachncol)  # hence removes NULL items silently (no error or warning), #842.
   if (ncol==0L) return(null.data.table())
   nrow = max(eachnrow)
+  # only check the atomic and nonNULL type because NULL will be ignored while non-atomic (e.g., list) is always recycled
+  atomic_nonnull = vapply(x, function(xi) is.atomic(xi) && !is.null(xi), TRUE)
```
MichaelChirico, line 156: we have internal helper `vapply_1b` to wrap `vapply(., ., TRUE)`

MichaelChirico:
> Maybe we don't need a "less breaking manner" because that was a warning already.

This should be mentioned in the NEWS item then, right? Something like "Previously this resulted in a warning"

shrektan:
I was busy for a while. Now the tests have been fixed. I'm going to remove the WIP label.

BTW, there's a failure on the test coverage but it looks like having little to do with PR, so I gonna ignore it.

MichaelChirico:
LGTM

jangorecki:
This email seems slightly related: https://stat.ethz.ch/pipermail/r-devel/2020-May/079572.html

mattdowle (approved):
Your examples really helped, thanks.

mattdowle:
I reran all revdeps to check the impact. After resolving unrelated issues, only one possible impact (`RstoxData`). Even that one might be something else in dev and not this change. I'll file a separate issue. 

```
> status()
Installed data.table to be tested against: 1.14.1 2021-04-28 04:39:30
CRAN:
 ERROR   :    2 : RstoxData traitdataform 
 WARNING :    1 : fwildclusterboot 
 NOTE    :  350 
 OK      :  698 
 TOTAL   : 1051 / 1051

> cran()
tools::CRAN_check_results() returned 226,625 rows in 5.584s elapsed (1.249s cpu) 
R-release is used for revdep checking so comparing to CRAN results for R-release
            Package r-release-linux-x86_64 r-release-macos-x86_64 r-release-windows-ix86+x86_64
1:        RstoxData                     OK                   NOTE                            OK
2:    traitdataform                  ERROR                     OK                         ERROR
3: fwildclusterboot                   WARN                     OK                          WARN
```

inst/tests/tests.Rraw
```diff
@@ -17346,3 +17346,14 @@ test(2170.2, DT[A > -1, which = NA], 1L)
 test(2170.3, DT[A > -1 | is.na(A), which = NA], integer())
 test(2170.4, DT[A > 10, which = NA], seq_len(nrow(DT)))
 test(2170.5, DT[!(A > 1), which = NA], c(1:3,6L)) # matches DT[A <= 1, which = NA]
+
+# data.table() zero-nrow result if any non-null & atomic element is length 0, #3727
+test(2171.1, data.table(A=double(), B=1:2), data.table(A=double(), B=integer()))
```
jan-glx, line 17351: This seems to cover #2590

PR #4264, opened 2020-02-27
fcase support scalar condition, vectorize default and lazy-eval default

### LINKED ISSUE #4258: Shouldn't fcase() recycle?
Would be nice if either of these would work

```R
fcase(
  iris$Sepal.Length > 5, ">5",
  iris$Sepal.Length < 4, "<4",
  TRUE, as.character(iris$Sepal.Length)
)
# Results in an error


fcase(
  iris$Sepal.Length > 5, ">5",
  iris$Sepal.Length < 4, "<4",
  default = as.character(iris$Sepal.Length) # Length of 'default' must be 1.
)
```

My current solution (maybe I am missing some neater solution):
```R
fcase(
  iris$Sepal.Length > 5, ">5",
  iris$Sepal.Length < 4, "<4",
  rep(TRUE, nrow(iris)), as.character(iris$Sepal.Length)
)

dplyr::case_when(
  iris$Sepal.Length > 5 ~ ">5",
  iris$Sepal.Length < 4 ~ "<4",
  TRUE ~  as.character(iris$Sepal.Length)
)


```

shrektan:
Closes #4258 

Three features:

1. Scalar condition
1. Vectorized default value
1. Lazy-eval default value

In short, the `default` should be regarded as just a special condition-value pair argument. So it should be vectorized and lazily evaluated. 

I believe with the three features the `fcase()` interface will have a closer-natural-fits to what our users expect.


jangorecki:
would be nice to measure overhead as I see some of the constant values declaration to be moved into loops.

shrektan:
The impact should be minimal when comparing to the scalar default. 

```r
library(data.table)
set.seed(100)
n <- 1e5
cond1 <- sample(c(T, F, NA), n, replace = TRUE)
cond2 <- sample(c(T, F, NA), n, replace = TRUE)
val1 <- rnorm(n)
val2 <- rnorm(n)
default <- rnorm(n)
val1_str <- as.character(val1)
val2_str <- as.character(val2)
val1_list <- as.list(val1)
val2_list <- as.list(val2)
  fcase(cond1, val1, cond2, val2, default = 0.0),
  fcase(cond1, val1_str, cond2, val2_str, default = 'abc'),
  fcase(cond1, val1_list, cond2, val2_list, default = list(1:2)),
  times = 1e3
)
```

### Before the PR

```
Unit: microseconds
                                                           expr      min       lq     mean   median
                   fcase(cond1, val1, cond2, val2, default = 0)  635.247 1048.592 1283.355 1091.572
       fcase(cond1, val1_str, cond2, val2_str, default = "abc") 1518.133 1946.129 2913.089 2005.545
 fcase(cond1, val1_list, cond2, val2_list, default = list(1:2)) 1386.652 1761.742 2158.458 1816.270
       uq       max neval
 1144.882  59054.09  1000
 2136.983 235036.02  1000
 1913.051  77624.62  1000
```

### After the PR

```
Unit: microseconds
                                                           expr      min       lq     mean   median
                   fcase(cond1, val1, cond2, val2, default = 0)  797.861 1204.731 1833.934 1240.831
       fcase(cond1, val1_str, cond2, val2_str, default = "abc") 1706.479 2115.319 2781.135 2175.959
 fcase(cond1, val1_list, cond2, val2_list, default = list(1:2)) 1512.438 1856.679 2120.990 1904.642
       uq      max neval
 1297.599 141721.1  1000
 2299.117 236917.5  1000
 1974.804  56486.5  1000
```

src/fifelse.c
```diff
@@ -257,13 +257,13 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
...
       const int *restrict pouts = INTEGER(outs);
       int *restrict pans = INTEGER(ans);
-      const int pna = nonna ? INTEGER(na)[0] : NA_INTEGER;
       for (int64_t j=0; j<len2; ++j) {
         idx = imask ? j : p[j];
         if (pcons[idx]==1) {
           pans[idx] = pouts[idx & amask];
         } else {
           if (imask) {
+            const int pna = nonna ? INTEGER(na)[scalarna?0:j] : NA_INTEGER;
```
MichaelChirico: we should avoid repeatedly calling `INTEGER` inside the loop by assigning the `int *` pointer to `INTEGER(na)` before the loop starts

MichaelChirico:
Looks good for recycling `default`!

But the issue is a bit more general -- this still fails:

```
fcase(
  iris$Sepal.Length > 5, ">5",
  iris$Sepal.Length < 4, "<4",
  TRUE, as.character(iris$Sepal.Length)
)
```

shrektan:
@2005m 

> I would have written it differently by directly duplicating the default vector to `ans`. I can post my suggestion if you want. 

You mean just assign the default value to `ans` first? It looks a better choice. 

> Also you want to replace that `%d` by `%"PRId64"` on line 183 as it can create warning with the new rtools. 

Ok. I'm not aware of the reason. Because it's a long type?

shrektan:
> Looks good for recycling `default`!
> 
> But the issue is a bit more general -- this still fails:
> 
> ```
> fcase(
>   iris$Sepal.Length > 5, ">5",
>   iris$Sepal.Length < 4, "<4",
>   TRUE, as.character(iris$Sepal.Length)
> )
> ```

Not sure if it's desired since we've already had the `default` argument. But for the `dplyr`-syntax-familiar users, having this feature seems good.

2005m:
@shrektan , yes to your first question. I can send you the code tonight or commit something if you want. I just need to write it but it is easy. Regarding your second question, yes it is for long type. Look at line 43 how it is done for fifelse.

2005m:
Test that and let me know if any issue (I did not check it)

```C
SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
  int n=length(args);
  if (n % 2) {
    error(_("Received %d inputs; please supply an even number of arguments in ..., "
            "consisting of logical condition, resulting value pairs (in that order). "
            "Note that the default argument must be named explicitly, e.g., default=0"), n);
  }
  int nprotect = 0, l = 0;
  int64_t len0=0, len1=0, len2=0, idx=0, lena = xlength(na);
  SEXP ans = R_NilValue, value0 = R_NilValue, tracker = R_NilValue, cons = R_NilValue, outs = R_NilValue;
  PROTECT_INDEX Icons, Iouts;
  PROTECT_WITH_INDEX(cons, &Icons); nprotect++;
  PROTECT_WITH_INDEX(outs, &Iouts); nprotect++;
  SEXPTYPE type0;
  bool nonna = !isNull(na), imask = true, amask = false;
  int *restrict p = NULL;
  n = n/2;
  for (int i=0; i<n; ++i) {
    REPROTECT(cons = eval(SEXPPTR_RO(args)[2*i], rho), Icons);
    REPROTECT(outs = eval(SEXPPTR_RO(args)[2*i+1], rho), Iouts);
    if (isS4(outs) && !INHERITS(outs, char_nanotime)) {
      error(_("S4 class objects (except nanotime) are not supported. Please see https://github.com/Rdatatable/data.table/issues/4131."));
    }
    if (!isLogical(cons)) {
      error(_("Argument #%d must be logical."), 2*i+1);
    }
    const int *restrict pcons = LOGICAL(cons);
    if (i == 0) {
      len0 = xlength(cons);
      len2 = len0;
      type0 = TYPEOF(outs);
      value0 = outs;
      amask = lena != len0;
      if (nonna) {
        if (lena != 1 && amask) {
          error(_("Length of 'default' must be 1 or %"PRId64"."), len0);
        }
        SEXPTYPE tn = TYPEOF(na);
        if (tn == LGLSXP && LOGICAL(na)[0]==NA_LOGICAL && lena==1) {
          nonna = false;
        } else {
          if (tn != type0) {
            error(_("Resulting value is of type %s but 'default' is of type %s. "
                     "Please make sure that both arguments have the same type."), type2char(type0), type2char(tn));
          }
          if (!R_compute_identical(PROTECT(getAttrib(outs,R_ClassSymbol)), PROTECT(getAttrib(na,R_ClassSymbol)), 0)) {
            error(_("Resulting value has different class than 'default'. "
                     "Please make sure that both arguments have the same class."));
          }
          UNPROTECT(2);
          if (isFactor(outs)) {
            if (!R_compute_identical(PROTECT(getAttrib(outs,R_LevelsSymbol)), PROTECT(getAttrib(na,R_LevelsSymbol)), 0)) {
              error(_("Resulting value and 'default' are both type factor but their levels are different."));
            }
            UNPROTECT(2);
          }
        }
      }
      if (lena == len0 && nonna) {
        ans = PROTECT(duplicate(na)); nprotect++;
      } else {
        ans = PROTECT(allocVector(type0, len0)); nprotect++;
      }
      copyMostAttrib(outs, ans);
      tracker = PROTECT(allocVector(INTSXP, len0)); nprotect++;
      p = INTEGER(tracker);     
    } else {
      imask = false;
      l = 0;
      if (xlength(cons) != len0) {
        error(_("Argument #%d has a different length than argument #1. "
                 "Please make sure all logical conditions have the same length."),
                 i*2+1);
      }
      if (TYPEOF(outs) != type0) {
        error(_("Argument #%d is of type %s, however argument #2 is of type %s. "
                 "Please make sure all output values have the same type."),
                 i*2+2, type2char(TYPEOF(outs)), type2char(type0));
      }
      if (!R_compute_identical(PROTECT(getAttrib(value0,R_ClassSymbol)),  PROTECT(getAttrib(outs,R_ClassSymbol)), 0)) {
        error(_("Argument #%d has different class than argument #2, "
                 "Please make sure all output values have the same class."), i*2+2);
      }
      UNPROTECT(2);
      if (isFactor(value0)) {
        if (!R_compute_identical(PROTECT(getAttrib(value0,R_LevelsSymbol)),  PROTECT(getAttrib(outs,R_LevelsSymbol)), 0)) {
          error(_("Argument #2 and argument #%d are both factor but their levels are different."), i*2+2);
        }
        UNPROTECT(2);
      }
    }
    len1 = xlength(outs);
    if (len1 != len0 && len1 != 1) {
      error(_("Length of output value #%d must either be 1 or length of logical condition."), i*2+2);
    }
    int64_t amask = len1>1 ? INT64_MAX : 0;
    switch(TYPEOF(outs)) {
    case LGLSXP: {
      const int *restrict pouts = LOGICAL(outs);
      int *restrict pans = LOGICAL(ans);
      const int pna = nonna ? LOGICAL(na)[0] : NA_LOGICAL;
      for (int64_t j=0; j<len2; ++j) {
        idx = imask ? j : p[j];
        if (pcons[idx]==1) {
          pans[idx] = pouts[idx & amask];
        } else {
          if (imask && amask) {
            pans[j] = pna;
          }
          p[l++] = idx;
        }
      }
    } break;
    case INTSXP: {
      const int *restrict pouts = INTEGER(outs);
      int *restrict pans = INTEGER(ans);
      const int pna = nonna ? INTEGER(na)[0] : NA_INTEGER;
      for (int64_t j=0; j<len2; ++j) {
        idx = imask ? j : p[j];
        if (pcons[idx]==1) {
          pans[idx] = pouts[idx & amask];
        } else {
          if (imask && amask) {
            pans[j] = pna;
          }
          p[l++] = idx;
        }
      }
    } break;
    case REALSXP: {
      const double *restrict pouts = REAL(outs);
      double *restrict pans = REAL(ans);
      const double na_double = Rinherits(outs, char_integer64) ? NA_INT64_D : NA_REAL;
      const double pna = nonna ? REAL(na)[0] : na_double;
      for (int64_t j=0; j<len2; ++j) {
        idx = imask ? j : p[j];
        if (pcons[idx]==1) {
          pans[idx] = pouts[idx & amask];
        } else {
          if (imask && amask) {
            pans[j] = pna;
          }
          p[l++] = idx;
        }
      }
    } break;
    case CPLXSXP: {
      const Rcomplex *restrict pouts = COMPLEX(outs);
      Rcomplex *restrict pans = COMPLEX(ans);
      const Rcomplex pna = nonna ? COMPLEX(na)[0] : NA_CPLX;
      for (int64_t j=0; j<len2; ++j) {
        idx = imask ? j : p[j];
        if (pcons[idx]==1) {
          pans[idx] = pouts[idx & amask];
        } else {
          if (imask && amask) {
            pans[j] = pna;
          }
          p[l++] = idx;
        }
      }
    } break;
    case STRSXP: {
      const SEXP *restrict pouts = STRING_PTR(outs);
      const SEXP pna = nonna ? STRING_PTR(na)[0] : NA_STRING;
      for (int64_t j=0; j<len2; ++j) {
        idx = imask ? j : p[j];
        if (pcons[idx]==1) {
          SET_STRING_ELT(ans, idx, pouts[idx & amask]);
        } else {
          if (imask && amask) {
            SET_STRING_ELT(ans, idx, pna);
          }
          p[l++] = idx;
        }
      }
    } break;
    case VECSXP: {
      const SEXP *restrict pouts = SEXPPTR_RO(outs);
      const SEXP pna = SEXPPTR_RO(na)[0];
      for (int64_t j=0; j<len2; ++j) {
        idx = imask ? j : p[j];
        if (pcons[idx]==1) {
          SET_VECTOR_ELT(ans, idx, pouts[idx & amask]);
        } else {
          if (imask && nonna && amask) {
            SET_VECTOR_ELT(ans, idx, pna);
          }
          p[l++] = idx;
        }
      }
    } break;
    default:
      error(_("Type %s is not supported."), type2char(TYPEOF(outs)));
    }
    if (l==0) {
      break;
    }
    len2 = l;
  }
  UNPROTECT(nprotect);
  return ans;
}
```

shrektan:
@2005m Thanks. Much cleaner implementation!

Only one small issue that the new var `amask` should be renamed to `namask` because `amask` has already been used. I've fixed this when committing.

jangorecki:
Those new changes seems to invalidate benchmark. If there is any difference, does this benchmark address the most pessimistic scenario?

shrektan:
> Those new changes seems to invalidate benchmark. If there is any difference, does this benchmark address the most pessimistic scenario?

Yes. There's an overhead of always copy the whole vector of the `default` arg. 

In addition, after adding the supports for the scalar condition, I just realized the `default` argument may not be really necessary. I think **we should just convert the R function's default into a `condition = TRUE, value = default`**. The arguments are:

- the default will also be lazy evaluated
- we can avoid the cost of evaluation and copying of the default vector

How do you think?

shrektan:
DONE. The one thing I'm not satisfied with is kind of the lines below. I have to declare a pointer w/o initialized it. The code is OK but if there's a better way to avoid this, I'm happy to make the change.

```C
const int *restrict pouts;
if (!naout) pouts = LOGICAL(outs); // the content is not useful if out is NA_LOGICAL scalar
```

2005m:
I'll be honest I don't really like this scalar `TRUE` condition. I find confusing and not clear. I prefer `default`.  I am not sure how close we want to be to `dplyr`.

shrektan:
But I can’t see any drawback for providing this, either. We can still use default.

2005m:
I think in my initial implementation we decided not to use lazy evaluation for default because it was not adding any value. The default argument is used very early in the process when the first set of condition has `FALSE` values.

shrektan:
@2005m Thanks for the previous good work on both `fifelse()` and `fcase()`!  Your code is very clean and I absolutely love that.

As of my personal experience, I find a vectorized & lazy-evaluated `default` very useful. I will use `fcase()` as a "vectorized" and "more powerful" version of `switch()`. 

- The reason for a Lazy-evaluated default: 
   As the `default` in `switch()` could be used to claim that previous conditions have covered all the cases, I hope `fcase()` could use `default` to throw errors when unexpected cases happen. 
- The reason for a vectorized default: 
   I usually have a use case that is to modify an existing column. When all conditions are missed, I would like the old value is kept as it is. A vectorized `default` will enable this easily.

Thanks for your understanding.

2005m:
No worries ;-) . I am just sharing my view that is all. As i said i can understand the need for a vectorised default arguement. The rest a bit less but it does not bother me. With regards to switch, I think a `fswitch` can be implemented in a much efficient way than `fcase`. My initial experiments were showing a x5 speed-up.

jangorecki:
Let's keep it open for some time so we can gather more feedback before making decision.

What I like
- `default` to be lazily evaluated
- simpler the C's implementation
- NA handling without the need for NA_class_

What I don't like
- scalar TRUE as else condition
- extra overhead related to eval and copy of `default` arg

Is it possible to just use the last argument of `...` as a default, when the length of `...` is odd? Then we don't need to provide scalar `TRUE` or name the `default`. Although it may be not verbose enough to encode 3 types of arguments inside `...`: conditions, values, and default.

shrektan:
There should be no “extra overhead related to eval and copy of default arg” now, I’m afraid, since it’s evaluated only once. I’m ok and it’s possible to remove the scalar TRUE condition, which I admit it’s not very useful and quite odd. I like the idea to just let the last odd arg be the default, if default is not explicitly provided.

MichaelChirico:
@shrektan would you like to update this to `master` for review targeting the next release?

MichaelChirico:
I think I botched the merge :see_no_evil:

MichaelChirico:
Merge is less botched, but not 100% sure it's faithfully merged...

MichaelChirico:
OK I think we've got the merge to `master` part done here at least.

> Is it possible to just use the last argument of ... as a default, when the length of ... is odd? Then we don't need to provide scalar TRUE or name the default.

I am not a fan of this TBH, requiring `default=` seems best way to encourage readable code.

```r
fcase(A, B, C, D, E, F, G, H, I, J)
```

Can you tell quickly whether a default was supplied here?

Great coding style can help get around this, but it's a bad idea IMO to leave the door open to write confusing code.

Supplying `default=` can be immediately understood & requires 4 extra keystrokes with tab completion.

> scalar TRUE as else condition

AIUI in current implementation, that's just an implementation detail, right? `default=` in R call is translated to `TRUE, default` when passing to C. That doesn't seem so bad, am I missing something?

cc @shrektan again in case you want to drop in again now that we're merged to `master` :)

src/fifelse.c
```diff
@@ -201,105 +201,104 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
-        UNPROTECT(2);
-      }
-    }
-  }
-
+  int nprotect = 0, l = 0;
+  int64_t len0=0, len1=0, len2=0, idx=0;
+  SEXP ans = R_NilValue, value0 = R_NilValue, tracker = R_NilValue, whens = R_NilValue, thens = R_NilValue;
+  PROTECT_INDEX Iwhens, Ithens;
+  PROTECT_WITH_INDEX(whens, &Iwhens); nprotect++;
```
HughParsonage, line 215: I don't understand why `PROTECT_WITH_INDEX` is being used. Could a comment be added explaining what these variables are and why this protection method was chosen?

src/fifelse.c
```diff
@@ -201,105 +201,104 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+  int *restrict p = NULL;
   const int n = narg/2;
   for (int i=0; i<n; ++i) {
-    SEXP cons = PROTECT(i==0 ? cons0 : eval(SEXPPTR_RO(args)[2*i], rho)); // protect cons0 again for easy unprotect at the end of this loop
-    SEXP outs = PROTECT(i==0 ? value0 : eval(SEXPPTR_RO(args)[2*i+1], rho));
-    if (isS4(outs) && !INHERITS(outs, char_nanotime)) {
+    idefault = i == (n - 1); // mark if the current eval is the `default` on R side
+    REPROTECT(whens = eval(SEXPPTR_RO(args)[2*i], rho), Iwhens);
+    REPROTECT(thens = eval(SEXPPTR_RO(args)[2*i+1], rho), Ithens);
+    if (isS4(thens) && !INHERITS(thens, char_nanotime)) {
```
HughParsonage, line 226: Is PROTECTion required in this branch (before erroring)?

src/fifelse.c
```diff
@@ -358,12 +360,13 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
       }
     } break;
     case STRSXP: {
-      const SEXP *restrict pouts = STRING_PTR(outs);
-      const SEXP pna = nonna ? STRING_PTR(na)[0] : NA_STRING;
+      const SEXP *restrict pthens;
```
HughParsonage, line 363: Is `restrict` legal here? I'm honestly not sure. I'm 70% sure it's fine but would like a third pair of eyes. My concern is to do with the global string pool.

src/fifelse.c
```diff
@@ -201,105 +201,104 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+      const int *restrict pthens;
+      if (!naout) pthens = LOGICAL(thens); // the content is not useful if out is NA_LOGICAL scalar
       int *restrict pans = LOGICAL(ans);
-      const int pna = nonna ? LOGICAL(na)[0] : NA_LOGICAL;
+      const int pna = NA_LOGICAL;
       for (int64_t j=0; j<len2; ++j) {
         const int64_t idx = imask ? j : p[j];
-        if (pcons[idx]==1) {
-          pans[idx] = pouts[idx & amask];
+        if (pwhens[idx & conmask]==1) {
```
HughParsonage, line 300: I question the decision to use `amask` `conmask` over a simple branch for the length-one cases. Was this chosen because the alternative would have been too repetitive? I won't insist on a change here as obviously it works.

src/fifelse.c
```diff
@@ -201,105 +201,104 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
-    const int *restrict pcons = LOGICAL(cons);
-    const bool imask = i==0;
-    int64_t l=0; // how many this case didn't satisfy; i.e. left for next case
-    switch(TYPEOF(outs)) {
+    int64_t amask = len1>1 ? INT64_MAX : 0, conmask = xlength(whens)>1 ? INT64_MAX : 0;
+    switch(TYPEOF(ans)) {
     case LGLSXP: {
-      const int *restrict pouts = LOGICAL(outs);
+      const int *restrict pthens;
+      if (!naout) pthens = LOGICAL(thens); // the content is not useful if out is NA_LOGICAL scalar
```
HughParsonage, line 295: (Applies to other cases too): Should the `if (naout)` branch be handled outside and before the switch statement?

src/fifelse.c
```diff
@@ -373,29 +376,27 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
...
-          if (imask && nonna) {
-            SET_VECTOR_ELT(ans, idx, pna);
-          }
           p[l++] = idx;
         }
       }
     } break;
     default:
-      error(_("Type '%s' is not supported"), type2char(TYPEOF(outs)));
+      error(_("Type '%s' is not supported."), type2char(TYPEOF(ans)));
```
HughParsonage, line 393: I think we should mention what argument the "Type" is referring to.

src/fifelse.c
```diff
@@ -201,105 +201,104 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
-    SEXP outs = PROTECT(i==0 ? value0 : eval(SEXPPTR_RO(args)[2*i+1], rho));
-    if (isS4(outs) && !INHERITS(outs, char_nanotime)) {
+    idefault = i == (n - 1); // mark if the current eval is the `default` on R side
+    REPROTECT(whens = eval(SEXPPTR_RO(args)[2*i], rho), Iwhens);
+    REPROTECT(thens = eval(SEXPPTR_RO(args)[2*i+1], rho), Ithens);
+    if (isS4(thens) && !INHERITS(thens, char_nanotime)) {
       error(_("S4 class objects (except nanotime) are not supported. Please see https://github.com/Rdatatable/data.table/issues/4131."));
     }
-    if (!isLogical(cons)) {
+    if (!isLogical(whens)) {
```
HughParsonage, line 229: We should say what type we encountered, rather than simply it was not type logical (so user can more easily identify the error).

tdhock (approved):
I like the new functionality/tests, but I don't think I can judge the implementation, as I have not yet hacked this file.

src/fifelse.c
MichaelChirico: One thing here is getting a consistent argument to make translation easier:

https://github.com/search?q=repo%3ARdatatable%2Fdata.table+%2F%22Type+%27%25s%27+is+not+supported%2F&type=code

So I would move this to a follow-up where we try and improve that message more generally, maybe as "Type '%s' is not supported in argument %s".

src/fifelse.c
MichaelChirico: The basic idea is to re-use memory as we go along, I tried naming these `when`/`then` to evoke `CASE WHEN A THEN B WHEN C THEN D ...`. So `PROTECT_INDEX` is to make the memory for `A`, `C`, `E`, ... is in the same place, same for `B`, `D`, `F`, ...

(AIUI)

I am not sure the benefits/tradeoffs here. The first commit in this branch was much simpler https://github.com/Rdatatable/data.table/pull/4264/commits/3334d0c3a137270d83c60108353c30a6b6e2e4d0

src/fifelse.c
TysonStanley: Agreed, added the typeof for the arg

src/fifelse.c
MichaelChirico: I am still noob at `PROTECT_WITH_INDEX()` so I turned to LLM for some explanations...

https://g.co/gemini/share/85cdc98bb748
https://chatgpt.com/share/a4792977-f24a-47da-95c0-da48a1ad7c4c
Claude (permalinks not supported?) Claude Pt 1 
Claude Pt 2

The main themes are pretty consistent -- the alternative involves doing `UNPROTECT(2)` and then `PROTECT()` twice again in each iteration, hence potentially lots of RAM churn that we avoid with `PROTECT_WITH_INDEX()`.

src/fifelse.c
MichaelChirico: We don't need to right? That's my read of WRE/what I'm used to elsewhere (also [on Garbage Collection](https://cran.r-project.org/doc/manuals/R-exts.html#Garbage-Collection)):

> Note that the pointer protection stack balance is restored automatically on non-local transfer of control (See [Condition handling and cleanup code](https://cran.r-project.org/doc/manuals/R-exts.html#Condition-handling-and-cleanup-code).), as if a call to UNPROTECT was invoked with the right argument.

src/fifelse.c
MichaelChirico: I had a similar thought about reorganizing the code... but probably better to explore as a follow-up.

src/fifelse.c
MichaelChirico: I also _think_ it's fine, but OTOH, I can't tell whether `STRING_PTR()` is part of the API would be dinged eventually...

src/fifelse.c
MichaelChirico: Ditto here, let's defer to follow-up in order to get this over the line. It's a great & highly-requested feature, we can improve implementation+readability iteratively.

MichaelChirico:
@ben-schwen think you can take a pass at review here?

src/fifelse.c
```diff
@@ -201,105 +201,106 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
-        if (!R_compute_identical(PROTECT(getAttrib(value0,R_LevelsSymbol)),  PROTECT(getAttrib(na,R_LevelsSymbol)), 0)) {
-          error(_("Resulting value and 'default' are both type factor but their levels are different."));
-        }
-        UNPROTECT(2);
-      }
-    }
-  }
-
+  int nprotect=0, l=0;
+  int64_t len0=0, len1=0, len2=0, idx=0;
```
ben-schwen, line 212: idx is never used below but always masekd with `const int64_t idx`
```suggestion
  int64_t len0=0, len1=0, len2=0;
```

src/fifelse.c
```diff
@@ -358,12 +362,13 @@ SEXP fcaseR(SEXP na, SEXP rho, SEXP args) {
...
+      const SEXP pna = NA_STRING;
       for (int64_t j=0; j<len2; ++j) {
         const int64_t idx = imask ? j : p[j];
-        if (pcons[idx]==1) {
-          SET_STRING_ELT(ans, idx, pouts[idx & amask]);
+        if (pwhens[idx & whenMask]==1) {
+          SET_STRING_ELT(ans, idx, naout ? pna : pthens[idx & thenMask]);
         } else {
           if (imask) {
             SET_STRING_ELT(ans, idx, pna);
```
ben-schwen, line 374: I know that for the imask case we have that `idx = j` but I would prefer we either do `pans[j] = pna` in the `if (imask)` of the `INTEGER`, `REAL`, `COMPLEX` case or `pans[idx] = pna` (just in string syntax) for the `STRING` and `VECTOR` case for consistency

MichaelChirico: makes sense. the comments in the `VECSXP` branch tell me we shouldn't expect parity in that branch b/c of how "missing" for list columns is a bit different, so I left that be.

src/fifelse.c
```diff
@@ -201,105 +201,106 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+      len2 = len0;
+      type0 = TYPEOF(thens);
+      value0 = thens;
+      ans = PROTECT(allocVector(type0, len0)); nprotect++;
+      copyMostAttrib(thens, ans);
+      tracker = PROTECT(allocVector(INTSXP, len0)); nprotect++;
+      p = INTEGER(tracker);     
+    } else {
+      imask = false;
+      l = 0;
```
ben-schwen, line 244: should be moved before `if (i==0) {} else {}` branching

ben-schwen:
Basically LGTM. However, there are many things I would want to rename/remodel for clarity purpose. Right now I find this piece of our codebase quite hard to grasp in comparison to its complexity.

ben-schwen (approved):

MichaelChirico:
> Right now, I find this piece of our codebase quite hard to grasp in comparison to its complexity.

I think the same can be said of a lot of our codebase TBH 😆

I think adding a lot of performance regression tests is the most important piece here, I always worry about accidentally introducing a slowdown when refactoring, even if our great test suite _is_ passing.

Filed #6322.

src/fifelse.c
MichaelChirico: I think you're just referring to `l=0`; It works because `l=0` is also set to the initial value.

I agree it's clearer to just always initialize `l=0` since it's needed in each iteration.

MichaelChirico:
Thanks @shrektan!!

PR #4271, opened 2020-03-01
print(col.names="none") more robust

### LINKED ISSUE #4270: col.names='none' works strangely for wide output
Filing @tdhock 's comment, https://github.com/Rdatatable/data.table/issues/1523#issuecomment-592171556, as its own issue.

In summary, if the print object is wider than `getOption('width')`, the `col.names='none'` logic fails to remove the column names register except in the first row.

MichaelChirico:
Closes #4270

Requesting Jan for the `test()` issue


inst/tests/tests.Rraw
```diff
@@ -16846,3 +16846,7 @@ A = data.table(A=c(complex(real = 1:3, imaginary=c(0, -1, 1)), NaN))
 test(2138.3, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
 A = data.table(A=as.complex(rep(NA, 5)))
 test(2138.4, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
+
+# column names shouldn't print even if wrapping, #4270
+DT = setDT(replicate(getOption('width'), 1, simplify = FALSE))
+test(2139, {print(DT, col.names='none'); TRUE}, notOutput="v")
```
MichaelChirico: @jangorecki, I had to add the awkward extra `TRUE` because of what seems to be a bug in `test`, I think related to double-`print`ing in this line:

```
out = capture.output(print(x <- suppressMessages(withCallingHandlers(tryCatch(x, error=eHandler), warning=wHandler, message=mHandler))))
```

the `tryCatch(x)` is doing `print` with the correct options, but the `print` in `test()` doesn't keep those options. Then the test fails because the inner `print()` returns `DT`, so the outer print just runs `print` again but without `col.names='none'`, so the test fails. Any idea for a fix? Maybe add `...` option to `test` compatible with `output`/`notOutput` that gets passed to `print`, then we could write the test as:

```
test(2139, DT, notOutput = 'v', col.names = 'none')
```

?

jangorecki: yes, I am aware of it... no idea about fixing

inst/tests/tests.Rraw
```diff
@@ -17596,3 +17596,7 @@ DT = data.table(a=1L)
 test(2186, DT[, if (TRUE) .(a=1L) else .(a=1L, b=2L)], DT,
      warning='j may not evaluate to the same number of columns for each group')
 
+# col.names='none' should apply when wrapping too, #4270
+DT = setDT(replicate(getOption('width'), 1, simplify = FALSE))
+test(2187, {print(DT, col.names='none'); TRUE}, notOutput="v")
```
mattdowle, line 17601: I checked this test fails correctly in dev before this PR. Good.
Even though `notOutput=` is case insensitive and the test does work, I feel that capital `"V"` would leave a little less scope for doubt since the column names in this test are lots of default `V<num>`. So I'll change that `"v"` to be `"V"`

PR #4275, opened 2020-03-03
detect new edge case in auto-naming j

### LINKED ISSUE #4274: auto-names throwing warnings+error with if/else list
## Description

My package [`spatsoc`](https://github.com/ropensci/spatsoc/) v0.1.11 was returning two warnings and one error from data.table recently. I narrowed it down to version 1.12.6 vs 1.12.8, potentially the new auto-name j (#3802). These are internal warnings so not something I'm too confident diagnosing, but I did find a simple workaround that seems to make them disappear - so that might be informative for you. 

The warnings:

`longer object length is not a multiple of shorter object length`

`Different branches of j expression produced different auto-named columns: [NA!=distance]; using the most "last" names`

and the error:

` Internal error: jvnames is length 2 but ans is 4 and bynames is 1`

## The fix
[The fix](https://github.com/ropensci/spatsoc/issues/23#issuecomment-593551920): assign the if/else lists and explicitly return it. [Here](https://github.com/ropensci/spatsoc/pull/25/commits/81a4e1542602d4238586f6245d96eb755aecb7ac) is a commit that makes that change. 

The warnings and error are present in v1.12.8 but not in 1.12.6 and disappear completely with the above fix. 

## Minimal reproducible example
This will require an install of a couple packages, as I'm not sure how to reproduce it outside of my situation but the warnings and error can be reproduced using one of `spatsoc`'s function's examples from the man. 

```r
# Install spatsoc v0.1.11
remotes::install_github('ropensci/spatsoc@v0.1.11')

# Load packages
library(data.table)
library(spatsoc)

# Read example data
DT <- fread(system.file("extdata", "DT.csv", package = "spatsoc"))

# Cast the character column to POSIXct
DT[, datetime := as.POSIXct(datetime, tz = 'UTC')]

# Temporal grouping
group_times(DT, datetime = 'datetime', threshold = '20 minutes')

# Edge list generation
# NOTE: this returns the two warnings
edge_nn(DT, id = 'ID', coords = c('X', 'Y'),
        timegroup = 'timegroup')

# Edge list generation, returning distance between nearest neighbours
# NOTE: this returns the two warnings and the error
edge_nn(DT, id = 'ID', coords = c('X', 'Y'),
        timegroup = 'timegroup', threshold = 100,
        returnDist = TRUE)
```

## Output of `sessionInfo()`

```
R version 3.6.2 (2019-12-12)
Running under: Arch Linux

Matrix products: default
BLAS:   /usr/lib/libopenblasp-r0.3.8.so
LAPACK: /usr/lib/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_CA.UTF-8      
 [2] LC_NUMERIC=C              
 [3] LC_TIME=en_CA.UTF-8       
 [4] LC_COLLATE=en_CA.UTF-8    
 [5] LC_MONETARY=en_CA.UTF-8   
 [6] LC_MESSAGES=en_CA.UTF-8   
 [7] LC_PAPER=en_CA.UTF-8      
 [8] LC_NAME=C                 
 [9] LC_ADDRESS=C              
[10] LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_CA.UTF-8
[12] LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils    
[5] datasets  methods   base     

other attached packages:
[1] spatsoc_0.1.11    data.table_1.12.8

[1] compiler_3.6.2 tools_3.6.2   
[3] packrat_0.5.0 
```

MichaelChirico:
Closes #4274 

I'm not _super_ happy w this "solution", any thoughts for improvement Jan?


inst/tests/tests.Rraw
```diff
@@ -16846,3 +16846,8 @@ A = data.table(A=c(complex(real = 1:3, imaginary=c(0, -1, 1)), NaN))
 test(2138.3, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
 A = data.table(A=as.complex(rep(NA, 5)))
 test(2138.4, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
+
+# compatible branches might seem incompatible if the condition is global: #4274
+DT = data.table(a=1L)
+test(2139, DT[ , if (TRUE) .(a=1) else .(a=1, b=2)], DT,
```
jangorecki: `else` branch is ever evaluated?

MichaelChirico: we evaluate its names, and the logic is to take the "last" names from j; in this case, it'll be the else branch.

in the test example it's obvious the second branch is spurious, but in general it's hard to tell lazily which branches will be evaluated

PR #4282, opened 2020-03-06
better error message for missing j in cube

MichaelChirico:
Ran into this today -- before this fix, the missing `j` was being passed as `jj` to `groupingsets` and we'd get an error from `all.vars` referencing a missing `jj`. Better to reference `j` since that's part of `args(cube.data.table)`.


NEWS.md
```diff
@@ -145,6 +145,8 @@ unit = "s")
 
 7. Added more explanation/examples to `?data.table` for how to use `.BY`, [#1363](https://github.com/Rdatatable/data.table/issues/1363).
 
+8. Cube catches a missing `j` argument earlier to give friendlier output.
```
jangorecki: `cube` data.table method...

PR #4288, opened 2020-03-07
change fread default na.strings from 'NA' to ''

MichaelChirico:
2 years removed from the declared intention to do so:

https://github.com/Rdatatable/data.table/commit/29e2d461942b620ee04d638e941bb965df29b87e

I made the quick change, but it breaks a lot of tests.

Some are nontrivial to fix -- looking at 880/881, it's hard to get consistency vs. `write.table`/`read.csv` -- I'll have to take a bit more time to understand how those functions distinguish `x,` and `x,''` for string columns. At first glance I didn't see a way.

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -1,6 +1,6 @@
 fread = function(
 input="", file=NULL, text=NULL, cmd=NULL, sep="auto", sep2="auto", dec=".", quote="\"", nrows=Inf, header="auto",
-na.strings=getOption("datatable.na.strings","NA"), stringsAsFactors=FALSE, verbose=getOption("datatable.verbose",FALSE),
+na.strings=getOption("datatable.na.strings",""), stringsAsFactors=FALSE, verbose=getOption("datatable.verbose",FALSE),
 skip="__auto__", select=NULL, drop=NULL, colClasses=NULL, integer64=getOption("datatable.integer64","integer64"),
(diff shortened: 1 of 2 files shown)


shrektan:
Not fully understand the whole topic. How about making the default `na.strings` be two strings: "" and "NA"?  I mean both "" and "NA" are regarded as `NA`.

SamuelAllain:
Hello,

That change would be great ! I spent quite a lot of time trying to understand why a table was altered by a fwrite-fread. My reprex :

```
library(data.table) # 1.14.0
d1 <- data.table(character = c("Text", NA), numeric = c(NA, 1))
fwrite(d1, "toDelete.csv")
d2 <- fread("toDelete.csv")
d3 <- fread("toDelete.csv", na.strings = "")
d1
d2

all.equal(d1, d2) #"Column 'character': 'is.NA' value mismatch: 0 in current 1 in target"
all.equal(d1, d3) # TRUE
file.remove("toDelete.csv")
```
I expected d1 and d2 to be equal. Thanks a lot for your work.

davidbudzynski:
Have there been any progress made on this? `fread`'s help says that it's due to change from `"NA"` to `""`, and the [NEWS section for v1.11.0 (01 May 2018)](https://rdatatable.gitlab.io/data.table/news/index.html#notice-of-intended-future-potential-breaking-changes) says it was supposed to happen in the next release. As of now, the default still is `"NA"`.

PR #4289, opened 2020-03-07
fifelse() coerces NA to other types and supports vectorized na argument

### LINKED ISSUE #4277: Allow coercion for NA input to fifelse
```
fifelse(c(TRUE, FALSE), NA, 1L)
# Error in fifelse(c(TRUE, FALSE), NA, 1L) : 
#   'yes' is of type logical but 'no' is of type integer. Please make sure that both arguments have the same type.
```

that `typeof(NA)` is `logical` I think is an implementaton detail we can obscure from users. we should coerce `NA` to `NA_integer_` in this case.

### LINKED ISSUE #4286: the NA argument in fifelse should support vector
```r
data.table::fifelse(c(T, F, NA, T, F, NA), 1:6, 11:16, 101:106)
# Error in data.table::fifelse(c(T, F, NA, T, F, NA), 1:6, 11:16, 101:106) : 
#  Length of 'na' is 6 but must be 1
```

### LINKED ISSUE #4287: the `na` argument in fifelse should be coerced
This is valid:

```r
data.table::fifelse(c(T, F, NA), 1L, 2.0, 3.0)
# [1] 1 2 3
```

But not this:

```r
data.table::fifelse(c(T, F, NA), 1L, 2L, 3.0)
# Error in data.table::fifelse(c(T, F, NA), 1L, 2L, 3) : 
#  'yes' is of type integer but 'na' is of type double. Please make sure that both arguments have the same type.
```

shrektan:
Closes #4287 
Closes #4286 
Closes #4277 

Two points:

1. `na` supports vectorized input: The `na` argument should be no differences from `yes` and `no` due to the fact that `TRUE`, `FALSE` and `NA` are all valid R's logical values
1. The logical value of `NA`will be coerced to other types automatically:
    1. The `na` argument already performs this behavior. As stated in the 1 point, `yes` and  `no` should behave the same;
    1. `data.table` does this auto-coercion in many other functions
    1. Should only have a very small performance impact


inst/tests/tests.Rraw
```diff
@@ -15704,7 +15704,7 @@ test(2072.009, fifelse(test_vec, rep(1L,11L), rep(0L,10L)),      error="Length o
 test(2072.010, fifelse(test_vec, rep(1,10L), rep(0,11L)),        error="Length of 'yes' is 10 but must be 1 or length of 'test' (11).")
 test(2072.011, fifelse(test_vec, rep(TRUE,10L), rep(FALSE,10L)), error="Length of 'yes' is 10 but must be 1 or length of 'test' (11).")
 test(2072.012, fifelse(0:1, rep(TRUE,2L), rep(FALSE,2L)),        error="Argument 'test' must be logical.")
-test(2072.013, fifelse(test_vec, TRUE, "FALSE"),                 error="'yes' is of type logical but 'no' is of type character. Please")
+test(2072.013, fifelse(test_vec, TRUE, "FALSE"),                 error="'no' is of type character but expect logical")
```
jangorecki: expect logical carry less information than saying that 'yes' is logical and types doesn't match

src/fifelse.c
```diff
@@ -11,127 +11,168 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
+    const int na = NA_INTEGER;
     #pragma omp parallel for num_threads(getDTthreads())
     for (int64_t i=0; i<len0; ++i) {
-      pans[i] = pl[i]==0 ? pb[i & bmask] : (pl[i]==1 ? pa[i & amask] : pna);
+      pans[i] = pl[i]==0 ?
+                (na_b ? na : pb[i & bmask]) :
+                pl[i]==1 ? 
+                (na_a ? na : pa[i & amask]) : 
+                (na_n ? na : pna[i & nmask]);
     }
```
jangorecki: overhead should be measured, most pessimistic case

shrektan: These are pretty fast checks and should be close-to-zero overheads and the benchmark validates this:

### CODE

```r
set.seed(100)
n = 1e6
test_num = runif(n) 
test_true = test = rep(TRUE, n)
test_na = rep(NA, n)
test[test_num <= 1/3] <- FALSE
test[test_num >= 1/3] <- NA
yes = rnorm(n)
no = rnorm(n)

  data.table::fifelse(test_true, yes, no, NA),
  data.table::fifelse(test, yes, no, NA), 
  data.table::fifelse(test_na, yes, no, NA), 
  times = 1e3
)
```

### Before the PR

image


### After the PR

image

inst/tests/tests.Rraw
shrektan: Addressed by https://github.com/Rdatatable/data.table/pull/4289/commits/2b69bc4c3dc3f284ea274c8bb9e41cc92f6a13e0

src/fifelse.c
```diff
@@ -11,127 +11,180 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
   const int64_t len0 = xlength(l);
   const int64_t len1 = xlength(a);
   const int64_t len2 = xlength(b);
+  const int64_t len3 = xlength(na);
   SEXPTYPE ta = TYPEOF(a);
   SEXPTYPE tb = TYPEOF(b);
+  SEXPTYPE tn = TYPEOF(na);
+  bool na_a = ta==LGLSXP && LOGICAL(a)[0]==NA_LOGICAL;
```
mattdowle, line 18: @shrektan Great PR. Fully agree.
Just at this point, `a` could be length-0, and if so this `LOGICAL(a)[0]` would be an out-of-bound memory access.
Maybe the length checks later in this function could be moved up to be done first?

mattdowle (changes requested):

mattdowle (approved):

PR #4297, opened 2020-03-11
should use the long int 2 to avoid integer overflow

### LINKED ISSUE #4295: Failed to allocate counts or TMP when assigning g in gforce
I'm working with a data.table of `1423657324` rows and `14` columns.

There's an integer group (`grp`) with 3790 unique values.

When I do the following

```r
stats <- ft[, .(
      col1 = median(col1, na.rm = TRUE),
      col2 = median(col2, na.rm = TRUE),
      col3 = median(col3, na.rm = TRUE)
    ), keyby = grp]
```

The following error occurs:

```
Error in gforce(thisEnv, jsub, o__, f__, len__, irows) :
  Internal error: Failed to allocate counts or TMP when assigning g in gforce
```

This does not occur when a shorter version of the data (1/3 of the size) is done with the same aggregation.

shrektan:
Closes #4295


jangorecki:
@shrektan could you please include NEWS item as well? mentioning both issues that are being resolved by this fix. Thanks

shrektan:
Sure, added.

PR #4299, opened 2020-03-11
identifying / unifying similar messages for streamlined translation

MichaelChirico:
You can see from the diff here that there are lots of messages that are very close in different parts of the source:

```
verbose must be TRUE or FALSE
# vs
'verbose' must be TRUE or FALSE
```

Both look quite similar but will create different `msgid` in the `po` template, potentially 1000s of lines apart -- this led to some duplicate efforts already. hoping to at least make the maintenance burden smaller going forward by unifying some similar messages I found.

Approach is pretty manual so far:

```
f = readLines('po/data.table.pot')
idx = grep('^msgid', f)
msg = character(length(idx))

for (ii in seq_along(idx)) {
  start = idx[ii]
  end = start
  while (!grepl('^msgstr', f[end])) end = end + 1L
  msg[ii] = paste(gsub('(?:^msgid )?"(.*)"$', '\\1', f[start:(end - 1L)]), collapse = '')
}

dist = adist(msg)
diag(dist) = NA_real_

nearest = apply(dist, 1L, min, na.rm = TRUE)

for (ii in seq_along(msg)) {
  cat(sprintf(
    'Original message:\n%s\nCandidate match (distance %d):\n%s\n\n********\n\n', 
    msg[ii], nearest[ii], 
    paste(msg[which(dist[ , ii] == nearest[ii])],
          collapse = '\n---------\n')
  ))
}
```

Any suggestions for a semantic string matcher rather than just using `adist` would be great. I'm also thinking of putting this type of function into a package `potools` for working with translations as a dev.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -121,9 +121,9 @@ static int _selfrefok(SEXP x, Rboolean checkNames, Rboolean verbose) {
     if (verbose) Rprintf(_(".internal.selfref ptr is NULL. This is expected and normal for a data.table loaded from disk. Please remember to always setDT() immediately after loading to prevent unexpected behavior. If this table was not loaded from disk or you've already run setDT(), please report to data.table issue tracker.\n"));
     return -1;
   }
-  if (!isNull(p)) error(_("Internal error: .internal.selfref ptr is not NULL or R_NilValue")); // # nocov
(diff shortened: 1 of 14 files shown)


MichaelChirico:
coverage dropped because I added a bunch of newlines before errors in src -- glad to have caught them! might leave them here and address all such cases in a separate PR

PR #4301, opened 2020-03-13
Add errors on separate lines for coverage

MichaelChirico:
```
if (FALSE) error("failed");
```

Is not caught by Codecov because it doesn't build a proper AST for C code. So we have to put `error` branches on their own line to be properly detected by Codecov.

Only done this for `error` so far. A few more patterns could be included:

 - [ ] `warning`
 - [ ] Verbose output? `Rprintf` etc, not sure how important it is to cover these
 - [ ] `STOP` in `fread.c`
 - ...

### DIFF
--- a/src/dogroups.c
+++ b/src/dogroups.c
@@ -12,13 +12,19 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
   clock_t tstart=0, tblock[10]={0}; int nblock[10]={0};
   const bool verbose = LOGICAL(verboseArg)[0]==1;
 
-  if (!isInteger(order)) error(_("Internal error: order not integer vector")); // # nocov
-  if (TYPEOF(starts) != INTSXP) error(_("Internal error: starts not integer")); // # nocov
-  if (TYPEOF(lens) != INTSXP) error(_("Internal error: lens not integer")); // # nocov
+  if (!isInteger(order))
+    error(_("Internal error: order not integer vector")); // # nocov
+  if (TYPEOF(starts) != INTSXP)
+    error(_("Internal error: starts not integer")); // # nocov
+  if (TYPEOF(lens) != INTSXP)
+    error(_("Internal error: lens not integer")); // # nocov
   // starts can now be NA (<0): if (INTEGER(starts)[0]<0 || INTEGER(lens)[0]<0) error(_("starts[1]<0 or lens[1]<0"));
-  if (!isNull(jiscols) && LENGTH(order) && !LOGICAL(on)[0]) error(_("Internal error: jiscols not NULL but o__ has length")); // # nocov
-  if (!isNull(xjiscols) && LENGTH(order) && !LOGICAL(on)[0]) error(_("Internal error: xjiscols not NULL but o__ has length")); // # nocov
-  if(!isEnvironment(env)) error(_("'env' should be an environment"));
+  if (!isNull(jiscols) && LENGTH(order) && !LOGICAL(on)[0])
+    error(_("Internal error: jiscols not NULL but o__ has length")); // # nocov
+  if (!isNull(xjiscols) && LENGTH(order) && !LOGICAL(on)[0])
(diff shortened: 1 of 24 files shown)


MichaelChirico:
wow! 62% (176/285) of these branches were uncovered. looks like it's paying off already in terms of understanding spurious coverage.

inst/tests/tests.Rraw
```diff
@@ -6312,15 +6312,13 @@ options(datatable.optimize = Inf)
 
 # fread dec=',' e.g. France
 test(1439, fread("A;B\n1;2,34\n", dec="12"), error="nchar(dec) == 1L is not TRUE")
-test(1440, fread("A;B\n8;2,34\n", dec="1"), data.table(A=8L, B="2,34"))
```
MichaelChirico: `!dec %chin% c('.', ',')` is implied incorrect by L1179 of `fread.c` @ master:

> dec='' not allowed. Should be '.' or ','

`?fread` is not as strict:

> The decimal separator as in utils::read.csv. If not "." (default) then usually ",". See details.

Took a conservative approach for now, and blocked `!dec %chin% c(',', '.')`. But that broke two tests (`1440` and `1444.2` here).

`dec='1'` and `dec='*'` ran without errors... happy to restore this to work, and change the error/message found in `fread.c`.

src/forder.c
```diff
@@ -1328,7 +1348,8 @@ SEXP binary(SEXP x)
 {
   char buffer[69];
   int j;
-  if (!isReal(x)) error(_("x must be type 'double'"));
+  if (!isReal(x))
```
MichaelChirico: This function `binary(` isn't used anywhere... just for debugging?

MichaelChirico:
Don't know how to cover this line so leaving this PR here:

```
      if (!firstJumpEnd)
        STOP(_("Single column input contains invalid quotes. Self healing only effective when ncol>1"));
```

MichaelChirico:
Heads up that there will be a ton of conflicts between this one and #4306 (I count 20 at the moment). Let's merge this one first, since this one adds a bunch of internal errors.

MichaelChirico:
@jangorecki any  thoughts on the third point -- should we put verbose messaging on its own line as well?

jangorecki:
@MichaelChirico I don't think it is so important. I would just keep that in mind when coding new stuff, but not necessarily touch all the existing one.

src/dogroups.c
```diff
@@ -77,7 +77,7 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
   // using <- in j (which is valid, useful and tested), they are repointed to the .SD cols for each group.
   SEXP names = PROTECT(getAttrib(SDall, R_NamesSymbol)); nprotect++;
   if (length(names) != length(SDall))
-    error(_("length(names)!=length(SD)"));
+    error(_("Internal erro: length(names)!=length(.SD)")); // # nocov
```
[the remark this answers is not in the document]
MichaelChirico: nice catch... might as well just leave it to #4306

src/forder.c
mattdowle: Yes

mattdowle:
What would it take to make C coverage cover branches within one line; i.e. like R coverage already does?
I don't know which tool that would be -- whether it's something inside `covr` or inside `gcc` toolchain itself, @jimhester? 
But that's the root cause. There would be much more bang-for-the-time to work on that?  That way, coverage for all R packages using C code would be improved, not just data.table.
If there is something deeply insurmountable about doing that (e.g. a `covr` or `gcc` developer stating that it's not just difficult but not possible in C) then I'd be much more comfortable splitting all our C lines given that justification I can point to. But I hope it's actually possible to get coverage within same-line C branches implemented. That would be awesome. Could even lead to a paper and conference presentation.

MichaelChirico:
Too many conflicts here. This is being handled incrementally for now. If needed, when the PR queue is emptier, we can start this from scratch & add a linter to prevent backsliding.

PR #4334, opened 2020-04-01
Fix round.IDate weeks

artemklevtsov:
Hi,

`round.IDate` for the weeks (`IDateTime.R#81`) seems inaccurate: first week of year have a 6 days instead 7 (fix: should be `yday(x) - 1L`):

```r
> rle(unclass(fun_old(as.IDate(0:21), "week")))
Run Length Encoding
  lengths: int [1:4] 6 7 7 2
  values : int [1:4] 0 7 14 21
> rle(unclass(fun_new(as.IDate(0:21), "week")))
Run Length Encoding
  lengths: int [1:4] 7 7 7 1
  values : int [1:4] 0 7 14 21
```


MichaelChirico:
I'm actually not sure I understand the intended behavior of `round.IDate`. I see @tshort implemented this long ago but it's not well-documented what the specific behavior of "rounding" is for IDate class (and we don't have a base equivalent to compare to since `round.Date` does not accept `units=`.

At a glance, it _looks_ like the intention is more like `trunc.IDate`, i.e., always round _down_ to the "start" of the week (with weeks always relative to January 1). I'm not sure that's what people would expect from a "round" method, where we might instead expect dates at the "end" of a week to "round" to the next week.

But many questions remain -- why set Jan 1 as the start, instead of the nearest Monday? Or Sunday? What about rounding to ISO weeks instead for stability across years? What do we do with Wednesdays (Sunday-based week, i.e. day 4 of 7), should it be rounded "up" or "down"?

Anyway, the fix you provide looks good _relative to the current definition of `round.IDate`_, so maybe the best move is to merge your PR and then revisit the purpose of this method in a follow-up. The current `round.IDate` is definitely buggy in the way you describe, so it should be fixed regardless of what we do with `round.IDate`.

MichaelChirico:
Looked around for some guidance on how else this is implemented, came across [`lubridate::round_date()`](https://lubridate.tidyverse.org/reference/round_date.html). They solve the problem I noted above by adding a second parameter, `week_start`, to govern, under `round(unit="week")`, to which day of the week to round. Maybe that's the most sensible approach.

MichaelChirico:
Hmm things seem to have been borked by trying to update the branch. Not sure what happened / how to fix it. Part of the issue is likely that the change was created on `master` in the fork, which makes things harder to work with even though we have write permissions. Note that it's much preferable to make work on branches other than `master`. Closing this and resuming work in #6024

PR #4340, opened 2020-04-02
setorderv gets new arg neworder, closes #4012

### LINKED ISSUE #4012: setorder[v] could take index vector too
It was requested at least twice
https://stackoverflow.com/questions/37878620/reorder-rows-in-data-table-in-a-specific-order/
https://stackoverflow.com/questions/56759013/is-there-are-version-of-setorder-that-behaves-like-setcolorder
I posted an answer in one of the questions as a new function which uses data.table internals
```r
setroworder <- function(x, neworder) {
    .Call(data.table:::Creorder, x, as.integer(neworder), PACKAGE = "data.table")
    invisible(x)
}
```
Instead of new function we could just add new argument `neworder` that would redirect to this functionality.

I just need such functionality now to randomly reorder data.
Having big DT and doing `DT[sample(.N)]` is costly because of copy. Doing `setorder(DT, neworder=sample(nrow(DT)))` would be much nicer.

jangorecki:
closes #4012

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -81,6 +81,20 @@ unit = "s")
 
 14. Added support for `round()` and `trunc()` to extend functionality of `ITime`. `round()` and `trunc()` can be used with argument units: "hours" or "minutes". Thanks to @JensPederM for the suggestion and PR.
 
+15. Function `setorderv` gets new argument `neworder` where user can specify custom ordering directly, [#4012](https://github.com/Rdatatable/data.table/issues/4012).
+
+```r
+DT = data.table(id1 = c("a","b","c","d"), v1 = rnorm(4))
+
+# move first row to the end
+setorderv(DT, neworder = c(2:4,1L))
+DT
+
+# random order
+setorderv(DT, neworder = sample(nrow(DT)))
+DT
+```
+
 ## BUG FIXES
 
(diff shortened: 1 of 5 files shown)


jangorecki:
There is an internal function that is not being used anywhere in the package.
It is basically pointing to the internals which are used in `setorderv`. Seems to be safe to remove this function.
```r
# reorder a vector based on 'order' (integer)
# to be used in fastorder instead of x[o], but in general, it's better to replace vector subsetting with this..?
# Basic checks that all items of order are in range 1:n with no NAs are now made inside Creorder.
# FOR INTERNAL USE ONLY
setreordervec = function(x, order) .Call(Creorder, x, order)
```

jangorecki:
failure of CI in this PR
```
* checking Rd cross-references ... WARNING
Missing link or links in documentation object 'frank.Rd':
  'forder'
Missing link or links in documentation object 'openmp-utils.Rd':
  'forder'
Missing link or links in documentation object 'setorder.Rd':
  'forder'
See section 'Cross-references' in the 'Writing R Extensions' manual.
```
is related to moving alias `forder` into new manual file. That will happen in #4458, so we may want to wait with merging this one till #4458 is merged.

PR #4351, opened 2020-04-03
non-ascii tests

### LINKED ISSUE #1343: Finish DtNonAsciiTests package and submit to CRAN
Investigate why the change in R-devel on 18 Sep 2015 caused 2 of data.table's non-ASCII tests (1092 & 1093) to fail on R-devel on Windows. data.table v1.9.4 on CRAN was very stable in 'OK' state for nearly a year before this fail.  Decided it's good to have non-ASCII tests in a separate package just to take a belt-and-braces approach.
See https://developer.r-project.org/blosxom.cgi/R-devel/NEWS
The change to match() is most likely I'd guess.

jangorecki:
attempt to close #1343

I ported unit tests from https://github.com/Rdatatable/DtNonAsciiTests/blob/master/tests/testthat/tests.R
unfortunately they are not passing.
The only problem seems to be that LHS of `:=` is not able to recognize utf-8 column name.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -3332,7 +3332,59 @@ DT[,`:=`(last.x=tail(x,1L),last.x1=tail(x1,1L)),by=y]
 test(1086, class(DT$last.x), c("POSIXct", "POSIXt"))
 test(1087, class(DT$last.x1), "ITime")
 
-# Tests 1088-1093 were non-ASCII. Now in DtNonAsciiTests
+# Bug fix for #5159 - chmatch and character encoding (for some reason this seems to pass the test on a mac as well)
+a = c("a","\u00E4","\u00DF","z")
+au = iconv(a,"UTF8","latin1")
+test(1164.1, chmatch(a, au), match(a, au))
+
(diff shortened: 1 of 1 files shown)


shrektan:
BTW, there's no easy way to support the non-ASCII colnames reliably, AFAIK. This is different from non-ASCII character values. The reason is R will always try to convert the symbol names to the native encoding. The symbol names I mean include the variable names inside of an environment. Since we need to eval these symbols in an environment, we can't avoid that. This is actually an R problem.

jangorecki:
Thanks for your valuable insight. If so, then it make sense to convert column names to native encoding as early as possible, possibly with warning/message.

jangorecki:
Just found this
> One related example is R symbols. They need to have a unique representation defined by a pointer stored in the R symbol table. For any effective implementation, they need to be in the same encoding, which now is the native encoding.

https://developer.r-project.org/Blog/public/2020/05/02/utf-8-support-on-windows/index.html

jangorecki:
#4708 another example

x-hey:
Maybe related to #4665

PR #4352, opened 2020-04-03
test.data.table manual says to run in clean environment

### LINKED ISSUE #3705: tests fail if z in .GlobalEnv
```
library(data.table)
z = 4
test.data.table()
```

```
Error in test.data.table() : 
  3 errors out of 9118 in 00:01:00 elapsed (00:02:40 cpu) on Thu Jul 11 08:46:42 2019. [endian==little, sizeof(long double)==16, sizeof(pointer)==8, TZ=Asia/Singapore, locale='LC_CTYPE=en_US.UTF-8;LC_NUMERIC=C;LC_TIME=en_US.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONETARY=en_US.UTF-8;LC_MESSAGES=en_US.UTF-8;LC_PAPER=en_US.UTF-8;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=en_US.UTF-8;LC_IDENTIFICATION=C', l10n_info()='MBCS=TRUE; UTF-8=TRUE; Latin-1=FALSE', getDTthreads()='omp_get_num_procs()==8; R_DATATABLE_NUM_PROCS_PERCENT=="" (default 50); R_DATATABLE_NUM_THREADS==""; omp_get_thread_limit()==2147483647; omp_get_max_threads()==8; OMP_THREAD_LIMIT==""; OMP_NUM_THREADS==""; data.table is using 4 threads. This is set on startup, and by setDTthreads(). See ?setDTthreads.; RestoreAfterFork==true']. Search tests/tests.Rraw for test numbers: 1894.09, 1894.13, 1894.14
```

```
if (exists("z")) rm(z)
test(1894.09, DT[, sum(z)*..z], error="Variable 'z' is not found in calling scope. Looking in calling scope because this symbol was prefixed with .. in the j= parameter.")
z = 3L
test(1894.10, DT[, sum(z)*..z], 72L)
setnames(DT, "z", "..z")
test(1894.11, DT[, sum(y)*..z], INT(105,120,135))
rm(z)
test(1894.12, DT[, sum(y)*..z], INT(105,120,135))
setnames(DT, "..z", "z")
test(1894.13, DT[, sum(y)*..z], error="Variable 'z' is not found in calling scope")
..z = 4L
test(1894.14, DT[, sum(y)*..z], 60L)  # 'manual' prefix (we have recommended in the past) still works, for now.
```

There's `if (exists('z')) rm(z)` but that didn't work?

jangorecki:
closes #3705

### DIFF
--- a/man/test.data.table.Rd
+++ b/man/test.data.table.Rd
@@ -17,7 +17,7 @@ test.data.table(script = "tests.Rraw", verbose = FALSE, pkg = ".",
 \item{showProgress}{ Output 'Running test <n> ...\\r' at the start of each test? }
 }
 \details{
-  Runs a series of tests. These can be used to see features and examples of usage, too. Running test.data.table will tell you the full location of the test file(s) to open.
+  Runs a series of tests. These can be used to see features and examples of usage, too. Running test.data.table will tell you the full location of the test file(s) to open. Test should be run in a clean environment.
 
   Setting \code{silent=TRUE} sets \code{showProgress=FALSE} too, via the default of \code{showProgress}.
 }


mattdowle:
What does "clean environment" mean precisely?  Should users `rm(*)`, or does it mean "fresh new R session"? If the latter, what if the very reason the user wants to run `test.data.table()` is if they want to test in that particular session; maybe they think some packages might be causing bad interaction effects with data.table, or that their R session may be corrupted or something.

I really think #3705 should be fixed directly. I thought it already ran in its own independent scope, but clearly it must still be able to reach .GlobalEnv. If there's not a way to isolate it from .GlobalEnv, then test 1894 could be made robust to this.

PR #4353, opened 2020-04-04
int i and nomatch arg

### LINKED ISSUE #3109: Respect `nomatch = 0L` for integer i
Currently `nomatch` seems to be ignored if `i` in `dt[i, j, nomatch = 0L]` is integer. Values of `i` which are out of bounds or `NA` result in rows filled with `NA`:
```r
dt = data.table(x = 1:10)
dt[c(1L, 11L, NA_integer_), x, nomatch = 0L]
> 1 NA NA
```
I see no reason to not support `nomatch = 0L` for integer `i`.

Also note that the current handling of negative integer `i` seems to be inconsistent with not-joins:
```r
# mixing negation/not-join with nomatch
dt[-1L, nomatch = 0L] # works, nomatch ignored
dt[!list(1L), on = "x", nomatch = 0L] # exception

# negation/not-join with out-of-bounds values
dt[-11L] # warning
dt[!list(11), on = "x"] # no warning
```

### LINKED ISSUE #3666: Select only rows that exist?
When selecting rows with a numeric index, is there a good way to only select rows that exist? For example, I'd like:

``` r
library(data.table)

dt <- data.table(x = 1:2)
dt[1:4, ]
#>     x
#> 1:  1
#> 2:  2
#> 3: NA
#> 4: NA
```

I need this to match the semantics of `dplyr::slice()` (https://github.com/tidyverse/dtplyr/issues/10). 

I also need to do this by-group, so I came up with this:

``` r
library(data.table)

dt <- data.table(x = 1:4, y = c(1, 2, 1, 2))
dt[dt[, na.omit(.I[1:3]), by = "y"]$V1]
#>    x y
#> 1: 1 1
#> 2: 3 1
#> 3: 2 2
#> 4: 4 2
```

The problem with this approach is that it refers to `dt` twice; this is problematic if `dt` is actually a compound expression because it will require `dt` to be evaluated twice. I could assign it to a variable, but I was hoping to find some way to do it in a single sequence of call.

jangorecki:
closes #3109
closes #3666 
also refactors code for better code coverage, thus is likely to get some failure for codcov check if any of those expanded lines were not covered before


src/subset.c
```diff
@@ -124,39 +124,62 @@ const char *check_idx(SEXP idx, int max, bool *anyNA_out, bool *orderedSubset_ou
   return NULL;
 }
 
-SEXP convertNegAndZeroIdx(SEXP idx, SEXP maxArg, SEXP allowOverMax)
+SEXP convertNegAndZeroIdx(SEXP idx, SEXP maxArg, SEXP allowOverMax, SEXP nomatchArg)
```
mattdowle, line 127: the name of the new argument (`noMatchArg`) needs knowledge of where this function is being called from to understand. A better name might be `allowNA` or `removeNA` to more directly convey what it does, iiuc.

src/subset.c
```diff
@@ -124,39 +124,62 @@ const char *check_idx(SEXP idx, int max, bool *anyNA_out, bool *orderedSubset_ou
...
 {
   // called from [.data.table to massage user input, creating a new strictly positive idx if there are any negatives or zeros
   // + more precise and helpful error messages telling user exactly where the problem is (saving user debugging time)
   // + a little more efficient than negativeSubscript in src/main/subscript.c (it's private to R so we can't call it anyway)
   // allowOverMaxArg is false when := (test 1024), otherwise true for selecting
+  // nomatchArg is true (nomatch==0L) when we want to skip nomatch'ing rows, it also excludes NAs #3109
```
mattdowle, line 133: this argument excludes NAs, iiuc, so I'm looking at 'it also' and wondering what else it does

mattdowle:
Both before and after my changes, this PR turns on `nomatch=0` as well as `nomatch=NULL`: 

```R
DT = data.table(x = 1:4)
> DT[c(1L, 5L, NA_integer_)]
       x
   <int>
1:     1
2:    NA
3:    NA
> DT[c(1L, 5L, NA_integer_), nomatch=NULL]  # new behavior NEWS item mentions
       x
   <int>
1:     1
> DT[c(1L, 5L, NA_integer_), nomatch=0]  # nomatch=0 can be used as well
       x
   <int>
1:     1
```

The new tests don't cover `nomatch=0`, but iiuc and what Jan writes, we want to only allow `nomatch=NULL` for this new feature since `nomatch=0` is only for backwards compatibility as per news item 5 in v1.12.0 (Jan 2019) :

image

So we don't want to open the door to folk starting to use `nomatch=0L` for this (which is #3109's request but Jan correctly [nudges](https://github.com/Rdatatable/data.table/issues/3109#issuecomment-430036916) towards `NULL`.)

The difficulty to overcome is that `[.data.table` converts `NULL` to `0L` near the top so that `nomatch` has two states from then on. If we're going to turn this new feature on only for `nomatch=NULL` then we might need to carry 3 states through. Or, we could put in the warning that `0L` is deprecated (with option to turn off) as that NEWS item warned, and that way folk trying to use this new feature with `nomatch=0L` will be nudged towards `nomatch=NULL`.

mattdowle:
Merging  now to make progress. As always, everything is always open for discussion and reconsideration until release.

PR #4356, opened 2020-04-05
unit test for fread meaningful error of non existing txt.gz

### LINKED ISSUE #3136: fread doen't work for .txt.gz but it does for the same file with .csv.gz extension
Hello!

I tried to read a .txt.gz file with `fread` using the new feature for reading .gz files (#717 ), but I got a warning and a 0 row table. But **when I manually change the extension from .txt.gz to .csv.gz I have no problem** and I obtain the dataset that I have compressed. 

I think this could be a bug. For now I changed the name of my files using batch, but it would be nice if `fread` was less strict with the extensions. 

I was using version 1.11.8

Example:

`   data = fread(input = "file.txt.gz",
               nrows = 100,
               colClasses = 'character',
               fill=TRUE)  `

` 
Taking input= as a system command ('file.txt.gz') and a variable has been used in the expression passed to 'input='. Please use fread(cmd=...). There is a security concern if you are creating an app, and the app could have a malicious user, and the app is not running in a secure envionment; e.g. the app is running as root. Please read item 5 in the NEWS file for v1.11.6 for more information and for the option to suppress this message.
"files" no se reconoce como un comando interno o externo,
programa o archivo por lotes ejecutable.
Warning messages:
1: running command 'C:\Windows\system32\cmd.exe /c (file.txt.gz) > C:\Documents\file.txt.gz' had status 1 
2: In (if (.Platform$OS.type == "unix") system else shell)(paste0("(",  :
  '(file.txt.gz) > C:\Users\rununez\AppData\Local\Temp\RtmpGaf6n4\file393417ff2352' execution failed with error code 1
3: In fread(input = "file.txt.gz", nrows = 100, colClasses = "character", fill = TRUE) :
  File 'C:\Users\rununez\AppData\Local\Temp\RtmpGaf6n4\file393417ff2352' has size 0. Returning a NULL data.table.
`

This is the output I obtain when I print the resulting dataset:

`
data 
`
`
Null data.table (0 rows and 0 cols)
`

Manually changing the file name to "file.csv,gz" (and changing the input variable accordingly) solves the problem

`
 data = fread(input = "file.csv.gz",
               nrows = 100,
               colClasses = 'character',
               fill=TRUE)  `

jangorecki:
Closes #3136

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16846,3 +16846,7 @@ A = data.table(A=c(complex(real = 1:3, imaginary=c(0, -1, 1)), NaN))
 test(2138.3, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
 A = data.table(A=as.complex(rep(NA, 5)))
 test(2138.4, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
+
+# fread attempt to read non-existing txt.gz file gives meaningful error message #3136
+test(2139.1, fread("not_existing_file.txt.gz", nrows = 100, colClasses = 'character', fill=TRUE), error="File.*does not exist or is non-readable")
+


mattdowle:
Test 1452 already covers this.
This is a case where we really need to confirm the test reproduced the problem, to make sure that the test is covering that problem.
I went back to the issue and could not reproduce their report using v1.11.8. I'll add a comment in the issue as to what I guess happened.

PR #4359, opened 2020-04-07
test by dot with verbose=T

### LINKED ISSUE #3196: replacedot might fail in by when referring to unscoped variable & verbose
```
x = data.table(
  i1 = c(234L, 250L, 169L, 234L, 147L, 96L, 96L, 369L, 147L, 96L),
  i4 = c(79L, 113L, 270L, -121L, 113L, 113L, -121L, 179L, -228L, 113L),
  v = 0
)

x[1:5, sum(v), by = .(i5 = 1:5 %% 2L), verbose = TRUE]
```

> Error in .(i5 = 1:5%%2L) : could not find function "."

But, works:

```
x[1:5, sum(v), by = .(i5 = 1:5 %% 2L)]
x[1:5, sum(v), by = list(i5 = 1:5 %% 2L), verbose = TRUE]
```

jangorecki:
closes #3196


mattdowle:
Ideally we should pinpoint when exactly this was fixed. So that we can confirm that the test correctly fails before that fix. Sometimes we get surprised and it actually passes before the fix. That causes us to realize there's a mistake in the test, or something. For example, that happened recently here: https://github.com/Rdatatable/data.table/pull/4356#issuecomment-843573997.
But I realize doing so takes time and there are other priorities. So what I did was test the test in v1.11.8 and it did correctly fail. Therefore it was fixed sometime between 1.11.8 and now.

MichaelChirico:
@mattdowle ran bisect, it was fixed here:

https://github.com/Rdatatable/data.table/commit/a694e91389513299c23e647f642cbd65b79a16ef

At a glance I don't immediately see why it fixed this issue, but at least it's related to `by`

mattdowle:
@MichaelChirico Great, thanks. I confirmed 1.12.2 fixed it, and tagged #3196 accordingly.

PR #4361, opened 2020-04-07
eval and := fail fixed

### LINKED ISSUE #1181: (some?) eval expressions have issues with :=
``` R
dt1 = data.table(a = 1, key = 'a')
dt2 = data.table(c = 1, fn = list(quote(5*a)), key = 'c')

# this works
dt1[, eval(dt2[a]$fn[[1]], .SD)]
[1] 5

# this doesn't
dt1[, n := eval(dt2[a]$fn[[1]], .SD)]
#Error in eval(expr, envir, enclos) : object 'a' not found
```

Relevant SO post - http://stackoverflow.com/a/30899285/817778

jangorecki:
closes #1181


mattdowle:
Shouldn't be a bug fix news item in the current release.
I went back in previous versions to see when it was fixed.  So far I can reproduce the bug in v1.9.6 (Sep 2015) but it seems fixed in v1.9.8 (Nov 2016).

PR #4366, opened 2020-04-08
nested dt addcol, tests for #1629

### LINKED ISSUE #1629: Nested calls to data.table() do not avoid shallow copies
For details, please see this [SO question](http://stackoverflow.com/questions/36434717/adding-column-to-nested-r-data-table-by-reference). The gist is that nested `data.table`s can trigger shallow copies when modified, but do so inconsistently depending on how they were constructed. Obviously this presents an issue w.r.t. writing functions that want to modify them. If we can depend on an update be reference code is cleaner. Otherwise recursive functions need to always return stuff and suppress the shallow copy warning messages.

This reproduces it.

```
DT = data.table(d = list(data.table(a=1))); DT$d[[1]][, new_col := NA]
```

Thanks much.

jangorecki:
closes #1629


inst/tests/tests.Rraw
```diff
@@ -16846,3 +16846,15 @@ A = data.table(A=c(complex(real = 1:3, imaginary=c(0, -1, 1)), NaN))
...
 test(2138.4, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
+
+# Nested calls to data.table() do not avoid shallow copies #1629
+add_col1 = function(dt) {
+  if (is.data.table(dt)) dt[, new_col:=NA]
+  if (is.list(dt)) lapply(dt, add_col1)
+  invisible()
+}
+dt = data.table(a=c(1,2), b=list(data.table(d=c("a", "b"), e=c(100, 200))))
+add_col1(dt)
```
mattdowle, line 16857: Using v1.12.2 (see https://github.com/Rdatatable/data.table/issues/1629#issuecomment-846121806) I ran the new test to confirm it fails correctly.  New tests .2 and .3 fail (good), .1 passes (also fine as we often have extra tests in a set of tests which test the test setup).  However, the `add_col1(dt)` produces warnings in v1.12.2 which aren't being checked are not produced.  So it would go the extra mile to wrap the `add_col1()` call with `test()` so that it is checked that no warnings are produced by it (it doesn't produce warnings in v1.12.4 onwards). 
```
> add_col1(dt)
Warning messages:
1: In `[.data.table`(dt, , `:=`(new_col, NA)) :
  Invalid .internal.selfref detected and fixed by taking a (shallow) copy of the data.table so that := can add this new column by reference. At an earlier point, this data.table has been copied by R (or was created manually using structure() or similar). Avoid names<- and attr<- which in R currently (and oddly) may copy the whole data.table. Use set* syntax instead to avoid copying: ?set, ?setnames and ?setattr. If this message doesn't help, please report your use case to the data.table issue tracker so the root cause can be fixed or this message improved.
2: In `[.data.table`(dt, , `:=`(new_col, NA)) :
  Invalid .internal.selfref detected and fixed by taking a (shallow) copy of the data.table so that := can add this new column by reference. At an earlier point, this data.table has been copied by R (or was created manually using structure() or similar). Avoid names<- and attr<- which in R currently (and oddly) may copy the whole data.table. Use set* syntax instead to avoid copying: ?set, ?setnames and ?setattr. If this message doesn't help, please report your use case to the data.table issue tracker so the root cause can be fixed or this message improved.
```

Also the test could be moved into 2049 to be next to the other one.

I'll do this and then merge.

PR #4372, opened 2020-04-10
export uniq function (uniqlist) before

### LINKED ISSUE #900: Export data.table:::uniqlist for external use
`uniqlist` is useful to detect whether there's a change in value in an R's data frame. (For example, check if a variable change values in time series data.)

See this StackOverflow [question](http://stackoverflow.com/questions/21266015/determine-when-columns-of-a-data-frame-change-value-and-return-indices-of-the-ch), which shows that there's no readable alternative to `uniqlist`

```r
x = data.frame(cnt=1:10, code=rep('ELEMENT 1',10), val0=rep(5,10),  val1=rep(6,10),val2=rep(3,10))
x[4,]$val0=6

data.table:::uniqlist(x[, 3:5])
# [1] 1 4 5
```

jangorecki:
closes #900
TODO: tests

### DIFF
--- a/R/uniqlist.R
+++ b/R/uniqlist.R
@@ -1,18 +1,28 @@
+nrow2 = function(x) {
+  if (!length(x)) return(0L)
+  if (is.data.table(x)) nrow(x) else if (is.list(x)) length(x[[1L]]) else stop("nrow2 expects data.table or list")
+}
 
-uniqlist = function (l, order = -1L)
-{
-  # Assumes input list is ordered by each list item (or by 'order' if supplied), and that all list elements are the same length
-  # Finds the non-duplicate rows. Was called duplist but now grows vector - doesn't over-allocate result vector and
-  # is >2x times faster on numeric types
-  # TO DO: Possibly reinstate reverse argument :
-  #    FALSE works in the usual duplicated() way,  the first in a sequence of dups, will be FALSE
-  #    TRUE has the last in a sequence of dups FALSE (so you can keep the last if that's required)
-  # l = list(...)
-  if (!is.list(l))
-    stop("l not type list")
-  if (!length(l))  return(list(0L))
-  ans = .Call(Cuniqlist, l, as.integer(order))
-  ans
+uniqlist = function (l, order=integer()) {
+  # used in "[.data.table" when doing groupby (!byjoin) to find the groups using byval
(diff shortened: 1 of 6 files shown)


src/uniqlist.c
```diff
@@ -356,16 +367,27 @@ SEXP uniqueNlogical(SEXP x, SEXP narmArg) {
...
+      return ScalarInteger(0);
+    } else {
+      return ScalarInteger(1);
+    }
+  }
   Rboolean second = ix[i];
   // we've found 2 different values (first and second). Which one didn't we find? Then just look for that.
   // NA_LOGICAL == INT_MIN checked in init.c
-  const int third = (first+second == 1) ? NA_LOGICAL : ( first+second == INT_MIN ? TRUE : FALSE );
+  int third;
```
jangorecki, line 380: this is not a `const` anymore, but to address compiler warnings I would need to `int tmp; if (...) tmp = ... else tmp = ...; const int third = tmp;` so I think it is OK

MichaelChirico:
@jangorecki WDYT, worth reviving this? I still don't have a great idea for a name of the exported function. IINM the current implementation still assumes it receives a sorted input -- for the exported function, I think the expectation should be that we ensure sorting.

jangorecki:
Ensure sorted will suffocate its performance. I found another SO that praises its speed
https://stackoverflow.com/questions/76582182/efficiently-find-the-first-of-the-last-1s-sequence/76585097#76585097
If we document that then I don't think we should worry about order.

R/uniqlist.R
```diff
@@ -1,18 +1,41 @@
...
+
+uniq = function(x, order=integer()) {
+  if (!is.list(x))
+    stop("'x' must be a data.table type object");
+  if (!is.integer(order)) {
+    if (is.numeric(order))
+      order = as.integer(order)
+    else
+      stop("'order' must be an integer")
+  }
```
MichaelChirico, line 29:
```suggestion
  if (!is.numeric(order))
    stopf("'%s' must be an integer", "order")
  order = as.integer(order)
```

MichaelChirico:
> Ensure sorted will suffocate its performance.

Not necessarily right? We can have the signature be `uniq(x, cols, index)`, if providing `cols` we can check that input is keyed/indexed & sort if not; `index` allows direct access to provide sort order. WDYT?

Issue #4346, opened 2020-04-03
forder should not accept list, but forderv should


jangorecki:
Recent issue reported by @MichaelChirico https://github.com/Rdatatable/data.table/issues/4214 reminded me some old discussion that happened in https://github.com/Rdatatable/data.table/pull/3817.

To summarize it.

- `forder` should be a replacement for `order`
- `forderv` should expect data.table/list input
- whenever user uses `forder(DT)` there should be a (verbose?) message pointing to use `forderv`.

also we should first resolve https://github.com/Rdatatable/data.table/issues/3447

jangorecki:
I would like to propose for `forderv` to have a default `retGrp`, that means secondary indices would carry that attribute as well, be a little bit more heavy, but it opens more possibilities to avoid heavy re-computation. One of many examples
```
# TODO: could check/reuse secondary indices, but we need 'starts' attribute as well!
```

PR #4373, opened 2020-04-11
export is.sorted

### LINKED ISSUE #2325: [Request] export is.sorted
I'd like to test whether my input table is sorted and error out if not. The base tools for this [are not very appealing](https://stackoverflow.com/questions/7599146/testing-if-rows-of-a-matrix-or-data-frame-are-sorted-in-r). 

For now, I have a workaround:

```
library(data.table)
is_sorted = function(DT, cols = names(DT)){
  e = as.call(lapply(c("order", cols), as.name))
  identical(seq_len(nrow(DT)), DT[eval(e), which = TRUE])
} 

DT = data.table(id = 1:3, v = c(2,1,3))
is_sorted(DT) # TRUE
is_sorted(DT, "v") # FALSE
```

This is similar to a couple answers on SO, but implicitly uses `forder` instead of `order`. I guess `data.table:::is.sorted` is more efficient for this, though.

jangorecki:
add argument retOrd to return order

unfold C for codecov

closes #2325

### DIFF
--- a/R/setkey.R
+++ b/R/setkey.R
@@ -155,16 +155,24 @@ setreordervec = function(x, order) .Call(Creorder, x, order)
 # The others (order, sort.int etc) are turned off to protect ourselves from using them internally, for speed and for
 # consistency; e.g., consistent twiddling of numeric/integer64, NA at the beginning of integer, locale ordering of character vectors.
 
-is.sorted = function(x, by=seq_along(x)) {
+is.sorted = function(x, by=seq_along(x), retOrd=FALSE) {
   if (is.list(x)) {
-    warning("Use 'if (length(o <- forderv(DT,by))) ...' for efficiency in one step, so you have o as well if not sorted.")
+    # for efficient use via retOrd argument see note in ?is.sorted
     # could pass through a flag for forderv to return early on first FALSE. But we don't need that internally
     # since internally we always then need ordering, an it's better in one step. Don't want inefficiency to creep in.
-    # This is only here for user/debugging use to check/test valid keys; e.g. data.table:::is.sorted(DT,by)
-    0L == length(forderv(x,by,retGrp=FALSE,sort=TRUE))
-  } else {
+    o = forderv(x,by,retGrp=FALSE,sort=TRUE)
+    ans = 0L == length(o)
+    if (isTRUE(retOrd))
+      ans = setattr(copy(ans), "order", o)
+    ans
+  } else if (is.null(x)) { # NULL does not satisfy C isVectorAtomic
+    NA
+  } else if (is.atomic(x)) {
+    if (isTRUE(retOrd)) stop("retOrd works only for data.table/list input")
     if (!missing(by)) stop("x is vector but 'by' is supplied")
     .Call(Cfsorted, x)
+  } else {
+    stop("'x' argument is of unsupported type")
   }
   # Cfsorted could be named CfIsSorted, but since "sorted" is an adjective not verb, it's clear; e.g., Cfsort would sort it ("sort" is verb).
   # Return value of TRUE/FALSE is relied on in [.data.table quite a bit on vectors. Simple. Stick with that (rather than -1/0/+1)
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -56,6 +56,7 @@ export(nafill)
 export(setnafill)
 export(.Last.updated)
 export(fcoalesce)
+export(is.sorted)
 
 S3method("[", data.table)
 S3method("[<-", data.table)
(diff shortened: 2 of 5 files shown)


R/setkey.R
```diff
@@ -155,16 +155,24 @@ setreordervec = function(x, order) .Call(Creorder, x, order)
...
-  } else {
+    o = forderv(x,by,retGrp=FALSE,sort=TRUE)
+    ans = 0L == length(o)
+    if (isTRUE(retOrd))
+      ans = setattr(copy(ans), "order", o)
+    ans
+  } else if (is.null(x)) { # NULL does not satisfy C isVectorAtomic
+    NA
+  } else if (is.atomic(x)) {
+    if (isTRUE(retOrd)) stop("retOrd works only for data.table/list input")
```
ColeMiller1, line 171: If I saw this error message, I would make my vector a list. Should we automatically do ```list(x)``` for convenience?

R/setkey.R
```diff
@@ -155,16 +155,24 @@ setreordervec = function(x, order) .Call(Creorder, x, order)
 # The others (order, sort.int etc) are turned off to protect ourselves from using them internally, for speed and for
 # consistency; e.g., consistent twiddling of numeric/integer64, NA at the beginning of integer, locale ordering of character vectors.
 
-is.sorted = function(x, by=seq_along(x)) {
+is.sorted = function(x, by=seq_along(x), retOrd=FALSE) {
   if (is.list(x)) {
-    warning("Use 'if (length(o <- forderv(DT,by))) ...' for efficiency in one step, so you have o as well if not sorted.")
+    # for efficient use via retOrd argument see note in ?is.sorted
```
ColeMiller1, line 160: What is the use case of ```is.sorted(..., retOrd = TRUE)```? The order seems better dealt with exporting ```forder```.

man/is.sorted.Rd
```diff
@@ -0,0 +1,50 @@
...
+\title{ Checks if input is sorted }
+\description{
+  Checks if input is sorted.
+}
+\usage{
+  is.sorted(x, by=seq_along(x), retOrd=FALSE)
+}
+\arguments{
+  \item{x}{ data.table type object or atomic vector. }
+  \item{by}{ data.table columns used to check if \code{x} is sorted by those columns. }
```
ColeMiller1, line 12: ```by = seq_along(x)``` implies an integer argument. Since a character vector is supported, we may want to point it out.

man/is.sorted.Rd
```diff
@@ -0,0 +1,50 @@
...
+\preformatted{
+  if (!is.sorted(DT, by="Sepal.Length"))
+    DT = DT[order(Sepal.Length)]
+}
+
+  could be written as
+
+\preformatted{
+  if (!s <- is.sorted(DT, by="Sepal.Length", retOrd=TRUE))
+    DT = DT[attr(s, "order")]
```
ColeMiller1, line 31: This use case, wouldn't a user do ```DT[order(Sepal.Length)]``` and have nearly identical performance?

R/setkey.R
jangorecki: On one hand it would be nice, but fsorted that we use for atomic vectors is likely to be much more efficient. So by default it definitely not good to wrap it in list.nooooooooooooooooooooooooooooooooooooooooooooooooo!

R/setkey.R
jangorecki: Agree, use case address the warning without exporting forder

man/is.sorted.Rd
jangorecki: For iris, yes.
This is basically to address @mattdowle comment in #2325 and the warning that got removed.

R/setkey.R
ColeMiller1: Only wrap it if ```retOrd == TRUE```.  That arg would seem to trump performance considerations of fsorted.

jangorecki:
@ColeMiller1 maybe I should just export `forder` at the same time... but there is a pending change to it described in #4346

man/is.sorted.Rd
```diff
@@ -0,0 +1,50 @@
...
+\usage{
+  is.sorted(x, by=seq_along(x), retOrd=FALSE)
+}
+\arguments{
+  \item{x}{ data.table type object or atomic vector. }
+  \item{by}{ data.table columns used to check if \code{x} is sorted by those columns. }
+  \item{retOrd}{ logical, when \code{TRUE} it will set an attribute \code{"order"} on the returned value, providing an order of \code{x}. Works only for data.table type \code{x}, not for atomic vector. }
+}
+\details{
+  Checks if the input is object is sorted. Can check also by a subset of columns provided in \code{by} argument. Can also return an order used in computation when using \code{retOrd} argument.
```
jangorecki, line 16: Should possibly mention integer() trick in case for 1:n

jangorecki: It is actually explained later on, so maybe just a pointer to section in the manual

jangorecki:
Argument retOrd is overlapping with #3447 and could possibly be removed from this PR

Issue #4348, opened 2020-04-03
support OpenMP with system macOS toolchain


kevinushey:
With R 4.0.0, R (as well as packages on CRAN) will be built with the system toolchain (Apple Clang), which unfortunately mean OpenMP support is not provided natively by the compiler. Fortunately, it is still possible to use OpenMP with the system toolchain if you:

1. Instruct `clang` to enable OpenMP with `-Xpreprocessor -fopenmp`,
2. Explicitly provide include paths to `libomp` during compilation,
3. Explicitly provide library paths to `libomp` during link.

Unfortunately, `data.table` fails to detect this case as it invokes the compiler with just `-fopenmp` here:

https://github.com/Rdatatable/data.table/blob/b1b1832b0d2d4032b46477d9fe6efb15006664f4/configure#L64

It would be helpful for users on macOS if support for OpenMP on macOS could be detected + used.

Sorry, I recognize this is kind of a pain but figured it was worth filing since OpenMP is somewhat vital to data.table's performance.

See also:

http://r-sig-mac.29524.n8.nabble.com/R-SIG-Mac-Apple-Clang-does-support-OpenMP-if-libomp-is-available-td458.html#a461

(+ some other recent discussion on R-SIG-Mac)

MichaelChirico:
I think Jim Heater raised this same issue, let me see...

krivard:
I am pasting my working version of `Makevars` here with edits from [the version in the install docs](https://github.com/Rdatatable/data.table/wiki/Installation) marked, so that google may find it and help a future traveller:

```
# Added: -Xpreprocessor
CC=/usr/bin/clang -Xpreprocessor -fopenmp

# Added: -Xpreprocessor -lomp -I/usr/local/include
CXX=/usr/bin/clang++ -Xpreprocessor -fopenmp -lomp -I/usr/local/include
# -O3 should be faster than -O2 (default) level optimisation ..
CFLAGS=-g -O3 -Wall -pedantic -std=gnu99 -mtune=native -pipe
CXXFLAGS=-g -O3 -Wall -pedantic -std=c++11 -mtune=native -pipe

# Added: -lomp
LDFLAGS=-L/usr/local/opt/gettext/lib -L/usr/lib -Wl,-rpath,/usr/lib -lomp
CPPFLAGS=-I/usr/local/opt/gettext/include -I/usr/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include
```

[...]

PR #4374, opened 2020-04-11
check if R CMD SHLIB can compile OpenMP directly

kevinushey:
Part of https://github.com/Rdatatable/data.table/issues/4348.

The intention here is that if the user has already modified `src/Makevars` such that OpenMP "just works", then that should be detected by explicitly validating that `R CMD SHLIB` can compile with OpenMP.

This allows `data.table` to compile using OpenMP for me when using the following Makevars:

```
CPPFLAGS     += -I/usr/local/opt/libomp/include -Xpreprocessor -fopenmp
LDFLAGS      += -L/usr/local/opt/libomp/lib -lomp

SHLIB_OPENMP_CFLAGS   =
SHLIB_OPENMP_CXXFLAGS =
```

Asking whether we can set those flags (or similar) in R 4.0.0 is a separate question, I think.


mattdowle:
Thanks, @kevinushey! As Jan suggested, I've invited you to be project member, and will add you to contributors list.

PR #4404, opened 2020-04-27
unload namespace before reinstalling, #4403

### LINKED ISSUE #4403: data.table::update.dev.pkg() warning or error
`data.table v 1.12.8`:  

applying 
```
data.table::update.dev.pkg()
```
gives the following message in R v. 4.0.0, R Studio v 1. 2.5042 both with Admin rights:
```
R data.table package has been updated to NA (1.12.8)
Warning message:
In file.copy(savedcopy, lib, recursive = TRUE) :
  problem copying C:\R\R-4.0.0\library\00LOCK\data.table\libs\x64\datatable.dll to C:\R\R-4.0.0\library\data.table\libs\x64\datatable.dll: Permission denied
```

```
> sessionInfo()
R version 4.0.0 (2020-04-24)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:

[1] compiler_4.0.0    tools_4.0.0       data.table_1.12.8
```

jangorecki:
closes #4403
```r
#R version 4.0.0 (2020-04-24) -- "Arbor Day"
#Copyright (C) 2020 The R Foundation for Statistical Computing
#Platform: x86_64-w64-mingw32/x64 (64-bit)
#...
data.table::update.dev.pkg()
#trying URL 'https://Rdatatable.gitlab.io/data.table/bin/windows/contrib/4.0/data.table_1.12.9.zip'
#Content type 'application/zip' length 2675878 bytes (2.6 MB)
#downloaded 2.6 MB
#
#package 'data.table' successfully unpacked and MD5 sums checked
#
#The downloaded binary packages are in
#        C:\Users\jan\AppData\Local\Temp\Rtmps5UOqq\downloaded_packages
#R data.table package has been updated to dd7609e83132e19c9be80d71c73e8a8f95e19b27 (1.12.9)
```


jangorecki:
codecov fail is false positive

jangorecki:
We might not even need `on.exit()` but just `unloadNamespace` alone, but better to leave it.
If PR would not always fix the issue, then we could add `Sys.sleep(2)` after `unloadNamespace` and before `install.packages`. Testing on windows R 4.0.0 would be appreciated.
```sh
R -q
unload = function() {unloadNamespace("pkg"); !"pkg"%in%loadedNamespaces()}
utils::package.skeleton("pkg", list="unload")
q("no")
rm pkg/man/unload.Rd
R CMD build pkg
R CMD INSTALL pkg_1.0.tar.gz 
R -q
library(pkg)
unload()
#[1] TRUE
```

R/devel.R
```diff
@@ -17,22 +17,33 @@ dcf.repo = function(pkg, repo, field, type){
...
-              utils::packageVersion(pkg, lib.loc=lib)))
+  if (una)
+    cat(sprintf("No revision information found in DESCRIPTION file for %s package. Unsure '%s' is correct field in PACKAGES file in your package repository '%s'. Otherwise package will be re-installed every time, proceeding to installation.\n",
+                pkg, field, contrib.url(repo, type=type)))
+  # see if Revision is different then currently installed Revision, note that installed package will have Revision info only when it was installed from remote devel repo
+  upg = una || !identical(ups, dcf.lib(pkg, field, lib.loc=lib))
+  # update.dev.pkg fails on windows R 4.0.0, we have to unload package namespace before installing new version #4403
+  on.exit({
+    if (upg) {
+      unloadNamespace(pkg)
```
mattdowle, line 38: Great.  Is the idea of the `unloadNamepsace()` to relese the dll lock on Windows?  If so, adding a comment to to that effect would be good.

R/devel.R
```diff
@@ -17,22 +17,33 @@ dcf.repo = function(pkg, repo, field, type){
...
+      utils::install.packages(pkg, repos=repo, type=type, lib=lib, ...)
```
mattdowle, line 39: Maybe wrap this with a tryCatch()?  Just so that a more helpful error message can be produced instead; e.g. "Please close all R sessions to release the dll lock on Windows, and try again. If the dll lock is still open, try a reboot which should release the lock on Windows. Also, if you have a .Rpofile which loads data.table on startup, turn that off temporarily in order to upgrade the dll on Windows."

jangorecki: .Rprofile should not be an issue, because DT is already loaded at this point.

jangorecki: I would lean towards keeping `on.exit` chunk as simple as possible

PR #4408, opened 2020-04-30
Add . and .. aliases to ?data.table

### LINKED ISSUE #2145: Describe "dot dot prefix" in help text, FAQ and vignette
In [New Features for 1.10.2](https://github.com/Rdatatable/data.table/blob/master/NEWS.md#changes-in-v1102--on-cran-31-jan-2017) the `..` prefix is introduced. However, it is not mentioned in `?data.table`. 

Furthermore, if you do a mistake in your subsetting, as examplified e.g. in [this nice answer](http://stackoverflow.com/questions/41051890/r-data-table-1-10-0-why-does-a-named-column-index-value-not-work-while-a-int/41052245#41052245), an informative error message is generated:

> Error in `[.data.table`(DT, , myCols) : 
>   j (the 2nd argument inside [...]) is a single symbol but column name 'myCols' is not found. Perhaps you intended DT[,..myCols] or DT[,myCols,with=FALSE]. This difference to data.frame is deliberate and explained in FAQ 1.1. 

From the error message one could easily get the impression that `..` is explained in FAQ 1.1. But it's not. 

In the release notes it says that `..` "is experimental", so perhaps this omission is intentional? Still, I think this feature deserves to be described in the help text and FAQ (and in the vignette), possibly with an "experimental" reservation.

### LINKED ISSUE #4385: ?".." should return a page about ..
Users reading someone else's `data.table` code & unfamiliar with `..` might naturally try `?".."`, that should link to a relevant page.

### LINKED ISSUE #4407: ?"." should return the help page where the . alias is described
Related: #4385 

We see this kind of question more than once:

https://stackoverflow.com/questions/61514355/how-to-understand-the-meaning-of-in-data-table

Indeed, it's hard to search for. `?"."` (and maybe `?".("` or `?".()"`) should return the page for `?"[.data.table"`

MichaelChirico:
Closes #4407 
Closes #4385 
Closes #2145 ( #4862 already closed the FAQ element and this PR refines FAQ 1.5 further )


MichaelChirico:
not sure how I changed coverage 🤔

avimallu:
I fancied checking earlier open issues, and came across #2145, which is currently partly covered already in `?data.table`'s examples as:

```r
colNum = 2                     # to refer vars in `j` from the outside of data use `..` prefix
DT[, ..colNum]                 # same, equivalent to DT[, .SD, .SDcols=colNum]
```
It is also covered in the introduction vignette under the header _Select columns named in a variable using the `..` prefix_.

Lastly, this PR adds the `..` operator to the help text, effectively, so I feel that this PR should probably close that issue.

PR #4427, opened 2020-05-04
merge sort arg document better

### LINKED ISSUE #2574: sort = FALSE does't generate equivalent results with data.frame.merge
Here is sample/test code:

```
A <- data.frame(a=1:4)
AB <- data.frame(a=2:4, b=1)
merge(A,AB,all.x = TRUE,sort=FALSE)
  a  b
1 2  1
2 3  1
3 4  1
4 1 NA

A <- data.table::data.table(a=1:4)
AB <- data.table::data.table(a=2:4, b=1)
merge(A,AB,all.x = TRUE,sort=FALSE)
   a  b
1: 1 NA
2: 2  1
3: 3  1
4: 4  1
```
This can have profound effects. Merge depends on type of `A` and `data.frame`/`data.table` are interchangeable in many scenario's. Changing type of `A` can end up causing fireworks at unexpected places. 

I don't know what the correct behavior should be, but I do think the results should be the same.

### LINKED ISSUE #2594: forderv orders NA's different than base::order by default
Probably, this is known and intended:
```
x <- c(3,NA,6,NaN, 2, 1)
x[data.table:::forderv(x)]
# [1]  NA NaN   1   2   3   6
x[order(x)]
#  1   2   3   6  NA NaN
```
The reason is the different default for the `na.last` argument.
Is there a reason for this?
The result is that `merge.data.frame` and `merge.data.table` have different row orders since the `setkey` used for sorting has the default NA ordering of data.table:
```
df1 <- data.frame(x = c(2,NA,1, NaN), y = 1:4)
df2 <- data.frame(x = c(NA,1), y = 5:8)
merge(df1, df2, by = "x", sort = TRUE)
# x y.x y.y
# 1  1   3   6
# 2  1   3   8
# 3 NA   2   5
# 4 NA   2   7
merge(setDT(df1), setDT(df2), by = "x", sort = TRUE)
# x y.x y.y
# 1: NA   2   5
# 2: NA   2   7
# 3:  1   3   6
# 4:  1   3   8

### PR DESCRIPTION
**Pull request by jangorecki**
explain difference of sorting NAs in data.table merge

closes #2574
closes #2594

### CODE REVIEW

File: man/merge.Rd [hunk 1]
```diff
@@ -54,18 +55,9 @@ as any \code{by.x}.}
...
 \code{merge.data.frame} method or \code{merge.data.table} method depending on
-the class of its first argument. Note that, unlike \code{SQL}, \code{NA} is
+the class of its first argument. Note that, unlike \code{SQL} join, \code{NA} is
 matched against \code{NA} (and \code{NaN} against \code{NaN}) while merging.
 
-In versions \code{<= v1.9.4}, if the specified columns in \code{by} were not the
-key (or head of the key) of \code{x} or \code{y}, then a \code{\link{copy}} is
-first re-keyed prior to performing the merge. This was less performant as well as memory
-inefficient. The concept of secondary keys (implemented in \code{v1.9.4}) was
-used to overcome this limitation from \code{v1.9.6}+. No deep copies are made
```
MichaelChirico, line 64: can this sentence be included elsewhere in the Rd?

 > No deep copies are made..., thereby improving performance and memory efficiency.

jangorecki: That was an extra deep copy that we needed for setkey

MichaelChirico: OK

Issue #4424, opened 2020-05-04
internal anyNamed no longer returns false?


jangorecki:
According to codecov the follow line is no longer visited:
https://github.com/Rdatatable/data.table/blob/dd7609e83132e19c9be80d71c73e8a8f95e19b27/src/assign.c#L665
Most likely the function may no longer return false (unless we have poor coverage of it), which is probably a result of switching codecov environment from R 3.6 to R 4.0. We might want to revisit use of `anyNamed` function.

tlapak:
Unfortunately, I don't have the time to implement and test a fix but I did have the time to look through the sources and was curious. Maybe this helps.

The function can still return false, I guess, just not when it's called on a non-empty list. But that's currently all that it's used for. Thing is that with reference counting turned on, MAYBE_NAMED checks the reference count and SET_VECTOR_ELT bumps up the reference count of the element it sticks into the list so that it's always at least one. With reference counting turned off, the named attribute _doesn't_ get set. So you'd need to branch based on reference counting yes/no. Or rip out the test one way or another.

PR #4428, opened 2020-05-04
fixes R-devel foverlaps failures due to c.POSIXct change

jangorecki:
fixes
```
  Test 1390.3 produced 1 errors but expected 0
  Expected:
  Observed: 'origin' must be supplied
  Test 1390.4 produced 1 errors but expected 0
  Expected:
  Observed: 'origin' must be supplied
  Test 1390.5 produced 1 errors but expected 0
  Expected:
  Observed: 'origin' must be supplied
  Test 1985.1 produced 1 errors but expected 0
  Expected:
  Observed: 'origin' must be supplied
  Test 1993.1 produced 1 errors but expected 0
  Expected:
  Observed: 'origin' must be supplied
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -147,6 +147,8 @@ unit = "s")
 
 7. Added more explanation/examples to `?data.table` for how to use `.BY`, [#1363](https://github.com/Rdatatable/data.table/issues/1363).
 
+8. Change of `c.POSIXct` method planned for R 4.1.0 impacted `foverlaps` function that could raise `'origin' must be supplied` error. Fix for planned change has been provided in [#4428](https://github.com/Rdatatable/data.table/pull/4428).
+
 
 # data.table [v1.12.8](https://github.com/Rdatatable/data.table/milestone/15?closed=1)  (09 Dec 2019)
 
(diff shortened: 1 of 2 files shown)


mattdowle:
Great. Needs news item. For example, when this change in R-devel is released, and users start getting this error message in data.table <= 1.12.8, then if they search for the error, hopefully they'll find the news item and find out they need to upgrade data.table.

jangorecki:
@mattdowle done

mattdowle:
This missed coverage line is odd, doesn't seem related to this PR. Will let it go. If it's truly now uncovered then master coverage will show it and it won't get forgotten.
image

jangorecki:
@mattdowle see https://github.com/Rdatatable/data.table/issues/4424

eddelbuettel:
Are we sure this only affects r-devel?  It seems to also affect R-patches if I am reading the [CRAN results page for `data.table`](https://cloud.r-project.org/web/checks/check_results_data.table.html) correctly (but not R-release).

It affects us in Debian as the release data.table fails its tests with R 4.0.0, effectively block transition of r-base 4.0.0 (and now r-base 4.0.1) from unstable into testing.  See [#962497](bugs.debian.org/962497).

jangorecki:
Quite possibly R-patched as well. It is actually surprising that base R decided to make a breaking change like this without first giving a warning for at least one version.
We are planning to publish to CRAN soon, so hopefully it will not last for long.

eddelbuettel:
And the actual fix is a one-liner to maybe the Debian maintainer for data.table can carry that over. I probably would if it were my package (though this speculating as there may be other things too betweeb R 4.0.0 and data.table 0.12.8).

I think the change to `c.POSIXct` was pretty late-breaking as well.  Oh well.  Spilled milk now.

PR #4440, opened 2020-05-10
Remove deep copy of indices from shallow()

### LINKED ISSUE #4311: Major performance drop of keyed := when index is present
The performance of `dt[selector, foo := bar]` on key could significantly drop when an index is present. Following is my use case and reproducible example:

```r
library(data.table)

dt <- data.table(symbol = rep(1:1000, each = 5000))
dt[, date := seq_len(.N), by = symbol]
setkeyv(dt, c("symbol", "date"))

flag_dt <- data.table(symbol = sample.int(500, 5000, replace = TRUE))
flag_dt[, start_date := sample.int(3000, .N, replace = TRUE)]
flag_dt[, end_date := start_date + sample.int(3000, .N, replace = TRUE)]
flag_dt[, id := seq_len(.N)]

calendar <- dt[, sort(unique(date))]
```

When `dt` has no index, the following code that repeatedly using `symbol, date` selector to modify `flag` is fast enough.

```r
system.time({
  dt[, flag := 0L]
  flag_dt[, {
    dates <- calendar[calendar %between% c(start_date, end_date)]
    if (length(dates)) {
      selector <- list(symbol, dates)
      dt[selector, flag := 1L]
    }
    NULL
  }, by = id]
})
```

```
   user  system elapsed 
 26.189   0.399   3.344 
```

```
  user  system elapsed 
 24.648   0.220   3.119 
```

However, if an index is created intentionally, or in many cases unintentionally (auto index triggered by `dt[flag0 == 1, ...]`), the performance of the above code significantly decreases and could be unstable:

```r
dt[, flag0 := sample(0:1, .N, replace = TRUE)]
setindexv(dt, "flag0")
system.time({
  dt[, flag := 0L]
  flag_dt[, {
    dates <- calendar[calendar %between% c(start_date, end_date)]
    if (length(dates)) {
      selector <- list(symbol, dates)
      dt[selector, flag := 1L]
    }
    NULL
  }, by = id]
})
```

```
   user  system elapsed 
386.415  27.380  52.938
```

```
   user  system elapsed 
212.908   7.289  27.665
```

I also tried explicitly writing `dt[selector, flag := 1L, on = .(symbol, date)]`, still no luck.

Avoiding creating an index or disabling auto-index could avoid this problem but I'm still curious if there's something that significantly adds the overhead of keyed `:=` while there's an index.

tlapak:
Close #4311.

This ~still needs a news item~ might need some benchmarking to confirm that the other occurrences of `shallow()` had the same impact as they did on joins. But they should. Those would be
- most setops
- frankv
- dcast
- foverlaps
- duplicated.data.table

And indirectly of course everything that calls these. But the overhead is only really noticeable when there are many calls as in the original report. I don't know when I'll be able to finish this so I'm putting it up as is for now.

In the code there was a reference to `example(merge)` breaking if the sorted attribute wasn't copied. That seems to have been due to a pattern of calling setnames internally in the past. But that pattern hasn't been in use for about five years and nothing currently breaks if I don't make that copy. I do it anyway and have added a test for it as I think it's expected to work, even if `shallow()` isn't exported.


jangorecki:
This is great, I was recently worrying about using too much of `shallow` and being affected by this issue.

src/assign.c
```diff
@@ -152,13 +152,21 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   R_len_t i,l;
   int protecti=0;
   SEXP newdt = PROTECT(allocVector(VECSXP, n)); protecti++;   // to do, use growVector here?
-  //copyMostAttrib(dt, newdt);   // including class
-  DUPLICATE_ATTRIB(newdt, dt);
+  SHALLOW_DUPLICATE_ATTRIB(newdt, dt);
```
jangorecki, line 155: `SHALLOW_DUPLICATE_ATTRIB` may not be available in R 3.1.0, this is the file where that macro is defined in current R version, it seems to be missing in 3.1 https://github.com/wch/r-source/blob/R-3-1-branch/src/main/coerce.c

src/assign.c
```diff
@@ -152,13 +152,21 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
...
   //        so that the next change knows to duplicate.
-  //        Does copyMostAttrib duplicate each attrib or does it point? It seems to point, hence DUPLICATE_ATTRIB
-  //        for now otherwise example(merge.data.table) fails (since attr(d4,"sorted") gets written by setnames).
+  //        keepattr() also merely points to the entire attrbutes list and thus doesn't allow replacing
+  //        some of its elements.
+  
+  // We copy all attributes that refer to column names so that calling setnames on either
+  // the original or the shallow copy doesn't break anything.
+  SEXP index = PROTECT(getAttrib(dt, sym_index)); protecti++;
+  setAttrib(newdt, sym_index, shallow_duplicate(index));
```
jangorecki, line 165: `shallow_duplicate` was AFAIK designed to be used on `VECSXP`, `LISTSXP`, etc. You call it here on `INTSXP`. It would be good to figure out proper/better? way to shallow copy index. PS. I do exactly the same in #4386.
Hope @mattdowle will have good idea how to handle that. Maybe we need to change index structure, and wrap it into extra list?

jangorecki:
We need tests that uses `address` function, those tests need to fail on current master, and pass on this PR.

src/assign.c
tlapak: Looks like you're right, it doesn't show up in Rinternals.h for the 3.1 branch. It is only available starting from the 3.3 branch. Do you think lifting the code from https://github.com/wch/r-source/blob/bf4cf8912596162a3ae5a2b7da1a74220105b172/src/main/memory.c (I believe this is the actual definition?) would be an option? I had tested with `copyMostAttrib` which seems to be functionally the same but of course it doesn't check the object and S4 bits and leaves out the names. But at least here it would work I think. I was just a little worried there and opted for `SHALLOW_DUPLICATE_ATTRIB` because that the closest to how it was before.

jangorecki:
```c
SET_ATTRIB(newdt, shallow_duplicate(ATTRIB(dt)));
```
could possibly work. `ATTRIB` gives `LISTSXP` so will play well with `shallow_duplicate`.

src/assign.c
tlapak: Looking at https://github.com/wch/r-source/blob/trunk/src/main/duplicate.c that doesn't seem to be an issue. It starts out exactly like `duplicate` in that it makes a copy of the `INTSXP` and then a shallow copy of the attributes. And the flow looks the same for R 3.1 as far as I can tell.
Initially, I had done this more manually by allocating a zero length `INTSXP` and making a shallow copy of the attributes but this seemed more succinct.

src/assign.c
jangorecki: You may find this example useful
https://github.com/Rdatatable/data.table/pull/4370/files#diff-293a9e4361f16756700c7a5baddbabafR19-R34

tlapak:
Yeah that makes sense. So comparing `address(attr(attr(dt, 'index'), '__a'))` to the shallow copy should do it I think.
Oh yeah, I wasn't too worried, only a bit mystified by nqrecreateindices.c showing up in the report. But I reckon this is due to me working off of the branch before the last merge.

src/assign.c
tlapak: Yes, that is essentially how `SHALLOW_DUPLICATE_ATTRIB` is defined so it ought to work (minus the ATTRIB on newdt) and it's what you do in your link. Thanks for that btw. So I'll change that line.

src/assign.c
```diff
@@ -152,13 +152,24 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
...
+  
+  SEXP sorted = PROTECT(getAttrib(dt, sym_sorted)); protecti++;
+  setAttrib(newdt, sym_sorted, duplicate(sorted));
```
MichaelChirico, line 171: any reason to use deep copy here if we're doing shallow copy for the other attributes?

jangorecki: This is just character vector `c("key_col1","key_col2")`. I don't think it cost anything

tlapak: They are exactly the same in this situation. Both make a copy of the atomic vector, including a string vector and only differ in how they treat the attributes. This seemed to me to be the more defensive approach.

MichaelChirico: works for me! thanks

src/assign.c
```diff
@@ -152,13 +152,24 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
...
   //        also increases truelength. Perhaps make that distinction, then, and split out, but marked
   //        so that the next change knows to duplicate.
-  //        Does copyMostAttrib duplicate each attrib or does it point? It seems to point, hence DUPLICATE_ATTRIB
-  //        for now otherwise example(merge.data.table) fails (since attr(d4,"sorted") gets written by setnames).
+  //        keepattr() also merely points to the entire attrbutes list and thus doesn't allow replacing
+  //        some of its elements.
+  
+  // We copy all attributes that refer to column names so that calling setnames on either
+  // the original or the shallow copy doesn't break anything.
+  SEXP index = PROTECT(getAttrib(dt, sym_index)); protecti++;
```
jangorecki, line 167: Protect AFAIK is not needed. `dt` would need to be garbage collected before we will assign index as an attribute of `newdt`.

tlapak: You are probably right but I wasn't quite sure so decided to follow the pattern of just a few lines later and take the cautious approach. Looking through the rest of the file, there are a few instances where the call to `getAttrib` is not `PROTECT`ed and many where it is. I also don't think that `dt` could be garbage collected.

Anyway, I wouldn't mind taking it out but I don't think it hurts and then it begs the question of what to do further down in the function or in the rest of the code.

jangorecki:
If you need to benchmark this PR on a bigger machine, please prepare optimistic and pesimistic scenario cases.

src/assign.c
jangorecki: Agree, I spotted in the code there were protects sometimes where it was not needed. It is not really an issue, but it spreads the pattern of overprotecting, as you see on yourself. I used to do it as well. Still, I might be wrong, but this will be verified when preparing release for CRAN and running strict memory tests.

tlapak:
It's not so much about needing a bigger machine, it's more about finding the time to come up with meaningful benchmarks. Just for `shallow()` it's clear enough so I'll add a small example below. But other operations take so much more time that they drown out the index copying in most situations and the effect only really becomes apparent in loops, similar to the original report. The results aren't massive, but they're there.

```R
shallow <- data.table:::shallow
set.seed(1L)
n <- 1e8
dt1 <- data.table(a=sample(n, n))
setindex(dt1, a)

system.time({
    dt2 <- shallow(dt1)
})
# before
# user  system elapsed 
# 0.11    0.03    0.15 

# now
# user  system elapsed 
#    0       0       0 
```
And for some functions:
```R
library('data.table')

shallow <- data.table:::shallow
set.seed(1L)
n <- 5e6
dt1 <- data.table(a=sample(n, n))
setindex(dt1, a)

dt2 <- data.table(a=sample(1e2, 1e2))
setindex(dt2, a)

gc()
system.time({
  for (i in seq_len(1e2))
    fintersect(dt2, dt1)
})
# before
#   user  system elapsed 
#  72.35    9.45   54.41 

# now
#   user  system elapsed 
#  69.65    8.80   49.03 

system.time({
  for (i in seq_len(1e2))
    duplicated(dt1)
})

# before
#   user  system elapsed 
#  16.25    4.11   14.28 

# now
#   user  system elapsed 
#  17.13    4.53   14.76 

system.time({
  for (i in seq_len(1e2))
    frankv(dt1, 'a')
})
# before
#   user  system elapsed 
#  65.17    7.88   50.44 

# now
#   user  system elapsed 
#  64.72    6.45   48.38 

system.time({
  for (i in seq_len(1e2))
    all.equal(dt1, dt1)
})
# before
#   user  system elapsed 
#  19.09    7.73   28.00 

# now
#   user  system elapsed 
#  15.23    5.46   21.61 
```

src/assign.c
tlapak: I've removed the PROTECT statements and also went ahead and removed the one on what is now line 174 for consistency. I considered just going through the entire file while I was at it, but that seemed out of scope for the PR.

mattdowle: I had the same thoughts and logic as you both. But `rchk` revealed that `getAttrib` can sometimes allocate inside it, perhaps more so now with ALTREP. Search for "A common source of true errors is a failure to protect the result of getAttrib when retrieving an attribute that may be automatically generated/converted (e.g. names, dimnames)" in https://github.com/kalibera/rchk/blob/master/doc/USAGE.md. Also search https://github.com/kalibera/rchk/blob/master/doc/INTERNALS.md for `getAttrib`.
I suspect in data.table's usage of the R API, we will never see `getAttrib` allocate, but R API is more general. So to pass `rchk` (as per steps in CRAN_Release.cmd, and as required by CRAN under additional checks) we have to protect getAttrib calls.
The variance in why some `getAttrib` calls are not protected in data.table, may be that `rchk` knows some cases of `getAttrib` do not need to be protected depending on what the 2nd argument is.
In the past I've gone through and protected any `getAttrib` calls that `rchk` spots until it passes.

src/assign.c
jangorecki: I made a dedicated issue to have it well discussed: https://github.com/Rdatatable/data.table/issues/4467

src/assign.c
tlapak: Very interesting, thanks for that. I don't mind reverting those changes at all. I just have two thoughts:
1) From reading the docs you linked, the seemingly inconsistent calls to `PROTECT` may be related to garbage collection only being triggered by allocating calls as not all instances of `getAttrib(*, R_NamesSymbol)` are currently protected.
2) As far as I can tell, the only instances where it does allocate are for rownames of a data.frame and with `R_NamesSymbol` when the first argument is a pairlist or a language object but then in the latter case it still calls `MARK_NOT_MUTABLE`, which I _thought_ protected the object from garbage collection.

mattdowle: 1. I'm not fully following there. But I did a grep and yes I see that you're right and not all `getAttrib(*, R_NamesSymbol)` are protected, for example in `rbindlist.c`. It could be that those will be picked up when I rerun `rchk` before release (perhaps those lines were added or changed in dev since 1.12.8),  or more likely, from prior discussions with Thomas I suspect that `rchk` looks afterwards to the usage of the unprotected `getAttrib`. If there is no possible GC between the getAttrib and its last usage, or if the result is protected by an assign using `setAttrib`, then `rchk` is clever enough (I guess) to not raise it. It does an extreme amount of tracking to spot unbalanced protection, for example, all statically (without runtime tests) so I've learnt not to underestimate how advanced it is.
2. That's one case.  But isn't it possible too that column names are also some kind of ALTREP object of 1,000,000 columns, where even column names are not materialized either as well as the column data. Even `STRING_ELT` can allocate if the vector it is passed is an ALTREP.

src/assign.c
mattdowle: This absolutely needs to be a deep copy, iiuc. That's what a comment in the code was trying to convey. Consider when a user uses `setnames()` to change a column's name, and that column is involved in a key or index. `setnames` looks to see if that column is involved in the key or any indexes, and if so updates the corresponding name in the key or index name. If it wasn't a deep copy of names, that change-of-column-name in the copy would break the original table.  Better is to always shallow copy and put a flag on it so that the next update knows to deep copy, however we didn't get to that yet.  Plus full reference counting in R came along which we may be able to tie in with rather than do our own thing.

src/assign.c
mattdowle: This `INTSXP` being copied here is an empty one though (the dummy object upon which the indexes are attached as attributes). So I'm not sure what's improper here in the PR as it stands now. @jangorecki Do you mean that it's possible that `shallow_duplicate()` might not work like this in future in R on this case of being passed INTSXP? If so, that's a good spot but on balance it feels right to use it here for this. And the tests in CI will spot if and when R-devel makes any change to `shallow_duplicate()`.

jangorecki: Thanks Matt. Good spot. Sorry for confusion. Indeed copy of `integer()` (a placeholder to which we are attaching indices) is not an issue here. The issue is copying of indices itself, which is not happening here. So the related items I was linking here are not really matching use case here.

PR #4442, opened 2020-05-12
dll name uses underscore

jangorecki:
This PR changes dll (or `so`) name from `datatable.dll` to `data_table.dll`. It is to be consistent on how `tools::package_native_routine_registration_skeleton` handles dot in a package name.

I was able to further simplify it to avoid use of `data_name` and just use `data.table` instead. Moreover, it was possible to remove dll renames from `Makevars.[in|win]`. It did work and pass pkg check, but it is against WRE:
> DLL needs to be both a valid file name and valid as part of a C entry point (e.g. it cannot contain ‘.’)

----

Minimal working example below (it produces, and successfully uses `r.gitlab.ci.[so|dll]`):

[`src/init.c: void R_init_[pkgname]`](https://github.com/jangorecki/r.gitlab.ci/blob/a2c48a357c8578f232a2d6c5c40dfff6e3dfe53d/src/init.c#L13)
```c
void R_init_r_gitlab_ci(DllInfo *dll)
```
[`NAMESPACE: useDynLib`](https://github.com/jangorecki/r.gitlab.ci/blob/a2c48a357c8578f232a2d6c5c40dfff6e3dfe53d/NAMESPACE#L1)
```
useDynLib(r.gitlab.ci, .registration=TRUE)
```
[`R/onLoad.R: library.dynam.unload`](https://github.com/jangorecki/r.gitlab.ci/blob/a2c48a357c8578f232a2d6c5c40dfff6e3dfe53d/R/hello.R#L4-L6)
```r
.onUnload = function(libpath) {
  library.dynam.unload("r.gitlab.ci", libpath)
}
```


R/onLoad.R
```diff
@@ -19,13 +19,12 @@
   # Runs when loaded but not attached to search() path; e.g., when a package just Imports (not Depends on) data.table
   if (!exists("test.data.table", .GlobalEnv, inherits=FALSE)) {
     # check when installed package is loaded but skip when developing the package with cc()
-    dllV = if (is.loaded("CdllVersion",PACKAGE="datatable")) .Call(CdllVersion) else "before 1.12.0"
-    #                                              ^^ no dot as this is the name of the dll file, #3282
```
MichaelChirico, line 23: @jangorecki what of the referenced issue?

> Local testing on Ubuntu shows there has to be no dot in the package name (that argument is really the name of the dll/so which has the dot removed).

MichaelChirico:
basically LGTM, checking for other usage (still needs fine-tuning)

```
grep -Er "[^'\"$]\bdatatable\b[^-]" ~/github/data.table --exclude=*.{git,o,so,Rproj,Rraw,po,mo,md,Rmd,yml,Rd}* --exclude=*{onLoad,test.data.table}.R | grep -Ev "datatable[.](?:optimize|alloccol|verbose|showProgress)"
```

jangorecki:
Appveyor build is failing (R 4.0), GitLab windows build is passing (R 3.6)

R/onLoad.R
jangorecki: according to documentation it should be a dll name, so `data_table` now, but using that leads to failure linux
```
Error: package or namespace load failed for 'data.table':
 .onLoad failed in loadNamespace() for 'data.table', details:
  call: NULL
  error: invalid version specification 'before 1.12.0'
```
which is caused by the fact that it still uses `data.table/libs/data.table.so`.

Ultimately I think that R should raise and exception not allowing to create such dll because it is against WRE
> DLL needs to be both a valid file name and valid as part of a C entry point (e.g. it cannot contain ‘.’)

so looks like we need Makevars to rename dll

mattdowle:
> It is to be consistent on how tools::package_native_routine_registration_skeleton handles dot in a package name.

I looked at `?package_native_routine_registration_skeleton` but couldn't see anything about `.` and `_` documented there. So I started wondering if we should be changing something like this based on current behaviour of an R function that isn't documented.

But then I looked at the full WRE paragraph mentioned :

> Note that there are some implicit restrictions on this mechanism as the basename of the DLL needs to be both a valid file name and valid as part of a C entry point (e.g. it cannot contain ‘.’): for portable code it is best to confine DLL names to be ASCII alphanumeric plus underscore. **If entry point R_init_lib is not found it is also looked for with ‘.’ replaced by ‘_’.**

Where the last sentence (my bolding) convinces me it is worth changing to be consistent with.

PR #4445, opened 2020-05-14
gmin/gmax properly support integer64

### LINKED ISSUE #4444: Not ignoring NAs with 64-bit integers and grouped operations?
## Problem

When I attempt to find the grouped minima of a 64-bit integer (using `na.rm = T`), it appears as though `data.table` is ignores the `na.rm = T`. This does not happen for other grouped operations on 64-bit integers (*e.g.*, `max` or `sum`).

## Minimal reproducible example

```r
# Load packages
library(bit64)
library(data.table)
# Create data table
tmp = data.table(x = as.integer64(c(1, 2, NA)), id = 1)
# Grouped min does not work
tmp[, min(x, na.rm = T), by = id]
# Grouped max does work
tmp[, max(x, na.rm = T), by = id]
```

**Output:**

```
> tmp[, min(x, na.rm = T), by = id]
   id   V1
1:  1 <NA>
> # Grouped max does work
> tmp[, max(x, na.rm = T), by = id]
   id V1
1:  1  2
```

## Output of `sessionInfo()`

```r
R version 3.6.3 (2020-02-29)
Running under: macOS Catalina 10.15.4

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] bit64_0.9-7       bit_1.1-14        data.table_1.12.9

[1] compiler_3.6.3 tools_3.6.3   
```

MichaelChirico:
Closes #4444


src/gsumm.c
```diff
@@ -761,28 +761,62 @@ SEXP gmin(SEXP x, SEXP narm)
...
+        for (i=0; i<n; i++) {
+          thisgrp = grp[i];
+          ix = (irowslen == -1) ? i : irows[i]-1;
+          if (xp[ix] == NA_INTEGER64) continue;
+          if (ansp[thisgrp] == NA_INTEGER64 || xp[ix] < ansp[thisgrp])
+            ansp[thisgrp] = xp[ix];
+        }
+        for (i=0; i<ngrp; i++) {
+          if (ansp[i] == NA_INTEGER64) {
+            warning(_("No non-missing values found in at least one group. Returning 'MAX_INTEGER64' for such groups to be consistent with bit64::max.integer64"));
```
MichaelChirico, line 778: Maybe better to do `Returning '%d'` and substitute `MAX_INTEGER64` but I can never remember the format code for this input type...

MichaelChirico: nvm, done

src/gsumm.c
```diff
@@ -896,39 +930,62 @@ SEXP gmax(SEXP x, SEXP narm)
     break;
   case REALSXP:
```
MichaelChirico, line 931: There's a `TODO` above:

> // TODO rework gmax in the same way as gmin and remove this *update

I _partially_ addressed that in this PR, only for the `REALSXP` case

inst/tests/tests.Rraw
```diff
@@ -17435,3 +17435,22 @@ exid = data.table(id=1, expected)
...
+  DT = data.table(grp=c(1L, 1L, 1L, 2L), i64=as.integer64(c(NA, 1:3)))
+  old = options(datatable.optimize=2L)
+  test(2183.1, DT[ , min(i64), by=grp], data.table(grp=1:2, V1=as.integer64(c(NA, 3))))
+  test(2183.2, DT[ , min(i64, na.rm=TRUE), by=grp], data.table(grp=1:2, V1=as.integer64(c(1, 3))))
+  test(2183.3, DT[ , max(i64), by=grp], data.table(grp=1:2, V1=as.integer64(c(NA, 3))))
+  test(2183.4, DT[ , max(i64, na.rm=TRUE), by=grp], data.table(grp=1:2, V1=as.integer64(2:3)))
+  # force an all-NA group
+  DT[ , i64 := rev(i64)]
+  i64_default_min = suppressWarnings(min(as.integer64(NA), na.rm=TRUE))
+  i64_default_max = suppressWarnings(max(as.integer64(NA), na.rm=TRUE))
```
jangorecki, line 17450: I would add tests for that as well so in case bit64 will change this behavior we will be notified

MichaelChirico: good idea. actually it turns out they _aren't_ warning, which differs from the equivalent base behavior. So the test is a bit wonky, WDYT?

NEWS.md
```diff
@@ -86,7 +86,11 @@
 
 13. `melt` with a list for `measure.vars` would output `variable` inconsistently between `na.rm=TRUE` and `FALSE`, [#4455](https://github.com/Rdatatable/data.table/issues/4455). Thanks to @tdhock for reporting and fixing.
 
+<<<<<<< HEAD
 14. `GForce` handles `integer64` input correctly for `min` and `max` with missing values, [#4444](https://github.com/Rdatatable/data.table/issues/4444). Thanks @go-see for the report.
+=======
```
jangorecki: @MichaelChirico

PR #4462, opened 2020-05-20
convenience macros for char->digit in fread

MichaelChirico:
Minor PR. Noticed this used a lot, hopefully the code is more readable with these macros.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -103,7 +103,8 @@ static void Field(FieldParseContext *ctx);
 #define ASSERT(cond, msg, ...) \
   if (!(cond)) STOP(_("Internal error in line %d of fread.c, please report on data.table GitHub:  " msg), __LINE__, __VA_ARGS__) // # nocov
 
-
+#define AS_DIGIT(x) (uint_fast8_t)(x - '0')
+#define IS_DIGIT(x) AS_DIGIT(x) < 10
 
 //=================================================================================================
 //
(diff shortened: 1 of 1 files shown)


src/fread.c
```diff
@@ -910,7 +911,7 @@ static void parse_double_hexadecimal(FieldParseContext *ctx)
     acc <<= (13 - ndigits) * 4;
     ch += 1 + (Eneg = ch[1]=='-') + (ch[1]=='+');
     uint64_t E = 0;
-    while ((digit = (uint8_t)(*ch-'0')) < 10) {
+    while ((digit = AS_DIGIT(*ch)) < 10) {
```
MichaelChirico, line 914: note the difference of `uint8_t` --> `uint_fast8_t`. I assume `uint8_t` was unintentional

src/fread.c
```diff
@@ -943,7 +944,7 @@ static void parse_bool_numeric(FieldParseContext *ctx)
 {
   const char *ch = *(ctx->ch);
   int8_t *target = (int8_t*) ctx->targets[sizeof(int8_t)];
-  uint8_t d = (uint8_t)(*ch - '0');  // '0'=>0, '1'=>1, everything else > 1
+  uint_fast8_t d = AS_DIGIT(*ch);  // '0'=>0, '1'=>1, everything else > 1
```
MichaelChirico, line 947: same here: `uint8_t` --> `uint_fast8_t`

Issue #3160, opened 2018-11-26
IDateTime could optionally track different time zones for different rows


MichaelChirico:
simple `base` R time objects (`POSIXct`) are ill-suited to handling regional datasets (`tzone` attribute is atomic); similarly, `POSIXlt` objects are overly broad (_every_ component is a vector).

What's really needed is a structure that's just a little bit more general than that offered by `IDateTime` at the moment:

```
idate itime tz
17861 32415  1
17861 32415  2
```

Where I'm suppressing the `print` methods for each column to show the simple integer structure of each column.

`idate` and `itime` are as before, but the objects (optionally? `NA` by default?) gain a `tz` column stored as a `factor` whose labels name different `OlsonNames()` time zones, e.g.

```
tz = factor(c('Asia/Singapore', 'Asia/Jakarta'))
```

Related methods would then handle printing/formatting/etc appropriately accounting for the `tz` column.

Related: #2402, #1451

MichaelChirico:
So far I've been doing things `by = tz`; "producing a report" is exactly the sort of functionality I have in mind, but also, filtering rows based on local time (`x[hour_local == 9L]`, e.g.).

Functionality I have in in mind handles this logic internally to be less of a fuss. I'll add `Low` label since it's a nice-to-have.

mattdowle:
If there are mixed timezones in an input dataset, it's best practice to convert everything to consistent UTC up front. Leaving them mixed in the same column is going to lead to bugs really quickly, slow it down even if it works, and at best result in a new class that can't be passed to other packages.
If the use case is really to do something like `x[hour_local == 9L]` then a column can be added for that using plain `integer`, or `ITime` or similar.  In other words users don't have to use `POSIXct`. I most often prefer two columns anyway: date and time separately. That way grouping by date/month/day-of-week need only touch the smaller 4 byte date column,  rolling joins are a bit easier to find first and last record by day and don't join to the last row of the previous day (penultimate column, date, must match). Most often in time series we're doing time deltas within the day, not across days.
I'm just trying to say there other better ways to represent varying local time zone data, if that is really needed, than attempting to incorporate a vector-timezone into POSIXct. Those ways are already possible without needing to add anything into data.table.
If you agree, can we close this issue?

MichaelChirico:
[...]

Issue #1162, opened 2015-05-27
[question] when will sep2 in fread be implemented?


zachmayer:
I can't wait :grinning:

dselivanov:
@mattdowle , @arunsrinivasan , do you have any suggestions for implementation of this feature? I will try to work on it.

zx8754:
:+1:

zachmayer:
I think the idea is it will be something like this:
```{R}
1:87595672-88977120
1:92704584-94054584
dt <- fread("myFile.txt", sep = ":", sep2 = "-")
> dt
   V1                V2
1:  1 87595672,88977120
2:  1 92704584,94054584
```

dselivanov:
@zx8754, no. For your case you can replace second type of delimiter in shell and use `fread` as usual. See for example this [thread](http://askubuntu.com/questions/20414/find-and-replace-text-within-a-file-using-commands)

l-jaye:
hello! Just wondering if there is an update for sep2? is it currently available? would love to use it!

MichaelChirico:
update: https://stackoverflow.com/questions/49061517/convert-string-to-list-and-nest-in-data-table

kprimice:
Any update on sep2?

mattdowle:
Contributions welcome!
Generally to potential contributors:  I understand that `fread.c` is one of the hardest pieces of C to work on (for example), especially now that it's multi-threaded.  But I would have thought that would make it _more_ attractive to learn C and contribute to, for the very reason it is impressive; e.g. enhancing your CV.

jangorecki:
@MichaelChirico it is much easier to answer particular questions about the code than writing generally enough so it can be followed by readers, and low-level enough to be useful.
It is much easier to experiment with C code when you are equipped with good tools. `cc()` and `dd()` are very useful.

jangorecki:
when implemented post an answer in https://stackoverflow.com/questions/65141208/read-csv-five-with-array-cols-in-r

PR #4464, opened 2020-05-20
Add support for native parsing of iso8601 dates/timestamps in fread

MichaelChirico:
Very common use case for me -- `fread` string timestamps in ISO8601 format like `2020-05-01T03:14:18.343Z`.

To do so, treat the parsing as a series of subparsing jobs -- `StrtoI32` for year, month, day, hour, minute, timezone hour offset, timezone minute offset; then `parse_double_regular` for seconds. Hence splitting off the "core" of those functions as `_core` functions, and leaving the original as wrappers to feed the `FieldParseContext` correctly.

Somewhat related in intent to #1656 (smoothing/speeding up date/time I/O), but that issue would make round trips even faster (no conversion to character on output, and "simple" numeric parsing on input)

TODO:
 - [x] Also parse timezone offsets like `+07:00` `+0630`
 - [x] Return `class`'d object, not just underlying value
 - [x] Augment `colClasses` API to support type overrides from user
 - [x] Add NEWS, documentation
 - [x] tests
 - [x] benchmark
- [x] add option to restore old behaviour for those who updgraded and need a quick fix without downgrading
- [x] `fread("a,b\n2015-01-01,2015-01-01", colClasses="POSIXct")` reads as character because direct POSIXct requires time to be present (use verbose=TRUE to see the bumps from type A to B to C).  Relax direct POSIXct parser to allow date-only.
- [x] I don' think test 2150.15 should be passing, but it is because something in our test environment (test.data.table()?) appears to be setting locale timezone to UTC (more investigation needed).  Done (it was prior test 2124) : e89d159411f9cd179000183c671efe6a0cb1decc
- [x] the real potential break with previous versions of data.table, is that for those using `colClasses="POSIXct"` or `as.POSIXct()` after the fread call, datetime values are now interpreted as UTC, not local timezone as `as.POSIXct` does by default. The news item gives focus to the tzone attribute of the resultant POSIXct column, but for those usages, there will be a silent shift in interpretation; i.e. they might be pleased after upgrade that they had POSIXct before and they still have POSIXct (just much faster) with no code changes, but the datetimes will have shifted. For those using `fasttime` after their `fread` calls, there will be no shift. In short, we need to convey the *interpretation* of datetime values will change for those using colClasses="POSIXct". Perhaps a warning for calls to fread which have `colClasses=POSIXct`, or a way to create a warning so those calls can be controlled. Or, extend the parser to interpret the datetime in local timezone, so that there is no break from previous versions of data.table, nor base R defaults. Sadly, I know that's hard. `fasttime` doesn't do local time for example, which may be why it was never promoted to R.  Update: now reads UTC-marked datetime as written by fwrite by default, and unmarked datetime still get read by as.POSIXct in local time.
- [x] last 3 test fails. All relate to date-only being read as POSIXct by colClasses="POSIXct"


jangorecki:
this is very related to #1656 
does that iso number is the same as "unix epoch" time?

MichaelChirico:
You're right this doesn't close #1656 _per se_, it's just very related.

#1656 wants to `fwrite` to use the underlying numeric representation of `POSIXct` columns, and `fread` offer a way for users to signal "this numeric column, apply `POSIXct` at the end". It's notable I think that this one basically _cannot_ be auto-detected (unless we want to turn into Excel and produce a ton of false positives 😃).

This PR offers a different way of producing `POSIXct` columns from `fread` via string parsing.

Would need to think a bit more about how to implement both at the same time...

MichaelChirico:
The ISO part refers to the string formatting:

https://en.wikipedia.org/wiki/ISO_8601

IINM the relation to `unixtime` (=epoch time) is an implementation detail.

Same in intent to [Python's `datetime.fromisoformat`](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) or [Presto's `from_iso8601_timestamp`/`from_iso8601_date`](https://prestodb.io/docs/current/functions/datetime.html). SparkQL's `timestamp()` `CAST` wrapper also recognizes this format.

inst/tests/tests.Rraw
```diff
@@ -8609,6 +8609,8 @@ if (test_R.utils) {
 # fix for #1573
 ans1 = fread(testDir("issue_1573_fill.txt"), fill=TRUE, na.strings="")
 ans2 = setDT(read.table(testDir("issue_1573_fill.txt"), header=TRUE, fill=TRUE, stringsAsFactors=FALSE, na.strings=""))
+date_cols = c('SD2', 'SD3', 'SD4')
+ans2[ , (date_cols) := lapply(.SD, as.IDate), .SDcols = date_cols]
```
MichaelChirico, line 8613: first broken old test -- `%Y-%m-%d` columns in this file are now parsed as `IDate`

inst/tests/tests.Rraw
```diff
@@ -10720,7 +10722,7 @@ test(1743.08, sapply(fread("a,b,c\n2017-01-01,1,1+3i", colClasses=c("Date", "int
 test(1743.09, sapply(fread("a,b,c\n2017-01-01,1,1+3i", colClasses=c("Date", "integer", "complex")), class), c(a="Date", b="integer", c="complex"))
 test(1743.10, sapply(fread("a,b,c,d\n2017-01-01,1,1+3i,05", colClasses=c("Date", "integer", "complex", NA)), class), c(a="Date",b="integer",c="complex",d="integer"))
 test(1743.11, sapply(fread("a,b,c,d\n2017-01-01,1,1+3i,05", colClasses=c("Date", "integer", "complex", "raw")), class), c(a="Date",b="integer",c="complex",d="raw"))
-test(1743.12, x = vapply(fread("a,b\n2015-01-01,2015-01-01", colClasses = c(NA, "IDate")), inherits, what = "IDate", FUN.VALUE = logical(1)), y = c(a=FALSE, b=TRUE))
+test(1743.12, x = vapply(fread("a,b\n1+3i,2015-01-01", colClasses = c(NA, "IDate")), inherits, what = "IDate", FUN.VALUE = logical(1)), y = c(a=FALSE, b=TRUE))
```
MichaelChirico, line 10725: I'm not quite clear the purpose of this test. But before, `colClasses=NA` led the first column type to be guessed; now it is correctly guessed as `IDate`, but the test expected `!inherits(., 'IDate')`

inst/tests/tests.Rraw
```diff
@@ -10771,9 +10773,9 @@ test(1743.241, fread("a,b,c\n2,2,f", colClasses = list(character="c", integer="b
 test(1743.242, fread("a,b,c\n2,2,f", colClasses = c("integer", "integer", "factor"), drop="a"), data.table(b=2L, c=factor("f")))
 
 ## POSIXct
-test(1743.25, fread("a,b,c\n2015-06-01 11:00:00,1,ae", colClasses=c("POSIXct","integer","character")), data.table(a=as.POSIXct("2015-06-01 11:00:00"),b=1L,c="ae"))
+test(1743.25, fread("a,b,c\n2015-06-01 11:00:00,1,ae", colClasses=c("POSIXct","integer","character")), data.table(a=as.POSIXct("2015-06-01 11:00:00", tz='UTC'),b=1L,c="ae"))
```
MichaelChirico, line 10776: three more tests broken because I forced `tz='UTC'` in the implementation.

this is making me reconsider that -- should we default to not setting `'tzone'`?

MichaelChirico:
Not really sure what to do for the `colClasses` API come to think of it. Drawing a blank...

inst/tests/tests.Rraw
jangorecki: but you changed fread input, no? maybe that change should not be here but in another PR, no?

MichaelChirico:
Here's a benchmark:

```
library(data.table)
library(fasttime)
fastIDate = function(x) as.IDate(fastPOSIXct(x))

# comparison to fasttime:
#  (1) truncates down (milliseconds not supported)
#  (2) doesn't support TZ offsets
#  (3) doesn't support times before 1970-01-01

NN = 1e7
rdate = format(.Date(sample(-20000:20000, NN, TRUE)))
rts = .POSIXct(runif(NN, -2*1590029545, 2*1590029545), tz = 'UTC')
rtz = sample(OlsonNames(), NN, TRUE)
rtime_iso8601 = format(rts, format = '%FT%H:%M:%OSZ')
rtime_utc = format(rts, format = '%FT%T%z')

DT = data.table(rdate, rts, rtime_iso8601, rtime_utc, rtz)
DT[ , by = rtz, rtime_mixed := format(rts, format = "%FT%T%z", tz = .BY$rtz)]

f = tempfile()

fwrite(DT[ , 'rdate'], f)
  times = 10L,
  fromChar = as.IDate(rdate),
  fasttime = fastIDate(rdate),
  fread = fread(f)$rdate
)
# Unit: milliseconds
#      expr                       min                        lq                      mean
#  fromChar 35882.7227990000028512441 39291.1388519999964046292 42673.7721455999999307096
#  fasttime  2081.4537019999997937703  2324.4893710000001192384  4764.3731585999994422309
#     fread   374.9090039999999817155   393.2138810000000148648   609.2863710000000310174
#                     median                        uq                      max neval
#  43118.8744290000031469390 44802.1115659999995841645 50509.715732999997271691    10
#   4846.0090810000001511071  5695.4341949999998178100  9010.839712999999392196    10
#    409.9370655000000169821   848.5971970000000510481  1055.998761000000058630    10

  times = 10L,
  fromChar = as.IDate(rdate),
  fasttime = fastIDate(rdate),
  fwrite_fread = {fwrite(data.table(rdate), f); fread(f)$rdate}
)
# Unit: milliseconds
#          expr                       min                        lq                      mean
#      fromChar 35845.2031050000005052425 39328.8093050000024959445 39655.5725804999965475872
#      fasttime  2132.8088490000000092550  4906.1557210000000850414  5918.7371296999999685795
#  fwrite_fread   879.3674630000000433938   914.6356719999999995707   984.0583596999999826949
#                     median                       uq                      max neval
#  39571.6334754999988945201 40353.395442000000912230 42844.634253999996872153    10
#   6160.8297330000004876638  6443.078268999999636435 12707.056249999999636202    10
#    943.1287370000000009895  1004.976440000000025066  1227.542063000000098327    10

fwrite(DT[ , 'rtime_iso8601'], f)
  times = 10L,
  fromChar = as.POSIXct(rtime_iso8601, format = '%FT%H:%M:%OSZ', tz = 'UTC'),
  fasttime = fastPOSIXct(rtime_utc, tz = 'UTC'),
  fread = fread(f)$rtime_iso8601
)
# Unit: milliseconds
#      expr                       min                        lq                      mean
#  fromChar 53267.5157200000030570664 54396.0746129999970435165 57902.1551069999986793846
#  fasttime   374.4050359999999955107   407.7941289999999980864   438.6393107000000100015
#     fread   827.6846910000000434593   853.0744369999999889842  1561.1232625000000098225
#                     median                        uq                       max neval
#  57430.6017244999966351315 58453.2611570000008214265 69509.8057030000054510310    10
#    459.2325749999999970896   462.6702389999999809334   483.3714049999999815554    10
#    875.0970604999999977736  1743.7432160000000749278  5516.9704620000002250890    10

fwrite(DT[ , 'rtime_utc'], f)
  times = 10L,
  fromChar = as.POSIXct(rtime_utc, format = '%FT%T%z', tz = 'UTC'),
  fasttime = fastPOSIXct(rtime_utc, tz = 'UTC'),
  fread = fread(f)$rtime_utc
)
# Unit: milliseconds
#      expr                      min                       lq                    mean
#  fromChar 6461.8654249999999592546 6520.2935799999995651888 8031.751119100000323670
#  fasttime  400.2528889999999819338  424.2101149999999734064 1599.063912200000004304
#     fread  887.4787450000000035288  895.9816120000000410073 1001.675952400000028319
#                    median                        uq                      max neval
#  6779.6675299999997150735 10312.0697949999994307291 11352.484152000000904081    10
#   459.7312014999999973952  3933.7474739999997837003  4715.532877999999982421    10
#   942.4043510000000196669   966.9638159999999516003  1627.955359000000044034    10

parse_mixed_time = function(x) {
  utc = as.POSIXct(substr(x, 1L, 19L), format = '%FT%T', tz = 'UTC')
  hour_offset = as.numeric(substr(x, 20L, 22L))
  minute_offset = as.numeric(substr(x, 23L, 24L))
  utc - 3600*hour_offset - 60*minute_offset
}
fasttimeMixed = function(x) {
  utc = fastPOSIXct(substr(x, 1L, 19L), tz = 'UTC')
  hour_offset = as.numeric(substr(x, 20L, 22L))
  minute_offset = as.numeric(substr(x, 23L, 24L))
  utc - 3600*hour_offset - 60*minute_offset
}
fwrite(DT[ , 'rtime_mixed'], f)
  times = 10L,
  fromChar = parse_mixed_time(rtime_utc),
  fasttime = fasttimeMixed(rtime_utc),
  fread = fread(f)$rtime_mixed
)
# Unit: milliseconds
#      expr                       min                        lq                     mean
#  fromChar 14486.1231759999991481891 17218.5587549999982002191 19018.817376399998465786
#  fasttime  6913.8239929999999731081  7051.4129789999997228733  8682.411599700000806479
#     fread   947.8858159999999770662   982.3679939999999533029  1093.509279300000116564
#                     median                       uq                      max neval
#  17722.9039320000010775402 20225.369352000001526903 30088.444311999999627005    10
#   7434.6908979999998337007 10910.573624000000563683 11846.584786999999778345    10
#    997.5085490000000163491  1023.713062000000036278  1958.662417000000004919    10

# restriction to since 1970-01-01 [fasttime validity]
fwrite(DT[rdate >= .Date(0), 'rdate'], f)
  times = 10L,
  fromChar = as.IDate(rdate),
  fasttime = fastIDate(rdate),
  fread = fread(f)$rdate
)
# Unit: milliseconds
#      expr                       min                        lq                     mean
#  fromChar 34985.4997430000003078021 40414.9324960000012652017 42401.060834800002339762
#  fasttime  2408.1196479999998700805  6147.3126640000000406872  8354.270583600000463775
#     fread   192.6099480000000028213   206.9921659999999974389  1135.318660500000078173
#                     median                        uq                      max neval
#  42407.7967409999982919544 45023.2967720000015106052 49840.856562000000849366    10
#   8293.6175894999996671686 11095.8809689999998226995 12965.062056999999185791    10
#    439.6589885000000208493   580.2351089999999658176  4295.531719999999950232    10

fwrite(DT[rtime_iso8601 >= '1970-01-01', 'rtime_iso8601'], f)
  times = 10L,
  fromChar = as.POSIXct(rtime_iso8601, format = '%FT%H:%M:%OSZ', tz = 'UTC'),
  fasttime = fastPOSIXct(rtime_utc, tz = 'UTC'),
  fread = fread(f)$rtime_iso8601
)
# Unit: milliseconds
#      expr                       min                        lq                     mean
#  fromChar 48567.7034510000012232922 49745.8525749999971594661 52436.273207099999126513
#  fasttime   411.2232799999999883767   449.2160329999999817119  1047.987805499999922176
#     fread   412.2872570000000109758   427.8790230000000178734   636.538313300000027084
#                     median                        uq                       max neval
#  52135.1380439999993541278 53402.8323680000030435622 61943.1473170000026584603    10
#    457.8400750000000130058   493.2715469999999982065  3636.4618930000001455483    10
#    442.7219159999999646971   956.0641170000000101936   997.2641069999999672291    10

# restriction to single thread

setDTthreads(1)
fwrite(DT[ , 'rdate'], f)
  times = 10L,
  fromChar = as.IDate(rdate),
  fasttime = fastIDate(rdate),
  fread = fread(f)$rdate
)
# Unit: milliseconds
#      expr                       min                        lq                     mean
#  fromChar 32983.0609610000028624199 33186.8137110000025131740 35771.825230000002193265
#  fasttime  1869.0543990000001031149  2209.9826649999999972351  4563.082333800000014890
#     fread   361.3851700000000164437   372.4046319999999923311   619.459307200000012017
#                     median                        uq                      max neval
#  35033.6785275000002002344 37213.1049540000021806918 43132.509016999996674713    10
#   5490.6435404999992897501  5726.2085029999998369021  8833.638363000000026659    10
#    392.3812080000000150903   962.5612579999999525171  1035.250911999999971158    10

fwrite(DT[ , 'rtime_iso8601'], f)
  times = 10L,
  fromChar = as.POSIXct(rtime_iso8601, format = '%FT%H:%M:%OSZ', tz = 'UTC'),
  fasttime = fastPOSIXct(rdate, tz = 'UTC'),
  fread = fread(f)$rtime_iso8601
)
# Unit: milliseconds
#      expr                       min                        lq                      mean
#  fromChar 48383.1732800000027054921 49138.0654100000028847717 50472.0632050999993225560
#  fasttime   223.2523239999999873362   268.2423759999999788306   696.4542721999999912441
#     fread   768.1861649999999599459   805.1989630000000488508  1118.3323049000000537490
#                     median                        uq                      max neval
#  49871.5922500000015133992 52306.8768859999981941655 53327.665379999998549465    10
#    285.4133984999999711363   306.9988180000000284053  4484.410552000000279804    10
#    828.8447465000000420332  1596.4692489999999907013  1885.647050000000035652    10
```

MichaelChirico:
So in summary we blow `base` methods out of the water. `fasttime` is more competitive, but only for `POSIXct` output. I suspect the `as.IDate` step is the bottleneck there.

For `POSIXct` output, we do ~about the same when restricted to the range of valid `fasttime` inputs (1970-01-01 throught) -- `fasttime` "outperforms" on more general input, I assume because it just fails (returns `NA`) as soon as it detects a time outside that range.

`fasttime` is also not flexible enough to handle timezone offsets -- a wrapper that parses the offset & applies it slows down `fasttime` considerably.

inst/tests/tests.Rraw
MichaelChirico: Hmm shall I just remove the test?

Maybe the test is: `as(x, 'IDate')` gets applied because it was in `colClasses`, but the other column stays `character`. In which case a better test would be to re-tailor the test toward `complex` instead of `IDate`...

MichaelChirico:
Just realized `readr` offers their own native parsing (for some reason I thought it was just applying `as.Date` ex-post). Here are side-by-side benchmarks vs `readr`:

```
library(data.table)
library(readr)

NN = 1e7
rdate = format(.Date(sample(-20000:20000, NN, TRUE)))
rts = .POSIXct(runif(NN, -2*1590029545, 2*1590029545), tz = 'UTC')
rtz = sample(OlsonNames(), NN, TRUE)
rtime_iso8601 = format(rts, format = '%FT%H:%M:%OSZ')
rtime_utc = format(rts, format = '%FT%T%z')

DT = data.table(rdate, rts, rtime_iso8601, rtime_utc, rtz)
DT[ , by = rtz, rtime_mixed := format(rts, format = "%FT%T%z", tz = .BY$rtz)]

f = tempfile()

fwrite(DT[ , 'rdate'], f)
  times = 20L,
  readr = suppressMessages(read_csv(f, progress=FALSE)),
  fread = fread(f, showProgress=FALSE)
)
# Unit: milliseconds
#   expr       min        lq      mean    median        uq       max neval
#  readr 1386.4638 1438.7039 2034.6879 1545.2631 1655.4217 5036.6778    20
#  fread  106.1061  115.3648  142.5847  146.2171  158.1482  201.9532    20

fwrite(DT[ , 'rtime_iso8601'], f)
  times = 20L,
  readr = suppressMessages(read_csv(f, progress=FALSE)),
  fread = fread(f, showProgress=FALSE)
)
# Unit: milliseconds
#   expr       min        lq      mean   median        uq      max neval
#  readr 2514.8418 2587.5590 3105.4354 2667.628 2982.5975 7262.127    20
#  fread  223.4602  228.2907  379.0271  235.442  259.4958 2483.842    20

fwrite(DT[ , 'rtime_utc'], f)
  times = 20L,
  readr = suppressMessages(read_csv(f, progress=FALSE)),
  fread = fread(f, showProgress=FALSE)
)
# Unit: milliseconds
#   expr       min        lq      mean    median       uq       max neval
#  readr 3420.7929 3676.1125 4250.4507 3857.4463 4336.010 7842.4694    20
#  fread  248.9622  258.3923  295.5566  270.8553  313.008  434.5843    20

fwrite(DT[ , 'rtime_mixed'], f)
  times = 20L,
  readr = suppressMessages(read_csv(f, progress=FALSE)),
  fread = fread(f, showProgress=FALSE)
)
# Unit: milliseconds
#   expr       min        lq      mean    median        uq       max neval
#  readr 3565.8941 3642.8500 4006.1577 3855.8094 4004.9712 7204.2305    20
#  fread  262.7315  275.2454  292.0794  280.8626  289.1907  380.9208    20

# restriction to single thread
setDTthreads(1)
fwrite(DT[ , 'rdate'], f)
  times = 20L,
  readr = suppressMessages(read_csv(f, progress=FALSE)),
  fread = fread(f, showProgress=FALSE)
)
# Unit: milliseconds
#   expr      min        lq      mean    median        uq      max neval
#  readr 1372.249 1385.0602 1415.0192 1395.2615 1409.3363 1590.724    20
#  fread  378.426  382.4896  536.4007  385.2831  389.0673 3295.760    20

fwrite(DT[ , 'rtime_iso8601'], f)
  times = 20L,
  readr = suppressMessages(read_csv(f, progress=FALSE)),
  fread = fread(f, showProgress=FALSE)
)
# Unit: milliseconds
#   expr       min        lq     mean    median       uq      max neval
#  readr 2562.9606 2597.7621 2837.967 2684.0372 3028.164 3725.742    20
#  fread  799.8739  813.6693 1336.382  907.6082 1123.258 5626.535    20
```

So "out of the box" (4 cores on my machine) this branch is about 10x faster. Single-threaded is still 3-5x faster.

MichaelChirico:
@statquant suggests we read timestamp columns as `nanotime` instead. In principle this is just a matter of converting the units to nanoseconds by multiplying everything by 1e9, being careful to keep things in the right range.

This could in principle be coupled with the `integer64` argument to `fread` -- return `nanotime` by default, or turn it off and return `POSIXct` if `integer64='double'`. Any thoughts?

jangorecki:
Any POSIXct fields in csv should not be automatically turned into nanotime, but if a time field has higher precision than POSIXct offers, then make sense to automatically turn it into nanotime. The same way we do for numeric/integer64. And using existing argument to control that behavior looks fine for me.

MichaelChirico:
It's the opposite right?

For other fields, we try int -> int64 -> float; `POSIXct` doesn't really have a possibility for `int` (unless we try to map `ITime` first? I'd say not in scope as of now).

`nanotime` has range of `-2^63` nanoseconds --> `2^63` nanoseconds:

```
.POSIXct(c(-1,1)*9223372036854775807/1e9, tz="UTC")
# [1] "1677-09-21 00:12:44.145223 UTC" "2262-04-11 23:47:16.854776 UTC"
```

i.e., we can represent nanoseconds exactly in that range -- if we encounter timestamps outside that range, then we bump to double.

statquant:
What I was suggesting was that columns with precision up to millisecond would be converted to POSIXct and sub-milli to nanotime. Maybe an option could specify the behavior, {"auto", "POSIX", "nanotime"} where last 2 would just convert to POSIXct or nanotime

inst/tests/tests.Rraw
jangorecki: isn't this going to be a breaking change? maybe we could relax that, and keep old behavior for coming release and change it in next release?

MichaelChirico: yes i agree. not even 100% sure we should change the behavior -- if so we need to be very explicit that it's a break with `base`

NEWS.md
```diff
@@ -4,6 +4,10 @@
...
...re returned as `data.table`'s `integer`-backed `IDate` class (see `?IDate`), and times are returned as `POSIXct` columns with time zone set to `UTC` (with offset indicated by `%z` applied). To display this column in the time zone of your choosing, recall that time zones in `POSIXct` vectors are just printing artifacts -- all `POSIXct` columns use numeric seconds since 1970-01-01 00:00:00 UTC as the underlying representation, and the time zone is only applied when `print` (or `format`) is invoked. As such, changing time zones can be done near instantaneously & with no copies by `setattr(tzcol, 'tzone', 'Asia/Yangon')` (or any other time zone available in `OlsonNames()`). Since this is technically a breaking change (i.e. existing code may depend on dates and times being read as character), an option is provided to restore the old behaviour (`options(datatable.old.fread.datetime.character=TRUE)`). Or, you may wish to simply downgrade to the old version until you can update your `fread` calls. In addition to convenience, `fread` is now significantly faster in the presence of date and time columns. Since this is a potentially breaking change, the minor version number is bumped: v1.13.0.
```
jangorecki, line 9: maybe we could set `datatable.old.fread.datetime.character=TRUE` as default and release as 1.12.10, and then after release change this default to `FALSE`?

jangorecki: This will also help users to prepare for migration if we mention now that this will be a breaking change from the next release. Users can now upgrade their code during 1.12.10 cycle, and not being stuck on 1.12.8 before they upgrade.

mattdowle: My gut feel is that would be taking on too much work (releasing to CRAN is a lot of work) due to people who can't just downgrade or set an option. The same people who don't read NEWS. Why do the majority have to have the inconvenience of having to set an option for something that is long-overdue and much needed, just because the minority upgrade production systems directly without testing first, find it too hard to downgrade, and won't even read the first item of the NEWS file to find out the option to set.
There will be problems, bugs and changes to defaults. By nudging straight away, though, we get through those faster. It's a .0 release (to convey be-aware) and the NEWS item is right up at the top in it's own section in capitals. And the change is long-overdue, much needed, and compelling.

R/fread.R
```diff
@@ -295,7 +295,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir())
              "complex" = as.complex(v),
              "raw" = as_raw(v),  # Internal implementation
              "Date" = as.Date(v),
-             "POSIXct" = as.POSIXct(v),
+             # "POSIXct" = as.POSIXct(v),  # v1.13.0 now reads POSIXct directly, #4464
```
jangorecki, line 298: doesn't that mean it will dispatch to `methods::as(v, new_class)`?

mattdowle: but it's already a POSIXct so as() will just return `v` straight back.  As with any explicit class where fread has already read that class directly.

mattdowle: on second thought, why was the as.POSIXct there in the first place; i.e. why didn't we leave it to `as(v, new_class)`? I seem to remember `as(v, "POSIXct")` is slower than `as.POSIXct()`. Perhaps to avoid method lookup.
If we did add the line back and cover it, iiuc, we'd need to construct a datetime field that `fread` did not convert to `POSIXct` directly, but could still be converted with `as.POSIXct`. Or, maybe a date column which fread reads directly as `IDate`, but user explicitly passes `colClasses=POSIXct` intending a date-only stored as POSIXct.

MichaelChirico: cc @HughParsonage , I seem to recall your PR started with using `as(., class)`, any insight on using `as.POSIXct` directly?

HughParsonage: It simply doesn't work: 

``` r
x <- "2018-05-01 15:55:01"
as(x, "POSIXct")
#> Error in as(x, "POSIXct"): no method or default for coercing "character" to "POSIXct"
as.POSIXct(x)
#> [1] "2018-05-01 15:55:01 AEST"
```

NEWS.md
mattdowle: CRAN revdeps results may well make the decision for us. If a lot of packages fail due to this change, we might have to give those packages more time by having this new feature off by default. The difference with CRAN packages are that checks are run with latest version and they don't have any choice about that.  Out in the wild is different because users have a choice to upgrade or not, downgrade, or use the new option.

R/fread.R
mattdowle: Thanks, @HughParsonage.
I tried putting back the as.POSIXct and covering it, but it turns out to be unreachable. For the reason that because POSIXct is now a recognized direct type, specifying `colClasses=POSIXct` for a date-only column (how I was trying to cover it), overrides the auto-detected IDate reader and bumps to use the POSIXct direct reader directly at C level before reading. (Which currently fails on date-only, and therefore bumps to character in out-of-sample, but that's a separate minor issue to fix which I added to the list at the top of this issue.)
So this switch should not contain any cases that have direct readers at C level. I'll add a comment about this to the code pointing here.

mattdowle: resolved by eaf2d57

inst/tests/tests.Rraw
mattdowle: In eaf2d57 I reinstated the original test but changed the result, kept your new test and added another.  So 1743.12 becomes 3: 1743.12[1-3]. Think that covers everything here.

MichaelChirico:
Fixed the last point and added two new test cases:

```
fread("a,b\n2015-01-01,2015-01-01", colClasses="POSIXct")
#             a          b
#        <POSc>     <POSc>
# 1: 2015-01-01 2015-01-01
```

also partial-Date partial-POSIXct columns are detected & read as POSIXct:

```
fread("a,b\n2015-01-01,1.1\n2015-01-02 01:02:03,1.2")
#                      a     b
#                 <POSc> <num>
# 1: 2015-01-01 00:00:00   1.1
# 2: 2015-01-02 01:02:03   1.2
```

eddelbuettel:
Hi all -- I am coming to this late as I was just poking around the GH repo to see when/if a new release may come and finally thought of looking at milestones...

Now, and I say that with the utmost respect for both all your coding chops, and general taste in API design which are both really well documented -- I think the parsing to UTC by default is not the right choice here.  There is a reason _e.g._ R parses by default to TZ if one is set (and tries fairly hard to get a TZ value).  I emulate that in `anytime()` and also offer alternates as `utctime()`.  The latter are almost never used.

A basic/naive user of `fread()` who hands a csv file to will expect the `format()`-ed time to look the same way.  Not with the UTC offset.

```r
R> format(as.POSIXct("2020-07-11 13:28:12"))
[1] "2020-07-11 13:28:12"
R> format(fasttime::fastPOSIXct("2020-07-11 13:28:12"))
[1] "2020-07-11 08:28:12"
R> 
```

mattdowle:
Thanks, @eddelbuettel. Yes that's the way I was leaning already and why I had added to the top comment: "Or, extend the parser to interpret the datetime in local timezone, so that there is no break from previous versions of data.table, nor base R defaults. Sadly, I know that's hard. fasttime doesn't do local time for example, which may be why it was never promoted to R."

mattdowle:
Regarding this that I put in top comment : 

> I don' think test 2150.15 should be passing, but it is because something in our test environment (test.data.table()?) appears to be setting locale timezone to UTC (more investigation needed).

It seems to be test 2124 and it's down to a material difference between TZ environment variable being unset vs empty, at least on my Linux box.

Test 2124 does this : 
```R
oldtz=Sys.getenv('TZ')
Sys.setenv(TZ='Asia/Jakarta') # UTC+7
...
Sys.setenv(TZ=oldtz)
```

and observe the following :

```R
$ R --vanilla
Sys.getenv('TZ',unset=NA)
# [1] NA
Sys.getenv('TZ')
# [1] ""
as.POSIXct("2010-05-15 01:02:03")
# [1] "2010-05-15 01:02:03 MDT"
Sys.setenv(TZ="")
as.POSIXct("2010-05-15 01:02:03")
# [1] "2010-05-15 01:02:03 UTC"
Sys.unsetenv("TZ")
as.POSIXct("2010-05-15 01:02:03")
# [1] "2010-05-15 01:02:03 MDT"
```

So if TZ was unset before that test, TZ will be set to `""` after the test. And `TZ='""'` appears to mean UTC, not local.
So that explains why 2150.15 is passing when it shouldn't: it's running under UTC due to the prior test setting TZ to `""`.
If a user runs test.data.table() when their TZ is unset, it'll be set to `""`, so that needs fixing anyway so that `test.data.table()` doesn't impact user environment.

MichaelChirico:
> extend the parser to interpret the datetime in local timezone

This gets to a broader point that I also got stuck at when looking into https://github.com/Rdatatable/data.table/issues/1162 (`sep2`) -- From a data structures point of view, how would one implement such a thing in `fread`? In particular I got stuck at making such infra R-agnostic so it can work with `pydatatable`.

As it relates to this issue, if we see `2020-01-01 01:02:03 Asia/Bahrain`, I guess the return type would have to be a `struct`? One piece the parsed time-in-seconds, one piece the time zone string?

In any case I don't think it's something that we should try to address for now.

eddelbuettel:
> how would one implement such a thing 

I don't know either but I have _e.g._ been wrapping Boost date_time for a decade or so (first in `RcppBDT`, now in `anytime`) and it simply does not parse timezones, see [its (long) docs page](https://www.boost.org/doc/libs/1_73_0/doc/html/date_time/date_time_io.html#date_time.format_flags) and scroll down about 1/6 or search for "(output only)".   :cry:

MichaelChirico:
More directly for the issue at hand, I am a bit of two minds about time zones. I recognize `base` is trying very hard to figure out the system time zone, but I have in mind reproducibility, with [what Kurt Hornik wrote about changing `stringsAsFactors`](https://developer.r-project.org/Blog/public/2020/02/16/stringsasfactors/index.html) in mind.

If the system time zone is used, co-authors in different locales can conceivably get different downstream results on their respective machines. Hopefully as innocuous as a shifter but I'm not sure how badly that could spiral out in a complicated analysis. Obviously it's ideal if everyone is fastidious about setting time zones if their doing time series stuff, but certainly R-core agrees this type of thing is a big enough threat for it to be a major reason to change `stringsAsFactors` default. Certainly all of our production data at Grab is using UTC timestamps for a reason 😃

Lastly, for ISO8601 timestamps like `2020-01-02T03:04:05+04:30`, that `+04:30` is clearly relative to UTC time. And Zulu time (`Z`) is UTC. The original thrust of this PR was to read ISO8601 timestamps which AFAICT are UTC-centric; it just happened that reading other timestamps was trivially easy to include at the same time, leading to the ambiguity here.

That's my two cents on why I went with UTC as a default. I agree with the need to anticipate more potentially breaking code, especially as base R doesn't (currently?) accept `as.POSIXct(<POSIXct>, tz='New/Timezone')` as a syntax for changing time zone. So code that was using `as.POSIXct` to a non-system time zone may also be broken.

MichaelChirico:
> it simply does not parse timezones

That's a bummer. When I started this PR I thought it would be out of scope and wanted to stick with Zulu-time (`Z`) only, but actually the standard `+mm:nn` and `+mm` versions are easy enough to handle. The real monster is parsing timezone names & figuring out what is/is not in `OlsonNames()` -- that could reasonably be left until the R level.

The next can of worms -- `POSIXct` only accepts one time zone per column. Hence https://github.com/Rdatatable/data.table/issues/3160. For another day...

mattdowle:
> As it relates to this issue, if we see 2020-01-01 01:02:03 Asia/Bahrain, I guess the return type would have to be a struct? One piece the parsed time-in-seconds, one piece the time zone string?

This almost never occurs.  We'll never support a string like Asia/Bahrain in the data like that.  UTC offsets is all we need, as you've implemented already.  The real issue is blank timezone; i.e. 2020-01-01 01:02:03.  That is interpreted as local timezone by R by default,  and as Dirk pointed out, that's what users expect and we need to maintain.

This is why `fwrite` only writes UTC datetimes, and it includes the Z character, unlike R's write.csv which writes timestamps in local time.

I'm nearly done with wrapping this up so it can be merged for this release.

MichaelChirico:
I think I missed something so just noting it here once I caught up.

Original proposal was UTC-dominant. Assumed any timestamp-alike on input was UTC time, so e.g.

```
2020-01-01 00:00:00
```

Is read as a time written in UTC, and parsed as UTC.

But maybe it is written in local time, then how should we parse? `fasttime` has a similar issue where it assumes every input string is UTC.

And in general if we don't know the timezone, it's hard to adjust ex-post anyway. Because of daylight savings, timezone jumps, etc, the UTC offset vs. "locale time" can be discontinuous. Instead the user has to rely on `as.POSIXct` to handle the heavy lifting in such cases.

I'm not even sure a `tz` argument to `fread` would help us here, unless we plan to shoehorn in all the complicated timezone logic at the parsing level -- the benefit is in allowing users to say `tz='UTC'` to force "fast parsing" on, then worry about time zone quirks on their own.

I do think it's common to find CSVs without timezone information there directly (e.g. [this benchmark](https://twitter.com/wesmckinn/status/1271810106921975811) of `arrow` on the NYC taxi data set -- my guess is their CSV reader has some built in assumptions about UTC time) so it's unfortunate we can't help in this case (as we've seen such a big performance improvement because loading all those unique strings chokes up the character cache). But hopefully production systems are producing ISO8601 timestamps (certainly `fwrite` is).

mattdowle:
> I'm not even sure a tz argument to fread would help us here
Read the NYC data with fread( , tz="UTC") and set TZ env variable to "UTC". Done. Work away in UTC but it will feel like whatever timezone the data was written.
If the NYC data was written using local time, then it should exhibit the gap and overlap on EST/EDT changes which would be a data quality exercise to perform, and a problem indeed since that dataset is continuous 24 hours. But that's nothing to do with R, unless it was R that was used to create the file and its default local time was used.

MichaelChirico:
> all that analysis and investigation can be done by reading the datetimes as UTC and going from there.

In that case, shouldn't we plan to set `tz='UTC'` by default (maybe not this release, but next release?)

mattdowle:
> In that case, shouldn't we plan to set tz='UTC' by default (maybe not this release, but next release?)
No, because of backwards compatibility with past data.table versions (colClasses=POSIXct using as.POSIXct), and expectations over matching base R default.  The R local time default is ok for people working in one time zone, saving their own data, and loading it up again themselves in their timezone. But for everyone writing and maintaining production systems, or working across timezones,  we set the TZ environment variable and work in UTC.  What we could do is default tz= to 'UTC" when the timezone of the R session is UTC.  In that case local time is UTC so reading the unmarked datetimes as UTC is the same as reading as UTC.   As this PR stands now,  as.POSIXct would still be used when TZ is UTC.

mattdowle:
Ok all tests and coverage pass. I'm done and ready to merge. Ok by you, @MichaelChirico?
We can add `tz=` to `fread` as a follow up. Since that's not too hard, I'd say in this release. Default tz="" (local time) and allow tz="UTC" only, not any other timezone strings.

MichaelChirico:
OK by me. BTW I would use `data.table:::is_utc` rather than checking `'UTC'` only

MichaelChirico:
What about `tz=Sys.timezone()` as the default? I know it's not an exact map to the default by at least it signals to the user where to look better than `''` does

mattdowle:
I see why `is_utc` is useful, but doesn't allowing so may alternatives open the door straight away to using a whole load of equivalent names? It's a new argument, so isn't it a good time to use one standard: UTC? Then in user code, when grep'ing for example, we know it's either tz="UTC" or tz="". Anything else would be an error explaining either "UTC" or "".
`Sys.timezone()` returns "America/Denver" for me.  I agree the call `Sys.timezone()` conveys better as a default, but its value leads to having to deal with timezone strings like that.  R uses tz=="" to mean local, so it's in keeping with that.  But we could use tz="local" for better clarity.    Or,  call the argument utc=FALSE/TRUE, maybe.

PR #4465, opened 2020-05-20
Remove 'typedef R_xlen_t RLEN' in data.table.h

jangorecki:
The comments there show it was about supporting R < 3.1.0 when R_xlen_t was introduced in R.
Since we now depend on R 3.1.0, this can be removed. In fact, it would have caused hard to track memory problems as we start to support long vectors.


mattdowle:
Nice find! Avoids some future pain.

PR #4469, opened 2020-05-21
keep.rownames=key="str" works for as.data.table.data.frame

### LINKED ISSUE #4468: as.data.table.data.frame keep.rownames & key
This works fine:

```
as.data.table(mtcars, keep.rownames = TRUE, key = 'rn')
```

But it's a bit arcane to know that `rn` is going to be the rownames' name. More readable might be:

```
as.data.table(mtcars, keep.rownames = 'model', key = 'model'
```

But that's an error:

```
Error in setkeyv(ans, key) : some columns are not in the data.table: model
```

Haven't looked yet but it must be setkeyv is applied before the setnames.

Note that this approach already works for `setDT`:

```
setDT(DT, keep.rownames = 'model', key = 'model')
```

MichaelChirico:
Closes #4468


mattdowle:
Good - coverage is back in PRs.

PR #4480, opened 2020-05-23
is.valid

jangorecki:
no manual yet, lets agree first on the current API and checks to be in scope.


jangorecki:
Waiting already 5 years in the queue without much attention, closing for now.

PR #4481, opened 2020-05-24
update and clarify := docs

### LINKED ISSUE #3908: := does not update by reference existing column if i is missing
I found a behavior that was not what I was expecting. When using `:=`  to update an existing column without setting `i` the column is copied. I have never seen that documented. In all the documentation (examples and vignettes) `:=` is always used to add a new column or to update using `i`.

If this behavior is expected I suggest to add this case in the examples and in the vignettes.

``` r
library(data.table)
DT = data.table(a = LETTERS[c(3L,1:3)], b = 4:7)
DT[, d := 9L]
address(DT$d)
#> [1] "0x56402b48d7b8"
DT[2, d := -8L]
address(DT$d)
#> [1] "0x56402b48d7b8"
DT[2, d := 10L]
address(DT$d)
#> [1] "0x56402b48d7b8"
DT[1:.N, d := d*2L]
address(DT$d)
#> [1] "0x56402b48d7b8"
DT[, d := d*2L] # I was not expecting that one
address(DT$d)
#> [1] "0x56402a0cae98"
```

<sup>Created on 2019-09-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jangorecki:
closes #3908


man/assign.Rd
```diff
@@ -63,7 +63,7 @@ For additional resources, please read \href{../doc/datatable-faq.html}{\code{vig
...
+Unlike \code{<-} for \code{data.frame}, the (potentially large) LHS is not coerced to match the type of the (often small) RHS. Instead the RHS is coerced to match the type of the LHS, if necessary. Where this involves double precision values being coerced to an integer column, a warning is given (when fractional data is truncated). It is best to get the column types correct up front and stick to them. Changing a column type is possible but deliberately harder: provide a whole column as the RHS. This RHS is then \emph{plonked} into that column slot and we call this \emph{plonk syntax}, or \emph{replace column syntax} if you prefer. By needing to construct a full length vector of a new type, you as the user are more aware of what is happening, and it is clearer to readers of your code that you really do intend to change the column type. When a column is \emph{plonked}, the original column is not updated by reference, because that needs to update every single element of that column.
```
mattdowle, line 66: good spot on the "whether or not" being wrong!

mattdowle (commented):
I'm only leaving a comment, but GitHub error says "I need to leave a comment when requesting changes". But I'm not requesting changes. Seems like a possible GitHub bug so I'm leaving this comment about the comment here as well as the actual comment to see if GitHub accepts it ...

mattdowle:
Coverage CI fail is just because I didn't merge main into this branch. Solely a `.Rd` doc change so no need.

Issue #856, opened 2014-10-03
Implement comment.char argument in fread


arunsrinivasan:
Similar to `read.table`.

mattdowle:
Update this question :  http://stackoverflow.com/questions/18920777/fread-data-table-in-r/

eddelbuettel:
Bump.  Needing this right now.

RLathamBAH:
+1

DavidArenburg:
Actually `fread` seem to assume it already has this implemented as it is mentioning `comment.char` in its warnings. This is a warning I've recently saw (using the dev 1.10.5 version)

> Warning in fread(file_x, skip = startind - 1L, header = TRUE, fill = TRUE) :
>   Stopped early on line 2. Expected 168 fields but found 235. Consider fill=TRUE and comment.char=.

eddelbuettel:
Interesting.  Can you chase the comment and find an author and commit per `git blame`?  
I am mostly using the CRAN version so I work around the issue (when I have to, which is not that often).

DavidArenburg:
@eddelbuettel You can find it here https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L2351

eddelbuettel:
Thanks.   Which one can click on for `git blame` so yield ... 
> Better skip= and nrow= (#2623)

by Matt just one day ago (!!)

map2085:
Bump

Berghopper:
Bump, needing this in package development. Working with very large data with an unusual format. comment selection would be great.

jangorecki:
@Berghopper consider using `fread(cmd="grep pattern file.csv")` till this feature in not available.

Berghopper:
@jangorecki Already made my own custom function for parsing and loading the file. It also used fread, yes. Thanks anyway however :).

bpbond:
Bump; would be super useful.

joost823:
Also bumping. Even though `fread(cmd='grep -v "#" table.csv')` works fine in general, it's not cross platform compatible and makes your code a bit harder to read.

ggirelli:
Bumping too. And if you want to skip only lines with hash at the beginning `cmd='grep -v "^#" table.csv'`.

cimbusch:
Yeah bumping... Would be great to have that feature!

ArthurPERE:
[...]

PR #4486, opened 2020-05-25
RFC/WIP for fread comment.char

MichaelChirico:
Very simple version of how we could implement a `comment.char` argument for `fread` (#856). (just testing with a fixed `comment.char='#'` for now)

This addresses line-initial comments, but not line-trailing comments:

```
fread('a,b
#a comment
1,2
#another comment
3,4')
#       V1    V2
#    <int> <int>
# 1:     3     4

fread('a,b#line-trailing comment
1,2')
#        a b#line-trailing comment
#    <int>                   <int>
# 1:     1                       2

fread('a,b
1,2#trailing after numeric')
#        a                        b
#    <int>                   <char>
# 1:     1 2#trailing after numeric
```

Is this good enough to file as a solution for now? How common is the line-trailing use case? Any efficiency concern of this approach when `comment.char` is not used?

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -247,6 +247,9 @@ static inline void skip_white(const char **pch) {
  */
 static inline bool eol(const char **pch) {
   const char *ch = *pch;
+  if (*ch == '#') {
+    while (*ch != '\r' && *ch != '\n') ch++;
+  }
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Superseded by #7375

Issue #4101, opened 2019-12-11
implicit type promotion in fcoalesce and fifelse


BenoitLondon:
Not sure I understand how the fcoalesce function is meant to be used but I found it not very user-friendly.

e.g I have a data.table coming from some computation or database and if one column is full of NA it will be of type Logical and then cant be used directly inside fcoalesce.

```
> fcoalesce(NA, 1L, 1.0)
Error in fcoalesce(NA, 1L, 1) : 
  Item 2 is type integer but the first item is type logical. Please coerce before coalescing.
> fifelse(TRUE, NA, 1.0)
Error in fifelse(TRUE, NA, 1) : 
  'yes' is of type logical but 'no' is of type double. Please make sure that both arguments have the same type.
```

2005m:
I think it is easy to do for `fifelse`. The question is do we want to do it?

jaapwalhout:
I think these error messages could be expected as `NA` is by default of class logical in R (see also `?NA`):

```
> class(NA)
[1] "logical"
```

Luckily you can force missing value to be of a certain type, by using `NA_integer_`, `NA_real_`, `NA_complex_` or `NA_character_`. Applying this knowledge to your `fifelse`-example, the following works:
```
> fifelse(TRUE, NA_real_, 1.0)
[1] NA
> fifelse(TRUE, NA_integer_, 1.0)
[1] NA
```

jangorecki:
I think that fixing data from the upstream provider is the most appropriate step, it doesn't have to be anything sophisticated.
Adding extra line to ensure class is matching
```r
dt = as.data.table(dbGetQuery("select * from users"))
dt[, `:=`(id=as.integer(id), name=as.character(name), rate=as.numeric(rate))]
myifelse(dt, ...)
```
Or a sketch of handling that more generally by a wrapper function having `colClasses`.
```
dbGetQuery2 = function(..., colClasses=NULL) {
  ans = as.data.table(dbGetQuery(...))
  if (is.null(colClasses)) return(ans)
  for (col in names(colClasses)) set(ans, i=NULL, j=col, value=do.call(paste0("as.", colClasses[[col]]), list(ans[[col]])))
  ans
}
dt = dbGetQuery2("select * from users", colClasses=c(id="integer", name="character", rate="numeric"))
myifelse(dt, ...)
```
This way any logical `NA` will go through `as.[class]` function.

----

[...]

Issue #3594, opened 2019-05-24
Extend functionality of nafill to use 'fill' argument for all 'type's


ben519:
Suppose I have this vector

```
x = c(NA,1,NA,NA,5,3,NA,NA)
```

and I want to fill NAs with the preceding non NA values. I can do this

```
nafill(x = c(NA,1,NA,NA,5,3,NA,NA), type = "locf")
[1] NA  1  1  1  5  3  3  3
```

Great, but sometimes I want to specify a fill value to catch the NA(s) at the front of the vector. I tried this which seemed obvious to me,

```
nafill(x = c(NA,1,NA,NA,5,3,NA,0), type = "locf", fill = -1)
```

but it didn't work and instead gave a warning, "argument 'fill' ignored, only make sense for type='const'".

My request is to extend the method so that 'fill' is applied to the front/back of the vector for types 'locf' and 'nocb' respectively. Thanks 

# Output of `sessionInfo()`

```
R version 3.6.0 (2019-04-26)
Platform: x86_64-apple-darwin18.5.0 (64-bit)
Running under: macOS Mojave 10.14.4

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /usr/local/Cellar/openblas/0.3.6/lib/libopenblasp-r0.3.6.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.12.3

loaded via a namespace (and not attached):
[1] compiler_3.6.0 tools_3.6.0
```

jangorecki:
Thanks. For now the workaround would probably be
```r
nafill(nafill(x = c(NA,1,NA,NA,5,3,NA,0), type = "locf"), fill = -1)
```

saraswatmks:
@jangorecki if no one else is working on this, can I take it up ?

MichaelChirico:
@saraswatmks I just assigned you, go for it!

jangorecki:
@saraswatmks note that nafill tests are in `inst/tests/nafill.Rraw`, so new tests should goes there. It can be useful, to avoid merge conflicts and to easily test only this script `test.data.table(script="inst/tests/nafill.Rraw")`

[...]

Issue #3992, opened 2019-10-25
nafill, setnafill for character, factor and other types


mik3y64:
`nafill` and `setnafill` only supports numeric types. 
```
library(data.table)

x = letters[1:10]
x[c(1:2, 5:6, 9:10)] = NA

nafill(x, "locf")
# Error in nafill(x, "locf") : 
# 'x' argument must be numeric type, or list/data.table of numeric types

setnafill(x, "locf")
# Error in setnafill(x, "locf") : 
#   'x' argument is atomic vector, in-place update is supported only for list/data.table
```

Both `zoo::na.locf` and `tidyr::fill` support all data types.

```
zoo::na.locf(x, na.rm = FALSE)
# [1] NA  NA  "c" "d" "d" "d" "g" "h" "h" "h"

tidyr::fill(tibble::as_tibble(x), value)
# A tibble: 10 x 1
#    value
#    <chr>
#  1 NA   
#  2 NA   
#  3 c    
#  4 d    
#  5 d    
#  6 d    
#  7 g    
#  8 h    
#  9 h    
# 10 h
```

Feature request for supporting character, factor and other types. It's useful in many cases in merging and cleaning data.tables/data.frames. Unlike matrix, data.frames/data.tables usually contain columns of arbitrary types, and it's very common for data.tables/data.frames to have `NA` values filled in columns of arbitrary types, especially characters. Right now I have to import the specific functions from other packages just for that functionalities and those functions are much slower than the ones created by `data.table`.

Thank you for creating this superior `data.table` package.

chinsoon12:
Maybe as a temp workaround:
```
DT <- data.table(X=c("A",NA,NA,"B",NA,"C"))
DT[, X := X[nafill(replace(.I, is.na(X), NA), "locf")]]
```

jangorecki:
Thanks @mik3y64 for the request. It is on the list to do, it was just not the part of the initial functionality. We have to first finish https://github.com/Rdatatable/data.table/pull/3765 then the logic here can re-use changes from that PR.
Thanks @chinsoon12 for interesting workaround.

mik3y64:
Great!! I am enjoying the performance of `data.table`. Totally understand the hard work in writing fast codes, unlike many other packages out there where they're mostly just some sort of "wrappers". Thank you very much `data.table` team for the awesome works. Thanks @chinsoon12 for the very interesting workaround. I would've never thought of doing it that way.

tdhock:
[...]

PR #4491, opened 2020-05-25
Internal class aware coerceAs. Already used in nafill and froll

jangorecki:
closes:
- [x] class-aware C coercion function, no-alloc coerce where possible #3913
- [x] use new `coerceAs` in nafill
- [x] use new `coerceAs` in froll
- [x] (**this is logically a distinct issue**) Extend functionality of nafill to use 'fill' argument for all 'type's #3594
- [ ] ~~as of now partially: implicit type promotion in fcoalesce and fifelse #4101~~
- [x] verbose message from `memrecycle` requires `verbose=3L` or higher
- [x] `memrecycle` no longer mentions column position and name when those are not provided
- [ ] ~~`coerceAs` will suppress warning from `memrecycle` related to valid coercion #4503~~


src/fifelse.c
```diff
@@ -16,17 +16,15 @@ SEXP fifelseR(SEXP l, SEXP a, SEXP b, SEXP na) {
...
-    } else {
-      error(_("'yes' is of type %s but 'no' is of type %s. Please make sure that both arguments have the same type."), type2char(ta), type2char(tb));
+    if (TYPEORDER(ta)<TYPEORDER(tb)) {
+      a = PROTECT(coerceAs(a, b, ScalarLogical(TRUE))); nprotect++;
+      ta = tb;
+    } else if (TYPEORDER(ta)>TYPEORDER(tb)) {
+      b = PROTECT(coerceAs(b, a, ScalarLogical(TRUE))); nprotect++;
+      tb = ta;
     }
+    if (TYPEOF(a)!=TYPEOF(b))
```
MichaelChirico, line 26: i feel really not necessary. 

```
  SEXPTYPE ta = TYPEOF(a);
  SEXPTYPE tb = TYPEOF(b);
```

is 2 lines before

src/utils.c
```diff
@@ -363,3 +308,53 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
...
+  // copyArg does not update in place, but only IF an object is of the same type-class as class to be coerced, it will return with no copy
+  if (!isVectorAtomic(x))
+    error("'x' is not atomic");
+  if (!isVectorAtomic(as))
+    error("'as' is not atomic");
+  if (isNewList(as))
+    error("'as' is a list");
+  int protecti=0;
+  bool copy = LOGICAL(copyArg)[0];
+  if (!copy && TYPEOF(x)==TYPEOF(as) && CLASSOF(x)==CLASSOF(as))
```
MichaelChirico, line 352: doesn't `CLASSOF` equality imply `TYPOF` equality?

src/fifelse.c
jangorecki: it ensure that `coerceAs` produces matched types, won't cost much

src/utils.c
jangorecki: yes but it is more expensive, this needs improvement anyway because any non handled class will be returned as just integer/double/etc. So two different (unhandled) classes will be interpretted as equal.

MichaelChirico: OK agree

inst/tests/nafill.Rraw
```diff
@@ -203,3 +204,127 @@ test(7.07, setnafill(DT, fill=0, cols=1L), copy(DT)[ , a := ans1])
...
+test(10.22, coerceAs(a, b), structure(1L, class="b"), output="integer[a] into integer[b]")
+a = structure(1, class="a")
+b = structure(2, class="b")
+test(10.23, coerceAs(a, b), structure(1, class="b"), output="double[a] into double[b]")
+a = structure(1, class="a")
+b = structure(2L, class="b")
+test(10.24, coerceAs(a, b), structure(1L, class="b"), output="double[a] into integer[b]")
+if (test_bit64) {
+  x = as.integer64(1L)
+  test(10.81, coerceAs(x, 1), 1, output="double[integer64] into double[numeric]")
```
MichaelChirico, line 280: can we add some tests with `NA` (coercing to/from `NA` in the int64 types)? I'm not sure `memrecycle` does the trick for these because there's no conversion `NA_INTEGER64 <-> NA_REAL` right?

jangorecki: AFAIR those are above near `# coerceAs int/numeric/int64 as used in nafill`

GreenGrassBlueOcean:
I was wondering if there are any plans to add this pull request to the upcoming data.table release? An improved nafill for character and boolean's  columns as mentioned in ([#3992](https://github.com/Rdatatable/data.table/issues/3992)) would be highly appreciated.

mattdowle:
The last bullet is ticked (memrecycle no longer warns on valid coercion #4503) and I commented in that issue. But when I ran the exhibit in #4503 it still warns, which I'm thinking is good.  But that would mean, as of now in this PR, #4503 is not actually done then iiuc?
```
> data.table(a=1:2, b=0L)[2L, b:="2"][]
       a     b
   <int> <int>
1:     1     0
2:     2     2
Warning message:
In `[.data.table`(data.table(a = 1:2, b = 0L), 2L, `:=`(b, "2")) :
  Coercing 'character' RHS to 'integer' to match the type of the target column (column 2 named 'b').
> 
```

mattdowle:
Can you also write a bit about why this PR it titled "class aware coerceAs, nafill".   The news item is about nafill now working when both fill= and type="locf" are supplied, where `fill` applies to the end NAs before the first non-NA, #3594. That's great and I see that in bullet item 4. But note the title of #3594 mismatches the contents of its first comment.  The contents of the first comment matches the news item for this PR though.  In any case, what's coerceAs got to do with nafill's fill= and locf=,  so that the supplied fill= can be coerced?

src/utils.c
```diff
@@ -374,6 +319,63 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
   return(ans);
 }
 
+// class1 is used by coerseAs only, which is used by frollR.c and nafill.c only
```
jangorecki, line 322: Note that there are multiple issues that are "scheduled" to be resolved by using `coerceAs`, so this list is likely to grow soon.

mattdowle: why can't we use existing coerceVector?

jangorecki: to handle int64 NAs

jangorecki: it is also explained in https://github.com/Rdatatable/data.table/issues/3913

jangorecki:
Yes, coerceAs is used to coerce `fill` argument to nafill and froll. Additionally it is used in froll to coerce `x` input to double.

mattdowle:
Have seen the missed coverage in new coerceAs. Sounds like future PRs will address that.

PR #4493, opened 2020-05-25
Non-equi joins are allow.cartesian = TRUE

### LINKED ISSUE #4489: non-equi join with .EACHI fails unexpectedly (seemingly related to values in join columns)
I have some time series with values. Apart from different values, the structure of the data sets are the same. A small example:

```
set.seed(2)
d = data.table(time = 1:8, v = sample(8))
d
#    time v
# 1:    1 5
# 2:    2 7
# 3:    3 6
# 4:    4 1
# 5:    5 8
# 6:    6 4
# 7:    7 2
# 8:    8 3
```

For each time, I want to determine the number of rows after the focal time where the values are larger than the focal value. E.g. for time 2 (value 7) there is one subsequent value which is higher (8 at time 5).   

There may be better ways to reach my goal, but here's the code I used and which gave me the desired result:

```
d[d, on = .(time > time, v > v), .N, by = .EACHI]
#    time v N
# 1:    1 5 3
# 2:    2 7 1
# 3:    3 6 1
# 4:    4 1 4
# 5:    5 8 0
# 6:    6 4 0
# 7:    7 2 1
# 8:    8 3 0
``` 

The code works for most data, but in some rare instances it errors. After digging back and forth, I thought I had found the reason: the error was triggered by setting the value 'v' equal to the time on two rows. E.g.:

```
d[time == 2L, v := 2L]
d[time == 7L, v := 7L]

d[d, on = .(time > time, v > v), .N, by = .EACHI, verbose = TRUE]

# This chunk of the message is same in the code above:

# i.time has same type (integer) as x.time. No coercion needed.
# i.v has same type (integer) as x.v. No coercion needed.
# Non-equi join operators detected ... 
# forder took ... forder.c received 8 rows and 2 columns
# 0.020s elapsed (0.000s cpu) 
# Generating non-equi group ids ... done in 0.000s elapsed (0.000s cpu) 
# Recomputing forder with non-equi ids ... Assigning to all 8 rows
# RHS_list_of_columns == false
# RHS for item 1 has been duplicated because NAMED==7 MAYBE_SHARED==1, but then is being plonked. length(values)==8; length(cols)==1)
# forder.c received 8 rows and 3 columns
# done in 0.000s elapsed (0.000s cpu)

# From here there are some differences in the message:
 
# Found 3 non-equi group(s) ...     # Above: Found 4 non-equi group(s) ...
# Starting bmerge ...
# forder.c received 8 rows and 2 columns
# bmerge done in 0.000s elapsed (0.000s cpu) 
# forder.c received 14 rows and 3 columns     # Above: forder.c received 11 rows and 3 columns
# Constructing irows for '!byjoin || nqbyjoin'

# Error in vecseq(f__, len__, if (allow.cartesian || notjoin || !anyDuplicated(f__,  : 
# Join results in 18 rows; more than 16 = nrow(x)+nrow(i).
# Check for duplicate key values in i each of which join to the same group in x over and over again.
# If that's ok, try by=.EACHI to run j for each group to avoid the large allocation.
```

However, if the value at time 5 instead of time 7 was changed, it works...

```
set.seed(2)
d = data.table(time = 1:8, v = sample(8))
d[time == 2L, v := 2L]
d[time == 5L, v := 5L]

d[d, on = .(time > time, v > v), .N, by = .EACHI, verbose = TRUE]
```
What's going on here? (most parsimonous guess: user error ;) I think I'm slightly confused by the error message. Although `Join results in 18 rows; more than 16 = nrow(x)+nrow(i).` is true for the join _per se_, I use `.EACHI` to aggregate the data to have _less_ rows than the `nrow(x)+nrow(i)` limit. In addition, `try by=.EACHI` is indeed a great suggestion, but I already use it. I clearly have misunderstood something here. 

--------

Tried on Windows, `R version 3.6.3`, `data.table 1.12.9 IN DEVELOPMENT built 2020-02-20`, and on Windows, `R version 4.0.0`, `data.table 1.12.8`

ColeMiller1:
Close #4489 

If any non-equi joins are detected ```c(">", ">=", "<", "<=", "!=")```, ```allow.cartesian``` is set to TRUE per Arun's suggestion. Debated including warning about this ```allow.cartesian``` change but for non-equi joins it seems like users should expect a one to many join.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -109,6 +109,8 @@ unit = "s")
 
 13. A relatively rare case of segfault when combining non-equi joins with `by=.EACHI` is now fixed, closes [#4388](https://github.com/Rdatatable/data.table/issues/4388).
 
+14. Non-equi joins now automatically set allow.cartesian to `TRUE`. Closes [4489](https://github.com/Rdatatable/data.table/issues/4489)
+
 ## NOTES
 
 0. Retrospective license change permission was sought from and granted by 4 contributors who were missed in [PR#2456](https://github.com/Rdatatable/data.table/pull/2456), [#4140](https://github.com/Rdatatable/data.table/pull/4140). We had used [GitHub's contributor page](https://github.com/Rdatatable/data.table/graphs/contributors) which omits 3 of these due to invalid email addresses, unlike GitLab's contributor page which includes the ids. The 4th omission was a PR to a script which should not have been excluded; a script is code too. We are sorry these contributors were not properly credited before. They have now been added to the contributors list as displayed on CRAN. All the contributors of code to data.table hold its copyright jointly; your contributions belong to you. You contributed to data.table when it had a particular license at that time, and you contributed on that basis. This is why in the last license change, all contributors of code were consulted and each had a veto.
--- a/NEWS.md
+++ b/NEWS.md
@@ -109,7 +109,7 @@ unit = "s")
 
 13. A relatively rare case of segfault when combining non-equi joins with `by=.EACHI` is now fixed, closes [#4388](https://github.com/Rdatatable/data.table/issues/4388).
 
-14. Non-equi joins now automatically set allow.cartesian to `TRUE`. Closes [4489](https://github.com/Rdatatable/data.table/issues/4489)
+14. Non-equi joins now automatically set allow.cartesian to `TRUE`. Closes [4489](https://github.com/Rdatatable/data.table/issues/4489). Thanks to @Henrik-P for reporting.
 
 ## NOTES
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -109,6 +109,8 @@ unit = "s")
 
 13. A relatively rare case of segfault when combining non-equi joins with `by=.EACHI` is now fixed, closes [#4388](https://github.com/Rdatatable/data.table/issues/4388).
 
+14. Non-equi joins now automatically set allow.cartesian to `TRUE`. Closes [4489](https://github.com/Rdatatable/data.table/issues/4489)
```
jangorecki, line 112: It would be good to mention github handler of a reported of that issue. And if you wish also your, as contributor who provided PR.

inst/tests/tests.Rraw
```diff
@@ -16861,3 +16861,12 @@ A = data.table(A=c(complex(real = 1:3, imaginary=c(0, -1, 1)), NaN))
 test(2138.3, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
 A = data.table(A=as.complex(rep(NA, 5)))
 test(2138.4, rbind(A,B), data.table(A=c(as.character(A$A), B$A)))
+
+# Set allow.cartesian = TRUE when non-equi #4489
+
+set.seed(2)
+dt = data.table(time = 1:8, v = sample(8))
+dt[time == 2L, v := 2L]
+d[time == 7L, v := 7L]
```
jangorecki, line 16870: `d` is missing

PR #4502, opened 2020-05-27
some sapply->vapply replacements

MichaelChirico:
Noticed while working on #4501.

Some regex:

```
grep -Enr "(?:any|all)\(sapply\(" R
grep -Enr "sapply\(.*is\." R
```


jangorecki:
Personaly I don't like our internal `vapply_*` family.
I would prefer to have something like this
```r
vapply2 = function(FUN.VALUE, ...) vapply(FUN.VALUE=FUN.VALUE, ...)
```
and use it
```r
vapply2(NA, x, is.numeric)
vapply2(1L, y, length)
vapply2("", z, paste, collapse="")
```
or just `vapply` is good enough

mattdowle:
On @jangorecki's comment the only minor thoughts I have on `vapply2` is that I generally don't remember what the version `2` is for on such versioned function names and prefer instead a letter or word or different name .. like `vapply_1b` I remember `1b` means length-1 boolean, and `sapply` I remember the `s` stands for simplify. Also the `*apply*` family taking the first argument as what to apply to is pretty strong in my mind.

R/groupingsets.R
```diff
@@ -60,7 +60,7 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
     stop("All columns used in 'sets' argument must be in 'by' too. Columns used in 'sets' but not present in 'by': ", brackify(setdiff(sets.all.by, by)))
   if (id && "grouping" %chin% names(x))
     stop("When using `id=TRUE` the 'x' data.table must not have a column named 'grouping'.")
-  if (any(sapply(sets, anyDuplicated)))
+  if (any(vapply_1i(sets, anyDuplicated)))
```
mattdowle, line 63: shouldn't this one be `vapply_1b`?

mattdowle: I forgot `anyDuplicated` returns integer index of first dup, `0L` otherwise.  So `_1i` is correct then.

PR #4506, opened 2020-05-27
catch some partial matching issues in examples

MichaelChirico:
Needs a NEWS item?


man/melt.data.table.Rd
```diff
@@ -87,52 +87,58 @@ An unkeyed \code{data.table} containing the molten data.
...
 DT <- data.table(
-      i_1 = c(1:5, NA),
-      i_2 = c(NA,6,7,8,9,10),
-      f_1 = factor(sample(c(letters[1:3], NA), 6, TRUE)),
-      f_2 = factor(c("z", "a", "x", "c", "x", "x"), ordered=TRUE),
-      c_1 = sample(c(letters[1:3], NA), 6, TRUE),
-      d_1 = as.Date(c(1:3,NA,4:5), origin="2013-09-01"),
-      d_2 = as.Date(6:1, origin="2012-01-01"))
+  i_1 = c(1:5, NA),
+  n_1 = c(NA, 6, 7, 8, 9, 10),
```
MichaelChirico, line 91: `i_2` was actually `numeric` so `i` is a bit confusing. further it's useful to have a `numeric` column (see below), so not just converting to `integer`

man/melt.data.table.Rd
```diff
@@ -87,52 +87,58 @@ An unkeyed \code{data.table} containing the molten data.
...
+melt(DT, id.vars=1:2, measure.vars="f_1")
+melt(DT, id.vars=c("i_1", "n_1"), measure.vars=3) # same as above
+melt(DT, id.vars=1:2, measure.vars=3L, value.factor=TRUE) # same, but 'value' is factor
+melt(DT, id.vars=1:2, measure.vars=3:4, value.factor=TRUE) # 'value' is *ordered* factor
 
 # preserves attribute when types are identical, ex: Date
-melt(DT, id=3:4, measure=c("d_1", "d_2"))
-melt(DT, id=3:4, measure=c("i_1", "d_1")) # attribute not preserved
+melt(DT, id.vars=3:4, measure.vars=c("d_1", "d_2"))
+melt(DT, id.vars=3:4, measure.vars=c("n_1", "d_1")) # attribute not preserved
```
MichaelChirico, line 111: `Date` is a numeric class -- this was warning about type promotion with `i_1`. no such warning with `n_1` --> clarity of use case

man/melt.data.table.Rd
```diff
@@ -87,52 +87,58 @@ An unkeyed \code{data.table} containing the molten data.
...
-melt(DT, id=1, measure=c("c_1", "f_1")) # value is char
-melt(DT, id=1, measure=c("c_1", "i_2")) # i2 coerced to char
+melt(DT, id.vars=1, measure.vars=c("c_1", "f_1")) # value is char
+suppressWarnings(
+  melt(DT, id.vars=1, measure.vars=c("c_1", "n_1")) # n_1 coerced to char, with warning
+)
 
 # on na.rm=TRUE. NAs are removed efficiently, from within C
-melt(DT, id=1, measure=c("c_1", "i_2"), na.rm=TRUE) # remove NA
+melt(DT, id.vars=1, measure.vars=c("c_1", "c_2"), na.rm=TRUE) # remove NA
```
MichaelChirico, line 126: change to `c_2` (also added this column) to focus on `na.rm`, otherwise there's a type promotion warning as well

.dev/CRAN_Release.cmd
```diff
@@ -232,6 +232,23 @@ require(data.table)
...
+options(
+  warn = 2L,
+  warnPartialMatchArgs = TRUE,
+  warnPartialMatchAttr = TRUE,
+  warnPartialMatchDollar = TRUE
+)
+for (f in list.files('man', full.names = TRUE)) {
+    l = readLines(f)
+    if (!any(grepl(r'(\examples)', l, fixed=TRUE))) next
+    nm = gsub(r'(\\name\{([^}]*)\})', '\\1', grep(r'(\\name)', l, value=TRUE))
```
jangorecki, line 246: this `r'(\\name\{([^}]*)\})'` requires R 4.0.0 so stopifnot would be useful just before entering the loop

man/dcast.data.table.Rd
```diff
@@ -95,20 +95,20 @@ dcast(DT, v1 + v2 + v3 ~ ., value.var = "v4")
 dcast(DT, v1 + v2 ~ ., value.var = "v4", fun.aggregate = sum)
 
 # fill and types
-dcast(DT, v2 ~ v3, value.var = 'v1', fill = 0L)  #  0L --> 0
-dcast(DT, v2 ~ v3, value.var = 'v4', fill = 1.1) # 1.1 --> 1L
+dcast(DT, v2 ~ v3, value.var = 'v1', fun.aggregate=length, fill = 0L)  #  0L --> 0
+dcast(DT, v2 ~ v3, value.var = 'v4', fun.aggregate=length, fill = 1.1) # 1.1 --> 1L
```
jangorecki, line 99: and what about line width of manual examples? partial matching is nice here so we don't have to break down calls into multiple lines

.dev/CRAN_Release.cmd
MichaelChirico: hmm I wasn't worried about that since this is in CRAN_release script, hopefully that's run on latest release

jangorecki: Yeah, that is no problem but to avoid confusion if reached here on older R, it is good to have stopifnot somewhere.
I am still on 3.6.3, will move to 4.0.1 directly.

PR #4510, opened 2020-05-28
build and install stuff

### LINKED ISSUE #4441: configure script exits too early
package installation output does not print `gcc` version even if user uses `gcc` because configure scripts escaped to early.
https://github.com/Rdatatable/data.table/pull/4162/files#r423629641

jangorecki:
- [x] closes #4441: moved printing gcc version to the top of configure script
- [x] new make target to more easily store git revision to be used before `make build && make install`
- [x] extra output to file of `CC` and `CFLAGS` in configure to retain compilation information so we can access that later after installation:
```r
cat(readLines(system.file(package="data.table", "cc")),sep="\n")
#CC=gcc -std=gnu99
#CFLAGS=-O3 -mtune=native
```
- [x] properly detect `CC` string to start with `gcc`, current condition will incorrectly match `CC=cc` as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,3 +42,6 @@ test:
 check:
 	_R_CHECK_CRAN_INCOMING_REMOTE_=false $(R) CMD check data.table_1.12.9.tar.gz --as-cran --ignore-vignettes --no-stop-on-test-error
 
+.PHONY: revision
+revision:
+	echo "Revision: $(shell git rev-parse HEAD)" >> DESCRIPTION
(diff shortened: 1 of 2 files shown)


jangorecki:
If user does not have own ~/.R/Makevars this PR actually let us #4441 
```r
cat(readLines(system.file(package="data.table", "cc")),sep="\n")
CC=gcc
CFLAGS=-g -O2
```
this is what I got when compilated with no Makevars

PR #4517, opened 2020-05-30
Add some new "status quo" atime tests for common operations

### LINKED ISSUE #3797: add timing test for many .SD cols
Follow up to #3777.  Add a test in CI somewhere to ensure it doesn't come back.  I almost added it to the main test suite but didn't since we need a timing test suite anyway, and the main test suite running on CRAN should be correctness only and quick.

### LINKED ISSUE #4687: Continuous Benchmarking
As already mentioned in multiple issues and over email/slack, we need automated tests that will be able to track performance regression.
This issue is meant to define scope.

Related useful project is planned in [conbench](https://github.com/conbench/conbench). Once it will be working, I think we should use it. Unfortunately it does not seem to happen anytime soon, or even in a more distant future.
Anyway, keeping scope minimal should make it easier to eventually move to _conbench_ later on.
Another related work is my old project [macrobenchmarking](https://gitlab.com/jangorecki/macrobenchmarking).
And recent draft PR #4517.

----

#### Dimensions by which we will track timings

- environment (allow to lookup hardware configuration)
- R version
- git sha of data.table (lookup date and version)
- benchmark script (probably fixed to `benchmark.Rraw`)
- query
- version of a query (in case we modify existing query for some reason)
- description

#### Dimensions that for now I propose to not include in scope

- operating system
- linux kernal version
- compiler
- compiler version
- R compilation flags
- data.table compilation flags
- metrics (memory usage, etc.)
- `datatable.optimize` option
- [ ] number of threads (?)
- [ ] multiple runs of single query (?)

----

#### Store timings

In current infrastructre we do not have any processes that appends artifacts (timings in context of CB). Each CB run has to store results somewhere and re-use them later on.

- [ ] timings storage in csv for simplicity (?)

#### Signalling a regression

- [ ] Should we compare only to the previous timings, or to an average timings from longer period (last month, etc.)?
- [ ] What is the tolerance threshold? for cheap queries 5-25% variance will be common.
- [ ] In case of exceeding threshold  we may want to run benchmark few times and take an average to ensure before signalling regression. Then we need a second threshold for such average.

#### Environment

To reduce number of false regression signals we need to use private dedicated infrastructure.
Having dedicated machine may not be feasible, so we need to have a mechanism of signalling to jenkins (or other orchestration process) that particular machine is in use in an exclusive mode.
- [ ] We may use the same machine that runs [db-benchmark](https://github.com/h2oai/db-benchmark)

#### Pipeline

In the most likely case of not having a dedicated machine, CB may ended up being queued for a longer while (up to multiple days). Therefore it make sense to have it in a separate pipeline rather than in our data.table GLCI. Such CB pipeline could be scheduled to run daily or weekly instead of running on each commit.
- [ ] We could eventually move publishing artifacts (package, website, docker images) to dedicated daily pipeline. This is not strictly related to CB but would make it easier to publish CB results as well.

#### Versioning

- [ ] Should all elements of CB be included as a part of `data.table` project? or a separate project
  - test script `inst/tests/benchmark.Rraw`
  - new function `benchmark()` that meant to be used like `test()`, and `benchmark.data.table()` to be used like `test.data.table()`
  - any extra orchestration could be in `.ci/`

----

## Example test cases

- `[[` on a list column by group #4646
- joining in a loop, order of different types #3928
- simple access `DT[10L]`, `DT[, 3L]` #3735
- use of `.SD` for many columns #3797
- calling `setDT` in a loop #4476
- multithreaded function calls by group `DT[, uniqueN(a), by=b]`, should stress new throttle feature #4484
- more cases in existing [benchmark.Rraw](https://github.com/Rdatatable/data.table/blob/b417000b570133d1d014a4745163142d0c68e195/inst/tests/benchmark.Rraw) file

jangorecki:
After having discussion with Matt on slack we decided to narrow down scope of #4687. So new benchmarking feature can be more usable and not introduce extra maintenance burden that tracking historical timings, and other features initially listed here, would require.

As a starting point I took system.time tests that have been already taken out from tests.Rraw file (to reduce run time of main test script). Those tests have been moved to a new benchmark() function that meant to replace test() function when system.time is needed.
To keep things simple, we don't need new benchmark.data.table(), as we can just call `test.data.table("benchmarks.Rraw")` or `cc("benchmarks.Rraw")` in dev-mode. It will already recognize benchmarks calls in the test script.
This is still very much a starting point so any feedback is very welcome.

Ideas for improvement:
- include `times` argument to run expression multiple times and take mean/median to compare. This will allow to make a more tight tolerance (test 1110).
- test for available memory at start and stop early if less than necessary

----

<details>
  <summary>Initial proposal at bc8a8becc41f46d3f199b74b7b5daec5e49ca23a</summary>

PR brings new set of scripts, and internal functions, to measure performance of data.table. They are not run in any of our workflow as of now, but rather should be run manually. For now there is no point to merge this branch. Opening PR to more easily document and refer to amongst gh issues.

For example, addressing "add timing test for many .SD cols #3797" for which scripts are defined in `benchmarks.Rraw` file. Yet to close #3797 we need to add a rules to be checked after all benchmarks, to confirm optimize=0 is not that much different than optimize=Inf.
```
data.table:::benchmark.data.table(libs=list.dirs("library/gcc", recursive=FALSE))
benchmark.data.table() running: benchmarks.Rraw
R_LIBS_USER=library/gcc/O0 R_DATATABLE_NUM_THREADS=1 R_DATATABLE_NUM_PROCS_PERCENT=100 Rscript inst/benchmarks/benchmarks.Rraw
R_LIBS_USER=library/gcc/O0 R_DATATABLE_NUM_THREADS=4 R_DATATABLE_NUM_PROCS_PERCENT=100 Rscript inst/benchmarks/benchmarks.Rraw
...
R_LIBS_USER=library/gcc/O0 R_DATATABLE_NUM_THREADS=40 R_DATATABLE_NUM_PROCS_PERCENT=100 Rscript inst/benchmarks/benchmarks.Rraw
R_LIBS_USER=library/gcc/O0-g R_DATATABLE_NUM_THREADS=1 R_DATATABLE_NUM_PROCS_PERCENT=100 Rscript inst/benchmarks/benchmarks.Rraw
...
R_LIBS_USER=library/gcc/O0-g R_DATATABLE_NUM_THREADS=40 R_DATATABLE_NUM_PROCS_PERCENT=100 Rscript inst/benchmarks/benchmarks.Rraw
R_LIBS_USER=library/gcc/O2 R_DATATABLE_NUM_THREADS=1 R_DATATABLE_NUM_PROCS_PERCENT=100 Rscript inst/benchmarks/benchmarks.Rraw
...
```
```
> data.table:::summary.benchmark()[-c(1:3)]
               cflags   num    fun          args         desc    th user_self sys_self elapsed
               <char> <num> <char>        <char>       <char> <int>     <num>    <num>   <num>
 1:               -O0  1.01      [ DT,,.SD,by=st  optimize=0L     1     1.395    0.230   1.625
 2:               -O0  1.02      [ DT,,.SD,by=st optimize=Inf     1     1.600    0.238   1.838
...
85: -O3 -mtune=native  1.01      [ DT,,.SD,by=st  optimize=0L     1     1.336    0.231   1.566
86: -O3 -mtune=native  1.02      [ DT,,.SD,by=st optimize=Inf     1     1.484    0.279   1.763
...
```
</details>


jangorecki:
Should also cover comment in #4666

MichaelChirico:
cc @Anirban166 this old draft PR has some potential new benchmarking tests for #6078. If we extract the good tests from here I think we could close this PR too.

MichaelChirico:
@tdhock what would the format be for pulling examples like this into {atime}, exactly? thus far all of our examples are like slow/fast pairs, but this is structured as standalone "status quo" tests we'd like to prevent from slowing in the future.

Would it be appropriate to add them as tests with just one label? "Fast" or "Status quo" maybe?

tdhock:
sure it is possible to omit the Slow and just include Fast (=Status Quo).
Actually you don't have to include any historical references in an atime test, but I think it is probably the best use of our dev time to focus on examples that somebody complained about in the past. (= that person depends on performance for some real application)

MichaelChirico:
@jangorecki I trimmed away a lot of what was here & just pulled out a few common operations that are ubiquitous to serve as baseline performance tests, PTAL.

@tdhock I think since there's no reference to any commit hash in this branch, we can proceed to delete this branch after merging, right?

.ci/atime/tests.R
```diff
@@ -286,5 +286,38 @@ test.list <- atime::atime_test_list(
...
+      expr = for (ii in seq_len(nrow(DT))) data.table:::`[.data.table`(DT, ii, `:=`(i, ii)),
+      `Status quo` = "8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0"), # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0)
+
+    "Serial value setting with set" = atime::atime_test(
+      setup = {
+        set.seed(34893)
+        DT = data.table(i=N:1)
+      },
+      expr = for (ii in seq_len(nrow(DT))) set(DT, ii, "i", ii),
+      `Status quo` = "8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0") # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0)
```
MichaelChirico:
```suggestion
      `Status quo` = "8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0"), # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0)
```

.ci/atime/tests.R
```diff
@@ -286,5 +286,38 @@ test.list <- atime::atime_test_list(
...
+        DT = data.table(grp = sample.int(10L, N, TRUE), v=rnorm(N), i=sample(N), vna=rnorm(N))
+        DT[1:3, vna := NA_real_]
+      },
+      expr = data.table:::`[.data.table`(DT, j=.(mean(v), sum(v), mean(i), sum(i), mean(vna, na.rm=TRUE), sum(vna, na.rm=TRUE))),
+      `Status quo` = "8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0"), # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0)
+
+    "GForce aggregation (benchmark across # groups)" = atime::atime_test(
+      setup = {
+        set.seed(34893)
+        DT = data.table(grp = sample.int(1e7, N, TRUE), v=rnorm(1e7), i=sample(1e7))
```
MichaelChirico:
```suggestion
        DT = data.table(grp = sample.int(1e7, N, TRUE), v=rnorm(1e7), i=sample(1e7), vna=rnorm(N))
```

.ci/atime/tests.R
```diff
@@ -286,5 +286,38 @@ test.list <- atime::atime_test_list(
...
+      },
+      expr = for (ii in seq_len(nrow(DT))) data.table:::`[.data.table`(DT, ii, `:=`(i, ii)),
+      `Status quo` = "8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0"), # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0)
+
+    "Serial value setting with set" = atime::atime_test(
+      setup = {
+        set.seed(34893)
+        DT = data.table(i=N:1)
+      },
+      expr = for (ii in seq_len(nrow(DT))) set(DT, ii, "i", ii),
```
MichaelChirico:
```suggestion
      expr = for (ii in seq_len(nrow(DT))) data.table::set(DT, ii, "i", ii),
```

.ci/atime/tests.R
```diff
@@ -445,6 +445,39 @@ test.list <- atime::atime_test_list(
     Fast = "ed2df986da6d3a4ff35bec1b0f75db2b767e3eb2", # Merge commit of the PR that uses a much faster implementation
     expr = data.table::isoweek(x)),
 
+    "GForce aggregation (benchmark across # rows)" = atime::atime_test(
+      setup = {
+        set.seed(34893)
+        DT = data.table(grp = sample.int(10L, N, TRUE), v=rnorm(N), i=sample(N), vna=rnorm(N))
+        DT[1:3, vna := NA_real_]
+      },
+      expr = data.table:::`[.data.table`(DT, j=.(mean(v), sum(v), mean(i), sum(i), mean(vna, na.rm=TRUE), sum(vna, na.rm=TRUE))),
```
ben-schwen, line 454: wouldn't we need a by for GForce to kick in?

.ci/atime/tests.R
```diff
@@ -445,6 +445,39 @@ test.list <- atime::atime_test_list(
...
+      `Status quo` = "8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0"), # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8f5ffa8bb8f3f5861020a6e32f897c30e42eeab0)
+
+    "GForce aggregation (benchmark across # groups)" = atime::atime_test(
+      setup = {
+        set.seed(34893)
+        DT = data.table(grp = sample.int(1e7, N, TRUE), v=rnorm(1e7), i=sample(1e7), vna=rnorm(N))
+      },
+      expr = data.table:::`[.data.table`(DT, j=.(mean(v), sum(v), mean(i), sum(i), mean(vna, na.rm=TRUE), sum(vna, na.rm=TRUE))),
```
ben-schwen, line 462: same as above

PR #4523, opened 2020-06-03
Support positional specifiers on Windows for translations

mattdowle:
As per https://github.com/Rdatatable/data.table/issues/4402#issuecomment-638453605
Attempt at #4402

Issue #4402, opened 2020-04-27
rbind unmatched columns crashes R session (Windows, dev version) 


shrektan:
Test 450 will crash R session with dev version of data.table (but not with CRAN version) on Windows.

R version: 3.6.2
Platform: Windows 10

```r
library(data.table)
DT = data.table(a=1:3,b=4:6)
rbind(DT,list(c=4L,a=7L))
```

jangorecki:
On linux it is
```
Error in rbindlist(l, use.names, fill, idcol) : 
  Column 1 ['c'] of item 2 is missing in item 1. Use fill=TRUE to fill with NA (NULL for list columns), or use.names=FALSE to ignore column names.
```

shrektan:
This line leads to the crash.

https://github.com/Rdatatable/data.table/blob/dd7609e83132e19c9be80d71c73e8a8f95e19b27/src/rbindlist.c#L210-L211

shrektan:
I can confirm it's due to the Chinese translation as if I set the language to `en`, it works well.

jangorecki:
fy @MichaelChirico

shrektan:
Looks like we can't use `_()` with `snprintf()` on Windows, because when I change the code from

https://github.com/Rdatatable/data.table/blob/dd7609e83132e19c9be80d71c73e8a8f95e19b27/src/rbindlist.c#L210-L212

to 

```c
if (usenames==TRUE) error( _("Column %d ['%s'] of item %d is missing in item %d. Use fill=TRUE to fill with NA (NULL for list columns), or use.names=FALSE to ignore column names.%s"),
              w2+1, str, i+1, missi+1, extra);
```

The crash will no longer happen.

If that's the case, we may have to check all the combination of `snprintf()` and `_()`.

shrektan:
I'm almost sure that `snprintf()` can't be used with `_()` on Windows but I don't know the reason.

As you can see that the following code (another place uses `snprintf()` with `_()`), on Windows (compared to Mac), although it doesn't crash the R session, it doesn't interpolate the values like `%3$d`.

```r
library(data.table)
DT = data.table(a=1:3,b=2)
rbind(DT,list(b=5,a=3), use.names = F)
```
### Mac

<img width="531" alt="image" src="https://user-images.githubusercontent.com/8368933/80618804-6f854380-8a76-11ea-97b7-c0f06bd21b28.png">

### Windows

[...]

PR #4524, opened 2020-06-04
attempt to add test of Chinese error message

mattdowle:
For #4402, follow up to #4523

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -16915,13 +16915,19 @@ test(2140, dt[dt, on=.(time>time, v>v), .N, by=.EACHI], data.table(time=1:8, v=I
 DT = data.table(a=1:3,b=4:6)
 test(2141.01, rbind(DT,list(c=4L,a=7L)), error="Column 1 ['c'] of item 2 is missing in item 1")
 old = Sys.getenv("LANGUAGE")
+print(old)
+print(.libPaths())
 if (.Platform$OS.type=="windows") {
   i = 2L
   for (lang in c("Chinese_China.1252", "Chinese_China.936", "Chinese (Simplified)_China.936", "Chinese (Simplified)_China.1252",
                "zh-cn", "zh-CN", "zh_CN", "zh", "ZH", "zh_cn", "Chinese - China", "Chinese_China", "Chinese-China")) {
-    Sys.setenv(LANGUAGE = lang)
-    Sys.setenv(LANG = lang)
-    test(2141+i/100, rbind(DT,list(c=4L,a=7L)), error="2.*1.*c.*1") # arguments different order
+    test(2141+i/100, system2("R", input="q('no')", args="--no-save", env=paste0("LANGUAGE='",lang,"'")), 0L)
+    test(2142+i/100, system2("R", input="q('no')", args="--no-save", env=paste0("LC_ALL='",lang,"'")), 0L)
+    test(2143+i/100, system2("R", input="q('no')", args="--no-save", env=paste0("LANGUAGE=",lang)), 0L)
+    test(2144+i/100, system2("R", input="q('no')", args="--no-save", env=paste0("LC_ALL=",lang)), 0L)
+    #Sys.setenv(LANGUAGE = lang)
+    #Sys.setenv(LANG = lang)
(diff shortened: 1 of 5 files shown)


inst/tests/tests.Rraw
```diff
@@ -16910,3 +16910,14 @@ dt = data.table(time = 1:8, v = INT(5,7,6,1,8,4,2,3))
 dt[time == 2L, v := 2L]
 dt[time == 7L, v := 7L]
 test(2140, dt[dt, on=.(time>time, v>v), .N, by=.EACHI], data.table(time=1:8, v=INT(5,2,6,1,8,4,7,3), N=INT(3,5,2,4,0,1,0,0)))
+
+# repeat of test 450 for #4402
+DT = data.table(a=1:3,b=4:6)
+test(2141.1, rbind(DT,list(c=4L,a=7L)), error="Column 1 ['c'] of item 2 is missing in item 1")
+old = Sys.getenv("LANGUAGE")
+Sys.setenv(LANGUAGE="zh_CN")
```
MichaelChirico, line 16918: I tried to trigger the #4402 error in the [`snprintf-windows-test`](https://github.com/Rdatatable/data.table/tree/snprintf-windows-test) branch earlier but wasn't getting anywhere, you can see the 5 commits there used various versions of this, I may have been doing something wrong

mattdowle:
@shrektan Thanks for testing master and I understand it doesn't work. Just so we can get a test going, do you know how to get R on Windows to produce Chinese?  I've tried all kinds of variations of the LANG and LANGUAGE in this PR and none of them seem to generate any Chinese at all.  Is it the wrong string, the wrong environment variable?  Maybe R on Windows must be set to a different language on startup and can't be done at runtime like this?

shrektan:
Thanks for all your efforts. Frankly speaking, I don't really know and I'm not sure if it's possible without the Windows' language being set up to Chinese.

But I will try to find a way (hopefully) this day later when I get back home, where I can set up a Windows machine with English locale to mimic what's happening in the CI.

mattdowle:
https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Localization-of-messages contains (bold mine) : 

> It is usually possible to change the language once R is running via **(not Windows)** Sys.setlocale("LC_MESSAGES", "new_locale"), or by setting an environment variable such as LANGUAGE

mattdowle:
Still no Chinese present in main.Rout. Sigh.
This artile mentioned `chs` and `simplified-chinese` but R, on AppVeyor at least, still remains English.
https://withr.github.io/configure-character-encoding-for-r-under-linux-and-windows/

shrektan:
@mattdowle @MichaelChirico The below code works on my PC (Windows 10, English Language, with Chinese Language installed).
```r
x <- list(
  LC_COLLATE = "Chinese (Simplified)_China.936",
  LC_CTYPE = "Chinese (Simplified)_China.936",
  LC_MONETARY = "Chinese (Simplified)_China.936",
  LC_NUMERIC = "C",
  LC_TIME = "Chinese (Simplified)_China.936"
)
for (i in seq_along(x)) {
  Sys.setlocale(names(x)[[i]], x[[i]])
}
Sys.setenv('LANGUAGE' = 'zh_CN')
```

shrektan:
Looks this approach works. The "main.Rout.fail" prints the expected Chinese message before it gets segfault.

image

mattdowle:
Oh wow! You did it! Not only Chinese but it reproduces the crash too ... awesome!

For future reference, this is what I see in AppVeyor web display; i.e., not Chinese but definitely not ascii, so the translation is kicking in and this is just a web-display issue of the output in AppVeyor that we can ignore for this purpose.
image

mattdowle:
Will merge shortly so that @shrektan can test master to see if it works now please. If it does work now, we can go through and tidy/review in a follow up. If it still doesn't work then we can throw this wrapper away. Although it's a bit hairy, at least it's contained in one file and not too long. And being plain C99, we can be confident it should work on all Windows machines, old and new, MinGW, MSVC, etc.

MichaelChirico:
Regardless of the outcome, thanks for the herculean effort to support this! Awesome stuff.

mattdowle:
@MichaelChirico yes very good point. `%1$` might match  to the 2nd character of `%%1$` which wouldn't be correct. Ok will fix that and add a test.

PR #4529, opened 2020-06-06
when as.data.frame dispatches to list method, force input to list

### LINKED ISSUE #4526: as.data.table(dfidx) fails
The class `dfidx` (from the package `dfidx`) exists. Here the idea is to glue and idx `data.frame` to another `data.frame`, and when working with the desired `data.frame`, the idx object always is glued to it. This is for example used in the `mlogit()` package. Conversion of this type to `data.table` fails.

```
library(mlogit)
data("TravelMode",package="AER") 
Mtravel <- dfidx(data     = TravelMode,         
                       shape    = "long",           
                       choice   = "choice",         
                       chid.var = "individual",     
                       alt.var  = "mode")
DTtravel <- as.data.table(Mtravel)
```
Gives the following error message:
```
Error in .subset2(x, i, exact = exact) : 
  attempt to select less than one element in get1index
```

The `mlogit` package requires such an object to carry out the desired regression. dfidx objects also come with their own '['. 
More on this here:
https://cran.microsoft.com/web/packages/dfidx/vignettes/dfidx.html#more_advanced_use_of_dfidx

One could always use `dcast()` and `melt()` to change `data.table`s from wide to long etc. and just use dfidx right before the estimation. Another alternative would be to write a function like the following (which of course is terribly inefficient)

```
my.as.dt <- function(x){
  wide = ncol(x) - 1
  long = nrow(x)
  
  DTprep = list()
  for(i in 1:wide){
    DTprep[[i]] = x[1:long, i][1:long]
  }
  
  DT = as.data.table(DTprep)
  colnames(DT) = colnames(x)[1:wide]
  return(DT)
}
```

Another approach would be to use setDT, which also delivers a warning:

```
DT = setDT(Mtravel)
Warning message:
In setDT(Mtravel) :
  Some columns are a multi-column type (such as a matrix column): [8]. setDT will retain these columns as-is but subsequent operations like grouping and joining may fail. Please consider as.data.table() instead which will create a new column for each embedded column.
```

And the following call of that object will deliver an error:

```
DT
Error in `[.data.table`(x, i, , ) : 
  Column 8 ['idx'] is a data.frame or data.table; malformed data.table.
```

Hopefully you can just give `as.data.table()` and `setDT()` the power to overwrite this and drop the idx `data.frame` (potentially with a warning that some commands which require it in the future may fail).

MichaelChirico:
Closes #4526 

I'm not sure if there's an efficiency implication to the `as.list` but it feels right to do regardless.


NEWS.md
```diff
@@ -14,6 +14,8 @@
 
 1. `by=.EACHI` when `i` is keyed but `on=` different columns than `i`'s key could create an invalidly keyed result, [#4603](https://github.com/Rdatatable/data.table/issues/4603) [#4911](https://github.com/Rdatatable/data.table/issues/4911). Thanks to @myoung3 and @adamaltmejd for reporting, and @ColeMiller1 for the PR. An invalid key is where a `data.table` is marked as sorted by the key columns but the data is not sorted by those columns, leading to incorrect results from subsequent queries.
 
+2. `as.data.table()` more robust to unrecognized subclasses of `data.frame`, [#4526](https://github.com/Rdatatable/data.table/issues/4526). Thanks @RicoDiel for the report.
```
mattdowle, line 17: A little more detail in the news item please:
i) what was not robust before; e.g. was it an error or warning before, or was the result wrong?
ii) what is an example of an 'unrecognized subclass'?

mattdowle (changes requested):

NEWS.md
MichaelChirico: had another go at it, LMK WYT

mattdowle: @MichaelChirico I don't see any changes. Did you commit and push?

MichaelChirico: i must have forgotten to push, done now

mattdowle: strange, I still don't see the news item updated.

MichaelChirico: Hmm I am also seeing it rendered the same in this comment thread. But on the `Changes` page it's different. I guess a thread is anchored to a commit in chase the anchoring spot is moves or disappears. Marking this as resolved to hopefully remove the confusion.

mattdowle: I had been looking at the files-changed tab, and the changes button, but I was expecting a bigger change to the news item so my eyes were seeing the first 'more robust to subclasses' and thinking it wasn't changed. Then the several commits were labelled merges which threw me off.   Ok, on the same page now.

The idea is to be very specific in news items, e.g. including an example error message so it comes up in when folk search for parts of the error message.  I'll have a go...

mattdowle: Done. I added your name too given #4959 and I'll add a comment there too.

mattdowle (approved):

MichaelChirico:
thanks! looking better

PR #4534, opened 2020-06-08
use gettextf to escape % in error message

MichaelChirico:
Got an error from `tools::update_pkg_po('.')`, it seems it was having trouble understanding that the `%` in the R message is not a formatting specifier. Tried playing around with a few things but changing to use `gettextf` is the only thing that works apparently. Message formatting looks nicer anyway.


MichaelChirico:
This PR is superseded; closing

PR #4542, opened 2020-06-13
alloc on heap to avoid stack overflow

jangorecki:


MichaelChirico:
I see these other potential hits, any that should be fixed for sure?

```
# removing a few false positives manually from:
grep -Er "^\s*[a-zA-Z_0-9]+(?:\s+[*]?\s*|\s*[*]?\s+)[a-zA-Z_0-9]+\[[^]0-9]" src/*.c -n
src/forder.c:843:    uint8_t o[my_n];
src/forder.c:910:      int TMP[my_n];  // on stack fine since my_n is very small (<=256)
src/forder.c:1025:    int my_gs[ngrp==0 ? 256 : ngrp];  // ngrp==0 when sort and skip==true; we didn't count the non-zeros in my_counts yet in that case
src/forder.c:1199:  int my_gs[ngrp];
src/frollR.c:98:  int* ikl[nk];                                                 // pointers to adaptive window width
src/fsort.c:132:  double mins[nBatch], maxs[nBatch];
src/nafill.c:117:  double* dx[nx];
src/nafill.c:118:  int32_t* ix[nx];
src/nafill.c:119:  int64_t* i64x[nx];
src/nafill.c:120:  uint_fast64_t inx[nx];
```

jangorecki:
thanks @MichaelChirico! all the remaining ones should be fine

PR #4551, opened 2020-06-17
a pass at clarifying ?froll

MichaelChirico:
Mainly because it wasn't clear to me whether `n=3` or `n=7` is needed for a weekly moving average


man/froll.Rd
```diff
@@ -12,71 +12,67 @@
...
-  Setting \code{algo="exact"} will make rolling functions to use
-  compute-intensive algorithm that suffers less from floating point
-  rounding error. It also handles \code{NaN, +Inf, -Inf} consistently to
+  When \code{algo="fast"} an \emph{"on-line"} algorithm is used, and
+  all of \code{NaN, +Inf, -Inf} are treated as \code{NA}.
+  Setting \code{algo="exact"} will make rolling functions to use a more
+  computationally-intensive algorithm that suffers less from floating point
+  rounding error (the same consideration applies to \code{\link[base]{mean}}).
+  \code{algo="exact"} also handles \code{NaN, +Inf, -Inf} consistently to
   base R. In case of some functions (like \emph{mean}), it will additionally
```
MichaelChirico, line 60: Jan this second sentence looks like the same as the one before, is it redundant or am I missing something?

jangorecki: What do you mean by the same applies to base mean? Afair base mean do rounding correction always.
If we don't want to repeat anything then better remove the one before.

MichaelChirico: Just to convey that `algo='exact'` should be matching `base::mean` which does error correction right?

man/froll.Rd
```diff
@@ -12,71 +12,65 @@
...
+  \item{\dots}{ Extra arguments passed to \code{FUN} in \code{frollapply}. }
 }
 \details{
-  \code{froll*} functions accepts vectors, lists, data.frames or
-  data.tables. They always return a list except when the input is a
-  \code{vector} and \code{length(n)==1} in which case a \code{vector}
-  is returned, for convenience. Thus rolling functions can be used
-  conveniently within data.table syntax.
+  \code{froll*} functions accept vectors, lists, \code{data.frames} or
+  \code{data.tables}. They always return a list except when the input is a
```
jangorecki, line 42: I would avoid wrapping plural of data.table in code formatting. I have seen many people being confused by DataTables.

mattdowle: ok, done in e35bea2

PR #4564, opened 2020-06-20
fread parse Y/N as bool

### LINKED ISSUE #4563: parse [Yy]/[Nn] fields as logical (with option?)
related to: https://twitter.com/evalparse/status/1274011477125431296

there are a couple logical fields in the file stored as Y/N (and I've come across this in other data sets previously as well).

https://loanperformancedata.fanniemae.com/lppub-docs/FNMA_SF_Loan_Performance_Glossary.pdf?_ga=2.40004134.548517782.1592626534-439404991.1592626534

I guess the performance impact won't be large since there are only two unique strings to add, but there is still a memory benefit.

Could be on by default and controlled by `logical01`? Maybe have that argument accept a vector or even change the name to be more general.

MichaelChirico:
Closes #4563 

The basic idea is here, but it seems more complicated than I'd hoped because this interferes with the `header='auto'` logic whenever there's a column named `[YyNn]`, which happens in a few of our tests.

Options

 1. Adjust `header='auto'` logic to allow YN-as-bool as well as strings to be considered OK for recognizing a header
 2. Only allow `Y` and `N` (and _not_ `y` or `n`) to be recognized as boolean. This would solve all our current tests but may not be as good "in the wild"
 3. Don't pursue a YN-as-bool parser owing to this issue

PS it would be easy to support `[Yy][Ee][Ss]`/`[Nn][Oo]`-as-bool in this parser as well, but I haven't seen such things in the wild. I would leave that out until there would be demand.


MichaelChirico:
I think the way forward here is to put this behind a `logicalYN` argument. That will enable back-compatibility.

MichaelChirico:
Suggesting @tdhock as reviewer, I'm not sure who's the most familiar with `fread()` now. Feel free to pass along :)

src/fread.c
```diff
@@ -75,8 +75,8 @@ static freadMainArgs args = {0};  // global for use by DTPRINT; static implies =
   static int mmp_fd = -1;
 #endif
 
-const char typeName[NUMTYPE][10] = {"drop", "bool8", "bool8", "bool8", "bool8", "bool8", "int32", "int64", "float64", "float64", "float64", "int32", "float64", "string"};
-int8_t     typeSize[NUMTYPE]     = { 0,      1,       1,       1,       1,       1,       4,       8,       8,         8,         8,         4,       8       ,  8      };
+const char typeName[NUMTYPE][10] = {"drop", "bool8", "bool8", "bool8", "bool8", "bool8", "bool8", "int32", "int64", "float64", "float64", "float64", "int32", "float64", "string"};
+int8_t     typeSize[NUMTYPE]     = { 0,      1,       1,       1,       1,       1,      1,       4,       8,       8,         8,         8,         4,       8       ,  8      };
```
tdhock, line 79: I don't understand this change. Can you please add a comment?

MichaelChirico: Also a few comments, PTAL.

src/fread.c
```diff
@@ -1154,6 +1154,21 @@ static void parse_bool_lowercase(FieldParseContext *ctx)
...
+    *(ctx->ch) = ch + 1;
+  } else if (ch[0] == 'N' || ch[0] == 'n') {
+    *target = 0;
+    *(ctx->ch) = ch + 1;
+  } else {
+    *target = NA_BOOL8;
+  }
+}
 
 /* How to register a new parser
```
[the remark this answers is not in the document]
tdhock: ok that is fine thanks

tdhock (approved):
great thanks

Issue #1007, opened 2015-01-14
rolling joins with roll=TRUE and value -Inf


arunsrinivasan:
Discovered due to #1006.

```R
require(data.table) ## 1.9.5
DT = data.table(x=c(-Inf, 3), y=1:2, key="x")
#       x y
# 1: -Inf 1
# 2:    3 2

# wrong result.. should be 2,1
DT[J(2), roll=TRUE]
#    x  y
# 1: 2 NA

PR #4566, opened 2020-06-21
bmerge rework

jangorecki:
This PR is a subset of changes I made in [bmerge-dev](https://github.com/Rdatatable/data.table/tree/bmerge-dev) branch where bmerge was made parallel ([timings](https://gist.githubusercontent.com/jangorecki/0cf9170bee01ee6763719ada418c81e7/raw/4589f2cf8137024b3ea769fc8f4162ad53da9a6e/time.out), note that bmerge is not well parallelized there yet). Unfortunately it segfaults there in some cases, thus I decided to push those changes in smaller steps.

In this PR we have:
- declaration of variables more local in `bmerge_r`
- removed all data pointer macros inside `bmerge_r`
- reduce overhead related of calling `ENC2UTF8` twice: once in `bmerge_r` and once in `StrCmp`
- AFAIU `col=-1`, a nqGrpId col, can be of type integer only, thus redundant condition removed from REALSXP case: https://github.com/Rdatatable/data.table/pull/4566/files#diff-a40725b0264adfb86e09dd63d64d1a011f919ea3f7b8128b753a1cb9b21a1188L374
- reduce repetition by adding new variables `t = TYPEOF(xc)`, `isDataCol = col>-1`, `isLastCol = col==ncol-1`


jangorecki:
CI failure here is caused by #4769

src/bmerge.c
```diff
@@ -1,20 +1,20 @@
...
-  o NA,NAN,-Inf,+Inf are distinct values and can be joined to
-  o type double is joined within tolerance (apx 11 s.f.)
-  o join to prevailing value (roll join a.k.a locf), forwards or backwards
-  o join to nearest
-  o roll the beginning and end optionally
-  o limit the roll distance to a user provided value
+  * list of vectors (key of many columns) of different types
+  * ties (groups)
+  * NA,NAN,-Inf,+Inf are distinct values and can be joined to
+  * type double is joined within tolerance (apx 11 s.f.) -- is this info up to date?
```
jangorecki, line 12: note the question added at the end of this line

src/bmerge.c
```diff
@@ -1,20 +1,20 @@
...
+  * join to prevailing value (roll join a.k.a locf), forwards or backwards
+  * join to nearest
+  * roll the beginning and end optionally
+  * limit the roll distance to a user provided value
+  * non equi joins (no != yet) since 1.9.8
```
jangorecki, line 17: this single line is new here

src/bmerge.c
```diff
@@ -491,48 +495,48 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
...
-        if (o) k=o[j]-1; else k=j;
+        int k = o ? o[j]-1 : j;
         retFirst[k] = retFirst[ir];
         retLength[k]= retLength[ir];
       }
     }
   }
   switch (op[col]) {
   case EQ:
-    if (ilow>ilowIn && (xlow>xlowIn || ((roll!=0.0 || op[col] != EQ) && col==ncol-1)))
```
mattdowle, line 533: Odd that the `op[col] != EQ` was here when it was inside case EQ. Nice spot.

src/bmerge.c
```diff
@@ -280,21 +280,21 @@ void bmerge_r(int xlowIn, int xuppIn, int ilowIn, int iuppIn, int col, int thisg
         /* Regular roll=TRUE|+ve|-ve */                                                           \
         if ((( roll>0.0 && (!lowmax || xlow>xlowIn) && (xupp<xuppIn || !uppmax || rollends[1]))   \
           || ( roll<0.0 && xupp==xuppIn && uppmax && rollends[1]) )                               \
-          && ( isinf(rollabs) || ((LOWDIST)-(TYPE)rollabs <= 0) ))                                \
+          && ( isinf(rollabs) || ((LOWDIST)-(TYPE)rollabs <= (TYPE)1e-6) ))                       \
```
jangorecki, line 283: Could you explain this change?

mattdowle: When TYPE is int or int64_t, (TYPE)1e-6 casts down to 0 as before. When TYPE is double, it is 1e-6 as before. So it's just a change to make the macro-ization work.  The bigger change vs master on this line is the `isinf()` first.  That's for efficiency to avoid the comparison happening at all, but also to simplify the solution for #1007 and reduce an extra comparison. All tests pass (there is no change to tests.Rraw in this PR at all) so I'm assuming no behavior has changed. If it has then we are missing a test for that behavior.

mattdowle:
The original intent and title of this PR was a first step towards parallelization by taking dataptr out of bmerge_r. (There wasn't any DATAPTR as such to be removed because that was already done when USE_RINTERNALS was removed before. So dataptr meant REAL, INTEGER, STRING_ELT, etc.) There were so many other good changes in this PR that are retained that it became more of a rework, hence title change.

Of the 5 bullets in the original comment, 4 are retained. The local variables in loops and StrCmp doing ENC2UTF8 twice were particularly good spots. The two bullets noticing redundancy and repetition were retained and continued in my commits.

Overall bmerge.c has reduced from 555 lines to 454 by moving long-standing common repeated logic into a macro. That's exchanging one type of complexity for another. On the positive, we no longer have a risk where logic in one case can be wrong but right in another case; i.e. test coverage for one type covers other types too. It's easier to add new types, as evidenced in this PR now by dealing with integer64 separately in a much cleaner way and being able to remove i64twiddle() as per the todo. But on the negative it's harder to see how the macro works.  At least we can see what's different between the cases now and continue to work on that.  Removing dtwiddle in future would further simplify.

Back to parallelization.  There are two issues at play:  i) INTEGER(), REAL(), etc have higher overhead since ALTREP so need to be taken outside loops,  ii) because they are R-API functions we think we have to take them outside parallel regions. The 1st is true.  The 2nd may not be true.  So long as we've checked that the columns are not ALTREP columns (which we do in many places and un-ALTREP columns whenever we encounter them), then I'm thinking it should be fine for them to appear inside a parallel region.  Not inside a loop due to their overhead, but inside a parallel region before a loop like they are in bmerge_r. TYPEOF is thread-safe too I guess but we should check. What isn't thread-safe is ENC2UTF8 which allocates the UTF8 result. All we need to do is write a StrCmp that returns the same result by comparing on a byte level in an ENC2UTF8 way, without actually calling ENC2UTF8. Then bmerge_r would be thread-safe, provided that no column is ALTREP is checked up front in bmerge(). Although not as much as a direct loop, bmerge_r is recursive so the overhead of INTEGER, REAL could still reduce performance. I doubt it would be large enough to be worth working on. But if it was, then allocating 
a vector to hold the pointers up front as this PR originally did, could be reconsidered. Not because it is needed for parallelizaton, but to avoid the overhead of being called once by each call of bmerge_r.

jangorecki:
I see many great improvements.
As for `dt_t` and `column_t` types, I was planning to re-use them in a follow up PR(s), should I avoid them or is it fine to introduce them in another PR?

mattdowle:
I'm not really comfortable with `dt_t` and `column_t` because that feels like a thin layer.  I like working with R-API directly rather than going via a layer.  But if you have other branches that use them then maybe I should look at them so I can understand the benefits.
Insulating ourselves from future changes in R is a good point. Maybe then, when bmerge is parallel. But I envisage a layer specific for bmerge_r there, not a general layer that is used throughout the whole code base. Having said that, maybe a layer would have avoided all the work we've done in moving these functions outside loops; we could have replaced the R API calls with calls to a non-ALTREP layer.
Checking enc2utf first, and then setting nth=1 presumably, yep good idea. #4386 is one of the 5 PRs marked High which I'm working through. That one is 'lazy forder' though so I'm not sure how that relates to bmerge yet.

jangorecki:
This is bmerge parallel dev branch: https://github.com/Rdatatable/data.table/compare/bmerge-dev
Note that it only batch LHS and to each batch joins full RHS, this is sub-efficient and should be replaced by batching RHS as well. RHS batches should be computed by making non-equi joins for boundaries of LHS batches.

In smerge (sort merge) branch it currently works for single integer column, to make it work for any type multicol I wanted to re-use `dt_t` and `column_t`: https://github.com/Rdatatable/data.table/compare/smerge

PR #4576, opened 2020-06-26
unit tests for #3779

### LINKED ISSUE #3779: Issue in X[Y,,.EACHI] when Y contains integer64
I was performing a left join with a right table with integer64 and it failed.
I might be misunderstanding .EACHI but I would expct the following to work, and it does when Y$y is not integer64

```
X = data.table(x=1:3)                          
Y = data.table(x=1:2, y=as.integer64(c(10,20)))
X[Y, `:=`(y = i.y), on = "x", by = .EACHI]     
Error in `[.data.table`(X, Y, `:=`(y = i.y), on = "x", by = .EACHI) :                          
  Internal error: memrecycle source is integer64 but target is real and not integer64; target should be type integer64
```

jangorecki:
closes #3779
no NEWS entry because it was already fixed in 1.12.8, just to ensure no regression in future


mattdowle:
I tested the test but it passed in 1.12.6. So then I found it was actually fixed in 1.12.4 (not 1.12.8) and the test correctly fails in 1.12.2.

Issue #4577, opened 2020-06-26
New check note and warning in R-devel


jangorecki:
Observed just now: https://rdatatable.gitlab.io/data.table/web/checks/data.table/test-dev-cran-lin/00check.log
```
* checking Rd \usage sections ... NOTE
 Error in unique.default(env_list) : 
   LENGTH or similar applied to environment object
 Calls: <Anonymous> ... .get_S3_generics_as_seen_from_package -> unique -> unique.default
 Execution halted
 The \usage entries for S3 methods should use the \method markup and not
 their full name.
 See chapter ‘Writing R documentation files’ in the ‘Writing R
 Extensions’ manual.
```
```
 * checking S3 generic/method consistency ... WARNING
 Error in unique.default(env_list) : 
   LENGTH or similar applied to environment object
 Calls: <Anonymous> ... .get_S3_generics_as_seen_from_package -> unique -> unique.default
 Execution halted
 See section ‘Generic functions and methods’ in the ‘Writing R
 Extensions’ manual.
```
same on 1.12.8

jangorecki:
Following commits might have cause the issue: https://github.com/wch/r-source/commit/69de92b9fb1b7f2a7c8d1394b8d56050881a5465

jangorecki:
It is regression in R-devel, and hopefully will be fixed. We can safely close this issue.

PR #4584, opened 2020-06-30
R-devel check detects soft issues now

jangorecki:
On our R-devel linux build "soft issues" that were not causing job status to fail were not signalled loudly enough. They required a human to look into log output to spot the issue (#4577).
This PR address that. If there is a breaking change or regression in R-devel, we will see that with an orange color of pipeline badge, even when it is just a new NOTE or a new compilation warning.
Thanks to @mattdowle for idea.

----

when there as an early failure, and no check artifacts are available, job (`dev-cran-lin`) is missing on our check summary: https://jangorecki.gitlab.io/-/data.table/-/jobs/618377341/artifacts/bus/integration/cran/web/checks/check_results_data.table.html

when failure happens due to out strict checks on check.out, then job (`dev-cran-lin`) is listed on our check summary and we can preview install output and check log: [link soon]

### DIFF
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -9,7 +9,7 @@ On each Pull Request opened in GitHub we run Travis CI and Appveyor to provide p
 Test jobs:
 - `test-rel-lin` - `r-release` on Linux, most comprehensive test environment, `-O3 -flto -fno-common -Wunused-result`, extra check for no compilation warnings, includes testing [_with other packages_](./../inst/tests/other.Rraw) ([extended suggests](./../inst/tests/tests-DESCRIPTION))
 - `test-rel-cran-lin` - `--as-cran` on Linux, `-g0`, extra check for final status of `R CMD check` where we allow one NOTE (_size of tarball_).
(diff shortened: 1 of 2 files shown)


jangorecki:
We can now see orange pipeline badge in https://gitlab.com/Rdatatable/data.table/pipelines
It should turn green when we rebuild R-devel docker image. It is is rebuild daily, statuses are here: https://gitlab.com/jangorecki/dockerfiles/pipelines
The image is build using R-devel sources from `https://stat.ethz.ch/R/daily/R-devel.tar.gz`.
The issue that is causing the orange badge now has been fixed [just 11 hours ago](https://github.com/wch/r-source/commit/1b867bb7d3026c89dfbbbdbee7af24b793bcf3d4).
So in order to be green, stat.ethz.ch needs to provide recent enough R-devel, at the time of rebuild of docker image.
Just clarifying that so there will be no confusion, why we are stll orange.

I will try to figure out if orange badge can be also on README rather than green...

PR #4586, opened 2020-07-01
coerceAs applied to fill in dcast

### LINKED ISSUE #4561: dcast on missing int64 value generates 9218868437227407266 instead of NA
When I dcast a data.table that contains an int64 column, it fills in 9218868437227407266 for any missing values instead of the usual NA. 

This is related to Issues [#488](https://github.com/Rdatatable/data.table/issues/488), [#1385](https://github.com/Rdatatable/data.table/issues/1385), and [#3723](https://github.com/Rdatatable/data.table/issues/3723), but none of those fixes have resolved this error from arising when using `dcast()`. This has been asked on StackOverflow [here (no MWE and no response)](https://stackoverflow.com/questions/50803331/integer64-not-assigned-na-values-during-data-tabledcast-but-defaults-to-921) and [here](https://stackoverflow.com/questions/55870524/unexpected-behavior-of-dcast-with-integer64-numbers) where the latter answer simply proposed a workaround, but should have been filed as an issue.

```
library(data.table)
apple <- fread("id, time, y
               a, 1, 12345678901234
               b, 1, 70
               b, 2, 20")
dcast(apple, id ~ time, value.var = "y")
   id              1                   2
1:  a 12345678901234 9218868437227407266
2:  b             70                  20
```

`#` `Output of sessionInfo()`
> R version 4.0.1 (2020-06-06)
Running under: Ubuntu 20.04 LTS

>Matrix products: default

>locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

>attached base packages:

>other attached packages:
[1] data.table_1.12.9 RPostgres_1.2.0  

>loaded via a namespace (and not attached):
 [1] compiler_4.0.1  bit_1.1-15.2    hms_0.5.3       tools_4.0.1     DBI_1.1.0      
 [6] Rcpp_1.0.4.6    bit64_0.9-7     vctrs_0.3.1     blob_1.2.1      pkgconfig_2.0.3
[11] rlang_0.4.6

MichaelChirico:
Closes #4561


inst/tests/tests.Rraw
```diff
@@ -15653,7 +15653,7 @@ test(2071.09, any_na(data.table(expression(1))), error="Unsupported column type
 test(2071.10, dcast(data.table(a=1, b=1, l=list(list(1))), a ~ b, value.var='l'),
               data.table(a=1, `1`=list(list(1)), key='a'))
 test(2071.11, dcast(data.table(a = 1, b = 2, c = 3), a ~ b, value.var = 'c', fill = '2'),
-              data.table(a=1, `2`=3, key='a'))
+              data.table(a=1, `2`=3, key='a'), warning="Coercing 'character' RHS to 'double'")
```
MichaelChirico, line 15656: I believe this warning is consistent with alerting user to string->double coercion elsewhere, but could be considered a breaking change, WDYT?

tdhock: consistent with what I wrote in #5549, I think this warning is confusing, because fill is not actually used in the output. I would like to merge that other PR, and if we do, then this warning will go away (and I guess we will have to modify this test).

If you want to test the coercion warning, I would suggest a more complicated case when the fill is actually used in the output, such as
```
> dcast(data.table(a = c(1,1,2), b = c(1,2,1), c = 3), a ~ b, value.var = 'c', fill='2')
Key: <a>
       a     1     2
   <num> <num> <num>
1:     1     3     3
2:     2     3     2
```
Does that make sense?

MichaelChirico: Makes perfect sense. IINM these tests are just about coverage, so your modified example will be better for now. Then hopefully this warning will go away after #5549 -- in fact we should make sure to include this as a test case there.

tdhock: lgtm

tdhock (approved):
lgtm

vlulla:
It appears that this still does not work with `colSums`/`rowSums`.

```R
> i64v = as.integer64(c(12345678901234, 70, 20, NA))
> apple = data.table(id = c("a", "b", "b"), time = c(1L, 1L, 2L), y = i64v[1:3])
> colSums(dcast(apple, id ~ time, value.var = "y")[,-1,with=FALSE])
            1             2
6.099576e-311 9.881313e-323
> colSums(dcast(apple, id ~ time, value.var = "y")[,-1,with=FALSE], na.rm=TRUE)
            1             2
6.099576e-311 9.881313e-323
> rowSums(dcast(apple, id ~ time, value.var = "y")[,-1,with=FALSE], na.rm=TRUE)
[1] 6.099576e-311 4.446591e-322
```
How do I get `colSums`/`rowSums` to work regardless of whether the column type is integer64 or numeric type? Thanks in advance.

MichaelChirico:
@vlulla that's not a data.table issue, same will happen for data.frame of integer64. Please file a bit64 bug.

PR #4595, opened 2020-07-10
uniqueN(by=character()) returns nrow, as does by=NULL

### LINKED ISSUE #4594: uniqueN() fails for zero-length vectors in by
If in `uniqueN` a zero-length vector is passed to `by`, the function throws an error. I would expect it to return 0.

```
library(data.table)
DT <- data.table(idx=1:4, value="val")
uniqueN(DT, character(0))
# Error in forderv(x, by = by, retGrp = TRUE, na.last = if (!na.rm) FALSE else NA) :
# Internal error: DT has 2 columns but 'by' is either not integer or is length 0
```

This was unexpected, as instead the following works:
```
uniqueN(DT[, .SD, .SDcols=character(0)])
# [1] 0
```

#### Output of `sessionInfo()`
```
R version 3.6.3 (2020-02-29)
Running under: Ubuntu 18.04.4 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:

attached base packages:

other attached packages:
[1] data.table_1.12.8

[1] compiler_3.6.3
```

MichaelChirico:
Closes #4594 

Not 100% sure this is the 'perfect' fix, but I think consistency with `by=NULL` makes sense, at least.


ColeMiller1:
Should ```unique.data.table()``` behave similar as well as any other duplicated type of operation take ```by = character(0L)``` as ```NULL```? If so, ```uniqueN()``` and ```unique()``` need work as well. I would lean towards resolving everything in the same PR for consistency...

``` r
library(data.table)
DT <- data.table(idx=1:4, value="val")

uniqueN(DT, by = character(0))
#> Error in forderv(x, by = by, retGrp = TRUE, na.last = if (!na.rm) FALSE else NA): Internal error: DT has 2 columns but 'by' is either not integer or is length 0
uniqueN(DT, by = NULL)
#> [1] 4

unique(DT, by = character(0L))
#> Error in forderv(x, by = by, sort = FALSE, retGrp = TRUE): Internal error: DT has 2 columns but 'by' is either not integer or is length 0
unique(DT, by = NULL)
#>    idx value
#> 1:   1   val
#> 2:   2   val
#> 3:   3   val
#> 4:   4   val

duplicated(DT, by = character(0L))
#> logical(0)
duplicated(DT, by = NULL)
#> [1] FALSE FALSE FALSE FALSE

anyDuplicated(DT, by = character(0L))
#> [1] 0
anyDuplicated(DT, by = NULL)
#> [1] 0
```

inst/tests/tests.Rraw
```diff
@@ -13299,8 +13299,7 @@ test(1962.004, duplicated(DT, by = -1L),
      error = 'specify non existing column*.*-1')
 test(1962.005, duplicated(DT, by = 'y'),
      error = 'specify non existing column*.*y')
-test(1962.0061, duplicated(data.table(NULL)), logical(0L))
-test(1962.0062, duplicated(data.table(a = 1L), by = character()), logical())
+test(1962.006, duplicated(data.table(NULL)), logical(0L))
```
MichaelChirico, line 13302: this `duplicated(., by=character())` was tested against for coverage before, we might count that as a breaking change?

cc @ColeMiller1 

if so, maybe want to sneak it into the upcoming release as "potentially breaking"

mattdowle:
We seem to have lost Travis CI, and therefore coverage.
I saw the news about Travis ... have we hit a quota or something?
@jangorecki any magic you can work?

PR #4609, opened 2020-07-17
Add row names to null data.table on assign

### LINKED ISSUE #4597: data.table not compatible with ggplot when it was generated from empty data.table
Hi,

I think I found a bug case where data.table does not work with ggplot. 
Here's an example:

```
dt = data.table()
dt[, x := seq(1,100)]
dt[, y := seq(100,1)]
head(dt)
   x   y
1: 1 100
2: 2  99
3: 3  98
4: 4  97
5: 5  96
6: 6  95

# generate ggplot

ggplot(dt, aes(x,y)) + geom_point()

Error in `$<-.data.frame`(x, name, value) : 
  replacement has 1 row, data has 0
```

A workaround was
```
dt %>% setDF %>% setDT # convert to DF and convert it back

ggplot(dt, aes(x,y)) + geom_point() # now it works
```
which doesn't seem convenient.

I'm not certain if it is coming from `data.table` or `ggplot`, for `plot(dt)` just works fine, but I guessed it was from `data.table` side because `setDT` somehow resolves the issue. Or was I naive to use empty `data.table` from scratch?

ColeMiller1:
Closes #4597.

I will do the news release after 1.13.0 is released. Also, will update test tomorrow to make this pass.

``` r
library(data.table)

dt = data.table()
dt[, new := 1:5]
attr(dt, "row.names") ##fixed!
#> [1] 1 2 3 4 5

## warning for now - easy to change to error with feedback. 
## data.table::groupingsets depended on this so there may be other issues with a hard change
dt = data.table(x = integer())
dt[, new := 1:5]

## Warning message:
## In `[.data.table`(dt, , `:=`(new, 1:5)) :
##   The data.table has 0 rows but there are 5 rows being assigned. The value will be ignored but new column(s) will be 
##   added. This warning will be upgraded to error in next release
dt
#> Empty data.table (0 rows and 2 cols): x,new
```


ColeMiller1:
@jangorecki  let me know if I should split this up. The updating of row.names of `data.table()[, new_var := 1:5]` makes sense. However, I'm uncertain about `data.table(x = integer())[, new_var := 1]`. Current behavior is a zero-length variable being assigned. Proposed behavior is a warning with long-term aim of being an error. I did it as a single PR because they were relatively closely related.

mattdowle:
Great fix, thanks.
As you alluded to above, and influenced by recent #4262, I went ahead and removed the new warning and reverted the changed tests. I'm (now) viewing `data.table(x = integer())[, new_var := 1]` as recycling of the length-1 to match the length-0 data. Happily, this is already what the tests tested, so my change to this PR brings that side of it back to current behavior so it can be merged to fix the `row.names` bug. Can always reconsider, discuss, and have another PR in future as you said.

PR #4610, opened 2020-07-17
Translations update

MichaelChirico:
I think we're just about code frozen for the release, so I ran `tools::update_pkg_po('.')` to expose the delta of messages since the original / main translations PR. I'll go to the translation team now & ask for help.


po/R-zh_CN.po
```diff
@@ -1896,29 +1905,40 @@ msgstr ""
...
+
 msgid ""
-"**********\n"
 "This installation of data.table has not detected OpenMP support. It should "
 "still work but in single-threaded mode."
 msgstr ""
-"**********\n"
 "data.table的安装未检测到OpenMP支持。在单线程模式下应该仍能运行"
 
+msgid "sysname"
```
MichaelChirico, line 1917: I think it's an R bug that these show up as translatable. I filed: https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17860

For now just supplied identity translations

PR #4613, opened 2020-07-17
Onattach translations

MichaelChirico:
A quick follow up to #4610 (despite the large diff).

Basically trying to address the issue raised in https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17860 by removing non-strings from translation.

Also added some notes for release about bumping the po header version tag.


R/onAttach.R
```diff
@@ -18,20 +18,24 @@
   }
   dev = as.integer(v[1L, 3L]) %% 2L == 1L  # version number odd => dev
   if (!isTRUE(getOption("datatable.quiet"))) {   # new option in v1.12.4, #3489
-    packageStartupMessage("data.table ", v, if(dev)paste0(" IN DEVELOPMENT built ",d,g),
-                          " using ", getDTthreads(verbose=FALSE), " threads (see ?getDTthreads).  Latest news: r-datatable.com")
+    dev_tag = if (dev) gettextf(" IN DEVELOPMENT built %s%s", d, g, domain="R-data.table") else ""
```
MichaelChirico, line 21: a bit ugly but `xgettext` ignores the strings passed to `...` of `gettextf` for translation -- even if those strings are `gettext` themselves. so we have to pull this translation out of the `gettextf` call for it to work.

ditto below

po/R-zh_CN.po
```diff
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
-"Project-Id-Version: data.table 1.12.5\n"
-"POT-Creation-Date: 2020-07-17 14:38\n"
+"Project-Id-Version: data.table 1.12.9\n"
```
MichaelChirico, line 3: strange that this doesn't bump automatically. it seems to do so in the .pot file.

MichaelChirico:
Superseded by #5056

PR #4623, opened 2020-07-21
OSX Travis

mattdowle:
I turned on OSX Travis as per CRAN_Release.
The output appears pretty helpful on first glance so I started following its advice, but as you scroll down there lots of this "*may* need to add" language so I guess it's progressing ok, but I'm not sure.
Why it fails at the last step is a mystery, or what we should do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
 language: r
-dist: trusty
-sudo: required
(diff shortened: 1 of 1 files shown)


mattdowle:
Any ideas by any chance, @MarcusKlik?

PR #4627, opened 2020-07-21
GCC / Fortran conflict in OSX Travis build

MarcusKlik:
In respons to #4623:

Hi @mattdowle, it seems there is a conflict between the `llvm` and `fortran` installations, see [this line](https://travis-ci.org/github/Rdatatable/data.table/jobs/710240029#L642-L648) in the Travis output. 

To fix, simplify and update the Travis build, I've:

* added a line to the _before\_install_ section with a command to remove the `fortran` symlink.
* moved the `llvm` installation to a new _brew\_packages_ section (only a single line needed to install `llvm`).
* updated the linux image to the more recent _bionic_ (if there was a specific reason to keep using _trusty_, my apologies for that :-))

I believe that [should fix](https://travis-ci.org/github/MarcusKlik/data.table.test/jobs/710337647) the OSX build!

all the best

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
 language: r
-dist: trusty
-sudo: required
+dist: bionic
 cache: packages   # to rebuild cache see tweet thread ending here https://twitter.com/jimhester_/status/1115718589804421121
 warnings_are_errors: true
-
-branches:
-  only:
-    - "master"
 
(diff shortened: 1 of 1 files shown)


mattdowle:
Thanks, @MarcusKlik!! You're the best!

PR #4631, opened 2020-07-22
reworked fmelt.c:concat to survive gctorture

### LINKED ISSUE #4628: strict gctorture results to work through
With R-devel compiled with clang-10, strict barrier and gctorture too (see CRAN_Release.cmd for commands). Hasn't finished yet, but this is output so far is below.
The 1036.* fails probably are knock-on effects from the first 1036.02.
Unlikely to be a memory fault related to gctorture. More likely some extra strict checks turned on when strict barrier is on too, possibly new in latest R-devel.
```
> print(Sys.time()); started.at<-proc.time(); try(test.data.table()); print(Sys.time()); print(timetaken(started.at))
[1] "2020-07-21 00:39:28.074113 MDT"
getDTthreads(verbose=TRUE):
test.data.table() running: /home/mdowle/build/R-devel-strict-clang/library/data.table/tests/tests.Rraw 
Running test id 1036.02      Test 1036.02 didn't produce the correct warning :
Expected: All measure variables not of type 'character' will be coerced 
Observed: type 31 is unimplemented in 'type2char' 
Test 1036.02 produced 1 errors but expected 0
Expected: 
Observed: CAR/CDR/TAG or similar applied to unknown type #31 object
Running test id 1036.03      Test 1036.03 ran without errors but failed check that x equals y:
> x = length(levels(ans$variable)) 
First 1 of 1 (type 'integer'): 
[1] 100
> y = 317L 
First 1 of 1 (type 'integer'): 
[1] 317
Mean relative difference: 2.17
Running test id 1036.04      Test 1036.04 ran without errors but failed check that x equals y:
> x = levels(ans$variable)[c(1, 2, 316, 317)] 
First 4 of 4 (type 'character'): 
[1] "V1" "V2" NA   NA  
> y = tt <- c("Geography", "Estimate; SEX AND AGE - Total population",      "Percent; HISPANIC OR LATINO AND RACE - Total housing units",      "Percent Margin of Error; HISPANIC OR LATINO AND RACE - Total housing units") 
First 4 of 4 (type 'character'): 
[1] "Geography"                                                                 
[2] "Estimate; SEX AND AGE - Total population"                                  
[3] "Percent; HISPANIC OR LATINO AND RACE - Total housing units"                
[4] "Percent Margin of Error; HISPANIC OR LATINO AND RACE - Total housing units"
'is.NA' value mismatch: 0 in current 2 in target
Running test id 1036.05      Test 1036.05 ran without errors but failed check that x equals y:
> x = range(as.integer(ans$variable)) 
First 2 of 2 (type 'integer'): 
[1]   1 100
> y = INT(1, 317) 
First 2 of 2 (type 'integer'): 
[1]   1 317
Mean relative difference: 2.17
Running test id 1036.06      Test 1036.06 ran without errors but failed check that x equals y:
> x = as.vector(table(table(as.integer(ans$variable)))) 
First 1 of 1 (type 'integer'): 
[1] 100
> y = INT(309, 8) 
First 2 of 2 (type 'integer'): 
[1] 309   8
Numeric: lengths (1, 2) differ
Running test id 1036.07      Test 1036.07 ran without errors but failed check that x equals y:
> x = sapply(ans, class) 
First 2 of 2 (type 'character'): 
 variable     value 
 "factor" "integer" 
> y = c(Id = "character", Id2 = "integer", variable = "factor", value = "character") 
First 4 of 4 (type 'character'): 
         Id         Id2    variable       value 
"character"   "integer"    "factor" "character" 
Names: 2 string mismatches
Lengths (2, 4) differ (string compare on first 2)
1 string mismatch
Running test id 1372.26  

... still running

```

mattdowle:
Closes #4628 
Confirmed that all parts of test 1036 now pass full strength gctorture(TRUE) with latest R-devel compiled strictly.
concat was marked as a hack anyway so needed attention. If I was forced to guess I'd guess it was that `SETCAR(t, mkString(", "));` doesn't protect the result of `mkString` but I'd be surprised if it doesn't. Initially I protected the result of concat before it was passed on to warning/Rprintf which definitely needed doing, but it wasn't just that (it still failed torture). Not sure what else it could be as I don't see anything else missing protection. The dev cycle time to resolve this one was significant due to the slowness of torture so I just went for the rework approach which needed doing anyway.   The new approach for concat is simpler and cleaner, but we need to make sure not to use it twice in the same warning/Rprintf call (comment added).
Several of the fmelt tests weren't checking the dynamic parts of their messages; i.e. the result of concat being correct, so I expanded those tests too. I checked that the expanded fmelt tests also pass v1.12.8 to ensure the reworked concat returns the same string as before. concat is only used in warning and verbose messages, and only in fmelt, so the risk of this code rework at this stage in release is limited. Reran rchk just to make sure nothing new introduced, and all fine.

### DIFF
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -61,32 +61,40 @@ SEXP whichwrapper(SEXP x, SEXP val) {
   return which(x, LOGICAL(val)[0]);
 }
 
-// hack by calling paste using eval. could change this to strcat, but not sure about buffer size for large data.tables... Any ideas Matthew?
-SEXP concat(SEXP vec, SEXP idx) {
-
-  SEXP s, t, v;
-  int nidx=length(idx);
-
-  if (TYPEOF(vec) != STRSXP) error(_("concat: 'vec must be a character vector"));
+static const char *concat(SEXP vec, SEXP idx) {
+  if (!isString(vec)) error(_("concat: 'vec must be a character vector"));
   if (!isInteger(idx) || length(idx) < 0) error(_("concat: 'idx' must be an integer vector of length >= 0"));
+  
+  static char ans[1024];  // so only one call to concat() per calling warning/error
+  int nidx=length(idx), nvec=length(vec);
+  ans[0]='\0';
+  if (nidx==0) return ans;
   const int *iidx = INTEGER(idx);
-  for (int i=0; i<length(idx); ++i) {
-    if (iidx[i] < 0 || iidx[i] > length(vec))
-      error(_("Internal error in concat: 'idx' must take values between 0 and length(vec); 0 <= idx <= %d"), length(vec)); // # nocov
+  for (int i=0; i<nidx; ++i) {
+    if (iidx[i]<1 || iidx[i]>nvec)
+      error(_("Internal error in concat: 'idx' must take values between 1 and length(vec); 1 <= idx <= %d"), nvec); // # nocov
   }
-  PROTECT(v = allocVector(STRSXP, nidx > 5 ? 5 : nidx));
-  for (int i=0; i<length(v); ++i) {
-    SET_STRING_ELT(v, i, STRING_ELT(vec, iidx[i]-1));
+  if (nidx>4) nidx=4;  // first 4 following by ... if there are more than 4
+  int remaining=1018;  // leaving space for ", ...\0" at the end of the 1024, potentially
+  char *pos=ans;
+  int i=0;
+  for (; i<nidx; ++i) {
+    SEXP this = STRING_ELT(vec, iidx[i]-1);
+    int len = length(this);
+    if (len>remaining) break; 
+    strncpy(pos, CHAR(this), len);
+    pos+=len;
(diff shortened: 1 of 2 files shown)


src/fmelt.c
```diff
@@ -61,32 +61,40 @@ SEXP whichwrapper(SEXP x, SEXP val) {
...
 
-// hack by calling paste using eval. could change this to strcat, but not sure about buffer size for large data.tables... Any ideas Matthew?
-SEXP concat(SEXP vec, SEXP idx) {
-
-  SEXP s, t, v;
-  int nidx=length(idx);
-
-  if (TYPEOF(vec) != STRSXP) error(_("concat: 'vec must be a character vector"));
+static const char *concat(SEXP vec, SEXP idx) {
+  if (!isString(vec)) error(_("concat: 'vec must be a character vector"));
```
MichaelChirico, line 65: should be `'vec'`, i'll file follow-up

PR #4632, opened 2020-07-23
small typo

MichaelChirico:
Follow-up to #4631


mattdowle:
Good spot. If we have to resubmit to CRAN (there's always something) I'll put in.
Does this typo affect translation lookup?

MichaelChirico:
no. either (1) this is a new message & there's no template for it yet or
(2) the existing template simply won't be found (it's the same as if it's a
new message, just with some wasted space in the lookup files)

PR #4635, opened 2020-07-23
fwrite matrix input without as.data.table

### LINKED ISSUE #3188: fwrite: avoid converting matrix input to data.table
PR #3125 implemented #2613 to enable to `fwrite` to accept `matrix`.  This was merged in v1.12.0. However it converts matrix to data.table which can be costly.
The convert can be avoided as follows.
`fwriteR.c` contains this : 
```
    SEXP column = VECTOR_ELT(DFcoerced, j);
    if (args.nrow != length(column))
      error("Column %d's length (%d) is not the same as column 1's length (%d)", j+1, length(column), args.nrow);
    int32_t wf = whichWriter(column);
    if (wf<0) {
      error("Column %d's type is '%s' - not yet implemented in fwrite.", j+1, type2char(TYPEOF(column)));
    }
    args.columns[j] = (wf==WF_CategString ? column : (void *)DATAPTR(column));
    args.whichFun[j] = (uint8_t)wf;
    if (TYPEOF(column)==VECSXP && firstListColumn==0) firstListColumn = j+1;
  }
```
This loop populates the pointers in `arg.columns[]` which are then passed to `fwrite.c`.  That DATAPTR returns a pointer to where the data (e.g. `int *`, `double *`) for the R vector starts.  In the case of matrix, this `args.columns[]` just needs to be populated with offsets into matrix vector. In R a matrix is a single very long vector just with dimension attribute attached. A matrix in R is columnar, just like a data.table,  so there doesn't need to be a transpose. It should be fairly simple to do and work well.

Careful to add tests for a factor matrix as well as a character matrix.

MichaelChirico:
Closes #3188

still quite exploratory, not any real progress, just parking this here to get it off this computer. anyone else feel free to pick up the baton if i don't get to it.

### DIFF
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -31,9 +31,8 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
   nThread = as.integer(nThread)
   # write.csv default is 'double' so fwrite follows suit. write.table's default is 'escape'
   # validate arguments
-  if (is.matrix(x)) { # coerce to data.table if input object is matrix
-    message("x being coerced from class: matrix to data.table")
-    x = as.data.table(x)
+  if (is.matrix(x)) {
+    x = list(x)
   }
   stopifnot(is.list(x),
     identical(quote,"auto") || isTRUEorFALSE(quote),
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Don't think there's anything here that can't be pretty easily re-devised when we get to it.

Issue #4643, opened 2020-07-26
Add a API definining header hiding the function pointer as commonly done


eddelbuettel:
Congratulations on getting 1.3.0 out. It is impressive to watch your well-oiled release process which I should do more often :)

It is lovely that the beginnings of a C API are now out.  Big thanks to @lsilvest for getting it going and all of you for making it happen with #3751 and building on the older #1694.  

This is immediatly useable in something like the following (and I apologise for bring C++ and Rcpp in but it makes one-liners, here broken for readability, easier to test)

```r
R> Rcpp::cppFunction(' SEXP mysub(SEXP x, SEXP rows, SEXP cols) { 
  SEXP(*fun)(SEXP, SEXP, SEXP) = (SEXP(*)(SEXP,SEXP,SEXP)) R_GetCCallable("data.table", "CsubsetDT"); 
  return fun(x,rows,cols);
}')
R>
```

As your documentation states, if one is familiar with what WRE says or willing to read, this is easy to use as we proxy here from R giving us the desired subsetter from C(++).

```r
R> library(data.table)
R> dt <- data.table(iris)
R> mysub(dt, 1:4, 1:4)
   Sepal.Length Sepal.Width Petal.Length Petal.Width
1:          5.1         3.5          1.4         0.2
2:          4.9         3.0          1.4         0.2
3:          4.7         3.2          1.3         0.2
4:          4.6         3.1          1.5         0.2
R> 
```

Now, it is not uncommon to export APIs such as this---and fully understood it is subject to change and will change---by an API defining header.  I have done that twice in packages exporting C API portions from R itself ([RApiDatetime](https://github.com/eddelbuettel/rapidatetime), [RApiSerialize](https://github.com/eddelbuettel/rapiserialize)) and added it to other packages ([xts](https://github.com/joshuaulrich/xts).  In each case a header was added which client packages include and which makes the _C function pointer assignment and SEXP-counting_ internal to a simpler helper function.  With it (and I have a version here which I could PR) the C(++) snippet becomes 

```r
R> Rcpp::cppFunction("SEXP mysub2(SEXP x, SEXP rows, SEXP cols) { return subsetDT(x,rows,cols); }", 
     include="#include <datatableAPI.h>", 
     depends="data.table")
R> mysub2(dt, 1:4, 1:4)
   Sepal.Length Sepal.Width Petal.Length Petal.Width
1:          5.1         3.5          1.4         0.2
2:          4.9         3.0          1.4         0.2
[...]

PR #4645, opened 2020-07-26
Add header to declare C level API 

eddelbuettel:
This PR follows the discussion in #4643 and adds a (very simply and short) file declaring a C level function to access the registered-as-callable function `subsetDT`.  But providing a function which shields the somewhat unwieldy C function pointer setup, wider use is facilitated.   It also offers a benefit for maintenance as it provides another user facing layer allowing the data.table team to redefine or rename the internal function registers as C callable and called here.

At present the function can called three ways

- "as is":  `DT_subsetDT()` with an explicit prefix (and we could consider dropping the suffix...)
- "with an additional define":  `subsetDT()` which is a little cleaner but may clash, this is an opt-in
- "from C++":  `dt::subsetDT()` where we put a simple C++ namespace around it

Please free to rename or realign as you see fit.  I have tested all three access mechanisms (in a simple Rcpp-based test).  Doing a proper unit test is likely more work but maybe something we should do.  I can help...

### DIFF
--- a/inst/include/datatableAPI.h
+++ b/inst/include/datatableAPI.h
@@ -1,0 +1,45 @@
+
+/* This header file provides the interface used by other packages,
+   and should be included once per package.                        */
+
+#ifndef _R_data_table_API_h_
+#define _R_data_table_API_h_
+
+/* number of R header files (possibly listing too many) */
+#include <R.h>
+
+#ifdef HAVE_VISIBILITY_ATTRIBUTE
+    # define attribute_hidden __attribute__ ((visibility ("hidden")))
+#else
+    # define attribute_hidden
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* provided the interface for the function exported in
+   ../src/init.c via R_RegisterCCallable()		*/
+
+SEXP attribute_hidden DT_subsetDT(SEXP x, SEXP rows, SEXP cols) {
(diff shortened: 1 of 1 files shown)


mattdowle:
Thanks, @eddelbuettel! I'll invite you to be a project member so you can create branches directly in the main project in future if you wish, and I'll add you to DESCRIPTION in a follow up commit to master (slightly quicker for me than adding that to your PR).

inst/include/datatableAPI.h
```diff
@@ -0,0 +1,45 @@
...
+extern "C" {
+#endif
+
+/* provided the interface for the function exported in
+   ../src/init.c via R_RegisterCCallable()		*/
+
+SEXP attribute_hidden DT_subsetDT(SEXP x, SEXP rows, SEXP cols) {
+     static SEXP(*fun)(SEXP, SEXP, SEXP) =
+       (SEXP(*)(SEXP,SEXP,SEXP)) R_GetCCallable("data.table", "CsubsetDT");
+     return fun(x,rows,cols);
```
jangorecki, line 27: is it necessary to create `fun` object here? couldn't this be a one liner to returning `R_GetCCallable()` call?

eddelbuettel: Yes: `fun` is what is invoked when you call.  The `static` makes the assignment stick, and the `R_GetCCallable()` ges the identifier from the _other_ package.

Issue #4654, opened 2020-07-31
bit64 update on CRAN breaks data.table


mattdowle:
No sooner have we started to pass tests on CRAN, are they broken again.
data.table updated to v1.13.0 on 24 July.  Then bit64 updated to 4.0.2 on 30th July.
I wrote to Kurt and Jens:
> Looked further and I suppose it's because CRAN policy states that the reverse dependencies are returned by `tools::package_dependencies(reverse = TRUE)`. The default for its `which=` does not include Suggests. data.table suggests bit64 which is why I assume data.table as a revdep of bit64 was not tested by bit64, nor CRAN's revdep testing upon submission of bit64. When I run revdep checks I set `which='all'` but that's not in CRAN policy to do so.  I think that's a failing of policy because Suggests are needed for packages to pass their tests; that's what the Suggests category is there for amongst other things.
Kurt - do you think CRAN policy could be updated to include Suggests in revdep testing?

ColeMiller1:
@mattdowle there seem to be two issues that led to the break.

1. ```bit``` 4.0.0 included a new ```copy()``` function which masked ```data.table::copy``` and resulted in our test scripts to fail.
2. ```bit64``` 4.0.0 included a new ```all.equal.integer64``` method that is more accurate - specifically, it no longer provides incorrect results with respect to ```NA``` values. This fix uncovered two unit tests which had wrong expected results.

#4656 passes on our CI builds. Hopefully that will help get us back to passing.

mattdowle:
@ColeMiller1  Thanks for investigating. I had correspondence on email with Jens too.  He's kindly suggested to renamed his copy, which would be better for users since data.table::copy is long standing.

---

Jens,

I've looked at the two tests you highlighted: 2127.26 and 2127.27.
They were passing before because in previous version of bit64 :

> all.equal( as.integer64(3), as.integer64(4) )
[1] TRUE
> all.equal( as.integer64(3), as.integer64(NA) )
[1] TRUE

That seems to be a bug that you've fixed in the new version. New version result :

> all.equal( as.integer64(3), as.integer64(4) )
[1] "Mean relative difference: 0.3333333"
> all.equal( as.integer64(3), as.integer64(NA) )
[1] "'is.NA' value mismatch: 1 in current 0 in target"

I'm surprised this is classified as a 'new feature' in bit64 v4.0.0.
All the NEWS file says is: NEW FEATURES

    o new method all.equal.integer64
[...]

PR #4655, opened 2020-08-01
[[ by group performance

### LINKED ISSUE #4424: internal anyNamed no longer returns false?

### LINKED ISSUE #4646: [[ by group takes forever (24 hours +) with v1.13.0 vs 4 seconds with v1.12.8
I have a script that takes only 4 seconds to process in V 1.12.8, and more than 24+ hours when using V 1.13.0. All other variables are the same (computer, r and RStudio versions, dataset, etc.)

All relevant and detailed information about this issue is contained in [Stack Overflow](https://stackoverflow.com/questions/63105711/why-data-table-unnesting-time-grows-with-the-square-of-number-of-rows-for-a-spec).

https://stackoverflow.com/questions/63105711/why-data-table-unnesting-time-grows-with-the-square-of-number-of-rows-for-a-spec
I am not a programmer, so I am not able to go through the code. Below you will find the sessionInfo() reg\rding thje environment AFTER changing back to V 1.12.8.

Please let me know whether I can be of any further help.

BTW, congratulations for this amazing gamecahnger package.

Regards,

Fabio.

> sessionInfo()
R version 4.0.2 (2020-06-22)

Matrix products: default

locale:
[1] LC_COLLATE=Portuguese_Brazil.1252  LC_CTYPE=Portuguese_Brazil.1252    LC_MONETARY=Portuguese_Brazil.1252
[4] LC_NUMERIC=C                       LC_TIME=Portuguese_Brazil.1252    

attached base packages:

other attached packages:
[1] microbenchmark_1.4-7 data.table_1.12.8    lubridate_1.7.9      stringi_1.4.6        runner_0.3.7         e1071_1.7-3         
[7] ggplot2_3.3.2        stringr_1.4.0        magrittr_1.5        

 [1] Rcpp_1.0.5       pillar_1.4.6     compiler_4.0.2   class_7.3-17     tools_4.0.2      digest_0.6.25    packrat_0.5.0    evaluate_0.14   
 [9] lifecycle_0.2.0  tibble_3.0.3     gtable_0.3.0     pkgconfig_2.0.3  rlang_0.4.7      rstudioapi_0.11  yaml_2.2.1       xfun_0.16       
[17] withr_2.2.0      dplyr_1.0.0      knitr_1.29       generics_0.0.2   vctrs_0.3.2      grid_4.0.2       tidyselect_1.1.0 glue_1.4.1      
[25] R6_2.4.1         rmarkdown_2.3    purrr_0.3.4      scales_1.1.1     ellipsis_0.3.1   htmltools_0.5.0  colorspace_1.4-1 tinytex_0.25    
[33] munsell_0.5.0    crayon_1.3.4

### LINKED ISSUE #4658: 1.13.0 slow down in a repeated loop on list column
I have noticed a huge performance drop in data.table loop operations, possibly due to the new version upgrade

```
library(data.table)
library(microbenchmark)

dt <- data.table('id'=1:20000,
                 'list_col'=sample(c('', '', 'a', 'a:b', 'a:b:c'), 20000, TRUE))
feature <- 'list_col'

microbenchmark(
  long_dt <- dt[, c("id", feature), with = FALSE][
    , feature_names := {
      x <- get(feature)
      stringr::str_split(x, ':')
    }][
      , .(
        feature_names = paste0(feature, "_", unlist(feature_names))
      )
      , by = "id"]
  , times = 10
  , unit = 'ms'
)
```
data.table 1.12.8, default settings, using 6 threads:
```
      min       lq     mean   median       uq      max neval
 122.2447 149.6991 173.3268 183.5777 193.9876 201.7234    10
```
data.table 1.13.0, default settings, using 6 threads:

```
      min      lq     mean   median      uq      max neval
 12820.75 12913.1 12989.59 13007.94 13065.1 13097.85    10
```

Also, I have tried several different threads and throttle combinations, but have seen no improvements at all

ColeMiller1:
Closes #4646
Closes #4658
Closes #4424
``` r
library(data.table)

set.seed(123L)
n = 500L
n_nested = 40L
dt = data.table(ordered_id = seq_len(n),
                unordered_id = sample(n),
                value = replicate(n, data.table(val1 = sample(n_nested)), simplify = FALSE))

bench::mark(
    dt[, value[[1L]], by = ordered_id]
    , dt[, value[[1L]], by = unordered_id]
    , check = FALSE
    , time_unit = "ms" 
)

## pre-regression (1.12.8)

##  expression                             min median `itr/sec` mem_alloc
##  <bch:expr>                           <dbl>  <dbl>     <dbl> <bch:byt>
##1 dt[, value[[1L]], by = ordered_id]   0.755  0.838     1131.     401KB
##2 dt[, value[[1L]], by = unordered_id] 0.796  0.892     1033.     409KB

## Post-regression (CRAN 1.13.0)

#> # A tibble: 2 x 6
#>   expression                               min  median `itr/sec` mem_alloc
#>   <bch:expr>                             <dbl>   <dbl>     <dbl> <bch:byt>
#> 1 dt[, value[[1L]], by = ordered_id]   102.    126.         8.02    53.7MB
#> 2 dt[, value[[1L]], by = unordered_id]   0.630   0.669   1261.     409.2KB
#> # ... with 1 more variable: `gc/sec` <dbl>

## after patch

##  expression                             min median `itr/sec` mem_alloc `gc/sec`
##  <bch:expr>                           <dbl>  <dbl>     <dbl> <bch:byt>    <dbl>
##1 dt[, value[[1L]], by = ordered_id]   0.634  0.691     1358.     401KB     15.0
##2 dt[, value[[1L]], by = unordered_id] 0.651  0.778     1187.     409KB     12.8
```

The root cause appears to be in ```memrecycle``` which is in ```assign.c```. I can get similar performance by commenting out the ```duplicate``` line but then there are many errors during the testing. I'm including this in case someone is able to move forward with fixing the root cause.

https://github.com/Rdatatable/data.table/blob/9d3b9202fddb980345025a4f6ac451ed26a423be/src/assign.c#L707-L723


ColeMiller1:
Both the Bionic and MacOS builds are failing on 

```
dt1 = copy(dt)
## Error in copy(dt) : SEXP is not atomic vector
```

After looking into it, these errors seem related to #4654 and the bit64 update.

mattdowle:
I followed Cole's first inclination; i.e. removed the copy from `memrecycle` as he did which seems right to me too.
That fixes the slowdown problem in subsetting but leaves us with a different problem to solve in a different way: all the test fails were to do with static vectors not being copied, which is why that copy was there in memrecycle. So I moved that copy into dogroups.c to be close to where it's relevant. `dogroups` now uses negative truelength to mark the specials (.SD, .BY, .I). Just before assigning `jval` to a list column, it checks to see if any specials are present and copies them if so. Although it's still relatively complex, it's cleaner in the sense that the complexity is now focused and localized to dogroups.c. This should avoid unexpected effects like the slowdown occurring due to the copy in the central place (memrecycle) which is now gone.
Extensive comments and example added to new `anySpecialStatic()` at the top of dogroups.c which replaces the `anyNamed()` in assign.c.
Other tidy ups:
Removed `origSDnrow` as it was equal to `maxGrpSize`, adding a check that that's true.
`origIlen` could also be removed because `I` was allocated with `maxGrpSize` anyway.

jangorecki:
this will also resolve #4424

mattdowle:
You mean .NGRP? That's constant across groups so is the odd one out.  But tested anyway (2153.3) and referred to in comment in anySpecialStatic.

PR #4660, opened 2020-08-04
Error handling for using a factor with `i`

### LINKED ISSUE #1632: handling i which is a factor
It happens quite often to me that column which I passed as `i`, turns out to be a factor. Current behavior [is to throw an error](https://github.com/Rdatatable/data.table/blob/master/R/data.table.R#L705): "_i has not evaluated to logical, integer or double_", so I have to manually convert it to the same class as the `key` of the target table.
1. One option would be just to add this conversion to `"[.data.table"` function code. I guess there are some reasons against as it have not been done yet, would be glad to hear it.
2. When it happened first time, it took me some time to figure out what was the problem. It would be helpful to make error message more informative, just add `class(i)` at the end of message in `stop()` call.

ben-schwen:
Closes #1632.

Added an additional check for `is.factor` to avoid the confusing error "i has evaluated to type integer. Expecting logical, integer or double." when executing `setkey(data.table(a=factor("a")))[factor("a")]`.


mattdowle:
Replaced by #5120

PR #4671, opened 2020-08-15
Add inline to function definition

eddelbuettel:
This is something I should have picked up ages but didn't because most often these API include headers are used only in one file / compilation unit.

And when for once it is not, we get duplicate symbols :-/

So this just adds a missing `inline` that should have been present in the original PR.

### DIFF
--- a/inst/include/datatableAPI.h
+++ b/inst/include/datatableAPI.h
@@ -21,7 +21,7 @@ extern "C" {
 /* provided the interface for the function exported in
    ../src/init.c via R_RegisterCCallable()		*/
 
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Poking evyerone  This is trivial one-word fix, and as @jangorecki pointed out in #4672 the fail is unrelated to the PR but due to some macOS setup issue.   I would appreciate some feedback.

eddelbuettel:
Another gentle poke.  This is a one-line one-word change adding the missing `inline` qualifier.

eddelbuettel:
Thanks guys, and truly sorry for having missed `inline` last round.  

I really did write these headers the wrong way without for years before being bitten by a dual inclusion.  Silly me!

PR #4692, opened 2020-09-03
mirror fix to base R for as.IDate

### LINKED ISSUE #4676: as.IDate error that can be fixed
```r
require(data.table) # 1.13.1 from today
x <- c("2020-01-01", "")
y <- c("", "2020-01-01")
as.IDate(x) # works fine
as.IDate(y)
# Error in charToDate(x) : 
#   character string is not in a standard unambiguous format
```

MichaelChirico:
Closes #4676 

Actually I wonder if we should branch on the base R capability... for a user on bleeding-edge R-devel, won't they end up running this line of code twice?

```
is.na(x) = !nzchar(x)
```

Once in `as.IDate.default` and then once in `as.Date.character`?

In that case, we can branch on:

```
is.na(tryCatch(as.Date(''), error=function(e) 1L))
```

Or we can store the result of that test in the `.global` environment.


MichaelChirico:
Looks like there was some change to the coverage metrics for C code cc @jangorecki

MichaelChirico:
@jangorecki it looks like all (or most) of the changes are things like

```
default:
    stop(_("Internal error")); // # nocov
```

The `default:` line was covered before, isnt' now.

We can either add a `// # nocov` to the `default:` line, or we can move the `stop()` to the `default` line. Any opinion on which is better?

jangorecki:
I would start by checking if it is a regression or an enchantment in codecov.

MichaelChirico:
@jangorecki AFAICT no difference in Codecov (not sure how to tell). There may have been a change in `gcov` or maybe we/they changed `gcc` version? Also not sure how to tell that.

I don't see anything particularly noteworthy recently in gcov.c, I don't know what else might be the cause:

https://github.com/gcc-mirror/gcc/commits/master/gcc/gcov.c

https://gcc.gnu.org/git/?p=gcc.git&a=search&h=HEAD&st=commit&s=gcov

inst/tests/tests.Rraw
```diff
@@ -17537,3 +17537,8 @@ vr = "Species"
 IDT[, virginca := get(vr) == "virginica"]
 ans = data.table(round = c(3, 3, 3, 2, 2, 4, 2, 4), k = c(6, 7, 8, 5, 7, 7, 6, 8), kar = structure(c(3L, 3L, 3L, 3L, 3L, 3L, 3L, 3L), .Label = c("setosa", "versicolor", "virginica"), class = "factor"), N = c(24L, 14L, 4L, 1L, 1L, 1L, 3L, 2L))
 test(2184.5, IDT[(virginca), .N, by = .(round(Sepal.Width), k = round(Sepal.Length), kar = get(vr))] , ans)
+
+# initial '' doesn't crash as.IDate, #4676;
+#   also imitates base R as from 4.0.2 patched
+y = c('', '2020-01-01')
+test(2185, as.IDate(y), rev(as.IDate(rev(y))))
```
mattdowle, line 17544: This test compares `as.IDate` to itself, so it could pass incorrectly. Tests should compare to a fixed value.

This was the main point of me requesting-changes on this PR.  Then the other relatively minor requests I added too while I was at it.

R/IDateTime.R
```diff
@@ -7,6 +7,10 @@ as.IDate = function(x, ...) UseMethod("as.IDate")
 
 as.IDate.default = function(x, ..., tz = attr(x, "tzone", exact=TRUE)) {
   if (is.null(tz)) tz = "UTC"
+  # 4676 mimics for back-compatibility a similar patch applied to as.Date.character in r79119
```
mattdowle, line 10: since changes are needed anyway, this comment seems more appropriate inside if the if().  Since the if() has been elongated to 3 lines anyway, may as well provide an extra reason for the elongation by putting the comment inside.

NEWS.md
```diff
@@ -88,6 +88,8 @@
 
 14. `by=...get()...` could fail with `object not found`, [#4873](https://github.com/Rdatatable/data.table/issues/4873) [#4981](https://github.com/Rdatatable/data.table/issues/4981). Thanks to @sindribaldur for reporting, and @OfekShilon for fixing.
 
+15. `as.IDate` works on character input where the first input is `''` by treating it as `NA`, [#4676](https://github.com/Rdatatable/data.table/issues/4676). This is consistent with how `''` is treated when it appears later in the input; the underlying inconsistency in `as.Date.character` was fixed in R version 4.0.3; we applied the change in `data.table` as well because we depend only on a much earlier version of R.
```
mattdowle, line 91: The words 'works' needs replacing with detail. Overall the news item could be maybe half as long. The skeleton that springs to mind is something like
> `as.IDate(c('',somedate))` now returns ... instead of .... This is a backport of rXXX from R 4.0.3 so that we can continue to support and test data.table on R 3.1.0+.

inst/tests/tests.Rraw
```diff
@@ -17537,3 +17537,8 @@ vr = "Species"
 IDT[, virginca := get(vr) == "virginica"]
 ans = data.table(round = c(3, 3, 3, 2, 2, 4, 2, 4), k = c(6, 7, 8, 5, 7, 7, 6, 8), kar = structure(c(3L, 3L, 3L, 3L, 3L, 3L, 3L, 3L), .Label = c("setosa", "versicolor", "virginica"), class = "factor"), N = c(24L, 14L, 4L, 1L, 1L, 1L, 3L, 2L))
 test(2184.5, IDT[(virginca), .N, by = .(round(Sepal.Width), k = round(Sepal.Length), kar = get(vr))] , ans)
+
+# initial '' doesn't crash as.IDate, #4676;
+#   also imitates base R as from 4.0.2 patched
```
mattdowle, line 17542: this comment can be on one line please

mattdowle (changes requested):

inst/tests/tests.Rraw
MichaelChirico: thinking here was that the bug is about inconsistency in the first place -- the order within `y` matters for the output. so this test confirms the order doesn't matter.

for that reason i'll leave it, but i agree we should also test against a fixed value, so I'll add a second test.

mattdowle:
@michaelchirico seems like this one is ready, great thanks, but just checking as it's still marked WIP

MichaelChirico:
indeed! sorry about that. removed WIP

NEWS.md
mattdowle: I suggested `as.IDate(c('',somedate))` like that but I don't see that.  I see 'error' in the edit but not what the error was. And the news item is still long.  No worries - I'll have a go then.

inst/tests/tests.Rraw
mattdowle: in your revision you still used `as.IDate` in the y value.  I changed it to a truly fixed value.

mattdowle (approved):

R/IDateTime.R
```diff
@@ -7,6 +7,10 @@ as.IDate = function(x, ...) UseMethod("as.IDate")
 
 as.IDate.default = function(x, ..., tz = attr(x, "tzone", exact=TRUE)) {
   if (is.null(tz)) tz = "UTC"
+  if (is.character(x)) {
+    # 4676 mimics for back-compatibility a similar patch applied to as.Date.character in r79119
```
jangorecki, line 11: `# 4676` this will not be caught by a regex looking for issue number

mattdowle: Good spot. Done in df8009d. Since it was just a comment tweak I did that directly without a PR

PR #4707, opened 2020-09-16
Step 1 in fixing OpenMP detection in configure to work in cases where "-fopenmp" in required

fgp:
There seems to be an oversight in the current code.

It doesn't reset `R_NO_OPENMP` back to 0  before trying to compile `test-omp.c` with -fopenmp. As a result, even if compilation succeeds, `R_NO_OPENMP` is still 1 afterwards, and `configure` thus assumes OpenMP support is not available.


mattdowle:
Thanks. Related to https://github.com/Rdatatable/data.table/issues/4348#issuecomment-732431383.
This PR is a step in the right direction so merging, but note that the comment `# TODO: and then nothing seems to be done with this outcome` still stands.

PR #4712, opened 2020-09-20
Backport str2lang

MichaelChirico:
`str2lang` is a more efficient version of `parse(text=.)` but it's only available since R 3.6.0.

We are allowed to use an object from our namespace in `.onLoad` right? If not, we should at least use `keep.source=FALSE` there.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -34,15 +34,6 @@ which.first = function(x)
   match(TRUE, x)
 }
 
-# which.last
-which.last = function(x)
-{
-  if (!is.logical(x)) {
-    stop("x not boolean")
-  }
-  length(x) - match(TRUE, rev(x)) + 1L
-}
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -34,13 +34,10 @@ which.first = function(x)
   match(TRUE, x)
 }
 
-# which.last
-which.last = function(x)
```
MichaelChirico, line 38: this is not used internally (anymore at least)

MichaelChirico:
#4713 fixes the main internal usage of `parse(text=.)` to just use metaprogramming instead, h/t @jangorecki for the comment on #4711 supporting this.

The other three uses are pretty minor (two run once in `.onLoad`, one run in `test.data.table`), so just closing & moving on

PR #4714, opened 2020-09-22
Merge info

### LINKED ISSUE #4677: feature request: output from merge
It would be great to have some output on the merge operation like:

- Rows in x
- Rows in y
- Matched rows 
- Total rows after merge

This feature would make it easier to understand what happened.

A verbose option could do: merge(...,verbose = T) for example

ycphs:
Fixes #4677 

provides the requested information.

### DIFF
--- a/R/merge.R
+++ b/R/merge.R
@@ -1,9 +1,11 @@
 merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FALSE, all.x = all,
-               all.y = all, sort = TRUE, suffixes = c(".x", ".y"), no.dups = TRUE, allow.cartesian=getOption("datatable.allow.cartesian"), ...) {
+               all.y = all, sort = TRUE, suffixes = c(".x", ".y"), no.dups = TRUE, allow.cartesian=getOption("datatable.allow.cartesian"), info = FALSE, ...) {
   if (!sort %in% c(TRUE, FALSE))
     stop("Argument 'sort' should be logical TRUE/FALSE")
   if (!no.dups %in% c(TRUE, FALSE))
     stop("Argument 'no.dups' should be logical TRUE/FALSE")
+  if (!info %in% c(TRUE, FALSE))
+    stop("Argument 'info' should be logical TRUE/FALSE")
   class_x = class(x)
   if (!is.data.table(y)) {
     y = as.data.table(y)
@@ -98,5 +100,15 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
 
   # retain custom classes of first argument that resulted in dispatch to this method, #1378
   setattr(dt, "class", class_x)
-  dt
+  if(info) {
+    cat("Rows in x:" , x[, .N, verbose = F], "\n")
+    cat("Rows in y:" , y[, .N, verbose = F], "\n")
+    cat("Matched rows:" , y[x, nomatch = NULL, on = by, allow.cartesian =
+                              allow.cartesian, verbose = F][, .N, verbose = F], "\n")
+    cat("Total rows after merge:" , dt[, .N, verbose = F], "\n")
+    dt
+  }  else {
+    dt
+  }
+
 }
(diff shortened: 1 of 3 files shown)


R/merge.R
```diff
@@ -98,5 +100,15 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
 
   # retain custom classes of first argument that resulted in dispatch to this method, #1378
   setattr(dt, "class", class_x)
-  dt
+  if(info) {
+    message("Rows in x: " , x[, .N, verbose = F],"\n",
+   "Rows in y: " , y[, .N, verbose = F],"\n",
+    "Matched rows: " , y[x, nomatch = NULL, on = by, allow.cartesian =
+                              allow.cartesian, verbose = F][, .N, verbose = F],"\n",
+    "Total rows after merge: " , dt[, .N, verbose = F])
```
ColeMiller1, line 108: ```nrows(x)```, ```nrows(y)``` and ```nrows(dt)``` would be simpler than using ```.N```. However, on the join line, my limited testing indicates you could simply put ```j = .N``` instead of chaining another call.

ColeMiller1:
What is matched rows? If this is a cross join with a many-to-many relationship, I would expect matched rows to be equal to or less than the amount of rows in either table.

As an example, here's a simple join. I believe the output of the PR would display 8 rows. My feeling is that the answer should be 4 rows (e.g., all rows were matched). 

``` r
library(data.table)
x = data.table(a = rep(2:1, 2),
               c = 1:4)
y = data.table(a = rep(1:2, 2),
               b = 4:1)

x[y, on = 'a']
#>        a     c     b
#>    <int> <int> <int>
#> 1:     1     2     4
#> 2:     1     4     4
#> 3:     2     1     3
#> 4:     2     3     3
#> 5:     1     2     2
#> 6:     1     4     2
#> 7:     2     1     1
#> 8:     2     3     1
```

R/merge.R
```diff
@@ -1,9 +1,11 @@
 merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FALSE, all.x = all,
-               all.y = all, sort = TRUE, suffixes = c(".x", ".y"), no.dups = TRUE, allow.cartesian=getOption("datatable.allow.cartesian"), ...) {
+               all.y = all, sort = TRUE, suffixes = c(".x", ".y"), no.dups = TRUE, allow.cartesian=getOption("datatable.allow.cartesian"), info = FALSE, ...) {
```
jangorecki, line 2: `verbose` seems a perfect fit here rather than new `info` arg

R/merge.R
```diff
@@ -98,5 +100,15 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
 
   # retain custom classes of first argument that resulted in dispatch to this method, #1378
   setattr(dt, "class", class_x)
-  dt
+  if(info) {
+    message("Rows in x: " , x[, .N, verbose = F],"\n",
```
jangorecki, line 104: `message` goes to STDERR. Any reason why using that rather than `cat`?

R/merge.R
```diff
@@ -98,5 +100,15 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+   "Rows in y: " , y[, .N, verbose = F],"\n",
```
jangorecki, line 105: using `F` or `T` instead of full names will fail strict package checks

jangorecki (commented):
IMO the place where those information should be in in `[` or `bmerge`, then it will work when joining with `[` and `merge.data.table` as well. From the performance point of view it does't make sense to do join twice just to print those information, like it is made in this PR.

ycphs:
Thank you for your helpful review. I will try to move it to bmerge and change the PR accordingly.

The idea of having the info flag is would be a solution to have different levels of details for the verbose output.

MichaelChirico:
@ycphs would you like to resume this PR?

MichaelChirico:
Closing this, please feel free to re-start the work later.

PR #4718, opened 2020-09-25
new OS dep required by pkgdown

### LINKED ISSUE #4717: CI failure due to pkgdown installation failure
Currently GL CI fails when attempting to produce artifacts (pkgdown website). It is because one of pkgdown dependencies is failing to be installed.
Reproducible
```sh
docker run -it registry.gitlab.com/jangorecki/dockerfiles/r-builder Rscript -e 'install.packages("systemfonts")'
```
error
```
------------------------- [ERROR MESSAGE] ---------------------------
<stdin>:1:10: fatal error: fontconfig/fontconfig.h: No such file or directory
compilation terminated.
--------------------------------------------------------------------
ERROR: configuration failed for package ‘systemfonts’
* removing ‘/usr/local/lib/R/site-library/systemfonts’

The downloaded source packages are in
        ‘/tmp/RtmplIvXGb/downloaded_packages’
Warning message:
In install.packages("systemfonts") :
  installation of package ‘systemfonts’ had non-zero exit status
```
This is quite unfortunate because even we fix this one it is likely that similar issues will come back again in future due to the complexity of dependencies of pkgdown. What I think could make sense is to fix CRAN source to a specific date snapshot using MRAN, and keep using older pkgdown always.

jangorecki:
closes #4717

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -301,7 +301,7 @@ integration: ## merging all artifacts to produce single R repository, documentat
     R_DEVEL_BIN_VERSION: "4.1"
   script:
     ## pkgdown installs pkgs from "." so run at start to have clean root dir
-    - apt-get update -qq && apt-get install -y libxml2-dev
+    - apt-get update -qq && apt-get install -y libxml2-dev libfontconfig1-dev ## fontconfig1 for #4717
     - mkdir -p /tmp/pkgdown/library
     - R_LIBS_USER=/tmp/pkgdown/library Rscript -e 'install.packages("pkgdown", repos=Sys.getenv("CRAN_MIRROR"), quiet=TRUE); pkgdown::build_site(override=list(destination="./pkgdown"))'
     ## html manual, vignettes, repos, cran_web, cran_checks


.gitlab-ci.yml
```diff
@@ -301,7 +301,7 @@ integration: ## merging all artifacts to produce single R repository, documentat
     R_DEVEL_BIN_VERSION: "4.1"
   script:
     ## pkgdown installs pkgs from "." so run at start to have clean root dir
-    - apt-get update -qq && apt-get install -y libxml2-dev
```
MichaelChirico, line 304: could you link the dockerfile if it's public

eddelbuettel: To be used by the the yaml they have to be _in_ the yaml so are you looking for these two (and I am quoting):

- `image: registry.gitlab.com/jangorecki/dockerfiles/r-builder`
- `image: registry.gitlab.com/jangorecki/dockerfiles/r-base-dev`

A top-level `apt` call, however, would be at the level of the CI script / vm, not the container.

jangorecki: All images are built and published on GL, see jangorecki/dockerfiles, in menu you will find container registry there

PR #4719, opened 2020-09-26
Make tests 168,2043 work for any LC_TIME

### LINKED ISSUE #3450: Should fix the tests that depends on certain locale
For example, test 168.1, I have to set the locale to "C" in order to parse the `datetimes` correctly (otherwise they are all NAs) because my default locale is Chinese (Simplified)_People's Republic of China.936.

 Another example is the test 1966, it simply throws an error on my computer complains the file name is invalid.

https://github.com/Rdatatable/data.table/blob/676c9b28b6f7970542455ac0df69ce63a395662f/inst/tests/tests.Rraw#L468-L472

https://github.com/Rdatatable/data.table/blob/676c9b28b6f7970542455ac0df69ce63a395662f/inst/tests/tests.Rraw#L13012-L13014

tdhock:
Closes #3450
hi @jangorecki @shrektan @2005m @MichaelChirico this is a fix for #3450 which you commented on.
The problem in that issue is that some of the computed values depend on the LC_TIME locale value, whereas the expected values in the test are always English/C locale (Jan/NOV).
On my system (ubuntu 18.04) I have the following locales, the important one is `LC_TIME=fr_FR.UTF-8`
```
Fri Sep 25 16:41:56 2020  endian==little, sizeof(long double)==16, longdouble.digits==64, sizeof(pointer)==8, TZ==unset, Sys.timezone()=='America/Phoenix', Sys.getlocale()=='LC_CTYPE=en_US.UTF-8;LC_NUMERIC=C;LC_TIME=fr_FR.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONETARY=fr_FR.UTF-8;LC_MESSAGES=en_US.UTF-8;LC_PAPER=fr_FR.UTF-8;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=fr_FR.UTF-8;LC_IDENTIFICATION=C', l10n_info()=='MBCS=TRUE; UTF-8=TRUE; Latin-1=FALSE', getDTthreads()=='omp_get_num_procs()==2; R_DATATABLE_NUM_PROCS_PERCENT==unset (default 50); R_DATATABLE_NUM_THREADS==unset; R_DATATABLE_THROTTLE==unset (default 1024); omp_get_thread_limit()==2147483647; omp_get_max_threads()==2; OMP_THREAD_LIMIT==unset; OMP_NUM_THREADS==unset; RestoreAfterFork==true; data.table is using 1 threads with throttle==1024. See ?setDTthreads.'
```
with current master I get
```
(base) tdhock@maude-MacBookPro:~/R$ R CMD INSTALL data.table && R --vanilla < data.table/tests/main.R 
...
  7 errors out of 9996. Search tests/tests.Rraw.bz2 for test numbers: 168.1, 168.2, 168.3, 2042.2, 2042.3, 2127.26, 2127.27.
```
After applying the commit in this PR I get
```
All 10001 tests in tests/tests.Rraw completed ok in 00:01:50 elapsed (00:01:46 cpu)
```
I could have fixed this by just setting LC_TIME=C during the tests, but I figured it would be better to not alter the environment. So instead I just changed the expected values so that they are generated based on the current locale (e.g. "janv." and "nov." instead of "Jan" and "NOV"). Is this approach ok?

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -472,7 +472,7 @@ try(graphics.off(),silent=TRUE)
 
 # IDateTime conversion methods that ggplot2 uses (it calls as.data.frame method)
 # Since %b is e.g. "nov." in LC_TIME=fr_FR.UTF-8 locale, we need to
-# have the target/y value in these tests depend on the locale as well.
+# have the target/y value in these tests depend on the locale as well, #3450.
 NOV = format(strptime("2000-11-01", "%Y-%m-%d"), "%b")
 x = c("09:29:16","10:42:40","23:47:12","01:06:01","11:35:34","11:51:09")
 datetimes = paste0("2011 ", NOV, c(18,18,18,19,19,19), " ", x)
@@ -15070,7 +15070,7 @@ test(2042.1, DT[ , as.character(mean(date)), by=g, verbose=TRUE ],
              data.table(g=c("a","b"), V1=c("2018-01-04","2018-01-21")),
      output=msg<-"GForce is on, left j unchanged.*Old mean optimization is on, left j unchanged")
 # Since %b is e.g. "janv." in LC_TIME=fr_FR.UTF-8 locale, we need to
-# have the target/y value in these tests depend on the locale as well.
+# have the target/y value in these tests depend on the locale as well, #3450.
 Jan.2018 = format(strptime("2018-01-01", "%Y-%m-%d"), "%b-%Y")
 test(2042.2, DT[ , format(mean(date),"%b-%Y")], Jan.2018)
 test(2042.3, DT[ , format(mean(date),"%b-%Y"), by=g, verbose=TRUE ],  # just this case generated the error
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -471,10 +471,13 @@ test(167.3, DT[,plot(b,f),by=.(grp)], data.table(grp=integer()))
 try(graphics.off(),silent=TRUE)
 
 # IDateTime conversion methods that ggplot2 uses (it calls as.data.frame method)
-datetimes = c("2011 NOV18 09:29:16", "2011 NOV18 10:42:40", "2011 NOV18 23:47:12",
-              "2011 NOV19 01:06:01", "2011 NOV19 11:35:34", "2011 NOV19 11:51:09")
+# Since %b is e.g. "nov." in LC_TIME=fr_FR.UTF-8 locale, we need to
+# have the target/y value in these tests depend on the locale as well.
+NOV = format(strptime("2000-11-01", "%Y-%m-%d"), "%b")
+x = c("09:29:16","10:42:40","23:47:12","01:06:01","11:35:34","11:51:09")
+datetimes = paste0("2011 ", NOV, c(18,18,18,19,19,19), " ", x)
 DT = IDateTime(strptime(datetimes,"%Y %b%d %H:%M:%S"))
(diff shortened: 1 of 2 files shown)


MichaelChirico:
LGTM thanks!

tdhock:
great thanks! @mattdowle would you consider merging please?

MichaelChirico:
hi Toby, any particular reason to prioritize?

As you can see we've got ourselves a bit of  a PR backlog at the moment

tdhock:
haha no rush

inst/tests/tests.Rraw
```diff
@@ -471,10 +471,13 @@ test(167.3, DT[,plot(b,f),by=.(grp)], data.table(grp=integer()))
 try(graphics.off(),silent=TRUE)
 
 # IDateTime conversion methods that ggplot2 uses (it calls as.data.frame method)
-datetimes = c("2011 NOV18 09:29:16", "2011 NOV18 10:42:40", "2011 NOV18 23:47:12",
-              "2011 NOV19 01:06:01", "2011 NOV19 11:35:34", "2011 NOV19 11:51:09")
+# Since %b is e.g. "nov." in LC_TIME=fr_FR.UTF-8 locale, we need to
+# have the target/y value in these tests depend on the locale as well.
```
mattdowle, line 475: Great comment. Just needs ", #3450." on the end too so we can find the issue in future. I've just added a sentence to contributing guide point 1 stating that now.

inst/tests/tests.Rraw
```diff
@@ -15065,10 +15068,13 @@ test(2041.2, DT[, median(time), by=g], DT[c(2,5),.(g=g, V1=time)])
 # 'invalid trim argument' with optimization level 1; #1876
 test(2042.1, DT[ , as.character(mean(date)), by=g, verbose=TRUE ],
              data.table(g=c("a","b"), V1=c("2018-01-04","2018-01-21")),
-             output=msg<-"GForce is on, left j unchanged.*Old mean optimization is on, left j unchanged")
-test(2042.2, DT[ , format(mean(date),"%b-%Y")], "Jan-2018")
+     output=msg<-"GForce is on, left j unchanged.*Old mean optimization is on, left j unchanged")
+# Since %b is e.g. "janv." in LC_TIME=fr_FR.UTF-8 locale, we need to
+# have the target/y value in these tests depend on the locale as well.
```
mattdowle, line 15073: and ", #3450." here too please.

mattdowle (changes requested):
Great. Had a look. It just needs a short news item (notes section), and the ", #3450" adding on these lines please.

mattdowle:
As it happens, since CRAN is in error status and this release is pressing, the timing is good at the moment for any quick and easy PRs (like this one) to be prioritized to get them out of the way in 1.13.2.  Most of the PRs currently scheduled for next release (milestone 1.13.1) will probably get bumped to 1.13.3 milestone.

tdhock:
hi @mattdowle thanks for the quick review. here are the issue/PR mentions, is that OK?

mattdowle:
Thanks @tdhock.  News items have developed a silent standard we should probably write down.  We just reference the issue and not the PR too, because the issue itself links to the PR. That keeps the news item briefer, but also 1 link rather than 2 helps the CRAN check times because CRAN checks that every link in NEWS.md works and that can take over 20 mins. Also the name of the user-visible function affected should be present, in this case `test.data.table()`, and up front, to aid users quickly scanning the news items.  It's ok I can tweak the news item, just explaining what's in my mind when I tweak them.  We should write this stuff down.

mattdowle:
@tdhock is the revised news item ok?  I'm not sure about the "`Jan` vs `janv.`" part.  Is it J vs j like that, and the French has a dot too?

mattdowle (approved):

tdhock:
great thanks for the info

PR #4720, opened 2020-09-26
support missing values in measure.vars arg to melt

### LINKED ISSUE #4027: Ability to explicitly pass index names in melt.data.table measure.vars argument
Currently melt with multiple melting patterns works as expected only when data has a regular grid pattern. If some of the variables are missing the order of the output is scrambled.

For example:

```R
(dt <- data.table(id = 1, a.1 = 1, a.3 = 3, b.1 = 1, b.2 = 2, b.3 = 3))
##    id a.1 a.3 b.1 b.2 b.3
## 1:  1   1   3   1   2   3

melt.data.table(dt, "id" , measure.vars = list(a = 2:3, b = 4:6), variable.name = "ix")
##    id ix  a b
## 1:  1  1  1 1
## 2:  1  2  3 2
## 3:  1  3 NA 3
```

But what I really want is the ix of `a.3` to be 3,  not 2.

```R
##    id ix  a b
## 1:  1  1  1 1
## 2:  1  2 NA 2
## 3:  1  3  3 3
```

I was hoping that explicit index names in the measure.vars would be taken into account but they are currently not:

```R
melt.data.table(dt, "id" ,
                measure.vars = list(a = c("1" = 2, "3" = 3),
                                    b = c("4" = 4, "5" = 5, "6" = 6)),
                variable.name = "ix")
##    id ix  a b
## 1:  1  1  1 1
## 2:  1  2  3 2
## 3:  1  3 NA 3
```

I am fairly confident that this is related (or maybe even the same as #1739) but the example and discussion there is a bit cryptic.

Would you consider adding the ability to explicitly pass index names?

Thanks!

<details>

```
R Under development (unstable) (2019-08-06 r76924)
Running under: Ubuntu 18.04.3 LTS

Matrix products: default
BLAS:   /home/vspinu/bin/R-latest/lib/libRblas.so
LAPACK: /home/vspinu/bin/R-latest/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=de_DE.UTF-8    LC_MESSAGES=C              LC_PAPER=de_DE.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            

attached base packages:
[1] stats     graphics  utils     datasets  grDevices methods   base     

other attached packages:
[1] data.table_1.12.4

[1] compiler_3.7.0
```

</details>

tdhock:
Closes #4027
hi @vspinu @Henrik-P @jangorecki @mrdwab tagging you because you commented on the related issue #4027 which I hope this PR will close. In current master I get these errors because NA is unsupported in the measure.vars argument to melt:
```
> DT.wide = data.table(a2=2, b1=1, b2=2)
> melt(DT.wide, measure.vars=list(a=c(NA,1), b=2:3))
Error in melt.data.table(DT.wide, measure.vars = list(a = c(NA, 1), b = 2:3)) : 
  One or more values in 'measure.vars' is invalid.
> melt(DT.wide, measure.vars=list(a=c(NA,"a2"), b=c("b1","b2")))
Error in melt.data.table(DT.wide, measure.vars = list(a = c(NA, "a2"),  : 
  One or more values in 'measure.vars' is invalid.
```

I added a couple of simple tests which explain the new behavior implemented in this PR:
```
> DT.wide = data.table(a2=2, b1=1, b2=2)
> melt(DT.wide, measure.vars=list(a=c(NA,1), b=2:3))
   variable  a b
1:        1 NA 1
2:        2  2 2
> melt(DT.wide, measure.vars=list(a=c(NA,"a2"), b=c("b1","b2")))
   variable  a b
1:        1 NA 1
2:        2  2 2
```
It took me a while to read through and understand the fmelt code, and I added some comments which explain the meaning of the different variables/functions. If those comments are not welcome I can delete them.
All tests pass on my computer (Ubuntu 18.04, R-4.0.2).
@MichaelChirico can you code review and maybe suggest additional tests?


tdhock:
We should add a test for id.vars NOT null, since there is separate handling of that case in checkVars. (I only implemented the new feature for the case of id.vars null)
A simple fix would be to just copy the same logic I used over to the NOT null case.
But before doing that I wanted to check in with you guys and see if you would be open to a different approach: refactoring which reduces duplication between the two cases? It seems like there is a lot of duplicated code/logic in checkVars currently, and it could be simplified.

tdhock:
by the way did you notice the funny coincidence that the original issue number #4027 anagrams to this PR number #4720, how often does that happen!?

src/fmelt.c
```diff
@@ -265,7 +268,7 @@ SEXP checkVars(SEXP DT, SEXP id, SEXP measure, Rboolean verbose) {
       tmp = PROTECT(unlist_(tmp2)); protecti++;
     }
     for (i=0; i<length(tmp); i++) {
-      if (INTEGER(tmp)[i] <= 0 || INTEGER(tmp)[i] > ncol)
+      if (invalid_measure(INTEGER(tmp)[i], ncol))
```
tdhock, line 271: here is a fix for the case of id.vars NOT null, which just repeats the same logic as used for the case for id.vars IS null.

NEWS.md
```diff
@@ -6,6 +6,14 @@
 
 ## NEW FEATURES
 
+1. `melt.data.table()` now supports `NA` entries when specifying a
+   list of `measure.vars`, which translate into runs of missing values
```
jangorecki, line 10: We don't need new lines here

tdhock: ok fixed

vspinu:
@tdhock I don't see how this PR fixes #4027. Could you please use it with the original example?

tdhock:
```r
remotes::install_github("Rdatatable/data.table@fix4027")
#> Skipping install of 'data.table' from a github remote, the SHA1 (4c5810c2) has not changed since last install.
#>   Use `force = TRUE` to force installation
library(data.table)
dt <- data.table(id = 1, a.1 = 1, a.3 = 3, b.1 = 1, b.2 = 2, b.3 = 3)
melt(
  dt, variable.name="ix",
  measure.vars=list(a=c("a.1", NA, "a.3"), b=c("b.1", "b.2", "b.3")))
#>    id ix  a b
#> 1:  1  1  1 1
#> 2:  1  2 NA 2
#> 3:  1  3  3 3
```

tdhock:
should the new tests be moved to the "melt section" starting at test number 1035?

src/fmelt.c
```diff
@@ -105,7 +105,7 @@ SEXP measurelist(SEXP measure, SEXP dtnames) {
     SEXP x = VECTOR_ELT(measure, i);
     switch(TYPEOF(x)) {
       case STRSXP  :
-        SET_VECTOR_ELT(ans, i, chmatch(x, dtnames, 0));
+        SET_VECTOR_ELT(ans, i, chmatch(x, dtnames, NA_INTEGER));
```
mattdowle, line 108: The new `chmatch_na` function in this PR that I just removed I initially thought was necessary because this PR needed `NA`s in `x` not to match to `NA`s in `table`. So I started to add `bool matchNAtoNA` argument to `chmatchMain`. Then it turned out that just passing `nomatch=NA_INTEGER` on this line (the only line where `chmatch_na` was used) instead of `nomatch=0`, seemed to work and passed the new tests. I left `matchNAtoNA` as a comment in `chmatchMain` in case we need to come back to that in future.
@tdhock Please check and let me know if I got anything wrong here. If so, a new test is probably needed please.

PR #4730, opened 2020-10-01
add na.rm argument to first/last and gfirst/glast

### LINKED ISSUE #4239: Add a na.rm argument to first()/last()
Personally, I'd find it useful if `gfirst()`/`glast()` would accept a `na.rm` argument. This would then return the first/last non missing element of a vector (within each group).

For consistency I guess also `first()`/`last()` should then accept a `na.rm` argument. Although what this would mean if a non-atomic object (i.e. a `data.frame`) was passed, I don't know. Maybe the flag should only be honored if `is.atomic(x) && nargs() == 1L`?

### LINKED ISSUE #4446: FR: na.rm in first/last
```
DT  = data.table(a = c(1, 2, 3, NA))
last(DT$a)
# [1] NA
```

vs. 

```
last(DT$a, na.rm = TRUE)
# [1] 3
```

ben-schwen:
Closes #4446
Closes #4239 

Adds an `na.rm` argument with default `na.rm=FALSE` to `first()`/`last()` and `gfirst()`/`glast()`. 

Here should be noted that `gfirst()` and `glast()` return the first resp. last non NA value if such a value exists. If such a value does not exist `gfirst()`/`glast()` still return NA similar to `gmedian()`, `gvar()` and `gsd()`.


R/last.R
```diff
@@ -1,8 +1,22 @@
...
   verbose = isTRUE(getOption("datatable.verbose", FALSE))
+
+  stopifnot(isTRUEorFALSE(na.rm))
+
+  if (na.rm) {
+    if (verbose) 
+      cat("last: na.rm=TRUE\n")
+    
+    if (is.vector(x) && length(x)) {
+      x = x[complete.cases(x)]
```
MichaelChirico, line 14: why not `x[!is.na(x)]`

R/last.R
```diff
@@ -1,8 +1,22 @@
...
 # reworked to avoid loading xts namespace (#3857) then again to fix dispatching of xts class (#4053)
-last = function(x, n=1L, ...) {
+last = function(x, n=1L, na.rm=FALSE, ...) {
   verbose = isTRUE(getOption("datatable.verbose", FALSE))
+
+  stopifnot(isTRUEorFALSE(na.rm))
+
+  if (na.rm) {
+    if (verbose) 
+      cat("last: na.rm=TRUE\n")
```
MichaelChirico, line 11: please do follow the codecov bot & add tests for the uncovered lines

src/gsumm.c
```diff
@@ -1013,6 +1017,16 @@ SEXP glast(SEXP x) {
       if (isunsorted) k = oo[k]-1;
       k = (irowslen == -1) ? k : irows[k]-1;
       ians[i] = ix[k];
+      if (narm){
+        j = 0;
+        while(j < grpsize[i]){
```
MichaelChirico, line 1022: what happens when the group is all-`NA`?

src/gsumm.c
```diff
@@ -1037,6 +1061,16 @@ SEXP glast(SEXP x) {
       if (isunsorted) k = oo[k]-1;
       k = (irowslen == -1) ? k : irows[k]-1;
       dans[i] = dx[k];
+      if (narm){
+        j = 0;
+        while(j < grpsize[i]){
+          if (dx[k-j] != NA_REAL){
```
MichaelChirico, line 1067: this is usually written with `!ISNAN(.)` instead of comparison to `NA_REAL`

src/gsumm.c
```diff
@@ -1049,6 +1083,17 @@ SEXP glast(SEXP x) {
       if (isunsorted) k = oo[k]-1;
       k = (irowslen == -1) ? k : irows[k]-1;
       dans[i] = dx[k];
+      if (narm){
+        j = 0;
+        while(j < grpsize[i]){
+          const Rcomplex elem = dx[k-j];
+          if (!(ISNAN(elem.i) && ISNAN(elem.r))){
```
MichaelChirico, line 1090: i don't think we need to do an allocation each iteration, just `ISNAN(dk[k-j].r)`. Actually in this case, I would also use the `||` condition, since it's typically the case that both real & imaginary parts are simultaneously `NA`. So using `||` would short circuit and `dk[k-j].i` wouldn't even need to be run.

MichaelChirico: alternatively, you could flip the condition:

```
if (ISNAN(.) || ISNAN(.)) { j++; next; }
dans[i] = dx[k-j];
break;
```

src/gsumm.c
ben-schwen: Good question. Also raised this question in #4239 

> Whats happens in other `gfunctions` for those groups?
> 
> * `gsum` returns 0
> * `gmedian`, `gvar` and `gsd` return NA
> * `gmean` returns NaN
> * `gprod` returns 1
> * `gmax`/`gmin` return Inf

Personally I find `NA` intuitive since I would not want groups to disappear by grouping functions.

src/gsumm.c
ben-schwen: Have to agree flipping the conditional makes it look cleaner

src/gsumm.c
ben-schwen: changed

R/last.R
ben-schwen: Already used `complete.cases` for the `data.frame` case. However, seeing `complete.cases` failing for non standard types likes `nanotime` `!is.na` seems more reasonable.

src/gsumm.c
```diff
@@ -1020,16 +1020,13 @@ SEXP glast(SEXP x, SEXP narmArg) {
       if (narm){
         j = 0;
         while(j < grpsize[i]){
-          if (ix[k-j] != NA_LOGICAL){
-            ians[i] = ix[k-j];
-            break;
-          }
-          j++;
+          if (ix[k-j] == NA_LOGICAL){j++; continue;}
```
jangorecki, line 1023: I would expand this to few lines for better code coverage

src/gsumm.c
```diff
@@ -1049,26 +1074,50 @@ SEXP glast(SEXP x) {
...
     ans = PROTECT(allocVector(VECSXP, ngrp));
     for (i=0; i<ngrp; i++) {
       k = ff[i]+grpsize[i]-2;
       if (isunsorted) k = oo[k]-1;
       k = (irowslen == -1) ? k : irows[k]-1;
       SET_VECTOR_ELT(ans, i, VECTOR_ELT(x, k));
+      if (narm){
+        j = 0;
+        while(j < grpsize[i]){
+          if (VECTOR_ELT(x, k-j) == R_NilValue){j++; continue;}
```
jangorecki, line 1114: AFAIR we prefer to use `isNull()` rather than `== R_NilValue`

src/gsumm.c
```diff
@@ -1037,9 +1061,19 @@ SEXP glast(SEXP x) {
       if (isunsorted) k = oo[k]-1;
       k = (irowslen == -1) ? k : irows[k]-1;
       dans[i] = dx[k];
+      if (narm){
```
jangorecki, line 1064: A minor thing is to align to the coding style convention of a file you are editing. As you can see around, add extra space between in cases like:
- `if (narm){` -> `if (narm) {`
- `while(...){` -> `while (...) {`

jangorecki (commented):
last comment about styling, other than that it looks good

ben-schwen:
Closing here since incompatible with merged #5109

PR #4737, opened 2020-10-03
fix for melt with na.rm=TRUE and list column value

tdhock:
hi again, I have been trying to get 100% coverage in fmelt.c and some of the uncovered lines revealed a bug. This PR adds a test for this case, and will eventually add a fix. The problem is shown below:
```r
library(data.table)
melt(data.table(a1=1, b1=1, b2=2), na.rm=TRUE, measure.vars=list(a="a1", b=c("b1","b2")))
#>    variable a b
#> 1:        1 1 1
# only one row as expected (since a2 is missing, b2 is removed as well).
melt(data.table(a1=1, b1=list(1:2), b2=list(c('foo','bar'))), na.rm=TRUE, measure.vars=list(a="a1", b=c("b1","b2")))
#>    variable  a       b
#> 1:        1  1     1,2
#> 2:        2 NA foo,bar
# two rows, second a value missing, incorrect since na.rm=TRUE was specified.
```


tdhock:
There was some code which treated lists as a special case, and always used data->narm=FALSE in the C code, even when the user specified na.rm=TRUE in the R code. Seems to me that this is undesirable, due to the inconsistency with the non-list case. If that behavior is desirable to keep, then I would have at least expected some kind of message telling the user that their request of na.rm=TRUE was ignored (seems like there was a message but only if verbose=TRUE).

tdhock:
This behavior is mentioned on the ?melt.data.table man page:
```r
# return 'NA' for missing columns, 'na.rm=TRUE' ignored due to list column
melt(DT, id=1:2, measure=patterns("l_", "c_"), na.rm=TRUE)
```

mattdowle:
I guess the news item in #4720 covers this one too.

PR #4738, opened 2020-10-03
WIP: vector support for fsetdiff/funion/fintersect and fsetequal

### LINKED ISSUE #3752: add fsetdiff(), fintersect(), funion() supports for vectors
Hi,

data.table has already implemented fantastic `fsetdiff(), fintersect(), funion()` for data.table objects. I would like to proposal their supports on vectors as well. The reason is the equivalent functions in base R will stripe off all the attributes and is bad for Date (or any other inputs with attributes).

``` r
vec_date1 <- as.Date(c('1999-01-01', '1999-01-02'))
vec_date2 <- as.Date(c('1999-01-02', '1999-01-03'))
intersect(vec_date1, vec_date2)
#> [1] 10593
setdiff(vec_date1, vec_date2)
#> [1] 10592
union(vec_date1, vec_date2)
#> [1] 10592 10593 10594
```

<sup>Created on 2019-08-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

ben-schwen:
Closes #3752

Merely a wrapper for vectors to use `fintersect`, `funion`, `fsetdiff` and `fsetequal` also with vectors.

Whats interesting are the benchmarks.

```
set.seed(13)
x = round(rnorm(1e7), 4)
y = round(rnorm(1e7), 4)

	base = intersect(x,y),
	dplyr = dplyr::intersect(x,y),
	dt = fintersect(x,y),
	times=10L)

Unit: milliseconds
  base 1014.8031 1033.9171 1056.6386 1067.0683 1073.6028 1089.5941    10
 dplyr 1027.2003 1038.3417 1084.2351 1082.1406 1114.7926 1198.5054    10
    dt  732.0577  745.6789  755.5956  753.7491  765.2026  782.2758    10
```
Some speedup for reals due to threads (slower for single thread).

```
a = as.character(x)
b = as.character(y)
	base = intersect(a,b),
	dplyr = dplyr::intersect(a,b),
	dt = fintersect(a,b),
	times=10L)

Unit: milliseconds
  expr      min        lq      mean    median        uq       max neval
  base 1061.917 1113.0419 1168.0659 1148.5322 1225.2934 1327.6311    10
 dplyr 1058.967 1149.0933 1152.5475 1158.7641 1184.9405 1212.4005    10
    dt  438.239  453.3726  536.0707  541.5542  592.0239  622.7527    10
```
Major speedup for characters (also faster for single thread).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 ## NEW FEATURES
 
+1. Set operations `fintersect`, `funion`, `fsetdiff` and `fseteuqal` now also support vectors as input. #3752(https://github.com/Rdatatable/data.table/issues/3752). Thanks to @shrektan for the request and thanks to Benjamin Schwendinger for the PR. 
+
 ## BUG FIXES
 
 1. `test.data.table()` could fail the 2nd time it is run by a user in the same R session on Windows due to not resetting locale properly after testing Chinese translation, [#4630](https://github.com/Rdatatable/data.table/pull/4630). Thanks to Cole Miller for investigating and fixing.
--- a/R/setops.R
+++ b/R/setops.R
@@ -56,7 +56,15 @@ funique = function(x) {
   if (.seqn && ".seqn" %chin% names(x)) stop("None of the datasets should contain a column named '.seqn'")
 }
 
+vec2dt = function(fun, x, y, all=FALSE){
+  fun(setDT(list(x)), setDT(list(y)), all)[[1]]
+}
+
 fintersect = function(x, y, all=FALSE) {
+  if (is.vector(x) && is.vector(y)){
+    return(vec2dt(fintersect, x, y, all))
+  }
+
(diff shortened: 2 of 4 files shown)


jangorecki:
This kind of wrappers are good as user defined functions or a third party package functions. We should ideally try to use our internals instead.

ben-schwen:
> This kind of wrappers are good as user defined functions or a third party package functions. We should ideally try to use our internals instead.

@jangorecki I understand. For the example of `fintersect`: I guess that using `bmerge` together with `setDT` is also less favorable than writing a 2nd version of `bmerge` for vectors?

jangorecki:
Another version of bmerge does not make sense to support just vectors because logic is quite complex. I would love to rework interface to bmerge so it is easier to call it from C, then such use cases could be much simpler.

PR #4739, opened 2020-10-05
Update URLs to avoid notes in R CMD check on R-devel

jimhester:
As you may be aware recently CRAN has become more strict with 301 redirects in URLs, insisting that they all are changed to their final location. These additional checks are currently only available in R-devel.

In addition I believe because data.table has so many URLs CRAN has been disabling the checks for data.table in particular, so some of these may have been missed.

I recently extracted the code from R-devel, backported it to R 3.3+ and put it in a package, https://github.com/jimhester/urlchecker. Additionally I added support for concurrent requests, which makes the check take much less time for packages with lots of links (like data.table). The package also has a function (`url_update()`) to automatically update redirected links.

I used data.table as a stress test to verify it was working as intended, and a side benefit was fixing the links. So I thought you might appreciate the changes.

I see some of the redirects were from `http://r-datatable.com` to `https://rdatatable.gitlab.io/data.table`, which may be undesirable. If you could change that to a `302` redirect instead of a `301` that would avoid the check note.


README.md
```diff
@@ -72,7 +72,7 @@ DT[Petal.Width > 1.0, mean(Petal.Length), by = Species]
 
 ### Getting started
 
-* [Introduction to data.table](https://cloud.r-project.org/web/packages/data.table/vignettes/datatable-intro.html) vignette
+* [Introduction to data.table](https://cran.r-project.org/package=data.table/vignettes/datatable-intro.html) vignette
```
jimhester, line 75: This was the only flagged URL that was not a redirect issue. I updated it to use the 'canonical CRAN url'

DESCRIPTION
```diff
@@ -67,7 +67,7 @@ Suggests: bit64, curl, R.utils, xts, nanotime, zoo, yaml, knitr, rmarkdown
 SystemRequirements: zlib
 Description: Fast aggregation of large data (e.g. 100GB in RAM), fast ordered joins, fast add/modify/delete of columns by group using no copies at all, list columns, friendly and fast character-separated-value read/write. Offers a natural and flexible syntax, for faster development.
 License: MPL-2.0 | file LICENSE
-URL: http://r-datatable.com, https://Rdatatable.gitlab.io/data.table, https://github.com/Rdatatable/data.table
+URL: https://rdatatable.gitlab.io/data.table, https://Rdatatable.gitlab.io/data.table, https://github.com/Rdatatable/data.table
```
MichaelChirico, line 70: better to just remove this link? presumably no need to include both `https://rdatatable.gitlab.io/data.table` and `https://Rdatatable.gitlab.io/data.table`

MichaelChirico:
LGTM thanks for this!

mattdowle:
Yes indeed, thanks @jimhester. Sorry I didn't see this PR at the time otherwise I would have thanked you in the news item too. In the end it was note 2 in v1.14.0 (Feb 2021) which resolved this and contained more background info in case you didn't see it and find it interesting. 

image

PR #4743, opened 2020-10-08
Fix building from source on Mac OSX

### LINKED ISSUE #4742: After #4735, I can't build data.table on Mac from source
## The configure file was modified in PR #4735 

However, the two following lines can't be executed properly on OSX :

https://github.com/Rdatatable/data.table/blob/ba2d7bb170bb0aa120192b0daa65605bc4b038c8/configure#L107-L108

## It will cause the failure of installing data.table from source on Mac OSX, with two errors:

截屏2020-10-08 下午9 27 04

## This is because the `sed` used on OSX is not GNU `sed`

and `sed -i` must be followed by an extension argument on Mac

image

shrektan:
Fixes #4742 as the sed used on OSX is not the GNU version. In particular, the `sed -i` command is different.

By correcting the two lines, now I can build data.table from source on my Mac.

### DIFF
--- a/configure
+++ b/configure
@@ -104,7 +104,7 @@ else
   sed -e "s|@openmp_cflags@|\$(SHLIB_OPENMP_CFLAGS)|" src/Makevars.in > src/Makevars
 fi
 # retain user supplied PKG_ env variables, #4664. See comments in Makevars.in too.
-sed -i "s|@PKG_CFLAGS@|$PKG_CFLAGS|" src/Makevars
-sed -i "s|@PKG_LIBS@|$PKG_LIBS|" src/Makevars
+sed -i "" "s|@PKG_CFLAGS@|$PKG_CFLAGS|" src/Makevars
+sed -i "" "s|@PKG_LIBS@|$PKG_LIBS|" src/Makevars
 
 exit 0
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -104,7 +104,7 @@ else
   sed -e "s|@openmp_cflags@|\$(SHLIB_OPENMP_CFLAGS)|" src/Makevars.in > src/Makevars
 fi
 # retain user supplied PKG_ env variables, #4664. See comments in Makevars.in too.
-sed -i "s|@PKG_CFLAGS@|$PKG_CFLAGS|" src/Makevars
-sed -i "s|@PKG_LIBS@|$PKG_LIBS|" src/Makevars
+sed -i "" "s|@PKG_CFLAGS@|$PKG_CFLAGS|" src/Makevars
```
MichaelChirico, line 107: I thought I remembered `sed -i` requires a non-default `sed` on Mac.

To be extensible I think we need to `sed` to a temp file, then `mv` it afterwards & purge the temp

shrektan: Yes, I agree. That may be the reason of the CI fails...

jangorecki:
Looks good, thanks. I renamed filename to keep it as Makevars and use extension to describe the state of file.

mattdowle (approved):
Thanks for testing before release. Great.

PR #4751, opened 2020-10-11
fread() throws correct non-ASCII messages

### LINKED ISSUE #4747: The msg from fread should mark txt as declared encoding
**On Windows**, when the text is UTF-8 encoded and the printed message from `fread()` contains some text, the message will be displayed as garbage letters. The cause I believe is we didn't mark the txt as the declared encoding "UTF-8".

### Code

```r
txt <- "A,B\n中文1,中文2\n中文3"
txt <- enc2utf8(txt)
data.table::fread(text = txt, encoding = 'UTF-8')
```

### Output

```
       A     B
1: 中文1 中文2
Warning message:
In data.table::fread(text = txt, encoding = "UTF-8") :
  Discarded single-line footer: <<涓枃3>>
```

### Code

```r
txt <- "A,B\n中文1,中文2\n中文3"
data.table::fread(text = txt)
```

### Output

```
       A     B
1: 中文1 中文2
Warning message:
In data.table::fread(text = txt) : Discarded single-line footer: <<中文3>>
```

## session Info

```
> sessionInfo()
R version 4.0.2 (2020-06-22)

Matrix products: default

locale:
[1] LC_COLLATE=Chinese (Simplified)_China.936  LC_CTYPE=Chinese (Simplified)_China.936   
[3] LC_MONETARY=Chinese (Simplified)_China.936 LC_NUMERIC=C                              
[5] LC_TIME=Chinese (Simplified)_China.936    

attached base packages:

other attached packages:
[1] magrittr_1.5      data.table_1.13.0
```

---

## Another example on Mac

```r
x <- "fa\xE7ile"
Encoding(x) <- "latin1"
txt <- sprintf("A,B\n%s,%s\n%s", x, x, x)
Encoding(txt) <- "UTF-8"

data.table::fread(text = txt, encoding = 'UTF-8')

txt2 <- iconv(txt, "UTF-8", "latin1")
data.table::fread(text = txt2, encoding = 'Latin-1')
```

shrektan:
Fixes #4747 

Before the PR the last line throw warnings with garbage letters. This PR fixes it.

### Example code

```r
x <- "fa\xE7ile"
Encoding(x) <- "latin1"
txt <- sprintf("A,B\n%s,%s\n%s", x, x, x)
Encoding(txt) <- "UTF-8"
data.table::fread(text = txt, encoding = 'UTF-8')
txt2 <- iconv(txt, "UTF-8", "latin1")
data.table::fread(text = txt2, encoding = 'Latin-1')
```

### Before 

```
Warning message:
In data.table::fread(text = txt2, encoding = "Latin-1") : Discarded single-line footer: <<fa�ile>>
```

### After

```
Warning message:
In data.table::fread(text = txt, encoding = "UTF-8") :
  Discarded single-line footer: <<façile>>
```


src/fread.h
```diff
@@ -7,8 +7,13 @@
 #include "myomp.h"
 #ifdef DTPY
   #include "py_fread.h"
+  #define ENC2NATIVE(s) (s)
```
MichaelChirico, line 10: cc @st-pasha , not sure if there's something to do here on the Python side

src/freadR.h
```diff
@@ -2,6 +2,7 @@
 #define dt_FREAD_R_H
 #define STRICT_R_HEADERS   // https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Error-handling
 #include <R.h>
+#include <Rinternals.h>
```
jangorecki, line 5: I understand we need this for `translateChar(mkCharCE(s, ienc))`, right?
It may be good to ask R core for public API for this functionality.

mattdowle: Maybe but I have the impression from WRE that Rinternals.h is already a public API, at least partially. For example, the string "#include <Rinternals.h>" is exhibited 13 times in WRE. We need it even for just SEXP (which is stated too in WRE). I did find one use of the word 'public' in the context of Rinternals.h too:
image
Note 144 in full is "see [The R API](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#The-R-API): note that these are not all part of the API."
I remember seeing some debate on R-devel mailing list on this over the years. Some R-core members believe that the R API is anything mentioned in WRE. But that would exclude many of the exported (i.e. non-hidden) functions that many packages use. I believe that `#define USE_RINTERNALS` before "#include <Rinternals.h>" (as data.table used to do in the past) does access truly internal non-API functions/structures. But including Rinternals.h is not necessarily accessing non-public API, in fact it's necessary to include it to access all the public API.
However, with that said, `translateChar` and `mkCharCE` are both highlighted by name in WRE (section 5.15) and for that reason there would be no disagreement by any R-core member that those are already public API.

MichaelChirico: that seems... needlessly confusing!

mattdowle: I agree. All those that have thought so have engaged. Then what you discover is that R-core members disagree between themselves. Tangent issues (such as data.table using SET_TRUELENGTH) are quickly raised. I assume what you would like is for R.h to be public and Rinternals.h to be not-public? Underlying that seemingly simple task is i) change to R over time, and ii) in our case of data.table, achievements by using SET_TRUELENGTH that some core members wish was never accessible.
Careful what you wish for.

jangorecki: Yes, sorry for fuss. When I read Rinternals I had in mind USE_INTERNALS switch. So definitely confusing.

PR #4753, opened 2020-10-12
C data.table docs and minor rename

jangorecki:
Improved docs for recently provided header file for exported functions.

Note that there is a rename of exported C function.


man/cdt.Rd
```diff
@@ -2,18 +2,25 @@
...
 \details{
-  For details how to use those see \emph{Writing R Extensions} manual \emph{Linking to native routines in other packages} section.
+  Details how to use those can be found in \emph{Writing R Extensions} manual \emph{Linking to native routines in other packages} section.
+  An example use with \code{Rcpp}:
+\preformatted{
+  dt = data.table::as.data.table(iris)
+  Rcpp::cppFunction("SEXP mysub2(SEXP x, SEXP rows, SEXP cols) { return DT_subsetDT(x,rows,cols); }",
+    include="#include <datatableAPI.h>",
+    depends="data.table")
+  mysub2(dt, 1:4, 1:4)
```
jangorecki, line 19: This Dirk example is really nice and shows how new header can be nicely used. I would like to provide pure C example as well but cannot get it to work.
```
dt = data.table::as.data.table(iris)
mysub1 = inline::cfunction(sig=c(x="SEXP", rows="integer", cols="integer"),
  "return DT_subsetDT(x,rows,cols); ",
  includes = "#include <datatableAPI.h>",
  cppargs = paste0("-I",system.file("include", package="data.table")),
  language = "C")
```
@eddelbuettel any idea how to effectively use new header file from plain C? if possible just base R, not even `inline`, for simplicity

eddelbuettel: @jangorecki Can you please rephrase that question?  What is your are asking?  Header files are using during compilation only (and we can all blame R's `LinkingTo:` for confusion).  Are you aiming for an example in R using a header file but no compiler?  I can't quite square it ...

eddelbuettel (approved):
Looks good to me (modulo the one inline comment). I'd have to run a test compilation to see "for sure" if all locations of the old identifier have been renamed but I trust @jangorecki on this.

man/cdt.Rd
eddelbuettel: Are you asking if I have a package hanging around that uses the new C header?  No, as it is brandnew.

But you could take aim at for example [RcppXts](https://cloud.r-project.org/web/packages/RcppXts/index.html) which was set up (well over seven and a half years ago !!) to do the same `xts` and the export header there.  

It would be possible (and in fact very valuable) to have a similar examples package for `data.table`.  But right now we have ... one function.  Little stretched for an example package.

PR #4766, opened 2020-10-18
add check for trunc.cols

TysonStanley:
Simple update to `print.data.table()` to add a check for `trunc.cols` to be logical. In response to @jangorecki on #4074.


TysonStanley:
It doesn't look like any of the errors had anything to do with the print method, but rather with some other functions (specifically `fcase()`.

R/print.data.table.R
```diff
@@ -15,6 +15,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
   # trunc.cols - should only the columns be printed that can fit in the console? (FALSE)
   if (!col.names %chin% c("auto", "top", "none"))
     stop("Valid options for col.names are 'auto', 'top', and 'none'")
+  if (!is.logical(trunc.cols))
```
jangorecki, line 18: Could check if NA as well

ColeMiller1: Related: could use ```!isTRUEorFALSE(trunc.cols)```

TysonStanley: Now added check for NA. Hadn't ever used `isTRUEorFALSE()` so I didn't use that this time.

TysonStanley:
Regarding the test coverage, do we want to add checks for this? It is called in so many functions that I'm not sure if you want me to add to those but I can.

MichaelChirico:
a simple test of getting an error for the cases described is warranted. generally if there's an error that wasn't caught by our sweet already that we're fixing, we should add a new test for that error

TysonStanley:
Added three small checks.

R/print.data.table.R
```diff
@@ -15,6 +15,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
   # trunc.cols - should only the columns be printed that can fit in the console? (FALSE)
   if (!col.names %chin% c("auto", "top", "none"))
     stop("Valid options for col.names are 'auto', 'top', and 'none'")
+  if (!is.logical(trunc.cols) | is.na(trunc.cols))
```
jangorecki, line 18: operator `|` will not do a good job, it is a vectorized "or", and because is.logial and is.na are vectorized as well, we may ended up passing a logical vector to `if`.
because we want this to be scalar, the safe approach can be testing for `isTRUE || isFALSE`

jangorecki (commented):
tests look good, just arg validation can be more strict

R/print.data.table.R
TysonStanley: Fixed.

jangorecki (approved):

Issue #3618, opened 2019-05-30
add rchk and/or torture CI jobs


jangorecki:
followup of https://github.com/Rdatatable/data.table/issues/3615#issuecomment-497387143
- long time of torture run might be an issue
- rchk setup with singularity and not docker might be issue to
in both cases most likely solution would be to use private CI runner using ssh, same as we do for windows jobs now.
Another thing is that if those tests returns false positives often? if so it might be difficult to make them useful.

mattdowle:
torture doesn't have false positives.  Although rchk does. it will still be very useful to have rchk results to hand without me having to run it. It doesn't need to be a pass/fail in CI.

jangorecki:
Public GLCI runners have 3h timeout, so we will need to plug private runner to handle more time consuming jobs.

PR #4768, opened 2020-10-20
CI improvements

### LINKED ISSUE #4780: DT devel R package repo should not include non-dependencies packages
As of now we publish as well packages required for `other.Rraw` test script, which is far more than user needs for reproducibility.
We need two local repositories, and the one for `other.Rraw` shouldn't be included in artifacts for website. It also doesn't need binaries for Windows (3.6, 4.0, 4.1).

### LINKED ISSUE #4781: Remove plm tests from other.Rraw
AFAIU this change https://github.com/GeoBosh/Rdpack/commit/aa46920289016169be13f9beeed84351abc63d5a in one of the dependencies of `plm` (our extended tests dependency) introduced new dependency which itself has a new OS dependency. Adding new OS dependency just for an extra integration test of one package (2 small unit tests) doesn't look to be a best solution. My suggestion is to remove `plm` from our extended tests `other.Rraw` file.

jangorecki:
This PR does not
- include heavy jobs to be run on schedule: #3618
- make separate pipelines for artifacts/heavy jobs

Both are blocked on GLCI ability to run multiple pipelines for same branch.

This PR
- adds new Windows job "oldrel" that runs R 3.6, produce and publish binaries
- prepare Check table to support proving information from scheduled jobs (different version, commit sha)
- improves consistency to CRAN about quality/quantity of information provided (thanks @sindribaldur)
- updates CI yaml for newer, more robust syntax

----

Also removes plm from our extended suggests other.Rraw as it require new OS dependency which we don't want, closes #4781.
Also closes #4780 by not publishing extended suggests package to final R repo on website.


.ci/publish.R
```diff
@@ -1,12 +1,17 @@
 format.deps <- function(file, which) {
   deps.raw = read.dcf(file, fields=which)[[1L]]
   if (all(is.na(deps.raw))) return(character())
+  deps.raw = gsub("\n", " ", deps.raw, fixed=TRUE)
   deps.full = trimws(strsplit(deps.raw, ", ", fixed=TRUE)[[1L]])
   deps = trimws(sapply(strsplit(deps.full, "(", fixed=TRUE), `[[`, 1L))
+  deps.full = gsub(">=", "&ge;", deps.full, fixed=TRUE)
+  deps.full = gsub("<=", "&le;", deps.full, fixed=TRUE)
+  if (any(grepl(">", deps.full, fixed=TRUE), grepl("<", deps.full, fixed=TRUE), grepl("=", deps.full, fixed=TRUE)))
```
MichaelChirico, line 9: probably `grepl("[<=>]", deps.full)` is better than three `fixed` greps

.ci/publish.R
```diff
@@ -26,6 +31,39 @@ format.bins <- function(ver, bin_ver, cran.home, os.type, pkg, version, repodir)
   paste(ans[fe], collapse=", ")
 }
 
+format.entry <- function(field, dcf, url=FALSE) {
+  if (field %in% colnames(dcf)) {
+    value = gsub("\n", " ", dcf[,field], fixed=TRUE)
+    if (url) {
+      urls = trimws(strsplit(value, ",", fixed=TRUE)[[1L]])
+      value = paste(sprintf("<a href=\"%s\">%s</a>", urls, urls), collapse=", ")
```
MichaelChirico, line 39: i have learned to love `toString` for this :smiley:

.ci/publish.R
```diff
@@ -186,30 +308,36 @@ check.index <- function(pkg, jobs, repodir="bus/integration/cran") {
...
-               sprintf("<h2>Package Check Results for Package <a href=\"../packages/%s/index.html\"> %s </a> </h2>", pkg, pkg),
-               sprintf("<p>Last updated on %s.</p>", format(Sys.time(), usetz=TRUE)),
-               sprintf("<table border=\"1\" summary=\"CRAN check results for package %s\">", pkg),
-               "<tr><th>Test job</th><th>OS type</th><th>Install</th><th>Check</th><th>Rout.fail</th><th>Memtest</th></tr>",
-               tbl,
-               "</table>",
-               "</body>",
-               "</html>"),
-             file)
+  writeLines(c(
```
MichaelChirico, line 324: is this modeled off how it's done for CRAN itself?

.ci/publish.R
MichaelChirico: for readability as much as efficiency (though maybe i'm just too used to regex :) )

inst/tests/other.Rraw
```diff
@@ -155,15 +155,6 @@ if (loaded[["knitr"]]) {
   test(11, kable(DT), output="x.*y.*1.*2")
 }
 
-# for plm package
```
MichaelChirico, line 158: perhaps we could offer to add this test to plm suite instead?

jangorecki: palm does not depends on DT so I don't think it is worth it

mattdowle:
Have seen the failed appveyor-devel job is due to the `-alike` change in latest R-devel, which is fixed in master by #4770. So proceeding to merge.

jangorecki:
New syntax for defining dependencies introduced in this PR gave us new DAG graph: https://gitlab.com/Rdatatable/data.table/-/pipelines/210901332/dag
That more precisely visualize dependencies between jobs.
It also let us to download a little less artifacts between different jobs. For deploy docker build jobs we didn't need website artifacts from integration stage but we do still need to wait for this stage to finish.

jangorecki:
Another nice thing is that check summary table provides now more details:
https://rdatatable.gitlab.io/data.table/web/checks/check_results_data.table.html

AdrianAntico:
@jangorecki 

Hey guys, not sure if this recent update is causing the warning below or if my machine is acting weird. Nonetheless, I started receiving a warning today when running library(data.table). Code info is below:

```
> library(data.table)
data.table 1.13.2 using 32 threads (see ?getDTthreads).  Latest news: r-datatable.com
Warning message:
package ‘data.table’ was built under R version 4.0.3

> sessionInfo()
R version 4.0.2 (2020-06-22)

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] data.table_1.13.2

[1] compiler_4.0.2 tools_4.0.2   
```

jangorecki:
This upgrade is not a cause of warning you are getting.
First of all you are using 1.13.2 which must have been installed from CRAN therefore nothing related to our CI pipeline which releases only odd version numbers.
AFAIU warning is caused by the fact that CRAN ships data.table built on r-release (4.0.3) and you use previous version 4.0.2. If you want to get rid of the warning, I think `install.packages("data.table", type="source")` should be sufficient. Alternatively you can also upgrade your R version to 4.0.3, that will also resolve the warning.

AdrianAntico:
@jangorecki  Thanks for getting back to me. I just reinstalled with type="source" and I'm no longer seeing the warning.

PR #4770, opened 2020-10-21
update unit test err msg, closes #4769

### LINKED ISSUE #4769: R-devel changed error message breaks 1199.2
Recent release 1.13.2 of data.table was nicely passing cran checks for a couple of hours: https://cran.r-project.org/web/checks/check_results_data.table.html
15 hours ago there was a commit to R https://github.com/wch/r-source/commit/138a8c50ff56310d61169496fe94e488704c13c0 that broke test 1199.2.
~~My suggestion is to fix it for 1.13.4 and submit straight away to CRAN, so no need for revdeps, ASAN checks.~~ Actually it is not that urgent as any code isn't actually broken. Unit test was against base R. Revdeps and users are not affected.

jangorecki:
closes #4769
passed on 2020-10-20 r79356 https://gitlab.com/jangorecki/data.table/-/jobs/803180097
R change 2020-10-20 r79353

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 ## NOTES
 
-1. Within hours after 1.13.2 release to CRAN there was a change in error message of base R that broke one of our unit tests, [#4769](https://github.com/Rdatatable/data.table/issues/4769).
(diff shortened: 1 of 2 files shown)


mattdowle:
I changed the pattern to be minimal in case of any similar minor changes to R's message in future.  We should probably review all tests that look for R's messages, and similarly relax them. Not sure how to find them though.

For the news item, I took it as an opportunity to give credit to CRAN.

MichaelChirico:
> Not sure how to find them though.

suggestion -- run an R session in chinese & run tests. i think you'll need to tweak test.data.table() since we made that change for non-english sessions

PR #4776, opened 2020-10-24
Merge no_keyed x with keyed y

### LINKED ISSUE #4772: three data.table merge behavior inconsistency
i looked around for related discussions and didn't find anything specific to this?  i posted on SO and the accepted answerer there provided more detail about how it's likely a bug and also an easy fix--  https://stackoverflow.com/questions/64464386/three-data-table-merge-behavior-inconsistency/64464931#64464931

thanks very very much for this excellent software!

	library(data.table)

	x <- data.table( a = 1:3 )
	y <- data.table( a = 2:4 )
	z <- data.table( a = 3:5 )

	# works
	merge( x , y )
	# works
	merge( y , z )

	# fails
	merge( x , merge( y , z ) )
	# Error in merge.data.table(x, merge(y, z)) :
	#   A non-empty vector of column names for `by` is required.

	# works
	merge( merge( x , y ) , z )

ben-schwen:
Closes #4772 

`base::intersect(x,y)` returns an argument of `typeof(y)`. This leads to the behaviour that `intersect(NULL, "a") = character(0)` while `intersect("a", NULL) = NULL`. The previous checks for the `by` argument led to an unnecessary error when `intersect` evaluated to `character(0)`.

**Note:** This leads also to the change of test 1962.020 which explicitly checked for `by = character(0L)` and led to an error before.


MichaelChirico:
The underlying issue was already fixed in #5183. I'll change this PR to just add more regression tests.

MichaelChirico (approved):
Thanks! Glad it's fixed in the meantime

Issue #1500, opened 2016-01-15
as.Date.IDate won't change the class if xts package loaded


jangorecki:
To convert your IDate into Date you currently need to go through `as.integer`.
```r
library(data.table)
id = as.IDate("2016-01-15")
class(id)
#[1] "IDate" "Date" 
class(as.Date(id))
#[1] "IDate" "Date" 
class(as.Date(as.integer(id)))
#[1] "Date" 
```
when resolved update this [SO](http://stackoverflow.com/a/34812810/2490497).

franknarf1:
I see different output:

    > library(data.table)
    > id = as.IDate("2016-01-15")
    > class(id)
    [1] "IDate" "Date" 
    > #[1] "IDate" "Date" 
    > class(as.Date(id))
    [1] "Date"
    > #[1] "IDate" "Date" 
    > class(as.Date(as.integer(id)))
    Error in as.Date.numeric(as.integer(id)) : 'origin' must be supplied

    > sessionInfo()
    R version 3.2.3 (2015-12-10)
    Platform: i386-w64-mingw32/i386 (32-bit)
    Running under: Windows 7 x64 (build 7601) Service Pack 1

    locale:
    [1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
    [3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
    [5] LC_TIME=English_United States.1252    

    attached base packages:
    [1] stats     graphics  grDevices utils     datasets  methods   base     

    other attached packages:
    [1] data.table_1.9.7

    loaded via a namespace (and not attached):
    [1] chron_2.3-47

and the same with 64-bit (`Platform: x86_64-w64-mingw32/x64 (64-bit)`).

jangorecki:
Just tested latest data.table and I can confirm it is working as expected:
```r
library(data.table)
id = as.IDate("2016-01-15")
class(id)
#[1] "IDate" "Date" 
class(as.Date(id))
#[1] "Date"
class(as.Date(as.integer(id)))
#Error in as.Date.numeric(as.integer(id)) : 'origin' must be supplied
sessionInfo()
#R version 3.2.3 (2015-12-10)
#Platform: x86_64-pc-linux-gnu (64-bit)
#Running under: Ubuntu 14.04.3 LTS
#
#locale:
[...]

PR #4782, opened 2020-10-27
do not export as.Date.IDate anymore

### LINKED ISSUE #4777: S3 method lookup found 'as.Date.IDate' on search path
One of [CRAN machines reports](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-clang/data.table-00check.html) an issue caused by a workaround implemented to resolve https://github.com/Rdatatable/data.table/issues/1500 in 4f6908af4efee3a23bee27a8b8e51ae716d3941b

```
  S3 method lookup found 'as.Date.IDate' on search path
  S3 method lookup found 'as.Date.IDate' on search path
```
According to changelog of zoo, their as.Date method has been changed in 1.8-1 (Jan 2018), so it is likely that workaround is no longer needed.

jangorecki:
closes #4777


MichaelChirico:
FWIW i've got a mirror of CRAN locally, statically checked if any revdeps (recursive depends/imports) were using the previous export, didn't find any.

eddelbuettel:
FWIW comes up too when letting GitHub search in the CRAN mirror:  https://github.com/search?q=org%3Acran+as.Date.IDate

NAMESPACE
```diff
@@ -151,8 +151,11 @@ S3method("+", IDate)
 S3method("-", IDate)
 S3method(as.character, ITime)
 S3method(as.data.frame, ITime)
-S3method(as.Date, IDate)   # note that zoo::as.Date masks base::as.Date. Both generic.
-export(as.Date.IDate)      # workaround for zoo bug, see #1500. Removing this export causes CI pipeline to fail on others.Rraw test 6, but I can't reproduce locally.
+if (getRversion() >= "3.6.0") {
+  S3method(as.Date, IDate)
```
mattdowle, line 155: @jangorecki Why >= "3.6.0"?  Won't `as.Date(x)` break in R < 3.6.0 if the method is not exported?
Also could the comments be put back so we know there's some history on these lines pointing to #1500.

mattdowle (changes requested):

NAMESPACE
mattdowle: Higher up in the file the switch on 3.6.0 is about delayed registration (i.e. the `pkg::` prefix on the first argument to `S3method`) but that's not the case here with `as.Date`.

mattdowle: I figured the >=3.6 was copied from higher up. So I just went ahead and removed it, to merge and let's see if it passes CI. If this was wrong then we can always do a follow up PR.

mattdowle: When I added the suggest version for zoo, I did bit64 and bit at the same time (although nothing to do with this PR).

mattdowle (approved):

PR #4785, opened 2020-10-27
fwrite() UTF-8 csv file

### LINKED ISSUE #1770: fwrite UTF8

shrektan:
Fixes #1770 

## TODO

-  ~~Consider other encodings? (maybe not to reduce the complexity)~~
- [x] change the argument name from `fileEncoding` to `encoding`, the default values and availables should be close to what  fread() already has
- [x] Add encoding="native" to ensure the strings written are in native encoding.
- [x] Confirm that `bom=TRUE` works great
- ~~Consider export new functions `fwrite8()` / `fread8()` which forces the use of UTF-8 encoding (maybe a seperate PR?)~~
- [x] Update Rd file for the new argument
- [x] Add tests
- [x] Update NEWS 

## Example

``` r
library(data.table)
latin1 = "fa\xE7ile"
Encoding(latin1) = "latin1"
utf8 = iconv(latin1, "latin1", "UTF-8")
text = c(latin1, utf8, "aaaaaaaa")
dt = data.table(A = text, B = as.factor(text))
csvfile = tempfile(fileext = ".csv")

fwrite(dt, csvfile, encoding = "UTF-8")
fread(csvfile, encoding = "UTF-8")
#>           A        B
#>      <char>   <char>
#> 1:   façile   façile
#> 2:   façile   façile
#> 3: aaaaaaaa aaaaaaaa

fwrite(dt, csvfile, encoding = "native")
fread(csvfile)
#>           A        B
#>      <char>   <char>
#> 1:   façile   façile
#> 2:   façile   façile
#> 3: aaaaaaaa aaaaaaaa

fwrite(dt, csvfile, encoding = "")
fread(csvfile)
#>            A         B
#>       <char>    <char>
#> 1: fa\xe7ile fa\xe7ile
#> 2:    façile fa\xe7ile
#> 3:  aaaaaaaa  aaaaaaaa
```

<sup>Created on 2020-10-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


jangorecki:
why argument name `fileEncoding` and not just `encoding`?

fileEncoding="native" sounds better than "".
as for BOM, why we cannot use just encoding utf8 and existing argument `bom=TRUE`?
fwrite8() / fread8() better as a follow up PR, they will use fread/fwrite anyway

shrektan:
> why argument name fileEncoding and not just encoding?

I took it from `utils::read.csv()` but I do not like it. If we are not bound to have the same argument names as base R, I'm happy to change the name.

> fileEncoding="native" sounds better than "".

Yes, agreed.

> as for BOM, why we cannot use just encoding utf8 and existing argument bom=TRUE?

Ha, I missed this. Just did a simple test and setting bom=TRUE seems works great. Will confirm.

> fwrite8() / fread8() better as a follow up PR, they will use fread/fwrite anyway

Agreed.

jangorecki:
We have already `encoding` in fread, it is probably align to that one instead of base R.

src/fwriteR.c
```diff
@@ -5,18 +5,23 @@
 #define DATETIMEAS_EPOCH     2
 #define DATETIMEAS_WRITECSV  3
 
+#define NEED2NATIVE(s) !(IS_ASCII(s) || (s)==NA_STRING)
```
jangorecki, line 8: minor stuff...
I think checking for NA is cheaper than for ASCII so changing order of that check could give some speed up
```
(s)!=NA_STRING && !IS_ASCII(s)
```

src/fwriteR.c
```diff
@@ -5,18 +5,23 @@
...
+#define ENCODED_CHAR(s) (utf8 && NEED2UTF8(s) ? translateCharUTF8(s) : \
+  (native && NEED2NATIVE(s)) ? translateChar(s) : CHAR(s))
```
jangorecki, line 10: I don't get parenthesis here... does it work as expected and I am just unable to read it properly? :/
why not like this?
```
(utf8 && NEED2UTF8(s)) ? translateCharUTF8(s) : ((native && NEED2NATIVE(s)) ? translateChar(s) : CHAR(s))
```

shrektan: I just think it's too long to fit in a single line... nothing special. I'm ok to put it to one line.

jangorecki: it is not just single line but my option has different parentheses, two lines are no problem

jangorecki: ah ok, I think I know how that works. It just doesn't have "else" condition into new parentheses. I would personally add those parentheses so it is easier to see branches in this ternary operator. And multiline is fine.

src/fwriteR.c
```diff
@@ -5,18 +5,18 @@
...
 #define DATETIMEAS_WRITECSV  3
 
-#define NEED2NATIVE(s) !(IS_ASCII(s) || (s)==NA_STRING)
-#define ENCODED_CHAR(s) (utf8 && NEED2UTF8(s) ? translateCharUTF8(s) : \
-  (native && NEED2NATIVE(s)) ? translateChar(s) : CHAR(s))
+static bool utf8=false;
+static bool native=false;
+#define TO_UTF8(s) (utf8 && NEED2UTF8(s))
+#define TO_NATIVE(s) (native && (s)!=NA_STRING && !IS_ASCII(s))
+#define ENCODED_CHAR(s) (TO_UTF8(s) ? translateCharUTF8(s) : TO_NATIVE(s) ? translateChar(s) : CHAR(s))
```
jangorecki, line 12: thanks, now is more readable, still it is not obvious if
```
TO_UTF8(s) ? translateCharUTF8(s) : TO_NATIVE(s)
```
is meant to evaluate to boolean condition for second ternary operator.
If we just add extra `(` there it is then very clear that it isn't:
```
 (TO_UTF8(s) ? translateCharUTF8(s) : (TO_NATIVE(s) ? translateChar(s) : CHAR(s)))
```

shrektan: Ha. I sometimes write this kind of multiple ternary operator, splitting to multiple lines actually looks clear to me (maybe I just get used to it). I'm ok to add the extra `()`.

jangorecki: Thank you for your patience, I can imagine so many minor stuff can be annoying :)

jangorecki (approved):
PR looks very good. If I understand correctly checks for !ascii and !utf8 are performed in loops, then would be good to measure its overhead for a regular ascii table. The overhead is just from testing for `native` and `utf8` boolean vars, which will be extremely fast so I don't think it will be problem, but when doing in loops it can adds up little bit. We are interested in the most pessimistic case.

R/fwrite.R
```diff
@@ -11,8 +11,12 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
            compress = c("auto", "none", "gzip"),
            yaml = FALSE,
            bom = FALSE,
-           verbose=getOption("datatable.verbose", FALSE)) {
+           verbose=getOption("datatable.verbose", FALSE),
+           encoding = "") {
```
MichaelChirico, line 15: nit: could have signature `encoding = c("", "UTF-8", "native")` and use `match.args` instead of the check below

man/fwrite.Rd
```diff
@@ -59,6 +60,7 @@ fwrite(x, file = "", append = FALSE, quote = "auto",
   \item{yaml}{If \code{TRUE}, \code{fwrite} will output a CSVY file, that is, a CSV file with metadata stored as a YAML header, using \code{\link[yaml]{as.yaml}}. See \code{Details}. }
   \item{bom}{If \code{TRUE} a BOM (Byte Order Mark) sequence (EF BB BF) is added at the beginning of the file; format 'UTF-8 with BOM'.}
   \item{verbose}{Be chatty and report timings?}
+  \item{encoding}{ The encoding of the strings written to the CSV file. Default is \code{""}, which means writting raw bytes without considering the encoding. Other possible options are \code{"UTF-8"} and \code{"native"}. }
```
MichaelChirico, line 63: if this matches the behavior of `write.table` (or other `base` function), we should mention that here

src/fwriteR.c
```diff
@@ -5,18 +5,23 @@
 #define DATETIMEAS_EPOCH     2
 #define DATETIMEAS_WRITECSV  3
 
+static bool utf8=false;
+static bool native=false;
+#define TO_UTF8(s) (utf8 && NEED2UTF8(s))
+#define TO_NATIVE(s) (native && (s)!=NA_STRING && !IS_ASCII(s))
```
MichaelChirico, line 11: @jangorecki I think you recommended putting `IS_ASCII` last right? We should do that in `data.table.h` then too right?

```
#define NEED2UTF8(s) !(IS_ASCII(s) || (s)==NA_STRING || IS_UTF8(s))
```

src/fwriteR.c
```diff
@@ -5,18 +5,23 @@
 #define DATETIMEAS_EPOCH     2
 #define DATETIMEAS_WRITECSV  3
 
+static bool utf8=false;
+static bool native=false;
+#define TO_UTF8(s) (utf8 && NEED2UTF8(s))
```
MichaelChirico, line 10: nit: `TO_UTF8` reads to me like "convert `s` to UTF-8" but IINM it's a logical check. Unfortunately the most logical choice is `NEED2UTF8` which is taken... `REALLY_NEED2UTF8`? :stuck_out_tongue: 

Nothing comes to mind unfortunately, worth considering some more though.

src/fwriteR.c
```diff
@@ -5,18 +5,23 @@
...
+#define TO_NATIVE(s) (native && (s)!=NA_STRING && !IS_ASCII(s))
+#define ENCODED_CHAR(s) (TO_UTF8(s) ? translateCharUTF8(s) : (TO_NATIVE(s) ? translateChar(s) : CHAR(s)))
```
MichaelChirico, line 12: note that `TO_UTF8` and `TO_NATIVE` both run `TO_ASCII` and `S==NA_STRING`, so I think there's some redundancy here. 

how about

```
s == NA_STRING ? CHAR(s) : ( IS_ASCII(s) ? CHAR(s) : ( native ? translateChar(s) : IS_UTF8(s) ? translateCharUTF8(s) : CHAR(s) ) )
```

based on this truth table (the other 22/32 combinations are impossible because `utf8` :arrow_right: `!native` and `IS_ASCII` :arrow_right: `!IS_NATIVE` (and _vice versa_) 

Truth table of all possible IS_ASCII, IS_UTF8, s==NA_STRING, utf8, native values

R/fwrite.R
shrektan: The encoding arg of `fread()` doesn't use this signature, either. The reason is the empty string seems not work well with `match.arg()` :

``` r
f <- function(x = c("", "a", "b")) { 
  match.arg(x)
}
f()
#> [1] ""
f("a")
#> [1] "a"
f("b")
#> [1] "b"
f("") # this throws
#> Error in match.arg(x): 'arg' should be one of "", "a", "b"
```

<sup>Created on 2020-10-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

man/fwrite.Rd
shrektan: It's a little bit different and the `fileEncoding` section of `base::write.table()` seems not very clear to me.

src/fwriteR.c
shrektan: Yes, you're right but after a 2nd thought, I still prefer the current version. The reasons are:

1. In terms of speed, the original version doesn't do anything unnecessary. For the default encoding value "", checking two boolean values should be much faster than checking NA_STRING and IS_ASCII. 

1. In terms of clarity, despite there's a small redundancy, the current version is easier to be understood.

R/fwrite.R
shrektan: Filed the bug to https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17959

mattdowle:
I did a quick benchmark to test for a slow down on regular ascii, master vs this branch.

```R
ids = as.character(as.hexmode(1:1e6))
DT = data.table(A=sample(ids, 1e8, replace=TRUE))
tables()
#    NAME        NROW NCOL  MB COLS KEY
# 1:   DT 100,000,000    1 816    A    
# Total: 816MB
system.time(fwrite(DT, tempfile()))   # ~2.5s in both master and this PR
#    user  system elapsed 
#   7.823   0.323   2.512 
```

Looks good. I was thinking the extra branch could possibly have an impact, but of course each string (and its header in the global cache) has to be fetched anyway to write it, and that fetch dominates any branch or logic considerations.

Issue #4572, opened 2020-06-23
fwrite output as a character vector


mrdwab:
Given the speed of `fwrite`, it can be used in conjunction with `fread` as an alternative to `do.call(paste, ...)` to flatten multiple columns into a character vector. It would be nice to be able to capture the output of `fwrite` directly as a character vector.

It is **[much faster](https://stackoverflow.com/a/62540449/1270695)** than some of the other idiomatic approaches that are often considered.

Here's the behavior I'm hoping to be able to replicate:

```r
fpaste <- function(dt, sep = ",") {
  x <- tempfile()
  fwrite(dt, file = x, sep = sep, col.names = FALSE)
  fread(x, sep = "\n", header = FALSE)
}

d <- data.frame(a = 1:3, b = c('a','b','c'), c = c('d','e','f'), d = c('g','h','i')) 
cols = c("b", "c", "d")

fpaste(d[cols], "-")
#       V1
# 1: a-d-g
# 2: b-e-h
# 3: c-f-i
```   
Here's a comparison with a straightforward `paste` in a `data.table`:

    set.seed(1) 
    d2 <- d[sample(1:3,1e6,TRUE),]
    d3 <- as.data.table(d2)

    bench::mark(fpaste(d2[cols], "-")$V1, d3[, paste(b, c, d, sep = "-")])
    ## # A tibble: 2 x 13
    ##   expression                          min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
    ##   <bch:expr>                      <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
    ## 1 fpaste(d2[cols], "-")$V1         90.2ms  93.8ms     10.8     8.41MB     3.60     3     1      278ms
    ## 2 d3[, paste(b, c, d, sep = "-")] 220.9ms 223.2ms      4.43   30.55MB     0        3     0      678ms
    ## # … with 4 more variables: result <list>, memory <list>, time <list>, gc <list>

MichaelChirico:
for the record I tried using `capture.output` instead of disk I/O and it's way, way slower (I gave up running the benchmark)

mrdwab:
@MichaelChirico I had forgotten to mention in my original post that I tried with `capture.output` and also gave up, and then tried with `R.utils::captureOutput` which performed much better, but still slower than `fpaste`.

ColeMiller1:
[...]

PR #4805, opened 2020-11-11
use writeLines for fread(text=.)

MichaelChirico:
Cleaner & equivalent behavior

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -35,7 +35,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="")
     if (!is.character(text)) stop("'text=' is type ", typeof(text), " but must be character.")
     if (!length(text)) return(data.table())
     if (length(text) > 1L) {
-      cat(text, file=(tmpFile<-tempfile(tmpdir=tmpdir)), sep="\n")  # avoid paste0() which could create a new very long single string in R's memory
+      writeLines(text, tmpFile<-tempfile(tmpdir=tmpdir))  # avoid paste0() which could create a new very long single string in R's memory
       file = tmpFile
       on.exit(unlink(tmpFile), add=TRUE)
     } else {


R/fread.R
```diff
@@ -35,7 +35,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="")
     if (!is.character(text)) stop("'text=' is type ", typeof(text), " but must be character.")
     if (!length(text)) return(data.table())
     if (length(text) > 1L) {
-      cat(text, file=(tmpFile<-tempfile(tmpdir=tmpdir)), sep="\n")  # avoid paste0() which could create a new very long single string in R's memory
+      writeLines(text, tmpFile<-tempfile(tmpdir=tmpdir))  # avoid paste0() which could create a new very long single string in R's memory
```
MichaelChirico, line 38: We could also use `useBytes=TRUE` here to facilitate writing UTF-8 strings on Windows:

https://stackoverflow.com/q/10675360/3576984

mrdwab: @MichaelChirico This isn't something I've tested, but maybe `fwrite(as.data.table(text), tmpFile<-tempfile(tmpdir = tmpdir), col.names = FALSE)` is also worth exploring. I'm basing this assumption on the speed that I observed [here](https://github.com/Rdatatable/data.table/issues/4572#issue-644031582).

MichaelChirico: In general I think it may warrant some benchmarking. Actually I have run into an issue recently where `cat()` is very slow for long input. In my case I was `cat`-ing a whole file by character (e.g. `cat(strsplit(readChar(f, file.size(f)), NULL)[[1]], sep = '', file = f)` is quite slow)

MichaelChirico:
A benchmark of approaches (h/t @mrdwab for suggesting `fwrite`):

```
library(data.table)
library(microbenchmark)
library(withr)

mtcars = as.data.table(mtcars, keep.rownames = "make")
rows = do.call(paste, c(unname(mtcars), list(sep = ",")))
meta = data.table(
  n_rows = c(1e2, 1e5, 1e8),
  n_reps = c(1e3, 1e2, 5)
)
for (ii in seq_len(nrow(meta))) {
  cat(sprintf("timing for emitting %d rows of input\n", meta$n_rows[ii]))
  txt = sample(rows, meta$n_rows[ii], TRUE)
  tmp <- tempfile()
  print(microbenchmark::microbenchmark(
    times = meta$n_reps[ii],
    cat = with_tempfile(tmp, cat(txt, file = tmp, sep = "\n")),
    writeLines = with_tempfile(tmp, writeLines(txt, tmp)),
    fwrite = with_tempfile(tmp, fwrite(list(rows), tmp, quote=FALSE))
  ))
}
```

output:

```
timing for emitting 100 rows of input
Unit: microseconds
       expr     min       lq     mean   median       uq      max neval cld
        cat 355.504 363.2115 381.2248 366.9645 374.1155 2365.032  1000   c
 writeLines  92.257  96.9025 104.1850  99.6420 101.9755  268.826  1000 a  
     fwrite 129.107 140.6755 156.1185 144.9870 150.6950 2355.349  1000  b 
timing for emitting 100000 rows of input
Unit: microseconds
       expr        min         lq       mean    median        uq       max neval cld
        cat 290911.277 295095.536 303709.532 305186.89 308638.26 330139.37   100   c
 writeLines  21014.126  25466.695  32991.918  36208.78  37545.37  48450.52   100  b 
     fwrite    158.899    281.687   9117.068  11869.05  12574.75  17852.94   100 a  
timing for emitting 100000000 rows of input
Unit: microseconds
       expr          min          lq        mean      median          uq         max neval cld
        cat 2.841903e+08 284357875.9 289614328.1 290874438.7 294160661.5 294488373.2     5   c
 writeLines 2.261030e+07  24129058.5  27682231.0  26859014.4  28238340.4  36574441.3     5  b 
     fwrite 2.593020e+02    412013.1    336822.1    419335.1    421485.6    431017.2     5 a  
```

So `fwrite()` is clearly a winner if people are using `fread` to ingest huge character vectors, but I don't think that's a primary use case here. I lean to keeping `writeLines` for simplicity.

PR #4806, opened 2020-11-11
testing is more robust to changes in base messages

MichaelChirico:
Related: #4770 

To figure out which messages are produced by base:

First, I edited `R/test.data.table.R` to force `foreign=FALSE`, then I commented out the `setnames` call between tests 1904.5 and 1904.6 (it relies on an object that doesn't get created due to test failure), then I ran:

```
# note: `es` is a language where R has translations but data.table does not,
#   meaning DT messages are in English but base are in Spanish
LANGUAGE=es Rscript -e "library(data.table); test.data.table()" > test_failures.txt
```

There were two cases where the approach here wouldn't work:

 1. #4989 -- because `.checkTypos` relies on English messages, in non-English `.checkTypos` can never return suggested fixes.
 2. test 1671 is testing an error that's generated in C, I don't see a way to generate that directly from R short of writing a short C function solely for this purpose which seems like a bit much. The `gettext()` approach here is robust to other languages, but not to R's own message changing.


mattdowle:
@MichaelChirico Good idea to force `foreign=FALSE`.  But LANGUAGE=zh_CN so our own messages are being translated. How about picking a language that `data.table` is not translated for (and so data.table will emit English and will pass our English tests) but base R is translated for.  Then we will see the base R messages in non-English not matching our English tests which should be a much smaller list.

mattdowle:
It's especially nice that the base messages appear as `base_messages` in the `test()` call now. That conveys more, reads well, and can be searched. Great.

PR #4810, opened 2020-11-17
extend g[ to work for tail queries like x[.N]

### LINKED ISSUE #4809: Slow .SD[.N] compare to last(.SD) with groupby
I noticed there is significant speed difference between 
`DT[,.SD[.N], by=col]` and 
`DT[,.last(.SD), by=col]`

for relatively large data (10 * 1.8M rows). 

image

`DT[,.SD[1],by=col]` and `DT[,first(.SD), by=col]` did not show difference in performance, however.

image

MichaelChirico:
Closes #4809 

The idea is to detect `x[.N - k]` cases and send them to `g[`, which gets a new argument `fromLast` (default `FALSE`), so `x[1L]` becomes `"g["(x, 1L, FALSE)` and `x[.N-2L]` becomes `"g["(x, 2L, TRUE)`. Logic for `x[.N]` is basically the same as for `glast`.

Two things for feedback:

 - Any suggestions for improvement of the R implementation? In particular the part where we edit `jsub` feels somewhat redundant (running roughly same somewhat long list of checks on `jsub` a second time)
 - The edge case here has different semantics than for the old `fromLast=FALSE` cases -- in particular `DT[,x[0],id]` behaves differently than `DT[,x[.N+1],id]` -- any suggestions for how to implement this cleanly? Maybe it should be an error instead of allowing it and returning `NA`? That's not very nice (e.g. the workaround of suggesting the `base::` alternative is quite clunky for `[`) & would probably break some code.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 ## NEW FEATURES
 
+1. Group-wise queries like `x[.N]` or `.SD[.N-1L]` (i.e. an offset of `k>=0` from the end of the group at index `.N`) are now optimized, [#4809](https://github.com/Rdatatable/data.table/issues/4809) and part of [#735](https://github.com/Rdatatable/data.table/issues/4809). Previously only fixed positive offsets like `x[1L]` or `.SD[2L]` were optimized. Thanks to @matthewgson for the report.
+
 ## BUG FIXES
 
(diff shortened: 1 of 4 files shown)


MichaelChirico:
@ben-schwen any high-level comments here before I try and update to current `master`? given your recent heavy work in this file

ben-schwen:
> @ben-schwen any high-level comments here before I try and update to current `master`? given your recent heavy work in this file

Nah I guess lessons learned in #5548 and #5887 should be enough and actually should make this a lot smoother.

Ideally, `"g["(x, n, TRUE)` should also vector when `n` is of length longer than 1 (similar approach as in `head/tail`), but I guess this is another issue

PR #4815, opened 2020-11-19
runlock doesnt try to unlock functions

### LINKED ISSUE #4814: Aggregation fails for list columns with closures that don't have class "function"
```r
f <- function(x) x
f2 <- f
class(f2) <- "x"
dt <- data.table(f, f2)
f <- function(x) x
f2 <- f
class(f2) <- "x"
dt <- data.table(id = 1, f, f2)
dt[, .(f), by = id]
#>    id             f
#> 1:  1 <function[1]>
dt[, .(f2), by = id]
#> Error in as.vector(x, "list") : 
#>   cannot coerce type 'closure' to vector of type 'list'
```
I would expect the result of `.(f2)` aggregation to be similar to the `.(f)` aggregation.

(Maybe this is related to the fact that `as.list(f)` gives a result and `as.list(f2)` fails? But I don't think it *should* fail since the content of the `f` column is not the result of `as.list(f)`.)

```r
> sessionInfo()
R Under development (unstable) (2020-11-14 r79432)
Running under: Fedora 32 (Thirty Two)

Matrix products: default
BLAS:   /home/user/R-devel/lib64/R/lib/libRblas.so
LAPACK: /home/user/R-devel/lib64/R/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.utf8       LC_NUMERIC=C             
 [3] LC_TIME=en_US.utf8        LC_COLLATE=en_US.utf8    
 [5] LC_MONETARY=en_US.utf8    LC_MESSAGES=en_US.utf8   
 [7] LC_PAPER=en_US.utf8       LC_NAME=C                
 [9] LC_ADDRESS=C              LC_TELEPHONE=C           
[11] LC_MEASUREMENT=en_US.utf8 LC_IDENTIFICATION=C      

attached base packages:

other attached packages:
[1] data.table_1.13.2

[1] compiler_4.1.0
```

MichaelChirico:
Closes #4814


jangorecki:
LGTM

mattdowle:
Travis CI is stuck in queued status again. Proceeding without. GLCI will check after merge and covers it anyway.

PR #4822, opened 2020-11-26
Updated error message in rbindlist to display the class attributes of mismatching columns

logworthy:
Thought it would be helpful to provide a little more context to this error message.

Old:  `Class attribute on column 1 of item 2 does not match with column 1 of item 1.`
New:  `Class attribute on column 1 (POSIXct, POSIXt) of item 2 does not match with column 1 (Date) of item 1.`

I've added some tests and attempted to update the translations.


src/rbindlist.c
```diff
@@ -310,10 +312,14 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
-                                   PROTECT(getAttrib(firstCol, R_ClassSymbol)),
-                                   0)) {
-            error(_("Class attribute on column %d of item %d does not match with column %d of item %d."), w+1, i+1, firstw+1, firsti+1);
+          thisClass = PROTECT(getAttrib(thisCol, R_ClassSymbol));
+          firstClass = PROTECT(getAttrib(firstCol, R_ClassSymbol));
+          if (!R_compute_identical(thisClass, firstClass, 0)) {
+            char* thisClassStr = joinCharVec(thisClass, ", ");
+            char* firstClassStr = joinCharVec(firstClass, ", ");
+            error(_("Class attribute on column %d (%s) of item %d does not match with column %d (%s) of item %d."), w+1, thisClassStr, i+1, firstw+1, firstClassStr, firsti+1);
+            free(thisClassStr);
```
MichaelChirico, line 321: UNPROTECT() not free()

MichaelChirico: actually, isn't the UNPROTECT() just below this branch targeting the two allocations?

src/rbindlist.c
```diff
@@ -270,6 +270,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
   }
 
   SEXP coercedForFactor = NULL;
+  SEXP thisClass = R_NilValue;
```
MichaelChirico, line 273: any reason to declare these here instead of closer to their usage?

src/utils.c
```diff
@@ -374,3 +374,37 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
...
+char *joinCharVec (SEXP x, const char *sep)
+{
+    char *joined = NULL;                /* pointer to joined string w/sep */
+    size_t lensep = strlen (sep),       /* length of separator */
+        sz = 0;                         /* current stored size */
+    int first = 1;                      /* flag whether first term */
+
+    /* check that a character vector has been passed */
+    if (TYPEOF(x) != STRSXP)
+      error(_("Internal error: unsupported type '%s' passed to joinCharVec()"), type2char(TYPEOF(x)));
```
MichaelChirico, line 389: add // # nocov

src/rbindlist.c
MichaelChirico: nvm

src/rbindlist.c
```diff
@@ -310,10 +312,14 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
-                                   0)) {
-            error(_("Class attribute on column %d of item %d does not match with column %d of item %d."), w+1, i+1, firstw+1, firsti+1);
+          thisClass = PROTECT(getAttrib(thisCol, R_ClassSymbol));
+          firstClass = PROTECT(getAttrib(firstCol, R_ClassSymbol));
+          if (!R_compute_identical(thisClass, firstClass, 0)) {
+            char* thisClassStr = joinCharVec(thisClass, ", ");
+            char* firstClassStr = joinCharVec(firstClass, ", ");
+            error(_("Class attribute on column %d (%s) of item %d does not match with column %d (%s) of item %d."), w+1, thisClassStr, i+1, firstw+1, firstClassStr, firsti+1);
+            free(thisClassStr);
+            free(firstClassStr);
```
jangorecki, line 322: How those free are ever reached when there is error call above?

src/utils.c
```diff
@@ -374,3 +374,37 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
   return(ans);
 }
 
+// Convert a character vector into a single string, e.g. for printing error messages
+// adapted from https://stackoverflow.com/a/58163237
+// make sure to *free* the returned string after use
+char *joinCharVec (SEXP x, const char *sep)
```
jangorecki, line 380: I would prefer not using join word for this kind of operation because we use it commonly in context of R's merge. Maybe just concatenate?

src/utils.c
```diff
@@ -374,3 +374,37 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
...
+{
+    char *joined = NULL;                /* pointer to joined string w/sep */
```
jangorecki, line 382: Please use 2 whitespace indentation, unless rest of the file you are editing uses different indentation.

jangorecki (commented):
Thanks for PR. Some minor feedback left, haven't really looked at the logic.

src/utils.c
logworthy: Makes sense.  I've renamed to `concatCharVec`.

src/rbindlist.c
logworthy: Good catch!  I've taken another shot at it, copying to a fixed-size buffer before the error.

src/utils.c
```diff
@@ -374,37 +374,39 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
   return(ans);
 }
 
-// Convert a character vector into a single string, e.g. for printing error messages
+// Concatenate a character vector into a single string, e.g. for printing error messages
 // adapted from https://stackoverflow.com/a/58163237
 // make sure to *free* the returned string after use
-char *joinCharVec (SEXP x, const char *sep)
+// # nocov start
```
jangorecki, line 380: why whole function is escaped from nocov? aren't we able emit error that uses this function in a reproducible way?

logworthy:
@jangorecki, @MichaelChirico thanks for the quick feedback!  Please let me know if there's anything else.  I've converted back to draft since I still need to update the translation files - will do those next if everything else is looking good.

MichaelChirico:
don't bother with the translations part, we handle that in one go before release.

src/rbindlist.c
```diff
@@ -310,10 +310,21 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       if (firsti==-1) { firsti=i; firstw=w; firstCol=thisCol; }
       else {
         if (!factor && !int64) {
-          if (!R_compute_identical(PROTECT(getAttrib(thisCol, R_ClassSymbol)),
-                                   PROTECT(getAttrib(firstCol, R_ClassSymbol)),
-                                   0)) {
-            error(_("Class attribute on column %d of item %d does not match with column %d of item %d."), w+1, i+1, firstw+1, firsti+1);
+          SEXP thisClass = PROTECT(getAttrib(thisCol, R_ClassSymbol));
```
MichaelChirico, line 313: @shrektan can you comment on if we need to do anything with encoding here? class can potentially be non-UTF8 right?

src/rbindlist.c
```diff
@@ -310,10 +310,21 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+          if (!R_compute_identical(thisClass, firstClass, 0)) {
+            #define BUFSIZE 1024 // 'error' function length limit in R's errors.c is 8192
+            char thisClassBuf[BUFSIZE];
+            char* thisClassStr = concatCharVec(thisClass, ", ");
+            strncpy(thisClassBuf, thisClassStr, BUFSIZE);
+            thisClassBuf[BUFSIZE-1] = '\0';
+            free(thisClassStr);
+            char firstClassBuf[BUFSIZE];
+            char* firstClassStr = concatCharVec(firstClass, ", ");
+            strncpy(firstClassBuf, firstClassStr, BUFSIZE);
```
MichaelChirico, line 324: `base` `do_paste` uses `strcpy`, is that more appropriate? In particular I think `strcpy` handles the terminal `'\0'` right? I am not an expert on the difference.

logworthy: `strcpy` will attempt to copy the source string in its entirety, even if the destination array is too small.  `strncpy` adds a size argument and will stop copying when that size limit is reached.  If the string terminates before the size limit is reached, then the `'\0'` byte will be copied, otherwise it needs to be added manually.

Since we don't know how large the string returned by `concatCharVec` will be, this seemed like the right thing to do.  An alternative option would be to add a `max_size` argument to `concatCharVec` to limit the size of the output.

MichaelChirico: thanks for the explanation, makes sense. I don't think a new argument is necessary.

src/rbindlist.c
shrektan: In theroy, yes... and if not native encoded, the message would become garbage letters (the errors and warnings API of R expects native encoding)... However, I don't think anyone would actually use a non-ASCII string as the class name... 

Anyway, if this is needed, we should add the translation (making the native encoded string) in `concatCharVec()` . I mean, adding an extra argument `encoding` ( 0: do nothing 1: to utf-8 2: to native encoding ) there.

MichaelChirico: OK, as long as it won't crash R, I think it's fine to leave that for future users to report :smiley: 

My guess is "ASCII in class name" is a broader assumption in the source code, can't think of any related issues I've seen.

logworthy: For what it's worth, I played around with this, but it didn't quite work the way I expected.
image
  

Using `CHAR()` or `Rf_translateCharUTF8()` gives this result:
image

Using `Rf_translateChar()` or `Rf_translateChar0()` gives this:
image

If I try to write a test it fails either way:
image
image


Simple errors don't seem to display correctly either:
image

🤷‍♂️

src/rbindlist.c
```diff
@@ -310,10 +310,21 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
-          if (!R_compute_identical(PROTECT(getAttrib(thisCol, R_ClassSymbol)),
-                                   PROTECT(getAttrib(firstCol, R_ClassSymbol)),
-                                   0)) {
-            error(_("Class attribute on column %d of item %d does not match with column %d of item %d."), w+1, i+1, firstw+1, firsti+1);
+          SEXP thisClass = PROTECT(getAttrib(thisCol, R_ClassSymbol));
+          SEXP firstClass = PROTECT(getAttrib(firstCol, R_ClassSymbol));
+          if (!R_compute_identical(thisClass, firstClass, 0)) {
+            #define BUFSIZE 1024 // 'error' function length limit in R's errors.c is 8192
+            char thisClassBuf[BUFSIZE];
+            char* thisClassStr = concatCharVec(thisClass, ", ");
```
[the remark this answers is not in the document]
logworthy: We could add a `length` argument, but on its own I don't think that wouldn't make the code more concise.

The real issue here is the dynamic allocation of the array within `concatCharVec()`.  I think that memory needs to be freed, but the array also needs to be passed to the `error()` function, and I'm not sure if there is a way to free it after `error()` has been called.  

Hence we're copying the values to the intermediate statically allocated arrays before the dynamic arrays are freed, and then passing those static arrays to `error()`.

I'm open to feedback if there is a better way to handle this.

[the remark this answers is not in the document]
shrektan: I think return a SEXP object should solve your concerns as SEXP object could be freed by R.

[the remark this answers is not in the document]
logworthy: Ooh, that makes things much simpler!  Thanks :)

logworthy:
> I think this needs to be `UNPROTECT(1)`

What needs to be `UNPROTECT(1)`?

src/utils.c
```diff
@@ -374,3 +374,38 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
...
+      strcpy (concatenated + sz, sep);    /* copy separator */
+      sz += lensep;                 /* update stored size */
+    }
+    strcpy (concatenated + sz, Rf_translateChar(STRING_ELT(x, i)));     /* copy string to concatenated */
+    first = 0;                      /* unset first flag */
+    sz += len;                      /* update stored size */
+  }
+
+  SEXP concatenatedCharVec = PROTECT(mkChar(concatenated));
+  free(concatenated);
```
MichaelChirico, line 409: sorry I think the comment got misplaced from mobile.

This `free` applies to the `SEXP` right? if so then it needs to be `UNPROTECT`ed

logworthy: Oh, okay.  No, it's actually being called on a `char*` which is dynamically allocated and used only within `concatCharVec()`.

mattdowle (changes requested):
Great idea and improvement.
There is already `concat` in `fmelt.c`. Can that be enhanced and reused for this usage? The main enhancement needed being able to be called twice from one error/warning message. It constrains the maximum length of the string which is needed, and this PR doesn't do currently. When its `idx` is `NULL`, rather than error as it does now, it could concat all the items in order as needed in this PR.
Or, if it were me, I would have just included the first item of the class vector in the message; no concat would be needed. In this example, `POSIXct` != `Date` is already clear and helpful to the user. They don't need to know that POSIXct inherits from POSIXt too. I can only think of esoteric constructed examples where the message might state that myClass != myClass (and thus be confusing to the user) because myClass somehow didn't inherit the same way.
If you decide to include all the classes in the message and reuse `fmelt.c:concat`, its `static` could be removed and its signature added to `data.table.h` so it could be used here.
Also please remove the translation changes from PR as Michael said it will be done as part of release procedures in one shot.  Or, since it has already been translated, maybe @MichaelChirico  is ok to leave the translation in this PR this time? Up to him on that.

MichaelChirico:
yes, please revert the translations since unfortunately they're incorrect 😄 -- we'll let a native speaker handle updating them (and other new/changed messages) in one go before release. thanks!

MichaelChirico:
Hi @logworthy would you like to resume this PR? Please take note of related PRs like #5874 -- is the functionality here still needed given recent enhancements?

cc @ben-schwen again.

ben-schwen:
Definitely an improvement, but would go with Matt's comment and only return the first item of class vector, which makes it much simpler.

It won't interfere with my recent PRs since they try to solve the problem and not just improve the error message.

MichaelChirico:
Thanks... marking as draft until #5570 is merged, unless someone wants to rebase this PR relative to that one first.

PR #4829, opened 2020-12-03
Fix segfaults when assigning all-NA-vector to factor column

### LINKED ISSUE #4824: set function crashes the R session
A very bad usage of `set` leads to the systematic crash of the Rsession. 

`#` `minimal reproductible example` :
```
library(data.table)
n<-1e6
dt<-
  data.table(
    fact=factor(rep("A",n)),
    case=rep(TRUE,n),
    coeff=rep(5,n))
x<-"fact"
 set(
    dt,
    i= which(dt$case==TRUE),
    j= "fact",
    value=dt[(case),..x]*dt$coeff
    )
```
For the crash to happen :
- n should be big (n=1e4 : no crash, n=1e5 : crash)
- `dt$fact` needs to be a `factor`
- argument `i` needs to be used in `set`
- argument `j` needs to be exactly "fact" (needs to overwrite column "fact")
- one should use `..x` and not "fact" in `dt[(case),..x]`

`#` `Output of sessionInfo()`
> sessionInfo()
R version 4.0.0 (2020-04-24)
Running under: Windows 10 x64 (build 18362)Matrix products: defaultlocale:
[1] LC_COLLATE=French_France.1252  LC_CTYPE=French_France.1252
[3] LC_MONETARY=French_France.1252 LC_NUMERIC=C
[5] LC_TIME=French_France.1252attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   baseother attached packages:
[1] data.table_1.13.0loaded via a namespace (and not attached):
[1] compiler_4.0.0 tools_4.0.0

shrektan:
Closes #4824 

```r
library(data.table)
dt = data.table(FACTOR = factor(rep("a", 100L))) 
set(dt, i = 1:99, j = "FACTOR", value = rep(NA, 99L)) 
dt
str(dt)
```

Before the PR, `print(dt)` leads to an error that `Error in as.character.factor(x) : malformed factor`. By inspecting the value of `dt$FACTOR`, you should find that some of the values are garbaged. 

This PR fixes the issue.


mattdowle (approved):
Great fix! Thanks for chasing it down.

Issue #4832, opened 2020-12-03
1.13.4 release checks


mattdowle:
Since the changes in this release were minimal I went for an accelerated approach: submitting to CRAN early since CRAN does revdep checking. I haven't received those results yet.
In the meantime I continued with local revdep testing, and a full refresh of my revdeplib, using R-release again instead of R-devel, for example (see commits to .dev).
Has now completed and there's just one impact (`akc`) due to `as.Date.IDate` no longer exported, news note 2.
```
CRAN:
 ERROR   :    4 : akc lava reproducible wyz.code.metaTesting 
 WARNING :    2 : bRacatus crfsuite 
 NOTE    :  330 
 OK      :  632 
 TOTAL   :  968 / 968 
```

[fail.log](https://github.com/Rdatatable/data.table/files/5639866/fail.log)

- [x] `akc`  due to `as.Date.IDate` no longer exported. PR submitted https://github.com/hope-data-science/akc/pull/4
- [x] `lava`  not due to data.table update as it occurs for me with 1.13.2 too https://github.com/kkholst/lava/issues/10
- [x] `reproducible`  same error `object 'split_chain' not found` on CRAN
- [x] `wyz.code.metaTesting`  same error `3L not equal to 2` as on CRAN unrelated to data.table, no change since https://github.com/Rdatatable/data.table/issues/4758
- [x] `bRacatus`  Same CRS warning occurs on CRAN so that's not due to data.table update https://github.com/EduardoArle/bRacatus/issues/3
- [x] `crfsuite`  not data.table related. gcc 10.2.0 locally appears to reveal a `-Wstringop-overflow` https://github.com/bnosac/crfsuite/issues/14

eddelbuettel:
AFAICT you may not get rev.dep results because the package is already in state 'archived' which to the best of my knowledge means 'dead' whereas rev.dep results only come at the subsequent step(s).

```sh
edd@rob:~$ cranIncoming.r data.table        # simple littler wrapper for a foghorn function
    package version cran_folder                time    size
 data.table  1.13.4     archive 2020-12-01 19:15:00 5192530
edd@rob:~$ 
```

mattdowle:
[...]

PR #4833, opened 2020-12-05
urls

mattdowle:
For 1.13.4 #4832 
* two moved-to articles
* removed http: reference that has no https:. It contained advice about base R read.table as background info, so was easier to remove than fight with CRAN maintainers about http addresses.
* Removed http://r-datatable.com from everywhere (other than already-translated string messages) that CRAN maintainers look at and require https. DomainMonster only offer forwarding for http not https.


mattdowle:
The fails are to do with latest R-devel not this PR. Proceeding to merge.

PR #4834, opened 2020-12-05
amend for change in R-devel

jangorecki:
fixes tests failures on R-devel
merge only when this pipelines passes: https://gitlab.com/jangorecki/data.table/-/pipelines/225679419


jangorecki:
superseded by #4835 that has better description

PR #4835, opened 2020-12-05
cater for latest change in R-devel today which breaks test 2022

mattdowle:
In tests 2022.[1-4], the function value assigned has an environment attached. That environment isn't present in the RHS y value constructed by the `data.table()` call. R-devel's `all.equal` is now checking environments, on by default. More detail in https://github.com/wch/r-source/commit/4329ff7d48dcb745028b5cc1c651e373b4e7258c by @mmaechler.

Adding `check.environments=FALSE` to the `all.equal()` call in `test()` gets passed through to `all.equal.function` and fixes it. Since `all.equal` takes an `...` input, passing the new argument in old versions of R still works.

Test 2022 ...
```
d = data.table(id=c("a","b"), f=list(function(x) x*2, function(x) x^2), key="id")
d[.("a"), f:=function(x)x^3]
y = data.table(id=c("a","b"), f=list(function(x) x^3, function(x) x^2), key="id")
d$f[[1]]
# function(x)x^3
# <environment: 0x560c268b6118>
y$f[[1]]
# function(x) x^3

# before the change in R-devel
all.equal(d$f[[1]], y$f[[1]])
# [1] TRUE

# after the change in R-devel
all.equal(d$f[[1]], y$f[[1]])
# [1] "Names: 2 string mismatches"                                  
#  [2] "Length mismatch: comparison on first 2 components"           
#  [3] "Component 1: Modes: numeric, list"                           
#  [4] "Component 1: Lengths: 1, 2"                                  
#  [5] "Component 1: names for current but not for target"           
#  [6] "Component 1: Attributes: < target is NULL, current is list >"
#  [7] "Component 1: target is numeric, current is data.table"       
#  [8] "Component 2: Modes: numeric, list"                           
#  [9] "Component 2: Lengths: 1, 2"                                  
# [10] "Component 2: names for current but not for target"           
# [11] "Component 2: Attributes: < target is NULL, current is list >"
# [12] "Component 2: target is numeric, current is data.table"       

all.equal(d$f[[1]], y$f[[1]], check.environments=FALSE)
# [1] TRUE
```

### DIFF
--- a/.dev/CRAN_Release.cmd
+++ b/.dev/CRAN_Release.cmd
@@ -301,7 +301,7 @@ tar xvf R-devel.tar.gz
 mv R-devel R-devel-strict-clang
 tar xvf R-devel.tar.gz
 
-cd R-devel  # used for revdep testing: .dev/revdep.R.
+cd R-devel  # may be used for revdep testing: .dev/revdep.R.
 # important to change directory name before building not after because the path is baked into the build, iiuc
 ./configure CFLAGS="-O2 -Wall -pedantic"
 make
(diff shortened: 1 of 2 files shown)


mattdowle:
included in v1.13.4 #4832

jangorecki:
Argument name just changed in R-devel
https://github.com/wch/r-source/commit/e29743e44114b413ff0903406179f80fdd9e9764

PR #4839, opened 2020-12-08
don't use base::order on data.frame in test as that's now an error in R-devel

### LINKED ISSUE #4838: xtfrm change in r-devel today breaks on data.frame
Email to Kurt copied here.

---

Dear Kurt,

I can't find you on GitHub, but this email is copied here so that everyone in data.table team knows the status: https://github.com/Rdatatable/data.table/issues/4838.
Your change to R-devel today appears to break two data.table tests (1319.1 & 1319.2).
https://github.com/wch/r-source/commit/8696236cdc7a2be91605f048164e23430aabf0b1
Here is an example using data.frame.  Did you intend for this to break?
This will affect the checks of the data.table submission which is waiting for you to manually accept the 3 spurious not-found URLs.
```
> xtfrm(data.frame(A=8:5))   # before the change
[1] 4 3 2 1
>

> xtfrm(data.frame(A=8:5))   # after the change
Error in xtfrm.default(data.frame(A = 8:5)) : cannot xtfrm 'x'
```

Best, Matt

---

mattdowle:
Closes #4838


mattdowle:
The missed line of coverage is strange to occur given the change in this PR.
image
I'm going to let this go this time. For many reasons.  i) it's not forgetting it complete because the missed line of coverage goes into the pool of overall missed coverage which we can get to another time,  ii) I've looked at the missed line and in my opinion is a safe one to let go this time; in fact it seems spurious,  iii) we've lost so much time already on this submission which solves many other more pressing issues on CRAN.  
Proceeding.

PR #4843, opened 2020-12-10
new helper function fctr to retain order of input in levels, closes #…

### LINKED ISSUE #4837: helper function: convert char to factor retaining order
I would like to propose to add and export new helper function for constructing factor from character, like base R but retaining original order.
This can be useful when doing plots, `dcast`, and many other cases where we want to retain order of elements in a vector. More use cases in linked PR.
I am currently using the following helper:
```r
fctr = function(x) factor(x, levels=unique(x))
```
I would like to propose to have it in data.table, eventually improving implementation by using `forder` to obtain unique values.

jangorecki:
closes #4837


jangorecki:
In fabafed7671284c7c51473a9e6f3cc69df7264db I added extra argument.
When doing plots with lattice in a grid layout, plots are being ordered by levels starting from bottom left corner. This is good for leff-to-right order of plots horizontally, yet for a vertical order of plots bottom-to-top might not be expected, and then levels needs to be reversed for plots to be produced in top-to-bottom order. This new argument will make that more convenient.
An example of such plotting can be found [in this report](https://github.com/h2oai/db-benchmark/blob/5638cd83ed29ac4a6e0114ea5b10f99abdf72905/_report/history.Rmd) and generated below plot. Having original order of levels "basic" would be at the bottom row in layout, while "advanced" at the top.
datatable groupby 1e9

R/utils.R
```diff
@@ -166,3 +166,10 @@ rss = function() {  #5515 #5517
   # nocov end
 }
 
+# convert char to factor retaining order #4837
+fctr = function(x, levels=unique(x), ..., rev=FALSE) {
```
jangorecki, line 170: We could push performance a bit by replacing `base::unique` with `funique`.
```r
funique = function(x, copy=FALSE) {
  o = forderv(x, sort=TRUE, retGrp=TRUE)
  if (attr(o, "maxgrpn", TRUE) <= 1L) { ## all unique
    if (copy) x = copy(x)
    return(x)
  }
  f = attr(o, "starts", exact=TRUE)
  if (length(o)) f = o[f]
  if (length(o <- forderv(f))) f = f[o] ## this rolls back to original order
  x[f]
}
```
but that adds a bit more code to maintain, and `fctr` is not kind of function to be commonly used in time critical paths. You call it once. Base `unique` is already fast, on 1e8 values we can push from 12s -> 6s, which is nice, but as said, don't want to introduce extra code to maintain but rather a convenience function on top of base R.

jangorecki:
Needs unit test for #5327 so it can close that one as well

MichaelChirico (commented):
Searching around on `cran`, I _do_ see calls like `factor(x, levels=unique(x))` pretty commonly:

https://github.com/search?q=org%3Acran%20%2Ffactor%5C(.*unique%5C(%2F&type=code

As well as more generally in `lang:R`:

https://github.com/search?q=lang%3AR+%2Ffactor%5C%28.*unique%5C%28%2F&type=code

I think that justifies a new function even though it's just a base wrapper.

We also see ample `rev=TRUE` cases.

I only quibble about the name, there's a tension between "its good to have something concise" and "readers should be able to infer something about it from its name".

Maybe `u_factor` / `factor_u` (**u**nsorted `factor()` / **u**nique `factor()`)? `uns_factor` / `unq_factor`?

Alternatively, maybe this is something to propose to `r-devel` first?

jangorecki:
> I only quibble about the name, there's a tension between "its good to have something concise" and "readers should be able to infer something about it from its name".

`fctr` is just like `factor` but without vowels, therefore should be very easy to infer

> Alternatively, maybe this is something to propose to r-devel first?

I doubt they would consider adding that

MichaelChirico:
> `fctr` is just like `factor` but without vowels, therefore should be very easy to infer

we can guess it's related to factors, but not anything else. is it fast alternative (as is most common for our helpers)? is it a related-but-similar S3 class?

> I doubt they would consider adding that

not as a separate function, surely, but _possibly_ as new argument(s) to factor()

jangorecki:
>  is it fast alternative (as is most common for our helpers)?

not really, therefore no **f** prefix :)

>  is it a related-but-similar S3 class?

`factor()` is not S3 class, so less likely users will think that way

MichaelChirico:
> factor() is not S3 class, so less likely users will think that way

really?

```r
is.object(factor())
# [1] TRUE

methods(class = 'factor')
#  [1] all.equal     as.character  as.data.frame as.Date       as.list       as.logical    as.POSIXlt    as.vector     c             coerce        droplevels    [<-          
# [13] [             [[<-          [[            format        initialize    is.na<-       length<-      levels<-      Math          Ops           plot          print        
# [25] relevel       relist        rep           show          slotsFromS3   summary       Summary       xtfrm        
# see '?methods' for accessing help and source code
```

MichaelChirico:
I would like more feedback from other @Rdatatable/committers here on whether (1) we want to export a simple {base} wrapper (2) the proposed naming

The closest I can see in terms of "nearness" to base R functionality would be `like()` which is typically invoked via `%like%` and friends; even then it builds in the faster implementation for factor input that's not present in older versions of R (for now: #7104). We could consider unexporting `like()` but we probably wouldn't.

TysonStanley:
Just my opinion: 1) it would have nearly no need for maintenance going forward so if it is useful for Jan and others, seems like a low lift to add. 2) I'm fine with the name. `fctr` is similar to the `fct_` prefix in `forcats` so there's obviously similar notation. Don't think the name is informative but also don't know how important that is given it's not doing a lot of special things.

MichaelChirico:
Jan WDYT about extending this to support the "normal" factor() default of sorted levels? that way users can have a one-stop shop instead of needing to switch between the proposed fctr() and factor()

jangorecki:
> > factor() is not S3 class, so less likely users will think that way
> 
> really?
> 
> ```r
> is.object(factor())
> # [1] TRUE
> 
> methods(class = 'factor')
> #  [1] all.equal     as.character  as.data.frame as.Date       as.list       as.logical    as.POSIXlt    as.vector     c             coerce        droplevels    [<-          
> # [13] [             [[<-          [[            format        initialize    is.na<-       length<-      levels<-      Math          Ops           plot          print        
> # [25] relevel       relist        rep           show          slotsFromS3   summary       Summary       xtfrm        
> # see '?methods' for accessing help and source code
> ```

uh sorry, I meant `factor` is not a S3 method... you are absolutely right

jangorecki:
by

> the "normal" factor() default of sorted levels?

you mean

`levels=sort(unique(x))` instead of `levels=unique(x)`

?

If so then I would add another arg `sort=FALSE`, because otherewise retaining original order, as was the purpose of this helper, will no longer be easier with `fctr()` than with `factor()`.

MichaelChirico:
> by
> 
> > the "normal" factor() default of sorted levels?
> 
> you mean
> 
> `levels=sort(unique(x))` instead of `levels=unique(x)`
> 
> ?
> 
> If so then I would add another arg `sort=FALSE`, because otherewise retaining original order, as was the purpose of this helper, will no longer be easier with `fctr()` than with `factor()`.

yes that's what I had in mind. basically fctr() becomes extended factor() with a different default.

Issue #3872, opened 2019-09-14
Handling of zlib fwrite dependency


MichaelChirico:
`#include <zlib.h>` might fail on some machines. In this case it would be helpful to fail more helpfully.

Currently on `rocker/r-ver:3.5.3` (e.g.), we get the following installation failure for dev `data.table`:

```
install.packages('data.table', type = 'source', repos='http://Rdatatable.github.io/data.table')
Installing package into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
trying URL 'http://Rdatatable.github.io/data.table/src/contrib/data.table_1.12.3.tar.gz'
Content type 'application/gzip' length 4960958 bytes (4.7 MB)
==================================================
downloaded 4.7 MB

* installing *source* package ‘data.table’ ...
** libs
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fopenmp -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c assign.c -o assign.o
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fopenmp -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c between.c -o between.o
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fopenmp -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c bmerge.c -o bmerge.o
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fopenmp -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c chmatch.c -o chmatch.o
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fopenmp -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c cj.c -o cj.o
[...]

PR #4844, opened 2020-12-12
makes dependency on zlib optional

jangorecki:
Follow-up of #3872
It is NOT meant to solve Solaris issue, but general issue when zlib is not available.

---

before PR
```
docker run -it rocker/r-ver:3.5.3 Rscript -e 'install.packages("data.table", repos="https://Rdatatable.gitlab.io/data.table"); data.table::fwrite(iris, "iris.gz")'
```
```
*** pkg-config is not installed.
*** Compilation will now be attempted and if it works you can ignore this message. In
*** particular, this should be the case on Mac where zlib is built in.
*** However, if compilation fails, try 'locate zlib.h zconf.h' and ensure the zlib
*** development library is installed :
***   deb: zlib1g-dev (Debian, Ubuntu, ...)
***   rpm: zlib-devel (Fedora, EPEL, ...)
***   There is a zlib in brew for OSX but the built in zlib should work.
*** Note that zlib is required to compile R itself so you may find the advice in the R-admin
*** guide helpful regarding zlib. On Debian/Ubuntu, zlib1g-dev is a dependency of r-base as
*** shown by 'apt-cache showsrc r-base | grep ^Build-Depends | grep zlib', and therefore
*** 'sudo apt-get build-dep r-base' should be sufficient too.
*** To silence this message, please ensure that :
***   1) 'pkg-config --exists zlib' succeeds (i.e. $? -eq 0)
***   2) 'pkg-config --libs zlib' contains -lz
*** Compilation will now be attempted ...
...
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fopenmp -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=f
ormat-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c fwrite.c -o fwrite.o
fwrite.c:10:49: fatal error: zlib.h: No such file or directory
 #include <zlib.h>      // for compression to .gz
```

---

after PR
```
docker run -it rocker/r-ver:3.5.3 Rscript -e 'install.packages("remotes", repos="https://cloud.r-project.org"); remotes::install_github("Rdatatable/data.table@nozlib"); data.table::fwrite(iris, "iris.gz")'
```
```
*** pkg-config is not installed.
*** Compilation will now be attempted and if it works you can ignore this message. In
*** particular, this should be the case on Mac where zlib is built in or pkg-config
*** is not installed. However, if compilation fails, try 'locate zlib.h zconf.h' and
*** ensure the zlib development library is installed :
***   deb: zlib1g-dev (Debian, Ubuntu, ...)
***   rpm: zlib-devel (Fedora, EPEL, ...)
***   There is a zlib in brew for OSX but the built in zlib should work.
*** Note that zlib is required to compile R itself so you may find the advice in the R-admin
*** guide helpful regarding zlib. On Debian/Ubuntu, zlib1g-dev is a dependency of r-base as
*** shown by 'apt-cache showsrc r-base | grep ^Build-Depends | grep zlib', and therefore
*** 'sudo apt-get build-dep r-base' should be sufficient too.
*** To silence this message, please ensure that :
***   1) 'pkg-config --exists zlib' succeeds (i.e. $? -eq 0)
***   2) 'pkg-config --libs zlib' contains -lz
*** Compilation will now be attempted ...
*** Compilation without compression support in fwrite
...
Error in data.table::fwrite(iris, "iris.gz") : 
  Compression in fwrite uses zlib library which was not installed at the time when data.table was compiled. To enable compression support in fwrite one must install zlib and re-install data.table.
```


configure
```diff
@@ -46,9 +48,9 @@ fi
 
 if [ $msg -ne 0 ]; then
   echo "*** Compilation will now be attempted and if it works you can ignore this message. In"
-  echo "*** particular, this should be the case on Mac where zlib is built in."
-  echo "*** However, if compilation fails, try 'locate zlib.h zconf.h' and ensure the zlib"
-  echo "*** development library is installed :"
+  echo "*** particular, this should be the case on Mac where zlib is built in or pkg-config"
+  echo "*** is not installed. However, if compilation fails, try 'locate zlib.h zconf.h' and"
+  echo "*** ensure the zlib development library is installed :"
   echo "***   deb: zlib1g-dev (Debian, Ubuntu, ...)"
```
MichaelChirico, line 54: doesn't this turn off zlib support on mac by default

jangorecki: If `pkg-config` is not available on Mac then yes, it turns off. Needs testing anyway.
There is a related comment above:
>  if pkg-config is not available then zlib will be disabled for higher chance of compilation success

jangorecki: Note that branch was hit before when pkg-config was not installed, just this console message was not mentioning that.

jangorecki: We could eventually try to check if zlib is available by attempting to compile dummy code, like we do for openmp. Docker image in your report didn't have either pkg-config and zlib. So then we could still compile with zlib support. For now I don't think it is worth. Messages are clear and points users in the proper direction. Installing pkg-config and/or zlib.

jangorecki:
I put it on label High so we can have it merged at the beginning on a development cycle to be well tested before it will land on CRAN.

mattdowle:
Changes I made in 08a2584:
* zlib runtime is still a requirement for those installing the binary package from CRAN. So reinstated the SystemRequirements line in DESCRIPTION. Not a problem because the zlib runtime is already required by R.  Further, `test.data.table()` will fail, and therefore so will `R CMD check`, if data.table has not been compiled with zlib headers because tests of `fwrite` compression will fail.
* `#define z_stream" empty line removed. I'm not sure that would work. I checked all uses of `z_stream` are inside `#ifndef NOZLIB blocks` blocks.
* distinction between library and headers made in the messages
* the error message now tells the user that reinstalling will output further guidance

PR #4850, opened 2020-12-16
extra print in test summary

jangorecki:
Output I am getting after first commit in this PR.
```
All 8259 tests (last 1913.24) in tests/tests.Rraw completed ok in 57.3s elapsed (00:01:12 cpu)
```
and after fixing the problem, in second commit.
```
All 9993 tests (last 2161.2) in inst/tests/tests.Rraw completed ok in 00:01:32 elapsed (00:01:49 cpu)
```

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -177,7 +177,7 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
   cat("10 longest running tests took ", as.integer(tt<-DT[, sum(time)]), "s (", as.integer(100*tt/(ss<-timings[,sum(time)])), "% of ", as.integer(ss), "s)\n", sep="")
   print(DT, class=FALSE)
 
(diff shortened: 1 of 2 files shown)


mattdowle:
> Output I am getting after first commit in this PR.

The first commit, which appears to be https://github.com/Rdatatable/data.table/pull/4850/commits/108404190da90003f29057ff69e665c8355b0717, appears to just change the type of quotes within a comment.  Because that change is in a comment, I don't see why that would change anything.   

> and after fixing the problem

what problem exactly?  I'm looking at the change to a comment and wondering why that changes anything.

mattdowle:
I don't see any reason why not to merge this PR, though.  In this case i don't need to understand the reason. Including `env$prevtest` in the last line is fine, whatever the reason.
Will merge after version bump.

jangorecki:
@mattdowle first commit was cb6cae9. Second commit was 1084041 that changed quotes in comment making test script to not quite when parsing.

PR #3763, opened 2019-08-13
Closes #3633 -- restore reshape2 redirection (with warning)

MichaelChirico:
#3633 

I looked at the `reshape2` `list` method for `melt`, it's quite unintuitive to me, so I wouldn't re-implement it as-is to `data.table`. We might want to add `melt.list` in the future but I don't think it would look like `reshape2`'s (which relies on `attr` of the `list` at various levels?)

Issue #3633, opened 2019-06-06
Handling how data.table::melt is not passing to reshape2::melt.list method anymore.


ProfFancyPants:
So a lot of my old code is breaking. This isn't critical but I wanted to bring it up here before I start making changes willynilly. 

By continuing to file this new issue / feature request, I confirm I have :
1. searched the news.

This is what the news says about the issue:

> [NEWS.0](https://github.com/Rdatatable/data.table/blob/d3ccd8139d3d592201e0f085c5f37a4fc5a426e3/NEWS.0.md)
> [L605](https://github.com/Rdatatable/data.table/blob/d3ccd8139d3d592201e0f085c5f37a4fc5a426e3/NEWS.0.md#L605)
> 4. `data.table` redefines `melt` generic and *suggests* `reshape2` instead of *import*. As a result we don't have to load `reshape2` package to use `melt.data.table` anymore. The reason for this change is that `data.table` requires R >=2.14, whereas `reshape2` R v3.0.0+. Reshape2's melt methods can be used without any issues by loading the package normally.
> [L609](https://github.com/Rdatatable/data.table/blob/d3ccd8139d3d592201e0f085c5f37a4fc5a426e3/NEWS.0.md#L609)
>  6. Because `reshape2` requires R >3.0.0, and `data.table` works with R >= 2.14.1, we can not import `reshape2` anymore. Therefore we define a `melt` generic and `melt.data.table` method for data.tables and redirect to `reshape2`'s `melt` for other objects. This is to ensure that existing code works fine.

## Minimal reproducible example

```
SCRIPT_LIST <- list(
  W3_ENGLISH = list(
    ## _W3_ENGLISH ####
    LANGUAGE = "English",
    NCOL_TXT = 1428L,
    NCHAR_LINE1 = 8366L,
    FILE_NAME_RE = "^[Ww][Jj][^Ee]",
    D_MACRO = "text"
  ),
  W3_SPANISH = list(
    ## _W3_SPANISH ####
    LANGUAGE = "Spanish",
    NCOL_TXT = 228L,
    NCHAR_LINE1 = 1363L,
    FILE_NAME_RE = "^[Ww][Mm][^Ee]",
    D_MACRO = "text"
  ),
  W4_ECAD = list(
    ## _W4_ECAD ####
    LANGUAGE = "English or Spanish",
    NCOL_TXT = 188L,
    NCHAR_LINE1 = 1780L,
[...]

PR #4864, opened 2021-01-02
Search registered S3 method for melt before redirecting to reshape2

odelmarcelle:
Follow-up of #3633 and #3763

The condition leading to `UseMethod` in the `melt` generic is not true for non-`data.table` object. Hence, S3 methods for `melt` defined by other packages may be skipped.

I initially suggested changing that condition in order to search if a S3 method exists for the current object before redirecting to `reshape2`.
After giving it a second thought, it appears more correct to performs the redirection to `reshape2` in a `default` method. This leaves the generic clean and allows registering new methods.

Since there were no tests on the generic `melt`, I left the test file as-is. 

After the change, using S3 method dispatch on a custom object works as intended:
```
library(data.table)
# Creating a S3 object and it's melt method
S3Object = list(dt1 = data.table(id = 1:2, a = 3L), dt2 = data.table(id = 1:2, b = 4:5))
class(S3Object) = "S3Class"
new_melt = function(data, ..., na.rm = FALSE, value.name = "value") {
  melt(Reduce(merge, data), id.vars = "id")
}
.S3method("melt", "S3Class", new_melt)

melt(S3Object) ## Working fine
```


R/fmelt.R
```diff
@@ -3,19 +3,19 @@
...
-    ns$melt(data, ..., na.rm=na.rm, value.name=value.name)
-  }
+  data_name = deparse(substitute(data))
+  ns = tryCatch(getNamespace("reshape2"), error=function(e)
+    stop("The melt generic in data.table has been passed a ", class(data)[1L],", but data.table::melt currently only has a method for data.tables. Please confirm your input is a data.table, with setDT(", data_name, ") or as.data.table(", data_name, "). If you intend to use a method from reshape2, try installing that package first, but do note that reshape2 is deprecated and you should be migrating your code away from using it."))
+  warning("The melt generic in data.table has been passed a ", class(data)[1L], " and will attempt to redirect to the relevant reshape2 method; please note that reshape2 is deprecated, and this redirection is now deprecated as well. To continue using melt methods from reshape2 while both libraries are attached, e.g. melt.list, you can prepend the namespace like reshape2::melt(", data_name, "). In the next version, this warning will become an error.")
```
jangorecki, line 17: I think we are no going into new major version so this code could be simplified for raising error, as mentioned in the warning already.

odelmarcelle: Do you mean stopping the redirection towards `reshape2`?

odelmarcelle:
Updating this again as I still think having melt as a generic method make sense. For example, it could serve to melt an S3 object containing multiple tables and matrices into a single long "melted" format.

I have added a note regarding the change in the NEWS and I added a test on melt.default in test.Rraw.

mattdowle:
Thanks @odelmarcelle. Please let me know if your name is different to your id, for the contributor list displayed on CRAN. Have invited you to be project member too: it should be a button in your GitHub profile or projects page that you need to click to accept.

odelmarcelle:
Thank you for reviewing and merging @mattdowle. Although I'm not sure I deserve to be listed in the contributor list for such a small input, my actual name is different from my Github username.

Could you perhaps correct my name to "Olivier Delmarcelle"? I corrected the DESCRIPTION file on my forked branch, but I'm not sure how to propagate this to the master branch now that this pull request is merged. Sorry for the trouble!

mattdowle:
Thanks @odelmarcelle, done in 76cfe11044ed0f773b1e9f6dd4d9dfab2126f74f.

PR #4881, opened 2021-01-25
https://r-datatable.com now works

mattdowle:
Follow up to #4833 
I moved domain from Domain Monster to Gandi. Thanks to Dirk in the news item, following his suggestion [here](https://github.com/Rdatatable/data.table/issues/4832#issuecomment-738454648).
302 (temporary) vs 301 (permanent) explained in news item.


mattdowle:
Travis CI seems to be stuck (jobs not starting). Proceeding to merge.

PR #4893, opened 2021-02-04
setDF deletes the index attribute

### LINKED ISSUE #4889: setDF should drop indices
[ff2.zip](https://github.com/Rdatatable/data.table/files/5919633/ff2.zip)

Attached is a small zip archive of a saved data.table that demonstrates a problem where sometimes the `[` operator gives misleading results.

```
load("./ff2")   # from the path you extracted the archive to
dim(ff[symbol == "ZM" & date == 20210128])
#  [1] 0 2     <---- wrong results, this subset is not empty
```
One way to show the correct result:
```
> dim(ff[symbol == "ZM"][date == 20210128])
# [1] 98  2
```

Many seemingly no-ops fix the data.table state. One example:
```
ff[, symbol:=symbol]
dim(ff[symbol == "ZM" & date == 20210128])
# [1] 98  2
```

Tested on both Windows and Ubuntu, and data.table versions 1.13.6 and 1.11.8 correspondingly.   (the sessionInfo below is of the windows machine)

`#` `Output of sessionInfo()`
R version 3.6.1 (2019-07-05)
Running under: Windows 10 x64 (build 19042)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:

other attached packages:
[1] data.table_1.13.6

[1] bit_4.0.4      compiler_3.6.1 tools_3.6.1    bit64_4.0.5

OfekShilon:
Closes #4889


MichaelChirico:
change looks great! please also add the following:

 1. NEWS item briefly describing the change & citing yourself for the fix (imitate the other items should be fine)
 2. add yourself to DESCRIPTION
 3. write a test to ensure this behavior doesn't regress later
 4. Denes mentioned a slight amendment to ?setDF as well, could you add that?

OfekShilon:
> change looks great! please also add the following:
> 
> 1. NEWS item briefly describing the change & citing yourself for the fix (imitate the other items should be fine)
> 2. add yourself to DESCRIPTION
> 3. write a test to ensure this behavior doesn't regress later
> 4. Denes mentioned a slight amendment to ?setDF as well, could you add that?

What change is needed to ?setDF ?  Where was it discussed?

ColeMiller1:
The tests are failing due to the external references changes. Those changes seem like a separate PR and outside the scope of removing the index attribute during `setDF`. That doesn't mean they are not worthwhile - it's just the index thing is a no-brainer and needs to be fixed. The external reference needs additional design and review (e.g., the tests are failing!) and maybe there are other ways to deal with it. 

See the comment here https://github.com/Rdatatable/data.table/issues/4889#issuecomment-773395179 for the `?setDF` enhancement. Here's a copy of the suggestion Denes had:

>All data.table attributes including any keys and indices of the input data.table are stripped off.

OfekShilon:
> The tests are failing due to the external references changes. Those changes seem like a separate PR and outside the scope of removing the index attribute during `setDF`. That doesn't mean they are not worthwhile - it's just the index thing is a no-brainer and needs to be fixed. The external reference needs additional design and review (e.g., the tests are failing!) and maybe there are other ways to deal with it.
> 

This made me learn the hard way that any commits I push to the same branch are automatically added to the same PR  :(
I never intended for the other commits to go in. Do I have a way out of this mess or should I delete the PR and start from scratch?

jangorecki:
PR is always based on a branch. If you don't want commits to go into same PR then you have to make new branch for them.
If you are not OK with recent commits you can do `git revert` and `git push`. If you are not sharing a branch to others, then `git reset` and `git push --force` are fine.

OfekShilon:
I can't see why tests are failing. Observe the changed files - I don't change any external references or do anything remotely exotic.  @ColeMiller1?

inst/tests/tests.Rraw
```diff
@@ -17261,3 +17261,11 @@ if (identical(x, enc2native(x))) {
 
 # fintersect now preserves order of first argument like intersect, #4716
 test(2163, fintersect(data.table(x=c("b", "c", "a")), data.table(x=c("a","c")))$x, c("c", "a"))
+
+# setDF now drops index attributes, #4889
+d <- data.table(a=1:100, b=1:100)
+invisible(d[a == 50])
```
MichaelChirico, line 17267: I think the test will be clearer to future readers if we run `setindex(d, a)`

src/utils.c
```diff
@@ -378,11 +378,11 @@ SEXP coerceUtf8IfNeeded(SEXP x) {
 #include <zlib.h>
 #endif
 SEXP dt_zlib_version() {
-  char out[51];
+  char out[71];
 #ifndef NOZLIB
-  snprintf(out, 50, "zlibVersion()==%s ZLIB_VERSION==%s", zlibVersion(), ZLIB_VERSION);
+  snprintf(out, 70, "zlibVersion()==%s ZLIB_VERSION==%s", zlibVersion(), ZLIB_VERSION);
```
MichaelChirico, line 383: is this intended to be here? was this causing an error on your build?

inst/tests/tests.Rraw
OfekShilon: You're right, fixing

src/utils.c
OfekShilon: This particular 50->70 isn't strictly necessary, but the out buffer _was_ too small and I got a build warning about it. Do you prefer I restore this one to 50?

MichaelChirico: I think it's fine to keep -- could you share with us the output, or perhaps make a note in the comments there that on some platforms it takes certain format so it's wider

OfekShilon: Sure. The warning was -
```
utils.c: In function ‘dt_zlib_version’:
utils.c:385:72: warning: ‘zlib header files were not found when data.table was compiled’ directive output truncated writing 61 bytes into a region of size 50 [-Wformat-truncation=]
   snprintf(out, 50, "zlib header files were not found when data.table was compiled");
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
In file included from /usr/include/stdio.h:862:0,
                 from dt_stdio.h:30,
                 from data.table.h:1,
                 from utils.c:1:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:10: note: ‘__builtin_snprintf’ output 62 bytes into a destination of size 50
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
It is generated by the _next_ line, which is >50 chars on all platforms:
```
snprintf(out, 50, "zlib header files were not found when data.table was compiled");
```

MichaelChirico: oh! indeed that buffer is too small. good catch.

MichaelChirico: no need to add a comment even

OfekShilon:
This is a small but rather important fix (`[.data.table` occasionally gives wrong results without it). Anything holding it back from being merged?

OfekShilon:
@MichaelChirico  is there anything I can do to help merge this?

PR #4894, opened 2021-02-06
fread tz= default changed from "" to "UTC"

mattdowle:
Follow up to #4464 

> In that case, shouldn't we plan to set tz='UTC' by default (maybe not this release, but next release?)

Yes, you were right, @MichaelChirico.

At least I did say at the bottom of [this comment](https://github.com/Rdatatable/data.table/pull/4464#issuecomment-658052032) that I wasn't as strongly against tz='UTC' by default and that we could ask the community.

News item contains the detail.

The next release containing this change would be 1.14.0.

Merging now straight away without review. We can see if it passes GLCI, I'll start a rerun of all revdeps just to make sure, and I'll tweet asking community to test dev and raise any objections. We can keep discussing and see if any objections are raised. We can always revert this PR if necessary. Merging is necessary so that folk can test master.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 **Benchmarks are regularly updated: [here](https://h2oai.github.io/db-benchmark/)**
 
-data.table [v1.13.7](https://github.com/Rdatatable/data.table/milestone/20)  (in development)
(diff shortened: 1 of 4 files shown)


mattdowle:
@renkun-ken Hi. Would you mind running your extra-strength go-first please? Wondering if you're ok with this default change.

PR #4895, opened 2021-02-07
use %iscall% in cedta()

MichaelChirico:
This should be a bit safer


jangorecki:
Safer is good but does it add extra overhead?

MichaelChirico:
basically adds one call to `is.call` in the case that (1) we're in a call to something from `utils` and (2) the call stack is at least 8 deep.

I think it's fine to add such a low-overhead call in that case.

PR #4896, opened 2021-02-11
Fix #4784

### LINKED ISSUE #4784: setDT modifies the class of the origin data.frame
Take this code:
```
> d1 <- data.frame(a=c(1,2,3,4), b=c(2,3,4,5))
> d2 <- d1
> setDT(d2)  
```
At this point d2 is a shallow copy of d1, meaning they are not the same R object (although they point to the same columns).
`d2`'s class is data.table, as expected, _but so is `d1`_
```
> class(d2)
[1] "data.table" "data.frame"
> class(d1)
[1] "data.table" "data.frame"
```
Unlike #4783 , I think this is an easy fix. In two places in `setDT`'s source you see the pair of lines -
```
setattr(x, "class", .resetclass(x, "data.frame"))
setalloccol(x)
```
`setalloccol` is what initiates the shallow copy. If the order of the rows were reversed, the class of only the current shallow copy (`d2` in the example above) would be modified, as expected.

`#` `Output of sessionInfo()`
R version 3.6.1 (2019-07-05)
Matrix products: default

OfekShilon:
Closes #4784

Small fix. Essentially just a change in order of two operations.


OfekShilon:
@MichaelChirico  how do you guys prefer to work? Shall I submit this PR (and another one coming) from a clean branch, without the previous PR, and assuming you accept - you'll merge the tests and NEWS?

MichaelChirico:
Thanks for another PR! Unless the fix here depends on the one in #4893 I would try and keep the branches as pristine as possible -- this allows more flexibility on the order when the two are merged. For example, there might be changes requested to one but not the other, so even if they are reviewed by Matt in quick succession, one might still end up merging sooner.

Put another way, ideally each should be ready to merge independently of the other; if one is in fact a dependency of the other, mark the "downstream" one as WIP and make note of the dependency in the PR description. Don't worry about the numbering in the NEWS, Matt will handle that in the final merge to master.

OfekShilon:
> Thanks for another PR! Unless the fix here depends on the one in #4893 I would try and keep the branches as pristine as possible -- this allows more flexibility on the order when the two are merged. For example, there might be changes requested to one but not the other, so even if they are reviewed by Matt in quick succession, one might still end up merging sooner.
> 
> Put another way, ideally each should be ready to merge independently of the other; if one is in fact a dependency of the other, mark the "downstream" one as WIP and make note of the dependency in the PR description. Don't worry about the numbering in the NEWS, Matt will handle that in the final merge to master.

@MichaelChirico  Thanks!  I'll delete this PR and submit a clean replacement soon.

PR #4902, opened 2021-02-13
Marking columns as externally-referenced - WIP

OfekShilon:
This PR is not mergeable yet, it is intended to start a conversation.   I hope to get your feedback on the basic approach: _marking columns with an attribute saying they are also referenced outside the DT, and then use this attribute to bypass `memrecycle`._ 
It solves the immediate issue, but:
(1) Many tests break, as the output of setDT is indeed changed (additional attribute).
(2) I'm not entirely comfortable with this invasive approach.  Perhaps it would be better to contain the change inside the DT itself (adding a list attribute to it), and not add attributes to raw data.  I'm not sure which modifications would be required to make this work - cbind? Elsewhere?  Any thoughts are welcome.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 2. `fintersect()` now retains the order of the first argument as reasonably expected, rather than retaining the order of the second argument, [#4716](https://github.com/Rdatatable/data.table/issues/4716). Thanks to Michel Lang for reporting, and Ben Schwen for the PR.
 
+3. After setDT, some (not all) data modifications leaked to other names bound to the origin data.frame. [#4783](https://github.com/Rdatatable/data.table/issues/4783). Thanks to Ofek Shilon for the report and PR.
+
 ## NOTES
 
 1. Compiling from source no longer requires `zlib` header files to be available, [#4844](https://github.com/Rdatatable/data.table/pull/4844). The output suggests installing `zlib` headers, and how (e.g. `zlib1g-dev` on Ubuntu) as before, but now proceeds with `gzip` compression disabled in `fwrite`. Upon calling `fwrite(DT, "file.csv.gz")` at runtime, an error message suggests to reinstall `data.table` with `zlib` headers available. This does not apply to users on Windows or Mac who install the pre-compiled binary package from CRAN.
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -2637,6 +2637,12 @@ address = function(x) .Call(Caddress, eval(substitute(x), parent.frame()))
   stop('Check that is.data.table(DT) == TRUE. Otherwise, := and `:=`(...) are defined for use in j, once only and in particular ways. See help(":=").')
 }
 
+mark.external.ref.cols <- function(x, val=TRUE) {
+  lapply(names(x), function(nm) {
+    setattr(x[[nm]], "referenced.externally", val)
+  })  
+}
+
 setDF = function(x, rownames=NULL) {
   if (!is.list(x)) stop("setDF only accepts data.table, data.frame or list of equal length as input")
   if (anyDuplicated(rownames)) stop("rownames contains duplicates")
@@ -2653,6 +2659,7 @@ setDF = function(x, rownames=NULL) {
     setattr(x, "class", "data.frame")
     setattr(x, "sorted", NULL)
     setattr(x, ".internal.selfref", NULL)
+    mark.external.ref.cols(x, NULL)
   } else if (is.data.frame(x)) {
     if (!is.null(rownames)) {
       if (length(rownames) != nrow(x))
@@ -2721,6 +2728,7 @@ setDT = function(x, keep.rownames=FALSE, key=NULL, check.names=FALSE) {
     if (check.names) setattr(x, "names", make.names(names(x), unique=TRUE))
     # fix for #1078 and #1128, see .resetclass() for explanation.
     setattr(x, "class", .resetclass(x, 'data.frame'))
+    mark.external.ref.cols(x)
     setalloccol(x)
     if (!is.null(rn)) {
       nm = c(if (is.character(keep.rownames)) keep.rownames[1L] else "rn", names(x))
@@ -2761,6 +2769,7 @@ setDT = function(x, keep.rownames=FALSE, key=NULL, check.names=FALSE) {
     }
     setattr(x,"row.names",.set_row_names(n_range[2L]))
     setattr(x,"class",c("data.table","data.frame"))
+    # mark.external.ref.cols(x)
     setalloccol(x)
   } else {
     stop("Argument 'x' to 'setDT' should be a 'list', 'data.frame' or 'data.table'")
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -17261,3 +17261,10 @@ if (identical(x, enc2native(x))) {
 
 # fintersect now preserves order of first argument like intersect, #4716
 test(2163, fintersect(data.table(x=c("b", "c", "a")), data.table(x=c("a","c")))$x, c("c", "a"))
+
+# Avoid leakage of data changes after setDT to origin DF, #4783
+d1 <- data.frame(a=1)
+d2 <- d1
+setDT(d2)
+d2[!is.na(a), a:=2] # used to leak into d1
+test(2164, d1[["a"]], 2)
--- a/src/assign.c
+++ b/src/assign.c
@@ -503,7 +503,9 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       continue;
     }
 
-    if (coln+1 > oldncol) {  // new column
+    const bool colReferencedExternally = !isNull(getAttrib(thisvalue, sym_referenced_externally)); // Fix #4783
+//    if (coln+1 > oldncol || colReferencedExternally) {  // new column
+   if (coln+1 > oldncol) {  // new column
       SET_VECTOR_ELT(dt, coln, targetcol=allocNAVectorLike(thisvalue, nrow));
       // initialize with NAs for when 'rows' is a subset and it doesn't touch
       // do not try to save the time to NA fill (contiguous branch free assign anyway) since being
--- a/src/data.table.h
+++ b/src/data.table.h
@@ -99,6 +99,7 @@ extern SEXP sym_verbose;
 extern SEXP SelfRefSymbol;
 extern SEXP sym_inherits;
 extern SEXP sym_datatable_locked;
+extern SEXP sym_referenced_externally;
 extern SEXP sym_tzone;
 extern SEXP sym_old_fread_datetime_character;
 extern double NA_INT64_D;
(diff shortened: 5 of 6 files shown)


jangorecki:
I like the idea, although I think we could see if we could utilize R C api ref count. This PR seems to reinvent what is there under NAMED/MAYBE_SHARED flag.

OfekShilon:
@jangorecki That sounds like a good idea, but I don't understand R's ref count and don't trust them ([this discussion in r-devel](https://stat.ethz.ch/pipermail/r-devel/2011-November/062653.html) by Matt Dowle shows it's more complicated than it seems).

This is what I get when I try to inspect it:
```
> a <- 1:100
> .Internal(inspect(a))
@5595afc47e60 13 INTSXP g1c0 [MARK,NAM(7)] (len=100, tl=0) 1,2,3,4,5,...
> b <- a
> .Internal(inspect(a))
@5595afc47e60 13 INTSXP g1c0 [MARK,NAM(7)] (len=100, tl=0) 1,2,3,4,5,...
```
So it seems the refcount (7?) doesn't count the number of names bound to the data.  This is for R3.6 - do you get different results?

MichaelChirico:
there were some changes in R ≥ 4.0 that may have an affect here (sorry I
don't know the full details)

OfekShilon:
This is what I get on R4.0.3:

```
> a <- 1:100
> .Internal(inspect(a))
@564a1336a8b0 13 INTSXP g0c0 [REF(65535)]  1 : 100 (compact)
> b <- a
> .Internal(inspect(a))
@564a1336a8b0 13 INTSXP g0c0 [REF(65535)]  1 : 100 (compact)
> a<- c(1,3,5)
> .Internal(inspect(a))
@564a1334bfb8 14 REALSXP g0c3 [REF(1)] (len=3, tl=0) 1,3,5
> b <- a
> .Internal(inspect(a))
@564a1334bfb8 14 REALSXP g0c3 [REF(2)] (len=3, tl=0) 1,3,5

```
I assume the difference is because 1:100 is now stored as ALTREP.

This seems encouraging, but is an R4.0 only solution acceptable?

MichaelChirico:
as long as the package works with R 3.1.0 it should be fine. it's ok for
the package to work better w newer versions of R.

OfekShilon:
It seems ref count is still hard to reason about in R4:
```
> a <- c(1,3,5)
> .Internal(inspect(a))
@5611e24877c8 14 REALSXP g0c3 [REF(1)] (len=3, tl=0) 1,3,5
> dt1 <- data.table(a=a, b=c(2,4,6))
> .Internal(inspect(dt1[["a"]]))     # 2 seems right
@5611e2487d18 14 REALSXP g0c3 [REF(2)] (len=3, tl=0) 1,3,5
> .Internal(inspect(dt1[["b"]]))     # but also for b?? It isn't referenced externally
@5611e2487d68 14 REALSXP g0c3 [REF(2)] (len=3, tl=0) 2,4,6
> .Internal(inspect(dt1[["b"]]))    # evaluating again it changes :(
@5611e2487d68 14 REALSXP g0c3 [REF(4)] (len=3, tl=0) 2,4,6
> .Internal(inspect(dt1[["a"]]))
@5611e2487d18 14 REALSXP g0c3 [REF(4)] (len=3, tl=0) 1,3,5
> a <- 1
> .Internal(inspect(dt1[["a"]]))   # dt1[[a]] is no longer referenced externally, but the refcount did not decrement:
@5611e2487d18 14 REALSXP g0c3 [REF(4)] (len=3, tl=0) 1,3,5
```
Am I missing something?   If not - 'manual' solutions are in order.

OfekShilon:
It seems ref counts are still very hard to reason about in R4:
```
> a <- c(1,3,5)
> .Internal(inspect(a))
@5611e3b71b78 14 REALSXP g0c3 [REF(1)] (len=3, tl=0) 1,3,5
> dt1 <- data.table(a=a, b=c(2,4,6))
> .Internal(inspect(dt1[["a"]]))    # 2 seems reasonable. But why was it copied??
@5611e3b71df8 14 REALSXP g0c3 [REF(2)] (len=3, tl=0) 1,3,5
> .Internal(inspect(dt1[["b"]]))    # Why 2? Not referenced externally
@5611e3b71e48 14 REALSXP g0c3 [REF(2)] (len=3, tl=0) 2,4,6
> a <- 1
> .Internal(inspect(dt1[["a"]]))     # not decremented :(
@5611e3b71df8 14 REALSXP g0c3 [REF(2)] (len=3, tl=0) 1,3,5
> 
> 
> df <- data.frame(a=11)
> .Internal(inspect(df[["a"]]))     # surprisingly high..
@5611e4099438 14 REALSXP g0c1 [REF(11)] (len=1, tl=0) 11
> dt <- setDT(df)
> .Internal(inspect(dt[["a"]]))     # at least it wasn't copied
@5611e4099438 14 REALSXP g0c1 [REF(12)] (len=1, tl=0) 11
> 
> rm(df)
> .Internal(inspect(dt[["a"]]))     # not decremented  :(
@5611e4099438 14 REALSXP g0c1 [REF(12)] (len=1, tl=0) 11
```
Am I missing something? It seems 'manual' solutions are still in order.

ColeMiller1:
Won’t this lead to all columns being copied as they are being assigned? All the columns are being marked externally referrenced which would lead to memrecycle. 

Maybe it would be better to have use cases of what `setDT` leads to copies or doesn’t. Related, there probably should be a way for advanced users to override this behavior like `setDT(..., copy = FALSE)`.

As for expected behavior, these are some examples for design. 
``` r
dt = setDT(list(x = 1)) # no copies needed

y = 2
dt = setDT(list(x = 1, y = y)) # y will need copies

DF = data.frame(x = 1)
setDT(DF) # no copies

y = 2
DF = data.frame(x = 1, y = y)
setDT(DF) # y will need copies

DF1 = data.frame(x = 1)
DF2 = DF1
setDT(DF1) # x will need copies, DF2 is unaffected

f = function (DF) {
  setDT(DF)
  DF[, x := 5]
  DF
}
DF3 = data.frame(x = 1)
DT = f(DF3) # DF3 is still data.frame but f() returns data.table. x will need copies

**Comment by OfekShilon** [comment]
> Won’t this lead to all columns being copied as they are being assigned? All the columns are being marked externally referrenced which would lead to memrecycle.

All columns in a `setDT`-generated DT will indeed be marked as externally referenced.  This doesn't mean they are always copied when assigned: if plonking takes place (i.e. `i` argument to `[` is empty and the column is replaced entirely) no copy takes place.

> Related, there probably should be a way for advanced users to override this behavior like `setDT(..., copy = FALSE)`.
Sounds good. I think I would prefer a global option personally.

> As for expected behavior, these are some examples for design.
> ...
> DF = data.frame(x = 1)
> setDT(DF) # no copies
Why not? 

> ...
We tried in-house some analysis of arguments to setDT (whether its argument is a function call, or whether its argument is a caller-argument etc.) and didn't get very far - as there was an overwhelming amount of corner cases.

Did anyone ever measure the performance impact of `memrecycle`? Personally I think giving it up altogether might be a worthy trade off for consistent semantics.

**Comment by ColeMiller1** [comment]
I like to design with use cases in mind. While I agree that it can be easy to get bogged down by edge cases, these are very common use cases.

>> As for expected behavior, these are some examples for design.
...
DF = data.frame(x = 1)
setDT(DF) # no copies

>Why not?

I expect objects that have no other references would not need to copy the vectors down the road. Why would we need to copy anything other than the pointers to the data? 

I can't really speak to the performance considerations of `memrecycle`.

**Comment by OfekShilon** [comment]
@ColeMiller1 examples of internal attempts to work around this limitation:
(1) We tried to detect if setDT is called on a function argument and assert. This is the common case of data-modification-leakage, but misses cases like `d1<-data.frame; d2<-d1; setDT(d2)`.
(2) We tried to _allow_ `setDT` to operate only on function calls, like `setDT(generateData(...))` - to force it to operate only on "new", unreferenced, data. This missed function calls like `cbind`, or even user-defined functions that leave column data in place.
And more.
In this context, 'design by use case' might mean - have the implementation work by the contract most of the time, and neglect some harder cases. As a user (of this library and others) I'm not in favor. In real-life size code bases this won't mean less bugs, but rather harder to isolate ones.  The right thing to do is to declare a contract and comply 100% of the time - or the package would be very hard to use.  This might mean changing the contract to make such compliance possible: a different example in DT is - it [cannot really apply to its argument by reference](https://github.com/Rdatatable/data.table/issues/4831#issuecomment-778640798), so I think it shouldn't try to. The same goes here: either `:=` modifications apply to all names bound to the date, or only to the current one. The current state, where _sometimes_ it's one and _sometimes_ the other - is extremely hard to work with, whether the 'other' is 10% or 0.1% of the cases.

I came across @jangorecki 's [benchmark](https://h2oai.github.io/db-benchmark/).  Would it be possible to use it to benchmark the main data.table against my fork?   (I can try myself but it would have to be on a machine serving many other users, not ideal for benchmarking...)

**Comment by jangorecki** [comment]
Yes, we can run db-benchmark tests for data.table branch other than master, it happened few times in the past already.

**Comment by OfekShilon** [comment]
> We can run db-benchmark tests for data.table branch other than master, it happened few times in the past already.
> I would like to hear @mattdowle and/or @arunsrinivasan comments on this idea. If it is a lot of work, then best to wait for their comments before investing more time into to.

Thanks @jangorecki.  If it's a lot of work on your side too then I'll try to find a dedicated machine to benchmark on myself.

**Comment by jangorecki** [comment]
Not at all, I just install branch to custom library location and run benchmark script.

**Comment by ColeMiller1** [comment]
@OfekShilon I had no intention that edge case would be fail - more so that there are expected behaviors that we should keep in mind. Specifically, if there is a way to design this so that columns unique to a data table are unaffected, that's what I would advocate for.  

This is tangentially relevant to the discussions and is at the very least a great reference:

https://github.com/RcppCore/Rcpp/issues/943

And this comment stood out as it is referring to copying pointers and not the underlying memory for Rcpp.

https://github.com/RcppCore/Rcpp/issues/943#issuecomment-469347159

**Comment by OfekShilon** [comment]
@jangorecki  I disabled in-place assignment completely in [this branch](https://github.com/OfekShilon/data.table/tree/4783_test). Could you please try and benchmark it?

**Comment by OfekShilon** [comment]
I just ran a quick and dirty benchmark myself, attached is the test script (note that in real life my R often crashes after `install_github` and I can't run the script in one piece).  The key code I run is - 
```r
...
dt <- setDT(df)
microbenchmark( dt[!is.na(a), a := newDat] )
```
Printing the microbenchmark results:
```
> mbVanillaIP
Unit: milliseconds
 vanilla i-p 202.3709 216.1531 245.5188 227.0159 280.0774 334.3973   100
> mbTestIP
Unit: milliseconds
 no in-place 247.6703 317.6447 321.4421 319.6458 321.6315 364.8844   100
```

So there is a noticeable - but I'd say *not* overwhelming - performance impact.
I think it's safe to say that in any form the assignment of data is never an interesting bottleneck - it is always its processing. Personally I'll be very, very happy to pay this small price in the initial creation of data if that means avoiding its leakage to places it doesn't belong.     (you can see again a demo of the problem and the solution in the attached script).

What do you guys think?
@jangorecki, @MichaelChirico, @ColeMiller1 

[testNoInPlaceDT.zip](https://github.com/Rdatatable/data.table/files/6178081/testNoInPlaceDT.zip)

OfekShilon:
The discussion diverged too much - I will close this PR and start a new one with the full suggested code in place.

PR #4908, opened 2021-02-17
Clarify status of reshape2

hadley:
Minor clarification — reshape2 is not deprecated, and we are committed to keeping it available on CRAN.


MichaelChirico (approved):
Thanks Hadley, LGTM

eddelbuettel:
There is a typo in the one line the PR changes. Apologies for posting an image but I think it drives the point home:

image

This presumably meant _superseded_ (though my instinct would still be to spell it _supercede_).

avimallu:
For completeness, I believe that the error message for `data.table::dcast` in the event something other than a `data.frame` or `data.table` is passed will also have to be changed:

```r
The dcast generic in data.table has been passed a factor and will attempt to redirect to the reshape2::dcast;
please note that reshape2 is deprecated, and this redirection is now deprecated as well. Please do this
redirection yourself like reshape2::dcast(as.factor(mtcars)). In the next version, this warning will
become an error.
```

I'm not very well versed with the `data.table` source code to suggest where to change this, but just pointing it out.

hadley:
I have corrected the typo, and updated all other mentions of reshape2 being deprecated (by grepping for reshape2).

mattdowle:
Thanks @hadley, will merge. Although this PR does touch code files, it's edits to messages, so I won't add you to the contributors list which would make you a copyright holder. There are one or two similar cases in the past so I'll be consistent. Any type of logic change, even just to a single line of R or C code, would be really welcome and I'd add you to contributors list at that point.

Btw, our use of the word 'deprecated' derives from your use of the word 'retired' in Dec 2017 here: https://github.com/hadley/reshape/commit/83ef5d25751dae1bf44a21f1d1ddb0ea66402d10.  
To this day, reshape's README starts :
> reshape2 is retired: only changes necessary to keep it on CRAN will be made. We recommend using tidyr instead.

To me, that is the definition of deprecation, but I see your https://github.com/Rdatatable/data.table/issues/3549#issuecomment-562164234 pointing to https://lifecycle.r-lib.org/articles/stages.html#retired and I see that you have since retired the word retired (footnote 1 in the doc). Would it make sense to update the word 'retired' in https://github.com/hadley/reshape/blob/master/README.md to 'superseded' then?

hadley:
@mattdowle I have no expectation of becoming a contributor based on this tiny contribution (which as you say, arises from our own confusion around messaging). I've fixed the reshape2 readme — thanks for pointing that out!

mattdowle:
@hadley I didn't think you had that expectation, sorry I didn't write that very well: I've seen comments from others who wonder why I haven't added you, so I was more documenting for them. And I was taking the opportunity to welcome you and really encourage you to contribute code.

PR #4909, opened 2021-02-17
First look for .datatable.aware in environment

hadley:
This is a PR requested by @MichaelChirico to kick off discussion of the issue described in https://github.com/tidyverse/dtplyr/issues/184 — i.e., it would be nice if dtplyr could declare the code that it generates and evaluates is data.table aware, independently of what package that is running that code.

If this idea seems interesting/promising I'm happy to finish off with a NEWS bullet, tests (although I'm not yet sure how to test this specific), etc. But I wanted to make sure this contribution was appropriate before doing more work.


R/cedta.R
```diff
@@ -22,7 +22,12 @@ cedta.pkgEvalsUserCode = c("gWidgetsWWW","statET","FastRWeb","slidify","rmarkdow
 # cedta = Calling Environment Data.Table-Aware
 cedta = function(n=2L) {
   # Calling Environment Data Table Aware
-  ns = topenv(parent.frame(n))
+  env = parent.frame(n)
+  if (isTRUE(ns$.datatable.aware)) {
```
MichaelChirico, line 26: @hadley `ns` doesn't exist here, is this supposed to be `env$.datatable.aware`?

hadley: Now I don't remember what I was thinking 😞 But yes using `env$.datatable.aware` seems reasonable.

mattdowle:
Thanks, @hadley. Great solution.
I'm not sure how to test this either from data.table's side. I'll add the news item and merge so you or your team can test either dev now or after release. I don't see much risk in adding this given it's just looking for `.datatable.aware` in the calling environment as well now. If it doesn't work we can always revisit.
I've invited you to be project member (it should be a button you need to click in your GitHub profile or projects page) and will add you to contributors list. This is our standard practice for any code change PRs. Thanks again.

hadley:
Got it, thanks!

PR #4750, opened 2020-10-11
speedup %like% for factors 

shrektan:
Ref to #4748 

We can avoid the coercing and matching by subsetting the result using factor value, which is equivalent to use the numeric code, according to ?`[` .

## Benchmark Code

```r
library(data.table)                                                                                            
library(microbenchmark)

set.seed(1L)                                                                                                   
n_disctinct <- 100L                                                                                            
n_letters <- 5L                                                                                                
x <- replicate(n_disctinct, paste(sample(LETTERS, n_letters, replace = TRUE), collapse = ""))                  
n_length <- 1e6L                                                                                               
x <- sample(x = x, size = n_length, replace = TRUE)                                                            
f <- factor(x)   
microbenchmark(
  f %like% "SSA"
)
``` 

## Before the PR

```
Unit: milliseconds
           expr      min       lq     mean   median       uq      max neval
 f %like% "SSA" 14.04802 14.90246 17.14793 17.00165 18.11933 58.64134   100
```

## After the PR

```
Unit: milliseconds
           expr      min       lq     mean   median       uq      max neval
 f %like% "SSA" 4.586466 4.994767 6.548575 5.773167 7.093326 49.07044   100
```

PR #4916, opened 2021-02-22
like uses a faster approach for factor input

MichaelChirico:
`%in%` always struck me as odd, only thinking of this approach now


avimallu:
I may be wrong, @MichaelChirico, but this looks identical to the recently merged #4750.

MichaelChirico:
You're right @avimallu, thanks!

PR #4924, opened 2021-03-05
internal cleanup -- new helper clip_msec

MichaelChirico:


MichaelChirico:
A bit strange that codecov is dinging us here. -10 lines, -10 hits, so no change really.

PR #4927, opened 2021-03-05
use paste0 in vignette

MichaelChirico:


MichaelChirico:
Strange... Travis says it passed: https://travis-ci.org/github/Rdatatable/data.table/builds/761521009

PR #4931, opened 2021-03-07
Implementation of %notin%

### LINKED ISSUE #4152: "not in" operator `%!in%`
FR to propose a not-in operator
```r
`%!in%` = function(x, y) !(x%in%y)
```
Motivation is to have a single call that we could optimise. Otherwise we have to investigate unevaluated expression tree more deeply (and in a much more complex way) to know when user is doing a not-in operation.
Pseudo code in R C api
```r
CAR(x %!in% 1:10)
#`%!in%`
CAR(! x %in% 1:10)
#`!`
```

mczek:
I implemented %notin% to close #4152 .


R/notin.R
```diff
@@ -0,0 +1,6 @@
+# Intended to be used to create %notin% operator
+notin = function(example, elements) {
+  return(!match(example, elements, nomatch = 0))
```
MichaelChirico, line 3: We should branch on `is.character(example)` and use `chmatch` for string input

R/notin.R
```diff
@@ -0,0 +1,6 @@
...
+}
+
+"%notin%" = function(example, elements) notin(example, elements)
```
MichaelChirico, line 6: Same branching here -- we wrote an api for `%chin%` that gives the `%chin%` output directly instead of wrapping `chmatch() > 0L`

R/notin.R
mczekanski1: @MichaelChirico Thank you, I just added this branching.

R/notin.R
mczekanski1: @MichaelChirico Could you explain this a little more? I was trying to mimic the implementation of `%like%` where there is a defined R function `like` that's wrapped make the operator. So if we have branching for characters in the `notin` function, then shouldn't that handle branching for the opreator as well?

MichaelChirico: `notin` does handle branching, but it dispatches to `chmatch`. If we branch in `%notin%`, we can dispatch to `%chin%` directly (which doesn't use `chmatch` in R).

This is different for character input because if we dispatch to `%in%`, it will just re-route to `match` anyway

mczekanski1: I'm sorry if I'm missing something, but it seems like if it's better to dispatch to `%chin%` than `chmatch`, then we should use `%chin%` in both `notin` and `%notin%` and then their definitions would be the same.

This is would be the operator implementation and the function would match:


```
"%notin%" = function(example, elements) {
  if (is.character(example)) {
    return(!(example %chin% elements))
  } else {
    return(!match(example, elements, nomatch = 0))
  }
}
```

MichaelChirico: I think that makes sense. Actually I am not sure if we need `notin()` -- for `like()`, it's a more extensible version of `%like%` since it allows to control several more options. whereas here, the signature is the same, it's just a difference of using it as an operator. 

FWIW, I think we should also look into trying to add an option in the C API like `invert` so that we don't need to `!` at the end . Then the function could look like 

```
"%notin%" = function(example, elements) {
  if (is.character(example)) {
    return(.Call(cchin, x, table, /* invert= */TRUE))
  } else {
    return(!match(example, elements, nomatch = 0))
  }
}
```

mczekanski1: That sounds like a good idea to me. I will try to change the `chin` function to include the invert parameter. It looks like this means altering `chmatchMain` which is also used for `chmatch` and `chmatchdup`, so those would also be easy to negate.

mczekanski1: I think I've implemented this change pretty efficiently. Please let me know if this looks good or if there are any suggestions.

R/notin.R
```diff
@@ -0,0 +1,7 @@
+"%notin%" = function(example, elements) {
```
MichaelChirico, line 1: please use the same signature as `%in%`, i.e., `function(x, table)`

man/notin.Rd
```diff
@@ -0,0 +1,34 @@
+\name{notin}
+\alias{\%notin\%}
+
+\title{
+Convenience for checking if an example is in a set of elements
+}
+
+\description{
+Intended to behave opposite to  \code{\link[=base]{in}}
```
MichaelChirico, line 9: "Opposite" doesn't feel very precise here (opposite how?), how about "Check whether an object is absent from a table, i.e., the logical inverse of `%in%`"?

man/notin.Rd
```diff
@@ -0,0 +1,34 @@
...
+example \%notin\% elements
+}
+
+\arguments{
+   \item{example}{ vector or \code{NULL}: value to be matched }
+   \item{elements}{ vector or \code{NULL}: values to check for a match }
+}
+
+\details{
+ Internally, \code{\%notin\%} is a wrapper around \code{\link[=base]{match}} and \code{\link[=data.table]{chmatch}}.
```
MichaelChirico, line 22: I'm not 100% sure the details section here is necessary

man/notin.Rd
```diff
@@ -0,0 +1,34 @@
...
+}
+
+\value{
+    Logical vector, \code{TRUE} indicating whether each \code{example} was not found in \code{elements}.
+}
+
+\seealso{ \code{\link[base]{match}} }
```
MichaelChirico, line 29: Also link `chmatch`

man/notin.Rd
```diff
@@ -0,0 +1,34 @@
...
+
+\examples{
+  11 \%notin\% 1:10
```
MichaelChirico, line 32: An example with both `TRUE` and `FALSE` in the output would be more instructive

R/notin.R
```diff
@@ -0,0 +1,7 @@
...
+  if (is.character(example)) {
+    return(.Call(Cchin, example, elements, TRUE))
```
MichaelChirico, line 3: this branch doesn't have any tests:

https://app.codecov.io/gh/Rdatatable/data.table/compare/4931/diff#D1L4

inst/tests/tests.Rraw
```diff
@@ -17263,3 +17263,10 @@ if (identical(x, enc2native(x))) {
 
 # fintersect now preserves order of first argument like intersect, #4716
 test(2163, fintersect(data.table(x=c("b", "c", "a")), data.table(x=c("a","c")))$x, c("c", "a"))
+
+
+# Test new feature %notin%, #4152
+test(2164.1, 11 %notin% 1:10, TRUE)
+test(2164.2, "a" %notin% c(), TRUE)
+test(2164.3, "a" %notin% c("a", "b", "c"), FALSE)
+test(2164.4, c(1, 2) %notin% c(1,2,3), c(FALSE, FALSE))
```
MichaelChirico, line 17272: could you add some edge case tests as well (`character()`, `integer()` and `NULL`)? Should be fine since we're building on `%in%` but just in case.

R/notin.R
MichaelChirico: Looking at the tests it should be covered. Maybe it hasn't re-run coverage yet.

mczekanski1: That link doesn't work for me, but I do have a couple of tests.

MichaelChirico: seems to have gone away on merge thanks

inst/tests/tests.Rraw
```diff
@@ -17273,3 +17273,12 @@ if (test_bit64) {
...
 }
 
+# Test new feature %notin%, #4152
+test(2165.1, 11 %notin% 1:10, TRUE)
+test(2165.2, "a" %notin% c(), TRUE)
+test(2165.3, "a" %notin% c("a", "b", "c"), FALSE)
+test(2165.4, c(1, 2) %notin% c(1,2,3), c(FALSE, FALSE))
+test(2165.5, "a" %notin% character(), TRUE)
+test(2165.6, "a" %notin% integer(), TRUE)
+test(2165.7, "a" %notin% NULL, TRUE)
```
MichaelChirico, line 17283: we need tests on `NA` as well. In particular `%in%` _always_ returns `TRUE`/`FALSE`, I guess `%notin%` should work analogously.

It raises a small puzzle -- if `NA %in% 1:5` is `FALSE`, should `NA %notin% 1:5` be `TRUE` or `FALSE`? `TRUE` would be the natural answer to ensure `%in%` and `%notin%` are inverses but I think there could be an argument it should be `FALSE`.

schrnstn: Just stumbled upon this as a long time data.table user.

From my perspective, `NA %notin% 1:5` should return `TRUE` to be analogous to `%in%`.
The key observations that leads me to this is that `%in%` treats `NA` as an allowed value in the right-hand-sided vector, in this case `1:5`. To see this, note that `NA %in% c(1:5, NA)` indeed returns `TRUE`.
Having this behaviour in mind, returning `FALSE` is unnatural in the use case you described.

mczekanski1: I agree with this suggestion and have added tests to reflect this behavior.

mczekanski1:
I just added the code from`data.table` repository so it passes the check.

mattdowle:
Thanks for working on this.
Just started to look at this and my first thought is wondering why `negate` has been added to `chmatchMain`. Why not leave `chmatchMain` as-is and iterate on its output afterwards to negate it in a C wrapper functon? Adding `negate` into `chmatchMain` is adding to its complexity slightly and I'm wondering why not keep negate isolated in a C wrapper function.

mczekanski1:
@mattdowle I was thinking that this would most efficiently implement the negation and be a starting point for negating the other dependencies of `chmatchMain`. I thought that the easiest way to implement the wrapper function is to create a `negate` function, which takes a logical `SEXP` vector and negates each element to make it reusable for other cases. That seemed a bit trivial and I thought that there would be a slight performance improvement by negating at the source in chmatchMain, especially when the difference is negating the values to be placed in the output. However, I'm open to changing it to a `negate` wrapper function.

mczekanski1:
> Thanks for working on this, @mczekanski1.
> Just started to look at this and my first thought is that adding `negate` into `chmatchMain` is adding to its complexity slightly and I'm wondering why not keep `negate` isolated in a C wrapper function which calls `chmatchMain` and then iterates on its output to negate it.

@mattdowle I looked into this a bit and profiled a few approaches. I'm not sure that my implementation is necessarily faster because it's a time savings of `O(len(x))`, which isn't particularly noticable, even with large x. I could instead implement a wrapper negate function instead that would basically be the function I use below or just wrap the `!` operator because it seems to perform similarly.

```
library(data.table)
library(microbenchmark)
library(inline)
 
negate_fun <- cfunction(sig = signature(x = "SEXP"),
body = '
if(TYPEOF(x) != LGLSXP) {
  error("not logical vector");
}
int n = length(x);
int nprotect = 0;
SEXP ans = PROTECT(allocVector(LGLSXP, n)); nprotect++;
int *ansd = INTEGER(ans);
for(int i=0;i<n;i++) {
  int val = LOGICAL(x)[i]?0:1;
  ansd[i] = val;
}
UNPROTECT(nprotect);
return ans;')
 
x <- sample(letters, 1e6, TRUE)
microbenchmark(
  "in" = letters %in% x,
  "!in" = !(letters %in% x),
  "!chin" = !(letters %chin% x),
  "cLoop" = negate_fun(letters %chin% x),
  "notin" = letters %notin% x
)
# Unit: milliseconds
#  expr      min       lq      mean    median        uq       max neval
#    in 8.791168 9.056458 12.444812 11.209896 14.159338 92.336123   100
#   !in 8.803188 9.031143 12.804303  9.329833 14.003813 93.118404   100
# !chin 5.282855 5.347939  5.468743  5.452525  5.554680  6.124725   100
# cLoop 5.286705 5.346260  5.490601  5.470050  5.571715  6.240186   100
# notin 5.280464 5.370345  5.520032  5.461920  5.585090  6.456345   100
 
missing_vals <- as.character(1:100)
microbenchmark(
  "in" = missing_vals %in% x,
  "!in" = !(missing_vals %in% x),
  "!chin" = !(missing_vals %chin% x),
  "cLoop" = negate_fun(missing_vals %chin% x),
  "notin" = missing_vals %notin% x
)
# Unit: milliseconds
#  expr      min       lq      mean    median        uq       max neval
#    in 8.845388 9.156558 13.213133 12.825747 14.051578 90.564042   100
#   !in 8.850098 9.066103 11.182928  9.437478 13.862617 15.598714   100
# !chin 5.364905 5.384320  5.483883  5.424180  5.579690  5.746956   100
# cLoop 5.370195 5.386195  5.510441  5.458745  5.580490  6.050866   100
# notin 5.367015 5.385040  5.477812  5.458620  5.563875  5.718045   100

```

mczek:
@MichaelChirico @mattdowle I just updated my fork to include changes to the original repo. I don't have approval for workflows, so I think the PR won't be marked as able to be merged automatically. I'm happy to work on this some more if you think it would still be helpful, I'm just not sure what the best direction is based on the profiling I did previously, given that using `!` isn't slower than a C function to negate each element.

MichaelChirico (approved):
Thanks for revisiting & the benchmarks!

mattdowle (changes requested):
Yes, thanks for the benchmarks and work on this, @mzek and @MichaelChirico.
I'm looking at the changes to `chmatch.c` and it's those line changes that increases its complexity more relative to what I had in mind and suggested; e.g. the extra `chinNoMatch` variable and the loop code repeated when `negate`.
When I wrote above the following:

> I'm wondering why not keep negate isolated in a C wrapper function which calls chmatchMain and then iterates on its output to negate it.

I didn't have in mind an `allocVector()` for that. Since the C wrapper function would be calling `chmatchMain`, it would know it is receiving new memory from it and therefore can iterate directly on its contents (negating it) by reference. So that's one thing wrong with the benchmark: memory allocation and garbage collection will muddy the waters when the data size is increased. More to the point, if the negation is done by reference it doesn't need the extra memory so will work whereas at some point as vector size increases the negation using the allocation will start to fail with out of memory. A contiguous loop through memory is almost insignificant; i.e. rarely worth saving. Especially where the downside of saving it is increased complexity which makes C code harder to maintain in future. Providing the loop is contiguous (not jumping around memory) and not doing anything meaty inside the loop such as referencing other memory which could be different memory on each iteration. The suggested loop to negate is about as simple as it gets, although it does involve a write. But since it's writing to the same cache line the CPU has just pulled into cache, it should be optimized by the CPU pipeline; e.g. (roughly speaking) I don't think it will broadcast a cache write for each of the 16 writes within the line; it'll wait for all the writes to that line and broadcast one cache line write because the pipeline knows it is a contiguous loop. Something like that.

If that all makes sense and I understood correctly, please revise this PR so that its a C wrapper which is one single simple loop which does the negation by reference on the output from `chmatchMain`. Perhaps a new benchmark with much bigger data than just 1e6 to check if I'm correct that the extra loop to negate-by-reference does not increase the time by anything significant. I think the diff lines (37 `chmatch.c`, 4 `assign.c`, 2 `subset.c`) should reduce dramatically. And since these particular diff lines in this case are increasing complexity then it's worth doing in this case I would say because I look at the new variables and logic and I don't relish that being in `chmatch.c` in future. It is already one of the trickiest parts of data.table C code due to the way it clobbers the STRSXP truelength. That is really complex logic but there is no other known way, and the speedup is so large that the complexity is worth it in that case. There would have to be an utterly correct benchmark (i.e. big data and no allocVector) of the approach I'm suggesting showing the loop was worth saving so that I have something to point to as justification for why the extra complexity was added.

R/notin.R
```diff
@@ -0,0 +1,9 @@
+"%notin%" = function(x, table) {
+  if (is.character(x) && is.character((table))) {
+    result = .Call(Cchin, x, table)
+  } else {
+    result = match(x, table, nomatch = 0) > 0
+  }
+  return(.Call(Cnegate, result))
```
mattdowle, line 7: Bit uncomfortable with a `.Call` doing this by reference. I suppose in this simple function as it stands it will work but there is nothing to indicate this changes the logical vector by reference to the reader of the code. But I see what you mean about reusing it to negate the output from `match`/`%in%` too which is nice and better than what I had in mind which was `Cchnotin` wrapper just for that branch. Ok so maybe naming it `CnegateByRef` would be clear so that a comment isn't as critical to remember in future if we use it elsewhere too.

mattdowle (changes requested):
Great. Thanks. Much more comfortable with this approach.

R/notin.R
```diff
@@ -0,0 +1,9 @@
+"%notin%" = function(x, table) {
+  if (is.character(x) && is.character((table))) {
+    result = .Call(Cchin, x, table)
+  } else {
+    result = match(x, table, nomatch = 0) > 0
```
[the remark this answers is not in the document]
MichaelChirico: look at print(`%in%`)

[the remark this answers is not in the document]
mattdowle: Oh :-)  So a direct `%in%` could be submitted to R which used half as much memory. Probably just a new flag to be passed to the internal C function.  Even when the data isn't big, saving half as many small objects on the heap should be beneficial to save garbage collection. Albeit, probably most usage of `x %in% y` is a length-1 `x` though and so that's likely using R's internal global true/false off heap, and global constant small integers. But outside those cases it would be beneficial. Shouldn't be any downsides really, other than actually getting it done.

For now, this `%notin%` could drop the `>0` part to save that allocation, and that could be dealt with by the `negate` wrapper by reference.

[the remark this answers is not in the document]
mczek: Thanks for taking a look @mattdowle ! I'm not sure which approach you're advocating for in your comment `%in%`. It might just be that I don't understand the R internals that well. Could you elaborate? As for clobbering the internal global TRUE/FALSE constant,  when I ran `1 %notin% 1:10` twice with this approach I would get `FALSE` the first time and `TRUE` the second time. Thank you for explaining why that was happening. My new approach solves this.

R/notin.R
mczek: I agree that it doesn't seem great to be altering vectors by reference from R, so I've moved the negation to C where it can be used more intuitively.

src/negate.c
```diff
@@ -0,0 +1,14 @@
+#include "data.table.h"
+
+SEXP negate(SEXP x) {
+  if(TYPEOF(x) != LGLSXP) {
+    error("not logical or integer vector");
+  }
+  int n = length(x);
+  int *ansd = INTEGER(x);
+  for(int i=0;i<n;i++) {
+    ansd[i] = LOGICAL(x)[i] == TRUE?0:1;
```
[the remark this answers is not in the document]
mczek: I like this approach to check `NA_LOGICAL`, so I've switched to using your first suggestion. I think it would be nice to be branchless, but I'm having a hard time understanding how the second approach would work. If the element is `NA_INTEGER` we would be setting the element of `ansd` to `NA_INTEGER xor true`, which doesn't intuitively have a value to me, but perhaps it can be evaluated?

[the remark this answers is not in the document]
mattdowle: Great. I like the `match() == 0` approach you've done in the latest commit so the negate isn't needed on that result. Minor improvement would be to use `0L` to avoid the `0` double to integer coercion.
On this branchless question, if the element is NA_INTEGER,  `!= NA_INTEGER` will be false (not true) so that `NA_INTEGER xor false` will do nothing and leave `NA_INTEGER` as `NA_INTEGER`. And now I realize I meant NA_LOGICAL of course (not NA_INTEGER). Even though they're the same currently in R I maintain hope that one day type logical will be changed to be a single byte with NA_LOGICAL then having the value -127. Best be tidy and future proof where we can and use LOGICAL here because that is the correct type here.

mattdowle:
Found the following project setting showing the default : 
image
Have now relaxed that as far as it lets me to the first option. So CI should just run again like before, without needing approval every time.

src/negate.c
```diff
@@ -0,0 +1,20 @@
+#include "data.table.h"
+
+void negateByRef(SEXP x) {
+  if(TYPEOF(x) != LGLSXP) {
+    error("not logical or integer vector");
+  }
+  int n = length(x);
```
mattdowle, line 7: Good to declare this `const int n =`. That may help the compiler in the `for()` loop telling it that `n` can't change so it might be able to optimize `i<n` on every iteration (by doing it in batches or something clever like that). Since we target all sorts of compilers on many architectures, perhaps some of them benefit from the `const` declaration.

src/negate.c
```diff
@@ -0,0 +1,20 @@
...
+  int *ansd = INTEGER(x);
+  for(int i=0;i<n;i++) {
+    if (ansd[i] != NA_LOGICAL) ansd[i] ^= 1;
+  }
+}
+
+
+SEXP notchin(SEXP x, SEXP table) {
+  SEXP result = chin(x, table);
```
mattdowle, line 16: Needs a PROTECT here. Because something inside negateByRef could trigger a GC which could free the result. It's hard to see why calling INTEGER() could possibly cause a GC (since it's just a pointer offset) but it is an R API call, and we can't be sure about any R API call. Sometimes every entry point in the R API is forced to GC (R can be compiled to do that to debug) and that might be why INTEGER() could GC. Or perhaps the object is an ALTREP and needs to be materialized which could GC. Regardless, `Rchk`'s static analysis would detect the unprotected return so this line wouldn't pass release procedures at that stage.

R/notin.R
```diff
@@ -0,0 +1,9 @@
+"%notin%" = function(x, table) {
+  if (is.character(x) && is.character(table)) {
+    result = .Call(Cnotchin, x, table)
+  } else {
+    result = match(x, table, nomatch = 0L) == 0
```
mattdowle, line 5: 0L on the RHS of == too pls

Also I would just drop the `result =` parts and the `return(result)`. The value of `if()` will be returned because that's the last value.

src/negate.c
```diff
@@ -0,0 +1,20 @@
...
+  int *ansd = LOGICAL(x);
+  for(int i=0;i<n;i++) {
+    ansd[i] ^= (ansd[i] != NA_LOGICAL); 
+  }
+}
+
+
+SEXP notchin(SEXP x, SEXP table) {
+  SEXP result = PROTECT(chin(x, table));
+  negateByRef(result);
```
mattdowle, line 17: Needs an UNPROTECT(1) here.

mattdowle (changes requested):

mattdowle:
@mczek Thanks for all your work here. Have invited you to be project member. It should be a button in your GitHub profile or projects page that you need to click to accept.

mattdowle (approved):

mczek:
> @mczek Thanks for all your work here. Have invited you to be project member. It should be a button in your GitHub profile or projects page that you need to click to accept.

Thanks @mattdowle! I'm sorry but I didn't get to the email quick enough and the invite has expired. I'd love to join as a project member. Could you please resend the link?

mattdowle:
@mczek Done. (Seems like GitHub introduced a 7 day expiry. Seems way too short.)

PR #4938, opened 2021-03-16
Fix typo in ?setDTthreads

bobjansen:
I hope this is not too minor


MichaelChirico (approved):
no such thing! thanks for the fix

PR #4944, opened 2021-03-26
Keep joins grouped by foreign columns from crashing

### LINKED ISSUE #4892: Segfault in `[.data.table`
The following code (see [attached file](https://github.com/Rdatatable/data.table/files/5927697/example.RData.gz) for the data)

```
library(data.table)
load("example.RData")
col = "ground.temp.min"
filter.zeros.thresh.K = 3
obs[.(check.regions$stn, check.regions$date), by = check.regions$g, (col) :=
     {v = get(col)
      if (!any(!is.na(v) &
              !zero &
              abs(v) <= filter.zeros.thresh.K))
          v[date == median(date)] = NA
      v}]
```

causes this segfault:

```
 *** caught segfault ***
address 0x56386f5fd8dc, cause 'memory not mapped'

Traceback:
 1: `[.data.table`(obs, .(check.regions$stn, check.regions$date),     by = check.regions$g, `:=`((col), {        v = get(col)        if (!any(!is.na(v) & !zero & abs(v) <= filter.zeros.thresh.K))             v[date == median(date)] = NA        v    }))
 2: obs[.(check.regions$stn, check.regions$date), by = check.regions$g,     `:=`((col), {        v = get(col)        if (!any(!is.na(v) & !zero & abs(v) <= filter.zeros.thresh.K))             v[date == median(date)] = NA        v    })]
```

`sessionInfo` is:

```
R version 4.0.2 (2020-06-22)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:

attached base packages:

other attached packages:
[1] data.table_1.13.6

[1] compiler_4.0.2
```

tlapak:
Closes #4892

So this fixes the reported segfault, but I'm not sure if there isn't something else that's going wrong. Slightly modifying @ColeMiller1's example, we get 
```R
library('data.table')
x = data.table(id = 1:4, grp = 1:4, key = 'id')
y = data.table(id = 2:5, key = 'id')
x[y, by = x$grp, .SD, verbose = TRUE]
>        x    id
>    <int> <int>
> 1:     1     2
> 2:     2     3
> 3:     3     4
> 4:     4    NA

x[y, by = grp, .SD, verbose = TRUE]
>      grp    id
>    <int> <int>
> 1:     2     2
> 2:     3     3
> 3:     4     4
> 4:    NA    NA
```
Is that the correct behaviour? They can't both be correct, right? I'm not really sure what the intended purpose of grouping by a non-member column is. Maybe it would be better to catch this further up.

Also just as a remark, it's not necessary for this to be keyed join to trigger the sagfault.


ColeMiller1:
Thanks for looking into it deeper. I'm not sure if it's just Windows, but segfaults are difficult to debug for me - shutting down R each time is disruptive.

Should expected behavior be to simply error on joins when something other than `by = .EACHI` is used? For true joins, `by = .EACHI` is the only thing supported for now, AFAIU.

tlapak:
I thought about it some more and I guess aggregating a join makes sense? But I can't find documentation on it and the behaviour is somewhat unexpected to me. I played around with it a bit more and it works out as shorthand for slightly different join + group by operations:
```R
library('data.table')

x = data.table(id = 1:4, grp = c(2L, 1L, 2L, 1L), key = 'id')
y = data.table(id = 2:5, grp = c(2L, 1L, 2L, 1L), key = 'id')

all.equal(x[y, by=grp, .SD],
          x[y, .(grp=x.grp, id=x.id)][, .SD, grp])
> [1] TRUE

all.equal(x[y, by=.(grp=x$grp), .SD],
          x[y, .(grp=i.grp, id=x.id)][, .SD, grp])
> [1] TRUE

all.equal(x[y, .SD],
          x[y, .(id=i.id, grp=x.grp)])
> [1] TRUE
```
Although I haven't tried it for more complex examples.
Interestingly, the columns of the i argument aren't available when `by` is passed a column.

> I'm not sure if it's just Windows, but segfaults are difficult to debug for me - shutting down R each time is disruptive.

Good old print debugging... The tooling on Windows can be a bit of a pain.

ColeMiller1:
Doh. Sorry for the mistake - I remembered that `by = i.col` but should have tested `by = x.col` before commenting.

I think as long as `data.table` allows for column use from the `parent.frame`, there will just be some edge cases. The only other thought I have is to probe the `by` expression to try to determine if the variable is in the `data.table`. If the variable is in the parent.frame and `!is.null(irows)` then a warning might be good as it's unclear what the expected output should be. That is, would `by = parent_frame_col` be subsetted as well or would it be used as is.

The last line of this is what explains the discrepancy. While `x` has been subsetted, the variable in the `parent.frame` has not been. 

https://github.com/Rdatatable/data.table/blob/ec1259af1bf13fc0c96a1d3f9e84d55d8106a9a4/R/data.table.R#L790-L800

mattdowle:
Great fix! I confirmed the new test crashed for me too before this fix.

Issue #3215, opened 2018-12-13
DT[TRUE] can lead to invalid key


mattdowle:
I didn't realize `DT[TRUE]` was a way to achieve a shallow copy.  shallow copy was only intended for internal use. Thanks to @renkun-ken for highlighting this in #3214.

In v1.11.8 we see this :   
```R
DT = data.table(id = 1:5, key="id")
DT1 = DT[TRUE]
key(DT1)
[1] "id"
DT1[3, id:=6L]
key(DT1)
# NULL
DT$id
# [1] 1 2 6 4 5     # should be 1:5
key(DT)
# [1] "id"              # invalid key
```

It only occurs after `DT[TRUE]`, iiuc, which hopefully folk have not discovered or relied on too much?!   I hope the usage out there is like @renkun-ken described to add new columns to the shallow copy,  not to change existing columns!

renkun-ken:
Yes, setkey, changing existing columns cannot be used on the shallow copy since columns themselves are not copied.

jangorecki:
If we won't allow to make shallow copy with `dt[TRUE]` this issue will be automatically resolved.

mattdowle:
Eventually. But in the meantime, we can't break @renkun-ken's workflow.
More detail here: https://github.com/Rdatatable/data.table/issues/3214#issuecomment-462490046

jangorecki:
the following code could be added to tests to ensure copy behaviour
```r
DT = data.table(a=c(1,2), b=c("b","a"))
address(DT)
address(DT[])
address(DT[, .SD])
address(DT[TRUE])
sapply(DT, address)
sapply(DT[], address)
sapply(DT[, .SD], address)
sapply(DT[TRUE], address)
```

MichaelChirico:
Now that we have regex search in GitHub we can get a better sense of how this is used downstream:

https://github.com/search?q=lang%3AR+%2F%28%5B%5E%21%3D%3E%3C%5D%3D%7C%3C-%29%5B+%5D*%5Ba-z0-9._%5D%2B%5C%5B%5B+%5D*TRUE%5B+%5D*%5C%5D%2F&type=code

(a bit hard to ensure it's `[.data.table` statically so this is all "assignment of `name[TRUE]`")

Really rare usage pattern indeed. Here's one usage that per the comment assumes `[TRUE]` is inducing a copy:

[...]

PR #4947, opened 2021-03-30
Fix #4783 with a designated option

### LINKED ISSUE #4783: Inconsistent semantics after setDT
```
> d1 <- data.frame(a=c(1,2,3,4), b=c(2,3,4,5))
> d2 <- d1
> setDT(d2)  # At this point d2 is a shallow copy of d1, pointing to the same columns
```

Do modifications to `d2` impact `d1`?  We could live with both 'yes' or 'no', but the answer is _sometimes_:
```
d2[, b:=3:7]         # (1) impacts only d2
d2[, c:=4:8]         # (2) impacts only d2
d2[!is.na(a), b:=31] # (3) impacts both
d2[, b:=30]          # (4) impacts both
```
In cases 1&2 `d2` 'plunks' the full columns into itself and `d1` isn't affected. In cases 3 & 4 it seems that operation-in-place optimization kicks in (`address(d2$b)` is unchanged), so there is no copy-on-write and data still pointed to by `d1` is overwritten.

These semantic discrepancies make (the otherwise great) setDT unusable to us except in the most trivial scripts.

`#` `Output of sessionInfo()`
R version 3.6.1 (2019-07-05)
Matrix products: default

OfekShilon:
This PR adds an option that disables in-place assignment - thereby solving #4783 and also indirectly the old #3215.
The option is currently initialized to TRUE (old behaviour) - but at least for us the right initialization is FALSE (fixed behaviour).

Measurements on various size DTs show that the added performance overhead (extra allocations) is definitely manageable.


OfekShilon:
This continues the discussion in [this PR](https://github.com/Rdatatable/data.table/pull/4902)

OfekShilon:
I'm not sure how to continue - all tests pass for me locally.  Could it be a merge conflict I missed?

inst/tests/tests.Rraw
```diff
@@ -17264,12 +17271,40 @@ if (identical(x, enc2native(x))) {
...
+
+# In-place assignment can leak to origin DF, #4783
+ orig.opt <- options(datatable.assign.inplace=TRUE)
+ d <- data.frame(a=c(1,2), b=c(1,2))
+ df <- d
+ setDT(d)
+ d[, a:=c(10, 11)]  # not leaking to df
+ test(2164.2, df$a[1], 1)
+ d[!is.na(a), b:=c(20,21)]  # leaks to df
+ test(2166, df$b[1], 20)
```
MichaelChirico, line 17291: the issue is test numbering. here is 2166, next is 2164.3 -- tests should be in increasing order in the file.

OfekShilon: Thanks! I just added another commit (with other improvements) and hopefully this is fixed.

PR #4959, opened 2021-04-20
credit Michael for PR in news item

tdhock:
also remove diff.prev because that was just the column name in the issue example (not related to the bugfix)


mattdowle:
Thanks @tdhock for clearing this up.
It seems that @MichaelChirico gave you credit for the PR in his PR so it seems like he meant for reporting. The background is that historically we haven't been crediting ourselves in the news items. That came from years ago when it was mainly Arun and I. Then when @jangorecki and @MichaelChirico have contributed regularly in recent years, I suggested they continue in the same way. However, now we have more regular contributors, perhaps this should be revised and we should credit ourselves in each news item going forward.  Maybe all the news items in the last few releases should be made consistent in this regard.

tdhock:
Ok, thanks, just wanted to make sure it was clear that I reported the issue but not the PR/fix.
My day job is mainly academic research at the university so maybe I'm too obsessed with "giving credit where it is due" haha.

PR #4978, opened 2021-05-06
Fix 4783 - by adding an allow.assign.inplace attribute

OfekShilon:
Previous fix attempts were too general (global option) or too specific (per-column attribute). This one, I hope, hits the right balance: `setDT` adds a per-dt attribute that disables in-place assignment semantics.  It was made from a fresh fork to ease merging.
I'll close the previous PRs.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -7491,12 +7491,12 @@ test(1542.06, DT1, data.table(id=1:5, newCol=6:10, key="id"))
 DT1[3, id:=6L]
 test(1542.07, DT1, data.table(id=INT(1,2,6,4,5), newCol=6:10))
 
-# current wrong behaviour (invalid key, #3215); root cause is the shallow exposed uniquely via DT[TRUE]
-test(1542.08, DT$id, INT(1,2,6,4,5))
-test(1542.09, key(DT), "id")
-# future correct behaviour :
-# test(1542.10, DT$id, 1:5)
-# test(1542.11, key(DT), "id")
+# past wrong behaviour (invalid key, #3215); root cause is the shallow exposed uniquely via DT[TRUE]
+# test(1542.08, DT$id, INT(1,2,6,4,5))
+# test(1542.09, key(DT), "id")
+# current correct behaviour :
+test(1542.10, DT$id, 1:5)
+test(1542.11, key(DT), "id")
 
 # rest of #1130 - merge doesn't copy, instead uses joins without keys.
 set.seed(1L)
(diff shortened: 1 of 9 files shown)


OfekShilon:
This suggested fix includes an addition of an internal-only argument to `setDT`. Is this acceptable?

inst/tests/tests.Rraw
```diff
@@ -9011,59 +9011,60 @@ set.seed(123)
 dt = data.table(x1 = rep(letters[1:2], 6), x2 = rep(letters[3:5], 4), x3 = rep(letters[5:8], 3), y = rnorm(12))
 dt = dt[sample(.N)]
 df = as.data.frame(dt)
+setDT_noattr = function(x) setDT(x, set.allow.inplace.attrib=FALSE)
```
mattdowle, line 9014: IIUC this `setDT_noattr` has been added because `test()` considers `x` and `y` different when one has the new attribute but the other doesn't.  If so, please modify `test()` instead to ignore the new attribute; i.e. add `setattr(xc,"allow.assign.inplace",NULL); setattr(yc,"allow.assign.inplace",NULL)` near line `test.data.table.R:429`. Then revert these test changes, leaving a much smaller set of true test changes.

mattdowle (changes requested):

mattdowle:
I added WIP label so you can remove it when the changes requested are either done, or I misunderstood and you want to send the PR back to me.
One of my to-do filters is https://github.com/Rdatatable/data.table/pulls?q=is%3Apr+is%3Aopen+review%3Achanges-requested+-label%3AWIP+,  so removing WIP will add it there.

mattdowle:
> Is an additional setDT argument for internal-calls-only considered acceptable style?
It would be unusual and I'm not sure we have any other examples of that. But don't worry about that yet. I suspect this PR will morph quite a bit. Just taking it one step at a time. If we need a new internal-only version of `setDT()`, then what springs to mind on first thought is to create a new function (e.g. `setDT_internal()`) and use that internally, rather than an argument or an option switch.  But that would be a relatively easy and quick change to make, so lets leave that till later.

OfekShilon:
> It would be unusual and I'm not sure we have any other examples of that. But don't worry about that yet. I suspect this PR will morph quite a bit. Just taking it one step at a time. If we need a new internal-only version of `setDT()`, then what springs to mind on first thought is to create a new function (e.g. `setDT_internal()`) and use that internally, rather than a user-visible argument or an internal option switch. But that would be a relatively easy and quick change to make, so lets leave that till later, if possible.

I completely understand why you suspect this PR would keep changing - and I apologize for changing it twice already - but conversation with you guys is the only way to make it mold.  The internals (modifications to `Cassign()` ) are all in place, the discussion is only about the proper API to use it - and I'm not sure what is the proper "data.tableism" here. 
Do you consider the `Cassign` change acceptable?  
Since writing this, a colleague of mine asked to have a DT with full copy-on-modify semantics - so I now think to communicate with `setDT` via an option and not an argument, so advanced users (and not just internal functions) could choose so themselves.   Sounds acceptable?  If so, I'll fix this PR soon.

mattdowle:
@OfekShilon Don't worry - there is no friction due to the PR changing. We don't mind changes in a PR at all.

> Do you consider the Cassign change acceptable?

A change of that nature to Cassign is acceptable yes, although I haven't gone through the details of it yet.  I can see it is looking at a new attribute and changing behavior depending on the attribute on the data. That's fine in principle.

Thanks for reducing the number of test changes but what I can see now is the new `option(data.table.prevent.inplace=FALSE)`. We aren't allowing any new options like that which change code behavior because if users set that option, they would change the behavior of `setDT` inside packages too, likely inadvertently. In other words the `options()` mechanism in R is global: both user space is affected and inside packages using data.table too. So we only use `option()` to control display and output (e.g. see news item new feature 7 in this dev release), and temporarily (i.e. we remove the option after a few years) to make it easier for users to migrate their code when we make backwards incompatible changes (and even then it isn't perfect because of the global nature of the option). That's why I wrote above "rather than a user-visible argument or an internal option switch". What you've added is a user visible option which is very unlikely to fly. Instead, I suggested above creating a new internal function and then using that internally.

But first, can you explain why you didn't do what I asked in https://github.com/Rdatatable/data.table/pull/4978#discussion_r629666259 please?  I think I must have misunderstood, so I need you to explain that. That's the stage I'm at currently.

inst/tests/tests.Rraw
OfekShilon: @mattdowle  Sorry, back from a long hiatus. Per your request I reverted the setDT_noattr workaround and modified `test.data.table` logic. It came down to more than a simple additions of `setattr( .., "allow.assign.inplace")` because test() didn't apply this logic to _lists_ of data tables. 
The fix isn't the most general it can be (theoretically it needs to apply for arbitrary nesting of data.tables in lists) but it is good enough for the existing tests.

OfekShilon:
@mattdowle  Sorry for the long leave. Some explanations on the order of changes:
(1) You requested that I remove the `setDT_noattr` workaround and replace it with internal `test()` fixes.
(2) I added global option to bypass in-place assignment. This made the `setDT_noattr` workaround redundant and I removed it entirely. 
(3) You commented here that the global-option approach is not acceptable.
(4) I now reverted the global-option code, which made your `setDT_noattr` reservation relevant again.
(5) I also applied your review, deleted `setDT_noattr` and modified internal `test()` logic. Note that it wasn't as straightforward as originally thought (for the first time had to apply some logic to _lists_ of data.tables), but all tests now pass.

Hopefully this can be merged now.

OfekShilon:
@mattdowle  I resolved your review long ago, somehow I still see 'changes requested'. Am I missing something?

Do you consider this change satisfactory?  Can it be included in 1.15?

OfekShilon:
@mattdowle  Is there anything unclear, or otherwise missing, to merge this fix?

MichaelChirico:
@mattdowle I agree it's far too long since we've heard from you here. Marked as 'High' to make sure this doesn't escape notice as we close in on 1.14.6

OfekShilon:
This might have become hard to merge in the 1.5Y since suggesting it. Would it help to start over in a new PR? Perhaps with better explanations?

MichaelChirico:
@OfekShilon I merged to `master` locally, but I don't have push permissions to `IstraResearch`, could you grant them? thanks

OfekShilon:
> @OfekShilon I merged to `master` locally, but I don't have push permissions to `IstraResearch/data.table`, could you grant it? thanks

Thanks! I'm afraid I'm not allowed to grand access to my organization - just pushed a merge of `master` to this branch.

jangorecki:
I don't think comfortable for having such a significant change merged shortly before preparing major release. I am putting it for the next milestone, as currently the priority is to have master branch released on CRAN.

jangorecki:
Like most of other PR we will have to merge/rebase to master and resolve conflicts. That's the cost of not actively maintaining the project. What's positive that at least it won't be as difficult as if project would be actively maintained for 2.5 years, because there were not that many changes over this time.

MichaelChirico:
Finally read the issue & comments carefully enough to agree that yes, there's something wanting with the current code.

However, I _don't_ want `prevent.assign.inplace` to be `TRUE` by default. In my experience it's very common to do things like `setDT(foo())` where `foo()` is some other I/O function, e.g. `DBI::dbGetQuery()`.

Preventing assignment on objects like that will be extremely user-unfriendly IMO.

I wonder if there's some way to use R internals here -- we really care about the case of `setDT(x)` where `x` _is already bound to some variable_. I'm not sure this is possible/advisable.

A much simpler solution is to just add the argument and keep it back-compatible -- users with a codebase where the "leaky" behavior tends to bite should use a wrapper function, although I'm not totally opposed to adding an option (I know Matt pushed back on that idea in the earliest review).

That said, as Ofek is no longer working much on R, I wonder who might sponsor this work to get it over the finish line. For now I'll bump the milestone; please chime in if this PR is heavily affecting you as well.

yihui:
This problem affects a friend of mine. I don't have the capability to work on the code in this PR, but I'll be happy to sponsor the work financially if it helps.

PR #4980, opened 2021-05-07
nafill character, logical, etc.

### LINKED ISSUE #3992: nafill, setnafill for character, factor and other types

jangorecki:
closes #3992

### DIFF
--- a/src/nafill.c
+++ b/src/nafill.c
@@ -108,8 +108,8 @@ SEXP nafillR(SEXP obj, SEXP type, SEXP fill, SEXP nan_is_na_arg, SEXP inplace, S
   if (obj_scalar) {
     if (binplace)
       error(_("'x' argument is atomic vector, in-place update is supported only for list/data.table"));
-    else if (!isReal(obj) && !isInteger(obj))
-      error(_("'x' argument must be numeric type, or list/data.table of numeric types"));
+    else if (!isReal(obj) && !isInteger(obj) && !isLogical(obj) && !isFactor(obj) && !isString(obj))
+      error(_("'x' argument must be numeric/integer/logical/factor/character/integer64, or list/data.table of such types"));
     SEXP obj1 = obj;
     obj = PROTECT(allocVector(VECSXP, 1)); protecti++; // wrap into list
     SET_VECTOR_ELT(obj, 0, obj1);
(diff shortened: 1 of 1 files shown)


src/nafill.c
```diff
@@ -86,6 +86,28 @@ void nafillInteger64(int64_t *x, uint_fast64_t nx, unsigned int type, int64_t fi
   if (verbose)
     snprintf(ans->message[0], 500, "%s: took %.3fs\n", __func__, omp_get_wtime()-tic);
 }
+void nafillString(const SEXP *x, uint_fast64_t nx, unsigned int type, SEXP fill, ans_t *ans, bool verbose) {
+  double tic=0.0;
+  if (verbose)
+    tic = omp_get_wtime();
+  if (type==0) { // const
+    for (uint_fast64_t i=0; i<nx; i++) {
+      ((SEXP*)ans->char_v)[i] = x[i]==NA_STRING ? fill : x[i];
```
jangorecki, line 95: @mattdowle can we assign to character vector (`ans->char_v`) this way? or we must use `SET_STRING_ELT`? We are assigning character that must be already in the global cache. `char_v` is `void*` that is why cast to `SEXP*` here.

mattdowle: SET_STRING_ELT decrements the reference count on the old CHARSXP and increments the reference count on the CHARSXP replacing it.  So we have to use it to make sure those reference counts are maintained. Unless you know for sure that the old and new are both permanent in the global cache; e.g. NA_STRING being replaced by a `sym_` which we `install()` in init.c. But even then, it would be seen as quite risky as it would be 'behind-the-write-barrier'. Such a write may be caught by `rchk` but even if not, it's the reference counts on CHARSXP to be aware of.

jangorecki: Thanks, it is now much more clear. I am trying to change code to use `SET_STRING_ELT` but I am keep getting `SET_STRING_ELT() can only be applied to a 'character vector', not a 'builtin'`. Tried redefining `ans->char_v` type and casting but haven't made it working yet. Should I expect this to be a simple change?

mattdowle: @jangorecki Yes it should be simpler. If you can't work out where the `builtin` is coming from, then it could be a possible corruption. Try running the branch through the ASAN and rchk steps in CRAN_Release.cmd.  ASAN would need these lines to be covered to detect anything. Because of the reference counting, SET_STRING_ELT cannot be used inside parallel regions: try setting threads to 1 and seeing if the same error occurs to see if that might be in play.

jangorecki: parallel is escaped when there is any character column on input, so that should not be an issue.

src/nafill.c
```diff
@@ -86,6 +86,28 @@ void nafillInteger64(int64_t *x, uint_fast64_t nx, unsigned int type, int64_t fi
   if (verbose)
     snprintf(ans->message[0], 500, "%s: took %.3fs\n", __func__, omp_get_wtime()-tic);
 }
+void nafillString(const SEXP *x, uint_fast64_t nx, unsigned int type, SEXP fill, ans_t *ans, bool verbose) {
+  double tic=0.0;
+  if (verbose)
+    tic = omp_get_wtime();
+  if (type==0) { // const
```
MichaelChirico, line 93: would an `enum` be more natural for `type`? not really related to this PR, just a thought on reading

jangorecki: maybe but in this simple case it is probably more lines of code, it is still very clear here

src/nafill.c
jangorecki: the problem was fixed in this line
https://github.com/Rdatatable/data.table/pull/4980/commits/b24956495cb98e508fa13960c6d900ee4a6b5d1c#diff-f9176ef08eac951aa792d88fe2e2c6c3264fc1393b8836def832c33592a71932R192

MichaelChirico:
Hey @jangorecki can I check what's the status of this PR? What makes it "WIP"?

jangorecki:
Unfortunately cannot recall, but it is fat from ready to merge. No news entry, last commit says char not yet implemented.
Could be possibly taken over from here.
But no need to rush for 1.18.0, I will put it on the next milestone

MichaelChirico:
@jangorecki what's the status here? looking for someone else to take over?

jangorecki:
Yes

MichaelChirico:
Everything that was here is now merged; the TODO here was `setnafill()` support for `STRSXP`.

Closing this and marking for follow-up.

PR #4987, opened 2021-05-10
Minor changes to Programming vignette

avimallu:
@jangorecki, thank you for the effort into adding a programmable interface to `data.table` . I was going through the vignette and noticed some minor errors to language or typos that I believe should be edited. Please let me know if I should perhaps file another issue and then a PR, or whether just a PR is fine.

I have not made any changes to code, so there is no need of a NEWS item or a change to the contributors list.

Further, there is one small section that is going over my head, no matter how I read it (as well as the examples after it), so I would like to suggest a rephrase:
https://github.com/Rdatatable/data.table/blob/79c3d3e1d2b1842edc948f2d8c895fe45a9070ac/vignettes/datatable-programming.Rmd#L264

### DIFF
--- a/vignettes/datatable-programming.Rmd
+++ b/vignettes/datatable-programming.Rmd
@@ -193,7 +193,7 @@ DT[filter_col %in% filter_val,
 
 ### Substitute variables and character values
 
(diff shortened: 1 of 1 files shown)


jangorecki:
Thanks a lot. Yes, English is not my native language as we can see :) as for the sentence you asked, I recall it was difficult to express the logic that is why I provided examples that illustrated this behavior.

avimallu:
Closing in favour of #4988.

PR #4992, opened 2021-05-12
Allow env dot alias in `substitute2` and `[.data.table`

tonyfischetti:
This PR allows for the use of `.()` as an alias to `list()` in the `env` arguments to `substitute2` and `[.data.table`, as a convenience feature. Just like in the `i`, `j`, and `by` arguments.

If this is of interest, I can submit another pull request I pretty much already completed that allows the `.()` alias in `env` to handle non-named list elements in a manner similar to `as.data.table`.
It would allow code like this:
```{r}
DT = as.data.table(iris)
tmp = "Species"
DT[, .N, tmp, env=.(tmp)]
```
Let me know if that PR is also of interest. I can also, of course, submit it as an issue, but since it all depends on the viability of this PR, I though I'd mention it here first.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,7 +18,7 @@
     DT = data.table(A=1:3, B=letters[1:3])
     DT[A>3,   .(ITEM='A>3', A, B)]  # (1)
     DT[A>3][, .(ITEM='A>3', A, B)]  # (2)
-    # the above are now equivalent as expected and return:  
+    # the above are now equivalent as expected and return:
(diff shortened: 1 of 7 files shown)


jangorecki:
Please submit an issue proposing such change so we can discuss it there rather than in PR. Then eventually we can always re-open.

tonyfischetti:
Sorry, I'm used to the "shut-up-and-show-me-the-code" attitude of Linux kernel development. Thought it would be better to just submit a PR

PR #5000, opened 2021-05-15
Localize remaining loop variables

mattdowle:
A lot were already done. This clears up the remaining. I started with :
`grep -En "for [(][^ ]+[ ]*=" src/*.c`
and went from there.  The ones remaining are intentional; e.g. they use the ending loop counter value afterwards, and now we can see that more clearly.

It's not for speed (although it might help depending on compiler and options) but for i) easier reading without needing to scroll up to the declaration, and ii) restrict variable scope to prevent accidentally using or changing a variable in other parts of the code.

Also modernized `anyNA` in `frank.c`: integer64 tidy, took `STRING_ELT()` outside loop, and saved the repeated `ans[0]=true` in each case.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -149,7 +149,6 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   // NEW: cols argument to specify the columns to shallow copy on. If NULL, all columns.
   // called from alloccol where n is checked carefully, or from shallow() at R level
   // where n is set to truelength (i.e. a shallow copy only with no size change)
-  R_len_t i,l;
   int protecti=0;
   SEXP newdt = PROTECT(allocVector(VECSXP, n)); protecti++;   // to do, use growVector here?
   SET_ATTRIB(newdt, shallow_duplicate(ATTRIB(dt)));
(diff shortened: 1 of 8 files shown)


src/assign.c
```diff
@@ -173,21 +172,20 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
 
   SEXP names = PROTECT(getAttrib(dt, R_NamesSymbol)); protecti++;
   SEXP newnames = PROTECT(allocVector(STRSXP, n)); protecti++;
+  const int l = isNull(cols) ? LENGTH(dt) : length(cols);
   if (isNull(cols)) {
-    l = LENGTH(dt);
-    for (i=0; i<l; i++) SET_VECTOR_ELT(newdt, i, VECTOR_ELT(dt,i));
+    for (int i=0; i<l; ++i) SET_VECTOR_ELT(newdt, i, VECTOR_ELT(dt,i));
```
MichaelChirico, line 177: why change from R_len_t to int?

MichaelChirico:
good stuff! would it make sense to take the remaining loops with a comment and then add something to CRAN_release?

mattdowle:
On comment on the remaining ones, I thought about a comment at the time, but figured they're quite clear as intentional: the ones in `froll.c` have the `int i` on the line before the `for()`, for example, and the `i` is used quite close after the loop. But now you mention it, yes a comment on the remaining ones would be good. Either for convenience in the grep output where the comment marker would show, or so the grep could auto filter them (if we used a consistent comment text). And yes something to CRAN_release.

src/assign.c
mattdowle: I'm kind of fed up with `R_len_t`. It's point (I assume, because it was a good idea) was so that R could change its definition from `int` to `int64_t` (for example) under the covers without affecting package space. But that wasn't the direction taken by R (possibly because use of `R_len_t` wasn't/couldn't be enforced so lots of `int` existed in package space which would have broken if `R_len_t` was changed from `int` to `int64_t`) and we have to make that change anyway and look for the one letter difference between `R_len_t` and `R_xlen_t`. We're using `int` more often now than `R_len_t`, and certainly `int64_t` more than `R_xlen_t`. When I tried last time to change everything to int64_t, my brain and greps found it much easier looking at `int`.

MichaelChirico: I see, I had been thinking (based on pure speculation) `R_len_t` and `R_xlen_t` were meant to be there, but that their point was platform robustness (akin to `int_fast32_t` and so on). Glad to know your reasoning.

PR #5003, opened 2021-05-15
tag non-local loops with #skip_loop_scope

MichaelChirico:
Follow-up to #5000

### DIFF
--- a/.dev/CRAN_Release.cmd
+++ b/.dev/CRAN_Release.cmd
@@ -203,6 +203,10 @@ grep allocVector *.c | grep -v PROTECT | grep -v SET_VECTOR_ELT | grep -v setAtt
 grep coerceVector *.c | grep -v PROTECT | grep -v SET_VECTOR_ELT | grep -v setAttrib | grep -v return
 grep asCharacter *.c | grep -v PROTECT | grep -v SET_VECTOR_ELT | grep -v setAttrib | grep -v return
 
+# Enforce local scope for loop index (`for (int i=0; ...)` instead of `int i; for (i=0; ...)`)
+#   exceptions are tagged with #skip_loop_scope
+grep -En "for\s*[(]\s*[a-zA-Z0-9_]+\s*=.*[+][+]" src/*.c | grep -Fv "#skip_loop_scope"
+
 cd ..
 R
 cc(test=TRUE, clean=TRUE, CC="gcc-10")  # to compile with -pedandic -Wall, latest gcc as CRAN: https://cran.r-project.org/web/checks/check_flavors.html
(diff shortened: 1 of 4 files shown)


.dev/CRAN_Release.cmd
```diff
@@ -203,6 +203,10 @@ grep allocVector *.c | grep -v PROTECT | grep -v SET_VECTOR_ELT | grep -v setAtt
 grep coerceVector *.c | grep -v PROTECT | grep -v SET_VECTOR_ELT | grep -v setAttrib | grep -v return
 grep asCharacter *.c | grep -v PROTECT | grep -v SET_VECTOR_ELT | grep -v setAttrib | grep -v return
 
+# Enforce local scope for loop index (`for (int i=0; ...)` instead of `int i; for (i=0; ...)`)
+#   exceptions are tagged with #skip_loop_scope
+grep -En "for\s*[(]\s*[a-zA-Z0-9_]+\s*=.*[+][+]" src/*.c | grep -Fv "#skip_loop_scope"
+
```
mattdowle, line 209: Not sure why you added the `[+][+]` here. Making the grep more restrictive like that will increase the chance of false negatives. And as luck would have it, it does happen to miss this one (which my grep caught) :

`src/dogroups.c:120:  for (s = ATTRIB(SD); s != R_NilValue && TAG(s)!=R_RowNamesSymbol; s = CDR(s));`

For the same reason, I think my `[^ ]+` was better than your `[a-zA-Z0-9_]+`. Searching for a specific set of characters like that increases the small chance of false negatives. You're probably right that those are the only characters allowed in C variable names, but it's feasible another character is allowed, and I'm having to thing about it. Compared with my `[^ ]`+ which has less chance of false negatives.

Also, how about the marker being `#loop_counter_not_local_scope_ok`. Just to be a bit more self describing.

mattdowle (changes requested):

.dev/CRAN_Release.cmd
MichaelChirico: TBH I was confused myself when looking at `[^ ]+` so I went with the more specific regex `[a-zA-Z0-9_]+` -- to be fair I have looked at the GNU manual for this myself recently so maybe I'm more comfortable with it:

https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Identifiers

(technically `[a-zA-Z_][a-zA-Z0-9_]*` would be more accurate)

Including `[+][+]` was in fact specifically to exclude the `s = ATTRIB(SD)` loop, I should have noted that up front. That loop looked like it's not in the spirit of what we're trying to identify with this check, but maybe that's not true for all similar loops... removing it now.

Thanks for the suggestion on the name. I should really stop trying to economize on space for this kind of thing :sweat_smile:

mattdowle: I prefer `[^ ]+` for the reasons I said, but I'll move on.

mattdowle (approved):

PR #5010, opened 2021-05-18
programming vignette tweak

jangorecki:


tonyfischetti:
How come it's "Starting from version 1.15.0" and not "Starting from version 1.14.1"?

jangorecki:
Odd minor version numbers are development version that are not being released to CRAN. In Cran you will find only even minor versions of data.table. So the next CRAN version will be either 1.15.0 or 1.14.2. I bet for 1.15.0.

tonyfischetti:
Got it! Thanks for clearing that up :)

mattdowle:
We typically bump the 2nd number when there's a potential-breaking-changes section. The way current dev is looking, it would be released as 1.14.2.  Putting 1.14.2 in the vignette would hedge the bet.

PR #5016, opened 2021-05-20
combine consecutive stopifnot calls

MichaelChirico:
reduce visual clutter of multiple calls


mattdowle:
I'm ok to merge this one, but I don't really agree there was 'visual clutter' before. So I'd prefer not to see a whole lot more of PRs like this one, or greps or linters about this. Maybe this is a one off unusual one, but just saying just in case it's not. It's not like a bunch of checks on arguments ranks highly on lines we need to scrutinize. I see these lines as similar to long error messages which scroll off my screen to the right out of the way which I like very much because I like the long detail of long messages out of the way of the actual logic (I can scroll over I want, or choose to set my editor to wrap if I want). Not that you've reduced line length in this PR, though, so I'm fearing that might be on your agenda. 3 stopifnot's instead of 1 wasn't clutter. At least, relative to the project's to do list.

PR #4278, opened 2020-03-04
A linter to continuously check code quality in CI

MichaelChirico:
Closes #4190

TODO:

 - [x] linter for `gsub`/`grep`/etc calls that can use `fixed=TRUE`
 - [x] linter for `paste` calls using `sep=""`
 - [x] Style Guide Lint: Use `=` for assignment
 - [x] Style Guide Lint: 2-space indentation (no tabs anywhere)
 - [x] Style Guide Lint: No trailing whitespace
 - [x] Style Guide Lint: No space around `=` in function calls
 - [x] Style Guide Lint: Spacing around `if`: `if (condition) {`
 - [x] New lint: don't quote keywords unless required (see e.g. #4924)
 - [x] New lint: Proper use of `&&` / `||` where possible
 - [ ] New lint: Block `paste(x, collapse=',')` in favor of `brackify(x)`
 - [x] New lints: `any(is.na())`/`any(is.duplicated())`
 - [x] New lint: Negate after aggregation: `any(!x)` -> `!all(x)`, `all(!x)` -> `!any(x)`
 - [ ] New lint: `stop()`/`warning()`/`message()` --> `stopf()`/`warningf()`/`messagef()`
 - [x] New lint: `test(SYMBOL, ...)` or `test(SYMBOL + ....)` linted (see #6041 for lint logic)
 - [x] Incorporate `# nolint` logic
 - [x] Include `grep`s of `src` (still as regex, from R)
 - [x] New src regex: `++i` instead of `i++` in all loops
 - [x] Apply lints to existing code
 - [x] Add to CI

PR #5017, opened 2021-05-20
Flip any(!x) to !all(x) and all(!x) to !any(x)

MichaelChirico:
Better to negate only once. Also added to #4278


mattdowle:
This though, is great.

Issue #5042, opened 2021-06-11
dplyr::arrange interfering with data.table's auto-indexing 


Waldi73:
# `Minimal reproducible example`
This is a follow-up on this [StackOverflow question/answer](https://stackoverflow.com/a/67943599/13513328).
`dplyr::arrange` seems to interfere with auto-indexing, leading to unexpected wrong results.
A work-around is to disable auto-indexing with `options(datatable.auto.index = FALSE)`. 
As this issue is quite tricky, a warning, a mention to it in FAQ or a protection against this happening would be useful! 
```
library(dplyr); 
library(data.table)
DT <-
  fread(
    "iso3c  country income
MOZ Mozambique  LIC
ZMB Zambia  LMIC
ALB Albania UMIC
MOZ Mozambique  LIC
ZMB Zambia  LMIC
ALB Albania UMIC
"
  )

codes <- c("ALB", "ZMB")
options(datatable.auto.index = TRUE) # Default
DT <- distinct(DT) %>%   as.data.table()

# Index creation because %in% is used for the first time
DT[iso3c %in% codes,verbose=T]

# Index mixed up by arrange
DT <- DT %>% arrange(iso3c) %>% as.data.table()

# this is wack because data.table uses the old index where row were rearranged:
DT[iso3c %in% codes,verbose=T]
#>    iso3c country income
#> 1:   ALB Albania   UMIC

# this works because (...) prevents the parser to use auto-index
DT[(iso3c %in% codes)]
#>    iso3c country income
#> 1:   ALB Albania   UMIC
#> 2:   ZMB  Zambia   LMIC
```

# `Output of sessionInfo()`
```` 
R version 4.1.0 (2021-05-18)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18362)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.14.0 dplyr_1.0.6

MichaelChirico:
could you report this with dplyr as well? at first glance it seems like a bug on that end

avimallu:
[...]

PR #5043, opened 2021-06-12
as.df removes index

jangorecki:
#5042


mattdowle:
Yes completely agree with this, especially when it's seen in context just after `sorted` is removed too.
So will merge this PR as-is (without news item) with the issue left open in this milestone. Seems like you want to confirm this fixes it first before writing news item and closing the issue, which would need folks to test using the dev version built off main branch.

mattdowle:
I added NEWS item because this is what's causing NMdata to fail its tests, #5075.  Easy one to resolve as a PR to NMdata which I'll do and point to the news item.
I see it reported in #5042 that this PR doesn't fix the issue. It's tagged this milestone so I'll look at that again before release.

mattdowle:
Looks like PR #4893 was earlier and did `setDF` too. So I updated news item 27 in v1.14.2 to reflect that.

PR #4350, opened 2020-04-03
Fix segfaults when assigning to list column

tlapak:
Closes #4166 
Closes #4667
Closes #4678
Closes #4729

The issue there was unrelated to the unlisting. A minimal example that also segfaults is
```R
dt <- data.table(a=list(1:2, 3, 4))
dt[, a:=1:4]
```
Turns out when LHS is a list column and RHS isn't, no length checks were being made. I am slightly surprised by @MichaelChirico's result which was maybe even intended.

While testing my fix I noticed that subassigning a vector of correct length also segfaults:
```R
dt <- data.table(a=list(1:2, 3, 4))
dt[1:2, a:=1:2]
```
I see three possible ways of handling this:

1. Error out with a coercion error
2. Treat RHS as length 1 and replicate
3. Coerce to list by wrapping elements of RHS in list()

I opted for 3. because why not. It seems the most intuitive to me besides 1. But happy to change this if people disagree. Required 'some' abuse of the BODY macro in memrecycle...

PR #5047, opened 2021-06-17
Don't call copyMostAttrib on ScalarLogical result

mattdowle:
Follow up to #4350

The first commit "reprex" in this PR reproduces the problem. I'm using R 4.0.3 currently. Running `R CMD check` or `test.data.table()` is necessary to make 2195.[3-8] fail because test 2190.6 needs to run first. When I ran tests 2195 at the dev prompt they always passed and I couldn't reproduce.

Test 2190.6 is subassigning a logical vector with an attribute attached to an existing list column :
```R
DT = data.table(a=list(1:2, 3, 4))
DT[1:2, a:=structure(c(TRUE, FALSE), att='t')]
```

This ended up attaching the `"att"` attribute with value `"t"` to R's internal global FALSE constant. This caused the `!identical(incomparables, FALSE)` in duplicated.R to fail. I debugged and confirmed that `print(FALSE)` showed that FALSE had indeed received the attribute whereas `incomparables` did not have the attribute. Presumably TRUE was also receiving the attribute but only FALSE was showing up so far in tests 2195.* on the FALSE value.

The error shows up as : 
```
Running test id 2195.3          Test 2195.3 produced 1 errors but expected 0
Expected: 
Observed: argument 'incomparables != FALSE' is not used (yet)
```

In #4595 I worked around the issue by changing duplicated.R to use `!isFALSE(incomparables)` instead of `!identical(incomparables, FALSE)` to ignore any attributes attached to FALSE.

This PR now fixes the root cause. `ScalarLogical(false)` returns R's internal global FALSE value; `allocVector()` has to be used to return a new allocation of a length-1 logical value and a new allocation is necessary if attributes are to be attached. After this PR, duplicated.R could use `!identical(incomparables, FALSE)` again, but having gone through this, I'm now thinking that ignoring attributes attached to FALSE is more robust anyway so will leave the `!isFALSE()` in place in duplicated.R.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -1071,16 +1071,21 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
         BODY(SEXP, &, SEXP, val, SET_VECTOR_ELT(target, off+i, cval))
     } else {
       switch (TYPEOF(source)) {
-      // no protect of CAST needed because SET_VECTOR_ELT protects it, and it can't get released by copyMostAttrib or anything else inside BODY
-      // copyMostAttrib is appended to CAST so as to be outside loop
-      case RAWSXP:  BODY(Rbyte,    RAW,        SEXP, ScalarRaw(val);    copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case LGLSXP:  BODY(int,      INTEGER,    SEXP, ScalarLogical(val);copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case INTSXP:  BODY(int,      INTEGER,    SEXP, ScalarInteger(val);copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case REALSXP: BODY(double,   REAL,       SEXP, ScalarReal(val);   copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case CPLXSXP: BODY(Rcomplex, COMPLEX,    SEXP, ScalarComplex(val);copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case STRSXP:  BODY(SEXP,     STRING_PTR, SEXP, ScalarString(val); copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
+      // allocVector instead of ScalarLogical to avoid copyMostAttrib on R's internal global TRUE/FALSE values; #4595. Then because
+      //   ScalarInteger may now or in future R also return R internal global small integer constants, the same for that. Then
+      //   because we do that here for logical and integer, use allocVeector too for the other types to follow the same pattern and possibly
+      //   in future R will also have some global constants for those types too.
+      // the UNPROTECT can be at the end of the CAST before the SET_VECTOR_ELT, because SET_VECTOR_ELT will protect it and there's no other code inbetween
+      // the PROTECT is now needed because of the call to LOGICAL() which could feasibly gc inside it.
+      // copyMostAttrib is inside CAST so as to be outside loop.  See the history in #4350 and its follow up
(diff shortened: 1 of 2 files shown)


src/assign.c
```diff
@@ -1071,16 +1071,21 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
...
-      case LGLSXP:  BODY(int,      INTEGER,    SEXP, ScalarLogical(val);copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case INTSXP:  BODY(int,      INTEGER,    SEXP, ScalarInteger(val);copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case REALSXP: BODY(double,   REAL,       SEXP, ScalarReal(val);   copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case CPLXSXP: BODY(Rcomplex, COMPLEX,    SEXP, ScalarComplex(val);copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
-      case STRSXP:  BODY(SEXP,     STRING_PTR, SEXP, ScalarString(val); copyMostAttrib(source,cval), SET_VECTOR_ELT(target,off+i,cval))
+      // allocVector instead of ScalarLogical to avoid copyMostAttrib on R's internal global TRUE/FALSE values; #4595. Then because
+      //   ScalarInteger may now or in future R also return R internal global small integer constants, the same for that. Then
+      //   because we do that here for logical and integer, use allocVeector too for the other types to follow the same pattern and possibly
+      //   in future R will also have some global constants for those types too.
```
MichaelChirico, line 1077: btw I believe base::pi counts as just such a constant for REALSXP already

src/assign.c
```diff
@@ -1071,16 +1071,21 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
...
+      // the PROTECT is now needed because of the call to LOGICAL() which could feasibly gc inside it.
+      // copyMostAttrib is inside CAST so as to be outside loop.  See the history in #4350 and its follow up
+      case RAWSXP:  BODY(Rbyte,    RAW,        SEXP, PROTECT(allocVector(RAWSXP, 1));RAW(cval)[0]=val;copyMostAttrib(source,cval);UNPROTECT(1),             SET_VECTOR_ELT(target,off+i,cval))
+      case LGLSXP:  BODY(int,      LOGICAL,    SEXP, PROTECT(allocVector(LGLSXP, 1));LOGICAL(cval)[0]=val;copyMostAttrib(source,cval);UNPROTECT(1),         SET_VECTOR_ELT(target,off+i,cval))
+      case INTSXP:  BODY(int,      INTEGER,    SEXP, PROTECT(allocVector(INTSXP, 1));INTEGER(cval)[0]=val;copyMostAttrib(source,cval);UNPROTECT(1),         SET_VECTOR_ELT(target,off+i,cval))
+      case REALSXP: BODY(double,   REAL,       SEXP, PROTECT(allocVector(REALSXP, 1));REAL(cval)[0]=val;copyMostAttrib(source,cval);UNPROTECT(1),           SET_VECTOR_ELT(target,off+i,cval))
+      case CPLXSXP: BODY(Rcomplex, COMPLEX,    SEXP, PROTECT(allocVector(CPLXSXP, 1));COMPLEX(cval)[0]=val;copyMostAttrib(source,cval);UNPROTECT(1),        SET_VECTOR_ELT(target,off+i,cval))
```
MichaelChirico, line 1085: I'm not sure, does assigning like this work as expected for complex? or do we have to assign the real/imaginary parts individually?

src/assign.c
mattdowle: interesting, where did you see that?

MichaelChirico: hmm on second thought I think I misunderstood. I was analogizing to base::T which isn't the same issue

src/assign.c
mattdowle: yes `Rcomplex` is a struct and in C you can assign structs; e.g. `grep NA_CPLX data.table/src/*`.

PR #5049, opened 2021-06-18
Changes buffer size for writing header in fwrite

### LINKED ISSUE #5048: fwrite gzip on OpenBSD error
On OpenBSD 6.9, released May 1, 2021, writing gzipped csv fails :

```
> library(data.table)
> test.data.table()
getDTthreads(verbose=TRUE):
test.data.table() running: /usr/local/lib/R/library/data.table/tests/tests.Rraw.bz2 

**** Suggested package bit64 is not installed. Tests using it will be skipped.
**** Suggested package xts is not installed. Tests using it will be skipped.
**** Suggested package nanotime is not installed. Tests using it will be skipped.
**** Suggested package R.utils is not installed. Tests using it will be skipped.
**** Suggested package yaml is not installed. Tests using it will be skipped.

Test 1658.53 produced 1 errors but expected 0
Expected: 
Observed: Compress gzip error: -9
Test 1760 not run because this session either has no OpenMP or has been limited to one thread (e.g. under UBSAN and ASAN)

Fri Jun 18 15:21:25 2021  endian==little, sizeof(long double)==16, longdouble.digits==64, sizeof(pointer)==8, TZ==unset, Sys.timezone()=='Europe/Paris', Sys.geTest 1658.53 produced 1 errors but expected 0
Expected: 
Observed: Compress gzip error: -9tlocale()=='C', l10n_info()=='MBCS=FALSE; UTF-8=FALSE; Latin-1=FALSE', getDTthreads()=='This installation of data.table has not been compiled with OpenMP support.; omp_get_num_procs()==1; R_DATATABLE_NUM_PROCS_PERCENT==unset (default 50); R_DATATABLE_NUM_THREADS==unset; R_DATATABLE_THROTTLE==unset (default 1024); omp_get_thread_limit()==1; omp_get_max_threads()==1; OMP_THREAD_LIMIT==unset; OMP_NUM_THREADS==unset; RestoreAfterFork==true; data.table is using 1 threads with throttle==1024. See ?setDTthreads.', zlibVersion()==1.2.3 ZLIB_VERSION==1.2.3
Error in test.data.table() : 
  1 error out of 9060. Search tests/tests.Rraw.bz2 for test number 1658.53.

```

Reproducible example:
```
> library(data.table)
> sessionInfo()
R version 4.0.5 (2021-03-31)
Platform: x86_64-unknown-openbsd6.9 (64-bit)
Running under: OpenBSD 6.9 GENERIC.MP#3 amd64

Matrix products: default
BLAS:   /usr/local/lib/R/lib/libRblas.so.37.0
LAPACK: /usr/local/lib/R/lib/libRlapack.so.37.0

locale:
[1] C

attached base packages:

other attached packages:
[1] data.table_1.14.0

[1] compiler_4.0.5
> a = data.table(i=1:100)
> fwrite(a, "test.csv.gz")
Error in fwrite(a, "test.csv.gz") : Compress gzip error: -9
Execution halted
```

philippechataignon:
This PR is an attempt to Resolve #5048.

Actually, buffer size for writing header only uses headerLen. headerLen is computed accurately and deflateBound returns a buffer size witch guarantees an one pass compression. But that's not true with OpenBSD which uses on old zlib version.

Here, we use by default the buffer size used for rows from the buffer size parameter (buffMB) and, if we have a very long header, we use headerLen.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -159,6 +159,8 @@
 
 7. Grouping by a `list` column has its error message improved stating this is unsupported, [#4308](https://github.com/Rdatatable/data.table/issues/4308). Thanks @sindribaldur for filing, and @michaelchirico for the PR. Please add your vote and especially use cases to the [#1597](https://github.com/Rdatatable/data.table/issues/1597) feature request.
 
+8. OpenBSD 6.9 released May 2021 apparently uses a an old version of zlib (v1.2.3 from 2005) which induces `Compress gzip error: -9` from `fwrite()`, [#5048](https://github.com/Rdatatable/data.table/issues/5048). Thanks to Philippe Chataignon for investigating and for the PR which attempts a solution.
+
 
 # data.table [v1.14.0](https://github.com/Rdatatable/data.table/milestone/23?closed=1)  (21 Feb 2021)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -159,7 +159,7 @@
 
 7. Grouping by a `list` column has its error message improved stating this is unsupported, [#4308](https://github.com/Rdatatable/data.table/issues/4308). Thanks @sindribaldur for filing, and @michaelchirico for the PR. Please add your vote and especially use cases to the [#1597](https://github.com/Rdatatable/data.table/issues/1597) feature request.
 
-8. OpenBSD 6.9 released May 2021 apparently uses a an old version of zlib (v1.2.3 from 2005) which induces `Compress gzip error: -9` from `fwrite()`, [#5048](https://github.com/Rdatatable/data.table/issues/5048). Thanks to Philippe Chataignon for investigating and for the PR which attempts a solution.
+8. OpenBSD 6.9 released May 2021 apparently uses a 16 year old version of zlib (v1.2.3 from 2005) which induces `Compress gzip error: -9` from `fwrite()`, [#5048](https://github.com/Rdatatable/data.table/issues/5048). Thanks to Philippe Chataignon for investigating and for the PR which attempts a solution.
 
 
 # data.table [v1.14.0](https://github.com/Rdatatable/data.table/milestone/23?closed=1)  (21 Feb 2021)
--- a/src/fwrite.c
+++ b/src/fwrite.c
@@ -738,14 +738,9 @@ void fwriteMain(fwriteMainArgs args)
           free(buff);                                    // # nocov
           STOP(_("Can't allocate gzip stream structure"));  // # nocov
         }
-        // accurate size
-        size_t zbuffSize = deflateBound(&stream, headerLen);
-        // alternate size to resolve #5048 (old openbsd zlib)
-        // this buffsize is the same used for writing rows
-        size_t alt_zbuffSize = deflateBound(&stream, buffSize);
-        // takes the max size in case of very long header
-        if (alt_zbuffSize >  zbuffSize)
-             zbuffSize = alt_zbuffSize;
+        // by default, buffsize is the same used for writing rows (#5048 old openbsd zlib)
+        // takes the max with headerLen size in case of very long header
+        size_t zbuffSize = deflateBound(&stream, headerLen > buffSize ? headerLen : buffSize);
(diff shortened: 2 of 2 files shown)


mattdowle:
Many thanks! If possible it would be great if you could raise an issue to the OpenBSD team (maybe they are not aware it is using a 16-year version of zlib) and then we can place a link to that question/report in our news item so that readers of our news can be reassured that it has been reported.

philippechataignon:
Thanks Matt to merge this PR without waiting my answer. I confirm this PR fixes the problem. I will send a issue to OpenBSD team.

philippechataignon:
Here is report on openbsd-misc list:
https://marc.info/?l=openbsd-misc&m=162445581925627&w=1

mattdowle:
Thanks @philippechataignon. But if I were them I would be unclear reading that. What I had in mind was more like:
    "Hi. Is it intentional that OpenBSD 6.9, released May 2021 uses a 16 year old version of zlib (v1.2.3; July 2005) when the latest version is v1.2.11 (Jan 2017) is available?"
Could you post a follow a follow up please and we'll link to your follow up post.  Although, perhaps OpenBSD team will miss the reply if it's in the same thread.  So a new thread would avoid that problem.

philippechataignon:
It's an old version but they applied regularly patches as you can see here : https://cvsweb.openbsd.org/src/lib/libz/?sortby=date#dirlist : it's not a real 1.2.3. For example of asking a new zlib version, you can see this answer : https://marc.info/?l=openbsd-misc&m=141066380405272&w=2. OpenBSD system is very conservative and I'm not sure they will jump to 1.7.11. 
Perhaps, if you disagree, you can send your message above as lead maintainer in a new thread and link it in news. Not a problem for me.

mattdowle:
> It's an old version but they applied regularly patches as you can see here : https://cvsweb.openbsd.org/src/lib/libz/?sortby=date#dirlist : it's not a real 1.2.3.

It's sort of a fork of zlib then but with no new name?

> For example of asking a new zlib version, you can see this answer : https://marc.info/?l=openbsd-misc&m=141066380405272&w=2

The subject was "no zlib from zlib.net" (where the word 'no' is the first word) so after reading that exchange a few times I suspect the responder may have been frustrated by the unhelpful and incorrect subject and the comparison made to Ubuntu and then responded bluntly to correct the incorrect 'no' and the reply ended up along the lines of "we in fact have v1.2.3 from 2005". I guess that was frustrating to the OP because they knew that and it was 2014. When busy people in a dev team scan a mailing list, each subject conveys how much time each OP took to make a good subject and therefore how much respect each OP has for the responders they are seeking a reply from.

> OpenBSD system is very conservative and I'm not sure they will jump to 1.2.11. Perhaps, if you disagree, you can send your message above as lead maintainer in a new thread and link it in news. Not a problem for me.

Essentially, I have complained about OpenBSD in our news item. It currently reads as follows :

> OpenBSD 6.9 released May 2021 apparently uses a 16 year old version of zlib (v1.2.3 from 2005) which induces Compress gzip error: -9 from fwrite(), #5048. Thanks to Philippe Chataignon for investigating and for the PR which attempts a solution.

I'm glad I used the word 'apparently' as indeed it seems 'it's not really 1.2.3'. We'll continue to revise this news item before release. It is frustrating to me that we are spending time fixing problems caused by new releases (OpenBSD 6.9 was released last month May 2021) that use 16 year old versions of a wildly popular and depended on library such as zlib.

Here's my attempt then :  https://marc.info/?l=openbsd-misc&m=162455479311886&w=1
* version numbers in title
* "16 year old" in title
* "please" in title
* First words of body are "is it", a question.
* link to background (this PR where I'm writing now)
* very short post; as short as I could make it. I spent time on making it short and removing superfluous information.
* "intentional or is there any good reason" is truly what I'm asking.

rdm:
This looks like it had been a protocol / format change, though one which apparently was obscure enough that many people did not notice.

But that makes it worth careful inspection, and the change in the compressed representation should be well documented.

PR #5058, opened 2021-06-30
proper handling list-like objects in env arg #5057

### LINKED ISSUE #5057: providing data.table with character column to `env = list(...)` gives error?
thank you *very* much for `env = ...`.  it makes programming much easier.  with the current code, however, if i have a data.table in `env = list(...)` which has a `character` column, i seem to get errors.

````
require(data.table)
# this works
substitute2(dt, env=list(dt = data.table(a=1:9, b=1:9)))
# this gives an error
substitute2(dt, env=list(dt = data.table(a=1:9, b=as.character(1:9))))
````
the error message is
````
Error in `[.data.table`(x, to.name) :
  i evaluates to a logical vector length 2 but there are 9 rows. Recycling of logical i is no longer allowed as it hides more bugs than is worth the rare convenience. Explicitly use rep(...,length=.N) if you really need to recycle.
````

(if the error itself is deemed acceptable, maybe the error message could give a better indication of what the programmer is doing wrong?)

btw, in case this helps: if i use `1:2` rather than `1:9` 
````
substitute2(dt, env=list(dt = data.table(a=1:2, b=as.character(1:2))))
````
the error is different:
````
Error in `[<-.data.table`(`*tmp*`, to.name, value = list(a = `2`, b = `2`)) :
  type 'symbol' cannot be coerced to 'integer'
````

`#` `Output of sessionInfo()`
````
> sessionInfo()
R version 4.1.0 (2021-05-18)
Running under: Arch Linux

Matrix products: default
BLAS:   /usr/lib/libopenblasp-r0.3.15.so
LAPACK: /usr/lib/liblapack.so.3.9.1

locale:
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=en_US.UTF-8
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_US.UTF-8

attached base packages:

other attached packages:
[1] data.table_1.14.1

````

### PR DESCRIPTION
**Pull request by jangorecki**
no news item because it is feature not yet released to CRAN
closes #5057

### DIFF
--- a/R/programming.R
+++ b/R/programming.R
@@ -6,6 +6,9 @@ rm.AsIs = function(x) {
   oldClass(x) = cl[cl!="AsIs"]
   x
 }
+only.list = function(x) {
+  identical(class(x), "list")
+}
 list2lang = function(x) {
   if (!is.list(x))
     stop("'x' must be a list")
@@ -22,7 +25,7 @@ list2lang = function(x) {
     x[to.name] = lapply(x[to.name], as.name)
   }
   if (isTRUE(getOption("datatable.enlist", TRUE))) { ## recursively enlist for nested lists, see note section in substitute2 manual
-    islt = vapply(x, is.list, FALSE)
+    islt = vapply(x, only.list, FALSE) #5057 nested DT that inherits from a list must not be turned into list call
     to.enlist = !asis & islt
     if (any(to.enlist)) {
       x[to.enlist] = lapply(x[to.enlist], enlist)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: R/programming.R [hunk 1]
```diff
@@ -22,7 +25,7 @@ list2lang = function(x) {
     x[to.name] = lapply(x[to.name], as.name)
   }
   if (isTRUE(getOption("datatable.enlist", TRUE))) { ## recursively enlist for nested lists, see note section in substitute2 manual
-    islt = vapply(x, is.list, FALSE)
+    islt = vapply(x, only.list, FALSE) #5057 nested DT that inherits from a list must not be turned into list call
```
MichaelChirico, line 28: nit: `vapply_1b` :upside_down_face:

R/programming.R
```diff
@@ -50,7 +53,7 @@ substitute2 = function(expr, env) {
     # null is fine, will be escaped few lines below
   } else if (is.environment(env)) {
     env = as.list(env, all.names=TRUE, sorted=TRUE)
-  } else if (!is.list(env)) {
+  } else if (!only.list(env) && !(is.AsIs(env) && only.list(rm.AsIs(env)))) {
```
MichaelChirico, line 56: is this part about `AsIs` being tested? Not clear from a glance at the tests

jangorecki: Some tests failed without this part here

PR #5064, opened 2021-07-04
Properly handle `cols` argument of `measure()`

### LINKED ISSUE #5063: Odd behaviour of `melt()` when using the `cols` argument in `measure()`
`measure()` is a great enhancement to `melt()`  - thanks a lot for this!

However, while experimenting with the new `measure()` function when trying to find an answer for [this question on SO](https://stackoverflow.com/q/68233050/3817004) I came across an odd behaviour of `melt()` when the `cols` argument is used with `measure()`:

```
df <- fread("Country    AXOX1991    AXOX1992    BXOX1991    BXOX1992    CXOX1991    CXOX1992
Afghanistan 1   2   3   4   5   6
USA 6   5   4   3   2   1")
```

This gives the expected result (note that the `AXO` columns intentionally have be left out of the reshape):

```
melt(df,  measure.vars = measure(var, year, 
                                 pattern = "([BC]XO)X(\\d{4})"))
```
```
       Country AXOX1991 AXOX1992 var year value
1: Afghanistan        1        2 BXO 1991     3
2:         USA        6        5 BXO 1991     4
3: Afghanistan        1        2 BXO 1992     4
4:         USA        6        5 BXO 1992     3
5: Afghanistan        1        2 CXO 1991     5
6:         USA        6        5 CXO 1991     2
7: Afghanistan        1        2 CXO 1992     6
8:         USA        6        5 CXO 1992     1
```

Now, when I include `cols` in the call to `measure()`

```
(cols <- names(df)[4:7])

[1] "BXOX1991" "BXOX1992" "CXOX1991" "CXOX1992"

melt(df,  measure.vars = measure(var, year, cols = cols,
                                 pattern = "([BC]XO)X(\\d{4})"))
```

I get an unexpected and obviously wrong result and a warning:

```
   BXOX1992 CXOX1991 CXOX1992 var year       value
1:        4        5        6 BXO 1991 Afghanistan
2:        3        2        1 BXO 1991         USA
3:        4        5        6 BXO 1992           1
4:        3        2        1 BXO 1992           6
5:        4        5        6 CXO 1991           2
6:        3        2        1 CXO 1991           5
7:        4        5        6 CXO 1992           3
8:        3        2        1 CXO 1992           4
```

> Warning message:
> In melt.data.table(df, measure.vars = measure(var, year,  :
>   'measure.vars' [Country, AXOX1991, AXOX1992, BXOX1991, ...] are not all of the same type. By order of hierarchy, the molten data value column will be of type 'character'. All measure variables not of type 'character' will be coerced too. Check DETAILS in ?melt.data.table for more on coercion.
> 

The documentation says on the `pattern` argument:

> Perl-compatible regex with capture groups to match to `cols`.

Perhaps, I have completely misunderstood the meaning of the `cols` argument. Unfortunately, I have found no example of the proper use of `cols` neither in the documentation of `measure()` nor in the reshape vignette.

--------------

Even stranger, when using a function to create the _character vector of column names_ in the call, I get an error message right away:

```
melt(df,  measure.vars = measure(var, year, cols = names(df)[4:7],
                                 pattern = "([BC]XO)X(\\d{4})"))
``` 

> Error in (function (fun.list, sep = "_", pattern, cols, multiple.keyword = "value.name",  : 
>   cols must be a character vector of column names

`#` `Output of sessionInfo()`

```
R version 4.1.0 (2021-05-18)

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] data.table_1.14.1

```

tlapak:
Closes #5063 

There are two separate issues reported here. One results from `col` not being properly evaluated at any point. The other from the fact that measure and melt assume that measure was passed the complete list of column names. (`measure` returns an index into `cols` and `melt` applies that to `names(data)`)

The fix feels slightly hacky but seems to be the simplest way of letting `measure()` know about the full list of column names without exposing this interaction to the user. I also think we should check that `cols` is only passed a subset of column names.

Todos:

- [ ] Currently fails test 2183.00020 because that call contains two errors and I'm now checking `is.character(cols)` further up
- [ ] Also fails 2183.78 because I use a slightly different error message for the same test which is now combined with checking that `cols` is a subset of `names(data)`
- [ ] Unit tests
- [ ] News item

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -128,13 +128,23 @@ eval_with_cols = function(orig_call, all_cols) {
   })
   if (!is.primitive(fun)) {
     named_call = match.call(fun, orig_call)
-    if ("cols" %in% names(formals(fun)) && !"cols" %in% names(named_call)) {
-      named_call[["cols"]] = all_cols
-    } else if (is.call(named_call[["cols"]])) {
-      named_call[["cols"]] = eval(named_call[["cols"]], parent)
+    offsets = NULL
+    if ("cols" %in% names(formals(fun))) {
+      if (!"cols" %in% names(named_call)) {
+        named_call[["cols"]] = all_cols
+      } else if (is.language(named_call[["cols"]])) {
+        named_call[["cols"]] = eval(named_call[["cols"]], parent)
+      }
+      # not strictly needed when cols == all_cols, which is probably the most common case
+      # but overhead is minimal and this simplifies control flow
+      offsets = which(all_cols %in% named_call[["cols"]])
     }
     named_call[[1L]] = fun
-    eval(named_call, parent)
+    ans = eval(named_call, parent)
+    if (!is.null(offsets)) {
+      setattr(offsets[ans], "variable_table", attr(ans, "variable_table"))
+    }
+    return(ans)
   }
 }
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -133,11 +133,14 @@ eval_with_cols = function(orig_call, all_cols) {
       if (!"cols" %in% names(named_call)) {
         named_call[["cols"]] = all_cols
       } else if (is.language(named_call[["cols"]])) {
-        named_call[["cols"]] = eval(named_call[["cols"]], parent)
+        cols = named_call[["cols"]] = eval(named_call[["cols"]], parent)
+      }
+      if (!all(cols %in% all_cols)) {
+        stop("cols must be a subset of the column names of data")
       }
       # not strictly needed when cols == all_cols, which is probably the most common case
       # but overhead is minimal and this simplifies control flow
-      offsets = which(all_cols %in% named_call[["cols"]])
+      offsets = which(all_cols %in% cols)
     }
     named_call[[1L]] = fun
     ans = eval(named_call, parent)
--- a/R/utils.R
+++ b/R/utils.R
@@ -130,6 +130,8 @@ eval_with_cols = function(orig_call, all_cols) {
     named_call = match.call(fun, orig_call)
     if ("cols" %in% names(formals(fun)) && !"cols" %in% names(named_call)) {
       named_call[["cols"]] = all_cols
+    } else if (is.call(named_call[["cols"]])) {
+      named_call[["cols"]] = eval(named_call[["cols"]], parent.frame(2L))
     }
     named_call[[1L]] = fun
     eval(named_call, parent)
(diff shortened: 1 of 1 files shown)


tdhock:
hi @UweBlock thanks for the PR.
The big idea in my design was that the pattern/sep arguments to `measure()` can and should be used to specify the subset of columns to melt, and the user should not have to specify the cols argument. 
So rather than changing the code, maybe we should change the documentation? to say something like, "cols will be automatically set by melt.data.table, so the user should not provide this argument."

tdhock:
or better: "cols: a character vector that is automatically provided, with all of the column names of the data table input to melt. The user should NOT provide this argument; to specify the columns to be measured, use the pattern argument."

MichaelChirico:
Maybe I'm missing something obvious, but if `cols` is not supposed to be supplied by the user, why keep it in the public API? Shouldn't we just remove it from the signature?

UweBlock:
Thank you for the detailed explanation, @tdhock !

My impression was that the `cols` argument is to be used _together_ with the `sep`/`patterns` argument to define a subset of columns of the dataset which need to be reshaped and on which `sep`/`patterns` is to be applied. 

This would help in cases where `sep` is to simplistic to define the columns to reshape. Also, it would make it less complex to write regular expressions for `patterns` (especially for users who are less acquainted with regular expressions). In my example #5063 I could have written  `pattern = "(\\w+)X(\\d{4})"` instead of `pattern = "([BC]XO)X(\\d{4})"` if `cols` would restrict the columns to reshape.

So, if `cols` does not have this expected (at least by me) functionality or any other meaning to the user, then I would second @MichaelChirico [suggestion](https://github.com/Rdatatable/data.table/pull/5064#issuecomment-874320479) to remove `cols` from the API.

tdhock:
well, I would be all for removing the cols argument, because indeed I don't think the user should ever have to use it.
but then how does `measure()` know what are the column names of the table passed to melt?
I based my design on the `patterns()` function which also has the cols argument:
```r
patterns = function(..., cols=character(0L)) {
```
for completeness the cols argument to patterns is documented as follows:
```
  \item{cols}{A character vector of names to which each pattern is matched.}
```

avimallu:
> ```r
> patterns = function(..., cols=character(0L)) {
> ```

Perhaps it could be mentioned along with the default argument to specify that it is not required? To clarify its use further, the help text

> Perl-compatible regex with capture groups to match to cols. Columns that match the regex are considered measure variables.

or that for the `cols` argument can be specified to not require `cols` to be user specified?

tdhock:
hi @tlapak would you agree that to move forward we should either close or substantially modify this PR? (remove changes to code, add new docs for cols argument)

PR #5072, opened 2021-07-13
Set default domain="R-data.table" explicitly for translation functions

MichaelChirico:
@eliocamp helped point out that messages within `print()` are not translated, apparently for similar reasons as why `onAttach()`/`onLoad()` aren't -- it seems because `print()` is a base method, it looks for messages in the base domain. That sounds weird to me, but it is what it is.

See this thread starting from Elio's post:

https://twitter.com/d_olivaw/status/1414673149875179523

And my subsequent post to r-devel pointing out the potential error:

https://stat.ethz.ch/pipermail/r-devel/2021-July/080913.html

Anyway, we can sidestep this by always supplying the `domain`. As a bonus, a quick benchmark shows it's about a 10x speedup to do so:

```
# on current master
microbenchmark::microbenchmark(times = 1e5, tryCatch(print(as.data.table(mtcars), col.names = 'ijiji'), error=identity))
# Unit: microseconds
#                                                                           expr
#  tryCatch(print(as.data.table(mtcars), col.names = "ijiji"), error = identity)
#      min       lq     mean   median      uq      max neval
#  155.327 164.9255 185.2734 169.9765 174.553 28580.92 1e+05
```

vs

```
# on this branch
microbenchmark::microbenchmark(times = 1e5, tryCatch(print(as.data.table(mtcars), col.names = 'ijiji'), error=identity))
# Unit: microseconds
#                                                                           expr
#  tryCatch(print(as.data.table(mtcars), col.names = "ijiji"), error = identity)
#     min     lq     mean median     uq      max neval
#  13.517 14.874 16.54797 15.306 15.895 3358.671 1e+05
```

PS we still have some `gettext()`/`gettextf()` calls (especially in setops.R) -- maybe we should write a quick `DTgettext` / `DTgettextf` wrapper for those to get default domains there too? The cost is yet another internal helper...

### DIFF
--- a/R/onAttach.R
+++ b/R/onAttach.R
@@ -18,24 +18,22 @@
   }
   dev = as.integer(v[1L, 3L]) %% 2L == 1L  # version number odd => dev
   if (!isTRUE(getOption("datatable.quiet"))) {   # new option in v1.12.4, #3489
-    # NB: we need to supply domain= for translation below since the below is technically not run in the data.table namespace
     nth = getDTthreads(verbose=FALSE)
     if (dev)
-      packageStartupMessagef(domain="R-data.table", "data.table %s IN DEVELOPMENT built %s%s using %d threads (see ?getDTthreads).  ", v, d, g, nth, appendLF=FALSE)
+      packageStartupMessagef("data.table %s IN DEVELOPMENT built %s%s using %d threads (see ?getDTthreads).  ", v, d, g, nth, appendLF=FALSE)
     else 
-      packageStartupMessagef(domain="R-data.table", "data.table %s using %d threads (see ?getDTthreads).  ", v, nth, appendLF=FALSE)
-    packageStartupMessagef(domain="R-data.table", "Latest news: r-datatable.com")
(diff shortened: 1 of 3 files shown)


mattdowle:
> PS we still have some gettext()/gettextf() calls ...

I'm not sure really.  According to `grep -R gettext | wc -l` there are 22 lines using either `gettext`, `gettextf` or `ngettext`, with varying requirements. Depending on how many lines touched I think I lean towards adding domain="R-data.table" to the ones where that is needed, adding a grep to CRAN_Release to make sure each gettextf is passed domain, and adding a grep to ensure there are no calls to gettext.  Otherwise wouldn't the new wrapper(s) be confusing when to use it or not?  If going for the wrapper, maybe stringf, or translatef reads a bit better than DTgettextf. I never understood really why `gettext` wasn't called `translate` in the first place.  Or maybe DTgettextf reads better to those who know gettextf already.

MichaelChirico:
agree, translate/translatef sound good& clear and make it easy to look for gettext/gettextf as a release check. will file follow up.

PR #5086, opened 2021-07-29
Tidy other.Rraw CI

mattdowle:
* one place for `other.Rraw`'s package; the two places had gotten out of line
* `TEST_DATA_TABLE_WITH_OTHER_PACKAGES=true` is now in my `.bash_aliases` (which is here in GitHub) so that I always run `other.Rraw` when I run `R CMD check` locally.
* partial argument name warnings from base and other packages are now displayed in `other.Rout` rather than just 'there were 14 warnings'
* it's now all or nothing for `other.Rraw` (i.e. `_R_CHECK_FORCE_SUGGESTS_` now ignored) to prevent skipping some packages that `other.Rraw` tests just because they're not installed.

### DIFF
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -7,7 +7,7 @@ On each Pull Request opened in GitHub we run Travis CI and Appveyor to provide p
 ### [GitLab CI](./../.gitlab-ci.yml)
 
 Test jobs:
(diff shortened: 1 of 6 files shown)


mattdowle:
@jangorecki what's the command to push to branch to gitlab to test GLCI before merging?  This time I'll just merge.

mattdowle:
I already see I missed mirror-other-packages, will do a follow-up.

jangorecki:
> @jangorecki what's the command to push to branch to gitlab to test GLCI before merging? This time I'll just merge.

Just push to branch from your local git and not github.

PR #5087, opened 2021-07-29
fixing "invalid permission" segfault from #5077 by only starting the amount of thread, memory was allocated for.

### LINKED ISSUE #5077: `[.data.table` crashes with segfault while grouping with more than 79 threads
I'm working at the IT department in a research institute, and recently one of our scientists came to me with mysteriously crashing R sessions on one (and only one) of our machines (designated for use with R and Rstudio) at the institute. While investigating the problem, I found that the main and deciding difference between the machines he tried it on is the CPU and more specifically the number of cores of the CPUs. The machine the code was working on was a 4 socket, 16 core machine and the machine where the crashes occurred was a 2 socket, 265 core (with SMT) machine. Further investigation revealed, this was important, because when importing `data.table`, by default 50% of the available cores are chosen for multithreaded operations.

From playing around with `setDTthreads(threads = N)`, I found that `N = 79` seamed to be the magical threshold for the crashes to occur. Below this number the code consistently works, above this number the code will crash in most (but not all) cases.

The error first occured with R 4.0.2 and data.table 1.13.0. But because #4892 looked somewhat similar, I also tested it with the latest development version 1.14.1.

Please find blow the crashing code, from what he told me, the code is trying to archive the follwing:
> What we are calculating is what is called “the nearest neighbour distance” (it is a term). The “AgentID” is an ID given to each “person” to identify the data belong to them. For each “Time” and for each “AgentID”, we are calculating what is the distance of closest person.

```R
library(data.table)

M  <- 160
M2 <- M / 2
M4 <- M / 4

J  <- 10001

setDTthreads(threads = 96)

d2 <- data.table(
        Time=rep(1:J, each=M),
        AgentID=rep(1:M4, times=4*J*M),
        i_phase=sample(0:1, J*M, replace=T)*2-1,
        State=sample(0:99,J*M, replace=T),
        SimulationID=sample(1:4, J*M, replace=T),
        N=rep(M4, times=J*M),
        discountFactor=rep(0, times=J*M)
)

findNNDistance <- function(x){
    return(Rfast::nth(x,2,descending = F))
}

d2[, nnDistance_inphase := apply(as.matrix(dist(State, method = "euclidean")), 1, FUN=findNNDistance), by=.(Time,i_phase,N,SimulationID,discountFactor), verbose=T]

```
## error
The spcifics of the segfault vary, depending on the number of threads, so here are the error for a variaty of thread numbers
<details>
  <summary>the result for `N = 80` threads</summary>
  
  ```
Argument 'by' after substitute: .(Time, i_phase, N, SimulationID, discountFactor)
Detected that j uses these columns: [State]
Finding groups using forderv ... forder.c received 256025600 rows and 5 columns
12.3s elapsed (00:03:20 cpu) 
Finding group sizes from the positions (can be avoided to save RAM) ... 0.002s elapsed (0.001s cpu) 
Getting back original order ... forder.c received a vector type 'integer' length 80008
*** Error in `/usr/local/R/4.0.2/lib64/R/bin/exec/R': double free or corruption (!prev): 0x00000000040ae990 ***
Aborted (core dumped)

  ```
  
</details>
<details>
<summary>the result for `N = 96` threads</summary>

```
Argument 'by' after substitute: .(Time, i_phase, N, SimulationID, discountFactor)
Detected that j uses these columns: [State]
Finding groups using forderv ... forder.c received 256025600 rows and 5 columns
13.2s elapsed (00:04:12 cpu) 
Finding group sizes from the positions (can be avoided to save RAM) ... 0.001s elapsed (0.001s cpu) 
Getting back original order ... forder.c received a vector type 'integer' length 80008

 *** caught segfault ***

 *** caught segfault ***
address 0x64ff818, cause 'memory not mapped'

 *** caught segfault ***
address 0x623f844, cause 'memory not mapped'

 *** caught segfault ***
address 0x61bf84c, cause 'memory not mapped'
address 0x64bf81c, cause 'memory not mapped'

 *** caught segfault ***
address 0x62ff838, cause 'memory not mapped'

 *** caught segfault ***
address 0x61ff848, cause 'memory not mapped'

 *** caught segfault ***
address 0x63ff828, cause 'memory not mapped'

 *** caught segfault ***
address 0x647f820, cause 'memory not mapped'

 *** caught segfault ***
address 0x627f840, cause 'memory not mapped'

 *** caught segfault ***
address 0x643f824, cause 'memory not mapped'

 *** caught segfault ***
address 0x62bf83c, cause 'memory not mapped'

 *** caught segfault ***
address 0x63bf82c, cause 'memory not mapped'

 *** caught segfault ***
address 0x633f834, cause 'memory not mapped'

 *** caught segfault ***
address 0x653f814, cause 'memory not mapped'

 *** caught segfault ***
address 0x637f830, cause 'memory not mapped'
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: bad value
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: bad value
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
*** stack smashing detected ***: <unknown> terminated
> Segmentation fault (core dumped)
```

</details>
<details>
  <summary>the result for `N = 128` threads</summary>
  
  ```
Argument 'by' after substitute: .(Time, i_phase, N, SimulationID, discountFactor)
Detected that j uses these columns: [State]
Finding groups using forderv ... forder.c received 256025600 rows and 5 columns
35.9s elapsed (00:49:31 cpu) 
Finding group sizes from the positions (can be avoided to save RAM) ... 0.001s elapsed (0.000s cpu) 
Getting back original order ... forder.c received a vector type 'integer' length 80008

 *** caught segfault ***
address 0x7f7178721000, cause 'invalid permissions'

Traceback:
 1: forderv(firstofeachgroup)
 2: `[.data.table`(d2, , `:=`(nnDistance_inphase, apply(as.matrix(dist(State,     method = "euclidean")), 1, FUN = findNNDistance)), by = .(Time,     i_phase, N, SimulationID, discountFactor), verbose = T)
 3: d2[, `:=`(nnDistance_inphase, apply(as.matrix(dist(State, method = "euclidean")),     1, FUN = findNNDistance)), by = .(Time, i_phase, N, SimulationID,     discountFactor), verbose = T]

Possible actions:
1: abort (with core dump, if enabled)
2: normal R exit
3: exit R without saving workspace
4: exit R saving workspace
Selection: 1
R is aborting now ...
Segmentation fault (core dumped)

  ```
  
</details>

Because the `M = 160` seamed related to the threshold of 79/80 cores, I also tried other `M`, specifically `M = 100` but this error threshold still remains the same.

# Output of `sessionInfo()`
```
R version 4.0.2 (2020-06-22)
Running under: openSUSE Leap 15.2

Matrix products: default
BLAS:   /usr/local/R/4.0.2/lib64/R/lib/libRblas.so
LAPACK: /usr/local/R/4.0.2/lib64/R/lib/libRlapack.so

locale:
 [3] LC_TIME=de_DE.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=de_DE.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=de_DE.UTF-8       LC_NAME=C                 

attached base packages:

other attached packages:
[1] data.table_1.14.1

[1] compiler_4.0.2

```

bennet0496:
Closes #5077

After first finding that the allocated amount of memory for the `TMP` buffer seamed too low, `0x13c0000` vs `80*UINT_16_MAX*sizeof(int) = 0x13ffec0` and was followed by a protected 0-page with `setDTthreads(threads = 80)`, it turned out that in L717-L718 the allocated memory was actually only `79*UINT_16_MAX*sizeof(int) = 0x13bfec4`, because `getDTthreads` was called with `throttle = true`. 

But later when starting the parallel threads, in L1235-L1238 the number of threads to start was determined with `throttle = false`, causing the segfault at L995 for threads over the throttled threshold, while trying to access memory past the `TMP` buffer.

Starting only the throttled amount of threads, fixes that. This also seamed fixed the random `double free or corruption` errors as they are currently not occurring anymore for 80 threads like before.

Unfortunately, it seems to be quite difficult to find a reasonably small dataset and requirement in thread count to write a test for, as quite some things need to happen, that this error occurs. The requirement I currently found are

- the dataset must be groupable into enough groups, that parallel execution will occur (L1195-L1243)
- groups need to have the correct size (<65535?) for the parallel region L1235-L1239 to be run
- groups or slices of the group (?) must be larger than 256 and smaller than 65535 for L936-L1036 (containing the critical L995) to run
- the number of configured threads to run need to be larger than the number of throttled threads (the ceiling of the number of groups divided by 1024 (`R_DATATABLE_THROTTLE`)), that less memory is allocated than used
- the group for the thread with a thread-number higher than the ceiling of the number of groups divided by 1024 must have enough elements for the out of bound access to happen on the next protected page.
- the number of throttled threads needs to be large enough, that the `malloc()` in L718 will internally use `mmap()` instead of the default `brk()`/`sbrk()` so that the protected page is not moved by new heap allocations

In other words it seems to be a very specific edge case and the scientist that created the dataset that causes the crash, should maybe start to gamble :)


mattdowle:
Thanks for all this great detail. LOL at the last sentence.
I just had a quick look at the proposed fix so far, and I can see why it works by applying the throttle which reduces the number of threads. But `false` seems like the correct value there since the loop is through `ngrp`:
```
#pragma omp parallel for schedule(dynamic) num_threads(getDTthreads(ngrp, false))
for (int i=0; i<ngrp; i++) {
```
The comments in `getDTthreads` for `throttle` are: 
```
// throttle==true  : a number of iterations per thread (DTthrottle) is applied before a second thread is utilized
// throttle==false : parallel region is already pre-chunked such as in fread; e.g. two batches intended for two threads
```

I'm thinking it's the `true` on L717 that needs to be `false`?

mattdowle:
This `grep` finds this one, and 4 others to review: all `getDTthreads(..., true)` calls not on `#pragma` lines : 
```
$ grep -n "getDTthreads(.*true" *.c --exclude=openmp-utils.c --exclude=init.c | grep -v pragma 
forder.c:717:  nth = getDTthreads(nrow, true);  // this nth is relied on in cleanup()
forder.c:1039:  int batchSize = MIN(UINT16_MAX, 1+my_n/getDTthreads(my_n, true));  // (my_n-1)/nBatch + 1;   //UINT16_MAX == 65535
fsort.c:116:  int nth = getDTthreads(xlength(x), true);
gsumm.c:82:  nBatch = MIN((nrow+1)/2, getDTthreads(nrow, true)*2);  // *2 to reduce last-thread-home. TODO: experiment. The higher this is though, the bigger is counts[]
subset.c:14:  int nth = getDTthreads(n, /*throttle=*/true);   // not const for Solaris, #4638
```

I was thinking that all `getDTthreads()` calls not on `#pragma` lines should be enforced to be `false`; i.e. the 5 calls above.  But then, for a large number of threads such as the 80 in issue #5077, working memory would be allocated for the 80 threads which would be wasteful when throttling.

Maybe the problem in general is that we call `getDTthreads()` again on `#pragma` lines, where we should instead use `nth` where there is one, to ensure the `throttle=true|false` choice matches the allocation.

bennet0496:
> LOL at the last sentence.

I mean, from what I've seen, it looks pretty unlikely to me to end up with a dataset fulfilling all the requirement to hit this error. And there is a saying where I'm from, that when you manage to do something very unlikely by pure chance, than you have enough luck to gamble or more specifically to play the lottery (even though mathematically this would make no sense). But that's besides the point :)

> But false seems like the correct value there since the loop is through ngrp
> [...]
> I'm thinking it's the true on L717 that needs to be false?

To be honest, I'm not completely sure which would be the correct value in either place. The code is quite hard to read, and it took me a good week to understand it enough to find this error. Also throttling the parallel region, to fix the problem, seamed least intrusive to me, as throttling seamed to be applied for a good reason in the first place.

> I was thinking that all getDTthreads() calls not on #pragma lines should be enforced to be false; i.e. the 5 calls above. But then, for a large number of threads such as the 80 in issue #5077, working memory would be allocated for the 80 threads which would be wasteful when throttling.

Fair enough, allocating the upper bound of what could possibly be used will definitely also solve the problem. It looks like, the overhead would be about 256kB per additional thread, which is arguably not optimal but probably also not too bad.

> Maybe the problem in general is that we call getDTthreads() again on #pragma lines, where we should instead use nth where there is one, to ensure the throttle=true|false choice matches the allocation.

I agree, that it would probably be best to reuse the value where possible and not calculate it again to prevent mismatching option and therefore calculations. I will take a look at that. Maybe using the minimum of `nth` and the calculated value in the pragma lines would also be an option, as the basis on which the values are calculated seam to be different. I case at some point it desirable to start even less threads than anticipated during memory calculation.

bennet0496:
Okay so I traced back the values and found, that

variable_dependecy

now for the first call of `radix_r()`, directly after the memory allocation, `my_n = nrow`. This means for the blue box (with a sufficiently large number of configured threads)

CodeCogsEqn

And for the yellow box

CodeCogsEqn (1)

While during allocation of `TMP` with a sufficiently large number of configured threads `nth` is calculated as

CodeCogsEqn (2)

So it looks like usually one more thread is started than memory was allocated for. So when the additional conditions to reach L995 are met, the buffer overflow occurs.

It also looks like this global `TMP` is used by the code around L995, so it seems to overhead in some (or most?) of the time anyway, so it probably also wouldn't hurt to allocate the additional 256kB per thread for this buffer to be safe that no overflow occurs, instead of throttling the number of threads to start.

mattdowle:
Have invited you to be project member so among other things you can create branches in the main project in future. The invite should be a button you need to click to accept in GitHub in either your projects or profile page. Have also added you to the contributors list. Thanks again!

src/forder.c
```diff
@@ -1222,8 +1227,14 @@ void radix_r(const int from, const int to, const int radix) {
     } else {
       // all groups are <=65535 and radix_r() will handle each one single-threaded. Therefore, this time
       // it does make sense to start a parallel team and there will be no nestedness here either.
+
+      int threads_to_run = MIN(nth, getDTthreads(ngrp, false));  // the parallel regions below might reach code
```
mattdowle, line 1231: Calling `getDTthreads()` again here still leaves the same door open that this value of throttle needs to be the same as the value used to allocate on line 718. We can use `nth` on the `#pragma` line, avoid the new variable, avoid the extra call to `getDTthreads()` and reduce from 3 places to 2 places. Will do...

bennet0496: I don't think calling it again here with the `MIN` limiting it to the value of `nth` is necessarily the same problem. But thinking about it again and seeing the last commits you added, I'd say it was redundant, as `getDTthreads` w/ `throttle=false`, basically only limits to the number of configured threads, which is already done w/ `nth`. So if configured threads are the limiting factor, this already resembled in `nth` so there is no need to apply it again to `ngrp` as it can then only play a role here if `ngrp < threads`, but then also `ngrp < nth` would be true. So doing like you did with `MIN(nth, grp)` is more efficient as it saves the function call.

src/forder.c
```diff
@@ -39,7 +39,7 @@ static int gs_n = 0;                // the number of groups found so far (how mu
 static int **gs_thread=NULL;        // each thread has a private buffer which gets flushed to the final gs appropriately
 static int *gs_thread_alloc=NULL;
 static int *gs_thread_n=NULL;
-static int *global_TMP=NULL;               // UINT16_MAX*sizeof(int) for each thread; used by counting sort in radix_r()
```
bennet0496: I renamed the variable, because it made it easier to distinguish from the other block-local `TMP`(s?) during debugging. But yes, for consistency I probably should have named it back, after I was done.

NEWS.md
```diff
@@ -153,7 +153,7 @@
...
 
-29. A segfault could occur when grouping or sorting in some circumstances with a large number of threads such as 80, [#5077](https://github.com/Rdatatable/data.table/issues/5077). Thanks to Bennet Becker for reporting, debugging at C level, and fixing.
+29. A segfault could occur when grouping or sorting in some circumstances with a large number of threads such as 80 on a server with 256 logical cores, [#5077](https://github.com/Rdatatable/data.table/issues/5077). Thanks to Bennet Becker for reporting, debugging at C level, and fixing.
```
bennet0496, line 156: When mentioning the processor used, I think it would be better to use 128 threads as the example, as this would be the default value on such a machine. The other values mentioned (such as 80 or 96) were manually configured to initially find the error threshold. And I don't know how common it is for people to manually set the amount of threads to be used.

bennet0496 (commented):
This was the wrong of the two comment buttons, I guess. I'm sorry

NEWS.md
mattdowle: Thanks, I see. I'm going to try and create a test which reliably fails under ASAN (the commands are in CRAN_Release.cmd). I suspect it's not failing very often currently because there's space after the end of TMP within the last page that's effectively ok to write to (or something like that). Turning on ASAN should trap that in more cases given a suitable test. I never like using the words "in some circumstances" in the news items because I sense the reader thinking "well, what circumstances, and might my circumstances be affected?". So once the test is ready I'll go back and replace that phrase and the number of threads and cores with more detail.

bennet0496: Yes with ASAN it should not be too hard to find a reliably failing example. The examples I have on my work PC might even produce the error with this, but I will not be able to access them until tomorrow noon (CEST time). But I think even this example required at least 6 threads or something like that, and I don't know what is possible in the test environment or what the limits are.

And I wouldn't call it "effectively ok to write to" but "possible to write to, without the environment noticing", as buffer overflows are always inherently problematic, as it may overwrites other data for other program parts or the overflown data is unexpectedly overwritten by other parts of the program, especially if the data is small enough for heap allocation to occur.

mattdowle: Yes that is better wording. I do know that.

mattdowle: This is enough to segfault and doesn't need ASAN:
```
DT = data.table(grp=sample(65536))  # >=65536 necessary
setDTthreads(throttle=nrow(DT))     # increase throttle to reduce threads to 1 for this nrow
DT[, .N, by=grp]                    # segfault
setkey(DT, grp)                     # segfault
```

bennet0496: Oh ok, I didn't know (or find) the throttle value is adjustable like this. I tried to find an example with the default 1024 value, which made it much harder, to find a small example that would even remotely reach the protected page.

PR #5091, opened 2021-08-03
fwrite allows sep=""

### LINKED ISSUE #4817: fwrite could allow sep=""
So it can be used when we need to paste some strings are write to csv later. We can avoid pasting them in R.
Use case explained in https://github.com/Rdatatable/data.table/issues/4572#issuecomment-731603120

ben-schwen:
Closes #4817

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -97,6 +97,8 @@
 
 15. New convenience function `%plike%` maps to `like(..., perl=TRUE)`, [#3702](https://github.com/Rdatatable/data.table/issues/3702). `%plike%` uses Perl-compatible regular expressions (PCRE) which extend TRE, and may be more efficient in some cases. Thanks @KyleHaynes for the suggestion and PR.
 
+16. `fwrite()` now accepts `sep=""`, [#4817](https://github.com/Rdatatable/data.table/issues/4817). The motivation is an example where the result of `paste0()` needs to be written to file but `paste0()` takes 40 minutes due to constructing a very large number of unique long strings in R's global character cache. Allowing `fwrite(, sep="")` avoids the `paste0` and saves 40 mins. Thanks to Jan Gorecki for the request, and Ben Schwen for the PR.
+
 ## BUG FIXES
 
 1. `by=.EACHI` when `i` is keyed but `on=` different columns than `i`'s key could create an invalidly keyed result, [#4603](https://github.com/Rdatatable/data.table/issues/4603) [#4911](https://github.com/Rdatatable/data.table/issues/4911). Thanks to @myoung3 and @adamaltmejd for reporting, and @ColeMiller1 for the PR. An invalid key is where a `data.table` is marked as sorted by the key columns but the data is not sorted by those columns, leading to incorrect results from subsequent queries.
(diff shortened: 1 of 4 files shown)


MichaelChirico:
can we benchmark any potential performance impact of those branches added in potentially tight loops?

I guess they could be optimized away, makes sense to test with/without compiler optimization on

mattdowle:
Yes good thought. I was just about to remove the branches...

mattdowle:
Yes: I'd be surprised if there was any measurable difference.
The things that might be worth trying are `bool` instead of `int` and declaring `sep` and `sepLen` locally as `const` inside that deep loop in the parallel region. That way the `const` is making it crystal clear to the compiler that each thread doesn't need to read the global `sep` and `sepLen` every time just in case they changed.

archgen:
Hello, I am getting the below error when trying to set sep='' for fwrite.. I was wondering if you can direct me to how to utilise the added fwrite sep='' functionality? 

Thanks very much for your help! 
Matthew

Error in fwrite(eigenstrat_file$geno, sep = "", quote = FALSE, row.names = FALSE,  : 
  is.character(sep) && length(sep) == 1L && nchar(sep) == 1L is not TRUE 


R version 4.1.2 (2021-11-01)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur 10.16

other attached packages:
data.table_1.14.2

ben-schwen:
@archgen 
This feature has not made it to CRAN yet and is only available in the development version.

You can install the development version using `data.table::update_dev_pkg()`

archgen:
Hi @ben-schwen thanks for your quick reply! I tried data.table::update_dev_pkg() but got the following error Error: 'update_dev_pkg' is not an exported object from 'namespace:data.table'

ben-schwen:
@archgen Then `data.table:::update_dev_pkg()` should work.

archgen:
Thanks again @ben-schwen , unfortunately, I get the error 

Error in get(name, envir = asNamespace(pkg), inherits = FALSE) : 
  object 'update_dev_pkg' not found

I'm running on a cluster if that's important to know.. 

Thanks again for your time and help! 

sessionInfo()
R version 3.6.0 (2019-04-26)
Running under: Red Hat Enterprise Linux Server 7.7 (Maipo)

Matrix products: default
BLAS/LAPACK: /apps/software/OpenBLAS/0.2.18-GCC-5.4.0-2.26-LAPACK-3.6.1/lib/libopenblas_haswellp-r0.2.18.so

locale:
 [1] LC_CTYPE=en_AU.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_AU.UTF-8        LC_COLLATE=en_AU.UTF-8    
 [5] LC_MONETARY=en_AU.UTF-8    LC_MESSAGES=en_AU.UTF-8   
 [7] LC_PAPER=en_AU.UTF-8       LC_NAME=C                 
[11] LC_MEASUREMENT=en_AU.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] data.table_1.14.2

[1] compiler_3.6.0 tools_3.6.0

jangorecki:
data.table:::update.dev.pkg()

The "_" function name is in 1.14.3+

PR #5097, opened 2021-08-06
added fread(file=URL) support

### LINKED ISSUE #4952: fread only allows URLs on `input`, not `file`
The documentation for `fread`'s argument `file` states

> [...] a URL starting http:// [...]

However, if I try reading a file via HTTP, I can only get it to work by passing it to the generic `input`, not `file`.

``` r
library(data.table)
packageVersion("data.table")
#> [1] ‘1.14.0’

x <- "http://dados.cvm.gov.br/dados/FI/DOC/INF_DIARIO/DADOS/inf_diario_fi_202012.csv"

fread(x, sep = ";")
#>         TP_FUNDO         CNPJ_FUNDO  DT_COMPTC VL_TOTAL  VL_QUOTA VL_PATRIM_LIQ
#>      1:       FI 00.017.024/0001-53 2020-12-01  1099408 27.493767       1098370
#>      2:       FI 00.017.024/0001-53 2020-12-02  1099497 27.493801       1098371
#>      [...]

fread(file = x, sep = ";")
#> Error in fread(file = x, sep = ";"): File 'http://dados.cvm.gov.br/dados/FI/DOC/INF_DIARIO/DADOS/inf_diario_fi_202012.csv' does not exist or is non-readable. getwd()=='[...]'
```

<sup>Created on 2021-04-12 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

`#` `Output of sessionInfo()`

```r
sessionInfo()
#> R version 4.0.5 (2021-03-31)
#> Running under: Windows 10 x64 (build 19041)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.1252 
#> [2] LC_CTYPE=English_United States.1252   
#> [3] LC_MONETARY=English_United States.1252
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.1252    
#> 
#> attached base packages:
#> 
#>  [1] fansi_0.4.2       crayon_1.4.1      utf8_1.2.1        digest_0.6.27    
#>  [5] backports_1.2.1   lifecycle_1.0.0   reprex_1.0.0      magrittr_2.0.1   
#>  [9] evaluate_0.14     pillar_1.5.1      highr_0.8         stringi_1.5.3    
#> [13] rlang_0.4.10      fs_1.5.0          vctrs_0.3.7       ellipsis_0.3.1   
#> [17] rmarkdown_2.7     styler_1.3.2      tools_4.0.5       stringr_1.4.0    
#> [21] glue_1.4.2        purrr_0.3.4       xfun_0.22         yaml_2.2.1       
#> [25] compiler_4.0.5    pkgconfig_2.0.3   htmltools_0.5.1.1 knitr_1.31       
#> [29] tibble_3.1.0
```

ben-schwen:
Closes #4952.

Refactors file download to separate function for better reuse/maintenance.


R/fread.R
```diff
@@ -36,6 +36,33 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
...
+      curl::curl_download(input, tmpFile, mode="wb", quiet = !showProgress)
+    }
+    else { #if (startsWith(input, "ftp://") || str7== "http://" || str7=="file://") {
+      method = if (str7=="file://") "internal" else getOption("download.file.method", default="auto")
+      # force "auto" when file:// to ensure we don't use an invalid option (e.g. wget), #1668
+      tmpFile = tempfile(fileext = paste0(".",tools::file_ext(input)), tmpdir=tmpdir)
+      download.file(input, tmpFile, method=method, mode="wb", quiet=!showProgress)
+      # In text mode on Windows-only, R doubles up \r to make \r\r\n line endings. mode="wb" avoids that. See ?connections:"CRLF"
+    }
+    assign("file", tmpFile, envir=parent.env(environment()))
```
mattdowle, line 58: Why not just return `tmpFile`?

R/fread.R
```diff
@@ -36,6 +36,33 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
...
+    do.call(
+      on.exit, list(substitute(unlink(tmpFile)), add = TRUE),
+      envir = parent.frame()
+    )
```
mattdowle, line 62: Why the do.call here; i.e. can't it just be `on.exit(unlink(tmpFile), add=TRUE)` as before?

R/fread.R
```diff
@@ -36,6 +36,33 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
   )
   nThread=as.integer(nThread)
   stopifnot(nThread>=1L)
+
+  is_url = function(x) any(startsWith(x, c("http://", "https://", "ftp://", "ftps://", "file://")))
```
mattdowle, line 40: I glanced at our backport of `startsWith` to support R prior to 3.3.0 (see utils.R) and it doesn't look like it supports a set like that. If so, our backport needs a tweak then.

mattdowle (commented):
Looks great!  Just those questions on first glance.

R/fread.R
ben-schwen: Maybe my view was locked in thinking about the different environments. A return is definitely the more straightforward way.

R/fread.R
ben-schwen: I think we need the `do.call` here, because we want to `unlink` the file only after the exiting the `parent.frame()` (which is the fread function). Using `on.exit(unlink(tmpFile), add=TRUE)` as before would lead to unlinking the file after leaving the newly added `download_file` function.

R/fread.R
ben-schwen: I changed the usage here, before fiddling with the backport. This also seems okayish since supported set of URLs will probably not change.

PR #5098, opened 2021-08-08
Fwrite integer rownames

### LINKED ISSUE #4957: fwrite substitutes class integer rownames by 1,2,3...

The following R code:

```r
library(data.table)
df <- data.frame(c1 = c(1, 2, 3), c2 = c("a",  "b",  "c"))
rownames(df) <- as.integer(c(10, 20, 30)) # class(attributes(df)$row.names) 
fwrite(df, "out_fwrite", quote = F, row.names = T, sep = "\t")
write.table(df, "out_wt", quote = F, row.names = T, sep = "\t") # added for comparison
```
generates two output files, whose content is shown below:

`out_fwrite`
```
	c1	c2
1	1	a
2	2	b
3	3	c
```
`out_wt`
```
c1	c2
10	1	a
20	2	b
30	3	c
```

I noticed `fwrite` is converting the actual rownames (i.e. 10, 20, 30) to 1, 2, 3 without warning. This does __not__ happen when `rownames(df)  <- as.numeric(c(10, 20, 30))` or `rownames(df) <- as.character(c(10, 20, 30))`. This is also different from the behaviour of `write.table`.

`#` `Output of sessionInfo()`

```
R version 4.0.3 (2020-10-10)
Running under: Scientific Linux 7.2 (Nitrogen)

Matrix products: default
BLAS:   /nfs/users2/rg/dgarrido/R/R-4.0.3/lib64/R/lib/libRblas.so
LAPACK: /nfs/users2/rg/dgarrido/R/R-4.0.3/lib64/R/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] data.table_1.13.6 optparse_1.6.6   

[1] compiler_4.0.3 tools_4.0.3    getopt_1.20.3 
```

ColeMiller1:
Closes #4957 .

The biggest change is behavior of `quote = auto` for default row names and for integer assigned row-names because this PR makes all row names characters.

``` r
library(data.table)
DT = data.table(foo=1:3,bar=c(1.2,9.8,-6.0))

# 1.14.0
fwrite(DT,row.names=TRUE, quote = 'auto')
## "",foo,bar
## "1",1,1.2
## "2",2,9.8
## "3",3,-6

#1.14.1 with PR. Double quote are lost
fwrite(DT,row.names=TRUE, quote = 'auto')
## "",foo,bar
## 1,1,1.2
## 2,2,9.8
## 3,3,-6
```
I believe this better matches existing behavior with `quote = 'auto'` as the current behavior will _not_ add double quotes to character row.names:
```
row.names(DT) = letters[1:3]
fwrite(DT,row.names=TRUE, quote = 'auto')

## "",foo,bar
## a,1,1.2
## b,2,9.8
## c,3,-6
```
Finally, based on this PR, this part of `fwrite.c` shown below is no longer used. In the issue, I [comment](https://github.com/Rdatatable/data.table/issues/4957#issuecomment-894667558) about how to do this more directly in C. I would be happy to move forward with a C approach but it just seems like a higher diff for little productivity. While maybe big data people who need performance use row names, overall I doubt it. 

https://github.com/Rdatatable/data.table/blob/831013a8892a6d39e12bbd45aeb559ccd258bc0a/src/fwrite.c#L876-L881


R/fwrite.R
```diff
@@ -36,6 +36,7 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
   nThread = as.integer(nThread)
   # write.csv default is 'double' so fwrite follows suit. write.table's default is 'escape'
   # validate arguments
+  rn = if (row.names) row.names(x) else NULL # allocate row.names in R to address integer row.names #4957
```
mattdowle, line 39: I like your reasoning for this; i.e. option 1 you detailed in issue. However I'm just a bit concerned about any inadvertent usage that ends up somehow calling row.names=TRUE without the user intending to. Then `1:nrow` will be coerced to character by this line and the global character cache gets clobbered. Would prefer not to leave that door open. Another door open would be benchmarkers who either deliberately or by accident display results with row.names = TRUE and show poor performance.
Hence I went for option 3 as you described.

ColeMiller1: Brilliant! Thanks for your time - changes look great!

PR #5101, opened 2021-08-18
isolate tests.Rraw from variables created by user in .GlobalEnv

### LINKED ISSUE #3705: tests fail if z in .GlobalEnv

mattdowle:
Closes #3705

Could change `z` to be an unusual name such as `foo6937265` and skip the test in the unusual case that even that was in .GlobalEnv.  However, trying to tackle the root cause by changing the parent of the env that tests run inside, so we never have this issue again.

First commit set parent to `parent.env(.GlobalEnv)` which is `search()[2]`. This worked in that test 1894 no longer sees the `z` in `.GlobalEnv`.  However, then other tests which load `bit64` fail because when they are loaded at the top of tests.Rraw they become the new position 2 and functions like `as.integer64()` are not found because the parent of the env is now position 3 (the old position 2).

Hence tests.Rraw now attaches suggests at the end of `search()`.


MichaelChirico:
pretty cool & straightforward if it works. otherwise, what about something like

(1) copy .GlobalEnv to tmpEnv
(2) run tests
(3) on.exit, copy tmpEnv to .GlobalEnv

risk being user has large objects which are expensive to copy...

mattdowle:
I guess there's a missing step (1.5) which removes everything in .GlobalEnv. If so, that not allowed by CRAN policy; e.g. what if tests fail or the on.exit somehow doesn't run, or the on.exit does run but only gets through half the restore.

MichaelChirico:
mitigated by iterating the overwrite but fair, it's not a great solution if we can avoid it.

mattdowle:
I checked R 3.1.0's `library()` function has all the arguments this PR uses: `character.only`, `logical.return`, `quietly`, `warn.conflicts` and `pos`.

PR #5104, opened 2021-08-19
add DT() functional form data.table query

### LINKED ISSUE #641: [R-Forge #1984] Functional Queries
Submitted by: Yike Lu; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=1984&group_id=240&atid=978)

Add a function, perhaps dtquery=function(DT, ...) DT[...]

I just checked, and this syntax indeed works (it was a very easy function program). This would allow higher order abstractions as part of the standard implementation.

### LINKED ISSUE #4872: Compatibility with the future native pipe
Right now, r-devel is implementing a native pipe that is incompatible with chaining in data.table. Using [the proposed `d => ` syntax](https://twitter.com/eddelbuettel/status/1349087820783243266) I get an error using [this](https://github.com/wch/r-source/commit/263d6bcf0ba5f2a87d6ba24a795125e19359c675) R version

```r
library(data.table)
data <- CJ(group1 = letters[1:5], group2 = letters[10:14])

data[, x := rnorm(.N)] |> 
  d => d[, mean(x), by = group1]
#> Error: function '[' not supported in RHS call of a pipe
```

If "[" won't be supported as the RHS of a pipe, then no syntax transformation can  make data.table compatible with the `|>` pipe, if I understand correctly. 

I see two ways of addressing this. 

On the one side, we could talk with r-core to present this issue and see if they can change their implementation to make it work. 

Another option would be to create a functional alias to `data.table:::"[.data.table"`, lets say `dt()`. This, for example, seems to work. 

```r
dt <- data.table:::"[.data.table"

data[, x := rnorm(.N)] |> 
   dt(, mean(x), by = group1)
#>    group1          V1
#> 1:      a  0.58876302
#> 2:      b -0.24705765
#> 3:      c  0.07676786
#> 4:      d -0.33047608
#> 5:      e  0.54832829
```

(in fact, it also works with `dt <- base::"["`)

Personally, I don't mind that notation at all. It is true that if such a simple fix resolves the issue, then each user could do it in their own scripts, but I think it might be preferable if data.table provided a standardised alias so that other people's code stays readable. For the record, I don't like `dt` very much, it's just the first thing that came in mind.

mattdowle:
Closes #4872
Closes #641

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -57,6 +57,7 @@ export(setnafill)
 export(.Last.updated)
 export(fcoalesce)
 export(substitute2)
+export(DT)  # mtcars |> DT(i,j,by)  #4872
 
 S3method("[", data.table)
 S3method("[<-", data.table)
--- a/NEWS.md
+++ b/NEWS.md
@@ -109,6 +109,12 @@
 
 21. `melt()` was pseudo generic in that `melt(DT)` would dispatch to the `melt.data.table` method but `melt(not-DT)` would explicitly redirect to `reshape2`. Now `melt()` is standard generic so that methods can be developed in other packages, [#4864](https://github.com/Rdatatable/data.table/pull/4864). Thanks to @odelmarcelle for suggesting and implementing.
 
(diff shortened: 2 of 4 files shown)


mattdowle:
Merging now so that folk installing dev v1.14.1 can experiment with it before release.
We can still change anything about it before release in follow-up PRs.

PR #5109, opened 2021-08-22
Gforce edge case creates segfault

### LINKED ISSUE #1994: Edge case bug in GForce
```r
require(data.table)
dt <- data.table(a=1, b=2)[0] # empty table
dt[rep(NA_integer_, 10), mean(a), by=list(vv=1:10)] # subset with NA_integer_
```
This results in segfault.. and would result in the same with `median`, `min`, `max` etc. Since they all depend on the input table, as opposed to `gN()` (equivalent of `.N`), for example.

Not sure, but I think this happens because `irows` is non-empty (all NAs), but `x` (input table) is. And I think this difference is not taken care of [in this line](https://github.com/Rdatatable/data.table/blob/master/R/data.table.R#L1880).

ben-schwen:
Closes #1994.

Considered functions:
- [x] `gsum`
- [x] `gmean`
- [x] `gvar`
- [x] `gsd`
- [x] `gmedian`
- [x] `gprod`
- [x] `gminmax`
- [x] `gfirst` / `ghead`
- [x] `glast` / `gtail`


src/gsumm.c
```diff
@@ -760,13 +762,15 @@ static SEXP gminmax(SEXP x, SEXP narm, const bool min)
         const int thisgrp = grp[i];
         if (ansd[thisgrp]==NA_STRING) continue;
         const int ix = (irowslen == -1) ? i : irows[i]-1;
-        if (xd[ix]==NA_STRING || (strcmp(CHAR(xd[ix]), CHAR(ansd[thisgrp]))<0)==min)
+        if (ix+1==NA_INTEGER) SET_STRING_ELT(ans, thisgrp, NA_STRING);
```
mattdowle, line 765: `INT_MAX+1` overflows to `INT_MIN` so if there's a way to check if `irows[i]==NA_INTEGER` directly that would be neater; e.g. in the extreme edge case of `nrow==INT_MAX` and `irowslen==-1`. Actually, maybe it's not as extreme if folk have code that batches into INT_MAX sizes.
Anyway, who am I to comment when this NA-in-i bug was here all along.

ben-schwen: I had the same thought yesterday, but then also thought that if overflowing from `INT_MIN-1` to `INT_MAX` works in the first place, then the return overflow for the if should also work.

The problem vectors reaching length `INT_MAX` is certainly a future problem but one that we have to keep in mind, e.g. switching to 64bit integer for indexes? For now I'm not convinced folks have that kind of big vectors since this would mean an ridiculous amount of RAM.
```
print(object.size(seq.int(2^31-1)), unit="MB")
8192 Mb
```

`base` itself seems to have problems with vectors being that long e.g. `x = seq.int(2^32); print(x)` leading to 
```
[ reached getOption("max.print") -- omitted -99999 entries ]
```

mattdowle: Yes overflow/underflow is reliable in practice but, iirc, the C standard only guarantees it for unsigned not signed. So the concern would be i) weird and wonderful architectures and compilers, and ii) to catch unintentional overflow. UBSAN would catch it in CRAN_Release.cmd before release, if a test covered the overflow. UBSAN is also in CRAN extra tests (https://cran.r-project.org/web/checks/check_issue_kinds.html) so compliance is required by CRAN. I've seen overflow caught be UBSAN before and fixed it, it was probably on smaller types like `int8_t` or `char` where a test covered it.

That problem is just the printing mechanism. Base R can create and use 'big' vectors but the capability varies by function. R's news file is a good place to search to see how they've increased support over time. But regardless, I had in mind a regular data.table with INT_MAX rows.

PR #5113, opened 2021-08-25
DT() follow up

### LINKED ISSUE #5106: An empty `DT()` doesn't print at the end of a pipe chain
I was testing out the new `DT()` function and was expecting it to cause printing much like the equivalent `[.data.table` syntax.

``` r
library(data.table)

test_df = data.table(x = 1:3, y = 1:3)

# `[` syntax
copy(test_df)[, double_x := x * 2][]
#>    x y double_x
#> 1: 1 1        2
#> 2: 2 2        4
#> 3: 3 3        6

# Using `DT()` (no printing)
copy(test_df) |>
  DT(, double_x := x * 2) |>
  DT()
```

Edit: One option would be to define `DT()` as:
```r
DT = function(x, ...) {
  x[...]
}
```

### LINKED ISSUE #5107: Invoking data.table queries on a `data.frame` (and accasionally on a `data.table` ) using the functional form `DT(...)` behaves quite strangely.
I tested the new function `DT()` for piping to see how closely it is related to `[.data.table` when piping a `data.frame` (or a `data.table).

In the data.table documentation `help(data.table)`, I found the following. 
 > A data.table query may be invoked on a data.frame using functional form DT(...), see examples. The class of the input is retained.

Because no restriction is specified on the type of query that can be invoked, this seems to suggest that, in general , data.table queries can be used on a data.frame; but I found that several queries work quite strangely when piping a data.frame or do not work at all.

	D <- copy(mtcars)

### Issues related to .SD. 
When .SD is used in `j`, it works only if `i` or `by` argument is specified.

	D |> DT(, .SD)               # does not work
	D |> DT(, .SD, .SDcols=5:8)  # does not work
	D |> DT(, lapply(.SD, sum))  # does not work
but following the operations work

	D |> DT(, .SD, by=cyl)                           # works
	D |> DT(1:20, .SD)                               # works
	D |> DT(, .SD, by=cyl, .SDcols=5:8)              # works
	D |> DT(1:20, .SD, .SDcols=5:8)        		 # works
	D |> DT(1:20, .SD, by=cyl, .SDcols=5:8)	         # works
	D |> DT(1:20, lapply(.SD, sum))                  # works
	D |> DT(1:20, c(N=.N, lapply(.SD, sum)), by=cyl) # works

### Issues related to subsetting in `i`. 
Subsetting based on one or multiple equality operator `==` does not always work.

	D |> DT(cyl==4)              # does not work but works if cyl is of type character (example: iris |> DT(Species=="setosa"))
	D |> DT(cyl==4 & vs==0)      # does not work
	D |> DT(cyl==4 & vs>0)       # works
	D |> DT(cyl>=4)              # works
	D |> DT(cyl!=4)              # works
	D |> DT(cyl!=4 & vs!=0)      # works
	iris |> DT(Sepal.Length==5.0 & Species=="setosa")      # works
	iris |> DT(Sepal.Length==5.0)                          # does not work
	iris |> DT(Species=="setosa")                          # works

### Issues related to column selection

It works only when `i` is specified

	D |> DT(, cyl)               # does not work
	D |> DT(1:2, cyl)            # works
	D |> DT(, list(cyl))         # does not work
	D |> DT(1:2, list(cyl))      # work

### Assignment and deletion

	D |> DT(, z:=sum(cyl))         # does not work
	D |> DT(, z:=sum(cyl), by=cyl) # works but the output is wrong. 
	D |> DT(1:3, z:=5, by=cyl)     # works and throw a warning message
	D |> DT(1:3, z:=NULL)          # does not work

### Issue related to `on` argument
`on` argument (and related arguments such as `nomatch`, `mult`, etc.) does not work at all.

	X <- data.table(cyl=4)
	# some examples
	D |> DT(X, on="cyl")                # does not work
	D |> DT(.(4), on="cyl")             # does not work
	iris |> DT("setosa", on="Species")  # does not work

### Issue related to printing a `data.table`.
Usually when using a data.table, the output can be printed via chaining like in `DT[, z:=sum(wt), by=cyl][]` where `[]` allows to print the output after assignment. By analogy, I expected `DT |> DT(, z:=sum(wt), by=cyl) |> DT()` to behave similarly but it does not print anything, even if the function `print` itself is called (even several times after the assignment). I mentioned this issue previously when discussing the compatibility issue (#4872) with the new pipe operator in the comment [HERE](https://github.com/Rdatatable/data.table/issues/4872#issuecomment-843614250). This also appears in #5106.

	DT = setDT(copy(mtcars))

	DT[, z:=sum(wt), by=cyl][]                        # print the data.table DT with a new column
	DT |> DT(, z:=sum(wt), by=cyl) |> DT()            # print nothing
	DT |> DT(, z:=sum(wt), by=cyl) |> print()         # print nothing
	DT |> DT(, z:=sum(wt), by=cyl) |> DT() |> print() # print nothing

mattdowle:
Follows #5104 

Closes #5106 by making `DT()` always print.  This printing can't be tested afaik because `test()` uses `capture.output()` which always invokes the printed output. That's why `tests/autoprint.R` exists because that's the only known way to test R's autoprinting mechanism (i.e. what  happens at the R prompt). However, we can't add a `|>` test to that because `|>` depends on R>4.1.0 and any sort of conditional like that will change the scripts output (`autoprint.Rout`) which gets compared against the fixed `autoprint.Rout.save`. However, since the always-print solution is simple and absolute, and there's no known way, then I think it's acceptable to proceed.

Closes #5107. All tests added.

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1326,13 +1326,12 @@ replace_dot_alias = function(e) {
           if (keylen && (ichk || is.logical(i) || (.Call(CisOrderedSubset, irows, nrow(x)) && ((roll == FALSE) || length(irows) == 1L)))) # see #1010. don't set key when i has no key, but irows is ordered and roll != FALSE
             setattr(ans,"sorted",head(key(x),keylen))
         }
-        setattr(ans, "class", class(x)) # fix for #64
+        setattr(ans, "class", class(x))  # retain class that inherits from data.table, #64
         setattr(ans, "row.names", .set_row_names(length(ans[[1L]])))
         setalloccol(ans)
       }
-
       if (!with || missing(j)) return(ans)
-
+      if (!is.data.table(ans)) setattr(ans, "class", c("data.table","data.frame"))  # DF |> DT(,.SD[...]) .SD should be data.table, test 2212.013
       SDenv$.SDall = ans
       SDenv$.SD = if (length(non_sdvars)) shallow(SDenv$.SDall, sdvars) else SDenv$.SDall
       SDenv$.N = nrow(ans)
@@ -1545,6 +1544,7 @@ replace_dot_alias = function(e) {
     #  TODO add: if (max(len__)==nrow) stopf("There is no need to deep copy x in this case")
     #  TODO move down to dogroup.c, too.
     SDenv$.SDall = .Call(CsubsetDT, x, if (length(len__)) seq_len(max(len__)) else 0L, xcols)  # must be deep copy when largest group is a subset
+    if (!is.data.table(SDenv$.SDall)) setattr(SDenv$.SDall, "class", c("data.table","data.frame"))  # DF |> DT(,.SD[...],by=grp) needs .SD to be data.table, test 2022.012
     if (xdotcols) setattr(SDenv$.SDall, 'names', ansvars[xcolsAns]) # now that we allow 'x.' prefix in 'j', #2313 bug fix - [xcolsAns]
     SDenv$.SD = if (length(non_sdvars)) shallow(SDenv$.SDall, sdvars) else SDenv$.SDall
   }
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18055,10 +18055,14 @@ for (col in c("a","b","c")) {
 if (base::getRversion() >= "4.1.0") {
   # we have to EVAL "|>" here too otherwise this tests.Rraw file won't parse in R<4.1.0
   if (exists("DTfun")) DT=DTfun  # just in dev-mode restore DT() in .GlobalEnv as DT object overwrote it in tests above
-  test(2212.01, EVAL("mtcars |> DT(mpg>20, .(mean_hp=round(mean(hp),2)), by=cyl)"),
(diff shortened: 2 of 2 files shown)


jangorecki:
> However, we can't add a |> test to that because |> depends on R>4.1.0 and any sort of conditional like that will change the scripts output (autoprint.Rout) which gets compared against the fixed autoprint.Rout.save

There is a trick I did in `cat tests/knitr.R`
```r
if (suppressPackageStartupMessages(requireNamespace("knitr", quietly = TRUE))) {
    require(knitr)
    knit("knitr.Rmd", quiet=TRUE)
    cat(readLines("knitr.md"), sep="\n")
    invisible(file.remove("knitr.md"))
} else {
    cat(readLines("knitr.Rout.mock", warn = FALSE), sep="\n")
}
```
and then mock file is just expected output. So in this case instead of `requireNamespace` we can test R version.

MichaelChirico:
We can also use .Rin files I believe:

> Test code can either be provided directly in a .R (or .r as from R 3.4.0) file, or via a .Rin file containing code which in turn creates the corresponding .R file

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-subdirectories

Issue #1812, opened 2016-08-15
fread problem with different number of columns


christellacaze:
hi there,
I've got a problem importing files where some lines end with double tab:

If the file's first line has extra columns and then other lines have missing columns, then fill=TRUE solves the problem.  
However i have one file where the extra columns happen in line 82:

`> test[80:84]
[1] "1053982\t2014-11-30\t-1\ttimetak\t14"                "1053982\t2014-11-30\t-1\tyearcompleted\t4"          
[3] "1053982\t24-11-2014\t-1\tAMOUNT OTHER WEBSITES\t6\t\t" "1053982\t24-11-2014\t-1\tBBC1_422\t1\t\t"             
[5] "1053982\t24-11-2014\t-1\tBBC1_501\t13\t\t"`  

In this case fill=TRUE isn't enough to get around the problem, and i get an error message saying there's text after processing all cols:

```
> test<-fread(filenames[9], header = FALSE, fill = TRUE, data.table=FALSE, quote='') # problem as double tab only starts in row 82
Error in fread(filenames[9], header = FALSE, fill = TRUE, data.table = FALSE,  : 
  Expecting 5 cols, but line 82 contains text after processing all cols. Try again with fill=TRUE. Another reason could be that fread's logic in distinguishing one or more fields having embedded sep='    ' and/or (unescaped) '\n' characters within unbalanced unescaped quotes has failed. If quote='' doesn't help, please file an issue to figure out if the logic could be improved.
```

Is there a way i could force fread to either only read the first 5 columns per line and dropping anything after that?
Failing that, can i specify the number of columns in advance to account for the lines with extra tabs, as this way fill=TRUE would work?

MichaelChirico:
Use the `autostart` option so that `fread` catches that there are "6" columns in the file.

Try `fread(..., autostart = 100)`

christellacaze:
thanks Michael.  I have hundreds of files that i need to import with thousands of lines each, and they may all have extra tabs in various places so i was hoping for a solution that would work for all of them.  I guess that's not possible to do with fread so I'll use readr::read_delim instead.


thank you for all your help, I really appreciate it.


Christel


________________________________

MichaelChirico:
should be possible to set autostart to be x% of the length of the file, perhaps with command line tools. anyway I still think `sed` is your best option.

arunsrinivasan:
Please provide a minimal reproducible example.. code that we can copy/paste to reproduce the issue.

[...]

Issue #4130, opened 2019-12-19
fread:  fill=true doesn't work. 


Befrancesco:
Good morning, 
I'd like to manage the csv of 3.5 gb, but I can't because in my case the `fill=TRUE` doesn't work.
My code is: 

> n_a<-fread("C:/xyz/file_names.csv",sep=";", fill = TRUE)

somebody can help me?

Thank you in advance. 
.

MichaelChirico:
Hi @Befrancesco thanks for using `data.table` We'll need a bit more to go on if we're to have any chance at solving your issue.

Can you please run `fread` again with `verbose=TRUE` and report the output?

Ideally, you would be able to share with us your file. Barring that, you could try to reproduce the issue on a smaller subset of the table, e.g. using `readLines` and `sample()`ing lines to find some set of rows for which the error results. With a smaller subset, it would be easier to anonymize the data so you could share a facsimile version.

Befrancesco:
Thank you for your support  :) :)
Unfotunaly I can't share my file, because is confidential. 
```
omp_get_num_procs()==8
R_DATATABLE_NUM_PROCS_PERCENT=="" (default 50)
R_DATATABLE_NUM_THREADS==""
omp_get_thread_limit()==2147483647
omp_get_max_threads()==8
OMP_THREAD_LIMIT==""
OMP_NUM_THREADS==""
data.table is using 4 threads. This is set on startup, and by setDTthreads(). See ?setDTthreads.
RestoreAfterFork==true
Input contains no \n. Taking this to be a filename to open
[01] Check arguments
  Using 4 threads (omp_get_max_threads()=8, nth=4)
  NAstrings = [<<NA>>]
  None of the NAstrings look like numbers.
  show progress = 1
  0/1 column will be read as integer
[02] Opening the file
  Opening file C:/xyz/file_names.csv
  File opened, size = 3.375GB (3624311815 bytes).
  Memory mapped ok
[03] Detect and skip BOM
[04] Arrange mmap to be \0 terminated
  \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
[05] Skipping initial rows if needed
  Positioned on line 1 starting: <<xxx;yyy>>
[06] Detect separator, quoting rule, and ncolumns
  Using supplied sep ';'
  sep=';'  with 29 fields using quote rule 0
  Detected 29 columns on line 1. This line is either column names or first data row. Line starts as: <<x;y>>
[...]

Issue #3436, opened 2019-02-28
improve fread behaviour on inconsistent number of columns


TobiasGold:
Somewhat related to #2267, although not precisely the same. Also #2263 would probably solve my issue. #2727 is also related.

I have csv data which is compiled from different sources, some chunks of the data are bad in that they apparently end their lines with an additional seperator (comma in this case). 
Good line: `AB,1234,9,9,7`
Bad line: `CD,1234,6,6,7,`

If I read it with `test <- fread("/home/user/onebigfile.csv", fill=TRUE)` then I would expect fread() to realize that there are lines with one column too many and just add a "V1" column which I could then just throw away. Instead fread() stops on the first bad line and does not continue (there would be good lines afterwards again). 

If I interpret the verbose output correctly, fread() "realizes" that there are bad lines, but somehow does not change its guessing of columns from 56 to 57. Instead it assumes an awkward jump and carries on?

```
> test <- fread("/home/user/onebigfile.csv", fill=TRUE)
omp_get_max_threads() = 12
omp_get_thread_limit() = 2147483647
DTthreads = 4
RestoreAfterFork = true
Input contains no \n. Taking this to be a filename to open
[01] Check arguments
  Using 4 threads (omp_get_max_threads()=12, nth=4)
  NAstrings = [<<NA>>]
  None of the NAstrings look like numbers.
  show progress = 1
  0/1 column will be read as integer
[02] Opening the file
  Opening file /home/user/onebigfile.csv
  File opened, size = 188.9MB (198091496 bytes).
  Memory mapped ok
[03] Detect and skip BOM
[04] Arrange mmap to be \0 terminated
  \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
[05] Skipping initial rows if needed
  Positioned on line 1 starting: <<Var1,Var2,Var3,Var4,Var5,>>
[06] Detect separator, quoting rule, and ncolumns
  Detecting sep automatically ...
  sep=','  with 56 fields using quote rule 0
  Detected 56 columns on line 1. This line is either column names or first data row. Line starts as: <<Var1,Var2,Var3,Var4,Var5,>>
  Quote rule picked = 0
  fill=true and the most number of columns found is 56
[07] Detect column types, good nrow estimate and whether first row is column names
  Number of sampling jump points = 100 because (198091495 bytes from row 1 to eof) / (2 * 15166 jump0size) == 6530
[...]

Issue #2691, opened 2018-03-20
fread fails with uneven number of columns when max columns in final row (with fill=TRUE and col.names set)


alexdthomas:
This may be related to issue   #1812, but as that one does not have a reproducible example to confirm, I thought it would be more appropriate to open a new issue.  

When a file with an uneven number of columns has the max number of columns in the final row `fread` fails with the following error: 

> Error in fread("foo", header = FALSE, fill = TRUE, sep = ",", col.names = paste("V",  : 
  Expecting 3 cols, but line 9 contains text after processing all cols. Try again with fill=TRUE. Another reason could be that fread's logic in distinguishing one or more fields having embedded sep=',' and/or (unescaped) '\n' characters within unbalanced unescaped quotes has failed. If quote='' doesn't help, please file an issue to figure out if the logic could be improved.

This occurs even with `fill = TRUE` and the maximum number of column names passed to `col.names `.

Here is a small example
```
text <- "12223, University\n12227, bridge, Sky\n12828, Sunset\n13801, Ground\n14853, Tranceamerica\n14854, San Francisco\n15595, shibuya, Shrine\n16126, fog, San Francisco\n16520, California, ocean, summer, golden gate, beach, San Francisco\n"
cat(text, file = "foo")
max.fields<-max(count.fields("foo", sep = ','))
fread("foo", header = FALSE, fill=TRUE, sep=",", col.names = paste("V", 1:max.fields, sep = ""))
```

However, when the row with the maximum number of fields is moved to the middle of the file (in this example row 6), `fread` behaves as expected. 
```
text <- "12223, University\n12227, bridge, Sky\n12828, Sunset\n13801, Ground\n14853, Tranceamerica\n16520, California, ocean, summer, golden gate, beach, San Francisco\n14854, San Francisco\n15595, shibuya, Shrine\n16126, fog, San Francisco\n"
cat(text, file = "foo")
max.fields<-max(count.fields("foo", sep = ','))
fread("foo", header = FALSE, fill=TRUE, sep=",", col.names = paste("V", 1:max.fields, sep = ""))
```

I included this caveat in my answer to this Stackoverflow [question](https://stackoverflow.com/questions/44714323/r-error-text-after-processing-all-cols-in-fread-data-table/49394375#49394375)

laptop session info
```
R version 3.4.1 (2017-06-30)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.3

Matrix products: default
[...]

PR #5119, opened 2021-08-28
fread: turn off sampling for fill

### LINKED ISSUE #2727: Do not suggests fread fill=TRUE if already used
current fread behavior.
```r
dt=fread("Rprofmem.out", header=FALSE)
#Warning message:
#In fread("Rprofmem.out", header = FALSE) :
#  Stopped early on line 2. Expected 3 fields but found 2. Consider fill=TRUE and comment.char=. First #discarded non-empty line: <<new page:"library" >>
dt=fread("Rprofmem.out", header=FALSE, fill=TRUE)
#Warning message:
#In fread("Rprofmem.out", header = FALSE, fill = TRUE) :
#  Stopped early on line 90. Expected 7 fields but found 8. Consider fill=TRUE and comment.char=. First #discarded non-empty line: <<new page:"close" "readRDS" "FUN" "lapply" "find.package" "system.file" #"library" >>
```
Before printing this warning we should check if `fill=TRUE` was used.

ben-schwen:
Closes #2727 .

Moreover, it enables the user to turn off sampling for detecting the number of columns which appears to be an issue in #2691 , #1812 , #4130 , #3436 and #2727 .

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -2,7 +2,7 @@ fread = function(
 input="", file=NULL, text=NULL, cmd=NULL, sep="auto", sep2="auto", dec=".", quote="\"", nrows=Inf, header="auto",
 na.strings=getOption("datatable.na.strings","NA"), stringsAsFactors=FALSE, verbose=getOption("datatable.verbose",FALSE),
 skip="__auto__", select=NULL, drop=NULL, colClasses=NULL, integer64=getOption("datatable.integer64","integer64"),
-col.names, check.names=FALSE, encoding="unknown", strip.white=TRUE, fill=FALSE, blank.lines.skip=FALSE, key=NULL, index=NULL,
+col.names, check.names=FALSE, encoding="unknown", strip.white=TRUE, fill=FALSE, sample.fill=TRUE, blank.lines.skip=FALSE, key=NULL, index=NULL,
 showProgress=getOption("datatable.showProgress",interactive()), data.table=getOption("datatable.fread.datatable",TRUE),
 nThread=getDTthreads(verbose), logical01=getOption("datatable.logical01",FALSE), keepLeadingZeros=getOption("datatable.keepLeadingZeros",FALSE),
 yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
@@ -22,7 +22,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     stopf("Argument 'encoding' must be 'unknown', 'UTF-8' or 'Latin-1'.")
   }
   stopifnot(
-    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill), isTRUEorFALSE(showProgress),
+    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill), isTRUEorFALSE(sample.fill), isTRUEorFALSE(showProgress),
     isTRUEorFALSE(verbose), isTRUEorFALSE(check.names), isTRUEorFALSE(logical01), isTRUEorFALSE(keepLeadingZeros), isTRUEorFALSE(yaml),
     isTRUEorFALSE(stringsAsFactors) || (is.double(stringsAsFactors) && length(stringsAsFactors)==1L && 0.0<=stringsAsFactors && stringsAsFactors<=1.0),
     is.numeric(nrows), length(nrows)==1L
@@ -79,7 +79,6 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
       if (w<=2L) { # https: or ftps:
         if (!requireNamespace("curl", quietly = TRUE))
           stopf("URL requires https:// connection for which fread() requires 'curl' package which cannot be found. Please install 'curl' using 'install.packages('curl')'.") # nocov
-        
         curl::curl_download(file, tmpFile, mode="wb", quiet = !showProgress)
       } else {
         method = if (w==5L) "internal"  # force 'auto' when file: to ensure we don't use an invalid option (e.g. wget), #1668
@@ -146,6 +145,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     }
     # whitespace at the beginning or end of na.strings is checked at C level and is an error there; test 1804
   }
+  if (!sample.fill & !fill) stopf("sample.fill=FALSE cannot be used without fill=TRUE.")
   if (yaml) {
     if (!requireNamespace('yaml', quietly = TRUE))
       stopf("'data.table' relies on the package 'yaml' to parse the file header; please add this to your library with install.packages('yaml') and try again.") # nocov
@@ -262,8 +262,8 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     if (identical(tt,"") || is_utc(tt)) # empty TZ env variable ("") means UTC in C library, unlike R; _unset_ TZ means local
       tz="UTC"
   }
-  ans = .Call(CfreadR,input,identical(input,file),sep,dec,quote,header,nrows,skip,na.strings,strip.white,blank.lines.skip,
-              fill,showProgress,nThread,verbose,warnings2errors,logical01,select,drop,colClasses,integer64,encoding,keepLeadingZeros,tz=="UTC")
+  ans = .Call(CfreadR,input,identical(input,file),sep,dec,quote,header,nrows,skip,na.strings,strip.white,blank.lines.skip,fill,
+              sample.fill,showProgress,nThread,verbose,warnings2errors,logical01,select,drop,colClasses,integer64,encoding,keepLeadingZeros,tz=="UTC")
   if (!length(ans)) return(null.data.table())  # test 1743.308 drops all columns
   nr = length(ans[[1L]])
   require_bit64_if_needed(ans)
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18121,3 +18121,6 @@ if (base::getRversion() >= "4.1.0") {
 # precision powers of 10^(-n), #4461
 test(2213, identical(fread(text="A\n0.8060667366\n")$A, 0.8060667366))
 
+# turning off sampling for detecting the number of columns #2691 #1812 #4130 #3436 #2727
+test(2214.1, fread(text = paste(rep(c("1,2\n", "1,2,3\n"), each=100), collapse=""), fill=TRUE, sample.fill=FALSE), data.table(1L, 2L, rep(c(NA,3L), each=100)))
+test(2214.2, fread(text = paste(rep(c("1,2\n", "1,2,3\n"), each=100), collapse=""), fill=TRUE, sample.fill=FALSE, select = 1:2), data.table(1L, rep(2L, 200)))
--- a/man/fread.Rd
+++ b/man/fread.Rd
@@ -17,7 +17,7 @@ skip="__auto__", select=NULL, drop=NULL, colClasses=NULL,
 integer64=getOption("datatable.integer64", "integer64"),
 col.names,
 check.names=FALSE, encoding="unknown",
-strip.white=TRUE, fill=FALSE, blank.lines.skip=FALSE,
+strip.white=TRUE, fill=FALSE, sample.fill=TRUE, blank.lines.skip=FALSE,
 key=NULL, index=NULL,
 showProgress=getOption("datatable.showProgress", interactive()),
 data.table=getOption("datatable.fread.datatable", TRUE),
@@ -54,6 +54,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC"
   \item{quote}{ By default (\code{"\""}), if a field starts with a double quote, \code{fread} handles embedded quotes robustly as explained under \code{Details}. If it fails, then another attempt is made to read the field \emph{as is}, i.e., as if quotes are disabled. By setting \code{quote=""}, the field is always read as if quotes are disabled. It is not expected to ever need to pass anything other than \"\" to quote; i.e., to turn it off. }
   \item{strip.white}{ default is \code{TRUE}. Strips leading and trailing whitespaces of unquoted fields. If \code{FALSE}, only header trailing spaces are removed. }
   \item{fill}{logical (default is \code{FALSE}). If \code{TRUE} then in case the rows have unequal length, blank fields are implicitly filled.}
+  \item{sample.fill}{logical (default is \code{TRUE}). Only applicable if fill=TRUE. If \code{FALSE} then all rows are used for detecting number of columns. }
   \item{blank.lines.skip}{\code{logical}, default is \code{FALSE}. If \code{TRUE} blank lines in the input are ignored.}
   \item{key}{Character vector of one or more column names which is passed to \code{\link{setkey}}. It may be a single comma separated string such as \code{key="x,y,z"}, or a vector of names such as \code{key=c("x","y","z")}. Only valid when argument \code{data.table=TRUE}. Where applicable, this should refer to column names given in \code{col.names}. }
   \item{index}{ Character vector or list of character vectors of one or more column names which is passed to \code{\link{setindexv}}. As with \code{key}, comma-separated notation like \code{index="x,y,z"} is accepted for convenience. Only valid when argument \code{data.table=TRUE}. Where applicable, this should refer to column names given in \code{col.names}. }
--- a/src/fread.c
+++ b/src/fread.c
@@ -55,6 +55,7 @@ static bool any_number_like_NAstrings=false;
 static bool blank_is_a_NAstring=false;
 static bool stripWhite=true;  // only applies to character columns; numeric fields always stripped
 static bool skipEmptyLines=false, fill=false;
+static bool sampleFill=true; // turn off sampling for determining number of columns
 
 static double NA_FLOAT64;  // takes fread.h:NA_FLOAT64_VALUE
 
@@ -162,6 +163,7 @@ bool freadCleanup(void)
   skipEmptyLines = false;
   eol_one_r = false;
   fill = false;
+  sampleFill = true;
   // following are borrowed references: do not free
   sof = eof = NULL;
   NAstrings = NULL;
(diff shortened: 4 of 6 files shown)


mattdowle:
It feels to me, on first glance, like `fill=TRUE` could and should just work rather than adding a new argument. If we did add this new argument now it will be hard to take away in future once it is in use.
Since data.table is over-allocated, and `fread` is creating a new one for the first time, it should be possible to add new columns as they are encountered out-of-sample when `fill=TRUE`. I saw someone suggested that fill=20 could be used if the user knows there are 20 columns, which seems a good way to control the over-allocation; e.g. if user knows there are not more than 100 columns, then they could specify fill=100 and if only 23 were found, that would be fine and the result would have 23 columns.  If there are 1,000,000 columns and the sample finds only 20,  then fill=TRUE would result in a few shallow copies internally as it found more and more columns out-of-sample and the over-allocated amount kept getting filled.  Those shallow copies could be avoided if the user specified a very large fill=1e7 for example.  But that's just an extreme edge case.

ben-schwen:
@mattdowle Indeed, providing an estimate for the number of columns seems a better choice than reading the file twice. I guess the only thing left to do is to delete the overallocated column or is that a user problem?

mattdowle:
@ben-schwen Wow - great! Will look. Overallocated columns can be left there. They only take up 8 bytes per slot (or 4 bytes on 32bit) and they're not normally user visible.

NEWS.md
```diff
@@ -141,6 +141,8 @@
 
 26. `base::droplevels()` gains a fast method for `data.table`, [#647](https://github.com/Rdatatable/data.table/issues/647). Thanks to Steve Lianoglou for requesting, and Jan Gorecki and Benjamin Schwendinger for the PR. `fdroplevels()` for use on vectors has also been added.
 
+27. `fread(..., fill=FALSE)` now also accepts an `integer` in addition to boolean values. `fread` with `fill=TRUE` stops reading when the automacially estimated number of columns is too low. Providing an `integer` as argument for `fill` serves as estimate for the number of columns. Thanks to @christellacaze for requesting, and Benjamin Schwendinger for the PR.
```
mcol, line 144: Typo: `automacially`

ben-schwen: Ty

ben-schwen:
@mattdowle I took another look at this and now clean up the "overallocated" columns since these get filled with `NA`s. Hence, it works now with `fill=nguess` as long the user provided guess `nguess` is high enough.

The only drawback I noticed so far is that cleaning up these overallocated and `NA` filled columns won't work for too high guesses with `sep== ` or `number_like_NAstrings`. (setting the new `max_col` smarter might solve this)

ben-schwen:
Added this and the to be fixed issues to 1.16.0 milestone since it fixes #1812 (which was already on 1.16.0)

ben-schwen:
> @ben-schwen is this ready for review? Please write a quick summary of the change in the PR description, it will give good context to first-time readers of the PR without having to read the many linked issues :)

Done!

R/fread.R
```diff
@@ -22,11 +22,12 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     stopf("Argument 'encoding' must be 'unknown', 'UTF-8' or 'Latin-1'.")
   }
   stopifnot(
-    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill), isTRUEorFALSE(showProgress),
+    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill) || is.numeric(fill) && length(fill)==1L && fill >= 0.0, isTRUEorFALSE(showProgress),
```
MichaelChirico, line 25: Most often, `fill=` should be provided as an integer, so assume that by default and coerce otherwise.

```suggestion
    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill) || is.numeric(fill) && length(fill)==1L && fill >= 0L, isTRUEorFALSE(showProgress),
```

inst/tests/tests.Rraw
```diff
@@ -18330,3 +18330,22 @@ if (test_bit64) {
   apple = data.table(id = c("a", "b", "b"), time = c(1L, 1L, 2L), y = i64v[1:3])
   test(2248, dcast(apple, id ~ time, value.var = "y"), data.table(id = c('a', 'b'), `1` = i64v[1:2], `2` = i64v[4:3], key='id'))
 }
+
+# fread(...,fill) can also be used to specify a guess on the maximum number of columns #2691 #1812 #4130 #3436 #2727
```
MichaelChirico, line 18334: I see one example here derived from the closing issues, but I think it'd be better if we included more of those as specific regression tests before closing the linked issue.

inst/tests/tests.Rraw
```diff
@@ -18330,3 +18330,22 @@ if (test_bit64) {
...
+dt_str = paste(rep(c("1,2\n", "1,2,3\n"), each=100), collapse="")
```
MichaelChirico, line 18335: I'd use `strrep()`, either:

```suggestion
dt_str = paste0(strrep("1,2\n", 100L), strrep("1,2,3\n", 100L))
dt_str = paste(rep(c("1,2\n", "1,2,3\n"), each=100), collapse="")
```

or

```suggestion
dt_str = paste(strrep(c("1,2\n", "1,2,3\n"), 100L), collapse="")
dt_str = paste(rep(c("1,2\n", "1,2,3\n"), each=100), collapse="")
```

(looks like comments don't accept multiple ```` ```suggest ```` blocks...)

inst/tests/tests.Rraw
```diff
@@ -18330,3 +18330,22 @@ if (test_bit64) {
...
+test(2249.01, fread(text = dt_str, fill=FALSE), data.table(1L, rep(2L, 100)), warning=".*Consider fill=TRUE.*")
+test(2249.02, fread(text = dt_str, fill=TRUE), data.table(1L, rep(2L, 100)), warning=".*Consider fill=3.*")
+test(2249.03, fread(text = dt_str, fill=2L), data.table(1L, rep(2L, 100)), warning=".*Consider fill=3.*")
+test(2249.04, fread(text = dt_str, fill=3L), data.table(1L, 2L, rep(c(NA,3L), each=100)))
+test(2249.05, fread(text = dt_str, fill=5L, verbose=TRUE), data.table(1L, 2L, rep(c(NA,3L), each=100)), output="Provided number of fill columns: 5 but only found 3\n  Dropping 2 overallocated columns") # user guess slightly too big
+test(2249.06, fread(text = dt_str, fill=1000L), data.table(1L, 2L, rep(c(NA,3L), each=100))) # user guess much too big
```
MichaelChirico, line 18341:
```suggestion
ans = data.table(1L, 2L, rep(c(NA, 3L), each=100L))
test(2249.01, fread(text = dt_str, fill=FALSE), ans[1:100, -3L], warning=".*Consider fill=TRUE.*")
test(2249.02, fread(text = dt_str, fill=TRUE), ans[1:100, -3L], warning=".*Consider fill=3.*")
test(2249.03, fread(text = dt_str, fill=2L), ans[1:100, -3L], warning=".*Consider fill=3.*")
test(2249.04, fread(text = dt_str, fill=3L), ans)
test(2249.05, fread(text = dt_str, fill=5L, verbose=TRUE), ans, output="Provided number of fill columns: 5 but only found 3\n  Dropping 2 overallocated columns") # user guess slightly too big
test(2249.06, fread(text = dt_str, fill=1000L), ans) # user guess much too big
```

inst/tests/tests.Rraw
```diff
@@ -18330,3 +18330,22 @@ if (test_bit64) {
...
+# example from 2691
+text = "12223, University\n12227, bridge, Sky\n12828, Sunset\n13801, Ground\n14853, Tranceamerica\n14854, San Francisco\n15595, shibuya, Shrine\n16126, fog, San Francisco\n16520, California, ocean, summer, golden gate, beach, San Francisco\n"
+test(2249.07, dim(fread(text)), c(6L, 3L), warning=c("fill=TRUE", "Discarded"))
+test(2249.08, dim(fread(text, fill=TRUE)), c(9L, 9L))
+text = "12223, University\n12227, bridge, Sky\n12828, Sunset\n13801, Ground\n14853, Tranceamerica\n16520, California, ocean, summer, golden gate, beach, San Francisco\n14854, San Francisco\n15595, shibuya, Shrine\n16126, fog, San Francisco\n"
```
MichaelChirico, line 18346: for readability

```suggestion
lines = c(
  "12223, University",
  "12227, bridge, Sky",
  "12828, Sunset",
  "13801, Ground",
  "14853, Tranceamerica",
  "14854, San Francisco",
  "15595, shibuya, Shrine",
  "16126, fog, San Francisco",
  "16520, California, ocean, summer, golden gate, beach, San Francisco",
  "")
text = paste(lines, collapse="\n")
test(2249.07, dim(fread(text)), c(6L, 3L), warning=c("fill=TRUE", "Discarded"))
test(2249.08, dim(fread(text, fill=TRUE)), c(9L, 9L))
text = paste(lines[c(1:6, 8L, 7L, 9L)], collapse="\n")
```

man/fread.Rd
```diff
@@ -53,7 +53,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC"
...
   \item{quote}{ By default (\code{"\""}), if a field starts with a double quote, \code{fread} handles embedded quotes robustly as explained under \code{Details}. If it fails, then another attempt is made to read the field \emph{as is}, i.e., as if quotes are disabled. By setting \code{quote=""}, the field is always read as if quotes are disabled. It is not expected to ever need to pass anything other than \"\" to quote; i.e., to turn it off. }
   \item{strip.white}{ default is \code{TRUE}. Strips leading and trailing whitespaces of unquoted fields. If \code{FALSE}, only header trailing spaces are removed. }
-  \item{fill}{logical (default is \code{FALSE}). If \code{TRUE} then in case the rows have unequal length, blank fields are implicitly filled.}
+  \item{fill}{logical or integer (default is \code{FALSE}). If \code{TRUE} then in case the rows have unequal length, number of columns is estimated and blank fields are implicitly filled. If an integer is provided this is used as upper bound guess for the number of columns. }
```
MichaelChirico, line 56:
```suggestion
  \item{fill}{logical or integer (default is \code{FALSE}). If \code{TRUE} then in case the rows have unequal length, number of columns is estimated and blank fields are implicitly filled. If an integer is provided it is used as an upper bound for the number of columns. }
```

src/fread.c
```diff
@@ -2611,8 +2636,13 @@ int freadMain(freadMainArgs _args) {
       else {
         ch = headPos;
         int tt = countfields(&ch);
-        DTWARN(_("Stopped early on line %"PRIu64". Expected %d fields but found %d. Consider fill=TRUE and comment.char=. First discarded non-empty line: <<%s>>"),
+        if (fill>0) {
+          DTWARN(_("Stopped early on line %"PRIu64". Expected %d fields but found %d. Consider fill=%d or even higher ncol estimate. First discarded non-empty line: <<%s>>"),
```
MichaelChirico, line 2640:
```suggestion
          DTWARN(_("Stopped early on line %"PRIu64". Expected %d fields but found %d. Consider fill=%d or even or more based on your knowledge of the input file. First discarded non-empty line: <<%s>>"),
```

src/fread.c
```diff
@@ -2509,6 +2515,25 @@ int freadMain(freadMainArgs _args) {
...
+    dropFill = (int *)malloc((size_t)ndropFill * sizeof(int));
+    int i=0;
+    for (int j=max_col; j<ncol; ++j) {
+      type[j] = CT_DROP;
+      size[j] = 0;
+      ndrop++;
+      nNonStringCols--;
+      dropFill[i++] = j;
+    }
+    dropFilledCols(dropFill, ndropFill);
```
MichaelChirico, line 2534: won't this break pydatatable since `dropFilledCols()` is only defined in the R header?

inst/tests/tests.Rraw
ben-schwen: `strrep` is introduced in R 3.3.0 feature

inst/tests/tests.Rraw
ben-schwen: done

inst/tests/tests.Rraw

src/fread.c
ben-schwen: AFAIU the `fread` in `pydatatable` does not link our repo but is based on a hard copy.

ben-schwen:
> Looks pretty good overall! One question, I see all of the linked issues marked as "closed" -- does that mean there's no other way to fulfill those requests? I.e., we might see this issue as providing a _workaround_ for the linked issues, not necessarily addressing them directly. WDYT?

Good point. I will recheck all issues and update the first post on whether we address it directly or provide an acceptable solution. AFAIR I couldn't come up with a better solution than user guesses without double reading (which is something we wanted to avoid)

inst/tests/tests.Rraw
MichaelChirico: Bummer. Let's add a TODO to use it. We can also use a local equivalent for now to make the switch easier later, e.g.

```r
strrep = function(x, times) mapply(function(x, times) paste(rep(x, times), collapse=""), rep_len(x, l <- pmax(length(x), length(times))), rep_len(times, l), USE.NAMES=FALSE)
```

(not strictly necessary)

src/fread.c
MichaelChirico: OK. Anyway, I suppose what's needed is just to add a Cython routine for doing the same but for Python data structure.

NEWS.md
```diff
@@ -22,7 +22,9 @@
...
+
+7. `fread`'s `fill` argument now also accepts an `integer` in addition to boolean values. `fread` always guesses the number of columns based on reading a sample of rows in the file. When `fill=TRUE`, `fread` stops reading when this estimate winds up too low, e.g. when the sampled rows happen to exclude some rows that are even wider, [#2727](https://github.com/Rdatatable/data.table/issues/2727) [#2691](https://github.com/Rdatatable/data.table/issues/2691) [#4130](https://github.com/Rdatatable/data.table/issues/4130) [#3436](https://github.com/Rdatatable/data.table/issues/3436). Providing an `integer` as argument for `fill` allows for a manual estimate of the number of columns instead, [#1812](https://github.com/Rdatatable/data.table/issues/1812) [#5378](https://github.com/Rdatatable/data.table/issues/5378). Thanks to @jangorecki, @christellacaze, @Yiguan, @alexdthomas, @ibombonato, @Befrancesco, @TobiasGold for reporting/requesting, and Benjamin Schwendinger for the PR.
```
MichaelChirico, line 27:
```suggestion
7. `fread`'s `fill` argument now also accepts an `integer` in addition to boolean values. `fread` always guesses the number of columns based on reading a sample of rows in the file. When `fill=TRUE`, `fread` stops reading and ignores subsequent rows when this estimate winds up too low, e.g. when the sampled rows happen to exclude some rows that are even wider, [#2727](https://github.com/Rdatatable/data.table/issues/2727) [#2691](https://github.com/Rdatatable/data.table/issues/2691) [#4130](https://github.com/Rdatatable/data.table/issues/4130) [#3436](https://github.com/Rdatatable/data.table/issues/3436). Providing an `integer` as argument for `fill` allows for a manual estimate of the number of columns instead, [#1812](https://github.com/Rdatatable/data.table/issues/1812) [#5378](https://github.com/Rdatatable/data.table/issues/5378). Thanks to @jangorecki, @christellacaze, @Yiguan, @alexdthomas, @ibombonato, @Befrancesco, @TobiasGold for reporting/requesting, and Benjamin Schwendinger for the PR.
```

R/fread.R
```diff
@@ -22,11 +22,12 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     stopf("Argument 'encoding' must be 'unknown', 'UTF-8' or 'Latin-1'.")
   }
   stopifnot(
-    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill), isTRUEorFALSE(showProgress),
+    isTRUEorFALSE(strip.white), isTRUEorFALSE(blank.lines.skip), isTRUEorFALSE(fill) || is.numeric(fill) && length(fill)==1L && fill >= 0L, isTRUEorFALSE(showProgress),
     isTRUEorFALSE(verbose), isTRUEorFALSE(check.names), isTRUEorFALSE(logical01), isTRUEorFALSE(keepLeadingZeros), isTRUEorFALSE(yaml),
     isTRUEorFALSE(stringsAsFactors) || (is.double(stringsAsFactors) && length(stringsAsFactors)==1L && 0.0<=stringsAsFactors && stringsAsFactors<=1.0),
     is.numeric(nrows), length(nrows)==1L
   )
+  fill=as.integer(fill)
```
MichaelChirico, line 30: this means that `fill = 1L` is handled identically to `fill=TRUE` right? I am not sure there's a use case for the former, but maybe it should be documented?

inst/tests/tests.Rraw
MichaelChirico: NVM, it's overkill.

MichaelChirico:
> > Looks pretty good overall! One question, I see all of the linked issues marked as "closed" -- does that mean there's no other way to fulfill those requests? I.e., we might see this issue as providing a _workaround_ for the linked issues, not necessarily addressing them directly. WDYT?
> 
> Good point. I will recheck all issues and update the first post on whether we address it directly or provide an acceptable solution. AFAIR I couldn't come up with a better solution than user guesses without double reading (which is something we wanted to avoid)

I'm going to go ahead and merge. Please do follow up with the linked issues and (suggestions, not a checklist) (1) file PRs for associated tests (2) comment on the issue itself if you see fit, could be "this issue can be fixed by supplying your own upper bound guess on the # of columns" "please file a follow-up if you're convinced the workaround with `fill=` can be further improved", etc. (3) possibly re-open if you see fit.

R/fread.R
MichaelChirico: Let's handle in follow-up if needed

MichaelChirico (approved):

MichaelChirico:
Great stuff! Not sure we've ever had a PR close _seven_ issues at once before :tada:

R/fread.R
ben-schwen: Yes but the handling is anyway that it uses the max of data.table estimate and user guess.

PR #5133, opened 2021-09-03
names<- retain key and option to turn new warning on

mattdowle:
The existing names<- test didn't check retaining key, fixed.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -1514,11 +1514,11 @@ test(488, data.table(stats::reshape(DT, idvar="ID", timevar="TIME", direction="w
           data.table(ID=1:3,X.1=INT(1,4,7),Y.1=INT(10,13,16),X.2=INT(2,5,8),Y.2=INT(11,14,17),X.3=INT(3,6,9),Y.3=INT(12,15,18)))
 
 # Test warnings for names<- and colnames<-,  but only warnings when caller is data.table aware.
-DT = data.table(a=1:3,b=4:6)
+DT = data.table(a=1:3,b=4:6,key="a")
 test(489, names(DT)[1]<-"A", "A")  # needs R >= 3.1.0, which is stated dependency now
-test(490, names(DT), c("A","b"))
+test(490, DT, data.table(A=1:3, b=4:6, key="a"))  # key wasn't retained in dev after #5084
 test(491, colnames(DT)[2]<-"B", "B")
-test(492, names(DT), c("A","B"))
+test(492, DT, data.table(A=1:3, B=4:6, key="a"))
 
 # Check setnames out of bounds errors
 test(493, setnames(DT,"foo","bar"), error="not found.*foo")


ben-schwen:
Shouldn't it be `key="A"` in y of test 490 and 492 since we change the name of a?

mattdowle:
Yes. Thanks.

ben-schwen:
@mattdowle retaining key is possible with defining
```
`names<-.data.table` = function(x, value) {
  if (!is.null(value)) setnames(x, value)
  else {
    setattr(x, "names", NULL)
    invisible(x)
  }
}
`colnames<-.data.table` = `names<-.data.table`

names.data.table = function(x) {
  copy(attr(x, "names"))
}
colnames.data.table = names.data.table
```
and registering them as `S3method`. 

BUT fiddling with the names of the table with e.g. `names(dt) <- NULL` will prevent calling `names(dt)<-` again since this times `setnames` is called and this call will error because of `if (length(names(x)) != ncol) stopf("x has %d columns but its names are length %d", ncol, length(names(x)))`

I can push to this branch with a build which passes all tests but makes the following changes to `tests.Rraw`
```
names(x) = c('a', 'b') #line 13966 becomes
setattr(x, "names", c('a', 'b'))
#test 1967.72, output becomes "i.a has same type (integer) as x.a. No coercion needed.")
```

I just didn't want to hijack your PR without asking first ;)

mattdowle:
@ben-schwen Thanks for looking.  We already have `names<-` and `S3method` for it. `colnames<-` reaches `names<-` too.  Did you see `names<-` in data.table.R: it's very similar to yours. The problem is the `shallow()` in `names<-` which now doesn't retain the key because R's copy via its `*tmp*` (R's asterix not mine) is copy that selfref detects. We could remove the `shallow()`, but it's the `shallow()` that is reinstating over-allocation after R's copy via `*tmp*` removed the over-allocation.

ben-schwen:
Thanks for the explanation. Removing `shallow` will also lead to segfaulting subsequent `set()` calls so removing it from `names<-` does not seem an option. Maybe we should even add the following testcases to counter the segfault? 

```
test(492.2, DT[, C := 1], data.table(A=1:3, B=4:6, C=1, key="A"))
test(492.3, set(DT, j="D", value = 2), data.table(A=1:3, B=4:6, C=1, D=2, key="A"))
```

mattdowle:
Yes, good point. Adding those tests good idea. Thanks.

It seems that R's `*tmp*` isn't a copy actually when all the names are changed. It's only a copy when a subset of names are changed.  So the tests need to cover those two cases. IIUC.

You can see it happening using `debug("names<-.data.table")` and looking at `address(x)` upon entry.  Or adding a first line to it to `print(address(x))`.

```
DT = data.table(a=1:3, b=4:6)
names(DT)[1] = "A"    # R's *tmp* is a copy;  address(x) != address(DT)
names(DT) = c("P", "Q")  # R's *tmp* isn't a copy; address(x) == address(DT)

DT = data.table(a=1:3)
names(DT)[1] = "A"    # R's *tmp* isn't a copy
```

If I got that right, it feels to me as though R's internals go through an extra step or function call when a subset of names is changed, to construct the new full-length names vector, and that extra step is where the copy comes from. Strange. Just a guess. Would be interesting to look at R's internals on this to see if that's the case.

src/assign.c
```diff
@@ -134,22 +134,27 @@ static int _selfrefok(SEXP x, Rboolean checkNames, Rboolean verbose) {
...
 {
-  // NEW: cols argument to specify the columns to shallow copy on. If NULL, all columns.
-  // called from alloccol where n is checked carefully, or from shallow() at R level
-  // where n is set to truelength (i.e. a shallow copy only with no size change)
+  // cols: the columns to shallow copy on; all if NULL
+  // nSpare: passed by user to setalloccol()/alloc.col(); otherwise <0 means the datatable.alloccol option 
+  if (TYPEOF(dt)!=VECSXP) error(_("dt passed to shallow isn't type VECSXP"));
+  if (!isNull(cols) && !isInteger(cols)) error(_("cols passed to shallow isn't NULL or integer"));
   int protecti=0;
+  const int n = (isNull(cols) ? length(dt) : length(cols)) + (nSpare<0 ? allocColOpt() : nSpare);
```
ben-schwen, line 157: Could move `const int l = isNull(cols) ? LENGTH(dt) : length(cols);` up from line from line 180 and already use it here.

src/assign.c
```diff
@@ -230,57 +245,52 @@ SEXP alloccol(SEXP dt, R_len_t n, Rboolean verbose)
...
-
-    // But don't clear dim and dimnames. Because as from 1.12.4 we keep the matrix column as-is and ask user to use as.data.table to
-    // unpack matrix columns when they really need to; test 2089.2
-    // setAttrib(VECTOR_ELT(ans, i), R_DimSymbol, R_NilValue);
-    // setAttrib(VECTOR_ELT(ans, i), R_DimNamesSymbol, R_NilValue);
-  }
+  const bool verbose = LOGICAL(verboseArg)[0];
+  if (length(nSpareArg)!=1 || (!isInteger(nSpareArg) && !isReal(nSpareArg)))
+    error(_("%s should be a single number but it is length %d and type '%s'"), "getOption('datatable.alloccol') (or an argument with this default)", length(nSpareArg), type2char(TYPEOF(nSpareArg)));
+  int nSpare = isInteger(nSpareArg) ? INTEGER(nSpareArg)[0] : (int)REAL(nSpareArg)[0];
```
ben-schwen, line 269: won't work for `bit64` integer such as in `alloc.col(DT, as.integer64(100))` (quite edgy though)

src/assign.c
mattdowle: Thanks Ben. I have to admit I struggled in this PR so it's great to have more eyes on it helping out. Will come back to this line.

src/assign.c
mattdowle: Indeed. We probably have this problem throughout many arguments in many functions then? On the one hand I agree it's edgy but now you've raised it, ugh, yep. Maybe we need a new helper to deal with this and sweep the whole codebase in a separate PR for int/real arguments that could be int64.

MichaelChirico:
Hey Matt, given the subtlety of the issues here, it would help to add a verbal summary of the PR to the main commit. Right now it's a bit scattered among your responses, the code comments, and the referenced issues -- it would be easier to follow the PR in this case if you could provide such an overview of the reasoning.

mattdowle:
Yes I see. But if I reach that stage to make it easy for you, then it will be done, and I won't need any help. Kind of the very point is that it is scattered and hard to follow, and that's what I need help with. Perhaps you could ask on specific lines for example.

src/assign.c
jangorecki: coerceAs helper was added to handle this kind of coercion

mattdowle: I can see `coerceAs` defined in `wrappers.R` but it doesn't seem in use at R level yet. Yes it would be good to start using it.

jangorecki: AFAIR I added R wrapper for easier tests coverage.

MichaelChirico:
> if I reach that stage to make it easy for you, then it will be done

That makes sense, but even an encapsulation of that in the main post would be helpful, along the lines of

> Here's a draft version, my thinking so far is [...], but I don't have much confidence because of [...]

Something to communicate the current status of the PR in words & not just code... it will be an added friction for you as author but will substantially help the reviewer.

inst/tests/tests.Rraw
```diff
@@ -1320,6 +1320,7 @@ old = getOption("datatable.alloccol")  # Test that unsetting datatable.alloccol
 options(datatable.alloccol=NULL) # In this =NULL case, options() in R 3.0.0 returned TRUE rather than the old value. This R bug was fixed in R 3.1.1.
                                  # This is why getOption is called first rather than just using the result of option() like elsewhere in this test file.
                                  # TODO: simplify this test if/when R dependency >= 3.1.1
+# try() outside test() otherwise test() itself fails on the timings[,:=] step which calls CsubsetDT and uses datatable.alloccol
```
MichaelChirico, line 1323: it might also help to split this PR up, there are a few lines like this that look independent of the fix, streamlining would also help to focus on what's essential to the `names<-` issue

mattdowle: maybe if this change stands then splitting it up would be good, yes, but I hope you can see that the comment itself explains why the change is not independent: the comment says how CsubsetDT and datatable.alloccol usage were invoked and those are integrated with this issue; for example maybe this change won't need to be made once the PR is finished.

ColeMiller1:
For `test.data.table()`, I see 28 errors while running on either RStudio or RGui for this branch. I can work on resolving the issues but I'm not sure if there are problems outside of the test suite. Likewise, the first failing test I see does not have the expected value that I would think and ideally it wouldn't have a key anymore (I've included it below). I also see why this PR needs to add an option for the warning because some of the 28 errors are due to warning.

As for better downstream protection, could the indices be columns instead of attributes? That way any reordering or subset would be protected and still be useable by `data.table` internals and we could still hide them when using `print()`. That would allow us to revisit the 5084 solution, but I'm also not sure if this PR is a direct result of #5084 or is independent of it. 

<details>
<summary>Test 104.2 error </summary>

``` r
# 104.2
library(data.table)
dt <- data.table(A = rep(1:3, each=4), B = rep(11:14, each=3), C = rep(21:22, 6), key = "A,B")
dt1 <- dt2 <- dt
dt1[,"A"] <- 3L;
## test(104.2, dt[, transform(.SD, B=sum(C)), by="A"], data.table(A=rep(1:3,each=4), B=86L, C=21:22, key="A"))
## While test expects A to have 1:3, dt1 was modified which impacts dt. 
dt
##    A  B  C
##  1: 3 11 21
##  2: 3 11 22
##  3: 3 11 21
##  4: 3 12 22
##  5: 3 12 21
##  6: 3 12 22
##  7: 3 13 21
##  8: 3 13 22
##  9: 3 13 21
## 10: 3 14 22
## 11: 3 14 21
## 12: 3 14 22
```

</code>
</details>

jangorecki:
Print is just tip of the iceberg. Imagine lapply( DT, ... )

src/assign.c
```diff
@@ -200,6 +205,16 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   return(newdt);
 }
 
+SEXP shallow(SEXP dt, SEXP cols) {
+  return _shallow(dt, cols, -1);
+}
+
+SEXP copy(SEXP x)
+{
+  return INHERITS(x, char_datatable) ? _shallow(x, R_NilValue, -1) : duplicate(x);
```
ColeMiller1, line 214: This change is responsible for 10 failed tests. Previously, `copy()` only returned `duplicate(x)`.  An additional flag to guarantee copying or maybe even a different function may be helpful, e.g. `SEXP copy(SEXP x, bool doCopy) { return doCopy || !INHERITS(x, char_datatable) ? duplicate(x) : _shallow(x, R_NilValue, -1);}`

One way the tests fail is that the shallow copy returned does not always copy when assigned correctly when `i` is provided. E.g., `copy(dt)[, x := new_val]` would not affect the original dt whereas `copy(dt)[1L, x:= new_val]` would affect the original dt and it would be modified. These examples would be helpful to have a flag to guarantee a copy - here are the 8 tests that fail (980.20,  982.20, 1419.03, 1419.05, 1419.07, 1419.10, 2068.20, 2233.36).

A second way the tests failed is that now `[<-` behaves more similarly to `dt[, x:=newval]` in that other objects pointing to the same address gets similar change when the replacement vector is a different length (i.e., assigning a scalar to a column that is more than 1 value). There are 2 tests like this (104.2 and 113 )

Just as interesting to note is that when reverting `copy` back to simply `return duplicate(x);` there are 3 new errors produced that aren't produced in the original PR (510, 2016.6, 2049.4).  The first and last test are due to #5084 detecting invalid self ref. Specifically, making `<-` assignments and then subsequent assignment calls don't build on one another. e.g., `dt = data.table(x = 1L, y = 1L); key(dt) <- "x"; dt[, y:=2L];` results in `data.table(x = 1L, y = 1L)`. In that example, the first test would typically be correct but then the next text would expect the data.table to be modified but it is still the original.

2016.6, is also related to the invalid self ref, but the cause is because the data.table is loaded instead of modifying data.tables in a non-canonical way. I'll look at this more next weekend. 

```r
## Tests 980.20,  982.20, 1419.03, 1419.05, 1419.07, 1419.10, 2068.20, 2233.36
dt = data.table(x = 1L)
copy(dt)[, x:= x+1L]
dt
## x
## <int>
##   1:     1

copy(dt)[1L, x:=x+1L] ## affects original dt. Any subsequent calls that assume unmodified dt cause errors
dt
## x
## <int>
##   1:     2
```
``` r
## Tests 104.20  113.00
dt = data.table(a = 1:2)
dt1 <- dt
dt1[, "A"] <- 3L ## if this were the same length as initial vector, a copy would be made and match original tests
dt
##        a
##    <int>
## 1:     3
## 2:     3
```

```r
## Reverting copy behavior to always duplicate **change to PR code on local**
## Tests 510, 2049.4
dt = data.table(x = 1L, y = 1L)
key(dt) <- "x"
dt[, y:= 2L] ## Warning about invalid ref
dt           ## No change
## Key: <x>
##        x     y
##    <int> <int>
## 1:     1     1

## Similar issue when loading data from storage
## This still modified to return duplicate copy
## 2016.6
DT = data.table(a=1:3)
save(list="DT", file=tt<-tempfile())
rm(DT)
name = load(tt)
DT[, a:=4L] ## warning due to invalid self ref.
DT
##        a
##    <int>
## 1:     1
## 2:     2
## 3:     3

Comment (jangorecki) [reply src/assign.c:214]: Thanks for digging into this
> whereas `copy(dt)[1L, x:= new_val]` would affect the original dt and it would be modified.

This indeed seems to be unexpected behavior.

**Comment by jangorecki** [comment]
This issue is currently main blocker for releasing master branch to CRAN. If anyone would like to take over this PR you are welcome and encouraged to do so.

File: src/assign.c (hunk 4 above)
Comment (jangorecki) [reply src/assign.c:214]: added tests 492.* to ensure about that

**Comment by MichaelChirico** [comment]
I still am a bit lost on this PR, can anyone at least confirm that we have good tests in place such that once tests go green the PR is ready to go?

**Comment by jangorecki** [comment]
This issue came up from revdeps testing so even if current tests won't be enough then revdeps will be enough.

**Comment by MichaelChirico** [comment]
> This issue came up from revdeps testing so even if current tests won't be enough then revdeps will be enough.

I guess in this case yes, where our main concern is un-breaking regressed functionality. but in general it's much preferable to have minimal tests written by authors who understand the issue more deeply.

**Comment by jangorecki** [comment]
I don't think merging to master is good at that point. IMO better to make PR complete, and then merge to master.

**Comment by jangorecki** [comment]
Rolling back is some idea, but problem that PR fixed was quite severe, silently returning wrong answer. Moreover it may not be limited to dplyr only but to whoever subsets DT without [ dispatch. There is huge amount of work already invested in resolving that, and I am worrying if we roll it back, it will slowly get abandoned. Now feels the best time to work that out. I cannot imagine documentation entry that warns about subset avoiding [ that would not sound scary to 95% of readers.

**Comment by ColeMiller1** [comment]
That is fair - it might still be good to have a general timeline (e.g., 3 months) where if this still isn't resolved, other options are looked at. 

Do you have thoughts on failed test 104.2? The existing test and behavior in 1.14.10 expect this:

``` r
library(data.table)
dt = data.table(A = 1:2, key = "A")
dt1 <- dt

dt1[, "A"] <- 3L
dt1
#>    A
#> 1: 3
#> 2: 3

## original DT unaffected
dt
#>    A
#> 1: 1
#> 2: 2
```
In this PR, the original DT is also updated on the `dt1[, "A"] <- 3L` assignment which causes a failed test. I have found that since many of the failed tests are related to non-idiomatic `data.table` usage, that I am not really sure what the expected results should be. If anything, I would have guessed that the new behavior is how it would have behaved all along.

MichaelChirico:
> In this PR, the original DT is also updated on the `dt1[, "A"] <- 3L` assignment which causes a failed test. I have found that since many of the failed tests are related to non-idiomatic `data.table` usage, that I am not really sure what the expected results should be. If anything, I would have guessed that the new behavior is how it would have behaved all along.

I think this is what Matt was getting at with `cedta()`. If `!cedta()`, the calling code might be running `dt1[, "A"] <- 3L` _without knowing that the `[<-.data.table` is being dispatched_. E.g., this could be in a package that has only checked `is.data.frame()` and so runs the same code on data.frame as data.table.

If `cedta()`, the author should know better, and we should throw a warning that we can later upgrade to an error. Otherwise, we should try to do what `[<-.data.frame` would do, rather than introduce reference semantics to code that is not aware of it (could have bad implications).

MichaelChirico:
Reading a bit more carefully, I see `[<-.data.table` already has a `!cedta()` branch:

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2139-L2146

In fact, several methods do:

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2202-L2206

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2234-L2239

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2243-L2245

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2265-L2269

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2290-L2293

`names<-.data.table` is thus a bit of an outlier for not having such a branch:

https://github.com/Rdatatable/data.table/blob/6782251bcbc38c5240bb41f3af1b5f7436dac948/R/data.table.R#L2253-L2263

PR #5134, opened 2021-09-05
shift type="cyclic"

### LINKED ISSUE #4451: shift type=cyclic
The Syntax of shift() is
`shift(x, n=1L, fill=NA, type=c("lag", "lead", "shift"), give.names=FALSE)`
For n=1, one can use fill = x[.N] to put the last value on top of the column, so one can "rotate" the values by 1.

When n>2, x[.N] will only repeat the last value in all the places to fill. An for me intuitive way to use shift would be 
`n = a, fill = x[(.N-a):.N] `
which should then continue this "rotating" behavior. If it is attempted, the result is

> Error : fill must be a vector of length 1 

I would greatly appreciate if you could incorporate this functionality.

ben-schwen:
closes #4451

- [x] R and C
- [x] tests
- [x] man
- [x] NEWS

### Benchmarks (against naive R implementation)

```R
library(microbenchmark)
library(data.table)

x = seq.int(1e5)
microbenchmark(times = 10L,
  shift(x, 1, type="cyclic"),
  c(tail(x, 1), head(x,-1))
)
#> Unit: microseconds
#>                          expr     min      lq     mean   median      uq     max  neval
#>  shift(x, 1, type = "cyclic") 157.968 174.893 201.3697 184.7235 215.843 334.139     10
#>    c(tail(x, 1), head(x, -1)) 722.819 748.573 782.2081 758.9530 793.041 966.274     10

x = seq.int(1e6)
microbenchmark(times = 10L,
  shift(x, 1, type="cyclic"),
  c(tail(x, 1), head(x,-1))
)
#> Unit: milliseconds
#>                          expr      min       lq     mean    median       uq        max neval
#>  shift(x, 1, type = "cyclic") 1.597685 1.855232  2.30239  1.946397  2.39731   4.476933    10
#>    c(tail(x, 1), head(x, -1)) 6.788656 9.704511 10.46109 10.582472 10.92852  12.863361    10

x = seq.int(1e7)
microbenchmark(times = 10L,
  shift(x, 1, type="cyclic"),
  c(tail(x, 1), head(x,-1))
)
#> Unit: milliseconds
#>                          expr      min       lq      mean   median        uq        max neval
#>  shift(x, 1, type = "cyclic") 17.63109 17.72950  21.77189 20.16767  24.55881   31.33855    10
#>    c(tail(x, 1), head(x, -1)) 79.33389 86.99153 101.25132 92.47582 122.72189  125.07466    10

x = seq.int(1e8)
microbenchmark(times = 10L,
  shift(x, 1, type="cyclic"),
  c(tail(x, 1), head(x,-1))
)
#> Unit: milliseconds
#>                          expr       min        lq      mean    median        uq        max neval
#>  shift(x, 1, type = "cyclic")  173.5632  189.0297  221.7498  210.6784  227.4175   368.0429    10
#>    c(tail(x, 1), head(x, -1)) 1041.4718 1050.3436 1076.6787 1065.8843 1088.9624  1151.8256    10

x = seq.int(1e9)
microbenchmark(times = 10L,
  shift(x, 1, type="cyclic"),
  c(tail(x, 1), head(x,-1))
)
#> Unit: seconds
#>                          expr       min        lq      mean    median        uq        max neval
#>  shift(x, 1, type = "cyclic")  1.643138  1.746724  2.367037  2.107993  2.550822   3.940544    10
#>    c(tail(x, 1), head(x, -1)) 10.599029 11.178644 11.620190 11.534894 11.882139  13.054914    10
```

### DIFF
--- a/R/shift.R
+++ b/R/shift.R
@@ -1,4 +1,4 @@
-shift = function(x, n=1L, fill=NA, type=c("lag", "lead", "shift"), give.names=FALSE) {
+shift = function(x, n=1L, fill=NA, type=c("lag", "lead", "shift", "cyclic"), give.names=FALSE) {
   type = match.arg(type)
   stopifnot(is.numeric(n))
   ans = .Call(Cshift, x, as.integer(n), fill, type)
@@ -9,7 +9,7 @@ shift = function(x, n=1L, fill=NA, type=c("lag", "lead", "shift"), give.names=FA
       else nx = paste0("V", if (is.atomic(x)) 1L else seq_along(x))
     }
     else nx = names(x)
-    if (type!="shift") {
+    if (!(type %chin% c("shift", "cyclic"))) {
       # flip type for negative n, #3223
       neg = (n<0L)
       if (type=="lead" && length(unique(sign(n))) == 3L) neg[ n==0L ] = TRUE   # lead_0 should be named lag_0 for consistency (if mixing signs of n, #3832)
--- a/R/shift.R
+++ b/R/shift.R
@@ -1,5 +1,7 @@
-shift = function(x, n=1L, fill=NA, type=c("lag", "lead", "shift", "cyclic"), give.names=FALSE) {
+shift = function(x, n=1L, fill, type=c("lag", "lead", "shift", "cyclic"), give.names=FALSE) {
   type = match.arg(type)
+  if (type == "cyclic" && !missing(fill)) warning(sprintf("Provided argument fill=%s will be ignored since type='shift'.",fill))
+  if (missing(fill)) fill = NA
   stopifnot(is.numeric(n))
   ans = .Call(Cshift, x, as.integer(n), fill, type)
   if (give.names && is.list(ans)) {
--- a/R/shift.R
+++ b/R/shift.R
@@ -1,6 +1,6 @@
 shift = function(x, n=1L, fill, type=c("lag", "lead", "shift", "cyclic"), give.names=FALSE) {
   type = match.arg(type)
-  if (type == "cyclic" && !missing(fill)) warning(sprintf("Provided argument fill=%s will be ignored since type='shift'.",fill))
+  if (type == "cyclic" && !missing(fill)) warning("Provided argument fill=", fill, " will be ignored since type='shift'.")
   if (missing(fill)) fill = NA
   stopifnot(is.numeric(n))
   ans = .Call(Cshift, x, as.integer(n), fill, type)
--- a/man/shift.Rd
+++ b/man/shift.Rd
@@ -10,14 +10,14 @@
 }
 
 \usage{
-shift(x, n=1L, fill=NA, type=c("lag", "lead", "shift", "cyclic"), give.names=FALSE)
+shift(x, n=1L, fill, type=c("lag", "lead", "shift", "cyclic"), give.names=FALSE)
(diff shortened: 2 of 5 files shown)


R/shift.R
```diff
@@ -9,7 +9,7 @@ shift = function(x, n=1L, fill=NA, type=c("lag", "lead", "shift"), give.names=FA
       else nx = paste0("V", if (is.atomic(x)) 1L else seq_along(x))
     }
     else nx = names(x)
-    if (type!="shift") {
+    if (!(type %chin% c("shift", "cyclic"))) {
```
MichaelChirico, line 12: should we error (warn?) if `fill` is no missing for cyclic?

src/shift.c
```diff
@@ -4,7 +4,7 @@
 SEXP shift(SEXP obj, SEXP k, SEXP fill, SEXP type)
 {
   int nprotect=0;
-  enum {LAG, LEAD/*, SHIFT, CYCLIC*/} stype = LAG; // currently SHIFT maps to LAG and CYCLIC is unimplemented (see comments in #1708)
+  enum {LAG, LEAD/*, SHIFT*/,CYCLIC} stype = LAG; // currently SHIFT maps to LAG (see comments in #1708)
```
MichaelChirico, line 7: this looks a bit odd -- the old implied implementation is that SHIFT would come before CYCLIC, the current change is not back-compatible w that design goal right? should SHIFT be moved after CYCLIC in the comment?

R/shift.R
ben-schwen: I also thought about a warning (thinking again about it, error seems a better choice) if both `fill` with an argument and `type="cyclic"` are used, but then I gave up the thought, because for me there is no ambiguity of how cyclic should work other than wrap around like Python's `numpy.roll`.

src/shift.c
ben-schwen: I thought about leaving it like it and not inserting CYCLIC immediately after LEAD because its also the ordering within the R function definition.

R/shift.R
MichaelChirico: agree it's pretty unambiguous, I am thinking of a programmatic edge case where the user inadvertently supplies both in a loop, etc. a warning can alert them to possibly unintended behavior from their code. (thinking again a warning is definitely better than an error)

jangorecki:
Very nice. I would just scale up benchmarks to have one at least at seconds scale. Now biggest is still microseconds.

man/shift.Rd
```diff
@@ -40,6 +40,8 @@ shift(x, n=1:2, fill=0, type="lag")
 # getting a window by using positive and negative n:
 shift(x, n = -1:1)
 shift(x, n = -1:1, type = "shift", give.names = TRUE)
+# cyclic shift where pad uses pushed out values
+shift(x, n = -1:1, type = "shift")
```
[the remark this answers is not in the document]
ben-schwen: True, copy-paste trapped me.

mattdowle:
Looks great, thanks. Looked at comments and it seems everything has been considered, thanks all. I saw you scaled up the benchmark and that result (11 vs 2 seconds) looks amazing.  All I can think is it would be nice to add that speedup to the news item as an example. If you want to, then as a follow-up PR please. Merging now to make progress.

PR #5172, opened 2021-09-22
clang 13 omp

### LINKED ISSUE #5122: omp.h before Rincludes.h
Received from Prof Brian Ripley today :

Subject: Packages including omp.h after Rincludes.h

That is

AMORE BNN BRISC OpenMPController RFCCA RFpredInterval RLT
RandomFields RandomFieldsUtils RandomForestsGLS RhpcBLASctl RxODE
STB TMB brnn coop data.table excursions float glinternet glmmTMB
gmGeostats graphscan hashr hutilscpp influenceR iq kit likeLTD
mRMRe magick meanShiftR meanr mnlogit poismf poppr praznik rFerns
randomForestSRC rasterKernelEstimates rehh simputation spBayes
spNNGP stringdist sylcount unfoldr

Compilation logs can be seen at
https://www.stats.ox.ac.uk/pub/bdr/clang13/ but beware these may only
show the first error.

The imminent clang 13.0.0 has an omp.h including
```
/* OpenMP 5.1 */
#if defined(_OPENMP) && _OPENMP >= 201811
  #pragma omp begin declare variant match(device={kind(host)})
  static inline int omp_is_initial_device(void) { return 1; }
  #pragma omp end declare variant
  #pragma omp begin declare variant match(device={kind(nohost)})
  static inline int omp_is_initial_device(void) { return 0; }
  #pragma omp end declare variant
#endif
```

whose use of match causes a compilation error if included after
Rinternals.h.  'Writing R Extensions' presents various workarounds, the
main one being the strong advice to include system headers before R headers.

Please correct before 2021-09-21 to safely retain your package on CRAN.

mattdowle:
Closes #5122

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: data.table
-Version: 1.14.1
+Version: 1.14.3
 Title: Extension of `data.frame`
 Authors@R: c(
   person("Matt","Dowle",           role=c("aut","cre"), email="mattjdowle@gmail.com"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 **Benchmarks are regularly updated: [here](https://h2oai.github.io/db-benchmark/)**
 
-# data.table [v1.14.1](https://github.com/Rdatatable/data.table/milestone/20)  (in development)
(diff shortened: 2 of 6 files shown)


mattdowle:
@jangorecki 1.14.2 will be just this patch. I'll backport it locally so I'm not intending to `git tag` v1.14.2. Hope that's ok.

Issue #1534, opened 2016-02-11
shift() in data.table v1.9.6 is slow for many groups


nachti:
Hi there!
For many different groups in `by`, `shift` is much slower than manual shifting.
See: http://stackoverflow.com/questions/35179911/shift-in-data-table-v1-9-6-is-slow-for-many-groups
and https://github.com/nachti/datatable_test/blob/master/leadtest.R for a detailed example.
Cheers,
Gerhard

arunsrinivasan:
That's not surprising. This'll go away when `gforce` is optimised for `:=`. It's on the list for this release, I believe.

ben519:
+1 for this performance enhancement.  `shift()` is the main bottleneck in a lot of my code.  Seems that for a fixed number of rows, the time it takes to run `shift()` is proportional to the number of groups in the data. 

```
library(data.table)

# Build table to store timings
timings <- CJ(RowCount = 10^7, Groups = 10^c(0:7))
timings[, SizePerGroup := RowCount/Groups]

# Loop through each experiment
for(i in 1:nrow(dt)){
  print(paste0("Iteration: ", i))
  
  # Build dataset
  timings_i <- timings[i]
  dt <- data.table(Grp = rep(seq_len(timings_i$Groups), each = timings_i$SizePerGroup))
  dt[, Value := sample(100, size = .N, replace = T)]
  
  # Measure the time it takes to insert a column indicating the previous value by group
  elapsed <- system.time(dt[, PrevValueByGrp := shift(Value, type = "lag"), by = Grp])["elapsed"]
  timings[i, Elapsed := elapsed]
}

library(ggplot2)
ggplot(timings, aes(x = Groups, y = Elapsed))+geom_line()+geom_point()
```

<img width="669" alt="screen shot 2018-11-10 at 1 08 15 pm" src="https://user-images.githubusercontent.com/723099/48305066-b7fb1b00-e4e9-11e8-82a0-193e84cea853.png">

franknarf1:
@ben519 Fyi, for the special case of when your code looks like that, there's a shortcut:

```
library(data.table)
dt <- data.table(Grp = rep(seq_len(1e6), each=10L))
dt[, Value := sample(100L, size = .N, replace = TRUE)]

system.time(dt[, PrevValueByGrp := shift(Value, type = "lag"), by = Grp][])
#    user  system elapsed 
#   19.50    0.80   20.34
[...]

PR #5177, opened 2021-09-25
shift cyclic: add benchmark to NEWS

ben-schwen:
Follow up to #5134 as mentioned [here](https://github.com/Rdatatable/data.table/pull/5134#issuecomment-926930540).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -156,7 +156,7 @@
     # [1] 5 1 2 3 4
 
     # Benchmark
-    x = seq.int(1e9) # 3.8 GB
+    x = sample(1e9) # 3.7 GB
     microbenchmark::microbenchmark(
       shift(x, 1, type="cyclic"),
       c(tail(x, 1), head(x,-1)),
@@ -165,8 +165,8 @@
     )
     # Unit: seconds
     #                          expr  min   lq mean  median   uq  max neval
-    #  shift(x, 1, type = "cyclic") 1.61 1.66 1.89    1.80 1.92 3.05    10
-    #    c(tail(x, 1), head(x, -1)) 8.94 9.08 9.18    9.13 9.30 9.53    10
+    #  shift(x, 1, type = "cyclic") 1.57 1.67 1.71    1.68 1.70 2.03    10
+    #    c(tail(x, 1), head(x, -1)) 6.96 7.16 7.49    7.32 7.64 8.60    10
     ```
 
 ## BUG FIXES
--- a/NEWS.md
+++ b/NEWS.md
@@ -144,6 +144,7 @@
 27. `shift()` now also supports `type="cyclic"`, [#4451](https://github.com/Rdatatable/data.table/issues/4451). Arguments that are normally pushed out by `type="lag"` or `type="lead"` are re-introduced at this type at the first/last positions. Thanks to @RicoDiel for requesting, and Benjamin Schwendinger for the PR.
 
     ```R
+    # Usage
     shift(1:5, n=-1:1, type="cyclic")
     # [[1]]
     # [1] 2 3 4 5 1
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -153,6 +154,19 @@
     #
     # [[3]]
     # [1] 5 1 2 3 4
+
+    # Benchmark
+    x = seq.int(1e9) # 3.8 GB
```
jangorecki, line 159: This is altrep, so timing suffers from materializing this object. This happens for both base R and data.table so I think we can ignore it. It just may be slower than if you would use altrep+1-1 as the input object.

mattdowle: Good point. But altrep expands on the first time though : 
```
> x = seq.int(1e9)
> .Internal(inspect(x))
@55cf162368e8 13 INTSXP g0c0 [REF(65535)]  1 : 1000000000 (compact)
> y = shift(x, 1, type="cyclic")
> .Internal(inspect(x))
@55cf162368e8 13 INTSXP g0c0 [MARK,REF(65535)]  1 : 1000000000 (expanded)
```
So after the first run, the expanded altrep will be used, iiuc. This is why, iiuc, the max of shift() is 3.05s, an outlier and the first run. I confirmed it is the first run by using system.time() 3 times. If the base R way was tested before shift, then its max would also be an outlier, iiuc, although harder to see because the op takes longer so the extra fixed time for the altrep expansion on the first run isn't as big as a percentage.

> This is an altrep, so timing suffers from materializing this object.

Only the first run suffers, which becomes the max.

> This happens for both base R and data.table so I think we can ignore it.

It doesn't happen for both. It happens for the first run of whichever is tested first. But it can be ignored if we ignore the max column.

@ben-schwen To eliminate the risk of suspicion in the mind of the reader of the news item, how about just using `x=sample(1e9)`. That removes the suspicion too that shift() could be using the property that the input is a sequence to not actually read all of the input.

On my laptop:
```
> x = sample(1e9) # 3.8 GB
  shift(x, 1, type="cyclic"),
  c(tail(x, 1), head(x,-1)),
  times = 5L,
  unit = "s"
)
Unit: seconds
 shift(x, 1, type = "cyclic") 1.427934 1.462322 1.487594 1.469971 1.536985 1.540759     5
   c(tail(x, 1), head(x, -1)) 6.359621 6.575269 6.622709 6.684869 6.746652 6.747135     5
```

mattdowle: I'm glad I peppered my statements with "iiuc", always good to do that. I went and double-checked, and actually the base R way does not expand the altrep.

```
> x = seq.int(1e9)
> .Internal(inspect(x))
@55cf16d3bdf0 13 INTSXP g0c0 [REF(65535)]  1 : 1000000000 (compact)
> system.time(y<-c(tail(x, 1), head(x,-1)))
   user  system elapsed 
  6.261   1.948   8.210 
> .Internal(inspect(x))
@55cf16d3bdf0 13 INTSXP g1c0 [MARK,REF(65535)]  1 : 1000000000 (compact)
```

Repeating the system.time gives a stable timing, including the expansion of the altrep each time.  I had been thinking my laptop had faster memory or something than Ben's (which was a bit puzzling why I was seeing 6.7s and he was seeing 9.1s, a big difference).  It actually seems that the expansion of the altrep takes 1.2s and that was unfair to the base R way as that's happening for each and every one of its runs, but was only happening for the first run of shift().

ben-schwen: I replaced it now with the suggested `x=sample(1e9)` and I agree that shifting with groups is probably the most common use-case. The speedup of 4x `shift` in comparison with a naive implementation is recognizable for vectors of length 1e5 and longer. Using `shift` with many groups is also a known bottleneck #1534 and it appears that `shift` is even worse than a naive implementation (I would guess due to the massive amount of `memmove`s). I'm already working on a version of `gshift`.

## Benchmarks for using shift with many groups (cylic and lag)
```R
N = 1e7
DT = data.table(x = sample(N), y = sample(1e5,N,TRUE))
  DT[, shift(x, 1, type="cyclic"), y],
  DT[, c(tail(x, 1), head(x,-1)), y],
  times = 10L,
  unit = "s"
)
# Unit: seconds
#                                   expr      min       lq     mean   median       uq      max neval
#  DT[, shift(x, 1, type = "cyclic"), y] 1.951673 1.981829 2.010449 2.001052 2.019000 2.118122    10
#    DT[, c(tail(x, 1), head(x, -1)), y] 1.666980 1.675234 1.693656 1.688608 1.712064 1.721558    10

N = 1e7
DT = data.table(x = sample(N), y = sample(1e5,N,TRUE))
  DT[, shift(x, 1, type="lag"), y],
  DT[, c(NA, head(x,-1)), y],
  times = 10L,
  unit = "s"
)
Unit: seconds
 DT[, shift(x, 1, type = "lag"), y] 1.888065 1.926722 2.036473 1.991713 2.180608 2.292505    10
        DT[, c(NA, head(x, -1)), y] 1.138030 1.156684 1.203427 1.179590 1.234704 1.338751    10

```

PR #5180, opened 2021-09-28
Raw support setkey as column (not key)

### LINKED ISSUE #5100: Fully support raw vectors
Raw vectors can be an attractive type for certain low-cardinality variables because of its low memory footprint. However, using raw vectors in `data.table` are not fully supported. For example:

```r
DT <- data.table(x = c(1, 2), y = raw(2))
setkey(DT, x)  # ok
setkey(DT, y)  # not supported

 DT <- data.table(x = c(1, 2, 1), y = raw(3))
setkey(DT, x)  # error Item 2 of list is type 'raw' which isn't yet supported (SIZEOF=1)
```

Is there a reason raw vectors can't be used?

ben-schwen:
Towards #5100.

While enabling the usage of `raw` in functions like `setkey` for columns seems a good idea, I don't see the potential/use case for fully support `raw` in every place of `data.table`. Not even `base` enables e.g. the use of `order` for `raw`. 

Maybe this changes when seeing more use cases of `raw`. The only one I'm currently aware of is its use for serializing/deserializing objects.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15941,11 +15941,17 @@ test(2067.8, shift(list(z, 1L:3L), n=-1, type = 'cyclic'), list(c(z[2:3], z[1]),
 DT = data.table(a = 2:1, z = complex(0, 0:1))
 test(2068.1, setkey(copy(DT), a), data.table(a=1:2, z=complex(0, 1:0), key='a'))
 test(2068.2, DT[ , abs(z), by=a], data.table(a=2:1, V1=c(0, 1)))
-# raw continues not to be supported
+# support for ordering tables with raw columns, #5100
 DT = data.table(ID=2:1, r=as.raw(0:1))
-test(2068.3, setkey(DT, ID), error="Item 2 of list is type 'raw'")
+test(2068.3, setkey(copy(DT), ID), data.table(ID=1:2, r=as.raw(1:0), key='ID'))
+DT = data.table(x=c(1, 2, 1), y=raw(3))
+test(2068.4, setkey(copy(DT), x), data.table(x=c(1,1,2), y=raw(3), key='x'))
+test(2068.5, DT[, y[.N], x], data.table(x=c(1,2), V1=raw(2)))
+# expression continue to be not supported
+DT = data.table(ID=2:1, r=expression(1, 2))
+test(2068.6, setkey(DT, ID), error="Item 2 of list is type 'expression'")
 # setreordervec triggers !isNewList branch for coverage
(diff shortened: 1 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -15941,11 +15941,17 @@ test(2067.8, shift(list(z, 1L:3L), n=-1, type = 'cyclic'), list(c(z[2:3], z[1]),
 DT = data.table(a = 2:1, z = complex(0, 0:1))
 test(2068.1, setkey(copy(DT), a), data.table(a=1:2, z=complex(0, 1:0), key='a'))
 test(2068.2, DT[ , abs(z), by=a], data.table(a=2:1, V1=c(0, 1)))
-# raw continues not to be supported
+# support for ordering tables with raw columns, #5100
 DT = data.table(ID=2:1, r=as.raw(0:1))
-test(2068.3, setkey(DT, ID), error="Item 2 of list is type 'raw'")
```
MichaelChirico, line 15946: keep an eye on codecov here -- one of the reasons I hesitate a bit on #5100 is I used raw several times as the go-to "coverage test for unsupported type". which isn't a good reason not to implement features, but means extra overhead of rewriting coverage tests

src/reorder.c
```diff
@@ -13,7 +13,7 @@ SEXP reorder(SEXP x, SEXP order)
     ncol = length(x);
     for (int i=0; i<ncol; i++) {
       SEXP v = VECTOR_ELT(x,i);
-      if (SIZEOF(v)!=4 && SIZEOF(v)!=8 && SIZEOF(v)!=16)
+      if (SIZEOF(v)!=1 && SIZEOF(v)!=4 && SIZEOF(v)!=8 && SIZEOF(v)!=16)
```
MichaelChirico, line 16: I might do this test last since it's the least common

inst/tests/tests.Rraw
ben-schwen: I saw and acknowledged that but also used `expression`s as replacement for raw to not drop coverage (see test 2068.6).

Maybe there is even a better replacement than `expression` when looking at [R.internals.h](https://github.com/wch/r-source/blob/0cda2520430ba3ffb80073503bbd2651c4bb8866/src/include/Rinternals.h#L143-L174)

src/reorder.c
ben-schwen: I also had this in mind but was not sure if this potential minor speedup was worth loosing some kind of cleanliness of the kind. Same goes for checking `SIZEOF(x)` and ofc adding `raw` not as first if but as else branch below.

Issue #2490, opened 2017-11-17
rolling join is unexpectedly slow


ben519:
It seems that joining two tables based on non-roll columns and then subsetting the result based on the roll condition works faster than doing it all in one step.  For example

```
set.seed(0)
dt <- data.table(
  A = sample(x = 10^2, size = 10^8, replace = TRUE),
  B = sample(x = 10^4, size = 10^8, replace = TRUE),
  C = sample(seq.Date(as.Date("2000-01-01"), as.Date("2010-12-31"), by = "day"), size = 10^8, replace = TRUE)
)
jt <- data.table(A = 94L, B = 4778L, C = as.Date("2005-01-01"))


# Join on (A, B), then subset rows >= 2005-01-01
setindex(dt, "A", "B")
system.time(result1 <- dt[jt[, list(A, B)], on=c("A", "B"), nomatch = 0][C >= as.Date("2005-01-01")])

# user  system elapsed 
# 0.002   0.000   0.002 


# Rolling join on (A, B, C)
setindex(dt, "A", "B", "C")
system.time(result2 <- jt[dt, on=c("A", "B", "C"), roll = TRUE, nomatch = 0])

# user  system elapsed 
# 10.943   0.809  11.793 

all.equal(result1, result2)  # TRUE
```

Why is the second join so much slower than the first in this example?  Thanks!

```
sessionInfo()
R version 3.4.2 (2017-09-28)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Sierra 10.12.6

```

franknarf1:
With `jt[dt]` you are looking up every row of `dt` in `jt`; and vice versa for `dt[jt]`. It makes sense that the former would take longer since `dt` is so much larger, right? 

I'd guess that the corresponding nonequi join should be faster than it is though:

```
system.time(result3 <- dt[jt, on=.(A,B,C>=C), .(A, B, C = x.C)]) 
# 17 sec vs 12 for your rolling join
identical(result1, result3) 
# TRUE
```

Looking at the verbose output, apparently the nonequi join doesn't use the index.

> Non-equi join operators detected ... 
>   forder took ... 10.53 secs
>   Generating group lengths ... done in 3.54 secs
>   Generating non-equi group ids ... done in 1.93 secs
>   Found 1 non-equi group(s) ...
> Starting bmerge ...done in 0 secs
> Detected that j uses these columns: A,B,x.C 
> 

[...]

PR #5187, opened 2021-09-30
faster bmerge numeric and roll

mattdowle:
Thanks to a report comparing a single column case to `base::findInterval`, I had a look.
Some low hanging fruit was that INHERITS() (an R API call looking up class attribute and looping through the class vector) was being calling deeply in the recursive call to support integer64.  Also the XIND branch, although a trinary, is called in many places and the branch could be raised. `dtwiddle()` has long been sand in the cogs given that the default rounding is 0 anyway, and there was a todo to get NA for double out of the way up front. Tackling those yielded a 2x speedup. There's too much recursion in the case of a large table rolling to small (many rows rolling to the same row) and I'm hoping for another 2x speedup on that.
Related #2490


MichaelChirico:
LMK if you'd like an extra pair of eyes on this / when it's ready

mattdowle:
Status so far in the branch on this particular large-to-small benchmark. Low hanging fruit without tackling the too-much-recursion yet. The code rework allows the too-much-recursion to be done next.

```R
setDTthreads(1)
n <- 1e7 
x.df <- data.frame(date = as.POSIXct("2000-1-1") + 1:n, price = rnorm(n))
x.dt <- data.table(x.df, key = "date")

n2 <- 100
y.df <- data.frame(date = sort(sample(x.df$date, n2)), adjustment = runif(n2))
y.dt <- data.table(y.df, key = "date")

falign <- function(x, y) {
  i <- findInterval(as.numeric(x$date), as.numeric(y$date))
  i[which(i == 0)] <- NA
  x$price.adj <- x[, "price"] + y[i, "adjustment"]
  x
}

falign.dt <- function(x, y) {
  i <- findInterval(as.numeric(x$date), as.numeric(y$date))
  i[which(i == 0)] <- NA
  x[, price.adj := x[, "price"] + y[i, "adjustment"]]
  x
}

microbenchmark(
  falign(x.df, y.df),
  falign.dt(x.dt, y.dt),
  x.dt[, price.adj := y.dt[x.dt, price+adjustment, roll=TRUE]],
  times=10, unit="s"
))
```

```
# v1.14.0
Unit: seconds
                                                falign(x.df, y.df) 0.1488798 0.1542454 0.1617588 0.1593861 0.1620447 0.1992783    10
                                             falign.dt(x.dt, y.dt) 0.2815458 0.2901276 0.2958419 0.2960433 0.3009906 0.3185344    10
 x.dt[, `:=`(price.adj, y.dt[x.dt, price + adjustment, roll = T])] 0.8177566 0.8206021 0.8365346 0.8349606 0.8461528 0.8593044    10

# this branch as of now (3rd line down from 0.83 to 0.57 but still a lot slower than `findInterval`) : 
                                                falign(x.df, y.df) 0.1460495 0.1486709 0.1554336 0.1537538 0.1616508 0.1732326    10
                                             falign.dt(x.dt, y.dt) 0.2766880 0.3006418 0.3010593 0.3025785 0.3060092 0.3161539    10
 x.dt[, `:=`(price.adj, y.dt[x.dt, price + adjustment, roll = T])] 0.5418650 0.5635949 0.5735462 0.5717389 0.5876390 0.5971523    10
```

MichaelChirico:
Anyone want to take over this PR? @jangorecki and @tlapak are the only two I see with commits to this file in recent memory.

jangorecki:
This PR will have conflicts with mergelist, which has been already merged and conflicts resolved to recent master

tdhock:
@anirban166 @DorisAmoakohene please investigate speedups with atime

MichaelChirico:
Marking as draft over `label:WIP`

tdhock:
typically it is better to add atime test in follow-up pr (with merge commit SHAs)

MichaelChirico:
> typically it is better to add atime test in follow-up pr (with merge commit SHAs)

Shouldn't we prefer to see the proposed performance change before merging?

This is perhaps best handled with a stacked-PR workflow (`jj`) -- the new atime commit(s) are set as child(ren) of the current PR so that the change & its performance are both visible and kept in sync.

OTOH, for something like this that's showing performance improvement of an existing use case, we _could_ also recommend merging the atime baseline first independently...

tdhock:
yes you could add the atime code in this Pr, but then we would need another PR later anyway to fix it to a merge commit.

So we don’t forget to use merge commits in atime test cases, I think it would be better to encourage a two-PR workflow. in this first Pr you can post the atime_versions() code and result in a PR comment, then in the second Pr, adapt the code as a new test case. 
stacked Prs are possible yes.

about adding an atime test for a performance improvement of existing use case, of course that is a good idea, but then you would need three PRs (first to add test, next for performance improvement, last for adding merge sha to test).

NEWS.md
```diff
@@ -12,6 +12,8 @@
 
 3. options `"datatable.old.matrix.autoname"` is now `FALSE` by default, meaning `names(data.table(x=1, cbind(1)))` is now `c("x", "V2")`. Toggle the option to retain the old behavior for now; future releases will work to remove this possibility. See the release notes for 1.18.0, item 1 under `NOTE OF INTENDED FUTURE POTENTIAL BREAKING CHANGES`.
 
+4. Joins on numeric (`double`, including `POSIXct`) columns are no longer affected by `setNumericRounding()`, and always use exact matching as with the default `setNumericRounding(0L)`. If you relied on `setNumericRounding(1L)` or `setNumericRounding(2L)` to match nearly-equal numeric values in joins, round those columns explicitly before joining. `setNumericRounding()` continues to affect grouping and ordering. Numeric joins, especially rolling joins, are faster as a result. Thanks @jangorecki for the PR.
```
jangorecki, line 15: I think credits needs to go to Matt in the last sentence here.

Also considering impact on join, I think we could put it in breaking changes section stating "if you join by double columns..."

PR #5189, opened 2021-10-01
shift fix type coercion for integer64

### LINKED ISSUE #4865: shift() with data type integer64
Hello data.table team,

Here's an issue with the shift operator when used with data type integer64.

With type integer there are no issues:
```
d1 <- data.table(a=1:4)
d1[,b:=shift(a,fill=0L)]
d1[]                                        ## works as expected
   a b
1: 1 0
2: 2 1
3: 3 2
4: 4 3

```
Now with type integer64:
```
library(bit64)
d2 <- data.table(a=as.integer64(1:4))
d2[,b:=shift(a,fill=as.integer64(0))]        ## as.integer64("0") or as.integer64(0L) does not work either
Error in shift(a, fill = as.integer64(0)) : 
  INTEGER() can only be applied to a 'integer', not a 'double'

```
Strictly speaking, the data type of whatever we're filling with should be of the type of the rest of the vector.  But this works:
```
d2 <- data.table(a=as.integer64(1:4))
d2[,b:=shift(a,fill=0L)]
d2[]
   a b
1: 1 0
2: 2 1
3: 3 2
4: 4 3

> sapply(d2,class)
          a           b 
"integer64" "integer64" 
```
Maybe it should throw an error instead?

```
> sessionInfo()
R version 4.0.2 (2020-06-22)

Matrix products: default

locale:

attached base packages:

other attached packages:
[1] bit64_4.0.5       bit_4.0.4         data.table_1.13.0

[1] compiler_4.0.2

```
I have the same issue on a linux box with older versions of R and data.table.

ben-schwen:
Closes #4865 

Also changes type coercion for the `integer` case to switch neatless between `integer`, `double` and `integer64`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -378,7 +378,7 @@
 
 45. `shift(xInt64, fill=0)` and `shift(xInt64, fill=as.integer64(0))` (but not `shift(xInt64, fill=0L)`) would error with `INTEGER() can only be applied to a 'integer', not a 'double'` where `xInt64` conveys `bit64::integer64`, `0` is type `double` and `0L` is type integer, [#4865](https://github.com/Rdatatable/data.table/issues/4865). Thanks to @peterlittlejohn for reporting and Benjamin Schwendinger for the PR.
 
-46. `DT[i, strCol:=classVal]` did not coerce using the `as.character` method for the class resulting in either an unexpected string value or an error like `To assign integer64 to a target of type character, please use as.character() for clarity`. Discovered during work on the previous issue, [#5189](https://github.com/Rdatatable/data.table/pull/5189).
+46. `DT[i, strCol:=classVal]` did not coerce using the `as.character` method for the class, resulting in either an unexpected string value or an error such as `To assign integer64 to a target of type character, please use as.character() for clarity`. Discovered during work on the previous issue, [#5189](https://github.com/Rdatatable/data.table/pull/5189).
 
     ```R
     DT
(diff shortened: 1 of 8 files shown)


NEWS.md
```diff
@@ -376,6 +376,8 @@
 
 44. In v1.13.2 a version of an old bug was reintroduced where during a grouping operation list columns could retain a pointer to the last group. This affected only attributes of list elements and only if those were updated during the grouping operation, [#4963](https://github.com/Rdatatable/data.table/issues/4963). Thanks to @fujiaxiang for reporting and @avimallu and Václav Tlapák for investigating and the PR.
 
+45. `shift(x, fill)` could error if `x` had type `bit64::integer64` and a `fill` argument was also provided, [#4865](https://github.com/Rdatatable/data.table/issues/4865). Thanks to @peterlittlejohn for reporting and Benjamin Schwendinger for the PR.
```
MichaelChirico, line 379: Does  `could` really mean `would` here? If so, maybe reword:

> `shift(x, fill)` no longer fails when `x` is `integer64` and `fill` is provided

`could` makes it seem like sometimes fails, sometimes doesn't, in which case we'd want to explicate better what those failure cases are (and are not)

MichaelChirico (approved):

MichaelChirico:
NB: coverage decrease is spurious

NEWS.md
ben-schwen: In this case its really a `could`, since it worked out `x` being `bit64::integer64` and `fill` being `integer32`. I tweaked the message a bit, but I'm not sure if `integer32` is an appropriate wording for NEWS.

mattdowle:
I tackled the news item by focusing it on the most easily conveyed example: `nafill(xInt64, fill=0)`. Also added the error message so that folk searching for the error message might find it, as per contributing guide point 2 which has gotten quite dense now.
I didn't see why there were two blocks for the tests:  first one being one loop, the 2nd being a nested loop.  I started playing with that and testing in v1.14.2 to see which of the new tests worked before.  I liked the nested loop so tried to make that the only loop to see what would happen.  The revised new test passes in v1.14.2 but now fails in this PR.  I committed anyway, and I'm thinking `coerceAs` could be enhanced now to make these pass again, since it seems to be singleton 0 and NA which could be coerced without warning from complex.  Maybe there'll still be some warnings which should be tested by the tests. I'll keep going on this, hence WIP.

PR #5190, opened 2021-10-01
run plot() in tempdir()

MichaelChirico:
Our tests run in an environment where the test executing directory is not writeable, so plain `plot()` fails when trying to create `Rplots.pdf` locally. Switching to `tempdir()` for the write solves this.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -510,10 +510,13 @@ test(165, subset(DT,a>2), DT[a>2])
 test(166, suppressWarnings(split(DT,DT$grp)[[2]]), DT[grp==2])
 
 # and that plotting works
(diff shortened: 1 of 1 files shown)


mattdowle:
ok got it. I moved the `setwd(tempdir())` up to `test.data.table()` to cover everything now and in the future. Does it still fix the problem at your end? If so will merge.

MichaelChirico:
Confirmed that test passes with this update.

I think I originally tried to keep it local to that one test with the thought in mind that `testthat` also considered moving to running tests in `tempdir()`, but decided against it; on re-read, though, I don't think their concerns are relevant to us (since IIUC they mainly arise from `testthat` being used as a testing framework built for generality): https://github.com/r-lib/testthat/pull/1088.

So yes, this looks better to me.

PR #5192, opened 2021-10-02
improve coverage for gfuns

ben-schwen:
Added testcases for increasing coverage of functions that are `gforce` optimized.

Also added two newlines decreasing coverage to be in line with PR #4299, although I'm not sure if we can really cover those.


MichaelChirico:
I lean towards reverting the newlines there for now, while we still don't have any good answer to the point raised by Matt here:

https://github.com/Rdatatable/data.table/pull/4301#issuecomment-903161605

MichaelChirico (approved):

ben-schwen:
> I lean towards reverting the newlines there for now, while we still don't have any good answer to the point raised by Matt here:
> 
> [#4301 (comment)](https://github.com/Rdatatable/data.table/pull/4301#issuecomment-903161605)

Perfect for me. I reverted now all newlines for `if (nrow != n)` to stay consistent.

PR #5199, opened 2021-10-05
exclude `...` from tables() search

### LINKED ISSUE #5197: tables() fails: argument "..." is missing, with no default
the following code generates an error.  possibly because the current environment has three dots ("...").

````
  a <- function(...) data.table::tables()
  a()
````

the error is

````
Error in get(x, envir = env) : argument "..." is missing, with no default
````

`#` `Output of sessionInfo()`

R version 4.1.0 (2021-05-18)
Running under: Arch Linux

Matrix products: default
BLAS:   /usr/lib/libopenblasp-r0.3.15.so
LAPACK: /usr/lib/liblapack.so.3.10.0

locale:
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=en_US.UTF-8
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_US.UTF-8

attached base packages:

[1] compiler_4.1.0    data.table_1.14.3

### PR DESCRIPTION
**Pull request by MichaelChirico**
Closes #5197

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -2,10 +2,11 @@
 MB = NCOL = NROW = NULL
 
 tables = function(mb=TRUE, order.col="NAME", width=80,
-                   env=parent.frame(), silent=FALSE, index=FALSE)
+                  env=parent.frame(), silent=FALSE, index=FALSE)
 {
   # Prints name, size and colnames of all data.tables in the calling environment by default
-  all_obj = objects(envir=env, all.names=TRUE)
+  # include "hidden" objects (starting with .) via all.names=TRUE, but exclude ... specifically, #5197
+  all_obj = grep("...", objects(envir=env, all.names=TRUE), invert=TRUE, fixed=TRUE, value=TRUE)
   is_DT = which(vapply_1b(all_obj, function(x) is.data.table(get(x, envir=env))))
   if (!length(is_DT)) {
     if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
--- a/R/tables.R
+++ b/R/tables.R
@@ -7,7 +7,7 @@ tables = function(mb=TRUE, order.col="NAME", width=80,
   # Prints name, size and colnames of all data.tables in the calling environment by default
   # include "hidden" objects (starting with .) via all.names=TRUE, but exclude ... specifically, #5197
   all_obj = grep("...", objects(envir=env, all.names=TRUE, sorted=order.col == "NAME"), invert=TRUE, fixed=TRUE, value=TRUE)
-  is_DT = vapply_1b(all_obj, function(x) is.data.table(get(x, envir=env)))
+  is_DT = vapply_1b(mget(all_obj, envir=env), is.data.table)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: R/tables.R [hunk 1]
```diff
@@ -2,10 +2,11 @@
 MB = NCOL = NROW = NULL
 
 tables = function(mb=TRUE, order.col="NAME", width=80,
-                   env=parent.frame(), silent=FALSE, index=FALSE)
+                  env=parent.frame(), silent=FALSE, index=FALSE)
 {
   # Prints name, size and colnames of all data.tables in the calling environment by default
-  all_obj = objects(envir=env, all.names=TRUE)
+  # include "hidden" objects (starting with .) via all.names=TRUE, but exclude ... specifically, #5197
+  all_obj = grep("...", objects(envir=env, all.names=TRUE), invert=TRUE, fixed=TRUE, value=TRUE)
```
MichaelChirico, line 9: btw any reason to use `objects()` here? it looks like an exact duplicate of `ls()`, but probably less well-known...

R/tables.R
```diff
@@ -2,12 +2,13 @@
...
 tables = function(mb=TRUE, order.col="NAME", width=80,
-                   env=parent.frame(), silent=FALSE, index=FALSE)
+                  env=parent.frame(), silent=FALSE, index=FALSE)
 {
   # Prints name, size and colnames of all data.tables in the calling environment by default
-  all_obj = objects(envir=env, all.names=TRUE)
-  is_DT = which(vapply_1b(all_obj, function(x) is.data.table(get(x, envir=env))))
-  if (!length(is_DT)) {
+  # include "hidden" objects (starting with .) via all.names=TRUE, but exclude ... specifically, #5197
+  all_obj = grep("...", objects(envir=env, all.names=TRUE, sorted=order.col == "NAME"), invert=TRUE, fixed=TRUE, value=TRUE)
```
MichaelChirico, line 9: we could also pass `pattern="^([^.]|[.][^.]|[.][.][^.])"` to do it in one call, but that's pretty ugly, and anyway `objects()` itself is just calling `grep()` so there's no efficiency benefit (i.e., `pattern` is not being handled in C), so wrapping in `grep()` does the same thing with a more readable `grep()` call.

R/tables.R
MichaelChirico: yep -- the same:

https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/library/base/R/attach.R#L200

R/tables.R
mattdowle: `objects()` only calls `grep()` when `pattern` is not `missing`, iiuc. But `pattern` is missing in this `objects()` call. So doesn't adding `grep` here add overhead of grep?

R/tables.R
```diff
@@ -2,12 +2,13 @@
...
+  is_DT = vapply_1b(mget(all_obj, envir=env), is.data.table)
```
mattdowle, line 10: changing from `get` to `mget` here meant that `inherits` was changed from TRUE to FALSE because the defaults are different. But probably it should be FALSE anyway so seems like a good change in that respect.

PR #5089, opened 2021-07-30
added ghead/gtail support for n>1

ben-schwen:
Closes #5060 

Adds Gforce support for ghead/gtail and n > 1.

**Note:** This leads to the change of test 3 test cases which explicitly checked the non support of ghead/gtail for n != 0.

Issue #5032, opened 2021-05-31
lapply GForce opt could work also without .SD


jangorecki:
```r
d = as.data.table(iris)

d[, lapply(.SD, sum), by=Species, .SDcols=c('Petal.Length','Sepal.Length'), verbose=TRUE]
#lapply optimization changed j from 'lapply(.SD, sum)' to 'list(sum(Petal.Length), sum(Sepal.Length))'
#GForce optimized j to 'list(gsum(Petal.Length), gsum(Sepal.Length))'
#Making each group and running j (GForce TRUE) ... gforce initial population of grp took 0.000

d[, lapply(list(Petal.Length, Sepal.Length), sum), by=Species, verbose=TRUE]
#lapply optimization is on, j unchanged as 'lapply(list(Petal.Length, Sepal.Length), sum)'
#GForce is on, left j unchanged
#Old mean optimization is on, left j unchanged.
#Making each group and running j (GForce FALSE) ... 
```
similar issue #3815

jangorecki:
also when having manually expanded `lapply` it does work
```r
d[, list(sum(Petal.Length), sum(Sepal.Length)), by=Species, verbose=TRUE]
#lapply optimization is on, j unchanged as 'list(sum(Petal.Length), sum(Sepal.Length))'
#GForce optimized j to 'list(gsum(Petal.Length), gsum(Sepal.Length))'
#Making each group and running j (GForce TRUE) ... gforce initial population of grp took 0.000
```
but it doesn't work if we will want to combine two lists like this (below is just one list, but using `c()` already switches of GForce):
```r
d[, c(list(sum(Petal.Length), sum(Sepal.Length))), by=Species, verbose=TRUE]
#lapply optimization is on, j unchanged as 'c(list(sum(Petal.Length), sum(Sepal.Length)))'
#GForce is on, left j unchanged
#Old mean optimization is on, left j unchanged.
#Making each group and running j (GForce FALSE) ... 
```
so the only thing that works is to manually combine both lists as well
```r
d[, list(sum(Petal.Length), sum(Sepal.Length), mean(Petal.Width), mean(Sepal.Width)), by=Species, verbose=TRUE]
lapply optimization is on, j unchanged as 'list(sum(Petal.Length), sum(Sepal.Length), mean(Petal.Width), mean(Sepal.Width))'
GForce optimized j to 'list(gsum(Petal.Length), gsum(Sepal.Length), gmean(Petal.Width), gmean(Sepal.Width))'
[...]

Issue #3815, opened 2019-09-03
GForce optimisation could be more smart


jangorecki:
```r
d = data.table(x=1:4, y=1:2)
options(datatable.verbose=TRUE)
d[, j=.(min(x)), by=y]                       ## GForce TRUE
d[, j=.(min(x), mean(x)/min(x)), by=y]       ## GForce FALSE
d[, j={.(min(x))}, by=y]                     ## GForce TRUE
d[, j={x<-x; .(min(x))}, by=y]               ## GForce FALSE
d[, j={mn<-min(x); .(mn, mean(x)/mn)}, by=y] ## GForce FALSE
options(datatable.verbose=FALSE)
```
all those cases could be currently optimised

ben-schwen:
I think another case of this also appears in the db-benchmark (now resurrected by duckdblabs at https://github.com/duckdblabs/db-benchmark) as Q2 advanced group by

Since `DT[, .(range_v1_v2=max(v1, na.rm=TRUE)-min(v2, na.rm=TRUE)), by=id3]` could be rewritten using chaining.

With current master this gives a good speedup
``` r
library(data.table)
n = 1e7
DT=data.table(v1=rnorm(n), v2=rnorm(n), id3=sample(1e4, n, TRUE))
bench::mark(
  now = DT[, .(range_v1_v2=max(v1, na.rm=TRUE)-min(v2, na.rm=TRUE)), by=id3],
  gforced = DT[, .(max(v1, na.rm=TRUE), min(v2, na.rm=TRUE)), id3][, .(range_v1_v2=V1-V2), by=id3]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 now           316ms    328ms      3.05    40.3MB     0   
#> 2 gforced       126ms    135ms      6.57     268MB     8.22
```

jangorecki:
I would keep it as it is, benchmark also compares query syntax

ColeMiller1:
@ben-schwen somewhat related to your note, #4412 has some aspects of this but is more focused on logical expression evaluations. From the PR (note the 3 gForce optimized functions combined via `|`):

``` r
bench::mark(
  new_hav = dt[,
               j = .SD,
               by = x,
[...]

Issue #1414, opened 2015-10-29
GForce should be able to work with `:=` as well.


arunsrinivasan:

franknarf1:
Just ran into this today looking at [a question on SO](http://stackoverflow.com/q/37303866/1191259):

    actions = data.table(User_id = c("Carl","Carl","Carl","Lisa","Moe"),
                         category = c(1,1,2,2,1),
                         value= c(10,20,30,40,50))
    users = actions[, other_var := 1, by=User_id]

    # verbose says: the following is not optimized
    users[, value_one := 0 ]
    users[actions[category==1], value_one := sum(value), on="User_id", by=.EACHI, verbose=TRUE]

    # verbose says: the following is optimized
    rbind( 
        actions[category==1], 
        unique(actions[,"User_id", with=FALSE])[, value := 0 ],
    fill=TRUE)[, sum(value), by=User_id, verbose=TRUE]

To me, the first way looks idiomatic, considering the variable needs to end up in `users` in the end.

brodieG:
Just wanted to emphasize that enabling this can allow using GForce effectively for complex expressions, albeit with some work.  For [example I show in this post](https://www.brodieg.com/2019/02/24/a-strategy-for-faster-group-statisitics/#blood-from-a-turnip) how to enable it for:

```
slope <- function(x, y) {
  x_ux <- x - mean(x)
  uy <- mean(y)
  sum(x_ux * (y - uy)) / sum(x_ux ^ 2)
}
```

By doing:

```
DT <- data.table(grp, x, y)
setkey(DT, grp)
DTsum <- DT[, .(ux=mean(x), uy=mean(y)), keyby=grp]
DT[DTsum, `:=`(x_ux=x - ux, y_uy=y - uy)]
DT[, `:=`(x_ux.y_uy=x_ux * y_uy, x_ux2=x_ux^2)]
DTsum <- DT[, .(x_ux.y_uy=sum(x_ux.y_uy), x_ux2=sum(x_ux2)), keyby=grp]
res.slope.dt2 <- DTsum[, .(grp, V1=x_ux.y_uy / x_ux2)]
```
Whereas if GForce was supported in `:=` we could do:
```
DT <- data.table(grp, x, y)
DT[, `:=`(ux=mean(x), uy=mean(y)), keyby=grp]
DT[, `:=`(x_ux=x - ux, y_uy=y - uy)]
[...]

PR #5205, opened 2021-10-09
gshift as gforce optimized shift

### LINKED ISSUE #1534: shift() in data.table v1.9.6 is slow for many groups

ben-schwen:
Closes #1534

- [x] R (optimization "dispatch")
- [x] C lag/lead
- [x] NSE evaluation for fill arguments e.g. `fill=x[.N]`
- [x] C cyclic
- [x] tests
- [x] benchmarks
- [x] NEWS

### Internal optimization
Currently `shift` is only GForce optimized when its fill argument is not a call, to avoid cases where `fill=x[1]`, `fill=x[.N]`, etc. since those would need evaluation within their group.

### General Remark
The gforce optimization of `shift` would be the first function to always return vectors of the same length as `by`. The first step in this direction was made in #5089 by allowing `ghead` to return multiple values per group. Before that every gforce function returned exactly 1 value per group.

The whole `datatable.optimize` optimization code branch has grown pretty fat and might need to be rewritten. This is also related to #1414 #5032 #3815 #735 #523.


src/gsumm.c
```diff
@@ -1162,3 +1167,83 @@ SEXP gprod(SEXP x, SEXP narmArg) {
   return(ans);
 }
 
+SEXP gshift(SEXP x, SEXP nArg, SEXP fillArg, SEXP typeArg) {
```
MichaelChirico, line 1170: shouldn't we prefer to re-use the code in shift.c here? otherwise it'll be a pain to always update code in both places going forward

ben-schwen: I can try to recycle as much as possible but there are some different things to consider about both implementations.

| Basic shift | gshift |
|---|---|
| works on data.tables/vectors/lists | operates on vectors |
| main workhouse is `memmove` | used memory may not be blocked |

src/gsumm.c
```diff
@@ -1162,3 +1162,92 @@ SEXP gprod(SEXP x, SEXP narmArg) {
...
+            const CTYPE val = cycle ? (nosubset ? xd[k] : (irows[k]==NA_INTEGER ? fill : xd[irows[k]-1])) : fill; \
+            ASSIGN;                                                                                               \
+          }                                                                                                       \
+        }                                                                                                         \
+      }                                                                                                           \
+    }
+    switch(TYPEOF(x)) {
+      case LGLSXP:  { int *ansd=LOGICAL(tmp);             SHIFT(int,     LOGICAL,   ansd[ansi++]=val); } break;
+      case INTSXP:  { int *ansd=INTEGER(tmp);             SHIFT(int,     INTEGER,   ansd[ansi++]=val); } break;
+      case REALSXP: { double *ansd=REAL(tmp);             SHIFT(double,  REAL,      ansd[ansi++]=val); } break;
```
mattdowle, line 1241: Was just about to merge (looks awesome!) and then just saw `integer64` missed here (and `test_bit64` added to tests).  Bug fix 45 starts `shift(xInt64, fill=0)` so now that `shift` is optimized I wonder if the `shift` problems of int64 would return. Maybe existing tests of shift int64 don't test it by group then.
Can't think of anything else.

mattdowle (changes requested):

src/gsumm.c
mattdowle: Maybe basic shift could be redirected to call gshift then?  It would have to call gforce prep function first, perhaps in a dummy way to trick gshift into working.  Then it would be common code and all existing tests of shift would cover the new gforce gshift too.
Perhaps not now. Happy to merge this PR as is, subject to integer64. A task could be created to redirect shift to gshift in future.

ben-schwen:
> A `sample(1e6, N, TRUE)` (where N is 1e7 as you have) would be good for the news item; i.e. 1 million groups of 10 rows. You only went to group size 100 so far. The example on S.O. has just 5 rows in each group after all. You can just cut to the chase and test group size 5. Or in the news item just use the test verbatim as reported on S.O. and link to it. That way nobody can say we made up an unrealistic test if we used the test reported from the wild.

Added now both, but either one of them should do.

inst/tests/tests.Rraw
```diff
@@ -18292,3 +18292,43 @@ DT = data.table(A=1:3, key="A")
...
+                i = sample(c(-2L,0L,3L,NA), 20, TRUE),
+                d = sample(c(1.2,-3.4,5.6,NA), 20, TRUE),
+                s = sample(c("foo","bar",NA), 20, TRUE),
+                c = sample(c(0+3i,1,-1-1i,NA), 20, TRUE),
+                l = sample(c(TRUE, FALSE, NA), 20, TRUE),
+                r = as.raw(sample(1:5, 20, TRUE)),
+if (test_bit64) i64 = as.integer64(sample(c(-2L,0L,2L,NA), 20, TRUE)))
+
+nn = list(1, 5, -1, -5, c(1,2), c(-1,1))
+cols = c("y", "i", "d", "s", "c", "l", if (test_bit64) "i64")
```
mattdowle, line 18309: 'r' missed here. i'll change cols to `names(x)[-1]` to avoid the repetition

ben-schwen: "r" is missing on purpose since `gshift` will not work on `RAWSXP`

inst/tests/tests.Rraw
```diff
@@ -18292,3 +18292,41 @@ DT = data.table(A=1:3, key="A")
...
+                c = sample(c(0+3i,1,-1-1i,NA), 20, TRUE),
+                l = sample(c(TRUE, FALSE, NA), 20, TRUE),
+                r = as.raw(sample(1:5, 20, TRUE)),
+if (test_bit64) i64 = as.integer64(sample(c(-2L,0L,2L,NA), 20, TRUE)))
+nn = list(1, 5, -1, -5, c(1,2), c(-1,1))
+testnum = 2224
+for (n in nn) {
+  for (c in names(x)[-1]) {
+    testnum = testnum + 0.001
+    test(testnum, EVAL(sprintf("DT[, shift(%s, %d, type='lag'), by=x]", c, n)),  EVAL(sprintf("DT[, esc(%s, %d, type='lag'), by=x]", c, n)))
```
mattdowle, line 18312: I see that comparing gforce against non-gforce makes sense given all the other existing tests of non-gforce. But that could change in future (e.g. https://github.com/Rdatatable/data.table/pull/5205#discussion_r729489615). `y` should really be a fixed result so the tests are isolated and independent. I see the difficulty in constructing `y` given it's a nice loop through many cases. So we can save an `.Rdata` and compare to that. Will do. This will also also RAW to be easily tested here without having to add RAW to the regular shift.c which was the difficulty here I guess.

inst/tests/tests.Rraw
```diff
@@ -18292,3 +18292,41 @@ DT = data.table(A=1:3, key="A")
...
+    test(testnum, EVAL(sprintf("DT[, shift(%s, %d, type='cyclic'), by=x]", c, n)), EVAL(sprintf("DT[, esc(%s, %d, type='cyclic'), by=x]", c, n)))
+    # check if shift with opposite type is same as shift with n*-1
+    testnum = testnum + 0.001
+    test(testnum, EVAL(sprintf("DT[, shift(%s, %d, type='lag'), by=x]", c, n)),  EVAL(sprintf("DT[, esc(%s, %d, type='lead'), by=x]", c, -n)))
+    testnum = testnum + 0.001
+    test(testnum, EVAL(sprintf("DT[, shift(%s, %d, type='lead'), by=x]", c, n)), EVAL(sprintf("DT[, esc(%s, %d, type='lag'), by=x]", c, -n)))
+  }
+}
+test(2224.51, DT[, shift(i, fill=1:10), by=x], error="fill must be a vector of length 1")
+test(2224.52, DT[, shift(i, type="shift"), by=x], DT[, esc(i, type="shift"), by=x])
```
mattdowle, line 18327: isn't 2224.52 already covered in the loop above (the loop does `c="i"` and type=`"shift"`)?

src/gsumm.c
mattdowle: Interesting. The `coerceAs` is for `fill` but `fill` isn't currently tested by the new gforce test 2224.  A loop of differently typed `fill` needs adding to the test then, and I suspect a specific case here for integer64 will then be needed to pass that?  Can I leave that to you, and the C99 warning, hence WIP again.

ben-schwen:
I extended the tests that we introduced in #5189 to also check the coercion between types for `gshift`. There we also test the casting of `fill` between NA_REAL and NA_INTEGER64 and it works.

In my current understanding it works since under the hood `integer64` is just a "shifted" REALSXP with a class attribute. If we directly write these values into the answer then no problem should occur and the output of `numeric` and `integer64` is depending whether the answer has the class label of `integer64`.

```R
library(bit64)
.Internal(inspect(lim.integer64()))
#> @55a458cea988 14 REALSXP g0c2 [OBJ,ATT] (len=2, tl=0) -4.94066e-324,nan
#> ATTRIB:
#>   @55a457f3c630 02 LISTSXP g0c0 [REF(1)] 
#>     TAG: @55a455431140 01 SYMSXP g1c0 [MARK,REF(34387),LCK,gp=0x6000] "class" (has value)
#>     @55a457c6f040 16 STRSXP g0c1 [REF(65535)] (len=1, tl=0)
#>       @55a458186ff8 09 CHARSXP g1c2 [MARK,REF(1114),gp=0x61] [ASCII] [cached] "integer64"
```

mattdowle (approved):

jangorecki:
I suspect this is causing hard to spot issue: https://github.com/Rdatatable/data.table/issues/5547
What was the reason to add NSE as a part of this PR?

PR #5215, opened 2021-10-11
Internal: fmelt.c clean up checkVars()

tdhock:
Closes 5213

... eventually. So far I have just renamed the tmp* variables to something more informative. Next step is to remove repetition in if/else branches.


tdhock:
OK, cleanup is done. 
- renamed tmp/tmp2/etc to more informative variable names.
- created helper functions to avoid repetition.
- kept original four branches after all.

Issue #4217, opened 2020-01-30
Refresh attributes of sfc column when subsetting from data.table


jplecavalier:
I like to use the `sf` package and `data.table` together. Instead of using the `sf` class, which is built on top of the `data.frame` class, I use the `data.table` class with a column of class `sfc` which contains a vector of simple features with some metadata kept in the object's attributes.

However, when I subset some rows of my `data.table`, those attributes are not automatically updated, which can cause some problems. When subsetting from an `sf`, those attributes are automatically updated.

I have a non-elegant way to get around the problem by manually updating the attributes of the `sfc` column, but I'm wondering if there is a way to make it under the hood.

Let's say I have the following

```
library(sf)
library(data.table)

set.seed(20200130)

data <- data.table(
  id = 1:5,
  point = st_sfc(replicate(5, st_point(c(runif(1), runif(1))), simplify = FALSE))
)

data[]
##     id                        point
##  1:  1 POINT (0.09302893 0.6560987)
##  2:  2  POINT (0.4387638 0.7161379)
##  3:  3 POINT (0.8535522 0.08598417)
##  4:  4   POINT (0.927848 0.3534847)
##  5:  5 POINT (0.9615244 0.07300738)
```

Let's look at the attributes of the `point` column.

```
attributes(data$point)
##  $class
##  [1] "sfc_POINT" "sfc"      
##  
##  $precision
##  [1] 0
##  
##  $bbox
##        xmin       ymin       xmax       ymax 
##  0.09302893 0.07300738 0.96152443 0.71613790 
##  
##  $crs
##  Coordinate Reference System: NA
##  
##  $n_empty
##  [1] 0
```

If I subset some elements, the attributes won't update.

```
subset <- data[1:2]
attributes(subset$point)
##  $class
##  [1] "sfc_POINT" "sfc"      
##  
##  $precision
##  [1] 0
##  
##  $bbox
##        xmin       ymin       xmax       ymax 
##  0.09302893 0.07300738 0.96152443 0.71613790 
##  
##  $crs
##  Coordinate Reference System: NA
##  
##  $n_empty
##  [1] 0
```

However, if I use an `sf` object, they will silently update (you can see it easily with the `bbox` attribute).

```
set.seed(20200130)

data <- st_sf(
  id = 1:5,
  point = st_sfc(replicate(5, st_point(c(runif(1), runif(1))), simplify = FALSE))
)

subset <- data[1:2, ]

attributes(subset$point)
##  $class
[...]

PR #5224, opened 2021-10-16
Add format_list_item() method for simple feature geometry columns

### LINKED ISSUE #2273: compatibility with sf library
I wonder if it would be possible to make `data.table` compatible with  the new `sf` library. The library `sf` is promising to be a game changer for spatial analysis in `R` so it sounds like good idea to bring together the power of both libraries.

Currently, the class of an `sf` object is `"sf"         "data.frame"`  and it brings a column named `geometry` of class `"sfc_MULTIPOLYGON" "sfc" `.

Right now, I think the main incompatibility between dt and sf is that when we convert an `sf` data.frame into a `data.table`  using `setDT()`, the geometry column is ruined and the object is not recognised anymore as an `sf` class.

I'm sure there are other points to take into account when making these two great libraries together so I just wanted to put the ball rolling if someone has not done this before.

JoshOBrien:
By adding `format_list_item.sfg()`, a method for the recently added
`format_list_item()` S3 generic, we can ensure that any simple feature
geometry columns (with elements of class `"sfg"`) in a data.table will
be pretty printed using the **sf** package's `format.sfg()`. To see
what that looks like, see the example below

```r
library(sf)
library(data.table)
nc <- st_read(system.file("shape/nc.shp", package = "sf"))
DT <- data.table(nc)
DT[1:3, .(NAME, FIPS, geometry)]
 ##         NAME  FIPS                       geometry
 ## 1:      Ashe 37009 MULTIPOLYGON (((-81.47276 3...
 ## 2: Alleghany 37005 MULTIPOLYGON (((-81.23989 3...
 ## 3:     Surry 37171 MULTIPOLYGON (((-80.45634 3...
```

Closes #2273 (starting [here](https://github.com/Rdatatable/data.table/issues/2273#issuecomment-942545252)) which includes a discussion of the pros and cons of adding this method, for a data type defined in the `sf` package, to `data.table`.


NEWS.md
```diff
@@ -173,6 +173,8 @@
 
 29. `setkey()` now supports type `raw` as value columns (not as key columns), [#5100](https://github.com/Rdatatable/data.table/issues/5100). Thanks Hugh Parsonage for requesting, and Benjamin Schwendinger for the PR.
 
+30. `format_list_item.sfg()`, an unexported S3 method for the recently added `format_list_item()` generic, supports nice printing of simple feature geometry data.table columns, [#2273](https://github.com/Rdatatable/data.table/issues/2273).
```
MichaelChirico, line 176: I would fold this into the earlier NEWS entry since both are improvements in dev

MichaelChirico (commented):
please add a test as well (unless it's hard to create an sfg object from scratch)

MichaelChirico:
can't we just build a list with class sfg & make a placeholder format.sfg function for the test? it will not be a very good test but it'll catch some potential regressions

MichaelChirico:
we could also add a more meaningful test to inst/tests/other.Rraw

JoshOBrien:
> 
> 
> can't we just build a list with class sfg & make a placeholder format.sfg function for the test? it will not be a very good test but it'll catch some potential regressions

Creating a column of sfg objects without loading **sf** turns out to be easy enough. I can just `dput()` a simple example to a text file in `"inst/tests/"` and then use a call to `dget()` in `"other.Rraw"` to read the object back in.

I can also create a placeholder function `format.sfg()`, though it's hard to do that without copying in a bunch of functions from **sf**. Formatting a an `sfg` object of type `MULTIPOLYGON` in the **sf** package, for example, involves calls to **sf** functions `format.sfg()`, `get_start()`, `st_as_text()`, `st_as_text.sfg()`, `prnt.MULTIPOLYGON()`, `WKT_name()`, `prnt.MatrixListList()`, `prnt.MatrixList()`, `prnt.Matrix()`, and `fmt()`. Formatting other types of `sfg` objects requires additional functions. 

If we only test a data.table containing `MULTIPOLYGON` objects, I can nest the functions listed above into a couple that together give the same result as would be produced by loading **sf**. Is this kind of what you are meaning by providing a placeholder `format.sfg()` function? If so, I could put those in a script file in `"inst/test/"`, to be sourced in in `"other.Rraw"` before the test I create. 

```r
format.sfg <- function (x, ..., width = 30) {
    get_start <- function (x, n = 30) {
        if (is.list(x)) 
            structure(lapply(x, get_start, n = n), class = class(x))
        else head(x, round(n/3))
    }

    if (is.null(width)) 
        width = 30
    if (object.size(x) > 1000 && width > 0) 
        x = get_start(x, n = width)
    pr = st_as_text.sfg(x, ...)
    if (width > 0 && nchar(pr) > width) 
        paste0(substr(pr, 1, width - 3), "...")
    else pr
}

st_as_text.sfg <- function (x, ...) {
    prnt.MULTIPOLYGON <- function (x, ..., EWKT = TRUE) {
        WKT_name <- function (x, EWKT = TRUE) {
            cls = class(x)
            zm = substr(cls[1], 3, 4)
            retval = if (zm == "") 
                cls[2]
            else paste(cls[2], substr(cls[1], 3, 4))
            if (EWKT && !is.null(attr(x, "epsg")) && !is.na(attr(x, "epsg"))) 
                paste0("SRID=", attr(x, "epsg"), ";", retval)
            else retval
        }
        prnt.MatrixListList <- function (x, ...) {
            if (length(x) == 0) 
                empty
            else paste0("(", paste0(unlist(lapply(x, prnt.MatrixList, 
                ...)), collapse = ", "), ")")
        }
	prnt.MatrixList <- function (x, ...) {
            if (length(x) == 0) 
                empty
            else paste0("(", paste0(unlist(lapply(x, prnt.Matrix, ...)), 
                collapse = ", "), ")")
        }
	prnt.Matrix <- function (x, nested_parens = FALSE, ...) {
            pf = function(x, ..., collapse) paste0(fmt(x, ...), collapse = collapse)
            if (nrow(x) == 0) 
                empty
            else if (nested_parens) 
                paste0("((", paste0(apply(x, 1, pf, collapse = " ", ...), 
                    collapse = "), ("), "))")
            else paste0("(", paste0(apply(x, 1, pf, collapse = " ", ...), 
                collapse = ", "), ")")
        }
	fmt <- function (x, ...) sub("^[ ]+", "", sapply(unclass(x), format, ...))
        paste(WKT_name(x, EWKT = EWKT), prnt.MatrixListList(x, ...))
    }
    switch(class(x)[2], MULTIPOLYGON = prnt.MULTIPOLYGON(x, ...),
           stop(paste("no print method available for object of class", 
                      class(x)[2])))
}
```

MichaelChirico:
to clarify, I think we are happy to add sf as a dependency to other.Rraw since it's not tested on our main CI pipeline (only on gitlab which only runs on commits to master). sf might present an interesting case because it'll probably require more system dependencies in the docker image. CMIIW @jangorecki

JoshOBrien:
I just checked, and it looks like 22 of those 29 packages loaded by **sf** also get loaded by **ggplot2**, which is also used in `"other.Rraw"`. So, maybe not so bad. FTR, the additional seven packages currently loaded by **sf** are: **Rcpp**, **units**, **KernSmooth**, **e1701**, **class**, **proxy**, and **classInt**.

So, to clarify, are you fine if all the tests associated with this PR go into `"inst/tests/other.Rraw"`?

MichaelChirico:
Please add a simple facsimile test (i.e., no need for `sf`) for coverage to `inst/tests/tests.Rraw`, and some "real" tests (i.e., loading `sf`) to `inst/tests/other.Rraw`.

I think the bigger concern than the small-ish marginal R package footprint is the relatively heavy set of SystemRequirements that comes with `sf`. I'm pretty sure we'll need to add those reqs to the docker image running the other.Rraw CI:

https://gitlab.com/jangorecki/dockerfiles/-/tree/master/r-base-dev

It may be worth looking at the build size comparison of images with/without those SystemRequirements... Hopefully Jan can weigh in when he gets a chance.

JoshOBrien:
@MichaelChirico and @jangorecki -- I just pushed a commit that adds a facsimile test to `inst/tests/tests.Rraw`. (It also incorporates Michael's other suggested edits and adds an alias for the new method to the appropriate man file.)

If or when you want me to add a test to `inst/tests/other.Rraw` or elsewhere, I'll be happy to do that.

MichaelChirico:
thanks Josh! thanks for your patience on this, it's a slightly odd situation. maybe sf would add tests if they've already got CI running on sf-friendly images?

thinking again, I don't think it makes sense for data.table to test the specific behavior of the sf formatting method any time there's an update to format.sfg, especially at the cost of substantially more complex CI for us. the facsimile test will be enough. maybe sf would add a test on their side?

JoshOBrien:
Hello @tlapak . FR #4217 discusses just that issue.

MichaelChirico:
I see, good news we found a workaround!

Unfortunately I think we'll need to add a branch for the test as it won't work on R<4:

https://github.com/wch/r-source/blob/b168e0004d881a5c522c87b0c6f46abc37b0b4a3/doc/NEWS.Rd#L1611-L1612

It looks like `.S3method` is just calling `registerS3method` under the hood, could you try to use that instead?

MichaelChirico:
thanks again for your patience & investigations. I'm pretty sure our current test suite is registering other methods so you can skip unregistering.

NEWS.md
```diff
@@ -99,7 +99,7 @@
...
+17. `data.table` printing now supports customizable methods for both columns and list column row items, part of [#1523](https://github.com/Rdatatable/data.table/issues/1523). `format_col` is S3-generic for customizing how to print whole columns; `format_list_item` is S3-generic for customizing how to print each row of a list column. The `format_list_item.sfg()` method supports nice printing of simple feature geometry data.table columns (from the sf package), [#2273](https://github.com/Rdatatable/data.table/issues/2273). Thanks to @mllg who initially filed [#3338](https://github.com/Rdatatable/data.table/pulls/3338) with the seed of the idea, @franknarf1 who earlier suggested the idea of providing custom formatters, @fparages who submitted a patch to improve the printing of timezones for [#2842](https://github.com/Rdatatable/data.table/issues/2842), @RichardRedding for pointing out an error relating to printing wide `expression` columns in [#3011](https://github.com/Rdatatable/data.table/issues/3011), and @MichaelChirico for implementing. See `?print.data.table` for examples.
```
MichaelChirico, line 102: @JoshOBrien please cite yourself! thanks

MichaelChirico:
Thanks Matt -- should we just mark the code as #nocov then? that was the main reason I was leaning towards a test in tests.Rraw

mattdowle:
Which lines aren't covered? `print.data.table.R` is 100% covered, contains `format_list_item` and has no `#nocov`.

mattdowle:
Yes exactly, looks good to me.

MichaelChirico:
there is new code in print.data.table though (i.e. format_list_item.sfg), which needs a test if print.data.table is to stay at 100%.

also, as noted above, if we add sf to that first line in other.Rraw, the image that's used to run the GLCI tests is missing the SystemRequirements of sf, so the build will fail as sf can't install.

We could add the requirements but they are quite heavy, I wouldn't be surprised if the image size grows 40% or more.

hopefully the conundrum is clearer

mattdowle:
@JoshOBrien Yes please. Thanks!

@MichaelChirico Why are we adding `format_list_item.sfg` on our side: shouldn't it be a method that is added to `sf`?

JoshOBrien:
@mattdowle That's actually, sort of where this started, with me making note [here](https://github.com/Rdatatable/data.table/issues/2273#issuecomment-942545252) that I'd prepared [this PR](https://github.com/JoshOBrien/sf/commit/e691e8a456ee969fa54946d2214b7dd9f6d6e07e) for the **sf** package. 

That PR itself is pretty ungainly, because it needs to register an S3 method for whose generic (`format_list_item`), is provided by a package that may or may not be loaded. Most of that PR is thus code run upon loading **sf** that whether or not **data.table** is already loaded, and then either registers `format_list_item.sfg` or sets up an 'on attach' hook that will register it if/when **data.table** is subsequently attached.

The other complication (should we go that route) is that I'll will need to further elaborate that code to check the version of **data.table**, and only try to register the method if it's data.table>=1.14.4.

TBH, I think ideally the method probably should be over there in **sf**. It's just that in practice, adding it to **data.table** seemed so much simpler (requiring just the addition of `format_list_item.sfg = function(x, ...) {format(x, ...)}` to the code base) that trying to do this on the **data.table** side seemed appealing. It's in setting up testing that the complexity really blew up, making this now look like a really questionable proposition!

mattdowle:
Thanks. I'm seeing the can of worms now.

I like @MichaelChirico suggestion in https://github.com/Rdatatable/data.table/issues/2273#issuecomment-944043230 : 

> Maybe format_list_item.default could check if there's a format method registered for the class of the object, and run that as a default, and only lacking that back up to the current <{class}[dim]> approach.

Since that would work for any package providing a format method, it's easy to see that's appropriate on the data.table side, and would work for `sf`. But then I see there were problems when that was attempted. Hm.

mattdowle:
@JoshOBrien on your investigation in https://github.com/Rdatatable/data.table/issues/2273#issuecomment-944623933, how about calling the `format` method for the class only if it is one (or null) dimensional?  If it's two+ dimensional then display `<class [nxm]>` as before. Using your test with `nc` just above, I checked  `dim(DT[1,geometry])` is `NULL` so that would work for `sfg`, iiuc.

MichaelChirico:
Thanks Matt, I like the sound of that. Important to remember this is a printing method so some lost efficiency is usually alright since typically were only looking at 10 rows of the table anyway.

so we can try lapply(col, format) and look at the lengths()-equivalent to decide if format is appropriate, falling back to the class[1] approach on current HEAD

MichaelChirico:
OK current version passes existing tests.

Honestly, it looks like we've got to basically choose inclusion or exclusion -- do we want to keep a list of included classes where we know formatting plays well, or do we want to keep a list of excluded classes where we know it doesn't?

Current implementation uses the latter, but it seems pretty fragile to me -- we have to escape S4 objects, then functions/builtins, then plain lists... notice `!class(x) %chin% ...` which is the exclusion rule pretty explicitly.

mattdowle:
Again: I had in mind only calling format if a format method exists for that class, as Josh had already in [this comment](https://github.com/Rdatatable/data.table/issues/2273#issuecomment-944623933) using `getS3method`, and dim() is 1 (or null) as well. I don't see that in this PR as of now. Your example of `list()`, which doesn't have a format method according to `getS3method("format", class="list", optional=TRUE)` makes me think we're not quite on the same page here. When I wrote "Maybe the format method could be run on the first item (or even all items in case there's any variance in output), and if it returns a length 1 character vector then use it." perhaps I should have added the words "as well" there as well to make it clear that was additional to not instead of using getS3method and dim(). I hope that resolves the last comment?

JoshOBrien:
Oops. I didn't mean to push that last (bunch of?) commits.

I've not yet got the hang of working with Git on a project like this, and was trying to get my `local add_format_list.sfg` branch and the one on GitHub under `JoshOBrien:add_format_list_item.sfg` even with the one at `Rdatatable:add_format_list_item.sfg`. I have no idea how to roll that back (if that's required), so will just keep hands off of this for now!

JoshOBrien:
@MichaelChirico Is there anything you'd like me to do to improve this PR (which incorporates Matt's suggestions in [this comment](https://github.com/Rdatatable/data.table/pull/5224#issuecomment-948199956)) ? It now seems to me close to an optimal solution, implemented at the right level of generality. 

Also, FTR, it works nicely for columns of simple feature geometry objects. I haven't added a test for those to `"other.Rraw"`, since that would require adding **sf** (and its dependencies), but am happy to do so if you'd like me to.

MichaelChirico:
Hi Josh, sorry I just haven't gotten around to reviewing it yet.

mattdowle:
I moved the logic up to the column class (`format_col`) rather than each item (`format_list_item`). For `sf` this means that `sf:::format.sfc` is now used rather than `sf:::format.sfg`. That felt right so that for example formatting arguments can be applied across the whole column; e.g. justification can involve padding items depending on the width of other items. Since `sf` has `sf:::format.sfc` (which is currently just a vapply of `format.sfg`) it seems right to leave that up to `sf` and use it. Also should be a bit faster since testing for a format method is done once now on the column rather than for each and every item, although it's understood that efficiency isn't a concern for formatting. If there is no format method for the column class, `format_list_item` still tests each item for a format method and uses that.

mattdowle:
Running test.data.table() a 2nd time fails because `registerS3method()` in `tests.Rraw` persists the method after `test.data.table()` finishes. I've been searching for a way to unregister S3 methods without luck. Calling `registerS3method()` with a NULL method in an attempt to unregister it, fails with an error that NULL is an invalid method.

grantmcdermott:
Thanks to @JoshOBrien and everyone else for sheparding this PR through. I want to make sure that I'm not missing something, however, because I just tested with the latest dev version and the geometry column doesn't print as expected. At least, I thought that formatted print for .sfc/.sfg columns types would be handled automatically. (Is that wrong?)

``` r
data.table::update.dev.pkg()
#> R data.table package is up-to-date at d8dc315eb66bc29c49ac0344550f976d16b153ee (1.14.3)

library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.3.1, PROJ 8.0.1
library(data.table)

nc <- st_read(system.file("shape/nc.shp", package = "sf"))
#> Reading layer `nc' from data source `/usr/lib/R/library/sf/shape/nc.shp' using driver `ESRI Shapefile'
#> Simple feature collection with 100 features and 14 fields
#> Geometry type: MULTIPOLYGON
#> Dimension:     XY
#> Bounding box:  xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> Geodetic CRS:  NAD27
DT <- data.table(nc)
DT[1:3, .(NAME, FIPS, geometry)]
#>         NAME  FIPS geometry
#> 1:      Ashe 37009  <XY[1]>
#> 2: Alleghany 37005  <XY[1]>
#> 3:     Surry 37171  <XY[1]>
```

<sup>Created on 2021-11-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mattdowle:
@grantmcdermott Thanks for testing! You're not missing something and yes it should just work as you expected. If I understand correctly, `data.table::update.dev.pkg()` grabs the last image build by one of the last steps in GLCI, which isn't completing currently (see the red badge on https://github.com/Rdatatable/data.table/wiki). If that's correct, then using devtools to build from latest master should work until the GLCI pipeline is working again.

grantmcdermott:
Super, thanks Matt. Can confirm that everything is working as expected after I installed through remotes (devtools).

Next, my quixotic quest to convince the DT team to enable `options(datatable.print.class=TRUE, datatable.print.keys=TRUE)` by default 😉

JoshOBrien:
@mattdowle This looks terrific, and works on everything I've so far thrown at it. Thrilled to see that it will be in the next release of **data.table**. Thanks!

PR #5227, opened 2021-10-18
groupingsets: new metaprogramming together with named by

### LINKED ISSUE #5206: groupingsets() adding an extra column with named vectors?
```R
groupingsets(
  data.table(iris),
  j = mean(Sepal.Length),
  by = c('Sp' = 'Species'),
  sets = list('Species')
)
     Sp    V1    Species
1: <NA> 5.006     setosa
2: <NA> 5.936 versicolor
3: <NA> 6.588  virginica
```

I'm using latest development version of `data.table` and `R 4.1.0`. This seems not to have been the behaviour about three months ago when I last ran the code that now broke.

ben-schwen:
Closes #5206 

I guess no NEWS is needed since we fixed this before rollout?

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18293,4 +18293,5 @@ test(2223.1, DT[.(4), nomatch=FALSE], data.table(A=integer(), key="A"))
 test(2223.2, DT[.(4), nomatch=NA_character_], data.table(A=4L, key="A"))
 
 # groupingsets by named by argument
-test(2224.1, groupingsets(data.table(iris), j = mean(Sepal.Length), by = c('Species'), sets = list('Species')), as.data.table(iris)[, mean(Sepal.Length), Species])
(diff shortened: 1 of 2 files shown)


inst/tests/tests.Rraw
```diff
@@ -18292,3 +18292,6 @@ DT = data.table(A=1:3, key="A")
 test(2223.1, DT[.(4), nomatch=FALSE], data.table(A=integer(), key="A"))
 test(2223.2, DT[.(4), nomatch=NA_character_], data.table(A=4L, key="A"))
 
+# groupingsets by named by argument
+test(2224.1, groupingsets(data.table(iris), j = mean(Sepal.Length), by = c('Sp'='Species'), sets = list('Species')), as.data.table(iris)[, mean(Sepal.Length), Species])
```
jangorecki, line 18296: These are just few rows so I would add fixed objects to compare to, and not another data.table query.

PR #5242, opened 2021-10-28
fread doc - remove unhosted bz2 files

### LINKED ISSUE #4727: fread doc .bz2 files
At end of `fread` documentation, the commands:
* `download.file("http://stat-computing.org/dataexpo/2009/2008.csv.bz2", destfile="2008.csv.bz2")`
* `fread("http://stat-computing.org/dataexpo/2009/1987.csv.bz2")`

are mentioned. However, both files are not hosted anymore, hence, both commands result in a 404 error.

ben-schwen:
Closes #4727

### DIFF
--- a/man/fread.Rd
+++ b/man/fread.Rd
@@ -271,42 +271,13 @@ system.time(DT <- fread("testbig.csv"))
 
 all(mapply(all.equal, DF, DT))
 
-
-# Real data example (Airline data)
-# https://stat-computing.org/dataexpo/2009/the-data.html
-
-download.file("https://stat-computing.org/dataexpo/2009/2008.csv.bz2",
-              destfile="2008.csv.bz2")
-# 109MB (compressed)
-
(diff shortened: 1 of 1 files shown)


MichaelChirico (approved):
would be better to find an alternative data set since the point was to show off performance on a decent/realistic data set... but barring that it's better to remove

ben-schwen:
> would be better to find an alternative data set since the point was to show off performance on a decent/realistic data set... but barring that it's better to remove

I totally agree that it would be better to find a different dataset but failed to find a suitable one. Possible criteria are that the dataset has bz2 format and its URL should have expected long time support and should not mind the additional web traffic. A good candidate would have been the UCI ML repository but I didn't a suitable dataset there.

MichaelChirico:
Let's see if anyone else can help: https://twitter.com/michael_chirico/status/1457482807232458757

PR #5246, opened 2021-11-01
Gforce weighted.mean

### LINKED ISSUE #3977: GForce optimize weighted.mean
I'm not sure if it is possible to GForce optimize `weighted.mean`? It's quite different from existing GForce functions because it takes two variables.

ben-schwen:
Closes #3977

As pointed in the issue its faster (at least on my multi-core machine) to double call `gsum` due to #3202 than providing an implementation similar to e.g. `gsd`/`gvar`.


jangorecki (commented):
Not enough tests. Add some for NA, maybe NaN as well (both in x and w). Test against fixed values rather than stats function call, especially when object is so tiny.

jangorecki:
This seems to fail on R 3.5 and older
https://gitlab.com/Rdatatable/data.table/-/jobs/1848594604

PR #5201, opened 2021-10-07
`melt`: Check the length of unlisted `measure.vars`

hsbadr:
Check whether the length of `measure.vars` does not change when unlisted (i.e., the length of each element of `measure.vars` equals `1L`). ~This PR also fixes reverse dependencies after https://github.com/Rdatatable/data.table/pull/5112.~

Example:
```
library(data.table)
DT.wide <- data.table(a2 = 2, b1 = 1, b2 = 2)
melt(DT.wide, measure.vars = list("b1", "b2"), value.name = "value")
```

Old behavior:
```
   a2 variable value1 value2
1:  2        1      1      2
```

New behavior:
```
   a2 variable value
1:  2       b1 1
2:  2       b2 2
```

PR #5247, opened 2021-11-02
melt(measure.vars=list) returns indices in variable column

### LINKED ISSUE #5209: melt with measure.vars=list of length=1 should return integer in variable column
melt() docs say that variable column should be an integer when measure.vars is a list,
```
variable.name: name (default ''variable'') of output column containing
          information about which input column(s) were melted. 
...
          If 'measure.vars' is a list of integer/character
          vectors, then each entry of this column contains an integer
          indicating an index/position in each of those vectors. 
```
In most cases that is true, for example 
```r
> library(data.table)
> (iris.row <- data.table(iris[1,]))
   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
1:          5.1         3.5          1.4         0.2  setosa
> melt(iris.row, measure.vars = list(Length=c("Petal.Length","Sepal.Length"), Width=c("Petal.Width","Sepal.Width")))
   Species variable Length Width
1:  setosa        1    1.4   0.2
2:  setosa        2    5.1   3.5
> melt(iris.row, measure.vars = list(Length="Petal.Length",Width="Petal.Width"))
   Sepal.Length Sepal.Width Species variable Length Width
1:          5.1         3.5  setosa        1    1.4   0.2
```
However if measure.vars is a list of length=1 then we get variable= character column name instead, 
```r
> melt(iris.row, measure.vars = list(Length="Petal.Length"))
   Sepal.Length Sepal.Width Petal.Width Species     variable Length
1:          5.1         3.5         0.2  setosa Petal.Length    1.4
```
This is not a big deal (probably not a lot of users which melt only one column), but it is inconsistent with the documentation, so I will work on a fix.

tdhock:
Closes #5209 

This was not a big problem, but it would be nice for consistency: whenever user specifies measure.vars=list, variable column should contains indices rather than column names. Increased consistency may help avoid issues such as https://github.com/Rdatatable/data.table/pull/5201 in which a user thought that `measure.vars=list(...)` and `measure.vars=c(...)` should yield same result (docs say otherwise).

Previous behavior: (master/CRAN)
```r
> melt(data.table(a=10, b=20), measure.vars=list("a"))
    b variable value
1: 20        a    10
```

New behavior:
```r
> melt(data.table(a=10, b=20), measure.vars=list("a"))
    b variable value
1: 20        1    10
```

Main fix involves checking if measure.vars is list in fmelt C code.
To make some old tests keep passing, I had to change 

- `patterns`: when one argument/pattern input, used to output a list with one element (integer vector of match indices), now returns just the integer vector (not in a list). When used like `melt(DT, measure.vars=patterns("^a")` we therefore still get the same result as before this PR (variable is column name, not index). 
- `.SDcols=patterns("^a")` again patterns returns vector instead of list so that needed to be special cased in `[.data.table` -- just use that vector instead of trying to pass it through `Reduce(intersect, ...)` -- again same results as before this PR.


mnazarov:
Another inconsistency with length-1 list in `measure.vars`, is that names are discarded, i.e.
```
# CRAN version
> melt(data.table(a=10, b=20), measure.vars=list("n" = "a"))
    b variable value
1: 20        a    10
```
And this PR seems to fix that too (even if it seems that the names *were* returned when PR was created in 2021):
```
# with this PR
> melt(data.table(a=10, b=20), measure.vars=list("n" = "a"))
    b variable  n
1: 20        1 10
```
Maybe adding an extra test with names could be useful.

tdhock:
hi @jangorecki @MichaelChirico I would like to include this fix in next release, and I believe that it is ready to merge into master, so could you please review and/or merge if you agree that it is ready? Thanks!

NEWS.md
```diff
@@ -36,6 +36,8 @@
 
 5. `fwrite(x, row.names=TRUE)` with `x` a `matrix` writes `row.names` when present, not row numbers, [#5315](https://github.com/Rdatatable/data.table/issues/5315). Thanks to @Liripo for the report, and @ben-schwen for the fix.
 
+6. `melt` with a list for `measure.vars` was inconsistent between length=1 list and length>1 list, [#5209](https://github.com/Rdatatable/data.table/issues/5209). Thanks to @tdhock for reporting and fixing.
```
MichaelChirico, line 39: Note: We usually write NEWS items in terms of the new behavior, not in terms of what used to be a problem, e.g. here "melt... is consistent for any length of `list()` input to `measure.vars`". It might also be prudent to describe what is now consistent, i.e. "melt returns an `integer` column for `variable` whenever `measure.vars` is a `list()`, consistent with the documented behavior".

R/fmelt.R
```diff
@@ -30,6 +30,7 @@ patterns = function(..., cols=character(0L)) {
   # replace with lengths when R 3.2.0 dependency arrives
   if (length(idx <- which(sapply(matched, length) == 0L)))
     stopf('Pattern(s) not found: [%s]', brackify(p[idx]))
+  if(length(matched)==1)return(matched[[1]])
```
MichaelChirico, line 33:
```suggestion
  if (length(matched) == 1L) return(matched[[1L]])
```

src/fmelt.c
```diff
@@ -293,7 +293,8 @@ struct processData {
     totlen,       // of output/long DT result of melt operation.
     nrow;         // of input/wide DT to be melted.
   SEXPTYPE *maxtype;
-  Rboolean narm;  // remove missing values?
+  Rboolean measure_is_list,
+    narm;  // remove missing values?
```
MichaelChirico, line 297:
```suggestion
  Rboolean measure_is_list, narm;  // remove missing values?
```

inst/tests/tests.Rraw
```diff
@@ -17275,7 +17275,13 @@ exid = data.table(id=1, expected)
 test(2182.3, melt(DTid, measure.vars=list(a=c(NA,1), b=2:3), id.vars="id"), exid)
 test(2182.4, melt(DTid, measure.vars=list(a=c(NA,"a2"), b=c("b1","b2")), id.vars="id"), exid)
 test(2182.5, melt(DT.wide, measure.vars=list(a=c(NA,1), b=2:3), na.rm=TRUE), data.table(variable=factor(2), a=2, b=2))
-test(2182.6, melt(DT.wide, measure.vars=list(b=c("b1","b2"))), data.table(a2=2, variable=factor(c("b1","b2")), b=c(1,2))) # measure.vars named list length=1, #5065
+test(2182.6, melt(DT.wide, measure.vars=list(b=c("b1","b2"))), data.table(a2=2, variable=factor(c("1","2")), b=c(1,2))) # measure.vars named list length=1, #5065
```
MichaelChirico, line 17278: why change the test?

src/fmelt.c
MichaelChirico: Oh, this is a `struct`, not declaring variables within a function body. ignore.

src/fmelt.c
```diff
@@ -302,6 +303,11 @@ static void preprocess(SEXP DT, SEXP id, SEXP measure, SEXP varnames, SEXP valna
   SEXPTYPE type;
   data->lmax = 0; data->totlen = 0; data->nrow = length(VECTOR_ELT(DT, 0));
   SET_VECTOR_ELT(data->RCHK, 0, vars = checkVars(DT, id, measure, verbose));
+  if(!isNull(measure) && isNewList(measure)){
+    data->measure_is_list = TRUE;
+  }else{
+    data->measure_is_list = FALSE;
+  }
```
MichaelChirico, line 310:
```suggestion
  data->measure_is_list = !isNull(measure) && isNewList(measure) ? TRUE : FALSE;
```

Alternatively, this could be a simple `bool` in the struct and we just set `data->measure_is_list = !isNull(measure) && isNewList(measure)`.

MichaelChirico:
Minor feedback only. It does look like it could be a breaking change, though? We can merge and see what revdep testing tells us.

NEWS.md
tdhock: sounds great, thanks for the suggestion.

inst/tests/tests.Rraw
tdhock: this is the new result using this PR
```r
> melt(DT.wide, measure.vars=list(b=c("b1","b2")))
      a2 variable     b
   <num>   <fctr> <num>
1:     2        1     1
2:     2        2     2
> melt(DT.wide, measure.vars=c("b1","b2"))
      a2 variable value
   <num>   <fctr> <num>
1:     2       b1     1
2:     2       b2     2
```

tdhock:
yes it could be a breaking change. I can follow up and suggest fixes for any revdeps that show up.

src/fmelt.c
```diff
@@ -302,6 +303,7 @@ static void preprocess(SEXP DT, SEXP id, SEXP measure, SEXP varnames, SEXP valna
   SEXPTYPE type;
   data->lmax = 0; data->totlen = 0; data->nrow = length(VECTOR_ELT(DT, 0));
   SET_VECTOR_ELT(data->RCHK, 0, vars = checkVars(DT, id, measure, verbose));
+  data->measure_is_list = !isNull(measure) && isNewList(measure);
```
MichaelChirico, line 306: huh, somewhat surprised we need the `!isNull()` test in addition to `isNewList()`, but it's right there in the definition of `isNewList()`:

https://github.com/r-devel/r-svn/blob/898b57332801c2ee3b26a3e41fea40aabe6a3fa2/src/include/Rinlinedfuns.h#L829-L832

```suggestion
  data->measure_is_list = !isNull(measure) && isNewList(measure); // NB: NULL passes isNewList() hence !isNull() too
```

MichaelChirico:
> yes it could be a breaking change. I can follow up and suggest fixes for any revdeps that show up.

Here are some places to look beyond CRAN packages:

https://github.com/search?q=lang%3AR+%2Fmeasure%5B.%5Dvars%5Cs*%3D%5Cs*list%5B%28%5D%5B%5E%2C%29%5D%2B%5B%29%5D%2F&type=code

MichaelChirico (approved):
Feel free to merge after addressing latest round of review. Thanks!

jangorecki:
I feel we are much less strict about backward compatibility than we used to be. Checking revdeps is not sufficient IMO, it is not sufficient by a huge factor. First, most of pkgs on CRAN does not have good test coverage. Second, we don't know how many revdeps there are which are not o CRAN or github, I would assume at least 3-4 times more than CRAN. Third, there are thousands of plain scripts. Current process of looking at cran/bioc revdeps, and checking github for the usage is as much as we can do, but IMO not enough to jump in with breaking changes. Let's try to give at least 1 release cycle for transition. We could have extra section in NEWS file "breaking changes in next release" which described how to switch to new behavior.

src/fmelt.c
tdhock: right

NEWS.md
```diff
@@ -36,6 +36,8 @@
 
 5. `fwrite(x, row.names=TRUE)` with `x` a `matrix` writes `row.names` when present, not row numbers, [#5315](https://github.com/Rdatatable/data.table/issues/5315). Thanks to @Liripo for the report, and @ben-schwen for the fix.
 
+6. `melt` returns an integer column for `variable` whenever `measure.vars` is a list, consistent with the documented behavior, [#5209](https://github.com/Rdatatable/data.table/issues/5209). Thanks to @tdhock for reporting and fixing.
```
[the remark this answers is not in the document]
tdhock: ok I mentioned length=1

tdhock:
> We could have extra section in NEWS file "breaking changes in next release" which described how to switch to new behavior.

great idea Jan, I added a sentence about that:

7. `melt` returns an integer column for `variable` when `measure.vars` is a list of length=1, consistent with the documented behavior, [#5209](https://github.com/Rdatatable/data.table/issues/5209). Thanks to @tdhock for reporting and fixing. Any users who were relying on this behavior can change `measure.vars=list("col_name")` (output `variable` was column name, now is column index/integer) to `measure.vars="col_name"` (`variable` still is column name).

PR #5249, opened 2021-11-04
Fwrite rounding

### LINKED ISSUE #5238: Incorrect date + time writing via fwrite (.: 00Z) ISO
Hey!

```r
data.table::fwrite(
  x    = data.table::as.data.table(
          anytime::anytime(1678296152.99999952316284179688)), 
  file = here::here('test.csv'))
```

If we transfer the following code with a digit to a normal date + time + milliseconds, and then write it to a file, then we get an incorrect record: `2023-03-08T17:22:32.:00Z`

Why does fwrite round like this? The problem is that I then need to transfer this record format to another system, and its developers are sure that such a record does not meet any standards (like me) and do not want to change their parsers for this.

On my side, I, of course, did rounding, but I would like to get the correct record right away, not `.:00Z`

tlapak:
Closes  #5238

I have also cleaned up the `digits.secs` option that was set in earlier fwrite tests.

Note that base R does not correctly handle printing of fractional seconds for negative timestamps (on Windows at least). (The issue is somewhere in as.POSIXlt.POSIXct)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -451,7 +451,7 @@
     
 47. `tables()` failed with `argument "..." is missing` when called from within a function taking `...`; e.g. `function(...) { tables() }`, [#5197](https://github.com/Rdatatable/data.table/issues/5197). Thanks @greg-minshall for the report and @michaelchirico for the fix.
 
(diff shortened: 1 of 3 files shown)


avimallu:
In the news section, I think you meant `fwrite` and not `fread`.

tlapak:
> In the news section, I think you meant `fwrite` and not `fread`.

Ah, thanks. Fixed.

PR #4694, opened 2020-09-05
fread nrows=0L fixed to work like nrows=0

ben-schwen:
Closes #4686. `fread` now also accepts `nrows=0L` and treats it like `nrows=0`. Beforehand `0L` was treated like `nrows=Inf`.

Issue #1267, opened 2015-08-13
[fread] broken functionality in 1.9.6


ziyadsaeed:
fread from v1.9.4 gives correct results but fread from dev 1.9.5 doesn't.
For eg

create a test.csv file 

``` R
cat("V1, V2, V3
1,2,3
V4, V5, V6, V7
4,5,6,7
8,9,10,11",
    file = "test.csv"
)
```

Notice there are different number of columns
fread from 1.9.4

``` R
fread("test.csv", nrows = 1, header = TRUE, skip = 0)
V1 V2 V3
1:  1  2  3
```

This is what I want

But in fread from v1.9.5

``` R
     fread("test.csv", nrows = 1, header = TRUE, skip = 0)
    V4 V5 V6 V7
    1:  4  5  6  7
    Warning messages:
    1: In fread("test.csv", nrows = 1, header = TRUE, skip = 0) :
     Starting data input on line 3 and discarded previous non-empty line: 1,2,3
    2: In fread("test.csv", nrows = 1, header = TRUE, skip = 0) :
    Stopped reading at empty line 5 but text exists afterwards (discarded): 8,9,10,11
```

Not the expected result.

ziyadsaeed:
Minimum reproducible code is now added.

ziyadsaeed:
This bug has now been released in v1.9.6.

jangorecki:
@ziyadsaeed you can self-close resolved issue

ziyadsaeed:
it isn't resolved. I'm mentioning that this bug was not fixed in the dev version and now the release version 1.9.6 has this bug.

arunsrinivasan:
The default behaviour is right as it is. I'd expect `skip=0` explicitly to work as you intend however.

ziyadsaeed:
what is the workaround to ask fread to just read the first line

everdark:
in 1.9.6 (and 1.9.7 to this point of time) `nrows` no longer forces `fread` to just read those rows.
this makes me frustrated on small files with only the last line having less columns.

`fread("1,2,3\n1,2", nrows=1)` wont work and there seems to be no work-around...
i'd like to achieve reading the entire file except for the unbalanced last line.

and this only happens in small files:

```r
for ( i in 100:1) { 
  lines <- paste0(paste(rep("1,2,3", i), collapse='\n'), "\n1,2")
  fread(lines, nrows=i)
}
```

[...]

PR #5253, opened 2021-11-06
fread(file, nrows=0) file with header does not determine types

### LINKED ISSUE #4029: when nrows = 0 in fread, field types returned are not correct
Hi All,
First off, thanks for your work on an amazing package.

In the documentation for fread, under the nrows argument; the following suggests to me that it is the intention for fread to return a 0 row data.table with correct classes. 
> 'nrows=0' returns the column names and typed empty columns determined by the large sample; useful for a dry run of a large file or to quickly check format consistency of a set of files before starting to read any of them.

However in the reprex below, all columns are returned as logicals.

```
library(data.table)
tmp <- tempfile(fileext = ".csv")    
fwrite(iris, tmp)    
str(fread(tmp, nrows = 0))    
#> Classes 'data.table' and 'data.frame':   0 obs. of  5 variables:
#>  $ Sepal.Length: logi 
#>  $ Sepal.Width : logi 
#>  $ Petal.Length: logi 
#>  $ Petal.Width : logi 
#>  $ Species     : logi 
```

Whereas, I was expecting the same output as the following:

```
str(fread(tmp)[0])
#> Classes ‘data.table’ and 'data.frame':	0 obs. of  5 variables:
#>  $ Sepal.Length: num 
#>  $ Sepal.Width : num 
#>  $ Petal.Length: num 
#>  $ Petal.Width : num 
#>  $ Species     : chr 
```

```
sessionInfo()
#> R version 3.6.1 (2019-07-05)
#> Running under: Ubuntu 18.04.3 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
#> LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1
#> 
#> locale:
#>  [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
#>  [4] LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
#>  [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
#> [10] LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.12.6
#> 
#> [1] compiler_3.6.1 tools_3.6.1   
```
Thanks!

ben-schwen:
Closes #4029

Follow up to #4694

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -1579,7 +1579,7 @@ int freadMain(freadMainArgs _args) {
   int ncol;  // Detected number of columns in the file
   const char *firstJumpEnd=NULL; // remember where the winning jumpline from jump 0 ends, to know its size excluding header
   const char *prevStart = NULL;  // the start of the non-empty line before the first not-ignored row (for warning message later, or taking as column names)
-  int jumpLines = (int)umin(100,nrowLimit);   // how many lines from each jump point to use. If nrowLimit is supplied, nJumps is later set to 1 as well.
+  int jumpLines = (int)umin(100, umax(1,nrowLimit));   // how many lines from each jump point to use. If nrowLimit is supplied, nJumps is later set to 1 as well. // sample at least 1 line for files with header and nrows==0L #4029
   {
   if (verbose) DTPRINT(_("[06] Detect separator, quoting rule, and ncolumns\n"));
 
--- a/src/fread.c
+++ b/src/fread.c
@@ -1579,7 +1579,7 @@ int freadMain(freadMainArgs _args) {
   int ncol;  // Detected number of columns in the file
   const char *firstJumpEnd=NULL; // remember where the winning jumpline from jump 0 ends, to know its size excluding header
   const char *prevStart = NULL;  // the start of the non-empty line before the first not-ignored row (for warning message later, or taking as column names)
-  int jumpLines = (int)umin(100, umax(1,nrowLimit));   // how many lines from each jump point to use. If nrowLimit is supplied, nJumps is later set to 1 as well. // sample at least 1 line for files with header and nrows==0L #4029
+  int jumpLines = nrowLimit==0 ? 100 : (int)umin(100, nrowLimit);   // how many lines from each jump point to use. If nrows>0 is supplied, nJumps is later set to 1. nrows=0 #4029
   {
   if (verbose) DTPRINT(_("[06] Detect separator, quoting rule, and ncolumns\n"));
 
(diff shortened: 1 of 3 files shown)


src/fread.c
```diff
@@ -1579,7 +1579,7 @@ int freadMain(freadMainArgs _args) {
   int ncol;  // Detected number of columns in the file
   const char *firstJumpEnd=NULL; // remember where the winning jumpline from jump 0 ends, to know its size excluding header
   const char *prevStart = NULL;  // the start of the non-empty line before the first not-ignored row (for warning message later, or taking as column names)
-  int jumpLines = (int)umin(100,nrowLimit);   // how many lines from each jump point to use. If nrowLimit is supplied, nJumps is later set to 1 as well.
+  int jumpLines = (int)umin(100, umax(1,nrowLimit));   // how many lines from each jump point to use. If nrowLimit is supplied, nJumps is later set to 1 as well. // sample at least 1 line for files with header and nrows==0L #4029
```
MichaelChirico, line 1582: Would it make sense to remove `nrowLimit` from here entirely? I have to imagine the first (or any random) row will give the incorrect column type a decent % of the time.

Scanning 100 rows to guess column types _should_ be pretty fast (it makes sense to test)...

ben-schwen: This was the default behavior pre #2623. It seems that it was user requested to only read `nrows` for sampling if `nrows` argument is supplied #1267. Bumping from 0 to 1 seems pretty safe to me (because I suspect files to have at least 1-2 "good" lines) while bumping from 0 to 100 might introduce unwanted behavior for those users?

Changing to `int jumpLines = 100` already breaks example from #1267
```r
fread("1,2,3\n1,2", nrows=1)
# > fread("1,2,3\n1,2", nrows=1)
#        1     2      3
#    <int> <int> <lgcl>
# 1:     1     2     NA
# Warning message:
# In fread("1,2,3\n1,2", nrows = 1) :
#   Detected 3 column names but the data has 2 columns. Filling rows automatically. Set fill=TRUE explicitly to avoid this warning.
fread("1,2,3\n1,2", nrows=1, header=FALSE)
# > fread("1,2,3\n1,2", nrows=1, header=FALSE)
#       V1    V2
#    <int> <int>
# 1:     1     2
```

While current fix gives
```r
fread("1,2,3\n1,2", nrows=1)
# > fread("1,2,3\n1,2", nrows=1)
#       V1    V2    V3
#    <int> <int> <int>
# 1:     1     2     3
```

MichaelChirico: I see, point taken. It really seems like a clash of purpose then -- `fread(., nrows=0)` is essentially asking for a different behavior vs. `fread(., nrows=1..100)`.

Maybe it makes sense to split that into a different argument? Or otherwise treat `n=0` differently than `n=1..100` -- we check up to 100 rows for `n=0` (because the user is trying to get a quick view of their data) but only `n` rows for `nrows=n`?

Seems somewhat confusing. Happy to approve this PR as is though.

cc @mattdowle for vis.

MichaelChirico (approved):

src/fread.c
ben-schwen: While a different argument could have made sense for this as e.g. `peak=TRUE`, I guess it would be pretty annoying for users who rely on this feature to change it now. 

Treating `n=0` differently, e.g. sampling 100 rows for `nrows=0` and otherwise take `min(100, nrows)`, seems like a good option to me and I'm happy to alter the PR in this direction.

PR #5256, opened 2021-11-06
set operations for DT containing x and y as column names

### LINKED ISSUE #5255: Set operations naming conflicts for x and y
Please have a look at the question and answer raised on [Stack overflow](https://stackoverflow.com/questions/69863758/unexpected-outcome-of-data-table-fintersect-all-true/69864348#69864348)

> It's related to naming conflicts. The function uses x and y as the names of the data.tables. You also happen to have columns 
> named x and y. So in the function, x[, ".seqn" := rowid(x)], in your example, data.table is using rowid(x) on the vector as
> opposed to the whole data.table. If you change your vectors to x1 and y1 (or change your data table names to DT1 as in your 
> working example), it works. A potential solution for the maintainers is to change the names of x and y to .x and .y or 
> something to reduce likelihood. –  Cole  1 hour ago 

Would it be possible to change the internal `x` and `y` to `.x` and `.y` as suggested, by Cole? The variables `x` and `y` are used very often like in the data.table [examples on set operations](https://rdatatable.gitlab.io/data.table/reference/setops.html)

ben-schwen:
Closes #5255

Problem is that `shallow(x)[, j]` gives precedence to columns names over object names and so `rowidv(x)` uses column and not the data.table `x`. Same argument holds for `y`.

https://github.com/Rdatatable/data.table/blob/e88826e94127991754a8821ac8149d15f1f9c3c2/R/setops.R#L58-L63

Fix avoids naming conflict by using the new programming environment. Could achieve the same by replacing `:=` with `set`, as below

```R
x = set(shallow(x), j=".seqn", value=rowidv(x))
y = set(shallow(y), j=".seqn", value=rowidv(y))
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -448,8 +448,10 @@
     # 2:          2021-02-03  # was 18661
     # 3: 4611686018427387906  # was error 'please use as.character'
     ```
-    
+
 47. `tables()` failed with `argument "..." is missing` when called from within a function taking `...`; e.g. `function(...) { tables() }`, [#5197](https://github.com/Rdatatable/data.table/issues/5197). Thanks @greg-minshall for the report and @michaelchirico for the fix.
+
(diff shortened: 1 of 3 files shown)


MichaelChirico (approved):
Thanks!

Issue #678, opened 2014-06-08
[R-Forge #205] Move TO DO comments in the code into tracker (use grep)


arunsrinivasan:
Submitted by: Matt Dowle; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=205&group_id=240&atid=978)

The marker "TO DO" is consistently used in the .R and .c comments.

---

As of this writing (April 15, 2019), here's the full list from `grep`:
- [ ] src/rbindlist.c:                // TO DO: options(datatable.pedantic=TRUE) to issue this warning :
- [ ] src/fastmean.c:              if(ISNAN(REAL(x)[i])) continue;  // TO DO: could drop this line and let NA propogate?
- [ ] src/uniqlist.c:                // TO DO: store previ twiddle call, but it'll need to be vector since this is in a loop through columns. Hopefully the first == will short circuit most often
- [ ] src/uniqlist.c:                // TO DO: store previ twiddle call, but it'll need to be vector since this is in a loop through columns. Hopefully the first == will short circuit most often
- [ ] src/forder.c:    // TO DO: at this point if the last count==n then it's all the same number and we can stop now.
- [ ] src/forder.c:static void *xtmp=NULL; int xtmp_alloc=0;                                   // TO DO: save xtmp if possible, see allocs in forder
- [ ] src/forder.c:static void alloc_xtmp(int n) {                                             // TO DO: currently always the largest type (double) but 
- [ ] src/forder.c:        // TO DO: could include extra bits to divide the first radix up more. Often the MSD has groups in just 0-4 out of 256.
- [ ] src/forder.c:    alloc_otmp(maxgrpn);   // TO DO: can we leave this to forder and remove these calls??
- [ ] src/forder.c:    alloc_xtmp(maxgrpn);   // TO DO: doesn't need to be sizeof(double) always, see inside
- [ ] src/forder.c:    /\* TO DO:  If nextradix==-1 AND no further columns from forder AND !retGrp, we're 
- [ ] src/forder.c:    // see comments in iradix for structure.  This follows the same. TO DO: merge iradix in here (almost ready)
- [ ] src/forder.c:    if (radix_xsuballoc < maxgrpn) {                                            // TO DO: centralize this alloc
- [ ] src/forder.c:        // TO DO: could include extra bits to divide the first radix up more. Often the MSD has groups in just 0-4 out of 256.
[...]

Issue #24, opened 2014-06-08
[R-Forge #5672] error when merging zero row data.table


arunsrinivasan:
Submitted by: Garrett See; Assigned to: Arun ; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5672&group_id=240&atid=975)

The following did not give an error in 1.9.2, but does in 1.9.3.  Using svn revision 1263:

``` S
library(data.table)
# data.table 1.9.3  For help type: help("data.table")
a <- data.table(BOD, key="Time")
b <- data.table(BOD, key="Time")[Time < 0] # zero row data.table
merge(a,b, all=TRUE) # works fine
#    Time demand.x demand.y
#1:    1      8.3       NA
#2:    2     10.3       NA
#3:    3     19.0       NA
#4:    4     16.0       NA
#5:    5     15.6       NA
#6:    7     19.8       NA
merge(b,a, all=TRUE) # error
# Error in setcolorder(dt, c(setdiff(names(dt), end), end)) :
#   neworder is length 2 but x has 3 columns.
```

Originally reported on [the mailing list](http://thread.gmane.org/gmane.comp.lang.r.datatable/2531)

PR #5263, opened 2021-11-19
rbindlist support fill=TRUE with use.names=FALSE and use it in merge.R ToDo of #678

### LINKED ISSUE #5262: R/merge.R:                # TO DO: use set() here instead..
Towards #678

Use `set` in `all.y` branch instead of `cbind`

ben-schwen:
Closes #5262

### merge.R
- [x] use `set` in merge.R instead of `cbind`

### DIFF
--- a/R/merge.R
+++ b/R/merge.R
@@ -82,8 +82,9 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
       othercolsx = setdiff(nm_x, by)
       if (length(othercolsx)) {
         tmp = rep.int(NA_integer_, length(missingyidx))
-        # TO DO: use set() here instead..
-        yy = cbind(yy, x[tmp, othercolsx, with = FALSE])
+        nx = c(names(yy), names(x[tmp, othercolsx, with = FALSE]))
+        nx = make.unique(nx)
+        set(yy, NULL, tail(nx, -ncol(yy)), x[tmp, othercolsx, with = FALSE])
       }
       # empty data.tables (nrow =0, ncol>0) doesn't skip names anymore in new rbindlist
       # takes care of #24 without having to save names. This is how it should be, IMHO.
--- a/R/merge.R
+++ b/R/merge.R
@@ -81,8 +81,7 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
       yy = y[missingyidx]
       othercolsx = setdiff(nm_x, by)
       if (length(othercolsx)) {
-        nx = c(names(yy), paste0("V",seq_len(length(othercolsx))))
-        nx = make.unique(nx)
+        nx = make.unique(c(names(yy), othercolsx))
         set(yy, NULL, tail(nx, -ncol(yy)), rep(list(NA), length(othercolsx)))
       }
       # empty data.tables (nrow =0, ncol>0) doesn't skip names anymore in new rbindlist
--- a/R/merge.R
+++ b/R/merge.R
@@ -78,15 +78,7 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
     # Perhaps not very commonly used, so not a huge deal that the join is redone here.
     missingyidx = y[!x, which=TRUE, on=by, allow.cartesian=allow.cartesian]
     if (length(missingyidx)) {
-      yy = y[missingyidx]
-      othercolsx = setdiff(nm_x, by)
-      if (length(othercolsx)) {
-        nx = make.unique(c(names(yy), othercolsx))
-        set(yy, NULL, tail(nx, length(othercolsx)), rep(list(NA), length(othercolsx)))
-      }
-      # empty data.tables (nrow =0, ncol>0) doesn't skip names anymore in new rbindlist
-      # takes care of #24 without having to save names. This is how it should be, IMHO.
-      dt = rbind(dt, yy, use.names=FALSE)
+      dt = rbind(dt, y[missingyidx], use.names=FALSE, fill=TRUE)
     }
   }
   # X[Y] syntax puts JIS i columns at the end, merge likes them alongside i.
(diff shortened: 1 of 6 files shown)


R/merge.R
```diff
@@ -82,8 +82,9 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
       othercolsx = setdiff(nm_x, by)
       if (length(othercolsx)) {
         tmp = rep.int(NA_integer_, length(missingyidx))
-        # TO DO: use set() here instead..
-        yy = cbind(yy, x[tmp, othercolsx, with = FALSE])
+        nx = c(names(yy), names(x[tmp, othercolsx, with = FALSE]))
+        nx = make.unique(nx)
+        set(yy, NULL, tail(nx, -ncol(yy)), x[tmp, othercolsx, with = FALSE])
```
MichaelChirico, line 87: is there a noticeable memory advantage to looping over the RHS columns instead of creating the whole with=FALSE subset table?

it looks like the advantage is we can do the tmp subset once & apply it to all columns...

MichaelChirico (approved):

R/merge.R
ben-schwen: I guessed that saving this tmp object and doing indexing twice is preferred over using additional memory. Rethinking it again, I might have figured out what the ToDo meant initially where using set omits this object.

ben-schwen:
@MichaelChirico whats the preferred process here: 1 PR per source file? Or 1 PR per todo? Or 1 megathread PR?

MichaelChirico:
I think generally one PR per TODO makes the most sense, but if you see a cluster of TODOs in closely related code, do combine them. basically TODOs that are logically independent should get their own PRs

R/merge.R
```diff
@@ -81,9 +81,9 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
       yy = y[missingyidx]
       othercolsx = setdiff(nm_x, by)
       if (length(othercolsx)) {
-        tmp = rep.int(NA_integer_, length(missingyidx))
-        # TO DO: use set() here instead..
-        yy = cbind(yy, x[tmp, othercolsx, with = FALSE])
+        nx = c(names(yy), paste0("V",seq_len(length(othercolsx))))
+        nx = make.unique(nx)
+        set(yy, NULL, tail(nx, -ncol(yy)), rep(list(NA), length(othercolsx)))
```
MichaelChirico, line 86: does this need to be NA_integer_?

ben-schwen: Nope, `NA` logical works fine here. `rbindlist` coerces types so `logical` columns of `yy`will be coerced to the types of `dt`. The reason behind the previous `tmp` used `NA_integer_` is that it was used for indexing the data.table and therefore was `integer`.

R/merge.R
```diff
@@ -81,9 +81,8 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
       yy = y[missingyidx]
       othercolsx = setdiff(nm_x, by)
       if (length(othercolsx)) {
-        tmp = rep.int(NA_integer_, length(missingyidx))
-        # TO DO: use set() here instead..
-        yy = cbind(yy, x[tmp, othercolsx, with = FALSE])
+        nx = make.unique(c(names(yy), othercolsx))
+        set(yy, NULL, tail(nx, length(othercolsx)), rep(list(NA), length(othercolsx)))
```
jangorecki, line 85: What if missingindex is 1:nrow(y) and potentially copy is not happening (or may not happening in future due to some regression). Would be good to have a test that ensure that. Otherwise we could ended up updating y table.

jangorecki (commented):
Btw. I recall cbindlist in a PR has option for controlling copy behavior.

jangorecki:
Uh, sorry for late feedback but I realized nrow(y)=1 can be useful to test against as well. I can imagine that there could be different handling of a copy when you subset all rows and the index is scalar at the same time. Kind of edge case but worth to have it as well.

ben-schwen:
> Uh, sorry for late feedback but I realized nrow(y)=1 can be useful to test against as well. I can imagine that there could be different handling of a copy when you subset all rows and the index is scalar at the same time (based on DT[TRUE] doing shallow copy). Kind of edge case but worth to have it as well.

Good point. I added test cases to check for accidental shallow copying.

AFAIU current behavior of shallow copying of #3215 (and friends) leads only to problems when pre-copy cols are altered in the copy since these cols point to the same address? Since we only add cols here and don't alter other cols in `yy`, I guessed there shouldn't be a problem here.

inst/tests/tests.Rraw
```diff
@@ -1875,6 +1877,16 @@ test(630.1, merge(DT1,DT2,all.x=TRUE), setkey(adt(merge(adf(DT1),adf(DT2),by="a"
 
 test(631, merge(DT1,DT2,all.y=TRUE), data.table(a=c(2,3,5),total.x=c(NA,1,1),total.y=c(5,1,2),key="a"))
 test(631.1, merge(DT1,DT2,all.y=TRUE), setkey(adt(merge(adf(DT1),adf(DT2),by="a",all.y=TRUE)),a))
+# ensure merge(x,y,all.y) does not alter input y
+# merge containing idx 1:nrow(y)
```
MichaelChirico, line 1881: this comment is a little unclear on its own -- ideally we can read the comment without any more context and know the point of the test. I think 'idx' refers to the implementation? best to be more explicit about what it means

MichaelChirico (approved):

mattdowle:
I tried to get to the root cause of this block of code. It took me a while to realize that all it was doing was adding on the right number of NA columns to `yy` so it could be `rbind`-ed using `use.names=FALSE`. So I made `rbind` support `use.names=FALSE` together with `fill=TRUE` and that block goes away.
I also removed the long standing comment about issue #24 after looking at it and thinking that comment is unlikely to be useful in the future now.
Left the new tests in place and unchanged.

PR #5270, opened 2021-11-21
5268 support non numeric types

### LINKED ISSUE #5268: as.xts.data.table is limited to numeric data types, but this is not an xts limitation
`xts` allows columns to have any type, as long as the columns are all the same type (as its really a matrix under the hood)

for example, you can have character columns:
`xts(x = cbind(letters, LETTERS), order.by = Sys.Date() + (1:26))`

logicals:
`xts(x = c(T,F), order.by = Sys.Date() + (1:2))`

`as.xts.data.table` currently only allows conversion for numeric columns. it should follow the same rules as `xts` and support any type

ethanbsmith:
add support for non numeric xts, while maintaining backward compatibility and an optimized common use case. fixes #5268


mattdowle:
Look good. Needs a test, man page update, and news item please.

PR #5275, opened 2021-11-29
print.class and print.keys now TRUE by default

mattdowle:
Thanks to a suggestion by @grantmcdermott in https://github.com/Rdatatable/data.table/pull/5224#issuecomment-981117084.

### DIFF
--- a/.dev/cc.R
+++ b/.dev/cc.R
@@ -22,8 +22,6 @@
 # c
 # test and step between R and C
 
-options(datatable.print.class = TRUE)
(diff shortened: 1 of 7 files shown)


grantmcdermott:
Thanks Matt! 

(Background: I've been sharing this tip with my students when I teach my [data.table lecture](https://raw.githack.com/uo-ec607/lectures/master/05-datatable/05-datatable.html) and get asked why it isn't the default.)

PR #5276, opened 2021-12-01
5268_as_xts_support_non_numerics

### LINKED ISSUE #5268: as.xts.data.table is limited to numeric data types, but this is not an xts limitation

ethanbsmith:
allow bypass of numeric only columns
use `as.matrix` for fast conversion
fixes #5268 (replaces pr: https://github.com/Rdatatable/data.table/pull/5270)


R/xts.R
```diff
@@ -15,11 +15,14 @@ as.data.table.xts = function(x, keep.rownames = TRUE, key=NULL, ...) {
...
 
-as.xts.data.table = function(x, ...) {
+as.xts.data.table = function(x, numeric.only = TRUE, ...) {
   stopifnot(requireNamespace("xts"), !missing(x), is.data.table(x))
   if (!xts::is.timeBased(x[[1L]])) stopf("data.table must have a time based column in first position, use `setcolorder` function to change the order, or see ?timeBased for supported types")
-  colsNumeric = vapply_1b(x, is.numeric)[-1L] # exclude first col, xts index
-  if (!all(colsNumeric)) warningf("Following columns are not numeric and will be omitted: %s", brackify(names(colsNumeric)[!colsNumeric]))
-  r = setDF(x[, .SD, .SDcols = names(colsNumeric)[colsNumeric]])
-  return(xts::as.xts(r, order.by = if ("IDate" %chin% class(x[[1L]])) as.Date(x[[1L]]) else x[[1L]]))
+  r <- x[, -1L]# exclude first col, xts index
```
MichaelChirico, line 21: sorry for the late review. would it be better to `shallow()`-copy `x` and then delete the first column?

NEWS.md
```diff
@@ -520,6 +520,8 @@
 
 51. `merge.data.table()` silently ignored the `incomparables` argument, [#2587](https://github.com/Rdatatable/data.table/issues/2587). It is now implemented and any other ignored arguments (e.g. misspellings) are now warned about. Thanks to @GBsuperman for the report and @ben-schwen for the fix.
 
+52. `as.xts.data.table` now supports non-numeric xts coredata matrixes, [5268](https://github.com/Rdatatable/data.table/issues/5268). Existing numeric only functionality is supported by a new `numeric.only` parameter, which defaults to `TRUE` for backward compatability and the most common use case. To convert non-numeric columns, set this parameter to `FALSE`. Conversions of `data.table` columns to a `matrix` now uses `data.table::as.matrix`, with all its performance benefits. Thanks to @ethanbsmith for the report and fix.
```
MichaelChirico, line 523: Do we plan to support this `numeric.only` option indefinitely? If not, we should state the migration plan up front.

ethanbsmith:
it is an extremely common use case. the numeric filter was in place for years and no one ever mentioned it as a problem. i only ran into the issue when i needed to use a boolean matrix
________________________________

PR #5301, opened 2021-12-31
stop loosing colnames with data.frame of ITime

### LINKED ISSUE #4673: create data.frame with ITime will lose the colname
Looks like a bug to me.

``` r
data.frame(COL = data.table::as.ITime('12:00:00'))
#>         V1
#> 1 12:00:00
```

<sup>Created on 2020-08-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

ben-schwen:
Closes #4673


R/IDateTime.R
```diff
@@ -211,6 +211,12 @@ as.character.ITime = format.ITime = function(x, ...) {
 }
 
 as.data.frame.ITime = function(x, ...) {
+  nm = names(get("x", envir=parent.frame()))
```
MichaelChirico, line 214: This assumes `as.data.frame()` is being called from `data.frame()`/`data.table()`, right? But it breaks direct call:

```r
as.data.frame(as.ITime('12:00:00'))
# Error in get("x", envir = parent.frame()) : object 'x' not found
```

I wonder if the ggplot2 issue cited is even still relevant...

MichaelChirico:
Superseded by #7108

PR #5335, opened 2022-02-16
try setting NOT_CRAN=true during revdeps to activate more tests

MichaelChirico:
This will activate tests using `testthat::skip_on_cran()`. Inspired by finding this:

https://github.com/kaufman-lab/intervalaverage/blob/c672cc4dafff051f50eded8c558a7e9bcea1a08b/tests/testthat/test-performance.R#L1-L10

This setting _may_ cause more headaches than it's worth as it will turn on a bunch of tests that aren't necessarily regularly passing already. But maybe worth giving it a try on next release.

### DIFF
--- a/.dev/revdep.R
+++ b/.dev/revdep.R
@@ -331,7 +331,7 @@ run = function(pkgs=NULL, R_CHECK_FORCE_SUGGESTS=TRUE, choose=NULL) {
   }
   if (!identical(pkgs,"_ALL_")) for (i in pkgs) system(paste0("rm -rf ./",i,".Rcheck"))
   SUGG = paste0("_R_CHECK_FORCE_SUGGESTS_=",tolower(R_CHECK_FORCE_SUGGESTS))
-  cmd = paste0("ls -1 *.tar.gz ", filter, "| TZ='UTC' OMP_THREAD_LIMIT=2 ",SUGG," parallel --max-procs 50% --timeout 1200 ",R," CMD check")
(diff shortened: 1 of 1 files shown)


jangorecki:
What's the motivation?
Problem about this option is that tests may not be even properly executed because they may expect to have more RAM, third party software installed, or even a local data files.

MichaelChirico:
it seems like a good way to catch some more stressful tests for possible regressions.

I share your suspicion that it may bring more headaches than it's worth, but think we should try anyway. for the cases mentioned, ideally maintainers would add additional skips for system state tests instead of sweeping everything under the skip_on_cran() rug... ideally...

MichaelChirico:
the more concerning issue is probably that it becomes a lot harder to test packages with unrelated failures.

now we can run tests with dev version against CRAN downstream. there should be few spurious failures as packages are passing on CRAN (or will soon be removed/updated so failures are transient)

with this, we'd need to do two runs+ (1) CRAN data.table+NOT_CRAN=true (2) dev data.table+NOT_CRAN=true. then compare the difference.

that makes me think we'll not go through with this as my recent experience shows me the skip_on_cran() tests have a way of breaking on non-maintainer machines...

MichaelChirico:
With some more experience it looks like a _ton_ of NOT_CRAN tests are broken in general :\ particularly owing to the fragility of snapshot tests. See e.g. https://github.com/r-lib/testthat/issues/1576. Seems insurmountable to me at the moment.

Issue #3439, opened 2019-03-02
fread() not handling null fields correctly


sz-cgt:
Documentation for `fread()` claims that (see the `na.strings` parameter documentation):
> A character vector of strings which are to be interpreted as NA values. By default, ",," for columns of all types, including type 'character' is read as NA for consistency. ,"", is unambiguous and read as an empty string. To read ,NA, as NA, set na.strings="NA". To read ,, as blank string "", set na.strings=NULL. When they occur in the file, the strings in na.strings should not appear quoted since that is how the string literal ,"NA", is distinguished from ,NA,, for example, when na.strings="NA".

However, that is not the case.

``` r
library(data.table)
dt <-
  data.table(a = c(NA, letters[2:5]),
             c = c(letters[1:2], NA, letters[4:5]),
             e = c(letters[1:4], NA))
fwrite(dt, "dt.csv")
readLines("dt.csv")
#> [1] "a,c,e" ",a,a"  "b,b,b" "c,,c"  "d,d,d" "e,e,"

all.equal(dt, fread("dt.csv"))
#> [1] "Column 'a': 'is.NA' value mismatch: 0 in current 1 in target"

all.equal(dt, fread("dt.csv", na.strings = ""))
#> [1] TRUE
```

<sup>Created on 2019-03-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

As you can see the files have sequences of delimiters with no characters between them, but `fread()` is not returning them as `NA` values.  Explicitly setting `na.strings = ""`, produces the expected behaviour, but this too violates the documentation, which says this should produce the blank string behaviour instead (third line in the quote above).

```
> sessionInfo()
R version 3.5.2 (2018-12-20)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 16299)

Matrix products: default

locale:
[1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252   
[3] LC_MONETARY=English_Australia.1252 LC_NUMERIC=C                      
[5] LC_TIME=English_Australia.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.12.0

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.0      ps_1.3.0        packrat_0.5.0   digest_0.6.18   rprojroot_1.3-2
[...]

PR #5353, opened 2022-03-22
Correct fread() help to reflect default na.strings behaviour

ysaidani:
This PR adjusts the help text for the `na.strings` parameter in the `fread()` function. No actual code is changed, hence no tests are affected and no NEWS entry is added.
### Problem
As described in #3439 , the `fread()` help isn't consistent with the function's actual default behaviour:
- What the `fread()` help **says**: `By default, ",," for columns of all types, including type 'character' is read as NA for consistency.`
- What `fread()` actuallly **does**: `na.strings=getOption("datatable.na.strings","NA")`; the `getOption` statement returns its default value (the second argument), i.e. `"NA"`. Thus, a ",NA," sequence in a file is read as `<NA>`, while a ",," sequence is read as `""`, an empty string - contradicting the description of the default behaviour given in the `fread()` help.
### Solution
#4288 adjusts the default behaviour to `na.strings=getOption("datatable.na.strings","")`, but doesn't pass a number of checks. Thus, as a minimally invasive fix, this PR adjusts the documentation to correctly reflect the current default behaviour.


man/fread.Rd
```diff
@@ -36,7 +36,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC"
...
+  \item{na.strings}{ A character vector of strings which are to be interpreted as \code{NA} values. By default, \code{",,"} is read as \code{NA} for columns of all types, except those of type \code{character}: Here, the default is to read \code{,NA,} as \code{NA} (\code{na.strings="NA"}). To read \code{,,} as \code(NA), set \code{na.strings=""} (to be implemented as the default in a future release). To read \code{,,} as blank string \code{""}, set \code{na.strings=NULL}. \code{,"",} is unambiguous and always read as an empty string. When they occur in the file, the strings in \code{na.strings} should not appear quoted since that is how the string literal \code{,"NA",} is distinguished from \code{,NA,}, for example, when \code{na.strings="NA"}. }
```
MichaelChirico, line 39:
```suggestion
  \item{na.strings}{ A character vector of strings which are to be interpreted as \code{NA} values. By default, \code{",,"} is read as \code{NA} for columns of all types besides \code{character}. For \code{characteR} columns, the current default is to read \code{,NA,} as \code{NA} (\code{na.strings="NA"}). To read \code{,,} as \code{NA}, set \code{na.strings=""}. We plan to make this the default in a future release. To read \code{,,} as a blank string \code{""}, set \code{na.strings=NULL}. \code{,"",} is unambiguous and always read as an empty string. When they occur in the file, the strings in \code{na.strings} should not appear quoted since that is how the string literal \code{,"NA",} is distinguished from \code{,NA,}, for example, when \code{na.strings="NA"}. }
```

MichaelChirico:
I added some nuance to your explanation -- in particular I found the part about `na.strings=NULL` maybe a bit misleading (since `na.strings="NA"`, the default, will also read `,,` as blank `""`).

More broadly, I worry this explanation in the usage section is too long-winded -- might it not be better to illustrate all this with examples in the `\examples{}` section instead?

MichaelChirico:
Hi @ysaidani could you PTAL at the review and follow up if you get time? Thanks!

MichaelChirico:
I'll close this pending request from other users for similar clarity, otherwise @ysaidani please feel free to take this back up if it's still important to you. Thank you!

PR #5383, opened 2022-05-13
clarify that list input->unnamed list output

MichaelChirico:


man/transpose.Rd
```diff
@@ -26,6 +26,8 @@ transpose(l, fill=NA, ignore.empty=FALSE, keep.names=NULL, make.names=NULL)
 }
 \value{
   A transposed \code{list}, \code{data.frame} or \code{data.table}.
+
+  \code{list} outputs will be unnamed.
```
ben-schwen, line 30: Not sure we can really say this although I understand the intention.

``` r
transpose(list(names=c('x','y'), c(1,2), c(3,4)), make.names="names")
#> $x
#> [1] 1 3
#> 
#> $y
#> [1] 2 4
```

MichaelChirico: how about 

```suggestion
  \code{list} outputs will only be named according to \code{make.names}.
```

tdhock: would be great to add that as an example too if there is not one yet.

ben-schwen (approved):
LGTM

PR #5399, opened 2022-06-02
Remove optimisation for na.strings = "" from fread.c

### LINKED ISSUE #5377: CRAN Warning on r-devel-linux-x86_64-debian-gcc
I noticed that `data.table` has been warning on CRAN for r-devel-linux-x86_64-debian-gcc. That image uses GCC 12 and since GCC 12, `-Waddress` now warns in more cases when comparing a non-constant pointer to NULL (gcc has always optimized this out), see [here](https://gcc.gnu.org/gcc-12/changes.html). This picked up on a missing dereference in `fread` (line 1287):
https://github.com/Rdatatable/data.table/blob/e9a323de01a17af70d5316016606fa8d35b25023/src/fread.c#L1284-L1289
I've looked at it a bit and I think I get what Matt was trying to do there with that optimisation but I don't have a test case that would demonstrate any speed improvement from it. For small files, I can't see any difference. So unless someone has a good test case, I'd suggest just removing this line and the associated check in `end_NA_string`.

(The comment in there is also currently incorrect, since the default is still `'NA'` and not `'\0'`. Also note that atm, adding in the dereference can lead to a NULL pointer dereference a bit further down.)

tlapak:
Closes #5377

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -603,6 +603,7 @@
 
 16. The options `datatable.print.class` and `datatable.print.keys` are now `TRUE` by default. They have been available since v1.9.8 (Nov 2016) and v1.11.0 (May 2018) respectively.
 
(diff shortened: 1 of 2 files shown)


MichaelChirico:
I think it warrants a NEWS item. Do we have `-Waddress` set in any CI that could have caught this sooner? GitLab?

tlapak:
I'm not fighting a NEWS item but the code never did anything as the comparison was already folded to false. 

It's not about -Waddress so much as GCC 12.1 which was only released in May. I'm pretty sure R CMD check sets -Wall. If I read the discussion right, Clang also misses it.

tlapak:
I've added a news item

mattdowle:
Thanks @tlapak. I did this in #5476 a bit differently by retaining the optimization and putting it later outside the loop with better comments.

PR #5401, opened 2022-06-03
Expand altrep in assign

### LINKED ISSUE #5400: Aliasing issue with `:=` affecting a different column
The below triggers the bug for me (note the assignment to `col2` changing the value of `col1`!):

```
library(data.table)

coalesce = function(x, ...) {
  for (y in list(...)) {
    idx = is.na(x)
    x[idx] = if (length(y) != 1) y[idx] else y
  }
  x
}

dt = data.table(id=1:64, col1=0, col2=0)
print(dt[1, .(col1, col2)])
#    col1 col2
# 1:    0    0
dt[, col1 := coalesce(col2, 111)]
dt[, col2 := 999]
print(dt[1, .(col1, col2)])
#    col1 col2
# 1:  999  999
```

And my `sessionInfo()` output:
```
R version 4.0.5 (2021-03-31)

Matrix products: default

locale:
[1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252    LC_MONETARY=English_Australia.1252 LC_NUMERIC=C                      
[5] LC_TIME=English_Australia.1252    

attached base packages:

other attached packages:
[1] data.table_1.14.2

[1] compiler_4.0.5 tools_4.0.5   
```

Basically it looks like col1 and col2 end up pointing at the same vector such that `:=` modifies them both; I'm guessing they are shared but the reference counts are off such that `:=` thinks it is safe to modify in-place. Not 100% clear to me if the actual underlying bug may be base R or data.table.

When trying to put together a minimal repro, I noticed a few different changes that make this bug disappear:

* Simply printing the data table between the col1 and col2 assignments makes the issue go away.

* It only manifests where the number of rows is at least 64. Perhaps that is used as a threshold at which some sort of copy-on-write optimization logic is kicking in somewhere?

* Also the problem seems to be related to the coalesce function used here, despite it not having any effect in this example. Eg replacing it with `coalesce = function(x, ...) x` avoids any issue. It seems as though base r is doing something weird with `[<-` with an all false logical subset; maybe the result is the same object but no longer marked as shared? Note that assigning to col1 after coalesce does not affect col2, only vice-versa. Alternatively returning `x[]` in coalesce bypasses the erroneous sharing by forcing a copy or bumping the ref count.

tlapak:
Closes #5400 

Note that I modified test 754.5 (now 754.05) because the character conversion (now tested in 754.07 and 754.08) gets deferred which results in the same behaviour as in the report, while `as.numeric` doesn't get deferred.

Tests 754.05, 754.07 and 754.09 depend on ALTREP behaviour of R and could break if that changes without `data.table` being broken, e.g. if `as.numeric` gets deferred at some point in the future, the test will break but everything will still work correctly.


inst/tests/tests.Rraw
```diff
@@ -2270,15 +2270,36 @@ test(753.1, DT[,c("x1","x2"):=4:6, verbose = TRUE], data.table(a=letters[1:3],x=
...
+# Expand ALTREPS in assign.c, #5400
+# String conversion gets deferred
+test(754.07, DT[,a:=as.character(a),verbose=TRUE], output="RHS for item 1 has been duplicated")
+# Executing the code inside of test expands the ALTREP so we repeat the code
+# in order to check the result after a further assignment
+DT = data.table(a=1:3)
+DT[,b:=as.character(a)]
+DT[, a:=5L]
+test(754.08, DT, data.table(a=5L, b=as.character(1:3)))
+# This function returns an ALTREP wrapper if the input is at least length 64
```
MichaelChirico, line 2291: that seems like an implementation detail coming from R itself? Is that guaranteed to be true in all future versions of R? If not, we should at least cite the source we're getting this from.

src/assign.c
```diff
@@ -498,11 +498,12 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
     if (length(rows)==0 && targetlen==vlen && (vlen>0 || nrow==0)) {
       if (  MAYBE_SHARED(thisvalue) ||  // set() protects the NAMED of atomic vectors from .Call setting arguments to 2 by wrapping with list
          (TYPEOF(values)==VECSXP && i>LENGTH(values)-1) || // recycled RHS would have columns pointing to others, #185.
-         (TYPEOF(values)!=VECSXP && i>0) // assigning the same values to a second column. Have to ensure a copy #2540
+         (TYPEOF(values)!=VECSXP && i>0) || // assigning the same values to a second column. Have to ensure a copy #2540
+         ALTREP(thisvalue)
```
MichaelChirico, line 502: a comment here on why this test exists would be appropriate IMO

MichaelChirico:
Hi @tlapak, would you like to revisit this PR for 1.17.0 (#6616)?

PR #5245, opened 2021-11-01
`:=` works with GForce 

ben-schwen:
Closes #1414

PR #5404, opened 2022-06-05
:=(...) with shift() - gforce follow-up #5245, #5348

### LINKED ISSUE #5403: Multiple column assignment using `:=`(...) with shift() creates list column types in 1.14.3
# Minimal Reprex

The following code works as expected

```
dt <- data.table(grp = c(1,1,1,2,2,2), x = 1:6)
dt[, x_prev := shift(x = x, type = "lag", n = 1), by = grp]
dt[, x_next := shift(x = x, type = "lead", n = 1), by = grp]

print(dt)
     grp     x x_prev x_next
   <num> <int>  <int>  <int>
1:     1     1     NA      2
2:     1     2      1      3
3:     1     3      2     NA
4:     2     4     NA      5
5:     2     5      4      6
6:     2     6      5     NA
```

but when I try to create `x_prev` and  `x_next` in one fell swoop, the resulting fields are lists of vectors

```
dt <- data.table(grp = c(1,1,1,2,2,2), x = 1:6)
dt[, `:=`(
  x_prev = shift(x = x, type = "lag", n = 1),
  x_next = shift(x = x, type = "lead", n = 1)
), by = grp]

print(dt)
     grp     x            x_prev            x_next
   <num> <int>            <list>            <list>
1:     1     1 NA, 1, 2,NA, 4, 5  2, 3,NA, 5, 6,NA
2:     1     2 NA, 1, 2,NA, 4, 5  2, 3,NA, 5, 6,NA
3:     1     3 NA, 1, 2,NA, 4, 5  2, 3,NA, 5, 6,NA
4:     2     4 NA, 1, 2,NA, 4, 5  2, 3,NA, 5, 6,NA
5:     2     5 NA, 1, 2,NA, 4, 5  2, 3,NA, 5, 6,NA
6:     2     6 NA, 1, 2,NA, 4, 5  2, 3,NA, 5, 6,NA
```

The same code works properly on data.table 1.14.2, so this appears to be a bug introduced by 1.14.3.

---

```
> sessionInfo()
R version 4.2.0 (2022-04-22)
Platform: x86_64-apple-darwin21.3.0 (64-bit)
Running under: macOS Monterey 12.4

Matrix products: default
LAPACK: /usr/local/Cellar/r/4.2.0/lib/R/lib/libRlapack.dylib

locale:

attached base packages:

other attached packages:
[1] data.table_1.14.3

[1] compiler_4.2.0  httr_1.4.3      R6_2.5.1        generics_0.1.2  tools_4.2.0     lubridate_1.8.0
```

ben-schwen:
Follow up to #5245, #5348 which implemented #1414
Closes #5403


jangorecki:
merging this PR could potentially create issues in revdeps therefore revdeps check should be repeated after merge

jangorecki (approved):

tdhock:
"revdeps check should be repeated after merge" -> revdeps check is done every evening for master branch. https://github.com/Rdatatable/data.table/wiki/Release-management-and-revdep-checks

inst/tests/tests.Rraw
```diff
@@ -18059,6 +18059,13 @@ for (opt in c(0,Inf)) {
   testnum = 2233.44
 }
 options(old)
+# optimized := with gforce functions that can return lists #5403
+DT = data.table(grp=1:2, x=1:4)
+test(2233.45, DT[, `:=`(y = shift(x, type = "lag", n = 1), z = shift(x = x, type = "lead", n = 1)), by = grp], data.table(grp=1:2, x=1:4, y=c(NA,NA,1:2), z=c(3:4, NA, NA)))
```
MichaelChirico, line 18064: why `shift(x` vs `shift(x = x`, is that testing something?

MichaelChirico, line 18064:
```suggestion
test(2233.45, DT[, `:=`(y = shift(x, type = "lag", n = 1), z = shift(x, type = "lead", n = 1)), by = grp], data.table(grp=1:2, x=1:4, y=c(NA, NA, 1:2), z=c(3:4, NA, NA)))
```

MichaelChirico, line 18064: maybe instead of `DT = ...` repeatedly, the test could be `copy(DT)[...]`, WDYT?

inst/tests/tests.Rraw
```diff
@@ -18059,6 +18059,13 @@ for (opt in c(0,Inf)) {
...
+DT = data.table(grp = 1:2, x=1:4)
+test(2233.46, DT[, l := shift(x, n=c(0, 0)), by=grp], data.table(grp=1:2, x=1:4, l=lapply(1:4, rep, 2)))
```
MichaelChirico, line 18066: any reason to use functional form `` `:=`(...)`` above, but operator form `x := y` here? Especially in the next test with length-2 LHS.

MichaelChirico, line 18066:
```suggestion
test(2233.46, DT[, l := shift(x, n=c(0, 0)), by=grp], data.table(grp=1:2, x=1:4, l=list(INT(1, 1), INT(2, 2), INT(3, 3), INT(4, 4))))
```

Reading I didn't correctly infer what the `lapply()` output would be, better to be explicit on the expected value. could also do `c(1L, 1L)` etc. to avoid `INT()`.

MichaelChirico:
I think we need more tests. What happens in the more ambiguous cases? e.g.

```r
DT <- data.table(a = 1:4, b = 5:8)
DT[, c('out1', 'out2') := shift(list(a, b), n = c(0, 0))]
```

`shift()` produces 2x2=4-item output in this case, is it valid to supply length-2 LHS to `:=`?

Perhaps more definitively, for:

```r
DT[, (...) := shift(list(a, b), n = -1:1)
```

What are valid lengths of LHS? 1, 2, 3, 6? Or only some of those.

inst/tests/tests.Rraw
ben-schwen: I guess I find the functional form more intuitive than wrapping the result inside a list when using multiple assignments. `shift` is a here a somewhat special case since it already returns a list.

MichaelChirico: got it, I see the difference now. both of the latter tests have only one call on RHS.

ben-schwen:
> I think we need more tests. What happens in the more ambiguous cases? e.g.
> 
> ```r
> DT <- data.table(a = 1:4, b = 5:8)
> DT[, c('out1', 'out2') := shift(list(a, b), n = c(0, 0))]
> ```
> 
> `shift()` produces 2x2=4-item output in this case, is it valid to supply length-2 LHS to `:=`?

Nope, it's not valid, but also never has been and I consider that actually a quite good safeguard.

```r
f = shift
DT[, c('out1', 'out2') := f(list(a, b), n = c(0, 0))]
# Error in `[.data.table`(DT, , `:=`(c("out1", "out2"), f(list(a, b), n = c(0,  : 
#   Supplied 2 columns to be assigned 4 items. Please see NEWS for v1.12.2.
DT[, c('out1', 'out2') := shift(list(a, b), n = c(0, 0))]
# Error in `[.data.table`(DT, , `:=`(c("out1", "out2"), shift(list(a, b),  : 
#   Supplied 2 columns to be assigned 4 items. Please see NEWS for v1.12.2.
```

MichaelChirico:
> Nope, it's not valid, but also never has been and I consider that actually a quite good safeguard.

Do we have regression tests for this?

ben-schwen:
> > Nope, it's not valid, but also never has been and I consider that actually a quite good safeguard.
> 
> Do we have regression tests for this?

There was one for supplying more columns than items. Now we also have the opposite direction

jangorecki:
This looks to be ready to merge, is that correct?

ben-schwen:
Yes.

jangorecki (approved):

PR #5416, opened 2022-07-06
use `pkg-config` to find compiler flags for `zlib`

### LINKED ISSUE #5622: Add pkg-config --cflags zlib to configure script
If zlib is installed in a non-standard location, then data.table will fail to install because it fails to find the header file zlib.h. This can be fixed by modifying ~/.R/Makevars, for example, but it would be easier to use pkg-config in the configure script. The configure script currently uses `pkg-config --libs zlib` and it could be modified to also use `pkg-config --cflags zlib`. I see there is an open pull request from last year that makes this change to the configure script and it works for me: https://github.com/Rdatatable/data.table/pull/5416

CreRecombinase:
Closes #5622

This pull requests ensures that R takes advantage of the info that `pkg-config`  shares regarding compilation of packages that use `zlib`


MichaelChirico:
Hi @CreRecombinase, sorry for the delay. Would you like to resume this PR?

MichaelChirico:
LGTM based on [LLM summary](https://chatgpt.com/share/613a27a5-6bc4-4a53-9a89-f2a853cec281) and my reading of `pkg-config` man page: https://linux.die.net/man/1/pkg-config

It should not break any existing usage, and only improve the functionality for some users where the linking is not done correctly.

MichaelChirico (approved):

PR #5432, opened 2022-08-07
minor programming vignette fixes

jangorecki:
this assumes next release is 1.14.4 (ant not 1.15.0)


jangorecki:
grep all files for "1.14." and fins possibly other places where version numbers have to be updated

PR #5443, opened 2022-08-22
assign 'clamp' to intermediate variable for error readability

### LINKED ISSUE #5442: Stack trace for allow.cartesian error is ugly
```
Error in vecseq(f__, len__, if (allow.cartesian || notjoin || !anyDuplicated(f__,  : 
  Join results in 6 rows; more than 5 = nrow(x)+nrow(i). Check for duplicate key values in i each of which join to the same group in x over and over again. If that's ok, try by=.EACHI to run j for each group to avoid the large allocation. If you are sure you wish to proceed, rerun with allow.cartesian=TRUE. Otherwise, please search for this error message in the FAQ, Wiki, Stack Overflow and data.table issue tracker for advice.
```

> Error in vecseq(f__, len__, if (allow.cartesian || notjoin || !anyDuplicated(f__,  : 

user doesn't need to see that / it adds a lot of distracting noise to the info we present the user.

I'm not aware of a way to set `call.=FALSE` for errors generated in C with `error()`, so instead I propose capturing that `if()` condition in a variable to at least make the error call simpler.

MichaelChirico:
Closes #5442. Not sure it warrants a NEWS item; excluded for now.


MichaelChirico:
@tdhock WDYT, worthy of a NEWS item?

jangorecki:
I would rebase that on top of mergelist or forderLazy branch. This branching is very complex and I was just recently resolving conflicts to latest master there.

MichaelChirico:
OK, marking as draft for now, pending #4370 (specifically, the #6437 chunk) or #4386

tdhock:
I don't really understand what this part of the code is doing, but if there are no user facing changes I would think NEWS item is not necessary.

MichaelChirico:
>no user facing changes

Somewhat ambiguous -- it's an error message, so it will indeed face some users. Anyway let's wait till unmarked draft to discuss further.

MichaelChirico:
> I would rebase that on top of mergelist or forderLazy branch. This branching is very complex and I was just recently resolving conflicts to latest master there.

Indeed #4370 makes this totally obsolete, already in `master` from cf7fb114ed93fe310d950f9372744371123cfd82 (#6436)

PR #5445, opened 2022-08-25
fix check as-cran note

jangorecki:
fixes
```
Found the following (possibly) invalid URLs:
  URL: https://www.h2o.ai/blog/fast-csv-writing-for-r/ (moved to https://h2o.ai/blog/fast-csv-writing-for-r/)
    From: man/fwrite.Rd
    Status: 200
    Message: OK
```


mattdowle:
@jangorecki A better PR title would have been better; i.e. mentioning something about what the note is about. It's the PR title that shows on master history. When anyone glances through the commit history it saves a little bit of time if the title is more specific but at the same time not too long. For example, in this case maybe 'Rd-only: html link update'.  I sometimes put "Rd-only" or "GLCI-only" right at the start of the title so that we know it's not a logic change. I spend quite a bit of time editing PR titles for this reason. This time I just merged but I really should have edited the title before merging or put it back to you.

PR #5446, opened 2022-08-25
Rbind allow binding of different class attributes

### LINKED ISSUE #3911: rbindlist could use an "ignore.attributes" argument
Binding `integer64` with integer/numeric works silently and correctly, whereas binding `IDate` with `Date` fails and so does binding of attribute cols where one/more tables are empty (0-rows).

### integer 64 case:

```r
require(data.table) # 1.12.2 current CRAN version, but I don't think things change with current devel)
require(bit64)

ll <- list(data.table(a=1, b=as.integer64(2)), data.table(a=2, b=1))
str(rbindlist(ll))
# Classes ‘data.table’ and 'data.frame':	2 obs. of  2 variables:
#  $ a: num  1 2
#  $ b:integer64 2 1 
```

The 2nd element's col `b` seems to be implicitly converted to `integer64` type without a message/warning.

### IDate case

```r
ll <- list(data.table(a=1, b=as.IDate(Sys.Date())), data.table(a=2, b=Sys.Date()))
rbindlist(ll)
# Error in rbindlist(ll) : 
#   Class attribute on column 2 of item 2 does not match with column 2 of item 1.
```
First element's `b` column is `IDate` and 2nd is `Date` and there's an error.

### 0-row data.table case:

```r
ll <- list(data.table(a=1, b=Sys.time()), data.table(a=numeric(0), b=numeric(0)))
rbindlist(ll)
# Error in rbindlist(ll) : 
#   Class attribute on column 2 of item 2 does not match with column 2 of item 1.
```
Even if the empty table has a wrong type, it could be skipped as it doesn't really make it to the final table. (This used to work in previous versions when `fill=TRUE` -- probably accidental).

---

In essence, I'd think it'd be immensely convenient to have an argument to *ignore strict attributes check*, e.g., `ignore.attr=FALSE` (together with a global option so that code that has been written with old+lenient behaviour work continue to work) which would simply retain the first element's class.

And if the default is to be strict, then, the first case should also error, for consistency.

### LINKED ISSUE #4934: `rbindlist` fails in the presence of AsIs
This might be by design, but I couldn't find an explicit mention of it in `NEWS`, issue reports or elsewhere. 
We're trying to upgrade `data.table` from 1.11.8 to 1.13.6, and some of our tests break on code that simplifies to this:
```
d <- list()
d[[1]] <- data.frame(a=1, b=I(2))
d[[2]] <- data.frame(a=3, b=4)
rbindlist(d, fill=TRUE)
```
The `rbindlist` call succeeds in 1.11.8, but in 1.13.6 fails with:

> Error in rbindlist(d, fill = TRUE) : 
  Class attribute on column 2 of item 2 does not match with column 2 of item 1.

We suspect it's a regression, as `AsIs` class should not account as a class change - just inhibit coercion.

-----------------------------------------------
Edit: turns out this surfaced due to an R3.4 -> R4.0.3 upgrade, not a data.table upgrade. Still, suggest it be considered for fixing.

`#` `Output of sessionInfo()`
> R version 4.0.3 (2020-10-10)
> Platform: x86_64-pc-linux-gnu (64-bit)
> Running under: Ubuntu 18.04.5 LTS
> 
> Matrix products: default
> BLAS:   /mnt/public/istra/apps/R/R-4.0.3.mkl/lib/libRblas.so
> LAPACK: /mnt/public/istra/apps/R/R-4.0.3.mkl/lib/libRlapack.so
> 
> locale:
>  [1] LC_CTYPE=en_US.UTF-8 LC_NUMERIC=C         LC_TIME=C            LC_COLLATE=C         LC_MONETARY=C        LC_MESSAGES=C       
>  [7] LC_PAPER=C           LC_NAME=C            LC_ADDRESS=C         LC_TELEPHONE=C       LC_MEASUREMENT=C     LC_IDENTIFICATION=C 
> 
> attached base packages:
> [1] datasets  utils     stats     graphics  grDevices methods   base     
> ...

### LINKED ISSUE #5309: Rbind in 1.14.3 doesn't like POSIX
Looks like rbind (and by extension merge(all = T)) doesn't like POSIXct type data in 1.14.3. To the extent this is a legit bug, I think its been introduced relatively recently (code running on an older version of 1.14.3 seemed to work just fine).

``` r
Sys.setenv(TZ = 'America/Los_Angeles')
library('data.table')
#> Warning: package 'data.table' was built under R version 4.1.2
t2 = data.table(a = NA_real_, b = NA_real_, c = NA_real_, d= as.Date(NA))
t1 = data.table(a = 1.1, b = 1.1, c = 1.1, d = as.Date('2021-10-05'), e = as.POSIXct("2021-10-06 13:58:00 UTC"))
r2 = rbind(t1,t2, fill = T, use.names = F)
#> Error in rbindlist(l, use.names, fill, idcol): Class attribute on column 5 of item 2 does not match with column 5 of item 1.

sessionInfo()
#> R version 4.1.0 (2021-05-18)
#> Running under: Windows Server x64 (build 19041)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.1252 
#> [2] LC_CTYPE=English_United States.1252   
#> [3] LC_MONETARY=English_United States.1252
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.1252    
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.14.3
#> 
#>  [1] ps_1.6.0        digest_0.6.27   withr_2.4.2     magrittr_2.0.1 
#>  [5] reprex_2.0.1    evaluate_0.14   highr_0.9       stringi_1.6.2  
#>  [9] rlang_0.4.11    cli_2.5.0       rstudioapi_0.13 fs_1.5.0       
#> [13] rmarkdown_2.11  tools_4.1.0     stringr_1.4.0   glue_1.4.2     
#> [17] xfun_0.29       yaml_2.2.1      fastmap_1.1.0   compiler_4.1.0 
#> [21] htmltools_0.5.2 knitr_1.33
```

Under 1.14.2:

``` r
Sys.setenv(TZ = 'America/Los_Angeles')
library('data.table')
#> Warning: package 'data.table' was built under R version 4.1.2
t2 = data.table(a = NA_real_, b = NA_real_, c = NA_real_, d= as.Date(NA))
t1 = data.table(a = 1.1, b = 1.1, c = 1.1, d = as.Date('2021-10-05'), e = as.POSIXct("2021-10-06 13:58:00 UTC"))
r2 = rbind(t1,t2, fill = T, use.names = F)
#> Warning in rbindlist(l, use.names, fill, idcol): use.names= cannot be FALSE when
#> fill is TRUE. Setting use.names=TRUE.

sessionInfo()
#> R version 4.1.0 (2021-05-18)
#> Running under: Windows Server x64 (build 19041)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.1252 
#> [2] LC_CTYPE=English_United States.1252   
#> [3] LC_MONETARY=English_United States.1252
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.1252    
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.14.2
#> 
#>  [1] ps_1.6.0        digest_0.6.27   withr_2.4.2     magrittr_2.0.1 
#>  [5] reprex_2.0.1    evaluate_0.14   highr_0.9       stringi_1.6.2  
#>  [9] rlang_0.4.11    cli_2.5.0       rstudioapi_0.13 fs_1.5.0       
#> [13] rmarkdown_2.11  tools_4.1.0     stringr_1.4.0   glue_1.4.2     
#> [17] xfun_0.29       yaml_2.2.1      fastmap_1.1.0   compiler_4.1.0 
#> [21] htmltools_0.5.2 knitr_1.33
```

### LINKED ISSUE #5391: Error related to merging on character vs factor column
Hello.
I've tested `data.table` dev version 1.14.3 as it has so many features and bugfixes.
Unfortunately package [`bupaverse/processmapR`](https://github.com/bupaverse/processmapR)) become broken due some changes in `data.table` internal logic. 
I suppose current issue can be a clarification of https://github.com/Rdatatable/data.table/issues/5361 as it influence @ work of another packages, not my personal code.

Error can be reproduced enven in public [`bupaR`](https://bupar.net/processmaps.html) sample:
```
library(bupaR)
patients %>%
    process_map()
#> Error in rbindlist(l, use.names, fill, idcol) : 
#>   Class attribute on column 7 of item 2 does not match with column 7 of item 1.
```

More precisely, I tried to debug and found that error raised at this line https://github.com/bupaverse/processmapR/blob/6dbf8d776204b6706432f2fdf81b22a1ab6dcaae/R/process_map.R#L198
during `data.table::merge` by ordered factor and character.

When I rollback `data.table` to version 1.14.2 everything returns to working state.

SessionInfo()
```
R version 4.2.0 (2022-04-22 ucrt)

Matrix products: default

attached base packages:

other attached packages:
 [1] petrinetR_0.2.1     processmonitR_0.1.0 xesreadR_0.2.3      processmapR_0.3.4   eventdataR_0.3.0   
 [6] edeaR_0.8.6         bupaR_0.4.4         fstcore_0.9.12      lgr_0.4.3           tsibble_1.1.1      
[11] furrr_0.3.0         future_1.25.0       dqrng_0.3.0         rTRNG_4.23.1-2      anytime_0.3.9      
[16] stringi_1.7.6       data.table_1.14.3   tictoc_1.0.1        datapasta_3.1.0     magrittr_2.0.3     
[21] forcats_0.5.1       stringr_1.4.0       dplyr_1.0.9         purrr_0.3.4         readr_2.1.2        
[26] tidyr_1.2.0         tibble_3.1.7        ggplot2_3.3.6       tidyverse_1.3.1    
```

### LINKED ISSUE #5542: revdeps new errors in rbindlist due to class mis-match
revdep heuristicsmineR has the following new error when running examples if data.table from github master is installed,
```
* checking examples ... ERROR
Running examples in 'heuristicsmineR-Ex.R' failed
The error most likely occurred in:

> ### Name: as.petrinet
> ### Title: Converts the object to a Petrinet
> ### Aliases: as.petrinet
> 
> ### ** Examples
> 
> data(L_heur_1)
> cn <- causal_net(L_heur_1, threshold = .8)
Error in rbindlist(l, use.names, fill, idcol) : 
  Class attribute on column 7 of item 2 does not match with column 7 of item 1.
Calls: causal_net ... merge -> merge.data.table -> rbind -> rbind -> rbindlist
Execution halted
```
git bisect says this started happening after merging https://github.com/Rdatatable/data.table/pull/5263
@ben-schwen can you please have a look? (you were author of that PR)

ben-schwen:
Closes #5309 (shadowing current approach for `int64` and `factor`)
Closes #3911 (automatically allows for mixing `Date` and `IDate` and adds `ignore.attr` argument)
Closes #4934 (only carries class of `AsIs` to result if `AsIs` is first in binding to stay conform with `do.call(rbind, list)` in this case)
Closes #5391
Closes #5542
Towards #5486 also needs #5569

- [x] Mix/fill dates (`Date` and `IDate`) with `atomic` columns
- [x] Mix/fill `POSIXct` with `atomic` columns
- [x] Mix/fill `ITime`
- [x] Mix/fill `AsIs`
- [x] Add `ignore.attr` argument `rbindlist` and `rbind` to manually deactivate check for equal classes of binding columns 
- [x] Tests
- [x] News


R/merge.R
```diff
@@ -97,7 +97,7 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
     # Perhaps not very commonly used, so not a huge deal that the join is redone here.
     missingyidx = y[!x, which=TRUE, on=by, allow.cartesian=allow.cartesian]
     if (length(missingyidx)) {
-      dt = rbind(dt, y[missingyidx], use.names=FALSE, fill=TRUE)
+      dt = rbind(dt, y[missingyidx], use.names=FALSE, fill=TRUE, ignore.attr=TRUE)
```
jangorecki, line 100: Do we have unit test that covers that? It looks like a breaking change, therefore should be introduced softly, having backward compatible default for 1-2 versions, before switching to new behavior.

jangorecki:
This PR resolves one of the issues on 1.15.0 milestone. But it is much bigger than just fixing regression, it adds new feature. I would therefore prefer to shift it to 1.15.99, and for 1.15.0 push only regression fixes.

MichaelChirico:
Agree with Jan, it would be nice to merge a minimal PR that fixes regression to send out as 1.15.0. I glanced at the diff here and didn't see an easy way to separate out the regression fix from the new functionality, is that possible @ben-schwen? Or should we work on that as a standalone PR?

ben-schwen:
#5309 has two issues in it:
1) `rbindlist(..., fill=TRUE, use.names=FALSE)`. This did not work previously and threw a warning changing the value of `use.names` to `TRUE`.
```r
Warning message:
In rbindlist(l, use.names, fill, idcol) :
  use.names= cannot be FALSE when fill is TRUE. Setting use.names=TRUE.
```
This error of #5309 is already fixed by #5468

2) merge used `rbind(dt, yy, use.name=FALSE, fill=FALSE)` and manually filled yy before binding.
There is a working version of merge without `rbindlist(fill=TRUE, use.names=FALSE)` in #5263 but I would even roll back merge to version pre #5263 (since it was only a minor). Will add tests of #5263 and file regression PR.

MichaelChirico:
Is this ready for review? If so, please resolve the R/merge.R conflict. Thanks!

R/merge.R
ben-schwen: We have now since #5857. I'm indifferent whether we add the change to merge or not. It makes the code much shorter and does not hide so much what we want to achieve but since many ppl rely on `merge` we should be careful about changes.

ben-schwen:
> Is this ready for review? If so, please resolve the R/merge.R conflict. Thanks!

Yes. Only need to decide whether we want to include the change in `merge` together with the test 2253.31 or not.

NEWS.md
```diff
@@ -431,6 +434,7 @@
 
 41. `tables()` is faster by default by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). `mb=` now accepts any function which accepts a `data.table` and returns a higher and better estimate of its size in bytes, albeit more slowly; e.g. `mb = utils::object.size`.
 
+
```
MichaelChirico, line 437:
```suggestion
```

inst/tests/other.Rraw
```diff
@@ -699,6 +699,14 @@ if (loaded[["nanotime"]]) {
   DT = data.table(time=nanotime(c(1,NA,3)))
   test(27, na.omit(DT), DT[c(1,3)])
 
+  # was 2239 in tests.Rraw, rbind with vectors with class attributes #5309
```
MichaelChirico, line 702: is there more context here? I see this as 2239 now (L18210):

```r
test(2239.1, shift(matrix(1:10, ncol = 1)), error="consider wrapping")
```

man/rbindlist.Rd
```diff
@@ -7,14 +7,15 @@
...
+rbindlist(l, use.names="check", fill=FALSE, idcol=NULL, ignore.attr=FALSE)
 # rbind(..., use.names=TRUE, fill=FALSE, idcol=NULL)
 }
 \arguments{
   \item{l}{ A list containing \code{data.table}, \code{data.frame} or \code{list} objects. \code{\dots} is the same but you pass the objects by name separately. }
   \item{use.names}{\code{TRUE} binds by matching column name, \code{FALSE} by position. `check` (default) warns if all items don't have the same names in the same order and then currently proceeds as if `use.names=FALSE` for backwards compatibility (\code{TRUE} in future); see news for v1.12.2.}
   \item{fill}{\code{TRUE} fills missing columns with NAs, or NULL for missing list columns. By default \code{FALSE}.}
   \item{idcol}{Creates a column in the result showing which list item those rows came from. \code{TRUE} names this column \code{".id"}. \code{idcol="file"} names this column \code{"file"}. If the input list has names, those names are the values placed in this id column, otherwise the values are an integer vector \code{1:length(l)}. See \code{examples}.}
+  \item{ignore.attr}{\code{TRUE} allows binding columns with different class attributes. By default \code{FALSE}.}
```
MichaelChirico, line 18:
```suggestion
  \item{ignore.attr}{Logical, default \code{FALSE}. When \code{TRUE}, allows binding columns with different class attributes.}
```

MichaelChirico: I would expect _all_ attributes to be ignored, not just 'class', can we clarify that?

src/rbindlist.c
```diff
@@ -2,16 +2,19 @@
...
     error(_("%s should be TRUE or FALSE"), "fill");
   if (!isLogical(usenamesArg) || LENGTH(usenamesArg)!=1)
     error(_("use.names= should be TRUE, FALSE, or not used (\"check\" by default)"));  // R levels converts "check" to NA
+  if (!isLogical(ignoreattrArg) || LENGTH(ignoreattrArg)!=1 || LOGICAL(ignoreattrArg)[0] == NA_LOGICAL)
+    error(_("ignore.attr= should be TRUE or FALSE"));
   if (!length(l)) return(l);
   if (TYPEOF(l) != VECSXP) error(_("Input to rbindlist must be a list. This list can contain data.tables, data.frames or plain lists."));
   Rboolean usenames = LOGICAL(usenamesArg)[0];
   const bool fill = LOGICAL(fillArg)[0];
+  const bool ignoreatt = LOGICAL(ignoreattrArg)[0];
```
MichaelChirico, line 17: nit: should this be `ignoreattr`?

src/rbindlist.c
```diff
@@ -305,14 +308,25 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       if (INHERITS(thisCol, char_integer64)) {
         if (firsti>=0 && !length(getAttrib(firstCol, R_ClassSymbol))) { firsti=i; firstw=w; firstCol=thisCol; } // so the integer64 attribute gets copied to target below
         int64=true;
+      } else if (INHERITS(thisCol, char_Date))  {
+        if (firsti>=0 && !length(getAttrib(firstCol, R_ClassSymbol))) { firsti=i; firstw=w; firstCol=thisCol; }
```
MichaelChirico, line 312: we don't need to `PROTECT()` the `getAttrib()` result because it's not stored, right?

inst/tests/tests.Rraw
```diff
@@ -18732,3 +18734,46 @@ x = list(
 N = 5e4
 y = x[rep(1:5, N)]
 test(2268, rbindlist(y, fill=TRUE), rbindlist(x, fill=TRUE)[rep(1:5, N)])
+
+# rbind with vectors with class attributes #5309
+x = data.table(a = 1L, b = as.Date("2020-01-01"))
+y = data.table(a = 2L, b = as.IDate("2021-01-01"))
+z = data.table(a = 3L, b = NA)
+test(2269.01, rbind(x, y), data.table(a=1:2,       b= as.Date(c("2020-01-01", "2021-01-01"))))
+test(2269.02, rbind(y, x), data.table(a=c(2L, 1L), b=as.IDate(c("2021-01-01", "2020-01-01"))))
```
MichaelChirico, line 18743:
```suggestion
test(2269.01, rbind(x, y), data.table(a=c(1L, 2L), b= as.Date(c("2020-01-01", "2021-01-01"))))
test(2269.02, rbind(y, x), data.table(a=c(2L, 1L), b=as.IDate(c("2021-01-01", "2020-01-01"))))
```

MichaelChirico: or, use `2:1`, `2:3`, `3:2` for consistency below

inst/tests/tests.Rraw
```diff
@@ -18732,3 +18734,46 @@ x = list(
...
+test(2269.07, rbind(x, z, fill=TRUE), data.table(a=c(1L, 3L), b= as.Date(c("2020-01-01", NA))))
+test(2269.08, rbind(z, x, fill=TRUE), data.table(a=c(3L, 1L), b= as.Date(c(NA, "2020-01-01"))))
+test(2269.09, rbind(y, z, fill=TRUE), data.table(a=c(2L, 3L), b=as.IDate(c("2021-01-01", NA))))
+test(2269.10, rbind(z, y, fill=TRUE), data.table(a=c(3L, 2L), b=as.IDate(c(NA, "2021-01-01"))))
+x = data.table(a=1L, b=as.POSIXct("2021-10-06 13:58:00 UTC"))
+test(2269.11, rbind(x, z, fill=TRUE), data.table(a=c(1L, 3L), b=as.POSIXct(c("2021-10-06 13:58:00 UTC", NA))))
+test(2269.12, rbind(z, x, fill=TRUE), data.table(a=c(3L, 1L), b=as.POSIXct(c(NA, "2021-10-06 13:58:00 UTC"))))
+x = data.table(c=1L, d=as.POSIXct("2021-10-06 13:58:00 UTC"))
+test(2269.13, rbind(x, z, fill=TRUE, use.names=FALSE), data.table(c = c(1L, 3L), d=as.POSIXct(c("2021-10-06 13:58:00 UTC", NA))))
+test(2269.14, setnames(rbind(z, x, fill=TRUE, use.names=FALSE), c("c", "d")), data.table(c=c(3L, 1L), d=as.POSIXct(c(NA, "2021-10-06 13:58:00 UTC"))))
```
MichaelChirico, line 18758: why use `setnames()` here?

inst/tests/tests.Rraw
```diff
@@ -18732,3 +18734,46 @@ x = list(
...
+x = data.table(a=1L, b=as.ITime(0))
+y = data.table(a=2L, b=NA)
+test(2269.15, rbind(x,y), data.table(a = c(1L, 2L), b=as.ITime(c(0, NA))))
+test(2269.16, rbind(y,x), data.table(a = c(2L, 1L), b=as.ITime(c(NA, 0))))
+y[, b := NULL]
+test(2269.17, rbind(x,y, fill = TRUE), data.table(a = c(1L, 2L), b=as.ITime(c(0, NA))))
+test(2269.18, rbind(y,x, fill = TRUE), data.table(a = c(2L, 1L), b=as.ITime(c(NA, 0))))
+x = data.table(a = 1L, b = as.Date("2020-01-01"))
+y = data.table(a = 2L, b = NA)
+test(2269.19, merge(x, y, by="a", all=TRUE), data.table(a=1:2, b.x=as.Date(c("2020-01-01", NA)), b.y=NA, key="a"))
```
MichaelChirico, line 18768: comment here about why `merge()` is being grouped in with the `rbind()` tests (probably just a cite of the related issues right?)

inst/tests/tests.Rraw
```diff
@@ -18732,3 +18734,46 @@ x = list(
 N = 5e4
 y = x[rep(1:5, N)]
 test(2268, rbindlist(y, fill=TRUE), rbindlist(x, fill=TRUE)[rep(1:5, N)])
+
+# rbind with vectors with class attributes #5309
```
MichaelChirico, line 18738: All of this uses `rbind()`, which test `rbindlist()` implicitly, but that's somewhat of an implementation detail, so let's add at least some tests on `rbindlist()` directly.

Also I don't see any test involving the `AsIs` logic that is included here.

MichaelChirico:
All LGTM, minor requests for further improvement. Defer to @jangorecki about `merge()`, I think it's fine though, AIUI the main concern was about including too much in a patch release, now we have "normal" release the issue is resolved.

jangorecki:
Probably full outer join do rbind

inst/tests/other.Rraw
ben-schwen: The original test case number for this PR was 2239, but yes it sounded confusing

man/rbindlist.Rd
ben-schwen: Yes, but we only checked for equality of class attributes. Clarified it!

src/rbindlist.c
ben-schwen: I think we do not have to `PROTECT` since it is not allocating something new but only a getter.
And they never use `PROTECT` for getAttrib in WRE

inst/tests/tests.Rraw
ben-schwen: I added an comment to mark the `AsIs` cases better and also included the `rbindlist` there

inst/tests/tests.Rraw
ben-schwen: good catch, was an attempt to lazy recycle y input

NEWS.md
```diff
@@ -40,6 +40,10 @@
 
 14. `fread` loads `.bgz` files directly, [#5461](https://github.com/Rdatatable/data.table/issues/5461). Thanks to @TMRHarrison for the request with proposed fix, and Benjamin Schwendinger for the PR.
 
+15. `rbindlist(l, use.names=TRUE)` and `rbind` allows now the automatic binding of columns with different class attributes for certain classes such as `Date`, `IDate`, `ITime`, `POSIXct` and `AsIs` with other columns of similar classes, e.g., `IDate` and `Date`. The conversion is done automatically and the class attribute of the final column is determined by the first encountered class attribute in the binding list, [#5309](https://github.com/Rdatatable/data.table/issues/5309), [#4934](https://github.com/Rdatatable/data.table/issues/4934), [#5391](https://github.com/Rdatatable/data.table/issues/5391).
```
MichaelChirico, line 43:
```suggestion
15. `rbindlist(l, use.names=TRUE)` and `rbind` now works correctly on columns with different class attributes for certain classes such as `Date`, `IDate`, `ITime`, `POSIXct` and `AsIs` with other columns of similar classes, e.g., `IDate` and `Date`. The conversion is done automatically and the class attribute of the final column is determined by the first encountered class attribute in the binding list, [#5309](https://github.com/Rdatatable/data.table/issues/5309), [#4934](https://github.com/Rdatatable/data.table/issues/4934), [#5391](https://github.com/Rdatatable/data.table/issues/5391).
```

NEWS.md
```diff
@@ -40,6 +40,10 @@
...
 
+15. `rbindlist(l, use.names=TRUE)` and `rbind` now works correctly on columns with different class attributes for certain classes such as `Date`, `IDate`, `ITime`, `POSIXct` and `AsIs` with other columns of similar classes, e.g., `IDate` and `Date`. The conversion is done automatically and the class attribute of the final column is determined by the first encountered class attribute in the binding list, [#5309](https://github.com/Rdatatable/data.table/issues/5309), [#4934](https://github.com/Rdatatable/data.table/issues/4934), [#5391](https://github.com/Rdatatable/data.table/issues/5391).
+
+`rbindlist(l, ignore.attr=TRUE)` and `rbind` also gained argument `ignore.attr` to manually deactivate the safety-net of binding columns with different column classes, [#3911](https://github.com/Rdatatable/data.table/issues/3911), [#5542](https://github.com/Rdatatable/data.table/issues/5542). Thanks to @dcaseykc, @fox34, @adrian-quintario, @berg-michael, @arunsrinivasan, @statquant, @pkress, @jrausch12, @therosko, @OfekShilon, @iMissile, @tdhock for the request and @ben-schwen for the PR.
```
MichaelChirico, line 45:
```suggestion
`rbindlist(l, ignore.attr=TRUE)` and `rbind` also gains argument `ignore.attr` to manually deactivate the safety-net of binding columns with different column classes, [#3911](https://github.com/Rdatatable/data.table/issues/3911), [#5542](https://github.com/Rdatatable/data.table/issues/5542). Thanks to @dcaseykc, @fox34, @adrian-quintario, @berg-michael, @arunsrinivasan, @statquant, @pkress, @jrausch12, @therosko, @OfekShilon, @iMissile, @tdhock for the request and @ben-schwen for the PR.
```

src/rbindlist.c
```diff
@@ -2,16 +2,19 @@
...
 
-SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
+SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignoreattrArg)
 {
   if (!isLogical(fillArg) || LENGTH(fillArg) != 1 || LOGICAL(fillArg)[0] == NA_LOGICAL)
     error(_("%s should be TRUE or FALSE"), "fill");
   if (!isLogical(usenamesArg) || LENGTH(usenamesArg)!=1)
     error(_("use.names= should be TRUE, FALSE, or not used (\"check\" by default)"));  // R levels converts "check" to NA
+  if (!isLogical(ignoreattrArg) || LENGTH(ignoreattrArg)!=1 || LOGICAL(ignoreattrArg)[0] == NA_LOGICAL)
+    error(_("ignore.attr= should be TRUE or FALSE"));
```
MichaelChirico, line 12:
```suggestion
    error(_("%s should be TRUE or FALSE"), "ignore.attr");
```

src/rbindlist.c
```diff
@@ -305,14 +308,25 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+      } else if (INHERITS(thisCol, char_ITime)) {
+        if (firsti>=0 && !length(getAttrib(firstCol, R_ClassSymbol))) { firsti=i; firstw=w; firstCol=thisCol; }
+        itime=true;
+      } else if (!asis && INHERITS(thisCol, char_AsIs)) {
+        asis=true;
       }
       if (firsti==-1) { firsti=i; firstw=w; firstCol=thisCol; }
       else {
-        if (!factor && !int64) {
+        if (!factor && !int64 && ((!date && !posixct) || (date && posixct)) && !itime &!asis) { // prohibit binding of date and posixct
```
MichaelChirico, line 325:
```suggestion
        if (!factor && !int64 && date == posixct && !itime &!asis) { // prohibit binding of date and posixct
```

src/rbindlist.c
```diff
@@ -305,14 +308,25 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+      } else if (INHERITS(thisCol, char_ITime)) {
+        if (firsti>=0 && !length(getAttrib(firstCol, R_ClassSymbol))) { firsti=i; firstw=w; firstCol=thisCol; }
+        itime=true;
+      } else if (!asis && INHERITS(thisCol, char_AsIs)) {
+        asis=true;
       }
       if (firsti==-1) { firsti=i; firstw=w; firstCol=thisCol; }
       else {
-        if (!factor && !int64) {
+        if (!factor && !int64 && date == posixct && !itime &!asis) { // prohibit binding of date and posixct
```
MichaelChirico, line 325: `&!` looks like a typo, it should be `&& !` right? Should have been caught by a test, please add

src/rbindlist.c
MichaelChirico: We do use `PROTECT(getAttrib())` in a few places, so I've marked this as something to look into in #6280. Will be easier after #6263

src/rbindlist.c
ben-schwen: kind of strange since we have an `asis` test

MichaelChirico: I wonder if it's just skirting by based on the bitwise comparison "just working"... let's just fix it and proceed unless you have any idea for a different test...

MichaelChirico, line 325:
```suggestion
        if (!factor && !int64 && date == posixct && !itime && !asis) { // prohibit binding of date and posixct
```

MichaelChirico (approved):

PR #5448, opened 2022-08-26
rbindlist protection stack overflow

### LINKED ISSUE #4536: rbindlist protection stack overflow upon tables with nested lists
I am having an issue with Rstudio and how you cannot adjust it's pp_size like you can normal R instances.  The problem results from working on large datasets, but these datasets are well within my computer's memory limits.  Data.table -- like usual -- is my saving grace because of its speed and stability, however, I have come into a situation where I am greedily leveraging rbindlist and erroring out.

The following is an example of a list where each data.table is the result of unstructured extracted JSON. As you can see it is pretty dirty to begin with. Many of the fields in the table are actually lists. I normally do a ton of work using list fields in data.tables so this is nothing new to me.  The reason I don't want the list fields to be blown out into their own columns yet because of name conflicts and just not wanting to deal with it until I am ready. 

```R
# Run from Rstudio ####
Cstack_info()
#     size    current  direction eval_depth 
# 15922790      20472          1          2 
library('data.table')
options(datatable.print.class = T)
l <- list(
  data.table(V1 = "a", V4 = 1L, V5 = "", V14 = T),
  data.table(V1 = "", V4 = 0L, V5 = ""),
  data.table(V2 = list(F), V6 = "", V7 = F, V8 = F, V9 = "", V10 = "", V11 = T),
  data.table(V2 = list(list(V2A = F, V2B = "", V2C = "" )),
             V6 = "", V7 = T, V8 = F, V12 = "", V9 = "", V15 = "", 
             V10 = "", V16 = F, V11 = F, V17 = F),
  data.table(V18 = "", V19 = "",
             V1 = list(list(V1A = "", V1B = "", V1C = "", V1D = "", V1E = "")),
             V20 = "", V21 = "",
             V3 = list(list(V3A = T, V3B = "", V3C = 1L, V3D = 0L, V3F = 1L)),
             V13 = "", V14 = T)
)
l <- l[sample(seq(l), 2e5, replace = T)]
d <- rbindlist(l, fill = T)
# Error: protect(): protection stack overflow
```
Now, I know the problem isn't exactly to do with data.table's call to `.Call(Crbindlist, l, use.names, fill, idcol)`. The reason being is that this process works effortlessly when I call it in a separate R instance using the max-ppsize switch (`--max-ppsize=500000`).
```R
## IN Rgui.exe --max-ppsize=500000 ##
system("\"C:\\Program Files\\R\\R-4.0.1\\bin\\x64\\Rgui.exe\" --max-ppsize=500000")
# In the instance:
# > Cstack_info()
#     size    current  direction eval_depth 
# 63737856      11288          1          2 
```
I have tried everything I can think, and everything I could find online (including playing with envir variables [Defn.h](https://github.com/wch/r-source/blob/55ca9df310f2a93879c8deeec1d277011af75500/src/include/Defn.h#L683)). You can't seem to change Rstudio's Cstack size.  This might be intentional on Rstudio's part seeing as they require you to pay for the pro version of RS server if you want to change this setting.  The following does nothing:
```R
system("\"C:\\Program Files\\RStudio\\bin\\rstudio.exe\" --max-ppsize=500000")
options("expressions" = 5e5)
memory.limit(size = 800000000)
```
After investigating with Rprofmem and tracemem there is really no place I can find to reduce the stack size programming-wise. Also, I couldn't find a way to change how Rstudio calls it's renv initially, or how to change R's embedded defaults. 

So I am out of easy ideas at this point. This is frustrating because I not really reaching the limits of this computer which I created for this kind of work. My workaround looks like this:
```R
library('foreach')
l_brks <-  seq(1L, length(l), length.out = 3)
l_cuts <- cut(seq(l), breaks = l_brks, include.lowest = T)
f_rbind <- function(...) {rbind(..., fill = T)}
d <- foreach(i = levels(l_cuts), .combine = f_rbind, .inorder = T) %do% {
  rbindlist(l[which(i == l_cuts)], fill = T)
}
```
The reason I used foreach is for the potential of dispatching jobs to R instances where I might be able to get more ppsize control when I learn more about it.  My solition isn't as inelegant when compared with a single call to `rbindlist(l, fill = T)`, but at least it does the job.

At the end of the day, I really don't know where (if any) the problem lies so I would be very grateful for your guys' insights on the matter.

ben-schwen:
Closes #4536

- [x] Fix
- [x] News
- [x] Test


src/rbindlist.c
```diff
@@ -513,13 +513,15 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+          bool listprotect = false;
           if ((TYPEOF(target)==VECSXP || TYPEOF(target)==EXPRSXP) && TYPEOF(thisCol)!=TYPEOF(target)) {
             // do an as.list() on the atomic column; #3528
-            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); nprotect++;
+            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); listprotect = true;
           }
           // else coerces if needed within memrecycle; with a no-alloc direct coerce from 1.12.4 (PR #3909)
           const char *ret = memrecycle(target, R_NilValue, ansloc, thisnrow, thisCol, 0, -1, idcol+j+1, foundName);
           if (ret) warning(_("Column %d of item %d: %s"), w+1, i+1, ret);
+          if (listprotect) UNPROTECT(listprotect); // earlier unproject rbindlist calls with lots of lists #4536
```
jangorecki, line 524: Typo in comment

jangorecki:
Adding code from issue that this PR closes won't fit for unit test for some reason?

ben-schwen:
> Thank you for fix. Adding code from issue that this PR closes won't fit for unit test for some reason?

Didn't want to add that one since it takes over 1s and testing already takes some time.

src/rbindlist.c
```diff
@@ -515,13 +515,15 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
...
+          bool listprotect = false;
           if ((TYPEOF(target)==VECSXP || TYPEOF(target)==EXPRSXP) && TYPEOF(thisCol)!=TYPEOF(target)) {
             // do an as.list() on the atomic column; #3528
-            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); nprotect++;
+            thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target))); listprotect = true;
           }
           // else coerces if needed within memrecycle; with a no-alloc direct coerce from 1.12.4 (PR #3909)
           const char *ret = memrecycle(target, R_NilValue, ansloc, thisnrow, thisCol, 0, -1, idcol+j+1, foundName);
           if (ret) warning(_("Column %d of item %d: %s"), w+1, i+1, ret);
+          if (listprotect) UNPROTECT(1); // earlier unprotect rbindlist calls with lots of lists #4536
```
MichaelChirico, line 526: Should the `UNPROTECT()` precede the `warning()` for the case when `options(warn=2)`?

MichaelChirico: Yes, I _think_ we're meant to always `UNPROTECT()` before signalling `error()` or `warning()`, e.g. from [WRE](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Parsing-R-code-from-C-1):

```c
#include <R.h>
#include <Rinternals.h>
#include <R_ext/Parse.h>

SEXP menu_ttest3()
{
    char cmd[256];
    SEXP cmdSexp, cmdexpr, ans = R_NilValue;
    ParseStatus status;
   ...
    if(done == 1) {
        cmdSexp = PROTECT(allocVector(STRSXP, 1));
        SET_STRING_ELT(cmdSexp, 0, mkChar(cmd));
        cmdexpr = PROTECT(R_ParseVector(cmdSexp, -1, &status, R_NilValue));
        if (status != PARSE_OK) {
            UNPROTECT(2);
            error("invalid call %s", cmd);
        }
        /* Loop is needed here as EXPSEXP will be of length > 1 */
        for(int i = 0; i < length(cmdexpr); i++)
            ans = eval(VECTOR_ELT(cmdexpr, i), R_GlobalEnv);
        UNPROTECT(2);
    }
    return ans;
}
```

IINM {rchk} is supposed to help ferret those out, so I'm not sure we have many violations for the case of `error()`, can you easily find any?

We might want to report to {rchk} maintainer T. Kalibera about possibly catching usages like `warning(); UNPROTECT();` as well by that package.

cc @jangorecki as well who has worked more extensively here.

inst/tests/tests.Rraw
```diff
@@ -18566,3 +18566,15 @@ test(2261.04, setNumericRounding(2L), 1L)
...
+
+# rbindlist with many nested lists leads to protection stack overflow #4536
+x = list(
+  data.table(V1="a", V4=1L, V5="", V14=TRUE),
+  data.table(V1="",  V4=0L, V5=""),
+  data.table(V2=list(FALSE), V6="", V7=FALSE, V8=FALSE, V9="", V10="", V11=TRUE),
+  data.table(V2=list(list(V2A=FALSE, V2B="", V2C="" )), V6="", V7=TRUE, V8=FALSE, V12="", V9="", V15="", V10="", V16=FALSE, V11=FALSE, V17=FALSE),
+  data.table(V18="", V19="", V1=list(list(V1A="", V1B="", V1C="", V1D="", V1E="")), V20="", V21="", V3 = list(list(V3A=TRUE, V3B="", V3C=1L, V3D=0L, V3F=1L)), V13="", V14=TRUE)
+)
+N = 5e4
```
MichaelChirico, line 18578: I missed the subtlety of this test on the first read, could you please add a comment on why this part is crucial to the test failing on `master` but fixed here? I think it will be useful for posterity.

MichaelChirico (approved):
LGTM, I will resolve the merge conflicts after the new comment if you'd like :)

src/rbindlist.c
ben-schwen: I just took a glance at `assign.c` and `between.c` but both contain those "violations".

MichaelChirico: ok let's file a follow-up

inst/tests/tests.Rraw
ben-schwen: done. I also simplified the test case (still failing but cut down to the essential part)

MichaelChirico (approved):

PR #5453, opened 2022-08-31
rbindlist(l, use.names=TRUE) handle different encodings for column names

### LINKED ISSUE #5452: `rbind` cannot handle different name encodings
In the following case, the bind does not work:

```R
x <- data.table(A = 1, B = 2, C = 3)
y <- copy(x)
setnames(x , c("Ä", "Ö", "Ü"))
setnames(y , iconv(c("Ä", "Ö", "Ü"), from = "UTF-8", to = "latin1"))
Encoding(names(x))
Encoding(names(y))
rbind(x,y)
```

Output:
```
[1] "UTF-8" "UTF-8" "UTF-8"
[1] "latin1" "latin1" "latin1"

Error in rbindlist(l, use.names, fill, idcol) (rbindbug.R#7): Column 1 ['Ä'] of item 2 is missing in item 1. Use fill=TRUE to fill with NA (NULL for list columns), or use.names=FALSE to ignore column names.
Show stack trace
```

```
> sessionInfo()
R version 4.1.2 (2021-11-01)
Running under: Windows 10 x64 (build 22000)

Matrix products: default

locale:
[1] LC_COLLATE=German_Germany.1252  LC_CTYPE=German_Germany.1252
[3] LC_MONETARY=German_Germany.1252 LC_NUMERIC=C
[5] LC_TIME=German_Germany.1252

attached base packages:

other attached packages:
[1] data.table_1.14.2

[1] compiler_4.1.2
```

ben-schwen:
Closes #5452

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -552,6 +552,8 @@
 
 53. `as.data.frame(DT, row.names=)` no longer silently ignores `row.names`, [#5319](https://github.com/Rdatatable/data.table/issues/5319). Thanks to @dereckdemezquita for the fix and PR, and @ben-schwen for guidance.
 
+54. `rbindlist(l, use.names=TRUE)` can now handle different encodings for the column names, [#5452](https://github.com/Rdatatable/data.table/issues/5452). Thanks to @MEO265 for the report, and Benjamin Schwendinger for the fix.
+
 ## NOTES
 
 1. New feature 29 in v1.12.4 (Oct 2019) introduced zero-copy coercion. Our thinking is that requiring you to get the type right in the case of `0` (type double) vs `0L` (type integer) is too inconvenient for you the user. So such coercions happen in `data.table` automatically without warning. Thanks to zero-copy coercion there is no speed penalty, even when calling `set()` many times in a loop, so there's no speed penalty to warn you about either. However, we believe that assigning a character value such as `"2"` into an integer column is more likely to be a user mistake that you would like to be warned about. The type difference (character vs integer) may be the only clue that you have selected the wrong column, or typed the wrong variable to be assigned to that column. For this reason we view character to numeric-like coercion differently and will warn about it. If it is correct, then the warning is intended to nudge you to wrap the RHS with `as.<type>()` so that it is clear to readers of your code that a coercion from character to that type is intended. For example :
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18814,3 +18814,14 @@ test(2238.6, "a" %notin% integer(), TRUE)
 test(2238.7, "a" %notin% NULL, TRUE)
 test(2238.8, NA %notin% 1:5, TRUE)
 test(2238.9, NA %notin% c(1:5, NA), FALSE)
+
+# rbindlist(l, use.names=TRUE) should handle different colnames encodings #5452
+x = data.table(a = 1, b = 2, c = 3)
+y = data.table(x = 4, y = 5, z = 6)
+setnames(x , c("Ä", "Ö", "Ü"))
+setnames(y , iconv(c("Ö", "Ü", "Ä"), from = "UTF-8", to = "latin1"))
+test(2239.1, rbindlist(list(x,y), use.names=TRUE), data.table(Ä = c(1,6), Ö=c(2,4), Ü=c(3,5)))
+test(2239.2, rbindlist(list(y,x), use.names=TRUE), data.table(Ö=c(4,2), Ü=c(5,3), Ä = c(6,1)))
+y[, Ä := NULL]
+test(2239.3, rbindlist(list(x,y), use.names=TRUE, fill=TRUE), data.table(Ä = c(1,NA), Ö=c(2,4), Ü=c(3,5)))
+test(2239.4, rbindlist(list(y,x), use.names=TRUE, fill=TRUE), data.table(Ö=c(4,2), Ü=c(5,3), Ä = c(NA,1)))
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -78,7 +78,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
       if (!length(cn)) continue;
       const SEXP *cnp = STRING_PTR(cn);
       for (int j=0; j<thisncol; j++) {
-        SEXP s = cnp[j];
+        SEXP s = ENC2UTF8(cnp[j]); // ignore encodings for use.names #5452
         if (TRUELENGTH(s)<0) continue;  // seen this name before
         if (TRUELENGTH(s)>0) savetl(s);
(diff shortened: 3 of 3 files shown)


MichaelChirico:
Could you write out in the PR description what are the three call sites being updated? Is there any possibility of using a helper instead to avoid relying on updating 3 places to maintain similar changes in the future? Also please ensure there are tests for each of the 3 call sites (I guess `codecov` would be telling the truth here, but it's currently broken).

IIUC, UTF-8 was chosen as the "winning" encoding, it seems like even if none of the inputs are UTF-8. If that's right, I _think_ it is fine, though it might be a breaking change, right? Please describe the behavior more explicitly in the NEWS item, and perhaps in `?rbindlist`, and possibly write some new tests.

NEWS.md
```diff
@@ -12,6 +12,8 @@
 
 2. `fwrite()` respects `dec=','` for timestamp columns (`POSIXct` or `nanotime`) with sub-second accuracy, [#6446](https://github.com/Rdatatable/data.table/issues/6446). Thanks @kav2k for pointing out the inconsistency and @MichaelChirico for the PR.
 
+3. `rbindlist(l, use.names=TRUE)` can now handle different encodings for the column names, [#5452](https://github.com/Rdatatable/data.table/issues/5452). Thanks to @MEO265 for the report, and Benjamin Schwendinger for the fix.
```
MichaelChirico, line 15:
```suggestion
3. `rbindlist(l, use.names=TRUE)` can now handle different encodings for the column names in different entries of `l`, [#5452](https://github.com/Rdatatable/data.table/issues/5452). Thanks to @MEO265 for the report, and Benjamin Schwendinger for the fix.
```

inst/tests/tests.Rraw
```diff
@@ -19063,3 +19063,14 @@ test(2280.3, foo(), error="Internal error in foo: broken")
 # fwrite respects dec=',' for sub-second timestamps, #6446
 test(2281.1, fwrite(data.table(a=.POSIXct(0.001)), dec=',', sep=';'), output="1970-01-01T00:00:00,001Z")
 test(2281.2, fwrite(data.table(a=.POSIXct(0.0001)), dec=',', sep=';'), output="1970-01-01T00:00:00,000100Z")
+
+# rbindlist(l, use.names=TRUE) should handle different colnames encodings #5452
+x = data.table(a = 1, b = 2, c = 3)
+y = data.table(x = 4, y = 5, z = 6)
+setnames(x , c("\u00e4", "\u00f6", "\u00fc"))
+setnames(y , iconv(c("\u00f6", "\u00fc", "\u00e4"), from = "UTF-8", to = "latin1"))
+test(2282.1, rbindlist(list(x,y), use.names=TRUE), data.table("\u00e4"=c(1,6), "\u00f6"=c(2,4), "\u00fc"=c(3,5)))
```
MichaelChirico, line 19072: Should we document that the output will _always_ have UTF-8 encoded names? Even if all inputs are `latin1`-encoded?

inst/tests/tests.Rraw
```diff
@@ -19063,3 +19063,14 @@ test(2280.3, foo(), error="Internal error in foo: broken")
 # fwrite respects dec=',' for sub-second timestamps, #6446
 test(2281.1, fwrite(data.table(a=.POSIXct(0.001)), dec=',', sep=';'), output="1970-01-01T00:00:00,001Z")
 test(2281.2, fwrite(data.table(a=.POSIXct(0.0001)), dec=',', sep=';'), output="1970-01-01T00:00:00,000100Z")
+
+# rbindlist(l, use.names=TRUE) should handle different colnames encodings #5452
+x = data.table(a = 1, b = 2, c = 3)
+y = data.table(x = 4, y = 5, z = 6)
+setnames(x , c("\u00e4", "\u00f6", "\u00fc"))
```
MichaelChirico, line 19070:
```suggestion
# a-umlaut, o-umlaut, u-umlaut
setnames(x , c("\u00e4", "\u00f6", "\u00fc"))
```

inst/tests/tests.Rraw
```diff
@@ -19063,3 +19063,14 @@ test(2280.3, foo(), error="Internal error in foo: broken")
...
+setnames(y , iconv(c("\u00f6", "\u00fc", "\u00e4"), from = "UTF-8", to = "latin1"))
+test(2282.1, rbindlist(list(x,y), use.names=TRUE), data.table("\u00e4"=c(1,6), "\u00f6"=c(2,4), "\u00fc"=c(3,5)))
+test(2282.2, rbindlist(list(y,x), use.names=TRUE), data.table("\u00f6"=c(4,2), "\u00fc"=c(5,3), "\u00e4"=c(6,1)))
+set(y, j="\u00e4", value=NULL)
+test(2282.3, rbindlist(list(x,y), use.names=TRUE, fill=TRUE), data.table("\u00e4"=c(1,NA), "\u00f6"=c(2,4), "\u00fc"=c(3,5)))
+test(2282.4, rbindlist(list(y,x), use.names=TRUE, fill=TRUE), data.table("\u00f6"=c(4,2),  "\u00fc"=c(5,3), "\u00e4"=c(NA,1)))
```
MichaelChirico, line 19076: Maybe add a test for `use.names=FALSE` too?

src/rbindlist.c
```diff
@@ -82,7 +82,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
       if (!length(cn)) continue;
       const SEXP *cnp = STRING_PTR_RO(cn);
       for (int j=0; j<thisncol; j++) {
-        SEXP s = cnp[j];
+        SEXP s = ENC2UTF8(cnp[j]); // convert different encodings for use.names #5452
```
MichaelChirico, line 85: I've lost my familiarity with all of `rbindlist()`, why do we need to do `ENC2UTF8()` 3 times?

MichaelChirico:
Saw my earlier comment. At least I'm consistent :)

inst/tests/tests.Rraw
ben-schwen: If `use.names = FALSE` we are executing a different branch of `rbindlist` since all these changes are the `usenames==TRUE || usenames==NA_LOGICAL` branch

inst/tests/tests.Rraw
ben-schwen: We actually do not convert the column names encodings but only convert for the comparisons. So the output will have the encoding of the first element in the list. This is consistent of base which also takes the encoding of the column names `x` for encoding the names of `rbind(x,y)`

inst/tests/tests.Rraw
ben-schwen: Done

ben-schwen:
Updated the description and also the code comments!

MichaelChirico (approved):

PR #5462, opened 2022-09-14
shift on matrix: news and improve error

### LINKED ISSUE #5287: 1.14.3 `shift` no longer works on matrix type
`shift` used to work on matrix columns, but now generates and error. i dont see anything in the NEWS. is this intentional as its a breaking change?

**works in 1.14.0**

```R
sessionInfo()
R version 4.1.2 (2021-11-01)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:

[1] compiler_4.1.2    tools_4.1.2       data.table_1.14.0 checkpoint_1.0.0 
> data.table::shift(matrix(1:10, ncol = 1))
 [1] NA  1  2  3  4  5  6  7  8  9
```

**not working under 1.14.3**

```R
 sessionInfo()
R version 4.1.2 (2021-11-01)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:

[1] compiler_4.1.2    tools_4.1.2       data.table_1.14.3 checkpoint_1.0.0 
> data.table::shift(matrix(1:10, ncol = 1))
Error in data.table::shift(matrix(1:10, ncol = 1)) : 
  'as' must not be matrix or array
```

jangorecki:
closes #5287


src/utils.c
```diff
@@ -348,7 +348,7 @@ SEXP coerceAs(SEXP x, SEXP as, SEXP copyArg) {
   if (!isNull(getAttrib(x, R_DimSymbol)))
     error(_("'x' must not be matrix or array"));
   if (!isNull(getAttrib(as, R_DimSymbol)))
-    error(_("'as' must not be matrix or array"));
+    error(_("input must not be matrix or array"));
```
MichaelChirico, line 351: If users are hit by this, maybe they'd like a more actionable error. If I do want to `shift()` a matrix, how?

maybe suggest wrapping with `data.frame()` or `data.table()` to convert to a column-wise list and `shift()` that?

MichaelChirico: Oh I see.

In that case it is a reason to error for this problem earlier inside shift directly rather than re-use the error message, WDYT?

jangorecki: Yes, that actually makes sense, from the point of providing more tailored error message. Other than that we are just duplicating the logic. Then other functions using coerceAs could raise own errors as well. Possibly some of them already doing it.

ben-schwen:
LGTM

PR #5463, opened 2022-09-14
memrecycle no snprintf overhead

### LINKED ISSUE #5366: significantly slower performance time based rolling 1.14.3
I am a multi year data.table user and today I've seen 3 times slower performance of data.table time based rolling aggregation compared with pandas rolling aggregation. Please advise if the behavior is expected and if here is a correct place to post reproducible examples. Thanks. 

for data.table time based rolling I use non-equip join syntax. Pandas rolling I use the rolling() function.

### LINKED ISSUE #5371: Memrecycle Performance Regression
Identified by @tlapak in #5366, there is a performance regression in 14.3 introduced by #4491 @jangorecki . I'm reporting here so we can separate the Pandas rolling time performance comparison from the regression issue.

The regression was specifically introduced by this commit: https://github.com/Rdatatable/data.table/pull/4491/commits/e793f53466d99f86e70fc2611b708ae8c601a451 . Basically, `snprintf()` performs poorly in spite of the better looking code. See also https://stackoverflow.com/questions/64600389/why-is-calling-snprintf-so-slow.

For performance, Stackoverflow seems to suggest `strcpy()`. I propose either rolling back the changes or using a helper function during the calls so that the string is only made when needed. Without any changes, users would see performance issues which would be most noticeable when there are many groups in non-Gforce j calls.

```r 
set.seed(123L)
n = 1e4L
dt = data.table(id = seq_len(n),
                val = rnorm(n))

### 1.14.2
system.time(dt[, .(vs = (sum(val))), by = .(id)])

##    user  system elapsed 
##    0.04    0.00    0.05

### after commit e793f53
system.time(dt[, .(vs = (sum(val))), by = .(id)])

##    user  system elapsed 
##    1.22    0.03    1.23 
```

### LINKED ISSUE #5424: Performance regression with `.N` and `:=`
On the latest Github master `dt[, N = .N, by = id]` became much slower than on the latest CRAN version. The output below is from an M1 Mac but the issue can be reproduced on x64 Linux and Windows as well. I included tests for all optimization levels. I haven't found similar issues with other optimized functions.

# CRAN (1.14.2)

``` r
library(data.table)

N <- 1e8
n <- 1e6

set.seed(1L)
dt <- data.table(g = sample(seq_len(n), N, TRUE),
                 x = runif(N),
                 key = "g")
#> forder.c received 100000000 rows and 2 columns
#> forder took 2.256 sec
#> reorder took 1.934s elapsed (4.338s cpu)

options(datatable.optimize = 0L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Detected that j uses these columns: <none> 
#> Finding groups using uniqlist on key ... 0.115s elapsed (0.097s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.004s elapsed (0.003s cpu) 
#> All optimizations are turned off
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.310s for 1000000 groups
#>   eval(j) took 0.300s for 1000000 calls
#> 1.412s elapsed (0.630s cpu)

options(datatable.optimize = 1L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Detected that j uses these columns: <none> 
#> Finding groups using uniqlist on key ... 0.112s elapsed (0.099s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.003s elapsed (0.003s cpu) 
#> lapply optimization is on, j unchanged as '.N'
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.312s for 1000000 groups
#>   eval(j) took 0.300s for 1000000 calls
#> 1.767s elapsed (0.610s cpu)

options(datatable.optimize = 2L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Detected that j uses these columns: <none> 
#> Finding groups using uniqlist on key ... 0.104s elapsed (0.096s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.004s elapsed (0.003s cpu) 
#> lapply optimization is on, j unchanged as '.N'
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.311s for 1000000 groups
#>   eval(j) took 0.298s for 1000000 calls
#> 1.442s elapsed (0.613s cpu)

options(datatable.optimize = 3L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Detected that j uses these columns: <none> 
#> Finding groups using uniqlist on key ... 0.114s elapsed (0.098s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.003s elapsed (0.003s cpu) 
#> lapply optimization is on, j unchanged as '.N'
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.308s for 1000000 groups
#>   eval(j) took 0.296s for 1000000 calls
#> 1.364s elapsed (0.595s cpu)
```

<sup>Created on 2022-07-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessionInfo()
#> R version 4.2.1 (2022-06-23)
#> Platform: aarch64-apple-darwin20 (64-bit)
#> Running under: macOS Monterey 12.5
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.14.2
#> 
#>  [1] digest_0.6.29   withr_2.5.0     magrittr_2.0.3  reprex_2.0.1   
#>  [5] evaluate_0.15   highr_0.9       stringi_1.7.8   rlang_1.0.4    
#>  [9] cli_3.3.0       fs_1.5.2        rmarkdown_2.14  tools_4.2.1    
#> [13] stringr_1.4.0   glue_1.6.2      xfun_0.31       yaml_2.3.5     
#> [17] fastmap_1.1.0   compiler_4.2.1  htmltools_0.5.3 knitr_1.39
```

</details>

# Github (master)

``` r
library(data.table)

N <- 1e8
n <- 1e6

set.seed(1L)
dt <- data.table(g = sample(seq_len(n), N, TRUE),
                 x = runif(N),
                 key = "g")
#> forder.c received 100000000 rows and 2 columns
#> forder took 2.672 sec
#> reorder took 1.965s elapsed (4.403s cpu)

options(datatable.optimize = 0L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Argument 'by' after substitute: g
#> Detected that j uses these columns: <none>
#> Finding groups using uniqlist on key ... 0.174s elapsed (0.105s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.014s elapsed (0.003s cpu) 
#> All optimizations are turned off
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.260s for 1000000 groups
#>   eval(j) took 0.249s for 1000000 calls
#> 2.103s elapsed (1.489s cpu)

options(datatable.optimize = 1L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Argument 'by' after substitute: g
#> Detected that j uses these columns: <none>
#> Finding groups using uniqlist on key ... 0.116s elapsed (0.099s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.004s elapsed (0.003s cpu) 
#> lapply optimization is on, j unchanged as '.N'
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.263s for 1000000 groups
#>   eval(j) took 0.249s for 1000000 calls
#> 2.294s elapsed (1.479s cpu)

options(datatable.optimize = 2L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Argument 'by' after substitute: g
#> Detected that j uses these columns: <none>
#> Finding groups using uniqlist on key ... 0.096s elapsed (0.094s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.004s elapsed (0.003s cpu) 
#> lapply optimization is on, j unchanged as '.N'
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.262s for 1000000 groups
#>   eval(j) took 0.249s for 1000000 calls
#> 2.096s elapsed (1.468s cpu)

options(datatable.optimize = 3L)
dt_mod <- copy(dt)
dt_mod[, N := .N, by = g]
#> Argument 'by' after substitute: g
#> Detected that j uses these columns: <none>
#> Finding groups using uniqlist on key ... 0.114s elapsed (0.098s cpu) 
#> Finding group sizes from the positions (can be avoided to save RAM) ... 0.003s elapsed (0.004s cpu) 
#> lapply optimization is on, j unchanged as '.N'
#> Making each group and running j (GForce FALSE) ... 
#>   memcpy contiguous groups took 0.261s for 1000000 groups
#>   eval(j) took 0.247s for 1000000 calls
#> 2.048s elapsed (1.464s cpu)
```

<sup>Created on 2022-07-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessionInfo()
#> R version 4.2.1 (2022-06-23)
#> Platform: aarch64-apple-darwin20 (64-bit)
#> Running under: macOS Monterey 12.5
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.14.3
#> 
#>  [1] digest_0.6.29   withr_2.5.0     magrittr_2.0.3  reprex_2.0.1   
#>  [5] evaluate_0.15   highr_0.9       stringi_1.7.8   rlang_1.0.4    
#>  [9] cli_3.3.0       fs_1.5.2        rmarkdown_2.14  tools_4.2.1    
#> [13] stringr_1.4.0   glue_1.6.2      xfun_0.31       yaml_2.3.5     
#> [17] fastmap_1.1.0   compiler_4.2.1  htmltools_0.5.3 knitr_1.39
```

</details>

jangorecki:
closes #5424, closes #5366, closes #5371
```
issue:  1.14.2 vs  master vs  branch
#5424:  0.939s vs  1.725s vs  0.958s
#5366: 14.786s vs 26.515s vs 14.621s
#5371:  0.058s vs  0.050s vs  0.025s
```
code:
```r
library(data.table)
set.seed(123L)
n = 1e4L
dt = data.table(id = seq_len(n),
                val = rnorm(n))

system.time(
  dt[, .(vs = (sum(val))), by = .(id)]
)


library(data.table)
N <- 1e8
n <- 1e6
set.seed(1L)
dt <- data.table(g = sample(seq_len(n), N, TRUE),
                 x = runif(N),
                 key = "g")
dt_mod <- copy(dt)
system.time(
  dt_mod[, N := .N, by = g]
)


library(data.table)
set.seed(1)
n <- 1e6
d1 <- abs(rnorm(n, sd = 4))
d2 <- as.integer(cumsum(d1))
tm <- as.POSIXct("2020-01-01 09:30:00") + d2
nIds <- 3
tmCol <- rep(tm, nIds)
idCol <- rep(c("a", "b", "c"), n)
f1 <- function() {
    dt <- data.table(tm = tmCol, v = 1, id = idCol)
    dt[, tm1 := tm - 40]
    dt[, tm2 := tm]
    dt[, rowNum := .I]
    dt[dt, .(vs = sum(v)), on = .(id, rowNum <= rowNum, tm >= tm1, tm < tm2), by = .EACHI]
}
system.time(
  f1()
)
```


ColeMiller1 (approved):
Thanks, Jan.  Simple yet effective approach to rollback the performance issues.

ben-schwen:
Why do we not need targetDesc in the `CHECK_RANGE` macro?https://github.com/Rdatatable/data.table/blob/ade6e57984f9e7f898a638d2049039cd886f2369/src/assign.c#L867-L881

or is created in every possible branch before?

jangorecki:
12 lines above is code which is still valid for the macro. Macro doesn't have own scope, it just substitutes code in existing scope, so the code 12 lines above is still used there.

MichaelChirico:
great stuff!

can we blow up the examples for the fixes to #5424 and #5371? I worry about benchmarks lasting <1s

src/assign.c
```diff
@@ -845,6 +851,8 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
     warning(_("Coercing 'list' RHS to '%s' to match the type of %s."), type2char(TYPEOF(target)), targetDesc);
     source = PROTECT(coerceVector(source, TYPEOF(target))); protecti++;
   } else if ((TYPEOF(target)!=TYPEOF(source) || targetIsI64!=sourceIsI64) && !isNewList(target)) {
+    static char targetDesc[501];
+    snprintf(targetDesc, 500, colnum==0 ? _("target vector") : _("column %d named '%s'"), colnum, colname);
```
MichaelChirico, line 855: shouldn't this one be declared inside the `GetVerbose()>=3` branch?

MichaelChirico:
Hey Jan, IIUC the issue is we are declaring a string buffer inside `memrecycle`, but it's inefficient because `memrecycle` might get called many (100s?) of times, so the fix is to only declare the buffer just as it's needed. This makes a bit of code duplication because we can't just re-use the one declaration of the buffer, but increases efficiency since the buffer is not typically needed.

Is that right? It would be helpful to summarize your fix in the PR description, this will especially make it easier when referring back to the PR in the future.

src/assign.c
```diff
@@ -730,6 +728,8 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
           for (int i=0; i<slen; ++i) {
             const int val = sd[i+soff];
             if ((val<1 && val!=NA_INTEGER) || val>nlevel) {
+              static char targetDesc[501];  // from 1.14.1 coerceAs reuses memrecycle for a target vector, PR#4491
```
MichaelChirico, line 731: maybe wrap this into a helper function to save some duplication?

jangorecki: Could be, but I am fine either way

src/assign.c
jangorecki: I think macro below needs it as well

jangorecki:
Yes, your are correct. Personally I could even remove this declaration all together by keeping verbose message more obscure.

mattdowle:
Great that you identified this, all the testing and benchmarks fit with the fix.

However, regarding the thought that it's the allocation?

> Is that right?
> Yes, your are correct.

No that's wrong. The allocation is declared `static`. Please look up what that is. We do that quite a lot in data.table and needs to be understood.  To copy that down and repeat conveys to readers of the code that the `static` keyword isn't understood so that needs to be put back at the top just once.  The slow down will all be in the `snprintf` call itself.

> Personally I could even remove this declaration all together by keeping verbose message more obscure.

Maybe. Or just repeat the ternary in each of the messages, or wrap that ternary in a macro. With the ternary I think I was trying to match previous messages maybe. Leaving implementation details aside do we agree that the `column %d named '%s'` part is better and more useful to provide to user when we can? It would be a shame to lose that part.

mattdowle (changes requested):
Great that you identified this, all the testing and benchmarks fit with the fix.

However, regarding the thought that it's the allocation?

> Is that right?
> Yes, your are correct.

No that's wrong. The allocation is declared `static`. Please look up what that is. We do that quite a lot in data.table and needs to be understood.  To copy that down and repeat conveys to readers of the code that the `static` keyword isn't understood. So that needs to be put back at the top just once. The slow down will all be in the `snprintf` call itself.

> Personally I could even remove this declaration all together by keeping verbose message more obscure.

Yes remove the declaration and `snprintf`, agree, but not make the message more obscure. Could repeat the ternary in each of the messages, or wrap that ternary in a macro. With the ternary I think I was trying to match previous messages maybe. Leaving implementation details aside do we agree that the `column %d named '%s'` part is better and more useful to provide to user when we can? It would be a shame to lose that part.

mattdowle (commented):
How does this `targetDesc` get populated now? If I'm right that it doesn't, does that mean we're missing a test or the test is insufficient?
[ GH won't give me "+" to comment on that because it's in a block that hasn't changed. This comment is supposed to be attached to the `targetDesc` inside the CHECK_RANGE definition. ]

mattdowle:
Btw, the commit `static declares once` would be better as `declare static once`. The commit messages are suppose to describe what the commit does and if I look back in future at that, I won't know what that means. Perhaps in your mind you're describing a fact in the commit message, whereas I'm looking for a description of what the commit does. Or maybe it was just English.

src/assign.c
mattdowle: but the macro only needs it inside its `if(COND)`. So isn't this `snprintf()` indeed wasteful?  This `snprintf` could be inside this `GetVerbose()>=3` and then appear inside the CHECK_RANGE macro as well inside its `if(COND)`. When I didn't see the `snprintf` inside the `CHECK_RANGE` that's why I wrote https://github.com/Rdatatable/data.table/pull/5463#pullrequestreview-1140163707. So it would read better if `targetDesc` was populated right there in `CHECK_RANGE` next to where it is used?

mattdowle:
I see now how that `targetDesc` in CHECK_RANGE gets populated, thanks. But now I reopened comment above and added further up there. https://github.com/Rdatatable/data.table/pull/5463#discussion_r995209117.  Adding this new comment to be sure you see the reopened comment up there.

mattdowle:
Assuming I'm right this time that that `snprintf` should be moved inside those `if()`s, then to avoid confusion in the future and to avoid the need to repeat the `snprintf` and in the correct way too, wherever `targetDesc` is passed as an argument currently to `error()`/`warning()`/`Rprintf()`/`snprintf()`, it could be a function call `targetDesc()` instead. That function call would populate the `targetDesc` storage and return `targetDesc`. That way there can be no doubt that the `snprintf()` to make `targetDesc` is only computed when needed, and is also the same and consistent across all usage.

jangorecki:
I pushed improvement for now to address your comment about `if` condition in `CHECK_RANGE`.
`targetDesc()` is a good idea, but till Monday I won't have time to touch that.

mattdowle:
Thanks Jan. I did the `targetDesc()`. All good?

mattdowle (approved):

PR #5467, opened 2022-09-16
Friendlier error in assignment with trailing comma

MichaelChirico:
e.g. ``DT[, `:=`(a = 1, b = 2,)``.

We could also allow this to "just work", as it would in Python and some SQL flavors. Allowing a trailing comma makes code more programmer-friendly since we can easily re-order/refactor entries with simple copy-paste/delete lines without regard to whether this happens to be the last/second-to-last entry.


MichaelChirico:
Ready for review.

NEWS.md
```diff
@@ -22,6 +22,8 @@
 
 3. data.table now depends on R 3.2.0 (2015) instead of 3.1.0 (2014). 1.17.0 will likely move to R 3.3.0 (2016). Recent versions of R have good features that we would gradually like to incorporate, and we see next to no usage of these very old versions of R.
 
+4. Erroneous assignment calls in `[` with a trailing comma get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
```
jangorecki, line 25: Showing one example here seems to be useful

jangorecki (approved):

PR #5468, opened 2022-09-16
rbindlist segfault for fill=TRUE and usenames=FALSE

### LINKED ISSUE #5444: rbindlist() crashing R session
```R
library(data.table)
# data.table 1.14.3 IN DEVELOPMENT built 2022-07-20 18:26:12 UTC

X = list(1, c(4.4, 3.5))
rbindlist(lapply(X, as.list), fill = TRUE, use.names = FALSE)
```

Crashes R (version 4.2.1) on Windows 10 both using Rstudio and a `-vanilla` session.

ben-schwen:
Closes #5444 

- [x] fix
- [x] tests
- [x] update news

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -14695,6 +14695,9 @@ test(2003.4, rbindlist(list(data.table(a=1:2,c=5:6), data.table(b=3:4)), fill=TR
              data.table(a=c(1:4), c=INT(5,6,NA,NA)))
 test(2003.5, rbindlist(list(data.table(a=1:2), data.table(b=3:4, c=5:6)), fill=TRUE, use.names=FALSE),
              data.table(a=c(1:4), c=INT(NA,NA,5,6)))
+# rbindlist segfault with fill=TRUE and usenames=FALSE #5444
+test(2003.6, rbindlist(list(list(1), list(2,3)), fill=TRUE, use.names=FALSE), data.table(c(1,2), c(NA, 3)))
+test(2003.7, rbindlist(list(list(1), list(2,factor(3))), fill=TRUE, use.names=FALSE), data.table(c(1,2), factor(c(NA, 3))))
 
 # chmatch coverage for two different non-ascii encodings matching; issues mentioned in comments in chmatch.c #69 #2538 #111
 x1 = "fa\xE7ile"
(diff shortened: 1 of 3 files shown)


ben-schwen:
I changed the behavior of test 2003.5 introduced in #5263 since for me it seems more natural if the `data.table` gains the column name of the table with the most columns and not a new one.

This is no breaking change since `fill=TRUE` with `usenames=FALSE` is only available in current dev version 1.14.3

jangorecki:
I think the names of columns in such case shoybe based on the order of occurrence. Unit test for that is pretty simple, doesn't cover for example case of 3 tables. But the results of the current test looks good (names are based on the order of occurrence of columns), better than before.

MichaelChirico:
> I changed the behavior of test 2003.5 introduced in #5263 since for me it seems more natural if the `data.table` gains the column name of the table with the most columns and not a new one.
> 
> This is no breaking change since `fill=TRUE` with `usenames=FALSE` is only available in current dev version 1.14.3

now that there's new behavior, I think that does warrant a NEWS edit (either adding to the existing item or getting its own)

inst/tests/tests.Rraw
```diff
@@ -18814,3 +18814,7 @@ test(2238.6, "a" %notin% integer(), TRUE)
 test(2238.7, "a" %notin% NULL, TRUE)
 test(2238.8, NA %notin% 1:5, TRUE)
 test(2238.9, NA %notin% c(1:5, NA), FALSE)
+
+# rbindlist segfault with fill=TRUE and usenames=FALSE #5444
```
MichaelChirico, line 18818: if this is fixing an issue caused by a change in dev, maybe the tests should be attached to the same suite as the original PR?

src/rbindlist.c
```diff
@@ -282,7 +282,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
     for (int i=0; i<LENGTH(l); ++i) {
       SEXP li = VECTOR_ELT(l, i);
       if (!length(li)) continue;
-      int w = usenames ? colMap[i*ncol + j] : j;  // colMap tells us which item to fetch for each of the final result columns, so we can stack column-by-column
+      int w = usenames ? colMap[i*ncol + j] : (j<length(li) ? j : -1);  // colMap tells us which item to fetch for each of the final result columns, so we can stack column-by-column // check if j exceeds length for fill=TRUE and usenames=FALSE #5444
```
MichaelChirico, line 285: nit: why the second comment marker (`//`)?

jangorecki: I looked at that as well. If comments are unrelated then better to have second // so it's clear for whoever reads that it's completely new comment. Otherwise comments could be merged into single one.

ben-schwen: Comments are unrelated, why I went for the second //

inst/tests/tests.Rraw
ben-schwen: Moved them accordingly

src/rbindlist.c
MichaelChirico: makes sense... i think elsewhere that's where we'd see

```c
foo(); // comment
       // another comment
```

side note -- for me personally, that's why I usually prefer anchoring the comment to the previous line rather than end-of-line, makes it easier: continuations get a hanging indent, new comments restart indentation:

```c
// this is a comment that
//   spills onto line 2
// this is an unrelated comment
```

jangorecki (approved):

PR #5474, opened 2022-09-28
fread read compressed .bgz

### LINKED ISSUE #5461: Add .bgz file decompression to data.table::fread() (compatible with .gz)
Currently, .bgz files are read as plain text, which fails due to invalid characters. [.bgz files are compatible with gunzip](http://www.htslib.org/doc/bgzip.html), and have the same data header (0x1F 0x8B). renaming *.bgz files to *.gz files allows them to be decompressed normally.

Adding .bgz to the list of files that can be decompressed by data.table::fread shouldn't require anything other than R.utils. I think adding `".bgz"` to the vector in this line:

https://github.com/Rdatatable/data.table/blob/c4a2085e35689a108d67dacb2f8261e4964d7e12/R/fread.R#L121

And checking for `w<=2` on this line

https://github.com/Rdatatable/data.table/blob/c4a2085e35689a108d67dacb2f8261e4964d7e12/R/fread.R#L124

Would allow fread to decompress .bgz files automatically. However, I haven't tested this.

ben-schwen:
Closes #5461


NEWS.md
```diff
@@ -42,6 +42,8 @@
 
 13. `dcast`gains `value.var.in.dots`, `value.var.in.LHSdots` and `value.var.in.RHSdots` arguments, [#5824](https://github.com/Rdatatable/data.table/issues/5824). This allows the `value.var` variable(s) in `dcast` to be represented by `...` in the formula (if not otherwise mentioned). Thanks to @iago-pssjd for the report and PR.
 
+14. `fread` now supports to read a compressed `.bgz` directly, [#5461](https://github.com/Rdatatable/data.table/issues/5461). Thanks to @TMRHarrison for the request with proposed fix, and Benjamin Schwendinger for the PR.
```
MichaelChirico, line 45:
```suggestion
14. `fread` loads `.bgz` files directly, [#5461](https://github.com/Rdatatable/data.table/issues/5461). Thanks to @TMRHarrison for the request with proposed fix, and Benjamin Schwendinger for the PR.
```

"now supports" felt like there was some new functionality added, but as noted, the change is much simpler than that -- a purely technical change from realizing reading `bgz` can re-used the same code as reading `gz`.

MichaelChirico (approved):
Awesome!

PR #5476, opened 2022-10-06
gcc-12 always-false in fread.c

mattdowle:
With gcc 12.1 as displayed on CRAN currently
```
Version: 1.14.2
Flags: --no-tests
Check: whether package can be installed
Result: WARN
    Found the following significant warnings:
     fread.c:1278:38: warning: the comparison will always evaluate as ‘false’ for the pointer operand
                               in ‘nastr + 8’ must not be NULL [-Waddress]
Flavor: r-devel-linux-x86_64-debian-gcc

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -1284,8 +1284,6 @@ int freadMain(freadMainArgs _args) {
   while (*nastr) {
     if (**nastr == '\0') {
       blank_is_a_NAstring = true;
-      // if blank is the only one, as is the default, clear NAstrings so that doesn't have to be checked
-      if (nastr==NAstrings && nastr+1==NULL) NAstrings=NULL;
       nastr++;
       continue;
     }
@@ -1324,6 +1322,10 @@ int freadMain(freadMainArgs _args) {
     if (args.skipString) DTPRINT(_("  skip to string = <<%s>>\n"), args.skipString);
     DTPRINT(_("  show progress = %d\n"), args.showProgress);
     DTPRINT(_("  0/1 column will be read as %s\n"), args.logical01? "boolean" : "integer");
+  }
+  if (*NAstrings==NULL ||                             // user specified that blank is to be considered "" not NA
+      (**NAstrings=='\0' && *(NAstrings+1)==NULL)) {  // blank is the only value to be considered NA, as is the default
+    NAstrings=NULL;  // clear NAstrings to save end_NA_string() dealing with these cases (blank_is_a_NAstring was set to true above)
   }
 
   stripWhite = args.stripWhite;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by jangorecki** [comment]
There was an issue and PR already to resolve this
https://github.com/Rdatatable/data.table/issues/5377

**Comment by mattdowle** [comment]
NB: turning on `-Waddress` in `cc.R` with gcc 11.2 doesn't detect it. I tried adding `-Waddress` to cc.R so that we wouldn't need gcc 12 to detect it. `-Wall` is already on in `cc.R` so I guess that `-Wall` already includes `-Waddress` but that something has improved in gcc 12 that now detects this; i.e. it's not just that `-Waddress` is now included by -Wall in gcc 12.

File: src/fread.c [hunk 1]
```diff
@@ -1284,25 +1284,22 @@ int freadMain(freadMainArgs _args) {
...
+      const char *ch = *nastr;
+      size_t nchar = strlen(ch);
+      if (isspace(ch[0]) || isspace(ch[nchar-1]))
+        STOP(_("freadMain: NAstring <<%s>> has whitespace at the beginning or end"), ch);
+      if (strcmp(ch,"T")==0    || strcmp(ch,"F")==0 ||
+          strcmp(ch,"TRUE")==0 || strcmp(ch,"FALSE")==0 ||
+          strcmp(ch,"True")==0 || strcmp(ch,"False")==0)
+        STOP(_("freadMain: NAstring <<%s>> is recognized as type boolean, this is not permitted."), ch);
+      if ((strcmp(ch,"1")==0 || strcmp(ch,"0")==0) && args.logical01)
+        STOP(_("freadMain: NAstring <<%s>> and logical01=%s, this is not permitted."), ch, args.logical01 ? "TRUE" : "FALSE");
```
MichaelChirico, line 1297: slightly confused by the templating in the string here am I missing something? in the error case logical01 is always TRUE?

src/fread.c
```diff
@@ -1284,25 +1284,22 @@ int freadMain(freadMainArgs _args) {
...
+      char *end;
+      errno = 0;
+      (void)strtod(ch, &end);  // careful not to let "" get to here (see continue above) as strtod considers "" numeric
```
MichaelChirico, line 1300: is "continue" referring to somewhere else in the file?

src/fread.c
mattdowle: good spot. agree logical01=TRUE could be in the string and the ternary isn't needed.

src/fread.c
mattdowle: indeed. good spot.

Issue #4936, opened 2021-03-16
Tests fail when digits option is set.


vlulla:
```{bash}
$ R --vanilla
R> library("data.table")
R> test.data.table()
R> # all tests pass.
R> quit()

$ R --vanilla
R> options(digits=17) # similar to python https://docs.python.org/3/tutorial/floatingpoint.html
R> library("data.table")
R> test.data.table()
Error in test.data.table() : 
  15 errors out of 10033. Search tests/tests.Rraw.bz2 for test numbers: 640, 1037.103, 1037.301, 1437.12, 1437.14, 1437.16, 1613.06, 1613.1, 1613.12, 1613.14, 1613.444, 1613.5, 1743.413, 1768, 1769.
R> quit()
```
I do this because I need `log(1000)/log(10)` to return the floating point instead of the rounded result printed by R. Compare R's output with python's `math.log(1000)/math.log(10)` to see the difference.

Output of `sessionInfo()`
```{r}
> sessionInfo()
R version 4.0.4 (2021-02-15)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux bullseye/sid

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/atlas/libblas.so.3.10.3
LAPACK: /usr/lib/x86_64-linux-gnu/atlas/liblapack.so.3.10.3

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] yaml_2.2.1        R.utils_2.10.1    R.oo_1.24.0       R.methodsS3_1.8.1
 [5] nanotime_0.3.2    xts_0.12.1        zoo_1.8-8         bit64_4.0.5      
 [9] bit_4.0.4         data.table_1.13.6

loaded via a namespace (and not attached):
[...]

Issue #4633, opened 2020-07-23
Tests failing due to setting datatable.print.keys option


vlulla:
I set the `options(datatable.print.keys=TRUE)` in my `.Rprofile`. This causes tests `1465.18` and `1872.09` to fail. Both of these  test failures can be prevented by _explicitly_ setting this option to `FALSE` before running these tests.

OfekShilon:
Fixed by https://github.com/Rdatatable/data.table/commit/22b51c510a614e9095b7cd6de57cf87aae0d8914

PR #5497, opened 2022-10-24
Fix #5285 - make tests more robust to options

### LINKED ISSUE #5285: Tests fail when run with some non-default options
```
options(warn=2)
test.data.table()
```
Would result in multiple failures. Also:
```
options(digits=17)
test.data.table()
```
Probably best to add to `oldOptions` in `test.data.table`:
```
...
 warn = 1,
 digits = 7
```

`#` `Output of sessionInfo()`
R 4.0.5, data.table 1.14.3

OfekShilon:
Fixes #5285 and #4936  (#4633 was fixed separately)


MichaelChirico:
This should wait until #5845, which might supersede it.

MichaelChirico:
Sorry I must have forgotten this PR exists, #6052 re-applied the change, so this is now obsolete. Thank you!

PR #5498, opened 2022-10-25
fix broken links in R html manuals

jangorecki:
Possibly change in R 4.2 made links to be resolved not directly but by a help/ dir which was previously not considered. This PR should also keep it and links should be resolved then.


MichaelChirico:
@jangorecki still relevant after #5690/#5632?

jangorecki:
it is relevant only if we keep GLCI.

jangorecki:
superseded by #5758

PR #5501, opened 2022-10-28
Limit deparse in name_dots to 1 line

OfekShilon:
Per @MichaelChirico 's [request](https://github.com/Rdatatable/data.table/pull/5493#pullrequestreview-1152996040), this improvement is separated to its own PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -550,6 +550,8 @@
 
 53. `as.data.frame(DT, row.names=)` no longer silently ignores `row.names`, [#5319](https://github.com/Rdatatable/data.table/issues/5319). Thanks to @dereckdemezquita for the fix and PR, and @ben-schwen for guidance.
 
+54. data.table construction uses `deparse` to try and automatically name un-named cols. The attempt is now limited to 1- line which gives substantial speedup in [some edge cases](https://github.com/Rdatatable/data.table/pull/5501). Thanks @OfekShilon for the report and fix, @michaelchirico for guidance.
+
 ## NOTES
 
 1. New feature 29 in v1.12.4 (Oct 2019) introduced zero-copy coercion. Our thinking is that requiring you to get the type right in the case of `0` (type double) vs `0L` (type integer) is too inconvenient for you the user. So such coercions happen in `data.table` automatically without warning. Thanks to zero-copy coercion there is no speed penalty, even when calling `set()` many times in a loop, so there's no speed penalty to warn you about either. However, we believe that assigning a character value such as `"2"` into an integer column is more likely to be a user mistake that you would like to be warned about. The type difference (character vs integer) may be the only clue that you have selected the wrong column, or typed the wrong variable to be assigned to that column. For this reason we view character to numeric-like coercion differently and will warn about it. If it is correct, then the warning is intended to nudge you to wrap the RHS with `as.<type>()` so that it is clear to readers of your code that a coercion from character to that type is intended. For example :
--- a/NEWS.md
+++ b/NEWS.md
@@ -550,8 +550,15 @@
 
 53. `as.data.frame(DT, row.names=)` no longer silently ignores `row.names`, [#5319](https://github.com/Rdatatable/data.table/issues/5319). Thanks to @dereckdemezquita for the fix and PR, and @ben-schwen for guidance.
 
-54. Certain flavours of data.table construction try to use `deparse` to automatically name un-named cols. The attempt is now limited to 1 line of the `deparse` input which gives substantial speedup in [some cases](https://github.com/Rdatatable/data.table/pull/5501). Thanks @OfekShilon for the report and fix, @michaelchirico for guidance.
+54. `data.table(...)` unnamed arguments are deparsed in an attempt to name the columns but when called from `do.call()`, the input data itself was deparsed taking a very long time, [#5501](https://github.com/Rdatatable/data.table/pull/5501). Many thanks to @OfekShilon for the report and fix, and @michaelchirico for guidance. Unnamed arguments to `data.table(...)` may now be faster in other cases not involving `do.call()` too; e.g. expressions spanning a lot of lines or other function call constructions that led to the data itself being deparsed.
 
+    ```R
+    DF = data.frame(a=runif(1e6), b=runif(1e6))
+    DT1 = data.table(DF)                 # 0.02s before and after
+    DT2 = do.call(data.table, list(DF))  # 3.07s before, 0.02s after
+    identical(DT1, DT2)                  # TRUE
+    ```
+    
 ## NOTES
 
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -97,7 +97,7 @@ name_dots = function(...) {
   if (any(notnamed)) {
     syms = vapply_1b(dot_sub, is.symbol)  # save the deparse() in most cases of plain symbol
     for (i in which(notnamed)) {
-      tmp = if (syms[i]) as.character(dot_sub[[i]]) else deparse(dot_sub[[i]])[1L]
+      tmp = if (syms[i]) as.character(dot_sub[[i]]) else deparse(dot_sub[[i]], nlines=1)[1L]
```
MichaelChirico, line 100: nlines=1L

MichaelChirico:
could you add a test where this matters? probably worth a NEWS item as well, this could potentially break some folks

OfekShilon:
> could you add a test where this matters? probably worth a NEWS item as well, this could potentially break some folks

I'm not aware of any case where this change makes an observable difference in behavior. The best I can suggest is a test documenting the performance improvement in an awkward call of the sort `transform` used to make:
```
t1 <- system.time(dt1 <- data.table(a=runif(1e5), b=runif(1e5)))
t2 <- system.time(dt2 <- do.call(data.table, c(list(dt1))))

test(2239.1, all.equal(dt1,dt2), TRUE)
test(2239.2, t2["elapsed"]/t1["elapsed"]<5, TRUE)   # On my system this is >30 before the fix, <1.5 after
```
Do you think this adds value?

MichaelChirico:
It matters here:

```r
asodijfoaisjdfoiajsdofijasodifjoaisdjfoiajsdoifjaosidjfoiajsdofijaosidjfoiasjdofijasdoifjaosidjfoiasjdfiuajsfuasiudfhoiauhjzxclkvjbhlszkjxhcliuasdhfgoiuahsdoufhlijxchlkzjhvkjhaljshdfiluhcubvjdfbvjkahsdfliuhawpioedhflaijsdh = function() 1L
data.table(c(
 
 asodijfoaisjdfoiajsdofijasodifjoaisdjfoiajsdoifjaosidjfoiajsdofijaosidjfoiasjdofijasdoifjaosidjfoiasjdfiuajsfuasiudfhoiauhjzxclkvjbhlszkjxhcliuasdhfgoiuahsdoufhlijxchlkzjhvkjhaljshdfiluhcubvjdfbvjkahsdfliuhawpioedhflaijsdh(),
 
 asodijfoaisjdfoiajsdofijasodifjoaisdjfoiajsdoifjaosidjfoiajsdofijaosidjfoiasjdofijasdoifjaosidjfoiasjdfiuajsfuasiudfhoiauhjzxclkvjbhlszkjxhcliuasdhfgoiuahsdoufhlijxchlkzjhvkjhaljshdfiluhcubvjdfbvjkahsdfliuhawpioedhflaijsdh(),
 
 asodijfoaisjdfoiajsdofijasodifjoaisdjfoiajsdoifjaosidjfoiajsdofijaosidjfoiasjdofijasdoifjaosidjfoiasjdfiuajsfuasiudfhoiauhjzxclkvjbhlszkjxhcliuasdhfgoiuahsdoufhlijxchlkzjhvkjhaljshdfiluhcubvjdfbvjkahsdfliuhawpioedhflaijsdh(),
 
 asodijfoaisjdfoiajsdofijasodifjoaisdjfoiajsdoifjaosidjfoiajsdofijaosidjfoiasjdofijasdoifjaosidjfoiasjdfiuajsfuasiudfhoiauhjzxclkvjbhlszkjxhcliuasdhfgoiuahsdoufhlijxchlkzjhvkjhaljshdfiluhcubvjdfbvjkahsdfliuhawpioedhflaijsdh()
))
```

There should be a more minimal version of this but hopefully illustrates the poitn

MichaelChirico:
> Do you think this adds value?

We shy away from timing tests in our main suite -- you never know when some system interference / sun spots can cause things to go haywire and now we have to deal with the threat of CRAN removal...

OTOH, there is this non-CRAN suite, couldn't hurt to throw a test like that in there:

https://github.com/Rdatatable/data.table/blob/master/inst/tests/benchmark.Rraw

I'd suggest a threshold of maybe 2 or 5 to be conservative, too, .2 vs .5 seems to tight and likely not to replicate across platforms / with high statistical certainty

OfekShilon:
> There should be a more minimal version of this but hopefully illustrates the point

Nice! `deparse` has a default `width.cutoff = 60L`, so these can be trimmed to 61-char names. 
I'll do that - and add a conservative timing test to benchmark.Rraw - tomorrow. Thanks.

OfekShilon:
@MichaelChirico I checked now and think your example also doesn't have an observable impact on behavior. With and without `nlines=1`, the generated name of the column is "V1". Notice [the line after the deparse](https://github.com/Rdatatable/data.table/blob/0895fa247afcf6b38044bd5f56c0d209691ddb31/R/utils.R#L101):
```
      tmp = if (syms[i]) as.character(dot_sub[[i]]) else deparse(dot_sub[[i]])[1L]
      if (tmp == make.names(tmp)) vnames[i]=tmp
```
`make.names` converts newlines, brackets and others to dots, and so does change pretty much anything that slips into this code path - except proper variable names - and causes the deparse to go unused. With and without nlines=1, including your example and anything I could come up with.
This is kind of good news - I probably won't be able to add a test, but the chances of breaking anyone are virtually zero.

I am adding the news item and benchmark test.

NEWS.md
```diff
@@ -550,6 +550,8 @@
 
 53. `as.data.frame(DT, row.names=)` no longer silently ignores `row.names`, [#5319](https://github.com/Rdatatable/data.table/issues/5319). Thanks to @dereckdemezquita for the fix and PR, and @ben-schwen for guidance.
 
+54. data.table construction uses `deparse` to try and automatically name un-named cols. The attempt is now limited to 1- line which gives substantial speedup in [some edge cases](https://github.com/Rdatatable/data.table/pull/5501). Thanks @OfekShilon for the report and fix, @michaelchirico for guidance.
+
```
mattdowle, line 554: > data.table construction

Does that mean a `data.table(...)` call with arguments passed to it?  If so, could "data.table construction" be replaced with `data.table(...)` ?

> The attempt is now limited to 1- line

Does that mean calls to `data.table(...)` split over several lines (which I guess is very common) will lose this feature in the 2nd line onwards?  i.e.

```
data.table( A = 1:5,
            mean(my_vec)      # column name will be different now?
)
```

mattdowle (changes requested):
Great find. Just on the news item, I try to imagine myself as user reading news, and this is the thought I would have.

NEWS.md
OfekShilon: I refined the phrasing a bit (see the [commit](https://github.com/Rdatatable/data.table/pull/5501/commits/d70ef652973c82b1192ae4d484f5e16470530ae7)) - it is now both more accurate and less understandable.. I'd be thankful for any suggestion, and am totally fine with omitting the NEWS item altogether. AFAICT this change has no user-facing consequences, beyond some speedup.

mattdowle: Thanks. Ok I played with it a bit and tweaked the news item. It's not just 'some' speedup, but it scales with the data size. It's a tremendous finding certainly deserving of a news item and example!

PR #5508, opened 2022-10-31
ensure that DCF does not have empty line at the end before appending …

jangorecki:
…to it, as that would result in build failure

follow up of comments in https://github.com/Rdatatable/data.table/commit/9acf6d974d297bbfeb5a6eda3de76db9accc7acf#comments

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -63,7 +63,7 @@ build: ## build data.table sources as tar.gz archive
   before_script:
     - Rscript -e 'install.packages(c("knitr","rmarkdown"), repos=file.path("file:",normalizePath("bus/mirror-packages/cran")), quiet=TRUE)'
     - rm -r bus
-    - sed -i '${/^$/d;}' ./DESCRIPTION ## remove last empty line in DCF
+    - sed -i '${/^$/d;}' ./DESCRIPTION ## make last line end abruptly; i.e. without a final \n
     - echo "Revision:" $CI_BUILD_REF >> ./DESCRIPTION
   script:
     - R CMD build .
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -63,6 +63,7 @@ build: ## build data.table sources as tar.gz archive
   before_script:
     - Rscript -e 'install.packages(c("knitr","rmarkdown"), repos=file.path("file:",normalizePath("bus/mirror-packages/cran")), quiet=TRUE)'
     - rm -r bus
+    - sed -i '${/^$/d;}' ./DESCRIPTION ## remove last empty line in DCF
```
mattdowle, line 66: An "empty line" to me means `\n\n` at the end. Just having a `\n` at the end of the last line is just a normal line ending, not an empty line being creating.

So, to be absolutely clear, suggest comment be :
```
make last line end abruptly; i.e. without a final \n
```

Further, I tested it, and if the file ends with `\n[ ]+` then the last empty line is not removed.
And if there are 2 or more `\n` at the end, only 1 is removed.

mattdowle (changes requested):

.gitlab-ci.yml
jangorecki: > Further, I tested it, and if the file ends with \n[ ]+ then the last empty line is not removed.
And if there are 2 or more \n at the end, only 1 is removed.

I think those two are clear issues that we should not auto-fix in CI.

mattdowle: `R CMD build` has no problem with spaces or newlines at the end so why do you view these as an issue at all, let alone a clear issue?

jangorecki: Because they serve no purpose. Last new line has no purpose as well, but it is easier to be added during editing DCF. Although now, after your change of closing brackets of authors being in the last line, I think we could skip `sed` in my PR altogether, as the last line will not have to be edited anymore.

mattdowle: Because they serve no purpose is a reason not to let them get in the way again in future. That is the goal here: to prevent the same problem happening again. The GLCI script is the one applying the `>>` operation to DESCRIPTION, so that is what should prepare for that operation the best it can. It's that simple. So that we don't have the same problem again in future.

I will change it to do what I asked, merge and move on.

PR #5509, opened 2022-11-02
Strengthen CRAN_Release.cmd to find UBSAN errors in data.table.Rcheck

### LINKED ISSUE #5496: test() could catch UBSAN errors
Follow up to https://github.com/Rdatatable/data.table/issues/5491#issuecomment-1287900278.
Perhaps `test()` could capture `stderr` or use `errno` or something like that.
A reproducible example would be to compile `Rdevel-strict-clang` 2022-10-18 r83118 as described in `CRAN_Release.cmd` and then run test 1189 of `as.ITime()`.

mattdowle:
Closes #5496

Made many attempts to capture UBSAN errors on stderr but nothing worked. sink()-ing type="message" does work but the UBSAN errors still appeared in the output rather than the sink'd file. Also tried a Linux only `stat /proc/self/fd/2` before and after the test to see if the timestamp of stderr had increased, with no luck. Setting ` UBSAN_OPTIONS=log_path=....` is possible but seems too UBSAN specific; I was hoping to capture anything on stderr to include ASAN too for example and any other tool that might arise in future.

So in the end, just strengthened CRAN_Release.cmd to do a full `R CMD check` under UBSAN/ASAN rather than just test.data.table(). So now .Rd examples and vignettes are covered too. And included a `grep` on data.table.Rcheck/*.Rout for the UBSAN errors. Can refine it in future.

### DIFF
--- a/.dev/CRAN_Release.cmd
+++ b/.dev/CRAN_Release.cmd
@@ -289,11 +289,11 @@ cd ~/build
 wget -N https://stat.ethz.ch/R/daily/R-devel.tar.gz
 rm -rf R-devel
 rm -rf R-devel-strict-*
-tar xvf R-devel.tar.gz
+tar xf R-devel.tar.gz
 mv R-devel R-devel-strict-gcc
-tar xvf R-devel.tar.gz
+tar xf R-devel.tar.gz
 mv R-devel R-devel-strict-clang
-tar xvf R-devel.tar.gz
+tar xf R-devel.tar.gz
 
 sudo apt-get -y build-dep r-base
 cd R-devel  # may be used for revdep testing: .dev/revdep.R.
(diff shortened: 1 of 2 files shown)


.dev/CRAN_Release.cmd
```diff
@@ -329,15 +333,22 @@ alias Rdevel-strict-gcc='~/build/R-devel-strict-gcc/bin/R --vanilla'
...
+# passed through to here. However, our configure script seems to get in the way and gets them from {R_HOME}/bin/R
+# So I needed to edit my ~/.R/Makevars to get CFLAGS the way I needed.
+Rdevel-strict-[gcc|clang] CMD check data.table_1.14.5.tar.gz
+# Use the (failed) output to get the list of currently needed packages and install them
+Rdevel-strict-[gcc|clang]
 isTRUE(.Machine$sizeof.longdouble==0)  # check noLD is being tested
 options(repos = "http://cloud.r-project.org")
-install.packages(c("bit64","xts","nanotime","R.utils","yaml")) # minimum packages needed to not skip any tests in test.data.table()
-# install.packages(c("curl","knitr"))                          # for `R CMD check` when not strict. Too slow to install when strict
+install.packages(c("bit64", "bit", "curl", "R.utils", "xts","nanotime", "zoo", "yaml", "knitr", "rmarkdown"))
```
MichaelChirico, line 345: maybe set Ncpus if this is really so slow?

MichaelChirico (commented):
LGTM. getting closer to what CRAN actually does seems the best route, even a few extra hours spent building/testing is worth it

.dev/CRAN_Release.cmd
mattdowle: It's not so slow anymore. Didn't time it but under 30 mins I think. Could be that clang's ubsan is faster than gcc's perhaps. I tended to use gcc in the past. Anyway, Ncpus is a good idea, should use it regardless.

PR #5513, opened 2022-11-04
control compression level for fwrite with gzip

### LINKED ISSUE #5506: Set compression level for gzipped output in fwrite
Can we have functionality to choose compression level for gzipped CSVs in `fwrite()`?

ben-schwen:
Closes #5506

Still need to discuss the name of new argument and argument order for `fwrite`

- [x] implementation
- [ ] tests
- [ ] NEWS 

``` r
library(data.table)
DT = data.table(a=sample(10L, 1000L, TRUE), b=rep(1:4,each=250))
fwrite(DT, file=f1<-tempfile(fileext=".gz"), gzip_ratio=1L)
fwrite(DT, file=f2<-tempfile(fileext=".gz"), gzip_ratio=2L)
fwrite(DT, file=f3<-tempfile(fileext=".gz"), gzip_ratio=3L)
fwrite(DT, file=f4<-tempfile(fileext=".gz"), gzip_ratio=4L)
fwrite(DT, file=f5<-tempfile(fileext=".gz"), gzip_ratio=5L)
fwrite(DT, file=f6<-tempfile(fileext=".gz"), gzip_ratio=6L)
fwrite(DT, file=f7<-tempfile(fileext=".gz"), gzip_ratio=7L)
fwrite(DT, file=f8<-tempfile(fileext=".gz"), gzip_ratio=8L)
fwrite(DT, file=f9<-tempfile(fileext=".gz"), gzip_ratio=9L)
file.size(f1, f2, f3, f4, f5, f6, f7, f8, f9)
#> [1] 1055 1021 1009  934  900  871  871  863  863
```


ben-schwen:
> I think compression ratio is something that fits well into use of a global option in R. 

That was also my first thought, but I think I read smth about discouraging the introduction of new global options. A new global argument would also make everything cleaner since I do not throw any errors if a gzip_ratio is chosen with another or no compression.

MichaelChirico:
Global option looks less appealing to me since this is only used in one place / doesn't appear to correspond neatly to any R-side concept (it's just an argument to `deflateInit2()` from the C library).

So we could do `gzip_ratio = getOption('datatable.gzip_ratio', NULL)`, but it's not clear where else that option would be relevant.

PS,

 (1) Why 'ratio'? The manual names the `deflateInit2()` argument as `level`, why not `gzip_level`? https://www.zlib.net/manual.html
 (2) Are we sure we're making a back-compatible change here? IINM the default per this PR is `0`, which is not `Z_DEFAULT_COMPRESSION`, which the manual gives as "currently 6" https://www.zlib.net/manual.html

ben-schwen:
> (1) Why 'ratio'? The manual names the `deflateInit2()` argument as `level`, why not `gzip_level`? https://www.zlib.net/manual.html (2) Are we sure we're making a back-compatible change here? IINM the default per this PR is `0`, which is not `Z_DEFAULT_COMPRESSION`, which the manual gives as "currently 6" https://www.zlib.net/manual.html

Ad (1) true that, `gzip_level` is a better name, will change
Ad (2) a compression level of 0 is currently set to `Z_DEFAULT_COMPRESSION` (in the `deflateInit` call) to stay in par if zlib ever changes the default level. I misread the documentation and thought that only values in `1:9` are valid, but apparently, 0 is also possible. Will change default value to -1 and set it to `Z_DEFAULT_COMPRESSION`

R/fwrite.R
```diff
@@ -10,6 +10,7 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
            buffMB=8, nThread=getDTthreads(verbose),
            showProgress=getOption("datatable.showProgress", interactive()),
            compress = c("auto", "none", "gzip"),
+           gzip_ratio = 0:9,
```
jangorecki, line 13: I think it's better to use camel case rather than underline here

MichaelChirico:
Superseded by #6393, PTAL there.

MichaelChirico:
It looks like the only differences are:

 1. Argument name, #6393 went with `compressLevel`, which I like a bit better
 2. Argument default, #6393 went with `6L` vs. `-1:9` here. I think just using `6L` is more straightforward, even though the one here _does_ illustrate all valid values concisely. OTOH, it suggests `0L` is valid, which it's not, really we'd want `c(-1L, 1:9)` which isn't so nice. In that case `c(6L, 1:5, 7:9)` might be better, but still I think just `6L` is fine.

PR #5520, opened 2022-11-10
Move ram tests

### LINKED ISSUE #5517: Reduce or move top 10 tests by RAM usage
Follow up to #5515 for #5507

More specifically, the union of the two lists of top 10 ids when `gc()` is before and after `ps_mem()`.

Keep repeating until we see, hopefully, the plot ending closer to the starting RAM usage. Currently it ends at 265MB. Perhaps with #5516 too it should be possible to end comfortably under 100MB. Since R itself plus data.table loaded uses 67MB.

mattdowle:
Closes #5517
Moved the 10 tests as planned. Ending RSS down from 265 to 200. So moving in the right direction but will need to rinse and repeat. 

```
      ID nTest   RSS diff
 1: 1147     7 164.9 26.9
 2: 1549     1 169.3 22.7
 3: 1835     1 183.5 17.9
 4:  646     1 119.8 17.8
 5:  861     1 136.5 16.4
 6: 1888     9 198.7 14.4
 7: 1760     2 167.7  8.2
 8: 1779    13 175.6  7.9
 9: 1035    55 142.6  6.2
10: 1378     5 149.1  4.5
```

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -17745,26 +17745,30 @@ DT = data.table(x = sample(letters[1:5], 20, TRUE),
                 c = sample(c(0+3i,1,-1-1i,NA), 20, TRUE),
                 l = sample(c(TRUE, FALSE, NA), 20, TRUE),
                 r = as.raw(sample(1:5, 20, TRUE)))
-load(testDir("test2224.Rdata")) # ans array
+load(testDir("test2224.Rdata")) # 47KB array 24x8 where each cell contains a length-20 result
 if (test_bit64) {
   DT[, i64:=as.integer64(sample(c(-2L,0L,2L,NA), 20, TRUE))]
 } else {
   ans = ans[, -match("i64",colnames(ans))]
 }
-test(2224.01, sapply(names(DT)[-1], function(col) {
-  sapply(list(1, 5, -1, -5, c(1,2), c(-1,1)), function(n) list(
-    # fill is tested by group in tests 2218.*; see comments in #5205
-    EVAL(sprintf("DT[, shift(%s, %d, type='lag'),    by=x]$V1", col, n)),
-    EVAL(sprintf("DT[, shift(%s, %d, type='lead'),   by=x]$V1", col, n)),
-    EVAL(sprintf("DT[, shift(%s, %d, type='shift'),  by=x]$V1", col, n)),
-    EVAL(sprintf("DT[, shift(%s, %d, type='cyclic'), by=x]$V1", col, n))
-  ))
-}), ans)
+i = 1L
+for (col in names(DT)[-1]) {
+  for (n in list(1, 5, -1, -5, c(1,2), c(-1,1))) {
+    for (type in c('lag','lead','shift','cyclic')) {
+      # fill is tested by group in tests 2218.*; see comments in #5205
+      # sapply(sapply()) changed to for(for(for())) to save 29MB, #5517
+      test(2224.1+i/10000,  # 192 tests here when test_bit64=TRUE; 168 when FALSE
+           EVAL(sprintf("DT[, shift(%s, %d, type='%s'), by=x]$V1", col, n, type)),
+           ans[[i]])
+      i = i+1L
+    }
+  }
+}
 a = 1:2  # fill argument with length > 1 which is not a call
-test(2224.02, DT[, shift(i, fill=a), by=x], error="fill must be a vector of length 1")
+test(2224.2, DT[, shift(i, fill=a), by=x], error="fill must be a vector of length 1")
 DT = data.table(x=pairlist(1), g=1)
 # unsupported type as argument
-test(2224.03, DT[, shift(x), g], error="Type 'list' is not supported by GForce gshift.")
+test(2224.3, DT[, shift(x), g], error="Type 'list' is not supported by GForce gshift.")
 
 # groupingsets by named by argument
 test(2225.1, groupingsets(data.table(iris), j=sum(Sepal.Length), by=c('Sp'='Species'), sets=list('Species')),
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -13980,14 +13980,7 @@ DT = data.table( id = 1:5 , val = letters[1:5] )
 test(1981.3, DT[, new_col := shift(val, "lead")],      error="is.numeric(n) is not TRUE")
 test(1981.4, DT[, new_col := shift(val, NA_integer_)], error="Item 1 of n is NA")
 
-# print of DT with many columns reordered them, #3306.
-DT = as.data.table(lapply(1:255, function(i)rep.int(i, 105L)))  # 105 to be enough for 'top 5 ... bottom 5' to print
-out = capture.output(print(DT))
-tt = out[grep("V",out)]
-tt = unlist(strsplit(gsub(" ","",tt), "V"))
-test(1982.1, tt[1L], "")
-tt = as.integer(tt[tt!=""])
-test(1982.2, tt, seq_along(tt))
+# 1982 moved to benchmark.Rraw, #5517
 
 # parse(text = 'list(`\\phantom{.}`)') fails, #3319
 DT <- data.table(x=1, y=1:5)
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -520,11 +520,7 @@ test(164, foo(f), DT[,mean(b),by=d])
 test(165, subset(DT,a>2), DT[a>2])
 test(166, suppressWarnings(split(DT,DT$grp)[[2]]), DT[grp==2])
 
-# and that plotting works
-test(167.1, DT[,plot(b,f)], NULL)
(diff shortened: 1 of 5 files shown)


mattdowle:
Current status 265 down to 160MB



```
      ID nTest   RSS diff
 1: 1750    36 141.1 14.5
 2: 2224     3 159.6 10.4
 3: 1300     1 115.4  6.9
 4: 1642    91 126.8  6.2
 5: 1826     2 147.0  5.8
 6: 1378     5 120.2  4.5
 7: 1850     1 146.9  3.7
 8: 2201     4 149.2  3.3
 9: 1241     1 106.5  3.0
10: 2114     8 145.9  2.7
```

inst/tests/tests.Rraw
```diff
@@ -9691,7 +9649,7 @@ nqjoin_test <- function(x, y, k=1L, test_no, mult="all") {
   gc()  # no longer needed but left in place just in case, no harm
 }
 
-dt1 = nq_fun(400L)
+dt1 = nq_fun(100L)  # 400 reduced to 100, #5517
```
jangorecki, line 9652: Few lines below we have set.seed which, I believe, together with this particular size of data was testing some particular code. Changing the size is likely to disable tests that were designed for this particular size.
I recall non equi join testing was a long process, and I was running many random data, compare against sqlite, and report back any issues to Arun. So changing the size is likely to ignore those precisely reported cases.

mattdowle: added note to benchmark.Rraw

mattdowle:
Current status 265 down to 130MB



```
      ID nTest   RSS diff
 1: 1538     2 116.3  9.4
 2: 1642    91 125.4  6.4
 3: 1378     5 108.0  4.5
 4: 1639   143 119.0  3.1
 5:   69     7 100.8  2.0
 6: 1989     1 127.8  1.9
 7: 1223   728 103.5  1.7
 8: 1437    36 109.2  1.2
 9: 1888     9 126.1  0.9
10: 1461     1 106.9  0.6
```

jangorecki:
Plot could have ylim starting from 0. There is big progress in reducing memory usage but without looking at the y scale plot is not much different.

mattdowle:
But I am looking at the y scale. Also I'm sometimes looking quite closely at the gaps which is easier when the range of data fills the y axis.

I'm also looking at the time `test.data.table(memtest=1)` takes. That's perhaps a crude method to simulate an artificially strained/overloaded/mismanaged server since every `gc()` will flush the cpu cache like a server that has too many tasks running concurrently which keeps context switching. That's a guess. Anyway that time is down from 13 mins to 9 mins on my laptop.

Btw, for anybody reading commits or comments now or in the future: I do know that when a change is made and I write "this saved 9MB" (for example) that that is not true. What is true is that the increase of 9MB in the plot of rss over time disappears from the top 10 when making that change, and that's the definition of "saving" that I mean: I'm writing roughly. All I know it's a change that lowers ram usage, that's all. The actual ram that is saved is something less than 9MB; rss has just stepped up to accommodate the various actual heaps. These are tipping points, that's all, and I'm using 9MB in the nominal sense (like 2x1 wood is not 2x1).

mattdowle:
1538 (#1 ID in last table above) was due to `tables()` so that was relieved by #5524 and its follow up without changing the test.

Then I was hoping #5501 might improve test 1642 (current #1 ID in the top 10) as that does construct calls albeit with `as.call` and `eval` rather than `do.call`, and not a `data.table()` call. But didn't get lucky. I've left the tracing in for now and it's really odd because the rss steps up on mere entry to `nqjoin_test` before it has done any work at all (and yes, there's a trace right before its call so I don't think I've missed something before its call). So I'm now thinking that functions defined within functions are something to do with memory usage, which the non-equi tests 1641-1652 do. I was pretty sure it was going to be the lapply through rows, but it isn't. Or at least, the step up is earlier on entry currently. And then when that is resolved, maybe it will then be the lapply. But if function definitions within functions turns out to be significant, then we should sweep the entire codebase to see where else in the internals we do that. Functions are byte compiled (we have that as TRUE in DESCRIPTION) but I don't know about functions within functions.

mattdowle:
Indeed, moving those functions outside `nqjoint_test` in 0f8e2d4 moved 1642 from top id with 9.9MB to 9th with 1.4MB. It didn't change the ending rss which is still 127MB but I'll keep the change and move on to the new top ID and keep going.
```
      ID nTest   RSS diff
 1: 1982     2 126.1  6.5
 2: 1378     5 108.4  4.6
 3: 1538     2 111.9  4.6
 4: 1639   143 116.2  3.6
 5: 1647    14 120.1  2.5
 6:   69     7 101.1  2.0
 7: 1223   728 103.8  1.8
 8: 1437    36 109.8  1.4
 9: 1642    91 117.6  1.4
10: 1989     1 127.4  1.3
```

mattdowle:
Ending RSS now 120.6MB



```
      ID nTest   RSS diff
 1: 1639   143 113.5  6.4
 2: 1378     5 108.1  4.5
 3: 1989     1 119.6  2.8
 4: 1647    14 116.7  2.1
 5:   69     7 100.8  2.0
 6: 1223   728 103.6  1.8
 7: 1437    36 109.7  1.6
 8: 1942    12 116.4  1.4
 9: 1641    14 114.6  1.1
10: 1888     9 116.0  1.0
```

1639 has 143 parts. I don't see anything obviously large there. Will trace the parts ...

mattdowle:
Another thing about changing the y-axis to start from zero, is that, afaik, `ylim=` needs the minimum and maximum. But here we would just want to set the lower bound to be 0 and let the function look at the data to find the upper bound. But that's not how it works: you have to provide the maximum to ylim too. So you either you make a crude guess for the maximum based on the existing plot and hard code that number, or you call max() on the data again (which introduces scope for a bug by using the wrong object there) and then there's doubt too about if you need to add a little bit to the max so that the maximum observation definitely doesn't get chopped off.   Does anyone know how to tell plot _just_ that the ylim should start at zero?   (Now that I'm pretty much done, it does now make sense to show a plot with y starting at 0 too.)

MichaelChirico:
I don't quite follow. if you're doing plot(x,y) you have y already, then plot(x, y, ylim=c(0, max(y))) is pretty normal and IIRC equivalent for the upper bound to the default

mattdowle:
Getting to a point of diminishing returns now. I'm scratching my head way too much over every test now. Things like `sapply(sapply(sapply(...)))` in tests generate a lot of intermediate objects but they should be small on these small datasets and should be released anyway (and maybe they are). It is still troubling why the steps up in RSS happen towards the end but at least they're under 5MB now. Two things it could be is the leak in fread (PR exists), and the finalizer on data.table objects. I'll do those next, see if it helps here, and either way get this on CRAN. Since it's still possible the CRAN Windows fail is not even related to tests.Rraw. If it still fails on CRAN Windows then we won't know whether this work helped but was still insufficient, or whether it's not just not tests.Rraw at all. Hoping we'll get lucky and it is tests.Rraw.

Now starting R with `$ R_GC_MEM_GROW=0 R` when running memtest. It doesn't make much difference but should use this option if ever doing this exercise again in future just to avoid bigger jumps up getting in the way.

Same plot but with ylim starting at 0 on the right



  

```
      ID nTest   RSS diff
 1: 1378     5 108.3  4.5
 2: 1639   149 111.2  4.0
 3: 1989     1 118.4  3.0
 4: 1647    14 114.9  2.2
 5:   69     7 101.0  2.1
 6: 1223   728 103.7  1.7
 7: 1641    14 112.7  1.5
 8: 1942    12 115.0  1.4
 9: 1437    36 109.5  1.2
10: 1834     2 113.9  1.2
```

mattdowle:
> I don't quite follow. if you're doing plot(x,y) you have y already, then plot(x, y, ylim=c(0, max(y))) is pretty normal and IIRC equivalent for the upper bound to the default

It's where `y` is not `y` but some complicated expression. Then you end up repeating the expression; e.g. a table name, a filter, and column name.

mattdowle:
adding support for `ylim.min`, yes, so it could be other minimums not just 0 while leaving `max` up to the function which I'm not sure is actually just `max(x)`, doesn't it depend on tick mark settings, margins, and things I don't want to know about.

mattdowle:
`fread` leak fix didn't help here. I guess that's because `fread` didn't leak when the data is small. Certainly under 100 rows it didn't leak, and maybe under 1,000 too because if the number of rows is less than the sample size then it doesn't need to estimate because the sample already reads all the rows and knows exactly how many rows there are before allocating. We didn't have any big test files anyway, and any large constructions have been moved to benchmark.Rraw.

image

```
      ID nTest   RSS diff
 1: 1378     5 108.0  4.4
 2: 1639   149 111.0  4.0
 3: 1989     1 118.1  2.7
 4: 1647    14 114.5  2.1
 5:   69     7 100.8  2.0
 6: 1223   728 103.5  1.7
 7: 1641    14 112.4  1.4
 8: 1437    36 109.3  1.2
 9: 1942    12 114.9  0.8
10: 1461     1 107.0  0.6
```

PR #5523, opened 2022-11-11
Use %chin% where possible

MichaelChirico:
Something we can consider adding to release procedures:

1. Add `` `%in%` = function(x, y) {if (is.character(x)) stop("Use %chin%"); match(x, y, nomatch = 0L) > 0L} `` to e.g. R/AllS4.R
2. Run test suites
3. Edit `%in%`➡️`%chin%` as needed

We could also just leave that in there permanently doing something like `if (doing_dev) "%in%" = ...`.


MichaelChirico:
@jangorecki, WDYT about how to discourage backsliding here?

jangorecki:
As long as input is not large I don't think it makes much difference...

jangorecki:
```
library(data.table)

data.table 1.14.10 using 4 threads (see ?getDTthreads).  Latest news: r-datatable.com
> x = rep("a", 1e6)

> system.time("a" %in% x)

   user  system elapsed 
  0.002   0.000   0.002 
> system.time("a" %chin% x)

   user  system elapsed 
  0.003   0.000   0.004 
``

**Comment by MichaelChirico** [comment]
That's precise too:

```r
microbenchmark(times = 1000, "a" %in% x, "a" %chin% x)
# Unit: microseconds
#          expr      min       lq     mean    median       uq        max neval cld
#    "a" %in% x  786.911  912.596 4075.828  981.5015 1145.422 143626.642  1000   b
#  "a" %chin% x 1856.862 1883.532 1917.688 1905.1475 1934.957   2792.704  1000  a 
```

Shouldn't that be a reason to improve `%chin%` then? `%in%` is even making two calls: `match()` and `>`.

jangorecki:
Could be but these are microseconds and no one ever complaints.

MichaelChirico:
In that case, being internally consistent (as in this PR) is the best choice IMO. We can improve `%chin%` as needed going forward.

PR #5524, opened 2022-11-12
tables(mb=type_size) faster lower bound MB by default

mattdowle:
* Long standing comment in `tables()`: `# object.size() is slow hence optional; TODO revisit`
* `mb=TRUE` still works and uses the default `type_size` function
* alternative functions can be passed to `mb=`, such as `object.size`
* motivated by test ID 1538 in https://github.com/Rdatatable/data.table/pull/5520#issuecomment-1312283904 but needed doing anyway

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -1,7 +1,24 @@
 # globals to pass NOTE from R CMD check, see http://stackoverflow.com/questions/9439256
 MB = NCOL = NROW = NULL
 
-tables = function(mb=TRUE, order.col="NAME", width=80,
+type_size = function(DT) {
+  # for speed and ram efficiency, a lower bound by not descending into character string lengths or list items
+  # if a more accurate and higher estimate is needed then user can pass object.size or alternative to mb=
+  # in case number of columns is very large (e.g. 1e6 columns) then we use a for() to avoid allocation of sapply()
+  ans = 0L
+  lookup = c("raw"=1L, "integer"=4L, "double"=8L, "complex"=16L)
+  for (i in seq_along(DT)) {
+    col = DT[[i]]
+    tt = lookup[storage.mode(col)]
+    if (is.na(tt)) tt = .Machine$sizeof.pointer
+    tt = tt*nrow(DT)
+    if (is.factor(col)) tt = tt + length(levels(col))*.Machine$sizeof.pointer
+    ans = ans + tt
+  }
+  ans + ncol(DT)*.Machine$sizeof.pointer  # column name pointers
+}
+
+tables = function(mb=type_size, order.col="NAME", width=80,
                   env=parent.frame(), silent=FALSE, index=FALSE)
 {
   # Prints name, size and colnames of all data.tables in the calling environment by default
@@ -13,6 +30,7 @@ tables = function(mb=TRUE, order.col="NAME", width=80,
     if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
     return(invisible(data.table(NULL)))
   }
+  if (isTRUE(mb)) mb=type_size  # can still use TRUE, although TRUE will now be the lower faster type_size method
   DT_names = all_obj[is_DT]
   info = rbindlist(lapply(DT_names, function(dt_n){
     DT = get(dt_n, envir=env)   # doesn't copy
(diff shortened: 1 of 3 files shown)


R/tables.R
```diff
@@ -1,7 +1,24 @@
...
+    tt = lookup[storage.mode(col)]
+    if (is.na(tt)) tt = .Machine$sizeof.pointer
+    tt = tt*nrow(DT)
+    if (is.factor(col)) tt = tt + length(levels(col))*.Machine$sizeof.pointer
+    ans = ans + tt
+  }
+  ans + ncol(DT)*.Machine$sizeof.pointer  # column name pointers
+}
+
+tables = function(mb=type_size, order.col="NAME", width=80,
```
MichaelChirico, line 21: a useful test would be checking that this gives the same answer as mb=object.size() for tables without string columns

man/tables.Rd
```diff
@@ -19,9 +19,9 @@ tables(mb=TRUE, order.col="NAME", width=80,
 \details{
 Usually \code{tables()} is executed at the prompt, where \code{parent.frame()} returns \code{.GlobalEnv}. \code{tables()} may also be useful inside functions where \code{parent.frame()} is the local scope of the function; in such a scenario, simply set it to \code{.GlobalEnv} to get the same behaviour as at prompt.
 
-Note that on older versions of \R, \code{object.size} may be slow, so setting \code{mb=FALSE} may speed up execution of \code{tables} significantly.
+`mb = utils::object.size` provides a higher and more accurate estimate of size, but may take longer. Its default `units="b"` is appropriate.
```
MichaelChirico, line 23: we should document the differences (esp. that char/list columns are skipped, and column attributes are skipped (except factor levels).

it may also make sense to leave mb=object.size as the default and set mb=type_size in our suite. my thinking is that in practice tables may have a large part of memory in string columns; hiding that by default may be confusing.

mattdowle: I think if people do use `tables()` they use it mainly to get a list of their data.table objects, number of rows and columns, column names, etc. Like a database has ways just to see the tables that exist.  I think they rarely use the MB column. So the slowness of getting a better MB figure definitely gets in the way of the more common desire just to list the tables.

mattdowle: under the `mb` argument further up in `.Rd` I did document the differences and esp. : 

> \item{mb}{ a function which accepts a \code{data.table} and returns its size in bytes. By default, \code{type_size} (same as \code{TRUE}) provides a fast lower bound by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). A column \code{"MB"} is included in the output unless \code{FALSE} or \code{NULL}. }

char/list columns are not completely skipped: the size of their pointers is included, but what those pointers point to is excluded. That's what I thought I conveyed quite well in mb argument there, no? The character string part applies to both character columns and the levels of factor columns.

It is missing that column attributes are skipped. That could be added. It does say 'fast lower bound', so perhaps words like "for example, by excluding ..." could be added.

R/tables.R
mattdowle: Yes good idea. With a caveat that if R-devel change object.size slightly it could break our test. Maybe find what object.size says now and match that but use that figure instead of the call. Then put the test that actually calls object.size somewhere off CRAN like maybe other.Rraw (by stretching how we think of other to include base).

man/tables.Rd
MichaelChirico: > I think they rarely use the MB column.

can't claim to speak for a typical user but personally that column is my main use case for tables()

e.g. when I'm doing exploratory stuff in a sandbox session and need to clear out some temp objects, tables() is a quickly one-liner and then I can run rm(x, tmp2, DT2) or w/e to axe the biggest memory hogs

> That's what I thought I conveyed quite well in mb argument there, no?

yes, conveyed well on re-read.

> perhaps words like "for example, by excluding ..." could be added, rather than trying to document exactly what the function does

yes this works for me, over-documenting would also be a mistake in this case -- fortunes(250): Use the Source, Luke!

> not exporting it at least for now

also agree w keeping it private, especially for now

R/tables.R
MichaelChirico: ok, marked for follow-up #5525

man/tables.Rd
mattdowle: yes even if I'm right about what's most common, there will be users like you who always use MB.  As a utility function, the mb= argument could be getOption("datatable.tables.mb", type_size) then?  Then you could set that option to object.size. Also a better name than `type_size` could be `fast_lower_bound`?  That way `tables(mb=fast_lower_bound,...` reads better in the argument list and .Rd.  Maybe the "MB" column could be named differently depending on the mb function used; e.g. "MB_fast_lower_bound" and "MB_object.size" ?  Although, those are a bit long?

mattdowle: Maybe just rename `type_size` as `lower_bound` to be a bit shorter so that `mb=lower_bound` and `"MB_lower_bound"` leaves no room for misleading anyone. (It can be left to inference that the reason for the lower_bound is to be faster.)

PR #5549, opened 2022-11-30
dcast only computes default fill if necessary

### LINKED ISSUE #5390: `first` function not allowed as `fun.aggregate` for `dcast`
I would expect I could do:

````r
DT <- data.table(v1 = rep(1:2, each = 6),
                 v2 = rep(rep(1:3, 2), each = 2),
                 v3 = rep(1:2, 6),
                 v4 = rnorm(6))
dcast(DT, v1 + v2 ~ ., value.var = "v4", fun.aggregate = data.table::first)
Error: Aggregating function(s) should take vector inputs and return a single value (length=1). However, function(s) returns length!=1. This value will have to be used to fill any missing combinations, and therefore must be length=1. Either override by setting the 'fill' argument explicitly or modify your function to handle this case appropriately.
```
However I get an error instead of
```r
dcast(DT, v1 + v2 ~ ., value.var = "v4", fun.aggregate = \(.).[1])
   v1 v2          .
1:  1  1 -1.3199733
2:  1  2  2.2407363
3:  1  3 -0.1241345
4:  2  1 -1.3199733
5:  2  2  2.2407363
6:  2  3 -0.1241345
```
Is it expected or a bug?

### LINKED ISSUE #5512: un-necessary warning for dcast on int columns with fun.aggregate=min/max
Hi! I noticed a warning which I think is potentially confusing and should be removed:
```r
> library(data.table)
data.table 1.14.5 IN DEVELOPMENT built 2022-11-01 15:48:48 UTC; root using 6 threads (see ?getDTthreads).  Latest news: r-datatable.com
> DT <- data.table(chr=c("a","b","b"), int=1:3)[, num := as.numeric(int)]
> dcast(DT, chr ~ ., list(min,max), value.var="int")
Key: <chr>
      chr int_min int_max
   <char>   <int>   <int>
1:      a       1       1
2:      b       2       3
Warning messages:
1: In dcast.data.table(DT, chr ~ ., list(min, max), value.var = "int") :
  NAs introduced by coercion to integer range
2: In dcast.data.table(DT, chr ~ ., list(min, max), value.var = "int") :
  NAs introduced by coercion to integer range
```
The result above is as expected, so I would suggest removing this warning to avoid confusion. (because of this warning, I mistakenly thought that the result was not computed correctly, but in fact the result is fine)

It seems that this issue is specific to when `value.var` is type integer, and `fun.aggregate` is `min` or `max`. No warning happens when we do min/max on numeric column:
```r
> dcast(DT, chr ~ ., list(min,max), value.var="num")
Key: <chr>
      chr num_min num_max
   <char>   <num>   <num>
1:      a       1       1
2:      b       2       3
```
And no warning happens when we do mean/sd on int column:
```r
> dcast(DT, chr ~ ., list(mean,sd), value.var="int")
Key: <chr>
      chr int_mean    int_sd
   <char>    <num>     <num>
1:      a      1.0        NA
2:      b      2.5 0.7071068
```

I searched the tests for this warning message and I found:
```
froll.Rraw:414:test(6000.114, frollmean(1:5, Inf), error="n must be positive integer values", warning="NAs introduced by coercion*")
froll.Rraw:416:test(6000.115, frollmean(1:5, c(5, Inf)), error="n must be positive integer values", warning="NAs introduced by coercion*")
nafill.Rraw:125:test(3.07, nafill(x, fill="asd"), x, warning=c("Coercing.*character.*integer","NAs introduced by coercion"))
```

tdhock:
Closes #5512 
Closes #5390 

In current master, `dcast(fill=NULL)` always computes a default fill value, even when there are no missing cells. For example this is the result of a new test case using current master

```r
> DT <- data.table(chr=c("a","b","b"), int=1:3)[, num := as.numeric(int)]
> dcast(DT, num ~ chr, min, value.var="int")
Key: <num>
     num     a     b
   <num> <int> <int>
1:     1     1    NA
2:     2    NA     2
3:     3    NA     3
Warning message:
In dcast.data.table(DT, num ~ chr, min, value.var = "int") :
  NAs introduced by coercion to integer range
```
In the code above it is normal to compute the fill value (`fill_value=as.integer(min(integer()))` which is NA) because it is used three times.

However the code below gives the following result using current master, indicating that a default fill value is computed, even though it is not used:
```r
> dcast(DT, . ~ chr, min, value.var="int")
Key: <.>
        .     a     b
   <char> <int> <int>
1:      .     1     2
Warning message:
In dcast.data.table(DT, . ~ chr, min, value.var = "int") :
  NAs introduced by coercion to integer range
```
Using this branch we get the output below (no warning), indicating that no default fill value was computed, because it is not necessary:
```r
> dcast(DT, . ~ chr, min, value.var="int")
Key: <.>
        .     a     b
   <char> <int> <int>
1:      .     1     2
> 
```


src/fcast.c
```diff
@@ -15,57 +22,54 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
...
-      const int *ithiscol = INTEGER(thiscol);
-      const int *ithisfill = INTEGER(thisfill);
       for (int j=0; j<ncols; ++j) {
         SET_VECTOR_ELT(ans, nlhs+j+i*ncols, target=allocVector(thistype, nrows) );
         int *itarget = INTEGER(target);
         copyMostAttrib(thiscol, target);
         for (int k=0; k<nrows; ++k) {
           int thisidx = idx[k*ncols + j];
-          itarget[k] = (thisidx == NA_INTEGER) ? ithisfill[0] : ithiscol[thisidx-1];
+          itarget[k] = (thisidx == NA_INTEGER) ? INTEGER(thisfill)[0] : INTEGER(thiscol)[thisidx-1];
```
ben-schwen, line 50: can't we just use the same code for `INTSXP`, `LGLSXP`, `REALSXP` and `CPLXSXP` with `coerceAs`?

tdhock: This is my first time hacking on this code, so I'm not sure, but I also had the feeling that it would be desirable to avoid the repeated logic in these switch cases. About usage of coerceAs, would that introduce unwanted overhead / performance penalty? I was thinking of solving that via a C macro. Anyway I would suggest saving that for another PR, though.

jangorecki: Value added from coerceAs is handling attributes and therefore classes like int64, not sure if relevant here

src/fcast.c
```diff
@@ -15,57 +22,54 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
...
     SEXPTYPE thistype = TYPEOF(thiscol);
     int nprotect = 0;
-    if (isNull(fill)) {
-      if (LOGICAL(is_agg)[0]) {
-        thisfill = PROTECT(allocNAVector(thistype, 1)); nprotect++;
-      } else thisfill = VECTOR_ELT(fill_d, i);
-    }
-    if (TYPEOF(thisfill) != thistype) {
-      thisfill = PROTECT(coerceVector(thisfill, thistype)); nprotect++;
+    if(some_fill){
```
shrektan, line 31:
```suggestion
    if (some_fill) {
```
comestic

src/fcast.c
```diff
@@ -3,6 +3,13 @@
 // #include <signal.h> // the debugging machinery + breakpoint aidee
 // raise(SIGINT);
 
+bool any_NA_int(int N_data, int *idx){
+  for (int data_i=0; data_i<N_data; ++data_i) {
+    if(idx[data_i] == NA_INTEGER)return true;
+  }
+  return false;
+}  
+
```
shrektan, line 12:
```suggestion
inline bool any_NA_int(int n, int *idx) {
  for (int i=0; i<n; ++i) {
    if (idx[i] == NA_INTEGER) return true;
  }
  return false;
}  

```
comestic and inline may better

src/fcast.c
```diff
@@ -96,92 +100,3 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
   UNPROTECT(1);
   return(ans);
 }
-
```
shrektan, line 99: why to remove those lines?

shrektan (approved):
Thanks. It looks good to me.

src/fcast.c
mattdowle: Yes agree `i` instead of `data_i` reads better.
The `int n` could be `const int n` too. Depending on architecture and compiler optimization abilities and flags, it can help speed to tell the compiler that the `n` is `const` so it can optimize the `i<n` computation on every loop iteration; e.g. loop unrolling. I don't really know and it's hard to test but the way I think about it is to use `const` when we can to give all compilers everywhere the best chance. The easier example is a loop condition like `i<LENGTH(x)`; that `LENGTH(x)` should be pre-saved into `const int n=LENGTH(x)` otherwise I suspect it's quite likely that `LENGTH(x)` will be recomputed on each and every iteration and that API call has overhead, albeit small. I know there's no API call in this example but even so just being in the habit of using `const` when we can might help speed and can sometimes help catch bugs by protecting that variable from writes.

mattdowle (changes requested):

src/fcast.c
```diff
@@ -15,57 +22,54 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
...
+      }
+      if (TYPEOF(thisfill) != thistype) {
+	thisfill = PROTECT(coerceVector(thisfill, thistype)); nprotect++;
+      }
     }
     switch (thistype) {
     case INTSXP:
     case LGLSXP: {
-      const int *ithiscol = INTEGER(thiscol);
-      const int *ithisfill = INTEGER(thisfill);
```
mattdowle, line 35: These are here like this to avoid the small overhead of the R API call `INTEGER()`. When that call (or the similar ones `REAL()`, `COMPLEX()`, etc) appear inside a loop body or loop condition, there is a noticeable slowdown. R doesn't use macros anymore for those API calls. Even though they are `inline` functions they are slightly heavier inside them due to the welcome ALTREP initiative. Perhaps the compiler is clever enough to branch-predict away the branches inside the inline functions, but that wasn't my experience. A few years ago we swept through the codebase and removed R API calls from loops.

NEWS.md
```diff
@@ -561,6 +561,8 @@
     identical(DT1, DT2)                  # TRUE
     ```
 
+55. `dcast(fill=NULL)` only computes default fill value if necessary, which eliminates some previous warnings which were potentially confusing, [#5512](https://github.com/Rdatatable/data.table/issues/5512). Thanks to Toby Dylan Hocking for the PR.
```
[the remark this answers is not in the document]
tdhock: ok, I updated the NEWS item to include the warning text.

src/fcast.c
```diff
@@ -15,57 +22,54 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
     SET_VECTOR_ELT(ans, i, VECTOR_ELT(lhs, i));
   }
   // get val cols
+  bool some_fill = any_NA_int(nrows*ncols, idx);
```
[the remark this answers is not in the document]
tdhock: I understand that the alternate approach would be to suppress the warnings when we call `fun` on the empty vector to get the fill value. That is possible, but seems to me wasteful/un-necessary, because when there are no NA in idx, then the fill value is not used (so computing it is not necessary, and confusing, which was actually the core issue -- if a value is not needed, then why would we compute it?).

src/fcast.c
tdhock: This is interesting, I will investigate.

src/fcast.c
tdhock: It is confusing to keep lots of commented code in main source. If somebody wants this code they can always go back and see them in git history.

src/fcast.c
tdhock: I did some time/memory benchmarks to compare usage of INTEGER before for loop, or in each iteration of for loop, source code: https://github.com/tdhock/atime/blob/dt-dcast/vignettes/data.table.Rmd
I observed that there is no significant difference in time/memory usage, see figures below. (median line, min/max ribbon over 10 timings for each input data size N) These data indicate that it does not matter if we use INTEGER before or inside the for loop, and that the speed/memory usage of this branch is comparable to master/release (no significant speedups nor slowdowns).
pr5549-no-agg-no-missing
pr5549-no-agg-half-missing
pr5549-min-agg

jangorecki: It would be much easier to understand the code if that would be just simple system.time()

jangorecki: Just looking at the last picture it seems that there is noticable difference. It is log y scale there, if you make it non log then it will be very clear. Also it make sense to benchmark only such loops rather than complete dcast call, as we are discussing those macros in general, rather than dcast in particular.

tdhock: Sorry that the code is not just a simple system.time -- this is necessary if we want to see the asymptotic scaling -- if there are any time differences as data size increases.
Actually there is no significant difference here (all median lines are overlapping min/max bands, over the 10 timings). 
If there was a difference on these log log plots it would look something like this https://github.com/Rdatatable/data.table/issues/5424#issuecomment-1197279832

jangorecki: Could you try running following code using `dataptr-loop` branch? for me macros are indeed much slower...
```r
test.dataptr = data.table:::dataptrLoop
set.seed(108)
x = 1:1e8
x[2] <- 3L ## ensure no altrep
system.time(test.dataptr(x, 1L)) ## dataptr outside the loop
#   user  system elapsed 
#  0.053   0.000   0.053 
system.time(test.dataptr(x, 2L)) ## dataptr inside the loop
#   user  system elapsed 
#  0.238   0.000   0.239 
```

tdhock: that is interesting thanks @jangorecki I will try that.
it seems to me that these data indicate that INTEGER before the for loop may result in speedups in some cases, but in the context of dcast it does not make a difference (other parts of the dcast code are the bottleneck, not this particular loop with INTEGER)

tdhock: pr5549-INTEGER-loop-alone
figure above is benchmark for Jan's test.dataptr, figure below is for two versions of dcast.
pr5549-fcast_in
These data indicate that there is some speedup if you move the INTEGER call before the loop, but using that technique does not result in speedups for dcast, because this loop is not the bottleneck.

tdhock: About Michael's point, I tried these benchmarks on both a windows and a linux machine, and got very similar results (windows results reported here).

MichaelChirico:
From #4586, let's include this as a test case & make sure it doesn't warn:

```r
test(xxxxxxx, dcast(data.table(a = 1, b = 2, c = 3), a ~ b, value.var = 'c', fill = '2'),
              data.table(a=1, `2`=3, key='a'))
```

tdhock:
I added that test case alongside a similar one.

MichaelChirico:
@tdhock is this PR ready for review? I'm wary of the conflict

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,13 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
   DT = data.table(x=sample(5,20,TRUE), y=sample(2,20,TRUE), z=sample(letters[1:2],20,TRUE), d1=runif(20), d2=1L)
   test(1102.38, names(dcast(DT, x ~ y + z, fun.aggregate=length, value.var = "d2", sep=".")),
                 c("x", "1.a", "1.b", "2.a", "2.b"))
+
+  # test for #5512, only compute default fill if needed.
+  DT = data.table(chr=c("a","b","b"), int=1:3)
+  test(1102.39, dcast(DT, int ~ chr, min, value.var="int"), data.table(int=1:3, a=c(1L,NA,NA), b=c(NA,2L,3L), key="int"), warning="NAs introduced by coercion to integer range")
```
tdhock, line 3735: hi @MichaelChirico I resolved the conflict, but now this test is failing. 
Instead of the usual base R warning (NAs introduced by coercion to integer range) we now get a dcast-specific warning, which I guess you added in that other PR? `inf (type 'double') at RHS position 1 out-of-range(NA) or truncated (precision lost) when assigning to type 'integer' (target vector)`
I was expecting the base R warning still.
Do you think this is desirable/expected? If so I will update the test to expect the new warning.

tdhock: in detail, min is being called for every combination of int/chr, including combinations for which there are no data:
```r
> min(c())
[1] Inf
Warning message:
In min(c()) : no non-missing arguments to min; returning Inf
```
Then Inf (double) is being converted to NA_integer_ ,
```r
> as.integer(min(c()))
[1] NA
Warning messages:
1: In min(c()) : no non-missing arguments to min; returning Inf
2: NAs introduced by coercion to integer range 
```

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 # data.table [v1.15.99](https://github.com/Rdatatable/data.table/milestone/30)  (in development)
 
+## BUG FIXES
+
+X. `dcast(fill=NULL)` only computes default fill value if necessary, which eliminates some previous warnings (for example, when fun.aggregate=min or max, warning was NAs introduced by coercion to integer range) which were potentially confusing, [#5512](https://github.com/Rdatatable/data.table/issues/5512). Thanks to Toby Dylan Hocking for the PR.
```
MichaelChirico, line 7: move this NEWS item and cite both closed bugs

inst/tests/tests.Rraw
MichaelChirico: warning comes from here. need to think more about why:

https://github.com/Rdatatable/data.table/blob/15c127e99f8d6aab599c590d4aec346a850f1334/src/assign.c#L873

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,15 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
   DT = data.table(x=sample(5,20,TRUE), y=sample(2,20,TRUE), z=sample(letters[1:2],20,TRUE), d1=runif(20), d2=1L)
   test(1102.38, names(dcast(DT, x ~ y + z, fun.aggregate=length, value.var = "d2", sep=".")),
                 c("x", "1.a", "1.b", "2.a", "2.b"))
+
+  # test for #5512, only compute default fill if needed.
+  DT = data.table(chr=c("a","b","b"), int=1:3, dbl=as.double(4:6))
+  test(1102.39, dcast(DT, int ~ chr, min, value.var="int"), data.table(int=1:3, a=c(1L,NA,NA), b=c(NA,2L,3L), key="int"), warning="NAs introduced by coercion to integer range")#as.integer(min(integer()) is Inf) is NA
+  test(1102.391, dcast(DT, int ~ chr, min, value.var="int", fill=NA), data.table(int=1:3, a=c(1L,NA,NA), b=c(NA,2L,3L), key="int"))
```
MichaelChirico, line 3736: this is the same output as the previous test, right? if so, consider doing `ans<-data.table(...)` in the previous test and just `ans` here. I think that will make the test series more readable as we don't need to mentally construct a new table when checking our intuition for how the test should work.

MichaelChirico, line 3736: minor point: I don't see a reason to subdivide the test number like this, I would just number 39,...,43 for the tests in this PR. IME adding a new digit to number tests is only done to squeeze new tests inside the flow of long-ago tests and avoid ballooning the diff just to keep tests numbered in serial, which won't apply to a series of highest-numbered tests like this.

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,15 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
...
+  test(1102.392, dcast(DT, int ~ chr, min, value.var="dbl"), data.table(int=1:3, a=c(4,Inf,Inf), b=c(Inf,5,6), key="int"))
+  test(1102.40, dcast(DT, . ~ chr, min, value.var="int"), data.table(.=".",a=1L,b=2L,key="."))
+  test(1102.41, dcast(data.table(a = 1, b = 2, c = 3), a ~ b, value.var = 'c', fill = '2'), data.table(a=1, `2`=3, key='a'))
```
MichaelChirico, line 3739: If I'm understanding this test correctly, I think it would be clearer to put this test first, it is closest to the spirit of the change described (don't calculate 'fill' if not needed). Then a quick comment transitioning to the tests about potential warnings emitted from `fun.aggregate=` in the edge case.

MichaelChirico: Does this change work with `fill = stop("shouldn't evaluate")`? if so, that might make the test as clear as possible. if not, why not?

src/fcast.c
```diff
@@ -15,24 +22,28 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
...
```
MichaelChirico, line 25: what are the performance implications of having to re-scan the full output for NA before applying `fill=`?

src/fcast.c
```diff
@@ -15,24 +22,28 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
...
+      if (isVectorAtomic(thiscol)) { // defer error handling to below, but also skip on list
+	thisfill = PROTECT(coerceAs(thisfill, thiscol, /*copyArg=*/ScalarLogical(false))); nprotect++;
+      }
     }
     switch (thistype) {
     case INTSXP:
     case LGLSXP: {
       const int *ithiscol = INTEGER(thiscol);
-      const int *ithisfill = INTEGER(thisfill);
+      const int *ithisfill = 0;
```
MichaelChirico, line 45: why declare `ithisfill` if it may not be used (`!some_fill`)?

src/fcast.c
```diff
@@ -96,92 +109,3 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
   UNPROTECT(1);
   return(ans);
 }
-
-// commenting all unused functions, but not deleting it, just in case
```
MichaelChirico, line 100: split this part to a separate PR?

src/fcast.c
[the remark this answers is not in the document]
MichaelChirico: Having a bit of trouble following the full context of Matt's comment here (probably elsewhere in the full comment history), but it looks similar to my reaction, which is some slight surprise that this is the approach for this functionality. I would have expected an approach that can tell on-the-fly (while the output is being constructed) whether the reshape is "complete" or "ragged" (the latter being where `fill=` is needed).

Is that not possible? If it's not possible, a comment here describing what was tried / why we think it can't be done would be great. If we're just settling on this approach for expedience and think it can be improved in the future, that may also be fine (pending performance check), but warrants a follow-up issue to revisit with instructions on what we're going for.

src/fcast.c
```diff
@@ -3,6 +3,13 @@
 // #include <signal.h> // the debugging machinery + breakpoint aidee
 // raise(SIGINT);
 
+inline bool any_NA_int(const int n, const int *idx) {
+  for (int i=0; i<n; ++i) {
+    if (idx[i] == NA_INTEGER) return true;
```
MichaelChirico, line 8: Slightly confused why we only need to look for `NA_INTEGER`, won't we ever need to look for non-`int` objects in the fill? Maybe I'm not fully understanding the approach/how it fits in to normal execution of `dcast()`.

src/fcast.c
tdhock: ok

src/fcast.c
tdhock: it needs to be declared (otherwise compilation error) since we use it below in
```c
itarget[k] = (thisidx == NA_INTEGER) ? ithisfill[0] : ithiscol[thisidx-1];
```
(unless we do a bigger if/else with more repetitive code but I don't think that is a good idea)

src/fcast.c
tdhock: My guess would be insignificant but I could run a performance test to verify.

src/fcast.c
tdhock: fcast has argument SEXP idxArg which is INTEGER describing which data to reshape (NA if not used)

inst/tests/tests.Rraw
tdhock: not it does not, because standard evaluation is used (fill is always evaluated, even if not used)

inst/tests/tests.Rraw
tdhock: ok

tdhock: ok

inst/tests/tests.Rraw
tdhock: ok I will change the test

inst/tests/tests.Rraw
tdhock: and I think this test is closest to the spirit of the change, so I put it first:
```r
test(1102.39, dcast(DT, . ~ chr, min, value.var="int"), data.table(.=".",a=1L,b=2L,key="."))#fill not used in output, so default fill not computed, and no warning emitted during computation of that default fill.
```

NEWS.md
tdhock: ok

src/fcast.c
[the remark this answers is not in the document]
tdhock: the idxArg comes from this line of R code in fcast.R:
```r
    idx = do.call("CJ", mapunique)[map, 'I' := .I][["I"]] # TO DO: move this to C and avoid materialising the Cross Join.
```
so we could do anyNA(idx) and pass that through as another argument to fcast, but I suppose there would be no efficiency gain, and then we would have another argument to fcast, which seems desirable to avoid.

inst/tests/tests.Rraw
tdhock: `fill = stop("shouldn't evaluate")` could be implemented if we hack fcast.R (this PR only touches fcast.c)

MichaelChirico: thanks for clarifying! i don't think there's strong reason to do that. By far the typical use case for `fill=` supplies a literal value; [I don't see any case where `fill = foo()`](https://github.com/search?q=lang%3AR+%2F%28%5B%5Ea-z0-9._%3A%5D%7Ce%3A%3A%29dcast%5B%28%5D%5B%5E%29%5D%2F+%2F%5B%5Ea-z0-9._%5Dfill%5Cs*%3D%5Cs*%5Ba-z0-9._%5D%2B%5B%28%5D%2F+NOT+%2Fgeom_%2F&type=code) where `foo()` might be slow & we'd benefit from lazy evaluation.

We can re-evaluate if there's a compelling FR later.

src/fcast.c
MichaelChirico: quick test:

```r
DT = CJ(a = letters, b = letters, c = letters, d = letters, e = LETTERS)
DT[, v := rnorm(.N)]

system.time(replicate(5, dcast(DT, a + b + c + d ~ e, value.var = 'v', fill = 0)))
## master
#    user  system elapsed 
#  14.877   1.437   8.186 
## branch fix5512 @ 3c7fb24483cbe05cf2db0638be8a71c8f5120299
#    user  system elapsed 
#  14.817   1.414   8.163 

## if there _are_ NA to fill
DT = DT[-sample(.N, 1e5)]

system.time(replicate(5, dcast(DT, a + b + c + d ~ e, value.var = 'v', fill = 0)))
## master
#    user  system elapsed 
#  14.093   1.558   7.674 
## branch fix5512 @ 3c7fb24483cbe05cf2db0638be8a71c8f5120299
#    user  system elapsed 
#  14.268   1.399   7.674 
```

So it looks like not a significant concern, though please to pipe in if you think that benchmark is not representative. Thanks!

src/fcast.c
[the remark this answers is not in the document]
MichaelChirico: If you think it's something worth pursuing in the longer term, please file a follow-up, but I'm happy for now.

NEWS.md
```diff
@@ -28,6 +28,8 @@
 
 3. Optimized `shift` per group produced wrong results when simultaneously subsetting, for example, `DT[i==1L, shift(x), by=group]`, [#5962](https://github.com/Rdatatable/data.table/issues/5962). Thanks to @renkun-ken for the report and Benjamin Schwendinger for the fix.
 
+4. `dcast(fill=NULL)` only computes default fill value if necessary, which eliminates some previous warnings (for example, when fun.aggregate=min or max, warning was NAs introduced by coercion to integer range) which were potentially confusing, [#5512](https://github.com/Rdatatable/data.table/issues/5512), [#5390](https://github.com/Rdatatable/data.table/issues/5390). Thanks to Toby Dylan Hocking for the fix.
```
MichaelChirico, line 31: >  `dcast(fill=NULL)` only computes default fill value if necessary

This reads to me like `fill` is evaluated lazily, which as discussed it's not, do we need to be more precise in the wording here?

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,16 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
   DT = data.table(x=sample(5,20,TRUE), y=sample(2,20,TRUE), z=sample(letters[1:2],20,TRUE), d1=runif(20), d2=1L)
   test(1102.38, names(dcast(DT, x ~ y + z, fun.aggregate=length, value.var = "d2", sep=".")),
                 c("x", "1.a", "1.b", "2.a", "2.b"))
+
+  # test for #5512, only compute default fill if needed.
+  DT = data.table(chr=c("a","b","b"), int=1:3, dbl=as.double(4:6))
+  test(1102.39, dcast(DT, . ~ chr, min, value.var="int"), data.table(.=".",a=1L,b=2L,key="."))#fill not used in output, so default fill not computed, and no warning emitted.
```
MichaelChirico, line 3735:
```suggestion
  test(1102.39, dcast(DT, . ~ chr, min, value.var="int"), data.table(.=".",a=1L,b=2L,key=".")) # fill not used in output, so default fill not computed, and no warning emitted.
```

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,16 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
...
+  ans <- data.table(int=1:3, a=c(1L,NA,NA), b=c(NA,2L,3L), key="int")
+  test(1102.40, dcast(DT, int ~ chr, min, value.var="int"), ans, warning="inf (type 'double') at RHS position 1 out-of-range(NA) or truncated (precision lost) when assigning to type 'integer' (target vector)")#warning emitted during computation of default fill since as.integer(min(integer()) is Inf) is NA.
```
MichaelChirico, line 3737:
```suggestion
  test(1102.40, dcast(DT, int ~ chr, min, value.var="int"), ans, warning="inf (type 'double') at RHS position 1 out-of-range(NA) or truncated (precision lost) when assigning to type 'integer' (target vector)") # warning emitted during computation of default fill since as.integer(min(integer()) is Inf) is NA.
```

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,16 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
...
+  test(1102.41, dcast(DT, int ~ chr, min, value.var="int", fill=NA), ans)
+  test(1102.42, dcast(DT, int ~ chr, min, value.var="dbl"), data.table(int=1:3, a=c(4,Inf,Inf), b=c(Inf,5,6), key="int"))
+  test(1102.43, dcast(data.table(a = 1, b = 2, c = 3), a ~ b, value.var = 'c', fill = '2'), data.table(a=1, `2`=3, key='a'))
```
MichaelChirico, line 3740: Here I'd either add a comment about the lack of warning, or perhaps add a second test where there _is_ a warning (i.e. `fill` is needed) to contrast. (or maybe I'm still not understanding this test)

src/fcast.c
```diff
@@ -3,6 +3,13 @@
 // #include <signal.h> // the debugging machinery + breakpoint aidee
 // raise(SIGINT);
 
+bool any_NA_int(const int n, const int *idx) {
```
MichaelChirico, line 6: I do find it a bit odd to use this helper only once... is there anywhere else in the code-base that could re-use this? otherwise, WDYT about just using `anyNA(idx)` at the R level to avoid re-inventing the wheel? (I guess requires a new argument to `fcast()` to pass on this result)

MichaelChirico:
Basically ready to go, some more minor feedback this round. Thanks!

inst/tests/tests.Rraw
tdhock: great

src/fcast.c
tdhock: not sure if it could be used elsewhere.
I would rather not add another argument to fcast, if that is OK with you?
I believe it is already very complicated to understand how the fcast C/R code work together, and adding another argument would only make that worse.

inst/tests/tests.Rraw
tdhock: I deleted this test, after all it seems un-necessary.

NEWS.md
tdhock: I'm not sure I understand. What wording would you suggest?

inst/tests/tests.Rraw
```diff
@@ -3732,12 +3732,15 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
 
   # test for #5512, only compute default fill if needed.
   DT = data.table(chr=c("a","b","b"), int=1:3, dbl=as.double(4:6))
-  test(1102.39, dcast(DT, . ~ chr, min, value.var="int"), data.table(.=".",a=1L,b=2L,key="."))#fill not used in output, so default fill not computed, and no warning emitted.
+  mymin <- function(x){
+    if(length(x)==0)stop("calling mymin on vector of length 0")
```
tdhock, line 3736: There are two operations which are relevant to test, (1) computing the default fill value by using fun.aggregate on a zero-length vector, and (2) coercing the result to the output type.
When fun.aggregate=min with integer(), we get a warning `NAs introduced by coercion to integer range` which comes from (2) whereas the other warning `no non-missing arguments to min; returning Inf` which would happen in (1) is suppressed in R code:
```r
fill.default = suppressWarnings(dat[0L][, eval(fun.call)])
```
But the point of this PR is that it would be preferable to do neither (1) nor (2) in the case where there are no missing entries to fill, so this is why I add tests involving mymin, which stops with an error if it is called on a zero-length vector.

src/fcast.c
tdhock: I appreciate this review. It helped me think of a better test involving fun.aggregate=mymin which stops with an error on zero-length input. Makes me wonder if we can handle this entirely on the R side now, I'm not sure.

tdhock: After all, it makes more sense to use anyNA in R, then use that to determine whether to compute the default fill, and pass the result to C.

inst/tests/tests.Rraw
```diff
@@ -3729,6 +3729,19 @@ test(1100, dt1[dt2,roll=-Inf,rollends=c(FALSE,TRUE)]$ind, INT(NA,NA,1,2,2,2,2,2,
   DT = data.table(x=sample(5,20,TRUE), y=sample(2,20,TRUE), z=sample(letters[1:2],20,TRUE), d1=runif(20), d2=1L)
   test(1102.38, names(dcast(DT, x ~ y + z, fun.aggregate=length, value.var = "d2", sep=".")),
                 c("x", "1.a", "1.b", "2.a", "2.b"))
+
+  # test for #5512, only compute default fill if needed.
+  DT = data.table(chr=c("a","b","b"), int=1:3, dbl=as.double(4:6))
+  mymin <- function(x){
+    if(length(x)==0)stop("calling mymin on vector of length 0")
```
MichaelChirico, line 3736:
```suggestion
    if (!length(x)) stop("calling mymin on vector of length 0")
```

inst/tests/tests.Rraw
MichaelChirico: PR making much more sense now, I had misunderstood -- I thought we were _also_ avoiding warning for a case like

```r
dcast(DT, a ~ b, value.var = "c", fill = 'xxx')
```

Where `c` is numeric. I.e., if `fill` is not needed we don't need the `as.numeric('xxx')` warning to show up.

Indeed, IINM that _does_ happen on this PR:

```r
DT = data.table(a = 1, b = 2, c = 3)
# warns on master, not on this branch
dcast(DT, a ~ b, value.var = 'c', fill = 'xxx')
```

Was that an unintentional improvement? If so we could add a test to make sure it stays

R/fcast.R
```diff
@@ -152,23 +152,22 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
...
-  fill.default = NULL
   if (is.null(fun.call)) {
     oo = forderv(dat, by=varnames, retGrp=TRUE)
     if (attr(oo, 'maxgrpn', exact=TRUE) > 1L) {
       messagef("'fun.aggregate' is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables %s used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). Check the resulting table for values larger than 1 to see which combinations were not unique. See ?dcast.data.table for more details.", brackify(varnames))
       fun.call = quote(length)
     }
   }
-  if (!is.null(fun.call)) {
+  dat_for_default_fill = dat
```
MichaelChirico, line 162: I don't think this will trigger a copy, but I'm nevertheless wary. WDYT about making this `dat_for_default_fill = dat[0L]` instead? Since that's all that's needed below.

R/fcast.R
```diff
@@ -211,7 +210,12 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
     }
     maplen = vapply_1i(mapunique, length)
     idx = do.call("CJ", mapunique)[map, 'I' := .I][["I"]] # TO DO: move this to C and avoid materialising the Cross Join.
-    ans = .Call(Cfcast, lhs, val, maplen[[1L]], maplen[[2L]], idx, fill, fill.default, is.null(fun.call))
+    fill.default = NULL
+    some_fill = anyNA(idx)
+    if (run_agg_funs && is.null(fill) && some_fill) {
+      fill.default = dat_for_default_fill[0L][, maybe_err(eval(fun.call))]
+    }
```
MichaelChirico, line 217:
```suggestion
    some_fill = anyNA(idx)
    fill.default = if (run_agg_funs && is.null(fill) && some_fill) dat_for_default_fill[0L][, maybe_err(eval(fun.call))]
```

R/fcast.R
```diff
@@ -211,7 +210,12 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
...
+    ans = .Call(Cfcast, lhs, val, maplen[[1L]], maplen[[2L]], idx, fill, fill.default, is.null(fun.call), some_fill)
```
MichaelChirico, line 218: Probably something to examine as a follow-up but I'm surprised we need all of `fill`, `fill.default`, and `some_fill` as arguments, it seems we should be able to remove at least one. Unless you're sure all three are necessary, I'll write the issue.

R/fcast.R
tdhock: great

R/fcast.R
tdhock: yes I think they are all necessary, I was surprised too. 
in particular I tried replacing if(some_fill) in C code on !isNull(fill_d) but that did not work (there are times when fill_d is NULL but there is apparently something to fill still, not exactly sure why)

inst/tests/tests.Rraw
tdhock: yes indeed I added a test.

MichaelChirico (approved):
thanks for your efforts!

tdhock:
you are welcome! thanks very much for your feedback, too!
However we have one failing test, and I've figured out which commit we need to revert, https://github.com/Rdatatable/data.table/pull/5549/commits/747c76cd8a65e46294dd42edf69294a60f3b9e94 but not sure why, any ideas? 
```
(base) tdhock@tdhock-MacBook:~/R/data.table(fix5512)$ git checkout 747c76cd8a65e46294dd42edf69294a60f3b9e94 && R CMD INSTALL . && R -e "library(data.table);dcast(data.table(id=1, grp=1, e=expression(1)), id ~ grp, value.var='e')"
HEAD est maintenant sur 747c76cd dat_for_default_fill is zero-row dt
Le chargement a nécessité le package : grDevices
* installing to library ‘/home/tdhock/lib/R/library’
** using staged installation
gcc 12.3.0
zlib 1.2.11 is available ok
** libs
using C compiler: ‘gcc (GCC) 12.3.0’
gcc -shared -L/home/tdhock/lib/R/lib -L/usr/local/lib -o data.table.so assign.o between.o bmerge.o chmatch.o cj.o coalesce.o dogroups.o fastmean.o fcast.o fifelse.o fmelt.o forder.o frank.o fread.o freadR.o froll.o frollR.o frolladaptive.o fsort.o fwrite.o fwriteR.o gsumm.o idatetime.o ijoin.o init.o inrange.o nafill.o negate.o nqrecreateindices.o openmp-utils.o programming.o quickselect.o rbindlist.o reorder.o shift.o snprintf.o subset.o transpose.o types.o uniqlist.o utils.o vecseq.o wrappers.o -fopenmp -lz -L/home/tdhock/lib/R/lib -lR
PKG_CFLAGS = -fopenmp
PKG_LIBS = -fopenmp -lz
installing to /home/tdhock/lib/R/library/00LOCK-data.table/00new/data.table/libs
** R
** inst
Le chargement a nécessité le package : grDevices
** help
*** installing help indices
** building package indices
Le chargement a nécessité le package : grDevices
** installing vignettes
Le chargement a nécessité le package : grDevices
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
Le chargement a nécessité le package : grDevices
** testing if installed package keeps a record of temporary installation path
* DONE (data.table)

R version 4.3.2 (2023-10-31) -- "Eye Holes"
Copyright (C) 2023 The R Foundation for Statistical Computing

R est un logiciel libre livré sans AUCUNE GARANTIE.
Vous pouvez le redistribuer sous certaines conditions.
Tapez 'license()' ou 'licence()' pour plus de détails.

R est un projet collaboratif avec de nombreux contributeurs.
Tapez 'contributors()' pour plus d'information et
'citation()' pour la façon de le citer dans les publications.

Tapez 'demo()' pour des démonstrations, 'help()' pour l'aide
en ligne ou 'help.start()' pour obtenir l'aide au format HTML.
Tapez 'q()' pour quitter R.

Le chargement a nécessité le package : grDevices
> library(data.table);dcast(data.table(id=1, grp=1, e=expression(1)), id ~ grp, value.var='e')
Erreur dans `[.data.table`(dat, 0) : 
  Internal error: column type 'expression' not supported by data.table subset. All known types are supported so please report as bug.
Appels : dcast -> dcast.data.table -> [ -> [.data.table
Exécution arrêtée
(base) tdhock@tdhock-MacBook:~/R/data.table((HEAD détachée sur 747c76cd))$ git checkout ee93c5fc7b7fde45b37ebf95d06ddf7f58af1c5e && R CMD INSTALL . && R -e "library(data.table);dcast(data.table(id=1, grp=1, e=expression(1)), id ~ grp, value.var='e')"
La position précédente de HEAD était sur 747c76cd dat_for_default_fill is zero-row dt
HEAD est maintenant sur ee93c5fc Update R/fcast.R
Le chargement a nécessité le package : grDevices
* installing to library ‘/home/tdhock/lib/R/library’
** using staged installation
gcc 12.3.0
zlib 1.2.11 is available ok
** libs
using C compiler: ‘gcc (GCC) 12.3.0’
gcc -shared -L/home/tdhock/lib/R/lib -L/usr/local/lib -o data.table.so assign.o between.o bmerge.o chmatch.o cj.o coalesce.o dogroups.o fastmean.o fcast.o fifelse.o fmelt.o forder.o frank.o fread.o freadR.o froll.o frollR.o frolladaptive.o fsort.o fwrite.o fwriteR.o gsumm.o idatetime.o ijoin.o init.o inrange.o nafill.o negate.o nqrecreateindices.o openmp-utils.o programming.o quickselect.o rbindlist.o reorder.o shift.o snprintf.o subset.o transpose.o types.o uniqlist.o utils.o vecseq.o wrappers.o -fopenmp -lz -L/home/tdhock/lib/R/lib -lR
PKG_CFLAGS = -fopenmp
PKG_LIBS = -fopenmp -lz
installing to /home/tdhock/lib/R/library/00LOCK-data.table/00new/data.table/libs
** R
** inst
Le chargement a nécessité le package : grDevices
** help
*** installing help indices
** building package indices
Le chargement a nécessité le package : grDevices
** installing vignettes
Le chargement a nécessité le package : grDevices
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
Le chargement a nécessité le package : grDevices
** testing if installed package keeps a record of temporary installation path
* DONE (data.table)

R version 4.3.2 (2023-10-31) -- "Eye Holes"
Copyright (C) 2023 The R Foundation for Statistical Computing

R est un logiciel libre livré sans AUCUNE GARANTIE.
Vous pouvez le redistribuer sous certaines conditions.
Tapez 'license()' ou 'licence()' pour plus de détails.

R est un projet collaboratif avec de nombreux contributeurs.
Tapez 'contributors()' pour plus d'information et
'citation()' pour la façon de le citer dans les publications.

Tapez 'demo()' pour des démonstrations, 'help()' pour l'aide
en ligne ou 'help.start()' pour obtenir l'aide au format HTML.
Tapez 'q()' pour quitter R.

Le chargement a nécessité le package : grDevices
> library(data.table);dcast(data.table(id=1, grp=1, e=expression(1)), id ~ grp, value.var='e')
Erreur dans dcast.data.table(data.table(id = 1, grp = 1, e = expression(1)),  : 
  Unsupported column type in fcast val: 'expression'
Appels : dcast -> dcast.data.table
Exécution arrêtée
```

MichaelChirico (approved):

MichaelChirico:
I think the codecov thing is spurious, merging.

R/fcast.R
```diff
@@ -152,23 +152,22 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
...
-    if (is.null(fill)) {
-      fill.default = suppressWarnings(dat[0L][, eval(fun.call)])
-      # tryCatch(fill.default <- dat[0L][, eval(fun.call)], error = function(x) stopf(errmsg))
-      if (nrow(fill.default) != 1L) stopf(errmsg)
+    maybe_err = function(list.of.columns) {
+      if (any(lengths(list.of.columns) != 1L)) stopf("Aggregating function(s) should take vector inputs and return a single value (length=1). However, function(s) returns length!=1. This value will have to be used to fill any missing combinations, and therefore must be length=1. Either override by setting the 'fill' argument explicitly or modify your function to handle this case appropriately.")
+      list.of.columns
     }
-    dat = dat[, eval(fun.call), by=c(varnames)]
+    dat = dat[, maybe_err(eval(fun.call)), by=c(varnames)]
```
jangorecki, line 170: I think this could possibly affect the code path in [.
We check for `eval` in `j` to handle that specially, not sure if we are checking for nested `eval`.
I would go with new `env` argument, or eventually modify `fun.call` by prefixing it with `maybe_err` call.

R/fcast.R
```diff
@@ -211,7 +210,12 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
     }
     maplen = vapply_1i(mapunique, length)
     idx = do.call("CJ", mapunique)[map, 'I' := .I][["I"]] # TO DO: move this to C and avoid materialising the Cross Join.
-    ans = .Call(Cfcast, lhs, val, maplen[[1L]], maplen[[2L]], idx, fill, fill.default, is.null(fun.call))
+    some_fill = anyNA(idx)
+    fill.default = if (run_agg_funs && is.null(fill) && some_fill) dat_for_default_fill[, maybe_err(eval(fun.call))]
+    if (run_agg_funs && is.null(fill) && some_fill) {
+      fill.default = dat_for_default_fill[0L][, maybe_err(eval(fun.call))]
```
jangorecki, line 216: same here

R/fcast.R
MichaelChirico: Are you thinking for efficiency? Otherwise passing tests ensures your concern is moot right?

jangorecki: possibly for efficiency but just touching the edge like that raises some concerns

PR #5556, opened 2022-12-07
Fix #5286

### LINKED ISSUE #5286: `shallow` adds .internal.selfref to an input data.frame
The non-exported `shallow` is supposed to operate on data.frame arguments since 1.14.2.  But:
```r
> df=data.frame(a=1:10)
> attributes(data.table:::shallow(df))
$names
[1] "a"

$class
[1] "data.frame"

$row.names
 [1]  1  2  3  4  5  6  7  8  9 10

$.internal.selfref
<pointer: 0x56250b748b10>
```
[It appears](https://github.com/Rdatatable/data.table/blob/master/src/assign.c#L198)  `setselfref` is called regardless of if the input is indeed a data.table.

Also, test 1984.27 was supposed to catch it - but the way data frames/tables comparison is coded in `test` currently it doesn't. The culprit are these lines in the comparison:
```
        setattr(xc,".internal.selfref",NULL)   # test 2212
        setattr(yc,".internal.selfref",NULL)
```

`#` `Output of sessionInfo()`
R4.0.5, data.table 1.14.3

OfekShilon:


inst/tests/tests.Rraw
```diff
@@ -14047,7 +14047,7 @@ test(1984.242, na.omit(data.table(A=c(1,NA,2)), cols=character()), data.table(A=
 test(1984.25, rbindlist(list(DT[1L], DT[2L]), idcol = TRUE), data.table(.id=1:2, a=1:2))
 test(1984.26, setalloccol(`*tmp*`), error='setalloccol attempting to modify `*tmp*`')
 DF = as.data.frame(DT)
-test(1984.27, shallow(DF), DF)  # shallow (which is not exported) works on DF from v1.14.2
+test(1984.27, identical(shallow(DF), DF), TRUE)  # shallow (which is not exported) works on DF from v1.14.2
```
MichaelChirico, line 14050:
```suggestion
test(1984.27, identical(shallow(DF), DF))  # shallow (which is not exported) works on DF from v1.14.2. identical() to force checking the selfref attribute
```

MichaelChirico:
Hi @OfekShilon thanks for the PR. Would you like to take it back up? If so, please:

 1. Add a better PR title, it should be self-contained. Only link other issues in the PR description.
 2. Add NEWS

MichaelChirico:
Feel free to re-open if you'd like to continue this PR. Thanks!

OfekShilon:
@MichaelChirico  I no longer work in IstraResearch and have no access to that fork.

MichaelChirico:
Got it. would you like to continue the PR if I publish the branch back to the main repo here?

OfekShilon:
I can try..

PR #5561, opened 2022-12-14
add favicons to webpage

### LINKED ISSUE #3956: website issues reported
Two issues with the website pointed out by Grant McDermott:

https://twitter.com/grant_mcdermott/status/1181597318920826880

https://twitter.com/grant_mcdermott/status/1181592771112493056

> Nice. Relatedly, I’m not sure if the (sub)subheadings on the first vignette (e.g. here https://rdatatable.gitlab.io/data.table/articles/datatable-intro.html#get-all-the-flights-with-jfk-as-the-origin-airport-in-the-month-of-june- ) should have a dash prefix. They aren’t bullet formatted and they don’t appear in the content menu either.

> Was just looking at this yesterday. So much easier to navigate than the old wiki setup! One minor thing I noticed, though: The RHS contents menu isn’t rendering some subheadings correctly.

Confirmed about both, though not sure what (if anything) we can do. sub-heads are rendering fine for this `tidyverse` vignette: https://tidyr.tidyverse.org/articles/pivot.html

While here it would also be nice to get the `data.table` logo to render as the tab icon (not sure what the lingo is for that thing):

Screenshot 2019-10-10 at 10 44 07 AM

ben-schwen:
Towards #3956


jangorecki:
I wonder if that could possibly live in `.graphics` directory? then we don't have extra new directory for exactly the same purpose.

jangorecki:
@ben-schwen yes, copy in CI feels easiest. Then we don't need to change pkgdown yaml, righr? Look at the _integration_ job. It publishes to a directory named pkgdown already, so either copy it to pkgdown/favicon (if pkgdown needs them for rendering) and then remove that directory. Or you just copy it after to pkgdown/pkgdown/favicon, if it is not needed during render website.

ben-schwen:
@jangorecki I adjusted the CI accordingly, now copying the files from `.graphics`. Previously we were building the pkgdown site into `./pkgdown` and this needs the folder to be empty (since we overwrite it). Therefore I changed the destination of the pkgdown build to `./website` (also in the subsequent places).

jangorecki:
Changing name to website looks good but I think you forget to copy icons from .graphics dir in CI, no?

ben-schwen:
It's done here but the line might be a bit bloated

https://github.com/Rdatatable/data.table/blob/ee600f0c23c2e353e9f41ca38f938a25caa9bb8d/.gitlab-ci.yml#L300

jangorecki:
Uh, yes, please put in own line, so each workaround has own line then

.gitlab-ci.yml
```diff
@@ -297,8 +297,9 @@ integration:
   script:
     - R --version
     - *install-deps ## markdown pkg not present in r-pkgdown image
+    - rm -rf ./pkgdown && mkdir -p ./pkgdown/favicon/ && cp .graphics/favicon/* ./pkgdown/favicon/ ## copy favicons
```
jangorecki, line 300:
```suggestion
    - mkdir -p ./pkgdown/favicon/ && cp .graphics/favicon/* ./pkgdown/favicon/ ## copy favicons
```

PR #5570, opened 2022-12-21
Rbindlist retains attributes

### LINKED ISSUE #5569: I had expected rbindlist() to preserve attributes, but it doesn't.  It'd be a nice feature; or an important limitation to disclose in the documentation.
``` r
library(data.table)
x <- data.table(0)
setattr(x,"xa",0)
y <- data.table(1)
setattr(y,"ya",1)
z <- rbindlist(list(x,y))
attributes(z)
#> $names
#> [1] "V1"
#> 
#> $row.names
#> [1] 1 2
#> 
#> $class
#> [1] "data.table" "data.frame"
#> 
#> $.internal.selfref
#> <pointer: 0x000001e86a87bf50>
```

<sup>Created on 2022-12-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

`#` `Output of sessionInfo()`
``` r
sessionInfo()
#> R version 4.2.2 (2022-10-31 ucrt)
#> Running under: Windows 10 x64 (build 22621)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_New Zealand.utf8  LC_CTYPE=English_New Zealand.utf8   
#> [3] LC_MONETARY=English_New Zealand.utf8 LC_NUMERIC=C                        
#> [5] LC_TIME=English_New Zealand.utf8    
#> 
#> attached base packages:
#> 
#>  [1] rstudioapi_0.14   knitr_1.41        magrittr_2.0.3    R.cache_0.16.0   
#>  [5] rlang_1.0.6       fastmap_1.1.0     stringr_1.5.0     styler_1.8.1     
#>  [9] highr_0.9         tools_4.2.2       xfun_0.34         R.oo_1.25.0      
#> [13] cli_3.4.1         withr_2.5.0       htmltools_0.5.3   yaml_2.3.6       
#> [17] digest_0.6.29     lifecycle_1.0.3   purrr_0.3.5       vctrs_0.5.0      
#> [21] R.utils_2.12.2    fs_1.5.2          glue_1.6.2        evaluate_0.18    
#> [25] rmarkdown_2.18    reprex_2.0.2      stringi_1.7.8     compiler_4.2.2   
#> [29] R.methodsS3_1.8.2
```

<sup>Created on 2022-12-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

ben-schwen:
Closes #5569

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1327,6 +1327,7 @@ has a better chance of working on Mac.
 
 46. Adding a `list` column containing an item of type `list` to a one row `data.table` could fail, [#3626](https://github.com/Rdatatable/data.table/issues/3626). Thanks to Jakob Richter for reporting.
 
+47. `rbind` and `rbindlist` silently dropped attributes of all arguments, now arguments of first input argument are retained, [#5569](https://github.com/Rdatatable/data.table/issues/5569). Thanks to @cthombor for reporting and Benjamin Schwendinger for the fix.
 ## NOTES
 
 1. `rbindlist`'s `use.names="check"` now emits its message for automatic column names (`"V[0-9]+"`) too, [#3484](https://github.com/Rdatatable/data.table/pull/3484). See news item 5 of v1.12.2 below.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1327,7 +1327,7 @@ has a better chance of working on Mac.
 
 46. Adding a `list` column containing an item of type `list` to a one row `data.table` could fail, [#3626](https://github.com/Rdatatable/data.table/issues/3626). Thanks to Jakob Richter for reporting.
 
-47. `rbind` and `rbindlist` silently dropped attributes of all arguments, now arguments of first input argument are retained, [#5569](https://github.com/Rdatatable/data.table/issues/5569). Thanks to @cthombor for reporting and Benjamin Schwendinger for the fix.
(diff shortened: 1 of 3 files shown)


jangorecki:
Are data.table attributes retained or columns attributes as well?
How about binding attributes with `c()`? I can imagine that preserving attributes of the first DT won't always be much useful.

ben-schwen:
> Are data.table attributes retained or columns attributes as well? How about binding attributes with `c()`? I can imagine that preserving attributes of the first DT won't always be much useful.

Currently, `rbindlist` always retains column attributes from the first appearance of a column, which is usually the first input of the list, but can also be a later one if using `fill=TRUE`.

I also thought about binding the attributes from all inputs with `c()` but this won't work since we also store keys and indices as attributes.

For example `rbind(x,y)` where `x` and `y` both have the same key `"a"` then binding would result in the key `c("a","a")`. Keys are even more problematic if we use `use.names=FALSE` because later keys or indices do not have to appear in the result data.table.

jangorecki:
I would prefer to spent more time on design of this feature to avoid breaking changes later on.
New argument could possibly accept a function, the `first` would just take column/table attributes of a first DT, while `c` would combine column/table attributes from all objects. Then another switch is to control behavior for column and table separately. Then another factor is to exclude particular attributes of a column/table from being retained, and here good defaults would be `key` and `index`.
Not sure if we want to rush just to have some implementation done.

PR #5629, opened 2023-04-26
add CODEOWNERS file

tdhock:
The CODEOWNERS file is documented here, https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
Basically, github uses this file to automatically notify people for code review, when a (non-draft) PR is opened.
Creating this explicit mapping from files to reviewers would be essential to enable a more de-centralized code review process (less work for @mattdowle).
In this first draft of CODEOWNERS, I have volunteered to be reviewer for files related to melt.
@MichaelChirico @jangorecki @ben-schwen would you please (1) add lines to CODEOWNERS to indicate files that you would like to review, and (2) @tag-other-people who you think may be interested to volunteer as reviewers?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,5 @@
+#https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+* @mattdowle
+/R/fmelt.R @tdhock
+/src/fmelt.c @tdhock
+/man/melt.data.table.Rd @tdhock


tdhock:
In this PR let's also do some re-organization of what objects are defined in which R/* files, creating new smaller files with more specific functionality. For example the new/proposed DT function is currently defined in data.table.R, which has a lot of other things in it too, so maybe we could move the DT defintion to its own file DT.R?

tdhock:
maybe wait for @MichaelChirico ?

MichaelChirico:
Hi folks, just back from vacation.

Question about this, I was thinking of adding some files that I have a relatively high amount of experience with (e.g. fread), but I wasn't clear about the syntax after reading the docs.

If we have

```
* @mattdowle
/src/fread.c @michaelchirico
```

Does that mean Matt's "ownership" is overridden by mine for `src/fread.c`?

I think that means we should use

```
* @mattdowle
/src/fread.c @mattdowle @michaelchirico
```

To ensure Matt can still be an owner for that file, is that right?

MichaelChirico:
I was also unsure my most highly-contributed parts of the repo so I did the following (in case it's useful to anyone else):

```r
stats = 'git log --author="michaelchirico4@gmail.com" --pretty=tformat: --numstat' |>
  system(intern=TRUE) |>
  paste(collapse = "\n") |>
  data.table::fread(na.strings = "-") |>
  data.table::setnames(c("added", "removed", "path"))
stats = stats[, by = path, .(
  added = sum(as.integer(added), na.rm = TRUE),
  removed = sum(as.integer(removed), na.rm = TRUE)
)]
stats[, total := added + removed]
stats[order(-total)]
```

ben-schwen:
> Question about this, I was thinking of adding some files that I have a relatively high amount of experience with (e.g. fread), but I wasn't clear about the syntax after reading the docs.
> 
> If we have
> 
> ```
> * @mattdowle
> /src/fread.c @michaelchirico
> ```
> 
> Does that mean Matt's "ownership" is overridden by mine for `src/fread.c`?
> 
> I think that means we should use
> 
> ```
> * @mattdowle
> /src/fread.c @mattdowle @michaelchirico
> ```
> 
> To ensure Matt can still be an owner for that file, is that right?

Good point. Yes, only the last match for a file counts, according to docs. So if we want Matt to be the code owner of every file we need to add him to every line. However, I thought the whole idea of this is to take smth off Matt's plate.

jangorecki:
Once we have interested contributors gathered in this PR, then we could consider some, ideally minor, file structure reorganizations. So it could possibly be easier to assign owners of files. For example R/data.table.R holds way too much stuff. If some parts of it would be under ownership of a particular contributor, then could make sense to take that parts out into new file.

tdhock:
thanks!

PR #5631, opened 2023-04-27
add subset and dogroups support for columns of type expression

### LINKED ISSUE #5596: column type 'expression' not supported by data.table subset
The example below produces the following error:
Error in `[.data.table`(data.table(a = 1, b = expression(1)), , b, by = a) : 
  Internal error: column type 'expression' not supported by data.table subset. All known types are supported so please report as bug.

`#` Minimal reproducible example
data.table(a=1,b=expression(1))[,b,by=a]

`#` `Output of sessionInfo()`
R version 4.2.2 (2022-10-31 ucrt)

Matrix products: default

locale:
[1] LC_COLLATE=Danish_Denmark.utf8  LC_CTYPE=Danish_Denmark.utf8    LC_MONETARY=Danish_Denmark.utf8 LC_NUMERIC=C                    LC_TIME=Danish_Denmark.utf8    

attached base packages:

other attached packages:
[1] data.table_1.14.6

[1] compiler_4.2.2  tools_4.2.2     rstudioapi_0.14

ben-schwen:
Closes #5596


NEWS.md
```diff
@@ -12,6 +12,8 @@
 
 2. `fwrite()` respects `dec=','` for timestamp columns (`POSIXct` or `nanotime`) with sub-second accuracy, [#6446](https://github.com/Rdatatable/data.table/issues/6446). Thanks @kav2k for pointing out the inconsistency and @MichaelChirico for the PR.
 
+3. Subsetting or aggregating columns of type `expression` works, [#5596](https://github.com/Rdatatable/data.table/issues/5596). Thanks to @tsp for the report, and @ben-schwen for the fix.
```
MichaelChirico, line 15: I guess we should clarify what operations are/aren't supported here. E.g. this still won't work:

```r
dt = data.table(a=1:2, b=expression(1,2))
dt[,b[1],a]
# Type 'expression' is not supported by GForce head/tail/first/last/`[`
```

Is it worth adding support for that, or just specifying when "expression-in-`j`-with-`by`" works?

ben-schwen: I would see this as different issue to not depatch to GForce in the first place for type `expression`

MichaelChirico (approved):

NEWS.md
MichaelChirico: I'm not sure anyone actually needs this anyway; we can revisit upon request.

MichaelChirico:
Thanks! Narrowed the NEWS item to signal `expression` is not fully-supported as a column type, and moved it to NEW FEATURES since I don't think it was really a bug before, just not supported.

Issue #5634, opened 2023-05-04
`fread` handling of row names with `header=TRUE`


MLopez-Ibanez:
```R
library(data.table)
txt <- '
  foo  foo2
1   0 false
2   1    NA
'
print(read.table(text = txt, header = TRUE))
print(fread(text = txt, header="auto"))
print(fread(text = txt, header = TRUE))
```

`read.table()` handles the row names as expected.
`fread(header="auto")` gives a warning and creates an extra column. Not great but can be fixed with `suppressWarnings()` and removing the extra column.
`fread(header=TRUE)` should do the same but instead gives:

```R
   1 0 false
1: 2 1  <NA>
```

which is completely wrong. 

Verbose output: 

```R
fread(text=txt, header="auto", verbose=TRUE)
  OpenMP version (_OPENMP)       201511
  omp_get_num_procs()            4
  R_DATATABLE_NUM_PROCS_PERCENT  unset (default 50)
  R_DATATABLE_NUM_THREADS        unset
  R_DATATABLE_THROTTLE           unset (default 1024)
  omp_get_thread_limit()         2147483647
  omp_get_max_threads()          4
  OMP_THREAD_LIMIT               unset
  OMP_NUM_THREADS                unset
  RestoreAfterFork               true
  data.table is using 2 threads with throttle==1024. See ?setDTthreads.
Input contains a \n or is ")". Taking this to be text input (not a filename)
[01] Check arguments
  Using 2 threads (omp_get_max_threads()=4, nth=2)
  NAstrings = [<<NA>>]
  None of the NAstrings look like numbers.
  show progress = 1
  0/1 column will be read as integer
[02] Opening the file
  `input` argument is provided rather than a file name, interpreting as raw text to read
[03] Detect and skip BOM
[04] Arrange mmap to be \0 terminated
  \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
[05] Skipping initial rows if needed
  Positioned on line 2 starting: <<  foo  foo2>>
[06] Detect separator, quoting rule, and ncolumns
  Detecting sep automatically ...
  sep=' '  with 2 lines of 3 fields using quote rule 0
  Detected 3 columns on line 2. This line is either column names or first data row. Line starts as: <<1   0 false>>
  Quote rule picked = 0
  fill=false and the most number of columns found is 3
[07] Detect column types, good nrow estimate and whether first row is column names
[...]

PR #5635, opened 2023-05-04
Link to ?read.delim in ?fread to give a closer analogue of expected behavior

MLopez-Ibanez:
See #5634


man/fread.Rd
```diff
@@ -2,7 +2,7 @@
 \alias{fread}
 \title{ Fast and friendly file finagler }
 \description{
-   Similar to \code{read.table} but faster and more convenient. All controls such as \code{sep}, \code{colClasses} and \code{nrows} are automatically detected.
+   Similar to \code{\link[utils:read.delim]{read.delim()}} but faster and more convenient. All controls such as \code{sep}, \code{colClasses} and \code{nrows} are automatically detected.
```
MichaelChirico, line 5: Hmm, I actually don't think this is right.

`read.csv()` and `read.delim()` are thin wrappers around `read.table()`. The former sets `sep=","`, the latter `sep="\t"`.

However, `fread()` uses `sep="auto"` by default, meaning it's more generic than either `read.csv()` or `read.delim()`. Thus I think `read.table()` is the best fit (even though `fread()`'s auto-detection is not available in `read.table()`).

Partially, this is splitting hairs -- `?read.table`, `?read.csv`, and `?read.delim` are all the same page.

```suggestion
   Similar to \code{\link[utils:read.table]{read.table()}} but faster and more convenient. All controls such as \code{sep}, \code{colClasses} and \code{nrows} are automatically detected.
```

MichaelChirico: Sorry, I've just read the referenced discussion (FFR: use `#<issue>` in the PR body to make it easier on reviewers -- numbers in the PR title aren't auto-linked).

Maybe the compromise here is "Similar to [read.csv()] or [read.delim()]", that way we convey that it's for delimited files but not implying the full generality of `read.table()`, WDYT?

MLopez-Ibanez: Sounds OK. Thanks.

MichaelChirico:
Thanks!

PR #5642, opened 2023-05-24
add auto-index skip

### LINKED ISSUE #1264: Auto indexing should be skipped if subset and modify are on the same column
See [this SO post](http://stackoverflow.com/q/31989067/559784) for more info. Doing:

``` r
DT[y %in% 5:10, y := 100]
```

will create an index, and get modified, which results in the index getting removed, and as a result will be slow (due to creating the index unnecessarily every time) on bigger data.

ben-schwen:
Closes #1264 

Timings for [SO example](http://stackoverflow.com/q/31989067/559784)
```r
set.seed(1)
DF = data.table(x=rep(c("a","b","c"),each=40000000), y=sample(c(1,3,6),40000000,T), v=1:9)
DF1=DF2=DF
system.time(DF[y==6,"y"]<-10)
#>    user  system elapsed 
#>   2.059   0.720   2.779
system.time(DF1$y[DF1$y==6]<-10)
#>    user  system elapsed 
#>   2.462   1.416   3.880
system.time(DF2[y==6, y := 10]) 
#>    user  system elapsed 
#>   7.638   0.975   2.896

# after PR
system.time(DF2[y==6, y := 10]) 
#>    user  system elapsed 
#>   1.322   0.347   1.277 
```


MichaelChirico:
<pending discussion in issue #1264>

PR #5644, opened 2023-05-25
namespace-qualify methods::as() as needed

MichaelChirico:
As identified in https://bugs.r-project.org/show_bug.cgi?id=18540


MichaelChirico:
Trivial change to Rd and test files, merging.

Issue #5654, opened 2023-06-12
Allow explicitly setting `.datatable.aware <- FALSE` to use base R data.frame syntax


dvg-p4:
# Feature request: Allow setting `.datatable.aware <- FALSE` to explicitly use base R data.frame syntax in a datatable-aware package

## Motivation

I'm coding a package that uses data.table features in some places but not in others. We import data.table in NAMESPACE in order to have access to `:=` etc. for the parts of our code that do use the full features of `data.table`. We would like to be able to write user-facing functions that take either a base `data.frame` or a `data.table` as an argument, and use them as input with some simple selecting and filtering. 

## Problem

data.table indexing semantics are by design different from data.frame semantics. So in many simple cases, it's difficult or impossible to write a call that will index both a data.frame and a data.table the same way.

As [this vignette](https://rdatatable.gitlab.io/data.table/articles/datatable-importing.html#data-table-in-imports-but-nothing-imported) explains, you can set `.datatable.aware = TRUE` in a package to use data.table indexing semantics without importing data.table in NAMESPACE. However, as far as I can tell, there's no way to do the reverse--disable data.table semantics and go back to base R data.frame indexing for a certain function call.

## Partial workarounds

A couple imperfect strategies I've considered to get around this:
* `as.data.table()` / `as.data.frame()`: Creates an unnecessary copy of the data--slow on big tables and antithetical to the core point of data.table
* `setDT()` / `setDF()`: Does not behave consistently/stably when called on a function argument, see https://github.com/Rdatatable/data.table/issues/5618
* `subset()`: Lets me do programmatic column selection and subsets, but verbose and not actually consistent in some edge cases

## Suggestion:

Allow setting `.datatable.aware <- FALSE` to tell data.table that it should use base R data.frame indexing semantics in an otherwise data.table-aware package. This does not work currently since `data.table:::cedta()` does a big OR of this variable with the other conditions. But it seems like adding an `if (isFALSE(ns$.datatable.aware)) return FALSE` check before the big OR might do the trick.

dvg-p4:
Implemented and merged into master (v1.15.99) by #5655

dvg-p4:
Another workaround that I found in the meantime:
```r
class(df) <- "data.frame"
```
appears to do a shallow copy, and thus can accomplish a similar effect.

PR #5655, opened 2023-06-12
Check for .datatable.aware being FALSE

dvg-p4:
Building a quick implementation of #5654 based off the code added in https://github.com/Rdatatable/data.table/pull/4909. Technically this won't check if the package sets `.datatable.aware = FALSE` in its top-level namespace, but I can't imagine a use case for that--why would you be importing data.table in NAMESPACE if you didn't want to use its syntax at all?

TODO:
- [x] Double-check https://rdatatable.gitlab.io/data.table/CONTRIBUTING.html to see if this list is missing any steps
- [x] See if `devtools::check()` passes on Linux (getting thread-related errors on my macbook)
- [x] Check if this works on the command line
- [x] Check if this works within a package function (as motivated)
- [x] Write a test case for this
- [x] Make sure the new test fails on master
- [x] Make sure all tests (including the new one) pass on the PR branch
- [x] Update NEWS
- [x] Mark as ready for review


R/cedta.R
```diff
@@ -29,6 +29,9 @@ cedta = function(n=2L) {
   if (isTRUE(env$.datatable.aware)) {  # dtplyr 184
     return(TRUE)
   }
+  if (isFALSE(env$.datatable.aware)) {  #5654
```
MichaelChirico, line 32: We have `isTRUEorFALSE`:

https://github.com/Rdatatable/data.table/blob/cde7333938a590a3cddbda1b02103650e2f55d15/R/utils.R#L14

So we could more concisely write:

```r
if (isTRUEorFALSE(env$.datatable.aware)) return(env$.datatable.aware)
```

MichaelChirico:
BTW, for this case, feel free to confirm it works locally only. It will be cumbersome to write a test case (I reckon) since AIUI it will require creating & installing a test package. I don't think we have any tests like that so far, and this functionality is very close to existing & well-worn functionality, so I don't expect it to do anything surprising. Therefore just testing locally & reporting back is sufficient IMO as reviewer.

dvg-p4:
> Therefore just testing locally & reporting back is sufficient IMO as reviewer.

Ok, that makes this a lot more straightforward--let me get on that and I'll see if I can get this ready for review today.

dvg-p4:
```
Test 2244.2 ran without errors but selfrefok(x) is FALSE
```

This is starting to go a bit over my head but I suspect this is a deeper error that I've revealed rather than caused, glancing at the comments above `_selfrefok()`'s definition:

https://github.com/Rdatatable/data.table/blob/cde7333938a590a3cddbda1b02103650e2f55d15/src/assign.c#L93-L96

dvg-p4:
And this works as intended in a test package! https://github.com/dvg-p4/DataTableAwareTestPkg/tree/master
```R
> library(data.table)
data.table 1.14.99 IN DEVELOPMENT built 2024-01-15 20:22:22 UTC; dgealow using 48 threads (see ?getDTthreads).  Latest news: r-datatable.com
> library(DataTableAwareTestPkg)
> dt <- data.table(foo = 1:3, bar = 4:6)
> df <- data.frame(foo = 1:3, bar = 4:6)
> get_first_row_of_datatable(dt)
     foo   bar
   <int> <int>
1:     1     4
> get_first_row_of_datatable(df)
  foo
1   1
2   2
3   3
> get_first_column_of_dataframe(dt)
     foo
   <int>
1:     1
2:     2
3:     3
> get_first_column_of_dataframe(df)
  foo
1   1
2   2
3   3
```

dvg-p4:
Yeah it looks like the issue with `selfrefok()` failing with datatable-unaware indexing is a preexisting problem, I can replicate it on master with

```r
> dt = data.table(foo = 1:3, bar = 4:6, baz = 7:9)
> res <- `[.data.frame`(dt, 2:3, 1:2)
> res
     foo   bar
   <int> <int>
1:     2     5
2:     3     6
> data.table:::selfrefok(res)
[1] 0
```

Or, with the https://github.com/dvg-p4/DataTableAwareTestPkg/tree/default-unaware branch of my test package:

```r
> dt = data.table(foo = 1:3, bar = 4:6, baz = 7:9)
> res = get_first_column_of_dataframe(dt)
> res
     foo
   <int>
1:     1
2:     2
3:     3
> data.table:::selfrefok(res)
[1] 0
```

dvg-p4:
I was wondering, since I don't see a milestone assigned to this issue or PR, should I expect this to make it into the next release if it gets approved, or would it probably not be until the next patch after that? or next major release? Trying to plan out whether I ought to spend time on a workaround for our package if this won't be incorporated into a DT release for a while.

inst/tests/tests.Rraw
```diff
@@ -18260,3 +18260,13 @@ test(2243.54, dt[, .I[j], x]$V1, c(1L, 3L), output="GForce TRUE")
 test(2243.55, dt[, .I[i], x]$V1, 1:4, output="GForce FALSE")
 test(2243.56, dt[, .I[1:2], x]$V1, 1:4, output="GForce FALSE")
 options(old)
+
+# allow specifying datatable-unaware behavior, #5654 / #5655
+dt = data.table(foo = 1:3, bar = 4:6)
+.datatable.aware = TRUE
+test(2244.1, dt[1], data.table(foo = 1L, bar = 4L)) # Single index should be interpreted as row
+.datatable.aware = FALSE
+# selfrefok() fails on unaware-indexed datatables; the copy() here sidesteps that issue.
```
MichaelChirico, line 18269: Hmm, if selfrefok() becomes `FALSE` after returning from `cedta()`, isn't that a broader issue?

Now users who declare themselves unaware of data.table are getting obscure internal warnings from data.table....

WDYT? Worth addressing in this PR? Or kick the can until a further issue is reported downstream?

MichaelChirico:
> I was wondering, since I don't see a milestone assigned to this issue or PR, should I expect this to make it into the next release if it gets approved, or would it probably not be until the next patch after that? or next major release? Trying to plan out whether I ought to spend time on a workaround for our package if this won't be incorporated into a DT release for a while.

Right, 1.15.0 is sealed up already.

Going forward, "major" release will be the default, i.e., 1.16.0 is the next release. I believe the timeline we discussed is every 6 months for such releases.

inst/tests/tests.Rraw
dvg-p4: I don't *think* it actually propagates to a user-visible warning in normal use--it's specifically data.table::test() that checks selfrefok() and throws an error if not.

MichaelChirico: OK, let's wait & see then.

dvg-p4:
> Would you mind resolving the conflicts? I set the merge target to the upcoming dev version of R.

Sorry for the delay---done

dvg-p4:
@MichaelChirico I'm guessing from the timestamps that this being closed was an automatic Github action due to you deleting the branch it was targeting--if so could you re-open and re-target whatever it should be targeting now?

MichaelChirico:
@dvg-p4 sorry about that, usually it auto-rebases to `master`. I think it failed to do so because the branch is on your fork... "Reopen and comment" is greyed out on my UI, so I'm not sure how to proceed...

dvg-p4:
Oh looks like you got it haha. I can take a look at those conflicts tomorrow.

MichaelChirico (approved):
Thanks for your patience! Should be ready to merge once CI passes.

tdhock:
hi @MichaelChirico about your comment "delete placeholder 1-15-99 branch" I would suggest to by default not delete branches of PRs that have been merged, please, unless you have a strong reason to do so. Leaving the branch (not deleting) is preferable for performance testing, where we want to actually run some code on old branches, and compare with current code, to make sure that current code is indeed faster then previous/regression code.

ben-schwen:
> hi @MichaelChirico about your comment "delete placeholder 1-15-99 branch" I would suggest to by default not delete branches of PRs that have been merged, please, unless you have a strong reason to do so. Leaving the branch (not deleting) is preferable for performance testing, where we want to actually run some code on old branches, and compare with current code, to make sure that current code is indeed faster then previous/regression code.

Is there any upside of checking out a branch for performance testing over checking out a single a commit? Both even work with `git checkout branchName/commitID`.

tdhock:
actually, yes for performance testing we do checkout single commits.
but when the branch is deleted, the commit can no longer be checked out.

MichaelChirico:
> I would suggest to by default not delete branches of PRs that have been merged

> when the branch is deleted, the commits in that branch can no longer be checked out.

FWIW `1-15-99` is not a good example since it was rebase-merged and all of the PR commits now survive to `master`.

But anyway, deleting merged branches is the norm in every repo I've seen -- that's why GitHub lets us do so in one click. For your concern about checking out PR commits, I don't think your statement is accurate, e.g. taking Jan's #5889 (branch deleted circa Jan 12):

```shell
git checkout master
git checkout -b tmp
git cherry-pick cca5b7912aeee1826792d983f531f36736c9fb91
git cherry-pick 14555b2b3ef3498242110bfd64c3fc8570f4f7aa
```

Those SHAs are from https://github.com/Rdatatable/data.table/pull/5889/commits. I.e., I don't have any trouble applying specific commits even though the branch is deleted.

jangorecki:
If you want to benchmark PRs then merge commit is fine, if you want to benchmark individual commits inside PR then you cannot delete branches. IMO first should be enough.

tdhock:
anyway, if you delete a branch, we can always restore it later thanks to github.

PR #5660, opened 2023-06-15
Allow early exit from check for eval/evalq in cedta

MichaelChirico:
Done in the browser+untested, please take a second look :)


jangorecki:
Let's come back to this during 1.15.99

MichaelChirico:
Ready for review.

jangorecki (approved):

PR #5673, opened 2023-08-10
add examples to insert new column as first/last column

### LINKED ISSUE #5668: Add a new parameter `after=` to `set()` to control column position
It is currently not possible to specify where to insert the new column when using `set(DT, j="ID", value=values)`. A very common operation is to insert a column at the first position or after a specific column. Thus, I propose the following:

```R
set(DT, j="ID", value = values, after=0) # Inserts at the first position, all other column shifted to the right.
set(DT, j="ID", value = values, after=3) # Inserts at position 4 (or the last position if ncol(DT) < 3
set(DT, j="ID", value = values, after="Name") # Inserts after the existing column "Name" (if not such column, error).
```

Click preview tab ^^^ above!

By continuing to file this new issue / feature request, I confirm I have :
1. searched the [live NEWS file](https://github.com/Rdatatable/data.table/blob/master/NEWS.md) to see if it has been fixed in dev already. If so, I tried the [latest dev version](https://github.com/Rdatatable/data.table/wiki/Installation#windows).
2. looked at the titles of all the issues in the [current milestones](https://github.com/Rdatatable/data.table/milestones) and am aware of those. (Adding new information to existing issues is very helpful and appreciated.)
3. [searched all issues](https://github.com/Rdatatable/data.table/issues) (i.e. not in a milestone yet) for similar issues to mine and will include links to them explaining why mine is different.
4. searched on [Stack Overflow's data.table tag](http://stackoverflow.com/questions/tagged/data.table) and there is nothing similar there.
5. read the [Support](https://github.com/Rdatatable/data.table/wiki/Support) and [Contributing](https://github.com/Rdatatable/data.table/wiki/Contributing) guides.
6. please don't tag your issue with text in the title; project members will add the appropriate tags later.

#### Thanks! Please remove the text above and include the two items below.

`#` `Output of sessionInfo()`

ben-schwen:
Follow up to #4691
Closes #5668


ben-schwen:
Added this to 1.15.0 since follow up to #4691 which is 1.15.0

MichaelChirico (approved):
Great!

PR #5681, opened 2023-09-01
checkPoTools found bad translations

MichaelChirico:
cc @hongyuanjia to review, especially the 2nd translation where I was kind of stuck


po/R-zh_CN.po
```diff
@@ -769,7 +769,7 @@ msgid ""
 "Assigning to 0 row subset of %d rows"
 msgstr ""
 "没有找到匹配 i 的行。无法增加新的列所以无法运算 RHS of :=\n"
-"指定一个 0 行的子集"
+"指定一个 0 行的子集从 %d 行"
```
hongyuanjia, line 772: I think `从 %d 行中分配 0 行` should be sufficient. `subset` here can be ignored during translation.

po/R-zh_CN.po
```diff
@@ -206,7 +206,7 @@ msgstr ""
 "列才可以联结"
 
 msgid "%s has same type (%s) as %s. No coercion needed."
-msgstr "%s 有 %s 的类型。不需要强制转换。"
+msgstr "%1$s 有 %3$s 的类型 (%2$s)。不需要强制转换。"
```
hongyuanjia, line 209: I would suggest `%1$s 与 %3$s 为相同类型 (%2$s)。不需要强制转换。`

hongyuanjia (approved):

MichaelChirico:
po-only, merging

PR #5690, opened 2023-09-20
setup pkgup GH actions workflow

jangorecki:
This uses https://github.com/jangorecki/pkgup template for simple and yet artifact rich GH Action release pipeline. Requires setting of Pages to Github Actions, rather than gh-pages branch, as most likely is set now..
Once merged we should update CI documentation in `.ci/README.md`. I feel it is already outdated.

### DIFF
--- a/.github/workflows/pkgup.yaml
+++ b/.github/workflows/pkgup.yaml
@@ -51,7 +51,6 @@ jobs:
           mv doc public/doc
           cp -r --parents library/*/{html,help,doc,demo,DESCRIPTION,README,NEWS,README.md,NEWS.md} public 2>/dev/null || :
           sed -i 's|"/doc/html/|"/data.table/doc/html/|g' public/library/data.table/doc/index.html 2>/dev/null || :
-          echo '<!DOCTYPE html>\n<html><head><meta http-equiv="refresh" content="0; url=library/data.table/html/00Index.html"><title>HTTP redirect</title></head><body></body></html>' > public/index.html
       - name: repo
         if: github.ref == 'refs/heads/master'
         run: |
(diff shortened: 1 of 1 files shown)


MichaelChirico:
@jangorecki @tdhock any thoughts on reviewing this now to make sure all the pending PRs have proper CI?

I lean yes but it's slightly less trivial +"no code" than the other PRs I'm reviewing now, want to double-check first.

MichaelChirico:
Also @jangorecki PTAL at #5632 which is closely related.

jangorecki:
If we leave GLCI then we need to move our pkgdown website to GHCI as well, but that is not a blocker. Having dev pkg being released somewhere that update_dev_pkg can use it is somehow  a blocker, so we can update url already in patch release 1.14.10..
We can merge this PR for the moment, and after 1.14.10 work on adding pkgdown, and setting redirection from r-datatable.com to GH page.

jangorecki:
In settings-pages I have switched github pages to be deployed from GH actions rather than gh-pages git branch.

.github/workflows/pkgup.yaml
MichaelChirico, line 1: At a glance one might confuse pkgdown/pkgup as being closely related, maybe rename this workflow e.g. 'upload-pkg-binary'?

MichaelChirico (approved):
change to 'pkgup' naming is not required, Approving now if you want to skip that. O/w will re-approve on update.

.github/workflows/pkgup.yaml
jangorecki: this is pkg sources only, and pkgup stands for putting a package up. I will document CI once we finish releasing to CRAN, so it will be clear.

PR #5700, opened 2023-10-04
custom data.frame classes redirected to as.data.frame before as.data.table

### LINKED ISSUE #5699: as.data.table should remove extra attributes from extended data.frames
Spotted when investigating #5698
```r
x = iris
class(x) = c("tbl","data.frame")
attr(x, "t1") = "a"
y = as.data.table(x)
attr(y, "t1")
#[1] "a"       ## should be NULL
class(y)
#[1] "data.table" "data.frame"
```
We should assume creator of custom class provides correct data.frame method that clear any extra attributes
```r
as.data.frame.tbl = function(x) {
  attr(x, "t1") = NULL
  class(x) = "data.frame"
  x
}
z = as.data.frame(x)
attr(z, "t1")
#NULL
```
So then consistency can be retained, as described in #5698

jangorecki:
closes #5699


R/as.data.table.R
```diff
@@ -214,6 +214,7 @@ as.data.table.list = function(x,
 }
 
 as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
+  if (!identical(class(x), "data.frame")) return(as.data.frame(x))
```
MichaelChirico, line 217: this returns a data.frame?

inst/tests/tests.Rraw
```diff
@@ -18111,3 +18111,15 @@ test(2238.9, NA %notin% c(1:5, NA), FALSE)
...
+x = data.frame(a=c(1,5,3), b=c(2,4,6))
+class(x) = c("tbl","data.frame")
+attr(x, "t1") = "a"
+as.data.frame.tbl = function(x) {
+  attr(x, "t1") = NULL
+  class(x) = "data.frame"
+  x
+}
+y = as.data.table(x)
+test(2240.1, attr(y, "t1", TRUE), NULL)
```
jangorecki, line 18125: this test happens to fail from R CMD check but not when doing interactive test from cc()

MichaelChirico:
Hi @jangorecki would you mind summarizing what needs to be done to get this ready for review? That way someone else could have a go at sewing it up if you don't get to it first. Thanks!

jangorecki:
Examine why test fails, if still fails.

inst/tests/tests.Rraw
MichaelChirico: I am missing something about S3 dispatch mechanics, but for whatever reason when run in batch dispatch to`as.data.frame.tbl()` doesn't happen.

It probably needs to be properly registered (e.g. with `registerS3method()`), but I'm a bit loath to mess with the session S3 dispatch table like that.

I amended the test to only check that the `as.data.table()` and `as.data.frame()` behavior with respect to that attribute agree.

MichaelChirico (approved):
I am OK to merge with Jan's ack (and a NEWS entry).

PR #5720, opened 2023-11-03
update_dev_pkg uses GH for R repo (in master branch as well)

jangorecki:
It is the same what https://github.com/Rdatatable/data.table/pull/5719 but this one merges to master


MichaelChirico (approved):

jangorecki:
We may first want to merge https://github.com/Rdatatable/data.table/pull/5690 then test change here and merge

PR #5724, opened 2023-11-05
update CRAN maintainer

### LINKED ISSUE #5710: update maintainer field in dsc file
As discussed in #5676 we will have to chose a person for a role of new maintainer on CRAN.

jangorecki:
closes #5710


TysonStanley (approved):
My information looks correct, should be good to go

DESCRIPTION
```diff
@@ -12,7 +12,8 @@ BugReports: https://github.com/Rdatatable/data.table/issues
 VignetteBuilder: knitr
 ByteCompile: TRUE
 Authors@R: c(
-  person("Matt","Dowle",           role=c("aut","cre"), email="mattjdowle@gmail.com"),
+  person("Tyson","Barrett",        role=c("aut","cre"), email="t.barrett88@gmail.com"),
+  person("Matt","Dowle",           role="aut",          email="mattjdowle@gmail.com"),
```
tdhock, line 16: I'm ok for going ahead on master branch with Tyson as CRAN maintainer, for next major feature release (non-patch).
Also for patch releases, I was assuming that Matt will continue doing that, until governance is decided.
did anybody email Matt about this? (I did not) Matt has to give his OK to CRAN maintainers on email, in order for this to be updated on CRAN.

tdhock (approved):

DESCRIPTION
jangorecki: 1.14.10 branch already has that change: https://github.com/Rdatatable/data.table/compare/1.14.8...hotfix1.14.10
so we may need to revert that change possibly.

From what you are suggesting it most likely means that Matt would have to still make 2 pkg submissions to CRAN: 1.14.10 and then 1.15.0. Because the latter one would be moving maintainership to Tyson.

PR #5725, opened 2023-11-05
Minimal fix for broken shift() test

MichaelChirico:


inst/tests/tests.Rraw
```diff
@@ -15150,7 +15150,8 @@ z = c(1:3) + c(3:1)*1i
 test(2067.1, shift(z), c(NA, z[1:2]))
 test(2067.2, shift(z, type = 'lead'), c(z[2:3], NA))
 test(2067.3, shift(z, fill = 1i), c(1i, z[1:2]))
-test(2067.4, shift(list(z, 1:3)), list(c(NA, z[1:2]), c(NA, 1:2)))
+## NB: as.complex(NA) here covers for a change in behavior in R on whether Im(as.complex(NA)) is '0' or 'NA' (#5695)
+test(2067.4, shift(list(z, 1:3)), list(c(as.complex(NA), z[1:2]), c(NA, 1:2)))
```
jangorecki, line 15154: Based on that it seems that as.complex(NA) produces different NA complex that c() function?

jangorecki: @mmaechler shouldn't c() automatically coerce NA to what as.complex(NA) gives?

MichaelChirico: Yes, I'm in the process of writing to r-devel about this. I need to double-check the documentation before I hit send :)

MichaelChirico: Basically it looks like `c(NA, 0+1i)` is run as `c(NA_complex_, 0+1i)` as opposed to `c(as.complex(NA), 0+1i)`.

jangorecki: I think the point is that we would like shift to produce NA_complex_ rather what as.complex(NA) gives

jangorecki: https://stat.ethz.ch/R-manual/R-devel/library/base/html/complex.html

jangorecki: For a patch release, that would be enough actually, fix is already in master. Assuming that behavior of as.complex(NA) will not change in R-devel then we should be fine with this commit. Then we also don't have to port it to master when shift produces NA_complex_ already

jangorecki (approved):

jangorecki:
Oh and please update new file while you are on this PR

inst/tests/tests.Rraw
MichaelChirico: Linking the r-devel thread for reference: https://stat.ethz.ch/pipermail/r-devel/2023-November/082997.html

MichaelChirico: > fix is already in master.

Yes, but I was thinking the minimal commit for patch release is actually quite large / affects several code aspects. Safer to make this change for this quick release rather than risk introducing new issues.

MichaelChirico:
> new file

? Going to assume NEWS

jangorecki (approved):

Issue #5727, opened 2023-11-06
revive/rework CI


jangorecki:
Shortly after (or even for the) 1.15.0 we should try revive/rework our CI, so PRs can be well tested, for different platforms, R versions, and code coverage.

Tests should cover:
- testing r-devel, r-release, r-oldrel
- testing stated R version
- testing linux, mac, windows
- testing vanilla R (having only base R packages installed)
- strict 0 NOTEs from R CMD check in a tailored job (using flags like `_R_CHECK_CRAN_INCOMING_TARBALL_THRESHOLD_`)
- testing for compilation warnings

Artifacts should cover:
- source package
- binary packages
- R repository
- documentation in pkgdown
- documentation base R html
- ~~docker images~~ anyone uses them? installing data.table is so easy and fast that I don't think we save a lot of effort

tdhock:
I do not use data.table docker images (I did not know we had them) but I have used R-hub docker images in the past for testing other R packages.

eddelbuettel:
> docker images anyone uses them? 

If you start from 'ubuntu: latest' as an image you then stick in container like r2u which helps once either your package is 'expensive' or all the test dependencies are.  The following is from a GH Actions yaml file of mine

```
    runs-on: ubuntu-latest
    container:
      image: rocker/r2u:latest
```

and you can of course stick other containers (rhub, ...) in as needed too.

jangorecki:
We already publish docker images from GLCI but because DT doesn't have dependencies and compiles fast there is not that much need like many other pkgs

eddelbuettel:
Yup, all good, and it's a good spot to be in.  But sometimes you want to, say, test reverse dependencies and getting them quick and easily is a bonus.  

This bonus switch to an existing container in an action is sometimes I learned only this year (or last?) after sometimes having done the requivalent of a partial container setup in an action.  So given the question above it felt like somethings worth sharing.  Feel free to delete the posts if you think they are noise.

Looking for to data.table releases.  It's been a while.  So thanks for getting back to it!

jangorecki:
Bringing back CI to life uncovered couple issues

https://github.com/Rdatatable/data.table/issues/5484
https://github.com/Rdatatable/data.table/issues/5760
https://github.com/Rdatatable/data.table/issues/5760
https://github.com/Rdatatable/data.table/pull/5749

for now workarounds in CI will be either extra `apt-get install` or `allow_failures: true`

jangorecki:
[...]

PR #5728, opened 2023-11-06
run workflow only on pushes to master

jangorecki:
eliminate failed job for each new PR


MichaelChirico:
I assume this is temporary while waiting for #5727 right?

MichaelChirico (approved):

jangorecki:
Not necessarily, could be kept later on as well as an alternative bleeding edge devel version that may not be passing every test job but basic r-release linux one (if we keep GLCI) Unless we will be running out of CI minutes it is always a good backup.

Issue #5745, opened 2023-11-20
trim suggested dependencies


jangorecki:
Complete installation of data.table + suggested dependencies results in following packages being installed in a library. While data.table doesn't have any hard dependencies, the soft ones could be trimmed to keep that to the minimum. Not sure how easy, and at what cost, would be possible.
```dcf
Suggests: bit64 (>= 4.0.0), bit (>= 4.0.4), curl, R.utils, xts, nanotime, zoo (>= 1.8-1), yaml, knitr, rmarkdown, markdown, parallel
```
including recursive deps, total 38 packages
```
 [1] "bit"         "R.oo"        "R.methodsS3" "zoo"         "bit64"      
 [6] "RcppCCTZ"    "Rcpp"        "RcppDate"    "evaluate"    "highr"      
[11] "xfun"        "yaml"        "bslib"       "fontawesome" "htmltools"  
[16] "jquerylib"   "jsonlite"    "knitr"       "stringr"     "tinytex"    
[21] "base64enc"   "cachem"      "memoise"     "mime"        "rlang"      
[26] "sass"        "digest"      "ellipsis"    "fastmap"     "cli"        
[31] "glue"        "lifecycle"   "magrittr"    "stringi"     "vctrs"      
[36] "fs"          "R6"          "rappdirs"   
```
and few of them are quite heavy.
`curl` package also imposes extra OS dependency , devel version of `libcurl`.

Broken down by each Suggested dep:
```
$bit64
[1] "bit"

$bit
character(0)

$curl
character(0)

$R.utils
[1] "R.oo"        "R.methodsS3"

$xts
[1] "zoo"

$nanotime
[1] "bit64"    "RcppCCTZ" "zoo"      "Rcpp"     "RcppDate" "bit"     

$zoo
character(0)

$yaml
character(0)

$knitr
[1] "evaluate" "highr"    "xfun"     "yaml"    

$rmarkdown
 [1] "bslib"       "evaluate"    "fontawesome" "htmltools"   "jquerylib"  
 [6] "jsonlite"    "knitr"       "stringr"     "tinytex"     "xfun"       
[11] "yaml"        "base64enc"   "cachem"      "memoise"     "mime"       
[16] "rlang"       "sass"        "digest"      "ellipsis"    "fastmap"    
[21] "highr"       "cli"         "glue"        "lifecycle"   "magrittr"   
[26] "stringi"     "vctrs"       "fs"          "R6"          "rappdirs" 
```

[...]

PR #5749, opened 2023-11-22
Remove `curl` from suggests

### LINKED ISSUE #1686: [bug] fread does not work with https url behind a proxy
Hi, 

I recently encounter a problem with https url to read a file with `fread`. I had the following error

``` r
#> Error in curl::curl_download(input, tt, mode = "wb", quiet = !showProgress): Timeout was reached
```

The problem is I am behind a proxy and `curl_download` does not seem to import current proxy setting on windows for IE. And I do not find a way to configure `curl` proxy setting before `fread`

However, with `download.file` function, downloading the file with https url works perfectly without configuring anything.
Looking at `fread`,  `curl::curl_download` is used for secure url whereas for non secure, `download.file` is used ? 

``` r
if (!is_secureurl(input)) {
      download.file(input, tt, mode = "wb", quiet = !showProgress)
    }
    else {
      if (!requireNamespace("curl", quietly = TRUE)) 
        stop("Input URL requires https:// connection for which fread() requires 'curl' package, but cannot be found. Please install the package using 'install.packages()'.")
      curl::curl_download(input, tt, mode = "wb", quiet = !showProgress)
    }
```

Could it be possible that now R handles https url with `download.file`, the suggested `curl` library is no longer need and would solves this problem? 

---

For those who could try to reproduce behin a proxy : 

``` r
url <- "https://d37djvu3ytnwxt.cloudfront.net/asset-v1:MITx+15.071x_3+1T2016+type@asset+block/songs.csv"
DT <- data.table::fread(url, verbose = T)
#> Error in curl::curl_download(input, tt, mode = "wb", quiet = !showProgress): Timeout was reached
```

and my `sessionInfo()`

``` r
#> R version 3.2.4 Revised (2016-03-16 r70336)
#> Running under: Windows 7 x64 (build 7601) Service Pack 1
#> 
#> locale:
#> [1] LC_COLLATE=French_France.1252  LC_CTYPE=French_France.1252   
#> [3] LC_MONETARY=French_France.1252 LC_NUMERIC=C                  
#> [5] LC_TIME=French_France.1252    
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.9.6
#> 
#>  [1] clipr_0.2.0       magrittr_1.5      formatR_1.3      
#>  [4] htmltools_0.3.5   tools_3.2.4       curl_0.9.7       
#>  [7] Rcpp_0.12.4       stringi_1.0-1     rmarkdown_0.9.5  
#> [10] knitr_1.12.3      stringr_1.0.0     digest_0.6.9     
#> [13] reprex_0.0.0.9001 chron_2.3-47      evaluate_0.8.3
```

ben-schwen:
Closes #1686 
Towards #5745 

- [x] fix
- [ ] ~tests~ (had to tests for downloading from secure urls before besides)
- [x] news 
- [ ] ~translations~ (done in release cycle)

Regarding translations: (@MichaelChirico)
I've never noticed these so far, but are they adapted with every release?


MichaelChirico:
> I've never noticed these so far, but are they adapted with every release?

Would help if you're more specific, but if you're talking about anything in the po/ or inst/po folders, those are updated as part of the (normal) release process -- no need to touch in this PR.

jangorecki:
AFAIU all what is needed still here is to escape test of that functionality for R 3.1.0.

ben-schwen:
We have never added a test to really download from an `url` when enabling this, so should we add one now? Must ensure then that the file is up and reachable.

jangorecki:
easy way to figure one which test needs to be escaped is to just leave the error for all versions of R, run test, and then you will see which tests failed, and those escape for R < 3.2.2

ben-schwen:
Regarding translations, we have this in our `.pot`:
https://github.com/Rdatatable/data.table/blob/5e8ca96f7a9505704289f008f8a940ed2eefe83e/po/R-data.table.pot#L1255-L1256

jangorecki:
No need to worry about translations during dev cycle. It is part of release cycle.

jangorecki (approved):

Issue #5718, opened 2023-11-03
update_dev_pkg not serving recent dev pkg anymore


jangorecki:
gitlab is not currently releasing dev pkg therefore update_dev_pkg not serving recent devel. We could point it to GH R repo if we have one. Could be very simple, only source pkg. It just needs some CI updates on GH, which could be covered by setting up pkgup in #5690 
Although website will still be a bit out of date.
that could possibly still qualify for 1.14.10 milestone.

PR #5758, opened 2023-11-24
GLCI fixes

jangorecki:
towards #5727

- removed 3.4.4, 3.5.0 test jobs
- updated urls to windows R binaries
- cleanup old comments
- remove docker builds


jangorecki:
if we merge it to master before hotfix release, we can revert #5719 because #5718 will be solved at the root

tdhock (approved):
hi Jan it looks ok to me, but I am not really familiar with this, so not sure if I would be able to catch any issues.

.gitlab-ci.yml
```diff
@@ -24,83 +24,87 @@ stages:
...
+    - mkdir -p bus/$CI_JOB_NAME/cran/src/contrib
+    - mv $(ls -1t data.table_*.tar.gz | head -n 1) bus/$CI_JOB_NAME/cran/src/contrib/.
     - Rscript -e 'tools::write_PACKAGES(contrib.url("bus/build/cran"), fields="Revision", addFiles=TRUE)'
-    - rm bus/$CI_BUILD_NAME/cran/src/contrib/PACKAGES.rds ## fallback to PACKAGES dcf so available.packages:3.4.4 works
   <<: *artifacts
 
+## install deps aliases
 .test-install-deps: &install-deps
-  - Rscript -e 'source(".ci/ci.R"); install.packages(dcf.dependencies("DESCRIPTION", which="most"), quiet=TRUE)'
+  - Rscript -e 'source(".ci/ci.R"); install.packages(dcf.dependencies("DESCRIPTION", which="all"), repos=file.path("file:", normalizePath("bus/mirror-packages/cran", mustWork=FALSE)), quiet=TRUE)'
```
MichaelChirico, line 69: I would start (e.g.) `.ci/scripts/` and add `.ci/scripts/test-install-deps.R` with:

```r
source(".ci/ci.R")
install.packages(
  dcf.dependencies("DESCRIPTION", which="all"),
  repos=file.path("file:", normalizePath("bus/mirror-packages/cran", mustWork=FALSE)),
  quiet=TRUE
)
```

This will be more readable and avoid duplication (and hence double maintenance/risk of getting out of sync) across `test-intall-deps` and `test-install-deps-win` jobs.

MichaelChirico: Or, we can find a way to put `RUN_SCRIPT=if(win) Rscript.exe else Rscript` into env and unify the jobs.

.gitlab-ci.yml
```diff
@@ -138,37 +127,47 @@ build: ## build data.table sources as tar.gz archive
 #  tags:
 #    - macosx
 
-test-rel-lin: ## most comprehensive tests, force all suggests, also integration tests, using gcc -O3 -flto -fno-common -Wunused-result
+## most comprehensive tests
+# force all suggests
+# flags: gcc -O3 -flto -fno-common -Wunused-result
+# tests for compilation warnings
+# measure memory usage during tests
+test-rel-lin:
```
MichaelChirico, line 135: `lin` is `linux` here? I would just write out `linux` I don't see much benefit to saving 2 characters

.gitlab-ci.yml
jangorecki: Quoting arguments from shell also differs

jangorecki: > I would start (e.g.) `.ci/scripts/` and add `.ci/scripts/test-install-deps.R` with:
> 
> ```r
> source(".ci/ci.R")
> install.packages(
>   dcf.dependencies("DESCRIPTION", which="all"),
>   repos=file.path("file:", normalizePath("bus/mirror-packages/cran", mustWork=FALSE)),
>   quiet=TRUE
> )
> ```
> 
> This will be more readable and avoid duplication (and hence double maintenance/risk of getting out of sync) across `test-intall-deps` and `test-install-deps-win` jobs.

Unless we will need distinction for some reason, then we ended up having two scripts :)

MichaelChirico:
> hi Jan it looks ok to me, but I am not really familiar with this, so not sure if I would be able to catch any issues.

Ditto, basically LGTM but I don't follow 100%. It's just CI though, issues should be easy to detect/fix later, effects on package code are second-order only (e.g. if a mistake in CI eventually lets a mistake pass through PR review unnoticed)

.gitlab-ci.yml
jangorecki: Yes, it is 3 letters so it aligns better when printing to win and mac. Maybe lnx then?

jangorecki:
These CI runs on master so after PR being merged already. Design was made that ghci gives fast feedback and let us do merge, while glci tests thoroughly and we don't have to wait for it.

.gitlab-ci.yml
MichaelChirico: > Quoting arguments from shell also differs

Can't we unify if there's no env inside `"`? Anyway, if the shell command is just `Rscript -e "source(...)"` it will still be easier to glance & see it looks ~OK.

> Unless we will need distinction for some reason, then we ended up having two scripts :)

Looks like not the case yet, we can do some `.Platform`/`Sys.info()` switching to handle the small differences. Still almost all of the logic between the two jobs is the same, I find it more confusing needing to read long one-line scripts without syntax highlighting.

.gitlab-ci.yml
MichaelChirico: That would be less ambiguous, yes

PR #5759, opened 2023-11-24
escape zlib dependent tests

### LINKED ISSUE #5753: pkg check fails when zlib1g-dev missing
`zlib1g-dev` is not required to use data.table, as this is an optional dependency required for compression in fwrite.

Note that `apt-get install --no-install-recommends r-base` does not install zlib due to `--no-install-recommends`.

It happens that currently zlib is required during pkg tests. We could therefore escape related tests when zlib1g-dev is not available.

Testing with
```sh
sudo docker run -it --rm debian:testing

apt-get update \
  && apt-get install -y --no-install-recommends \
    make \
    gcc \
    ca-certificates \
    wget \
    libc6-dev
echo "deb http://http.debian.net/debian sid main" > /etc/apt/sources.list.d/debian-unstable.list
apt-get update && apt-get install -y --no-install-recommends r-base

wget https://rdatatable.github.io/data.table/src/contrib/data.table_1.14.9.tar.gz
_R_CHECK_FORCE_SUGGESTS_=false R CMD check data.table_1.14.9.tar.gz --no-manual --ignore-vignettes --no-examples
#Compression in fwrite uses zlib library...
```

jangorecki:
closes #5753

### DIFF
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -115,3 +115,5 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
   invisible()
 }
 
+nozlib = function() identical(.Call(Cdt_zlib_version), "zlib header files were not found when data.table was compiled")
+
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -4,7 +4,6 @@ Title: Extension of `data.frame`
 Depends: R (>= 3.1.0)
 Imports: methods
 Suggests: bit64 (>= 4.0.0), bit (>= 4.0.4), curl, R.utils, xts, nanotime, zoo (>= 1.8-1), yaml, knitr, rmarkdown, markdown
-SystemRequirements: zlib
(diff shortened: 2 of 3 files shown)


jangorecki:
this is needed to avoid failure in https://gitlab.com/Rdatatable/data.table/-/jobs/5613277998
safe to merge as we only change tests

R/fwrite.R
```diff
@@ -115,3 +115,5 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
   invisible()
 }
 
+nozlib = function() identical(.Call(Cdt_zlib_version), "zlib header files were not found when data.table was compiled")
```
MichaelChirico, line 118: NB: this may fail under Chinese (or other locales eventually) session since this string is passed through `gettext()`.

https://github.com/Rdatatable/data.table/blob/50a3dc3744a8039b2215ad6a348bf2aea1e5b3d7/src/utils.c#L376-L378

Safer to test on the `NOZLIB` branch, possibly on `grepl("ZLIB_VERSION", ., fixed = TRUE)` in case that message is later marked for translation?

Alternatively we could make a companion function to `dt_zlib_version` like `dt_has_zlib` that just returns `TRUE`/`FALSE`.

Issue #5768, opened 2023-11-26
-Wformat clang/gcc warnings


jangorecki:
New compilation warnings on CRAN
https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/data.table-00check.html
Same warnings, but less, on GCC
```
checking whether package ‘data.table’ can be installed ... [57s/138s] WARNING
Found the following significant warnings:
  assign.c:470:83: warning: format specifies type 'int' but the argument has type 'R_xlen_t' (aka 'long') [-Wformat]
  assign.c:518:22: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  assign.c:794:96: warning: format specifies type 'int' but the argument has type 'unsigned long' [-Wformat]
  forder.c:107:30: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:121:19: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:255:27: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:270:23: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:272:21: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:283:18: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:284:23: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:304:25: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:322:17: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:340:14: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:404:3: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:426:7: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:428:7: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:430:7: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:443:5: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:445:5: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  forder.c:447:5: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
[...]

PR #5769, opened 2023-11-26
Fix -Wformat issues

MichaelChirico:
Part of #5768

Still missing are any changes for `-Wformat-security`.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -468,7 +468,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       // strong error message for now.
     else if (TRUELENGTH(names) != oldtncol)
       // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768
-      error(_("Internal error: selfrefnames is ok but tl names ["PRId64"] != tl ["PRId64"]"), (long long)TRUELENGTH(names), (long long)oldtncol);  // # nocov
+      error(_("Internal error: selfrefnames is ok but tl names [%"PRId64"] != tl [%"PRId64"]"), (long long)TRUELENGTH(names), (long long)oldtncol);  // # nocov
     SETLENGTH(dt, oldncol+LENGTH(newcolnames));
     SETLENGTH(names, oldncol+LENGTH(newcolnames));
     for (i=0; i<LENGTH(newcolnames); i++)
--- a/src/assign.c
+++ b/src/assign.c
@@ -467,8 +467,8 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       // Can growVector at this point easily enough, but it shouldn't happen in first place so leave it as
       // strong error message for now.
     else if (TRUELENGTH(names) != oldtncol)
-      // Use (long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings
-      error(_("Internal error: selfrefnames is ok but tl names [%ld] != tl [%ld]"), (long)TRUELENGTH(names), (long)oldtncol);  // # nocov
+      // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768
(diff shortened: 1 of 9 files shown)


MichaelChirico:
(PS it would be great if we could add these compiler warnings to CI to prevent regressions on this in the future)

MichaelChirico:
I poked around in r-devel a bit to see how they handle `R_xlen_t` in `printf()` formatted messages, but didn't see anything.

Therefore the approach in 0d57fd8f9f793755a93d4e268cda7dec88f506b3 seems like the best expedient solution for now -- on most installations it will be a no-op (`long` -> `long` cast), and wherever `R_xlen_t` is just `int`, it'll cast `int` -> `long`.

jangorecki:
> (PS it would be great if we could add these compiler warnings to CI to prevent regressions on this in the future)

If you would happen to that docker that reproduces that please share. It's probably just compiler version and flags but will make it easier to add.

jangorecki:
I added `-Wformat -Wformat-security` to `~/.R/Makevars` and used gcc 13.2.0 (same as [CRAN's gcc](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/data.table-00check.html)) but I am unable to reproduce those warnings from CRAN. Must be some extra flag then.

This is what we just talked about with @eddelbuettel about lack of CRAN reproducibility...

MichaelChirico:
h/t @HughParsonage for this r-package-devel thread getting directly to the issue:

https://stat.ethz.ch/pipermail/r-package-devel/2023q4/010123.html

That points out R itself has a new macro `R_PRIdXLEN_T`:

https://github.com/r-devel/r-svn/blob/a9ded8dd775dc0aa75b26a5c88c11983d6d884f9/src/include/Rinternals.h#L75-L79

But we're miles away from being able to depend on such a recent addition to R. We could define the macro ourselves; it will be fragile but seems like no better choice at the moment if we want to avoid `long long`.

jangorecki:
From the thread you linked it sounds `long long` is good enough, will work across different R versions.

eddelbuettel:
I eyeballed the thread too but didn't like it much. I prefer to a) not depend on a particular version or b) make the code more complicated via an `#if ...` for a given architecture when c) a simple cast takes care of it.  So I am with @jangorecki here: a simple cast does it. I used `unsigned int` to convert a (C++_ `sizeof()` that made R on Windows bark. All quiet now.

MichaelChirico:
thanks all. was only wary of `long long` given the comments in the CRAN release notes. Agree it's best otherwise.

jangorecki:
Considering it was commented out in CRAN release we should be good

jangorecki:
testing PR via
```
sudo docker run -it --rm rocker/r-devel /bin/bash
mkdir -p ~/.R
echo 'CFLAGS=-Wformat -Wformat-security' > ~/.R/Makevars
echo 'CXXFLAGS=-Wformat -Wformat-security' >> ~/.R/Makevars

## optional
wget https://cran.r-project.org/src/contrib/data.table_1.14.8.tar.gz
RD CMD INSTALL data.table_1.14.8.tar.gz
# warnings reproduced

## PR
wget https://github.com/Rdatatable/data.table/archive/refs/heads/format-warnings.zip
unzip format-warnings.zip
RD CMD build data.table-format-warnings --no-build-vignettes
RD CMD INSTALL data.table_1.14.9.tar.gz
```

src/assign.c
```diff
@@ -468,7 +468,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       // strong error message for now.
     else if (TRUELENGTH(names) != oldtncol)
       // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768, PRId64 didnt work
-      error(_("Internal error: selfrefnames is ok but tl names [%ld] != tl [%ld]"), TRUELENGTH(names), oldtncol);  // # nocov
+      error(_("Internal error: selfrefnames is ok but tl names [%ld] != tl [%d]"), TRUELENGTH(names), oldtncol);  // # nocov
```
MichaelChirico, line 471: oh, interesting, does that mean TRUELENGTH(names) and TRUELENGTH(dt) have different types??

I noticed the Wformat error only hit this line once & was wondering why.

jangorecki: It appears to be like that. I didn't try to understand that.

PR #5770, opened 2023-11-27
proper fix for #5753 to make zlib fully optional

jangorecki:
follow up of #5759, this time proper fix by adding new internal function


MichaelChirico (approved):
perfect thank you!

Issue #5790, opened 2023-12-04
-Wformat still warns on Windows


jangorecki:
confirmed on master branch
```
* checking whether package 'data.table' can be installed ... WARNING
Found the following significant warnings:
  assign.c:474:15: warning: format '%ld' expects argument of type 'long int', but argument 2 has type 'R_xlen_t' {aka 'long long int'} [-Wformat=]
  assign.c:812:21: warning: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'long long unsigned int' [-Wformat=]
  fread.c:149:19: warning: format '%d' expects argument of type 'int', but argument 2 has type 'DWORD' {aka 'long unsigned int'} [-Wformat=]
  gsumm.c:619:32: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  gsumm.c:676:17: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  gsumm.c:1119:19: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  init.c:163:33: warning: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'long long unsigned int' [-Wformat=]
  init.c:180:39: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  init.c:181:39: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  init.c:183:39: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  init.c:184:59: warning: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'long long unsigned int' [-Wformat=]
  init.c:185:47: warning: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'long long unsigned int' [-Wformat=]
  init.c:185:47: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
  init.c:186:38: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
[...]

PR #5791, opened 2023-12-05
Use %zu for sizeof() formats

MichaelChirico:
Part of #5790


tdhock:
yay ci is working again thanks @jangorecki

PR #5792, opened 2023-12-05
Fix two "unusual" format warnings

MichaelChirico:
Part of #5790

I don't see clear answers about `DWORD` but it seems like `%ld` will work:

https://copyprogramming.com/howto/what-is-the-format-specifier-for-dword-c

For `TRUELENGTH()` we wrote the comment about casting to `long long` but there is no cast. So trying to actually do the cast.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -471,7 +471,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       // strong error message for now.
     else if (TRUELENGTH(names) != oldtncol)
       // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768, PRId64 didnt work
-      error(_("Internal error: selfrefnames is ok but tl names [%ld] != tl [%d]"), TRUELENGTH(names), oldtncol);  // # nocov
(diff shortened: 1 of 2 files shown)


src/assign.c
```diff
@@ -471,7 +471,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       // strong error message for now.
     else if (TRUELENGTH(names) != oldtncol)
       // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768, PRId64 didnt work
-      error(_("Internal error: selfrefnames is ok but tl names [%ld] != tl [%d]"), TRUELENGTH(names), oldtncol);  // # nocov
+      error(_("Internal error: selfrefnames is ok but tl names [%ld] != tl [%d]"), (long long)TRUELENGTH(names), oldtncol);  // # nocov
```
ben-schwen, line 474: Shouldn't it be %lld then?

MichaelChirico:
```suggestion
      error(_("Internal error: selfrefnames is ok but tl names [%lld] != tl [%d]"), (long long)TRUELENGTH(names), oldtncol);  // # nocov
```

jangorecki:
testing here https://gitlab.com/Rdatatable/data.table/-/pipelines/1096182458

ben-schwen:
> testing here https://gitlab.com/Rdatatable/data.table/-/pipelines/1096182458

Well the warnings about fread.c:149:19 and assign.c:474:15 are gone!

PR #5795, opened 2023-12-06
Add a simple dockerfile for extending basic CI image to do dev stuff (first: git)

MichaelChirico:
(might need to iterate few times, sorry for the PR noise)


jangorecki:
In the next iteration put directory to .Rbuildignore

MichaelChirico:
Did 4fcf81c1e4ca002985d2d9862a10c699255bf7db not work?

jangorecki:
Probably did. My comment was based on a yesterdays state. Sorry

jangorecki:
@MichaelChirico any idea if path `./.devcontainer` is configurable in some way? we do have `./.dev` directory where such directory would fit very well. To declutter directory.

jangorecki:
? @MichaelChirico

MichaelChirico:
Ah thanks. I lost track of where you posted that Q.

Not according to the official spec:

https://containers.dev/implementors/spec/

MichaelChirico:
I also see this https://github.com/devcontainers/spec/issues/159

jangorecki:
Him, then looks we are stuck with two different dirs in a root dir related exclusively to .github.

Issue #5748, opened 2023-11-21
tests failures due to complex NA


jangorecki:
We are getting more tests failures due to changed behavior of complex NA.
https://cran.r-project.org/web/checks/check_results_data.table.html
```
 9 errors out of 9917. Search tests/tests.Rraw for test numbers: 1658.58, 1743.231, 1962.0473, 2040.1, 2040.2, 2060.6, 2060.601, 2060.602, 2060.604.
```

not fixed on master

```sh
sudo docker run -it --rm registry.gitlab.com/jangorecki/dockerfiles/r-devel /bin/bash
wget https://rdatatable.github.io/data.table/src/contrib/data.table_1.14.9.tar.gz
R CMD INSTALL data.table_1.14.9.tar.gz
```

Looks like complex NA support has to be fixed in:
- [ ] `fwrite`: 1658.58: https://github.com/Rdatatable/data.table/blob/50a3dc3744a8039b2215ad6a348bf2aea1e5b3d7/src/fwrite.c#L304
```r
library(data.table)
test = data.table:::test
DT = data.table(a=1:3, b=list(1:4, c(3.14, 100e10), c(3i,4i,5i)))
DT[ , b := c(1i, -1-1i, NA)]
test(1658.58, fwrite(DT), output='a,b\n1,0\\+1i\n2,-1-1i\n3,$')
#Test 1658.58 did not produce correct output:
#Expected: <<a,b\n1,0\\+1i\n2,-1-1i\n3,$>>
#Observed: <<a,b\n1,0+1i\n2,-1-1i\n3,+0i>>
```
- [ ] `fread`: 1743.231
```r
test(1743.231, fread("a,b,c\n2,1,4i", colClasses=list(complex="c", integer=2L), drop="a"), data.table(b=1L, c="4i")) ## + warning
```
- [ ] `fcoalesce`: 2060.6, 2060.601, 2060.602, 2060.604
- [ ] `shift` (again?): 2067.4

jangorecki:
@MichaelChirico you think you are able to look into this? some of complex NA handling were added by you.

I think we can adapt fwrite/fread/fcoalesce/shift to new 4.4 behavior and branch those tests to run only 4.4+ so we don't need to test pre 4.4 behavior which is now changing.

MichaelChirico:
This will be tough in the near future -- I don't have a dev environment handy where I can easily test C issues.

I will explore if GitHub Codespaces is suitable, been meaning to try that out.

[...]

PR #5796, opened 2023-12-06
Fix tests for complex

MichaelChirico:
Part of #5748 (fixed for master, not hotfix)

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -9935,7 +9935,7 @@ test(1658.56, fwrite(data.table(exp(1) - pi*1i)), output='2.718[0-9]*-3.141[0-9]
 ## formerly 1658.46
 DT = data.table(a=1:3, b=list(1:4, c(3.14, 100e10), c(3i,4i,5i)))
 test(1658.57, fwrite(DT), output='0+3i|0+4i|0+5i')
-DT[ , b := c(1i, -1-1i, complex(real = NA, imaginary = NA))]
+DT[ , b := c(1i, -1-1i, NA_complex_)]
 test(1658.58, fwrite(DT), output='a,b\n1,0\\+1i\n2,-1-1i\n3,$')
 
 # more coverage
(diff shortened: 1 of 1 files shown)


inst/tests/tests.Rraw
```diff
@@ -10964,7 +10964,7 @@ test(1743.217, sapply(fread("a,b,c,d,e,f\na,b,c,d,e,f", colClasses = list(factor
 test(1743.218, sapply(fread("a,b,c,d,e,f\na,b,c,d,e,f", colClasses = list(factor = c(1, 2, 4), factor = 3), select = c(5, 4, 2, 3)), class), y = c(e = "character", d = "factor", b = "factor", c = "factor"))
 
 test(1743.22, fread("a,b,c\n1999/01/01,2,f", colClasses=list(Date=1L), drop="a"), data.table(b=2L, c="f"))
-test(1743.231, fread("a,b,c\n2,1,4i", colClasses=list(complex="c", integer=2L), drop="a"), data.table(b=1L, c="4i"),
+test(1743.231, fread("a,b,c\n2,1,4j", colClasses=list(complex="c", integer=2L), drop="a"), data.table(b=1L, c="4j"),
```
MichaelChirico, line 10967: This is actually a substantial improvement in r-devel -- previously it seems `as.complex("4i")` would fail to coerce which seems wrong. Better to try and coerce to complex something that's not actually complex, to ensure the warning continues.

cc original author @HughParsonage in case I'm missing something behind the intent of this test.

HughParsonage: I think Matt added that specific test, though I added the tests around it, which looks to have been a back and forth that observed but did not mention the behaviour of as.complex("4i"):

https://github.com/Rdatatable/data.table/pull/2545/commits/c6e2b72dc2ab2211b8333094345d3bc36a84175f#diff-e3243f3780ce7d303c3317f73945310bfc37e45d193568246246aca20e3270ae

jangorecki (approved):

Issue #5715, opened 2023-11-02
version number conventions


tdhock:
as a part of #5676 I would like to discuss conventions we could adopt for version numbers.
As I understand the current system, even number in the third component/place is used for CRAN release (1.14.2 or 1.14.4), and odd number for github/dev code (1.14.1 or 1.14.3). 

I would suggest keeping that convention and augmenting it so that patch/hotfix releases (no new features, just fixes to satisfy CRAN requests) can be distinguished from regular/feature releases, and so that versions with new features will have larger version numbers than patch versions. Currently 1.14.9 is dev and 1.14.8 is CRAN, but if we submit a new patch version to CRAN that will be 1.14.10 which can be confusing because that number is greater than the current dev (1.14.9), which actually has more features. To fix that I would suggest increasing the second/minor component of the version number every time there are new features introduced in dev/github master. So current github master could become 1.15.1, and until that is released on CRAN as 1.15.2, we could submit any patch releases that CRAN requests as 1.14.even (1.14.10 or 1.14.12). Then patch releases would always have smaller version numbers than any github or cran versions with new features.

Additionally I would suggest bumping the major version to 2 to indicate the new governance of the project. So after new governance gets adopted, github master would become 2.0.1, and first regular CRAN release using new governance would be 2.0.2 (with patches 2.0.4, 2.0.6, etc). After that first regular CRAN release, github master would become 2.1.1, which would eventually be released to CRAN as 2.1.2 (with patches 2.1.4, 2.1.6, etc).

The only issue I see with this proposal is the lack of x.y.0 versions, which could be confusing to people who may be expecting that?

jangorecki:
> Currently 1.14.9 is dev and 1.14.8 is CRAN, but if we submit a new patch version to CRAN that will be 1.14.10 which can be confusing because that number is greater than the current dev (1.14.9)

After such patch release to CRAN we would normally increment master to 1.14.11 to avoid this problem. That is how it used to work, and there was never problem about it.

> To fix that

Then no fix is needed.

According to what you are proposing + trying to have (a non patch) release each 6 month, we will be then releasing 1.15, 1.16, etc. quite frequently, right? I recall discussion about this kind of versioning but somehow we haven't landed there. That generally make sense, but your suggestion would skip 1.15.0 and release 1.15.2 instead, which I think is not ideal. We can always increment to 1.15.0 from 1.14.51 (or whatever) just before release to CRAN, that is how it used to work.

[...]

PR #5803, opened 2023-12-06
follow version number conventions #5715

jangorecki:
I think procedure for a patch release added in this PR has to be corrected. @TysonStanley please have a look at that during release and commit directly to this branch.

Possibly we need different procedure for:
- patch release (x.y.2) from non-patch release (x.y.0)
- patch release (x.y.4) from previous patch release (x.y.2)


TysonStanley:
@jangorecki will definitely review this. I don't think this impacts today's patch release but would be good to have in CRAN_Release.cmd going forward. Can edit over the next day or so.

jangorecki:
Sure. Also before next cran submission everything will be already little bit more clearer.

jangorecki:
closes #5808

PR #5805, opened 2023-12-06
Transpose(dt) allows to return list without promoting elements to maxtype

### LINKED ISSUE #5639: `data.table::transpose` applied to lists changes element types
Applying `data.table::transpose` to a data.table that contains numerical and string values will convert everything to strings. By comparison `purrr::transpose` preserves the original types by creating a list of lists. Note that it is very easy to convert a list of lists to a list of vectors (that is, get the output of `data.table::transpose` from the output of `purrr::transpose`) but very hard the other way. 

```R
library(data.table)
dt <- data.table(num=runif(3), char = letters[1:3])
str(data.table::transpose(dt)) # I want a data.table, so fine.
# Classes ‘data.table’ and 'data.frame':	2 obs. of  3 variables:
#  $ V1: chr  "0.21412930265069" "a"
#  $ V2: chr  "0.0449177662376314" "b"
#  $ V3: chr  "0.299894759198651" "c"
str(data.table::transpose(as.list(dt))) # I want a list, so it should be a list of lists.
# List of 3
#  $ : chr [1:2] "0.21412930265069" "a"
#  $ : chr [1:2] "0.0449177662376314" "b"
#  $ : chr [1:2] "0.299894759198651" "c"
library(purrr)
str(purrr::transpose(dt)) # Exactly right!
# List of 3
#  $ :List of 2
#   ..$ num : num 0.214
#   ..$ char: chr "a"
#  $ :List of 2
#   ..$ num : num 0.0449
#   ..$ char: chr "b"
#  $ :List of 2
#   ..$ num : num 0.3
#   ..$ char: chr "c"
```

This is a very frequent operation, which is currently very hard to do with `data.table`. See 

https://stackoverflow.com/questions/3492379/data-frame-rows-to-a-list
https://stackoverflow.com/questions/27784076/r-data-table-to-list-of-rows-better-ways
https://stackoverflow.com/questions/41005943/turn-all-data-table-rows-into-a-list-r
https://stackoverflow.com/questions/56901903/convert-a-data-table-to-list-of-rows-that-are-data-tables-and-apply-a-function-t

(and more)

ben-schwen:
Closes #5639

Transpose now also supports transposing list columns.

- [x] Add feature
- [x] Docs
- [x] Tests (including fill)
- [x] News
- [x] Handle Factors consistently (do not copy levels)


tdhock:
Hi Ben thanks for sharing, looks interesting. I have never used transpose, so I was wondering what is the typical use case? with lists? The man page says
```
     This is particularly useful in tasks that require splitting a
     character column and assigning each part to a separate column.
     This operation is quite common enough that a function tstrsplit
     is exported.
```
I have used tstrsplit, so I guess I understand the use case with character, but I wonder if you could add an illustrative example, which would show how transpose is different / more flexible / useful with lists ?

ben-schwen:
The main use cases I have in my mind is when you receive a rowwise format or want to `lapply` some function rowwise. Not promoting to maximum type but returning lists would be favorable there to avoid additional casting.

```r
dt = data.table(name=c("Anna", "Bob"), age=c(30,20))
fun = function(x) sprintf("Hi %d year old %s", x[[2]], x[[1]])
transpose(dt, list.cols = TRUE)[, lapply(.SD, fun)]
#                     V1                 V2
#                 <char>             <char>
# 1: Hi 30 year old Anna Hi 20 year old Bob
```

MichaelChirico:
> Hi Ben thanks for sharing, looks interesting. I have never used transpose, so I was wondering what is the typical use case? with lists? The man page says
> 
> ```
>      This is particularly useful in tasks that require splitting a
>      character column and assigning each part to a separate column.
>      This operation is quite common enough that a function tstrsplit
>      is exported.
> ```
> 
> I have used tstrsplit, so I guess I understand the use case with character, but I wonder if you could add an illustrative example, which would show how transpose is different / more flexible / useful with lists ?

transpose() is the workhorse behind tstrsplit, tstrsplit is "just" strsplit()|>transpose().

but there are many other times we wind up with a list() of rows that we want as columns instead. base R code looks like ``lapply(seq_along(l), \(j) sapply(l, `[[`, j))``.  

transpose() is not only much cleaner but much faster and handles type conversion and fills ragged data

NEWS.md
```diff
@@ -293,6 +293,8 @@
 
 41. `tables()` is faster by default by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). `mb=` now accepts any function which accepts a `data.table` and returns a higher and better estimate of its size in bytes, albeit more slowly; e.g. `mb = utils::object.size`.
 
+42. `transpose` gains `list.cols=` argument, [#5639](https://github.com/Rdatatable/data.table/issues/5639). This enables to return output with list columns and avoids promoting type to maximum type. `Factor` columns are still promoted to `character` to avoid copying levels. Thans to @MLopez-Ibanez for the request, and Benjamin Schwendinger for the PR.
```
jangorecki, line 296: Does copying levels is a problem actually? Is there any piece of code that justify this exception?Underneath we copy pointers to strings on global character cache.

jangorecki: @ben-schwen that sounds much more reasonable, I would mention that in NEWS. Also this behavior deserves mention in manual page.

tdhock:
> transpose() is the workhorse behind tstrsplit, tstrsplit is "just" strsplit()|>transpose().
> 
> but there are many other times we wind up with a list() of rows that we want as columns instead. base R code looks like `` lapply(seq_along(l), \(j) sapply(l, `[[`, j)) ``.
> 
> transpose() is not only much cleaner but much faster and handles type conversion and fills ragged data

Can an example of this usage (and new usage) be please added to transpose.Rd in this PR?
(it would help me at least)

ben-schwen:
FWIW here are benchmarks version vs base solution

```r
l = list(sample(1e5), sample(letters, 1e5, TRUE))
bench::mark(
  base = lapply(seq(length(l[[1]])), function(x) lapply(l, `[[`, x)),
  dt = transpose(l, list.cols=TRUE)
)
# # A tibble: 2 × 13
#   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result           memory                  time            gc               
#   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>           <list>                  <list>          <list>           
# 1 base        277.2ms  282.9ms      3.53  786.52KB     5.30     2     3      566ms <list [100,000]> <Rprofmem [19,593 × 3]> <bench_tm [2]>  <tibble [2 × 3]> 
# 2 dt           15.7ms   25.6ms     31.2     3.05MB     7.81    16     4      512ms <list [100,000]> <Rprofmem [4 × 3]>      <bench_tm [16]> <tibble [16 × 3]>
```

jangorecki:
@ben-schwen base was run only two times while PR was run 16 times. Therefore I would look for a "max" statistic (missing) rather than min or median (presented in your post). You may find this post relevant: https://github.com/jangorecki/rollbench/pull/1#issuecomment-1826838156

man/transpose.Rd
```diff
@@ -6,14 +6,15 @@
...
-transpose(l, fill=NA, ignore.empty=FALSE, keep.names=NULL, make.names=NULL)
+transpose(l, fill=NA, ignore.empty=FALSE, keep.names=NULL, make.names=NULL, list.cols=FALSE)
 }
 \arguments{
   \item{l}{ A list, data.frame or data.table. }
   \item{fill}{ Default is \code{NA}. It is used to fill shorter list elements so as to return each element of the transposed result of equal lengths. }
   \item{ignore.empty}{Default is \code{FALSE}. \code{TRUE} will ignore length-0 list elements.}
   \item{keep.names}{The name of the first column in the result containing the names of the input; e.g. \code{keep.names="rn"}. By default \code{NULL} and the names of the input are discarded.}
   \item{make.names}{The name or number of a column in the input to use as names of the output; e.g. \code{make.names="rn"}. By default \code{NULL} and default names are given to the output columns.}
+  \item{list.cols}{Default is \code{FALSE}. \code{TRUE} will avoid promoting types and return columns of type \code{list} instead.}
```
jangorecki, line 17: we need factor behavior mentioned here

ben-schwen: Done. I also update the timings above including "max"

jangorecki (approved):

tdhock:
if you want to see how time/memory vary with N, please try using atime https://github.com/tdhock/atime/issues/15

image

above shows that both base R and dt are linear time, different by constant factors.

image

above shows the data size N which both can handle, given 1 second time limit or 1000kb memory limit.

NEWS.md
```diff
@@ -341,6 +341,8 @@
 
 41. `tables()` is faster by default by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). `mb=` now accepts any function which accepts a `data.table` and returns a higher and better estimate of its size in bytes, albeit more slowly; e.g. `mb = utils::object.size`.
 
+42. `transpose` gains `list.cols=` argument, [#5639](https://github.com/Rdatatable/data.table/issues/5639). This enables to return output with list columns and avoids promoting type to maximum type. The only exception are `factor` columns which are promoted to `character` so `transpose(, list.cols=TRUE)` and `transpose(, list.cols=FALSE)` keep same behavior for `factor`. Thanks to @MLopez-Ibanez for the request, and Benjamin Schwendinger for the PR.
```
MichaelChirico, line 344:
```suggestion
42. `transpose` gains `list.cols=` argument, [#5639](https://github.com/Rdatatable/data.table/issues/5639). Use this to return output with list columns and avoids type promotion (an exception is `factor` columns which are promoted to `character` for consistency between `list.cols=TRUE` and `list.cols=FALSE`). This is convenient for creating a row-major representation of a table. Thanks to @MLopez-Ibanez for the request, and Benjamin Schwendinger for the PR.
```

inst/tests/tests.Rraw
```diff
@@ -6921,10 +6921,23 @@ ll = sapply(ll, paste, collapse=",")
 test(1477.07, transpose(strsplit(ll, ",", fixed=TRUE)), tstrsplit(ll, ",", fixed=TRUE))
 test(1477.08, transpose(1:5), error="l must be a list")
 test(1477.09, transpose(list(as.complex(c(1, 1+5i)))), error="Unsupported column type")
-test(1477.10, transpose(list(list(1:5))), error="Item 1 of list input is")
+test(1477.10, transpose(list(x~y)), error="Item 1 of list input is") # changed behavior since we also support now list columns #5805
```
MichaelChirico, line 6924: In 5 years, this comment will make no sense. I would just remove it. It would have been useful as a PR comment or note in the PR description.

```suggestion
test(1477.10, transpose(list(x~y)), error="Item 1 of list input is")
```

inst/tests/tests.Rraw
```diff
@@ -6921,10 +6921,23 @@ ll = sapply(ll, paste, collapse=",")
...
 test(1477.11, transpose(as.list(1:5), fill=1:2), error="fill must be a length 1 vector")
 test(1477.12, transpose(as.list(1:5), ignore.empty=NA), error="ignore.empty should be logical TRUE/FALSE")
 test(1477.13, transpose(list()), list())
+# return list columns #5639
+la = list(as.list(1:3), list("a","b","c"))
+lb = list(list(1L,"a"), list(2L,"b"), list(3L,"c"))
+test(1477.21, transpose(list(1:3, c("a","b","c")), list.cols=TRUE), lb)
```
MichaelChirico, line 6931: What's happening with the test numbering? Skips 13->21 and then 28->41?

inst/tests/tests.Rraw
```diff
@@ -6921,10 +6921,23 @@ ll = sapply(ll, paste, collapse=",")
...
+test(1477.22, transpose(list(1:3, c("a","b","c")), list.cols=FALSE), lapply(lb, as.character))
```
MichaelChirico, line 6932: nit: `unlist()` is probably closer to what actually happens? since `unlist()` is more commonly associated with type promotion...

```suggestion
test(1477.22, transpose(list(1:3, c("a","b","c")), list.cols=FALSE), lapply(lb, unlist))
```

man/transpose.Rd
```diff
@@ -6,14 +6,15 @@
...
-transpose(l, fill=NA, ignore.empty=FALSE, keep.names=NULL, make.names=NULL)
+transpose(l, fill=NA, ignore.empty=FALSE, keep.names=NULL, make.names=NULL, list.cols=FALSE)
 }
 \arguments{
   \item{l}{ A list, data.frame or data.table. }
   \item{fill}{ Default is \code{NA}. It is used to fill shorter list elements so as to return each element of the transposed result of equal lengths. }
   \item{ignore.empty}{Default is \code{FALSE}. \code{TRUE} will ignore length-0 list elements.}
   \item{keep.names}{The name of the first column in the result containing the names of the input; e.g. \code{keep.names="rn"}. By default \code{NULL} and the names of the input are discarded.}
   \item{make.names}{The name or number of a column in the input to use as names of the output; e.g. \code{make.names="rn"}. By default \code{NULL} and default names are given to the output columns.}
+  \item{list.cols}{Default is \code{FALSE}. \code{TRUE} will avoid promoting types and return columns of type \code{list} instead. \code{factor} will always be casted to \code{character}.}
```
MichaelChirico, line 17:
```suggestion
  \item{list.cols}{Default is \code{FALSE}. \code{TRUE} will avoid promoting types and return columns of type \code{list} instead. \code{factor} will always be cast to \code{character}.}
```

src/transpose.c
```diff
@@ -18,23 +18,28 @@ SEXP transpose(SEXP l, SEXP fill, SEXP ignoreArg, SEXP keepNamesArg) {
   if (length(fill) != 1)
     error(_("fill must be a length 1 vector, such as the default NA"));
   R_len_t ln = LENGTH(l);
+  if (!isLogical(listColsArg) || LOGICAL(listColsArg)[0]==NA_LOGICAL)
```
MichaelChirico, line 21:
```suggestion
  if (!IS_TRUE_OR_FALSE(listColsArg))
```

src/transpose.c
```diff
@@ -18,23 +18,28 @@ SEXP transpose(SEXP l, SEXP fill, SEXP ignoreArg, SEXP keepNamesArg) {
...
     if (len>maxlen) maxlen=len;
     zerolen += (len==0);
     SEXPTYPE type = TYPEOF(li);
     if (isFactor(li)) type=STRSXP;
     if (type>maxtype) maxtype=type;
   }
+  if (listCol) maxtype=VECSXP; // need to keep preprocessing for zerolen
   fill = PROTECT(coerceVector(fill, maxtype)); nprotect++;
 
+
```
MichaelChirico, line 42: rm extra blank line

src/transpose.c
```diff
@@ -54,17 +59,10 @@ SEXP transpose(SEXP l, SEXP fill, SEXP ignoreArg, SEXP keepNamesArg) {
     const int len = length(li);
     if (ignore && len==0) continue;
     if (TYPEOF(li) != maxtype) {
-      li = PROTECT(isFactor(li) ? asCharacterFactor(li) : coerceVector(li, maxtype));
+      li = PROTECT(isFactor(li) ? (listCol ? coerceVector(asCharacterFactor(li), VECSXP) : asCharacterFactor(li)) : coerceVector(li, maxtype));
     } else PROTECT(li); // extra PROTECT just to help rchk by avoiding two counter variables
     switch (maxtype) {
-    case LGLSXP : {
```
MichaelChirico, line 60: any reason for this? if we already have the `LGLSXP` branch written out I would keep it... "cheating" and collapsing the `LGLSXP` and `INTSXP` branches can be forgiven but shouldn't be encouraged IMO

MichaelChirico:
`list.cols=TRUE` also works for `data.table` (/frame) input, though it's not clear from the current PR if that's intentional.

As I mentioned in the updated NEWS, transposing a table to get it in row-major form is indeed quite common & useful. This is _especially_ useful for I/O with non-column-major data structures (as is common for streaming applications which receive one row at a time). As noted in the linked issue, the workarounds were quite tedious before, after this PR, we can quite easily just `transpose(<row-major>) |> lapply(unlist) |> setDT()` to get a nice data.table out of something we receive, and `transpose(<column-major>, list.cols=TRUE)` to convert an R table into something to pass along.

In short, this feature is quite great and something I "didn't know I needed" all along!

tdhock:
would be great to add vignette documentation for this use case "As noted in the linked issue, the workarounds were quite tedious before, after this PR, we can quite easily just transpose(<row-major>) |> lapply(unlist) |> setDT() to get a nice data.table out of something we receive, and transpose(<column-major>, list.cols=TRUE) to convert an R table into something to pass along"

inst/tests/tests.Rraw
ben-schwen: Left some gaps on purpose for future fillings, e.g. BASIC style line numbering

src/transpose.c
MichaelChirico: See this and some other comments marked resolved without being addressed, is there maybe an unpublished commit or review?

inst/tests/tests.Rraw
MichaelChirico: thanks... I think so far our approach has been to use sequential numbering, then if we really want to jimmy in some new tests alongside old ones, "split them" e.g. 1477.16 becomes 1477.160, then the related tests are 1477.161, 1477.162, ...

With big gaps a future reader might waste time trying to find out what happened to the "removed" tests.

MichaelChirico:
> would be great to add vignette documentation for this use case "As noted in the linked issue, the workarounds were quite tedious before, after this PR, we can quite easily just transpose() |> lapply(unlist) |> setDT() to get a nice data.table out of something we receive, and transpose(, list.cols=TRUE) to convert an R table into something to pass along"

Do you have a specific vignette in mind? I don't see any mention of `transpose()` in the current vignettes. Maybe what we really need is a datatable-utils vignette demonstrating `transpose()` as well as `fcase()`, `fifelse()`, ... odds-and-end-functions we have

MichaelChirico (approved):

PR #5807, opened 2023-12-07
take default threads from CRAN env var

### LINKED ISSUE #5658: Best practice for using max 2 cores on CRAN?
CRAN has a policy that limits the number of cores that can be used during package checks, and they have recently implemented a mechanism for checking compliance. Upon recent CRAN submission of one of my R packages which imports data.table, I got the following rejection message from CRAN:
```
Flavor: r-devel-linux-x86_64-debian-gcc
Check: examples, Result: NOTE
  Examples with CPU (user + system) or elapsed time > 5s
                         user system elapsed
  aum_line_search      12.349  0.322   1.935
  aum_line_search_grid 10.033  0.308   1.781
  Examples with CPU time > 2.5 times elapsed time
                         user system elapsed ratio
  aum_line_search      12.349  0.322   1.935 6.548
  aum_line_search_grid 10.033  0.308   1.781 5.806
  aum_diffs_penalty     4.730  0.169   1.635 2.996
```
These messages can be suppressed by adding data.table::setDTthreads(1) at the start of each example.
Is there another/recommended way to avoid using too many cores when checking a CRAN package which uses data.table? The default number of threads in data.table is max cores/2 which is over the CRAN limit (max 2 cores I believe).
Also is there documentation about this somewhere? Probably would be good to mention in the datatable-importing vignette.

tdhock:
Closes #5658  
(already closed but actually still an issue)

DT currently takes default number of threads from env vars OMP_THREAD_LIMIT or OMP_NUM_THREADS, as defined in src/openmp-utils.c

CRAN does not allow using more than two CPUs during checks. Previously it had set OMP_THREAD_LIMIT=2 see https://github.com/Rdatatable/data.table/issues/5658#issuecomment-1598000921

Recently CRAN stopped setting OMP_THREAD_LIMIT, but it does set `_R_CHECK_EXAMPLE_TIMING_CPU_TO_ELAPSED_THRESHOLD_=2.5` which I propose in this PR to examine, round down, and use as a default for the DT number of threads. 

Currently packages which use DT in tests/examples/vignettes need to run `setDTthreads(2)` to get CRAN checks to pass. After merging this PR, the default would satisfy CRAN (package which use DT would no longer need to run `setDTthreads` in tests/examples/vignettes).


tanho63:
image

fwiw one of my pkgs got this email from cran today after implementing similar. interested to see how it turns out if you do end up giving it a go

jangorecki:
Thanks for letting us know. I think we will get sample reply. Those, and any others vars, should not be used to detect if pkg check runs on CRAN or not, so it is rather not surprising.
Having your example I think we can abandon this draft PR.

tdhock:
thanks for the info, that is helpful.

Issue #5840, opened 2023-12-20
Set datatable.logical01 to TRUE by default


MichaelChirico:
https://github.com/Rdatatable/data.table/blob/6c1fd839e0e0257d13f07975ead0fe6fdfee2f61/NEWS.1.md?plain=1#L1181

The plan to change the default has been in place since v1.11.0 (2018). Let's include it in 1.15.0 as a good opportunity to include such changes.

jangorecki:
deprecations IMO should be added at the beginning of dev cycle so there is more time to spot potential issues and overall impact, changingn default shortly before release is just extra risk.

MichaelChirico:
The reason is maintenance burden. Keeping around all these pending deprecations needlessly complicates the codebase. The change was discussed almost 6 years ago -- more than enough time.

There are already dozens or more downstreams that have begun setting `logical01=TRUE` manually -- clearly there's demand for this value:

https://github.com/search?q=%2Flogical01%5Cs*%3D%5Cs*TRUE%2F+lang%3AR&type=code

There are also (fewer) downstreams that have begun setting `logical01=FALSE` despite the default, possibly anticipating the planned deprecation:

https://github.com/search?q=%2Flogical01%5Cs*%3D%5Cs*FALSE%2F+lang%3AR&type=code

@tdhock's nightly revdep runs are meant to help mitigate the risk. I don't see a reason not to merge to `master` first. If the new error footprint turns out to be very large we can revert before release.

jangorecki:
Unfortunately revdeps are just a tiny fraction of code that uses data.table

MichaelChirico:
All .0 releases are the right time to include potentially breaking changes. Waiting another 6 months/year doesn't make sense to me if we really want to make the change.

revdeps aren't everything, but will give a good idea of what's broken anyway. We have seen time and again that even obscure behavior changes can get caught by revdeps. Something non-trivial as this will surely turn up something.

That said, the changes needed for #5843 didn't leave me with the most confidence that this feature is ready to be the default yet. For example, I'm not sure casting column of single value '1' or '0' to logical is the best behavior. Possibly we should require finding two unique values before casting. But in that case something like "read all files in this folder" may have unpleasant behavior like different column types for different files.

OTOH, maybe that's just an artifact of our tests on 1-row tables, and it's not very practically relevant.

IMO the best next step is to merge the change and see what revdeps turn up. I have a feeling we will be reverting.

tdhock:
[...]

PR #5841, opened 2023-12-20
Progress some planned deprecations

MichaelChirico:
#5840 will be filed separately since it's a new deprecation --> warrants its own PR.

These are simply progressing existing deprecation cycles to their next steps.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -609,5 +609,10 @@
 
 15. Thanks to @ssh352, Václav Tlapák, Cole Miller, András Svraka and Toby Dylan Hocking for reporting and bisecting a significant performance regression in dev. This was fixed before release thanks to a PR by Jan Gorecki, [#5463](https://github.com/Rdatatable/data.table/pull/5463). 
 
+16. `key(x) <- value` is now fully deprecated (from warning to error). Use `setkey()` to set a table's key. We started warning not to use this approach in 2012, with a stronger warning starting in 2019 (1.12.2). This function will be removed in the next release.
+
+17. Argument `logicalAsInt` to `fwrite()` now warns. Use `logical01` instead. We stated the intention to begin removing this option in 2018 (v1.11.0). It will be upgraded to an error in the next release before being removed in the subsequent release.
+
+18. Option `datatable.CJ.names` no longer has any effect, after becoming `TRUE` by default in v1.12.2 (2019). Setting it now gives a warning, which will be dropped in the next release.
 
 # data.table v1.14.10 (Dec 2023) back to v1.10.0 (Dec 2016) has been moved to [NEWS.1.md](https://github.com/Rdatatable/data.table/blob/master/NEWS.1.md)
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -609,5 +609,10 @@
 
 15. Thanks to @ssh352, Václav Tlapák, Cole Miller, András Svraka and Toby Dylan Hocking for reporting and bisecting a significant performance regression in dev. This was fixed before release thanks to a PR by Jan Gorecki, [#5463](https://github.com/Rdatatable/data.table/pull/5463). 
 
+16. `key(x) <- value` is now fully deprecated (from warning to error). Use `setkey()` to set a table's key. We started warning not to use this approach in 2012, with a stronger warning starting in 2019 (1.12.2). This function will be removed in the next release.
+
+17. Argument `logicalAsInt` to `fwrite()` now warns. Use `logical01` instead. We stated the intention to begin removing this option in 2018 (v1.11.0). It will be upgraded to an error in the next release before being removed in the subsequent release.
```
jangorecki, line 614: Possibly introduces revdeps failures, and slows down release by asking them to amend.

jangorecki (commented):
logicalAsInt we could move post release so revdeps (and non cran pkgs) have plenty of time to amend their code. Other looks good

NEWS.md
MichaelChirico: I disagree. Once again we should merge and check revdeps.

Moreover, in this case we can trivially check the entirety of GitHub for downstream usages, of which there are practically 0 -- most of the 41 results here are just {data.table} forks/copies:

https://github.com/search?q=lang%3AR%20%2FlogicalAsInt%5Cs*%3D%2F&type=code

MichaelChirico: There are no (0) CRAN packages using `logicalAsInt`:

https://github.com/search?q=lang%3AR+%2FlogicalAsInt%5Cs*%3D%2F+org%3Acran&type=code

jangorecki (approved):

MichaelChirico:
Refined the GitHub search and confirmed there's only two places using logicalAsInt, neither of which have any activity for >5 years. Nevertheless filed downstream issues about the pending switch.

jangorecki:
If packages were not active for 5 years, then it is more likely they will not manage to update their code within short period of 2 weeks that CRAN gives. Therefore we are risking this change to be reason for archival of two packages. I would say we should try yo avoid doing so. Giving them time of full 1.15.99 cycle doesn't guarantee they will be fixed but at least feels more appropriate, as one of the goals of DT, even officially defined recently, is backward compatibility.

MichaelChirico:
> If packages were not active for 5 years, then it is more likely they will not manage to update their code within short period of 2 weeks that CRAN gives. Therefore we are risking this change to be reason for archival of two packages. I would say we should try to avoid doing so. Giving them time of full 1.15.99 cycle doesn't guarantee they will be fixed but at least feels more appropriate, as one of the goals of DT, even officially defined recently, is backward compatibility.

to be clear these are not CRAN packages, they are only maintained on GitHub.

jangorecki:
It feels much safer then, yet not ideal.

PR #5692, opened 2023-09-23
rolling median

jangorecki:
This branch was forked from rollfuns branch #5682 so till it is merged to master it is best to look at diff here https://github.com/Rdatatable/data.table/compare/rollfuns..rollmedian

----

This is implementation of rolling median based on a novel the algorithm "sort-median" described by @suomela in 2014 in his paper [Median Filtering is Equivalent to Sorting](https://arxiv.org/abs/1406.1717). Thanks to Jukka for the paper and hints on extending the algo. "sort-median" scales very well, not only for size of input vector but also for size of rolling window. Implementation here is extended to address following limitations present in the paper:
- size of input vector must by divisible by window size
- window size must be uneven number

PR #5682, opened 2023-09-02
more rolling functions

jangorecki:
addressing more tasks from #2778

branched from adapt branch #5576 so till its merged to master it is best to look at diff here https://github.com/Rdatatable/data.table/compare/adapt..rollfuns

PR #5576, opened 2023-01-03
adapt uneven time series rolling window

jangorecki:
closes #3241

It is forked from https://github.com/Rdatatable/data.table/tree/frollapply branch, therefore should be rebased to master once #5575 will be merged. Till then it is probably best to look at the diff here: https://github.com/Rdatatable/data.table/compare/frollapply..adapt

PR #5575, opened 2023-01-03
frollapply rewritten, supports by.column=F

jangorecki:
closes #4887

It is forked from https://github.com/Rdatatable/data.table/tree/frollmax branch, therefore should be rebased to master once #5441 will be merged. Till then it is probably best to look at the diff here: https://github.com/Rdatatable/data.table/compare/frollmax..frollapply

PR #5845, opened 2023-12-20
Add options= to test(), convert most Rraw scripts

### LINKED ISSUE #5842: test() should get an options= argument to transparently & temporarily set options for a test
https://github.com/Rdatatable/data.table/blob/6c1fd839e0e0257d13f07975ead0fe6fdfee2f61/R/test.data.table.R#L252

Now, we set options sporadically in the text, e.g.

https://github.com/Rdatatable/data.table/blob/6c1fd839e0e0257d13f07975ead0fe6fdfee2f61/inst/tests/froll.Rraw#L306
https://github.com/Rdatatable/data.table/blob/6c1fd839e0e0257d13f07975ead0fe6fdfee2f61/inst/tests/froll.Rraw#L343

Having the option set 50, 100, or more lines away from where the value is needed harms readability. It's much better to set the options transparently in the call to `test()`.

Proposed API:

```r
test(..., options = list(key1 = "val1", ..., keyn = "valn"))

# impl
test <- function(..., options = NULL) {
  if (!is.null(options)) {
    old <- do.call("options", options)
    on.exit(options(old), add = TRUE)
  }
  ...
}
```

Possibly `tmp_options=` is a better name so as not to clash with `options()`, then we can do `do.call(options, tmp_options)`.

MichaelChirico:
Closes #5842.

Didn't touch tests.Rraw yet as there are 200+ existing calls to `options()`. Better to save that for a future clean-up. Converted the other scripts with a more manageable diff to demonstrate the new usage.


ben-schwen:
Looking great!

The only thing we might have to think about is where to put the `options` argument in a `test(...)` call. Current practice is AFAIC to have `test(num, call, expected)` so `test(num, options=options, call, expected)` differs from that.

MichaelChirico:
> Looking great!
> 
> The only thing we might have to think about is where to put the `options` argument in a `test(...)` call. Current practice is AFAIC to have `test(num, call, expected)` so `test(num, options=options, call, expected)` differs from that.

yes, this is a new style. basically this is a "set up" step as it might be called in other unit testing frameworks. so I think it's important to have it up front for clarity. that doesn't have a parallel among current arguments. I might even put it before the test number, but definitely not after the test outcome parts.

ben-schwen:
> yes, this is a new style. basically this is a "set up" step as it might be called in other unit testing frameworks. so I think it's important to have it up front for clarity. that doesn't have a parallel among current arguments. I might even put it before the test number, but definitely not after the test outcome parts.

I also thought about the setup step, so 2nd place makes definitely sense.

Pls, do not put it before the test number because in searching for tests I often use `test(testnum` if the test number occurs multiple times in `tests.Rraw`

tdhock:
I would like to see a similar `test(testnum, my_computed_value, my_expected_value, env=list(SOME_ENV_VAR="VALUE"))` -- I was recently working on #5807 which involves environment variables for getting default number of threads, and the tests would be much easier to understand/modify with such an enhancement to test.

jangorecki:
@tdhock I understand the use case, but if it is not as common as options, then I think it can be left to R's functions to set variables. Contrary R options are very common in tests, as Michael said, more than 200 times.

MichaelChirico:
> I would like to see a similar `test(testnum, my_computed_value, my_expected_value, env=list(SOME_ENV_VAR="VALUE"))` -- I was recently working on #5807 which involves environment variables for getting default number of threads, and the tests would be much easier to understand/modify with such an enhancement to test.

I think it's perfectly reasonable, but let's save it for a follow-up issue to keep PRs sharp

inst/tests/froll.Rraw
```diff
@@ -322,7 +321,7 @@ test(6000.0672, frollmean(c(1:2,NA,4:10), 4, hasNA=FALSE), c(rep(NA_real_, 6), 5
   "frollmean: processing algo 0 took.*",
   "frollfunR: processing.*took.*"
 ), warning="hasNA=FALSE used but NA.*are present in input, use default hasNA=NA to avoid this warning")
-test(6000.0673, frollmean(c(1:2,NA,4:10), 2, hasNA=FALSE), c(NA, 1.5, NA, NA, 4.5, 5.5, 6.5, 7.5, 8.5, 9.5), output=c(
+test(6000.0673, options = c(datatable.verbose=TRUE), frollmean(c(1:2,NA,4:10), 2, hasNA=FALSE), c(NA, 1.5, NA, NA, 4.5, 5.5, 6.5, 7.5, 8.5, 9.5), output=c(
```
jangorecki, line 324: I think many of those tests testing the output will be changed in another PR, therefore this PR will introduces many non-auto mergable lines. Personally I would prefer, as agreed before 1.14.10, to focus on clearing out PR queue, and then in next release create new PRs. This is to reduce overall work required to deal with conflicts.

MichaelChirico: How about excluding froll.Rraw then? tests.Rraw was excluded already for the same reason. Are any of the other scripts here involved in the major pending PRs?

A very conservative approach is to limit this PR to adding `options=` to `test()`, then do all conversions in follow-up(s).

jangorecki: I don't recall any others but it is up to 3 years old so not difficult to miss some.

Your conservative approach makes sense. We could use this option when adding new tests, but then eventually, when PRs are cleared, deal with the historical tests.

MichaelChirico: I am searching through all branches to find edits to each test file as well.

I think it is OK to ignore PRs in draft status, e.g. https://github.com/Rdatatable/data.table/pull/4517 which touches benchmark.Rraw.

MichaelChirico:
Here is a breakdown of the non-draft PRs touching each test script as of now:

 - inst/tests/tests.Rraw: #7068, #7048, #6916, #6899, #6789, #6624, #6610, #6181, #6136, #6111, #5907, #5441, #5362, #5269, #5168, #5167, #4978, #4868, #4372, #3093
 - inst/tests/froll.Rraw: #5898, #5897, #5896, #5895, #5893, #5892, #5891, #5890, #5692, #5682, #5576, #5575, #5441
 - inst/tests/benchmark.Rraw: #5441
 - inst/tests/other.Rraw: #5441, #5168, #4370
 - inst/tests/mergelist.Rraw #4370

inst/tests/froll.Rraw
MichaelChirico: I'll write a follow-up to #5996 to make the changes in {benchmark,programming,types}.Rraw & leave {froll,nafill,other}.Rraw for 1.17.0.

MichaelChirico:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/5845?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5845** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/5845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6015** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6015?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

MichaelChirico:
This PR is waiting for other PRs to clear before we can merge. In particular here are the PRs to close/merge before we can merge each test file included here:

 - froll.Rraw - #5576 #5575 #5441 #5890 #5891 #5892 #5893 #5895 #5896 #5897 #5682 #5692
 - nafill.Rraw - #4480 #4980
 - other.Rraw - #4480 #5168 #5446

Issue #5680, opened 2023-09-01
revdep simDAG example error


tdhock:
new revdep simDAG has the following error when checking using data.table master, which is not present when checking using data.table from CRAN:
```
* checking examples ... ERROR
Running examples in 'simDAG-Ex.R' failed
The error most likely occurred in:

> ### Name: long2start_stop
> ### Title: Transform a 'data.table' in the long-format to a 'data.table' in
> ###   the start-stop format
> ### Aliases: long2start_stop
> 
> ### ** Examples
> 
> library(simDAG)
> library(data.table)
> 
> # generate example data in long format
> long <- data.table(.id=rep(seq_len(10), each=5),
+                    .time=rep(seq_len(5), 10),
+                    A=c(rep(FALSE, 43), TRUE, TRUE, rep(FALSE, 3), TRUE,
+                            TRUE),
+                    B=FALSE)
> setkey(long, .id, .time)
> 
> # transform to start-stop format
> long2start_stop(data=long, id=".id", time=".time", varying=c("A", "B"))
Error: object 'max_t' not found
Execution halted
```
git bisect says this starts happening after merging https://github.com/Rdatatable/data.table/pull/5245 by @ben-schwen so could you please take a look and tell me if you think this should be fixed in simDAG or in data.table?

tdhock:
if necessary we can report an issue to the revdep here, https://github.com/RobinDenz1/siMDAG/issues

ben-schwen:
I don't have a computer with me but will check on Monday. There are already two follow ups to #5245 in particular #5348 and #5404, so one of those might already fix this

MichaelChirico:
@ben-schwen I'll let you close -- if you think it warrants a new regression test please add one

ben-schwen:
Problem in simDAG is covered by test 2243.01, since the evaluation of arguments type and fill both follow is_constantish.

Issue #5085, opened 2021-07-28
More tests detecting invalid index using selfrefok


mattdowle:
As per https://github.com/Rdatatable/data.table/issues/5042#issuecomment-888552669

- [ ] setkey.R::getindex()
- [ ] `[.data.table` at the start
- [ ] add tests preferably using base R like the test in #4893 so they can be in the main `tests.Rraw` rather than `other.Rraw`

PR #5849, opened 2023-12-21
names<- retains sorting attributes

### LINKED ISSUE #5125: revdep chicane affected
Emailed chicane's maintainer today, below.
3 other revdeps out of 1,111 break that I haven't looked at yet.
If there is a way to make `names()<-` continue to retain the key, then that would be better. On first glance I can't think of a way, but I'll have another go. Probably more `<-` methods are affected too.
In the meantime I figured it was best  to get the change request to the revdep since `setnames()` is better regardless.

--- 

Greetings,

revdep testing of the next data.table release (v1.14.2) shows that chicane will be affected.

I normally submit a PR to your project but in your case I can't find chicane on either R-Forge or GitHub. The DESCRIPTION file has a link to R-Forge, but not the project page on R-Forge, and searching for projects by name on R-Forge doesn't seem to work for any package, even when I'm logged in to R-Forge.

Fortunately in this case, the changes needed can be expressed over email. In add.covariates.R please change lines 61 and 73 as follows :

```
# from :
names(interaction.data)[ 'trans.count' == names(interaction.data) ] <- 'bait.trans.count';
# to:
setnames(interaction.data, "trans.count", "bait.trans.count")

# from :
names(interaction.data)[ 'trans.count' == names(interaction.data) ] <- 'target.trans.count';
# to :
setnames(interaction.data, "trans.count", "target.trans.count")
```

The reason is that the <- way goes via R's <- mechanism which copies the data and there's nothing I can do to avoid that copy when using R's <- mechanism.   That's why setnames() exists.  Further, setnames is more robust because you don't need to repeat the object name (interaction.data) a 2nd time which can cause bugs when the 2nd use happens to be another object with a similar name that gets used accidentally.  setnames() will also error if the column name being changed does not exist. The error suggests passing skip_absent=TRUE if it is intentional that sometimes the column might not exist. These features make setnames() a much better way to change column names of a data.table, regardless of <- being affected by the next release.

The reason the names(DT)<- breaks chicane in the next release of data.table is that the key is now dropped due to detecting R's copy.  The lack of key in the result causes 4 expect_equal tests to fail in test_add.covariates.R, lines 65, 93, 122 and 151.  I confirmed that R CMD check passes (with 2 unrelated notes) by changing to use setnames() in add.covarates.R.

You don't need to wait for v1.14.2 to be released, changing to setnames() will work before and after.

It is a consequence of bug fix 28 :

> dplyr::arrange(DT) uses vctrs::vec_slice which retains data.table's class but uses C to bypass [ method dispatch and does not adjust data.table's attributes containing the index row numbers, #5042. data.table's long-standing .internal.selfref mechanism to detect such operations by other packages was not being checked by data.table when using indexes, causing data.table filters and joins to use invalid indexes and return incorrect results after a dplyr::arrange(DT). Thanks to Waldi73 for reporting; avimallu, tlapak, MichaelChirico, jangorecki and hadley for investigating and suggestions; and mattdowle for the PR. The intended way to use data.table is data.table::setkey(DT, col1, col2, ...) which reorders DT by reference in parallel, sets the primary key for automatic use by subsequent data.table queries, and permits rowname-like usage such as DT["foo",] which returns the now-contiguous-in-memory block of rows where the first column of DT's key contains "foo". Multi-column-rownames (i.e. a primary key of more than one column) can be looked up using DT[.("foo",20210728L), ]. Using == in i is also optimized to use the key or indices, if you prefer using column names explicitly and ==. An alternative to setkey(DT) is returning a new ordered result using DT[order(col1, col2, ...), ].

Best, Matt

### LINKED ISSUE #5126: revdep CornerstoneR affected
Same as #5125.

It does this on line 232 of fitFunction.R : 
```
names(dtStart)[-seq_along(groups)] = par.names
```
which now drops `dtStart`'s key and results later in the errors below.
It's harder to see how this one can be converted to `setnames()` so I'm leaning towards finding a way to retain the key, rather than contact the revdep maintainer.

```
> test_check("CornerstoneR")
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure (test_fitFunction.R:95:3): Work with Invalid Variable Names ─────────
Check on res$coeff is not TRUE

`actual` is a character vector ('Contains missing values (column \'Coeff_a\', row 1)')
`expected` is a logical vector (TRUE)
Contains missing values (column 'Coeff_a', row 1)
Backtrace:
    █
 1. └─checkmate::expect_data_table(...) test_fitFunction.R:95:2
 2.   └─checkmate::makeExpectation(x, res, info, label)
 3.     └─testthat::expect_true(...)
```

which is because of these errors `try()`-ed and caught and places within `dtCoeff` due to missing key : 

```
$coeff
   X1.grp Coeff_a Coeff_b Coeff_c Coeff_d StdErr_a StdErr_b StdErr_c StdErr_d pseudoR2  RMSE Converged Iterations Tolerance StopCode
   <char>   <num>   <num>   <num>   <num>    <num>    <num>    <num>    <num>    <num> <num>    <lgcl>      <num>     <num>    <num>
1:      A      NA      NA      NA      NA       NA       NA       NA       NA       NA    NA        NA         NA        NA       NA
2:      B      NA      NA      NA      NA       NA       NA       NA       NA       NA    NA        NA         NA        NA       NA
                                                                                                                                                                                                                                                                                                                                                                 StopMessage
                                                                                                                                                                                                                                                                                                                                                                      <char>
1: Error : When i is a data.table (or character vector), the columns to join by must be specified using 'on=' argument (see ?data.table), by keying x (i.e. sorted, and, marked as sorted, see ?setkey), or by sharing column names between x and i (i.e., a natural join). Keyed joins might have further speed benefits on very large data due to x being sorted in RAM.\n
2: Error : When i is a data.table (or character vector), the columns to join by must be specified using 'on=' argument (see ?data.table), by keying x (i.e. sorted, and, marked as sorted, see ?setkey), or by sharing column names between x and i (i.e., a natural join). Keyed joins might have further speed benefits on very large data due to x being sorted in RAM.\n
```

### LINKED ISSUE #5127: revdep getDTeval affected
Same as #5125

On line 138 of getDTeval it does : 
```
names(the.result) <- gsub(pattern = "`", replacement = "", x = names(the.result))
```
which drops the key and causes : 

```
> test_check("getDTeval")
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure (test-getDteval.R:68:2): create a new column ────────────────────────
getDTeval(...) not equal to formulaic::snack.dat[, `:=`(age_decades, floor(get(age.name)/10))].
Datasets has different number of (non-excluded) attributes: target 2, current 3

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 4 ]
Error: Test failures
Execution halted
```

### LINKED ISSUE #5128: revdep maditr affected
Last of the 4/1111 revdeps and it's the same as #5125, #5126 and #5127

Line 121 of test_take_all.R is : 

```
names(res)[-1] = paste0("mean_", names(res)[-1])
```

which causes 

```
----- FAILED[attr]: test_take_all.R<117--124>
 call| expect_equal(current = take_all(iris, mean = mean, keyby = Species, 
 call| suffix = FALSE, .SDcols = -(1:2)), target = {
 call| res = dt_iris[, lapply(.SD, mean), keyby = Species, .SDcols = -(1:2)]
 call| names(res)[-1] = paste0("mean_", names(res)[-1])
 call| res
 call| })
 diff| Datasets have different keys. 'target': has no key. 'current': [Species].
Error: 1 out of 391 tests failed
Execution halted
```

MichaelChirico:
Follow up to #5084
Related to #5085
Closes #5125
Closes #5126
Closes #5127
Closes #5128

Minimized successor to #5133. That PR has a lot going on that I don't follow -- for now I'm focused on fixing the broken revdeps so we can release to CRAN.

I've confirmed {CornerstoneR}, {maditr}, {getDTeval}, and {chicane} have no test failures after this fix.

Revdep {simDAG} (as noted here: https://github.com/Rdatatable/data.table/issues/5128#issuecomment-1702133523) had 5 errors, this PR removes the 2 related to attributes. The remaining three are covered by #5680.


jangorecki:
Yes, feels related to NSE
https://github.com/RobinDenz1/simDAG/blob/814c2133acf87bccb492322d702171105fef6415/R/sim2start_stop.r#L136

inst/tests/tests.Rraw
```diff
@@ -18111,3 +18111,11 @@ test(2238.9, NA %notin% c(1:5, NA), FALSE)
 
 # shift actionable error on matrix input #5287
 test(2239.1, shift(matrix(1:10, ncol = 1)), error="consider wrapping")
+
+# names<- retains index and key, #5125, #5126, #5126, #5128
```
jangorecki, line 18115: Great work.
Could we add rest of unit tests added in #5133?
If some of them fails then possibly comment them out with explanation. So we have clarity what from #5133 is being covered.

MichaelChirico:
> Yes, feels related to NSE https://github.com/RobinDenz1/simDAG/blob/814c2133acf87bccb492322d702171105fef6415/R/sim2start_stop.r#L136

reading again probably these tests were already failing. this PR moves from 5 to 3 failed tests in simDAG, we only see output of 2 tests failing from names<- problem. I assume the other 3 are the same as here (haven't checked)

inst/tests/tests.Rraw
MichaelChirico: good idea. I skipped originally bc I wasn't sure whether output change tests are worth copying.

I'll read more carefully and copy over the worthwhile tests.

MichaelChirico: OK, copied over the tests. They caught two things:

 1. `colnames(x) <-` is now treated the same a `names(x) <-`. Requires a new S3 method registration.
 2. Doing `names(x) <- "..."; set(x, j=j, val=val)` is fixed with a simple `setalloccol()`. Maybe not ideal, but a good expedient.

MichaelChirico:
> > Yes, feels related to NSE https://github.com/RobinDenz1/simDAG/blob/814c2133acf87bccb492322d702171105fef6415/R/sim2start_stop.r#L136
> 
> reading again probably these tests were already failing. this PR moves from 5 to 3 failed tests in simDAG, we only see output of 2 tests failing from names<- problem. I assume the other 3 are the same as here (haven't checked)

confirmed: here's the 5 tests failing on data.table master:

```
══ Failed tests ═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
── Error (test_long2start_stop.r:17:3): general test case ──────────────────────
Error: object 'max_t' not found
Backtrace:
    ▆
 1. ├─simDAG::long2start_stop(...) at test_long2start_stop.r:17:3
 2. │ ├─...[]
 3. │ └─data.table::`[.data.table`(...)
 4. │   └─data.table:::gforce(thisEnv, jsub, o__, f__, len__, irows)
 5. └─data.table:::gshift(start, type = "lead", fill = max_t)
── Error (test_long2start_stop.r:38:3): non-logical time-varying variables ─────
Error: object 'max_t' not found
Backtrace:
    ▆
 1. ├─simDAG::long2start_stop(...) at test_long2start_stop.r:38:3
 2. │ ├─...[]
 3. │ └─data.table::`[.data.table`(...)
 4. │   └─data.table:::gforce(thisEnv, jsub, o__, f__, len__, irows)
 5. └─data.table:::gshift(start, type = "lead", fill = max_t)
── Error (test_sim2data_all_equal_last.r:23:3): start-stop: .all equal to .last ──
Error: object 'max_t' not found
Backtrace:
    ▆
 1. ├─simDAG::sim2data(sim, to = "start_stop", use_saved_states = TRUE) at test_sim2data_all_equal_last.r:23:3
 2. │ └─simDAG:::sim2start_stop(sim = sim, use_saved_states = use_saved_states)
 3. │   └─simDAG:::sim2start_stop.all(sim = sim)
 4. │     └─simDAG::long2start_stop(...)
 5. │       ├─...[]
 6. │       └─data.table::`[.data.table`(...)
 7. │         └─data.table:::gforce(thisEnv, jsub, o__, f__, len__, irows)
 8. └─data.table:::gshift(start, type = "lead", fill = max_t)
── Failure (test_sim2data_all_equal_last.r:35:3): long: .all equal to .last ────
`d_all` (`actual`) not equal to `d_last` (`expected`).

`attr(actual, 'sorted')` is absent
`attr(expected, 'sorted')` is a character vector ('.id', '.time')
── Failure (test_sim2long.all.r:52:3): general test case ───────────────────────
`out_dat` (`actual`) not equal to `expected` (`expected`).

`attr(actual, 'sorted')` is absent
`attr(expected, 'sorted')` is a character vector ('.id', '.time')

[ FAIL 5 | WARN 0 | SKIP 0 | PASS 402 ]
```

inst/tests/tests.Rraw
jangorecki: I don't see any tests being commented out. Does it mean we cover all tests from Matt's PR?

MichaelChirico: I didn't copy over anything where the only diff is in error/message output, i.e.

https://github.com/Rdatatable/data.table/pull/5133/files#diff-e3243f3780ce7d303c3317f73945310bfc37e45d193568246246aca20e3270aeR1336-R1341

https://github.com/Rdatatable/data.table/pull/5133/files#diff-e3243f3780ce7d303c3317f73945310bfc37e45d193568246246aca20e3270aeR1354-R1359

The new tests of actual code are copied over and pass:

https://github.com/Rdatatable/data.table/pull/5133/files#diff-e3243f3780ce7d303c3317f73945310bfc37e45d193568246246aca20e3270aeR1532-R1552

jangorecki (approved):

MichaelChirico:
Will go ahead and merge, but it would be helpful if some of the commenters on the original PR #5133  (@ben-schwen @ColeMiller1) could chime in on whether we're leaving something on the table vs. the original intent of that PR, in other words, what should we target as follow-up to achieve the original goals of #5133? Or should we consider the matter resolved & close that PR too?

PR #5862, opened 2023-12-28
Update .pot & .po files

MichaelChirico:
Since we're just about done 1.15.0, it is a good time to get translations ready for this update.

CL generated with:

```r
potools::po_extract()
potools::po_update()
```

 - This adds source references to the R files ({tools} helpers aren't built for this)
 - It also uses a set width for R messages ({tools} has unlimited width for R messages)


MichaelChirico:
No user-visible change here, merging

PR #5864, opened 2023-12-29
Update Mandarin translations

MichaelChirico:
A lot is due to being more intentional about using `stopf()` to ensure full messages are included (#5068), not fragments, so a lot of the fuzzed translations just require adding a `%s` or `%d`.

On the C side, a lot of the fuzz is due to #5790 -related fixes, just requires changing the format template specifier.

I got things started here but there's still a lot of fuzzy translations left.

Would appreciate if @hongyuanjia or any other Mandarin speakers could take over from here (as well as copy-edit my own attempts) 🙏

### DIFF
--- a/po/R-zh_CN.po
+++ b/po/R-zh_CN.po
@@ -1452,10 +1452,9 @@ msgid "x must be an atomic vector or data.frames/data.tables"
 msgstr "x 必须是原子向量或data.frame/data.table"
 
 #: fcast.R:7
-#, fuzzy, c-format
-#| msgid "' as value column. Use 'value.var' to override"
+#, c-format
 msgid "Using '%s' as value column. Use 'value.var' to override"
-msgstr "' 作为 value 列。可使用 'value.var' 修改"
+msgstr "用 '%s' 作为 value 列。可使用 'value.var' 修改"
 
 #: fcast.R:20 fmelt.R:16
 #, c-format
@@ -1521,10 +1520,9 @@ msgid "'drop' must be logical TRUE/FALSE"
 msgstr "'drop' 必须为逻辑 TRUE/FALSE"
 
 #: fcast.R:128
-#, fuzzy, c-format
-#| msgid "] not found or of unknown type."
+#, c-format
 msgid "Column [%s] not found or of unknown type."
-msgstr "] 无法找到或其类型未知。"
+msgstr "列 [%s]] 无法找到或其类型未知。"
 
 #: fcast.R:143
 #, c-format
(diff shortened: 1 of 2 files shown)


hongyuanjia:
@MichaelChirico, you have already done a great job! Only a few need minor updates. I will read through it and update it.

MichaelChirico:
I've really just started 😅

Search for `#, fuzzy` in the .po files for the full list of fuzzy messages:

https://github.com/Rdatatable/data.table/blob/master/po/data.table.pot
https://github.com/Rdatatable/data.table/blob/master/po/zh_CN.po

MichaelChirico:
Remember to poke around in the deprecated messages (lines starting `#~`) for fragments which were translated but no longer associated to a message, there could be some helpful info in there.

po/R-zh_CN.po
```diff
@@ -158,7 +152,7 @@ msgstr ""
 msgid ""
 "'between' arguments are all POSIXct but have mismatched tzone attributes: "
 "%s. The UTC times will be compared."
-msgstr "'between' 的参数均为 POSIXct 类型但时区属性（tzone）不匹配："
+msgstr "<fix fuzzy translation>'between' 的参数均为 POSIXct 类型但时区属性（tzone）不匹配："
```
MichaelChirico, line 155:
```suggestion
msgstr "<fix fuzzy translation>'between' 的参数均为 POSIXct 类型但时区属性（tzone）不匹配：%s。用 UTC 比较。"
```

@hongyuanjia I've added markers for all the pieces that need addressing, either with `<fix fuzzy translation>` as here or `<needs translation>` as e.g. L187.

My hope is we can use the `suggestion` feature on GitHub (as done here in this comment) to make it easier to finish the translations directly in the PR.

LMK if that makes sense to you, and if my sense that this is easier is right/wrong.

po/R-zh_CN.po
```diff
@@ -190,7 +184,7 @@ msgstr "或许你想用的是 %s？"
 msgid ""
 "RHS has length() %d; expecting length 2. %sThe first element should be the "
 "lower bound(s); the second element should be the upper bound(s)."
-msgstr ""
+msgstr "<needs translation>"
```
MichaelChirico, line 187:
```suggestion
msgstr "右边有长度 %d，但必须长度 2。 %s第一个元素该是下限，第二个该是上限。"
```

Another example for the `<needs translation>` types

po/R-zh_CN.po
hongyuanjia: I got it. I have already finished all `#, fuzzy` ones and added translations for those messages that are currently missing. I will be sure to fetch your updates to see if any more updates are needed.

I use GitHub Copilot to do the essential translations and check manually to ensure it makes sense and feels more natural. Currently, it did a really good job ;)

MichaelChirico:
> jangorecki added this to the 1.15.0 milestone 14 minutes ago

@jangorecki FYI, I didn't add this to the milestone earlier since it's a "nice-to-have". I didn't think it should be a blocker to finishing the release -- if Hongyuan (or other translators) happen not to have time, then we can just include the updates next time.

That's because this release has a bit of urgency given the lack of "real" releases for data.table for so long & because this request comes so close to finishing the milestone -- in typical release cycle, this request for updated translations would be sent out with plenty more time, and should be part of the milestone.

po/R-zh_CN.po
hongyuanjia: Indeed, there are quite a lot conflicts... I will open a new PR. Please see #5875.

MichaelChirico:
Obsolete by #5875

PR #5870, opened 2024-01-01
update rlib actions setup-r and pandoc to v2

ben-schwen:
Follows #5632
Updates Github CI pipeline since used version 1 is deprecated

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -39,11 +39,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: r-lib/actions/setup-r@v1
+      - uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.config.r }}
 
-      - uses: r-lib/actions/setup-pandoc@v1
+      - uses: r-lib/actions/setup-pandoc@v2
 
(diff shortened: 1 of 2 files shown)


MichaelChirico:
We might want to add dependabot to help automate upkeep of GHA versioning:

See e.g. https://github.com/r-lib/lintr/pull/2120

MichaelChirico (approved):

jangorecki (commented):
I wonder if we still need pandoc, can we try if it succeeds without it?

.github/workflows/R-CMD-check.yaml
```diff
@@ -39,11 +39,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: r-lib/actions/setup-r@v1
+      - uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.config.r }}
 
-      - uses: r-lib/actions/setup-pandoc@v1
+      - uses: r-lib/actions/setup-pandoc@v2
```
jangorecki, line 46:
```suggestion
```

jangorecki: our vignettes now are being rendered with markdown package which does not need pandoc

ben-schwen:
Great spot!

PR #5871, opened 2024-01-02
Switch to git-friendly ordered list in NEWS

MichaelChirico:
Markdown does not care about item numbers in lists, see e.g.

 - https://www.markdownguide.org/basic-syntax/#ordered-lists
 - https://spec.commonmark.org/0.30/#lists
 - https://daringfireball.net/projects/markdown/syntax#list

Using `1.` for every ordered item number is thus preferable because it is much friendlier for git -- we don't need to worry about mistaken git merges winding up with mis-numbered entries, or resolving conflicts by tediously enumerating the NEWS items before/after merge.

It also allows us to add/remove NEWS items in the middle of a block without creating a huge distracting diff as subsequent numbers are adjusted.

The only advantage I can see to keeping the actual item numbers is being able to see the numbers before markdown rendering, which could be useful for things like "See item N". But as noted that practice is also fragile, since numbers might move around in the future before release, while it's less likely we'll remember to update the reference to an outdated item number -- better to say "See the item about X".

---

If this is merged, it will likely create git conflicts in all pending PRs; I'll go through the "active" PRs to resolve these afterwards.

---

Old NEWS files are left un-touched as they're "relics" at this point.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,11 +10,11 @@
 
 1. `nafill()` now applies `fill=` to the front/back of the vector when `type="locf|nocb"`, [#3594](https://github.com/Rdatatable/data.table/issues/3594). Thanks to @ben519 for the feature request. It also now returns a named object based on the input names. Note that if you are considering joining and then using `nafill(...,type='locf|nocb')` afterwards, please review `roll=`/`rollends=` which should achieve the same result in one step more efficiently. `nafill()` is for when filling-while-joining (i.e. `roll=`/`rollends=`/`nomatch=`) cannot be applied.
 
-2. `mean(na.rm=TRUE)` by group is now GForce optimized, [#4849](https://github.com/Rdatatable/data.table/issues/4849). Thanks to the [h2oai/db-benchmark](https://github.com/h2oai/db-benchmark) project for spotting this issue. The 1 billion row example in the issue shows 48s reduced to 14s. The optimization also applies to type `integer64` resulting in a difference to the `bit64::mean.integer64` method: `data.table` returns a `double` result whereas `bit64` rounds the mean to the nearest integer.
+1. `mean(na.rm=TRUE)` by group is now GForce optimized, [#4849](https://github.com/Rdatatable/data.table/issues/4849). Thanks to the [h2oai/db-benchmark](https://github.com/h2oai/db-benchmark) project for spotting this issue. The 1 billion row example in the issue shows 48s reduced to 14s. The optimization also applies to type `integer64` resulting in a difference to the `bit64::mean.integer64` method: `data.table` returns a `double` result whereas `bit64` rounds the mean to the nearest integer.
 
-3. `fwrite()` now writes UTF-8 or native csv files by specifying the `encoding=` argument, [#1770](https://github.com/Rdatatable/data.table/pull/1770). Thanks to @shrektan for the request and the PR.
+1. `fwrite()` now writes UTF-8 or native csv files by specifying the `encoding=` argument, [#1770](https://github.com/Rdatatable/data.table/pull/1770). Thanks to @shrektan for the request and the PR.
 
-4. `data.table()` no longer fills empty vectors with `NA` with warning. Instead a 0-row `data.table` is returned, [#3727](https://github.com/Rdatatable/data.table/issues/3727). Since `data.table()` is used internally by `.()`, this brings the following examples in line with expectations in most cases. Thanks to @shrektan for the suggestion and PR.
+1. `data.table()` no longer fills empty vectors with `NA` with warning. Instead a 0-row `data.table` is returned, [#3727](https://github.com/Rdatatable/data.table/issues/3727). Since `data.table()` is used internally by `.()`, this brings the following examples in line with expectations in most cases. Thanks to @shrektan for the suggestion and PR.
 
     ```R
     DT = data.table(A=1:3, B=letters[1:3])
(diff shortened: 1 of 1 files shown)


jangorecki:
For me I more often look at md file sources rather than rendered version. Then having actual numbers in the source is rather preferred.

MichaelChirico:
> For me I more often look at md file sources rather than rendered version. Then having actual numbers in the source is rather preferred.

how often do you actually care about the item number though, really?

I actually prefer to have the NEWS entries unordered as there is no inherent order to bug fixes/features.

> I can imagine we won't be able refer to a n-th element inside news file, without having to look at it's rendered version.

again I think this is bad practice

> It's nice to have this git friendly style but just looking at the source of NEWS.md file and seeing all 1s is really terribly.

it's pretty common in my experience.

tdhock:
I tend to agree with Michael on this one -- the numbering is not meaningful. An alternative would be to instead use bullet points?

jangorecki:
not really happy but feels I have to give up

tdhock:
Jan made an important point that we need to have numbers if we want to refer to n-th element inside news file, without having to look at it's rendered version.
If that is still important then we should consider keeping the status quo.

jangorecki:
It's just I rarely look at rendered MD. Then from my perspective it is loss of information at the cost of convenience of git merge.

MichaelChirico:
> loss of information

AIUI the "lost information" is the item number, which is only useful for constructs like "See 5th New Feature in NEWS for v1.12.0".

We have only ever used something like this once before (`NEWS.0.md:3306`), from a scan (`grep -Enr "[0-9]+(st|th|nd|rd)" NEWS*`) of the NEWS files:

```
NEWS.0.md:1584:    *   .GRP is a new symbol available to j. Value 1 for the first group, 2 for the 2nd, etc. Thanks
NEWS.0.md:2384:        work, 3rd attempt. Thanks to Chris Neff for his
NEWS.0.md:3086:            DT[,.SD[2],by=colA] #  2nd row of each group
NEWS.0.md:3306:    objects. See 2nd new feature above.
NEWS.1.md:154:1. `test.data.table()` could fail the 2nd time it is run by a user in the same R session on Windows due to not resetting locale properly after testing Chinese translation, [#4630](https://github.com/Rdatatable/data.table/pull/4630). Thanks to Cole Miller for investigating and fixing.
NEWS.1.md:166:1. `bit64` v4.0.2 and `bit` v4.0.3, both released on 30th July, correctly broke `data.table`'s tests. Like other packages on our `Suggest` list, we check `data.table` works with `bit64` in our tests. The first break was because `all.equal` always returned `TRUE` in previous versions of `bit64`. Now that `all.equal` works for `integer64`, the incorrect test comparison was revealed. If you use `bit64`, or `nanotime` which uses `bit64`, it is highly recommended to upgrade to the latest `bit64` version. Thanks to Cole Miller for the PR to accommodate `bit64`'s update.
NEWS.1.md:168:    The second break caused by `bit` was the addition of a `copy` function. We did not ask, but the `bit` package kindly offered to change to a different name since `data.table::copy` is long standing. `bit` v4.0.4 released 4th August renamed `copy` to `copy_vector`. Otherwise, users of `data.table` would have needed to prefix every occurrence of `copy` with `data.table::copy` if they use `bit64` too, since `bit64` depends on (rather than importing) `bit`. Again, this impacted `data.table`'s tests which mimic a user's environment; not `data.table` itself per se.
NEWS.1.md:314:0. Retrospective license change permission was sought from and granted by 4 contributors who were missed in [PR#2456](https://github.com/Rdatatable/data.table/pull/2456), [#4140](https://github.com/Rdatatable/data.table/pull/4140). We had used [GitHub's contributor page](https://github.com/Rdatatable/data.table/graphs/contributors) which omits 3 of these due to invalid email addresses, unlike GitLab's contributor page which includes the ids. The 4th omission was a PR to a script which should not have been excluded; a script is code too. We are sorry these contributors were not properly credited before. They have now been added to the contributors list as displayed on CRAN. All the contributors of code to data.table hold its copyright jointly; your contributions belong to you. You contributed to data.table when it had a particular license at that time, and you contributed on that basis. This is why in the last license change, all contributors of code were consulted and each had a veto.
NEWS.1.md:1501:1. Some long-standing potential instability has been discovered and resolved many thanks to a detailed report from Bill Dunlap and Michael Sannella. At C level any call of the form `setAttrib(x, install(), allocVector())` can be unstable in any R package. Despite `setAttrib()` PROTECTing its inputs, the 3rd argument (`allocVector`) can be executed first only for its result to to be released by `install()`'s potential GC before reaching `setAttrib`'s PROTECTion of its inputs. Fixed by either PROTECTing or pre-`install()`ing. Added to CRAN_Release.cmd procedures: i) `grep`s to prevent usage of this idiom in future and ii) running data.table's test suite with `gctorture(TRUE)`.
```

Or is there some other information in the item number you think is worth keeping track of?

I don't want to bulldoze through any change that seems harmful to you, I just don't think I'm fully understanding your perspective here.

I also think the _inconvenience_ of manually merging NEWS item numbers is being undersold. We have 126 pending PRs, basically every time one is merged it invalidates the NEWS commit for a slew of other pending PRs --> toilsome manual work to resolve it.

That cost is way higher than the benefit of numbered items as I see it now.

jangorecki:
I recall referring to numbers in the news file to be pretty common practice, especially breaking changes, and AFAIR Matt's rebuttal to biased arrow timing vs fread. I think that grep you used may not cover most cases. I briefly went through newly rendered news file and even without much reading and looking for such cases, I seen Note 13 has a reference to another point.
In the end we can still refer to numbers even if we use only 1s, it is just less convenient to have to look at rendered version, and possibly little easier to end up out of sync.

tdhock:
some error messages refer to numbered news items, for example fread "Taking input= as a system command ('du -ks Ch14-for Ch14-named') and a variable has been used in the expression passed to `input=`. Please use fread(cmd=...). There is a security concern if you are creating an app, and the app could have a malicious user, and the app is not running in a secure environment; e.g. the app is running as root. Please read item 5 in the NEWS file for v1.11.6 for more information and for the option to suppress this message."

MichaelChirico:
Thanks both. I see this item/note `N` style much more commonly (`grep -Eir "(items?|entr(y|ies)|bullets?|notes?)\s*[0-9]+" NEWS*`):

```
grep -Eir "(items?|entr(y|ies)|bullets?|notes?)\s*[0-9]+" NEWS*
NEWS.0.md:  39. Thanks to Mark Landry and Michael Chirico for finding and reporting a problem in dev before release with auto `with=FALSE` (item 3 above) when `j` starts with with `!` or `-`, [#1864](https://github.com/Rdatatable/data.table/issues/1864). Fixed and tests added.
NEWS.1.md:3. `.rbind.data.table` (note the leading `.`) is no longer exported when `data.table` is installed in R>=4.0.0 (Apr 2020), [#5600](https://github.com/Rdatatable/data.table/pull/5600). It was never documented which R-devel now detects and warns about. It is only needed by `data.table` internals to support R<4.0.0; see note 1 in v1.12.6 (Oct 2019) below in this file for more details.
NEWS.1.md:1. In v1.13.0 (July 2020) native parsing of datetime was added to `fread` by Michael Chirico which dramatically improved performance. Before then datetime was read as type character by default which was slow. Since v1.13.0, UTC-marked datetime (e.g. `2020-07-24T10:11:12.134Z` where the final `Z` is present) has been read automatically as POSIXct and quickly. We provided the migration option `datatable.old.fread.datetime.character` to revert to the previous slow character behavior. We also added the `tz=` argument to control unmarked datetime; i.e. where the `Z` (or equivalent UTC postfix) is missing in the data. The default `tz=""` reads unmarked datetime as character as before, slowly. We gave you the ability to set `tz="UTC"` to turn on the new behavior and read unmarked datetime as UTC, quickly. R sessions that are running in UTC by setting the TZ environment variable, as is good practice and common in production, have also been reading unmarked datetime as UTC since v1.13.0, much faster. Note 1 of v1.13.0 (below in this file) ended `In addition to convenience, fread is now significantly faster in the presence of dates, UTC-marked datetimes, and unmarked datetime when tz="UTC" is provided.`.
NEWS.1.md:2. `fwrite()`'s mutithreaded `gzip` compression failed on Solaris with Z_STREAM_ERROR, [#4099](https://github.com/Rdatatable/data.table/issues/4099). Since this feature was released in Oct 2019 (see item 3 in v1.12.4 below in this news file) there have been no known problems with it on Linux, Windows or Mac. For Solaris, we have been successively adding more and more detailed tracing to the output in each release, culminating in tracing `zlib` internals at byte level by reading `zlib`'s source. The problem did not manifest itself on [R-hub](https://builder.r-hub.io/)'s Solaris instances, so we had to work via CRAN output. If `zlib`'s `z_stream` structure is declared inside a parallel region but before a parallel for, it appears that the particular OpenMP implementation used by CRAN's Solaris moves the structure to a new address on entering the parallel for. Ordinarily this memory move would not matter, however, `zlib` internals have a self reference pointer to the parent, and check that the pointers match. This mismatch caused the -2 (Z_STREAM_ERROR). Allocating an array of structures, one for each thread, before the parallel region avoids the memory move with no cost.
NEWS.1.md:12. The `datatable.old.unique.by.key` option has been removed as per the 4 year schedule detailed in note 10 of v1.12.4 (Oct 2019), note 10 of v1.11.0 (May 2018), and note 1 of v1.9.8 (Nov 2016). It has been generating a helpful warning for 2 years, and helpful error for 1 year.
NEWS.1.md:2. Compilation failed on CRAN's MacOS due to an older version of `zlib.h/zconf.h` which did not have `z_const` defined, [#3939](https://github.com/Rdatatable/data.table/issues/3939). Other open-source projects unrelated to R have experienced this problem on MacOS too. We have followed the common practice of removing `z_const` to support the older `zlib` versions, and data.table's release procedures have gained a `grep` to ensure `z_const` isn't used again by accident in future. The library `zlib` is used for `fwrite`'s new feature of multithreaded compression on-the-fly; see item 3 of 1.12.4 below.
NEWS.1.md:3. A missing item in `j` such as `j=.(colA, )` now gives a helpful error (`Item 2 of the .() or list() passed to j is missing`) rather than the unhelpful error `argument "this_jsub" is missing, with no default` (v1.12.2) or `argument 2 is empty` (v1.12.0 and before), [#3507](https://github.com/Rdatatable/data.table/issues/3507). Thanks to @eddelbuettel for the report.
NEWS.1.md:1. `rbindlist`'s `use.names="check"` now emits its message for automatic column names (`"V[0-9]+"`) too, [#3484](https://github.com/Rdatatable/data.table/pull/3484). See news item 5 of v1.12.2 below.
NEWS.1.md:10. The `datatable.old.unique.by.key` option has been warning for 1 year that it is deprecated: `... Please stop using it and pass by=key(DT) instead for clarity ...`. This warning is now upgraded to error as per the schedule in note 10 of v1.11.0 (May 2018), and note 1 of v1.9.8 (Nov 2016). In June 2020 the option will be removed.
NEWS.1.md:      See news item 5 in v1.12.2 for options to control this message.
NEWS.1.md:      See news item 5 in v1.12.2 for options to control this message.
NEWS.1.md:1. When upgrading to 1.12.0 some Windows users might have seen `CdllVersion not found` in some circumstances. We found a way to catch that so the [helpful message](https://twitter.com/MattDowle/status/1084528873549705217) now occurs for those upgrading from versions prior to 1.12.0 too, as well as those upgrading from 1.12.0 to a later version. See item 1 in notes section of 1.12.0 below for more background.
NEWS.1.md:9. Reminder that note 11 in v1.11.0 (May 2018) warned that `set2key()` and `key2()` will be removed in May 2019. They have been warning since v1.9.8 (Nov 2016) and their warnings were upgraded to errors in v1.11.0 (May 2018). When they were introduced in version 1.9.4 (Oct 2014) they were marked as 'experimental'.
NEWS.1.md:      or more RHS values contain fractions which have been lost; e.g. item 1 with value 3.140000 has
NEWS.1.md:18. Item 1 in NEWS for [v1.10.2](https://github.com/Rdatatable/data.table/blob/master/NEWS.md#changes-in-v1102--on-cran-31-jan-2017) on CRAN in Jan 2017 included :
NEWS.1.md:11. As `set2key()` and `key2()` have been warning since v1.9.8 (Nov 2016), their warnings have now been upgraded to errors. Note that when they were introduced in version 1.9.4 (Oct 2014) they were marked as 'experimental' in NEWS item 4. They will be removed in one year.
NEWS.md:      Item 2 has 0 rows but longest item has 1; filled with NA
NEWS.md:      Item 3 has 0 rows but longest item has 1; filled with NA
NEWS.md:      Item 2 has 0 rows but longest item has 3; filled with NA
NEWS.md:22. `DT[i, nomatch=NULL]` where `i` contains row numbers now excludes `NA` and any outside the range [1,nrow], [#3109](https://github.com/Rdatatable/data.table/issues/3109) [#3666](https://github.com/Rdatatable/data.table/issues/3666). Before, `NA` rows were returned always for such values; i.e. `nomatch=0|NULL` was ignored. Thanks Michel Lang and Hadley Wickham for the requests, and Jan Gorecki for the PR. Using `nomatch=0` in this case when `i` is row numbers generates the warning `Please use nomatch=NULL instead of nomatch=0; see news item 5 in v1.12.0 (Jan 2019)`.
NEWS.md:13. For nearly two years, since v1.12.4 (Oct 2019) (note 11 below in this NEWS file), using `options(datatable.nomatch=0)` has produced the following message :
```

Also, solving the git merge issue for NEWS does not also solve the issue for `inst/tests/tests.Rraw`, which is equally toilsome.

Therefore I think we can table this for now.

PR #5878, opened 2024-01-03
Fix mismatch of messages found by checkPoFiles()

MichaelChirico:
Related: #5681. I think the best approach is to ensure this test is run during CRAN release, since it only needs to be run occasionally.

cc @hongyuanjia to check I've tidied things up correctly. Thanks one last time!

### DIFF
--- a/po/R-zh_CN.po
+++ b/po/R-zh_CN.po
@@ -129,7 +129,7 @@ msgid ""
 "'between' function the 'x' argument is a POSIX class while '%s' was not, "
 "coercion to POSIX failed with: %s"
 msgstr ""
-"'between' 中的 'x' 参数为 POSIX 类，而 '%s' 并不是，将自动转换成 POSIX 失败：%s"
+"'between' 中的 'x' 参数为 POSIX 类，而 '%s' 并不是，将其强制转换成 POSIX 时失败：%s"
 
 #: between.R:27
 #, c-format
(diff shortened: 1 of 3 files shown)


po/R-zh_CN.po
```diff
@@ -129,8 +129,7 @@ msgid ""
 "'between' function the 'x' argument is a POSIX class while '%s' was not, "
 "coercion to POSIX failed with: %s"
 msgstr ""
-"'between' 中的 'x' 参数为 POSIX 类，而 'lower' 并不是，将 'lower' 自动转换成 "
-"POSIX 失败：%s"
+"'between' 中的 'x' 参数为 POSIX 类，而 '%s' 并不是，将自动转换成 POSIX 失败：%s"
```
hongyuanjia, line 132: The translation here sounds ambiguous. I would suggest:

```
"'between' 中的 'x' 参数为 POSIX 类，而 '%s' 并不是，将其强制转换成 POSIX 时失败：%s"
```

hongyuanjia:
@MichaelChirico, thanks for catching so many formatting errors in my PR. Every translation looks good to me, except one. Please see my comments.

po/R-zh_CN.po
MichaelChirico:
```suggestion
"'between' 中的 'x' 参数为 POSIX 类，而 '%s' 并不是，将其强制转换成 POSIX 时失败：%s"
```

MichaelChirico:
> @MichaelChirico, thanks for catching so many formatting errors in my PR.

In fact, I didn't catch them at all 😂 thanks to the R authors for developing `tools::checkPoFiles()` which did all the detective work here!

PR #5881, opened 2024-01-04
.SDcols is a negation only for _unary_ '-'

### LINKED ISSUE #5826: revdep simtrial test ERROR
revdep simtrial recently submitted an update which shows the error below using data.table master (but not with data.table from CRAN).
```
* checking tests ...
  Running 'testthat.R'
 ERROR
Running the tests in 'tests/testthat.R' failed.
Last 13 lines of output:
  == Failed tests ================================================================
  -- Failure ('test-independent_test_pvalue_maxcombo.R:83:3'): the p-values correspond to pvalue_maxcombo --
  `p1` not equal to `p2`.
  1/1 mismatches
  [1] 1 - 0.0716 == 0.928
  -- Failure ('test-independent_test_wlr.R:25:3'): the z values match with the correspondings in fh_weight --
  c(z1[1], z1[7:9]) not equal to `z2`.
  3/4 mismatches (average diff: 5.54)
  [1] 5.67 - -0.1777 == 5.85
  [2] 4.77 - -0.7132 == 5.48
  [3] 5.26 - -0.0381 == 5.30
  
  [ FAIL 2 | WARN 0 | SKIP 0 | PASS 182 ]
  Error: Test failures
  Execution halted
```
above output indicates the errors are numerical, which seems odd. But I double checked on another machine, the test failure is real.

git bisect says this PR is the cause https://github.com/Rdatatable/data.table/pull/4470 which seems to be related to .SDcols, which seems inconsistent with the numerical errors in the test failure. I don't understand, but maybe @jangorecki and @ColeMiller1 could comment/investigate please? (you authored/commented on that PR)

test source code is on github
* https://github.com/Merck/simtrial/blob/main/tests/testthat/test-independent_test_wlr.R#L25
* https://github.com/Merck/simtrial/blob/main/tests/testthat/test-independent_test_pvalue_maxcombo.R#L83

MichaelChirico:
Closes #5826

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -341,7 +341,7 @@
 
 23. `fread(fill=TRUE, verbose=TRUE)` would segfault on the out-of-sample type bump verbose output if the input did not contain column names, [5046](https://github.com/Rdatatable/data.table/pull/5046). Thanks to Václav Tlapák for the PR.
 
-24. `.SDcols=-V2:-V1` and `.SDcols=(-1)` could error with `xcolAns does not pass checks` and `argument specifying columns specify non existing column(s)`, [#4231](https://github.com/Rdatatable/data.table/issues/4231). Thanks to Jan Gorecki for reporting and the PR.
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Must admit I still don't understand the bug very well -- this incorrect assumption about unary `-` has been around since the first time `.SDcols` could accept `-` >10 years ago:

ac2180ef17b8a46245261349f05f5967318407cf

So what changed to make it an error now that we didn't notice before? 🤷

jangorecki (approved):
Nice it got detected. It could be a mess having this bug to end up on CRAN. Hidden costs of NSE...

PR #5883, opened 2024-01-06
Include note on plans re: logical01

### LINKED ISSUE #5856: Should logical01 be TRUE by default?
Follow-up to #5840.

When it was originally developed, the intention was to eventually turn on this argument by default.

The benefit is getting the storage type right by default. Columns with values `0/1/NA` _should_ be "logical" storage in R. R is also very good about converting back to numeric 0/1 when appropriate.

There are, however, a few downsides:

 - We can pretty easily get it wrong. In particular, the current logic just says basically "`if all(is.na(x) | x %in% c(0, 1)) as.logical(x) else x` i.e., any column with only values among 0/1/NA will be read as type logical by default. Where that rule comes across as funny to me is in cases where _there is only one unique value_ (or perhaps 2, if some entries are missing). The vector `[1 1 1 1 1 1 1 1 1]` _might_ be logical, but it could just be a constant numeric column, too. Same for a vector of zeroes.

  That hints at a potential improvement to the logic here, namely `if (all((is.na(x) | x %in% c(0, 1)) && any(x == 0) & any(x == 1)) as.logical(x) else x`, i.e., any column with only values 0/1/NA _that takes both 0 and 1 as values at least once_ will be read as type logical. E.g. we could call this behavior `logical01 = "auto"`.

  But that opens us up to a new problem -- reading directories where `"auto"` guesses differently for different files, and therefore the resulting tables can't be trivially `rbind()`'d. For example, consider one process doing `lapply(split(x, by = col), fwrite, ...)`, i.e., sharding `x` by some column `col`. There is a column `v` in `x` that only takes values 0/1, but in some shards it's constant, e.g. in the first file `all(v == 0)`, in the second file there's a mix, in the third file `all(v == 1)`. Then `logical01 = "auto"` will read `v` in the first & third files as numeric but as logical in the second file. Supplying `logical01 = TRUE/FALSE` explicitly is a workaround for this case, but it seems unpleasant if this comes up with any frequency. Auto-promoting logical->(integer|double) in `rbindlist()` would also work, but should be considered more carefully within `rbindlist()` itself whether this behavior makes sense in general.

- Back-compatibility issue. There are (as of this writing) 16 CRAN packages broken by changing `logical01=TRUE` by default, namely: NMdata, PeakSegDisk, Rdiagnosislist, antaresEditObject, antaresRead, bigsnpr, chicane, damr, googleCloudVisionR, growthcleanr, gtfsrouter, pepr, prioriactions, skynet, spatsoc, targets.

  It could be that this is a simple issue in tests, e.g. as happened in #5843 where our own tests had examples of one-row inputs requiring a simple `1`->`TRUE` change. One-row minimal examples for tests are not representative of end-user data, so the problem could be minimal. But it requires a bit more investigation to understand the severity of each individual breakage. On the flipside, this was always planned & stated in the NEWS for 6 years as the intention, and the initial phase would offer a simple global R `option()` to retain `logical01=FALSE` if desired, i.e., this is part of a normal deprecation cycle.

That's a dump of my thoughts on this issue for now. Opening this for discussion as we head into the next release.

MichaelChirico:
Closes #5856

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -615,4 +615,6 @@
 
 18. Option `datatable.CJ.names` no longer has any effect, after becoming `TRUE` by default in v1.12.2 (2019). Setting it now gives a warning, which will be dropped in the next release.
 
+19. In the NEWS for v1.11.0 (May 2018), section "NOTICE OF INTENDED FUTURE POTENTIAL BREAKING CHANGES" item 2, we stated the intention to eventually change `logical01` to be `TRUE` by default. After some consideration, reflection, and community input, we have decided not to move forward with this plan, i.e., `logical01` will remain `FALSE` by default in both `fread()` and `fwrite()`. See discussion in #5856; most importantly, we think changing the default would be a major disruption to reading "sharded" CSVs where data with the same schema is split into many files, some of which could be converted to `logical` while others remain `integer`. We will retain the option `datatable.logical01` for users who wish to use a different default -- for example, if you are doing input/output on tables with a large number of logical columns, where writing '0'/'1' to the CSV many millions of times is preferable to writing 'TRUE'/'FALSE'.
+
 # data.table v1.14.10 (Dec 2023) back to v1.10.0 (Dec 2016) has been moved to [NEWS.1.md](https://github.com/Rdatatable/data.table/blob/master/NEWS.1.md)


jangorecki (approved):
perfect!

PR #5887, opened 2024-01-07
Gforce eval vars

ben-schwen:
Follows #5548 

Problems are j arguments of the form `x[y]`.

- Adds check that `y` is indeed numeric. This was already partially done by enforcing `y>0` but didn't block `y==TRUE`.
- Enforce that y is not a column in the calling `data.table`

We could possible add a check to enforce `x` being a column in the data.table but this would deactive GForce for `.I[1]` expressions. Plus we already have the check that x should be atomic in the environment of the data.table.

### Revdep checks:
- [x] dtplyr tests
- [x] ahw tests + examples


inst/tests/tests.Rraw
```diff
@@ -18245,3 +18245,12 @@ test(2243.35, dt[, min(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(
 test(2243.36, dt[, max(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(g=c(1L,2L), V1=c(3L,4L)),        output="GForce FALSE")
 test(2243.37, dt[, var(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(g=c(1L,2L), V1=c(2,2)),          output="GForce FALSE")
 test(2243.38, dt[, sd(y,   na.rm=as.logical(j)),   g, verbose=TRUE], data.table(g=c(1L,2L), V1=c(sqrt(c(2,2)))), output="GForce FALSE")
+# revdeps
+dt = data.table(x = c(2,2,1,1), y = 1:4, z=letters[1:4])
+old = options(datatable.optimize=1L)
+test(2243.41, dt[, .I[TRUE], x]$V1, 1:4)
+test(2243.42, dt[, z[y], x], data.table(x=c(2,2,1,1), V1=c("a","b",NA,NA)))
+options(datatable.optimize=2L)
+test(2243.51, dt[, .I[TRUE], x]$V1, 1:4)
```
MichaelChirico, line 18254: verbose=TRUE to confirm whether GForce ran?

ben-schwen: Good point. Added + additional tests.

inst/tests/tests.Rraw
```diff
@@ -18245,3 +18245,18 @@ test(2243.35, dt[, min(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(
 test(2243.36, dt[, max(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(g=c(1L,2L), V1=c(3L,4L)),        output="GForce FALSE")
 test(2243.37, dt[, var(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(g=c(1L,2L), V1=c(2,2)),          output="GForce FALSE")
 test(2243.38, dt[, sd(y,   na.rm=as.logical(j)),   g, verbose=TRUE], data.table(g=c(1L,2L), V1=c(sqrt(c(2,2)))), output="GForce FALSE")
+# revdeps
+dt = data.table(x = c(2,2,1,1), y = 1:4, z=letters[1:4])
+i=c(1,2)
```
MichaelChirico, line 18250: is this intentionally double? I'm noticing the contrast to 1:2 below

inst/tests/tests.Rraw
```diff
@@ -18245,3 +18245,18 @@ test(2243.35, dt[, min(y,  na.rm=as.logical(j)),   g, verbose=TRUE], data.table(
...
+j=1L
```
MichaelChirico, line 18251: we also have the same contrast (in reverse) of 1 vs 1L

MichaelChirico (approved):
Minor point on tests, LGTM otherwise. feel free to merge

inst/tests/tests.Rraw
ben-schwen: Yes, I thought this way we test both.

inst/tests/tests.Rraw
ben-schwen: Same as above

MichaelChirico (approved):

PR #5889, opened 2024-01-08
frollmax1: frollmax, frollmax adaptive, left adaptive support

jangorecki:
- frollmax normal
- frollmax adaptive
- left adaptive support


man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+  see \emph{Adaptive rolling functions} section below for details.
 
-  When \code{algo="fast"} an \emph{"on-line"} algorithm is used, and
-  all of \code{NaN, +Inf, -Inf} are treated as \code{NA}.
-  Setting \code{algo="exact"} will make rolling functions to use a more
-  computationally-intensive algorithm that suffers less from floating point
-  rounding error (the same consideration applies to \code{\link[base]{mean}}).
-  \code{algo="exact"} also handles \code{NaN, +Inf, -Inf} consistently to
-  base R. In case of some functions (like \emph{mean}), it will additionally
+  When multiple columns or multiple windows width are provided, then they
```
ben-schwen, line 55:
```suggestion
  When multiple columns or multiple window widths are provided, then they
```

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
   }
 }
 
+/* fast rolling max */
+void frollmax(unsigned int algo, double *x, uint64_t nx, ans_t *ans, int k, int align, double fill, bool narm, int hasna, bool verbose) {
+  if (nx < k) {
+    if (verbose)
+      snprintf(end(ans->message[0]), 500, _("%s: window width longer than input vector, returning all NA vector\n"), __func__);
```
ben-schwen, line 404:
```suggestion
      snprintf(end(ans->message[0]), 500, _("%s: window width longer than input vector, returning all %d vector\n"), __func__, fill);
```

jangorecki: int64 is not supported, only real..
`%d` format is for integer

ben-schwen: My bad thought I read smth about int64 support. changed to `%lf`

R/froll.R
```diff
@@ -2,8 +2,25 @@ froll = function(fun, x, n, fill=NA, algo=c("fast", "exact"), align=c("right", "
   stopifnot(!missing(fun), is.character(fun), length(fun)==1L, !is.na(fun))
   algo = match.arg(algo)
   align = match.arg(align)
+  leftadaptive = isTRUE(adaptive) && align=="left"  ## support for left added in #5441
+  if (leftadaptive) {
+    rev2 = function(x) if (is.list(x)) sapply(x, rev, simplify=FALSE) else rev(x)
+    verbose = getOption("datatable.verbose")
+    if (verbose)
+      cat("froll: adaptive=TRUE && align='left' pre-processing for align='right'\n")
```
MichaelChirico, line 10:
```suggestion
      catf("froll: adaptive=TRUE && align='left' pre-processing for align='right'\n")
```

NEWS.md
```diff
@@ -296,6 +296,8 @@
 
 41. New function `%notin%` provides a convenient alternative to `!(x %in% y)`, [#4152](https://github.com/Rdatatable/data.table/issues/4152). Thanks to Jan Gorecki for suggesting and Michael Czekanski for the PR. `%notin%` uses half the memory because it computes the result directly as opposed to `!` which allocates a new vector to hold the negated result. If `x` is long enough to occupy more than half the remaining free memory, this can make the difference between the operation working, or failing with an out-of-memory error.
 
+42. New function `frollmax` has been implemented. It applies `max` over a rolling window. Request came from @gpierard who needed 1left aligned, adaptive, rolling max, [#5438](https://github.com/Rdatatable/data.table/issues/5438). Adaptive rolling functions did not have support for `align="left"`, therefore this feature has been added as well. It works for other adaptive rolling functions now. Adaptive `frollmax` has observed to be up to 50 time faster than second fastest solution using `max` and grouping `by=.EACHI`.
```
MichaelChirico, line 299: update merge target to 1-15-99 and fix NEWS numbering?

MichaelChirico, line 299:
```suggestion
42. New function `frollmax` has been implemented. It applies `max` over a rolling window. Request came from @gpierard who needs left-aligned, adaptive, rolling max, [#5438](https://github.com/Rdatatable/data.table/issues/5438). Adaptive rolling functions did not have support for `align="left"`, therefore this feature has been added as well. It works for other adaptive rolling functions now. Adaptive `frollmax` has been observed to be up to 50 times faster than the next fastest solution using `max` and grouping `by=.EACHI`.
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+  are run in parallel. The exception is for \code{algo="exact"}, which runs in
+  parallel already. By default data.table uses only half of available CPUs,
+  see \code{\link{setDTthreads}} for details.
+}
+\section{\code{hasNA} argument}{
```
MichaelChirico, line 60: this part is coming from a later PR?

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+    for (int i=0; i<nx; i++) {
+      ans->dbl_v[i] = fill;
+    }
+    return;
+  }
+  double tic = 0;
+  if (verbose)
+    tic = omp_get_wtime();
+  if (algo==0) {
```
MichaelChirico, line 413: should `algo` be an `enum`, and we use `switch()` here?

It seems if we might add more `algo` options later, `enum` is the right storage type. o/w, if we only plan `0`/`1`, just do `if (algo == 0) { ... } else { ... }`

jangorecki: I doubt we will have more than 2 options. This isn't really practical, unless some algo would be very memory efficient vs another time efficient, vs nested loop. Within rolling functions that are in the scope I doubt we will face such problem.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      snprintf(end(ans->message[0]), 500, _("%s: align %d, shift answer by %d\n"), __func__, align, -k_);
+    memmove((char *)ans->dbl_v, (char *)ans->dbl_v + (k_*sizeof(double)), (nx-k_)*sizeof(double));
+    for (uint64_t i=nx-k_; i<nx; i++) {
+      ans->dbl_v[i] = fill;
+    }
+  }
+  if (verbose)
+    snprintf(end(ans->message[0]), 500, _("%s: processing algo %u took %.3fs\n"), __func__, algo, omp_get_wtime()-tic);
+}
+inline void windowmax(double *x, uint64_t o, int k, double *w, uint64_t *iw) {
```
MichaelChirico, line 430: what's `w`? `iw`? Maybe longer/descriptive names are appropriate here.

src/froll.c
jangorecki: `else if (algo==1)` has the advantage of being more verbose, so looking at the smaller chunk of code it is clear what is happening. Often github coversation tab will not show enough lines and one has to go to "Files" to see bigger picture. So that may be useful instead of a comment

MichaelChirico: I think `if/else` keeps it simple then. `if (algo ==0) { ... } else { // algo==1` if we want to be so explicit.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+  for (int i=0; i<k-1; i++) {
+    //Rprintf("windowmax iteration %d, offset %d, first x val %f, testing x[o+i-k+1] >= w[0]: x[%d-%d+1] >= w[0]: %f >= %f: %d\n", i, o, x[o], i, k, x[o+i-k+1], w[0], x[o+i-k+1] >= w[0]);
+    if (x[o+i-k+1] >= w[0]) { // what if that is never satisfied? test!
+      iw[0] = o+i-k+1;
+      w[0] = x[iw[0]];
+    }
+  }
+}
+inline void windowmaxnarm(double *x, uint64_t o, int k, bool narm, int *nc, double *w, uint64_t *iw) {
```
MichaelChirico, line 439: here, what's `nc`?

src/froll.c
jangorecki: `w` for "window's max", and `iw` for "index of a max value in window".
I would rather keep those for easier reading code inside the function rather than having long variable names.

src/froll.c
jangorecki: "NA counter" - you will probably find tens of `nc` in froll.c and frolladaptive.c :)

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+ * should scale well for bigger window size, may carry overhead for small window, needs benchmarking
+ * TODO NA handling
+ */
+void frollmaxFast(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool narm, int hasna, bool verbose) {
+  if (verbose)
+    snprintf(end(ans->message[0]), 500, _("%s: running for input length %"PRIu64", window %d, hasna %d, narm %d\n"), "frollmaxFast", (uint64_t)nx, k, hasna, (int)narm);
+  double w = R_NegInf; // window max
+  uint64_t cmax = 0; // counter of nested loops for verbose
+  uint64_t iw = 0; // index of window max
+  bool truehasna = hasna>0;
```
MichaelChirico, line 468: how about `bhasna` for consistency? e.g. with `badaptive`

src/frolladaptive.c
```diff
@@ -364,3 +364,80 @@ void fadaptiverollsumExact(double *x, uint64_t nx, ans_t *ans, int *k, double fi
...
+  if (verbose)
+    snprintf(end(ans->message[0]), 500, _("%s: running in parallel for input length %"PRIu64", hasna %d, narm %d\n"), "fadaptiverollmaxExact", (uint64_t)nx, hasna, (int) narm);
+  if (hasna==-1) { // fastest we can get for adaptive max as there is no algo='fast', therefore we drop any NA checks when hasNA=FALSE
+    #pragma omp parallel for num_threads(getDTthreads(nx, true))
+    for (uint64_t i=0; i<nx; i++) {
+      if (i+1 < k[i]) {
+        ans->dbl_v[i] = fill;
+      } else {
+        double w = R_NegInf;
+        for (int j=-k[i]+1; j<=0; j++) { //Rprintf("x[%d+%d] > w: %f > %f: %d\n", i, j, x[i+j], w, x[i+j] > w);
```
ben-schwen, line 395: the reason why we do not reuse `windowmax` and `windowmaxnrm` is that we do not need the `iw` argument, right?

NEWS.md
```diff
@@ -296,6 +296,8 @@
 
 41. New function `%notin%` provides a convenient alternative to `!(x %in% y)`, [#4152](https://github.com/Rdatatable/data.table/issues/4152). Thanks to Jan Gorecki for suggesting and Michael Czekanski for the PR. `%notin%` uses half the memory because it computes the result directly as opposed to `!` which allocates a new vector to hold the negated result. If `x` is long enough to occupy more than half the remaining free memory, this can make the difference between the operation working, or failing with an out-of-memory error.
 
+42. New function `frollmax` has been implemented. It applies `max` over a rolling window. Request came from @gpierard who needs left-aligned, adaptive, rolling max, [#5438](https://github.com/Rdatatable/data.table/issues/5438). Adaptive rolling functions did not have support for `align="left"`, therefore this feature has been added as well. It works for other adaptive rolling functions now. Adaptive `frollmax` has been observed to be up to 50 times faster than the next fastest solution using `max` and grouping `by=.EACHI`.
```
MichaelChirico, line 299:
```suggestion
42. New window function `frollmax` computes the rolling maximum. Request came from @gpierard who needs left-aligned, adaptive, rolling max, [#5438](https://github.com/Rdatatable/data.table/issues/5438). Adaptive rolling functions did not have support for `align="left"`, therefore we added this feature as well for all adaptive rolling functions. We measure adaptive `frollmax` to be up to 50 times faster than the next fastest solution using `max` and grouping `by=.EACHI`.
```

MichaelChirico: > 50 times faster than the next fastest solution using `max` and grouping `by=.EACHI`

IIUC that means the 2nd fastest solution is also implemented with {data.table}, maybe that should be highlighted / also mention a 3rd fastest solution using another package? Or link to a benchmark, or include a small one in the NEWS item.

man/froll.Rd
MichaelChirico: Oh, this is `\section{}` not `\arguments{}`.

src/froll.c
MichaelChirico, line 430:
```suggestion
inline void windowmax(double *x, uint64_t o, int k, /* max of current window */ double *w, /* index of w within window */ uint64_t *iw) {
```

MichaelChirico: In that case we can annotate the parameter

```suggestion
inline void windowmax(double *x, uint64_t o, int k, /* max of current window */ double *w, /* index of w within window */ uint64_t *iw) {
```

src/froll.c
MichaelChirico: Similarly we can annotate the parameter

```suggestion
inline void windowmaxnarm(double *x, uint64_t o, int k, bool narm, /* NA counter */ int *nc, double *w, uint64_t *iw) {
```

NEWS.md
jangorecki: In jest PR benchmark is there in the news file

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
-  all of \code{NaN, +Inf, -Inf} are treated as \code{NA}.
-  Setting \code{algo="exact"} will make rolling functions to use a more
-  computationally-intensive algorithm that suffers less from floating point
-  rounding error (the same consideration applies to \code{\link[base]{mean}}).
-  \code{algo="exact"} also handles \code{NaN, +Inf, -Inf} consistently to
-  base R. In case of some functions (like \emph{mean}), it will additionally
+  When multiple columns or multiple window widths are provided, then they
+  are run in parallel. The exception is for \code{algo="exact"}, which runs in
+  parallel already. By default data.table uses only half of available CPUs,
+  see \code{\link{setDTthreads}} for details.
```
MichaelChirico, line 58: This seems fragile, if we change the default eventually we have to chase down all references and fix them. 

Better IMO to just link `?setDTthreads` for details about parallelism in {data.table}.

```suggestion
  parallel already. See \code{\link{setDTthreads}} for defaults and further details on parallelism in data.table.
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+}
+\section{\code{hasNA} argument}{
+  \code{hasNA} can be used to speed up processing in cases when it is known
+  if \code{x} contains \code{NA, NaN, +Inf, -Inf}.
```
MichaelChirico, line 62:
```suggestion
  whether \code{x} contains infinite values \code{NA, NaN, +Inf, -Inf}.
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+  \itemize{
+    \item{ Default \code{hasNA=NA} will use faster NA inaware implementation,
```
MichaelChirico, line 64:
```suggestion
    \item{ Default \code{hasNA=NA} will use faster \code{NA} agnostic implementation,
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+    but when NAs are detected it will re-run NA aware implementation. }
```
MichaelChirico, line 65:
```suggestion
    but when \code{NA}s are detected it will re-run \code{NA} aware implementation. }
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+    \item{ \code{hasNA=TRUE} will use NA aware implementation straightaway. }
+    \item{ \code{hasNA=FALSE} will use faster NA inaware implementation.
+    Then depending on the rolling function it will either
+    \itemize{
+      \item{ (\emph{mean, sum}) detect NAs, raise warning, re-run NA aware. }
+      \item{ (\emph{max}) not detect NAs and may silently produce incorrect
+      answer. }}
```
MichaelChirico, line 72:
```suggestion
    \item{ \code{hasNA=TRUE} will use \code{NA} aware implementation straightaway. }
    \item{ \code{hasNA=FALSE} will use faster \code{NA} agnostic implementation.
    Then depending on the rolling function it will either
    \itemize{
      \item{ (\emph{mean, sum}) detect \code{NA}s, raise warning, re-run \code{NA} aware. }
      \item{ (\emph{max}) not detect \code{NA}s and may silently produce an incorrect
      answer. }}
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+    Therefore \code{hasNA=FALSE} should be used with care.
+    }
+  }
+}
+\section{Implementation}{
+  \itemize{
+    \item{ \code{algo="fast"} uses \emph{"on-line"} algorithm, and
+  all of \code{NaN, +Inf, -Inf} are treated as \code{NA}. Not all functions
+  have \emph{fast} implementation available. As of now
+  \emph{max} and \code{adaptive=TRUE} does not have, therefore it will
```
MichaelChirico, line 82:
```suggestion
  \emph{max} and \code{adaptive=TRUE} do not have it, therefore it will
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+  automatically fall back to \emph{exact} implementation. \code{datatable.verbose}
+  option can be used to check that. }
+    \item{ \code{algo="exact"} will make rolling functions to use a more
```
MichaelChirico, line 85:
```suggestion
    \item{ \code{algo="exact"} will make rolling functions use a more
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+  computationally-intensive algorithm. For each observation from input vector
+  it will compute a function on a window from scratch (complexity \eqn{O(n^2)}).
+  Depeneding on the function, this algorithm may suffers less from
+  floating point rounding error (the same consideration applies to base \code{\link[base]{mean}}).
+  Algorithm also handles \code{NaN, +Inf, -Inf} consistently to
+  base R, unless - for some functions (e.g. \emph{max}) - \code{hasNA}is \code{FALSE}
```
MichaelChirico, line 91:
```suggestion
  base R, unless - for some functions (e.g. \emph{max}) - \code{hasNA} is \code{FALSE}
```

man/froll.Rd
```diff
@@ -45,36 +49,70 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
+  argument must be a list, then each list element must be an integer vector
+  of window sizes corresponding to every single observation in each column;
+  see Examples. Due to the logic or implementation of adaptive rolling
+  functions, the following restrictions apply:
   \itemize{
-    \item \code{align} only \code{"right"}. 
+    \item \code{align} does not support \code{"center"}.
     \item if list of vectors is passed to \code{x}, then all
       vectors within it must have equal length.
+    \item functionality in not suported in \code{frollapply}.
```
MichaelChirico, line 108:
```suggestion
    \item functionality is not suported in \code{frollapply}.
```

man/froll.Rd
```diff
@@ -84,32 +122,34 @@ frollapply(x, n, FUN, \dots, fill=NA, align=c("right", "left", "center"))
...
-}
-\note{
+\section{\code{zoo} package users notice}{
   Users coming from most popular package for rolling functions
   \code{zoo} might expect following differences in \code{data.table}
   implementation.
   \itemize{
-    \item rolling function will always return result of the same length as input.
-    \item \code{fill} defaults to \code{NA}. 
+    \item rolling function will always return result of the same length
```
MichaelChirico, line 129:
```suggestion
    \item rolling functions will always return result of the same length
```

jangorecki:
I am worried that by those small fixes we are just adding up merge conflicts to next 14 branches that forks from this one. Documentation probably was heavily rewritten 2 times in the there. At some point splitting frollapply into own manual page.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
   }
 }
 
+/* fast rolling max */
+void frollmax(unsigned int algo, double *x, uint64_t nx, ans_t *ans, int k, int align, double fill, bool narm, int hasna, bool verbose) {
+  if (nx < k) {
+    if (verbose)
+      snprintf(end(ans->message[0]), 500, _("%s: window width longer than input vector, returning all %f vector\n"), __func__, fill);
```
MichaelChirico, line 404: Maybe better for follow-up:

Wouldn't the expected behavior here be to just set `k=nx` instead & proceed?

Why should this have different output (pseudocode):

```r
x = c(1, 4, 5, 6, 2, 3, 4)
frollmax(x, k = 7) # right-aligned
# 1 4 5 6 6 6 6
frollmax(x, k = 8) # what I'd expect
# 1 4 5 6 6 6 6
```

MichaelChirico: Oh, I mis-read a bit. I see, it does give the same output, namely `rep(fill, length(x))` for both cases. CMIIW, resolving for now.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+  }
+  double tic = 0;
+  if (verbose)
+    tic = omp_get_wtime();
+  if (algo==0) {
+    frollmaxFast(x, nx, ans, k, fill, narm, hasna, verbose);
+  } else { // algo==1
+    frollmaxExact(x, nx, ans, k, fill, narm, hasna, verbose);
+  }
+  if (ans->status < 3 && align < 1) {
```
MichaelChirico, line 418: `ans->status < 3` doesn't really mean much to me as code reader, how about

```suggestion
if (ans->status < /* error= */ 3 && align < 1) {
```

or

```suggestion
if (ans->status < 3 && align < 1) { // see types.h
```

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      ans->dbl_v[i] = fill;
+    }
+  }
+  if (verbose)
+    snprintf(end(ans->message[0]), 500, _("%s: processing algo %u took %.3fs\n"), __func__, algo, omp_get_wtime()-tic);
+}
+inline void windowmax(double *x, uint64_t o, int k, /* max of current window */ double *w, /* index of w within window */ uint64_t *iw) {
+  for (int i=0; i<k-1; i++) {
+    //Rprintf("windowmax iteration %d, offset %d, first x val %f, testing x[o+i-k+1] >= w[0]: x[%d-%d+1] >= w[0]: %f >= %f: %d\n", i, o, x[o], i, k, x[o+i-k+1], w[0], x[o+i-k+1] >= w[0]);
+    if (x[o+i-k+1] >= w[0]) { // what if that is never satisfied? test!
```
MichaelChirico, line 433: Why `w[0]` over `*w`? Just style? Ditto for `iw[0]`

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      iw[0] = o+i-k+1;
+      w[0] = x[iw[0]];
+    }
+  }
+}
+inline void windowmaxnarm(double *x, uint64_t o, int k, bool narm, /* NA counter */ int *nc, double *w, uint64_t *iw) {
+  for (int i=0; i<k-1; i++) {
+    //Rprintf("windowmax iteration %d, offset %d, first x val %f, testing x[o+i-k+1] >= w[0]: x[%d-%d+1] >= w[0]: %f >= %f: %d\n", i, o, x[o], i, k, x[o+i-k+1], w[0], x[o+i-k+1] >= w[0]);
+    if (R_FINITE(x[o+i-k+1])) {
```
MichaelChirico, line 442: `o+i-k+1` is written three times here, store it?

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+ * TODO NA handling
+ */
+void frollmaxFast(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool narm, int hasna, bool verbose) {
+  if (verbose)
+    snprintf(end(ans->message[0]), 500, _("%s: running for input length %"PRIu64", window %d, hasna %d, narm %d\n"), "frollmaxFast", (uint64_t)nx, k, hasna, (int)narm);
+  double w = R_NegInf; // window max
+  uint64_t cmax = 0; // counter of nested loops for verbose
+  uint64_t iw = 0; // index of window max
+  bool truehasna = hasna>0;
+  if (!truehasna) {
```
MichaelChirico, line 469: IIUC this body is structured as

```c
if (!truehasna) {
  ...
}
if (truehasna) {
  ...
}
```

Why not `if / else`? Why not `if (truehasna) { ... } else ...` (i.e. flip branch order)?

MichaelChirico: NVM, I see `truehasna` might be edited within the first branch.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      iw[0] = o+i-k+1;
+      w[0] = x[iw[0]];
+    }
+  }
+}
+inline void windowmaxnarm(double *x, uint64_t o, int k, bool narm, /* NA counter */ int *nc, double *w, uint64_t *iw) {
+  for (int i=0; i<k-1; i++) {
+    //Rprintf("windowmax iteration %d, offset %d, first x val %f, testing x[o+i-k+1] >= w[0]: x[%d-%d+1] >= w[0]: %f >= %f: %d\n", i, o, x[o], i, k, x[o+i-k+1], w[0], x[o+i-k+1] >= w[0]);
+    if (R_FINITE(x[o+i-k+1])) {
+      if (x[o+i-k+1] >= w[0]) { // what if that is never satisfied? test!
```
MichaelChirico, line 443: why not add a helper for updating the max, e.g. roughly)

```c
void maybe_update_max(double *x, uint64_t o, double *w, uint64_t *iw) {
  if (x[o] >= *w) {
    *iw = o;
    *w = x[0]
  }
  return;
}
```

it looks like it would be re-used several times, and you can centralize the comment about using `>=` instead of `>` that's only used once below.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+  bool truehasna = hasna>0;
+  if (!truehasna) {
+    int i;
+    for (i=0; i<k-1; i++) { // #loop_counter_not_local_scope_ok
+      if (!R_FINITE(x[i])) {
+        truehasna = true;
+        break;
+      }
+      if (x[i] >= w) { // >= rather than > because we track most recent maximum using iw
+        iw = i;
```
MichaelChirico, line 477: i guess worst-case performance for this is on `seq_len(N)` since it needs to update two objects on every iteration. nothing to do in this PR, but making a note as relevant for any benchmarks.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+        w = x[iw];
+      } else {
+        // ensure iw ok, case when all window was -Inf or NA?
+      }
+      ans->dbl_v[i] = w;
+    } else {
+      truehasna = true;
+    }
+    if (!truehasna) {
+      for (uint64_t i=k; i<nx; i++) {
```
MichaelChirico, line 494: why is `i` `uint64_t` here but `int` above?

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+        truehasna = true;
+        break;
+      }
+      if (x[i] >= w) { // >= rather than > because we track most recent maximum using iw
+        iw = i;
+        w = x[iw];
+      }
+      ans->dbl_v[i] = fill;
+    }
+    if (!truehasna && !R_FINITE(x[i])) {
```
MichaelChirico, line 482: this is handling the case `i=k-1`, is that right? why is it separated out? seems like a comment is needed here

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      for (uint64_t i=k; i<nx; i++) {
+        if (!R_FINITE(x[i])) {
+          truehasna = true;
+          break;
+        }
+        if (x[i] >= w) {
+          //Rprintf("iteration %d, new max %f at %d, old max %f at %d\n", i, x[i], i, w, iw);
+          iw = i;
+          w = x[iw];
+        } else {
```
MichaelChirico, line 503: not a fan of the empty + nested `if` here, why not

```c
if (x[i] >= w) {
  ...
} else if (iw <= i-k) {
  ...
}
```

src/froll.c
jangorecki: Yes. When it is scalar I prefer [0] over *

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      ans->dbl_v[i] = narm ? R_NegInf : NA_REAL;
+    } else {
+      ans->dbl_v[i] = narm ? w : NA_REAL;
+    }
+    for (uint64_t i=k; i<nx; i++) {
+      if (R_FINITE(x[i])) {
+        if (x[i] >= w) {
+          iw = i;
+          w = x[iw];
+        } else {
```
MichaelChirico, line 575: ditto re: unnecessary nesting

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+          if (iw > i-k) { // max is still within window
+            // w and iw are still ok
+          } else { // max left the window, need to find max in this window
+            w = R_NegInf;
+            iw = i-k;
+            windowmaxnarm(x, i, k, narm, &nc, &w, &iw);
```
MichaelChirico, line 581: here, maybe the worst case is something like

```
1 0 0 0 0 0 ... 1 0 0 .... 1 0 0 ...
```

Where `1` only shows up every `k+1` entries, so that we always have to double-pass the vector.

Again nothing actionable in this PR, just checking my own understanding+filing note for benchmarking.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+            //Rprintf("iteration %d, windowmaxnarm found new max %f at %d\n", i, w, iw);
+            cmax++;
+          }
+        }
+        ans->dbl_v[i] = w;
+      } else {
+        nc++;
+      }
+      if (!R_FINITE(x[i-k])) {
```
MichaelChirico, line 590: add a comment for this? I've lost track of why this might be here.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      }
+    } else if (narm) {
+      nc[0]++;
+    } else {
+      w[0] = NA_REAL;
+    }
+  }
+}
+/* fast rolling max - fast
+ * fast online algorithm do single pass over elements keeping track of recent max and its index
```
MichaelChirico, line 455: Another algorithm that comes to mind and avoids the need to double-pass the vector ever, maybe you've seen it implemented elsewhere?

 1. Compute `o=order(x)`
 2. Set output to the value where `x[o[1]]` for `o[1], ..., o[1]+k-1`
 3. Find the next `o[i]` outside `o[1], ..., o[1]+k-1`
 4. Repeat until `o` is "empty"

Maybe it will perform poorly since it requires sorting --> best case it's O(n log n), whereas IINM this algorithm is best case `O(n)` (?)

But maybe it will do better in the worst case?

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+    snprintf(end(ans->message[0]), 500, _("%s: running in parallel for input length %"PRIu64", window %d, hasna %d, narm %d\n"), "frollmaxExact", (uint64_t)nx, k, hasna, (int)narm);
+  for (int i=0; i<k-1; i++) {
+    ans->dbl_v[i] = fill;
+  }
+  bool truehasna = hasna>0;
+  if (!truehasna || !narm) {
+    #pragma omp parallel for num_threads(getDTthreads(nx, true))
+    for (uint64_t i=k-1; i<nx; i++) {
+      if (narm && truehasna) {
+        continue;
```
MichaelChirico, line 620: Why `continue` instead of `break`? I might be missing something...

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
...
+      }
+      double w = R_NegInf;
+      for (int j=-k+1; j<=0; j++) {
```
MichaelChirico, line 623: Maybe more readable as `int j=k-1; j>=0; j--` and `x[i-j]` instead?

src/froll.c
MichaelChirico: Because `k` is set in R as `integer`, whereas `x` could be a long vector, right? Probably worth a comment.

src/froll.c
jangorecki: To draw more accurate conclusions it is probably better to look at final implementation in frollmax branch rather than frollmax1 as some parts have been simplified later on.

jangorecki: Also manual explains edge case where naive approach (nested loop) algo=exact could ended up to be faster than algo=fast

src/froll.c
jangorecki: Because it is inside parallel

src/froll.c
jangorecki: Yes, k must not be bigger than int32 and x length can be int64

src/froll.c
ben-schwen: `break` even gives a compile error

src/froll.c
MichaelChirico: I worry that "saving" 8 bytes on `uint64_t` for part of the loop here is over-optimizing at the expense of readability. Yes, we _can_ use `int` for the first `k` elements, but using a consistent storage type will prevent confusion/needing to know R implementation details like this.

In other words the cognitive load of using `int` is extremely high IMO. And "saving" is in quotes because it's more like "reducing RAM load by a tiny bit for a tiny fraction of time before the heavier loop kicks in".

src/froll.c
MichaelChirico: Thanks, noted. Resolving as nothing to do in this PR.

src/froll.c
MichaelChirico: A bit surprising. I guess it's supposed to be well-known for programming with openmp.

src/froll.c
MichaelChirico: That makes sense. Please add a reference comment here, though. I would have no clue about that if you hadn't mentioned it.

src/froll.c
jangorecki: The purpose was that the code to be reflecting what is precisely happening rather than saving memory. For understanding the code I think it is better to keep it int when int is the type really used/needed. I can imagine it may be additionally confusing when we start to use data types that are not matching use case. So the purpose is rather for the code to be self descriptive rather than a memory optimization.

src/froll.c
jangorecki: in `frollmean` function this line is commented well: https://github.com/Rdatatable/data.table/blob/0773c44edac01b326e06ecbaffe2914f4103851d/src/froll.c#L29
I don't think it make sense to copy comments to all function when implementation pattern is so similar. As explained before in https://github.com/Rdatatable/data.table/pull/5889/files/63ea485ed24eb3c7ec75f3e57f0f0b4681b76165#r1447373236

src/froll.c
jangorecki: this is addressed in later PRs: https://github.com/Rdatatable/data.table/blob/fff2b08a5a111f7115110aeb6ecc9457763247f9/src/froll.c#L431
this function has been rewritten and merged with non-narm version into single function

src/froll.c
jangorecki: `windowmaxnarm` (and in later PRs `wmax`) is already a helper (inline) function.
Where else you see it could be re-used? I don't see any other obvious place.

src/froll.c
jangorecki: it is not worst-case AFAIU because it doesn't have to recalculate window's max. Each next observation is new max so it does single pass. Updating two variables is not a problem, comparing to recalculating window's max, when current max going out of the window.

src/froll.c
jangorecki: first `k-1` are in their own loop to avoid extra branching in the main loop. The whole loop is split into two loops, rather than having single loop with 2 branches inside. This must be well commented in `frollmean`. The branch you are pointing here is an early check on non-finite values - if non-finite have been detected already in first `k-1` obs, there is no need to run non-nf main loop, but rather skip directly to an nf-aware main loop. This should be well commented in frollmean as well. It is the same pattern for all rolling functions except median.

src/froll.c
jangorecki: this was later removed, latest version does not have that extra nesting anymore: https://github.com/Rdatatable/data.table/blob/fff2b08a5a111f7115110aeb6ecc9457763247f9/src/froll.c#L499-L536C6

src/froll.c
jangorecki: this was later removed, latest version does not have that extra nesting anymore: https://github.com/Rdatatable/data.table/blob/fff2b08a5a111f7115110aeb6ecc9457763247f9/src/froll.c#L499-L536C6

src/froll.c
jangorecki: This implemented is still kind of "online"-ish implementation (single pass). From time to time (whenever max leaves current window) extra recomputation of max is called. Calling `order()` alone I think will already be more expensive than majority of cases using online-ish implementation, as `order` alone won't run in single pass.

src/froll.c
jangorecki: Could be. I recall I struggle with such a simple change somewhere, not sure if it was here though. Anyway I think it make sense to keep it as is to match implementation way in other rolling functions, and then eventually change them all at once.

src/froll.c
MichaelChirico: > From time to time

The key IMO is how often this happens in practice. A few times per input, OK; hundreds of times, could be quite costly.

Anyway I'll mark for follow-up: #5903

src/froll.c
jangorecki: that could ended up every `if (...) {` line to have comment like `// see frollmean comments for details`

MichaelChirico: Please don't push back on completely reasonable requests for better-documented code like this.

Just one comment at the top of the main entry point to individual rolling functions should be enough. That would be where a new reader could logically be expected to start reading about implementation of individual rolling function.

If that wasn't clear from my comment, please take initiative to come up with such compromises. Completely undocumented code is not acceptable; neither is expecting every new reader of the code to ingest 1000s of lines of C code to figure out what's going on / which secret compartment contains the real comments.

src/froll.c
MichaelChirico: Makes sense, thanks: #5904

src/froll.c
MichaelChirico: I don't find the code self-descriptive.

> int is the type really used/needed

How so? It's used two ways: (1) index a long array. (2) RHS of assignment to a `uint64_t` variable (`iw  = i`), e.g.

https://github.com/Rdatatable/data.table/pull/5889/files#diff-6eb444de657f71e8b89b3d686c4ec70e4ecff4e33154481d0b1b416d09449f1eR477

So it seems to me `uint64_t` is the type really used/needed.

> I can imagine it may be additionally confusing when we start to use data types that are not matching use case.

Not sure I follow, but if you're really married to `int`, please add a comment (inside `frollmax`)

src/froll.c
jangorecki: this line is where `i = k-1`, related comments are in https://github.com/Rdatatable/data.table/blob/45904a7c7d4b49775b4ef8f6b1c2d7b0967793f7/src/froll.c#L57-L75
although this has been rewritten in later PRs and now looks completely different: https://github.com/Rdatatable/data.table/blob/fff2b08a5a111f7115110aeb6ecc9457763247f9/src/froll.c#L467C8-L481

MichaelChirico: Thanks! I'll defer comments till then.

src/froll.c
jangorecki: `iw` is `uint64_t` so we don't need to re-declare it again. with `i` we declare it outside of the `for()` because it is needed after the loop finishes, and setting that to int64 upfront could give false impression that can be go over int64 range. I will add comments to explain that.

src/froll.c
jangorecki: added comments in https://github.com/Rdatatable/data.table/pull/5889/commits/03af0e30f1a6a9e75f82b5827c1078f42db48e45

MichaelChirico: Sorry, perhaps my suggestion was not clear enough. Adding a suggestion in a separate comment.

src/froll.c
```diff
@@ -397,6 +397,277 @@ void frollsumExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
   }
 }
 
+/* fast rolling max */
```
MichaelChirico, line 400:
```suggestion
/* fast rolling max */
// See frollmax() for detailed comments relevant to the below
```

src/froll.c
jangorecki: ok, added comment as well to frollsum top entry point

MichaelChirico (approved):
Thanks!

MichaelChirico:
Looks like `covr::package_coverage()` fails after this PR; `quiet=FALSE` gives a better error trace:

```
** testing if installed package can be loaded
Error: package or namespace load failed for ‘data.table’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/tmp/RtmpmqW3UL/R_LIBS35452f231af8/data.table/libs/data_table.so':
  /tmp/RtmpmqW3UL/R_LIBS35452f231af8/data.table/libs/data_table.so: undefined symbol: windowmax
```

I can't reproduce the issue outside {covr}, though, maybe it's a bug there? Something to do with an `inline void` declaration?

Anyway, just noting the issue for now -- presumably it is fixed later in the chain. Ignoring the covr reports in `1-15-99` for now.

jangorecki:
Yes, it was fixed in this commit https://github.com/Rdatatable/data.table/pull/5441/commits/c5caf2c7d5c78f62cf3e31afb5e8c28afecad418 later on

MichaelChirico:
Shall we just apply that as its own PR then? To get code coverage back working on `1-15-99`

PR #5890, opened 2024-01-08
frollmax2: code reorg, docs, tests

jangorecki:
- code reorganization by using C `frollfun` routing function
- news
- docs
- tests


MichaelChirico:
I've set the merge target here to frollmax1, would you mind doing similar for the subsequent PRs to make the relative diffs easier to see?

jangorecki:
Yes, we can do it, but are we going to merge from 10->9->...->1 ? Or we switch back to master before merging?

MichaelChirico:
> Yes, we can do it, but are we going to merge from 10->9->...->1 ? Or we switch back to master before merging?

once the merge target is deleted GH will automatically set master as the new target.

current form makes it hard to parallelize review, would only really make sense to review serially

NEWS.md
```diff
@@ -296,7 +296,40 @@
...
+42. Multiple improvements has been added to rolling functions. Request came from @gpierard who needed left aligned, adaptive, rolling max, [#5438](https://github.com/Rdatatable/data.table/issues/5438). There was no `frollmax` function yet. Adaptive rolling functions did not have support for `align="left"`. `frollapply` did not support `adaptive=TRUE`. Available alternatives were base R `mapply` or self-join using `max` and grouping `by=.EACHI`. As a follow up of his request, following features has been added:
+- new function `frollmax`, applies `max` over a rolling window.
+- support for `align="left"` for adaptive rolling function.
+- support for `adaptive=TRUE` in `frollapply`.
+- better support for non-double data types in `frollapply`.
+- better support for `Inf` and `-Inf` support in `algo="fast"` implementation.
+- `partial` argument to trim window width to available observations rather than returning `NA` whenever window is not complete.
```
jangorecki, line 305: This refers to future PR

src/frolladaptive.c
```diff
@@ -1,32 +1,53 @@
...
+    if (algo==0) {
+      frolladaptivesumFast(x, nx, ans, k, fill, narm, hasna, verbose);
+    } else if (algo==1) {
+      frolladaptivesumExact(x, nx, ans, k, fill, narm, hasna, verbose);
+    }
+    break;
+  case MAX :
+    if (algo==0 && verbose) {
+      //frolladaptivemaxFast(x, nx, ans, k, fill, narm, hasna, verbose); // frolladaptivemaxFast does not exists as of now
+      snprintf(end(ans->message[0]), 500, _("%s: algo %u not implemented, fall back to %u\n"), __func__, algo, (unsigned int) 1);
```
ben-schwen, line 32: algo 0 and algo 1 are less explanatory than algo fast and algo exact.
Especially for the case of frolladaptivemax users might be surprised what 0 and 1 mean.

jangorecki: Yes, it was like this for simplicity rather than being informative. Rolling functions don't have verbose argument for that reason, as those messages are more technical ones. Reason to introduce them was ability to test which variant is running.

jangorecki: If we want to change we could do it in separate PR as it is not related to changes in this PRs

src/froll.c
```diff
@@ -1,43 +1,65 @@
...
+    }
+    break;
+  case MAX :
+    if (algo==0) {
+      frollmaxFast(x, nx, ans, k, fill, narm, hasna, verbose);
+    } else if (algo==1) {
+      frollmaxExact(x, nx, ans, k, fill, narm, hasna, verbose);
+    }
+    break;
+  default:
```
MichaelChirico, line 47: AFAIK this should be excluded -- `rollfun_t` `typedef` enumerates every possible value. If we change the set of values in the future, IINM that will be a compiler warning if any type is omitted.

src/frolladaptive.c
MichaelChirico: marked for follow-up: #5902

inst/tests/froll.Rraw
MichaelChirico, line 1: seems like the tests should be their own PR?

MichaelChirico:
PR overall looks good (great change!), I'll give final approval once `frollmax1` is finalized to see the diff on NEWS/Rd more clearly

MichaelChirico:
> Still reviews could happen on 10 smaller PRs but none of them would needed to be merged

Unfortunately it's not so simple, with the evolving diffs inherited from the earlier PRs, it will be increasingly hard to keep track of what the actual current state is as a reviewer.

Did you get a chance to check out any of the recommended tools for PR stacking? The whole point there was to delegate these downstream merge conflicts to the tools, did that not work out as hoped?

jangorecki:
yes
https://graphite.dev/docs/squash-fold-split
`gt split --by-commit` is doing exactly what I made to split `frollmax` branch

MichaelChirico:
Looks like `gt modify` is the next step? https://graphite.dev/docs/update-mid-stack-branches

MichaelChirico:
If `gt modify` is not working out (😞), cherry-picking commits added during review should work right? possibly with some rebase+force push. If so I can try to handle that part since I insisted on breaking up the PRs :)

MichaelChirico:
Hi @jangorecki, we are getting close enough to the target release date (~3 months) that it would be nice to get an update on your availability to work on this & the other big PRs you have lined up for 1.16.0. Thanks!

jangorecki:
Does not look any better now. Already preparing all those PRs were weeks of my volunteering time and I cannot afford now more just for resolving conflicts. This is easy part that community could follow up with.

MichaelChirico:
> just for resolving conflicts

> gt split --by-commit is doing exactly what I made to split frollmax branch

Re-reading, it looks like you've just done manually what `gt split` does within the Graphite. But the advantage of the Graphite is that it manages full stacks, including automatically resolving conflicts in up-stack PRs as down-stack PRs evolve.

Time to do conflict resolution aside, there are substantive comments on #4386 to address, is that something you plan to dedicate time to?

jangorecki:
Yes, there are many comments there but they are purely cosmetic changes, like adding helper or paste0 instead or paste. Is there any change of the logic there?

MichaelChirico:
"Cosmetics" are also important, Jan. It is also your duty as code author to improve that aspect of your PRs as requested by reviewers.

jangorecki:

PR #5895, opened 2024-01-08
frollmax7: arg rename, use macros, Inf support

jangorecki:
- argument rename
- simplify code with macros
- Inf support


jangorecki:

PR #5896, opened 2024-01-08
frollmax8: deduplicate args handling

jangorecki:
- frollfun router function
- deduplicate N arg handling
- deduplicate partial and left adaptive logic


jangorecki:

PR #5899, opened 2024-01-08
ctb->aut for Committers

tdhock:
hi @jangorecki @MichaelChirico according to governance, we are committers, so we should have "aut" role in DESCRIPTION, as I propose here.


tdhock:
I don't believe it is necessary for us to list our emails, so I did not.

MichaelChirico (approved):

PR #5905, opened 2024-01-12
Move to R3.2.0

MichaelChirico:
Pared down #5838, per discussion https://github.com/Rdatatable/data.table/pull/5838#issuecomment-1881251500

PRs targeting `1-15-99` can now use R 3.2.0 features 🎉


jangorecki:
Change in GLCI is also needed to upgrade 3.1 job

Issue #4697, opened 2020-09-07
Be less conservative about ALTREP


MichaelChirico:
Discovered via SO Q&A:

https://stackoverflow.com/a/63770450/3576984

Currently `j` expands _all_ `ALTREP`s, which can lead to significant performance degradation:

```
system.time(for (i in 1:40732170) {})
#    user  system elapsed 
#   0.764   0.004   0.880 
data.table()[ , system.time(for (i in 1:40732170) {})]
#    user  system elapsed 
#   3.756   0.020   3.927
```

I don't know enough about `ALTREP` to say for sure which cases we can/can't avoid expansion, but hopefully we can do better than the above.

ColeMiller1:
Should we also be more aggressive about using ```ALTREP```? That is, ```.I``` is almost free now. ```irows``` could be refactored from being ```NULL``` and might (?) end up in simpler code.

R 3.5.0 is 2.5 years old so I think we're getting close to where incorporating it makes sense.

MichaelChirico:
Great talk from Gabe Becker here:

https://youtu.be/8i7ziLqsE2s

I think a low hanging fruit could be to set (first) keys to KNOWN_INCREASING, that could e.g. let R by very fast in doing `unique(DT$key)`.

jangorecki:
Unless there are NAs which R places at the end and not at the front.

MichaelChirico:
Actually no! one of the sortedness enum fit our case if I'm not mistaken: `SORTED_INCR_NA_1ST`

link to slides here:

https://www.bioconductor.org/help/course-materials/

ben-schwen:
Well if we manage to support altrep vectors as columns then out-of-memory data.table should be easier targetable.

tdhock:
is there a more convincing real example?
it seems like in a real example if you are computing something over a large loop you will also be returning a large number of rows, and you should have memory for that, so optimizing for ALTREP is not necessary.

jangorecki:
out-of-memory data.table is very desired feature due to severity of consequences when you just don't have enough memory. Together with long vector support these were two main points on the roadmap for data.table new features that Matt was working on. As he said, long vector support is not that great deal as long as there is no out-of-memory data.table.

PR #5907, opened 2024-01-13
implement frev as fast base::rev alternative

ben-schwen:
- [x] implement
- [ ] test
- [ ] docs


ben-schwen:
The goal is to export right? Should we also support `data.frame` and `data.table`? Matrix?

NEWS.md
```diff
@@ -293,7 +293,7 @@
 
 41. `tables()` is faster by default by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). `mb=` now accepts any function which accepts a `data.table` and returns a higher and better estimate of its size in bytes, albeit more slowly; e.g. `mb = utils::object.size`.
 
-42. `base::rev` gains a fast method for atomic vectors and classes solely consisting of a single atomic vector and attributes, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Thanks to Benjamin Schwendinger for suggesting and implementing.
+42. `base::rev` gains a fast method `frev(x, copy)` for atomic vectors, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Thanks to Benjamin Schwendinger for suggesting and implementing.
```
jangorecki, line 296:
```suggestion
42. `base::rev` gains a fast method `frev(x, copy)` for atomic vectors/list, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Thanks to Benjamin Schwendinger for suggesting and implementing.
```

ben-schwen:
Should be prohibit use on matrix?
```r
x = matrix(1:9, nrow=3L)
x
#      [,1] [,2] [,3]
# [1,]    1    4    7
# [2,]    2    5    8
# [3,]    3    6    9
frev(x)
#      [,1] [,2] [,3]
# [1,]    9    6    3
# [2,]    8    5    2
# [3,]    7    4    1
rev(x)
# [1] 9 8 7 6 5 4 3 2 1
```

NEWS.md
```diff
@@ -305,10 +305,10 @@
...
     )
     # Unit: seconds
-    #          expr        min         lq       mean     median         uq        max neval cld
-    #          base 1.37616865 1.39689379 1.86418194 1.54417404 1.91734152 4.27381571    10 a  
-    #     frev_copy 0.52941553 0.59072669 0.76882781 0.65942025 0.72746099 1.35119930    10  b 
-    #  frev_inPlace 0.06466391 0.06529963 0.06621053 0.06580863 0.06682197 0.06955259    10   c
+    #          expr   min    lq  mean median   uq   max neval cld
+    #          base 1.376 1.397 1.864 1.544 1.917 4.274    10 a  
+    #     frev_copy 0.529 0.591 0.769 0.659 0.727 1.351    10  b 
+    #  frev_inPlace 0.064 0.065 0.066 0.066 0.067 0.070    10   c
```
jangorecki, line 311: this is mixing two naming conventions in single name, I think it is ok to simply `fre_inplace`

NEWS.md
```diff
@@ -293,6 +293,24 @@
...
+      frev_copy = frev(x, copy=TRUE),
+      frev_inPlace = frev(x, copy=FALSE),
+      times = 10L,
+      unit = "s"
+    )
+    # Unit: seconds
+    #          expr   min    lq  mean median   uq   max neval cld
+    #          base 1.376 1.397 1.864 1.544 1.917 4.274    10 a  
+    #     frev_copy 0.529 0.591 0.769 0.659 0.727 1.351    10  b 
+    #  frev_inPlace 0.064 0.065 0.066 0.066 0.067 0.070    10   c
```
ben-schwen, line 311:
```suggestion
    #  frev_inplace 0.064 0.065 0.066 0.066 0.067 0.070    10   c
```

jangorecki:
> Should we prohibit use on matrix?
> 
> ```r
> x = matrix(1:9, nrow=3L)
> x
> #      [,1] [,2] [,3]
> # [1,]    1    4    7
> # [2,]    2    5    8
> # [3,]    3    6    9
> frev(x)
> #      [,1] [,2] [,3]
> # [1,]    9    6    3
> # [2,]    8    5    2
> # [3,]    7    4    1
> rev(x)
> # [1] 9 8 7 6 5 4 3 2 1
> ```

I like how frev handles the matrix but lack of consistency to base R in this may be a problem. I would say make sense to prohibit.

ben-schwen:
> > Should we prohibit use on matrix?
> > ```r
> > x = matrix(1:9, nrow=3L)
> > x
> > #      [,1] [,2] [,3]
> > # [1,]    1    4    7
> > # [2,]    2    5    8
> > # [3,]    3    6    9
> > frev(x)
> > #      [,1] [,2] [,3]
> > # [1,]    9    6    3
> > # [2,]    8    5    2
> > # [3,]    7    4    1
> > rev(x)
> > # [1] 9 8 7 6 5 4 3 2 1
> > ```
> 
> I like how frev handles the matrix but lack of consistency to base R in this may be a problem. I would say make sense to prohibit.

Me too, that's why I didn't prohibit it in the first place. Anyway, reversing a matrix does not seem to be a quite common operation (just checked that there is not even a single Matlab command for it, but flipup + transpose should do)

R/utils.R
```diff
@@ -165,4 +165,3 @@ rss = function() {  #5515 #5517
   round(ans / 1024, 1L)  # return MB
   # nocov end
 }
```
ben-schwen, line 167:
```suggestion
}

```

src/utils.c
```diff
@@ -424,3 +424,104 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
...
+    error(_("%s must be TRUE or FALSE."), "copy");
+  int n = LENGTH(x);
+  bool copy = LOGICAL(copyArg)[0];
+  int nprotect = 0;
+  if (copy) {
+    x = PROTECT(duplicate(x));
+    nprotect++;
+    copy = false;
+  }
+  if (!copy) {
```
jangorecki, line 442: Always true

tdhock:
Can you please give a specific example (with system.time/etc) where frev should be faster than base::rev? I would like to analyze the proposed code using atime, and compare it to base::rev to see how much less time/memory it uses.

jangorecki:
@tdhock Timings are already in this branch in NEWS.md file

NEWS.md
```diff
@@ -293,6 +293,24 @@
...
 41. `tables()` is faster by default by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). `mb=` now accepts any function which accepts a `data.table` and returns a higher and better estimate of its size in bytes, albeit more slowly; e.g. `mb = utils::object.size`.
 
+42. `base::rev` gains a fast method `frev(x, copy)` for atomic vectors/list, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Thanks to Benjamin Schwendinger for suggesting and implementing.
+
+    ```R
+    x = sample(2e8)
+    microbenchmark::microbenchmark(
+      base = rev(x),
+      frev_copy = frev(x, copy=TRUE),
+      frev_inPlace = frev(x, copy=FALSE),
```
jangorecki, line 303: This inPlace also probably need an update

R/wrappers.R
```diff
@@ -16,3 +16,5 @@ isRealReallyInt = function(x) .Call(CisRealReallyIntR, x)
 isReallyReal = function(x) .Call(CisReallyReal, x)
 
 coerceAs = function(x, as, copy=TRUE) .Call(CcoerceAs, x, as, copy)
+
+frev = function(x, copy=TRUE) .Call(Cfrev, x, copy)
```
jangorecki, line 20: in-place should be returned invisibly

ben-schwen:
> Can you please give a specific example (with system.time/etc) where frev should be faster than base::rev? I would like to analyze the proposed code using atime, and compare it to base::rev to see how much less time/memory it uses.

It should always use less memory than `base` since we do not allocate memory for the reverse indices, hence, the allocated memory saving should be exactly length(x)*4 bytes.

Regarding speed, it should always be faster than `base` since it is basically a more specialized version of `subsetVectorRaw` which used to be faster than base.

jangorecki (approved):

MichaelChirico:
Be sure to set 1-15-99 as the target if you want to merge before 1.15.0 is finished esp. w the new delay

src/utils.c
```diff
@@ -424,3 +424,102 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
...
+    } break;
+    case VECSXP: {
+      const SEXP *xd = SEXPPTR_RO(x);
+      #pragma omp parallel for num_threads(getDTthreads(n, true))
+      for (uint64_t i=0; i<n/2; ++i) {
+        const int k = n-1-i;
+        const SEXP tmp = xd[i];
+        SET_VECTOR_ELT(x, i, xd[k]);
+        SET_VECTOR_ELT(x, k, tmp);
+      }
```
HughParsonage, line 494: Are `SET_STRING_ELT` and `SET_VECTOR_ELT` thread-safe? I'd be more comfortable if these cases were not parallelized.

src/utils.c
```diff
@@ -424,3 +424,102 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
...
+    nprotect++;
+    copy = false;
+  }
+  if (n==0) {
+    UNPROTECT(nprotect);
+    return x;
+  }
+  switch (TYPEOF(x)) {
+    case LGLSXP: case INTSXP: {
+      int *restrict xd = INTEGER(x);
```
HughParsonage, line 448: I believe `INTEGER` is not permitted on type `LGLSXP`, even if the underlying type is `int`

src/utils.c
```diff
@@ -424,3 +424,102 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
...
 
+SEXP frev(SEXP x, SEXP copyArg) {
+  if (INHERITS(x, char_dataframe) || INHERITS(x, char_datatable))
+    error(_("'x' should not be data.frame or data.table."));
+  if (!isNull(getAttrib(x, R_DimSymbol)))
+    error(_("'x' should not be matrix or array"));
+  if (!IS_TRUE_OR_FALSE(copyArg))
+    error(_("%s must be TRUE or FALSE."), "copy");
+  bool copy = LOGICAL(copyArg)[0];
+  int n = LENGTH(x);
```
HughParsonage, line 435: I think we should use `R_xlen_t n = xlength(x)` unless there's a good reason not to, especially as we use `uint64_t` for the loop counter. That's a data.table wide thing though, so not a necessary change.

inst/tests/tests.Rraw
```diff
@@ -18260,3 +18260,51 @@ test(2243.54, dt[, .I[j], x]$V1, c(1L, 3L), output="GForce TRUE")
...
+test(2244.09, frev(list(1, "a"),    copy=TRUE), list("a", 1))
+test(2244.11, frev(c(0L, NA),       copy=FALSE), c(NA, 0L))
+test(2244.12, frev(1:3,             copy=FALSE), 3:1)
+test(2244.13, frev(d,               copy=FALSE), c(-Inf, Inf, NaN, NA))
+test(2244.14, frev(c(NA, 1, 0+2i),  copy=FALSE), c(0+2i, 1, NA))
+test(2244.15, frev(as.raw(0:1),     copy=FALSE), as.raw(1:0))
+test(2244.16, frev(NULL,            copy=FALSE), NULL)
+test(2244.17, frev(character(5),    copy=FALSE), character(5))
+test(2244.18, frev(integer(0),      copy=FALSE), integer(0))
+test(2244.19, frev(list(1, "a"),    copy=FALSE), list("a", 1))
```
HughParsonage, line 18283: Recommend adding a test for ALTREP vectors (e.g. `1:1e6`) and long vectors (e.g. 1:1e10).

HughParsonage (changes requested):

src/utils.c
jangorecki: Correct, this must not be parallel

ben-schwen: I thought this is the special case of permutation which @mattdowle mentioned in one of his presentations (I think it was useR2018) which is also thread safe?

jangorecki: Hm, that make sense, but then you have to look at the source of setkey, to ensure it is made similarly as there, and comment well.

jangorecki: I can imagine Matt used something else than SET_STRING_ELT which likely must not be called from multiple threads.

inst/tests/tests.Rraw
jangorecki: Tests needs to be light, therefore I would comment out those tests after running and confirming they work as expected.

src/utils.c
ben-schwen: We also have this in `subset.c` and `gsumm.c` so might want to remove it there too!

jangorecki: Then, I don't think it is actually necessary. It should have been bite us before.

inst/tests/tests.Rraw
ben-schwen: For making `1:1e10` work we would need to support ALTREP. Supporting ALTREP for `copy=FALSE` seems rather straightforward but creating new ALTREP objects seems to be a bit clunky.

```r
  if (ALTREP(x) && R_altrep_data2(x) == R_NilValue) {
    SEXP info = R_altrep_data1(x);
    R_xlen_t LENGTH = (R_xlen_t)REAL0(info)[0];
    int FIRST = (int)REAL0(info)[1];
    int INCR = (int)REAL0(info)[2];
    REAL0(info)[1] = FIRST+(LENGTH-1)*INCR;
    REAL0(info)[2] = INCR*(-1);
    R_set_altrep_data1(x, info);
    return x;
  }
```

HughParsonage: Altrep is easiest done at the R level (see below), but if there's currently no interface (i.e. not `is_altrep(x)` in data.table I withdraw my recommendation to include it in the test suite here.

```r
frev <- function(x) {
  if (is_altrep(x)) {
    return(last(x):first(x))
  }
....
}
```

HughParsonage (approved):

inst/tests/tests.Rraw
ben-schwen: @HughParsonage Thats actually such a big brain move. Ty for that!

jangorecki: Till now we always unpack altrep in DT.

R/utils.R
```diff
@@ -167,6 +167,10 @@ rss = function() {  #5515 #5517
...
 
 frev = function(x, copy=TRUE) {
-    ans = if (isAltrep(x)) last(x):first(x) else .Call(Cfrev, x, copy)
-    if (isTRUE(copy)) ans else invisible(ans)
+    if (!isTRUEorFALSE(copy))
+      stopf("'copy' must be TRUE or FALSE.")
+    altrep = isAltrep(x)
+    if (altrep && !copy) warning("Given vector is altrep and will not be modified in-place.")
```
jangorecki, line 173: We cannot accept such unpredictable behavior. It may happen that a workflow (that depends on update in place) will work fine for a month, and then suddenly getting another type (altrep) of input it will fail.

inst/tests/tests.Rraw
jangorecki: Till now we always unpacked altrep in DT.

ben-schwen:
@tdhock When timing this be careful to expand altrep, since we are now blazingly fast for altrep. And can also do shenanigans like `frev(seq.int(1e12))` :D

R/utils.R
ben-schwen: Good call. We also do not know how `first/last` might be dispatched, hence, relying on them seems error-prone.

MichaelChirico: relevant broader issue #4697. I would approach the broader issue separately from this PR. we might plan out a considered approach to adding ALTREP that minimizes breakage risk.

jangorecki:
I would advise to avoid force push if others are already looking at the branch. I think it also wipe out some of the in-code comments. merge is easiest alternative.

ben-schwen:
Sorry for the force push, I just wanted to reset the last changes (where ALTREP was added).

Agreed PR is ready.

tdhock:
atime benchmark looks very good for this PR, I approve.
image
```r
pkg.edit.DT <- function(old.Package, new.Package, sha, new.pkg.path){
  pkg_find_replace <- function(glob, FIND, REPLACE){
  }
  Package_regex <- gsub(".", "_?", old.Package, fixed=TRUE)
  Package_ <- gsub(".", "_", old.Package, fixed=TRUE)
  pkg_find_replace(
    "DESCRIPTION", 
  pkg_find_replace(
    file.path("src","Makevars.*in"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}
(version.expr.list <- atime::atime_versions_exprs(
  pkg.path="~/R/data.table",
  pkg.edit.fun=pkg.edit.DT,
  expr={
    ## x = sample(2e8)
    ## microbenchmark::microbenchmark(
    ##   base = rev(x),
    ##   frev_copy = frev(x, copy=TRUE),
    ##   frev_inplace = frev(x, copy=FALSE),
    ##   times = 10L,
    ##   unit = "s"
    ## )
    data.table::frev(x, copy=COPY)
  },
  PR5907="48ded0b8f2c4df8c8466b698e19cffe8d77c778e"))
  
atime.grid.args <- c(
  list(list(COPY=c(TRUE,FALSE))),
  version.expr.list)
(expr.list <- do.call(
  atime::atime_grid,
  atime.grid.args))
expr.list$base <- quote(rev(x))

library(data.table)
atime.result <- atime::atime(
  N=as.integer(10^seq(1, 8, by=0.5)),
  setup={
    set.seed(1)
    x = sample(N)
  },
  seconds.limit=1,
  expr.list=expr.list)
plot(atime.result)+
  ggplot2::scale_x_log10(limits=c(NA, 1e9), breaks=10^seq(1, 10))

```

.github/workflows/R-CMD-check.yaml
```diff
@@ -3,12 +3,12 @@
 on:
   push:
     branches:
-      - main
       - master
+      - 1-15-99
```
tdhock, line 7: please remove

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: data.table
-Version: 1.14.99
+Version: 1.15.99
```
tdhock, line 2: please remove

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
...
 Title: Extension of `data.frame`
-Depends: R (>= 3.1.0)
+Depends: R (>= 3.2.0)
```
tdhock, line 4: isn't this unrelated? please remove

tdhock:
I'm confused, why are there so many changes related to version 1.14.99 -> 1.15.99 in this PR?

MichaelChirico:
> I'm confused, why are there so many changes related to version 1.14.99 -> 1.15.99 in this PR?

This is a side-effect of targeting the `1-15-99` branch while the CRAN release was still pending.

R/wrappers.R
```diff
@@ -16,3 +16,5 @@ isRealReallyInt = function(x) .Call(CisRealReallyIntR, x)
 isReallyReal = function(x) .Call(CisReallyReal, x)
 
 coerceAs = function(x, as, copy=TRUE) .Call(CcoerceAs, x, as, copy)
+
+frev = function(x, copy=TRUE) if (isTRUE(copy)) .Call(Cfrev, x, copy) else invisible(.Call(Cfrev, x, copy))
```
MichaelChirico, line 20: WDYT about a new function in utils `visible = identity` then

`frev = function(x, copy=TRUE) (if (isTRUE(copy)) visible else invisible)(.Call(Cfrev,x, copy))`

Or we can just use `identity()`? Current code is a bit hard to read since my instinct is to scan letter-by-letter for what's the difference between the `if`/`else` branches.

MichaelChirico:
> > I'm confused, why are there so many changes related to version 1.14.99 -> 1.15.99 in this PR?
> 
> This is a side-effect of targeting the `1-15-99` branch while the CRAN release was still pending.

@tdhock this is now resolved. Please flag if there are any other PRs with this tedious work to do.

R/wrappers.R
jangorecki: using `(expr eval to fun)(fun args)` is less frequently used to may be less readable at first

ben-schwen: done

NEWS.md
```diff
@@ -16,6 +16,24 @@
 
 2. `cedta()` now returns `FALSE` if `.datatable.aware = FALSE` is set in the calling environment, [#5654](https://github.com/Rdatatable/data.table/issues/5654).
 
+3. `base::rev` gains a fast method `frev(x, copy)` for atomic vectors/list, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Thanks to Benjamin Schwendinger for suggesting and implementing.
```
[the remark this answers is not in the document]
ben-schwen: done

inst/tests/tests.Rraw
```diff
@@ -18330,3 +18330,51 @@ if (test_bit64) {
...
+test(2249.12, frev(as.raw(0:1),     copy=FALSE), as.raw(1:0))
+test(2249.13, frev(NULL,            copy=FALSE), NULL)
+test(2249.14, frev(character(5),    copy=FALSE), character(5))
+test(2249.15, frev(integer(0),      copy=FALSE), integer(0))
+test(2249.16, frev(list(1, "a"),    copy=FALSE), list("a", 1))
+test(2249.17, frev(1:1e2,           copy=TRUE),  rev(1:1e2))
+# copy arguments
+x = 1:3
+test(2249.21, {frev(x, copy=TRUE);  x}, 1:3)
+test(2249.22, {frev(x, copy=FALSE); x}, 3:1)
```
MichaelChirico, line 18357: This is OK, but shouldn't we also have a test showing `frev()` output and `x` have the same address?

MichaelChirico:
One problem I see with this new export: it's the first function from data.table with a `copy=` argument. This violates our usual pattern of by-reference functionality being restricted to `set*` functions.

The pattern chosen for `nafill()` was different: `nafill()` always copies, and by-reference behavior was exposed _via_ `setnafill()`. I think by-reference behavior causes enough headaches for users that the paradigm "only in `set*` functions" is useful, especially for code _reviewers_ who may be less familiar with data.table.

Should we do the same here? `frev()` always copies; `setrev()` is by reference. It's also a bit out-of-pattern since `set*` functions also accept data.tables (`frev()` doesn't, as of now).

man/frev.Rd
```diff
@@ -0,0 +1,32 @@
+\name{frev}
+\alias{frev}
+\alias{rev}
+\title{Fast reverse}
+\description{
+  Similar to \code{base::rev} but \emph{much faster}.
```
MichaelChirico, line 6:
```suggestion
  Similar to \code{\link[base]{rev}} but \emph{much faster}.
```

inst/tests/tests.Rraw
MichaelChirico: Should we also add a test like `frev(x[1:2], copy=FALSE); x` to make sure the original `x` is not altered when `frev()` is run on a subset? This is more like a test of base R behavior, but would be good to be cautious.

src/utils.c
```diff
@@ -424,3 +424,100 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
   return ScalarLogical(false);
 }
 
+SEXP frev(SEXP x, SEXP copyArg) {
+  if (INHERITS(x, char_dataframe) || INHERITS(x, char_datatable))
```
MichaelChirico, line 428: In proper usage, `inherits('data.table')` should imply `inherits('data.frame')`, so I would drop the latter test. Users violating the package's own S3 setup cannot expect normal behavior.

src/utils.c
```diff
@@ -424,3 +424,100 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
...
+    error(_("'x' should not be matrix or array"));
+  if (!IS_TRUE_OR_FALSE(copyArg))
+    error(_("%s must be TRUE or FALSE."), "copy");
+  bool copy = LOGICAL(copyArg)[0];
+  R_xlen_t n = xlength(x);
+  int nprotect = 0;
+  if (copy) {
+    x = PROTECT(duplicate(x));
+    nprotect++;
+    copy = false;
```
MichaelChirico, line 440: obsolete? `copy` is not referenced after this

src/utils.c
```diff
@@ -424,3 +424,100 @@ SEXP startsWithAny(const SEXP x, const SEXP y, SEXP start) {
...
+  if (!isNull(getAttrib(x, R_DimSymbol)))
+    error(_("'x' should not be matrix or array"));
+  if (!IS_TRUE_OR_FALSE(copyArg))
+    error(_("%s must be TRUE or FALSE."), "copy");
+  bool copy = LOGICAL(copyArg)[0];
+  R_xlen_t n = xlength(x);
+  int nprotect = 0;
+  if (copy) {
+    x = PROTECT(duplicate(x));
+    nprotect++;
```
MichaelChirico, line 439: IINM there's only one `PROTECT()` here; the `nprotect` seems like overkill. `UNPROTECT(1)` should do fine.

NEWS.md
```diff
@@ -30,6 +30,24 @@
 
 4. `dcast(fill=NULL)` only computes default fill value if necessary, which eliminates some previous warnings (for example, when fun.aggregate=min or max, warning was NAs introduced by coercion to integer range) which were potentially confusing, [#5512](https://github.com/Rdatatable/data.table/issues/5512), [#5390](https://github.com/Rdatatable/data.table/issues/5390). Thanks to Toby Dylan Hocking for the fix.
 
+3. New `frev(x, copy=TRUE)` as a faster analogue to `base::rev()` for atomic vectors/lists, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Thanks to Benjamin Schwendinger for suggesting and implementing.
+
+    ```R
+    x = sample(2e8)
+    microbenchmark::microbenchmark(
```
MichaelChirico, line 37: how many threads went into this benchmark? It's probably most fair to cite a single-threaded version. Then we can mention it will only get faster with more threads.

inst/tests/tests.Rraw
```diff
@@ -18330,3 +18330,51 @@ if (test_bit64) {
...
+test(2249.32, frev(as.IDate(1:10), copy=TRUE), as.IDate(10:1))
+test(2249.33, frev(as.IDate(1:10), copy=TRUE), as.IDate(10:1))
+# names
+x = c(a=1L, b=2L, c=3L)
+test(2249.41, frev(x, copy=TRUE), rev(x))
+test(2249.42, frev(x, copy=FALSE), x)
+# attributes
+x = structure(1:10, class = c("IDate", "Date"), att = 1L)
+test(2249.51, attr(frev(x, copy=TRUE), "att"), 1L)
+test(2249.52, attr(frev(x, copy=FALSE), "att"), 1L)
```
MichaelChirico, line 18370: Note that this will break e.g. the `gregexpr()` output, where the attributes are created to match the actual object order. Nothing we can do here besides document the potential issue, I think.

src/utils.c
ben-schwen: We have a similar check in assign.c
https://github.com/Rdatatable/data.table/blob/54f904831cafc37ba841958fbdbc614b8baac4c8/src/assign.c#L307-L309

inst/tests/tests.Rraw
ben-schwen: True, but this problem also appears with `base::rev`?

MichaelChirico: OK, let's just document that we copy base::rev behavior for attributes?

src/utils.c
ben-schwen: Maybe even `UNPROTECT(copy)` to cater for the copy/non-copy cases?

ben-schwen:
> One problem I see with this new export: it's the first function from data.table with a `copy=` argument. This violates our usual pattern of by-reference functionality being restricted to `set*` functions.
> 
AFAIR I copied the argument name from `coerceAs = function(x, as, copy=TRUE)`. The only other case that comes to my mind immediately is `droplevels.data.table = function(x, except = NULL, exclude, in.place = FALSE, ...)` which has an `in.place` argument.

> The pattern chosen for `nafill()` was different: `nafill()` always copies, and by-reference behavior was exposed _via_ `setnafill()`. I think by-reference behavior causes enough headaches for users that the paradigm "only in `set*` functions" is useful, especially for code _reviewers_ who may be less familiar with data.table.
>
So what about `frev` and `setrev` similar to `fcoalesce` and `setcoalesce`?

> Should we do the same here? `frev()` always copies; `setrev()` is by reference. It's also a bit out-of-pattern since `set*` functions also accept data.tables (`frev()` doesn't, as of now).

That's exactly the reason `copy=TRUE` is the default. Users being smart enough to use an argument with a default parameter should be smart enough to understand the argument.

src/utils.c
MichaelChirico: I think I'd still prefer `UNPROTECT(1)`, memory safety is hard enough to keep track of mentally that it's nice to have a literal value we can use here -- having to mentally map `copy == FALSE` --> `UNPROTECT(0)`, `copy = TRUE` --> `UNPROTECT(1)` is a tax on the reader IMO

MichaelChirico:
> That's exactly the reason `copy=TRUE` is the default. Users being smart enough to use an argument with a default parameter should be smart enough to understand the argument.

Possibly, but it's easy to think "faster --> unequivocally better" and use it somewhat carelessly / forgetting the full implications of the choice. And that's just on the part of the code author. We also need to consider code reviewers and future readers/maintainers.

It's with that (and more) in mind that we really want to limit the API surface of in-place operations of data.table IMO. Anxiety about getting bitten by mutability issues is quite a common reason to avoid touching {data.table}; this problem gets worse if readers/writers of code using {data.table} need to be wary of reference semantics in "new"/"unexpected" places. By sticking to a simple rule: "By-reference updates can only happen when you see `:=` or a `set*` function", we can ease this burden as much as possible.

I missed `droplevels(in.place=TRUE)` -- wish I'd given the same feedback there :)

> So what about frev and setrev similar to fcoalesce and setcoalesce?

Good catch, I'd forgotten about `setcoalesce()`. I believe @jangorecki originally requested that. To clarify, you means not exposing a `copy=FALSE` version to users, and only using that ourselves internally, but both are just thin wrappers to the C routine with a different flag passed? If so then yes, that looks like a good direction to me.

https://github.com/Rdatatable/data.table/blob/54f904831cafc37ba841958fbdbc614b8baac4c8/R/wrappers.R#L5-L6

src/utils.c
jangorecki: That's an internal error, so situation that should not happen

jangorecki:
Let's add setdroplevels then. We could have dot prefixed non exported function that had copy arg and then exported ones that goes via this .droplevels

src/utils.c
ben-schwen: I guess I'm on the wrong track here, but are you suggesting:
```r
  if (copy) {
    x = PROTECT(duplicate(x));
    UNPROTECT(1);
  }
```
and remove the other UNPROTECT calls? I thought that's invalid because we modify x afterwards?

NEWS.md
```diff
@@ -34,6 +34,24 @@
...
+      base = rev(x),
+      frev_copy = frev(x, copy=TRUE),
+      frev_inplace = frev(x, copy=FALSE),
+      times = 10L,
+      unit = "s"
+    )
+    #       expr      min    lq  mean median   uq   max neval cld
+    #       base    1.240 1.263 1.301 1.291 1.329 1.437    10 a  
+    #  frev_copy    0.533 0.538 0.590 0.576 0.639 0.667    10  b 
+    #  frev_inplace 0.099 0.101 0.130 0.102 0.107 0.260    10   c
```
MichaelChirico, line 52:
```suggestion
      frev = frev(x)
      times = 10L,
      unit = "s"
    )
    #  expr      min    lq  mean median   uq   max neval cld
    #  base    1.240 1.263 1.301 1.291 1.329 1.437    10  a
    #  frev    0.533 0.538 0.590 0.576 0.639 0.667    10   b
```

MichaelChirico: OTOH maybe a benchmark is overkill with only 2 options and where we are only 2x faster? We can just say "Twice as fast as `base::rev()` on large inputs, and faster with more threads"

inst/tests/tests.Rraw
MichaelChirico: Hmm but `rev()` drops attributes?

```
names(attributes(rev(structure(1:10, class = c("IDate", "Date"), att = 1L))))
# [1] "class"

names(attributes(rev(regexpr("[a-m]", letters))))
# NULL
```

src/utils.c
MichaelChirico: I see what you mean now, sorry, `UNPROTECT(copy)` works for me, otherwise the current `nprotect` is also OK.

jangorecki:
I think calling by reference operation on a vector alone is more likely to cause troubles, that possibly could be mentioned with some example, like changing R constant value

MichaelChirico:
> I think calling by reference operation on a vector alone is more likely to cause troubles, that possibly could be mentioned with some example, like changing R constant value

Note that in the current PR, there's no user-facing way to call `setrev()`, I think that's for the best.

inst/tests/tests.Rraw
ben-schwen: Good spot, I just checked with `names` and `class` but apparently those are special attributes.

NEWS.md
ben-schwen: true, removed benchmark and added wording about being twice as fast (which holds for elapsed time being big enough)

inst/tests/tests.Rraw
ben-schwen: I added details in the docs about attributes being retained.

MichaelChirico: I think we should rather match `base::rev()` behavior and strip (non-"special") attributes. As mentioned this can have subtly broken implications -- user wanting to keep attributes around should be responsible for how to do so.

ben-schwen: Not sure I'm on the same side here. Retaining attributes with all operations on a data.table is an often mentioned feature, users would like.

I know that it's not user-facing now, but what would be the behavior for `setrev` for stripping attributes? Strip them as a side effect or keep them?

ben-schwen: We drop attributes now on `frev`. For `setrev` we still keep them but can change if wanted.

For the special case of zero length vectors, attributes are still copied (`rev` behaves here the same).

NEWS.md
```diff
@@ -18,6 +18,8 @@
 
 6. `between()` gains the argument `ignore_tzone=FALSE`. Normally, a difference in time zone between `lower` and `upper` will produce an error, and a difference in time zone between `x` and either of the others will produce a message. Setting `ignore_tzone=TRUE` bypasses the checks, allowing both comparisons to proceed without error or message about time zones.
 
+7. New `frev(x, copy=TRUE)` as a faster analogue to `base::rev()` for atomic vectors/lists, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Twice as fast as `base::rev()` on large inputs, and faster with more threads. Thanks to Benjamin Schwendinger for suggesting and implementing.
```
```suggestion
7. New `frev(x)` as a faster analogue to `base::rev()` for atomic vectors/lists, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Twice as fast as `base::rev()` on large inputs, and faster with more threads. Thanks to Benjamin Schwendinger for suggesting and implementing.
```

inst/tests/tests.Rraw
```diff
@@ -21192,3 +21193,58 @@ test(2319.1, !is.null(attr(dt_get, ".internal.selfref")))
 dt_get0 = data.frame(a = 1:3, b = letters[1:3])
 setDT(get0("dt_get0"))
 test(2319.2, !is.null(attr(dt_get0, ".internal.selfref")))
+
+# 5885 implement frev
+d = c(NA, NaN, Inf, -Inf)
```
MichaelChirico: maybe these should all be structured as parity tests? `test(., frev(c(FALSE, NA)), rev(c(FALSE, NA)))` etc. WDYT?

inst/tests/tests.Rraw
```diff
@@ -21192,3 +21193,58 @@ test(2319.1, !is.null(attr(dt_get, ".internal.selfref")))
...
+test(2320.00, frev(c(FALSE, NA)),     c(NA, FALSE))
+test(2320.01, frev(c(0L, NA)),        c(NA, 0L))
+test(2320.02, frev(d),                c(-Inf, Inf, NaN, NA))
+test(2320.03, frev(c(NA, 1, 0+2i)),   c(0+2i, 1, NA))
+test(2320.04, frev(as.raw(0:1)),      as.raw(1:0))
+test(2320.05, frev(NULL),             NULL)
+test(2320.06, frev(character(5)),     character(5))
+test(2320.07, frev(integer(0)),       integer(0))
+test(2320.08, frev(list(1, "a")),     list("a", 1))
+test(2320.09, setrev(c(0L, NA)),      c(NA, 0L))
```
MichaelChirico: Should these be structured more to get after the side-effects, like

```r
test(2320.09, {x = c(0L, NA); setrev(x); x},      c(NA, 0L))
```

More generally, these days I shy away from tests of private API by default (e.g. https://softwareengineering.stackexchange.com/q/380287/221381), should we just drop the `setrev()` tests?

man/frev.Rd
```diff
@@ -0,0 +1,32 @@
+\name{frev}
+\alias{frev}
+\alias{rev}
+\title{Fast reverse}
+\description{
+  Similar to \code{\link[base]{rev}} but \emph{much faster}.
```
MichaelChirico: quibble: is 2x _much faster_? it's not an order-of-magnitude shift... maybe we soften "often much faster" or just "faster"?

man/frev.Rd
```diff
@@ -0,0 +1,32 @@
...
+
+\usage{
+frev(x)
+}
+\arguments{
+  \item{x}{ An atomic \code{vector} or \code{list}. }
+}
+
+\details{
+  \code{frev} does not retain attributes (similar to \code{\link[base]{rev}}).
```
MichaelChirico:
```suggestion
  \code{frev} does not retain attributes other than names and class (similar to \code{\link[base]{rev}}).
```

inst/tests/tests.Rraw
```diff
@@ -21192,3 +21193,58 @@ test(2319.1, !is.null(attr(dt_get, ".internal.selfref")))
...
+test(2320.51, attr(frev(x), "att"), attr(rev(x), "att"))
+test(2320.52, class(frev(x)), class(rev(x)))
+test(2320.53, attr(setrev(x), "att"), 1L)
+test(2320.54, class(setrev(x)), c("IDate", "Date"))
+x = structure(integer(0), att = 1L)
+test(2320.55, attr(frev(x), "att"), attr(rev(x), "att"))
+# errors
+test(2320.61, frev(data.table()), error="should not be data.frame or data.table")
+test(2320.62, frev(expression(1)), error="is not supported by frev")
+test(2320.63, frev(matrix(1)), error="should not be matrix or array")
```
MichaelChirico: Maybe allow this? It enables this idiom:

```r
M[] = frev(M)
```

https://github.com/search?q=lang%3AR+%2F%5Cw%5C%5B%5C%5D%5Cs*%28%3D%7C%3C-%29%5Cs*rev%5C%28%2F&type=code

`rev()` itself works, just dropping the dim attribute.

jangorecki:
Why not `frev` and `setfrev`? feels to be simpler naming pattern just by prefixing with `set`, and not removing extra `f`.

R/wrappers.R
```diff
@@ -21,3 +21,6 @@ fitsInInt32 = function(x) .Call(CfitsInInt32R, x)
 fitsInInt64 = function(x) .Call(CfitsInInt64R, x)
 
 coerceAs = function(x, as, copy=TRUE) .Call(CcoerceAs, x, as, copy)
+
+frev   = function(x) .Call(Cfrev, x, TRUE)
+setrev = function(x) invisible(.Call(Cfrev, x, FALSE))
```
MichaelChirico: There are only a handful of call sites in the code base -- let's use `setrev()` in this PR.

```
R/bmerge.R:      if (x_merge_type=="integer64") { w=i; wc=icol; wclass=i_merge_type; } else { w=x; wc=xcol; wclass=x_merge_type; nm=rev(nm) }  # w is which to coerce
R/cedta.R:  for (ii in length(calls):1) { # nolint: seq_linter. rev(seq_len(length(calls)))? See https://bugs.r-project.org/show_bug.cgi?id=18406.
R/as.data.table.R:  val = rev(dimnames(provideDimnames(x)))
R/as.data.table.R:    setattr(val, 'names', paste0("V", rev(seq_along(val))))
R/as.data.table.R:  setcolorder(ans, c(rev(head(names(ans), -1L)), "N"))
R/as.data.table.R:  val = rev(val)
R/as.data.table.R:    setattr(val, 'names', paste0("V", rev(seq_along(val))))
R/as.data.table.R:    stopf("Argument 'value.name' should not overlap with column names in result: %s", brackify(rev(names(val))))
R/as.data.table.R:  dims = rev(head(names(ans), -1L))
R/utils.R:  length(x) - match(TRUE, rev(x)) + 1L
```

We should use it whenever the in-place modifee is also totally internal, e.g. here:

https://github.com/Rdatatable/data.table/blob/037bc89f80ebb8f2703e54ebc864ce465c8f7a44/R/bmerge.R#L107-L108.

We could also consider structuring this like so

```r
frev   = function(x) {o = setrev(x, copy=TRUE); o}
setrev = function(x, copy) invisible(.Call(Cfrev, x, copy))
```

That way we can add a linter to disable using `rev()` internally, so that we always use `setrev()` and just think through in each case whether to use `copy=TRUE/FALSE`?

jangorecki: Looks OK, but if would ever decide to export "set" version of `frev`, then my preference would be `setfrev`.

inst/tests/tests.Rraw

inst/tests/tests.Rraw
ben-schwen: done, since we might want to export in the future

man/frev.Rd
ben-schwen: done

man/frev.Rd
ben-schwen: done

inst/tests/tests.Rraw
ben-schwen: done and added a test for rotate180 idiom

NEWS.md
```diff
@@ -46,6 +46,8 @@
 
 10. `data.table()` and `as.data.table()` with `keep.rownames=TRUE` now extract row names from named vectors, matching `data.frame()` behavior. Names from the first named vector in the input are used to create the row names column (default name `"rn"` or custom name via `keep.rownames="column_name"`), [#1916](https://github.com/Rdatatable/data.table/issues/1916). Thanks to @richierocks for the feature request and @Mukulyadav2004 for the implementation.
 
+9. New `frev(x)` as a faster analogue to `base::rev()` for atomic vectors/lists, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Twice as fast as `base::rev()` on large inputs, and faster with more threads. Thanks to Benjamin Schwendinger for suggesting and implementing.
```
```suggestion
11. New `frev(x)` as a faster analogue to `base::rev()` for atomic vectors/lists, [#5885](https://github.com/Rdatatable/data.table/issues/5885). Twice as fast as `base::rev()` on large inputs, and faster with more threads. Thanks to Benjamin Schwendinger for suggesting and implementing.
```

R/utils.R
```diff
@@ -226,7 +226,7 @@ fctr = function(x, levels=unique(x), ..., sort=FALSE, rev=FALSE) {
   if (!isTRUEorFALSE(rev))
     stopf("argument 'rev' must be TRUE or FALSE")
   if (sort) levels = sort(levels)
-  if (rev) levels = rev(levels)
+  if (rev) setfrev(levels)
```
MichaelChirico: we can't do that safely, right?

```r
levels = rev(letters)
fctr(letters, levels)
```

will have a side-effect

I suppose we could use `setfrev()` if we've already done overwrite under `sort`, but doubt it's worth the complication

MichaelChirico:
```suggestion
  if (rev) levels = frev(levels)
```

src/utils.c
```diff
@@ -542,6 +542,111 @@ SEXP perhapsDataTableR(SEXP x) {
...
+  levels = PROTECT(getAttrib(x, R_LevelsSymbol));
+  nprotect += 3;
+  if (copy) {
+    SET_ATTRIB(x, R_NilValue);
+    setAttrib(x, R_NamesSymbol, names);
+    setAttrib(x, R_ClassSymbol, klass);
+    setAttrib(x, R_LevelsSymbol, levels);
+  }
+  if (!isNull(names)) {
+    frev(names, ScalarLogical(FALSE));
```
MichaelChirico, line 646: is it safe to do this by-reference when `copy=TRUE`? I don't see this tested, please add a relevant test

src/utils.c
```diff
@@ -542,6 +542,111 @@ SEXP perhapsDataTableR(SEXP x) {
...
+        const Rbyte tmp = xd[i];
+        xd[i] = xd[k];
+        xd[k] = tmp;
+      }
+    } break;
+  default:
+    error(_("Type '%s' is not supported by frev"), type2char(TYPEOF(x)));
+  }
+  names = PROTECT(getAttrib(x, R_NamesSymbol));
+  klass = PROTECT(getAttrib(x, R_ClassSymbol));
```
MichaelChirico: `klass`+`levels` are only needed under `copy=TRUE` right? do we even need the temporary variable?

MichaelChirico (approved):
LGTM after this last round of review

src/utils.c
ben-schwen: Yes, since duplicate also duplicates the attributes!

Added a test though so we notice when this should change

src/utils.c
ben-schwen: yes. and we need the temp variables since we clear all attributes at copy!

PR #5908, opened 2024-01-14
Add a GHA for linting code

MichaelChirico:
Pared down version of #4278 to only add the GHA, and some whitespace-only code changes.

With this, we can start enforcing rules that we already obey ourselves.

Next, we can add a separate 'lint-new' workflow to enforce rules on _new_ code, i.e. lints already in `master` won't throw, but lints in new PRs will need to be fixed.

Eventually, all lints on master can be fixed and we can remove the 'lint-new' workflow.


jangorecki:
This PR touches many files and is likely to cause conflicts to long queue of PRs we have. We could consider adding just linter for now, and now fixing what it detects, and before 1.16.0, when, hopefully most, of PRs will be merged, then merge rest of this PR.

MichaelChirico:
> This PR touches many files and is likely to cause conflicts to long queue of PRs we have. We could consider adding just linter for now, and now fixing what it detects, and before 1.16.0, when, hopefully most, of PRs will be merged, then merge rest of this PR.

My thinking with only applying whitespace changes (besides vignettes, which don't have much in the queue anyway, and the changes are ~mostly whitespace-only too) is it's not likely to cause any troubling merge conflicts. It also lets us turn on more repo-wide linters up-front.

jangorecki:
Is location of .lintr.R configurable? we could put it into ".ci" dir for example

MichaelChirico:
> Is location of .lintr.R configurable? we could put it into ".ci" dir for example

Yes, per https://lintr.r-lib.org/reference/read_settings.html I think `env` is the better way for CI vs. `options()`.

.github/workflows/lint.yaml
```diff
@@ -30,3 +30,4 @@ jobs:
         shell: Rscript {0}
         env:
           LINTR_ERROR_ON_LINT: true
+          R_LINTR_LINTER_FILE: .ci/.lintr
```
jangorecki, line 33: missing file extension, no?

MichaelChirico: No -- {lintr} looks for `${R_LINTR_LINTER_FILE}.R` then `${R_LINTR_LINTER_FILE}`:

https://github.com/r-lib/lintr/blob/bad1632f23b77f41cec13fb6f1c883302f64f506/R/settings_utils.R#L90-L91

MichaelChirico:
> Is location of .lintr.R configurable? we could put it into ".ci" dir for example

Looks like some sort of {lintr} bug here I can't quite figure out, so this did complicate our workflow a bit (now we need `:::` access to {lintr}): https://github.com/r-lib/lintr/issues/2512

MichaelChirico:
cc @jangorecki again for review

MichaelChirico:
@tdhock it would be great to get this merged so we can start enforcing more incremental code quality going forward, PTAL. thanks!

MichaelChirico:
cc @Anirban166 for the trailing whitespace changes here / possible need to update #6094

.ci/.lintr.R
```diff
@@ -0,0 +1,103 @@
...
+  # )),
+  # undesirable_function_linter(modify_defaults(
+  #   default_undesirable_functions,
+  #   ifelse = "Use fifelse instead.",
+  #   Sys.setenv = NULL,
+  #   library = NULL,
+  #   options = NULL,
+  #   par = NULL,
+  #   setwd = NULL
+  # )),
```
tdhock, line 21: why is all this code commented?
if you want to keep, please explain why in a TODO comment? (as below which I think is easier to understand)

tdhock (changes requested):

.ci/.lintr.R
MichaelChirico: IIRC this is here to encode the data.table parameterization of default linters to make it easier to turn them on incrementally with the right settings. Added a comment to that effect, thanks!

tdhock: great thx

tdhock (approved):

PR #5911, opened 2024-01-15
Frollmax2 rebase

MichaelChirico:
Hey @jangorecki I had a go at rebasing #5890 to `1-15-99`, PTAL and I'll force-push there if it looks like I succeeded.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -319,7 +319,40 @@
 
 41. `tables()` is faster by default by excluding the size of character strings in R's global cache (which may be shared) and excluding the size of list column items (which also may be shared). `mb=` now accepts any function which accepts a `data.table` and returns a higher and better estimate of its size in bytes, albeit more slowly; e.g. `mb = utils::object.size`.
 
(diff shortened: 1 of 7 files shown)


jangorecki:
Till late Feb/early March I may not have enough time to work on all PRs so I am considering it may be better to revert frollmax1, as it is possibly buggy, and come back to it later. Haven't look at the content of this PR yet but some commits seem to be missing authored/committed info.

MichaelChirico:
> missing authored/committed info.

Yea, I saw that. Weird. I only did `git rebase 1-15-99` while checked out from this branch, then pushed, so not sure what happened. Alas.

> better to revert frollmax1

OK, happy to that then we just roll forward later.

MichaelChirico:
Hmm GitHub UI looks to make `revert` complicated, I guess because it's merged to `1-15-99`. I will try reverting again once `1-15-99` is merged to `master`. For now the code is not touching master so it's no hurry anyway.

jangorecki:
You can event force push to revert it invisibly

Issue #5913, opened 2024-01-15
Use classed conditions as often as possible


MichaelChirico:
#5912 is the latest example, but generally speaking it is a good idea to create some classed conditions.

The first step is to add support to `stopf()` and friends for making the resulting condition with extended class.

Thereafter we can work towards a taxonomy of condition classes.

MichaelChirico:
We'll also benefit from putting `...` into the signature of `test()`, with entries referring to condition classes, then we can do `test(n,x,y, custom_warning = ...)`. Implementation gets a bit messier from there but should be doable.

jangorecki:
I would also wait with this one till 1.17.0 at least, when first round of pending PRs will get resolved

MichaelChirico:
WDYT about merging #5914 now?

We don't need to use classed conditions everywhere yet, but could at least start for some common errors. Changing `stopf()` (etc) lines shouldn't cause too many conflicts with other PRs.

Even if not, with #5914 we can start writing custom classes for new functionality.

Most important is to pick a framework for errors, i.e. should all errors be classed like `c("dt_${SPECIFIC_CLASS}", "dt_warn", "warning")` (similarly for `dt_error`, `dt_message`, `dt_startup_message`)? Something else? We'll want to pick that first to avoid needing to change it when we scale up eventually.

Mukulyadav2004:
Hi  @MichaelChirico @jangorecki ,
I've been reviewing this issue and the related discussions. With the merge of #5914, it successfully implemented the infrastructure with `raise_condition()` and updated `stopf()` , `warningf()` , `message()` to support custom classes. So what I'm understanding from discussion is that the current need is to identify and implement specific condition classes throughout the codebase which enables users to programmatically handle specific warnings - like the melt coercion case(#5912). 
So whether this issue is open for contributions at this time, or if it's being reserved for a future release. If it is open to being worked on, I would be interested in helping.

MichaelChirico:
Thanks for the interest Mukul!

I think the main difficulty at this point is one of design -- once we pick a custom class for a condition, it becomes more difficult to change it later (since there may be code like `tryCatch(..., custom_class=)` downstream that would be broken if we do so.

So ideally we proceed carefully & with ample consideration about how to structure the classes. So far, that's meant we only introduce custom classes in the most obvious/pressing/requested places.

I think that incremental approach is still suitable, but if you have further thoughts / any overarching proposal, I'd love to hear it!

[...]

PR #5914, opened 2024-01-15
class= argument for condition calls

MichaelChirico:
Towards #5913

No NEWS needed here as there's no user-facing change. I expect we'll add some items as we start actually using the custom classes.


MichaelChirico:
Friendly ping here too @tdhock thanks!

tdhock (approved):
lgtm

PR #5916, opened 2024-01-16
gmin parallizle per group

### LINKED ISSUE #5900: locating performance improvement in NEWS
I've noticed a performance improvement between 1.14.10 and the current dev version but having a quick grep through the news nothing jumps out as obvious. Could someone flag the PR that gave the improvement / highlight in the NEWS if I've missed it (apologies if so)?

``` r
library(data.table)

N <- 1e8
DT <- data.table(group = sample(letters, N, replace = TRUE), value = rnorm(N))
```

**1.14.10:**
```r
microbenchmark::microbenchmark(DT[, .(min(value), max(value)), by = group], times = 10)
#> Unit: seconds
#>                                         expr      min     lq     mean   median       uq      max neval
#>  DT[, .(min(value), max(value)), by = group] 1.536686 1.5923 1.635291 1.629424 1.664668 1.733363    10
```

**data.table 1.14.99 IN DEVELOPMENT built 2024-01-09 04:18:05 UTC:**
```r
microbenchmark::microbenchmark(DT[, .(min(value), max(value)), by = group], times = 10, unit = "seconds")
#> Unit: seconds
#>                                         expr       min        lq     mean    median        uq       max neval
#>  DT[, .(min(value), max(value)), by = group] 0.5830119 0.6556398 0.687627 0.6822015 0.7115891 0.8280193    10

```

ben-schwen:
Closes #5900

- [x] implement
- [ ] NEWS

### DIFF
--- a/src/gsumm.c
+++ b/src/gsumm.c
@@ -768,9 +768,9 @@ static SEXP gminmax(SEXP x, SEXP narm, const bool min)
     const SEXP *restrict xd = STRING_PTR(x);
     if (!LOGICAL(narm)[0]) {
       const SEXP init = min ? char_maxString : R_BlankString;  // char_maxString == "\xFF\xFF..." in init.c
+      for (int i=0; i<ngrp; ++i) SET_STRING_ELT(ans, i, init);
       #pragma omp parallel for num_threads(getDTthreads(ngrp, false))
       for (int i=0; i<ngrp; ++i) {
-        SET_STRING_ELT(ans, i, init);
         for (int j=ff[i]-1; j<ff[i]-1+grpsize[i]; ++j){
           if (ansd[i]==NA_STRING) break; // once NA has been obsered group is finished
           const int k = isunsorted ? oo[j]-1 : j;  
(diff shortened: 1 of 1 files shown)


jangorecki:
Benchmark need to compare small groups and big groups, and it needs to test for parallelism just after crossing throttle threshold (for small groups parallel loop will possibly for single threaded). Therefore some C level debugging may be needed.

jangorecki:
@ben-schwen could you please include best and worst case scenarios?

tdhock:
hi ben I ran a benchmark of this PR vs master and 1.14.0, and I see some speedups using 4 or 8 thread on my windows machine, see below

image

however you can also see that this PR using 1 thread is actually a little slower than current master/1.14.0, do you know why that may be? I was expecting it to be the same.
source code here: https://github.com/tdhock/atime/issues/22#issuecomment-1906957225

jangorecki:
There were performance issues with openmp on windows in the past therefore better we use linux for benchmarks. If anyone is after performance seriously I don't think staying on windows is good choice for he/she.

ben-schwen:
> however you can also see that this PR using 1 thread is actually a little slower than current master/1.14.0, do you know why that may be? I was expecting it to be the same. source code here: [tdhock/atime#22 (comment)](https://github.com/tdhock/atime/issues/22#issuecomment-1906957225)

I expected it to be slower than master for a single thread, since we didn't do it per group but all groups concurrently before. This should be faster since we do have to get group indices with `ff` and `oo`.

It should be faster single-threaded for the case where have NAs in the groups and `min(..., na.rm=FALSE)` since we can finish these groups early.

tdhock:
I ran it on linux, here are the results, which are mostly consistent with windows. This PR with 1 thread is still a little slower than master, but there are speedups whenever you increase the number of cores, which is good.
image

Issue #5924, opened 2024-01-30
Please roll the main branch to 1.5.99


eddelbuettel:
Congratulations on getting 1.5.0 shipped to CRAN!  A true milestone.

So now that version is ahead of `master` and eg as a consequence the convenience wrapper `DT` is hiding.  I could re-enable (and another micro-issue is that I didn't immediately find the options ...) or I could just rebuild from source -- but for that source needs to be ahead of release.

```r
> sessionInfo()
R version 4.3.2 (2023-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 23.10

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.11.0 
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.11.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8      
 [8] LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

time zone: America/Chicago
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.15.0

loaded via a namespace (and not attached):
[1] compiler_4.3.2   parallel_4.3.2   colorout_1.3-0.1 fortunes_1.5-4  
> 
```

ben-schwen:
We have already set up the new dev [branch](https://github.com/Rdatatable/data.table/tree/1-15-99).

AFAIK the only thing left before switching master to the new branch is to tag `1.15.0` accordingly.

eddelbuettel:
Ahhhh. Because I had a fork and the real upstream as `parentrepo` I was only following `master` there. Corrected, and installed from 1-15-99.  (That left the branch in 'detached HEAD' which I don't like so I didn't 'remain' in the branch. I trust 'master' will get update soon.)  Thanks for the hint!

TysonStanley:
Will be tagging the release today.

PR #5925, opened 2024-01-30
Release changes 1.15

TysonStanley:
The changes made during the release process to CRAN (minor adjustments and version bump for the tag).

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -353,6 +353,6 @@ SEXP initLastUpdated(SEXP var) {
 
 SEXP dllVersion(void) {
   // .onLoad calls this and checks the same as packageVersion() to ensure no R/C version mismatch, #3056
-  return(ScalarString(mkChar("1.15.99")));
+  return(ScalarString(mkChar("1.15.0")));
 }
 
--- a/src/init.c
+++ b/src/init.c
@@ -353,6 +353,6 @@ SEXP initLastUpdated(SEXP var) {
 
 SEXP dllVersion(void) {
   // .onLoad calls this and checks the same as packageVersion() to ensure no R/C version mismatch, #3056
-  return(ScalarString(mkChar("1.15.0")));
+  return(ScalarString(mkChar("1.15.99")));
 }
 
--- a/src/init.c
+++ b/src/init.c
@@ -353,6 +353,6 @@ SEXP initLastUpdated(SEXP var) {
 
 SEXP dllVersion(void) {
   // .onLoad calls this and checks the same as packageVersion() to ensure no R/C version mismatch, #3056
-  return(ScalarString(mkChar("1.14.99")));
+  return(ScalarString(mkChar("1.15.0")));
 }
 
--- a/.dev/CRAN_Release.cmd
+++ b/.dev/CRAN_Release.cmd
@@ -598,7 +598,11 @@ Rdevel -q -e "packageVersion('xml2')"   # ensure installed
 Rdevel CMD check data.table_1.16.0.tar.gz --as-cran  # use latest Rdevel as it may have extra checks
 bunzip2 inst/tests/*.Rraw.bz2  # decompress *.Rraw again so as not to commit compressed *.Rraw to git
 
+
 # Resubmit to winbuilder (R-release, R-devel and R-oldrelease)
+# Submit to R-Hub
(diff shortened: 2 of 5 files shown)


src/init.c
```diff
@@ -353,6 +353,6 @@ SEXP initLastUpdated(SEXP var) {
 
 SEXP dllVersion(void) {
   // .onLoad calls this and checks the same as packageVersion() to ensure no R/C version mismatch, #3056
-  return(ScalarString(mkChar("1.14.99")));
+  return(ScalarString(mkChar("1.15.0")));
```
tdhock, line 356: I thought we did not want 1.15.0 (or any even number release version) to get on master branch?

tdhock (changes requested):

src/init.c
TysonStanley: For the tag, it must be changed to match the tag itself. This matches the release to CRAN so should not produce any issues. However, not sure if this should be merged to master (rather the 1.15.99).

tdhock: yes I think it is normal to have the version number match the tag name, but I thought we kept that code on a separate branch? (not master) 
correct me if I'm wrong.

TysonStanley: I think that is what we did previously, so yes, let's not include this last commit in the PR. I'll remove those changes from this PR.

TysonStanley:
Added version bump to 1.15.99 for merging to master

ben-schwen:
@TysonStanley could you also update the NEWS file to 1.15.0, stating when it was submitted to CRAN and then retag 1.15.0 (at least that's what Matt always did)

TysonStanley:
Using this branch for the tag. Would recommend merging at least the first two commits here for the master branch when we bring 1.15.99 into master.

MichaelChirico:
> Using this branch for the tag. Would recommend merging at least the first two commits here for the master branch when we bring 1.15.99 into master.

planning to rebase the 1-15-99 branch once this is in master to have the cleanest possible history

TysonStanley:
Sounds good. The tag is created. Do you want me to create a new PR with just the relevant commits for master?

ben-schwen:
I guess we are good with this PR

Closes #5931 #5924

tdhock:
Hi, I don't know if we still care to follow CRAN_Release.cmd, which says the following:
```
# DO NOT commit or push to GitHub. Leave 4 files (.dev/CRAN_Release.cmd, DESCRIPTION, NEWS and init.c) edited and not committed. Include these in a single and final bump commit below.
# DO NOT even use a PR. Because PRs build binaries and we don't want any binary versions of even release numbers available from anywhere other than CRAN.
# Leave milestone open with a 'release checks' issue open. Keep updating status there.
# ** If on EC2, shutdown instance. Otherwise get charged for potentially many days/weeks idle time with no alerts **
# If it's evening, SLEEP.
# It can take a few days for CRAN's checks to run. If any issues arise, backport locally. Resubmit the same even version to CRAN.
# CRAN's first check is automatic and usually received within an hour. WAIT FOR THAT EMAIL.
# When CRAN's email contains "Pretest results OK pending a manual inspection" (or similar), or if not and it is known why not and ok, then bump dev.

###### Bump dev for NON-PATCH RELEASE
# 0. Close milestone to prevent new issues being tagged with it. The final 'release checks' issue can be left open in a closed milestone.
# 1. Check that 'git status' shows 4 files in modified and uncommitted state: DESCRIPTION, NEWS.md, init.c and this .dev/CRAN_Release.cmd
# 2. Bump minor version in DESCRIPTION to next odd number. Note that DESCRIPTION was in edited and uncommitted state so even number never appears in git.
```
The last line is what I'm worried about (if we still care to follow these instructions). The "even number never appears in git" means that we should not push any commits with version 1.15.0 to master branch, and instead master should go from 1.14.99 to 1.15.99 -- does that make sense? Also it says below that the specific commit message we should use:
```r
# 8. Push to master with this consistent commit message: "1.14.8 on CRAN. Bump to 1.14.10"
```
so if you agree, let's close this PR and instead merge #5936 ?

jangorecki:
The other PR looks better but there are good updates to release procedure here in this PR that make sense to include as well, but probably better not in version bump PR anyway

PR #5929, opened 2024-01-31
add cast to xlength

### LINKED ISSUE #5935: data.table 1.15.0 two failing tests on CRAN
CRAN contacted me today with what looks like two failing tests. 

> Dear maintainer,
> 
> Please see the problems shown on <https://cran.r-project.org/web/checks/check_results_data.table.html>.
> Please correct before 2024-02-27 to safely retain your package on CRAN.
>
> Note that this will be the *final* reminder.
>
> The CRAN Team

ben-schwen:
Follow up to #5768
Closes #5935 
Fixes additional M1mac issue in [CRAN check results](https://cran.r-project.org/web/checks/check_results_data.table.html)


MichaelChirico:
Shall we target a new 1-15-2 patch branch for this? @TysonStanley have they e-mailed you about needing to fix this quickly?

TysonStanley:
> Shall we target a new 1-15-2 patch branch for this? @TysonStanley have they e-mailed you about needing to fix this quickly?

They haven't. And the checks all are saying ok even tho that issue is reported below so not sure what their protocol is for that. It popped up in the 1.14.10 release right at the same time we were submitting 1.15.

jangorecki:
As long as CRAN doesn't ask for fix I think we can keep it in 1.15.99. Over time there may be some extra stuff needed for patch release and then we can push that together.

ben-schwen:
AFAIA CRAN is trying out new checks since some time which are not hard checks right now but might become in the future.

Keeping it in 1.15.99 should be fine.

tdhock:
there was a related message on R-devel today, quoted below.
```
Apologies for not answering the question you asked, but is this about
hdf5r and problems printing R_xlen_t [*] that appeared in 1.3.8 and you
tried to solve in 1.3.9?

We had a thread about this last November:
https://stat.ethz.ch/pipermail/r-package-devel/2023q4/010123.html

To summarise, there is no single standard C format specifier that can be
used to print R_xlen_t. As an implementation detail, it can be defined
as int or ptrdiff_t (or something completely different in the future),
and ptrdiff_t itself is usually defined as long or long long (or, also,
something completely different on a weirder platform). All three basic
types can have different widths and cause painful stack-related
problems when a mismatch happens.

In R-4.4, there will be a macro R_PRIdXLEN_T defining a compatible
printf specifier. Until then (and for compatibility with R-4.3 and
lower), it's relatively safe to cast to (long long) or (ptrdiff_t) and
then use the corresponding specifier, but that's not 100% future-proof.
Also, mind the warnings that mingw compilers sometimes emit for "new"
printf specifiers despite UCRT is documented to support them.
```

MichaelChirico (approved):

MichaelChirico:
Also cherry-picked to the 1-15-2 branch, where I also added a NEWS item:

https://github.com/Rdatatable/data.table/commit/52220da877f53653c195d6c6ef26540730fac656

https://github.com/Rdatatable/data.table/commit/e4912e9b5118d4c3e0763e2f6d30b8494d58c394

PR #5930, opened 2024-02-01
add R CMD check M1 runner

ben-schwen:
As mentioned in #5929

              Would adding an arm64 runner have helped here? Newly supported in GHA:

https://fosstodon.org/@gaborcsardi/111850319359214546

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/issues/5929#issuecomment-1920018767_
            
Also supposed to check if it really finds the current -Wformat warning


ben-schwen:
Superseded by #6389

PR #5910, opened 2024-01-15
Make declarations static for covr

MichaelChirico:
As noted https://github.com/Rdatatable/data.table/pull/5889#issuecomment-1890991129

PR #5937, opened 2024-02-15
merge 1-15-99 branch into master

tdhock:
hi @MichaelChirico @jangorecki I believe now we can merge the dev branch 1-15-99 into master, right?


MichaelChirico:
Not quite, we need to `rebase` first, not merge. I've just `--force` pushed the result of rebase, PTAL and make sure the history looks correct.

Crucially, we should _not_ squash commits during this merge so that each commit (which was squashed from a PR into the `1-15-99` branch) survives to `master`.

.github/workflows/R-CMD-check.yaml
```diff
@@ -3,12 +3,12 @@
 on:
   push:
     branches:
-      - main
       - master
+      - 1-15-99
```
MichaelChirico, line 7: once the current run (https://github.com/Rdatatable/data.table/actions/runs/7938397022, https://github.com/Rdatatable/data.table/actions/runs/7938397023) passes, let's remove this carve-out for `1-15-99` from CI as well.

NAMESPACE
```diff
@@ -51,6 +51,7 @@ S3method(cube, data.table)
 S3method(rollup, data.table)
 export(frollmean)
 export(frollsum)
+export(frollmax)
```
MichaelChirico, line 54: @jangorecki should we roll back #5889 before merging?

.dev/CRAN_Release.cmd
```diff
@@ -195,15 +195,15 @@ R CMD build .
 export GITHUB_PAT="f1c.. github personal access token ..7ad"
 # avoids many too-many-requests in --as-cran's ping-all-URLs step (20 mins) inside the `checking CRAN incoming feasibility...` step.
 # Many thanks to Dirk for the tipoff that setting this env variable solves the problem, #4832.
-R CMD check data.table_1.14.99.tar.gz --as-cran
```
MichaelChirico, line 198: this commit is somewhat strange to survive to `master`, but the changes here still look useful. @TysonStanley shall we add this commit (perhaps with a better commit message) directly to `master` instead of merging along with `1-15-99`?

MichaelChirico:
Hmm, it appears `rebase` messes around with authorship? Now I show up as author of most of the commits :\

Is there a better way to do this, or do I just need to manually edit the commit metadata to restore the "correct" author for each commit & `--force` push again?

.github/workflows/R-CMD-check.yaml
MichaelChirico: `drop`'d the relevant commit & force-pushed again

.dev/CRAN_Release.cmd
MichaelChirico: #5938

NAMESPACE
jangorecki: Probably better

MichaelChirico: OK, I hope it doesn't do anything too catastrophic with all the `--force` pushes going around. Since the diffs are still visible in the #5889 web UI, at worst we can reconstitute the changes manually.

MichaelChirico: NB: had to remove the #5910 changes to src/froll.c. Will need to reconstitute those.

MichaelChirico:
OK, current state LGTM. I'll leave it to someone else to merge (again, **not** squash!)

MichaelChirico (approved):

tdhock:
not sure what is the right way to do that... 
"Create a merge commit" or "Rebase and merge" ? What is the difference?

MichaelChirico:
Not 100% sure, but I went with "Rebase and merge". Seems to have worked -- I believe the former would have added a new commit like "merge branch 1-15-99 to master", which I don't think we want in this case.

PR #5941, opened 2024-02-19
New quiet option for cc()

MichaelChirico:
Split off from #4370 for clarity

### DIFF
--- a/.dev/cc.R
+++ b/.dev/cc.R
@@ -31,7 +31,7 @@ sourceDir = function(path=getwd(), trace = TRUE, ...) {
   if(trace) cat("\n")
 }
 
-cc = function(test=FALSE, clean=FALSE, debug=FALSE, omp=!debug, cc_dir, path=Sys.getenv("PROJ_PATH"), CC="gcc") {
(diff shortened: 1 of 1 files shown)


.dev/cc.R
```diff
@@ -81,7 +81,7 @@ cc = function(test=FALSE, clean=FALSE, debug=FALSE, omp=!debug, cc_dir, path=Sys
     .GlobalEnv[[Call$name]] = Call$address
   for (Extern in xx$.External)
     .GlobalEnv[[Extern$name]] = Extern$address
-  sourceDir(file.path(path, "R"))
+  sourceDir(file.path(path, "R"), trace=!quiet)
   if (base::getRversion()<"4.0.0") rm(list=c("rbind.data.table", "cbind.data.table"), envir=.GlobalEnv) # 3968 follow up
```
MichaelChirico, line 85: (for potential follow-up)

is it still useful to have such a condition in a script for dev work? should any dev work still be happening with now-several-years-old R? If this is for back-compatibility testing, wouldn't we want some flexibility in the `R CMD` calls invoked by `cc()` to instead be some other R binary?

MichaelChirico:
(manually approving Jan's authored change)

PR #5952, opened 2024-02-23
ignore any project-specific Rprofile

MichaelChirico:
.dev/.Rprofile discusses setting it as a _global_ profile, but this may interfere with development of other packages, or with scripting work. So some developers (e.g. me) might want to put it in the data.table project folder instead to run `cc()`, in which case we should ignore it in git.


MichaelChirico:
Doesn't touch any actual code, going ahead & merging, please feel free to comment if we need any follow-up.

PR #5953, opened 2024-02-23
Use GForce on data.table::.N

### LINKED ISSUE #5951: data.table::.N should not disable GForce
Apparently, I was too slow, just noticed that you didn't use = for assignments.

Maybe we should also document that `data.table::.N` will always be evaluated as `NULL`.

_Originally posted by @ben-schwen in https://github.com/Rdatatable/data.table/issues/5944#issuecomment-1959895514_

MichaelChirico:
Closes #5951

This entailed a bit more moving around than anticipated; in the process of splitting up the change now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # History files
 .RData
 .Rhistory
-.Rprofile
 .Rapp.history
 Rplots.pdf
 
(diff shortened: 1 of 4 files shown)


MichaelChirico:
(growing pains with trying out Graphite -- continued in #5955)

PR #5956, opened 2024-02-23
Trigger GHA on any PR

MichaelChirico:
Per docs:

https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-including-branches

Our current config is set up to _only_ run GHA on PRs that target `master` (or `main`, which we don't use). But I think we don't need that restriction, especially given recent experience with chained PRs targeting each other instead of `master` directly.


tdhock:
great thanks

PR #5950, opened 2024-02-22
gshift() of one column returns a vector, not list(vector)

MichaelChirico:
Closes #5939 

No NEWS since the NEWS item belongs in the patch branch.

PR #5963, opened 2024-02-25
gshift works in subset queries

### LINKED ISSUE #5962: gshift gives wrong results in subsetting queries
I encounter this problem too with the following example:

```
library(data.table)

dt <- expand.grid(
  date = 1:10,
  time = 1:10,
  id = 1:5
)

setDT(dt, key = c("date", "time", "id"))
dt[, x := runif(.N)]
dt[time == 1, x1 := shift(x, type = "lead"), by = id]
```

previous producing

```
Key: <date, time, id>
Index: <time>
      date  time    id         x        x1
     <int> <int> <int>     <num>     <num>
  1:     1     1     1 0.1217285 0.2041882
  2:     1     1     2 0.2499479 0.3632787
  3:     1     1     3 0.9992822 0.3553096
  4:     1     1     4 0.9851168 0.1797294
  5:     1     1     5 0.2273191 0.2967562
 ---                                      
496:    10    10     1 0.2503363        NA
497:    10    10     2 0.2226453        NA
498:    10    10     3 0.3320792        NA
499:    10    10     4 0.3255982        NA
500:    10    10     5 0.1976716        NA
```

but now producing

```
> dt
Key: <date, time, id>
Index: <time>
      date  time    id          x                                                                    x1
     <int> <int> <int>      <num>                                                                <list>
  1:     1     1     1 0.74905819 0.59769795,0.30605044,0.74046541,0.08694826,0.24060779,0.89059279,...
  2:     1     1     2 0.51506472 0.59769795,0.30605044,0.74046541,0.08694826,0.24060779,0.89059279,...
  3:     1     1     3 0.82395315 0.59769795,0.30605044,0.74046541,0.08694826,0.24060779,0.89059279,...
  4:     1     1     4 0.81986857 0.59769795,0.30605044,0.74046541,0.08694826,0.24060779,0.89059279,...
  5:     1     1     5 0.78196531 0.59769795,0.30605044,0.74046541,0.08694826,0.24060779,0.89059279,...
 ---                                                                                                   
496:    10    10     1 0.53337552                                                                      
497:    10    10     2 0.03719923                                                                      
498:    10    10     3 0.73477434                                                                      
499:    10    10     4 0.03107033                                                                      
500:    10    10     5 0.72256881                               
```

_Originally posted by @renkun-ken in https://github.com/Rdatatable/data.table/issues/5939#issuecomment-1962291854_

MichaelChirico:
Closes #5962. Split off from #5950.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,8 @@
 
 1. `unique()` returns a copy the case when `nrows(x) <= 1` instead of a mutable alias, [#5932](https://github.com/Rdatatable/data.table/pull/5932). This is consistent with existing `unique()` behavior when the input has no duplicates but more than one row. Thanks to @brookslogan for the report and @dshemetov for the fix.
 
+2. Optimized `shift` per group produced wrong results when simultaneously subsetting, for example, `DT[i==1L, shift(x), by=group]`, [#5962](https://github.com/Rdatatable/data.table/issues/5962). Thanks to @renkun-ken for the report and Benjamin Schwendinger for the fix.
+
 ## NOTES
 
 1. `transform` method for data.table sped up substantially when creating new columns on large tables. Thanks to @OfekShilon for the report and PR. The implemented solution was proposed by @ColeMiller1.
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -17832,6 +17832,9 @@ test(2224.2, DT[, shift(i, fill=a), by=x], error="fill must be a vector of lengt
 DT = data.table(x=pairlist(1), g=1)
 # unsupported type as argument
 test(2224.3, DT[, shift(x), g], error="Type 'list' is not supported by GForce gshift.")
+# gshift with subset #5962
+DT = data.table(x = 1:3, g = rep(1:2, each=3L))
+test(2224.4, DT[g==1L, shift(x), by=g, verbose=TRUE], data.table(g=1L, V1=c(NA,1L,2L)), output="Making each group and running j (GForce TRUE)")
 
 # groupingsets by named by argument
 test(2225.1, groupingsets(data.table(iris), j=sum(Sepal.Length), by=c('Sp'='Species'), sets=list('Species')),
--- a/src/gsumm.c
+++ b/src/gsumm.c
@@ -1203,7 +1203,7 @@ SEXP gshift(SEXP x, SEXP nArg, SEXP fillArg, SEXP typeArg) {
   bool lag;
   const bool cycle = stype == CYCLIC;
 
-  R_xlen_t nx = xlength(x), nk = length(nArg);
+  R_xlen_t nk = length(nArg);
   if (!isInteger(nArg)) error(_("Internal error: n must be integer")); // # nocov
   const int *kd = INTEGER(nArg);
   for (int i=0; i<nk; i++) if (kd[i]==NA_INTEGER) error(_("Item %d of n is NA"), i+1);
@@ -1220,7 +1220,7 @@ SEXP gshift(SEXP x, SEXP nArg, SEXP fillArg, SEXP typeArg) {
     }
     R_xlen_t ansi = 0;
     SEXP tmp;
-    SET_VECTOR_ELT(ans, g, tmp=allocVector(TYPEOF(x), nx));
+    SET_VECTOR_ELT(ans, g, tmp=allocVector(TYPEOF(x), n));
     #define SHIFT(CTYPE, RTYPE, ASSIGN) {                                                                         \
       const CTYPE *xd = (const CTYPE *)RTYPE(x);                                                                  \
       const CTYPE fill = RTYPE(thisfill)[0];                                                                      \


TysonStanley:
Looks like this is the last piece before we can ship 1.15.2. If this can be approved tomorrow, I can submit by EOD.

MichaelChirico:
> Looks like this is the last piece before we can ship 1.15.2. If this can be approved tomorrow, I can submit by EOD.

@TysonStanley please submit regardless -- if this isn't merged in time, we'll simply ship it with a new patch 1.15.4 afterwards. 1.15.2 already includes the urgent fix that CRAN wants, this PR is good-to-have but not strictly required.

TysonStanley:
Will do. Planning to submit Monday EOD (eastern time).

jangorecki (approved):
LGTM

MichaelChirico:
Merging to `master` now. I'll start the `1-15-4` patch branch once the `1.15.2` tag is up.

ben-schwen:
I added the small news from #5950

MichaelChirico:
> I added the small news from #5950

the NEWS doesn't (won't) belong to the 1.16.0 release, though, so I intentionally left it out.

This is a slightly strange aspect of the move to patch releases -- cc @jangorecki @tdhock is there anything different we should do here? To summarize the above, we merge this (and generally, any patch-targeted) change to `master` and there will be no mention of it in the NEWS for some indefinite period of time (until the corresponding patch is on CRAN). Is that expected?

Related but similar issue -- is it possible for a patch release to be superseded by a "normal" release (e.g. we have a branch `1-15-4` going, but wind up ready to submit `1-16-0` first)? Or would we always push out the patch branch first before proceeding with the normal release?

ben-schwen:
As long as we do not mix up numbers between patch release and normal release I do not see anything against a normal release superseding a patch release.

I thought that the intended workflow was that we merge everything against master and then cherry pick some merged (smashed) PRs for patch release. If we merge against master without NEWS then we have to keep a trail of non-completed NEWS which seems tedious and error prone

MichaelChirico:
> We could merge patch branch to master when it is on CRAN, then at least we will not have changes added without news entities.

that means excluding the fixes from `master` until a patch is on CRAN?

MichaelChirico:
> If we merge against master without NEWS then we have to keep a trail of non-completed NEWS which seems tedious and error prone

I agree, but the NEWS entries belong under the patch release heading (1.15.2), so it seems the alternative to me is to have NEWS entries for a release that does not exist (at least on CRAN).

One more option is migrating the NEWS items from "under development" to "patch" in a post-release commit. That may be the best option but also doesn't feel ideal.

jangorecki:
> > We could merge patch branch to master when it is on CRAN, then at least we will not have changes added without news entities.
> 
> that means excluding the fixes from `master` until a patch is on CRAN?

Not even excluding but simply not merging to master but to a patch branch, and once patch is on CRAN then merging it to master with its news.

MichaelChirico:
> > > We could merge patch branch to master when it is on CRAN, then at least we will not have changes added without news entities.
> > 
> > 
> > that means excluding the fixes from `master` until a patch is on CRAN?
> 
> Not even excluding but simply not merging to master but to a patch branch, and once patch is on CRAN then merging it to master with its news.

Still not sure I follow. To clarify that means:

Day 0, we get a report about a regression bug in a published version
Day 1, we write the PR to fix it and merge it to `patch`
(repeat the above for more patches depending on the urgency)
Day `n`, we submit `patch` to CRAN
Day `n+1`, when `patch` is accepted, merge `patch` to `master`.

That means from days `1` through `n+1`, the bug remains unfixed on `master`. Am I missing something? That doesn't seem very desirable.

jangorecki:
Yes, it is exactly what I meant. We could be merging patch branch to master even before release to CRAN if it is necessary, but then we need to merge it again after CRAN release. In such case master will temporarily contain patch NEWS items (let's say 1.15.2) before they are on CRAN, and on top of them will be devel NEWS entire (1.15.99).
That way we don't need to cherry pick individual commits but rather whole (or parts of) patch releases.

MichaelChirico:
I'm not a fan of that approach -- it leaves `master` with potentially serious regression issues unfixed, where the same issues are fixed on a patch branch.

It's not ideal for us, but I think this approach is the friendliest to users:

 1. Target `master` with regression fix PRs. Include a NEWS item under the 'devel' heading.
 2. After review, cherry-pick to `patch` branch, _being sure to move the NEWS item to the corresponding patch heading_
 3. After release, edit the NEWS to shuffle the released NEWS items around

IINM, this is also more similar to how things have been done historically.

ben-schwen:
> It's not ideal for us, but I think this approach is the friendliest to users:
> 
> 1. Target `master` with regression fix PRs. Include a NEWS item under the 'devel' heading.
> 2. After review, cherry-pick to `patch` branch, _being sure to move the NEWS item to the corresponding patch heading_
> 3. After release, edit the NEWS to shuffle the released NEWS items around
> 
> IINM, this is also more similar to how things have been done historically.

If that's our way to go then we should probably incorporate it into the CRAN_release checklist

MichaelChirico:
> If that's our way to go then we should probably incorporate it into the CRAN_release checklist

Definitely! Just want to get some consensus on the approach before applying it here and documenting it.

MichaelChirico:
OK I'm taking the two 👍 on https://github.com/Rdatatable/data.table/pull/5963#issuecomment-1982299746 as OK to proceed for now. We can easily revisit.

PR #5966, opened 2024-02-27
Set me + Jan as the default owners

MichaelChirico:
This should reduce the noise in PRs of always assigning Matt. I think Jan and I have the most context on the repo to be able to delegate PRs to an appropriate reviewer, so I put us as the default owners. Over time, this will be less relevant as we get more file-level ownership built up.

cc other committers: @tdhock @mattdowle @arunsrinivasan for viz.


tdhock:
fine with me

Issue #5351, opened 2022-03-16
"total" label in cube/rollup/groupingsets


markseeto:
Thanks for the data.table package.

I often find that after using cube/rollup/groupingsets, I want to fill in the NAs in the "total" rows. It would be nice if this could be done as part of the call to cube/rollup/groupingsets.

Example:

```
DT <- CJ(State = c("s1", "s2", "s3"),
         Group = c("g1", "g2"))[rep(1:6, c(2, 4, 3, 0, 1, 3))]

Counts <- cube(DT, .N, by = c("State", "Group"))

Counts
#     State Group  N
#  1:    s1    g1  2
#  2:    s1    g2  4
#  3:    s2    g1  3
#  4:    s3    g1  1
#  5:    s3    g2  3
#  6:    s1  <NA>  6
#  7:    s2  <NA>  3
#  8:    s3  <NA>  4
#  9:  <NA>    g1  6
# 10:  <NA>    g2  7
# 11:  <NA>  <NA> 13

# Fill in the NAs
for (v in c("State", "Group")) {
  Counts[is.na(get(v)), (v) := "Total"]
}

Counts[]
#     State Group  N
#  1:    s1    g1  2
#  2:    s1    g2  4
#  3:    s2    g1  3
#  4:    s3    g1  1
#  5:    s3    g2  3
#  6:    s1 Total  6
#  7:    s2 Total  3
#  8:    s3 Total  4
#  9: Total    g1  6
# 10: Total    g2  7
# 11: Total Total 13

# Maybe this could be done as part of the call to cube(), e.g.:
cube(DT, .N, by = c("State", "Group"), total.label = "Total")

# And it could allow different labels for different variables:
cube(DT, .N, by = c("State", "Group"),
     total.label = list(State = "National", Group = "Total"))


sessionInfo()

# R version 4.1.2 (2021-11-01)
# Platform: x86_64-w64-mingw32/x64 (64-bit)
# Running under: Windows 10 x64 (build 19044)

# Matrix products: default

# locale:
# [1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252
# [3] LC_MONETARY=English_Australia.1252 LC_NUMERIC=C
# [5] LC_TIME=English_Australia.1252

# attached base packages:
# [1] stats     graphics  grDevices utils     datasets  methods   base

# other attached packages:
# [1] data.table_1.14.2

# loaded via a namespace (and not attached):
# [1] compiler_4.1.2
```
[...]

PR #5973, opened 2024-03-03
Add total.label argument for groupingsets, cube, rollup

markseeto:
Closes https://github.com/Rdatatable/data.table/issues/5351

Added a `total.label` argument to the `rollup.data.table`, `cube.data.table`, and `groupingsets.data.table` functions.

A couple of comments:
- It was sometimes not clear what should be an error, what should be a warning, and what should be neither. I made choices that seem reasonable to me, but there would be alternatives that are also reasonable.
- I wondered whether `all.label` might be a better name than `total.label`, since the thing being calculated won't always be a total. But I left it as `total.label` because the documentation uses the terminology "(sub-)totals", and the word "total" is used in other software, so "total" seems to be widely accepted.


jangorecki:
As for the arg name I would keep it as "label"

R/groupingsets.R
```diff
@@ -57,6 +57,16 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
     stopf("Argument 'sets' must be a list of character vectors.")
   if (!is.logical(id))
     stopf("Argument 'id' must be a logical scalar.")
+  if (!(is.null(total.label) ||
+        (is.character(total.label) && length(total.label) == 1L) ||
+        (is.list(total.label) && all(vapply_1b(total.label, is.character)) &&
```
jangorecki, line 62: We can rule out list because AFAIU chsr vector will be sufficient. Pleasr correct me if I am wrong.

markseeto:
> As for the arg name I would keep it as "label"

@jangorecki Thanks for your comment. Do you mean change it from "total.label" to "label", or keep it as "total.label"?

R/groupingsets.R
jangorecki: Yes, as long as each element in the list is scalar then list is not necessary and character vector is sufficient.

markseeto: @jangorecki Thanks, I'll try that.

jangorecki:
Yes, change to shorter "label" groupings sets are all about (sub-)totals so we don't need to be so explicitly in  naming variable so verbosely. Possibly even better than "label" will he better to use "labels" so it is clear that multiple values are supported.

R/groupingsets.R
jangorecki: I think I was wrong and labels must not be character type, right? Then we need a list type. Would be good to use such little bit  more complex example so it comes up straight away.

jangorecki:
Very good PR. What would be even more useful is minimal example usage in NEWS entry. Including different data types for labels.

markseeto:
> Yes, change to shorter "label" groupings sets are all about (sub-)totals so we don't need to be so explicitly in naming variable so verbosely. Possibly even better than "label" will he better to use "labels" so it is clear that multiple values are supported.

@jangorecki I prefer "total.label" or "total.labels" because they make it easier to guess where the label is going to be used, but I don't feel strongly about this, so I'm happy to change it to "labels".

markseeto:
> Very good PR. What would be even more useful is minimal example usage in NEWS entry. Including different data types for labels.

@jangorecki Thanks for the feedback. I can add an example to the NEWS entry.

R/groupingsets.R
markseeto: @jangorecki No, currently the label values have to be character, and they're only applied to variables that are character or factor.

markseeto: Do you think total labels for other variable types should be allowed, and if so, which types?

For dates, I assume the user would be more likely to want a word like "Total" than to put in a date. For integers, they might want an integer code for the total, but if it was me I'd be more likely to want a word. I guess the function could convert the variable to character in the output, or leave it to the user to convert to character before using the function.

jangorecki: It has to support whatever 'by' supports AFAIU. I should have some time this week so can look at it from my laptop. Recently I am on mobile only.

jangorecki:
> > Yes, change to shorter "label" groupings sets are all about (sub-)totals so we don't need to be so explicitly in naming variable so verbosely. Possibly even better than "label" will he better to use "labels" so it is clear that multiple values are supported.
> 
> @jangorecki I prefer "total.label" or "total.labels" because they make it easier to guess where the label is going to be used, but I don't feel strongly about this, so I'm happy to change it to "labels".

Reason why I prefer single word argument name is it does not enforce any style, like total.label or totalLabel or total_label will do.

R/groupingsets.R
markseeto: Thanks @jangorecki. Currently it only supports character and factor grouping variables. For other types, it's left as NA.

In https://github.com/Rdatatable/data.table/issues/5351 you wrote
> Then we of course want total.label="Total" to be recycled to list(State = "Total", Group = "Total"). Of course only if all columns are character.

I interpreted this as meaning the label argument should only apply to variables of type character, but now I realise that maybe I interpreted incorrectly and you might have meant that `total.label = "Total"` should only be recycled to character variables, but non-character labels for non-character variables are still allowed in a list.

A couple of options:

1. Only apply the label argument to grouping variables of type character or factor, and require `total.label = c(Region = "National", Product = "Total")` instead of `total.label = list(Region = "National", Product = "Total")`.
2. Allow the label argument to apply to grouping variables that aren't character or factor, and allow total.label to be a list so that different types can be specified for different variables.

I prefer option 1, but I can try option 2 if you think that's better. As I mentioned above, if I had a date or integer grouping variable, I would probably still want a character total label (e.g. "Total") rather than a date or integer total label.

markseeto:
> > > Yes, change to shorter "label" groupings sets are all about (sub-)totals so we don't need to be so explicitly in naming variable so verbosely. Possibly even better than "label" will he better to use "labels" so it is clear that multiple values are supported.
> > 
> > 
> > @jangorecki I prefer "total.label" or "total.labels" because they make it easier to guess where the label is going to be used, but I don't feel strongly about this, so I'm happy to change it to "labels".
> 
> Another reason why I prefer single word argument name is it does not enforce any style, like total.label or totalLabel or total_label will do.

OK, I can change it. Maybe "label" is better than "labels", since "labels" is a base function.

R/groupingsets.R
jangorecki: Changing column type is not an option, column type needs to be controlled in j argument and not in labels (so no for option B). Option 2. We cannot apply it to all variables because we must match column types, we don't just print results but return a data.table so saying it is only for output won't be useful. Our output is data.table and column types cannot depend on labels arg. It is the labels arg has to match to column types in the output 'by'.

markseeto: By "output", I meant the data.table that is returned. Sorry for the confusion.

Thanks for clarifying that the column types have to stay the same. That makes it simpler.

Do you think the shorter form `label = "Total"` should be allowed, or should `label` always be a named list? If the shorter form is allowed, should it always have to be character, or would something like `label = 999L` be allowed and apply to all the integer grouping variables?

jangorecki: Should be allows but for non-char column will result an NA, which anyway is likely to be expected.

markseeto: Thanks, I'll give that a try.

markseeto:
@jangorecki I've updated the functions, tests, documentation, and news, based on our discussions. When you have time, please let me know what you think.

I wasn't sure about the correct way to update a PR. Apologies if I've done it incorrectly.

ben-schwen:
> I don't understand the reason for the red cross next to the News commit. When I click on the cross, it says "R-CMD-check / windows-latest (devel) (pull_request)" was cancelled, but in the list of checks, it says that check was successful.

As soon as you push another commit to the branch, we abort the CI runners to save computing minutes. Hence, you pushed your commit before the "R-CMD-check / windows-latest (devel) (pull_request)" job finished.

markseeto:
@ben-schwen Thanks for the explanation.

MichaelChirico:
@jangorecki any final review here? I can help merge to `master` & other tidying.

NEWS.md
```diff
@@ -30,6 +30,40 @@
...
+8. `groupingsets.data.table()`, `cube.data.table()`, and `rollup.data.table()` gain a `label` argument, which allows the user to specify a label for each grouping variable, to be included in the grouping variable column in the output in rows where the variable has been aggregated, [#5351](https://github.com/Rdatatable/data.table/issues/5351). Thanks to @markseeto for the request, @jangorecki and @markseeto for specifying the desired behaviour, and @markseeto for implementing.
+
+    ```R
+    DT = data.table(V1 = rep(c("a1", "a2"), each = 5),
+                    V2 = rep(rep(c("b1", "b2"), c(3, 2)), 2),
+                    V3 = rep(c("c1", "c2"), c(3, 7)),
+                    V4 = rep(1:2, c(6, 4)))
+
+    # Call groupingsets() and specify a label for V1, a different label for character grouping variables
+    # other than V1, and a label for integer grouping variables.
```
jangorecki, line 42: please rephrase, second line feels not clear.

ignore this comment if you are native eng speaker (I am not)

NEWS.md
```diff
@@ -30,6 +30,40 @@
...
+    
+    groupingsets(DT, .N, by = c("V1", "V2", "V3", "V4"),
+                 sets = list(c("V1", "V2", "V3"), c("V1", "V2", "V4"), c("V1", "V4"), "V2", "V3"),
+                 label = list(V1 = "All values", character = "Total", integer = 999L))
```
jangorecki, line 46: what about other types? in case of double type we could nicely present use of NaN, which may be well fitting this use case.

R/groupingsets.R
```diff
@@ -85,6 +135,33 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
     set(empty, j = "grouping", value = integer())
     setcolorder(empty, c("grouping", by, setdiff(names(empty), c("grouping", by))))
   }
+  # Define variables related to label
+  if (!is.null(label)) {
+    total.vars = intersect(by, unlist(lapply(sets, function(u) setdiff(by, u))))
+    if (is.list(label)) {
+      by.vars.not.in.label = setdiff(by, names(label))
+      by.vars.not.in.label.class1 = vapply_1c(x[, by.vars.not.in.label, with=FALSE], function(u) class(u)[1])
```
jangorecki, line 143: 1. please avoid calling `[.data.table` here
2. use 1L rather than 1

R/groupingsets.R
```diff
@@ -85,6 +135,33 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
...
+      labels.by.vars.not.in.label =
+        structure(label[by.vars.not.in.label.class1[by.vars.not.in.label.class1 %in% label.names.not.in.by]],
+                  names = by.vars.not.in.label[by.vars.not.in.label.class1 %in% label.names.not.in.by])
+      label.expanded = c(label[label.names.in.by], labels.by.vars.not.in.label)
+      label.expanded = label.expanded[intersect(by, names(label.expanded))] # reorder
+    } else {
+      by.vars.matching.scalar.class1 = by[vapply_1c(x[, by, with=FALSE], function(u) class(u)[1]) ==
+                                          class(label)[1]]
```
jangorecki, line 151: same

R/groupingsets.R
```diff
@@ -105,6 +182,10 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
       missing.int64.by.cols = setdiff(int64.by.cols, by.set)
       if (length(missing.int64.by.cols)) r[, (missing.int64.by.cols) := bit64::as.integer64(NA)]
     }
+    if (!is.null(label))
+      for (total.var in intersect(setdiff(by, by.set), names(label.use))) {
+        r[, (total.var) := label.use[[total.var]]]
```
jangorecki, line 187: We don't want to loop over `[.data.table`. I haven't looked closed but it feels it could be vectorized (handling of int64 2 lines above is indeed vectorized, therefore [ make sense there), but probably even simpler is to just loop with `set()`.

man/groupingsets.Rd
```diff
@@ -55,6 +60,11 @@ rollup(DT, j = sum(value), by = c("color","year","status")) # default id=FALSE
 rollup(DT, j = sum(value), by = c("color","year","status"), id=TRUE)
 rollup(DT, j = lapply(.SD, sum), by = c("color","year","status"), id=TRUE, .SDcols="value")
 rollup(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"), id=TRUE)
+rollup(DT, j = sum(value), by = c("color","year","status"),
+       label = list(color = "total", year = as.Date("3000-01-01"), status = factor("total"))) # specify label by variable name
+rollup(DT, j = sum(value), by = c("color","year","status"),
+       label = list(color = "total", Date = as.Date("3000-01-01"), factor = factor("total"))) # specify label by variable name and first element of class
+rollup(DT, j = sum(value), by = c("color","year","status"), label = "total") # label is character scalar so applies to color only
```
jangorecki, line 67: I would like also examples of using NA and NaN in label arg

jangorecki (changes requested):
haven't closely looked at the logic and test coverage but found few places which doesn't look perfect

NEWS.md
markseeto: I am a native English speaker but I've changed the comment to try to make it clearer:

> Call groupingsets() and specify a label for V1, a different label for the other character grouping variables, a label for the integer grouping variables, and a label for the numeric grouping variable.

NEWS.md
markseeto: I've added a NaN label in the example.

R/groupingsets.R
markseeto: I've changed this.

R/groupingsets.R
markseeto: I've changed this.

R/groupingsets.R
markseeto: Thanks, I've removed the loop.

man/groupingsets.Rd
markseeto: I've added these.

markseeto:
@jangorecki Thanks for reviewing. I've made the changes you asked for.

Regarding the code-quality / lint-r check, I fixed some of them but I hope it will be OK to leave long variable names because they're more descriptive that way.

jangorecki:
Please avoid force push when PR is already under reviews

jangorecki (approved):

markseeto:
> Please avoid force push when PR is already under reviews

Thanks for approving and thanks for this feedback. I don't know what "force push" means but will try to find out so I can avoid doing them.

MichaelChirico:
> I don't know what "force push" means

Here's what we're seeing:

image

Most of the time when this happens you get an error when trying `git push` about mismatch of remote&local, I believe the error tells you to retry with `--force`. If you saw something about "incompatible histories" I've also been running into the same of late, not sure why. Just try and be cognizant that `--force` should be avoided when possible, it makes the history harder to track.

MichaelChirico:
One other note, I see this is your `markseeto:master` branch in your fork, in my experience it's smart to make fork edits in a dedicated branch, there are some subtle headaches to dealing with PRs in fork branches named `master`.

Anyway, you'll get an invite to be a data.table project member, so you'll be able to create branches directly on this repo going forward, so the point is kind of moot for data.table specifically, but it's good advice for PRs to other packages.

NEWS.md
```diff
@@ -27,6 +27,42 @@ rowwiseDT(
 
 2. Limited support for subsetting or aggregating columns of type `expression`, [#5596](https://github.com/Rdatatable/data.table/issues/5596). Thanks to @tsp for the report, and @ben-schwen for the fix.
 
+3. `groupingsets.data.table()`, `cube.data.table()`, and `rollup.data.table()` gain a `label` argument, which allows the user to specify a label for each grouping variable, to be included in the grouping variable column in the output in rows where the variable has been aggregated, [#5351](https://github.com/Rdatatable/data.table/issues/5351). Thanks to @markseeto for the request, @jangorecki and @markseeto for specifying the desired behaviour, and @markseeto for implementing.
+
+    ```R
```
MichaelChirico, line 32: consistency (I don't _think_ capitalization matters here...)

```suggestion
    ```r
```

R/groupingsets.R
```diff
@@ -57,6 +57,14 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
     stopf("Argument 'sets' must be a list of character vectors.")
   if (!is.logical(id))
     stopf("Argument 'id' must be a logical scalar.")
+  if (!(is.null(label) ||
+        (is.atomic(label) && length(label) == 1L) ||
+        (is.list(label) && all(vapply_1b(label, is.atomic)) && all(lengths(label) == 1L) && !is.null(names(label)))))
+    stopf("Argument 'label', if not NULL, must be a scalar or a named list of scalars.")
+  if (is.list(label) && !is.null(names(label)) && ("" %chin% names(label) || anyNA(names(label))))
```
MichaelChirico, line 64: is this `!is.null(names(label))` check redundant? Since we have `is.list(label) && ... !is.null(names(label))` in the above requirement?

MichaelChirico: I do find it a bit surprising that the above check requires "a named list of scalars" but we have a separate test for "all list elements must be named", maybe best to add in the check for `""`/`NA` names to the above condition?

R/groupingsets.R
```diff
@@ -66,6 +74,43 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
     stopf("Character vectors in 'sets' list must not have duplicated column names within a single grouping set.")
   if (length(sets) > 1L && (idx<-anyDuplicated(lapply(sets, sort))))
     warningf("'sets' contains a duplicate (i.e., equivalent up to sorting) element at index %d; as such, there will be duplicate rows in the output -- note that grouping by A,B and B,A will produce the same aggregations. Use `sets=unique(lapply(sets, sort))` to eliminate duplicates.", idx)
+  if (is.list(label)) {
+    other.allowed.names = c("character", "integer", "numeric", "factor", "Date", "IDate")
+    allowed.label.list.names = c(by, vapply_1c(x[, by, with=FALSE], function(u) class(u)[1]),
```
MichaelChirico, line 79:
```suggestion
    allowed.label.list.names = c(by, vapply_1c(.shallow(x, by), function(u) class(u)[1]),
```

R/groupingsets.R
```diff
@@ -66,6 +74,44 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, ...)
...
+    label.names.in.by.classes = label.classes[label.names.in.by]
+    x.label.names.in.by.classes = lapply(.shallow(x, label.names.in.by), class)
+    label.names.in.by.classes.match = vapply_1b(label.names.in.by,
+                                                function(u) identical(label.names.in.by.classes[[u]],
+                                                                      x.label.names.in.by.classes[[u]]))
+    label.names.not.in.by.classes1 = vapply_1c(label.classes[label.names.not.in.by], function(u) u[1])
+    label.names.not.in.by.classes1.match = (label.names.not.in.by == label.names.not.in.by.classes1)
+    if (!all(label.names.in.by.classes.match)) {
+      label.names.in.by.classes.mismatch.info =
+        paste0(label.names.in.by[!label.names.in.by.classes.match],
```
MichaelChirico, line 97: See #6503, #6482 -- this approach to building the error message makes it hard to translate.

MichaelChirico: (resolving with be0377680d4dd266ef552ea1c43d2710277aa067)

MichaelChirico:
> I fixed some of them but I hope it will be OK to leave long variable names because they're more descriptive that way.

Honestly I think it's overdoing it here. Those variables have very limited scope for an error branch. I just used simple names `idx`/`info` here, I don't think the code is noticeably less readable.

R/groupingsets.R
MichaelChirico: We can address this as a small follow-up PR if you agree, don't want to hold the PR back further.

MichaelChirico (approved):
Great, thanks!

markseeto:
@MichaelChirico Thanks for checking and tidying the code.

> Here's what we're seeing:
> 
> image
> 
> Most of the time when this happens you get an error when trying `git push` about mismatch of remote&local, I believe the error tells you to retry with `--force`. If you saw something about "incompatible histories" I've also been running into the same of late, not sure why. Just try and be cognizant that `--force` should be avoided when possible, it makes the history harder to track.

Thanks for the explanation. I see that too. I used the GitHub GUI in the browser rather than command-line git (I haven't learned command-line git yet). I didn't explicitly select "force push", but just synced my fork and then made the changes, and GitHub automatically did the force push. Maybe I should have just changed the files without syncing the fork first.

> One other note, I see this is your `markseeto:master` branch in your fork, in my experience it's smart to make fork edits in a dedicated branch, there are some subtle headaches to dealing with PRs in fork branches named `master`.

Thanks, I'll keep that in mind.

> I guess this PR was created before the invite.

Yes, this PR was started before I was a member.

R/groupingsets.R
markseeto: > is this `!is.null(names(label))` check redundant?

Yes, I think you're right.

> I do find it a bit surprising that the above check requires "a named list of scalars" but we have a separate test for "all list elements must be named", maybe best to add in the check for `""`/`NA` names to the above condition?

Maybe, but with separate checks the error messages can be more specific, the second one being for the situation where `label` is a named list but not all elements have a name. If we combine the error messages into one, it would be something like "Argument 'label', if not NULL, must be (1) a scalar, or (2) a named list with each element being named and each element being a scalar." Or "Argument 'label', if not NULL, must be (1) a scalar, or (2) a named list with no names being "" or NA and each element being a scalar." I think these are less clear than having separate error messages depending on the situation.

Issue #5969, opened 2024-02-29
fread cleanup prints error message in webR


maek-ies:
data.table is working quite fine in webR, but when using fread an error occurs (_**System errno 8 unmapping file: Bad file descriptor**_). Interestingly, the file loads but it throws an error neverthless. Same error occurs also for .gz files. Perhaps, it's just the fact that this runs in browser using wasm is not recognized, resulting in an error.

```r
install.packages("data.table")
library("data.table")

dt = data.table(1:10)
fwrite(dt,"dt.csv")
list.files()
dtcsv = fread("dt.csv", verbose = T)
print(dtcsv)

sessionInfo()
```

Resulting output:

`> install.packages("data.table")
Downloading webR package: data.table

> library("data.table")
data.table 1.14.7 IN DEVELOPMENT built 2023-12-19 09:51:47 UTC using 1 threads (see ?getDTthreads).  Latest news: r-datatable.com
**********
This development version of data.table was built more than 4 weeks ago. Please update: data.table::update_dev_pkg()
**********
**********
This installation of data.table has not detected OpenMP support. It should still work but in single-threaded mode.
This is Emscripten. This warning should not normally occur on Windows or Linux where OpenMP is turned on by data.table's configure script by passing -fopenmp to the compiler. If you see this warning on Windows or Linux, please file a GitHub issue.
**********

> dt = data.table(1:10)

> fwrite(dt,"dt.csv")

> list.files()
[1] "dt.csv"

> dtcsv = fread("dt.csv", verbose = T)
This installation of data.table has not been compiled with OpenMP support.
  omp_get_num_procs()            1
  R_DATATABLE_NUM_PROCS_PERCENT  unset (default 50)
  R_DATATABLE_NUM_THREADS        unset
  R_DATATABLE_THROTTLE           unset (default 1024)
  omp_get_thread_limit()         1
  omp_get_max_threads()          1
  OMP_THREAD_LIMIT               unset
  OMP_NUM_THREADS                unset
  RestoreAfterFork               true
  data.table is using 1 threads with throttle==1024. See ?setDTthreads.
freadR.c has been passed a filename: dt.csv
[01] Check arguments
  Using 1 threads (omp_get_max_threads()=1, nth=1)
  NAstrings = [<<NA>>]
  None of the NAstrings look like numbers.
  show progress = 1
  0/1 column will be read as integer
[02] Opening the file
  Opening file dt.csv
  File opened, size = 24 bytes.
  Memory mapped ok
[...]

PR #5974, opened 2024-03-04
Wasm: Keep file descriptor open in `fread.c` until finished with `mmap()`

georgestagg:
See #5969 for discussion.


src/fread.c
```diff
@@ -66,6 +66,9 @@ static size_t fileSize;
 static int8_t *type = NULL, *tmpType = NULL, *size = NULL;
 static lenOff *colNames = NULL;
 static freadMainArgs args = {0};  // global for use by DTPRINT; static implies ={0} but include the ={0} anyway just in case for valgrind #4639
+#ifdef __EMSCRIPTEN__
+  static int mmp_fd = -1; // Under Wasm we must keep the fd open until we're done with the mmap, due to an Emscripten bug (#5969).
```
MichaelChirico, line 70:
```suggestion
  // Under Wasm we must keep the fd open until we're done with the mmap, due to an Emscripten bug (#5969).
  // TODO(emscripten-core/emscripten#20459): revert this
  static int mmp_fd = -1;
```

BTW, what does this mean in terms of dependencies? I.e., if/when the Emscripten bug is fixed, if we remove this patch, it induces a version dependency on Emscripten, right? What would be the proper way to express that? Something within webR only? Something in our DESCRIPTION file?

MichaelChirico:
Thanks for the PR!! Really appreciate it. Please write a NEWS item to cite yourself, I guess this is a bug fix? Otherwise a note.

Lastly, can you think of a way to regression test this? It's probably not worth the effort, but maybe you're familiar with an easy way to do so.

src/fread.c
georgestagg: > if we remove this patch, it induces a version dependency on Emscripten

Technically true, but in any case packages built for a specific version of webR already strictly depend on a specific version of Emscripten, since Emscripten does not guarantee ABI compatibility between versions.

I think such a strong dependency link should probably be handled by webR itself and the [rwasm](https://github.com/r-wasm/rwasm) package, since it will be the same for all R packages built for a certain version of webR anyway.

georgestagg:
> Lastly, can you think of a way to regression test this? It's probably not worth the effort, but maybe you're familiar with an easy way to do so.

Not easily, sorry. We have thought a little about what automated testing could look like inside the Wasm environment, but the infrastructure is not there yet.

MichaelChirico (approved):
Thanks so much for handling this!

PR #5976, opened 2024-03-04
Improved warning message in fcast.R

### LINKED ISSUE #5217: Improve "Aggregate function missing, defaulting to 'length'" message
My guess is that most often, this happens by mistake, in which case a new/inexperienced user's response is likely along the lines of "aggregation? what aggregation?".

The culprit is here:

https://github.com/Rdatatable/data.table/blob/c6f6dd80b7ad565d9a78221fbe32ea0c5e0bf547/R/fcast.R#L156-L162

Almost certainly we can use that `forderv` output to provide a bit more context for diagnosing the issue.

Nj221102:
This PR Closes #5217 

   1. By adding a better warning message which informs the user seeing the warning about the cause of issue and how to solve it.
    
   2. Hence making User experience better for beginners :)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,11 +18,13 @@
 
 3. `split.data.table` also accepts a formula for `f`, [#5392](https://github.com/Rdatatable/data.table/issues/5392), mirroring the same in `base::split.data.frame` since R 4.1.0 (May 2021). Thanks to @XiangyunHuang for the request, and @ben-schwen for the PR.
 
-3. Namespace-qualifying `data.table::shift()`, `data.table::first()`, or `data.table::last()` will not deactivate GForce, [#5942](https://github.com/Rdatatable/data.table/issues/5942). Thanks @MichaelChirico for the proposal and fix. Namespace-qualifying other calls like `stats::sum()`, `base::prod()`, etc., continue to work as an escape valve to avoid GForce, e.g. to ensure S3 method dispatch.
+4. Namespace-qualifying `data.table::shift()`, `data.table::first()`, or `data.table::last()` will not deactivate GForce, [#5942](https://github.com/Rdatatable/data.table/issues/5942). Thanks @MichaelChirico for the proposal and fix. Namespace-qualifying other calls like `stats::sum()`, `base::prod()`, etc., continue to work as an escape valve to avoid GForce, e.g. to ensure S3 method dispatch.
 
 ## BUG FIXES
 
 1. `unique()` returns a copy the case when `nrows(x) <= 1` instead of a mutable alias, [#5932](https://github.com/Rdatatable/data.table/pull/5932). This is consistent with existing `unique()` behavior when the input has no duplicates but more than one row. Thanks to @brookslogan for the report and @dshemetov for the fix.
+
+2. `dcast` handles coercion of `fill` to `integer64` correctly, [#4561](https://github.com/Rdatatable/data.table/issues/4561). Thanks to @emallickhossain for the bug report and @MichaelChirico for the fix.
 
 ## NOTES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,7 +36,7 @@
 
 4. Erroneous assignment calls in `[` with a trailing comma (e.g. ``DT[, `:=`(a = 1, b = 2,)]``) get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
 
-5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary does better explaining the behavior, which can be confusing if duplicates were unexpected, and suggesting alternatives. cified aggregate function, offering clearer guidance, especially for beginners, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
+5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary, which could be confusing if duplicates were unexpected, does better explaining the behavior and suggesting alternatives, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
 
 # data.table [v1.15.0](https://github.com/Rdatatable/data.table/milestone/29)  (30 Jan 2024)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,8 @@
 
 4. Erroneous assignment calls in `[` with a trailing comma (e.g. ``DT[, `:=`(a = 1, b = 2,)]``) get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
 
+5. Improved the console message within the `dcast.data.table` function in the `fcast.R` file. This enhancement clarifies situations where duplicate row/column combinations exist in the data table without a specified aggregate function, offering clearer guidance, especially for beginners. Solution implemented by @Nj221102.
+
 # data.table [v1.15.0](https://github.com/Rdatatable/data.table/milestone/29)  (30 Jan 2024)
 
 ## BREAKING CHANGE
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,7 +36,7 @@
 
 4. Erroneous assignment calls in `[` with a trailing comma (e.g. ``DT[, `:=`(a = 1, b = 2,)]``) get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
 
-5. Improved the console message within the `dcast.data.table` function in the `fcast.R` file. This enhancement clarifies situations where duplicate row/column combinations exist in the data table without a specified aggregate function, offering clearer guidance, especially for beginners. Solution implemented by @Nj221102.
+5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary does better explaining the behavior, which can be confusing if duplicates were unexpected, and suggesting alternatives. cified aggregate function, offering clearer guidance, especially for beginners, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
 
 # data.table [v1.15.0](https://github.com/Rdatatable/data.table/milestone/29)  (30 Jan 2024)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,7 +36,9 @@
 
 4. Erroneous assignment calls in `[` with a trailing comma (e.g. ``DT[, `:=`(a = 1, b = 2,)]``) get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
 
-5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary, which could be confusing if duplicates were unexpected, does better explaining the behavior and suggesting alternatives, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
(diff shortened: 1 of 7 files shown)


MichaelChirico:
Thanks so much for the PR!

Reading now. At a high level, please be sure to:

 1. Add a NEWS entry (under Notes), being sure to cite yourself!
 2. Add a regression test in inst/tests/test.Rraw

See https://github.com/Rdatatable/data.table/blob/master/.github/CONTRIBUTING.md for more details.

R/fcast.R
```diff
@@ -156,7 +156,7 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
   if (is.null(fun.call)) {
     oo = forderv(dat, by=varnames, retGrp=TRUE)
     if (attr(oo, 'maxgrpn', exact=TRUE) > 1L) {
-      messagef("Aggregate function missing, defaulting to 'length'")
+      messagef("Warning: Duplicated Combinations Detected. This occurs when combinations of rows and columns are not unique, meaning certain combinations representing rows and columns in the new data table repeat in the original data table, pointing to more than one value in original data table. However, we only require a single representative value for each combination in new data table. This can be achieved using an aggregation function, which consolidates multiple values into one as needed. Functions like mean or sum are suitable for this purpose. but If no aggregation function is explicitly specified by user, then the length of values (i.e., the number of different values associated with each combination) is used as the unique value. To customize this behavior, user can input their desired aggregation function to manipulate the values accordingly. For further details, type `?dcast` in the terminal.")
```
MichaelChirico, line 159: This is a really great start and already a substantial improvement over the original message!

NB: Using markdown quotes (`>`) instead of a suggestion (```` ```suggestion ````) because the message is so long, I think it'll be easier to read than an inline diff of a very wide line.

Some high level comments about what changes I'm trying to make:

 1. Since this is a `message()`, "warning" is not the right framing of this output
 2. We should also make clear up front that this message is only coming because `fun.aggregate` was not provided
 3. "This occurs... more than one value in original data table." is a bit too wordy for me, see my attempt to rephrase and make sure it still conveys the idea to you (I also think we should make clear these rows/columns are derived from the `formula=` argument)
 4. We can also include `varnames` in the message to really hammer the point home

> 'fun.aggregate' is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables (%s) used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). See `?dcast.data.table` for more details.

Template value `%s` would be `brackify(varnames)`.

I'm still thinking of the right way to present insights about _what_ combinations are causing the multiplicity issue. The following is a very basic attempt to give something useful, but it suffers from not being robust enough

```r
print(dat[, .N, by=varnames][N > 1L])
```

The issue is that `varnames` _need not all be in `dat`_! Continuing the example from the issue comments (https://github.com/Rdatatable/data.table/issues/5217#issuecomment-1053960019)

```r
grp3 = DT$grp2
dcast(DT, grp1 ~ grp3, value.var="val")
```

Which complicates things a fair amount.

Nj221102:
@MichaelChirico  I am done with the changes , if this needs further changes plz suggest : )

Nj221102:
@MichaelChirico @jangorecki  can you plz review this PR

inst/tests/tests.Rraw
```diff
@@ -13593,7 +13593,7 @@ test(1962.087, dcast(DT, a ~ b, value.var = 'b'),
 test(1962.088, dcast(DT[0L, ], a ~ c, value.var = 'b'), data.table(a=numeric(), key="a")) #1215
 test(1962.089, dcast(DT, a ~ c, value.var = 'b'),
      data.table(a = c(1, 2), `2` = c(0L, 2L), `4` = c(2L, 0L), key = 'a'),
-     message = 'Aggregate function missing')
+     message = "fun.aggregate is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables [a, c] used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). See ?dcast.data.table for more details.")
```
tdhock, line 13596: [a, c] is helpful, good

tdhock:
please edit your initial comment in this PR so that is contains the following (which is recognized specially and will automatically close the issue when this PR is merged)

Closes #5217

Nj221102:
> please edit your initial comment in this PR so that is contains the following (which is recognized specially and will automatically close the issue when this PR is merged)
> 
> Closes #5217

done :)

NEWS.md
```diff
@@ -36,6 +36,8 @@
 
 4. Erroneous assignment calls in `[` with a trailing comma (e.g. ``DT[, `:=`(a = 1, b = 2,)]``) get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
 
+5. Improved the console message within the `dcast.data.table` function in the `fcast.R` file. This enhancement clarifies situations where duplicate row/column combinations exist in the data table without a specified aggregate function, offering clearer guidance, especially for beginners. Solution implemented by @Nj221102.
```
MichaelChirico, line 39:
```suggestion
5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary does better explaining the behavior, which can be confusing if duplicates were unexpected, and suggesting alternatives. cified aggregate function, offering clearer guidance, especially for beginners, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
```

R/fcast.R
```diff
@@ -154,10 +154,10 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
...
   fill.default = NULL
   if (is.null(fun.call)) {
-    oo = forderv(dat, by=varnames, retGrp=TRUE)
-    if (attr(oo, 'maxgrpn', exact=TRUE) > 1L) {
-      messagef("Aggregate function missing, defaulting to 'length'")
-      fun.call = quote(length)
+  oo = forderv(dat, by=varnames, retGrp=TRUE)
+  if (attr(oo, 'maxgrpn', exact=TRUE) > 1L) {
+    message(sprintf("fun.aggregate is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables %s used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). See ?dcast.data.table for more details.", paste(brackify(varnames), collapse = ", ")))
+    fun.call = quote(length)
```
MichaelChirico, line 160:
```suggestion
    oo = forderv(dat, by=varnames, retGrp=TRUE)
    if (attr(oo, 'maxgrpn', exact=TRUE) > 1L) {
      message(sprintf("fun.aggregate is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables %s used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). See ?dcast.data.table for more details.", paste(brackify(varnames), collapse = ", ")))
      fun.call = quote(length)
```

inst/tests/tests.Rraw
MichaelChirico: I would not test against the full contents of the message -- that will require us to change the test for minor issues like small copyediting/grammar/wording changes that don't materially affect the message.

Better would be a regex that captures the most important part of the message, e.g.

"fun.aggregate is NULL.*defaulting to length\\(\\).*\\[a, c\\]"

po/R-data.table.pot
```diff
@@ -1339,7 +1339,7 @@ msgstr ""
 
 #: fcast.R:159
 #, c-format
-msgid "Aggregate function missing, defaulting to 'length'"
+msgid "fun.aggregate is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables [a, c] used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). See ?dcast.data.table for more details."
```
MichaelChirico, line 1342: please revert this -- we edit .po/.pot files all at once before release, no need to make these changes for individual PRs.

R/fcast.R
```diff
@@ -156,7 +156,7 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
   if (is.null(fun.call)) {
     oo = forderv(dat, by=varnames, retGrp=TRUE)
     if (attr(oo, 'maxgrpn', exact=TRUE) > 1L) {
-      messagef("Aggregate function missing, defaulting to 'length'")
+      messagef("fun.aggregate is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables %s used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). See ?dcast.data.table for more details.", brackify(varnames))
```
MichaelChirico, line 159:
```suggestion
      messagef("'fun.aggregate' is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables %s used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). Check the resulting table for values larger than 1 to see which combinations were not unique. See ?dcast.data.table for more details.", brackify(varnames))
```

MichaelChirico: Adding one more sentence about using the results to find the duplicates. I think this is the best we can hope for given the issues with `formula=` inputs not necessarily being found directly in `data`.

po/R-data.table.pot
Nj221102: done 👍

inst/tests/tests.Rraw
Nj221102: > I would not test against the full contents of the message -- that will require us to change the test for minor issues like small copyediting/grammar/wording changes that don't materially affect the message.
> 
> Better would be a regex that captures the most important part of the message, e.g.
> 
> "fun.aggregate is NULL._defaulting to length\(\)._\[a, c\]"

done 👍

NEWS.md
```diff
@@ -36,6 +36,8 @@
 
 4. Erroneous assignment calls in `[` with a trailing comma (e.g. ``DT[, `:=`(a = 1, b = 2,)]``) get a friendlier error since this situation is common during refactoring and easy to miss visually. Thanks @MichaelChirico for the fix.
 
+5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary does better explaining the behavior, which can be confusing if duplicates were unexpected, and suggesting alternatives. cified aggregate function, offering clearer guidance, especially for beginners, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
```
MichaelChirico, line 39:
```suggestion
5. `dcast()` message about `fun.aggregate` defaulting to `length()` when aggregation is necessary, which could be confusing if duplicates were unexpected, does better explaining the behavior and suggesting alternatives, [#5217](https://github.com/Rdatatable/data.table/issues/5217). Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
```

MichaelChirico:
Looks great, thanks for taking this on!

PR #5983, opened 2024-03-06
New rtools43 URL

### LINKED ISSUE #5979: win-dev and win-rel CI jobs failed
Root cause not identified but I wasn't looking into it, just noticed they failed. Maybe simply url upgrade to recent R? not sure.

We need additionaly to ensure that jobs are listed in https://rdatatable.gitlab.io/data.table/web/checks/check_results_data.table.html where currently those jobs are missing

MichaelChirico:
Closes #5979 (I think)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,11 +14,11 @@ variables:
              ## a non-UTC timezone, although, that's what we do routinely in dev.
   R_REL_VERSION:  "4.3"
   R_REL_WIN_BIN:  "https://cloud.r-project.org/bin/windows/base/old/4.3.2/R-4.3.2-win.exe"
-  RTOOLS_REL_BIN: "https://cloud.r-project.org/bin/windows/Rtools/rtools43/files/rtools43-5863-5818.exe"
+  RTOOLS_REL_BIN: "https://cloud.r-project.org/bin/windows/Rtools/rtools43/files/rtools43-5958-5975.exe"
   RTOOLS43_HOME:  "/c/rtools"
   R_DEV_VERSION:  "4.4"
   R_DEV_WIN_BIN:  "https://cloud.r-project.org/bin/windows/base/R-devel-win.exe"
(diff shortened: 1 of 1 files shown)


MichaelChirico:
@jangorecki should we make this `curl` invocation `--fail` if the file is not found to make the error clearer in the future?

https://github.com/Rdatatable/data.table/blob/d09d74105fa3553b3ee3e43501130e27b4395bae/.gitlab-ci.yml#L215

MichaelChirico:
Testing: https://gitlab.com/Rdatatable/data.table/-/commit/8f5e086890de90758c6d549b009f0b60023a087c

MichaelChirico:
Hmm `curl --fail` changed the failure message but it's only a tiny improvement:

```
Start-Process : This command cannot be run due to the error: The system cannot find the file specified.
At line:301 char:50
+ ... BIN --fail; Start-Process -FilePath ..\rtools.exe -ArgumentList "/VER ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Start-Process], InvalidOperationException
    + FullyQualifiedErrorId : InvalidOperationException,Microsoft.PowerShell.Commands.StartProcessCommand
```

I think because `;` means the `Start-Process` command runs even if `curl.exe` fails. Not very familiar with PowerShell, but maybe `&&`/`||` work similarly enough to Linux (per https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_pipeline_chain_operators?view=powershell-7.4)...

MichaelChirico:
This fixes the issues with `win-rel` and `win-dev` which are now passing, see

https://gitlab.com/Rdatatable/data.table/-/jobs/6337150032
https://gitlab.com/Rdatatable/data.table/-/jobs/6337150033

Merging

PR #6002, opened 2024-03-14
Remove exported [.data.table

MichaelChirico:
As noted this was for `DT()`; we can (possibly) restore the export if we restore `DT()`. For now, I find it strange to export such a method.


MichaelChirico:
This breaks related tests:

https://github.com/Rdatatable/data.table/blob/475115fd00e82008ff739406b0e4efe133c8b73e/inst/tests/tests.Rraw#L17653-L17707

Should we:

 1. Comment out those tests
 2. Add `` `[.data.table` = data.table:::`[.data.table` `` alias at the top of the suite

jangorecki:
Why do we need to export it anyway? Wasn't there issue asking for that, which has been closer when that was added?

MichaelChirico:
I haven't looked into it, presumably git blame will tell us some more history

MichaelChirico:
Here's the relevant commit, presumably there's a smarter way to construct the call:

https://github.com/Rdatatable/data.table/pull/5176/files

ColeMiller1:
We should probably comment out both the export and the tests. Since they are related to `DT()` and that was pulled back, it makes sense to clean up the namespace. 

As far as simplifying the function call construct, I am unsure. While I think we will adjust to R 4.3 `dt |> _[...]`, it might be nice to clean up `DT()` to only support data.tables. It very well might not simplify much, but it is reasonable to expect users to use a data.table with data.table method such as `[.data.table`.

MichaelChirico:
Nobody is calling this directly, it is pretty much fully safe to unexport it:

https://github.com/search?q=lang%3AR+%2F%5Cn%5B%5E%23%5Cn%5D*%5B%5E%3A%5D%5B%27%22%60%5D%5C%5B%5C.data%5C.table%5B%60%27%22%5D%5C%28%2F&type=code

MichaelChirico:
@tdhock PTAL if you have any concerns about merging here, thanks!

tdhock:
lgtm thanks!

PR #6004, opened 2024-03-15
Remove importClassesFrom entry in NAMESPACE

MichaelChirico:
As discussed in the r-package-devel thread:

https://stat.ethz.ch/pipermail/r-package-devel/2024q1/010581.html

The proper approach for the downstream is to use `.datatable.aware=TRUE` to accomplish what they're after (ensuring `[.data.table` dispatches as intended).

I filed a PR with downstream to correct the issue:

https://github.com/yikeshu0611/do/pull/2

I think we should go ahead and merge this for 1.16.0, but let's leave it pending for now while we give downstream maintainer some time to respond.


tdhock:
thanks for investigating!

MichaelChirico:
Going ahead to merge here, no news from downstream.

PR #6005, opened 2024-03-15
Documentation for char.trunc('datatable.prettyprint.char')

### LINKED ISSUE #5994: options(datatable.prettyprint.char) is not documented
Thanks for the investigation! I confirm your findings and agree that it may not be so simple as just changing to `type='width'`. `substr()` is also involved; note this from `?substr`:

> These functions are often used with `nchar` to truncate a display. That does not really work (you want to limit the width, not the number of characters, so it would be better to use `strtrim`), but at least make sure you use the default `nchar(type = "c")`.

PS, it seems `options(datatable.prettyprint.char)` is completely undocumented, so we should address that too.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/issues/5096#issuecomment-1987501682_

Nj221102:
This PR Closes #5994  by adding the required documentation within ` man/print.data.table.Rd ` : )


MichaelChirico:
cc @joshhwuu who most recently had to learn this parameter from scratch :)

man/print.data.table.Rd
```diff
@@ -43,6 +45,7 @@
   \item{col.names}{ One of three flavours for controlling the display of column names in output. \code{"auto"} includes column names above the data, as well as below the table if \code{nrow(x) > 20}. \code{"top"} excludes this lower register when applicable, and \code{"none"} suppresses column names altogether (as well as column classes if \code{class = TRUE}. }
   \item{print.keys}{ If \code{TRUE}, any \code{\link{key}} and/or \code{\link[=indices]{index}} currently assigned to \code{x} will be printed prior to the preview of the data. }
   \item{trunc.cols}{ If \code{TRUE}, only the columns that can be printed in the console without wrapping the columns to new lines will be printed (similar to \code{tibbles}). }
+  \item{trunc.char}{If a number is provided as the value for \code{"datatable.prettyprint.char"}, it sets a limit for the length of strings processed by \code{char.trunc}. Strings longer than this limit have their excess characters removed, and "..." is appended to indicate omitted content.}
```
joshhwuu, line 48:
```suggestion
  \item{trunc.char}{ If \code{"datatable.prettyprint.char"} is a number greater than 0, \code{data.table.print} limits the lengths of strings to be equal to or less than \code{"datatable.prettyprint.char"}. Strings longer than this limit have their excess characters removed, and "..." is appended to the end of the string to indicate omitted content. Strings shorter than this limit are left unchanged. }
```

man/print.data.table.Rd
```diff
@@ -53,6 +56,8 @@
   \code{format_col} returns a \code{length(x)}-size \code{character} vector.
 
   \code{format_list_item} returns a length-1 \code{character} scalar.
+
+  \code{char.trunc} returns \code{x} if its length is within the limit else it returns the shrunk down version of \code{x} with "..." at end.
```
joshhwuu, line 60:
```suggestion
  \code{char.trunc} returns \code{x} if its length is within or equal to \code{"datatable.prettyprint.char"} else it returns the truncated version of \code{x} with "..." concatenated to the end.
```

man/print.data.table.Rd
```diff
@@ -98,6 +103,16 @@
   print(DT, trunc.cols=TRUE)
   options(old_width)
 
+  # `char.trunc` will shorten the string if its length exceed the give limit in `datatable.prettyprint.char`
```
joshhwuu, line 106:
```suggestion
  # `char.trunc` will truncate the string if its length exceeds the given limit: `datatable.prettyprint.char`
```

man/print.data.table.Rd
joshhwuu: Not sure if there are any rules regarding this but my suggested description is quite long - could omit some parts of it if we want to keep it simple. @MichaelChirico any thoughts?

Something like:

```R
\item{trunc.char}{ If \code{"datatable.prettyprint.char"} is a number greater than 0, \code{data.table.print} limits the lengths of printed strings to be equal to or less than \code{"datatable.prettyprint.char"} by removing excess characters and concatenating "..." to the end.
```

R/print.data.table.R
```diff
@@ -7,6 +7,7 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
                col.names=getOption("datatable.print.colnames"),
                print.keys=getOption("datatable.print.keys"),
                trunc.cols=getOption("datatable.print.trunc.cols"),
+               trunc.char=getOption("datatable.prettyprint.char"),
```
MichaelChirico, line 10: Hmm I didn't really mean to expose it as a new parameter. I just wanted to describe the option. If there is need to expose it I would wait for a specific FR to that end. So for now, just describing the parameter in `?print.data.table` is enough.

Nj221102: well this error occurs if we don't expose it as a parameter,

```
odoc mismatches from documentation object 'print.data.table':
print.data.table
  Code: function(x, topn = getOption("datatable.print.topn"), nrows =
                 getOption("datatable.print.nrows"), class =
                 getOption("datatable.print.class"), row.names =
                 getOption("datatable.print.rownames"), col.names =
                 getOption("datatable.print.colnames"), print.keys =
                 getOption("datatable.print.keys"), trunc.cols =
                 getOption("datatable.print.trunc.cols"), quote =
                 FALSE, timezone = FALSE, ...)
  Docs: function(x, topn = getOption("datatable.print.topn"), nrows =
                 getOption("datatable.print.nrows"), class =
                 getOption("datatable.print.class"), row.names =
                 getOption("datatable.print.rownames"), col.names =
                 getOption("datatable.print.colnames"), print.keys =
                 getOption("datatable.print.keys"), trunc.cols =
                 getOption("datatable.print.trunc.cols"), trunc.char =
                 getOption("datatable.prettyprint.char"), quote =
                 FALSE, timezone = FALSE, ...)
  Argument names in docs not in code:
    trunc.char
    ```

Comment (MichaelChirico) [reply R/print.data.table.R:10]: Right, `trunc.char` should be an argument neither in `?print.data.table` nor in the signature defined in `print.data.table.R`. Instead we should just describe how character columns are printed within the existing `?print.data.table`, and mention the option.

Comment (Nj221102) [reply R/print.data.table.R]: Done : ) ,  added description and removed it from `alias, usage, agruments`

File: man/print.data.table.Rd [hunk 5]
```diff
@@ -13,6 +13,8 @@
   Key enhancements include automatic output compression of many observations and concise column-wise \code{class} summary.
 
   \code{format_col} and \code{format_list_item} generics provide flexibility for end-users to define custom printing methods for generic classes.
+
+  \code{trunc.char} This feature compresses strings exceeding the \code{datatable.prettyprint.char} limit. Strings longer than this limit have excess characters removed, with "..." added to indicate truncation. Shorter strings remain unchanged.
```
MichaelChirico, line 17: trunc.char is the name of an internal function; the benefit of that is we can change it on a moment's notice without worrying about breaking anything.

So, let's stick to describing the _behavior_ without mentioning anything about the _implementation_ (e.g. the name of private functions/algorithms used to accomplish this)

man/print.data.table.Rd
```diff
@@ -13,6 +13,8 @@
   Key enhancements include automatic output compression of many observations and concise column-wise \code{class} summary.
 
   \code{format_col} and \code{format_list_item} generics provide flexibility for end-users to define custom printing methods for generic classes.
+
+  Truncate all strings exceeding the \code{datatable.prettyprint.char} limit, every string which exceeds the limit are trucated to limit of characters by removing excess characters.
```
MichaelChirico, line 17:
```suggestion
  Note also the option \code{datatable.prettyprint.char}; character columns entries exceeding this limit will be truncated, with \code{...} indicating the truncation.
```

man/print.data.table.Rd
```diff
@@ -98,6 +100,16 @@
...
 
+  # `char.trunc` will truncate the strings if their lengths exceed the given limit: `datatable.prettyprint.char`
+  # For example:
+
+  options(datatable.prettyprint.char = 5L)
+  DT = data.table(x=1:2, y=c("abcdefghij", "klmnopqrstuv"))
+  DT
+  #     x        y
+  # 1:  1     abcde...
+  # 2:  2     klmno...
```
MichaelChirico, line 111:
```suggestion
  old = options(datatable.prettyprint.char=5L)
  DT = data.table(x=1:2, y=c("abcdefghij", "klmnopqrstuv"))
  DT
  #     x        y
  # 1:  1     abcde...
  # 2:  2     klmno...
  options(old)
```

MichaelChirico (approved):
Thanks!

Nj221102:
> Thanks!

well everything was basically done by you, thanks for your help and guidance :), will try to improve myself in future.

PR #6016, opened 2024-03-20
Updated warning for referencing a non-existent value during creation of new column

### LINKED ISSUE #5648: Unintuitive warning when referencing a value that doesn't exist during creation of new column in J
This is hard to describe in words so here is a simple reprex:

```R
library(data.table)

foo<-data.table(col1=letters[1:10])

foo[,col3:=foo[col1=="a",]$col2]

##Returns the below error:

Warning message:
In `[.data.table`(foo, , `:=`(col3, foo[col1 == "a", ]$col2)) :
  Column 'col3' does not exist to remove
```

Because there is no `col2` at all the value retrieved is `NULL`, which gets interpreted as trying to delete `col3` rather than create it. In isolation this is a pretty obvious "duh" moment but in practice the reason this is going to happen in the first place is because someone's accidentally lost track of what order they've done things in.

I think a minor tweak to the error message would be a big help for anyone who's gotten themselves mixed up like that:

```
In `[.data.table`(foo, , `:=`(col3, foo[col1 == "a", ]$col2)) :
  value passed to `:=` is NULL but column 'col3' does not exist to remove
  If you are trying to create 'col3' check if value exists
```

joshhwuu:
Closes #5648 

This PR changes the old warning message when referencing a value that doesn't exist:

> ```r
> library(data.table)
> 
> foo<-data.table(col1=letters[1:10])
> 
> foo[,col3:=foo[col1=="a",]$col2]
> 
> # Warning message:
> # In `[.data.table`(foo, , `:=`(col3, foo[col1 == "a", ]$col2)) :
> # Column 'col3' does not exist to remove
> ```

The above issue may be easy to spot when isolated but the previous warning message isn't very helpful when someone just accidentally loses track of what order they put things in, especially within a larger context.

The new warning message:

https://github.com/Rdatatable/data.table/blob/ac9943998fb66a618ad0921583fe7f0046825ebc/src/assign.c#L433
```r
# Tried to assign NULL to column 'col3', but this column does not exist to remove.
```
should be clearer and more helpful in debugging.

Also, I updated tests 316, 944.1, and 944.3 to test for this new warning message.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -430,7 +430,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       if (newcolnum<0 || newcolnum>=length(newcolnames))
         error(_("Internal error in assign.c: length(newcolnames)=%d, length(names)=%d, coln=%d"), length(newcolnames), length(names), coln); // # nocov
       if (isNull(thisvalue)) {
-        warning(_("Column '%s' does not exist to remove"),CHAR(STRING_ELT(newcolnames,newcolnum)));
+        warning(_("Tried to assign NULL to column '%s', but column '%s' does not exist to remove"),CHAR(STRING_ELT(newcolnames,newcolnum)),CHAR(STRING_ELT(newcolnames,newcolnum)));
         continue;
       }
       // RHS of assignment to new column is zero length but we'll use its type to create all-NA column of that type
--- a/src/assign.c
+++ b/src/assign.c
@@ -430,7 +430,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       if (newcolnum<0 || newcolnum>=length(newcolnames))
         error(_("Internal error in assign.c: length(newcolnames)=%d, length(names)=%d, coln=%d"), length(newcolnames), length(names), coln); // # nocov
       if (isNull(thisvalue)) {
-        warning(_("Tried to assign NULL to column '%s', but column '%s' does not exist to remove"),CHAR(STRING_ELT(newcolnames,newcolnum)),CHAR(STRING_ELT(newcolnames,newcolnum)));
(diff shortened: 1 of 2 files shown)


src/assign.c
```diff
@@ -430,7 +430,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       if (newcolnum<0 || newcolnum>=length(newcolnames))
         error(_("Internal error in assign.c: length(newcolnames)=%d, length(names)=%d, coln=%d"), length(newcolnames), length(names), coln); // # nocov
       if (isNull(thisvalue)) {
-        warning(_("Column '%s' does not exist to remove"),CHAR(STRING_ELT(newcolnames,newcolnum)));
+        warning(_("Tried to assign NULL to column '%s', but column '%s' does not exist to remove"), CHAR(STRING_ELT(newcolnames,newcolnum)), CHAR(STRING_ELT(newcolnames,newcolnum)));
```
MichaelChirico, line 433: Saying the column name twice feels a bit redundant, what about?

```suggestion
        warning(_("Tried to assign NULL to column '%s', but this column does not exist to remove"), CHAR(STRING_ELT(newcolnames,newcolnum)));
```

FYI, if you really need to re-use an input to a templated message, it's better to do so in the template itself: `...column '%1$s', but column '%1$s' does not...` avoids the need to call `CHAR(...)` twice, and will make it clearer to translators that both entries are the same without needing to refer back to the package source.

MichaelChirico, line 433: Another suggestion

```suggestion
        warning(_("Skipping attempt to assign NULL to column '%s', which does not exist to remove"), CHAR(STRING_ELT(newcolnames,newcolnum)));
```

joshhwuu: I like this change, naming the column twice did feel redundant. Will need to update the tests now.

MichaelChirico:
Great, thanks!

PR #6017, opened 2024-03-20
Add `ignore.attr` Argument to `rbindlist` Function

### LINKED ISSUE #3911: rbindlist could use an "ignore.attributes" argument

Nj221102:
**Description:**

This pull request addresses issue #3911 by adding an `ignore.attributes` argument to the `rbindlist` function. When set to `TRUE`, this argument will enable ignoring strict attributes checking, allowing for more lenient behavior when combining data tables.

**Proposed Changes:**

- Added `ignore_attr` argument to the `rbindlist` function.
- When `ignore_attr` is set to `TRUE`, strict attributes checking is bypassed.
- Updated documentation and tests accordingly.
- Added tests.

**Related Issues:**
- Closes #3911


ben-schwen:
This seems similar addressing the same as #5446

Nj221102:
> This seems similar/addressing the same as #5446

oh no i didn't notice it, sorry let me close this one.

Issue #2687, opened 2018-03-18
documentation on how openMP is used are lacking 


malcook:
The only docs I could find on how openMP is actually used are from "old news" file [Changes in v1\.9\.8 \(on CRAN 25 Nov 2016\)](https://github.com/Rdatatable/data.table/blob/master/NEWS.0.md) which reads:

> Added setDTthreads() and getDTthreads() to control the threads used in data.table functions that are now parallelized with OpenMP on all architectures including Windows (fwrite(), fsort() and subsetting). 

it would be great if this could be in the FAQ with a little info on how (much) this helps what sort of operations.

Or am I missing something?

tdhock:
I read ?setDTthreads with the expectation of finding some indication of how openMP is used for parallelization, so I could get some idea about what kinds of data would result in speedups using more than one thread. The closest I found to an answer was:
```
     Internally parallelized code is used in the following places:
        • ‘between.c’ - between()
        • ‘cj.c’ - CJ()
        • ‘coalesce.c’ - fcoalesce()
        • ‘fifelse.c’ - fifelse()
        • ‘fread.c’ - fread()
        • ‘forder.c’, ‘fsort.c’, and ‘reorder.c’ - forder() and related
        • ‘froll.c’, ‘frolladaptive.c’, and ‘frollR.c’ - froll() and family
        • ‘fwrite.c’ - fwrite()
        • ‘gsumm.c’ - GForce in various places, see GForce
        • ‘nafill.c’ - nafill()
        • ‘subset.c’ - Used in ‘[.data.table’ subsetting
        • ‘types.c’ - Internal testing usage
```
There are links to these functions, so I expected to find more details in the linked man pages, but I did not find a sufficiently detailed description of how openMP is used / what kinds of data/operations would result in speedups when using multiple threads.
For example on ?fread the only mention of threads is:
```
 nThread: The number of threads to use. Experiment to see what works
          best for your data on your hardware.
```
Can someone please add details about how multi-threading is used, and when speedups should be expected?
for example, something like the following:
```
 nThread: The number of threads to use in the for loop over columns. 
          (THIS IS JUST AN EXAMPLE, I DO NOT KNOW IF THIS IS TRUE)
          Speedups should be expected when there are a large number of columns. 
          Experiment to see what works best for your data on your hardware.
```
The linked blog post has some benchmarking of fread and fwrite (computation times for some particular numbers of rows and columns), but it would be useful to have some description like this on the man pages.

[...]

PR #6018, opened 2024-03-23
Documented the use of OpenMP and parallelization in C code

Anirban166:
I did some research and tried to understand the use of OpenMP in the C files after filtering for [#pragma omp](https://github.com/search?q=repo%3ARdatatable%2Fdata.table+%23pragma&type=code) - From what I saw and understood, I wrote a bit for each of the twelve use cases mentioned to be using parallelization that you see when using `?setDTthreads`, or namely for these:
```
‘between.c’ - between()
‘cj.c’ - CJ()
‘coalesce.c’ - fcoalesce()
‘fifelse.c’ - fifelse()
‘fread.c’ - fread()
‘forder.c’, ‘fsort.c’, and ‘reorder.c’ - forder() and related
‘froll.c’, ‘frolladaptive.c’, and ‘frollR.c’ - froll() and family
‘fwrite.c’ - fwrite()
‘gsumm.c’ - GForce in various places, see GForce
‘nafill.c’ - nafill()
‘subset.c’ - Used in ‘[.data.table’ subsetting
‘types.c’ - Internal testing usage
```
I used a more straightforward 'Parallelism is used here to ...' and 'OpenMP is used here to ...' to start with before enlisting my key observations, but I tried to avoid repetition elsewhere whenever possible. I also included a small section on some factors that can contribute to improved speedup below those cases. Let me know if this is good enough, or if what I wrote aligns with the intended use!

This is in hopes of helping with or potentially closing #2687 which shed some light on having inadequate documentation for the same.

I made modifications to the `openmp-utils` manual, and for reproducing them (to test formatting or checking things in the man-page style), one can use `rstudioapi::previewRd('path-to-your-data.table-installation/man/openmp-utils.Rd')` to view inside RStudio after integrating my changes.

<details>
  <summary>(or click this in case you want to see how it looks here itself :)</summary>
  How my changes look like in the man page
</details>


jangorecki:
Regarding openmp utils manual. I would include the date there, something like "as of March 2024". It is because it doc will be getting outdated over time.

Anirban166:
At which spot in the file would it be ideal to place the date?

I think all documentation could use that, or a 'last updated' time point (considering that everything is subject to change in the future).

MichaelChirico:
I agree with Jan. This is a welcome change but very prone to becoming outdated.

> I think all documentation could use that, or a 'last updated' time point (considering that everything is subject to change in the future).

Not really. In particular, this is essentially documenting implementation details, not really user-facing behavior, which makes it much more likely to escape the normal PR process and wind up out of date, whereas other documentation will be updated by any relevant PRs.

Honestly, I would remove all of the "Parallelism is used here to..." sections. Interested users should be reading the source to learn these things, and such granular detail of implementation detail is _especially_ prone to become outdated (every time a loop is added/edited we're meant to update this documentation?). We could also devise some regression testing (a new GHA script) to at least ensure the list of routines where OpenMP is used is reflected in the cited list.

With those reservations in mind I will re-assign to @tdhock who gave the most extensive comments on the linked issue #2687 and I think has a clearer idea in mind of what he wants the documentation to look like.

Anirban166:
> Not really. In particular, this is essentially documenting implementation details, not really user-facing behavior, which makes it much more likely to escape the normal PR process and wind up out of date, whereas other documentation will be updated by any relevant PRs.

Thanks for explaining! That makes sense, and I see the distinction now.

> Interested users should be reading the source to learn these things, and such granular detail of implementation detail is _especially_ prone to become outdated (every time a loop is added/edited we're meant to update this documentation?). 

I agree as in essence, the regular user would only be concerned with the usage and notes on how to extract more benefit, and not these implementation-specific details (even when keeping this updated *could be* possible).

> We could also devise some regression testing (a new GHA script) to at least ensure the list of routines where OpenMP is used is reflected in the cited list.

This sounds like a good idea, but I'm not sure how that will pertain to testing regression standalone. Do you mean to update this list/manual by a grep for the functions where OpenMP is used for the GHA that does regression testing? (On a related note, I will share it this week in a new PR, just waiting for some feedback from Toby)

> With those reservations in mind I will re-assign to @tdhock who gave the most extensive comments on the linked issue #2687 and I think has a clearer idea in mind of what he wants the documentation to look like.

Sure, and Toby mentioned this as something to work on earlier so it would certainly be appropriate.

MichaelChirico:
> This sounds like a good idea, but I'm not sure how that will pertain to testing regression standalone. Do you mean to update this list/manual by a grep for the functions where OpenMP is used for the GHA that does regression testing? (On a related note, I will share it this week in a new PR, just waiting for some feedback from Toby)

In {lintr}, we have this repo-meta-tests action that does some consistency checks on the repo/codebase as a whole that aren't ideally suited to the package test suite:

https://github.com/r-lib/lintr/blob/main/.github/workflows/repo-meta-tests.yaml

Anirban166:
> In {lintr}, we have this repo-meta-tests action that does some consistency checks on the repo/codebase as a whole that aren't ideally suited to the package test suite:
https://github.com/r-lib/lintr/blob/main/.github/workflows/repo-meta-tests.yaml

Thanks for sharing - I went ahead and wrote a [first draft](https://github.com/Anirban166/Autocomment-atime-results/issues/31) for a step that updates such a list with filtered filenames (getting the names of routines or functions would be slightly trickier, but doable if you want me to proceed with this - let me know!). 

It makes sense to check such as part of the metadata as a consistency check, but I'm still not sure how it would be directly helpful for regression testing or what you mentioned earlier:

> We could also devise some regression testing (a new GHA script) to at least ensure the list of routines where OpenMP is used is reflected in the cited list.

tdhock:
From a user perspective I would like the docs to tell us if we get parallel speedups with large nrow or ncol. I guess the answer is usually nrow, but sometimes I observe speedups with large ncol, for example see below https://rcdata.nau.edu/genomic-ml/atime/vignettes/compare-data.table-tidyverse.html with nrow=10 and ncol varying. 
Untitled
Would be nice to have some explanation about why/when we would expect speedups with large ncol/ncol.

MichaelChirico:
> From a user perspective I would like the docs to tell us if we get parallel speedups with large nrow or ncol.

That seems perfectly reasonable to me, and is also something that's not likely to change over time.

Anirban166:
> From a user perspective I would like the docs to tell us if we get parallel speedups with large nrow or ncol.

> Would be nice to have some explanation about why/when we would expect speedups with large ncol/ncol.

I ran a [benchmark](https://github.com/Anirban166/Autocomment-atime-results/issues/32) and explored a bit to provide more information on that - Can you please check and review everything I wrote? (especially to check if I clarified enough in terms of the user-facing details, or more specifically what the user could take note of or try for better speedup gains)

Anirban166:
> From a user perspective I would like the docs to tell us if we get parallel speedups with large nrow or ncol.

> Would be nice to have some explanation about why/when we would expect speedups with large ncol/ncol.

I ran a [benchmark](https://github.com/Anirban166/Autocomment-atime-results/issues/32) and explored a bit to provide more information on that - Can you please review everything I wrote in this PR? (especially to check if I clarified enough in terms of the user-facing details, or more specifically what the user could take note of or try for better speedup gains)

tdhock:
@Nj221102 can you please review?

Anirban166:
@ben-schwen If possible, would you like to review this Ben?

Nj221102:
> @Nj221102 can you please review?

oh i am so sorry, i just saw this now, don't know how i missed the notification of the ping :(, 

yeah i will give it a try but i have limited knowledge in openmp stuff, so getting reviewed by @ben-schwen as well sounds great.

Nj221102 (approved):
Everything looks good to me :)

man/openmp-utils.Rd
```diff
@@ -38,20 +38,121 @@
...
+      \item The conditional checks within parallelized loops
+      }
+    
+    Significant speedup can be expected for more number of columns here, given that this function operates efficiently across multiple columns to find non-NA values.
+    
     \item\file{fifelse.c} - \code{\link{fifelse}()}
+    
+    For logical, integer, and real types, OpenMP is being used here to parallelize loops that perform conditional checks along with assignment operations over the elements of the supplied logical vector based on the condition (\code{test}) and values provided for the remaining arguments (\code{yes}, \code{no}, and \code{na}).
+    
+    Better speedup can be expected for more number of columns here as well, given that this function operates column-wise with independent vector operations.
```
ben-schwen, line 77: I don't understand what is meant by more number of columns

man/openmp-utils.Rd
```diff
@@ -38,20 +38,121 @@
...
+    
     \item\file{fread.c} - \code{\link{fread}()}
+    
+    OpenMP is used here to:
+
+    \itemize{
+      \item Avoid race conditions or concurrent writes to the output \code{data.table} by having atomic operations on the string data
```
ben-schwen, line 84: I would also mention that we use openmp here for the chunked reading of the file

man/openmp-utils.Rd
```diff
@@ -38,20 +38,121 @@
...
+    
+    Similar to \code{\link{fread}()}, this function is highly efficient in parallely processing data with large numbers of both rows and columns, but it has more notable speedups with an increased number of rows.
+    
     \item\file{gsumm.c} - GForce in various places, see \link{GForce}
+    
+    Functions with GForce optimization are internally parallelized to speed up grouped summaries over a large \code{data.table}. OpenMP is used here to parallelize operations involved in calculating group-wise statistics like sum, mean, and median (implying faster computation of \code{sd}, \code{var}, and \code{prod} as well). 
+    
+    These optimized grouping operations benefit more in terms of speedup if the input data contains a large number of rows (since they are often used to aggregate data across groups).
```
ben-schwen, line 120: The core advantage of using GForce comes not with many rows but with many groups.

Just imagine this:
You have a lot of groups. Then `data.table` has to evaluate this grouped operation for each of them, hence, it needs a lot of calls and a lot of new environments where each group is evaluated. 

With GForce it just iterate over all groups in one go, while keeping track what the current status for each group is. The groups are calculated with openmp and the scatter&gather pattern. For some operations (sum and mean) it also uses the gather&scatter pattern for the operation itself.

ben-schwen:
I would also mention our throttling here

man/openmp-utils.Rd
Anirban166: Absolutely! Not sure why I initially thought that the parallel region was only dealing with the progress updates, but it makes complete sense now that I re-read the code. (For starters, I should have clearly noticed the [directive](https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/src/fread.c#L2222) besides the [loop](https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/src/fread.c#L2266) before at least)

man/openmp-utils.Rd
Anirban166: I meant that relative to columns (or that the speedup is better for more number of columns as opposed to having more number of rows in the input data) but you're right that it isn't obvious without the mention of the rows vs. columns thing. I also don't like how I'm repeating myself on that note for every function..do you think it would better to have a statement towards the end of the document enlisting all the functions that perform better with more number of rows vs the ones that perform better with more number of columns?

Anirban166: I meant that relative to columns (or that the speedup is better for more number of columns as opposed to having more number of rows in the input data) but you're right that it isn't obvious without the mention of the rows vs. columns thing. I also don't like how I'm repeating myself on that note for every function..do you think it would be more preferable to have a statement towards the end of the document enlisting all the functions that perform better with more number of rows vs the ones that perform better with more number of columns?

man/openmp-utils.Rd
Anirban166: Noted. I did realize earlier that we need to have more number of groups to leverage GForce due to for a single pass on evaluating that grouped operation as opposed to individual evaluations but I was too hard set on the rows vs columns thing at the time of writing. I appreciate the detailed explanation! (Gave me more insight, especially about the pattern being used)

Anirban166: Noted. I did realize earlier that we need to have more number of groups to leverage GForce for a single pass on evaluating that grouped operation as opposed to individual evaluations but I was too hard set on the rows vs columns thing at the time of writing. I appreciate the detailed explanation though! (Gave me more insight, especially about the pattern being used)

man/openmp-utils.Rd
```diff
@@ -149,6 +149,8 @@ Apart from increasing the size of the input data, function-specific parameters w
       \item Having more and/or complex conditional logic when using \code{\link{fifelse}()} or \code{\link{subset}()}
       }
 
+\code{data.table} also incorporates throttling mechanisms to prevent the excessive creation of threads (can lead to significant overhead) for small data tasks.
```
tdhock, line 152: I don't think this is specific enough to be useful. we already have
```
  \item{throttle}{ 1024 (default) means that, roughly speaking, a single thread will be used when nrow(DT)<=1024, 2 threads when nrow(DT)<=2048, etc. The throttle is to speed up small data tasks (especially when repeated many times) by not incurring the overhead of managing multiple threads. Hence the number of threads is throttled (restricted) for small tasks. }
```

Anirban166: True

Anirban166:
@MichaelChirico Also, could you review this? (once more now, but for either merging/closing - I'm cool either way!)

MichaelChirico:
I'm still a bit torn on how to proceed. Obviously this is nice research and it'd be a shame to waste it. OTOH I don't think most of this info should be recorded/shared with users -- it's really developer-facing utility. I think user only cares when a function is taking 2+-dimensional input, which dimension is parallelized? (most often rows vs. columns).

So the question becomes where's most appropriate for this to live. Some options I can think of:

 - Put it in a Wiki. This is still sort of user-facing.
 - Put it in a document like CRAN_release in .dev folder
 - Put it in the comments above the corresponding routine in the src/ code

Right now I think the 3rd option is best -- the comments then live closest to where they're useful. Any other thoughts?

tdhock:
> user only cares when a function is taking 2+-dimensional input, which dimension is parallelized? (most often rows vs. columns).

yes please I would like to see something on man pages in this case. (fread and fwrite)

for other cases, comments are ok.

jangorecki:
Afair non exported functions can have manual as well. Then we can keep that format and just have it hidden from the ordinary user.

Anirban166:
> I'm still a bit torn on how to proceed. Obviously this is nice research and it'd be a shame to waste it. OTOH I don't think most of this info should be recorded/shared with users -- it's really developer-facing utility. I think user only cares when a function is taking 2+-dimensional input, which dimension is parallelized? (most often rows vs. columns).
> 
> So the question becomes where's most appropriate for this to live. Some options I can think of:
> 
> * Put it in a Wiki. This is still sort of user-facing.
> * Put it in a document like CRAN_release in .dev folder
> * Put it in the comments above the corresponding routine in the src/ code
> 
> Right now I think the 3rd option is best -- the comments then live closest to where they're useful. Any other thoughts?

I agree, and I like options 1 and 3 (with more preference towards the latter as well)

Putting it in the Wiki would be easier to maintain as anybody can edit (and I agree that such implementation details should not be user-facing in general, but some users might be keen to know more) and is not inappropriate as a place to read some extra details, but having that information organized in their respective separate files near their code sections offers convenience, especially in the reader not having to skim through all the other functions altogether (which I think is sweet for helping save time). Thanks for making the changes!

man/openmp-utils.Rd
```diff
@@ -41,17 +41,19 @@
...
+    \item\file{fwrite.c} - \code{\link{fwrite}(). Parallelized across rows.}
+    \item\file{gsumm.c} - GForce in various places, see \link{GForce}. Parallelized across groups.
     \item\file{nafill.c} - \code{\link{nafill}()}
     \item\file{subset.c} - Used in \code{\link[=data.table]{[.data.table}} subsetting
     \item\file{types.c} - Internal testing usage
   }
+
+  We endeavor to keep this list up to date, but note that the canonical reference here is the source code itself.
 }
+
```
tdhock, line 56: please undo addition of empty lines

Anirban166:
One thing that was not there in that PR was the section for speedup gains I wrote:
```Rd
In general, or as applicable to all the aforementioned use cases, better speedup can be expected when dealing with large datasets.

Having such data when using \code{\link{fread}()} or \code{\link{fwrite}()} (ones with significant speedups for larger file sizes) also means that while one part of the data is being read from or written to disk (I/O operations), another part can be simultaneously processed using multiple cores (parallel computations). This overlap reduces the total time taken for the read or write operation (as the system can perform computations during otherwise idle I/O time).

Apart from increasing the size of the input data, function-specific parameters when considered can benefit more from parallelization or lead to an increase in speedup. For instance, these can be:

    \itemize{
      \item Having a large number of groups when using \code{\link{forder}()} or a multitude of combinations when using \code{\link{CJ}()}
      \item Having several missing values in your data when using \code{\link{fcoalesce}()} or \code{\link{nafill}()}
      \item Using larger window sizes and/or time series data when using \code{\link{froll}()}
      \item Having more and/or complex conditional logic when using \code{\link{fifelse}()} or \code{\link{subset}()}
      }
```
And then the respective lines I had for each routine that advise on using more rows vs columns for speedup. I'll proceed to add that piece of information for fread and fwrite in their respective man files as Toby suggested above, but do you guys think rest of that information (or for the other routines) is worth retaining?

Anirban166:
> yes please I would like to see something on man pages in this case. (fread and fwrite)

I'm assuming you still want them in their man pages right? or would what Michael did here suffice?

MichaelChirico:
> do you guys think rest of that information (or for the other routines) is worth retaining?

I wasn't sure there's a lot of information there besides what should be clear from reading the source code, which is easier now that the comment appears alongside the implementation (in the man page it was more useful as there's more separation between the comment & the logic). But if you feel strongly feel free to add back into the comments.

Anirban166:
> in the man page it was more useful as there's more separation between the comment & the logic

Same thoughts (especially at the time of writing)

> But if you feel strongly feel free to add back into the comments.

Not at all, I was just curious to know if they were required or not (since aside from the use of OpenMP, Toby wanted to have information for the user on whether stuff like more rows vs. columns in the input data would be better in terms of speedup, for each routine)

MichaelChirico:
> I'm assuming you still want them in their man pages right? or would what Michael did here suffice?

That's a good point. Yes, let's make the mention in the corresponding man page directly, that will be most useful to users.

man/openmp-utils.Rd
```diff
@@ -41,17 +41,18 @@
...
     \item\file{coalesce.c} - \code{\link{fcoalesce}()}
     \item\file{fifelse.c} - \code{\link{fifelse}()}
-    \item\file{fread.c} - \code{\link{fread}()}
+    \item\file{fread.c} - \code{\link{fread}(). Parallelized across row-based chunks of the file.}
     \item\file{forder.c}, \file{fsort.c}, and \file{reorder.c} - \code{\link{forder}()} and related
     \item\file{froll.c}, \file{frolladaptive.c}, and \file{frollR.c} - \code{\link{froll}()} and family
-    \item\file{fwrite.c} - \code{\link{fwrite}()}
-    \item\file{gsumm.c} - GForce in various places, see \link{GForce}
+    \item\file{fwrite.c} - \code{\link{fwrite}(). Parallelized across rows.}
+    \item\file{gsumm.c} - GForce in various places, see \link{GForce}. Parallelized across groups.
```
tdhock, line 48: for me this is sufficient documentation of whether the parallelization happens on rows or columns

tdhock (approved):

MichaelChirico:
Thanks all!

man/openmp-utils.Rd
Anirban166: I agree! (for me too that's enough)

> That's a good point. Yes, let's make the mention in the corresponding man page directly, that will be most useful to users.

Appreciate you for considering this route!
Since I based that upon what Toby wanted here:

> yes please I would like to see something on man pages in this case. (fread and fwrite)

And now that he agrees, I think we can move ahead on this consensus then :)

Issue #5971, opened 2024-03-01
1.15.4 release


MichaelChirico:
1.15.2 release is now on CRAN, but we didn't get a fix for #5962 included in time. We should aim to release another patch with this regression fix in maybe another month or so, including any further issues that arise in the meantime.

Filing this issue as an earmark to that end.

The patch release branch is https://github.com/Rdatatable/data.table/tree/patch-1.15.4

TysonStanley:
Works for me.

MichaelChirico:
I think this patch is urgent enough to start getting a release ready next week if you'll have the time @TysonStanley -- the "`shift` in subsets issue" is bad for those affected by it, e.g. https://stackoverflow.com/a/78149204/3576984. Nothing else came up from the CRAN checks so I think the patch is as ready as it will be, anyway.

TysonStanley:
@MichaelChirico Will plan on it!

MichaelChirico:
We're also erroring on CRAN as of about 12 hours ago as the language comparison change is now in R-devel:

https://github.com/r-devel/r-svn/commit/f111326ae1a1200b76e4c74f280fac2673117b6f

https://cran.r-project.org/web/checks/check_results_data.table.html

TysonStanley:
CRAN response

>  Found the following (possibly) invalid URLs: URL: https://csvy.org/ From: man/fread.Rd

I just tried the site and it looks dead at the moment.

TysonStanley:
I haven't been able to find its new home (if there is one). Do we have a recommended approach for dead links like this one?

TysonStanley:
Short-term fix to remove the issue with the dead link in d88fd42 to get patch fixed for CRAN

MichaelChirico:
> I haven't been able to find its new home (if there is one). Do we have a recommended approach for dead links like this one?

checking with the csvy GitHub owner:

https://fosstodon.org/@michaelchirico/112131985274138229

we could also link the associated GitHub repo.

as for general approach, I'd say (1) try and find the same page relocated (2) try and find an equivalent page (3) delete it, possibly temporarily with a TODO if we've reason to expect the page to resurrect later.

TysonStanley:
Thanks, appreciate the feedback. And thanks for reaching out to the owner. I can wait a day or two to see if we get an answer (our patch isn't due until April 1).

TysonStanley:
@MichaelChirico any word on csvy?

MichaelChirico:
[...]

PR #6021, opened 2024-03-24
Use early exit to get auto-print output for 'main' branch

MichaelChirico:
Noticed in https://github.com/Rdatatable/data.table/issues/5971#issuecomment-2016987766 that we appear to be missing some output because the failure happens inside `if()...` :arrow_right: auto-print doesn't do anything.

We should prefer auto-print to happen in the "main" execution path for these scripts (i.e. the one that runs on CRAN by default), so here I re-arranged the branches to get this to happen.


tests/knitr.R
```diff
@@ -1,9 +1,9 @@
...
-    cat(readLines("knitr.Rout.mock", warn = FALSE), sep="\n")
+if (!suppressPackageStartupMessages(requireNamespace("knitr", quietly=TRUE))) {
+  cat(readLines("knitr.Rout.mock", warn=FALSE), sep="\n")
+  q('no')
 }
 
+require(knitr)
+knit("knitr.Rmd", quiet=TRUE)
+cat(readLines("knitr.md"), sep="\n")
+invisible(file.remove("knitr.md"))
```
jangorecki, line 9: https://github.com/Rdatatable/data.table/blob/master/tests/knitr.Rout.mock
no need to update here?

MichaelChirico: Good point about updating the goldens. I'm not sure how to generate the `.mock` correctly, working on it.

jangorecki (approved):

PR #6029, opened 2024-03-28
Make tests robust to exact phrasing & language of base condition messages

### LINKED ISSUE #4789: review and relax test strings on R messages
As per https://github.com/Rdatatable/data.table/pull/4770#issuecomment-719825102

MichaelChirico:
Closes #4789


tdhock (approved):
this seems reasonable. not sure how often base R would make a change to its messages?
at least this refactoring makes it clear in our test file which error messages are from us, and which are from base R.

MichaelChirico:
> not sure how often base R would make a change to its messages?

not _terribly_ often, but it's no fun to get kicked off CRAN for it when it does happen, as I just dealt with in a patch release for {lintr}.

it's also a better testing practice in principle: only directly test the code we own, though as seen it's a bit more tedious to design the tests & slightly cumbersome to avoid regression. on balance I think avoiding CRAN issues is worth these overheads which can themselves be mitigated.

PR #6038, opened 2024-03-30
Mention the optimized descending sort operation case when using order()

### LINKED ISSUE #1920: Incomplete dox in `?setorder`
In `?setorder`::**Usage**, we have:

https://github.com/Rdatatable/data.table/blob/566bff0fe1a10d94a494026c59eb611b90b4dc04/man/setorder.Rd#L32-L33

But `x[order(., decreasing = TRUE)]` is also optimized. Using the example from `?setorder`:

```
set.seed(45L)
DT = data.table(A=sample(3, 10, TRUE), 
         B=sample(letters[1:3], 10, TRUE), C=sample(10))
DT[order(C, decreasing = TRUE), verbose = TRUE]
# order optimisation is on, i changed from 'order(...)' to 'forder(DT, ...)'.
#     A B  C
#  1: 1 b 10
#  2: 2 a  9
#  3: 2 b  8
#  4: 1 a  7
#  5: 2 b  6
#  6: 1 a  5
#  7: 1 c  4
#  8: 1 c  3
#  9: 2 b  2
# 10: 1 b  1
```

So I guess the documentation should mention this.

Anirban166:
Closes #1920 

(Thanks to @MichaelChirico for pointing this out and clarifying the changes to be made!)

### DIFF
--- a/man/setorder.Rd
+++ b/man/setorder.Rd
@@ -31,6 +31,7 @@ setorder(x, \dots, na.last=FALSE)
 setorderv(x, cols = colnames(x), order=1L, na.last=FALSE)
 # optimised to use data.table's internal fast order
 # x[order(., na.last=TRUE)]
+# x[order(., decreasing=TRUE)]
 }
 \arguments{
 \item{x}{ A \code{data.table}. }


MichaelChirico (approved):
Thanks!

PR #6040, opened 2024-04-01
Allow running a subset of tests by pattern

### LINKED ISSUE #6039: test() should have an option to start at certain test
Using `cc()` and `test.data.table()` works like a charm for rapid developing. However, when running `test.data.table()` I often know at which particular test I want to start testing (since this was the test failing before). This might become even more important since our testcase database is growing (or we might come up with a better partitioning into different test files).

MichaelChirico:
Closes #6039.

---

The simple version 0b683fb54494b49a12d6e91921c369add70ec05a doesn't work right because of our non-hermetic tests. A version with just `try({x,y}, silent=TRUE)` worked but spewed a ton of output. In any case, it still takes ~30 seconds to run the suite on my computer for `testPattern='1111'` -- again because our tests are so non-hermetic and we wind up needing to evaluate the `x`&`y` parts for every test because of this.

Why is this the case? think of tests like:

```r
test(1, x, y <- 2)
test(2, z, y)

a <- foo(y)
```

We need to evaluate `y<-2` in order for `test(2)` to pass, and moreover we need `y` for `a` outside of tests so even early return of `test()` doesn't help.


MichaelChirico:
@ben-schwen PTAL if this suits your needs

MichaelChirico:
> > @ben-schwen PTAL if this suits your needs
> 
> Basically yes. However, I just noticed that it has problems when the test uses `fread`.

Could you elaborate? I just tried `test.data.table(testPattern="1552|2251")` and it worked fine. I suspect it's probably related to my [comment](https://github.com/Rdatatable/data.table/pull/6040/files#diff-45a72648454af4a94eaa1072baecf2e50098cbeb3c317794e07fb8bd487c920fR176)...

ben-schwen:
> Could you elaborate? I just tried `test.data.table(testPattern="1552|2251")` and it worked fine. I suspect it's probably related to my [comment](https://github.com/Rdatatable/data.table/pull/6040/files#diff-45a72648454af4a94eaa1072baecf2e50098cbeb3c317794e07fb8bd487c920fR176)...

I tried `test.data.table(testPattern='2025')` which ends up with
```r
Running 3 of 6832 tests matching '2025'
Error in parse(n = -1, file = file, srcfile = NULL, keep.source = FALSE) : 
  184:1: unexpected '}'
183: test(2025.01, fread(testDir("issue_3400_fread.txt"), skip=1, header=TRUE), data.table(A=INT(1,3,4), B=INT(2,2,5), C=INT(3,1,6)))
184: }
     ^
```
but maybe thats because these tests use the test_no + 0.1 instead of fixed numbers?

ben-schwen:
What if we simply add a global option for the testPattern and add this option to the signature of the test function. Then we could set this global parameter with `test.data.table`?

MichaelChirico:
> Maybe another more simple approach: What if we add a global option for the testPattern and add this option to the signature of the test function. Then we could set this global parameter with `test.data.table`.

that was done in https://github.com/Rdatatable/data.table/commit/0b683fb54494b49a12d6e91921c369add70ec05a as elaborated in PR description. it doesn't work well with our test setup :\

MichaelChirico:
> I tried `test.data.table(testPattern='2025')` which ends up with

the problem with 2025 is here:

https://github.com/Rdatatable/data.table/blob/d35dcebc9ea022a0a80f55e0abbcc84f4adbcffc/inst/tests/tests.Rraw#L14712-L14733

Some but not all of the suite have literal '2025' in the name, we wind up with that dangling `}` as part of the "setup code".

We can fix this by making more changes in #6041, but the shortcomings of the static approach remain.

MichaelChirico:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6040?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6040** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6040?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6041** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6041?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

MichaelChirico:
@ben-schwen restacked this onto #6041, with the latest change there `test.data.table(testPattern='2025')` now works.

The fundamental issue remains; as a workaround, I added a step to at least check the subsetted script parses, if not we back up to running the full suite instead of the cryptic error.

MichaelChirico:
This doesn't affect CI or `R CMD check`, so taking :+1: from @ben-schwen as GTG. Let's iterate more to improve this feature as needed later.

PR #6043, opened 2024-04-01
Note that R files are about to be loaded in cc()

MichaelChirico:
Slightly better than just listing out all the R file names


ben-schwen:
LGTM! TY!

PR #6046, opened 2024-04-03
add seal of approval and nc to README

### LINKED ISSUE #5723: Building a data.table community of packages with “Seal of Approval”
With the goal of building a community of packages that have similar philosophies and syntax that are separate from data.table (and outside of data.table scope #5722), we would like to set up a “Seal of Approval” (play on the mascots of `data.table`) process. The process for a package receiving the Seal of Approval could be: 

* The package developer submits a standard application, indicating that their package is on CRAN, is actively maintained, and has at least one of the features listed below.
* A small number of volunteers from the data.table community review the application to ensure that the criteria are met.  (This includes no testing or vetting of the package functionality itself; only checking that it's an appropriate choice for the Seal.)
* There is some variety of public community vote to approve the package. 

Approval will include being listed as a Seal of Approval package on the data.table repository and an SVG of the “seal” that they can include on their own repository/package logo. The initial idea would be packages that do at least one of the following:

* Built on the same principles as data.table (#5693)
* Extends the functionality of data.table to other contexts (e.g., to databases) with similar syntax
* Uses data.table on the backend of the package

Possible examples of this could include packages that have few dependencies (e.g. `tinytest`), extend functionality (e.g. `dtplyr`, `tidytable`, `tidyfast`), and packages that use data.table on the backend (e.g. `modelsummary`).
 
This process would hopefully help other developers feel more connected to data.table and be more likely to want to support it. Things for us to decide on are:

1. Does this idea resonate with the data.table community?
3. If so, does the list of criteria make sense for the purposes of the Seal of Approval?
5. What else should be considered in designing the Seal of Approval?

tdhock:
Closes #5723 

This is a draft implementation of the Seal of Approval idea.
Preview at https://github.com/Rdatatable/data.table/blob/seal-of-approval/README.md#seal-of-approval

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ The Seal of Approval is a curated list of packages that use `data.table` interna
 
 * [nc](https://github.com/tdhock/nc) is for parsing text data using named capture regular expressions, returning the results as a `data.table` with one row per match, and one column per capture group.
 
-To obtain the Seal of Approval, please (1) open a PR that adds your package to the list above, and (2) open a PR that adds a blog post about your package to [The-Raft](https://github.com/rdatatable-community/The-Raft/).
+To obtain the Seal of Approval, please (1) [open a PR in this repo](https://github.com/Rdatatable/data.table/edit/master/README.md) that adds your package to the list above, and (2) open a PR in [The-Raft](https://github.com/rdatatable-community/The-Raft/) repo that adds a blog post about your package.
 
 ### Stay up-to-date
 
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ DT[Petal.Width > 1.0, mean(Petal.Length), by = Species]
 
 ### Seal of Approval
 
-Below is a curated list of packages that use `data.table` internally, and/or provide complementary features that are outside of `data.table` scope. To obtain the Seal of Approval, please (1) open a PR that adds your package to the list below, and (2) open a PR that adds a blog post about your package to [The-Raft](https://github.com/rdatatable-community/The-Raft/).
+The Seal of Approval is a curated list of packages that use `data.table` internally, and/or provide complementary features that are outside of `data.table` scope. To obtain the Seal of Approval, please (1) open a PR that adds your package to the list below, and (2) open a PR that adds a blog post about your package to [The-Raft](https://github.com/rdatatable-community/The-Raft/#category=seal of approval).
 
 * [nc](https://github.com/tdhock/nc) is for parsing text data using named capture regular expressions, returning the results as a `data.table` with one row per match, and one column per capture group.
 
--- a/README.md
+++ b/README.md
@@ -87,14 +87,9 @@ DT[Petal.Width > 1.0, mean(Petal.Length), by = Species]
 
 ### Seal of Approval
 
-Below is a curated list of packages that use `data.table` internally,
-and/or provide complementary features that are outside of `data.table`
-scope.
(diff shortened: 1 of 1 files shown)


jangorecki:
I don't think that readme file is good place for such list. This is readme about data.table, not other pkgs

MichaelChirico:
> I don't think that readme file is good place for such list. This is readme about data.table, not other pkgs

Hmm, I didn't have any issue. It is under the "Community" heading after all. But on second thought, if there would be dozens of packages with the seal it would quickly get distracting. So maybe best is to link to a dedicated page instead.

OTOH, a plaintext "check out the seals of approval+link" will likely be too hard to find & thus undermine the whole point. Possibly we could give it prominence on README in another way, e.g. once there's a logo, add it next to the data.table hex? Or put the hex next to the link?

ben-schwen:
What if we make it as a vignette and link the vignette in the readme?
Vignettes get shipped with the package and have a prominent place on our homepage as well as on the CRAN page. The only other option which comes to my mind is to maintain an additional package but this seems a bit overkill for referring to other packages (looking at you tidyverse).

TysonStanley:
I vote for @ben-schwen 's idea, would still give prominence without having too much space in the README.

jangorecki:
For me vignette also does not seem to fit. Wiki page probably the most...

tdhock:
@jangorecki thanks for sharing your opinion, but can you please clarify why you think the vignette would not be a good fit? 
I believe it should be a good fit, because it is highly visible / easy for users to find/access (for example on CRAN), and edits would have to be reviewed in a PR.
I believe the wiki is not a good fit, because it is less visible, and anyone can edit it (no PR necessary).
Could you please reconsider? or suggest an alternative that satisfies reviewability and visibility?

jangorecki:
Vignettes are about package usage, not revdeps usage nor even listing "notable" revdeps. IMO this fits into new meta package, or if ewe want to have it in data.table repo, then wiki page.

MichaelChirico:
I agree vignette is probably not appropriate, vignette as I see it should be teaching user something, seal of approval is a bit of a stretch of that purpose.

> this fits into new meta package

Maybe there's some misunderstanding here, I don't see how a meta package would be relevant here.

My understanding is the purpose of the 'seal of approval' initiative is simply to highlight some downstream packages that align with data.table by (1) importing data.table (clearly) and (2) generally following some key data.table package design principals like minimal deps, high performance, etc.

In order for this initiative to be useful (as I see it) there needs to be some incentive to participate, mainly through recognition & highlighting by data.table itself. I only [see meaningful traffic](https://github.com/Rdatatable/data.table/graphs/traffic) to the Installation wiki page, which is almost certainly being linked directly from search engines. So relegating this project _only_ to the wiki pages will not be successful, IMO.

The README, by contrast, is visited often; reaching there could provide a meaningful boost to visibility for "sealed" packages.

> This is readme about data.table, not other pkgs

I don't see "seal of approval" as being "not about data.table", in fact, it's highlighting _other problems that data.table can help solve_. The `Community` section of the Wiki already is in the same direction.

In sum, as I see it, realistically this has to be part of README. The decision then is (1) Just don't pursue seal of approval or (2) Decide exactly to place seal of approval in README in a way that balances (a) focus of README on data.table without distracting with references to other packages and (b) seal purpose of highlighting specific noteworthy downstreams.

tdhock:
closing in favor of Kelly's proposal here https://github.com/Rdatatable/data.table/issues/5723#issuecomment-2099409950 
First we will work on the first blog posts on The Raft
Then we will make a new PR

PR #6047, opened 2024-04-04
Deprecate key="a,b" and by="a,b"

### LINKED ISSUE #4357: deprecate multiple columns to be defined as scalar character
For now I identified two cases, there are probably some more:

- `key` argument in `data.table` should not accept `key="A,B"`, which implies that `setkey` should not accept it as well
- `by` argument in `[.data.table` should not accept `by="A,B"` (so `keyby`)

In both cases `c("A","B")` should be used instead.
I understand that it has been added as a nice feature that makes life (very little bit) simpler, but it breaks consistency, introduces new challenges (#4292), and it is also confusing to users (like _does the column has a comma in its name?_).
WDUT @mattdowle

MichaelChirico:
Closes #4357 

@tdhock is it possible to kick off revdeps from a branch? That way we don't need to have this in `master` which may break anyone depending on the dev version & this behavior. Obviously depending on our dev version comes with such risks, but if we can avoid it all the better.

After merging, I'll open a revert PR marked as draft but tagged with 1.16.0 milestone to help ensure this is handled before release, whether it's giving up on the deprecation entirely or else making it more "proper" as mentioned in the NEWS.


tdhock:
no it is not possible currently to do revdeps from a branch
not sure how complicated that would be to implement.
easiest for testing revdeps is if we just merge into master

R/fread.R
```diff
@@ -340,7 +340,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     if (!is.character(key))
       stopf("key argument of data.table() must be a character vector naming columns (NB: col.names are applied before this)")
     if (length(key) == 1L) {
-      key = strsplit(key, split = ",", fixed = TRUE)[[1L]]
+      if (key != strsplit(key,split=",")[[1L]]) stopf("Usage of string-separated literals in %s is deprecated, please split such entries yourself before passing to data.table", "key=")
```
tdhock, line 343: change string-separated to comma-separated?
do you want to add a test for this error message?

tdhock (approved):
looks good to me

MichaelChirico (commented):
> change string-separated to comma-separated?
do you want to add a test for this error message?  

this won't be the last PR here, in particular we will not hit release with such a hard deprecation (i.e. stop()), it will just be opt-in for 1.16.0.

will add tests in the follow-up PR that is release-ready.

PR #5995, opened 2024-03-13
Changed trunc.char to better handle combining chars and full-width multibyte chars

joshhwuu:
Closes #5096 

The old version of `trunc.char` in `data.table.print` used `nchar(x)` internally which works fine in most cases except when dealing with combining characters such as `"á"` (The Latin "a" and accent) as `nchar(x)` recognizes two distinct readable characters. In this case using `nchar(x, 'width')` works as intended. 

This PR changes `trunc.char` to recognize when the char is combining or full-width and indexes accordingly. Additionally, `strtrim()` is used now in place of `substr()` because as mentioned in `?substr`: 

> These functions are often used with `nchar` to truncate a display. That does not really work (you want to limit the width, not the number of characters, so it would be better to use `strtrim`), but at least make sure you use the default `nchar(type = "c")`.

PR #6048, opened 2024-04-04
Changed char.trunc to better handle combining and full-width multibyte characters

### LINKED ISSUE #5096: truncate.char misbehaves for multibyte characters
This is a follow-up to the comments in #3414

Basically print.data.table internals currently use nchar() for printing in a way that won't work well with multibyte characters. it should be as simple as using nchar(., type='width') in the internal truncate.char and then writing tests.

joshhwuu:
Closes #5096 

Second PR, see: #5995.

The old version of `trunc.char` in `data.table.print` used `nchar(x)` internally which works fine in most cases except when dealing with combining characters such as `"á"` (The Latin "a" and accent) as `nchar(x)` recognizes two distinct readable characters. In this case using `nchar(x, 'width')` works better 

This PR changes `trunc.char` to recognize when the char is combining or full-width and indexes accordingly. Additionally, `strtrim()` is used now in place of `substr()` because as mentioned in `?substr`: 

> These functions are often used with `nchar` to truncate a display. That does not really work (you want to limit the width, not the number of characters, so it would be better to use `strtrim`), but at least make sure you use the default `nchar(type = "c")`.

Tasks:

- [x] Find a way to make tests work on all encodings and pass build tests (Works now because Appveyor dropped?)
- [x] Refactor tests for readability
- [x] Update NEWS.md
- [x] Add comment describing potential issues and suggestions to char.trunc


tdhock:
this looks good to me.
make sure to add your name as a contributor in DESCRIPTION, and add an item to NEWS.md

joshhwuu:
Done! Code coverage isn't passing for updating DESCRIPTION, there shouldn't be an issue I'm assuming?

MichaelChirico:
> Works now because Appveyor dropped?

Interesting that it passes on GHA-on-Windows but not Appveyor, hmm. Let's try submitting this to CRAN's winbuilder service to see if it passes there at least.

MichaelChirico:
OK, it passes on winbuilder:

https://win-builder.r-project.org/vpDSszNIo19V/00check.log

I am happy to merge and revisit this if someone else encounters the issue. It may be related to GHA and CRAN using UCRT-enabled machines. I see all the Windows checkers on CRAN use `utf8`, and in fact there are no non-UTF8 checkers left: https://cran.r-project.org/web/checks/check_flavors.html#r-release-windows-x86_64

Likely there's an issue in the tests for non-UTF8 users, but it will be easier to work with a user encountering this who can do fast iteration on a fix themselves.

inst/tests/tests.Rraw
```diff
@@ -18439,3 +18439,43 @@ dt = data.table(a = 1L)
...
+accented_a = "\u0061\u0301"
+ja_ichi = "\u4E00"
+ja_ni = "\u4E8C"
+ja_ko = "\u3053"
+ja_n = "\u3093"
+dots = "..."
+clean_regex = "^\\d+:\\s+" # removes row numbering from beginning of output
+# Tests for combining character latin a and acute accent, single row
+DT = data.table(strrep(accented_a, 4L))
+test(2253.01, gsub(clean_regex, "", capture.output(print(DT))[-1L]), strrep(accented_a, 4L), options=list(datatable.prettyprint.char = 4L))
```
MichaelChirico, line 18453: stylistic request before we merge -- use style `test(n, options=, ...)` to make `options=` clear up-front.

Can you also try going back to the simple `output=` approach of the tests? It will make it a lot cleaner to read. Hopefully the simpler test also passes on windows GHA, if not we can revert back to this functioning version.

joshhwuu: I'll do this once I get home later!

joshhwuu: For the first few tests, the simplified `output=` works well:

```R
...
test(2253.04, options=list(datatable.prettyprint.char = 4L), DT, output=strrep(ja_ichi, 4L))
test(2253.05, options=list(datatable.prettyprint.char = 3L), DT, output=paste0(strrep(ja_ichi, 3L), dots))
test(2253.06, options=list(datatable.prettyprint.char = 1L), DT, output=paste0(strrep(ja_ichi, 1L), dots))
...
```

But for the later tests (Once we have more columns/rows), I find that using this approach requires something like:
```R
test(2253.07, options=list(datatable.prettyprint.char = 4L), DT, output="     V1\n1:    á\n2:   áá\n3:  ááá\n4: áááá")
```
I'm not a huge fan of how the expected output looks, but if you think this way is more readable I can refactor the rest of the tests. For now, I'll commit the options syntax style change and using the simple approach with the first few tests.

MichaelChirico: ah, I had something else in mind but it looks like the force push erased the history in the other branch. oh well. I won't belabor it any further :)

MichaelChirico (approved):

Issue #5980, opened 2024-03-05
revdep neonPlantEcology example/vignette failures with new dcast coercion


tdhock:
looks like #4586 caused a revdep to start failing checks:
```
* checking examples ... ERROR
Running examples in 'neonPlantEcology-Ex.R' failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: npe_summary
> ### Title: Get plant biodiversity information for NEON plots
> ### Aliases: npe_summary
> 
> ### ** Examples
> 
> data("D14")
> plot_level <- neonPlantEcology::npe_summary(neon_div_object = D14, scale = "plot")
Loading required namespace: vegan
Error in dcast.data.table(`_DT14`[, .(site, plotID, subplotID, eventID,  : 
  'x' is not atomic
Calls: <Anonymous> ... as_tibble -> dt_eval -> eval_tidy -> dcast -> dcast.data.table
Execution halted
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes in 'inst/doc' ... OK
* checking running R code from vignettes ...
  'neonPlantEcology.Rmd' using 'UTF-8'... OK
  'using_npe.Rmd' using 'UTF-8'... OK
 NONE
* checking re-building of vignette outputs ... ERROR
Error(s) in re-building vignettes:
  ...
--- re-building 'neonPlantEcology.Rmd' using rmarkdown

Quitting from lines 107-122 [nspp] (neonPlantEcology.Rmd)
Error: processing vignette 'neonPlantEcology.Rmd' failed with diagnostics:
'x' is not atomic
--- failed re-building 'neonPlantEcology.Rmd'

--- re-building 'using_npe.Rmd' using rmarkdown

Quitting from lines 107-122 [nspp] (using_npe.Rmd)
Error: processing vignette 'using_npe.Rmd' failed with diagnostics:
'x' is not atomic
--- failed re-building 'using_npe.Rmd'

SUMMARY: processing the following files failed:
  'neonPlantEcology.Rmd' 'using_npe.Rmd'

Error: Vignette re-building failed.
Execution halted
```

tdhock:
@MichaelChirico any idea why this may be happening?

MichaelChirico:
It looks like here's an MRE:

```r
DT <- data.table(a = c(1,2,1), b = 1:3, c = 0)
dcast(DT, a~b, value.var="c", fill=list(0))

## ON master
[...]

PR #6051, opened 2024-04-05
Retain earlier dcast(fill=list(...)) behavior relying on base coercion behavior for lists

MichaelChirico:
This is #5980.

Not closing since `label:revdep` issues should be closed when revdep check officially removes it, but confirmed locally that {neonPlantEcology} now passes its tests.

### DIFF
--- a/src/fcast.c
+++ b/src/fcast.c
@@ -21,14 +21,16 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
     SEXP thisfill = fill;
     const SEXPTYPE thistype = TYPEOF(thiscol);
     int nprotect = 0;
-    if(some_fill){
+    if (some_fill) {
       if (isNull(fill)) {
-	if (LOGICAL(is_agg)[0]) {
-	  thisfill = PROTECT(allocNAVector(thistype, 1)); nprotect++;
-	} else thisfill = VECTOR_ELT(fill_d, i);
+        if (LOGICAL(is_agg)[0]) {
+          thisfill = PROTECT(allocNAVector(thistype, 1)); nprotect++;
+        } else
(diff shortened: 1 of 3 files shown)


src/fcast.c
```diff
@@ -21,14 +21,17 @@ SEXP fcast(SEXP lhs, SEXP val, SEXP nrowArg, SEXP ncolArg, SEXP idxArg, SEXP fil
...
+        if (LOGICAL(is_agg)[0]) {
+          thisfill = PROTECT(allocNAVector(thistype, 1)); nprotect++;
+        } else
+          thisfill = VECTOR_ELT(fill_d, i);
       }
       if (isVectorAtomic(thiscol)) { // defer error handling to below, but also skip on list
-	thisfill = PROTECT(coerceAs(thisfill, thiscol, /*copyArg=*/ScalarLogical(false))); nprotect++;
+        // #5980: some callers used fill=list(...) and relied on R's coercion mechanics for lists, which are nontrivial, so just dispatch and double-coerce.
+        if (isNewList(thisfill)) { thisfill = PROTECT(coerceVector(thisfill, TYPEOF(thiscol))); nprotect++; }
+        thisfill = PROTECT(coerceAs(thisfill, thiscol, /*copyArg=*/ScalarLogical(false))); nprotect++;
```
tdhock, line 34: so in the example with c=3 and fill=list(0L), thisfill starts as VECSXP, then coerceVector turns it into INTSXP, then coerceAs turns it into REALSXP?

tdhock (approved):

src/fcast.c
MichaelChirico: Yes -- not efficient, but we keep the `coerceAs()` behavior which is an improvement upon `coerceVector()`, the point of #4586 originally.

Specifically, we can have `dcast(x, y~z, fill=list(NA))` get cast to integer64 correctly if needed. That's probably worth a test.

PR #6053, opened 2024-04-07
Move S4 tests to own script

### LINKED ISSUE #3808: isolate S4 unit tests into new test script
Goal is to allow to `cc()`, equivalent to `cc("tests.Rraw")`, run without having to uninstall data.table. If S4 could be put into own Rraw test script, then running only that new S4 test script, `cc("S4.Rraw")`, would require to uninstall data.table.

MichaelChirico:
Closes #3808.

### DIFF
--- a/.dev/cc.R
+++ b/.dev/cc.R
@@ -42,14 +42,11 @@ sourceImports = function(path=getwd(), quiet=FALSE) {
   if (!quiet && length(nsParsedImports)) cat(sprintf("Ensuring objects from %d import entries in NAMESPACE resolve correctly\n", length(nsParsedImports)))
   for (ii in seq_along(nsParsedImports)) {
     entry = nsParsedImports[[ii]]
-    if (paste0("package:", entry[[1L]]) %in% search()) next # not strictly needed since a redundant 'require()' is just skipped, but helpful for reducing noise in !quiet case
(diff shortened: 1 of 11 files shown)


MichaelChirico:
Whoops, somehow I pushed this commit to master in 7b749b1e1649f94736e7ac672bde6bc2b2ed3bf3. I reverted it to make sure this goes through normal review.

MichaelChirico:
Hmm, AFAICT, these lines about needing to uninstall data.table are already irrelevant:

https://github.com/Rdatatable/data.table/blob/a2213177283f0f15823e1ff823c1fdf63746da3d/inst/tests/tests.Rraw#L4-L7

This commit which added that error: 335368b3e

also fixed the tests so that `cc()` runs twice.

So I'm not sure we _need_ to make this change -- we can just remove the part about uninstalling data.table since multiple `cc()` already works.

That said, keeping S4 tests in a separate script makes sense on its own without the above consideration, so I might proceed with this PR anyway.

MichaelChirico:
OK, this is ready for review. It's actually 2 PRs in one: (1) get tests.Rraw in shape to pass under `R_DEFAULT_PACKAGES=NULL` (2) move out the S4 tests. Happy to split these up if needed.

jangorecki (approved):

PR #6057, opened 2024-04-08
cc() works under R_DEFAULT_PACKAGES=NULL

### LINKED ISSUE #6056: cc() assumes NAMESPACE import* entries are already fulfilled
Thus `cc()` doesn't work under `R_DEFAULT_PACKAGES=NULL` since package usages of base functions may not resolve correctly.

MichaelChirico:
Closes #6056

### DIFF
--- a/R/AllS4.R
+++ b/R/AllS4.R
@@ -3,21 +3,21 @@ if ("package:data.table" %in% search()) stopf("data.table package loaded. When d
 
 ## Allows data.table to be defined as an object of an S4 class,
 ## or even have data.table be a super class of an S4 class.
-setOldClass(c('data.frame'))
-setOldClass(c('data.table', 'data.frame'))
+methods::setOldClass(c('data.frame'))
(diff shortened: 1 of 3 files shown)


R/AllS4.R
```diff
@@ -3,21 +3,21 @@ if ("package:data.table" %in% search()) stopf("data.table package loaded. When d
 
 ## Allows data.table to be defined as an object of an S4 class,
 ## or even have data.table be a super class of an S4 class.
-setOldClass(c('data.frame'))
-setOldClass(c('data.table', 'data.frame'))
+methods::setOldClass(c('data.frame'))
```
jangorecki, line 6: I would just import them once at start of the script by

setOldClass = methods::setOldClass

MichaelChirico: that has the effect of basically making this `importFrom(methods, setOldClass)`, whereas `setOldClass()` is only a build-time dependency and not needed by the installed package at all.

jangorecki (approved):

PR #6059, opened 2024-04-08
row separator "---" is no longer in the rownames column when row.names=FALSE

### LINKED ISSUE #4083: print: --- separator is in the rownames register even if row.names=FALSE
```
print(as.data.table(iris), row.names=FALSE)
```

```
    Sepal.Length Sepal.Width Petal.Length Petal.Width   Species
           <num>       <num>        <num>       <num>    <fctr>
             5.1         3.5          1.4         0.2    setosa
             4.9         3.0          1.4         0.2    setosa
             4.7         3.2          1.3         0.2    setosa
             4.6         3.1          1.5         0.2    setosa
             5.0         3.6          1.4         0.2    setosa
---                                                            
             6.7         3.0          5.2         2.3 virginica
             6.3         2.5          5.0         1.9 virginica
             6.5         3.0          5.2         2.0 virginica
             6.2         3.4          5.4         2.3 virginica
             5.9         3.0          5.1         1.8 virginica
```

Note the `---` is still in the column where the row names would have been. This looks a bit strange, and seems to defeat the purpose of `row.names=FALSE`. I would put `---` in the first column here.

joshhwuu:
Closes #4083 

Added simple check for `row.names`. If `row.names=FALSE`, passes a row to `rbind()` with separator `"---"` in the first column, and empty strings in the rest. 

Changed the following existing tests that check printing behavior when `row.names=FALSE`:

- 1138.2
- 2125.03
- 2125.05

These are just changes to the spacing in the expected output of the tests, as well as shifting the `"---"` separator to be right-aligned with the first column instead of the `rownames` column.


joshhwuu:
I agree that `rep("---", ncol(toprint))` may look better in this case. I'll make the changes and if anyone has objections, we can revert.

R/print.data.table.R
```diff
@@ -111,7 +111,11 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     toprint = toprint_subset(toprint, cols_to_print)
   }
   if (printdots) {
-    toprint = rbind(head(toprint, topn + isTRUE(class)), "---"="", tail(toprint, topn))
+    if (isFALSE(row.names)) {
+      toprint = rbind(head(toprint, topn + isTRUE(class)), rep("---", ncol(toprint)), tail(toprint, topn)) # 4083
```
MichaelChirico, line 115:
```suggestion
      toprint = rbind(head(toprint, topn + isTRUE(class)), "---", tail(toprint, topn)) # 4083
```

R will handle recycling `"---"` appropriately

MichaelChirico: But see below also...

inst/tests/tests.Rraw
```diff
@@ -16496,18 +16508,18 @@ test(2125.02, capture.output(print(DT, trunc.cols=TRUE)),
...
-       "    0 bbbbbbbbbbbbb ccccccccccccc",
-       "    0 bbbbbbbbbbbbb ccccccccccccc",
-       "    0 bbbbbbbbbbbbb ccccccccccccc",
+     c("   a             b             c",
+       "   0 bbbbbbbbbbbbb ccccccccccccc",
+       "   0 bbbbbbbbbbbbb ccccccccccccc",
+       "   0 bbbbbbbbbbbbb ccccccccccccc",
+       "   0 bbbbbbbbbbbbb ccccccccccccc",
+       "   0 bbbbbbbbbbbbb ccccccccccccc",
+       " ---           ---           ---",
```
MichaelChirico, line 16517: WDYT about narrowing the column in this case? `toprint` is already small and type character, what about something like

```r
col_widths = pmax(nchar(names(toprint)), apply(toprint, 2L, function(x) max(nchar(x))))
n_dashes = pmin(col_widths, 3L)
dashes = substr(rep("---", ncol(toprint)), 1L, n_dashes)
```

joshhwuu: IMO, the current output looks the best. I'm just thinking in terms of situations where we have many columns with 1-3 character-long names, the different number of dashes may look a bit strange, whereas having the 3 dashes for every column is a bit more consistent.

MichaelChirico: OK. It will be easy enough to change if requested later.

MichaelChirico (approved):
Thanks for taking care of this!

MichaelChirico:
Whoops, we forgot a NEWS item :upside_down_face:

This should get a NEWS item, @joshhwuu could you please file a follow-up PR adding one?

joshhwuu:
Of course!

PR #6062, opened 2024-04-09
Follow-up to PR #6059, added entry to NEWS.md

joshhwuu:


NEWS.md
```diff
@@ -70,6 +70,8 @@
 
 10. `test.data.table()` runs correctly in more sessions, in particular those where the `digits` or `warn` settings are not their defaults (`7` and `0`, respectively), [#5285](https://github.com/Rdatatable/data.table/issues/5285). Thanks @OfekShilon for the report and suggested fix and @MichaelChirico for the PR.
 
+11. Using `print.data.table` with `row.names = FALSE` now prints the separator `"---"` row in every value column instead of the `rownames` column, [#4083](https://github.com/Rdatatable/data.table/issues/4083). Thanks @MichaelChirico for the report and @joshhwuu for the fix.
```
MichaelChirico, line 73:
```suggestion
11. Using `print.data.table` when truncation is needed with `row.names = FALSE`, prints the indicator `---` in every value column instead of adding a blank column where the `rownames` would have been just to include `---`, [#4083](https://github.com/Rdatatable/data.table/issues/4083). Thanks @MichaelChirico for the report and @joshhwuu for the fix.
```

MichaelChirico:
thanks! sorry i didn't catch this on initial review :+1:

PR #6075, opened 2024-04-11
Use %s in 'should be TRUE or FALSE' messages

### LINKED ISSUE #6066: Unify %s should be TRUE or FALSE messages
This is a pretty common error, better for translation would be `stopf("%s should be TRUE or FALSE", "skip_absent")`, but that's out of scope for this PR.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/pull/6044#discussion_r1558485970_

Currently:

```
grep -Fr "should be TRUE or FALSE" po/*.pot
po/data.table.pot:msgid "narm should be TRUE or FALSE"
po/data.table.pot:msgid "fill= should be TRUE or FALSE"
```

joshhwuu:
Closes #6066

### DIFF
--- a/R/duplicated.R
+++ b/R/duplicated.R
@@ -4,7 +4,7 @@ duplicated.data.table = function(x, incomparables=FALSE, fromLast=FALSE, by=seq_
     .NotYetUsed("incomparables != FALSE")
   }
   if (nrow(x) == 0L || ncol(x) == 0L) return(logical(0L)) # fix for bug #28
-  if (is.na(fromLast) || !is.logical(fromLast)) stopf("'fromLast' must be TRUE or FALSE")
+  if (is.na(fromLast) || !is.logical(fromLast)) stopf("'%s' must be TRUE or FALSE", "fromLast")
   if (!length(by)) by = NULL  #4594
   query = .duplicated.helper(x, by)
 
--- a/R/duplicated.R
+++ b/R/duplicated.R
@@ -4,7 +4,7 @@ duplicated.data.table = function(x, incomparables=FALSE, fromLast=FALSE, by=seq_
     .NotYetUsed("incomparables != FALSE")
   }
   if (nrow(x) == 0L || ncol(x) == 0L) return(logical(0L)) # fix for bug #28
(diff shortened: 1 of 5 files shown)


R/duplicated.R
```diff
@@ -4,7 +4,7 @@ duplicated.data.table = function(x, incomparables=FALSE, fromLast=FALSE, by=seq_
     .NotYetUsed("incomparables != FALSE")
   }
   if (nrow(x) == 0L || ncol(x) == 0L) return(logical(0L)) # fix for bug #28
-  if (is.na(fromLast) || !is.logical(fromLast)) stopf("'fromLast' must be TRUE or FALSE")
+  if (is.na(fromLast) || !is.logical(fromLast)) stopf("'%s' must be TRUE or FALSE", "fromLast")
```
MichaelChirico, line 7: Oh, didn't realize this one is an R message, sorry. Actually this won't lead to any consolidation since R- and C-level messages are in separate domains (.pot files). I don't see any similar enough messages at the R level, so let's revert this.

MichaelChirico, line 7:
```suggestion
  if (is.na(fromLast) || !is.logical(fromLast)) stopf("'fromLast' must be TRUE or FALSE")
```

src/rbindlist.c
```diff
@@ -5,7 +5,7 @@
 SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
 {
   if (!isLogical(fillArg) || LENGTH(fillArg) != 1 || LOGICAL(fillArg)[0] == NA_LOGICAL)
-    error(_("fill= should be TRUE or FALSE"));
+    error(_("%s should be TRUE or FALSE"), "fill=");
```
MichaelChirico, line 8:
```suggestion
    error(_("%s should be TRUE or FALSE"), "fill");
```

MichaelChirico:
Please add one more: `src/nafill.c:103:    error(_("nan_is_na must be TRUE or FALSE")); // # nocov`

MichaelChirico (approved):
thanks!

PR #6077, opened 2024-04-11
measure supports cols arg

### LINKED ISSUE #5063: Odd behaviour of `melt()` when using the `cols` argument in `measure()`

tdhock:
Closes #5063 in which some users wanted to provide the cols arg to measure, which may be useful/convenient instead of using pattern arg.

This is an alternative to #5115 which would make using the cols arg an error. Moving forward with that other PR would be complicated since there is quite a bit of code which is using the cols arg with patterns, thanks @MichaelChirico for investigating.

So this PR may be considered if we want to add support for cols arg (seems like users want it in #5063 at least).

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -3159,6 +3159,9 @@ test(1034, as.data.table(x<-as.character(sample(letters, 5))), data.table(V1=x))
   DT_missing_l_2 = data.table(num_1=1, num_2=2, list_1=list(1), list_3=list(3))
   test(1035.0186, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_"), na.rm=TRUE), data.table(char="1", num=1, list=list(1)))
   test(1035.0187, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_"), na.rm=FALSE), data.table(char=c("1","2","3"), num=c(1,2,NA), list=list(1,NA,3)))
+  # measure supports cols arg, #5063
+  test(1035.0188, melt(DT_missing_l_2, measure.vars=measure(prefix, char, sep="_", cols=c("list_1","list_3"))), data.table(num_1=1,num_2=2,prefix="list",char=c("1","3"),value=list(1,3)))
+  test(1035.0188, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_", cols=c("list_1","list_3"))), data.table(num_1=1,num_2=2,char=c("1","3"),list=list(1,3)))
 
   ans1 = cbind(DT[, c(1,2,8), with=FALSE], variable=factor("l_1"))
   ans1[, value := DT$l_1]
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -3160,8 +3160,13 @@ test(1034, as.data.table(x<-as.character(sample(letters, 5))), data.table(V1=x))
   test(1035.0186, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_"), na.rm=TRUE), data.table(char="1", num=1, list=list(1)))
   test(1035.0187, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_"), na.rm=FALSE), data.table(char=c("1","2","3"), num=c(1,2,NA), list=list(1,NA,3)))
   # measure supports cols arg, #5063
-  test(1035.0188, melt(DT_missing_l_2, measure.vars=measure(prefix, char, sep="_", cols=c("list_1","list_3"))), data.table(num_1=1,num_2=2,prefix="list",char=c("1","3"),value=list(1,3)))
-  test(1035.0189, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_", cols=c("list_1","list_3"))), data.table(num_1=1,num_2=2,char=c("1","3"),list=list(1,3)))
+  lcols = c("list_1","list_3")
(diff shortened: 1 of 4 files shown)


inst/tests/tests.Rraw
```diff
@@ -3159,6 +3159,9 @@ test(1034, as.data.table(x<-as.character(sample(letters, 5))), data.table(V1=x))
   DT_missing_l_2 = data.table(num_1=1, num_2=2, list_1=list(1), list_3=list(3))
   test(1035.0186, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_"), na.rm=TRUE), data.table(char="1", num=1, list=list(1)))
   test(1035.0187, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_"), na.rm=FALSE), data.table(char=c("1","2","3"), num=c(1,2,NA), list=list(1,NA,3)))
+  # measure supports cols arg, #5063
+  test(1035.0188, melt(DT_missing_l_2, measure.vars=measure(prefix, char, sep="_", cols=c("list_1","list_3"))), data.table(num_1=1,num_2=2,prefix="list",char=c("1","3"),value=list(1,3)))
+  test(1035.0189, melt(DT_missing_l_2, measure.vars=measure(value.name, char, sep="_", cols=c("list_1","list_3"))), data.table(num_1=1,num_2=2,char=c("1","3"),list=list(1,3)))
```
MichaelChirico, line 3164: What happens when we specify both `patterns=` and `cols=`?

tdhock: works as expected, same result, I just added a couple of tests.

tdhock: so do you agree that the approach in this PR (supporting cols arg) is preferable to the other PR?

MichaelChirico: I think I prefer this one. `patterns(cols=)` indeed already works as I'd expect:

```r
DT=data.table(a=1,b=2,c=3, aa=4)
DT[, .SD, .SDcols=patterns("^a", cols=c("a", "b"))]
#        a
#    <num>
# 1:     1
```

So deprecating `cols=` just removes a functionality, whereas this PR maintains that & adds new functionality.

tdhock: agree

tdhock:
ok tests are passing and I added a NEWS item, so I believe this one is ready to merge.

MichaelChirico (approved):
Excellent :)

PR #6078, opened 2024-04-11
Added a GitHub Action to check for performance regressions within the incoming changes from PRs, along with two test cases based on historical regressions

### LINKED ISSUE #4687: Continuous Benchmarking

### LINKED ISSUE #6065: A GitHub Action to perform automated performance regression testing on pull requests
In an effort to potentially address #4687, I created a [GitHub Action](https://github.com/Anirban166/Autocomment-atime-results/blob/main/action.yml) to facilitate performance testing of the incoming changes that are introduced via pull requests to `data.table`. My motivation in doing so was to help ensure that `data.table` maintains its code efficiency or high-performance standards (core values of the project!) as PRs keep coming and getting integrated frequently (meaning they need to be monitored for performance regressions, and an automatic way to do that would be ideal, noh?).

Onto more details:
- My action utilizes [atime](https://github.com/tdhock/atime) to run predefined tests (as specified in `inst/atime/tests.R`) on different versions of `data.table`. These 'tests' are based on historical regressions that have been documented/discussed via issues and PRs here.
- It uses [cml](https://github.com/iterative/cml) to publish information in a comment on the pull request thread via the GitHub bot (which operates using the `GITHUB_TOKEN` I provide to authenticate itself and interact with the GitHub API within the scope of this step that I defined in my GHA workflow).
- Contents of the comment include a plot consisting of subplots for each test case depicting the time and memory trends across different `data.table` versions, the commit (SHA) that generated it, the link to download the artifact containing all of the `atime`-generated results for that run, the time elapsed from the job's start to the completion of the installation steps, and the time it took to run all the tests.
- To avoid cluttering, only one comment will exist in the PR thread at all times (after the initial comment is generated from the first commit on the PR branch, subsequent pushes will simply update that existing comment's body).
- The different versions of `data.table` (as defined within `atime`) include:
  - HEAD (PR source)
  - base (PR target)
  - merge-base (common ancestor between base and HEAD)
  - CRAN (latest version on the platform)
  - Before (a commit predating the onset of the performance regression)
  - Regression (a commit that is affected by the regression)
  - Fixed (a commit where the regression has been fixed; this should be the same or better as 'Before' in terms of performance)  

To demonstrate, I've created two examples on my fork of `data.table` that can be compared against each other:
- [Using my GitHub Action to observe the performance regression introduced by PR #4491 and fixed by PR #5463](https://github.com/Anirban166/data.table/pull/6)
- [Using my GitHub Action to observe the performance regression fixed by PR #4440](https://github.com/Anirban166/data.table/pull/8)

Both of them replicate PRs that are associated with historical `data.table` regressions (introducing/fixing one) and showcase the expected results in their respective plots:

- For the test case titled 'Test regression fixed in #​5463' one can see that base, merge-base, CRAN, Fixed, and Before are fast while Regression and HEAD are slow for the first PR that replicates the regression *introducing* PR, and in the second PR where it should not be affected, all commits are fast with the expected exception of Regression.

- For 'Test regression fixed in #​4440', the first PR illustrates the reverse case with all commits being fast except the one associated with Regression (as expected since that PR should not affect that test case), and in the second PR where it replicates the regression *fixing* PR, we see that Fixed, HEAD, CRAN, and Before are fast while Regression, merge-base, and base are slow as we would expect.

Note that the first PR's GHA workflow uses my action from the [Marketplace](https://github.com/marketplace/actions/autocomment-atime-results) while the second one is run via a self-contained version of my workflow, or with the entire logic in the file under `.github/workflows/` within the target branch. (I made them different on purpose to illustrate the two ways!)

So a question that I have here is: Which among these two approaches would be preferable if I make a PR to integrate my workflow?

The first approach allows for more direct customization while the second requires comparatively less maintenance (as I update things for my action, only the version numbers need to be changed here). 

Another question I have is regarding the duration. 

The total time taken to run the CI job can be roughly seen as an aggregation of two fairly time-consuming tasks:
- The time it takes to set up everything that is required. A core chunk of the time here is taken in installing `atime` and its dependencies. Roughly 12 minutes on average.
- The time it takes to run the tests, which would vary based on the number and complexity of them. The two tests that were used in the examples above take roughly 3 minutes on average.

Is this acceptable in terms of how long it takes to run the workflow? (and I presume we can skip running the action always; for e.g., changes that only deal with documentation)

Finally, I want to mention that I made this action specifically for `data.table` (again, in hopes that this helps to detect regressions and maintain the quality of code contributions performance-wise!), but it also works on any other GitHub repository for an R package. For example, last month I made a [PR](https://github.com/Anirban166/binsegRcpp/pull/2) between branches on my fork of [binsegRcpp](https://github.com/Anirban166/binsegRcpp) that introduces a performance regression (intentionally placed `Sys.sleep()`) in one of the R files, and it was depicted in the commented plot. (Note that it was run using an older version of my action's code)

Ready for feedback!

Anirban166:
This adds a GHA that will get triggered by every commit in a pull request to the master branch, creating and updating one comment on the PR thread.

As discussed [here](https://github.com/Rdatatable/data.table/issues/6065), it aims to aid in detecting performance degradation for the code changes introduced in PRs using test cases based on historical regressions that we would be maintaining in `inst/atime/tests.R` (added the file for starters, containing the two cases that I used for the examples I mentioned there; more test cases are to be added in the future)

Should probably be skipped for certain PRs, like for changes not based on code (such as documentation or comments) or minor things that are known to not have an impact on performance.

Note: I couldn't come up with a good name for my workflow file so I just named it `autocomment` (hopefully not confusing since at present there isn't another GHA workflow being used in `data.table` that makes comments!), but please feel free to suggest a better name or change it later.

Closes #6065 
Closes #4687


tdhock:
thanks! can you please change the name of the file from autocomment to performance-tests?

Anirban166:
> thanks! can you please change the name of the file from autocomment to performance-tests?

Sure!

inst/atime/tests.R
```diff
@@ -0,0 +1,69 @@
...
+    pkg.edit.fun = pkg.edit.fun,
+    N = 10^seq(3,8),
+    setup = quote({
+      set.seed(1L)
+      dt <- data.table(a = sample(N, N))
+      setindex(dt, a)
+    }),
+    expr = quote(data.table:::shallow(dt)),
+    "Before" = "9d3b9202fddb980345025a4f6ac451ed26a423be", # This should be changed later. Currently, the source of regression (or the particular commit that led to it) is not clear. In addition, older versions of data.table are having problems when being installed in this manner. (This includes commits from before Mar 20, 2020 or when the issue that discovered or first mentioned the regression was created)
+    "Regression" = "752012f577f8e268bb6d0084ca39a09fa7fbc1c4", # A commit that is affected by the regression: https://github.com/Rdatatable/data.table/commit/752012f577f8e268bb6d0084ca39a09fa7fbc1c4
```
tdhock, line 47: this documentation of where the commit ID comes from is not sufficient. 
Right now the comment contains a link to that commit ID, which is redundant with the commit ID itself (does not add any useful information).
We need to know where this commit ID was found. Typically the Regression commit can be the parent of the first commit in the PR which provides the fix, in this case https://github.com/Rdatatable/data.table/pull/4440/commits
The first commit in that PR is https://github.com/Rdatatable/data.table/commit/0f0e7127b880df8459b0ed064dc841acd22f5b73 and on its commit page we can see that the parent is b1b1832b0d2d4032b46477d9fe6efb15006664f4 (not 752012f577f8e268bb6d0084ca39a09fa7fbc1c4 which is the current commit ID in the code, why?).
So a sufficient documentation code/comment would be:
```r
Regression="b1b1832b0d2d4032b46477d9fe6efb15006664f4", # parent of the first commit (https://github.com/Rdatatable/data.table/commit/0f0e7127b880df8459b0ed064dc841acd22f5b73) in the PR (https://github.com/Rdatatable/data.table/pull/4440/commits) that fixes the regression
```

tdhock: please can you properly document all of the other commit IDs, as I suggested above? You may consult https://github.com/DorisAmoakohene/PerformanceTest_data.table since I believe Doris has already done some documentation/comments like this.

Also please remove un-necessary quotes, change "Regression" to Regression in the code. (quotes are not necessary unless the argument names is non-syntactic, meaning starts with a non-letter or contains special characters)

tdhock (changes requested):

inst/atime/tests.R
Anirban166: > Right now the comment contains a link to that commit ID, which is redundant with the commit ID itself (does not add any useful information). 

I agree, and I thought the same (that it does not add any useful information). That's why I did not have those links at the beginning for my examples. But then yesterday you [wrote](https://github.com/Anirban166/data.table/pull/6#discussion_r1559703661) about including 'links' here, and for the [second](https://github.com/Anirban166/data.table/pull/6#discussion_r1559704932) example you gave it had one such redundant link for 'Before' which made me think that was acceptable for you, or what you wanted. 

I already looked up the two test cases I'm using here on that repository yesterday before committing and they only had those type of links (for commit 'x', `https://github.com/Rdatatable/data.table/commit/x`) for both cases in the comments, and nothing else (Now I see that it got updated three hours again, but that wasn't the case yesterday). I thought I should preserve what Doris did and assumed that you both probably discussed this earlier since I didn't find the same suggestion (to remove such links) on her repository.

> We need to know where this commit ID was found/ how we can verify that it is correct.

Yes, agreed.

> So a sufficient documentation code/comment would be:
```r
Regression="b1b1832b0d2d4032b46477d9fe6efb15006664f4", # parent of the first commit (https://github.com/Rdatatable/data.table/commit/0f0e7127b880df8459b0ed064dc841acd22f5b73) in the PR (https://github.com/Rdatatable/data.table/pull/4440/commits) that fixes the regression
```

Thanks for the example.

> please can you properly document all of the other commit IDs, as I suggested above?
> Also please remove un-necessary quotes, change "Regression" to Regression in the code.

Certainly, I will do that soon (currently debugging some errors while developing my function to get the optimal thread count). I didn't want to change the code for the tests (since that's not what I worked on, or I believe it's Doris' part in discussion with you from earlier) so I left the quotes there but I'll change it now along with the other things you mentioned.

Anirban166: > Right now the comment contains a link to that commit ID, which is redundant with the commit ID itself (does not add any useful information). 

I agree, and I thought the same (that it does not add any useful information). That's why I did not have those links at the beginning for my examples. But then yesterday you [wrote](https://github.com/Anirban166/data.table/pull/6#discussion_r1559703661) about including 'links' here, and for the [second](https://github.com/Anirban166/data.table/pull/6#discussion_r1559704932) example you gave it had one such redundant link for 'Before' which made me think that was acceptable for you, or what you wanted. 

I already looked up the two test cases I'm using here on that repository yesterday before committing and they only had those type of links (for commit 'x', `https://github.com/Rdatatable/data.table/commit/x`) for both cases in the comments, and nothing else (now I see that it got updated four hours back, but that wasn't the case yesterday). I thought I should preserve what Doris did and assumed that you both probably discussed this earlier since I didn't find the same suggestion (to remove such links) on her repository.

> We need to know where this commit ID was found/ how we can verify that it is correct.

Yes, agreed.

> So a sufficient documentation code/comment would be:
```r
Regression="b1b1832b0d2d4032b46477d9fe6efb15006664f4", # parent of the first commit (https://github.com/Rdatatable/data.table/commit/0f0e7127b880df8459b0ed064dc841acd22f5b73) in the PR (https://github.com/Rdatatable/data.table/pull/4440/commits) that fixes the regression
```

Thanks for the example.

> please can you properly document all of the other commit IDs, as I suggested above?
> Also please remove un-necessary quotes, change "Regression" to Regression in the code.

Certainly, I will do that shortly (currently debugging some errors while developing my function to get the optimal thread count). I didn't want to change the code for the tests (since that's not what I worked on, or I believe it's Doris' part in discussion with you from earlier) so I left the quotes there but I'll change it soon now along with the other things you mentioned.

Anirban166: > But then yesterday you https://github.com/Anirban166/data.table/pull/6#discussion_r1559703661 about including 'links' here, and for the https://github.com/Anirban166/data.table/pull/6#discussion_r1559704932 example you gave it had one such redundant link for 'Before' which made me think that was acceptable for you, or what you wanted.

Never mind this, I reread and saw the correct changes right now - my bad! (The new commits made me think I was looking at another test case)

Anirban166: > But then yesterday you https://github.com/Anirban166/data.table/pull/6#discussion_r1559703661 about including 'links' here, and for the https://github.com/Anirban166/data.table/pull/6#discussion_r1559704932 example you gave it had one such redundant link for 'Before' which made me think that was acceptable for you, or what you wanted.

Never mind this, I just reread and saw the corrected changes right now - my bad! (The new commits made me think I was looking at another test case)

MichaelChirico:
> cases based on historical regressions that we would be maintaining in inst/atime/tests.R

Is there a particular reason we need to ship this with the package? My instinct would be to put it in e.g. the `.ci` folder.

I only ask because we already generate a CRAN `NOTE` based on the size of the package, a large part of which is coming from the huge text files in `inst/tests`. As the set of performance regression tests grows over time, it will only add more to this burden, but I don't see the benefit of every user having these tests available (whereas we _do_ benefit from every user being able to report `test.data.table()` results).

Anirban166:
@jangorecki Hello Jan!

I was wondering if you know of the commit/source that introduced the regression discussed in #4311 (since you mentioned in a comment there that the regression was already in 1.12.8)

MichaelChirico:
> Should probably be skipped for certain PRs, like for changes not based on code (such as documentation or comments) or minor things that are known to not have an impact on performance.

Yes, I'd be fine only acting on PRs that touch files in R/ or src/. That means a [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

Anirban166:
> Is there a particular reason we need to ship this with the package? My instinct would be to put it in e.g. the `.ci` folder.

I read the [manual](https://cran.r-project.org/web/packages/atime/atime.pdf) and under 'Details' for the `atime_pkg` function (on page 5), it says:
```
There should be a file named pkg.path/inst/atime/tests.R which defines test.list, a list with names
corresponding to different tests.
```
So I presume yes, unfortunately. But I believe @tdhock can change this path to be customizable and then we can redirect?

> As the set of performance regression tests grows over time, it will only add more to this burden, but I don't see the benefit of every user having these tests available (whereas we _do_ benefit from every user being able to report `test.data.table()` results).

I completely agree.

You're right,and unfortunately

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,21 @@
...
+      - synchronize
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    container: ghcr.io/iterative/cml:0-dvc2-base1
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      repo_token: ${{ secrets.GITHUB_TOKEN }}
+      R_KEEP_PKG_SOURCE: yes
```
MichaelChirico, line 19: I would add a quick comment on why this is needed as it's not immediately obvious

Anirban166:
> Yes, I'd be fine only acting on PRs that touch files in R/ or src/. That means a [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

Right on! 
I'll update my workflow with one shortly. (I've used it before in other projects, [here](https://github.com/tdhock/directlabels/blob/c1824df3a760d1faa2f10e7d75e6cb88c94d04f3/.github/workflows/Update_ghpages.yml#L7) for e.g.)

Anirban166:
> Is there a particular reason we need to ship this with the package? My instinct would be to put it in e.g. the `.ci` folder.

I read the [manual](https://cran.r-project.org/web/packages/atime/atime.pdf) and under 'Details' for the `atime_pkg` function (on page 5), it says:
```
There should be a file named pkg.path/inst/atime/tests.R which defines test.list, a list with names
corresponding to different tests.
```
So I presume yes, unfortunately. But I believe @tdhock can change this path to be customizable and then we can redirect?

> As the set of performance regression tests grows over time, it will only add more to this burden, but I don't see the benefit of every user having these tests available (whereas we _do_ benefit from every user being able to report `test.data.table()` results).

I completely agree.

inst/atime/tests.R
```diff
@@ -0,0 +1,69 @@
+pkg.edit.fun = function(old.Package, new.Package, sha, new.pkg.path) {
```
MichaelChirico, line 1: Could you add pseudo-roxygen markup to describe the purpose of this function? It needn't be detailed, but a quick blurb about why we need it (I guess it's consumed by atime?) would be helpful for future readers

.github/workflows/performance-tests.yml
Anirban166: I think I had it in my initial 'use my action from the marketplace' workflow for debugging purposes, but it's not required otherwise so I'll just omit it

inst/atime/tests.R
```diff
@@ -0,0 +1,69 @@
...
+        file.path("src", "init.c"),
+        paste0("R_init_", Package_regex),
+        paste0("R_init_", gsub("[.]", "_", new.Package_)))
+      pkg_find_replace(
+        "NAMESPACE",
+        sprintf('useDynLib\\("?%s"?', Package_regex),
+        paste0('useDynLib(', new.Package_))
+    }
+
+test.list <- list(
```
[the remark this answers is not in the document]
Anirban166: > One thing I see just glancing is: are entries required to be associated with a specific historical regression? 

Nope, and related to what you asked above:

> Are all of the fields required?

Specifying 'Before', 'Regression', or 'Fixed' is optional (you can also specify other label names and any number of them I believe), meaning that if you don't specify them, there are no historical references (and then they would just be regular performance tests, without stemming from comparisons with past versions)

inst/atime/tests.R
tdhock: this function is called for each version that we want to compare in the performance test (HEAD, master, historical regression, etc) and its job is to edit files so that we get an installable package with a name like data.table.abcdef1234567890 (SHA1 hash appended to package name) -- this is the mechanism that atime uses to test different versions of the code against each other, and this custom pkg.edit.fun is required because there are a few non-standard/repetitive places where "data.table" package name needs to be changed in the source files. This is documented in https://github.com/tdhock/atime/blob/main/vignettes/data.table.Rmd#L40 "It is not necessary to
specify `pkg.edit.fun` for typical packages (no compiled code, or
Rcpp), but `data.table` is an interesting example use case for
`pkg.edit.fun` since it specifies a custom shared object file name in
`Makevars`, and it has some custom version checking code in
`onLoad.R`."

tdhock:
> Is there a particular reason we need to ship this with the package? My instinct would be to put it in e.g. the .ci folder.

this is possible, see https://github.com/tdhock/atime/pull/41 for an implementation, would that be OK?

inst/atime/tests.R
```diff
@@ -0,0 +1,69 @@
...
+    setup = quote({
+      n <- N/100
+      set.seed(1L)
+      dt <- data.table(
+        g = sample(seq_len(n), N, TRUE),
+        x = runif(N),
+        key = "g")
+      dt_mod <- copy(dt)
+    }),
+    Before = "19b7866112614db53eb3e909c097407d91cd6738", # Parent of the regression commit (https://github.com/Rdatatable/data.table/commit/0895fa247afcf6b38044bd5f56c0d209691ddb31), which is the parent of the first commit in the PR that causes the issue (https://github.com/Rdatatable/data.table/pull/5493/commits)
```
tdhock, line 66: "regression commit (...), which is the parent of the first commit in the PR that causes the issue" -> seems to imply that the regression commit is the parent of the first commit in the PR that causes the issue, but that can not be the case, because the regression commit must be in the PR that causes the issue (not before that). please rephrase/simplify to "Parent of the first commit (...) in the PR that causes the issue (...)" unless I am mis-understanding something?

inst/atime/tests.R
```diff
@@ -0,0 +1,69 @@
...
+      n <- N/100
+      set.seed(1L)
+      dt <- data.table(
+        g = sample(seq_len(n), N, TRUE),
+        x = runif(N),
+        key = "g")
+      dt_mod <- copy(dt)
+    }),
+    Before = "19b7866112614db53eb3e909c097407d91cd6738", # Parent of the regression commit (https://github.com/Rdatatable/data.table/commit/0895fa247afcf6b38044bd5f56c0d209691ddb31), which is the parent of the first commit in the PR that causes the issue (https://github.com/Rdatatable/data.table/pull/5493/commits) 
+    Regression = "0895fa247afcf6b38044bd5f56c0d209691ddb31", # The regression commit is the parent of the first commit in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5493/commits)
```
tdhock, line 67: please re-phrase to the standard format, "parent of the first commit (...) in the PR that fixed the issue (...)"

tdhock (changes requested):
better comments, thanks! please revise again though.

inst/atime/tests.R
MichaelChirico: thanks, pointing to the right place in atime documentation would be sufficient for me

MichaelChirico:
> this is possible, see [tdhock/atime#41](https://github.com/tdhock/atime/pull/41) for an implementation, would that be OK?

perfect, no need to address anything in this PR then

inst/atime/tests.R
Anirban166: Nope, you understood right as per the relative clause, and I was confused too but I went ditto on those modifications that Doris made and you mentioned yesterday, assuming they were correct and thinking you were okay with them respectively. ([source](https://github.com/Anirban166/data.table/pull/6#discussion_r1559704932))

> seems to imply that the regression commit is the parent of the first commit in the PR that causes the issue, but that can not be the case, because the regression commit must be in the PR that causes the issue (not before that).

That, and I just checked that 19b7866112614db53eb3e909c097407d91cd6738 is the merge-commit of the PR that fixed that regression (#5463) which cannot be 'Before', but instead a candidate for 'Fixed' only (for which here we are using either the last commit or the merge-commit of the regression fixing PR for both test cases, whichever works)

I'm reverting 'Before' there to what I had originally since that makes sense and not this. The source is known, or regression occurs from e793f53466d99f86e70fc2611b708ae8c601a451 so isn't it natural to have the 'Before' set to the commit before it (which is be2f72e6f5c90622fe72e1c315ca05769a9dc854, and it showed correctly on the plot too), or a commit predating that? (in fact, all the commits for this test case were correct before..that example seems like a different case)

inst/atime/tests.R
Anirban166: Initially while making the most recent commit, I made it 'Parent of the first commit (https://github.com/Rdatatable/data.table/pull/5463/commits/ade6e57984f9e7f898a638d2049039cd886f2369) in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5463/commits)" but I realized that doesn't match and the comment too was for '5493'..turns out that's a different case

Anirban166: Initially while making the most recent commit, I made it 'Parent of the first commit (https://github.com/Rdatatable/data.table/commit/ade6e57984f9e7f898a638d2049039cd886f2369) in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5463/commits)" but I realized that doesn't match and the comment above too had *for '5493'*..turns out that's a different case

Anirban166: Initially while making the most recent commit, I made it 'Parent of the first commit (https://github.com/Rdatatable/data.table/commit/0895fa247afcf6b38044bd5f56c0d209691ddb31) in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5493/commits)" but I realized that doesn't match and the comment above too had *for '5493'*..turns out that's a different case.

I can make it "Parent of the first commit (https://github.com/Rdatatable/data.table/commit/ade6e57984f9e7f898a638d2049039cd886f2369) in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5463/commits)" and change Regression to be c4a2085e35689a108d67dacb2f8261e4964d7e12 instead of the regression causing commit or https://github.com/Rdatatable/data.table/commit/e793f53466d99f86e70fc2611b708ae8c601a451 - do you want that?

Anirban166: Initially while making the most recent commit, I made it "Parent of the first commit (https://github.com/Rdatatable/data.table/commit/0895fa247afcf6b38044bd5f56c0d209691ddb31) in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5493/commits)" but I realized that doesn't match and the comment above too had *for '5493'*..turns out that's a different case.

I can make it "Parent of the first commit (https://github.com/Rdatatable/data.table/commit/ade6e57984f9e7f898a638d2049039cd886f2369) in the PR that fixed the issue (https://github.com/Rdatatable/data.table/pull/5463/commits)" and change Regression to be c4a2085e35689a108d67dacb2f8261e4964d7e12 instead of the regression causing commit or https://github.com/Rdatatable/data.table/commit/e793f53466d99f86e70fc2611b708ae8c601a451 - do you want that?

inst/atime/tests.R
```diff
@@ -0,0 +1,108 @@
...
+      pkg_find_replace(
+        "NAMESPACE",
+        sprintf('useDynLib\\("?%s"?', Package_regex),
+        paste0('useDynLib(', new.Package_))
+    }
+
+# A list of performance tests.
+# 
+# Each entry in this list corresponds to a performance test and contains a sublist with three mandatory arguments:
+# - N: A numeric sequence of data sizes to vary.
```
MichaelChirico, line 60: off-topic / thinking aloud: @tdhock have you considered extending {atime} for timing surfaces in 2 dimensions? i'm wondering if we'd benefit from `(nrow, ncol)` 2D benchmarking in some case. Now we are restricted to marginal approach (`nrow`, then `ncol`), but the interaction effects can sometimes be interesting.

inst/atime/tests.R
```diff
@@ -0,0 +1,108 @@
...
+# @note Please check https://github.com/tdhock/atime/blob/main/vignettes/data.table.Rmd for more information.
+test.list <- list(
+  # Performance regression discussed in: https://github.com/Rdatatable/data.table/issues/4311 
+  # Fixed in: https://github.com/Rdatatable/data.table/pull/4440
+  "Test regression fixed in #4440" = list(
+    pkg.edit.fun = pkg.edit.fun,
+    N = 10^seq(3,8),
+    setup = quote({
+      set.seed(1L)
+      dt <- data.table(a = sample(N, N))
```
MichaelChirico, line 81:
```suggestion
      dt <- data.table(a = sample.int(N))
```

inst/atime/tests.R
```diff
@@ -0,0 +1,108 @@
...
+      setindex(dt, a)
```
MichaelChirico, line 82:
```suggestion
      setindexv(dt, "a")
```

inst/atime/tests.R
```diff
@@ -0,0 +1,108 @@
...
+    Regression = "b1b1832b0d2d4032b46477d9fe6efb15006664f4", # Parent of the first commit (https://github.com/Rdatatable/data.table/commit/0f0e7127b880df8459b0ed064dc841acd22f5b73) in the PR (https://github.com/Rdatatable/data.table/pull/4440/commits) that fixes the regression
+    Fixed = "9d3b9202fddb980345025a4f6ac451ed26a423be"), # Merge commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/4440)
+
+  # Test based on: https://github.com/Rdatatable/data.table/issues/5424
+  # Performance regression introduced from a commit in: https://github.com/Rdatatable/data.table/pull/4491
+  # Fixed in: https://github.com/Rdatatable/data.table/pull/5463    
+  "Test regression fixed in #5463" = list(
+    pkg.edit.fun = pkg.edit.fun,
+    N = 10^seq(3, 8),
+    expr = quote(data.table:::`[.data.table`(dt_mod, , N := .N, by = g)),
```
MichaelChirico, line 95: nit: I like having `expr` after `setup` for readability

inst/atime/tests.R
```diff
@@ -0,0 +1,108 @@
...
+    setup = quote({
+      n <- N/100
+      set.seed(1L)
```
MichaelChirico, line 98: would hate to wind up over-anchoring on weird artefacts of a single "unlucky" seed used in all the regression tests:

```suggestion
      set.seed(2L)
```

inst/atime/tests.R
Anirban166: Agreed, I also prefer that for consistency as the first test case does it but this one doesn't (just didn't want to modify the original code as I didn't work on this)

inst/atime/tests.R
Anirban166: Agreed, I wouldn't want specific patterns or behaviors to emerge with this seed only, or have other inputs skew results in a non-representative way. (presuming this change you suggested here is to test whether the results remain consistent)

MichaelChirico:
I am happy with the current state, I'll let Toby give final approval. Thanks again!

inst/atime/tests.R
tdhock: atime is inherently 1 dimensional (N varies), but you can make nrow and ncol vary with N

inst/atime/tests.R
```diff
@@ -0,0 +1,108 @@
...
+  "Test regression fixed in #4440" = list(
+    pkg.edit.fun = pkg.edit.fun,
+    N = 10^seq(3,8),
+    setup = quote({
+      set.seed(1L)
+      dt <- data.table(a = sample.int(N))
+      setindexv(dt, "a")
+    }),
+    expr = quote(data.table:::shallow(dt)),
+    Before = "9d3b9202fddb980345025a4f6ac451ed26a423be", # This needs to be changed later. Currently assigned to the merge commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/4440) as the source of regression (or the particular commit that led to it) is not clear. In addition, older versions of data.table are having problems when being installed in this manner. (This includes commits from before Mar 20, 2020 or when the issue that discovered or first mentioned the regression was created)
```
tdhock, line 85: that does not make sense to call "Before" any commit in the PR for the fix.

tdhock: if you remove please keep a comment about why we don't have a Before commit for this regression.

tdhock (approved):

tdhock:
thanks!

tdhock:
@Anirban166 can you please make a follow-up PR which adds another test, and moves tests under .ci? (I should have asked you to do that before merging, but forgot)

PR #6085, opened 2024-04-12
link measure from the list of valid measure.vars values

### LINKED ISSUE #6079: measure() needs to be mentioned+linked from ?melt
Currently, `?melt.data.table` only gives two examples of using `measure.vars=measure()`, without giving any hint as to how it works.

I think the best approach is to add `measure()` as an item in the list of valid `measure.vars` values, and link to the dedicated `?measure` page.

tdhock:
Closes #6079 

Is this what you had in mind @MichaelChirico ?

### DIFF
--- a/man/melt.data.table.Rd
+++ b/man/melt.data.table.Rd
@@ -27,7 +27,7 @@ non-measure columns will be assigned to it. If integer, must be positive; see De
     \item{ When missing, \code{measure.vars} will become all columns outside \code{id.vars}. }
     \item{ Vector can be \code{integer} (implying column numbers) or \code{character} (column names). }
     \item{ \code{list} is a generalization of the vector version -- each element of the list (which should be \code{integer} or \code{character} as above) will become a \code{melt}ed column. }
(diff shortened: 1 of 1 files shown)


MichaelChirico (approved):
LGTM thanks!

man/melt.data.table.Rd
```diff
@@ -27,7 +27,7 @@ non-measure columns will be assigned to it. If integer, must be positive; see De
     \item{ When missing, \code{measure.vars} will become all columns outside \code{id.vars}. }
     \item{ Vector can be \code{integer} (implying column numbers) or \code{character} (column names). }
     \item{ \code{list} is a generalization of the vector version -- each element of the list (which should be \code{integer} or \code{character} as above) will become a \code{melt}ed column. }
-    \item{ Pattern-based column matching can be achieved with the regular expression-based \code{\link{patterns}} syntax; multiple patterns will produce multiple columns. }
+    \item{ Pattern-based column matching can be achieved with the regular expression-based \code{\link{patterns}} (regex without capture groups; matching column names are used in the \code{variable.name} output column), or \code{\link{measure}} (regex with capture groups; each capture group becomes an output column).
```
MichaelChirico, line 30:
```suggestion
    \item{ Pattern-based column matching can be achieved with the regular expression-based \code{\link{patterns}} (regex without capture groups; matching column names are used in the \code{variable.name} output column), or \code{\link{measure}} (regex with capture groups; each capture group becomes an output column). }
```

PR #6087, opened 2024-04-12
Add "na.print" as a new argument to "print.data.table"

### LINKED ISSUE #3152: print.data.table should honour na.print
In `?print.default`:

> `na.print`	
> a character string which is used to indicate NA values in printed output, or NULL (see ‘Details’).

This works for base R data frames, but not for `data.table`s.

``` r
print(data.frame(x = c(NA, "a", "b")),
      na.print = ".")
#>   x
#> 1 .
#> 2 a
#> 3 b

library(data.table)
print(data.table(x = c(NA, "a", "b")), 
      na.print = ".")
#>       x
#> 1: <NA>
#> 2:    a
#> 3:    b
```

I think it would be worthwhile to offer the default `na.print` as an option, unless the current behaviour is intended.

<sup>Created on 2018-11-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

joshhwuu:
Closes #3152

- [x] Confirm we want to honor `na.print` by exposing it as a new argument to `print.data.table`, instead of handling it through `...` like `print.data.frame`.
- [x] Discuss whether we want to handle error throwing/type checking within `print.data.table` or allow `print.default` to handle it.
- [x] Add additional tests

Currently, `na.print` allows for either `NULL` or a character string to replace `NA` values when printing, but `print.data.table` does not handle this argument. I am suggesting we add the `na.print` argument to `print.data.table` and pass it to existing `print.default` calls, although I think it is possible to handle `na.print` through `...`, although we will have to change some documentation and an extra check.

As for error throwing and type checking, I think the `print.default` error is clear enough:

> invalid 'na.print' specification

However if we want to handle this error through `print.data.table` (maybe for clarity, translation, etc.), it should be easy enough to add. Currently `quote` is handled by `print.default` so I think doing the same for `na.print` would make sense.


man/print.data.table.Rd
```diff
@@ -47,6 +48,7 @@
   \item{trunc.cols}{ If \code{TRUE}, only the columns that can be printed in the console without wrapping the columns to new lines will be printed (similar to \code{tibbles}). }
   \item{quote}{ If \code{TRUE}, all output will appear in quotes, as in \code{print.default}. }
   \item{timezone}{ If \code{TRUE}, time columns of class POSIXct or POSIXlt will be printed with their timezones (if attribute is available). }
+  \item{na.print}{ The string to be printed in place of \code{NA} values. }
```
MichaelChirico, line 51:
```suggestion
  \item{na.print}{ The string to be printed in place of \code{NA} values, as in \code{print.default}. }
```

inst/tests/tests.Rraw
```diff
@@ -18454,3 +18454,9 @@ test(2256.6, fread('a;b\n1,14;5', verbose=TRUE), data.table(a=1.14, b=5L), outpu
 # helpful error about deleting during grouping, #1873
 DT = data.table(id = c(1, 1, 2, 2), a = 1:4, b = 5:8)
 test(2257, DT[ , c("c", "a") := .(a + 1, NULL), by=id], error="it's not possible to delete parts of a column")
+
+DT = data.table(x = c(NA, "a", "b"))
+test(2258.1, capture.output(print(DT, na.print = ".")), c("   x", "1: .", "2: a", "3: b"))
+test(2258.2, capture.output(print(DT, na.print = "_")), c("   x", "1: _", "2: a", "3: b"))
+test(2258.3, capture.output(print(DT, na.print = "NA")), c("    x", "1: NA", "2:  a", "3:  b"))
+test(2258.4, capture.output(print(DT, na.print = TRUE)), error="invalid 'na.print' specification")
```
MichaelChirico, line 18462:
```suggestion
DT = data.table(x=c(NA, "a", "b"))
test(2258.1, capture.output(print(DT, na.print=".")), c("   x", "1: .", "2: a", "3: b"))
test(2258.2, capture.output(print(DT, na.print="_")), c("   x", "1: _", "2: a", "3: b"))
test(2258.3, capture.output(print(DT, na.print="NA")), c("    x", "1: NA", "2:  a", "3:  b"))
test(2258.4, capture.output(print(DT, na.print=TRUE)), error="invalid 'na.print' specification")

```

R/print.data.table.R
```diff
@@ -118,9 +119,9 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     rownames(toprint) = format(rownames(toprint), justify="right")
     if (col.names == "none") {
-      cut_colnames(print(toprint, right=TRUE, quote=quote))
+      cut_colnames(print(toprint, right=TRUE, quote=quote, na.print=na.print))
```
MichaelChirico, line 122: we really shouldn't need all this repetition. but that's not relevant to this PR.

MichaelChirico: #6089

MichaelChirico:
Implementation looks good, including yes to including it in the `print.data.table` signature and to letting `print.default` to error handling. I think we just need NEWS.

inst/tests/tests.Rraw
```diff
@@ -18454,3 +18454,10 @@ test(2256.6, fread('a;b\n1,14;5', verbose=TRUE), data.table(a=1.14, b=5L), outpu
 # helpful error about deleting during grouping, #1873
 DT = data.table(id = c(1, 1, 2, 2), a = 1:4, b = 5:8)
 test(2257, DT[ , c("c", "a") := .(a + 1, NULL), by=id], error="it's not possible to delete parts of a column")
+
+DT = data.table(x=c(NA, "a", "b"))
+test(2258.1, capture.output(print(DT, na.print=".")), c("   x", "1: .", "2: a", "3: b"))
+test(2258.2, capture.output(print(DT, na.print="_")), c("   x", "1: _", "2: a", "3: b"))
+test(2258.3, capture.output(print(DT, na.print="NA")), c("    x", "1: NA", "2:  a", "3:  b"))
+test(2258.4, capture.output(print(DT, na.print=TRUE)), error="invalid 'na.print' specification")
+test(2258.5, capture.output(print(DT, na.print=".", quote=TRUE)), c('   "x"', "1:   .", '2: "a"', '3: "b"'))
```
MichaelChirico, line 18463: missing terminal newline

inst/tests/tests.Rraw
```diff
@@ -18454,3 +18454,10 @@ test(2256.6, fread('a;b\n1,14;5', verbose=TRUE), data.table(a=1.14, b=5L), outpu
 # helpful error about deleting during grouping, #1873
 DT = data.table(id = c(1, 1, 2, 2), a = 1:4, b = 5:8)
 test(2257, DT[ , c("c", "a") := .(a + 1, NULL), by=id], error="it's not possible to delete parts of a column")
+
+DT = data.table(x=c(NA, "a", "b"))
+test(2258.1, capture.output(print(DT, na.print=".")), c("   x", "1: .", "2: a", "3: b"))
```
MichaelChirico, line 18459: We should try and get coverage for every call site, here I think we only cover one.

The remaining 3 require:

 - `col.names="none"`
 - `printdots`, with/without `col.names="none"`

inst/tests/tests.Rraw
MichaelChirico: since we test `quote=TRUE`, it can't hurt to test interaction with `right` as well since they're all supplied to the same `print()` call.

inst/tests/tests.Rraw
joshhwuu: Done! PS: I'm assuming it is expected that `na.print` doesn't work with non-character datatables, as this is consistent with the behavior for printing dataframes as well:
```R
DF = data.frame(x = c(NA, 1))
print(DF, na.print=".")
#    x
# 1 NA
# 2  1
DF = data.frame(x = c(NA,"e"))
print(DF, na.print=".")
#   x
# 1 .
# 2 1
```
I originally wrote a test that was a combination of NAs and integers, and I was wondering for a very long time why it wasn't passing until I realized that we probably wouldn't want strings in our numerical datatables anyway :sob:.

inst/tests/tests.Rraw
joshhwuu: Done! Although `right` shouldn't affect the output as we always pass in `right=TRUE` anyways.

inst/tests/tests.Rraw
MichaelChirico: see also

https://bugs.r-project.org/show_bug.cgi?id=18701

and

https://bugs.r-project.org/show_bug.cgi?id=18702

inspired by this PR :)

inst/tests/tests.Rraw
```diff
@@ -18454,3 +18454,20 @@ test(2256.6, fread('a;b\n1,14;5', verbose=TRUE), data.table(a=1.14, b=5L), outpu
...
+test(2258.2, capture.output(print(DT, na.print="_")), c("   x", "1: _", "2: a", "3: b"))
+test(2258.3, capture.output(print(DT, na.print="NA")), c("    x", "1: NA", "2:  a", "3:  b"))
+test(2258.4, capture.output(print(DT, na.print=TRUE)), error="invalid 'na.print' specification")
+test(2258.5, capture.output(print(DT, na.print=".", quote=TRUE)), c('   "x"', "1:   .", '2: "a"', '3: "b"'))
+test(2258.6, capture.output(print(DT, na.print=".", right=TRUE)), c("   x", "1: .", "2: a", "3: b"))
+# tests for other call sites
+# col.names="none"
+test(2258.7, capture.output(print(DT, na.print=".", col.names="none")), c("1: .", "2: a", "3: b"))
+# printdots=TRUE, col.names="none"
```
MichaelChirico, line 18469:
```suggestion
# table requires splitting, col.names="none"
```

inst/tests/tests.Rraw
```diff
@@ -18454,3 +18454,20 @@ test(2256.6, fread('a;b\n1,14;5', verbose=TRUE), data.table(a=1.14, b=5L), outpu
...
+DT = data.table(x = c(NA, "e", "b", "j", "w", NA))
+test(2258.8, capture.output(print(DT, na.print=".", topn=2, col.names="none")), c(" 1: .", " 2: e", "---  ", " 5: w", " 6: ."))
+# printdots=TRUE, col.names!="none"
```
MichaelChirico, line 18472:
```suggestion
# table requires splitting, col.names!="none"
```

MichaelChirico (approved):

PR #6088, opened 2024-04-12
Use options= to avoid failing to reset verbose

MichaelChirico:
Similar to #6064, `test.data.table()` has continued to spew a ton of output. Traced it back to these tests, which fail to reset `options(datatable.verbose=TRUE)` because `options(old)` doesn't capture that setting.

Rather than fix the `options(old)` call, instead make another incremental improvement towards setting `options=` on a test-by-test level to avoid this pitfall.


tdhock (approved):
nice!

PR #6090, opened 2024-04-13
New regression test re: column plonk in magrittr pipeline

### LINKED ISSUE #2651: Allow column type changes when the as.* call comes at the end of a pipe chain
From here: https://stackoverflow.com/a/49061816/3576984

I used highly functional code for the desired outcome; for such cases I think the `magrittr` syntax increases readability substantially. I would have written the answer as:

```
library(magrittr)
DT[ , c := gsub('[][]', '', c) %>% 
      strsplit(',', fixed = TRUE) %>%
      lapply(as.integer) %>% as.list]
```

But this still fails on column type mismatch despite explicit use of `as.list`. The version where we force `as.list` to the top of the call list looks odd:

```
DT[ , c := as.list(
  gsub('[][]', '', c) %>% 
    strsplit(',', fixed = TRUE) %>%
    lapply(as.integer)
)]
```

And the pure R version is quite unreadable:

```
DT[ , c := as.list(lapply(
  strsplit(gsub('[][]', '', c), ',', fixed = TRUE),
  as.integer))]
```

MichaelChirico:
Closes #2651

### DIFF
--- a/inst/tests/other.Rraw
+++ b/inst/tests/other.Rraw
@@ -722,3 +722,8 @@ if (FALSE) {  # moved from tests.Rraw in #5517 and not yet back on; wasn't sure 
   }
 }
 
+if (loaded[["dplyr"]]) {
+  # regression test for converting character->list column in a magrittr (dplyr) pipe
+  DT = data.table(a = 1, b = 2, c = '[1, 2, 3, 4]', d = 4)
+  test(30, DT[, c := gsub('[][]', '', c) %>%  strsplit(',', fixed = TRUE) %>% lapply(as.integer) %>% as.list]$c, list(1:4))
+}


MichaelChirico:
> could be simplified at all?

Good point, I tried to stay as close to the original issue as possible -- since we're not sure precisely what changed to fix it, it's possible we'll oversimplify and wind up with a test that's not actually a regression test. That said, the first step with the ugly `gsub()` can certainly be dropped, so I did that. I think it's pretty straightforward now, WDYT?

> Also is other.Rraw regularly checked? (I guess not on CRAN, but on CI?)

It's in our main test suite, but only runs with the right `env`:

https://github.com/Rdatatable/data.table/blob/b9d51f1034cc0fdd06d324e62bdd05b9dcb84cef/tests/other.R#L2

This comment says it's run in GLCI, but I don't see that `env` set in our config, nor any reference to 'other.Rraw':

https://github.com/Rdatatable/data.table/blob/b9d51f1034cc0fdd06d324e62bdd05b9dcb84cef/inst/tests/other.Rraw#L4

@jangorecki am I missing something?

tdhock (approved):

jangorecki:
Yes, history of glci yaml will have the answer, of and at which point it was switched off

MichaelChirico:
#5758 removed here but without comment about why other.Rraw was excluded

jangorecki:
AFAIR Matt removed DESCRIPTION-others file and dependencies where not automatically resolved in CI, therefore it was switched off

PR #6091, opened 2024-04-13
Refactor calls to "print.default" within "print.data.table"

### LINKED ISSUE #6089: Avoid repetitive calls to print() [default] inside print.data.table()
we really shouldn't need all this repetition. but that's not relevant to this PR.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/pull/6087#discussion_r1563199703_
            
As illustrated in the linked PR, it's a pain to maintain since we have to remember to update 4 separate call sites to `print.default()`, all with the same signature. This should be unified.

joshhwuu:
Closes #6089

Need discussion on what to name the factored function, currently `print_default()`, and whether to put it inside the `print.data.table` body or outside as a helper

### DIFF
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -111,6 +111,7 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     #   function below adds those back when necessary
     toprint = toprint_subset(toprint, cols_to_print)
   }
+  print_args = list(right = TRUE, quote = quote, na.print = na.print)
   if (printdots) {
     if (isFALSE(row.names)) {
       toprint = rbind(head(toprint, topn + isTRUE(class)), "---", tail(toprint, topn)) # 4083
@@ -119,9 +120,9 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     rownames(toprint) = format(rownames(toprint), justify="right")
     if (col.names == "none") {
-      cut_colnames(print_default(toprint, quote, na.print))
+      cut_colnames(do.call(print, c(list(toprint), print_args)))
     } else {
-      print_default(toprint, quote, na.print)
+      do.call(print, c(list(toprint), print_args))
     }
     if (trunc.cols && length(not_printed) > 0L)
       # prints names of variables not shown in the print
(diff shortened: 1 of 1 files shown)


R/print.data.table.R
```diff
@@ -111,37 +111,27 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     #   function below adds those back when necessary
     toprint = toprint_subset(toprint, cols_to_print)
   }
+  print_default = function(x) {
+    if (col.names != "none") { cut_colnames = identity }
+    cut_colnames(print(x, right=TRUE, quote=quote, na.print=na.print))
+    # prints names of variables not shown in the print
+    if (trunc.cols && length(not_printed) > 0L) trunc_cols_message(not_printed, abbs, class, col.names)
```
MichaelChirico, line 118: let's redefine `trunc.cols` above:

```
if (isTRUE(trunc.cols)) {
  ...
  trunc.cols <- length(not_printed) > 0L
}
```

Then we can simplify the condition here: `if (trunc.cols) trunc_cols_message(...)`.

Alternatively, we can seed `not_printed=NULL`, then handle that directly in the `trunc_cols_message()` helper:

```r
not_printed=NULL
if (trunc.cols) {
  not_printed=...
}

trunc_cols_message=function(not_printed, ...) {
  if (!length(not_printed)) return(invisible())
  ...
}
```

Take a look at both approaches, my instinct is the latter will look more readable in the end.

joshhwuu: I think if we're just doing this to simplify the condition, the former method is cleaner IMO

R/print.data.table.R
```diff
@@ -110,6 +110,13 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     # When nrow(toprint) = 1, attributes get lost in the subset,
     #   function below adds those back when necessary
     toprint = toprint_subset(toprint, cols_to_print)
+    trunc.cols <- length(not_printed) > 0L
+  }
+  print_default = function(x) {
+    if (col.names != "none") { cut_colnames = identity }
```
MichaelChirico, line 116:
```suggestion
    if (col.names != "none") cut_colnames = identity
```

MichaelChirico (approved):
Looks great, thanks!

Issue #5426, opened 2022-07-31
setDT extremely slow for very wide input


MichaelChirico:
```r
DT = replicate(1e7, 1, simplify = FALSE)
system.time(setDT(DT))
#    user  system elapsed 
#  23.043   0.216  23.261 
```

PR #6094, opened 2024-04-18
Moved tests to the more appropriate directory and added a test case based on the performance improvement to be brought by #5427

Anirban166:
A follow-up to #6078 that:
- moves the atime-based test cases under `.ci` and correspondingly updates the version of my action in the `performance-tests` workflow to accommodate that change.
- adds a test for the performance improvement (with respect to #5426) implemented in #5427. Bumps up the total time taken to run the tests (3) to 4 minutes on average.

(For reference, the current state of the plot with these test cases looks like [this](https://github.com/Anirban166/data.table/pull/9#issuecomment-2050835653))


tdhock:
thanks!
I was expecting to see an auto comment with the new plot, but guess we are not seeing that because the modifications are outside the R/ and src/ directories?
Can you please change it so that modifications to the atime/tests.R file trigger the performance test auto comment action? (then in PRs which change the performance tests, we would be able to see what the new test result plot looks like)

.ci/atime/tests.R
```diff
@@ -104,5 +104,17 @@ test.list <- list(
     expr = quote(data.table:::`[.data.table`(dt_mod, , N := .N, by = g)),    
     Before = "be2f72e6f5c90622fe72e1c315ca05769a9dc854", # Parent of the regression causing commit (https://github.com/Rdatatable/data.table/commit/e793f53466d99f86e70fc2611b708ae8c601a451) in the PR that introduced the issue (https://github.com/Rdatatable/data.table/pull/4491/commits)
     Regression = "e793f53466d99f86e70fc2611b708ae8c601a451", # Commit responsible for regression in the PR that introduced the issue (https://github.com/Rdatatable/data.table/pull/4491/commits)
-    Fixed = "58409197426ced4714af842650b0cc3b9e2cb842") # Last commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/5463/commits)    
+    Fixed = "58409197426ced4714af842650b0cc3b9e2cb842"), # Last commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/5463/commits)
+
+  # Issue reported in: https://github.com/Rdatatable/data.table/issues/5426
+  # To be fixed in: https://github.com/Rdatatable/data.table/pull/5427
+  "Test performance improvement implemented in #5427" = list(
```
tdhock, line 111: This looks like a reasonable new test, but now I notice that all of the test names start with Test, and some include the keyword performance, both of which could be applied to any performance test, so to reduce repetition, can you please remove those keywords from the test names?

tdhock (changes requested):

tdhock:
Also I see Fast/Slow in grey, which should be fixed by my recent PR to atime https://github.com/tdhock/atime/pull/48/files (not yet on CRAN, plan to submit next week, after which time we can consider merging this PR)
68747470733a2f2f61737365742e636d6c2e6465762f633562646638316334386366386261393032356337626438656238376633633862306662656431303f636d6c3d706e672663616368652d6279706173733d30373863636263662d383465392d343561342d386133

Anirban166:
> I was expecting to see an auto comment with the new plot, but guess we are not seeing that because the modifications are outside the R/ and src/ directories?

Yup

I made the changes and we'll get to see the plot soon (two times as I made two pushes; although for one workflow or `R-CMD-check` in particular I noticed it has preempted runs for the first commit for two made in quick succession)

> Also I see Fast/Slow in grey, which should be fixed by my recent PR to atime https://github.com/tdhock/atime/pull/48/files (not yet on CRAN, plan to submit next week, after which time we can consider merging this PR) 

Perfect!
And yes let's wait till then.

Anirban166:
Just saw the plot, not sure why `merge-base` is missing..

MichaelChirico:
> Comparison Plot
> 
> Generated via commit [28483e8](https://github.com/Rdatatable/data.table/commit/28483e8f8f184ab04dbf3804ef0d70c8ebf0143a)
> 
> Download link for the artifact containing the test results: [↓ atime-results.zip](https://github.com/Rdatatable/data.table/actions/runs/8744615247/artifacts/1427514765)
> 
> Time taken to finish the standard R installation steps: 11 minutes and 56 seconds
> 
> Time taken to run `atime::atime_pkg` on the tests: 3 minutes and 52 seconds
> 
> 

@Anirban166 how will faceting work if there are, e.g. 10 plots? will it wind up being 2x20 plot or will it start breaking into new rows soon?

.github/workflows/performance-tests.yml
```diff
@@ -10,7 +10,8 @@ on:
       - synchronize
     paths:
       - 'R/**'
-      - 'src/**'       
+      - 'src/**'
+      - '.ci/atime/tests.R'
```
MichaelChirico, line 14:
```suggestion
      - '.ci/atime/**'
```

Seems more appropriate

Anirban166:
@Anirban166 how will faceting work if there are, e.g. 10 plots? will it wind up being 2x20 plot or will it start breaking into new rows soon?

Good question in terms of visual scalability. I agree (if you're thinking the same) that we would likely want to vertically append those subplots too if it's too much on the horizontal. I'll need to investigate, but the answer should be in the code that generates these plots using `atime`.

Anirban166:
> @Anirban166 how will faceting work if there are, e.g. 10 plots? will it wind up being 2x20 plot or will it start breaking into new rows soon?

Good question in terms of visual scalability. I agree (if you're thinking the same) that we would likely want to vertically append those subplots too if it's too much on the horizontal. I'll need to investigate, but the answer should be in the code that generates these plots using `atime`.

Anirban166:
@MichaelChirico I looked it up and it seems those subplots will be appended

Anirban166:
@MichaelChirico I looked it up and it seems that adding to the test cases will keep on appending horizontally only as per [this line](https://github.com/tdhock/atime/blob/1dac9ef8056a751095f88264aad24d7ed30aa851/R/versions.R#L296): (and for each test case the time and memory plots are vertically arranged as per [this line](https://github.com/tdhock/atime/blob/1dac9ef8056a751095f88264aad24d7ed30aa851/R/atime.R#L221))
```r
ggplot2::facet_grid(unit ~ expr.name, scales="free")
```
So yes, at present it'll be 2x20 for twenty test cases with each row only corresponding to a different unit for each of those columns.

MichaelChirico:
I filed https://github.com/tdhock/atime/issues/49 to handle this first upstream, nothing to do in data.table for now.

tdhock:
> Just saw the plot, not sure why merge-base is missing..

merge-base is missing if it is the same as another commit already shown (typically master)

tdhock:
why does the Fast curve memory go up so sharply for large N?

.ci/atime/tests.R
```diff
@@ -55,13 +55,13 @@ pkg.edit.fun = function(old.Package, new.Package, sha, new.pkg.path) {
     }
 
 # A list of performance tests.
-# 
+#
 # Each entry in this list corresponds to a performance test and contains a sublist with three mandatory arguments:
 # - N: A numeric sequence of data sizes to vary.
 # - setup: An expression evaluated for every data size before measuring time/memory.
-# - expr: An expression that will be evaluated for benchmarking performance across different git commit versions. 
+# - expr: An expression that will be evaluated for benchmarking performance across different git commit versions.
```
tdhock, line 62: please try to avoid making insignificant changes like this if possible, because it makes it more difficult for others to review the diff

MichaelChirico: Note that these changes were made in #5908. They were added here too because the file moving locations will make it harder to keep the change synced between PRs. Not sure the ideal approach.

tdhock: @Anirban166 atime 2024.4.17 with support for Fast/Slow colors is now on CRAN, so can you please commit a revert of these white space changes, which will hopefully trigger a new performance test plot with updated colors?

Anirban166: > which will hopefully trigger a new performance test plot with updated colors?

Yup (you can see the updated colors now)

tdhock: ok sorry

.ci/atime/tests.R
```diff
@@ -70,7 +70,6 @@ pkg.edit.fun = function(old.Package, new.Package, sha, new.pkg.path) {
 # - sha.vec: Named character vector or a list of vectors that specify data.table-specific commit SHAs for testing across those different git commit versions.
 #            For historical regressions, use 'Before', 'Regression', and 'Fixed' (otherwise something like 'Slow' or 'Fast' ideally).
 # @note Please check https://github.com/tdhock/atime/blob/main/vignettes/data.table.Rmd for more information.
-# nolint start: undesirable_operator_linter. ':::' needed+appropriate here.
```
tdhock: @MichaelChirico did you actually want to keep this?

MichaelChirico: Yes, it's needed, it will make sense for benchmarking to design tests on private interface to avoid overhead of only using public API.

.ci/atime/tests.R
```diff
@@ -105,6 +105,18 @@ test.list <- list(
     expr = quote(data.table:::`[.data.table`(dt_mod, , N := .N, by = g)),
     Before = "be2f72e6f5c90622fe72e1c315ca05769a9dc854", # Parent of the regression causing commit (https://github.com/Rdatatable/data.table/commit/e793f53466d99f86e70fc2611b708ae8c601a451) in the PR that introduced the issue (https://github.com/Rdatatable/data.table/pull/4491/commits)
     Regression = "e793f53466d99f86e70fc2611b708ae8c601a451", # Commit responsible for regression in the PR that introduced the issue (https://github.com/Rdatatable/data.table/pull/4491/commits)
-    Fixed = "58409197426ced4714af842650b0cc3b9e2cb842") # Last commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/5463/commits)
+    Fixed = "58409197426ced4714af842650b0cc3b9e2cb842"), # Last commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/5463/commits)
```
MichaelChirico, line 108: Another aside: @tdhock let's support `test.list` having a `NULL` final entry, that way we can have like:

```r
test.list <- list(
  "A" = list(...),
  "B" = list(...),
  NULL
)
```

And avoid the pesky git diff on irrelevant lines every time a test is added (as in here, where we add `,` --> diff)

.ci/atime/tests.R
tdhock: ok @Anirban166 please re-add this

Anirban166: Michael took care of that

.ci/atime/tests.R
tdhock: great idea

MichaelChirico (approved):
Great stuff!

.ci/atime/tests.R
tdhock: implemented in atime github, https://github.com/tdhock/atime/blob/ea591e37a8fbf69a06a0cdbb8a51108cb73290fb/R/test.R#L14 I will file a follow up PR next week after that gets on CRAN

tdhock (approved):

PR #6097, opened 2024-04-22
Fix other.Rraw tests to run locally

MichaelChirico:
Follow-up to #6090 -- since it seems this suite has not been run lately in CI, I tried locally and found it's not passing, fixed here.

Note that `options(warn=1)` in `tests/other.R` was doing nothing, since `test.data.table()` always sets `warn=0`:

https://github.com/Rdatatable/data.table/blob/d420afe916aa78c41f8d9857630f8c4de8abf537/R/test.data.table.R#L102


tdhock (approved):
thanks for your diligence Michael

Issue #6082, opened 2024-04-12
test.data.table() needs to initialize/reset setNumericRounding()


MichaelChirico:
Fails some tests:

```r
setNumericRounding(2)
test.data.table()
# Error: 2 error(s) out of 11157. Search inst/tests/tests.Rraw for test number(s) 1196.1, 1
```

After #1898 this could be a lot worse. Best to treat numeric rounding as another initial state that we want to control and then restore to user's setting on exit.

markseeto:
@MichaelChirico I'd like to try doing this.

MichaelChirico:
great, thanks for volunteering! assigned you, please reach out if you get stuck.

markseeto:
Thanks @MichaelChirico.

PR #6103, opened 2024-04-25
Add numeric rounding value initialisation in test.data.table()

markseeto:
Closes https://github.com/Rdatatable/data.table/issues/6082

Added `setNumericRounding(0)` to initialise the numeric rounding value in test.data.table() to avoid failed tests if the user has set a different value. The user's value is restored on exit.


markseeto:
@MichaelChirico Could you please help me understand the reason for the failed checks?

Before creating this pull request, I built the package with my changed files, then ran `R CMD check data.table_1.15.99.tar.gz` and it finished with `Status: OK`.

I also installed the package and ran
```
setNumericRounding(2)
test.data.table()
```
and that worked, with
`All 11183 tests (last 2261.1) in tests/tests.Rraw completed ok in 00:01:11 elapsed (00:01:01 cpu)`.

For the `R-CMD-check / ubuntu-20.04 (release) (pull_request)` check, in Details, it says `Error: Error in loadNamespace(x) : there is no package called ‘rcmdcheck’`. Do you know what I have to do to fix this?

For the `Autocomment atime-based performance regression analysis on PRs / comment (pull_request)` check, in Details, it has
```
Switched to a new branch 'master'
branch 'master' set up to track 'origin/master'.
fatal: invalid reference: testdatatable
Error: Process completed with exit code 128.
```
so I'm guessing that the problem might be caused by the PR being based on a non-master branch. If you can tell me how to fix this, it would be greatly appreciated.

tdhock:
it is normal that `Autocomment atime-based performance regression analysis on PRs / comment (pull_request)` fails for PRs that come from forks.

markseeto:
Thanks for the explanation @tdhock.

NEWS.md
```diff
@@ -56,6 +56,8 @@
 
 7. `melt` returns an integer column for `variable` when `measure.vars` is a list of length=1, consistent with the documented behavior, [#5209](https://github.com/Rdatatable/data.table/issues/5209). Thanks to @tdhock for reporting and fixing. Any users who were relying on this behavior can change `measure.vars=list("col_name")` (output `variable` was column name, now is column index/integer) to `measure.vars="col_name"` (`variable` still is column name).
 
+8. `test.data.table()` now initialises the numeric rounding value to 0 using `setNumericRounding(0)` to avoid failed tests if the user has set a different value, [#6082](https://github.com/Rdatatable/data.table/issues/6082). The user's value is restored on exit. Thanks to @MichaelChirico for the report and for describing the solution, and @markseeto for implementing.
```
MichaelChirico, line 59: Let's put this under NOTES

R/test.data.table.R
```diff
@@ -79,6 +79,9 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
   # sample method changed in R 3.6 to remove bias; see #3431 for links and notes
   # This can be removed (and over 120 tests updated) if and when the oldest R version we test and support is moved to R 3.6
 
+  userNumericRounding = getNumericRounding()
+  setNumericRounding(0)  # Initialise to 0 in case the user has set it to a different value. Restore to user's value when finished.
```
MichaelChirico, line 83: we should probably return the old rounding instead of `NULL` in `setNumericRounding()`, to match similar behavior by `setwd()`, `options()`, etc. Out of scope for this PR.

inst/tests/tests.Rraw
```diff
@@ -18532,3 +18532,10 @@ test(2260.15, inherits(tryCatch(messagef("x", class="test_message"), condition=i
 test(2260.16, inherits(tryCatch(messagef("x", class="test_message"), condition=identity), "message"))
 test(2260.17, inherits(tryCatch(packageStartupMessagef("x", class="test_psm"), condition=identity), "test_psm"))
 test(2260.18, inherits(tryCatch(packageStartupMessagef("x", class="test_psm"), condition=identity), "packageStartupMessage"))
+
+# Test setting of default numeric rounding to 0 in test.data.table(), #6082
+# The file issue_6082_tests.Rraw contains tests 1196.1 and 1196.2. Check that these pass when numeric rounding value has been set to 2 before running test.data.table().
+old_rounding = getNumericRounding()
+setNumericRounding(2)
+test(2261.1, as.logical(test.data.table(script = "issue_6082_tests.Rraw")))
```
MichaelChirico, line 18540: this is essentially a regression test on `test.data.table()` and/or `test()`, which we don't have in the suite yet (the whole file is marked `# nocov`). It doesn't quite match the normal way of doing tests for us, and I'm not an immediate fan of nesting `test.data.table()` within `test.data.table()`... let's just revert the new tests in this PR.

I'll file a follow-up that we should really write some meta-tests of `test.data.table()` itself, at some point.

R/test.data.table.R
MichaelChirico: #6108

inst/tests/tests.Rraw
MichaelChirico: #6109

MichaelChirico:
LGTM! Two small tweaks & we can merge.

markseeto:
@MichaelChirico Thanks for reviewing. I've deleted the test, deleted the file issue_6082_tests.Rraw, and moved the News item from Bug Fixes to Notes.

MichaelChirico (approved):
Thanks!

MichaelChirico:
Thanks @markseeto! You should have received an invite to become a project Member. The main thing this does is give you rights to create branches on this repo directly to facilitate future collaboration -- please do so as it makes the review process a bit easier. Welcome!

markseeto:
@MichaelChirico Thanks for approving and thanks for the invitation.

PR #6106, opened 2024-04-29
Refactor colClasses handling for readability

MichaelChirico:
Towards #6105. This PR is a simple refactor to reduce code nesting / improve readability. It makes the actual implementation of #6105, #6107, much simpler.

PR #6107, opened 2024-04-29
Defer as.Date() coercion to R level

### LINKED ISSUE #6105: fread: performance difference with/without specifying colClasses
Hello,
I noticed a large performance difference in `fread()` with and without specifying `colClasses()`.
If i specify the classes of the toy dataset below, the execution takes considerably longer than without specifying any classes. 
There is a similar question on [StackOverflow](https://stackoverflow.com/questions/68703718/r-data-table-fread-performance-with-and-without-colclasses) (currently without answer), but in that case it is not entirely clear whether the issue is relative to the calculation also present in the script or due to `fread()` itself. 
So, i created the script below which deals exclusively with `fread()`.
In one case, I also specify the keys when reading in the data, but that does not seem to make any difference.

```
set.seed(1234)
vote_id <- sample(x = 100000:200000, size = 20000, replace = FALSE)
pers_id <- sample(x = 100000:200000, size = 900, replace = FALSE)
dates <- sample(x = seq(as.Date('2010/01/01'), as.Date('2011/01/01'), by="day"), size = 250, replace = FALSE)
result <- sample(x = -1:1, size = 1000000, replace = TRUE)
big_dt <- data.table::data.table( 
  vote_id = vote_id,
  pers_id = pers_id,
  dates = dates,
  result = result)
# write data  
data.table::fwrite(x = big_dt, "big_dt.csv")
# becnmark fread
res = bench::mark(
  fread_classes_keys = data.table::fread(
    file = "big_dt.csv",
    colClasses = list(
      integer = c("vote_id", "pers_id", "result"),
      Date = c("dates") ), 
    key = c("dates", "vote_id", "pers_id")),
  fread_classes = data.table::fread(
    file = "big_dt.csv",
    colClasses = list(
      integer = c("vote_id", "pers_id", "result"),
      Date = c("dates") )),
  fread = data.table::fread(
    file = "big_dt.csv"),
  check = FALSE, iterations = 10)
print(res)
sessionInfo()
```

Results
```
> source("/home/scipima/Documents/github/ep_rollcall/test_scripts/fread_test.R", encoding = "UTF-8")
# A tibble: 3 × 13
  expression     min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr>  <bch:> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 fread_clas…  2.54s   2.55s     0.390   159.3MB    0.780    10    20      25.6s
2 fread_clas…  2.51s   2.51s     0.397   146.9MB    0.794    10    20      25.2s
3 fread       8.75ms 12.59ms    83.4      16.8MB    8.34     10     1    119.9ms
# ℹ 4 more variables: result <list>, memory <list>, time <list>, gc <list>
Warning messages:
1: Item 2 has 900 rows but longest item has 1000000; recycled with remainder. 
2: Some expressions had a GC in every iteration; so filtering is disabled. 
```

`Output of sessionInfo()`
```
sessionInfo()
R version 4.4.0 (2024-04-24)
Platform: x86_64-pc-linux-gnu
Running under: Linux Mint 21.3

Matrix products: default

locale:
 [5] LC_MONETARY=nl_BE.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=nl_BE.UTF-8       LC_NAME=C                 
[11] LC_MEASUREMENT=nl_BE.UTF-8 LC_IDENTIFICATION=C       

time zone: Europe/Brussels
tzcode source: system (glibc)

attached base packages:

 [1] compiler_4.4.0    magrittr_2.0.3    bench_1.1.3       cli_3.6.2        
 [5] tools_4.4.0       pillar_1.9.0      glue_1.7.0        tibble_3.2.1     
 [9] utf8_1.2.4        fansi_1.0.6       vctrs_0.6.5       data.table_1.15.0
[13] jsonlite_1.8.8    lifecycle_1.0.4   pkgconfig_2.0.3   rlang_1.1.3      
[17] profmem_0.6.0    
```

MichaelChirico:
Closes #6105

As seen in the tests, this _might_ be a breaking change if any downstreams depend on the output being _specifically_ Date (and not IDate). I am not sure it's possible -- `inherits(., "Date")` is still `TRUE`, and relevant methods should just back up to Date methods if not available for IDate. I lean towards just going ahead with this change unless revdeps finds anything concerning, WDYT @jangorecki?

cc also @HughParsonage who may have some extra insights.

---

One alternative I looked into that _won't_ work is adding `"Date"` as an alias for `"IDate"` in the `type` hierarchy:

https://github.com/Rdatatable/data.table/blob/d19bfef7026f25bb2d36c17879187d09bcb2b2c3/src/freadR.c#L27-L29

The problem is that there _are_ files where our rudimentary date parser doesn't detect the date, but `as.Date()` does, meaning `fread.c` returns a string --> the hierarchy is broken as there's an attempt to cast string->int32. See in particular this test:

https://github.com/Rdatatable/data.table/blob/d19bfef7026f25bb2d36c17879187d09bcb2b2c3/inst/tests/tests.Rraw#L11043


MichaelChirico:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6107?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6107** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6106** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

MichaelChirico:
WIP: need to investigate why `colClasses=<vector>` returns Date, while `colClasses=<list>` returns IDate.

jangorecki:
Yes, let's see revdeps

.ci/atime/tests.R
```diff
@@ -118,7 +118,17 @@ test.list <- atime::atime_test_list(
...
+    Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12",  # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"), # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+  "fread(colClasses=Date) improved in #6107" = atime::atime_test(
+    N = 10^seq(1, 7),
+    setup = {
+      DT = data.table(date=.Date(sample(20000, N, replace=TRUE))),
+      tmp_csv = tempfile()
+      fwrite(DT, tmp_csv)
+    },
+    expr = fread(tmp_csv, colClasses=list(Date='date')),
```
MichaelChirico, line 130: cc @Anirban166 @tdhock one consideration here -- actually there are 3 independent codepaths (at present) for potential regression:

 1. As here, with `colClasses = list(Date = ...)`
 2. With "anonymous" columns, i.e. `fread(f, colClasses="Date")`, required when column names are not known in advance
 3. With `select=` (see description in `?fread`)

I think ideally we'd combine some of the logic, but for now, it looks like we'd want 3 separate regression tests for each of these, WDYT?

Also, I am wondering what happens to `tmp_csv` -- when is it cleaned up? I don't see any `cleanup=` step in `atime_versions()`, do we need to wait for the R session to end for those (potentially large) files to be removed?

.ci/atime/tests.R
```diff
@@ -107,7 +107,7 @@ test.list <- atime::atime_test_list(
     Fixed = "58409197426ced4714af842650b0cc3b9e2cb842"), # Last commit in the PR that fixed the regression (https://github.com/Rdatatable/data.table/pull/5463/commits)
 
   # Issue reported in: https://github.com/Rdatatable/data.table/issues/5426
-  # To be fixed in: https://github.com/Rdatatable/data.table/pull/5427
+  # Ffixed in: https://github.com/Rdatatable/data.table/pull/5427
```
Anirban166, line 110: Fixed

.ci/atime/tests.R
Anirban166: > I think ideally we'd combine some of the logic, but for now, it looks like we'd want 3 separate regression tests for each of these, WDYT?

I agree (three separate regression tests would be better for now)

> Also, I am wondering what happens to `tmp_csv` -- when is it cleaned up? I don't see any `cleanup=` step in `atime_versions()`, do we need to wait for the R session to end for those (potentially large) files to be removed?

I'm not sure about the cleanup (Toby would know better), but I'm wondering too if we can have something like a `file.remove` or `unlink` call on `tmp_csv`

.ci/atime/tests.R
```diff
@@ -118,7 +118,17 @@ test.list <- atime::atime_test_list(
...
-    Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
-    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15") # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+    Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12",  # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"), # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+  "fread(colClasses=Date) improved in #6107" = atime::atime_test(
+    N = 10^seq(1, 7),
+    setup = {
+      DT = data.table(date=.Date(sample(20000, N, replace=TRUE)))
+      tmp_csv = tempfile()
+      fwrite(DT, tmp_csv)
```
MichaelChirico, line 128: @jangorecki I think we should run `fread(DT)` once here in the `setup` because of cacheing, right? Or do we need to run in a subprocess? Here the benchmark is really about what happens _after_ the fread.c code is run, only care about differences emerging (1) in freadR.c and/or (2) fread.R post-processing. WDYT?

.ci/atime/tests.R
Anirban166: > I think ideally we'd combine some of the logic, but for now, it looks like we'd want 3 separate regression tests for each of these, WDYT?

I agree (three separate regression tests would be better for now)

> Also, I am wondering what happens to `tmp_csv` -- when is it cleaned up? I don't see any `cleanup=` step in `atime_versions()`, do we need to wait for the R session to end for those (potentially large) files to be removed?

I'm not sure about the cleanup (Toby would know better), but I'm wondering too if we can have something like a `file.remove` or `unlink` call on `tmp_csv` as well

Anirban166: > I think ideally we'd combine some of the logic, but for now, it looks like we'd want 3 separate regression tests for each of these, WDYT?

I agree (three separate regression tests would be better for now)

> Also, I am wondering what happens to `tmp_csv` -- when is it cleaned up? I don't see any `cleanup=` step in `atime_versions()`, do we need to wait for the R session to end for those (potentially large) files to be removed?

I'm not sure about the cleanup (Toby would know better), but I'm wondering too if we can have something like a `file.remove` or `unlink` call on `tmp_csv`

tdhock: right three separate tests.
yes we need to wait for the R session to end for those files to be removed.
they probably won't be too large because of the 0.01 second time limit.
also please make sure to do data.table::fread instead of just fread inside expr.

.ci/atime/tests.R
MichaelChirico: (regardless, we see substantial improvement in all 3 cases already)

src/freadR.c
```diff
@@ -335,7 +335,7 @@ bool userOverride(int8_t *type, lenOff *colNames, const char *anchor, const int
               type[i]=CT_STRING; // e.g. CT_ISO8601_DATE changed to character here so that as.POSIXct treats the date-only as local time in tests 1743.122 and 2150.11
               SET_STRING_ELT(colClassesAs, i, tt);
             }
-          } else {
+          } else if (type[i] != CT_ISO8601_DATE || tt != char_Date) {
             type[i] = typeEnum[w-1];                           // freadMain checks bump up only not down
             if (w==NUT) SET_STRING_ELT(colClassesAs, i, tt);
           }
```
HughParsonage, line 341: What is the consequence of a dangling else here? (As in, if the new condition evaluates to false, it would previously be evaluated.) I think a comment for this line would be useful here.

MichaelChirico: Thanks for the call-out, this is something I wasn't understanding well myself.

The key is we just skip updating `type[i]` for that case. We might consider rewriting this then as:

```c
type[i] = (type[i] == CT_ISO8601_DATE && tt == char_Date) ? type[i] : typeEnum[w-1];
```

The "trouble" comes in the next line, because `w==NUT` in the relevant case here (i.e., `"Date"` is not a known class for C-side coercion).

So with the simple fix, we're back to R-side `as.Date()` coercion.

As noted, we _might_ prefer that (retaining fully back-compatible `Date` output columns), but it'll require a corresponding change in the `colClasses=list()` branch.

So it's down again to whether the result of this PR should be "cols parsed as IDate and requested as Date are returned as IDate, fully avoiding any coercion" or "we now skip the middle step in IDate->char->Date coercion; for maximum efficiency, request IDate instead of Date".

.ci/atime/tests.R
```diff
@@ -118,7 +118,37 @@ test.list <- atime::atime_test_list(
...
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"), # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+  "fread(colClasses=list(Date)) improved in #6107" = atime::atime_test(
+    N = 10^seq(1, 7),
+    setup = {
+      DT = data.table(date=.Date(sample(20000, N, replace=TRUE)))
+      tmp_csv = tempfile()
+      fwrite(DT, tmp_csv)
+    },
+    expr = data.table::fread(tmp_csv, colClasses=list(Date='date')),
+    Slow = "d19bfef7026f25bb2d36c17879187d09bcb2b2c3",
```
Anirban166, line 131: @MichaelChirico Can you please add a comment where the SHAs came from? (reviewing with Toby)

Anirban166 (changes requested):
The performance tests look great! (but please add the comments)

.ci/atime/tests.R
MichaelChirico: `Slow` I pulled from `master` around when I filed this PR. `Fast` is the first commit of this PR. Is that worth documenting? Is there some better approach I should be aware of?

Anirban166: Sounds reasonable enough to me, but I think @tdhock prefers to have comments mentioning that information beside those lines for consistency with how the other tests are documented or for others to know quickly why those commit SHAs are being used

tdhock: to make it easy to double check where the commit IDs come from, it is best if they can be referenced relative to a PR, like the existing comments, "last commit in PR#xyz (URL)" etc

MichaelChirico: Per recent discussion, dropped the atime edit here; I have it ready in branch `fread-date-idate-atime` to submit as a follow-up. That'll resolve the issues with git hash.

MichaelChirico:
My recollection is this was ready to submit besides the issue of tidying up the atime part (deferred to follow-up now), and we held off on the potential breaking change near release; submitting now.

PR #6112, opened 2024-04-30
setNumericRounding() returns old rounding value, instead of "NULL"

### LINKED ISSUE #6108: setNumericRounding() should return the initial value
we should probably return the old rounding instead of `NULL` in `setNumericRounding()`, to match similar behavior by `setwd()`, `options()`, etc. Out of scope for this PR.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/pull/6103#discussion_r1583889460_

joshhwuu:
Closes #6108

Changed `setNumericRounding()` to invisibly return the old value, allowing for this assignment:

```R
old = setNumericRounding(2)     # store old value and initiate new value
setNumericRounding(old)         # restore old value
```

Should I include a change to `test.data.table` as outlined in the issue as well? Also, there are many instances where we use `getNumericRounding()` to set/restore numeric rounding and can be refactored following this change, are they worth changing?


MichaelChirico:
> there are many instances where we use getNumericRounding() to set/restore numeric rounding and can be refactored following this change, are they worth changing?

Sorry, I'm not seeing those, unless you're referring to usage outside data.table itself?

MichaelChirico:
> Should I include a change to test.data.table as outlined in the issue as well?

Yes, the total change size is still pretty small --> safe to include here too. Thanks!

inst/tests/tests.Rraw
```diff
@@ -18532,3 +18532,15 @@ test(2260.15, inherits(tryCatch(messagef("x", class="test_message"), condition=i
...
+test(2261.02, getNumericRounding(), 0L)
+test(2261.03, setNumericRounding(1L), 0L)
+test(2261.04, getNumericRounding(), 1L)
+test(2261.05, setNumericRounding(2L), 1L)
+test(2261.06, getNumericRounding(), 2L)
+test(2261.07, setNumericRounding(1L), 2L)
+test(2261.08, getNumericRounding(), 1L)
+test(2261.09, setNumericRounding(0L), 1L)
+test(2261.10, getNumericRounding(), 0L)
+test(2261.11, capture.output(setNumericRounding(2L)), character(0))
```
MichaelChirico, line 18546: terminal newline missing

inst/tests/tests.Rraw
```diff
@@ -18532,3 +18532,15 @@ test(2260.15, inherits(tryCatch(messagef("x", class="test_message"), condition=i
 test(2260.16, inherits(tryCatch(messagef("x", class="test_message"), condition=identity), "message"))
 test(2260.17, inherits(tryCatch(packageStartupMessagef("x", class="test_psm"), condition=identity), "test_psm"))
 test(2260.18, inherits(tryCatch(packageStartupMessagef("x", class="test_psm"), condition=identity), "packageStartupMessage"))
+
+test(2261.01, setNumericRounding(0L), 0L)
+test(2261.02, getNumericRounding(), 0L)
+test(2261.03, setNumericRounding(1L), 0L)
```
MichaelChirico, line 18538: it looks like some of these tests are redundant, can you explain why we need 11 tests?

joshhwuu: yeah it was in the commit message, i just had these for a sanity check, but can certainly cut it down

inst/tests/tests.Rraw
```diff
@@ -18532,3 +18532,15 @@ test(2260.15, inherits(tryCatch(messagef("x", class="test_message"), condition=i
 test(2260.16, inherits(tryCatch(messagef("x", class="test_message"), condition=identity), "message"))
 test(2260.17, inherits(tryCatch(packageStartupMessagef("x", class="test_psm"), condition=identity), "test_psm"))
 test(2260.18, inherits(tryCatch(packageStartupMessagef("x", class="test_psm"), condition=identity), "packageStartupMessage"))
+
+test(2261.01, setNumericRounding(0L), 0L)
```
MichaelChirico, line 18536: better to initialize this -- it could be that some other test earlier in the suite is changed with adverse effects here. e.g.

```r
old_rounding = getNumericRounding()
setNumericRounding(0L)
test(...)
...
setNumericRounding(old_rounding)
```

joshhwuu:
> Sorry, I'm not seeing those, unless you're referring to usage outside data.table itself?

Yeah, I meant in `tests.Rraw`, a bunch of tests do this

inst/tests/tests.Rraw
MichaelChirico, line 18546: WDYT about `test(2261.11, setNumericRounding(2L), notOutput=".*")`?

inst/tests/tests.Rraw
MichaelChirico: Let's try and make it more minimal, thanks. If we think there's added value of tests that might appear redundant, a good comment will help future readers.

MichaelChirico:
> Yeah, I meant in tests.Rraw, a bunch of tests do this

Oh, I see it now. I guess tests.Rraw is not indexed for regex search by GitHub.

Let's mark it for follow-up since there are so many. Thanks!

MichaelChirico:
Implementation looks good, only nit-picky details left to fix, thanks!

joshhwuu:
Should we include a NEWS entry as well? Or is it minimal enough where it's not required?

MichaelChirico:
> Should we include a NEWS entry as well? Or is it minimal enough where it's not required?

I think it warrants a NOTE

inst/tests/tests.Rraw
joshhwuu: Upon trying this locally, this doesn't work, I'm getting:
```R
Test 2261.05 produced output but should not have:
  Expected absent (case insensitive): <<.*>>
  Observed: <<[1] 2>>
```
I think this is expected behaviour because `notOutput` calls `print` I believe

This happens with the previous version too:
```R
data.table:::test(1, getNumericRounding(), notOutput=".*")
```

MichaelChirico: thanks for investigating, please comment the purpose of the test in that case 👎

MichaelChirico (approved):
LGTM!

Issue #6114, opened 2024-05-01
mutation testing results: tests to add


tdhock:
Hi, @agroce and did mutation testing of data.table, and here are some changes/mutants that we found, that resulted in tests and CRAN checks still passing, and they look like significant changes, for which we may want to add tests. 
This was run on 1.15.0 so the line numbers/links refer to that version of the code (first line original, second line mutated).

[as.data.table.R:221](https://github.com/Rdatatable/data.table/blob/1.15.0/R/as.data.table.R#L221)
```
ans = data.table(rn=rownames(x), x, keep.rownames=FALSE, key=key)
ans = data.table(rn=rownames(x), keep.rownames=FALSE, x, key=key)
```

[bmerge.R:92](https://github.com/Rdatatable/data.table/blob/1.15.0/R/bmerge.R#L92)
```
if (xclass==""""integer64"""" || iclass==""""integer64"""") {
if (xclass>=""""integer64"""" || iclass==""""integer64"""") {
```

[data.table.R:770](https://github.com/Rdatatable/data.table/blob/1.15.0/R/data.table.R#L770)
```
if (!is.symbol(bysub) && (length(bysubl)!=2L || !is.symbol(bysubl[[2L]]) || !(bysubl[[1L]] %chin% c(""""."""",""""c"""",""""list""""))))
if (!is.symbol(bysub) && (length(bysubl)!=2L || !is.symbol(bysubl[[-1L]]) || !(bysubl[[1L]] %chin% c(""""."""",""""c"""",""""list""""))))
```

[data.table.R:2490](https://github.com/Rdatatable/data.table/blob/1.15.0/R/data.table.R#L2490)
```
r = do.call(""""CJ"""", c(ul, sorted=sorted, unique=TRUE))
r = do.call(""""CJ"""", c( sorted=sorted,ul, unique=TRUE))
```

[data.table.R:2536](https://github.com/Rdatatable/data.table/blob/1.15.0/R/data.table.R#L2536)
```
lapply(ll, split.data.table, drop=drop, by=by[-1L], sorted=sorted, keep.by=keep.by, flatten=flatten)
lapply(ll, drop=drop, split.data.table, by=by[-1L], sorted=sorted, keep.by=keep.by, flatten=flatten)
```

[data.table.R:2969](https://github.com/Rdatatable/data.table/blob/1.15.0/R/data.table.R#L2969)
```
} else if (name %iscall% c('$', '[[') && is.name(name[[2L]])) {
[...]

PR #6115, opened 2024-05-01
test fread(key=character())

tdhock:
this was identified by mutation testing, #6114 

[fread.R:341](https://github.com/Rdatatable/data.table/blob/1.15.0/R/fread.R#L341)
```
if (length(key) == 1L) {
if (length(key) < 1L) {
```
changing == to < above resulted in all passing tests, whereas with this new test, there should be a failure with that change.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -13319,7 +13319,7 @@ test(1958.01, fread('\U0001f64d', encoding = 'UTF-16'), error = "Argument 'encod
 test(1958.02, fread('a,b\n1,2', nrows = NA_real_),      data.table(a = 1L, b = 2L))
 test(1958.03, fread('a,b\n1,2', nrows = -1),            data.table(a = 1L, b = 2L))
 test(1958.04, fread('a,b\n1,2', key = 1),               error = 'must be a character vector naming columns')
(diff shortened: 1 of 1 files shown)


inst/tests/tests.Rraw
```diff
@@ -13319,6 +13319,7 @@ test(1958.01, fread('\U0001f64d', encoding = 'UTF-16'), error = "Argument 'encod
 test(1958.02, fread('a,b\n1,2', nrows = NA_real_),      data.table(a = 1L, b = 2L))
 test(1958.03, fread('a,b\n1,2', nrows = -1),            data.table(a = 1L, b = 2L))
 test(1958.04, fread('a,b\n1,2', key = 1),               error = 'must be a character vector naming columns')
+test(1958.041, fread('a,b\n1,2', key = character()),    data.table(a = 1L, b = 2L), warning="cols is a character vector of zero length. Removed the key, but use NULL instead, or wrap with suppressWarnings() to avoid this warning.")
```
MichaelChirico, line 13322: I think the warning test may be too exact, can we weaken it and still get the point across?

tdhock: sure

MichaelChirico (approved):

PR #6117, opened 2024-05-02
Debuggin codecov issue

MichaelChirico:
<WIP, feel free to edit with ideas>


MichaelChirico:
Per here, it looks like there's a {covr} bug interacting with R 4.4.0+ here. Closing this investigation -- the next step is to bisect the R-devel change that caused the breakage.

https://github.com/r-lib/covr/issues/567

Issue #6116, opened 2024-05-01
file path issues when testing on windows


tdhock:
Hi! I expected tests should pass on windows, but I am observing two failures, related to https://github.com/Rdatatable/data.table/issues/1668 (same test numbers, not same failure)
```
th798@cmp2986 MINGW64 ~/R/data.table (fread-key-length-0)
$ R -e 'library(data.table);test.data.table()'

R Under development (unstable) (2024-01-23 r85822 ucrt) -- "Unsuffered Consequences"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-w64-mingw32/x64

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

Loading required namespace: BiocManager
> library(data.table);test.data.table()
getDTthreads(verbose=TRUE):
  OpenMP version (_OPENMP)       201511
  omp_get_num_procs()            12
  R_DATATABLE_NUM_PROCS_PERCENT  unset (default 50)
  R_DATATABLE_NUM_THREADS        unset
  R_DATATABLE_THROTTLE           unset (default 1024)
  omp_get_thread_limit()         2147483647
  omp_get_max_threads()          12
  OMP_THREAD_LIMIT               unset
  OMP_NUM_THREADS                unset
  RestoreAfterFork               true
  data.table is using 6 threads with throttle==1024. See ?setDTthreads.
test.data.table() running: C:/Program Files/R/R-devel/library/data.table/tests/tests.Rraw
'file:' is not recognized as an internal or external command,
operable program or batch file.
Test 1378.2 produced 2 warnings but expected 0
Expected: 
Observed: '(file://C:/Program Files/R/R-devel/library/data.table/tests/russellCRLF.csv) > C:\Users\th798\AppData\Local\Temp\RtmpGMzTle\file1680248362b7' execution failed with error code 1
 Test 1378.2 produced 2 warnings but expected 0
Expected: 
Observed: File 'C:\Users\th798\AppData\Local\Temp\RtmpGMzTle\file1680248362b7' has size 0. Returning a NULL data.table.
[...]

PR #6118, opened 2024-05-02
Multiple observed warnings combined in display

MichaelChirico:
As seen in #6116. Before/after:

```r
test(1, {warning(1); warning(2); TRUE})
## BEFORE
Test 1 produced 2 warnings but expected 0
Expected: 
Observed: 1
 Test 1 produced 2 warnings but expected 0
Expected: 
Observed: 2

## AFTER
Test 1 produced 2 warnings but expected 0
Expected: 
Observed: 1
Observed: 2
```


tdhock (approved):
great thanks

PR #6119, opened 2024-05-03
Replace unnecessary calls to getNumericRounding() in tests.Rraw where possible

### LINKED ISSUE #6113: Use new setNumericRounding() return value in tests where possible.
> Yeah, I meant in tests.Rraw, a bunch of tests do this

Oh, I see it now. I guess tests.Rraw is not indexed for regex search by GitHub.

Let's mark it for follow-up since there are so many. Thanks!

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/issues/6112#issuecomment-2087729793_

joshhwuu:
Closes #6113 

Since #6112, we can now assign the old numeric rounding value when calling `setNumericRounding()`, this behavior better aligns with `old = options(...)` and can improve readability in certain tests


inst/tests/tests.Rraw
```diff
@@ -13137,8 +13131,7 @@ test(1942.02, DT[,sum(v),keyby=id1,verbose=TRUE], data.table(id1=c("D","A","C"),
 setindex(DT,id2)
 test(1942.03, DT[,sum(v),keyby=id2,verbose=TRUE], data.table(id2=INT(3,9,2), V1=INT(6,6,3), key="id2"), output="Finding groups using uniqlist on index 'id2")
 setindex(DT,id3)
-oldnr = getNumericRounding()  # TODO: return old
-setNumericRounding(0)
+oldnr = setNumericRounding(0)  # TODO: return old
```
MichaelChirico, line 13134:
```suggestion
oldnr = setNumericRounding(0)
```

At least I think we fulfilled this TODO

MichaelChirico:
Thanks for taking this on! Please make the one small change, I think GitHub struggles to 'apply suggestion' to a huge file like tests.Rraw so I couldn't do it in the UI :\

MichaelChirico (approved):

PR #6124, opened 2024-05-04
Activate + apply some linters

MichaelChirico:
Still not trying to touch inst/tests/tests.Rraw too much, but these small-scale changes should be OK to move forward.


tdhock (approved):
lgtm

PR #6127, opened 2024-05-08
Another batch of low-volume, low-complexity lint fixes

MichaelChirico:
Follow-up to #6124, here are some more lint fixes that can get our codebase more consistent without needing too much review effort.

### DIFF
--- a/R/fcast.R
+++ b/R/fcast.R
@@ -251,5 +251,5 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
     setattr(ans, 'names', c(lhsnames, allcols))
     setDT(ans); setattr(ans, 'sorted', lhsnames)
   } else stopf("Internal error -- empty rhsnames in dcast; please report") # nocov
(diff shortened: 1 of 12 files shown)


R/fcast.R
```diff
@@ -251,5 +251,5 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
     setattr(ans, 'names', c(lhsnames, allcols))
     setDT(ans); setattr(ans, 'sorted', lhsnames)
   } else stopf("Internal error -- empty rhsnames in dcast; please report") # nocov
-  return (ans)
+  return(ans)
```
tdhock, line 254: is there a linter to avoid return at end of function? (I only recommend using return for exiting early in R code)

tdhock (approved):
lgtm

R/fcast.R
MichaelChirico: yes, return_lintee() 😄 

currently deactivated

PR #6128, opened 2024-05-08
Activate 5 more linters, including return_linter

MichaelChirico:
Another in the series (prev: #6127).

@tdhock mentioned wanting `return_linter()` activated, so here it is. Also added `missing_argument_linter()` which only has one hit.


tdhock (approved):
that looks a lot better thanks

PR #6131, opened 2024-05-10
More fixes for occasional CI test based on new failures

MichaelChirico:
Based on latest failures: https://github.com/Rdatatable/data.table/actions/runs/9032914230

 - plain `locale-gen lv_LV` picks a different default encoding:

    ```shell
    sudo locale-gen lv_LV
    # Generating locales (this might take a while)...
    #   lv_LV.ISO-8859-13... done
    # Generation complete.
    ```

 - Need to be robust to {rcmdcheck} being unavailable, https://github.com/r-lib/rcmdcheck/issues/220
 - Need to specifically install some packages for other.Rraw. Previously, this was done by a dummy DESCRIPTION file, however I find we can pretty easily just parse the file itself since the required packages are on the first line.
 - Need to install system curl manually for {curl} to install, not sure why this doesn't affect the other GHA CI...

### DIFF
--- a/.github/workflows/R-CMD-check-occasional.yaml
+++ b/.github/workflows/R-CMD-check-occasional.yaml
@@ -101,8 +101,11 @@ jobs:
       - name: Install dependencies
         run: |
           remotes::install_deps(dependencies = TRUE)
+          other_deps_expr = parse('inst/tests/other.Rraw', n=1L)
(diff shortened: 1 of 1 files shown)


MichaelChirico:
OK, I think that's enough fixes to get us pretty well situated for tomorrow. Submitting.

Issue #995, opened 2015-01-01
retain attributes while chaining


jangorecki:
Question / Feature request.
Any tricky way to retain attributes when chaining? of course not all should be retained ("index","sorted","names", etc.).

``` r
library(data.table)
DT <- data.table(a = rep(1:5,2), b = rnorm(10))
setattr(DT,"my_custom_attr","attr_value")
str(DT)

# I would like to reuse attribute while chaining and also after chaining
DT[,attr(.SD,"my_custom_attr")] # NULL
attr(DT[,lapply(.SD,sum),by="a"],"my_custom_attr") # NULL
```

So the Feature request would be about retaining all user defined (non-DT related: "sorted","index", etc.) attributes.
Or maybe retain attrs defined by name, something like

``` r
options("datatable.retain.attr"=c("my_custom_attr","my_custom_attr2"))
```

This would allow to store custom metadata together with data.table, possibility reuse them while processing, or manipulate when using `DT[,f(.SD)]`.

arunsrinivasan:
Have marked as FR for now.

jangorecki:
new SO about that: https://stackoverflow.com/questions/33441632/retaining-metadata-when-subsetting-in-data-table
this SO highlights it is a gap comparing to data.frame

assaron:
Guy, is there are any problems with a solution like the following?

First, adding a function that will copy required attributes:

```{r}
retainAttributes <- function(x, ans) {
    if ("retain.attributes" %in% names(attributes(x))) {
        for (a in c(attr(x, "retain.attributes"), "retain.attributes")) {
            attr(ans, a) <- attr(x, a)
        }
    }
    ans
}
```

And then using it is as a wrapper when neccessary in `[.data.table`:
```
diff --git a/R/data.table.R b/R/data.table.R
index d18dd3f..dda4985 100644
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1343,7 +1343,7 @@ chmatch2 <- function(x, table, nomatch=NA_integer_) {
             setattr(ans, "class", class(x)) # fix for #5296
             setattr(ans, "row.names", .set_row_names(nrow(ans)))
             
-            if (!with || missing(j)) return(alloc.col(ans))
+            if (!with || missing(j)) return(alloc.col(retainAttributes(x, ans)))
 
             SDenv$.SDall = ans
[...]

PR #6136, opened 2024-05-15
all.equal.data.table gets ignore.indices argument

### LINKED ISSUE #6134: Optionally ignore indices in all.equal
Different tables that contain the same data and also behave pretty much the same will not be `all.equal()` when they have different indices. Indices can sometimes appear automatically depending on what operations are performed with tables, so I think it can often be reasonable to ignore differences in indices.
```r
x <- data.table(a = 1)
y <- data.table(a = 1)
all.equal(x, y)
#> [1] TRUE
x[a == 1]
#>        a
#>    <num>
#> 1:     1
all.equal(x, y)  # would be nice to have a way to get 'TRUE' here
#> [1] "Datasets have different indices. 'target': [a]. 'current': has no index."
```
While one can set `check.attributes = FALSE`, this also ignores other more important things like column names
```r
all.equal(x, y, check.attributes = FALSE)
#> [1] TRUE
all.equal(x, data.table(b = 1), check.attributes = FALSE)  # this is too lenient
#> [1] TRUE
```
The user can set / unset indices in tables before checking with `all.equals()`, but that gets unnecessarily complicated when the tables are buried inside larger objects.

It would therefore be useful to have an option `check.indices` or `ignore.indices` that makes `all.equal()` ignore data.table indices.

My use case is that I use `all.equal.data.table` a lot in unit tests, where I verify that objects created in different ways still contain the same data. Tables may have different indices depending on how objects are created, but because these indices make no relevant difference in object behaviour I would like to ignore them.

mb706:
Indices can now be ignored in `all.equal()` when `ignore.indices` is set:
```r
x <- data.table(a = 1)
y <- data.table(a = 1)
x[a == 1]
#>        a
#>    <num>
#> 1:     1
all.equal(x, y)
#> [1] "Datasets have different indices. 'target': [a]. 'current': has no index."
all.equal(x, y, ignore.indices = TRUE)
#> [1] TRUE
```

closes #6134


R/setops.R
```diff
@@ -104,8 +104,8 @@ fsetequal = function(x, y, all=TRUE) {
 
 # all.equal ----
 
-all.equal.data.table = function(target, current, trim.levels=TRUE, check.attributes=TRUE, ignore.col.order=FALSE, ignore.row.order=FALSE, tolerance=sqrt(.Machine$double.eps), ...) {
-  stopifnot(is.logical(trim.levels), is.logical(check.attributes), is.logical(ignore.col.order), is.logical(ignore.row.order), is.numeric(tolerance), is.data.table(target))
+all.equal.data.table = function(target, current, trim.levels=TRUE, check.attributes=TRUE, ignore.indices=FALSE, ignore.col.order=FALSE, ignore.row.order=FALSE, tolerance=sqrt(.Machine$double.eps), ...) {
```
ben-schwen, line 107: Any reason why we are adding a new argument in the middle? This might break downstream code.

ben-schwen:
What about keys?

mb706:
From my view the case for ignoring keys is less strong, since they do not appear automatically (I believe?) and also change the behaviour of the data.table object in more relevant ways. Ultimately that would be your decision, I can also implement ignoring keys if you want me to.

R/setops.R
mb706: I decided to group ignore.indices with check.attributes as they seem most related, as indices are attributes. I would say that users should always pass all.equal's optional arguments by name, since these are usually passed on recursively, which makes positional argument passing risky with all.equal beyond the first two. If you want me to, I can move ignore.indices to the end.

ben-schwen:
Nah, keys are more intrusive since they also change the layout of how the data is stored, so I share your opinion that they are substantial and it might be less common to ignore them.

What just crossed my mind is instead of adding a new argument we could extend `check.attributes` to also accept character vectors to specify attributes that should be ignored. This seems more versatile.

mb706:
> `check.attributes` to also accept character vectors to specify attributes that should be ignored.

It would be nicer if it were called `ignore.attributes`. IDK how you handle these cases, you could introduce an argument `ignore.attributes` with default `!check.attributes` and deprecate `check.attributes`?

In any case, the `index` attribute would still get special treatment. In this PR the line
```r
   if (!ignore.indices) {
```
would need to be changed to
```r
   if (!"index" %in% ignore.attributes) {
```

mb706:
Any opinions on which route you want to go with this? Should `check.attributes` be a vector of attribute names to be ignored? Should it be renamed `ignore.attributes` or something? or should I leave `ignore.indices` (and if so, should I move it to the end of the argument list)?

MichaelChirico:
I lean towards new argument `ignore.attributes` with default value `!check.attributes`. I agree `check.attributes = "index"` would be awkward and it's not worth supporting `check.attributes = !"index"`. Also requiring `check.attributes=<vector of names>` is a bit unwieldy too -- I'd rather operate on an exclusion principle here, I think that's more natural.

I believe we get `check.attributes` from the same argument to `attr.all.equal()`, so there's also something to be said for now overloading the argument name with functionality that's not present in base.

It's pretty common in other packages to use the same argument name in their `all.equal()` methods:

https://github.com/cran/bit64/blob/53a5d8eaf9ef55554fd44f91396f51745dc8243d/man/all.equal.integer64.rd#L18
https://github.com/cran/spam/blob/69c5411e812bfa2f57550c1c510c2c61d31a149f/man/allequal.Rd#L29
https://github.com/cran/nanotime/blob/a57b523997ca25f7b57ee40edf81dc73bdabf951/man/all.equal.nanoduration.Rd#L16
https://github.com/cran/nanotime/blob/a57b523997ca25f7b57ee40edf81dc73bdabf951/man/all.equal.nanoival.Rd#L8-L10
https://github.com/cran/nanotime/blob/a57b523997ca25f7b57ee40edf81dc73bdabf951/man/all.equal.nanoperiod.Rd#L8-L10
https://github.com/cran/nanotime/blob/a57b523997ca25f7b57ee40edf81dc73bdabf951/man/all.equal.nanotime.Rd#L16-L27
https://github.com/cran/kutils/blob/cec4bcd983f84aeb6700b33ff71c0d0100c40c15/man/all.equal.key.Rd#L7
https://github.com/cran/kutils/blob/cec4bcd983f84aeb6700b33ff71c0d0100c40c15/man/all.equal.keylong.Rd#L7
https://github.com/cran/nat/blob/189b954c314255d7906503a727bba228653c9007/man/all.equal.dotprops.Rd#L10
https://github.com/cran/nat/blob/189b954c314255d7906503a727bba228653c9007/man/all.equal.neuron.Rd#L11

One other option would be to allow something like `check.attributes="shared"`, meaning we compare attributes only for `intersect(names(attributes(current)), names(attributes(target)))`.

MichaelChirico:
Checking residual interest in continuing this PR given the above feedback.

jangorecki:
> One other option would be to allow something like `check.attributes="shared"`, meaning we compare attributes only for `intersect(names(attributes(current)), names(attributes(target)))`.

rather than changing argument type from logical to character we could as well allow `check.attributes = NA` which would be something between TRUE and FALSE, like TRUE for common attributes, FALSE for present on one side only.

jangorecki:
Btw. I have a WIP branch of rewrite all.equal to address https://github.com/Rdatatable/data.table/issues/7434

jangorecki:
I am not sure if adding new arg is a good way to go here.

Another option, then mentioned before, is allowing check.attributes to be character, a names of attributes that needs to be checked. And then check.attributes="key" will check key but not index. I think it should cover majority of the cases because usually there are not many attributes we want to compare.

mb706:
> names of attributes that needs to be checked

I would argue that it would be better to pass a vector of known-ignorable attributes than to pass a vector of known-important attributes. If a new attribute comes along (be it from data.table, or from user code that adds some attribute somewhere -- data.table respects user-defined attributes elsewhere, e.g. #995 ), that new attribute should be checked by default, not ignored by default, by code that was written with intent to ignore the index. I.e., in that case, I would prefer passing `ignore.attributes = "index"` to passing `check.attributes = c("sorted", <... whatever other attributes might come up>)`.

jangorecki:
No a bad idea

MichaelChirico:
> I would argue that it would be better to pass a vector of known-ignorable attributes than to pass a vector of known-important attributes.

FWIW, that's exactly the API of {testthat} (_via_ `waldo::compare(ignore_attr=)`):

https://waldo.r-lib.org/reference/compare.html

PR #6137, opened 2024-05-17
exclude test for systems without longdouble support

ben-schwen:
Found this test failing while developing with WSL1

No news since test only change


inst/tests/tests.Rraw
```diff
@@ -17878,7 +17878,9 @@ if (test_bit64) {
   test(2226.1, base::prod(2147483647L,2L), 4294967294)  # just to illustrate that base returns double
   DT = data.table(x=c(lim.integer64(), 2, 1, NA, NA, -2, 4), g=INT(1,2,1,2,1,2,3,3))
   test(2226.2, DT[, prod(x), g],            data.table(g=1:3, V1=as.integer64(c(NA,NA,-8L))))
-  test(2226.3, DT[, prod(x,na.rm=TRUE), g], data.table(g=1:3, V1=as.integer64(c(NA,"9223372036854775807",-8L))))
+  if (test_longdouble) {
```
MichaelChirico, line 17881: what happens on `!longdouble` systems?

I saw a segfault in the `r-cmd-check-occasional` logs yesterday, not sure if it's related:

https://github.com/Rdatatable/data.table/actions/runs/9140204969/job/25133296565

Probably best to leave the reasoning here in a comment.

ben-schwen: On my `!longdouble` with 53 long doubles the values of `lim.integer64()` get rounded to -(2^63) resp. 2^63 and then converted to `NA` here:

https://github.com/Rdatatable/data.table/blob/4045e8bafe2d569de37ac22a71bc0774f6c82526/src/gsumm.c#L1167

src/gsumm.c
```diff
@@ -1139,8 +1139,10 @@ SEXP gprod(SEXP x, SEXP narmArg) {
         const int64_t elem = nosubset ? xd[i] : (irows[i]==NA_INTEGER ? NA_INTEGER64 : xd[irows[i]-1]);
         if (elem==NA_INTEGER64) {
           if (!narm) s[thisgrp] = NA_REAL;
+          Rprintf("CONT: %d %Lf\n", i, s[thisgrp]);
```
MichaelChirico, line 1142: looks like this was included unintentionally

inst/tests/tests.Rraw
MichaelChirico: thanks!

MichaelChirico (approved):
ready to go after reverting incidental change

MichaelChirico (approved):
thanks!

PR #6138, opened 2024-05-17
More changes to occasional CI

MichaelChirico:
Latest failures here:

https://github.com/Rdatatable/data.table/actions/runs/9044008315/job/24852194788

There's still a lot wrong, trying to fix a few where I see the fix easily.


tdhock:
lgtm thanks

PR #6142, opened 2024-05-19
More fixes to occasional CI

MichaelChirico:
Today's failures:

https://github.com/Rdatatable/data.table/actions/runs/9140204969

 - Use `_R_CHECK_FORCE_SUGGESTS_=false` for e.g. `xts` which now has a R>=3.6.0 dependency


MichaelChirico:
A few steps closer :)

PR #6144, opened 2024-05-20
better locale robustness

### LINKED ISSUE #6140: coerceAs may be doing factor coercion incorrectly
As seen in `r-cmd-check-occasional`:

https://github.com/Rdatatable/data.table/blob/c8c35f3dc23b5d4ad6d6fce7bb624f349486480c/inst/tests/nafill.Rraw#L277

When run in a locale where `y` sorts before `x` we get:

```
> x = coerceAs(1:2, factor(c("x", "y"))) 
First 2 of 2 (type 'integer'): 
[1] y x
Levels: y x
> y = factor(c("x", "y")) 
First 2 of 2 (type 'integer'): 
[1] x y
Levels: y x
2 string mismatches
```

I am not sure the intended behavior in this case, but that output does look potentially off to me. We can fix the test if the behavior is intentional, want to check with Jan first.

MichaelChirico:
Closes #6140; follow-up to #6074

Looking more carefully, `coerceAs()` will receive a factor as-is from R unless it's doing some lazy evaluation. It's only job is to map what its input into the factor structure it receives.

So I don't think `coerceAs()` is doing anything wrong, rather, the test is poorly structured:

```r
# return a vector where the first element is the first level of the factor, the second is the second level
coerceAs(1:2, factor(c("x","y")))
# coerce c("x", "y") to a factor, picking up levels automatically
factor(c("x","y"))
```

i.e. in the second case, 'x' will always come first regardless of which level it is, while in the first, the first level always comes first.

Therefore I think the proposed edit (to always supply levels explicitly) is appropriate -- it abstracts the base-owned logic of generating levels away.

An uglier alternative is to use `coerceAs(order(c('x', 'y')), factor(c('x', 'y')))`. I don't see any real advantage of this approach.

### DIFF
--- a/inst/tests/nafill.Rraw
+++ b/inst/tests/nafill.Rraw
@@ -290,7 +290,7 @@ test(10.16, coerceAs(1:2, xy_factor), xy_factor, output="integer[integer] into i
 test(10.17, coerceAs(1:3, xy_factor), output="integer[integer] into integer[factor]", error="factor numbers.*3 is outside the level range")
 test(10.18, coerceAs(c(1,2,3), xy_factor), output="double[numeric] into integer[factor]", error="factor numbers.*3.000000 is outside the level range")
 test(10.19, coerceAs(factor("x"), xy_factor), factor("x", levels=c("x","y")), output="integer[factor] into integer[factor]")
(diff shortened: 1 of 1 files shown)


jangorecki (approved):
LGTM

PR #5867, opened 2023-12-31
Make test robust to R configuration/version/platform issue

MichaelChirico:
Closes #5866

@jameslamb are you able to test the conda build against this PR? Would be great to confirm it's WAI on your end. If so we can merge and try to have this shipped in CRAN version 1.15.0.

PR #6146, opened 2024-05-20
Warning will not throw again now that it's captured in base_messages

### LINKED ISSUE #6133: Test suite error for i386 architecture
Hi,

after upgrading the Debian packaged data.table to version 1.15.4 our CI tests uncovered a test error for i386 architecture.  (The previously packaged version was 1.14.10 where this problem did not happened.)    We are running [this script in CI](https://salsa.debian.org/r-pkg-team/r-cran-data.table/-/blob/master/debian/tests/run-unit-test?ref_type=heads) which for the problematic error boils down to

    LC_ALL=C.UTF-8 R --no-save < tests/main.R

The issue is reported [in the Debian bug tracking system](https://bugs.debian.org/1070839) and points to a [complete log](https://ci.debian.net/packages/r/r-cran-data.table/testing/i386/46485246/) where you can find starting at line 908:

    Fri May 10 06:13:18 2024  endian==little, sizeof(long double)==12, longdouble.digits==64, sizeof(pointer)==4, TZ==unset, Sys.timezone()=='Etc/UTC', Sys.getlocale()=='LC_CTYPE=C.UTF-8;LC_NUMERIC=C;LC_TIME=C.UTF-8;LC_COLLATE=C.UTF-8;LC_MONETARY=C.UTF-8;LC_MESSAGES=C.UTF-8;LC_PAPER=C.UTF-8;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=C.UTF-8;LC_IDENTIFICATION=C', l10n_info()=='MBCS=TRUE; UTF-8=TRUE; Latin-1=FALSE; codeset=UTF-8', getDTthreads()=='OpenMP version (_OPENMP)==201511; omp_get_num_procs()==2; R_DATATABLE_NUM_PROCS_PERCENT==unset (default 50); R_DATATABLE_NUM_THREADS==unset; R_DATATABLE_THROTTLE==unset (default 1024); omp_get_thread_limit()==2147483647; omp_get_max_threads()==2; OMP_THREAD_LIMIT==unset; OMP_NUM_THREADS==unset; RestoreAfterFork==true; data.table is using 1 threads with throttle==1024. See ?setDTthreads.', zlibVersion()==1.3 ZLIB_VERSION==1.3
    Error: 1 error(s) out of 9702. Search tests/tests.Rraw.bz2 for test number(s) 2150.025. Duration: 35.0s elapsed (34.1s cpu).
    Execution halted

So the problem seems to be in `inst/tests/tests/Rraw.bz2` with the according code

    Sys.unsetenv("TZ")
    # Notes:
    #  - from v1.14.0 tz="" needed
    #  - as.POSIXct puts "" on the result (testing the write.csv version here with missing tzone)
    #  - Per #5866, this test is complicated by some interaction with very specific builds of R where
    #    as.POSIXct() might fail for very old timestamps when 'tz=""' is used. On such systems,
    #    as.POSIXct() failure means 'times' is returned as a character, hence no 'tzone' attribute.
    #    fread() will also throw a warning, one substring of which will be the reproduced base R error.
    test(2150.025,
         attr(fread(tmp, colClasses=list(POSIXct="times"), tz="")$times, "tzone"),
         if (is.null(base_messages$maybe_invalid_old_posixct)) "" else NULL,
         warning=base_messages$maybe_invalid_old_posixct)

As you can see on the [Debian CI matrix](https://ci.debian.net/packages/r/r-cran-data.table/) the problem seems to exist for 32bit architectures.

I hope the log files with all installed package versions are sufficient to track down the problem.

Kind regards
      Andreas.

MichaelChirico:
Closes #6133 

Since the warning from R is only thrown once per session, there will be no `warning=` to match, since the warning is already generated when `base_messages$maybe_invalid_old_posixct` is thrown.

@tillea are you able to test that this PR resolves the issue?


MichaelChirico:
Thanks Andreas, could you elaborate on the urgency of a fix here?

My instinct is to just include this PR with the next data.table release rather than triggering a patch release. That would happen sometime in July.

tillea:
Am Mon, May 20, 2024 at 02:18:43PM -0700 schrieb Michael Chirico:
> Thanks Andreas, could you elaborate on the urgency of a fix here?

Its pretty urgent since data.table has a lot of dependencies and these
are all in trouble currently.
 
> My instinct is to just include this PR with the next data.table release rather than triggering a patch release. That would happen sometime in July.

It would be of great help if you could release something in May.

Thanks a lot for your cooperation, Andreas.

tdhock (commented):
I'm not sure I understand -- is there some way we can alter our CI to catch this kind of issue, or is it too obscure?

MichaelChirico:
> Its pretty urgent since data.table has a lot of dependencies and these are all in trouble currently.

Do I understand that this only affects 32-bit architectures?

Is there an issue on another tracker we might be missing? It would seem that there's been no other requests to have this issue addressed in 2 months since 1.15.4 was released. Are we sure this is affecting many users?

If I'm not mistaken, and please correct me if I'm wrong, but that test is the same on 1.15.0, so that would make it since Jan 30 that we haven't seen any other issues filed. Or perhaps there was only an attempt to ship 1.15.0+ version of data.table on your end starting very recently & it's not picked up automatically from CRAN updates?

eddelbuettel:
I can apply to two-line patch to the (Debian-internal) git repo of the package.  We apply upstream (interim) all the time time so I am not sure I understand why a new release would be required to fix a Debian-only issue (on a non-release architecture).

eddelbuettel:
Uggh.  There is apparently _another_ autotest regression on armhf, see [#1071737](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1071737).

ben-schwen:
@eddelbuettel sorry to bother you, but can you also include #6137 in your patch? This should do the trick!

eddelbuettel:
Yes I realized that that was the likely course of action. But will await test result on first patch.

eddelbuettel:
Looks like it failed again on i386:  https://ci.debian.net/packages/r/r-cran-data.table/testing/i386/47062403/

This is not my package, and I (by my choice) do _not_ run autopkgtest on the 100 or so CRAN packages I look after in Debian as I trust CRAN.  So if it were my package I'd turn the autopkgtests off.

eddelbuettel:
@ben-schwen et al:  Advice on how to uncork the [current autopkgtest error on i386](https://ci.debian.net/packages/r/r-cran-data.table/testing/i386/47062403/)? I have a hard time reading the (somewhat non-standard) test runners you use and I am also unfamiliar with the autopkgtest framework (as I do not bother with for the CRAN packages I maintain).  Advice would be appreciated, the package still has a target data of June 11 on its back for removal from Debian unstable.

ben-schwen:
@eddelbuettel AFAIU here is the timeline:
1) Michael added a layer of complexity to this test in #5867
2) He also refactored our test suite in #6029 to make it more robust, hence, the initial change of #5867 to the failing test `2150.025` is not necessary any more

I resetted the test to its initial behavior and also made a conda environment as suggested by @jameslamb in https://github.com/Rdatatable/data.table/issues/5867#issuecomment-1874771367, but couldn't reproduce the TZ problem reported there (the newly created conda environment gave me only `R 4.2` and not `R 4.2.3`)

eddelbuettel:
( Debian uses R 4.4.0, not 4.2.3 )

What changes do you suggest I monkey-patch the bz2'ed test file?

ben-schwen:
I assumed we are patching from `1.15.4` and created a branch for that https://github.com/Rdatatable/data.table/compare/1.15.4...debian_patch

But the test with the `test_longdouble` problem should have already failed with `1.15.0` and `1.15.2` on armhf?

eddelbuettel:
Missed the branch.  If I may, you are looking at the wrong page.  While Debian has 1.14.10 in testing the fight we are fight here is on the belatedly-added (after I filed a bugreport) 1.15.4 (and .0 and .2 where skipped) which is now failing the requored steps in 'unstable' to migrate to 'testing' where it would replace 1.14.10.  Pages I have open include

- the i386 bug report https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1070839
- a i386 fail with AFAICT is the latest https://ci.debian.net/packages/r/r-cran-data.table/testing/i386/47062403/
- the autopkgtest summary with a mix of green and red and (damnit!) fresh red on -2 for loong64 https://ci.debian.net/packages/r/r-cran-data.table/
- the overall build status (very green, this does not involve the tests) https://buildd.debian.org/status/package.php?p=r-cran-data.table
- the tracker page with bth 1.15.4 under unstable (left column) and the 'fail' summary in the middle https://tracker.debian.org/pkg/r-cran-data.table

eddelbuettel:
Thanks @ben-schwen -- I 'manually' adjusted the already patches file to that new branch and uploaded a new version.  We should know in a day or two if this gets us over the goal line.

ben-schwen:
> Thanks @ben-schwen -- I 'manually' adjusted the already patches file to that new branch and uploaded a new version. We should know in a day or two if this gets us over the goal line.

@eddelbuettel there seems to be a typo (ttest instead of test) with test 2226

eddelbuettel:
@ben-schwen et al: No luck.  Now it dies oon amd64 (!!) over

```
 88s Error in ttest(2226.3, DT[, prod(x, na.rm = TRUE), g], data.table(g = 1:3,  : 
 88s   could not find function "ttest"
```

Logs at https://ci.debian.net/packages/r/r-cran-data.table/unstable/amd64/47098902/

eddelbuettel:
@ben-schwen Yes, just went checking myself (before checking GitHub notifications).  So that is likely me having fat-fingered it?  If so can fix.

(It is now named 1.15.4-2.1 as I opted for a so-called non-maintainer upload so more clearly signal that I am not really involved with the package.)

eddelbuettel:
@ben-schwen That works -- the package is now back in 'testing'.  Thanks for all the help.  Will you merge the change for any next release?

ben-schwen:
The armhf fix is already in master #6137 (we found that one 2 weeks ago) and once this PR is merged, both of the manual patches will be in the next release.

@eddelbuettel I saw that we are failing on loong64 due to libcurl4, is that something we can ignore?

eddelbuettel:
Yes, apparently looong64 is not what is called a 'release architecture' so the fail there does not matter.  The transition actually happened today (but not without at the last moment triggering mass email to the maintainer of all packages dependending on r-cran-data.table because of 'pending autoremoval on June 11' -- I got ten or so).  

So I think we really are good and done here.

Thanks a bloody much for the help (and my vote for keeping the file in the source tarball uncompressed as in the repo).

tillea:
Am Sat, Jun 01, 2024 at 09:21:23AM -0700 schrieb Dirk Eddelbuettel:
> Thanks a bloody much for the help (and my vote for keeping the file in the source tarball uncompressed as in the repo).
Thanks to you both (and I also like the uncompressed file in tarball).  Andreas.

MichaelChirico:
Thanks a ton @eddelbuettel for the assist here, your expertise is much appreciated.

> Ah. I missed that the test file is bz2 compressed.

Yes, this is a CRAN thing to keep our binary a bit smaller. We still get the "5MiB" `NOTE` on certain builds. Alas.

Is there any way to ship to Debian a different binary than CRAN (Or is that a bad idea anyway)?

Or maybe we should revisit just saying "don't care" about the "too large" CRAN scold.

MichaelChirico:
Merging this to `master` given the reviews+confirmation here, thanks all!

eddelbuettel:
> Is there any way to ship to Debian a different binary than CRAN (Or is that a bad idea anyway)?

A package maintainer can easily accomodate such differences in the _installed_ package but we generally take _sources_ as given.

PR #6151, opened 2024-05-27
Update fifelse documentation to reflect evaluation behavior

### LINKED ISSUE #4549: fifelse did not work in recursive function
I am trying to use `data.table::fifelse` to implement a recursive function to calculate greatest common divisors mentioned in this [StackOverflow](https://stackoverflow.com/questions/21502181/finding-the-gcd-without-looping-r). However, I found that in this case, `data.table::fifelse` cannot be a replacement for `base::ifelse`.

## Minimal reproducible

```r
# use base::ifelse
gcd_base <- function(x,y) {
  r <- x%%y;
  return(ifelse(r, gcd(y, r), y))
}

gcd_dt <- function(x,y) {
  r <- x%%y;
  return(data.table::fifelse(r, gcd(y, r), y))
}

gcd_base(10, 1)
# [1] 1
gcd_dt(10, 1)
# Error: C stack usage  7971780 is too close to the limit
```

Any insights? Thanks!

joshhwuu:
Closes #4549

`fifelse` evaluates both `yes` and `no` arguments, regardless of the value of `test`. This makes it unfit for recursive functions as the evaluation will cause an infinite loop of calls.

Changed documentation to better reflect this behavior, along with a simple example to showcase the difference against base `ifelse`:

```R
ifelse(1 == 1, print("yes"), print("no"))
# [1] "yes"
fifelse(1 == 1, print("yes"), print("no"))
# [1] "yes"
# [1] "no"
```
Also mentioned that `fcase` can be used instead for recursive calls. Does `fcase` documentation also need updating? I noticed that the `when` and `value` pair is evaluated, even if the `when` doesn't bind, which means:

```R
gcd_dt <- function(x,y) {
  r <- x%%y;
  return(data.table::fcase(!r, y, r, gcd_dt(x, y)))
}
```
works, but

```R
gcd_dt <- function(x,y) {
  r <- x%%y;
  return(data.table::fcase(!r, y, r, gcd_dt(x, y))) # (when value pair swapped)
}
```
doesn't, due to the same reason that `fifelse` doesn't.

### DIFF
--- a/man/fcase.Rd
+++ b/man/fcase.Rd
@@ -63,8 +63,7 @@ fcase(
 # fcase can be used for recursion, unlike fifelse
 gcd_dt = function(x,y) {
   r = x\%\%y
-  # Recursive call must be in the last when-value pair
-  return(fcase(!r, y, r, gcd_dt(x, y)))
+  fcase(!r, y, r, gcd_dt(x, y)) # Recursive call must be in the last when-value pair
 }
 gcd_dt(10L, 1L)
 }
--- a/man/fcase.Rd
+++ b/man/fcase.Rd
@@ -12,9 +12,9 @@
 \item{default}{ Default return value, \code{NA} by default, for when all of the logical conditions \code{when1, when2, ..., whenN} are \code{FALSE} or missing for some entries. }
 }
 \details{
-	\code{fcase} evaluates each when-value pair in order, until it finds a \code{when} that is \code{TRUE}. It then returns the corresponding \code{value}. During evaluation, \code{value} will be evaluated regardless of whether the corresponding \code{when} is \code{TRUE} or not, which means recursive calls should be placed in the last when-value pair, see \code{Examples}.
+\code{fcase} evaluates each when-value pair in order, until it finds a \code{when} that is \code{TRUE}. It then returns the corresponding \code{value}. During evaluation, \code{value} will be evaluated regardless of whether the corresponding \code{when} is \code{TRUE} or not, which means recursive calls should be placed in the last when-value pair, see \code{Examples}.
 
-	\code{default} is always evaluated, regardless of whether it is returned or not.
+\code{default} is always evaluated, regardless of whether it is returned or not.
 }
 \value{
   Vector with the same length as the logical conditions (\code{when}) in \code{...}, filled with the corresponding values (\code{value}) from \code{...}, or eventually \code{default}. Attributes of output values \code{value1, value2, ...valueN} in \code{...} are preserved.
--- a/man/fcase.Rd
+++ b/man/fcase.Rd
@@ -61,10 +61,11 @@ fcase(
 )
 
 # fcase can be used for recursion, unlike fifelse
-gcd_dt <- function(x,y) {
-  r <- x%%y;
-  return(fcase(!r, y, r, gcd_dt(x, y))) # Recursive call must be in the last when-value pair
+gcd_dt = function(x,y) {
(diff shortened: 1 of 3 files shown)


man/fifelse.Rd
```diff
@@ -15,12 +15,16 @@
 }
 \details{
 In contrast to \code{\link[base]{ifelse}} attributes are copied from the first non-\code{NA} argument to the output. This is useful when returning \code{Date}, \code{factor} or other classes.
+
+\code{fifelse}, contrary to \code{\link[base]{ifelse}}, evalutes both \code{yes} and \code{no} arguments for type checking regardless the result of \code{test}. This means that neither \code{yes} nor \code{no} should be recursive function calls, use \code{fcase} instead.
```
Anirban166, line 19: evaluates*
regardless of*
semicolon is more appropriate instead of comma in between 'This means that neither yes nor no should be recursive function calls' and 'use fcase instead'
Also 'fifelse, contrary to ifelse, ...' can be rephrased to something simpler like 'Unlike base ifelse, fifelse evaluates ...'

man/fifelse.Rd
```diff
@@ -15,12 +15,16 @@
 }
 \details{
 In contrast to \code{\link[base]{ifelse}} attributes are copied from the first non-\code{NA} argument to the output. This is useful when returning \code{Date}, \code{factor} or other classes.
+
+Unlike \code{\link[base]{ifelse}}, \code{fifelse} evaluates both \code{yes} and \code{no} arguments for type checking regardless of the result of \code{test}. This means that neither \code{yes} nor \code{no} should be recursive function calls. For recursiveness, use \code{fcase} instead.
```
tdhock, line 19: recursiveness -> recursion

tdhock: yes please add an example of recursion with fcase, to clarify what you mean here

tdhock (changes requested):

joshhwuu:
From R-CMD-Check:
```R
  > # fcase can be used for recursion, unlike fifelse
  > gcd_dt <- function(x,y) {
  +   r <- x
  +   return(fcase(!r, y, r, gcd_dt(x, y))) # Recursive call must be in the last when-value pair
  + }
  > gcd_dt(10, 1)
  Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
  Execution halted
```
not sure why mod operator isn't showing up here, will investigate

man/fcase.Rd
```diff
@@ -54,5 +59,12 @@ fcase(
 	x > 5L, 3L,
 	default = 5L
 )
+
+# fcase can be used for recursion, unlike fifelse
+gcd_dt <- function(x,y) {
+  r <- x%%y;
+  return(fcase(!r, y, r, gcd_dt(x, y))) # Recursive call must be in the last when-value pair
```
tdhock: please remove return (not necessary)

tdhock (changes requested):

man/fcase.Rd
```diff
@@ -54,5 +59,12 @@ fcase(
 	x > 5L, 3L,
 	default = 5L
 )
+
+# fcase can be used for recursion, unlike fifelse
+gcd_dt = function(x,y) {
```
tdhock, line 64: i guess this is greatest common denominator? can you please add a comment to explain?

tdhock (changes requested):

tdhock (approved):
looks good to me

Issue #6153, opened 2024-05-27
Debian testing failure for gprod test


MichaelChirico:
Uggh.  There is apparently _another_ autotest regression on armhf, see [#1071737](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1071737).

_Originally posted by @eddelbuettel in https://github.com/Rdatatable/data.table/issues/6146#issuecomment-2134088259_

MichaelChirico:
cc @ben-schwen.

From the failure log, it looks like the testing environment has a strange setup?

```
sizeof(long double)==8,
longdouble.digits==, sizeof(pointer)==4, TZ==unset
```

`longdouble.digits` is empty? For me that's

```r
.Machine$longdouble.digits
# [1] 64
```

That test is behind `if (test_longdouble)` which in turn examines `capabilities("long.double")`. Any idea what might be the reason @eddelbuettel?

eddelbuettel:
Sorry, I wish I could help but I have particular insight here.  My world is amd64.  Debian has porter teams.  Maybe the Debian maintainers for data.table can help you here and contact them -- I was just trying to unplug this situation as the data.table thread of removal also threatens 10+ packages I look after.

MichaelChirico:
From how `.Machine` is populated, it's strange to see nothing in `longdouble.digits`:

https://github.com/r-devel/r-svn/blob/5c0e367225ec0ed9e1a95864302a6311708533ee/src/main/machine.c#L186

That `#ifdef HAVE_LONG_DOUBLE` branch is the same as populates `.Machine$sizeof.longdouble`. So there should be _some_ integer there. I don't follow what this routine is doing to define that `it` value that winds up as `longdouble.digits`:

https://github.com/r-devel/r-svn/blob/5c0e367225ec0ed9e1a95864302a6311708533ee/src/main/machar.c#L95

Anyway, time to follow up with the maintainers.

eddelbuettel:
For some (but not all?) architectures we can get shell access, the table is [here](https://db.debian.org/machines.cgi?sortby=purpose&sortorder=dsc) and the first one on the list armhf.  So if guys have something testable I can probably find a moment to run a quick test.

MichaelChirico:
Hmm just reporting back `.Machine` and `capabilities()` on that system would be helpful.

As a workaround we can probably just make `test_longdouble` also look at `length(.Machine$longdouble.digits) > 0L` also.

eddelbuettel:
[...]

PR #6154, opened 2024-05-27
Slightly more robust pull of .Machine entries

MichaelChirico:
Related to #6153, getting `longdouble.digits==` is not the most helpful, I think there's a couple ways we could get that. I assume based on the base R code that we're seeing `NULL` there, but this will help clarify the actual value / rely less on inference.

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -225,10 +225,12 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
   # does show the full file output these days, so the 13 line limit no longer bites so much. It still bit recently
   # when receiving output of R CMD check sent over email, though.
   tz = Sys.getenv("TZ", unset=NA)
+
+  maybe_unknown = function(x) if (length(x)) x else "<unknown>"
   cat("\n", date(),   # so we can tell exactly when these tests ran on CRAN to double-check the result is up to date
     "  endian==", .Platform$endian,
-    ", sizeof(long double)==", .Machine$sizeof.longdouble,
-    ", longdouble.digits==", .Machine$longdouble.digits, # 64 normally, 53 for example under valgrind where some high accuracy tests need turning off, #4639
+    ", sizeof(long double)==", maybe_unknown(.Machine$sizeof.longdouble),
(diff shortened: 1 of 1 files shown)


MichaelChirico:
`format()` will at least let us distinguish `integer()` from `NULL`.

ben-schwen:
Maybe also add `capabilities()["long.double"]`?

MichaelChirico:
I _guess_ that's overkill, here's how that element of `capabilities()` is defined:

https://github.com/r-devel/r-svn/blob/e0f5c5e1c60bca2b7b2044eb0b4770e287cd770e/src/main/platform.c#L2627-L2628

Here's the `LDOUBLE` macro:

https://github.com/r-devel/r-svn/blob/1f537a2469956b505ad10135eeee282fe33a8475/src/include/Defn.h#L2407-L2411

In turn I think `HAVE_LONG_DOUBLE` gets set by `./configure`.

Here's how `.Machine$sizeof.longdouble` is defined:

https://github.com/r-devel/r-svn/blob/1f537a2469956b505ad10135eeee282fe33a8475/src/main/machine.c#L139-L144

Here's the `SIZEOF_LONG_DOUBLE` macro:

https://github.com/r-devel/r-svn/blob/1f537a2469956b505ad10135eeee282fe33a8475/src/gnuwin32/fixed/h/config.h#L1182-L1191

So there might be _some_ daylight between the two, WDYT?

R/test.data.table.R
```diff
@@ -227,8 +227,8 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
   tz = Sys.getenv("TZ", unset=NA)
   cat("\n", date(),   # so we can tell exactly when these tests ran on CRAN to double-check the result is up to date
     "  endian==", .Platform$endian,
-    ", sizeof(long double)==", .Machine$sizeof.longdouble,
-    ", longdouble.digits==", .Machine$longdouble.digits, # 64 normally, 53 for example under valgrind where some high accuracy tests need turning off, #4639
+    ", sizeof(long double)==", format(.Machine$sizeof.longdouble),
```
MichaelChirico, line 230:
```suggestion
    ", sizeof(long double)==", format(.Machine$sizeof.longdouble),
    ", capabilities('long.double')==", capabilities('long.double'), # almost certainly overkill, but that's OK; see #6154
```

PR #6158, opened 2024-05-29
fread blank.lines.skip gains new value "none"

### LINKED ISSUE #5611: blank.lines.skip won't work when the top lines is empty
I want to keep the total line number. Following code should return 3 rows, but it always return 1 row
```
file2 <- tempfile()
writeLines(c("", "", NA), file2)
data.table::fread(file2, blank.lines.skip = FALSE)
```
image

joshhwuu:
WIP towards #5611 

Closes #5611

Currently, `fread()` starts reading at the first non-empty line of the file, regardless of `blank.lines.skip`, this aims to add a new value for `blank.lines.skip`, allowing users to skip "none" of the blank lines (read all lines of the file).

old:
```R
file <- tempfile()
writeLines(c("", "", NA), file)
data.table::fread(file, blank.lines.skip = FALSE)
#>        V1
#>    <lgcl>
#> 1:     NA
```
new:
```R
file <- tempfile()
writeLines(c("", "", NA), file)
data.table::fread(file, blank.lines.skip = "none")
#>        V1
#>    <lgcl>
#> 1:     NA
#> 2:     NA
#> 3:     NA
```
This does mean that users should be able to read in fully white/empty files if `blank.lines.skip == "none"`, at least that behavior made sense to me.

Marking this as WIP as there are two lines I'm unsure of:
```R
ASSERT(lastncol>0, "first non-empty row should always be at least one field; %c %d", sep, quoteRule); // # nocov

if (!fill && tt!=ncol) STOP(_("Internal error: first line has field count %d but expecting %d"), tt, ncol); // # nocov
```
these two assertions don't work with the changes for some reason I'm not so sure about, so I got around them by only having the assertions when `blank.lines.skip != "none"`, but this feels a bit hacky and I'm not sure what kind of error these two assertions are supposed to check for. If someone is able to help me out with these that would be greatly appreciated!

TODO:
- [ ] More tests
- [ ] Figure out assertion issues
- [ ] Documentation + cleanup


joshhwuu:
@ben-schwen Could you possibly take a look at this when you have time? Not sure if this was what you had in mind but it made sense to me, let me know if you have any suggestions or if you can provide some insight on the issue I have rn, thanks!

Anirban166:
> ```r
> ASSERT(lastncol>0, "first non-empty row should always be at least one field; %c %d", sep, quoteRule); // # nocov
> 
> if (!fill && tt!=ncol) STOP(_("Internal error: first line has field count %d but expecting %d"), tt, ncol); // # nocov
> ```
> 
> these two assertions don't work with the changes for some reason I'm not so sure about, so I got around them by only having the assertions when `blank.lines.skip != "none"`, but this feels a bit hacky and I'm not sure what kind of error these two assertions are supposed to check for. If someone is able to help me out with these that would be greatly appreciated!

Thoughts regarding the first assertion - Since `fread` is not skipping empty lines and we are dealing with blank data for the first two rows, it might be that it is somehow considering the first blank line as a valid row (bug?), in which case the number of fields in it are zero (causing the first assertion to fail)

Btw, the variable names seem to be a bit confusing with regards to the error message (I'd expect `lastncol` to be something like the number of columns in the last non-empty row that has been read for example)

Anirban166:
Also, is the option `blank.lines.skip` primarily for performance improvement? (since irrespective of it being `TRUE` or `FALSE` I get the same output for the 'old code' example)

Anirban166:
Also, is the option `blank.lines.skip` primarily for performance improvement? (since irrespective of it being `TRUE` or `FALSE` I get the same output for the 'old' code example)

joshhwuu:
> Thoughts regarding the first assertion - Since `fread` is not skipping empty lines and we are dealing with blank data for the first two rows, it might be that it is somehow considering the first blank line as a valid row (bug?), in which case the number of fields in it are zero (causing the first assertion to fail)

This was my theory too

> Also, is the option blank.lines.skip primarily for performance improvement? (since irrespective of it being TRUE or FALSE I get the same output for the 'old' code example)

Not sure what the intended purpose is but I believe `blank.lines.skip` doesnt count NA as blank lines since its an entry, and with the old behavior (skipping all leading empty rows by default) either true or false produces the same result, try something like:

file <- tempfile()
writeLines(c("", "", NA, "", "a"), file)

joshhwuu:
```R
Test 2264.1 produced 1 errors but expected 0
 Expected: 
 Observed: Internal error: first line has field count 0 but expecting 1
 Test 2264.2 produced 1 errors but expected 0
 Expected: 
 Observed: Internal error: first line has field count 0 but expecting 1
 Test 2264.3 produced 1 errors but expected 0
 Expected: 
 Observed: Internal error: first line has field count 0 but expecting 1
```

Looks like both assertions are very similar, failing because they expect the first line to have at least one field. I wonder what part of fread logic requires the first row to be non-empty. Perhaps for column writing?

joshhwuu:
> Also, is the option `blank.lines.skip` primarily for performance improvement? (since irrespective of it being `TRUE` or `FALSE` I get the same output for the 'old' code example)

@Anirban166, looks like `blank.lines.skip` is automatically TRUE whenever we have ncol > 1, not sure why but I'm getting the similar output for both true and false, except when `blank.lines.skip=TRUE` we keep the column names.

```R
f = tempfile()
input = "a,b\n\n\n1,3\n2,5\n4,6"
writeLines(input, f)
fread(f, blank.lines.skip=T)
#    a b
# 1: 1 3
# 2: 2 5
# 3: 4 6

fread(f, blank.lines.skip=F)
#    V1 V2
# 1:  1  3
# 2:  2  5
# 3:  4  6
```
Not sure if this is intended

PR #6159, opened 2024-05-29
Update strip.white documentation in fread

### LINKED ISSUE #2492: fread's skip.white=FALSE behavior
The documentation for this parameter says the following:
```
strip.white: default is ‘TRUE’. Strips leading and trailing whitespaces
          of unquoted fields. If ‘FALSE’, only header trailing spaces
          are removed.
```
So, it appears that, according to the documentation, when the flag is false the white space should be left intact EXCEPT for the trailing whitespace in the headers (and it is unclear whether the documentation talks about all fields in the header, or just the last one). I'm not sure why such exception is warranted, but I tried to see how it works:

(1) It appears that the trailing whitespace in the header isn't removed after all (and neither does the leading whitespace):
```
> data.table::fread("A  ,B  \n1,2\n3,4\n", strip.white=F) -> f
> colnames(f)
[1] "A  " "B  "
> data.table::fread("  A,  B\n1,2\n3,4\n", strip.white=F) -> f
> colnames(f)
[1] "  A" "  B"
```
I think this is good, just the documentation needs to be corrected.

(2) Now what about the data? It appears the flag is not respected when the data is numeric:
```
> data.table::fread("A,B,C,D\n  1.0  ,  2  ,  x , true \n  3.7  ,  4  ,  y\t, false   \n", strip.white=F) -> f
> str(f)
Classes ‘data.table’ and 'data.frame':	2 obs. of  4 variables:
 $ A: num  1 3.7
 $ B: int  2 4
 $ C: chr  "  x " "  y\t"
 $ D: logi  TRUE FALSE
```
Thus, it appears the flag only applies to character fields and is ignored for all others. I'm not sure whether this is the intentional behavior or not, but the documentation doesn't mention it at all...

(Update) Cross-checking with the documentation of `read.csv`, they mention the following:
```
strip.white: logical. Used only when ‘sep’ has been specified, and
          allows the stripping of leading and trailing white space from
          unquoted ‘character’ fields (‘numeric’ fields are always
          stripped).
```
So it appears that the behavior of `fread` is (almost) consistent with that of `read.csv`, and then it's just the documentation issue. The only discrepancy in the behavior is that `read.csv` strips both spaces and tabs, while `fread` only spaces.

joshhwuu:
Closes #2492

Small update to `fread` documentation:

```R
d = fread("A  ,B  \n1,2\n3,4\n", strip.white=F)
colnames(d)
# [1] "A  " "B  "
d = fread("  A,  B\n1,2\n3,4\n", strip.white=F)
colnames(d)
# [1] "  A" "  B"
```
This behavior was updated a while back, probably as a byproduct of another change, but the documentation was never updated


Anirban166 (approved):

man/fread.Rd
```diff
@@ -52,7 +52,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC"
...
   \item{encoding}{ default is \code{"unknown"}. Other possible options are \code{"UTF-8"} and \code{"Latin-1"}.  Note: it is not used to re-encode the input, rather enables handling of encoded strings in their native encoding. }
   \item{quote}{ By default (\code{"\""}), if a field starts with a double quote, \code{fread} handles embedded quotes robustly as explained under \code{Details}. If it fails, then another attempt is made to read the field \emph{as is}, i.e., as if quotes are disabled. By setting \code{quote=""}, the field is always read as if quotes are disabled. It is not expected to ever need to pass anything other than \"\" to quote; i.e., to turn it off. }
-  \item{strip.white}{ default is \code{TRUE}. Strips leading and trailing whitespaces of unquoted fields. If \code{FALSE}, only header trailing spaces are removed. }
+  \item{strip.white}{ default is \code{TRUE}. If \code{TRUE}, strips leading and trailing whitespaces from unquoted \code{"character"} fields. \code{"Numeric"} fields are always stripped of leading and trailing whitespaces.}
```
tdhock, line 55: why did you use captital N in Numeric?
if no good reason, I would suggest lower-case numeric for consistency with for example as.numeric

joshhwuu: Right, that makes sense. Thanks!

man/fread.Rd
```diff
@@ -52,7 +52,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC"
...
   \item{encoding}{ default is \code{"unknown"}. Other possible options are \code{"UTF-8"} and \code{"Latin-1"}.  Note: it is not used to re-encode the input, rather enables handling of encoded strings in their native encoding. }
   \item{quote}{ By default (\code{"\""}), if a field starts with a double quote, \code{fread} handles embedded quotes robustly as explained under \code{Details}. If it fails, then another attempt is made to read the field \emph{as is}, i.e., as if quotes are disabled. By setting \code{quote=""}, the field is always read as if quotes are disabled. It is not expected to ever need to pass anything other than \"\" to quote; i.e., to turn it off. }
-  \item{strip.white}{ default is \code{TRUE}. Strips leading and trailing whitespaces of unquoted fields. If \code{FALSE}, only header trailing spaces are removed. }
+  \item{strip.white}{ default is \code{TRUE}. If \code{TRUE}, strips leading and trailing whitespaces from unquoted \code{"character"} fields. \code{"numeric"} fields are always stripped of leading and trailing whitespaces.}
```
MichaelChirico, line 55:
```suggestion
  \item{strip.white}{ Logical, default \code{TRUE}, in which case leading and trailing whitespace is stripped from unquoted \code{"character"} fields. \code{"numeric"} fields are always stripped of leading and trailing whitespace.}
```

MichaelChirico (approved):
Thanks!

PR #6160, opened 2024-05-29
adding an atime test case Performance Regression with .N and := #PR5463

DorisAmoakohene:
This atime test case discusses the issue of [Performance Regression with .N and := (https://github.com/https://github.com/Rdatatable/data.table/issues/5424)

The cause of the regression is related to the addition of the snprintf function in the assign.c. https://github.com/Rdatatable/data.table/pull/4491

The Regression was fixed by creating targetDesc function and adding snprintf in assign.c [Fixes Regression](https://github.com/Rdatatable/data.table/commit/e793f53466d99f86e70fc2611b708ae8c601a451)

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -123,7 +123,7 @@ test.list <- atime::atime_test_list(
 
   #issue reported in: https://github.com/Rdatatable/data.table/issues/5424
   #Fixed in: https://github.com/Rdatatable/data.table/pull/5463
-   "Performance Regression with .N and := #PR5463" = atime::atime_test(
+   ".N and := regression fixed in #5463" = atime::atime_test(
      N = 10^seq(1, 4),
      setup = {
        set.seed(123L)
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -119,6 +119,22 @@ test.list <- atime::atime_test_list(
       data.table:::setDT(L)
     },
     Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
-    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15") # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"),# Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+
+  #issue reported in: https://github.com/Rdatatable/data.table/issues/5424
+  #Fixed in: https://github.com/Rdatatable/data.table/pull/5463
+   "Performance Regression with .N and := #PR5463" = atime::atime_test(
+     N = 10^seq(1, 4),
+     setup = {
+       set.seed(123L)
+       dt <- data.table(
+         id = seq_len(N),
+         val = rnorm(N))
+       dt
+     },
+     expr = data.table:::`[.data.table`(dt, , .(vs = (sum(val))), by = .(id)),
+     Before = "be2f72e6f5c90622fe72e1c315ca05769a9dc854",
+     Regression = "e793f53466d99f86e70fc2611b708ae8c601a451",
+     Fixed = "58409197426ced4714af842650b0cc3b9e2cb842")
 )
 # nolint end: undesirable_operator_linter.


.ci/atime/tests.R
```diff
@@ -119,6 +119,22 @@ test.list <- atime::atime_test_list(
...
+       set.seed(123L)
+       dt <- data.table(
+         id = seq_len(N),
+         val = rnorm(N))
+       dt
+     },
+     expr = data.table:::`[.data.table`(dt, , .(vs = (sum(val))), by = .(id)),
+     Before = "be2f72e6f5c90622fe72e1c315ca05769a9dc854",
+     Regression = "e793f53466d99f86e70fc2611b708ae8c601a451",
+     Fixed = "58409197426ced4714af842650b0cc3b9e2cb842")
```
Anirban166, line 138: Please add comments beside 'Before', 'Regression' and 'Fixed' (like the other test cases) as @tdhock wants that (and wait till he reviews that part)

.ci/atime/tests.R
```diff
@@ -119,6 +119,22 @@ test.list <- atime::atime_test_list(
       data.table:::setDT(L)
     },
     Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
-    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15") # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"),# Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+
+  #issue reported in: https://github.com/Rdatatable/data.table/issues/5424
+  #Fixed in: https://github.com/Rdatatable/data.table/pull/5463
+   "Performance Regression with .N and := #PR5463" = atime::atime_test(
```
Anirban166, line 126: Based on the reviews I received from @tdhock on the three existing test cases, the title would be of the form '<Function/Operator/Symbol> improved in #PRNumber' for performance improvement cases ('Fast' and 'Slow' labels) or '<Function/Operator/Symbol> regression fixed in #PRNumber' for performance regressions ('Before', 'Regression', and 'Fixed' labels)

(also the term 'Performance' is omitted from the titles since it's a given and would be repeated for each case if we were to use it)

.ci/atime/tests.R
```diff
@@ -119,6 +119,22 @@ test.list <- atime::atime_test_list(
       data.table:::setDT(L)
     },
     Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
-    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15") # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"),# Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+
+  #issue reported in: https://github.com/Rdatatable/data.table/issues/5424
+  #Fixed in: https://github.com/Rdatatable/data.table/pull/5463
```
tdhock, line 125: why does line 122 say PR #5427 fixes the issue, but line 125 says this is fixed in #5463? Please double check / fix so that these comments are consistent.

tdhock (changes requested):

.ci/atime/tests.R
DorisAmoakohene: @tdhock the line 122 is different test which was added already done  by @Anirban166 and has been merged, I added test from line 124 and named ".N and := regression fixed in #5463", for review so that it can also be merged together with the first two

tdhock: ok

.ci/atime/tests.R
tdhock: yes please add comments

tdhock (changes requested):
also please next time make your branch name more informative.
user-patch-number is a default from the github web interface right?
@Anirban166 can you please explain to @DorisAmoakohene about how to create branches with more informative names for PRs?

DorisAmoakohene:
@tdhock I updated the new branch name with @Anirban166 and it closed this pull request because the old branch name got deleted and you can see the new pull request for this test here https://github.com/Rdatatable/data.table/pull/6169

tdhock:
ok in the future try to make smaller branch names (if they are too large then the text is too small to read in the performance test plot for HEAD in the PR)

Anirban166:
Second that, also small thing but the title should be ' ... -regression-fixed-in-5463' (instead of 'adding-a-test-for-regression-for-5463')

PR #6164, opened 2024-05-30
Upgrade `dcast` behavior from message to warning for missing aggregate function

### LINKED ISSUE #5386: dcast: add argument to error on "Aggregate function missing, defaulting to 'length'"
We all _love_ seeing `Aggregate function missing, defaulting to 'length'` as it usually indicates a bug (possibly upstream resulting in unexpected input). I really need this to be an error but can't even use `options(warn = 2)` because `dcast` only gives a message and not a warning.

I suggest modifying `dcast.data.table`. One easy option would be a `strict` argument that defaults to `FALSE` but promotes the message to an error if set to `TRUE`. If we want to avoid additional arguments, we could have `fun.aggregate` accept `NA` as input with the same result.

Right now the best way to catch this case appears to be `stopifnot(anyDuplicated(DT[, .(LHS, RHS)]) == 0L)` but I dislike littering my code with this.

Nj221102:
closes #5386 

**Description**:
This PR enhances the behavior of the `dcast` function in the `data.table` package by issuing a warning instead of a message when the `fun.aggregate` argument is missing. The warning is now classed as `dt_missing_agg_function_warning`, providing users with more precise control over handling this condition in their code. This change improves clarity and helps users identify potential issues more effectively.

**Changes Made**:
- Modified the `dcast.data.table` function to issue a warning instead of a message when `fun.aggregate` is missing.
- Classed the warning as `dt_missing_agg_function_warning` to enable more targeted handling in user code.
- Updated related documentation.


tdhock:
@MichaelChirico I'm not sure there was consensus in the issue that we want to change to a warning -- would you approve merging this PR just to see what the revdeps say?

NEWS.md
```diff
@@ -96,6 +96,8 @@
 
 14. `setNumericRounding()` now invisibly returns the old rounding value instead of `NULL`, which is now consistent with similar behavior by `setwd()`, `options()`, etc. Thanks @MichaelChirico for the report and @joshhwuu for the fix.
 
+15. `dcast()` now issues a warning instead of a message when `fun.aggregate` is missing, defaulting to `length()`. This change makes potential issues more noticeable and easier to debug, [#5386](https://github.com/Rdatatable/data.table/issues/5386). The warning is classed as `dt_missing_agg_function_warning`, allowing for more targeted handling in user code. Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
```
MichaelChirico, line 99: We should be more specific here. "potential issues" is vague. As NEWS reader, my reaction is either (1) eyes glaze over & I miss something potentially important or (2) confusion as to whether I may be affected, and probably wind up spending too much time reading through details irrelevant to me --> frustration.

Better to offer enough details here so that the reader can quickly determine whether they need to look into this further for their own case.

The item is also _slightly_ incorrect -- the wording could be read to seem like we _always_ need to provide `fun.aggregate`.

I propose the following wording:

15. `dcast()` now issues a warning when `fun.aggregate` is used but not provided by the user. `fun.aggregate` defaults to `length` in this case; previously only a `message` was issued, but often relying on this default signals unexpected duplicates in the data, so a stricter class of signal was deemed more appropriate.

man/dcast.data.table.Rd
```diff
@@ -17,7 +17,7 @@
 \arguments{
   \item{data}{ A \code{data.table}.}
   \item{formula}{A formula of the form LHS ~ RHS to cast, see Details.}
-  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} with a message.
+  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} with a warning, which is classed as 'dt_missing_agg_function_warning', enabling more precise handling in user code.
```
MichaelChirico, line 20:
```suggestion
  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} with a 'dt_missing_agg_function_warning'.
```

R/fcast.R
```diff
@@ -182,7 +182,7 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
...
+      warningf("'fun.aggregate' is NULL, but found duplicate row/column combinations, so defaulting to length(). That is, the variables %s used in 'formula' do not uniquely identify rows in the input 'data'. In such cases, 'fun.aggregate' is used to derive a single representative value for each combination in the output data.table, for example by summing or averaging (fun.aggregate=sum or fun.aggregate=mean, respectively). Check the resulting table for values larger than 1 to see which combinations were not unique. See ?dcast.data.table for more details.", brackify(varnames), class= "dt_missing_agg_function_warning")
```
MichaelChirico, line 185: I wonder if `dt_missing_fun_aggregate_warning` is a better name. WDYT @tdhock ?

MichaelChirico:
> @MichaelChirico I'm not sure there was consensus in the issue that we want to change to a warning -- would you approve merging this PR just to see what the revdeps say?

Yes I'm A-OK with this. I think there is ~basically maintainer consensus here, from my read of the issue + comment reactions (upgrading to `warning` is OK, upgrading to `error` is not).

NEWS.md
Nj221102: done 👍

R/fcast.R
tdhock: yes or even `dt_missing_fun.aggregate_warning` if that is allowed?

MichaelChirico: it's allowed but I didn't feel great about mixing naming styles. WDYT about dt_missing_funaggregate_warning instead? since we use 'datatable' elsewhere omitting the '.'

tdhock: `dt_missing_fun_aggregate_warning` would be easier to read

Nj221102:
Hi @MichaelChirico and @tdhock from the conversation both of you had, i think `dt_missing_fun_aggregate_warning` is something you both agree on, so i am implementing it now and if any changes are needed later we can do that too :)

man/dcast.data.table.Rd
```diff
@@ -17,7 +17,7 @@
 \arguments{
   \item{data}{ A \code{data.table}.}
   \item{formula}{A formula of the form LHS ~ RHS to cast, see Details.}
-  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} with a message.
+  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} with a 'dt_missing_fun_aggregate_warning'.
```
tdhock, line 20: with a class

tdhock: warning of class

tdhock (changes requested):

tdhock (changes requested):

man/dcast.data.table.Rd
```diff
@@ -17,7 +17,7 @@
 \arguments{
   \item{data}{ A \code{data.table}.}
   \item{formula}{A formula of the form LHS ~ RHS to cast, see Details.}
-  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} with a 'dt_missing_fun_aggregate_warning'.
+  \item{fun.aggregate}{Should the data be aggregated before casting? If the formula doesn't identify a single observation for each cell, then aggregation defaults to \code{length} warning of class 'dt_missing_fun_aggregate_warning'.
```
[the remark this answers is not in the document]
Nj221102: sorry i didn't fully understand what you want to say with "with a warning" so i made it "with a warning of class" because i think telling people that 'dt_missing_fun_aggregate_warning' is class of the warning is better than just writing that it is the warning, but if it feels wrong to you we can try something else as well :)

NEWS.md
```diff
@@ -96,6 +96,8 @@
 
 14. `setNumericRounding()` now invisibly returns the old rounding value instead of `NULL`, which is now consistent with similar behavior by `setwd()`, `options()`, etc. Thanks @MichaelChirico for the report and @joshhwuu for the fix.
 
+15. `dcast()` now issues a warning when `fun.aggregate` is used but not provided by the user. `fun.aggregate` defaults to `length` in this case; previously, only a message was issued, but often Relying on this default signals unexpected duplicates in the data, so a stricter class of signal was deemed more appropriate, [#5386](https://github.com/Rdatatable/data.table/issues/5386). The warning is classed as `dt_missing_fun_aggregate_warning`, allowing for more targeted handling in user code. Thanks @MichaelChirico for the suggestion and @Nj221102 for the fix.
```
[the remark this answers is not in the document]
Nj221102: > `dcast()` now issues a warning when `fun.aggregate` is used but not provided by the user. `fun.aggregate` defaults to `length` in this case; previously, only a message was issued,

simply informing about the change of message to warning when fun.agg is missing.

> but often Relying on this default signals unexpected duplicates in the data

Updated `but often Relying on this default signals unexpected duplicates in the data` to `However, relying on this default often signals unexpected duplicates in the data.` 

 because "default signals" was a bit confusing even for me,  and this line just informs that when we rely on default ( which is length in this case) means that we have unexpected duplicates in the data, this line was suggested by michael to me so i think this must have some importance, also fixing the typo of making the R capital in this update
 
 > , so a stricter class of signal was deemed more appropriate, The warning is classed as `dt_missing_fun_aggregate_warning`, allowing for more targeted handling in user code.
 
 so we upgraded to warning and warning is classed as `dt_missing_fun_aggregate_warning` for targeted handling using try catch or `globalCallingHandlers`

tdhock (approved):

tdhock:
great thanks

PR #6165, opened 2024-05-30
"fwrite" headers are no longer automatically quoted when "na" is used.

### LINKED ISSUE #2964: fwrite is not consistent on quoting headers when `na` is used
```r
fwrite(data.table(foo = TRUE), na = 'asd')
#"foo"
#TRUE
fwrite(data.table(foo = TRUE))
#foo
#TRUE
```

joshhwuu:
Closes #2964 

Previously, the internal `doQuote` flag is set to true whenever `na` argument is given, (a non-empty string):

https://github.com/Rdatatable/data.table/blob/8ccadc685c4b75e7bbf89a6fb8a49cae4995a709/src/fwrite.c#L604-L606

Although this was good as to not misinterpret data, it had the side effect of writing headers with quotes. Changed code to add a new `quoteHeaders` flag which is set to the value of `doQuote` before it is changed by `na`, and set `doQuote` to this value temporarily when writing column names. This way, headers are quoted based on the preserved value of `doQuote` regardless of whether row fields are quoted due to the presence of `na` or not.

old: 
```R
fwrite(data.table(foo = TRUE), na = 'asd')
# "foo"
# TRUE
```

new:
```R
fwrite(data.table(foo = TRUE), na = 'asd')
# foo
# TRUE
```

Changed the following tests to reflect these changes:
1658.27
1658.35
1731.2
1731.4
1732.8
1751.3
1916.1

LMK if there should be an accompanying NEWS entry and/or documentation change.


inst/tests/tests.Rraw
```diff
@@ -13030,7 +13030,7 @@ test(1915, fread(str, select = c(5, 1, 3)),
 
 # unit tests for #2169 which was already fixed in ec05f8b/src/fwrite.c#L59
 test(1916.1, fwrite(data.table(foo=c(NA, TRUE, FALSE)), logical01=TRUE, na='EMPTY', verbose=FALSE),
-             output=c("\"foo\"","EMPTY","1","0")) # this might need to be changed after #2964 will be solved
+             output=c("foo","EMPTY","1","0")) # changed as of resolving #2964
```
Anirban166, line 13033: Imo, this comment can be avoided if it's resolved (which is the case now), as it only makes sense to have such if it's on the back burner (unless it's a critical thing and we must require a reference)

MichaelChirico:
Changed tests look good to me. For the implementation, is there no way to avoid adding the new flag? I'm not clear from the current diff on why exactly the change works / a flag is needed, so this will help check my understanding as well.

joshhwuu:
>  I'm not clear from the current diff on why exactly the change works / a flag is needed, so this will help check my understanding as well.

My knowledge of the file is still a bit hazy but I'll try my best to explain how I'm understanding this:

The line I referenced above:

https://github.com/Rdatatable/data.table/blob/8ccadc685c4b75e7bbf89a6fb8a49cae4995a709/src/fwrite.c#L604-L606

is set early in the main `fwrite` function, it checks if the `na` argument is non-empty and if the `quote` argument is given. If `na` is a non-empty string and `quote` isn't provided by the user, then the `doQuote` flag is set to "true" and `fwrite` will process every field, including headers with quotes. This makes sense because if we set `na` to a non-empty string, we should quote every field in case it contains the NA string.

See examples:

```R
fwrite(data.table(foo = "a"), na = 'asd', quote = F) # quote is false, is respected
# foo
# a
fwrite(data.table(foo = "a"), na = 'asd', quote = T) # quote is true, is respected
# "foo"
# "a"
fwrite(data.table(foo = "a"), na = 'asd') # quote isn't given, everything is quoted
# "foo"
# "a"
fwrite(data.table(foo = "a")) # na is empty, quoting is auto
# foo
# a
```
The main issue with the previous behavior is because headers being quoted when `quote` isn't given and `na` is non-empty isn't intuitive, as headers can't contain empty strings regardless. 

Because fields and headers aren't quoted by the same logic in every case, I thought it made sense to have a flag to track the intended quoting behavior of headers as well.

Initially, the `quoteHeaders` flag is set to the same value as the `doQuote` flag we get from `args.doQuote,` so if `quote` is ever provided by the user, the headers will respect the given `quote` value. If `quote` isn't given, then `quoteHeaders` will be the default "auto" value. This means that `quoteHeaders` won't be affected by the check I mentioned above, and therefore we won't unnecessarily quote headers.

With the flag, it is simple to just set `doQuote = quoteHeaders` before we call the internal helper to print column names and restore afterwards.

Of course, now that I think about it some more, it would be better to just use an internal variable to track `quoteHeaders`, as we never read the value of it outside of the main `fwrite` function.

@MichaelChirico LMK if this makes sense to you or if you see a better way to handle this behavior, thanks!

jangorecki (commented):
Change looks ok but the fact that so many tests needed an update feels like a not good sign. Let's test revdeps to get extra feedback.

joshhwuu:
> Let's test revdeps to get extra feedback.

Sure! Is this something I'm able to test and if so, how would I go about doing that?

ben-schwen:
> > Let's test revdeps to get extra feedback.
> 
> Sure! Is this something I'm able to test and if so, how would I go about doing that?

Although it's technically possible to do this locally (using the [revdep.R script](https://github.com/Rdatatable/data.table/blob/master/.dev/revdep.R)), our normal procedure here is to merge and wait for the revdeps to pop up on the [NAU Monsoon cluster](https://github.com/Rdatatable/data.table/wiki/Release-management-and-revdep-checks)

MichaelChirico:
> headers can't contain empty strings regardless.

Hmm, not _completely_ true. Maybe something we should try and enforce though:

```r
DT = data.frame(a=1, b=2)
names(DT) = c('c', NA_character_)
setDT(DT)
is.na(names(DT))
# [1] FALSE  TRUE
```

As for the `fwrite()` behavior:

```r
fwrite(DT)
# c,
# 1,2
fwrite(DT, quote=TRUE)
# "c",
# 1,2
fwrite(DT, quote=TRUE, na='xxx')
# "c",xxx
# 1,2
fwrite(DT, na='xxx')
# "c",xxx
# 1,2
```

This is on `master`, haven't checked on this branch. At a minimum we should make sure there's no error introduced.

joshhwuu:
> This is on `master`, haven't checked on this branch. At a minimum we should make sure there's no error introduced.

Here's the behavior on this branch:

```R
> DT = data.frame(a=1, b=2)
names(DT) = c('c', NA_character_)
setDT(DT)
is.na(names(DT))
[1] FALSE  TRUE

> fwrite(DT)
c,
1,2

> fwrite(DT, quote=TRUE)
"c",
1,2

> fwrite(DT, quote=TRUE, na='xxx')
"c",xxx
1,2

> fwrite(DT, na='xxx')
c,xxx
1,2

```

MichaelChirico (approved):

PR #6175, opened 2024-06-10
Function wrappers to get a dt without its keys or just the keys

### LINKED ISSUE #981: Return a data.table without the key
It would be nice to be able to return a data.table on the fly _without_ the key.  

This would be useful for things like regressions where you might want to keep a key, but you don't want to include it in the regression. Of course I could use my own function, but I would prefer to use something standard.
## Example function

Perhaps there's something more elegant / obvious?

``` R
keyless <- function(x){
    x[ , -which(colnames(x) %in% key(x)), with=FALSE]
}
```
## Example usage:

``` R
library(data.table)
## Example using the rock data, with an additional column ID which 
## in a real example may be used to join different data sets.
dt <- data.table(id=paste0("rock", sprintf("%02d", 1:48)), rock)
setkey(dt, id)

## View the structure:
str(dt)

# Classes ‘data.table’ and 'data.frame':  48 obs. of  5 variables:
#  $ id   : chr  "rock01" "rock02" "rock03" "rock04" ...
#  $ area : int  4990 7002 7558 7352 7943 7979 9333 8209 8393 6425 ...
#  $ peri : num  2792 3893 3931 3869 3949 ...
#  $ shape: num  0.0903 0.1486 0.1833 0.1171 0.1224 ...
#  $ perm : num  6.3 6.3 6.3 6.3 17.1 17.1 17.1 17.1 119 119 ...
#  - attr(*, "sorted")= chr "id"

## Define "keyless"
keyless <- function(x){
    x[ , -which(colnames(x) %in% key(x)), with=FALSE]
}
## Do a regression
## Obviously we want to exclude the column of identifiers, so we use keyless
lm(area~., keyless(dt))

# Call:
# lm(formula = area ~ ., data = keyless(dt))
# 
# Coefficients:
# (Intercept)         peri        shape         perm  
#    -407.069        2.193     2992.314        2.549  
```

I know that others have mentioned this, but I couldn't find an existing issue.

Thank you

joshhwuu:
Closes #981 

Small wrapper functions that allow the user to easily extract the keys/value columns of a data.table as a data.table object. Could also merge into one function, and allow an argument to choose whether we want to select keys or values. Names are also up for debate!

```R
dt = data.table(x = 1:5, y = letters[1:5], key = "x")

keys = keydata(dt)
# Key: <x>
#        x
#    <int>
# 1:     1
# 2:     2
# 3:     3
# 4:     4
# 5:     5

values = valuedata(dt)

#         y
#    <char>
# 1:      a
# 2:      b
# 3:      c
# 4:      d
# 5:      e
```

TODO:
- [ ] NEWS


man/coredata.Rd
```diff
@@ -0,0 +1,33 @@
...
+  \code{valuedata} returns a \code{data.table} containing the value columns of \code{x}.
+}
+\examples{
+# Create a data.table
+dt = data.table(x = 1:5, y = letters[1:5], key = "x")
+
+# Extract the key columns as a data.table
+keys = keydata(dt)
+
+# Exctract the value columns as a data.table
```
Anirban166, line 28: Extract*

MichaelChirico:
Thank you for the PR!

Honestly a bit hesitant about adding a new function here. Given the back-and-forth in the long-running issue discussion, I think it makes sense to chime in there first before a PR for such cases.

I look around on GitHub for similar cases and I find almost nothing:

https://github.com/search?q=lang%3AR+%2F%5C.SDcols%3F%5Cs*%3D%5Cs*%5B%21-%5D%3F%5Cs*key%5C%28%2F&type=code

Much more common appears to be usage like `[,,by=key(.)]`:

https://github.com/search?q=lang%3AR+%2F%5B%5Ea-z0-9._%5D%5B%21-%5D%3F%5Cs*key%5C%28%5Cs*%5Ba-z0-9._%5D%2B%5Cs*%5C%29%5Cs*%5C%5D%2F&type=code

So I lean towards, if this is something super-useful to you, just write a helper function for it. Especially given how easy the implementation is -- we're not hiding that much complexity from the user, especially if they're already familiar with the basics of `.SD`.

WDYT @tdhock + @jangorecki ?

joshhwuu:
Right, should've done that first, happy to deal with this however everyone sees fit!

MichaelChirico:
In this case, the PR was hopefully not that much trouble -- so thanks for lighting a fire! Just hoping to save you trouble in cases that require more investment to get to the PR.

tdhock:
i would rather not add a new function, if we can agree that we can just close the issue (maybe add documentation/vignette text to describe what to do with existing functions?)

joshhwuu:
As Michael mentioned, most users familiar with data.table should be able to figure something like this out, however it might be good to add a one line example in the key vignette/documentation

PR #6179, opened 2024-06-12
Add Verbose Argument to forder Function for Enhanced Logging Control (#4533)

### LINKED ISSUE #4533: C forder could take verbose argument
That would be handly when we have a C function `fun1` that uses `datatable.verbose` option, and this function calls `forder` internally. We are mostly interested in verbose output of `fun1` rather than `forder`. Yet good control could be provided by calling something like `forder(verbose = verbose-1L)`. That allows to print verbose messages from everything when bigger value is provided to `verbose`.
This feature is already implemented in #4491

Nj221102:
closes #4533

#### Description

This pull request addresses issue [#4533](https://github.com/Rdatatable/data.table/issues/4533) by adding a `verbose` argument to the `forder` function. This enhancement allows for better control over the verbosity of output, particularly when `forder` is called internally by other functions. This is useful for managing detailed logging in higher-level functions without overwhelming the output with messages from `forder`.

- This feature builds upon the logging enhancements implemented in [#4491](https://github.com/Rdatatable/data.table/pull/4491).

#### Checklist

- [x] Added `verbose` argument to `forder` function.
- [x] Updated internal calls to propagate `verbose` argument.
- [x] Adjusted logging mechanism to respect `verbose` level.
- [ ] Added tests for `verbose` functionality.

#### References

- Issue: [#4533](https://github.com/Rdatatable/data.table/issues/4533)
- Previous related PR: [#4491](https://github.com/Rdatatable/data.table/pull/4491)

### Implementation of the Verbose Argument in `forder`

Initially, I added a simple `verbose` argument to the `forder` function and set the `verbose` variable like this:

```c
const int Verbose = INTEGER(verbose)[0];
```

This straightforward approach allowed us to control the verbosity of `forder` by using a command such as `forder(verbose = verbose - 1L)`. However, this implementation revealed a significant issue: many tests started failing because `forder` no longer used the `GetVerbose()` function to retrieve the global verbosity settings. This change caused problems in scenarios where the global verbosity setting was expected to control the output.

For instance, when we attempted the following command:

```r
DT[order(C)[1:5], B, verbose=2L]
```

we did not receive the verbose output from `forder` as anticipated. This issue arose because the `forder` function was solely relying on the local `verbose` argument, ignoring the global verbosity settings controlled by `GetVerbose()`.

### Resolving the Issue: Combining Global and Local Verbosity

To address this problem and allow `forder` to respect both the global verbosity setting and the local `verbose` parameter, I modified the implementation using a ternary operator. This operator determines which verbosity value should be used based on the global setting and the local argument:

```c
const int verbose = GetVerbose() >= 2 ? GetVerbose() - 1L : INTEGER(verboseArg)[0];
```

With this modification, users can set the verbosity to a value greater than or equal to 2L to get verbose output from `forder`. The logic behind this approach is as follows:

1. **Global Verbosity Control**: If the global verbosity setting, retrieved using `GetVerbose()`, is greater than or equal to 2, the `forder` function will use this value. This ensures that the global setting takes precedence and reduces by one to adjust for the internal logic.
   
2. **Local Verbosity Control**: If the global verbosity setting is less than 2, the `forder` function will use the local `verbose` argument provided by the user. This allows for specific control over the verbosity for individual function calls.

### Benefits of the Combined Approach

This combined approach offers several advantages:

- **Flexibility**: Users can control verbosity at both a global and a local level, providing greater flexibility in managing output.
- **Backward Compatibility**: Existing code that relies on global verbosity settings continues to function as expected.

In summary, this improvement ensures that `forder` respects both global and local verbosity settings, allowing for comprehensive and flexible control over verbose output. This change resolves the issue where verbose output was not being generated as expected and ensures that all verbosity controls work harmoniously.

### DIFF
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -137,7 +137,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       }
       if (is.null(xo)) {
         if (verbose) {last.started.at=proc.time(); flush.console()}
-        xo = forderv(x, by = xcols, verbose=verbose-1L)
+        xo = forderv(x, by = xcols)
         if (verbose) {catf("Calculated ad hoc index in %s\n", timetaken(last.started.at)); flush.console()}
         # TODO: use setindex() instead, so it's cached for future reuse
       }
@@ -153,7 +153,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
     if (roll != FALSE) stopf("roll is not implemented for non-equi joins yet.")
     if (verbose) {last.started.at=proc.time();catf("  forder took ... ");flush.console()}
     # TODO: could check/reuse secondary indices, but we need 'starts' attribute as well!
-    xo = forderv(x, xcols, retGrp=TRUE, verbose=verbose-1L)
+    xo = forderv(x, xcols, retGrp=TRUE)
     if (verbose) {cat(timetaken(last.started.at),"\n"); flush.console()}
     xg = attr(xo, 'starts', exact=TRUE)
     resetcols = head(xcols, non_equi-1L)
@@ -161,7 +161,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       # TODO: can we get around having to reorder twice here?
       # or at least reuse previous order?
       if (verbose) {last.started.at=proc.time();catf("  Generating group lengths ... ");flush.console()}
-      resetlen = attr(forderv(x, resetcols, retGrp=TRUE, verbose=verbose-1L), 'starts', exact=TRUE)
+      resetlen = attr(forderv(x, resetcols, retGrp=TRUE), 'starts', exact=TRUE)
       resetlen = .Call(Cuniqlengths, resetlen, nrow(x))
       if (verbose) {catf("done in %s\n",timetaken(last.started.at)); flush.console()}
     } else resetlen = integer(0L)


src/forder.c
```diff
@@ -418,9 +418,12 @@ SEXP forder(SEXP DT, SEXP by, SEXP retGrpArg, SEXP sortGroupsArg, SEXP ascArg, S
   memset(stat,   0, 257*sizeof(uint64_t));
   TBEG()
 #endif
+  
+  if ((!isLogical(verbose) && !isInteger(verbose)) || LENGTH(verbose)!=1 || INTEGER(verbose)[0]==NA_INTEGER)
+    error(_("verbose option must be length 1 non-NA logical or integer"));
 
   int n_protect = 0;
-  const bool verbose = GetVerbose();
+  const int Verbose = INTEGER(verbose)[0];
```
Anirban166, line 426: Not sure if it's good practice to just capitalize the first letter of a variable name for a separate type of variable (bools vs int in this case)

inst/tests/tests.Rraw
```diff
@@ -16163,16 +16161,15 @@ test(2097.3, DT[J(NA_integer_)], error="Factor columns must join to factor or ch
...
+test(2098.4, DT[with(DT, do.call(forder, mget(groups))), verbose=TRUE], ans)
 old = options(datatable.optimize=0L)
-test(2098.5, DT[do.call(order, mget(groups)), verbose=TRUE],            ans, notOutput="forder.c")
-test(2098.6, DT[with(DT, do.call(order, mget(groups))), verbose=TRUE],  ans, notOutput="forder.c")
-test(2098.7, DT[do.call(forder, mget(groups)), verbose=TRUE],           ans, output=out)
-test(2098.8, DT[with(DT, do.call(forder, mget(groups))), verbose=TRUE], ans, output=out)
+test(2098.5, DT[do.call(order, mget(groups)), verbose=TRUE],            ans)
+test(2098.6, DT[with(DT, do.call(order, mget(groups))), verbose=TRUE],  ans)
+test(2098.7, DT[do.call(forder, mget(groups)), verbose=TRUE],           ans)
+test(2098.8, DT[with(DT, do.call(forder, mget(groups))), verbose=TRUE], ans)
```
Anirban166, line 16172: Include a brief comment explaining the differences among the tests

src/forder.c
Nj221102: the thing is that i actually didn't wanted to change the variable name, but needed to name the arguement with same name `verbose` hence thought of changing the variable name instead and wanted something that doesn't confuse people about what this variable's role is ? so thought `Variable ` would be fitting, what else could be better here ?

ben-schwen: Our common practice would be to rename the function argument to `verboseArg` and then use `verbose`

inst/tests/tests.Rraw
Nj221102: what type of difference are we talking about ?, these test were already there and i just modified them because their output changed due to this pr, Should i leave a comment informing this change??

src/assign.c
```diff
@@ -288,10 +288,11 @@ int checkOverAlloc(SEXP x)
 }
 
 SEXP alloccolwrapper(SEXP dt, SEXP overAllocArg, SEXP verbose) {
-  if (!IS_TRUE_OR_FALSE(verbose))
-    error(_("%s must be TRUE or FALSE"), "verbose");
+  if ((!isLogical(verbose) && !isInteger(verbose)) || LENGTH(verbose)!=1 || INTEGER(verbose)[0]==NA_INTEGER)
```
HughParsonage, line 291: This if condition is too terse to be readable and relies on the undocumented behaviour of `INTEGER()` on logical input. 

With integer verbose values, we should document the integer codes.

HughParsonage (changes requested):

inst/tests/tests.Rraw
Anirban166: Yes, a comment describing that change and/or what these set of tests do might be good to have.

src/assign.c
ben-schwen: We have the same condition here:
https://github.com/Rdatatable/data.table/blob/ff900d1e6a8bcfaa0385bd1304af7b90657d3c4d/src/init.c#L319-L321

Nj221102: yeah this if condition is already being used in many parts of the code base where verbose can be both `INTEGER` and `Logical`

MichaelChirico: that tells me we need a helper function :)

Anirban166: Same thoughts!

PR #6181, opened 2024-06-14
Return File Path after Writing in `fwrite`

### LINKED ISSUE #5706: `fwrite()` should (possibly invidibly) return the file path
Now `fwrite()` returns `invisible()`, but I think it might be more useful if it returned the path to the created file. It's a trivial change and I don't think it would create issues with backward compatibility, since it's probably not likely that any code relies on  `fwrite()` not returning anything (although a similar change to `ggsave()` [did break a few scripts](https://github.com/tidyverse/ggplot2/issues/3379) 🤣).

Nj221102:
#### Issue
closes #5706.

#### Description
The current implementation of `fwrite` returns an invisible value upon completion, which can be limiting in scenarios where users need to programmatically access the path of the created file for subsequent operations. This pull request aims to enhance the functionality of `fwrite` by modifying it to return the path of the file it writes to.

#### Changes Made
- **Functionality Enhancement**: Updated `fwrite` to return the file path after writing data to disk.
-  Updated tests accordingly.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -10059,7 +10059,7 @@ test(1658.56, fwrite(data.table(exp(1) - pi*1i)), output='2.718[0-9]*-3.141[0-9]
 DT = data.table(a=1:3, b=list(1:4, c(3.14, 100e10), c(3i,4i,5i)))
 test(1658.57, fwrite(DT), output='0+3i|0+4i|0+5i')
 DT[ , b := c(1i, -1-1i, NA_complex_)]
-test(1658.58, invisible(fwrite(DT)), output='a,b\n1,0+1i\n2,-1-1i\n3,\n[1] ""')
+test(1658.58, invisible(fwrite(DT)), output='a,b\n1,0+1i\n2,-1-1i\n3,\n')
 
 # more coverage
 test(1658.59, fwrite(data.table(a=list('a')), verbose=TRUE),


NEWS.md
```diff
@@ -44,6 +44,8 @@
 
 14. `fread` loads `.bgz` files directly, [#5461](https://github.com/Rdatatable/data.table/issues/5461). Thanks to @TMRHarrison for the request with proposed fix, and Benjamin Schwendinger for the PR.
 
+15. `fwrite` now returns the file path after writing data, [#5706](https://github.com/Rdatatable/data.table/issues/5706). This enhancement allows users to easily capture and utilize the file path for subsequent operations.Thanks to @Nj221102 for the PR.
```
Anirban166, line 47: space between 'operations.' and 'Thanks'

Anirban166: Also, @eliocamp should be credited

Nj221102: sure :)

NEWS.md
```diff
@@ -44,6 +44,8 @@
 
 14. `fread` loads `.bgz` files directly, [#5461](https://github.com/Rdatatable/data.table/issues/5461). Thanks to @TMRHarrison for the request with proposed fix, and Benjamin Schwendinger for the PR.
 
+15. `fwrite` now returns the file path after writing data, [#5706](https://github.com/Rdatatable/data.table/issues/5706). This enhancement allows users to easily capture and utilize the file path for subsequent operations. Thanks to  @eliocamp for the request and  @Nj221102 for the PR.
```
MichaelChirico, line 47:
```suggestion
15. `fwrite` now returns the file path instead of `NULL`, [#5706](https://github.com/Rdatatable/data.table/issues/5706). This can be useful to subsequent operations in scripts, especially some pipelines. Thanks to  @eliocamp for the request and  @Nj221102 for the PR.
```

R/fwrite.R
```diff
@@ -84,11 +84,11 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
     if (file.exists(file)) {
       suggested <- if (append) "" else gettextf("\nIf you intended to overwrite the file at %s with an empty one, please use file.remove first.", file)
       warningf("Input has no columns; doing nothing.%s", suggested)
-      return(invisible())
+      return(invisible(file))
```
MichaelChirico, line 87: Maybe here `NULL` should be returned, since the actual `file` is not affected at all by `fwrite()`? WDYT @eliocamp

inst/tests/tests.Rraw
```diff
@@ -9977,7 +9977,7 @@ test(1658.31, fwrite(ok_dt, qmethod=c("double", "double")), error="length(qmetho
 test(1658.32, fwrite(ok_dt, col.names="foobar"), error="isTRUEorFALSE(col.names)")
 
 # null data table (no columns)
-test(1658.33, fwrite(data.table(NULL)), NULL, warning="Nothing to write")
+test(1658.33, fwrite(data.table(NULL)), "", warning="Nothing to write")
```
MichaelChirico, line 9980: Hmm, wonder how useful this case is too -- when `fwrite()` to the terminal, maybe `NULL` should be returned as well? Not sure any inconsistency is worthwhile for this case though.

inst/tests/tests.Rraw
```diff
@@ -10009,9 +10009,9 @@ if (!haszlib()) {
 } else {
   test(1658.41, fwrite(data.table(a=c(1:3), b=c(1:3)), compress="gzip"), output='a,b\n1,1\n2,2\n3,3')  # compress ignored on console
   DT = data.table(a=rep(1:2,each=100), b=rep(1:4,each=25))
-  test(1658.421, fwrite(DT, file=f1<-tempfile(fileext=".gz"), verbose=TRUE), NULL,
+  test(1658.421, filepath <- fwrite(DT, file=f1<-tempfile(fileext=".gz"), verbose=TRUE), filepath,
```
MichaelChirico, line 10012: is `filepath` different from `f1`?

Ditto for `f2` below.

inst/tests/tests.Rraw
```diff
@@ -10059,7 +10059,7 @@ test(1658.56, fwrite(data.table(exp(1) - pi*1i)), output='2.718[0-9]*-3.141[0-9]
 DT = data.table(a=1:3, b=list(1:4, c(3.14, 100e10), c(3i,4i,5i)))
 test(1658.57, fwrite(DT), output='0+3i|0+4i|0+5i')
 DT[ , b := c(1i, -1-1i, NA_complex_)]
-test(1658.58, fwrite(DT), output='a,b\n1,0\\+1i\n2,-1-1i\n3,$')
+test(1658.58, fwrite(DT), output='a,b\n1,0+1i\n2,-1-1i\n3,\n[1] ""')
```
MichaelChirico, line 10062:
```suggestion
test(1658.58, invisible(fwrite(DT)), output='a,b\n1,0\\+1i\n2,-1-1i\n3,$')
```

inst/tests/tests.Rraw
```diff
@@ -10809,7 +10809,7 @@ if (test_bit64) {
   test(1731.1, class(DT[[1L]]), "integer64")
   test(1731.2, fwrite(DT,na="__NA__"), output=ans)
   f = tempfile()
-  test(1731.3, fwrite(DT, f, na="__NA__"), NULL)
+  test(1731.3, filepath <- fwrite(DT, f, na="__NA__"), filepath)
```
MichaelChirico, line 10812: `f`?

inst/tests/tests.Rraw
```diff
@@ -13775,9 +13775,9 @@ if (.Platform$OS.type=="windows") {
   test(1966.4, list.files(path = pth, pattern = "\\.csv$"), f)
   unlink(c(fp, file.path(pth, "\u00c3\u00b6.csv")))
   p = file.path(pth, "\u00fc"); dir.create(p); f = tempfile(tmpdir = p)
-  test(1966.5, fwrite(DT, enc2native(f)), NULL)
+  test(1966.5, filepath <- fwrite(DT, enc2native(f)), filepath)
```
MichaelChirico, line 13778: this case is more interesting, but note the test as you've written it is patently useless :)

In essence, you've written

```r
test(N, x, x)
```

That will always trivially be true!

Instead, we want `fwrite(DT, filepath<-enc2native(f)) to be compared to `filepath`.

Ditto the next test.

R/fwrite.R
eliocamp: That's a though one. Honestly this behaviour is kind of unexpected to me. I would've expected `fwrite` to either write an empty file or to throw an error. 

With `append == TRUE`, the file is left untouched, which would be the expected behaviour when writing an empty data.table, so returning the file would make sense, though.

inst/tests/tests.Rraw
Nj221102: done 👍

inst/tests/tests.Rraw
Nj221102: updated 👍

inst/tests/tests.Rraw
Nj221102: this seems to be failing the test 
```
 Test 1658.58 did not produce correct output:
  Expected: <<a,b\n1,0\\+1i\n2,-1-1i\n3,$>>
  Observed: <<a,b\n1,0+1i\n2,-1-1i\n3,\n[1] "">>
  ```
  and why does this test breaks to begin with?, its not like this PR changes anything related to output.

**Comment by tdhock** [comment]
I like returning NULL instead of file name in some cases, so user can test to see if anything was written

File: inst/tests/tests.Rraw (hunk 6 above)
Comment (MichaelChirico) [reply inst/tests/tests.Rraw:10062]: Oh, right, because `output=` test always runs explicit `print()` under the hood. So to reproduce locally you can run `print(fwrite(...))` and you'll see the `[1] ""` part.

That said, we don't need to test for that. I've updated the suggested edit.

**Comment by MichaelChirico** [comment]
>I like returning NULL instead of file name in some cases, so user can test to see if anything was written

That was my initial thought, though Elio points out it might be a bit awkward in the `append=` case. I'm imagining a pipeline like

```r
do_stuff() |>
  # ...
  last_step() |>
  fwrite(file, append=TRUE) |>
  manipulate_file() |>
  more_stuff()
```

Needing to ensure the file was written in order to get something useful out seems inconvenient here -- the file exists, after all.

(might not be the clearest explanation, I don't have a 100% clear thought here yet, but instinctually it's like the above)

Maybe we can do "return file path" :left_right_arrow: "the path exists on exit"?

eliocamp:
It's also awkward when saving within apply or even inside j

```r
DT[, fwrite(.SD), by = groups]
```
Returning `NULL` would essentially drop groups, wouldn't it?  

In general, I feel it's  kind of weird when functions return different things without any active decision from the user. It makes the function unpredictable and harder to program with. 

My ideal behaviour would be to always save whatever is being saved (even if its an empty data.table) and return the file path. For this PR, I don't think we can change what is saved, so at least for now I think a good solution would be to return the path if the file actually represents what's being saved so only returning `NULL` in the case of trying to write an empty data.table to an exiting file with `append = FALSE`.

PR #6182, opened 2024-06-14
Subset Copy Documentation

### LINKED ISSUE #5411: documentation of ":=" (or maybe rather "[" actually)
After creation of a data.table
```
library(data.table)
a=fread("
nr,a
1,a
2,b
")
a
```
a new one is created the following way:
```
b <- a[]
b
```
NB: I did not just assign `b <- a` but `b <- a[]`. According to the documentation (see below) I'd expect `b` to be independent from `a`, and changes not reflect back to `a`. On a first glance that seems to be the case:
```
b[1, 1] <- 4
b
a
all(b==a)
```
However when using `:=` to assign the new value that logic seems broken:
```
b <- a[]

b
b[1, nr:=4]
b
a
all(b==a)
```
What am I missing from the documentation, which says that the subsetting operation returns a _new_ `data.table`, which then in turn is updated by reference using the `:=` operator? 

The effect is the same when creating `b` as `b <- a[TRUE]`. However when using `b <- a[1:.N]` then both objects seem to be separate entities as I'd expect.

See **Note** section of `:=`:

> DT[a > 4, b := c] is different from DT[a > 4][, b := c]. The first expression updates (or adds) column b with the value c on those rows where a > 4 evaluates to TRUE. X is updated by reference, therefore no assignment needed.
> 
> The second expression on the other hand updates a **new data.table that's returned by the subset operation**. Since the subsetted data.table is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: ans <- DT[a > 4][, b := c]. 
> 

```
> sessionInfo()
R version 4.1.3 (2022-03-10)
Platform: i386-w64-mingw32/i386 (32-bit)

Matrix products: default

locale:
[1] LC_COLLATE=German_Germany.1252  LC_CTYPE=German_Germany.1252
[3] LC_MONETARY=German_Germany.1252 LC_NUMERIC=C
[5] LC_TIME=German_Germany.1252

attached base packages:

other attached packages:
[1] data.table_1.14.2

[1] compiler_4.1.3
```

joshhwuu:
Closes #5411 

Current `:=` documentation states: 
>`DT[a > 4, b := c]` is different from `DT[a > 4][, b := c]`. The first expression updates (or adds) column `b` with the value `c` on those rows where `a > 4` evaluates to `TRUE`. `X` is updated by reference, therefore no assignment needed.

>The second expression on the other hand updates a new `data.table` that's returned by the subset operation. Since the subsetted `data.table` is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: `ans <- DT[a > 4][, b := c]`.

While mostly correct about copying the `data.table` when using the subset operation, it doesn't mention that a copy isn't made when subset is empty, and instead we return a reference to the original `data.table`:

```R
DT = data.table(a = 1:5)
DT[, address(a)]
# 0x55dd1a8e3758

dt1 = DT[] # doesn't create a copy, same as 'dt1 = DT'
dt1[, address(a)]
# 0x55dd1a8e3758

dt1[, a := 2] # updates dt1 AND DT
dt1[, address(a)]
# 0x55dd1a8e3758

all.equal(DT, dt1) # since dt1 isn't a copy, DT is also updated by reference when using ':='
# TRUE

# --------------------------------------- Copies when subset is non-empty

dt2 = DT[1:.N] # copies rows 1 - nrow(DT)
dt2[, address(a)]
# 0x55dd16ebbdd8

dt2[, a := 2] # only updates dt2
dt2[, address(a)]
# 0x55dd16ebbdd8

all.equal(DT, dt2) # dt2 is a copy, therefore DT isn't updated by reference
# Column \'a\': Mean relative difference: 0.5384615
```

Updated documentation of `:=` under `Notes` section as this section states that a new `data.table` is made when doing subset operations, which it is, except when the subset operation is empty. 

`assign.Rd` is the only place I found talking about copying with subset, LMK if there's another area where this information would be useful to the user. (Maybe, `data.table.Rd`?), also happy to add the above code as examples in docs if we find the need.

Related: https://github.com/Rdatatable/data.table/blob/ff900d1e6a8bcfaa0385bd1304af7b90657d3c4d/NEWS.0.md?plain=1#L1772

### DIFF
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -85,11 +85,9 @@ Unlike \code{<-} for \code{data.frame}, the (potentially large) LHS is not coerc
 Since \code{[.data.table} incurs overhead to check the existence and type of arguments (for example), \code{set()} provides direct (but less flexible) assignment by reference with low overhead, appropriate for use inside a \code{for} loop. See examples. \code{:=} is more powerful and flexible than \code{set()} because \code{:=} is intended to be combined with \code{i} and \code{by} in single queries on large datasets.
 }
 \note{
-    \code{DT[a > 4, b := c]} is different from \code{DT[a > 4][, b := c]}. The first expression updates (or adds) column \code{b} with the value \code{c} on those rows where \code{a > 4} evaluates to \code{TRUE}. \code{X} is updated \emph{by reference}, therefore no assignment needed.
+    \code{DT[a > 4, b := c]} is different from \code{DT[a > 4][, b := c]}. The first expression updates (or adds) column \code{b} with the value \code{c} on those rows where \code{a > 4} evaluates to \code{TRUE}. \code{X} is updated \emph{by reference}, therefore no assignment needed.  Note that this does not apply when `i` is missing, i.e. \code{DT[]}.
 
     The second expression on the other hand updates a \emph{new} \code{data.table} that's returned by the subset operation. Since the subsetted data.table is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: \code{ans <- DT[a > 4][, b := c]}.
-
-    Regarding subset operations: \code{data.table} only creates a copy of its subset when the subset operation is non-empty. For example, \code{DT[a > 4]} returns a copy of \code{DT} where \code{a > 4} is \code{TRUE}. However, \code{DT[]} returns a reference to the original \code{DT}.
 }
 \value{
 \code{DT} is modified by reference and returned invisibly. If you require a copy, take a \code{\link{copy}} first (using \code{DT2 = copy(DT)}).
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -89,7 +89,7 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
 
     The second expression on the other hand updates a \emph{new} \code{data.table} that's returned by the subset operation. Since the subsetted data.table is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: \code{ans <- DT[a > 4][, b := c]}.
 
-    About subset operations: \code{data.table} only returns a copy of the subset when the subset operation is non-empty. For example, \code{DT[a > 4]} returns a copy of \code{DT} where \code{a > 4} is \code{TRUE}. However, \code{DT[]} returns a reference to the original \code{DT}.
+    About subset operations: \code{data.table} only creates a copy of its subset when the subset operation is non-empty. For example, \code{DT[a > 4]} returns a copy of \code{DT} where \code{a > 4} is \code{TRUE}. However, \code{DT[]} returns a reference to the original \code{DT}.
 }
 \value{
 \code{DT} is modified by reference and returned invisibly. If you require a copy, take a \code{\link{copy}} first (using \code{DT2 = copy(DT)}).
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -89,7 +89,7 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
 
     The second expression on the other hand updates a \emph{new} \code{data.table} that's returned by the subset operation. Since the subsetted data.table is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: \code{ans <- DT[a > 4][, b := c]}.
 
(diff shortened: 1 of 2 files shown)


man/assign.Rd
```diff
@@ -88,6 +88,8 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
     \code{DT[a > 4, b := c]} is different from \code{DT[a > 4][, b := c]}. The first expression updates (or adds) column \code{b} with the value \code{c} on those rows where \code{a > 4} evaluates to \code{TRUE}. \code{X} is updated \emph{by reference}, therefore no assignment needed.
 
     The second expression on the other hand updates a \emph{new} \code{data.table} that's returned by the subset operation. Since the subsetted data.table is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: \code{ans <- DT[a > 4][, b := c]}.
+
+    About subset operations: \code{data.table} only returns a copy of the subset when the subset operation is non-empty. For example, \code{DT[a > 4]} returns a copy of \code{DT} where \code{a > 4} is \code{TRUE}. However, \code{DT[]} returns a reference to the original \code{DT}.
```
Anirban166, line 92: Is 'About subset operations:' necessary? (and if so, 'Regarding' sounds a bit better than 'About' imo)

joshhwuu: > Is 'About subset operations:' necessary? 

IMO yes, as the documentation isn't _for_ subset operations, but is related so it would be nice if someone running into the issue above was able to quickly spot subset operations when scrolling through the documentation

Anirban166: Sounds good 👍🏻

man/assign.Rd
```diff
@@ -88,6 +88,8 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
     \code{DT[a > 4, b := c]} is different from \code{DT[a > 4][, b := c]}. The first expression updates (or adds) column \code{b} with the value \code{c} on those rows where \code{a > 4} evaluates to \code{TRUE}. \code{X} is updated \emph{by reference}, therefore no assignment needed.
 
     The second expression on the other hand updates a \emph{new} \code{data.table} that's returned by the subset operation. Since the subsetted data.table is ephemeral (it is not assigned to a symbol), the result would be lost; unless the result is assigned, for example, as follows: \code{ans <- DT[a > 4][, b := c]}.
+
+    Regarding subset operations: \code{data.table} only creates a copy of its subset when the subset operation is non-empty. For example, \code{DT[a > 4]} returns a copy of \code{DT} where \code{a > 4} is \code{TRUE}. However, \code{DT[]} returns a reference to the original \code{DT}.
```
tdhock, line 92: looks ok overall but I'm not sure about "subset operation is non-empty" maybe change to "`i` argument is provided" ?

MichaelChirico:
I seem to recall something like this came up before regarding the behavior of whether `DT[TRUE]` copies...

MichaelChirico:
> assign.Rd is the only place I found talking about copying with subset, LMK if there's another area where this information would be useful to the user. 

Did you check the vignettes as well? My first instinct is this one:

https://rdatatable.gitlab.io/data.table/articles/datatable-reference-semantics.html

man/assign.Rd
```diff
@@ -88,6 +88,8 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
     \code{DT[a > 4, b := c]} is different from \code{DT[a > 4][, b := c]}. The first expression updates (or adds) column \code{b} with the value \code{c} on those rows where \code{a > 4} evaluates to \code{TRUE}. \code{X} is updated \emph{by reference}, therefore no assignment needed.
```
MichaelChirico, line 88:
```suggestion
    \code{DT[a > 4, b := c]} is different from \code{DT[a > 4][, b := c]}. The first expression updates (or adds) column \code{b} with the value \code{c} on those rows where \code{a > 4} evaluates to \code{TRUE}. \code{X} is updated \emph{by reference}, therefore no assignment needed. Note that this does not apply when `i` is missing, i.e. \code{DT[]}.
```

MichaelChirico: I would be fine with this simpler change for this file at least.

MichaelChirico (approved):
LGTM. If any other reviewers want further refinement, let's do it as a follow-up

Issue #6180, opened 2024-06-13
Remove non-API calls in C


MichaelChirico:
We are getting these NOTEs on CRAN, e.g.

Found non-API calls to R: ‘SETLENGTH’, ‘SET_S4_OBJECT’,
      ‘SET_TRUELENGTH’, ‘UNSET_S4_OBJECT’

I'm really skeptical about the SETLENGTH and SET_TRUELENGTH parts, which have long been a key underlying feature of data.table internals. ~I propose we try and keep using those unless CRAN really forces our hands~ CRAN is really forcing our hands :). It be really great if someone wanted to do a performance comparison of the package with/without those calls.

Anyway, it should be easier to remove the non-API S4 calls.

The list does keep changing as this is an actively evolving piece of r-devel. Here's the current list of things we should (eventually) address:

- [x] LEVELS
- [x] OBJECT
- [x] Rf_GetOption
- [x] NAMED
- [x] SET_S4_OBJECT
- [x] SET_TYPEOF
- [x] STRING_PTR
- [x] UNSET_S4_OBJECT
- [x] isFrame
- [x] SETLENGTH
- [x] SET_TRUELENGTH
- [x] SET_GROWABLE_BIT
- [x] TRUELENGTH
- [x] findVar
- [x] SET_ATTRIB
- [x] ATTRIB

tdhock:
there is a web page which lists all of the API calls, grouped into 3 categories
https://yutannihilation.github.io/R-fun-API/
see also the R-devel thread [which is linked from that page](https://stat.ethz.ch/pipermail/r-devel/2024-June/083449.html)

ecoRoland2:
You should probably be proactive and contact R-core regarding your need of an API to `SETLENGTH` and `SET_TRUELENGTH`. I would hate seeing slower performance in data.table. It could cause some difficult decisions at my day job.

MichaelChirico:
if this is important for you (or anyone else reading) please take the time to construct a performance comparison.

ecoRoland2:
I would but unfortunately I think you need at least basic knowledge of C to do that.

ben-schwen:
AFAIU we can replace `SETLENGTH` calls with `Rf_lengthgets` calls and reassignment. But there is no safe way to handle `SET_TRUELENGTH` right?

tdhock:
[...]

Issue #4415, opened 2020-05-03
s4 objects in j groups are wrong


MichaelChirico:
Another follow-up of #4131 

`j` doesn't handle S4 objects properly:

```
library(data.table)
library(lubridate)
dt <- data.table(period_unit = c("days", "months", "weeks", "years"),
                 period_length = rep(1, times = 4))

dt[ , by = period_unit, do.call(.BY$period_unit, list(period_length))]
#    period_unit          V1
# 1:        days 1d 0H 0M 0S
# 2:      months 1d 0H 0M 0S
# 3:       weeks 1d 0H 0M 0S
# 4:       years 1d 0H 0M 0S
```

The `days` object has been cascaded to the other groups.

I _think_ this is a duplicate of #1777, but I'm not sure; if so, I would close that one since I wrote this to be very direct & clear about the issue.

tlapak:
Writing a somewhat lengthy comment here because I've been wanting _somewhere_ to put what I've read up on and this seems to be a useful spot.

I don't think this is a bug so much as a fundamental limitation of data.table. The only fully supported column types in `data.table`s are atomic vectors with scalar attributes<sup>1</sup> with some limited support for list columns tacked on.

Anything else you can (mostly<sup>2</sup>) stick in in one piece and get it out again in one piece but you can't do anything else safely.

The last answer in #1777 basically explains what the issue is. It's not strictly an S4 thing but that's where it mostly crops up. Internally, simple S4 objects are stored as a vector for the first slot with all further slots stored as attributes of that vector.<sup>3</sup> Now, `data.table` always operates on the underlying vector and assumes that all attributes are scalar.

Your example actually somewhat obscures the underlying issue. Replace the `rep` bit with `1:4` to get the exact same output. Also, the V1 column is actually malformed:
```R
str(dt[ , by = period_unit, do.call(.BY$period_unit, list(period_length))][, V1])
# Formal class 'Period' [package "lubridate"] with 6 slots
#   ..@ .Data : num [1:4] 0 0 0 0
#   ..@ year  : num 0
#   ..@ month : num 0
#   ..@ day   : num 1
#   ..@ hour  : num 0
#   ..@ minute: num 0
```
All of the slots should be length four like the first one. (NB: .Data == second)

The following are all related (in no particular order and I don't make any claims that the list is exhaustive.): #2273, #4315, #4217, #2948, #1777, #3388.

The last one on that list demonstrates that the issues aren't technically S4 specific.


[...]

PR #6183, opened 2024-06-15
Remove {UN,}SET_S4_OBJECT usage

MichaelChirico:
Part of #6180.

I would like to test the dogroups change. I don't _think_ there's any reason to do `UNSET_S4_OBJECT()`? How could the object wound up marked as S4?

Certainly we pass tests as is. Can anyone come up with an example broken by this change? :thinking:


MichaelChirico:
Reading a tiny bit more carefully how `keepattr()` is used in dogroups, it's part of `growVector()`, in turn called twice:

https://github.com/Rdatatable/data.table/blob/ff900d1e6a8bcfaa0385bd1304af7b90657d3c4d/src/dogroups.c#L401
https://github.com/Rdatatable/data.table/blob/ff900d1e6a8bcfaa0385bd1304af7b90657d3c4d/src/dogroups.c#L449

That's for places where we need to mess with the group memory size because of bad estimates up-front (?). So IINM to be broken you'd need to be doing hard-to-predict grouping operations on S4 columns. But already messing with S4 columns in `j` has very limited support (see e.g. #4415). So it's not clear anyone would be affected.

ben-schwen:
AFAIU `IS_S4_OBJECT` is also part of non exported API.

MichaelChirico:
I don't think that's right, here it is in Rinternals:

https://github.com/r-devel/r-svn/blob/53bbd920926a6263bf0e870fa142c6171f37ec01/src/include/Rinternals.h#L273

Well above the "Not API" section

And here it's mentioned in R-ints as the "correct" way to test S4:

https://github.com/r-devel/r-svn/blob/53bbd920926a6263bf0e870fa142c6171f37ec01/doc/manual/R-ints.texi#L1397-L1401

ben-schwen:
My bad, I should have the new section in [WRE](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Moving-into-C-API-compliance).

There we have it:
> This table lists some non-API functions used in packages and the API functions that should be used instead:
  IS_S4_OBJECT
  Use isObject and isS4.

MichaelChirico:
Nice. Slightly confused why this is note showing up in our `NOTE` (yet?), and about the organization of the header & R-ints ref, but assume that'll be ironed out over time.

ben-schwen:
LGTM, but lets see what revdeps will say

PR #6187, opened 2024-06-17
print.data.table Gets New Argument "indices"

### LINKED ISSUE #4280: Option to print indices as columns
`setindex` basically creates "shadow columns" -- `nrow(x)`-size vectors that can be used to sort `x` on the fly.

It would be helpful, besides `tables()` and `indices()`, to be able to show the indices alongside the rest of the tables columns.

This is nice for auditing our tables, as well as helping to surface/visualize what indices actually are/why they're useful.

e.g.

```
NN = 1e6
set.seed(39489)
DT = data.table(
  grp1 = sample(1e5, NN, TRUE),
  grp2 = sample(1e4, NN, TRUE),
  grp3 = sample(1e3, NN, TRUE)
)
setkey(DT, grp1, grp2)
setindex(DT, grp1, grp3)
DT
#            grp1 grp2 grp3
#       1:      1  544   67
#       2:      1  650  150
#       3:      1 1242  230
#       4:      1 1647  915
#       5:      1 6111  119
#      ---                 
#  999996: 100000 2699  494
#  999997: 100000 4844  188
#  999998: 100000 5299  964
#  999999: 100000 7166  819
# 1000000: 100000 8590   45
print(DT, show_indices = TRUE)
## same as
DT[ , paste0('index__', idx) := attr(attr(DT, 'index'), paste0('__', idx))][]
#            grp1 grp2 grp3 index__grp1__grp3
#       1:      1  544   67                 1
#       2:      1  650  150                 5
#       3:      1 1242  230                 2
#       4:      1 1647  915                 3
#       5:      1 6111  119                 6
#      ---                                   
#  999996: 100000 2699  494            999997
#  999997: 100000 4844  188            999995
#  999998: 100000 5299  964            999996
#  999999: 100000 7166  819            999999
# 1000000: 100000 8590   45            999998
```

add to #1523

joshhwuu:
Closes #4280

`print.data.table` gains a new argument to allow users to print its indices as a column without actually assigning a new column to the `data.table`. 

Note: does `set.seed()` work differently in the test script? for some reason I don't get the same values when using `sample` vs in my vscode or R in Windows

### DIFF
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -80,7 +80,9 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     toprint = rbindlist(list(head(x, topn), tail(x, topn)), use.names=FALSE)  # no need to match names because head and tail of same x, and #3306
     rn = c(seq_len(topn), seq.int(to=n_x, length.out=topn))
     printdots = TRUE
-    if (show.indices) toprint = cbind(toprint, rbindlist(list(head(index_dt, topn), tail(index_dt, topn)), use.names=FALSE))
+    idx = c(seq_len(topn), seq(to=nrow(x), length.out=topn))
+    toprint = x[idx, ]
+    if (show.indices) toprint = cbind(toprint, index_dt[idx, ])
   } else {
     toprint = x
     rn = seq_len(n_x)
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -71,7 +71,7 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
       show.indices = FALSE
     } else {
       index_dt <- as.data.table(attributes(attr(x, 'index')))
-      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt), fixed=TRUE))
+      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt)))
       setnames(index_dt, print_names)
     }
   }
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -7,10 +7,10 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
                col.names=getOption("datatable.print.colnames"),
                print.keys=getOption("datatable.print.keys"),
                trunc.cols=getOption("datatable.print.trunc.cols"),
+               print.indices=getOption("datatable.print.indices"),
                quote=FALSE,
                na.print=NULL,
-               timezone=FALSE,
-               indices=FALSE, ...) {
+               timezone=FALSE, ...) {
   # topn  - print the top topn and bottom topn rows with '---' inbetween (5)
   # nrows - under this the whole (small) table is printed, unless topn is provided (100)
   # class - should column class be printed underneath column name? (FALSE)
@@ -65,12 +65,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     return(invisible(x))
   }
-  if (indices) {
+  index_dt = NULL
+  if (print.indices) {
     if (is.null(indices(x))) {
-      indices = FALSE
+      print.indices = FALSE
     } else {
-      index_col_name = paste0("index__", indices(x))
-      index_col = attr(attr(x, 'index'), paste0('__', indices(x)))
+      index_dt = data.table()
+      index_names = indices(x)
+      if (length(index_names) > 1L) {
+        count = 1L
+        for (index_name in index_names) {
+          index_col = attr(attr(x, 'index'), paste0('__', index_name))
+          index_dt[, paste0("index", count, "__", index_name) := index_col]
+          count = count + 1L
+        }
+      } else {
+        index_dt[, paste0("index__", index_names) := attr(attr(x, 'index'), paste0('__', index_names))]
+      }
     }
   }
   n_x = nrow(x)
@@ -78,12 +89,12 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     toprint = rbindlist(list(head(x, topn), tail(x, topn)), use.names=FALSE)  # no need to match names because head and tail of same x, and #3306
     rn = c(seq_len(topn), seq.int(to=n_x, length.out=topn))
     printdots = TRUE
-    if (indices) set(toprint, j=index_col_name, value=c(head(index_col, topn), tail(index_col, topn)))
+    if (print.indices) toprint = cbind(toprint, rbindlist(list(head(index_dt, topn), tail(index_dt, topn)), use.names=FALSE))
   } else {
     toprint = x
     rn = seq_len(n_x)
     printdots = FALSE
-    if (indices) set(toprint, j=index_col_name, value=index_col)
+    if (print.indices) toprint = cbind(toprint, index_dt)
   }
   toprint=format.data.table(toprint, na.encode=FALSE, timezone = timezone, ...)  # na.encode=FALSE so that NA in character cols print as <NA>
   require_bit64_if_needed(x)
@@ -282,4 +293,8 @@ trunc_cols_message = function(not_printed, abbs, class, col.names){
     "%d variable(s) not shown: %s\n",
     n, brackify(paste0(not_printed, classes))
   )
+}
+
+hello = function(x) {
+  print(names(x))
 }
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -294,7 +294,3 @@ trunc_cols_message = function(not_printed, abbs, class, col.names){
     n, brackify(paste0(not_printed, classes))
   )
 }
-
-hello = function(x) {
-  print(names(x))
-}
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -9,7 +9,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
                trunc.cols=getOption("datatable.print.trunc.cols"),
                quote=FALSE,
                na.print=NULL,
-               timezone=FALSE, ...) {
+               timezone=FALSE,
+               indices=FALSE, ...) {
   # topn  - print the top topn and bottom topn rows with '---' inbetween (5)
   # nrows - under this the whole (small) table is printed, unless topn is provided (100)
   # class - should column class be printed underneath column name? (FALSE)
@@ -64,15 +65,25 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     return(invisible(x))
   }
+  if (indices) {
+    if (is.null(indices(x))) {
+      indices = FALSE
+    } else {
+      index_col_name = paste0("index__", indices(x))
+      index_col = attr(attr(x, 'index'), paste0('__', indices(x)))
+    }
+  }
   n_x = nrow(x)
   if ((topn*2L+1L)<n_x && (n_x>nrows || !topnmiss)) {
     toprint = rbindlist(list(head(x, topn), tail(x, topn)), use.names=FALSE)  # no need to match names because head and tail of same x, and #3306
     rn = c(seq_len(topn), seq.int(to=n_x, length.out=topn))
     printdots = TRUE
+    if (indices) set(toprint, j=index_col_name, value=c(head(index_col, topn), tail(index_col, topn)))
   } else {
     toprint = x
     rn = seq_len(n_x)
     printdots = FALSE
+    if (indices) set(toprint, j=index_col_name, value=index_col)
   }
   toprint=format.data.table(toprint, na.encode=FALSE, timezone = timezone, ...)  # na.encode=FALSE so that NA in character cols print as <NA>
   require_bit64_if_needed(x)
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -7,7 +7,7 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
                col.names=getOption("datatable.print.colnames"),
                print.keys=getOption("datatable.print.keys"),
                trunc.cols=getOption("datatable.print.trunc.cols"),
-               print.indices=getOption("datatable.print.indices"),
+               show.indices=getOption("datatable.show.indices"),
                quote=FALSE,
                na.print=NULL,
                timezone=FALSE, ...) {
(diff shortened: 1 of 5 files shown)


tdhock:
looks ok to me, please add NEWS

Anirban166 (approved):

R/print.data.table.R
```diff
@@ -9,7 +9,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
                trunc.cols=getOption("datatable.print.trunc.cols"),
                quote=FALSE,
                na.print=NULL,
-               timezone=FALSE, ...) {
+               timezone=FALSE,
+               indices=FALSE, ...) {
```
MichaelChirico, line 13: looks somewhat inconsistent to have `datatable.print.keys` option but not `datatable.print.indices`, should we make this controlled by an option?

inst/tests/tests.Rraw
```diff
@@ -18612,3 +18612,49 @@ test(2263.3, options=list(datatable.verbose=TRUE, datatable.optimize=0L), names(
...
+ "  2:    1   24   60                 2",
+ "  3:    2   26   32                 5",
+ "  4:    2   36   57                 3",
+ "  5:    2   51   30                 4",
+ " ---                                 ",
+ "196:   98   77   45               195",
+ "197:   98   87   70               197",
+ "198:  100   18   21               198",
+ "199:  100   36   51               199",
+ "200:  100   38   56               200"
```
MichaelChirico, line 18639: nit: either put the ')' on this line, or indent the table body, not a fan of having ')' line up with `"` in the previous line :)

inst/tests/tests.Rraw
```diff
@@ -18612,3 +18612,49 @@ test(2263.3, options=list(datatable.verbose=TRUE, datatable.optimize=0L), names(
...
+NN = 200
+set.seed(2024)
+DT = data.table(
+ grp1 = sample(100, NN, TRUE),
+ grp2 = sample(90, NN, TRUE),
+ grp3 = sample(80, NN, TRUE)
+)
+setkey(DT, grp1, grp2)
+setindex(DT, grp1, grp3)
+dt2 = copy(DT)
```
MichaelChirico, line 18626: nit: `DT2` for consistent variable naming

inst/tests/tests.Rraw
```diff
@@ -18612,3 +18612,49 @@ test(2263.3, options=list(datatable.verbose=TRUE, datatable.optimize=0L), names(
...
+ "196:   98   77   45               195",
+ "197:   98   87   70               197",
+ "198:  100   18   21               198",
+ "199:  100   36   51               199",
+ "200:  100   38   56               200"
+ )
+test(2264.1, print(DT, indices=TRUE), output=ans)
+setindex(DT, NULL) # clear indices
+# if no indices are set, simply ignore
+test(2264.2, capture.output(print(DT, indices=TRUE)), capture.output(print(DT, indices=FALSE)))
```
MichaelChirico, line 18644: I wonder if using `notOutput=` would be clearer and/or easier here, WDYT?

inst/tests/tests.Rraw
```diff
@@ -18612,3 +18612,49 @@ test(2263.3, options=list(datatable.verbose=TRUE, datatable.optimize=0L), names(
...
+DT = data.table(
+ grp1 = sample(100, NN, TRUE),
+ grp2 = sample(90, NN, TRUE),
+ grp3 = sample(80, NN, TRUE)
+)
+setkey(DT, grp1, grp2)
+setindex(DT, grp1, grp3)
+dt2 = copy(DT)
+ans = c(
+ "     grp1 grp2 grp3 index__grp1__grp3",
```
MichaelChirico, line 18628: I wonder if it's possibly beneficial to use a different separator for the `index` stub vs. the columns themselves? E.g. we could do `index__grp1_grp3` or `index_grp1__grp3`. For the former, we'd do `paste0("index__", vapply_1c(indices(x, vectors=TRUE), paste, collapse="_"))`. Uglier but not awful.

inst/tests/tests.Rraw
MichaelChirico, line 1: missing: a test for what output looks like when there are multiple indices set, e.g. `setindex(DT, grp3, grp1)`. IIUC that will give a new column `index__grp3__grp1`. Do you think it's an improvement to include a marker for index sequence? `index1__grp1__grp3`, `index2__grp3__grp1`?

R/print.data.table.R
joshhwuu: sure! should i rename the argument to `print.indices` for consistency as well?

inst/tests/tests.Rraw
joshhwuu: Sorry that wasn't actually supposed to be there, slipped through by accident

inst/tests/tests.Rraw
joshhwuu: Right, this makes sense, I'll see what I can do

joshhwuu: Looks like a helper function would be useful here! WDYT about something like:

```R
find_indices = function(x, topn) {
    idx_names = indices(x)
    if (!length(idx_names)) return(x)
    index_dt = data.table()
    if (length(idx_names)>1L) {
        count = 1
        for (idx_name in idx_names) {
            index_col_value = attr(attr(x, 'index'), paste0('__', idx_name))
            index_dt[, paste0("index", count, "__", idx_name) := index_col_value]
            count = count + 1
        }
    } else index_dt[, paste0("index__", idx_names) := attr(attr(x, 'index'), paste0('__', idx_names))]
    if (topn) {
        x = cbind(x, rbindlist(list(head(index_dt, topn), tail(index_dt, topn)), use.names=FALSE))
    } else {
        x = cbind(x, index_dt)
    }
    x
}
```
This allows us to assign `toprint` index columns and allows us to control the column naming easily, I thought it was a pretty elegant solution, LMK if this works.

joshhwuu: nvm, was just able to add this into the print function without problem, doesn't run unless `indices` is TRUE

inst/tests/tests.Rraw
joshhwuu: I like the look of the former better, but I think it would be beneficial for our column names to accurately reflect the output of `indices(x)`, like `'grp1__grp3'`. This also makes sense as the existing printing of indices show indices like: `Indices: <grp1__grp3>`

inst/tests/tests.Rraw
joshhwuu: Sorry, its not very obvious to me how to use notOutput here, this test is to check that when we try to print the indices of a dt without indices, the `indices` arg is ignored and the output is the same as if we just printed normally (without indices)

R/print.data.table.R
joshhwuu: changed to print.indices and added as a new option

inst/tests/tests.Rraw
MichaelChirico: Since it's a pseudo-column, we might want to do more to emphasize that the index can't actually be accessed as a normal column, maybe making the name non-syntactic is enough (and also helps w the issue raised here):

`index:grp1__grp3` or even `` `index:grp1__grp3` ``

inst/tests/tests.Rraw
MichaelChirico: `notOutput="index__grp1__grp3"` would be enough, no?

R/print.data.table.R
```diff
@@ -7,6 +7,7 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
                col.names=getOption("datatable.print.colnames"),
                print.keys=getOption("datatable.print.keys"),
                trunc.cols=getOption("datatable.print.trunc.cols"),
+               print.indices=getOption("datatable.print.indices"),
```
MichaelChirico, line 10: Hmm, this new name is worse IMO, because `print.keys` already handles printing the `Key:` and `Indices:` summaries above the table. WDYT about `show.indices` or even `show.indices.values`/`show.index.values`?

R/print.data.table.R
```diff
@@ -64,15 +65,36 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+    } else {
+      index_dt = data.table()
+      index_names = indices(x)
+      if (length(index_names) > 1L) {
+        count = 1L
+        for (index_name in index_names) {
+          index_col = attr(attr(x, 'index'), paste0('__', index_name))
+          index_dt[, paste0("index", count, "__", index_name) := index_col]
+          count = count + 1L
+        }
```
MichaelChirico, line 81:
```suggestion
        index_dt[, paste0("index", seq_along(index_names), "__", index_names) := attributes(attr(x, 'index'))]
```

R/print.data.table.R
```diff
@@ -64,15 +65,36 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
   if ((topn*2L+1L)<n_x && (n_x>nrows || !topnmiss)) {
     toprint = rbindlist(list(head(x, topn), tail(x, topn)), use.names=FALSE)  # no need to match names because head and tail of same x, and #3306
     rn = c(seq_len(topn), seq.int(to=n_x, length.out=topn))
     printdots = TRUE
+    if (print.indices) toprint = cbind(toprint, rbindlist(list(head(index_dt, topn), tail(index_dt, topn)), use.names=FALSE))
   } else {
     toprint = x
     rn = seq_len(n_x)
     printdots = FALSE
+    if (print.indices) toprint = cbind(toprint, index_dt)
```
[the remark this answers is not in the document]
MichaelChirico: there's some efficiency concern about materializing all `nrows()` index values only to throw all but `2*topn` away, but I think we can ignore that for now -- this is not the default way of printing, and should only be used interactively for inspection purposes, not performance-critical unless specifically requested.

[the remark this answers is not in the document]
MichaelChirico: NVM I'm misreading. Still not a fan of the repetitive ugly code `rbindlist(list(head(., topn), tail(., topn)), ...)`...

inst/tests/tests.Rraw
joshhwuu: Yes this looks great

inst/tests/tests.Rraw
joshhwuu: ah I see, my understanding of what `notOutput` did was wrong, thanks!

R/print.data.table.R
joshhwuu: I think I prefer `show.indices`, just because I think something like `show.index.values` might be a bit wordy to read if the user sets it as an option, ie `options(datatable.show.index.values=TRUE)`

R/print.data.table.R
joshhwuu: Much better, thanks!

R/print.data.table.R
```diff
@@ -64,15 +65,29 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     return(invisible(x))
   }
+  index_dt = NULL
+  if (show.indices) {
+    if (is.null(indices(x))) {
+      show.indices = FALSE
+    } else {
+      index_dt = data.table()
```
MichaelChirico, line 73:
```suggestion
```

R/print.data.table.R
```diff
@@ -64,15 +65,28 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
     return(invisible(x))
   }
+  index_dt = NULL
+  if (show.indices) {
+    if (is.null(indices(x))) {
+      show.indices = FALSE
+    } else {
+      index_names = indices(x)
+      index_dt <- as.data.table(attributes(attr(x, 'index')))
+      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_along(index_names) else "", ":", sub("__", "", names(index_dt)))
```
MichaelChirico, line 75:
```suggestion
      index_dt <- as.data.table(attributes(attr(x, 'index')))
      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt), fixed=TRUE))
```

R/print.data.table.R
```diff
@@ -64,15 +65,27 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     return(invisible(x))
   }
+  index_dt = NULL
+  if (show.indices) {
+    if (is.null(indices(x))) {
+      show.indices = FALSE
+    } else {
+      index_dt <- as.data.table(attributes(attr(x, 'index')))
+      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt), fixed=TRUE))
```
MichaelChirico, line 74: I'm wondering if it's worth maintaining the complexity of differentiating the n_index >1 vs =1 cases. Without it we can simplify:

```r
sprintf("index%d:%s", seq_len(ncol(index_dt)), sub("^__", "", names(index_dt)))
```

MichaelChirico, line 74:
```suggestion
      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt)))
```

R/print.data.table.R
```diff
@@ -64,15 +65,27 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt)))
+      setnames(index_dt, print_names)
+    }
+  }
   n_x = nrow(x)
   if ((topn*2L+1L)<n_x && (n_x>nrows || !topnmiss)) {
     toprint = rbindlist(list(head(x, topn), tail(x, topn)), use.names=FALSE)  # no need to match names because head and tail of same x, and #3306
     rn = c(seq_len(topn), seq.int(to=n_x, length.out=topn))
     printdots = TRUE
+    if (show.indices) toprint = cbind(toprint, rbindlist(list(head(index_dt, topn), tail(index_dt, topn)), use.names=FALSE))
```
MichaelChirico, line 83: two suggested alternatives to maybe make this a bit cleaner:

#### 1

```r
if (show.indices) index_dt = rbindlist(...)

if (show.indices) toprint = cbind(toprint, index_dt) # i.e. use the same expression just once, outside the if/else branch
```

#### 2

```r
idx = c(seq_len(topn), seq(to=nrow(x), length.out=topn))
toprint = x[idx, ]
if (show.indices) toprint = cbind(toprint, index_dt[idx, ])
```

---

At a glance the latter is a bit more readable and also more efficient than the `rbindlist()` approach:

```r
DT=setDT(lapply(1:10, \(.) rnorm(1e7)))
microbenchmark(times = 20,
  rbindlist=rbindlist(list(head(DT, 5), tail(DT, 5)), use.names=FALSE),
  seq={idx = c(seq_len(5), seq(nrow(DT)-5+1, length.out=5)); DT[idx, ]})
# Unit: microseconds
#       expr     min      lq     mean  median       uq     max neval cld
#  rbindlist 489.090 518.006 574.3189 537.609 589.5205 991.939    20   b
#        seq 171.324 188.492 215.6645 206.098 233.0890 314.695    20  a 
```

joshhwuu: the latter works for me!

R/print.data.table.R
joshhwuu: IMO the distinction feels a bit more polished and to keep it doesn't introduce a whole lot of complexity, so I'd rather keep it

R/print.data.table.R
```diff
@@ -64,15 +65,29 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     }
     return(invisible(x))
   }
+  index_dt = NULL
```
MichaelChirico, line 68: I wouldn't do this -- R lazy eval means we don't need to instantiate a variable in unvisited branches, so IMO it's just a distraction

MichaelChirico, line 68:
```suggestion
```

MichaelChirico (approved):
LGTM thanks!

joshhwuu:
Thanks Michael for the great review! Very insightful as always 🙂

PR #6189, opened 2024-06-18
Clarify Integer Key Behavior in `setkey.Rd`

### LINKED ISSUE #5074: Suggestion: Warn when making integer column key
Problem: When using `[` to perform a join between a data.table with an integer key and an integer vector, the result is not a merge.

Example:
```R
require(magrittr)
require(data.table)

## Construct a data.table with two columns of possible index values. 
## The columns are logically identical, but expressed differently.
## Column 1 "idchar" is CHARACTER
## Column 2 "idint" is INTEGER
ex <- data.table(idchar = rep(1:5, each=2) %>% sprintf("%02.0f", .),
                 idint = rep(1:5, each=2))
# > ex
#     idchar idint
# 1:      01     1
# 2:      01     1
# 3:      02     2
# 4:      02     2
# 5:      03     3
# 6:      03     3
# 7:      04     4
# 8:      04     4
# 9:      05     5
# 10:     05     5

## Both columns contain duplicates, which can be found with `duplicated`
dupes_char <- ex[idchar %>% duplicated, idchar]
dupes_int <- ex[idint %>% duplicated, idint]

## Doing a merge on the character duplicates gives the expected result:
setkey(ex, idchar)
ex[dupes_char]

# > ex[dupes_char]
#     idchar idint
# 1:      01     1
# 2:      01     1
# 3:      02     2
# 4:      02     2
# 5:      03     3
# 6:      03     3
# 7:      04     4
# 8:      04     4
# 9:      05     5
# 10:     05     5

## Doing a merge on the integer duplicates gives an **unexpected** result:
setkey(ex, idint)
ex[dupes_int]

# > ex[dupes_int]
#    idchar idint
# 1:     01     1
# 2:     01     1
# 3:     02     2
# 4:     02     2
# 5:     03     3
```

This example is easy to miss in more complex / larger examples. 

The problem happens because the integers are interpreted as index values,  rather than as a vector that should be compared to the key.

I think the *real* problem is that it's a bad idea to use an integer column as a key, because this could be misinterpreted in the future. 

I propose that the step `setkey(ex, idint)` issues some kind of warning. For example, it might say "The `[` syntax will fail when using integer keys"

Nj221102:
closes #5074 

### Description

This PR updates the documentation of `setkey.Rd` to emphasize the importance of using the dot (`.`) syntax when employing integer columns as keys. The documentation now explicitly advises users to use `.[(subset_value)]` instead of `[subset_value]` to ensure correct behavior in subsetting operations. This clarification aims to prevent potential issues where integers are misinterpreted as indices rather than values against the key column.

### DIFF
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -23,7 +23,7 @@ There are three reasons \code{setkey} is desirable:
 
 \code{NA}s are always first because:
 \itemize{
-  \item \code{NA} is internally \code{INT_MIN} (a large negative number) in R. Keys and indexes are always in increasing order so if \code{NA}s are first, no special treatment or branch is needed in many \code{data.table} internals involving binary search. It is not optional to place \code{NA}s last for speed, simplicity and rubustness of internals at C level.
+  \item \code{NA} is internally \code{INT_MIN} (a large negative number) in R. Keys and indexes are always in increasing order so if \code{NA}s are first, no special treatment or branch is needed in many \code{data.table} internals involving binary search. It is not optional to place \code{NA}s last for speed, simplicity and robustness of internals at C level.
   \item if any \code{NA}s are present then we believe it is better to display them up front (rather than hiding them at the end) to reduce the risk of not realizing \code{NA}s are present.
 }
 
@@ -87,7 +87,9 @@ required.)
 If you really wish to use column numbers, it is possible but
 deliberately a little harder; e.g., \code{setkeyv(DT,names(DT)[1:2])}.
 
-If you use integer columns as keys, it's crucial to ensure correct behavior in subsetting and joining operations. Integer keys should be used with the dot (\code{.}) syntax to explicitly match values against the key rather than interpreting them as indices.
+If you use integer columns as keys, it's crucial to ensure correct behavior in subsetting and joining operations.
+Integer keys should be used with the dot (\code{.}) syntax to explicitly match values against the key 
+rather than interpreting them as indices. see examples for better understanding.
 
 If you wanted to use \code{\link[base]{grep}} to select key columns according to
 a pattern, note that you can just set \code{value = TRUE} to return a character vector instead of the default integer indices.
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -89,7 +89,7 @@ deliberately a little harder; e.g., \code{setkeyv(DT,names(DT)[1:2])}.
 
 If you use integer columns as keys, it's crucial to ensure correct behavior in subsetting and joining operations.
 Integer keys should be used with the dot (\code{.}) syntax to explicitly match values against the key 
-rather than interpreting them as indices. see examples for better understanding.
+rather than interpreting them as indices. See examples for better understanding.
 
 If you wanted to use \code{\link[base]{grep}} to select key columns according to
 a pattern, note that you can just set \code{value = TRUE} to return a character vector instead of the default integer indices.
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -145,11 +145,11 @@ indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
 
 # Example illustrating the importance of using the dot (`.`) syntax with integer keys:
-DT <- data.table(id = 1:5, B = letters[5:1])
+DT <- data.table(id = 2:1)
 
 # Incorrect Usage:
 setkey(DT, id)
-subset_value <- 3
+subset_value <- 1
 DT[subset_value]  # Incorrect: treats `subset_value` as an index
 
 # Correct Usage:
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -148,9 +148,9 @@ indices(DT, vectors = TRUE) # get indices list
 DT = data.table(id = 2:1)
 setkey(DT, id)
 subset_value <- 1
-DT[subset_value]  # Incorrect: treats `subset_value` as an row number
+DT[subset_value]  # treats `subset_value` as an row number
 # To correctly subset based on the key column `id`, use the dot (`.`) syntax:
-DT[.(subset_value)]  # Correct: matches `subset_value` against `id`
+DT[.(subset_value)]  # matches `subset_value` against `id`
 
 }
 \keyword{ data }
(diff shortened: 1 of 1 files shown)


man/setkey.Rd
```diff
@@ -87,6 +87,10 @@ required.)
 If you really wish to use column numbers, it is possible but
 deliberately a little harder; e.g., \code{setkeyv(DT,names(DT)[1:2])}.
 
+If you use integer columns as keys, it's crucial to ensure correct behavior in subsetting and joining operations.
+Integer keys should be used with the dot (\code{.}) syntax to explicitly match values against the key 
+rather than interpreting them as indices. see examples for better understanding.
```
Anirban166, line 92: See*

Nj221102: Done 👍

tdhock (changes requested):
in your example there is no difference between the output with or without the dot. 
for clarity, please change to an example where there is a difference, for example:
```r
> DT <- data.table(id = 2:1)
> subset_value <- 1
> DT[subset_value]
      id
   <int>
1:     2
> setkey(DT, id)
> DT[.(subset_value)]
Key: <id>
      id
   <int>
1:     1
```

Nj221102:
> in your example there is no difference between the output with or without the dot. for clarity, please change to an example where there is a difference, for example:
> 
> ```r
> > DT <- data.table(id = 2:1)
> > subset_value <- 1
> > DT[subset_value]
>       id
>    <int>
> 1:     2
> > setkey(DT, id)
> > DT[.(subset_value)]
> Key: <id>
>       id
>    <int>
> 1:     1
> ```

done :)

man/setkey.Rd
```diff
@@ -87,6 +87,10 @@ required.)
 If you really wish to use column numbers, it is possible but
 deliberately a little harder; e.g., \code{setkeyv(DT,names(DT)[1:2])}.
 
+If you use integer columns as keys, it's crucial to ensure correct behavior in subsetting and joining operations.
+Integer keys should be used with the dot (\code{.}) syntax to explicitly match values against the key 
+rather than interpreting them as indices. See examples for better understanding.
```
tdhock, line 92: "indices" here means row number and that is different from the function `indices`
so can you please change indices to row numbers here and in the comments below?

tdhock (changes requested):

man/setkey.Rd
```diff
@@ -139,5 +143,18 @@ setindex(DT, A)
...
+DT <- data.table(id = 2:1)
+
+# Incorrect Usage:
+setkey(DT, id)
+subset_value <- 1
+DT[subset_value]  # Incorrect: treats `subset_value` as an row number
+
+# Correct Usage:
+# To correctly subset based on the key column `id`, use the dot (`.`) syntax:
+DT[.(subset_value)]  # Correct: explicitly matches `subset_value` against `id`
```
tdhock, line 157: I'm not sure about the word "explicit" here
I think `DT[.(subset_value), on="id"]` would be explicit and current code (without on arg) could be considered "implicit" can you please change here and above? or just delete explicit?

tdhock (changes requested):

man/setkey.Rd
Nj221102: done :)

Nj221102:
@tdhock  take a look if this needs further changes plz suggest :)

man/setkey.Rd
```diff
@@ -139,5 +143,18 @@ setindex(DT, A)
...
+
+# Example illustrating the importance of using the dot (`.`) syntax with integer keys:
+DT <- data.table(id = 2:1)
+
+# Incorrect Usage:
+setkey(DT, id)
+subset_value <- 1
+DT[subset_value]  # Incorrect: treats `subset_value` as an row number
+
+# Correct Usage:
```
tdhock, line 155: please delete correct/incorrect

tdhock (changes requested):
also please add a NEWS item

man/setkey.Rd
Nj221102: done 👍

Nj221102:
> also please add a NEWS item

news entry for a documentation update?
I am bit confused because i didn't had to do it for the previous PR's related to documentation.

man/setkey.Rd
```diff
@@ -139,5 +143,14 @@ setindex(DT, A)
...
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
+
+# Example illustrating the importance of using the dot (`.`) syntax with integer keys:
+DT = data.table(id = 2:1)
+setkey(DT, id)
+subset_value <- 1
+DT[subset_value]  # Incorrect: treats `subset_value` as an row number
+# To correctly subset based on the key column `id`, use the dot (`.`) syntax:
+DT[.(subset_value)]  # Correct: matches `subset_value` against `id`
```
tdhock, line 153: please remove correct/incorrect

tdhock (changes requested):

man/setkey.Rd
tdhock: please remove the words correct and incorrect from comments, because they could be confusing.
As I see it both lines of codes are correct but for different purposes (select by row number or by matching to key)

Nj221102: thanks for clarification, done 👍

man/setkey.Rd
```diff
@@ -144,13 +142,12 @@ setindex(DT, B)
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
 
-# Example illustrating the importance of using the dot (`.`) syntax with integer keys:
+# Use the dot .(subset_value) syntax with integer keys:
 DT = data.table(id = 2:1)
 setkey(DT, id)
 subset_value <- 1
-DT[subset_value]  # treats `subset_value` as an row number
-# To correctly subset based on the key column `id`, use the dot (`.`) syntax:
```
tdhock: deleted "correctly"

tdhock:
hi, I committed an attempt to further clarify, can you please review?

Nj221102:
> hi, I committed an attempt to further clarify, can you please review?

Yeah looks good to me, thanks for helping :)

tdhock:
great thanks

ben-schwen:
@tdhock could you please squash when merging? AFAIU this is our intended way to not clutter our commit history

tdhock:
sure, sorry about that

PR #6190, opened 2024-06-18
Added Comprehensive Examples for set Function in Documentation

### LINKED ISSUE #5031: add examples of `set` function in documentation

Nj221102:
## Description:
This PR enhances the documentation for the `set` function in the `data.table` package by adding detailed examples. These examples demonstrate the usage of `set` for assigning values to multiple columns and handling multiple values efficiently. The aim is to provide clear and comprehensive usage scenarios, thereby improving user understanding and ensuring best practices.

## Changes Made:
1. Added detailed examples to the `set` function documentation.
2. Demonstrated assignment to multiple columns of multiple values.

closes #5031

### DIFF
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -147,6 +147,10 @@ set(DT, 1L, "b", 10L)
 set(DT, 2L, c("b", "d"), list(20L, 30L))
 ## Set values by column indices
 set(DT, 3L, c(2L, 4L), list(40L, 50L))
+## Set value for an entire column without specifying rows
+set(DT, j = "b", value = 100L)
+## Set values for multiple columns without specifying rows
+set(DT, j = c("b", "d"), value = list(200L, 300L))
 
 \dontrun{
 # Speed example:
--- a/man/assign.Rd
+++ b/man/assign.Rd
@@ -153,7 +153,7 @@ set(DT, NULL, "b", 100L) # equivalent
 ## Set values for multiple columns without specifying rows
 set(DT, j = c("b", "d"), value = list(200L, 300L))
 ## Set values for multiple columns with multiple specified rows.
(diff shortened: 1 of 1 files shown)


tdhock:
Would be good to add an example using `set(DT, j=some_name, value=some_value)` which is how I typically use set (without i argument)

Nj221102:
> Would be good to add an example using `set(DT, j=some_name, value=some_value)` which is how I typically use set (without i argument)
Sure :)

man/assign.Rd
```diff
@@ -140,6 +140,18 @@ sq_col_idx = grep('d$', names(DT))
...
 
+# Examples using `set` function
+## Set value for single cell
+set(DT, 1L, "b", 10L)
+## Set values for multiple columns in a specific row
+set(DT, 2L, c("b", "d"), list(20L, 30L))
+## Set values by column indices
+set(DT, 3L, c(2L, 4L), list(40L, 50L))
+## Set value for an entire column without specifying rows
+set(DT, j = "b", value = 100L)
```
MichaelChirico, line 151:
```suggestion
set(DT, j = "b", value = 100L)
set(DT, NULL, "b", 100L) # equivalent
```

MichaelChirico:
LGTM. Should there be an example with >1 (but less than `nrow`) rows being assigned? Or is that obvious enough from the existing example?

Nj221102:
> LGTM. Should there be an example with >1 (but less than `nrow`) rows being assigned? Or is that obvious enough from the existing example?

Let me just add that as well, more examples will help people understand better, even though its pretty obvious but there is no harm in one  more example i think.

Nj221102:
> LGTM. Should there be an example with >1 (but less than `nrow`) rows being assigned? Or is that obvious enough from the existing example?

done :)

MichaelChirico (approved):
Great, thanks!

PR #6193, opened 2024-06-19
Clarify Usage of `.BY` in Special Symbols Documentation

### LINKED ISSUE #5389: Unexpected behavior in `.BY`
Hi, 

we are trying to extract the `.BY` information to a list object outside the current data.table. Unfortunately, we get unexpected results.
All elements of the list of length(by groups) contain the same value - the .BY information from the last group.
When using .GRP instead, the list is populated with different values - the correct group indices. 

Is this expected behavior, and how can I get a list of correct by-group values be extracted?

Thanks!

``` r
library(data.table)

dt = data.table(l = sample(letters[1:5], 100, replace = TRUE), b = rnorm(100))
setkey(dt, l)

byg = list()
grp = list()
dt[
  , m := {
    byg <<- append(byg, .BY)
    grp <<- append(grp, .GRP)
    mean(b)
  }
  , by = l
]

str(byg)
#> List of 5
#>  $ l: chr "e"
#>  $ l: chr "e"
#>  $ l: chr "e"
#>  $ l: chr "e"
#>  $ l: chr "e"
str(grp)
#> List of 5
#>  $ : int 1
#>  $ : int 2
#>  $ : int 3
#>  $ : int 4
#>  $ : int 5

sessionInfo()
#> R version 4.2.0 (2022-04-22)
#> Running under: Ubuntu 20.04.4 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
#> LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0
#> 
#> locale:
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.14.2
#> 
#>  [1] rstudioapi_0.13   knitr_1.39        magrittr_2.0.3    R.cache_0.15.0   
#>  [5] rlang_1.0.2       fastmap_1.1.0     fansi_1.0.3       stringr_1.4.0    
#>  [9] styler_1.7.0      highr_0.9         tools_4.2.0       xfun_0.31        
#> [13] R.oo_1.24.0       utf8_1.2.2        cli_3.3.0         withr_2.5.0      
#> [17] htmltools_0.5.2   ellipsis_0.3.2    yaml_2.3.5        digest_0.6.29    
#> [21] tibble_3.1.7      lifecycle_1.0.1   crayon_1.5.1      purrr_0.3.4      
#> [25] R.utils_2.11.0    vctrs_0.4.1       fs_1.5.2          glue_1.6.2       
#> [29] evaluate_0.15     rmarkdown_2.14    reprex_2.0.1      stringi_1.7.6    
#> [33] compiler_4.2.0    pillar_1.7.0      R.methodsS3_1.8.1 pkgconfig_2.0.3
```

<sup>Created on 2022-05-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Nj221102:
closes #5389 

**Description:**

This PR updates the `special-symbols` documentation to make it clear how to use `.BY` correctly in `data.table`. The main goal is to help users understand how to avoid issues when using `.BY` outside of its immediate context, as discussed in GitHub Issue #5389.

**Changes:**
- Added a Note in the `details` section to include advice on using `copy(.BY)` when extracting `.BY` outside its immediate context.
- Added an example to show the correct usage of `copy(.BY)`.


man/special-symbols.Rd
```diff
@@ -66,5 +68,25 @@ DT[{cat(sprintf('in i, .N is \%d\n', .N)); a < .N/2},
 # .I can be different in j and by, enabling rowwise operations in by
 DT[, .(.I, min(.SD[,-1]))]
 DT[, .(min(.SD[,-1])), by=.I]
+
+#These examples illustrate the importance of using copy(.BY) when capturing .BY values in a list outside the data.table context to avoid unexpected behavior.
+dt <- data.table(l = sample(letters[1:5], 100, replace = TRUE), b = rnorm(100))
```
tdhock, line 73: hi, this is a good start, but how about a simpler example, with fewer than 100 rows?
how about 4 rows, data.table(letter=c('a','a','b','b'),number=1:4)

tdhock (changes requested):

man/special-symbols.Rd
Nj221102: done :)

man/special-symbols.Rd
```diff
@@ -66,5 +68,25 @@ DT[{cat(sprintf('in i, .N is \%d\n', .N)); a < .N/2},
 # .I can be different in j and by, enabling rowwise operations in by
 DT[, .(.I, min(.SD[,-1]))]
 DT[, .(min(.SD[,-1])), by=.I]
+
+#These examples illustrate the importance of using copy(.BY) when capturing .BY values in a list outside the data.table context to avoid unexpected behavior.
```
Anirban166, line 72: Minor nit: Space after #

Anirban166 (approved):

man/special-symbols.Rd
Nj221102: Done :)

MichaelChirico:
Hi @Nj221102, FYI, there's no need to do the "Update" step for each individual PR every time `master` updates, it creates a lot of e-mail churn. If there are merge conflicts to resolve, go ahead and do so, otherwise your reviewer can do the update step easily when they review.

Nj221102:
> Hi @Nj221102, FYI, there's no need to do the "Update" step for each individual PR every time `master` updates, it creates a lot of e-mail churn. If there are merge conflicts to resolve, go ahead and do so, otherwise your reviewer can do the update step easily when they review.

Sure will keep that in mind from now on , sorry for the inconvenience caused

Nj221102:
@tdhock plz check this out as well.

man/special-symbols.Rd
```diff
@@ -66,5 +68,25 @@ DT[{cat(sprintf('in i, .N is \%d\n', .N)); a < .N/2},
 # .I can be different in j and by, enabling rowwise operations in by
 DT[, .(.I, min(.SD[,-1]))]
 DT[, .(min(.SD[,-1])), by=.I]
+
+# These examples illustrate the importance of using copy(.BY) when capturing .BY values in a list outside the data.table context to avoid unexpected behavior.
+dt <- data.table(l = letters[1:5], b = rnorm(5))
```
tdhock, line 73: again please use simpler data (1:4 instead of rnorm) as I mentioned in my previous comment

man/special-symbols.Rd
```diff
@@ -66,5 +68,25 @@ DT[{cat(sprintf('in i, .N is \%d\n', .N)); a < .N/2},
...
+byg <- list()
+dt[, m := {
+  byg <<- append(byg, .BY)  # Incorrect: Missing copy(.BY)
+  mean(b)
+}, by = l]
+str(byg)  # All elements of 'byg' will contain the .BY value of the last group
+
+## Correct usage
+byg <- list()
+dt[, m := {
```
tdhock, line 86: these examples are perhaps too long.
If I understand correctly, the important different is .BY versus copy(.BY) right?
So can you please delete as much other code (:= append, mean)  as possible to make the example simpler? ideally one line instead of 6 for both incorrect/correct usage

tdhock (changes requested):

MichaelChirico:
I've updated here to be as brief as possible, while also more directly addressing the root cause & generality of the issue.

This (AFAIK) is _not_ strictly related to `.BY`, `.BY` is just one example of the issue of expecting to use the special symbol values, which are generated in a `by=` context, outside of that context.

I think it's enough to call out the issue & mention the workaround (`copy()`).

MichaelChirico (approved):
Thank you!

PR #6195, opened 2024-06-20
Enhance Examples and Documentation for `fromLast` in `duplicated.Rd`

### LINKED ISSUE #5130: Improve the documentation for fromLast in unique
I have used `unique` functions quite some times, and I cannot figure out `fromLast = TRUE` means pick first or last record for unique by reading the help document. Every time I need to either run the example or test with code.

If you look at these two lines of example, from the surface people can read from intuition that it means picking duplicated rows fromLast, and picking unique rows fromLast, which will have totally reversed results.

```
duplicated(DT, by="B", fromLast=TRUE)
unique(DT, by="B", fromLast=TRUE)
```

The help document didn't really clarify which side was kept

> `fromLast`   logical indicating if duplication should be considered from the reverse side, i.e., the last (or rightmost) of identical elements would correspond to duplicated = FALSE.

Today I dug deeper and found out the source of confusion. The parameter and the help line came from base variant of `duplicated`. However the base variant of `unique` is a separate function and have slightly different wordings for same parameter, which is very clear and easy to understand.

> `fromLast`  logical indicating if duplication should be considered from the last, i.e., the last (or rightmost) of identical elements will be kept. 

Now in `data.table` they are grouped together and help document combined, so the help line will be very confusing for `unique` users if they haven't used base version before (I always use data.table so I only read the data.table version help).

I think we should improve the documentation to something like this:

> `fromLast`   logical indicating if duplication should be considered from the reverse side, i.e., the last (or rightmost) of identical elements would correspond to duplicated = FALSE in `duplicated`, and be kept in `unique`.

Nj221102:
closes #5130 

### Description

This PR enhances the documentation and examples for the `fromLast` parameter of `duplicated` and `unique` functions. It aims to provide clearer explanations and concise examples to help users understand the behavior of `duplicated` and `unique` functions with `fromLast` set to `TRUE` and `FALSE`.

#### Changes Made

1. **Documentation Update**:
   - Added a detailed explanation for the `fromLast` parameter, specifying its behavior for both `duplicated` and `unique` functions.
  
2. **Examples Update**:
   - Provided a concise example with a smaller `data.table` to illustrate all cases when `fromLast` is `TRUE` and `FALSE` for both `duplicated` and `unique` functions.


man/duplicated.Rd
```diff
@@ -40,9 +40,10 @@ uniqueN(x, by=if (is.list(x)) seq_along(x) else NULL, na.rm=FALSE)
 as well.}
 \item{\dots}{ Not used at this time. }
 \item{incomparables}{ Not used. Here for S3 method consistency. }
-\item{fromLast}{ logical indicating if duplication should be considered from
-the reverse side, i.e., the last (or rightmost) of identical elements would
-correspond to \code{duplicated = FALSE}.}
+\item{fromLast}{Logical indicating if duplication should be considered from
+the reverse side. For \code{duplicated}, this means the last (or rightmost) of identical elements 
+will correspond to \code{duplicated = FALSE}. For \code{unique}, this means the last 
+(or rightmost) of identical elements will be kept. Checkout examples for better understanding.}
```
MichaelChirico, line 46:
```suggestion
(or rightmost) of identical elements will be kept. See examples.}
```

man/duplicated.Rd
```diff
@@ -111,11 +112,17 @@ DT[,which.min(a)]            # row 10, the strictly smallest floating point valu
 identical(unique(DT),DT[1])  # TRUE, stable within tolerance
 identical(unique(DT),DT[10]) # FALSE
 
-# fromLast=TRUE
-DT <- data.table(A = rep(1:3, each=4), B = rep(1:4, each=3),
-                 C = rep(1:2, 6), key = c("A", "B"))
-duplicated(DT, by="B", fromLast=TRUE)
-unique(DT, by="B", fromLast=TRUE)
+# Explaining all usecases of fromLast in both 'duplicated' and 'unique' function.
```
MichaelChirico, line 115: I'm not sure any change to the examples is warranted. I find the new examples too wordy; OP only requested we update the description of `fromLast`, I don't think we need to update the examples.

cc @tdhock do you think there's anything worth adding to the examples? I find the original clear & concise enough.

Nj221102: Yeah if it's too wordy, we can remove them

tdhock: I think the unique example is helpful

tdhock (approved):

joshhwuu:
Getting this `NOTE` from `R CMD Check` following these changes:

```
* checking Rd line widths ... NOTE
Rd file 'duplicated.Rd':
  \examples lines wider than 100 characters:
     duplicated(DT, by="B", fromLast=FALSE) # Explanation: Returns TRUE for rows 4 and 5 as these are duplicates of the first occurrences of ... [TRUNCATED]
     duplicated(DT, by="B", fromLast=TRUE) # Explanation: Returns TRUE for rows 1 and 2 as these are duplicates of the last occurrences of ' ... [TRUNCATED]
     unique(DT, by="B", fromLast=FALSE) # Explanation: Keeps rows 1, 2, and 3, which are the first occurrences of each unique 'B' value
     unique(DT, by="B", fromLast=TRUE) # Explanation: Keeps rows  3, 4 and 5, which are the last occurrences of each unique 'B' value
```
Not breaking but I wonder if this will affect how the manual is rendered?

Nj221102:
> Getting this `NOTE` from `R CMD Check` following these changes:
> 
> ```
> * checking Rd line widths ... NOTE
> Rd file 'duplicated.Rd':
>   \examples lines wider than 100 characters:
>      duplicated(DT, by="B", fromLast=FALSE) # Explanation: Returns TRUE for rows 4 and 5 as these are duplicates of the first occurrences of ... [TRUNCATED]
>      duplicated(DT, by="B", fromLast=TRUE) # Explanation: Returns TRUE for rows 1 and 2 as these are duplicates of the last occurrences of ' ... [TRUNCATED]
>      unique(DT, by="B", fromLast=FALSE) # Explanation: Keeps rows 1, 2, and 3, which are the first occurrences of each unique 'B' value
>      unique(DT, by="B", fromLast=TRUE) # Explanation: Keeps rows  3, 4 and 5, which are the last occurrences of each unique 'B' value
> ```
> 
> Not breaking but I wonder if this will affect how the manual is rendered?

Thanks for info, working on it 👍

PR #6196, opened 2024-06-20
Clarify `T/F` versus `TRUE/FALSE` usage in `j` argument with new FAQ entry*

### LINKED ISSUE #4846: TRUE/FALSE and abbreviated variants T/F have different behavior when used in j
When using a logical vector in `j` to select columns dynamically ([StackOverflow](https://stackoverflow.com/questions/33301901/select-columns-in-data-table-based-on-logical-vector/33301945)), using the abbreviations `T`/`F` has different behavior than using `TRUE`/`FALSE`

`#` `Minimal reproducible example`

``` r
library(data.table)
#> Warning: package 'data.table' was built under R version 4.0.3

DT = data.table(x=rep(c("a","b","c"),each=3), y=c(1,3,6), v=1:9)
DT[, c(TRUE, TRUE, FALSE), verbose = TRUE]
#>    x y
#> 1: a 1
#> 2: a 3
#> 3: a 6
#> 4: b 1
#> 5: b 3
#> 6: b 6
#> 7: c 1
#> 8: c 3
#> 9: c 6
DT[, c(T, T, F), verbose = TRUE]
#> Detected that j uses these columns: <none>
#> [1]  TRUE  TRUE FALSE

# `with = FALSE` still works as expected, though should no longer be necessary
# per https://stackoverflow.com/a/33301945/6197649
DT[, c(T, T, F), with = FALSE, verbose = TRUE]
#>    x y
#> 1: a 1
#> 2: a 3
#> 3: a 6
#> 4: b 1
#> 5: b 3
#> 6: b 6
#> 7: c 1
#> 8: c 3
#> 9: c 6
```

`#` `Output of sessionInfo()`

```r
sessionInfo()
#> R version 4.0.2 (2020-06-22)
#> Running under: Windows 10 x64 (build 18363)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_Europe.1252  LC_CTYPE=English_Europe.1252   
#> [3] LC_MONETARY=English_Europe.1252 LC_NUMERIC=C                   
#> [5] LC_TIME=English_Europe.1252    
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] data.table_1.13.4
#> 
#>  [1] compiler_4.0.2  magrittr_1.5    tools_4.0.2     htmltools_0.5.0
#>  [5] yaml_2.2.1      stringi_1.4.6   rmarkdown_2.3   highr_0.8      
#>  [9] knitr_1.29      stringr_1.4.0   xfun_0.16       digest_0.6.25  
#> [13] rlang_0.4.7     evaluate_0.14
```

<sup>Created on 2020-12-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Nj221102:
closes #4846 

### Description

This pull request adds a new FAQ entry in `data.table.faq.Rmd` to clarify the behavior difference between `T/F` and `TRUE/FALSE` when used in the `j` argument of `data.table`. The FAQ entry explains why it's recommended to use `TRUE` and `FALSE` over `T` and `F` to avoid unexpected behavior.


Anirban166:
Does `T`/`F` have any utility besides being an abbreviated alias of `TRUE`/`FALSE`? (I'm wondering if they were introduced for some reason aside from the shorthand benefit)

Nj221102:
> Does `T`/`F` have any utility besides being an abbreviated alias of `TRUE`/`FALSE`? (I'm wondering if they were introduced for some reason aside from the shorthand benefit)

don't know much about them other than them being abbreviation but here some info i found online 

Historical Context: In early versions of S (the precursor to R), T and F were introduced as constants representing TRUE and FALSE, respectively. This convention has been carried over into R for consistency and familiarity with users who have experience with S.

Compatibility: Some packages or legacy code might still use T and F instead of TRUE and FALSE. Therefore, maintaining compatibility with these conventions ensures that older code continues to work seamlessly in newer R environments.

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,10 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
 
 However, features like basic S3 dispatch (pasting the function name together with the class name) is why some R folk love R. It's so simple. No complicated registration or signature is required. There isn't much needed to learn. To create the `merge` method for data.table all that was required, literally, was to merely create a function called `merge.data.table`.
 
+## Why do `T` and `F` behave differently from `TRUE` and `FALSE` in `data.table`?
+
+In R, `T` and `F` are global variables that default to `TRUE` and `FALSE`, respectively. However, they can be redefined in the user environment, leading to unexpected behavior. The `data.table` package might treat `T` and `F` as variable names rather than logical constants. To avoid this issue, always use `TRUE` and `FALSE` for logical indexing and column selection in `data.table`.
```
tdhock, line 465: logical indexing -> logical subsetting?
also the user could re-define TRUE and FALSE, so I think saying that T and F default to TRUE and FALSE, and can be redefined is misleading. better to say anything can be redefined in the user environment? but this is not specific to data table so I'm not sure that any FAQ is even needed
@jangorecki can you please review since you commented on the original issue?

Anirban166:
> Historical Context: In early versions of S (the precursor to R), T and F were introduced as constants representing TRUE and FALSE, respectively. This convention has been carried over into R for consistency and familiarity with users who have experience with S.
> 
> Compatibility: Some packages or legacy code might still use T and F instead of TRUE and FALSE. Therefore, maintaining compatibility with these conventions ensures that older code continues to work seamlessly in newer R environments.

It looks like it is something not just for `data.table` then (since it applies to R or other packages in general), in which case it probably doesn't warrant an FAQ entry.

Nj221102:
> > Historical Context: In early versions of S (the precursor to R), T and F were introduced as constants representing TRUE and FALSE, respectively. This convention has been carried over into R for consistency and familiarity with users who have experience with S.
> > Compatibility: Some packages or legacy code might still use T and F instead of TRUE and FALSE. Therefore, maintaining compatibility with these conventions ensures that older code continues to work seamlessly in newer R environments.
> 
> It looks like it is something not just for `data.table` then (since it applies to R or other packages in general), in which case it probably doesn't warrant an FAQ entry.

In discussion of issue #4846 decision was to add the faq, Hi @MichaelChirico WDYT ?

if this is not needed anymore, i can close this PR.

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,10 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
 
 However, features like basic S3 dispatch (pasting the function name together with the class name) is why some R folk love R. It's so simple. No complicated registration or signature is required. There isn't much needed to learn. To create the `merge` method for data.table all that was required, literally, was to merely create a function called `merge.data.table`.
 
+## Why do `T` and `F` behave differently from `TRUE` and `FALSE` in `data.table`?
```
MichaelChirico, line 463:
```suggestion
## Why do `T` and `F` behave differently from `TRUE` and `FALSE` in some `data.table` queries?
```

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,10 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
...
+
+In R, `T` and `F` are global variables that default to `TRUE` and `FALSE`, respectively. However, they can be redefined in the user environment, leading to unexpected behavior. The `data.table` package might treat `T` and `F` as variable names rather than logical constants. To avoid this issue, always use `TRUE` and `FALSE` for logical subsetting and column selection in `data.table`.
```
MichaelChirico, line 465: I think it's too many words. I would simply reproduce the example from the cited issue, then quickly explain what's happening and give the recommendation to avoid `T`/`F` shorthand in general.

We might even give a shoutout to [`T_and_F_symbol_linter()`](https://lintr.r-lib.org/reference/T_and_F_symbol_linter.html) 🙃

MichaelChirico:
> It looks like it is something not just for data.table then (since it applies to R or other packages in general), in which case it probably doesn't warrant an FAQ entry.

@Anirban166 no, not quite -- read the linked issue. Consider `mtcars[,1:3][, c(T,F,T)]`, which will "work as expected" for data.frames. The issue is data.table query analysis does not determine that `c(T,F,T)` is a logical query, IINM because `typeof(substitute(T))` and `typeof(substitute(TRUE))` are different and we'd have to evaluate `T` to see it's `TRUE`.

Moreover data.table scoping makes things complicated -- consider `data.table(T = 1, F = 2)`, what should `DT[, c(T, F, T)]` return? We'd rather not special-case these ill-advised symbols `T`/`F`.

Anirban166:
> @Anirban166 no, not quite -- read the linked issue. Consider `mtcars[,1:3][, c(T,F,T)]`, which will "work as expected" for data.frames. The issue is data.table query analysis does not determine that `c(T,F,T)` is a logical query, IINM because `typeof(substitute(T))` and `typeof(substitute(TRUE))` are different and we'd have to evaluate `T` to see it's `TRUE`.

> Moreover data.table scoping makes things complicated -- consider `DT=data.table(T = 1, F = 2)`, what should `DT[, c(T, F)]` return? We'd rather not special-case these ill-advised symbols `T`/`F`.

I think it's reasonable that `data.table` doesn't automatically recognize `T` and `F` as representing `TRUE` and `FALSE`, both because of how scoping works and in general. Yes, they would be columns in `DT` and not logical values, but that's what one should generally expect imo. I don't think that using those single letter variable names just to represent a shorthand of boolean values is a good practice or something that others should expect from a package to follow up on if that makes sense.

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,70 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
...
+#> 3: a 6
+#> 4: b 1
+#> 5: b 3
+#> 6: b 6
+#> 7: c 1
+#> 8: c 3
+#> 9: c 6
+
+# Using T/F may lead to unexpected behavior
+# `data.table` treats T and F as variable names, not logical constants
```
Anirban166, line 490: In comments `data.table` won't be rendered as code block so you can stick with data.table like you did in line 472 above

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,70 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
...
+DT[, .SD, .SDcols = c(T, T, F)]
+#> Detected that j uses these columns: <none>
+#> [1]  TRUE  TRUE FALSE
+
+# Using with = FALSE works as expected
+# This forces `data.table` to treat T/F as logical constants
```
Anirban166, line 496: Same here

Anirban166 (approved):
I'm all good with adding this FAQ entry for that purpose, and I think it would be helpful for those who stick to `T`/`F`. I just personally don't see much point in their existence in R, as sticking with `TRUE` or `FALSE` when needed sounds reasonable and good enough (in terms of typing convenience) to me. (In other words, people shouldn't generally use them even without us telling them I feel)

MichaelChirico:
> I'm all good with adding this FAQ entry, and I think it would be helpful for those who stick to `T`/`F`. I just personally don't see much point in their existence in R, as sticking with `TRUE` or `FALSE` when needed sounds reasonable and good enough (in terms of typing convenience) to me. (in other words, people shouldn't generally use them without us telling them I feel)

Agreed, but it's definitely a gotcha for beginners. Heaven knows why but tons of newbies (including myself!) start out writing R with `T` and `F` everywhere, and it's all over CRAN packages too-- we're stuck with it :(

Anirban166:
> Heaven knows why but tons of newbies (including myself!) start out writing R with T and F everywhere, and it's all over CRAN packages too-- we're stuck with it :(

Dang > < 
'Might as well hop aboard the train'

vignettes/datatable-faq.Rmd
Nj221102: done 👍

vignettes/datatable-faq.Rmd
```diff
@@ -462,49 +462,68 @@ However, features like basic S3 dispatch (pasting the function name together wit
...
-# In this case, it returns the values of columns T and F, which are 1 and 2.
-
-# To avoid these issues, always use TRUE and FALSE for logical subsetting and column selection in data.table.
+#>    x y
+#> 1: a 1
+#> 2: a 3
+#> 3: a 6
+#> 4: b 1
+#> 5: b 3
+#> 6: b 6
```
tdhock, line 504: please delete

vignettes/datatable-faq.Rmd
```diff
@@ -462,49 +462,68 @@ However, features like basic S3 dispatch (pasting the function name together wit
...
+#> 8: c 3
+#> 9: c 6
 ```
 
+ **Explanation:**
+1. **Using `TRUE`/`FALSE` works as expected:**
+   When you use `TRUE` and `FALSE` in `DT[, .SD, .SDcols = c(TRUE, TRUE, FALSE)]`, `data.table` correctly identifies them as logical constants. This selects the first two columns (`x` and `y`) and excludes the third column (`v`).
+
+2. **Using `T`/`F` leads to unexpected behavior:**
+   When you use `T` and `F` in `DT[, .SD, .SDcols = c(T, T, F)]`, `data.table` treats `T` and `F` as variable names rather than logical constants. Since these variables are not defined within the `data.table` scope, it results in incorrect behavior.
```
tdhock, line 515: this does not make sense to me, because the variable names are x, y, v, not T, F

vignettes/datatable-faq.Rmd
jangorecki: looks ok to me

vignettes/datatable-faq.Rmd
MichaelChirico: I just removed this section, I found it a bit redundant

MichaelChirico:
I took a pass at condensing this a bit, in case anyone wants to take a read with a 2nd pair of eyes before merging.

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,44 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
 
 However, features like basic S3 dispatch (pasting the function name together with the class name) is why some R folk love R. It's so simple. No complicated registration or signature is required. There isn't much needed to learn. To create the `merge` method for data.table all that was required, literally, was to merely create a function called `merge.data.table`.
 
+## Why do `T` and `F` behave differently from `TRUE` and `FALSE` in some `data.table` queries?
+
+Using `T` and `F` as abbreviations for `TRUE` and `FALSE` in `data.table` can lead to unexpected behavior. This is because `T` and `F` are global variables that can be redefined, which causes them to be treated as variable names rather than logical constants. This issue does not occur with `TRUE` and `FALSE`. Avoiding `T` and `F` is advice for using R generally, but it shows up in data.table in some perhaps surprising ways, for example:
```
Anirban166, line 465: > Using `T` and `F` as abbreviations for `TRUE` and `FALSE` in `data.table` can lead to unexpected behavior. This is because `T` and `F` are global variables that can be redefined, which causes them to be treated as variable names rather than logical constants. This issue does not occur with `TRUE` and `FALSE`.

Using `T` and `F` as abbreviations for `TRUE` and `FALSE` in `data.table` can lead to unexpected behavior as they are global variables that can be redefined, and are thus treated as variable names rather than reserved logical constants.

(Doesn't have to be what I rephrased above, but I would suggest something slightly less repetitive, and I don't think we need to explicitly mention the somewhat obvious resolution that the issue does not occur when using `TRUE`/`FALSE` instead)

> Avoiding `T` and `F` is advice for using R generally, but it shows up in data.table in some perhaps surprising ways, for example

`data.table`

vignettes/datatable-faq.Rmd
```diff
@@ -460,6 +460,44 @@ Happily, an internet search for "How does R method dispatch work" (at the time o
...
+#> Detected that j uses these columns: <none>
+#> [1]  TRUE  TRUE FALSE
+
+# Using with = FALSE works as expected
+# This forces data.table to treat T/F as logical constants
+DT[, .SD, .SDcols=c(T, T, F), with=FALSE]
+# same output as DT[, .SD, .SDcols=c(TRUE, TRUE, FALSE)]
+```
+
+As a general word of advice, `lintr::T_and_F_symbol_linter()` detects the usage of `T` and `F` and suggests replacing them with `TRUE` and `FALSE` to avoid such issues.
```
Anirban166, line 499: Nice plug for the linter :) 
But in seriousness, I like how condensed this is. Conveys the message clearly.

MichaelChirico (approved):

MichaelChirico:
Thanks!

PR #6204, opened 2024-06-25
set() adds new cols when rows aren't updated

### LINKED ISSUE #5409: set() that does not update any rows should still add cols
I would expect `set(x, i, j, val)` to work similarly to `x[i, j := val]` or `x[i, j] <- val` for integer `i`. However, there is a discrepancy when `i` is `0L`: While `<-` and `:=` add missing columns, `set()` does not.

```r
x <- data.table(a = 1)
x[0L, "b"] <- character(0)
x
#>        a      b
#>    <num> <char>
#> 1:     1   <NA>

x <- data.table(a = 1)
x[0L, b := character(0)]
x
#>        a      b
#>    <num> <char>
#> 1:     1   <NA>
```
vs.
```r
x <- data.table(a = 1)
set(x, 0L, "b", character(0))
x
#>        a
#>    <num>
#> 1:     1
```

## sessionInfo()
```
R version 4.1.3 (2022-03-10)
Running under: Fedora Linux 35 (Thirty Five)

Matrix products: default
BLAS/LAPACK: /usr/lib64/libflexiblas.so.3.2

locale:
 [1] LC_CTYPE=en_US.utf8       LC_NUMERIC=C             
 [3] LC_TIME=en_US.utf8        LC_COLLATE=en_US.utf8    
 [5] LC_MONETARY=en_US.utf8    LC_MESSAGES=en_US.utf8   
 [7] LC_PAPER=en_US.utf8       LC_NAME=C                
 [9] LC_ADDRESS=C              LC_TELEPHONE=C           
[11] LC_MEASUREMENT=en_US.utf8 LC_IDENTIFICATION=C      

attached base packages:

other attached packages:
[1] data.table_1.14.3

[1] compiler_4.1.3 tools_4.1.3   
```

joshhwuu:
Closes #5409 

Brings `set()` behavior closer to `:=`, so that `set()` doesn't return early even if no rows are updated.

[Question here: ](https://github.com/Rdatatable/data.table/issues/5409#issuecomment-2185353835) I noticed that `isString()` is used inside `SEXP assign` to differentiate between calls from `set()` and from `:=`, therefore this change worked. Although I just want to confirm that this is intended, as it isn't immediately obvious to me why `cols` from `set()` is always passed as a string while not from `:=`.


inst/tests/tests.Rraw
```diff
@@ -18698,3 +18698,10 @@ ans = c(
  "10:   83   64   41                 9                 9")
 # test where topn isn't necessary
 test(2264.8, print(DT, show.indices=TRUE), output=ans)
+
+# test that set() correctly adds new columns even if 'i' isn't valid
+dt = data.table(a=1L)
+test(2265.1, set(copy(dt), 0L,          "b", logical(0)), data.table(a=1L, b=NA))
+test(2265.2, set(copy(dt), NA_integer_, "b", NA),         data.table(a=1L, b=NA))
+test(2265.3, set(copy(dt), NA_integer_, "b", logical(0)), copy(dt)[0L, b := logical(0)])
+test(2265.4, set(copy(dt), 0L,          "b", NA),         copy(dt)[NA, b := NA])
```
Anirban166, line 18707: Can add one to test for updating an existing column (`a` in your case) after adding a new one (`b` as you went with) on an invalid row index (`0L`), something like:
```r
test(2265.5, { set(dt, 0L, "b", logical(0)); set(dt, 1L, "a", 2L); dt }, data.table(a = 2L, b = NA))
```

inst/tests/tests.Rraw
```diff
@@ -18698,3 +18698,11 @@ ans = c(
  "10:   83   64   41                 9                 9")
 # test where topn isn't necessary
 test(2264.8, print(DT, show.indices=TRUE), output=ans)
+
+# test that set() correctly adds new columns even if 'i' isn't valid
+dt = data.table(a=1L)
+test(2265.1, set(copy(dt), 0L,          "b", logical(0)), data.table(a=1L, b=NA))
+test(2265.2, set(copy(dt), NA_integer_, "b", NA),         data.table(a=1L, b=NA))
+test(2265.3, set(copy(dt), NA_integer_, "b", logical(0)), copy(dt)[0L, b := logical(0)])
```
MichaelChirico, line 18706: Shouldn't the `i=` values be reversed in tests 3/4? i.e. `[NA` in this test and `[0L` in the next? That looks more like "testing parity of `:=` and `set()`" to me.

src/assign.c
```diff
@@ -381,12 +381,12 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
...
       if ((rowsd[i]<0 && rowsd[i]!=NA_INTEGER) || rowsd[i]>nrow)
         error(_("i[%d] is %d which is out of range [1,nrow=%d]"), i+1, rowsd[i], nrow);  // set() reaches here (test 2005.2); := reaches the same error in subset.c first
-      if (rowsd[i]>=1) numToDo++;
+      if (rowsd[i]>=0) numToDo++;
     }
     if (verbose) Rprintf(_("Assigning to %d row subset of %d rows\n"), numToDo, nrow);
     // TODO: include in message if any rows are assigned several times (e.g. by=.EACHI with dups in i)
     if (numToDo==0) {
-      if (!length(newcolnames)) {
+      if (!length(newcolnames) && !isString(cols)) {
```
MichaelChirico, line 389: please add a comment explaining the new code here

inst/tests/tests.Rraw
joshhwuu: Something like `copy(dt)[0L, b := logical(0)]` -> `copy(dt)[NA, b := logical(0)]` ? Right that makes more sense.

MichaelChirico: right

inst/tests/tests.Rraw
```diff
@@ -18698,3 +18698,11 @@ ans = c(
  "10:   83   64   41                 9                 9")
 # test where topn isn't necessary
 test(2264.8, print(DT, show.indices=TRUE), output=ans)
+
+# test that set() correctly adds new columns even if 'i' isn't valid
+dt = data.table(a=1L)
+test(2265.1, set(copy(dt), 0L,          "b", logical(0)), data.table(a=1L, b=NA))
+test(2265.2, set(copy(dt), NA_integer_, "b", NA),         data.table(a=1L, b=NA))
+test(2265.3, set(copy(dt), 0L,          "b", NA),         copy(dt)[0L, b := NA])
+test(2265.4, set(copy(dt), NA_integer_, "b", logical(0)), copy(dt)[NA, b := logical(0)])
```
MichaelChirico, line 18707:
```suggestion
test(2265.4, set(copy(dt), NA_integer_, "b", logical(0)), copy(dt)[NA_integer_, b := logical(0)])
```

src/assign.c
```diff
@@ -381,12 +381,13 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
...
       if ((rowsd[i]<0 && rowsd[i]!=NA_INTEGER) || rowsd[i]>nrow)
         error(_("i[%d] is %d which is out of range [1,nrow=%d]"), i+1, rowsd[i], nrow);  // set() reaches here (test 2005.2); := reaches the same error in subset.c first
-      if (rowsd[i]>=1) numToDo++;
+      if (rowsd[i]>=0) numToDo++;
     }
     if (verbose) Rprintf(_("Assigning to %d row subset of %d rows\n"), numToDo, nrow);
     // TODO: include in message if any rows are assigned several times (e.g. by=.EACHI with dups in i)
     if (numToDo==0) {
-      if (!length(newcolnames)) {
+      // isString(cols) is exclusive to calls from set(), not :=, this part should only bind for set()
```
MichaelChirico, line 389: not sure I'm understanding:

 - `isString()` is for `set()`
 - this branch doesn't apply when `!isString()`
 - The comment says "this part should only bind for `set()`"

Maybe I'm just not understanding what we mean by "this part should bind". We could possible just exclude everything after `:=`, that may be good enough.

MichaelChirico (approved):

joshhwuu:
oops tried to resolve merge conflict on tests.rraw and screwed up the diff.. will try to fix now

tdhock (changes requested):
great thanks!
Is this behavior documented? If not please update docs.
Please add a NEWS item.

tdhock:
related docs I found are ?set
```
set is a low-overhead loop-able version of ‘:=’
...
       i: Optional. Indicates the rows on which the values must be
          updated with. If not provided, implies _all rows_. The ‘:=’
          form is more powerful as it allows _subsets_ and ‘joins’
          based add/update columns by reference. See ‘Details’.

          In set, only integer type is allowed in ‘i’ indicating
          which rows ‘value’ should be assigned to. ‘NULL’ represents
          all rows more efficiently than creating a vector such as
          ‘1:nrow(x)’.
```
I also find "docs" about what i is allowed to be in the error message below
```r
> DT=data.table(x=1)
> set(DT, i=-1L, j="b", value=5)
Error in set(DT, i = -1L, j = "b", value = 5) : 
  i[1] is -1 which is out of range [1,nrow=1]
> set(DT, i=0L, j="b", value=5)
> set(DT, i=NA_integer_, j="b", value=5)
> DT
       x
   <num>
1:     1
```
The error message above implies that the only valid values for i are [1,nrow] (but missing/NA and 0 in i are not mentioned on ?set).

It seems to me that allowing 0L and NA_integer_ are inconsistent with this error message.

Also below we see that `:=` allows NA_initeger_ and -1L, creating a new column with missing values, as when i=0L:
```r
> DT[-1L, b := 5]
> DT[NA_integer_, c := 5]
> DT[2L, d := 5]
Error in `[.data.table`(DT, 2L, `:=`(c, 5)) : 
  i[1] is 2 which is out of range [1,nrow=1]
> DT
       x     b     c
   <num> <num> <num>
1:     1    NA    NA
```
So to summarise, we have in current master

|   i |      `:=`      |  set |
|----------|-------------|------|
| -1L | add new col  | error |
| 0L|   add new col    |  ignore  |
| NA_integer_ |add new col  |  ignore   |
| >nrow | error | error |

For consistency and backward compatibility, we should probably change this to

|   i |      `:=`      |  set |
|----------|-------------|------|
| -1L | add new col  | add new col |
| 0L|   add new col    |  add new col  |
| NA_integer_ |add new col  |  add new col   |
| >nrow | error | error |

and change the error message to `i[1] is 2 but should not be larger than nrow=1`

or just error everywhere (but that would be a breaking change, maybe there will be revdep issues)

tdhock:
The only test case I see which specifies the behavior for weird `i` is 
```r
test(2005.02, set(DT, 4L, "b", NA), error="i[1] is 4 which is out of range [1,nrow=3]")
```
So i=-1, 0 and NA_integer_ are unspecified behavior in tests as well as docs.

tdhock:
wdyt we should do about this @joshhwuu @MichaelChirico @mb706 @ben-schwen ?

mb706:
I'd vote for your proposal making `set` consistent with `:=`. The current behaviour is what I stumbled over when replacing an instance of `:=` with `set()`, expecting them to behave the same as long as `i` is integer-valued.

Maybe another case to consider would be
```r
set(DT, i=integer(0), j="d", value=numeric(0))
```

tdhock:
you wrote
```r
set(DT, i=integer(0), j="d", value=numeric(0))
```
but it is simpler/preferable to write
```r
set(DT, i=0L, j="d", value=0)
```
literal numbers with L are integer, and without are numeric.

joshhwuu:
I also prefer that we bring `set` behavior closer to `:=`

ben-schwen:
@tdhock I think Martin actually meant zero length `integer` and `numeric` vectors as he wrote it.

NEWS.md
```diff
@@ -104,6 +104,8 @@
 
     We plan to export similar placeholders for `.` and `J` in roughly one year (e.g. data.table 1.18.0), but excluded them from this release to avoid back-compatibility issues. Specifically, some packages doing `import(plyr)` _and_ `import(data.table)`, and/or with those packages in `Depends`, will error when data.table starts exporting `.` (and similarly for a potential conflict with `rJava::J()`). We discourage using data.table (or any package, really) in Depends; blanket `import()` of package is also generally best avoided. See `vignette("datatable-importing")`.
 
+18. `set()` now adds new rows even if no rows are updated, [#5409](https://github.com/Rdatatable/data.table/issues/5409). This behavior is now consistent with `:=`, thanks to @mb706 for the report and @joshhwuu for the fix.
```
ben-schwen, line 107:
```suggestion
18. `set()` now adds new columns even if no rows are updated, [#5409](https://github.com/Rdatatable/data.table/issues/5409). This behavior is now consistent with `:=`, thanks to @mb706 for the report and @joshhwuu for the fix.
```

Shouldn't it be about the addition of columns?

ben-schwen:
Apart from news, LGTM

NEWS.md
joshhwuu: Ah right, thanks!

joshhwuu:
@tdhock WDYT about changing the error to `i[1] is 2 but should not be larger than nrow=1` as above, and mentioning in docs that NULL, 0L and -1L given in `i` are not used and ignored

tdhock:
> mentioning in docs that NA, 0L and -1L given in i are not used and ignored

yes please add docs: missing values and non-positive values are ignored, positive values greater than nrow cause error.

joshhwuu:
@tdhock upon further investigation, it seems that negative indexing values allowed in `DT[]` is for non-selection of rows, for example:
```R
DT = data.table(x = 1:5)
DT[-1L, b := 1]
DT
       x      b
   <int> <char>
1:     1   <NA>
2:     2      1
3:     3      1
4:     4      1
5:     5      1
```
Which I don't believe `set()` supports as is, and it would be better to use `:=` in this situation. See `?set`:

```
i: Optional. Indicates the rows on which the values must be
          updated with. If not provided, implies _all rows_. 
          The ‘:=’ form is more powerful as it allows _subsets_
          and ‘joins’ based add/update columns by reference. See
          ‘Details’.
```
I believe the correct way to go about this is to just mention missing values and 0L being ignored in the docs. LMK if you think otherwise.

tdhock:
yes that is moving in the right direction.
the more we dig the weirder it gets haha.
```r
> DT=data.table(x=1:2)
> set(DT,-1,"x",0)
Error in set(DT, -1, "x", 0) : 
  i[1] is -1 which is out of range [1,nrow=2]
In addition: Warning message:
In set(DT, -1, "x", 0) :
  Coerced i from numeric to integer. Please pass integer for efficiency; e.g., 2L rather than 2
> DT[NA, x := 0]
> DT
       x
   <int>
1:     1
2:     2
> DT[c(-1,1), x := 0]
Error in `[.data.table`(DT, c(-1, 1), `:=`(x, 0)) : 
  Item 1 of i is -1 and item 2 is 1. Cannot mix positives and negatives.
> DT[c(-1,NA), x := 0]
Error in `[.data.table`(DT, c(-1, NA), `:=`(x, 0)) : 
  Item 1 of i is -1 and item 2 is NA. Cannot mix negatives and NA.
> DT[c(NA,1), x := 0]
> DT
       x
   <int>
1:     0
2:     2
```

joshhwuu:
It would be difficult/time-consuming finding each of these cases and writing good concise documentation for them, so I agree that we can leave the documentation as-is for these niche cases

inst/tests/tests.Rraw
```diff
@@ -19002,3 +19002,12 @@ test(2277.2, DT[, closure(b), env=list(closure=var)], 0.5)
 test(2277.3, DT[, closure(b), env=list(closure=stats::var)], 0.5)
 test(2277.4, DT[, closure(b), env=list(closure=stats:::var)], 0.5)
 test(2277.5, DT[, lambda(b), env=list(lambda=function(x) sum(x))], 7L)
+
+# test that set() correctly adds new columns even if 'i' isn't valid
```
MichaelChirico, line 19006:
```suggestion
# test that set() correctly adds new columns even if now rows are updated
```

man/assign.Rd
```diff
@@ -35,7 +35,7 @@ set(x, i = NULL, j, value)
 \item{LHS}{ A character vector of column names (or numeric positions) or a variable that evaluates as such. If the column doesn't exist, it is added, \emph{by reference}. }
 \item{RHS}{ A list of replacement values. It is recycled in the usual way to fill the number of rows satisfying \code{i}, if any.  To remove a column use \code{NULL}. }
 \item{x}{ A \code{data.table}. Or, \code{set()} accepts \code{data.frame}, too. }
-\item{i}{ Optional. Indicates the rows on which the values must be updated with. If not provided, implies \emph{all rows}. The \code{:=} form is more powerful as it allows \emph{subsets} and \code{joins} based add/update columns by reference. See \code{Details}.
+\item{i}{ Optional. Indicates the rows on which the values must be updated with. If not provided, implies \emph{all rows}. Missing or zero values are ignored. The \code{:=} form is more powerful as it allows \emph{subsets} and \code{joins} based add/update columns by reference. See \code{Details}.
```
MichaelChirico, line 38:
```suggestion
\item{i}{ Optional. Indicates the rows on which the values must be updated. If not \code{NULL}, implies \emph{all rows}. Missing or zero values are ignored. The \code{:=} form is more powerful as it allows adding/updating columns by reference based on \emph{subsets} and \code{joins}. See \code{Details}.
```

MichaelChirico (approved):
Thanks!

PR #6210, opened 2024-06-28
fread cannot read columns with all NA values and colClasses="POSIXct"

### LINKED ISSUE #6208: fread column with only NA values as POSIXct returns warning and a character column
I would like to read a column of a `NA` values as `POSIXct` using `fread`. 

I would expect this to work: 
```
library(data.table)
fwrite(data.table(datetime = as.POSIXct(NA)), "test.csv")
fread("test.csv", colClasses = "POSIXct")
```
However, this gives the following warning: 

```
Warning message:
Column 'datetime' was requested to be 'POSIXct' but fread encountered the following error:
	character string is not in a standard unambiguous format
so the column has been left as type 'character' 
```

In contrast using `utils::write.csv` and `utils::read.csv` this works: 

```
write.csv(data.frame(datetime = as.POSIXct(NA)), "test.csv", row.names = FALSE)
read.csv("test.csv", colClasses = "POSIXct")
```

### Output of sessionInfo()
```
R version 4.4.1 (2024-06-14 ucrt)
Platform: x86_64-w64-mingw32/x64

Matrix products: default


locale:
[1] LC_COLLATE=Danish_Denmark.utf8  LC_CTYPE=Danish_Denmark.utf8    LC_MONETARY=Danish_Denmark.utf8 LC_NUMERIC=C                    LC_TIME=Danish_Denmark.utf8    

time zone: Europe/Copenhagen
tzcode source: internal

attached base packages:

other attached packages:
[1] data.table_1.15.4

[1] compiler_4.4.1    tools_4.4.1       rstudioapi_0.16.0
```

I have also posted this at [stackoverflow](https://stackoverflow.com/q/78676072/16697029)

ben-schwen:
Closes #6208


R/fread.R
```diff
@@ -302,6 +302,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
   for (j in which(nzchar(colClassesAs))) {       # # 1634
     v = .subset2(ans, j)
     new_class = colClassesAs[j]
+    if (new_class %chin% c("POSIXct")) v[!nzchar(v)] = NA_character_ # as.POSIXct/as.POSIXlt cannot handle as.POSIXct("") correctly #6208
```
MichaelChirico, line 305: are there any other classes where this might apply? is it possible for `new_class` to be missing? wondering why this isn't just `new_class == "POSIXct"`...

If it's really limited to POSIXct, maybe it's cleaner instead to use `as.POSIXct(replace(v, !nzchar(v), NA_character_)` below? Though that will induce a copy, right? WDYT about `"POSIXct" = {v[!nzchar(v)] = NA_character_; as.POSIXct(v)}`?

ben-schwen: It also does not work for `as.POSIXlt` since `as.POSIXct.character` calls `as.POSIXlt`

MichaelChirico: OK got it, hence the as.POSIXlt in the comment too. I think that's good enough.

MichaelChirico (approved):

PR #6211, opened 2024-06-30
change tags for linux

### LINKED ISSUE #6209: Gitlab CI wrong tags for runner
Our nightly builds get stuck daily in their first stage.

AFAIU we use tags in our `.gitlab-ci.yml` file and our runner does not have them listed.

@MichaelChirico If you could add [me](https://gitlab.com/ben-schwen) to the Gitlab repo I would try to fix it.

ben-schwen:
Closes #6209 

Apparently, our used tag `linux` is not available anymore at the [hosted runners](https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html)

This PR changes the tag from `linux` to `saas-linux-medium-amd64`. Another option would be to remove the tags which then uses the default of `saas-linux-small-amd64`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -46,7 +46,7 @@ stages:
 mirror-packages:
   stage: dependencies
   tags:
-    - linux
+    - saas-linux-medium-amd64
   image: registry.gitlab.com/jangorecki/dockerfiles/r-base-minimal
   cache:
     paths:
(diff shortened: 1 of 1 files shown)


ben-schwen:
@jangorecki Since you have experience with Gitlab CI, could you PTAL

ben-schwen:
I just restarted the pipeline and its running now! Lets say if other issue pop up

PR #6212, opened 2024-07-01
change tags for windows

ben-schwen:
Follow-up to #6211

Makes the same change of tags for [windows](https://docs.gitlab.com/ee/ci/runners/hosted_runners/windows.html)


jangorecki (approved):

MichaelChirico:
thanks for linking the docs! very helpful. if you happen to have seen a doc mentioning the deprecation of the old approach it'd be helpful to link it (don't spend more than a few minutes looking though)

PR #6121, opened 2024-05-03
Run linters on test scripts

MichaelChirico:
The default pattern for files to lint is `pattern = "(?i)[.](r|rmd|qmd|rnw|rhtml|rrst|rtex|rtxt)$"`, i.e. not `Rraw`. We don't need the other extensions, so trim them also for simplicity.

See:

https://lintr.r-lib.org/reference/lint.html

PR #6214, opened 2024-07-01
add bit64 package check

### LINKED ISSUE #6213: Ensure bit64 is loaded before use is.integer64
Enabling linters in #6121 we changed `if (class(col) == "integer64") {` to `if (is.integer64(col)) {` but do not check if `bit64` is loaded/available.

Error is detected in our [Gitlab CI](https://gitlab.com/Rdatatable/data.table/-/jobs/7231169413)

ben-schwen:
Closes #6213

Follow-up to #6121


MichaelChirico (approved):
nice find! linter also wound up improving the code here IMO.

Issue #6192, opened 2024-06-19
Odd/unexpected behaviour with `is.na`


iago-pssjd:
I maybe should expect this, but I didn't:
```r
dt = data.table(x = c(1, NA))
class(dt[, is.na(.SD)])
[1] "matrix" "array"
```
Shouldn't be a data.table?

Thanks!

iago-pssjd:
Well, actually, if it is equivalent to `is.na(dt)` (and there is no a `data.table` method for it), then I must expect that behaviour. There is no other example.

Thanks!

ben-schwen:
There is a `is.na.data.table` so the request would be for `is.na.data.table` to return a `data.table`?

MichaelChirico:
we do the same as data.frame here: is.na(mtcars)

iago-pssjd:
> There is a is.na.data.table so the request would be for is.na.data.table to return a data.table?
@ben-schwen  Well, it was more an expectation (actually, mainly for `dt[, is.na(.SD)]` and distinct to `is.na(dt)`), than a request, but the idea was that.

> we do the same as data.frame here: is.na(mtcars)

@MichaelChirico It's fine to me.

iago-pssjd:
@ben-schwen @MichaelChirico Although being the default same behaviour than for data.frame, it would be great having an option to return a data.table...

ben-schwen:
Could you provide an example where `is.na(x, as.data.table=TRUE)` is advantageous over `as.data.frame(is.na(x))`. Personally, I use `is.na` mainly for indexing so I like that it returns a matrix

iago-pssjd:
Indeed, my solution was to do `dt[, data.table(is.na(.SD))]`, but this implies a double conversion (copy) data.table -> matrix -> data.table. Not an example, but I will speculate  a bit. Wouldn't there be a more direct and efficient way (let's say, something "by reference") so that `is.na(x, as.data.table=TRUE)` there will not make a matrix to get a data.table?

MichaelChirico:
Note that `is.na()` primitive does not accept new arguments (there's no `...`), so I'm not sure it's possible (and probably not recommended) to extend it with such an argument.

Note that the implementation is pretty trivial:

https://github.com/Rdatatable/data.table/blob/6df30c39cc5d1e910e6f4ab84ccdcc693c64315c/R/data.table.R#L2378

So that if you want this, I recommend just doing `setDT(lapply(x, is.na))`.

[...]

PR #6217, opened 2024-07-03
Don't use function string matching for do.call(what=) or lapply(FUN=)

MichaelChirico:
Related: #6192. First caught my eye because I know this makes it harder for static analyzers, e.g. any code looking for `cbind()` usage will have to special-case the possibility that `cbind` is passed as a string, in an appropriate context, etc.

This is not a terribly big deal for base functions but is best practice when using dependencies since it can be hard to tell if downstreams get broken, statically.

It's also a bit slower -- `match.fun()` exits immediately when passed a function, takes an extra step when passed a string. That core part of function matching in `lapply()` is 10x slower. `do.call()` is roughly the same:

```r
microbenchmark(times=1e6, match.fun("is.na"), match.fun(is.na))
# Unit: nanoseconds
#                expr  min   lq     mean median   uq       max neval cld
#  match.fun("is.na") 6892 7183 8837.238   7290 7696 127587928 1e+06   b
#    match.fun(is.na)  575  677 1073.283    730  832  64672336 1e+06  a 
```

```r
l=list(1)
microbenchmark(times=1e6, do.call("identity", l), do.call(identity, l))
# Unit: microseconds
#                    expr   min    lq     mean median    uq      max neval cld
#  do.call("identity", l) 1.885 2.010 2.910809  2.094 2.451 42517.85 1e+06   a
#    do.call(identity, l) 1.850 1.966 2.915019  2.049 2.417 77993.08 1e+06   a
```

PS Shouldn't we use `NextMethod()`, rather than call the S3 method explicitly?

### DIFF
--- a/.dev/cc.R
+++ b/.dev/cc.R
@@ -62,8 +62,8 @@ cc = function(test=FALSE, clean=FALSE, debug=FALSE, omp=!debug, cc_dir, path=Sys
 
   xx = try(getDLLRegisteredRoutines("data_table",TRUE), silent=TRUE)
   if (!inherits(xx, "try-error")) {
-    remove(list=sapply(xx$.Call,'[[',"name"), pos=.GlobalEnv)
-    remove(list=sapply(xx$.External,'[[',"name"), pos=.GlobalEnv)
+    remove(list=sapply(xx$.Call, `[[`, "name"), pos=.GlobalEnv)
+    remove(list=sapply(xx$.External, `[[`, "name"), pos=.GlobalEnv)
     # if these objects aren't there to remove it's correctly an error (should always be there)
   }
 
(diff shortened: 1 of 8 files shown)


tdhock (approved):
looks good, thanks

ben-schwen:
LGTM, but I'm not sure I would apply it to test cases since some of them might explictly check for working characters, as e.g., 23c88e7a7b5b07ac34c26ee0cfb5435f58bfdd91

ben-schwen (approved):

Anirban166 (approved):

MichaelChirico:
> LGTM, but I'm not sure I would apply it to test cases since some of them might explictly check for working with characters, as e.g., [23c88e7](https://github.com/Rdatatable/data.table/commit/23c88e7a7b5b07ac34c26ee0cfb5435f58bfdd91)

Thanks for flagging! Indeed I skipped some test cases that are testing for strings usage specifically. In general you're right that altering test cases can be risky. Those changes I did include are clearly designed to test other things -- if behavior w.r.t. strings is intended, that should get its own specific test. The footprint here is small enough I'm confident we're OK.

PR #6218, opened 2024-07-04
Temporal change of NOTES count

ben-schwen:
As mentioned in #6180 we temporarily change the number of NOTES to 3, since issue with non-API calls is known.


MichaelChirico (approved):

tdhock:
great thanks

PR #6227, opened 2024-07-08
force loading bit64 earlier in print()

### LINKED ISSUE #6224: Incorrect printing of integer64 columns
Here is a minimal reproducible example:

```
library(data.table)

dt <- data.table(id = 1:10, int64 = bit64::as.integer64(1:10))
fst::write_fst(dt, "dt.fst")
```

Start a new session without `bit64` being loaded.

```
> library(data.table)

> dt <- fst::read_fst("dt.fst", as.data.table = TRUE)
fstcore package v0.9.18
(OpenMP was not detected, using single threaded mode)

> dt
       id         int64
    <int>         <i64>
 1:     1 4.940656e-324
 2:     2 9.881313e-324
 3:     3 1.482197e-323
 4:     4 1.976263e-323
 5:     5 2.470328e-323
 6:     6 2.964394e-323
 7:     7 3.458460e-323
 8:     8 3.952525e-323
 9:     9 4.446591e-323
10:    10 4.940656e-323

> dt
       id int64
    <int> <i64>
 1:     1     1
 2:     2     2
 3:     3     3
 4:     4     4
 5:     5     5
 6:     6     6
 7:     7     7
 8:     8     8
 9:     9     9
10:    10    10
```

It looks like `fst::read_fst` does not load `bit64` automatically if there are integer64 columns in the data. I'm not sure if it better handled on data.table side: if the table contains integer64 columns, `bit64` should be loaded in `print.data.table`.

MichaelChirico:
Closes #6224

My first thought was to add a `format_col.integer64` method, but it turns out the fix is much easier, just need to force loading {bit64} one line earlier.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -104,6 +104,7 @@
 
     We plan to export similar placeholders for `.` and `J` in roughly one year (e.g. data.table 1.18.0), but excluded them from this release to avoid back-compatibility issues. Specifically, some packages doing `import(plyr)` _and_ `import(data.table)`, and/or with those packages in `Depends`, will error when data.table starts exporting `.` (and similarly for a potential conflict with `rJava::J()`). We discourage using data.table (or any package, really) in Depends; blanket `import()` of package is also generally best avoided. See `vignette("datatable-importing")`.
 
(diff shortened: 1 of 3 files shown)


tdhock (approved):
great thanks
nice test

PR #6230, opened 2024-07-09
add one level of additional revdeps to mirror packages in CI

ben-schwen:
Currently we already mirror reverse dependencies, but these might also have reverse dependencies etc.

This results in that our CI reports things like `also installing the dependencies ‘R.oo’, ‘R.methodsS3’, ‘evaluate’, ‘highr’, ‘xfun’, ‘commonmark’`

This PR adds one extra level of mirroring also the `c("Imports", "Depends")` of the mirrored reverse dependencies (no recursion yet).

### DIFF
--- a/.ci/ci.R
+++ b/.ci/ci.R
@@ -43,12 +43,27 @@ function (repos, type = getOption("pkgType"), ver)
   res
 }
 
+local.extract_dependency_package_names = function (x) {
+  ## do not filter out R like tools:::.extract_dependency_package_names, used for web/$pkg/index.html
+  if (is.na(x))
+    return(character())
+  x <- unlist(strsplit(x, ",[[:space:]]*"))
+  x <- sub("[[:space:]]*([[:alnum:].]+).*", "\\1", x)
+  x[nzchar(x)]
+}
+
+rev.dependencies = function(pkg, available_pkgs, which=c("Imports", "Depends")) {
+  dependencies = available_pkgs[pkg, which]
+  x = dependencies[!is.na(dependencies)]
+  unlist(lapply(x, local.extract_dependency_package_names), use.names=FALSE)
+}
+
 ## returns dependencies for a package based on its DESCRIPTION file
 dcf.dependencies <-
 function(file = "DESCRIPTION",
          which = NA,
          except.priority = "base",
-         exclude = NULL) {
+         exclude = NULL, revdeps=FALSE) {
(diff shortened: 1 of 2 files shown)


.ci/ci.R
```diff
@@ -43,12 +43,27 @@ function (repos, type = getOption("pkgType"), ver)
...
+  if (is.na(x))
+    return(character())
+  x <- unlist(strsplit(x, ",[[:space:]]*"))
+  x <- sub("[[:space:]]*([[:alnum:].]+).*", "\\1", x)
+  x[nzchar(x)]
+}
+
+rev.dependencies = function(pkg, available_pkgs, which=c("Imports", "Depends")) {
+  dependencies = available_pkgs[pkg, which]
+  x = dependencies[!is.na(dependencies)]
```
tdhock, line 57: here can you please insert a comment explaining why tools::dependsOnPkgs is not sufficient?

tdhock:
does "mirroring" mean that the ci will keep a copy of these packages between runs, rather than downloading them in each run?

ben-schwen:
> does "mirroring" mean that the ci will keep a copy of these packages between runs, rather than downloading them in each run?

Exactly

.ci/ci.R
ben-schwen: Maybe I misunderstood something but `tools:dependsOnPkgs(x)` shows the relation for results depends on x (and only does so for installed packages). However, we want the inverse with x depends/imports result.

MichaelChirico: > only does so for installed packages

This part is easy to overcome: `tools::dependsOnPkgs("data.table", installed=available.packages())`.

Anyway, you're correct that's going in the wrong direction; `tools::package_dependencies()` traverses the graph in the right direction:

```r
tools::package_dependencies("data.table", recursive=TRUE, which="all") # overkill
# second-level instead
tools::package_dependencies("data.table", which="all") |>
  unlist() |> tools::package_dependencies() |> # could use which="all" here too?
  unlist(use.names=FALSE) |> setdiff(tools:::.get_standard_package_names()$base)
```

jangorecki:
What issue this PR is attempting to solve?
There was no job that meant to test revdeps, and I think, due to amount of revdeps it is not feasible to have job like that.

ben-schwen:
@jangorecki we had issues in our revdeps and I thought this might resolve it but as it turned out the problem was different one

PR #6232, opened 2024-07-09
Bump R dependency to R>=3.3.0

### LINKED ISSUE #6231: Bump R version dependency to 3.3
Our package is not particularly usable under R 3.2:

 - {knitr}, our VignetteBuilder, requires R >= 3.3.0
 - `R CMD check` does not get the `--ignore-vignettes` option until R 3.3.0

That means neither can we build the vignettes under R 3.3.0, nor can we tell `R CMD check` to skip checking vignettes.

I suppose package tarballs can still be made available, so we could technically distribute the package. But if we can't even check it using normal means, neither can we guarantee the package is in a working state.

We _could_ bend over backward to keep supporting R 3.2:

 - Install an older version of {knitr} on CI (knitr <= 1.37.3, at least 2 years old: https://github.com/yihui/knitr/commit/1da8e130add969c5222c7a24f807eb07537d5c04)
 - Set up a custom check environment that runs parts of `R CMD check` while skipping others, most importantly running `tests/` and examples

But I don't think it's worth it.

Discovered by the `R-cmd-check-occasional` job run today:

https://github.com/Rdatatable/data.table/actions/runs/9857920757/job/27218061001

MichaelChirico:
Closes #6231

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -196,6 +196,16 @@ base_messages = list(
   NULL
 )
 
+# Ensure an operation uses C-locale sorting (#3502). For test set-ups/comparisons that use base operations, which are
+#   susceptible to locale-specific sorting issues, but shouldn't be needed for data.table code, which always uses C sorting.
+# TODO(R>=3.3.0): use order(method="radix") as a way to avoid needing this helper
+with_c_collate = function(expr) {
+  old = Sys.getlocale("LC_COLLATE")
+  on.exit(Sys.setlocale("LC_COLLATE", old))
+  Sys.setlocale("LC_COLLATE", "C")
+  expr
+}
+
 ##########################
 .do_not_rm = ls()  # objects that exist at this point should not be removed by rm_all(); e.g. test_*, base_messages, Ctest_dt_win_snprintf, prevtest, etc
 ##########################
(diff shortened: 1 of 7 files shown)


joshhwuu:
If we bump to 3.3.0, could we also delete explicit definition of `strrep()` in the test script?

https://github.com/Rdatatable/data.table/blob/ac4bb6bbc40b718db98a8d5ffaa0bbcf91a963f1/inst/tests/tests.Rraw#L210-L213

MichaelChirico:
Please check the PR diff here, unless I'm missing something :)

joshhwuu:
> Please check the PR diff here, unless I'm missing something :)

Ah I must've missed it, apologies! Brain is a bit foggy, just came down with a nasty flu 😭

tdhock (approved):
nice NEWS entry

inst/tests/tests.Rraw
```diff
@@ -1853,10 +1837,10 @@ test(609, chorder(character()), base::order(character()))
 test(610, chorder(""), base::order(""))
 # Extra tests of chorder and chgroup
 x = sample(LETTERS)
-test(610.1, chorder(x), with_c_collate(base::order(x)))
+test(610.1, chorder(x), base::order(x, method="radix"))
 test(610.2, chgroup(x), seq_along(x))
 x = sample(LETTERS,1000,replace=TRUE)
-test(610.3, chorder(x), with_c_collate(base::order(x)))
+test(610.3, chorder(x), base::order(x, method="radix"))
```
jangorecki, line 1843: I would keep the old test. Radix implies C locale, or at least should, but we are not testing exactly the same anymore... Anyway test should be ideally vs generated output and not a function call

MichaelChirico: Thanks... I see your point about the former -- using `method="radix"` is actually a proxy way to go about testing what we want here, explicitly calling out `with_c_collate()` makes the intent clearer & more direct.

I don't agree about the latter point, in some tests we are explicitly trying to match a base equivalent, in the other we don't gain much by having a vector of randomly-ordered integers, function call to base `order()` is clearer.

jangorecki:
https://github.com/yihui/litedown/commit/7c3492185efdfcf935d00cd0749b721cd809a93a

Nice to see we are not the only ones to care about running on old R versions. What's even more interesting that litedown is so new package that it is not even yet on CRAN, so could easily choose some recent R dependency.

rikivillalba:
Hello.
Should  CRAN_Release.cmd be updated to test for R>=3.3.0 ?
https://github.com/Rdatatable/data.table/blob/6cee825fa1b05a8ccbf674a5ec5659559d674457/.dev/CRAN_Release.cmd#L253

PR #6235, opened 2024-07-10
Updating C API Entry Points: isFrame -> isDataFrame

TysonStanley:
This is starting a PR with some of the C API changes as reported in #6180. This is likely a growing list but current NOTEs

```
Found non-API calls to R: ‘LEVELS’, ‘NAMED’, ‘SETLENGTH’,
    ‘SET_GROWABLE_BIT’, ‘SET_S4_OBJECT’, ‘SET_TRUELENGTH’,
    ‘SET_TYPEOF’, ‘STRING_PTR’, ‘TRUELENGTH’, ‘UNSET_S4_OBJECT’
```

- [x] `isFrame` to `isDataFrame`: Easy first one (not on the notes but in WRE 6.21.8 Some backports)
- [ ]  `LEVELS`:
- [ ]  `NAMED`:
- [ ]  `SETLENGTH`:
- [ ]  `SET_GROWABLE_BIT`:
- [ ]  `SET_S4_OBJECT`:
- [ ]  `SET_TRUELENGTH`:
- [ ]  `SET_TYPEOF`:
- [ ]  `STRING_PTR`:
- [ ]  `TRUELENGTH`:
- [ ] `UNSET_S4_OBJECT`:

We can keep adding to this list as others come up in the notes. WRE 6.21 "Moving into C API compliance" says  "This section is a work in progress and will be adjusted as changes are made to the API."

Based on Kurt's useR! 2024 talk, they are aware this is a work in progress so I don't believe this should hold up any releases if we can't get to each of these right away.

### DIFF
--- a/src/data.table.h
+++ b/src/data.table.h
@@ -10,6 +10,9 @@
 #if !defined(R_VERSION) || R_VERSION < R_Version(3, 4, 0)
 #  define SET_GROWABLE_BIT(x)  // #3292
 #endif
+#if !defined(R_VERSION) || R_VERSION < R_Version(4, 5, 0)
+# define isDataFrame(x) isFrame(x)
+#endif
 #include <Rinternals.h>
 #define SEXPPTR_RO(x) ((const SEXP *)DATAPTR_RO(x))  // to avoid overhead of looped STRING_ELT and VECTOR_ELT
 #include <stdint.h>    // for uint64_t rather than unsigned long long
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Might be better to do these one entry type at a time instead of in a big PR.

MichaelChirico (approved):

MichaelChirico:
Note #6183 is also available for review :)

TysonStanley:
Thanks, probably smart to split it up. Got over ambitious! Will review the other

tdhock:
I would like to encourage creation of performance tests in PRs to fix this, please ask me or @Anirban166 or @DorisAmoakohene if you want help creating a performance test. My intuition is that changing some low-level C calls could have a big performance impact.

TysonStanley:
Yeah I think you are right for many of these. Don't think the changes in this PR are going to change anything but can still benchmark to be safe

MichaelChirico:
In this case, `isFrame()` is literally identical to `isDataFrame()`, so I would just go ahead and merge.

https://github.com/r-devel/r-svn/blob/09f3acb92dcc180636a7c4e257491302e9a22ae8/src/include/Rinlinedfuns.h#L905

MichaelChirico (approved):

PR #6239, opened 2024-07-11
Silence output of bit64 unload/reattach

MichaelChirico:
Follow-up to #6227 which made our test logs noisier


MichaelChirico:
No actual logic change here, merging

PR #6245, opened 2024-07-12
library(include.only=) call robust to versions missing that argument

MichaelChirico:
As identified by `R-cmd-check-occasional`:

https://github.com/Rdatatable/data.table/actions/runs/9908622817/job/27374800789

There are a few ways to do this, this seems the best to me that (1) continues the spirit of "minimally attach the libraries" and (2) working across versions while minimizing duplication.

Alternatives ignored: (1) Run `library()` directly, with/without `include.only=`, in branches according to the `"include.only" %in%` condition (2) Just drop `include.only=` usage altogether.


tdhock (approved):
great thanks

PR #6246, opened 2024-07-12
Remove two broken URLs

MichaelChirico:
As seen by `r-cmd-check-occasional`:

https://github.com/Rdatatable/data.table/actions/runs/9908622817/job/27374800789


tdhock (approved):
great thanks

PR #5989, opened 2024-03-08
Update guidance for patch releases

MichaelChirico:
Follows discussion in #5963

PR #6247, opened 2024-07-12
Add patch release NEWS to master

MichaelChirico:


tdhock:
great, why wasn't that there before?

MichaelChirico:
> great, why wasn't that there before?

an oversight of the "new" approach to versioning that's just building patches off the previous even version. so those items only wind up in patch branches. See #5989.

TysonStanley:
Sounds like a step I need to include in the patch release? Can add to the release docs.

MichaelChirico:
@TysonStanley PTAL at #5989! I had a pass at doing just that. appreciate your extra eyes + further refinement.

TysonStanley:
> @TysonStanley PTAL at #5989! I had a pass at doing just that. appreciate your extra eyes + further refinement.

Sorry, missed that notification. Will review now.

PR #6248, opened 2024-07-12
Issues with memory management in fmelt

MichaelChirico:
As identified by `rchk` on CRAN:

https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/data.table.out

```
Function checkVars
  [UP] allocating function uniq_diff may destroy its unprotected argument (measure_int_or_list <arg 1>), which is later used. data.table/src/fmelt.c:251
Function uniq_diff
  [UP] calling allocating function set_diff(V,?) with a fresh pointer (unique_col_numbers <arg 1>) data.table/src/fmelt.c:201
```

Note that the `cols_to_int_or_list()` output is indeed `PROTECT()`'d properly in other nearby branches. This makes sense because `measure_int_or_list` is a new `SEXP` not yet on the stack that must be marked against gc removal.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -62,6 +62,8 @@
 
 10. `dt[,,by=año]` (i.e., using a column name containing a non-ASCII character in `by` as a plain symbol) no longer errors with "object 'año' not found", #4708. Thanks @pfv07 for the report, and Michael Chirico for the fix.
 
+11. Fix some memory management issues in the C routine backing `melt()`, as identified by `rchk`. Thanks Tomas Kalibera and the CRAN team for setting up the `rchk` system, and @MichaelChirico for the fix.
+
 ## NOTES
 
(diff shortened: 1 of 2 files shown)


tdhock (approved):
great. I wrote this code and I guess forgot a PROTECT, thanks for looking at rchk results and providing a fix. add a NEWS item?

tdhock:
does this one fix handle all three rchk issues in fmelt.c?
```

Function checkVars
  [UP] allocating function uniq_diff may destroy its unprotected argument (measure_int_or_list <arg 1>), which is later used. data.table/src/fmelt.c:251
  [UP] calling allocating function uniq_diff with a fresh pointer (measure_int_or_list <arg 1>) data.table/src/fmelt.c:251
  [UP] unprotected variable measure_int_or_list while calling allocating function Rf_allocVector data.table/src/fmelt.c:254
```
I'm not used to reading / understanding these messages, are some redundant with each other?

MichaelChirico:
> does this one fix handle all three rchk issues in fmelt.c?

I'm also not sure, but yes that's my read.

I'll add NEWS in a bit

PR #6250, opened 2024-07-12
Add .libPaths() to the test.data.table() diagnostic

MichaelChirico:
Motivated by `r-cmd-check-occasional` still failing, which I no longer understand why. See `ubuntu-latest` log here:

https://github.com/Rdatatable/data.table/actions/runs/9908622817/job/27374799289

We get:

> test.data.table('other.Rraw') failed to attach required package(s): ggplot2, hexbin, plyr, dplyr, caret, gdata, nlme, sf, nanotime.

And indeed these packages are missing from "all installed packages". But the previous step (`Install dependencies`) has those packages installing without error, e.g.

> `* DONE (ggplot2)`

So now I'm wondering if it installed to a different library than is used by the subsequent step (?), otherwise that installed path is removed in between steps (it has `_temp` in it after all: `/home/runner/work/_temp/Library`). Anyway this diagnostic seems useful to include.


tdhock:
great thanks

Issue #6257, opened 2024-07-14
More potential rchk issues


MichaelChirico:
I just ran `rchk` locally and get

```
Function gmean
  [UP] allocating function gather may destroy its unprotected argument (x <arg 1>), which is later used. /rchk/packages/build/OFS5pfVE/data.table/src/gsumm.c:597
  [UP] calling allocating function gather with a fresh pointer (x <arg 1>) /rchk/packages/build/OFS5pfVE/data.table/src/gsumm.c:597
  [UP] unprotected variable x while calling allocating function Rf_allocVector /rchk/packages/build/OFS5pfVE/data.table/src/gsumm.c:598

Function rbindlist
  [UP] unprotected variable ans while calling allocating function Rf_getAttrib(?,S:names) /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:288
  [UP] unprotected variable coercedForFactor while calling allocating function Rf_getAttrib(?,S:names) /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:288
  [UP] unprotected variable ans while calling allocating function R_compute_identical /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:312
  [UP] unprotected variable coercedForFactor while calling allocating function R_compute_identical /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:312
  [UP] unprotected variable ans while calling allocating function Rf_mkChar /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:322
  [UP] unprotected variable coercedForFactor while calling allocating function Rf_mkChar /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:322
  [UP] unprotected variable ans while calling allocating function Rf_allocVector /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:327
  [UP] unprotected variable coercedForFactor while calling allocating function Rf_allocVector /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:327
  [UP] unprotected variable ans while calling allocating function Rf_copyMostAttrib /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:328
  [UP] unprotected variable coercedForFactor while calling allocating function Rf_copyMostAttrib /rchk/packages/build/OFS5pfVE/data.table/src/rbindlist.c:328
[...]

PR #6258, opened 2024-07-14
More careful memory management in shift()

MichaelChirico:
Related to #6257.

```
Function shift
  [UP] protect stack is too deep, unprotecting all variables, results will be incomplete
  [UP] protect stack is too deep, unprotecting all variables, results will be incomplete
  [UP] unsupported form of unprotect, unprotecting all variables, results will be incomplete /rchk/packages/build/OFS5pfVE/data.table/src/shift.c:174
```

Two things done here:

 - Rather than `PROTECT()` on `thisfill` for every iteration, only waiting till the end to `UNPROTECT()` all of the `thisfill`, instead we release it immediately after each iteration.
 - Don't do anything with `ans` after it's already `UNPROTECT()`'d. This is reminiscent of #6248 -- I'm not sure if there's any issue here, but the new approach does seem safer/more idiomatic.


MichaelChirico:
Adding `Rprintf("nprotect=%d\n", nprotect);` just before `UNPROTECT(nprotect);`.

Before this change:

```r
l <- lapply(seq_len(1e6), identity)
x <- shift(l)
# Error: protect(): protection stack overflow

x <- shift(l[1:1e4])
# nprotect=10001
```

After this change:

```r
l <- lapply(seq_len(1e6), identity)
x <- shift(l)
# nprotect=1

x <- shift(l[1:1e4])
# nprotect=1
```

MichaelChirico:
> Adding `Rprintf("nprotect=%d\n", nprotect);` just before `UNPROTECT(nprotect);`.
> 
> Before this change:
> 
> ```r
> l <- lapply(seq_len(1e6), identity)
> x <- shift(l)
> # Error: protect(): protection stack overflow
> 
> x <- shift(l[1:1e4])
> # nprotect=10001
> ```
> 
> After this change:
> 
> ```r
> l <- lapply(seq_len(1e6), identity)
> x <- shift(l)
> # nprotect=1
> 
> x <- shift(l[1:1e4])
> # nprotect=1
> ```

Based on this I think we've actually uncovered a bug, it's very similar to #5448 -- we can easily trigger a protection stack overflow on `master`. Test with `2e4` elements is 0.012 seconds, pretty slow for a single test but I think OK.

tdhock:
should we be running rchk on CI?

MichaelChirico:
> should we be running rchk on CI?

Let's discuss in #6257

MichaelChirico:
@ben-schwen one more before this first round of `rchk` fixes :)

ben-schwen:
LGTM. Should we link #6257 in the comment above the test?

PR #6264, opened 2024-07-15
Change to use SHALLOW_DUPLICATE_ATTRIB() since that's API since R3.3.0

MichaelChirico:
Follow-up to #6183, part of #6257. See also this comment thread: https://github.com/Rdatatable/data.table/commit/cd89cebd8d3df077c0849da4b964527046e6e83b#r144215033

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -151,10 +151,7 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   // where n is set to truelength (i.e. a shallow copy only with no size change)
   int protecti=0;
   SEXP newdt = PROTECT(allocVector(VECSXP, n)); protecti++;   // to do, use growVector here?
-  SET_ATTRIB(newdt, shallow_duplicate(ATTRIB(dt)));
-  SET_OBJECT(newdt, OBJECT(dt));
-  if (isS4(dt)) {newdt = PROTECT(asS4(newdt, TRUE, 1)); protecti++;} // To support S4 objects that include data.table
-  //SHALLOW_DUPLICATE_ATTRIB(newdt, dt);  // SHALLOW_DUPLICATE_ATTRIB would be a bit neater but is only available from R 3.3.0
+  SHALLOW_DUPLICATE_ATTRIB(newdt, dt);
 
   // TO DO: keepattr() would be faster, but can't because shallow isn't merely a shallow copy. It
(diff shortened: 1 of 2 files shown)


src/dogroups.c
```diff
@@ -482,8 +482,13 @@ SEXP keepattr(SEXP to, SEXP from)
   // Same as R_copyDFattr in src/main/attrib.c, but that seems not exposed in R's api
   // Only difference is that we reverse from and to in the prototype, for easier calling above
   SET_ATTRIB(to, ATTRIB(from));
```
ben-schwen, line 484: Note that `SET_OBJECT` will also be deprecated. Maybe we should switch to `SHALLOW DUPLICATE ATTRIBUTE` here

void SHALLOW_DUPLICATE_ATTRIB(SEXP to, SEXP from) {
    SET_ATTRIB(CHK(to), shallow_duplicate(CHK(ATTRIB(CHK(from)))));
    SET_OBJECT(CHK(to), OBJECT(from));
    IS_S4_OBJECT(from) ?  SET_S4_OBJECT(to) : UNSET_S4_OBJECT(to);
}

src/assign.c
```diff
@@ -153,7 +153,7 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   SEXP newdt = PROTECT(allocVector(VECSXP, n)); protecti++;   // to do, use growVector here?
   SET_ATTRIB(newdt, shallow_duplicate(ATTRIB(dt)));
   SET_OBJECT(newdt, OBJECT(dt));
```
ben-schwen, line 155: Backport `SHALLOW_DUPLICATE_ATTRIB` and use it here?

https://github.com/wch/r-source/blob/8e4314e5929e2a2f890fda48301324865b79d2c9/src/main/memory.c#L3944-L3948

MichaelChirico: Nice find... I think we can just go ahead and use it?

image

I would leave this as a TODO for now, given the experience with `isDataFrame()`/ #6235 -- let's wait for the `NOTE` to start throwing at least.

MichaelChirico: Per the comment just below, it's available from 3.3.0 (which we now use), so I went ahead and made the switch.

MichaelChirico:
ping @ben-schwen  :)

ben-schwen:
LGTM

PR #6265, opened 2024-07-15
PROTECT() names vector

MichaelChirico:
Part of #6257.

This _might_ resolve the issues there about `rbindlist`; the line numbers cited there don't seem correct, but this nearby _does_ look incorrect.

### DIFF
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -242,8 +242,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
 
   int nprotect = 0;
   SEXP ans = PROTECT(allocVector(VECSXP, idcol + ncol)); nprotect++;
-  SEXP ansNames;
-  setAttrib(ans, R_NamesSymbol, ansNames=allocVector(STRSXP, idcol + ncol));
+  SEXP ansNames = PROTECT(allocVector(STRSXP, idcol + ncol)); nprotect++;
+  setAttrib(ans, R_NamesSymbol, ansNames);
   if (idcol) {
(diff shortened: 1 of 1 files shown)


src/rbindlist.c
```diff
@@ -242,8 +242,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg)
 
   int nprotect = 0;
   SEXP ans = PROTECT(allocVector(VECSXP, idcol + ncol)); nprotect++;
-  SEXP ansNames;
-  setAttrib(ans, R_NamesSymbol, ansNames=allocVector(STRSXP, idcol + ncol));
+  SEXP ansNames = PROTECT(allocVector(STRSXP, idcol + ncol)); nprotect++;
+  setAttrib(ans, R_NamesSymbol, ansNames);
```
ben-schwen, line 246: Is this the 2nd case of

https://github.com/Rdatatable/data.table/blob/9e2601e04a88610f28c798d047ebeb6103bd6fb2/.dev/CRAN_Release.cmd#L167-L173

and therefore not needed?

ben-schwen: Relying on ii) also violates Tomas Kalibera suggestion on [callee-protection](https://blog.r-project.org/2019/04/18/common-protect-errors/)
> Historically, some core R API functions are callee-protect, they protect their own arguments and keep them protected for the whole duration of their call. It is better not to rely on this property, but it is often done and rchk tries to detect callee-protect functions and not report an error.

MichaelChirico: Filed #6280 to revisit this.

MichaelChirico:
ping @ben-schwen :)

ben-schwen:
LGTM, ty!

PR #6270, opened 2024-07-16
Use malloc() over variable-length array

### LINKED ISSUE #5528: Variable-length array compilation failure under -Werror
These declarations trigger the `vla` warning:

https://github.com/Rdatatable/data.table/blob/1b324637045c97b40960fa64a4dc4dc27b00596a/src/fsort.c#L128

See also https://stackoverflow.com/questions/61825661/why-does-clang-complain-about-using-variable-length-arrays-with-std-c99-flag

MichaelChirico:
Closes #5528


tdhock (approved):
great thanks

PR #6271, opened 2024-07-16
Consistently error when allocation request fails

### LINKED ISSUE #1115: Completion of error handling
Would you like to add more error handling for return values from functions like the following?
- [clock](http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock.html "Report CPU time used.") ⇒ [readfile](https://github.com/Rdatatable/data.table/blob/6601aab68559308654ab6dad7e2181f510a44b1f/src/fread.c#L416)
- [malloc](http://pubs.opengroup.org/onlinepubs/9699919799/functions/malloc.html "Memory allocation") ⇒ [preprocess](https://github.com/Rdatatable/data.table/blob/6601aab68559308654ab6dad7e2181f510a44b1f/src/fmelt.c#L300)

MichaelChirico:
Part of #1115

Also

 - More consistently put the corresponding error on its own line, good for code coverage
 - More consistently `free()` all objects when we allocate several consecutively
 - Don't do the checking for `Calloc()`, which is very explicitly advertised not to need this checking: https://cran.r-project.org/doc/manuals/r-release/R-exts.html#User_002dcontrolled-memory

TODO:

- [x] Just use `free(); free();` instead of `if (.) free(.)`, for conciseness
- [x] Consistency in error for translation purposes
- [x] Come up with a way to prevent backsliding


tdhock (approved):
this looks good to me. Thanks!
maybe add a NEWS item about the linter?

PR #6274, opened 2024-07-16
Explicitly request -Wvla to prevent backsliding

MichaelChirico:
Follow-up to #6270.

IIUC, per [here](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-pedantic-errors-1), we could also make this by supplying `-pedantic-errors` (instead of/in addition to `-pedantic` which we do already?).

I tried to find similar docs for clang/llvm but IME finding docs for compiler warnings is rarely fruitful...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -117,8 +117,8 @@ test-lin-rel:
     OPENBLAS_MAIN_FREE: "1"
   script:
     - *install-deps
-    - echo 'CFLAGS=-g -O3 -flto=auto -fno-common -fopenmp -Wall -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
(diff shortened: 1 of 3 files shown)


ben-schwen:
LGTM, but we should remove VLA before as in e.g.

https://github.com/Rdatatable/data.table/blob/a01f00f7438daf4612280d6886e6929fa8c8f76e/src/forder.c#L915

https://github.com/Rdatatable/data.table/blob/a01f00f7438daf4612280d6886e6929fa8c8f76e/src/forder.c#L1214

MichaelChirico:
Good point. There were a few others too. Now passing `R CMD check` under `-Wvla`. PTAL -- in particular on how `TMP` is `free()`'d in forder, I wasn't 100% confident there, could use the extra eyes.

jangorecki (approved):

ben-schwen:
CI is now complaining about these flags in:
- lin-dev-clang-cran
- lin-dev-gcc-strict-cran
- lin-rel-cran

```
Compilation used the following non-portable flag(s):
  ‘-Wvla’
```

PR #6276, opened 2024-07-16
Supply origin= in test failing on old R

MichaelChirico:
Observed on R3.5. I'm not sure what versions of R are affected.

https://github.com/Rdatatable/data.table/actions/runs/9957719789/job/27510348749


tdhock (approved):
great thanks

PR #6277, opened 2024-07-16
Optionally skip test relying on R>=4.1.0 behavior

MichaelChirico:
That's causing this observed issue:

```
Test 2247.3 ran without errors but failed check that x equals y:
> x = do.call(rbind, split(dt, ~y)) 
   x y  [Key= Types=int,int Classes=int,int]
1: 1 1                                      
2: 3 1                                      
3: 2 2                                      
4: 4 2                                      
> y = setDT(do.call(rbind, split(as.data.frame(dt), ~y))) 
First 1 of 1 (type 'character'): 
[1] "Error in unique.default(x, nmax = nmax) : \n  unique() applies only to vectors\n"
```

The error is from `~y` being coerced to factor, c.f. `as.factor(~y)`.

https://github.com/Rdatatable/data.table/actions/runs/9957719789/job/27510350948


tdhock:
great thanks

PR #4558, opened 2020-06-18
towards #4200, overhead in dogroups

jangorecki:
timings in https://github.com/Rdatatable/data.table/issues/4200#issuecomment-645980224
probably qualifies to close #4200

Issue #4200, opened 2020-01-26
groupby with dogroups (R expression) performance regression


jangorecki:
There is a performance regression (AFAIU) when doing by group computation where we run R's C eval by each group (q7 and q8 in db-benchmark). 
```
    in_rows question_group                    question 20181206_da98fb2 20190913_35b0de3 20191115_92abb70 20191205_eba8704 20191209_6808d2c 20191212_e0140ea 20191229_d52b0d8 20200124_c005296
 1:     1e9          basic               sum v1 by id1           21.144           10.026            8.375            9.362            9.060            9.082            9.366            9.271
 2:     1e9          basic           sum v1 by id1:id2           38.914           11.746            9.243            9.327            9.331            9.978           10.813            9.220
 3:     1e9          basic       sum v1 mean v3 by id3           99.517           14.487           12.044           12.291           13.496           14.325           13.191           13.169
 4:     1e9          basic           mean v1:v3 by id4           26.593           17.357           15.135           15.157           15.278           16.761           16.724           16.754
 5:     1e9          basic            sum v1:v3 by id6          122.214           14.569           13.454           14.035           14.046           14.842           14.400           15.019
 6:     1e9       advanced  median v3 sd v3 by id4 id5               NA               NA          121.742          110.925          106.340          113.837          111.984          123.411
 7:     1e9       advanced      max v1 - min v2 by id3               NA               NA           98.680           91.596           87.005           93.749           91.294          299.863
 8:     1e9       advanced       largest two v3 by id6               NA               NA          234.926          215.574          213.824          216.152          211.295          411.241
 9:     1e9       advanced regression v1 v2 by id2 id4               NA           72.466           81.297           76.121           75.311           74.769           72.571           41.157
10:     1e9       advanced     sum v3 count by id1:id6               NA          180.257          196.403          187.816          177.257          187.702          190.282          187.403
```
worth to note that at the same time q9 `x[, .(r2=cor(v1, v2)^2), by=.(id2, id4)]` got nice speed up

jangorecki:
```r
library(data.table)
N = 1e6L
set.seed(108)
d = data.table(id3 = sample(c(seq.int(N*0.9), sample(N*0.9, N*0.1, TRUE))), # 9e5 unq values
               v1 = sample(5L, N, TRUE),
               v2 = sample(5L, N, TRUE))
system.time(d[, max(v1)-min(v2), by=id3])
system.time(d[, max(v1)-min(v2), by=id3])

[...]

PR #6287, opened 2024-07-17
adding an atime test case; groupby with dogroups (R expression)  #PR4558

DorisAmoakohene:
This test case discusses the issue reported on performing group computations, specifically when running R's C eval on each group (q7 and q8) in the db-benchmark, indicating a slowness in the implementation of the code. https://github.com/Rdatatable/data.table/issues/4200

This is the https://github.com/Rdatatable/data.table/pull/4558 that discusses the Cause of the Regression: https://github.com/Rdatatable/data.table/issues/4200#issue-555186870 which appears that the regression occurred during the evaluation of C code within these particular groups, indicating a performance issue or slowness in the implementation of the code.

[The regression was fixed Regression by the addition of const int nth = getDTthreads](https://github.com/Rdatatable/data.table/pull/4558/files)


tdhock:
why is there another similar (identical?) PR https://github.com/Rdatatable/data.table/pull/6288 ?
If there is no difference, please only make one.

Issue #3735, opened 2019-07-30
Selecting from data.table by row is very slow


chnynf:
`allIterations <- data.frame(v1 = runif(1e5), v2 = runif(1e5))`
`DoSomething <- function(row) {
  someCalculation <- row[["v1"]] + 1
}
`

`> system.time(
       {
         for (r in 1:nrow(allIterations)) {
           DoSomething(allIterations[r, ])
         }
       }
     )`
   user  system elapsed 
   4.50    0.02    4.55 

`> allIterations <- as.data.table(allIterations)`
`> system.time(
       {
         for (r in 1:nrow(allIterations)) {
           DoSomething(allIterations[r, ])
         }
       }
     )`
   user  system elapsed 
  53.78   25.05   78.46 `

I'm working on a R project that involves applying fairly complicated functions across data.table or data.frame by rows.
In cases where vectorizing is not a good option, one might need to loop through rows, and that's when I realized selecting by row number from a data.table is actually much slower than from a data.frame.
I guess selecting by row number is not a recommended practice for data.table? Or would the team be interested in looking into this and optimize the performance?

I have more details about my test [here](https://github.com/chnynf/r-experiments/blob/master/speed-tests.R).

shrektan:
If loop on all the rows is unavoidable, I suggest you to use `purrr::pmap()`. 

``` r
df <- data.frame(v1 = runif(1e3), v2 = runif(1e3))
cal <- function(row) {
  row[['v1']] + 1
}
res1 <- res2 <- res3 <- res4 <- double(nrow(df))

t <- proc.time()
for (r in 1:nrow(df)) {
  res1[r] <- cal(df[r, ])
}
data.table::timetaken(t)
#> [1] "0.110s elapsed (0.110s cpu)"

dt <- data.table::as.data.table(df)
t <- proc.time()
for (r in 1:nrow(dt)) {
  res2[r] <- cal(dt[r, ])
}
data.table::timetaken(t)
#> [1] "0.530s elapsed (0.450s cpu)"

t <- proc.time()
res3 <- purrr::pmap_dbl(dt, ~ .x + 1)
data.table::timetaken(t)
#> [1] "0.040s elapsed (0.050s cpu)"

all.equal(res2, res1)
#> [1] TRUE
all.equal(res3, res1)
#> [1] TRUE
```

<sup>Created on 2019-07-31 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

[...]

PR #6289, opened 2024-07-17
adding an atime test case; new with optimization to allow avoid [ overhead #PR4488 

DorisAmoakohene:
This regression was reported in https://github.com/Rdatatable/data.table/issues/3735. They reported that selecting by row number from a data.table is much slower than from a data.frame.

The pull request that fixed the regression can be found https://github.com/Rdatatable/data.table/pull/4488. However, I couldn’t find specific details about the changes made to fix the regression. It’s likely that the changes involved optimizing the way data.table handles row selection to make it more efficient.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -119,6 +119,26 @@ test.list <- atime::atime_test_list(
       data.table:::setDT(L)
     },
     Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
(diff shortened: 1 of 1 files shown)


tdhock:
why did you open this PR and then immediately close?
Also please use more informative branch names. Current branch name is DorisAmoakohene-patch-2 which does not say anything about what you are trying to do in the PR.
something like perf-test-PR4488 would be better

DorisAmoakohene:
> why did you open this PR and then immediately close? Also please use more informative branch names. Current branch name is DorisAmoakohene-patch-2 which does not say anything about what you are trying to do in the PR. something like perf-test-PR4488 would be better

I didnt close it, I renamed the branch and it closed the PR, so the new one can be found here https://github.com/Rdatatable/data.table/pull/6290, I am still figuring out how to rename the branch name without having it duplicate the PR

tdhock:
git checkout -b your_branch_name

tdhock:
please use the command line git instead of the github web interface to submit PRs

Issue #6284, opened 2024-07-16
revdep scoringutils (.optmean) errors after non-ascii by fix


tdhock:
https://github.com/Rdatatable/data.table/pull/4711 broke revdep scoringutils https://rcdata.nau.edu/genomic-ml/data.table-revdeps/analyze/2024-07-16/
```
> system(paste(c("diff -u", Rcheck.list), collapse=" "))
--- R_Under_development_unstable_2024-07-15_r86902/release_1.15.4.Rcheck/00check.log	2024-07-16 01:24:12.695554623 -0700
+++ R_Under_development_unstable_2024-07-15_r86902/master_1.15.99.9e2601e04a88610f28c798d047ebeb6103bd6fb2.Rcheck/00check.log	2024-07-16 01:27:23.045737401 -0700
@@ -54,13 +54,75 @@
 * checking data for ASCII and uncompressed saves ... OK
 * checking installed files from 'inst/doc' ... OK
 * checking files in 'vignettes' ... OK
-* checking examples ... OK
+* checking examples ... ERROR
+Running examples in 'scoringutils-Ex.R' failed
+The error most likely occurred in:
+
+> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
+> ### Name: add_coverage
+> ### Title: Add coverage of central prediction intervals
+> ### Aliases: add_coverage
+> ### Keywords: scoring
+> 
+> ### ** Examples
+> 
+> library(magrittr) # pipe operator
+> ## Don't show: 
+>   data.table::setDTthreads(2) # restricts number of cores used on CRAN
+> ## End(Don't show)
+> score(example_quantile) %>%
++   add_coverage(by = c("model", "target_type")) %>%
++   summarise_scores(by = c("model", "target_type")) %>%
++   summarise_scores(fun = signif, digits = 2)
+The following messages were produced when checking inputs:
+1.  144 values for `prediction` are NA in the data provided and the corresponding rows were removed. This may indicate a problem if unexpected.
+Error in startsWith(names(expr)[3L], "na") : non-character object(s)
+Calls: %>% ... summarise_scores -> [ -> [.data.table -> lapply -> FUN -> startsWith
+Execution halted
 * checking for unstated dependencies in 'tests' ... OK
-* checking tests ... OK
+* checking tests ... ERROR
   Running 'testthat.R'
+Running the tests in 'tests/testthat.R' failed.
+Last 13 lines of output:
+  * plot_pit/plot-pit-quantile.svg
[...]

PR #6291, opened 2024-07-17
Handle mean(a,b) under R's startsWith()

MichaelChirico:
Part of #6284 

We used to have a backport:

```r
startsWith = function(x, stub) substr(x, 1L, nchar(stub))==stub
```

`substr()` coerces its argument to character, but `base::startsWith()` does not, hence the change in behavior. Anyway, the use of `startsWith()` was still error-prone; this PR adds further robustness checks as well.

Confirmed the test fails on `master`, passes on branch. Also confirmed that `as.character()` is not sufficient -- without `isTRUE()`, the same test errors as "missing value where TRUE/FALSE needed".

Also checked all other `startsWith()` call sites to be sure none of them might have silently been accepting non-character input.

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1987,11 +1987,11 @@ DT = function(x, ...) {  #4872
 
 .optmean = function(expr) {   # called by optimization of j inside [.data.table only. Outside for a small speed advantage.
   if (length(expr)==2L)  # no parameters passed to mean, so defaults of trim=0 and na.rm=FALSE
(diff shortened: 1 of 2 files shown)


tdhock:
nice work thanks! please add NEWS item.

MichaelChirico:
We are just restoring a behavior present on CRAN, so no NEWS. i.e. #6232 is a regression in dev that we've fixed.

PR #6295, opened 2024-07-17
adding an atime test case  for gshift as gforce optimized shift  #5205

DorisAmoakohene:
[In this test case the user mentions that the shift() function in data.table version 1.9.6 is slow when applied to many groups](https://github.com/Rdatatable/data.table/issues/1534). The exact cause or specific use case is not elaborated upon in the issue.[visit here to read more](https://github.com/Rdatatable/data.table/issues/1534)

[The pull request address the issue by allowing GForce optimization for lapply even without using .SD1. ](https://github.com/Rdatatable/data.table/pull/5205)


tdhock:
Hi I converted to draft because I think you should first merge your first PR, and then use what you learn from that experience to improve your other PRs, before asking others to review them.

DorisAmoakohene:
> Hi I converted to draft because I think you should first merge your first PR, and then use what you learn from that experience to improve your other PRs, before asking others to review them.

I have made most corrections to this PR  for review and merge

tdhock:
please resolve conflicts and fix typo shif -> shift

DorisAmoakohene:
> please resolve conflicts and fix typo shif -> shift

I have resolved this. Thank you

.ci/atime/tests.R
```diff
@@ -134,4 +135,42 @@ test.list <- atime::atime_test_list(
...
+    ## New DT can safely retain key.
+    expr = {
+      data.table:::`[.data.table`(L, , .SD)
+    },
+    Fast = "353dc7a6b66563b61e44b2fa0d7b73a0f97ca461", # Close-to-last merge commit in the PR (https://github.com/Rdatatable/data.table/pull/4501/commits) that fixes the issue 
+    Slow = "3ca83738d70d5597d9e168077f3768e32569c790", # Circa 2024 master parent of close-to-last merge commit (https://github.com/Rdatatable/data.table/commit/353dc7a6b66563b61e44b2fa0d7b73a0f97ca461) in the PR (https://github.com/Rdatatable/data.table/pull/4501/commits) that fixes the issue 
+    Slower = "cacdc92df71b777369a217b6c902c687cf35a70d"), # Circa 2020 parent of the first commit (https://github.com/Rdatatable/data.table/commit/74636333d7da965a11dad04c322c752a409db098) in the PR (https://github.com/Rdatatable/data.table/pull/4501/commits) that fixes the issue 
+
+  NULL)
+>>>>>>> master
```
tdhock, line 175: you need to remove >>> and === which are git annotations, not R code

.ci/atime/tests.R
```diff
@@ -171,6 +170,4 @@ test.list <- atime::atime_test_list(
     Slow = "3ca83738d70d5597d9e168077f3768e32569c790", # Circa 2024 master parent of close-to-last merge commit (https://github.com/Rdatatable/data.table/commit/353dc7a6b66563b61e44b2fa0d7b73a0f97ca461) in the PR (https://github.com/Rdatatable/data.table/pull/4501/commits) that fixes the issue 
     Slower = "cacdc92df71b777369a217b6c902c687cf35a70d"), # Circa 2020 parent of the first commit (https://github.com/Rdatatable/data.table/commit/74636333d7da965a11dad04c322c752a409db098) in the PR (https://github.com/Rdatatable/data.table/pull/4501/commits) that fixes the issue 
 
-  NULL)
```
tdhock, line 159: please undo deletion of this line which is required to terminate the test list

Anirban166 (approved):
Looks good to go!

tdhock:
currently we see this result 
image
there are two issues
1. there is not a huge separation between Regression and Fixed at largest N, but there is some separation at smaller N, why? Can you please investigate, perhaps with a larger N? and double check the historical commit IDs?
2. There is no Before, so please either add that, or change to Fast and Slow.

DorisAmoakohene:
> currently we see this result image
> 
> there are three issues
> 
> 1. there is not a huge time separation between Regression and Fixed at largest N, but there is some separation at smaller N, why? Can you please investigate, perhaps with a larger N? and double check the historical commit IDs?
> 2. There is no Before, so please either add that, or change to Fast and Slow.
> 3. why does Regression use less memory than Fixed?

Alright

DorisAmoakohene:
> currently we see this result image
> 
> there are three issues
> 
> 1. there is not a huge time separation between Regression and Fixed at largest N, but there is some separation at smaller N, why? Can you please investigate, perhaps with a larger N? and double check the historical commit IDs?
> 2. There is no Before, so please either add that, or change to Fast and Slow.
> 3. why does Regression use less memory than Fixed?

I investigated further with a larger sample size (N), but found no difference, it stops when it gets to 10000

. Even when I reverted to the  commit to the commit before the last commits, where the PR was merge nothing changes.

pic5205

tdhock:
the original benchmarks were reported by @ben-schwen in https://github.com/Rdatatable/data.table/pull/5205#issuecomment-943509212
there are two sizes in the benchmarks. 
* number of rows of DT
* number of values in sample (y column)

your current setup has ` DT = data.table(x = sample(N), y = sample(1e2, N, TRUE))`  so N controls number of rows of DT, not number of values in sample (y column) which is currently fixed at 1e2. Maybe you could try varying that instead?
Or varying both?

MichaelChirico:
@tdhock what's the status here?

tdhock:
let's close for now and open another pr with a similar test case.

PR #6296, opened 2024-07-17
Fix Performance of 'by' Operations when `verbose=TRUE`

### LINKED ISSUE #6286: 'by' operations much slower when verbose=TRUE
Seen when benchmarking #6228 , seems that by operations run _much_ slower with `verbose=TRUE`

```R
dt = data.table(a = 1:1000000)
system.time(copy(dt)[, 1, by = a])
#   user  system elapsed 
#  0.263   0.001   0.202

system.time(copy(dt)[, 1, by = a, verbose=TRUE])
#   user  system elapsed 
#  1.375   1.505   2.820
```

My initial guess for the culprits are the numerous calls to `clock()` when `verbose=TRUE`

`#` `Output of sessionInfo()`
```
R version 4.4.1 (2024-06-14)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 22.04.4 LTS

Matrix products: default

locale:

time zone: America/Vancouver
tzcode source: system (glibc)

attached base packages:

other attached packages:
[1] data.table_1.15.99 testthat_3.2.1.1   devtools_2.4.5     usethis_2.2.3     

 [1] miniUI_0.1.1.1    compiler_4.4.1    brio_1.1.5        promises_1.3.0   
 [5] Rcpp_1.0.12       stringr_1.5.1     callr_3.7.6       later_1.3.2      
 [9] fastmap_1.2.0     mime_0.12         R6_2.5.1          htmlwidgets_1.6.4
[13] desc_1.4.3        profvis_0.3.8     rprojroot_2.0.4   shiny_1.8.1.1    
[17] rlang_1.1.4       cachem_1.1.0      stringi_1.8.4     httpuv_1.6.15    
[21] fs_1.6.4          pkgload_1.3.4     memoise_2.0.1     cli_3.6.2        
[25] withr_3.0.0       magrittr_2.0.3    ps_1.7.6          processx_3.8.4   
[29] digest_0.6.35     rstudioapi_0.16.0 xtable_1.8-4      remotes_2.5.0    
[33] lifecycle_1.0.4   vctrs_0.6.5       glue_1.7.0        urlchecker_1.0.1 
[37] sessioninfo_1.2.2 pkgbuild_1.4.4    purrr_1.0.2       tools_4.4.1      
[41] ellipsis_0.3.2    htmltools_0.5.8.1

```

joshhwuu:
Closes #6286 

This PR changes calls to C's `clock()` to internal `wallclock()` in `dogroups.c`. Apparently calling `clock()` repeatedly on non-Windows platforms used to cause a significant overhead, leading to significantly longer processing times. I found this [SO](https://stackoverflow.com/questions/27178789/c-different-implementation-of-clock-in-windows-and-other-os) thread that discusses some of the differences between C's `clock()` on different platforms. 

- [ ] Add a performance test


Anirban166:
I created a performance test for this - [Result](https://github.com/Anirban166/data.table/pull/13#issuecomment-2235046139)

Could probably do with a better/succinct name for the title (open to suggestions)

I'm not using a 'Fast' label (as displayed in the latest plot) since it would only exist after this PR is merged onto master here and is available to be installed using the commit SHA associated with it (after this is merged, I can create a follow-up PR for the test). Doesn't seem to be the case where this was previously fast so 'Before' and related labels won't fit. I initially tested 1.15.4 vs 1.15.99 for checking on [this](https://github.com/Rdatatable/data.table/issues/6286#issuecomment-2231962058) - no change (I assume you switched operating systems in between the process of testing other versions yesterday?) so likely not a regression.

Anirban166 (approved):

tdhock:
> Not sure why this is happening, can anyone chime in?

please file an issue with additional details (OS, full input/output, traceback) https://github.com/tdhock/atime/issues

joshhwuu:
> no change (I assume you switched operating systems in between the process of testing other versions yesterday?) so likely not a regression.

Yup, I confirmed this [here](https://github.com/Rdatatable/data.table/issues/6286#issuecomment-2231986783)

Anirban166:
@MichaelChirico when you have some time, could you please review this? (looks merge-able to me; I will add a test following this)

tdhock:
if you want to make sure a PR is reviewed/merged before the next releaase, please tag it in the upcoming release milestone

joshhwuu:
There also may be reason to reevaluate all calls to `clock()` in the codebase, I may take some time and do some testing if I find the time

Anirban166:
> if you want to make sure a PR is reviewed/merged before the next releaase, please tag it in the upcoming release milestone

Noted! I'll keep that in mind, but I'll probably let Michael and the rest decide on those things given that they would know their workload and what things can be easily included or not the best.

For this, there is no urgency as in we want it to be within the next release, but I thought this might have been forgotten or in the gray area where a reader might think that discussion is pending.

MichaelChirico:
Thanks for the ping!

> Noted! I'll keep that in mind, but I'll probably let Michael and the rest decide on those things given that they would know their workload and what things can be easily included or not the best.

No problem at all throwing new things into the milestone -- go ahead! We will also have no problem bumping them to a different release later if needed :) That sort of thing might be discussed in the release issue.

As next non-patch release is not due for some months, putting the issue in the milestone now doesn't change the urgency IMO -- just earmarks a PR, whether it be for importance, simplicity, strong request, etc.

MichaelChirico:
I would add a NEWS entry under NOTES.

MichaelChirico:
The change looks fine, but it strikes me as odd for us to re-use `wallclock()` which is implemented in fread.c:

https://github.com/Rdatatable/data.table/blob/e73c2c849f921cf4ef51e3809842e0fee9b9f52c/src/fread.c#L379-L392

I typically think of fread.c as a separate library, since its code is also used by pydatatable. OTOH, the function is declared in data.table.h, _not_ fread.h, and we also have calls to `wallclock()` in other C files, e.g. `fsort()` and `gsumm()`. So maybe it's fine.

joshhwuu:
> and we also have calls to `wallclock()` in other C files, e.g. `fsort()` and `gsumm()`. So maybe it's fine.

This was my thought originally, maybe we could move implementation to a more centralized/clear location?

MichaelChirico:
My worry was `wallclock()` is shared by pydatatable, but it looks like they have their own implementation?

https://github.com/h2oai/datatable/blob/b1a87103c55433dc2069df28a460f3ad15450025/src/core/utils/misc.cc#L137-L144

Anyway we can move that to follow-up...

MichaelChirico:
Do you still plan to add a performance test here? I am OK omitting it, as noted performance of `verbose=TRUE` is not crucial, so I am fine not tracking that too carefully. OTOH, the performance improvement you noted here is pretty noteworthy.

tdhock:
this looks like a good start for performance test, https://github.com/Anirban166/data.table/pull/13#issuecomment-2235046139 because we can clearly see the difference between Slow and HEAD but please add a Fast commit from this PR

Anirban166:
I'm cool with omitting it too, given that the general and default use is with `verbose = FALSE`

Anirban166:
> Anirban166#13 (comment)

Alright then we're keeping it, and yes I will after this PR is merged.

Anirban166:
> this looks like a good start for performance test, [Anirban166#13 (comment)](https://github.com/Anirban166/data.table/pull/13#issuecomment-2235046139) because we can clearly see the difference between Slow and HEAD but please add a Fast commit from this PR

Alright then we're keeping it, and yes I will after this PR is merged.

Anirban166:
I need a name for the test though, so what would be your thoughts for the test case title @tdhock?

MichaelChirico (approved):

MichaelChirico:
OK I am fine with doing the atime test in a follow-up (actually I'm thinking that more generally that is the better approach).

@joshhwuu please add a NEWS entry

tdhock:
you are right, adding performance tests in separate PRs is a great idea.
smaller PRs are always better, easier for submitters, easier for reviewers.
plus Pr submitters maybe don't know how to create performance tests.

MichaelChirico:
The downside is it's often best to write the performance test immediately lest it fall through the cracks.

Another advantage though is once the PR is submitted, it's no longer an issue to pick a commit hash -- while a PR is live, that hash might change a few times.

OTOH, if the performance test reveals a regression, it would have made sense to fix it in the original PR.

Still, I think separation to another PR is on balance better.

NEWS.md
```diff
@@ -22,6 +22,8 @@
 
 4. The translations submitted for 1.16.0 are now actually shipped with the package -- our deepest apologies to the translators for the omission. We have added a CI check to ensure that the .mo binaries which get shipped with the package are always up-to-date.
 
+5. Grouping operations now use internal `wallclock()` for verbose reporting, reducing a significant overhead on non-Windows platforms introduced by C's native `clock()` function.
```
Anirban166, line 25: News LGTM. Thanks @joshhwuu!

Anirban166 (approved):

NEWS.md
MichaelChirico, line 25: Generally I think we should avoid writing about highly internal things like "wallclock" in the NEWS

```suggestion
5. Some grouping operations run much faster under `verbose=TRUE`, [#6286](https://github.com/Rdatatable/data.table/issues/6286). Thanks @joshhwuu for the report and fix. This overhead was not present on Windows. As a rule, users should expect `verbose=TRUE` operations to run more slowly, as extra statistics might be calculated as part of the report; here was a case where the overhead was particularly high and the fix was particularly easy.
```

Anirban166: I agree, abstraction might be better for such details

Anirban166: Also, nice edit!

Issue #4800, opened 2020-11-04
Print a warning/... when creating data.table with POSIXlt


iago-pssjd:
I was doing `setDT()` of a dataframe (or tibble) with a variable `POSIXlt`. It seems not to be any problem. Even I can do `str()` of that object. It does not show anything strange. Then errors started to appear when I tried to manipulate the object or simply do `head()`.

As I have [found](https://stackoverflow.com/questions/21487614/error-creating-r-data-table-with-date-time-posixlt) the problem, I would ask/suggest why do not produce the error when running `setDT` or `data.table` or, otherwise show a warning?, which would tell about the problem that there are POSIXlt variables and they are not supported. Possibly these variables could be removed or coerced to POSIXct or another better format for `data.table` and the warning would tell about that.

Any case, thank you for this great package.

jangorecki:
@iago-pssjd Thank you for your report. What would be even more better is to include minimal example (so we can just copy copy to observe issue you are describing) and your session info (to check if you might be using older version of DT and issue might have been fixed already).

iago-pssjd:
Thank you @jangorecki . I updated the issue and removed the reference to `data.table` which i toke just after look at the SO question, but I didn't check it. Now I see that this function does what I asked for, but `setDT` yet not.

jangorecki:
Thank you.
We can define an action point for this issue to add check for POSIXlt inside setDT.

shrektan:
It looks like a problem of `print.data.table()` to me and not `setDT()`.  `print.data.table()` should work with a data.table object that contains POSIXlt vectors. 

In my opinion, generally, we should not change the vector value of the object by `setDT()`, unless there're very strong reasons.

iago-pssjd:
@shrektan is not only a problem of `print.data.table()`, since if you want to do other operations on the object, it gives an error too (let me remember which specific operations, but I believe I was trying to create some variable, and I believe it had nothing to do with the time variable, bot sure also if using `by = `...)

shrektan:
I can't understand this because I think `POSIXlt` is nothing special and should not affect other usage, so an example is welcomed.

jangorecki:
If we change POSIXlt to POSIXct it won't be by reference anymore so I think the best is to raise error. Other operations will also warn on POSIXlt input.

shrektan:
[...]

Issue #2068, opened 2017-03-20
confusing POSIXlt warning


tdhock:
Hey guys thanks for all your hard work making data.table great!

Can you please clarify the warning that is issued when `strptime` is used inside of the second argument of `[.data.table`? Right now it says "POSIXlt column type detected and converted to POSIXct" even in cases when `strptime` does not operate on columns at all. 

For example this is a MRE which is potentially confusing:

```
> data.table(X=1)[, { strptime("1984-03-17", "%Y-%m-%d"); X}]
[1] 1
Warning message:
In strptime("1984-03-17", "%Y-%m-%d") :
  POSIXlt column type detected and converted to POSIXct. We do not recommend use of POSIXlt at all because it uses 40 bytes to store one date.
> devtools::session_info()
Session info -------------------------------------------------------------------
 setting  value                       
 version  R version 3.3.3 (2017-03-06)
 system   i686, linux-gnu             
 ui       X11                         
 language en_US                       
 collate  en_US.UTF-8                 
 tz       posixrules                  
 date     2017-03-20                  

Packages -----------------------------------------------------------------------
 package       * version    date       source                                  
 bitops          1.0-6      2013-08-17 CRAN (R 3.2.2)                          
 caTools         1.17.1     2014-09-10 CRAN (R 3.2.2)                          
 colorspace      1.3-0      2016-11-01 R-Forge (R 3.3.1)                       
 data.table    * 1.10.5     2017-03-20 Github (Rdatatable/data.table@ce00cb1)  
 devtools        1.12.0     2016-12-05 CRAN (R 3.3.3)                          
 digest          0.6.10     2016-08-02 CRAN (R 3.2.2)                          
 ggplot2       * 2.1.0      2016-11-09 Github (faizan-khan-iit/ggplot2@5fb99d0)
 gtable          0.2.0      2016-02-26 CRAN (R 3.2.2)                          
 lattice       * 0.20-34    2016-09-06 CRAN (R 3.3.3)                          
 memoise         1.0.0      2016-01-29 CRAN (R 3.2.2)                          
 munsell         0.4.3      2016-02-13 CRAN (R 3.2.2)                          
 namedCapture  * 2015.12.01 2015-12-21 Github (tdhock/namedCapture@0517592)    
 plyr            1.8.4      2016-06-08 CRAN (R 3.2.2)                          
 RColorBrewer  * 1.1-2      2014-12-07 CRAN (R 3.2.2)                          
[...]

PR #6299, opened 2024-07-17
automatically convert posixlt to posixct in set/:=/<-

### LINKED ISSUE #1724: := should as.POSIXct when RHS is POSIXlt
I am able to add a POSIXct column to a data.frame, but not data.table.

``` R
data <- read.table(header=TRUE, text='
 id weight
  1     20
  2     27
  3     24
')
dates <- c(strptime('160523-1600', '%y%m%d-%H%M'),strptime('160523-1600', '%y%m%d-%H%M'),strptime('160523-1600', '%y%m%d-%H%M'))
dt <- data.table(data)
# This works well
data$dates <- dates
# This results in an error
dt$dates <- dates
```

The error message I get is:

> Warning message:
> In `[<-.data.table`(x, j = name, value = value) :
>   Supplied 11 items to be assigned to 3 items of column 'dates' (8 unused)

And the `dates` field is not properly assigned to `dt`.

ben-schwen:
Closes #1724


tdhock:
related to an issue I was having many years ago https://github.com/Rdatatable/data.table/issues/2068#issuecomment-288197361
It sounds like this PR would fix that issue.
Briefly, I expected that any posixlt RHS of `:=` should be converted to posixct before storing in a data.table column, as when we call `data.table()`.
However I believe the current logic is: whenever strptime is used in `DT[, j]`, wrap it in as.POSIXct as we see below
```r
> data.table(x=strptime("2024-01-01","%Y-%m-%d"))
            x
       <POSc>
1: 2024-01-01
Warning message:
In as.data.table.list(x, keep.rownames = keep.rownames, check.names = check.names,  :
  POSIXlt column type detected and converted to POSIXct. We do not recommend use of POSIXlt at all because it uses 40 bytes to store one date.
> data.table()[, x := strptime("2024-01-01","%Y-%m-%d")]
Warning message:
In strptime("2024-01-01", "%Y-%m-%d") :
  strptime() usage detected and wrapped with as.POSIXct(). This is to minimize the chance of assigning POSIXlt columns, which use 40+ bytes to store one date (versus 8 for POSIXct). Use as.POSIXct() (which will call strptime() as needed internally) to avoid this warning.
> class(strptime("2024-01-01","%Y-%m-%d"))
[1] "POSIXlt" "POSIXt" 
```
So I think what you are doing is great, you may look at what is done in `data.table()` to make sure there is consistent messages/code.

NEWS.md
```diff
@@ -70,6 +70,8 @@
 
 14. `fread(x, colClasses="POSIXct")` now also works for columns containing only NA values, [#6208](https://github.com/Rdatatable/data.table/issues/6208). Thanks to @markus-schaffer for the report, and Benjamin Schwendinger for the fix.
 
+15. `DT[, j := value]` now automatically wraps `j` with `as.POSIXct(value)`, [#1724](https://github.com/Rdatatable/data.table/issues/1724). Thanks to @linzhp for the report, and Benjamin Schwendinger for the fix.
```
ben-schwen, line 73:
```suggestion
15. For `POSIXlt` vectors, `DT[, j := value]` now automatically wraps `j` with `as.POSIXct(values)`, [#1724](https://github.com/Rdatatable/data.table/issues/1724). Thanks to @linzhp for the report, and Benjamin Schwendinger for the fix.
```

src/assign.c
```diff
@@ -428,6 +428,10 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
   }
   if (any_duplicated(cols,FALSE)) error(_("Can't assign to the same column twice in the same query (duplicates detected)."));
   if (!isNull(newcolnames) && !isString(newcolnames)) error(_("newcolnames is supplied but isn't a character vector"));
+  if (Rf_inherits(values, "POSIXlt")) {
+    warning(_("Values of type POSIXlt detected and converted to POSIXct. We do not recommend the use of POSIXlt at all because it uses 40+ bytes to store one date. Use as.POSIXct() to avoid this warning."));
```
tdhock, line 432: compare with warning from data.table()

POSIXlt column type detected and converted to POSIXct. We do not recommend use of POSIXlt at all because it uses 40 bytes to store one date.

Can the two warnings be made exactly the same? (easier for translators probably)

inst/tests/tests.Rraw
```diff
@@ -18748,3 +18748,11 @@ test(2270, options=c(datatable.optimize=1L), DT[, mean(b, 1), by=a], data.table(
...
 DT2 = data.table(a=c(1, 1, 2, 2), b=1:4)
 test(2271, options=c(datatable.verbose=TRUE), copy(DT1)[DT2, on='a', v := 4], copy(DT1)[, v := 4], output="x.a.\nAssigning")
+
+# convert POSIXlt to POSIXct in set and := #1724
+dt = data.table(a=1:2)
+dates = as.POSIXlt(c("2024-01-01", "2024-01-02"), tz="")
+ans = data.table(a=1:2, dates=as.POSIXct(dates))
+test(2272.1, copy(dt)[, dates := dates], ans, warning="POSIXlt.*converted to POSIXct")
+test(2272.2, set(copy(dt), j="dates", value=dates), ans, warning="POSIXlt.*converted to POSIXct")
+test(2272.3, {dt1 = copy(dt); dt1$dates=dates; dt1}, ans, warning="POSIXlt.*converted to POSIXct")
```
tdhock, line 18758: what about a test like `DT[, new_var := strptime(something)]` ?
I would think that the code in this PR makes the existing code and warning (below) obsolete, what do you think?

strptime() usage detected and wrapped with as.POSIXct(). This is to minimize the chance of assigning POSIXlt columns, which use 40+ bytes to store one date (versus 8 for POSIXct). Use as.POSIXct() (which will call strptime() as needed internally) to avoid this warning.

src/assign.c
MichaelChirico: > (easier for translators probably)

Possibly, but this is thrown from C while the other is thrown from R, so currently they live in different domains (`R-data.table.pot` vs `data.table.pot`), so a given translator might not even know about the two different messages. I think there are two ways to align them:

 1. Make the R message thrown from a C wrapper like `posixlt_warning = function() .Call(Cposixlt_warning)`
 2. Call `gettext(., domain="data.table")` on the _exact_ string used in C.

The former adds a little clutter to our .so, while the latter is a bit fragile by depending on the exact wording. I'm not sure either is better, or if either is worth investing in in the first place.

inst/tests/tests.Rraw
ben-schwen: For the assignemnt part yes, but you do not have to assign in j, hence, we still need it for `DT[, strptime(something)]`

tdhock: do we need it? the strptime warning says "this is to minimize the chance of assigning POSIXlt columns" but that should no longer be possible after this PR, right?

NEWS.md
```diff
@@ -70,6 +70,8 @@
 
 14. `fread(x, colClasses="POSIXct")` now also works for columns containing only NA values, [#6208](https://github.com/Rdatatable/data.table/issues/6208). Thanks to @markus-schaffer for the report, and Benjamin Schwendinger for the fix.
 
+15. For `POSIXlt` vectors, `DT[, j := value]` now automatically wraps `j` with `as.POSIXct(values)`, [#1724](https://github.com/Rdatatable/data.table/issues/1724). Thanks to @linzhp for the report, and Benjamin Schwendinger for the fix.
```
MichaelChirico, line 73: this is likely to be a breaking change, right? e.g. tests assuming output with POSIXlt columns.

Also, the wording here looks off:

> automatically wraps `j` with `as.POSIXct(values)`

I think there's (1) collision of `j`, the LHS of the example with `j`, the argument to `[.data.table`; and (2) `value` vs. `values`. I'm also not sure we've described what happens precisely. WDYT about

> In `DT[, j := value]`, when `value` is class `POSIXlt`, we automatically coerce it to class `POSIXct` instead.

BTW I looked into the "40 bytes for a single date" thing and I think it's a lot worse? At least on my current system with R 4.1.3:

```r
format(object.size(as.POSIXlt(Sys.time())), 'B')
#  [1] "2344 bytes"
```

This took a shockingly long time to run so saving it here:

```r
DF = data.frame(t=Sys.time() + 0:1000000)
DF$tl = as.POSIXlt(DF$t)
DF$size_t = sapply(seq_len(nrow(DF)), \(i) as.double(object.size(head(DF$t, i))))
DF$size_tl = sapply(seq_len(nrow(DF)), \(i) as.double(object.size(head(DF$tl, i))))
summary(DF$size_tl/DF$size_t)
#    Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
#   6.345   6.500   6.500   6.500   6.500   7.797 

head(DF)
#                     t                  tl size_t size_tl
# 1 2024-07-18 17:15:33 2024-07-18 17:15:33    344    2344
# 2 2024-07-18 17:15:34 2024-07-18 17:15:34    352    2360
# 3 2024-07-18 17:15:35 2024-07-18 17:15:35    368    2464
# 4 2024-07-18 17:15:36 2024-07-18 17:15:36    368    2464
# 5 2024-07-18 17:15:37 2024-07-18 17:15:37    384    2640
# 6 2024-07-18 17:15:38 2024-07-18 17:15:38    384    2640
```

So maybe more useful to say "typically takes more than 6 times the storage as an equivalent `POSIXct` column"?

inst/tests/tests.Rraw
ben-schwen: True that.

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 # data.table [v1.15.99](https://github.com/Rdatatable/data.table/milestone/30)  (in development)
 
+## POTENTIALLY BREAKING CHANGES
+
+1. In `DT[, j := value]`, when value is class `POSIXlt`, we automatically coerce it to class `POSIXct` instead, [#1724](https://github.com/Rdatatable/data.table/issues/1724). Thanks to @linzhp for the report, and Benjamin Schwendinger for the fix.
```
tdhock, line 7: this PR looks great overall, but I would change `DT[, j := value]` to `DT[, variable := value]` to avoid confusion (j is both the variable name and the argument name)
```r
> data.table()[j= variable := 5][]
   variable
      <num>
1:        5
```

ben-schwen: good point!

src/assign.c
ben-schwen: Not sure the effort to align them is worth, if it is not ensured that both are translated by the same translator (or by the same translation string).

MichaelChirico:
Marking this for 1.17.0 -- I suspect there can be downstream breakages as a result. Better to defer till after release.

MichaelChirico (approved):
Let's see what revdeps find!

tdhock:
revdeps checked today and did not find anything, https://rcdata.nau.edu/genomic-ml/data.table-revdeps/analyze/2024-08-28/

iago-pssjd:
Does it also close https://github.com/Rdatatable/data.table/issues/4800?

ben-schwen:
@iago-pssjd Nope since this only covers operations on `data.table` and not the `setDT` conversion

PR #6306, opened 2024-07-21
Ensure we're not UNPROTECT()ing 'x' in gsumm

MichaelChirico:
`rchk` is still throwing issues around here:

```
Function gmean
  [UP] allocating function gather may destroy its unprotected argument (x <arg 1>), which is later used. /rchk/packages/build/CaNpi0PW/data.table/src/gsumm.c:600
  [UP] calling allocating function gather with a fresh pointer (x <arg 1>) /rchk/packages/build/CaNpi0PW/data.table/src/gsumm.c:600
  [UP] unprotected variable x while calling allocating function Rf_allocVector /rchk/packages/build/CaNpi0PW/data.table/src/gsumm.c:601
```

The only thing I can think of is that the `UNPROTECT()` is applying to `x`, not `as`:

> The protection mechanism is stack-based, so UNPROTECT(n) unprotects the last n objects which were protected. 

IINM that means we have to keep `as` protected as long as `x` is? Or we have to clutter the code to do `UNPROTECT(2)` (`x`, `as`), followed immediately by `PROTECT(x)` to get it back under protection? Not even sure that will work, and probably not worth the clutter.


ben-schwen:
> IINM that means we have to keep `as` protected as long as `x` is? Or we have to clutter the code to do `UNPROTECT(2)` (`x`, `as`), followed immediately by `PROTECT(x)` to get it back under protection? Not even sure that will work, and probably not worth the clutter.

That should work and apparently some packages are doing it that way.

https://github.com/search?q=org%3Acran+%2FUNPROTECT%5C%28%5B2-9%5D%2B%5C%29%3B%5Cs*PROTECT%5C%28%5Cw%2B%5C%29%3B%2F&type=code

MichaelChirico:
> That should work and apparently some packages are doing it that way.

Nice find! Great that GitHub regex search works across lines. Applied here now, with comment.

PR #6025, opened 2024-03-27
add week_start parameter to round.IDate

MichaelChirico:
add week_start parameter to round.IDate

'week_start' argument to round.IDate

PR #6310, opened 2024-07-25
Added Support for Locale-Specific Labels in `month()` and `wday()` Functions

### LINKED ISSUE #1286: text labels for months / weekdays
At this moment using `month(x)` stores the months as an integer. However, when I want to get the names of the months, I have to fall back to `months` from base R. An example borrowed from [this answer of me](http://stackoverflow.com/a/31927948/2204410) on StackOverflow.

The following code:

```
setDT(temp)[, lapply(.SD, mean), by=.(year(date), month(date))]
```

gives me:

```
   year month amount
1: 2011     2 42.610
2: 2011     3 23.195
3: 2011     4 61.891
```

In order to get the monthnames, I need:

```
setDT(temp)[,date:=as.IDate(date)][, lapply(.SD, mean), by=.(year(date), months(date))]
```

which gives:

```
   year   months amount
1: 2011 februari 42.610
2: 2011    maart 23.195
3: 2011    april 61.891
```

Wouldn't it be nice to get that same result with someting like this:

```
setDT(temp)[, lapply(.SD, mean), by=.(year(date), month(date, label=TRUE))]
```

In such an implementation the default behavior should off course be `label = FALSE` to preserve the speed of working with integers.

The same applies to the names of the weekdays.

Possibly related to [#750](https://github.com/Rdatatable/data.table/issues/750)

Nj221102:
closes #1286 

**Description:**

This PR introduces support for locale-specific labels in the `month()` and `wday()` functions. The `label` argument has been enhanced to accept `TRUE`, which returns month and weekday names in the regional language according to the user's current locale settings. This improvement aims to enhance the readability of date-related data by reflecting regional language preferences.

**Changes:**

- Updated `month()` function to support locale-specific labels when `label = TRUE`.
- Updated `wday()` function to support locale-specific labels when `label = TRUE`.

**Example Usage:**

```R
x <- c("2024-07-01", "2024-07-02", "2024-07-03")
month(x, label = TRUE)
# [1] "July" "July" "July"

Sys.setlocale("LC_TIME", "fr_FR.UTF-8")  # French (France)
# [1] "fr_FR.UTF-8"

wday(x, label = TRUE)
# [1] "Lundi"    "Mardi"    "Mercredi"
```


man/IDateTime.Rd
```diff
@@ -113,6 +113,7 @@ yearqtr(x)
   the S3 generic.}
   \item{units}{one of the units listed for truncating. May be abbreviated.}
   \item{ms}{ For \code{as.ITime} methods, what should be done with sub-second fractions of input? Valid values are \code{'truncate'} (floor), \code{'nearest'} (round), and \code{'ceil'} (ceiling). See Details. }
+  \item{label}{ for \code{wday} and \code{month} when set to true will return day and month names instead of integers, names will be in the regional language according to locale setting.}
```
Anirban166, line 116: 'for \code{wday} and \code{month} when set to true will return day and month names instead of integers' - Needs rephrasing (also, 'names will be in the regional language according to locale setting.' should be a separate sentence)

NEWS.md
```diff
@@ -44,6 +44,17 @@
 
 `rbindlist(l, ignore.attr=TRUE)` and `rbind` also gains argument `ignore.attr` to manually deactivate the safety-net of binding columns with different column classes, [#3911](https://github.com/Rdatatable/data.table/issues/3911), [#5542](https://github.com/Rdatatable/data.table/issues/5542). Thanks to @dcaseykc, @fox34, @adrian-quintario, @berg-michael, @arunsrinivasan, @statquant, @pkress, @jrausch12, @therosko, @OfekShilon, @iMissile, @tdhock for the request and @ben-schwen for the PR.
 
+16. `month()` and `wday()` functions now support locale-specific labels. By setting the `label` argument to `TRUE`, you can obtain month and weekday names which will be in regional language according to your locale settings, [#1286](https://github.com/Rdatatable/data.table/issues/1286). Thanks @jaapwalhout for the request and @Nj221102  for the PR.
```
Anirban166, line 47: Extra space before 'for the PR.'

man/IDateTime.Rd
Nj221102: change it to " \item{label}{ For \code{wday} and \code{month}, setting \code{label} to \code{TRUE} will return day and month names instead of integers. Names will be in the regional language according to the locale setting. }" is this good enough ?

MichaelChirico:
In general I'm not a big fan of making it easy to generate non-reproducible code -- now your analysis might depend on whose computer is running a certain script. If we want to allow this feature, we need to make this drawback crystal clear.

In fact I would prefer behavior like this:

 - `label=FALSE` : as previously
 - `label=TRUE` : use English locale always
 - `label="native"` : use locale-specific label. Open to alternative naming here.

The point is that if you are _going_ to introduce non-reproducible objects to your code, _any_ reader of the script should be able to tell as clearly as possible. `label=TRUE/FALSE` is not clear enough IMO.

I also really like @franknarf1's suggestion for the output here to be a factor, perhaps optionally, but I lean towards "factor with no option to return string".

ben-schwen:
> In fact I would prefer behavior like this:
> 
> * `label=FALSE` : as previously
> * `label=TRUE` : use English locale always
> * `label="native"` : use locale-specific label. Open to alternative naming here.

I second Michaels suggestion for the different label options.

R/IDateTime.R
```diff
@@ -356,7 +363,14 @@ isoweek = function(x) {
   1L + (nearest_thurs - year_start) %/% 7L
 }
 
-month   = function(x) convertDate(as.IDate(x), "month")
+month <- function(x, label = FALSE) {
+    dates <- as.IDate(x)
+    if (isTRUE(label)) {
+        return(format(dates, "%B"))
+    }
```
ben-schwen, line 370: Do we need to convert to `IDate` for using `format`?

R/IDateTime.R
```diff
@@ -356,7 +363,14 @@ isoweek = function(x) {
   1L + (nearest_thurs - year_start) %/% 7L
 }
 
-month   = function(x) convertDate(as.IDate(x), "month")
+month <- function(x, label = FALSE) {
```
ben-schwen, line 366: `data.table` coding style uses `=` instead of `<-`

tdhock:
I liked the factor idea because then we could do as.integer() to test the result (levels would be locale dependent but integer codes would be independent)
but you are right that it is confusing to sometimes return factor and other times return character, and I would rather avoid that inconsistency.

R/IDateTime.R
MichaelChirico: All of this family of functions use `as.IDate()` already, so need to note anything new here IMO.

MichaelChirico:
For now I went with changing the argument name to `local_name`. That leaves only the possibility of "integer day-of-{week,month}" and "locale-specific name of {week,month}", getting english-specific names always is not supported.

 - Plus side: clear argument name, simple implementation. Not clear there's an actual user request to get english-name-always.
 - Down side: not easy to extend to supporting english-name-always, route would be a 3rd argument like `english_name=TRUE/FALSE`.

MichaelChirico:
@tdhock WDYT?

man/IDateTime.Rd
```diff
@@ -88,11 +88,11 @@ second(x)
...
-wday(x)
+wday(x, local_name = FALSE)
 mday(x)
 week(x)
 isoweek(x)
-month(x)
+month(x, local_name = FALSE)
 quarter(x)
 year(x)
 yearmon(x)
```
tdhock, line 98: I have never used these helper functions, because I think strftime is sufficient. (do we need to add this functionality? it seems like it would less maintenance for us, if we just encouraged strftime/as.Date with format codes, which seems more standard anyway)

Also it seems strange to me that local_name can only be provided for month and wday, and not the others (minute, hour, etc).
what is the justification? because https://github.com/Rdatatable/data.table/issues/1286 only asked for those two?

inst/tests/tests.Rraw
```diff
@@ -18851,3 +18851,313 @@ t = "2024-07-01"
 test(2275.1, month(t, local_name=TRUE), format(as.Date(t, "%B")))
 test(2275.2, wday(t, local_name=TRUE), format(as.Date(t, "%A")))
 
+# lazy forder, #4386
+dd = data.table(a=1:2, b=2:1)
+d = copy(dd)
+test(2275.01, options=c(datatable.verbose=TRUE), forderv(d, "b"), 2:1, output="forder.*opt=-1.*took")
```
tdhock: this test is failing, did I not merge correctly?

man/IDateTime.Rd
MichaelChirico: Yea, I'm also not totally convinced. The advantage of these functions is they return an integer-valued answer, whereas `strftime()` is returning string. Re-reading the issue+comments, I see two reasons to prefer supporting this:

 1. These flexibly take string or date as input -- no need to check if `as.IDate()` is required first
 2. We could make life a lot easier by generating an factor [ref](https://github.com/Rdatatable/data.table/issues/1286#issuecomment-309570372). It does feel a little out-of-scope, though -- months is easy enough, but whether to start the week on Sunday or Monday is also locale-dependent. #6025 gets at the same issue -- do we really want a 3rd parameter here for a simple function, though?

WDYT @jangorecki, maybe we just close this FR/PR?

jangorecki:
I would close. Issue is there for 9 years and 0 upvotes. We can always reopen if it will suddenly be needed.

MichaelChirico:
OK. Thanks @Nj221102 for your work here & for pushing this to resolution!

PR #6311, opened 2024-07-26
More careful PROTECT()/UNPROTECT() in rbindlist for rchk

MichaelChirico:
I'm actually still not convinced this is actually meaningful, but the flood of `rchk` warnings (#6263) _does_ diminish (though not totally gone) with this change.

I'm not a fan of the redundancy, but I didn't see a way around it:

 - We can't move the `if (ret)` branch outside the `if (listprotect)` branch because `ret` will have the wrong scope. [1](https://github.com/Rdatatable/data.table/pull/6263/commits/ce8c256002ebf894b2c928ca99ab367966121ea5)
 - We also can't pre-declare `ret` like `const char *ret; ret=...` because that's non-`const` behavior.
 - simple `if (listprotect) PROTECT(...); ...; if (listprotect) UNPROTECT();` doesn't satisfy rchk [2](https://github.com/Rdatatable/data.table/actions/runs/10105791205/job/27946846780), i.e. the problem is _not_ just to do with how the `if` condition is constructed & `listprotect` is written

Open to other suggestions...


MichaelChirico:
PTAL @ben-schwen -- we can move to turn on the `rchk` GHA after this

src/rbindlist.c
```diff
@@ -244,9 +244,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
     ncol = length(VECTOR_ELT(l, first));  // ncol was increased as if fill=true, so reduce it back given fill=false (fill==false checked above)
   }
 
-  int nprotect = 0;
```
ben-schwen, line 247: I don't get the rationale behind this. IMO there should be no different between `UNPROTECT(nprotect); UNPROTECT(2)` in the end or `nprotect += 2; UNPROTECT(nprotect);`

src/rbindlist.c
```diff
@@ -534,22 +534,26 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
...
-          if (ret) warning(_("Column %d of item %d: %s"), w+1, i+1, ret);
           // e.g. when precision is lost like assigning 3.4 to integer64; test 2007.2
           // TODO: but maxType should handle that and this should never warn
         }
         ansloc += thisnrow;
       }
     }
   }
-  UNPROTECT(nprotect);  // ans, coercedForFactor, thisCol
+  UNPROTECT(nprotect); // coercedForFactor?
```
MichaelChirico, line 556:
```suggestion
  UNPROTECT(2 + nprotect); // ans, ansNames, coercedForFactor?
```

src/rbindlist.c
```diff
@@ -534,22 +534,26 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
...
+  UNPROTECT(2); // ans, ansNames
```
MichaelChirico, line 557:
```suggestion
  UNPROTECT(2 + nprotect); // ans, ansNames, coercedForFactor?
```

src/rbindlist.c
MichaelChirico: I like the clarity of a literal `2` -- I am worried that by always using the simple `UNPROTECT(1); nprotect++;` "trick" it makes it easier to obscure poor design where 100s of things wind up protected (#6258). My thinking is it's easier to catch this when we are forced to stare at the code a little more before deciding what to write in `UNPROTECT()`. I moved back to `2+nprotect`.

As to why `UNPROTECT(2); UNPROTECT(nprotect);` instead of `UNPROTECT(2+nprotect);`, well, that was just me trying to explore why rchk is giving us the current "error", turns out it's _not_ anything to do with mixing literal+variable there.

ben-schwen: Ok but then why not initialize with `nprotect = 2` and `UNPROTECT(nprotect)` in the end?

ben-schwen (approved):
AFAIU #6263 our pattern with `listprotect`, `PROTECT` and `UNPROTECT` inside the if cannot be checked with rchk.

However, we need this pattern because of the protection stack size(see #5448).

WDYT should we try to use a PROTECT_WITH_INDEX instead?

src/rbindlist.c
MichaelChirico: Oh, great idea.

MichaelChirico:
> WDYT should we try to use a PROTECT_WITH_INDEX instead?

It sounds good in principle, but I'm not seeing a straightforward way to do so in practice -- `thisCol` is first assigned "lazily" in the RHS of `||`, so we'd have to break up that statement to add further nesting to accomplish the `PROTECT_WITH_INDEX` approach IINM. Anyway `thisCol` only needs to be `PROTECT()`'d in the rare case of this weird behavior for list input -- usually it's just column `l[[i]][[w]]` from the input, so, already under protection.

I'd rather defer this to a follow-up where we try and simplify the code a bit more while we're at it.

ben-schwen:
Seems ok for me but in the end we can cut this PR down to?

```c
bool listprotect = (TYPEOF(target)==VECSXP || TYPEOF(target)==EXPRSXP) && TYPEOF(thisCol)!=TYPEOF(target);
// do an as.list() on the atomic column; #3528
if (listprotect) {
  thisCol = PROTECT(coerceVector(thisCol, TYPEOF(target)));
```

MichaelChirico:
No, rchk spews a lot more unless we do the repetitive thing done now :\

ben-schwen:
Well, then we are good to go!

PR #6312, opened 2024-07-26
Use STRING_PTR_RO, not STRING_PTR

MichaelChirico:
Part of #6180, thanks to recommendation in R-exts

```texi
@item STRING_PTR 
@itemx DATAPTR       
@itemx STDVEC_DATAPTR
       Use @code{STRING_PTR_RO} and @code{DATAPTR_RO}. Obtaining
        writable pointers to these data can violate the memory manager's
        integrity assumptions and is not supported.
```

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -799,7 +799,7 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
       if (sourceIsFactor) { sourceLevels=PROTECT(getAttrib(source, R_LevelsSymbol)); protecti++; }
       if (!sourceIsFactor || !R_compute_identical(sourceLevels, targetLevels, 0)) {  // !sourceIsFactor for test 2115.6
         const int nTargetLevels=length(targetLevels), nSourceLevels=length(sourceLevels);
-        const SEXP *targetLevelsD=STRING_PTR(targetLevels), *sourceLevelsD=STRING_PTR(sourceLevels);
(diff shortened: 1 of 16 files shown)


MichaelChirico:
Actually this was not available in R 3.3.0, so we'll just use `STRING_PTR()` on really old R:

https://github.com/wch/r-source/blob/tags/R-3-3-0/src/include/Rinternals.h

It first got included in R 3.5.0:

https://github.com/r-devel/r-svn/commit/cce33583750e631f772d687f9dd919c8169211ab
https://github.com/wch/r-source/blob/tags/R-3-5-0/src/include/Rinternals.h

HughParsonage (approved):

HughParsonage:
> Actually this was not available in R 3.3.0, so we'll just use `STRING_PTR()` on really old R:
> 
> https://github.com/wch/r-source/blob/tags/R-3-3-0/src/include/Rinternals.h
> 
> It first got included in R 3.5.0:
> 
> [r-devel/r-svn@cce3358](https://github.com/r-devel/r-svn/commit/cce33583750e631f772d687f9dd919c8169211ab) https://github.com/wch/r-source/blob/tags/R-3-5-0/src/include/Rinternals.h

I think this is fair. The objection to using this non-API call is because it would limit _future_ changes to R, not because it was hazardous to use in past versions.

PR #6313, opened 2024-07-26
Remove SET_TYPEOF() call

MichaelChirico:
Part of #6180 thanks to the helpful suggestion in R-exts:

```texi
An alternative way to construct the expression that will work in
any @R{} version is

@example
SEXP expr = LCONS(R_NilValue, allocList(2));
@end example
```


ben-schwen:
LGTM

PR #6316, opened 2024-07-27
Setdroplevels for in-place removal of unused levels

### LINKED ISSUE #6014: Add setdroplevels() to replace droplevels(in.place=TRUE)
Let's add setdroplevels then. We could have dot prefixed non exported function that had copy art and then exported ones that goes via this .droplevels

_Originally posted by @jangorecki in https://github.com/Rdatatable/data.table/issues/5907#issuecomment-2006261616_

We'll need to go through a deprecation cycle for `droplevels(in.place=TRUE)`, but we should return to the principle that _only_ `set*` functions and `:=` can do in-place modification.

Luckily, I don't see _any_ usage of this on GitHub, so we don't need to worry too much about breaking downstreams:

https://github.com/search?q=lang%3AR+%2Fdroplevels%5B%28%5D%2F+%2Fin%5B.%5Dplace%5Cs*%3D%2F+-path%3Afdroplevels.R&type=code

Therefore I recommend we start the deprecation cycle in 1.16.0.

MichaelChirico:
Closes #6014

Tests feel a bit sparse, but I think we're basically covered by existing `droplevels.data.table()` tests, didn't know how much time to spend on redundant testing.

Given that `setdroplevels()` is a pretty easy drop-in replacement for `droplevels(in.place=TRUE)`, and we didn't find any downstream usage, I didn't add an opt-out mechanism for this, LMYK if you think we should be more cautious.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -202,5 +202,5 @@ S3method(format_col, expression)
 export(format_list_item)
 S3method(format_list_item, default)
 
-export(fdroplevels)
+export(fdroplevels, setdroplevels)
 S3method(droplevels, data.table)
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 # data.table [v1.15.99](https://github.com/Rdatatable/data.table/milestone/30)  (in development)
 
+## BREAKING CHANGES
+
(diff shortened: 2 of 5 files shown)


ben-schwen:
LGTM, but would like to hear @jangorecki opinion on the depcrecation part.

jangorecki (approved):
Lgtm. Warning is fine.

PR #6317, opened 2024-07-28
make my_subset function example replicate subset

### LINKED ISSUE #6315: my_subset() does not actually replicate subset()
https://github.com/Rdatatable/data.table/blob/822631e985cfcbd29ddae9e63bb75b30c42c5816/vignettes/datatable-programming.Rmd#L69

subset() also drops NA values, so this helper should to. I think it will emphasize the point that writing with [[ can be very hard to read

TysonStanley:
Closes #6315

### DIFF
--- a/vignettes/datatable-programming.Rmd
+++ b/vignettes/datatable-programming.Rmd
@@ -62,16 +62,16 @@ There are multiple ways to work around this problem.
 
 #### Avoid *lazy evaluation*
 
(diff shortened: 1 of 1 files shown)


MichaelChirico (approved):
Thanks!

jangorecki (commented):
I don't really agree on this change. This is vignette aims for simplicity, and data is iris, where is no NA. This vignette is not meant to cover data.table way of subset on any other data than iris.

TysonStanley:
I understand both sides to this. Think if we are showing that this approach can get difficult to read quickly, then I like the addition. If it is just to show a simple case that is quite legible, then can revert.

PR #6324, opened 2024-07-29
PROTECT() more in forder+bmerge

MichaelChirico:
Follow-up to #4386 given the `rchk` results:

https://github.com/Rdatatable/data.table/actions/runs/10148033664/job/28059822291?pr=6263

```
Function bmerge
  [UP] calling allocating function forderReuseSorting(?,?,?,?,?,V,?,?) with argument allocated using Rf_ScalarInteger /rchk/packages/build/a4VtYr2D/data.table/src/bmerge.c:165

Function idxName
  [UP] calling allocating function subsetVector with a fresh pointer (dt_names <arg 1>) /rchk/packages/build/a4VtYr2D/data.table/src/forder.c:1546
```

for `idxName`, we might just keep `dt_names` protected and do `UNPROTECT(6)` for simplicity.

### DIFF
--- a/src/bmerge.c
+++ b/src/bmerge.c
@@ -162,7 +162,11 @@ SEXP bmerge(SEXP idt, SEXP xdt, SEXP icolsArg, SEXP xcolsArg, SEXP xoArg, SEXP r
   allGrp1[0] = TRUE;
   protecti += 2;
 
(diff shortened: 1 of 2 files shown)


jangorecki (approved):
Not completely understand why we don't increment on each protect but if it passes check then I think it is safe to merge

MichaelChirico:
> Not completely understand why we don't increment on each protect

That also works, my thinking is to be more aggressive releasing memory as soon as it's no longer needed within each routine. But probably not much difference to just keep those objects under `PROTECT()` until return.

PR #6331, opened 2024-07-31
Don't include po/ directory in bundled package

MichaelChirico:
I don't know why R doesn't do this by default, but certainly we don't need to include these files -- certainly `R CMD check` doesn't care.


tdhock (approved):
great thanks
maybe ask about this on R-devel or R-pkg-devel?

MichaelChirico:
I thought of it, but I think the reason is -- R doesn't want to/can't assume every package has its source tracked on GitHub/openly online. So including po/ by default is the easiest way to give end users access to the .po/.pot files of _every_ CRAN package.

Put another way, if I put my package on CRAN, but keep its sources private, end users can get .mo files but it's quite difficult for them to offer me new translations (basically they'd have to contact me individually and we share .pot/.po privately).

Therefore, while I think it's indeed best practice for packages to exclude it, I doubt R-core will consider this idea.

PR #6332, opened 2024-07-31
Use proper language name in R-pt_BR.po

rffontenelle:
While it shows "English" in the source message, the translation should say "português" (Portuguese) instead of "inglês" (English).

See https://github.com/Rdatatable/data.table/pull/6199#issuecomment-2261491327


MichaelChirico (approved):
thanks!

PR #6336, opened 2024-08-01
Improved Error Message for Assigning NULL to List Column Items

### LINKED ISSUE #5526: error with set() or := NULL to a list column item
```R
DT = data.table(A=1:3, B=list())
DT
      A      B
   <int> <list>
1:     1       
2:     2       
3:     3       
DT[2, B:=letters]     # ok
DT
       A               B
   <int>          <list>
1:     1                
2:     2 a,b,c,d,e,f,...      # ok
3:     3                
DT[2, B:=NULL]
Error in `[.data.table`(DT, 2, `:=`(B, NULL)) : 
  When deleting columns, i should not be provided
set(DT, 2L, "B", NULL)
Error in set(DT, 2L, "B", NULL) : 
  When deleting columns, i should not be provided

# Expected result: NULL in that list column item
DT
      A      B
   <int> <list>
1:     1       
2:     2          
3:     3       
```

When fixed can simplify `tables()` dd2134e9aad6fd9d432b1cc0a35ca9c33fcd5dca #5524

Nj221102:
**Description:**

This PR improves the error which occurs when attempting to assign `NULL` to a list column item using `set()` or `:=`. which is not supported, The new message guides users to use `list(list(NULL))` instead  for row-specific assignments, enhancing clarity and preventing common mistakes.

**Changes:**

- Updated the error message in `assign.c` to provide a clearer instruction.
- Updated tests to cover the new error message.

**New Error Message:**

```
Direct assignment of NULL to a list column item is not allowed. To set a list column item to NULL, use list(list(NULL)) in your query.
```

**Issue Reference:**

This PR closes #5526


src/assign.c
```diff
@@ -458,7 +458,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
     coln--;
     SEXP thisvalue = RHS_list_of_columns ? VECTOR_ELT(values, i) : values;
     vlen = length(thisvalue);
-    if (isNull(thisvalue) && !isNull(rows)) error(_("When deleting columns, i should not be provided"));  // #1082, #3089
+    if (isNull(thisvalue) && !isNull(rows)) error(_("Direct assignment of NULL to a list column item is not supported. If you intend to assign NULL to an item in the list column, use list(list(NULL)) in your query."));  // #1082, #3089
```
joshhwuu, line 461: Not sure if this is correct - IIUC, this would error whenever `i` is provided and `RHS` is `NULL`, but the column is not necessarily a list column, you might have to generalize the error message here or add a nested `if` clause for when the specific column is a list column.

Nj221102: yeah updated error message to cover both cases 

"Direct assignment of NULL when `i` is provided is not supported. If you intend to assign NULL to a column, use `NULL` without `i`. For list columns, if you intend to assign NULL to a specific item, use `list(list(NULL))` in your query.

src/assign.c
```diff
@@ -458,7 +458,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
     coln--;
     SEXP thisvalue = RHS_list_of_columns ? VECTOR_ELT(values, i) : values;
     vlen = length(thisvalue);
-    if (isNull(thisvalue) && !isNull(rows)) error(_("When deleting columns, i should not be provided"));  // #1082, #3089
+    if (isNull(thisvalue) && !isNull(rows)) error(_("Direct assignment of NULL when `i` is provided is not supported. If you intend to assign NULL to a column, use `NULL` without `i`. For list columns, if you intend to assign NULL to a specific item, use `list(list(NULL))` in your query."));  // #1082, #3089
```
tdhock, line 461: please remove backticks in string literals.
only use backticks in markdown on github issues/prs (not in code in git repo)

Nj221102: done 👍

Nj221102: can you please checkout if any more changes are required for this PR.

MichaelChirico: Hi Nitish, sorry for the delay. From my perspective, we are basically in code freeze until 1.16.0 is completed+released. Appreciate your patience for the next week or two, thanks!

Nj221102: sure, thanks for info

src/assign.c
```diff
@@ -458,7 +458,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
     coln--;
     SEXP thisvalue = RHS_list_of_columns ? VECTOR_ELT(values, i) : values;
     vlen = length(thisvalue);
-    if (isNull(thisvalue) && !isNull(rows)) error(_("When deleting columns, i should not be provided"));  // #1082, #3089
+    if (isNull(thisvalue) && !isNull(rows)) error(_("Direct assignment of NULL when i is provided is not supported. If you intend to assign NULL to a column, use NULL without i. For list columns, if you intend to assign NULL to a specific item, use list(list(NULL)) in your query."));  // #1082, #3089
```
MichaelChirico, line 461: Not sure I'm a fan of the wording chosen here:

> Direct assignment of NULL when i is provided is not supported.

Well, it's never supported to assign `NULL` to anything but a list column -- regardless of whether `i` is provided.

> If you intend to assign NULL to a column, use NULL without i

Is that right? Assigning `NULL` without `i` just deletes the column? And again, this only applies to `list` columns.

joshhwuu: Would it be ideal if we errored differently depending on whether LHS is a list column or not?

Something like:

```R
DT = data.table(A=1:3, B=list())

DT[2, A:=NULL] # ERROR: When deleting columns, i should not be provided

DT[2, B:=NULL} # ERROR: When deleting columns, i should not be provided. If you are trying to assign NULL to a list column, please use list(list(NULL)) in your query instead.
```
Something like that

MichaelChirico: yes that's exactly what I'm thinking. I ran out of time to check earlier -- IINM at that point in the code we haven't checked if the current column exists or not yet.

e.g. in your example

DT[2, C := NULL]

joshhwuu: WDYT about:

```C
if (isNull(thisvalue) && !isNull(rows)) {
  if (TYPEOF(VECTOR_ELT(dt, coln)) == VECSXP) {
    error(_("When deleting columns, i should not be provided. If you are trying to assign NULL to a list column, please use list(list(NULL)) instead."));
  } else {
    error(_("When deleting columns, i should not be provided."));
  }
}
```

For the case where the column doesn't exist, I believe the error is the same as current master, `ERROR: When deleting columns, i should not be provided`.

MichaelChirico: We also need to be sure `coln < length(dt)`, but that's roughly the idea. PTAL.

joshhwuu: Looks good!!

MichaelChirico (approved):
Thanks all!

PR #6342, opened 2024-08-02
Add a job for linting md files

MichaelChirico:
This was used to write 268dd5fa7ce8bc2b4d776e8f8aebe40748ae52c9 (committed directly to `master` by mistake, but not worth reverting & putting through review IMO).

We can add more checks over time.


TysonStanley (approved):
LGTM, this is a nice helper

PR #6348, opened 2024-08-03
Take a pass at condensing and tightening the NEWS

MichaelChirico:
Part of #6166


ben-schwen:
@MichaelChirico I took the liberty to fix the numbering. If it was indeed intended to leave one number after a number addressing multiple issues please feel free to restore (note that in this case NOTES 6 `dcast` is still missing this numbering)

MichaelChirico:
Thanks! #6342 will prevent backsliding here going forward...

PR #6355, opened 2024-08-05
Spanish translations

rikivillalba:
Add "es.po" and "R-es.po" spanish translations
Relevant thread:
https://github.com/Rdatatable/data.table/pull/6199#issuecomment-2261825742

PR #6281, opened 2024-07-16
Update Chinese translations for release

MichaelChirico:
These are the updated .po files.

@Rdatatable/chinese PTAL and help to update the translations.

 1. Look for `fuzzy` markers of translations that probably need to be fine-tuned as the source string changed slightly (or a lot). I counted 18 of those.
 2. Look for completely new messages, I count 10 of those.

```shell
msgfmt --statistics po/zh_CN.po
# 803 translated messages, 12 fuzzy translations, 4 untranslated messages.
msgfmt --statistics po/R-zh_CN.po
# 465 translated messages, 6 fuzzy translations, 6 untranslated messages.
```

Thank you all as always!

PR #6358, opened 2024-08-05
Add a code-quality check for translations

MichaelChirico:
Following #6355, #6199, and #6281, it will be nice to perform the (minimal) checks I've been doing manually, automatically. The tests are very lightweight, so we start by just doing them on every PR.


tdhock:
great idea! Thanks.

PR #6362, opened 2024-08-08
remove check for unavailabe dependencies on ancient R

ben-schwen:
Current CI ancient job is failing because new version of `xts`, and `evaluate` depend on `R 4.0.0`. This PR patches the CI job by installing fixed versions for certain packages, afterwards.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -209,6 +209,9 @@ test-lin-ancient-cran:
   <<: *test-lin
   script:
     - *install-deps
+    - Rscript -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/xts/pkgname_version.tar.gz", repos = NULL, type = "source")'
+    - Rscript -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/evaluate/pkgname_version.tar.gz", repos = NULL, type = "source")'
+    - Rscript -e 'install.packages("knitr")'
     - R CMD check --no-manual $(ls -1t data.table_*.tar.gz | head -n 1)
 
 .test-win-template: &test-win
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -211,7 +211,7 @@ test-lin-ancient-cran:
     - *install-deps
     - Rscript -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/xts/xts_0.12.2.tar.gz", repos = NULL, type = "source")'
     - Rscript -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/evaluate/evaluate_0.23.tar.gz", repos = NULL, type = "source")'
-    - Rscript -e 'install.packages("knitr")'
+    - Rscript -e 'install.packages("knitr", repos = "https://cloud.r-project.org")'
     - R CMD check --no-manual $(ls -1t data.table_*.tar.gz | head -n 1)
 
 .test-win-template: &test-win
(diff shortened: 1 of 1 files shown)


MichaelChirico:
This feels a bit fragile to me. I wonder what the practical use case we're testing for is exactly:

 1. User on ancient R is also working with ancient {knitr}
 2. User on ancient R doesn't need to check vignettes, doesn't care about {knitr}

I've worked towards the latter in the R-CMD-check-occasional GHA, and taken the approach to just skip stuff related to these Suggests deps that won't install:

https://github.com/Rdatatable/data.table/blob/83fd2c05ce2d8555ceb8ba417833956b1b574f7e/.github/workflows/R-CMD-check-occasional.yaml#L144-L148

(I don't really have a sense one way or the other what's more likely)

.gitlab-ci.yml
```diff
@@ -205,11 +205,14 @@ test-lin-dev-clang-cran:
...
-  image: registry.gitlab.com/jangorecki/dockerfiles/r-3.3.0
   <<: *test-lin
+  image: registry.gitlab.com/jangorecki/dockerfiles/r-3.3.0
+  variables:
+    _R_CHECK_FORCE_SUGGESTS_: "FALSE"
   script:
     - *install-deps
-    - R CMD check --no-manual $(ls -1t data.table_*.tar.gz | head -n 1)
+    - Rscript -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/xts/xts_0.12.2.tar.gz", repos = NULL, type = "source")'
+    - R CMD check --no-manual --no-build-vignettes --ignore-vignettes $(ls -1t data.table_*.tar.gz | head -n 1)
```
MichaelChirico, line 215: Conceivably, we'll catch up to an R version where knitr can be installed again in the future, so leave a note here IMO.

```suggestion
    # knitr requires evaluate, which requires R 3.6.0.
    # Restore checking vignettes if upgrading our R dependency means knitr can be installed.
    - R CMD check --no-manual --no-build-vignettes --ignore-vignettes $(ls -1t data.table_*.tar.gz | head -n 1)
```

.gitlab-ci.yml
```diff
@@ -205,11 +205,14 @@ test-lin-dev-clang-cran:
...
 test-lin-ancient-cran:
-  image: registry.gitlab.com/jangorecki/dockerfiles/r-3.3.0
   <<: *test-lin
+  image: registry.gitlab.com/jangorecki/dockerfiles/r-3.3.0
+  variables:
+    _R_CHECK_FORCE_SUGGESTS_: "FALSE"
   script:
     - *install-deps
-    - R CMD check --no-manual $(ls -1t data.table_*.tar.gz | head -n 1)
+    - Rscript -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/xts/xts_0.12.2.tar.gz", repos = NULL, type = "source")'
```
MichaelChirico, line 214: What happens if we don't install xts? Shouldn't `check` pass?

ben-schwen: Yes, with the added `_R_CHECK_FORCE_SUGGESTS_: "FALSE"`, `check` should pass!

MichaelChirico (approved):
LGTM! May as well merge #6383 first, right? Since we know CI will fail on 3.3.0 at current `master`.

ben-schwen:
> LGTM! May as well merge #6383 first, right? Since we know CI will fail on 3.3.0 at current `master`.

Yes, CI will only deliver new insights after merging #6383

PR #6363, opened 2024-08-12
Remove logical condition recycling from fcase()

### LINKED ISSUE #6352: Consider reverting fcase recycled conditions
@DavisVaughan offered a suggestion that we reconsider recycling conditions in `fcase()`: vectorized default should be enough to solve the user issue:

https://fosstodon.org/@davis/112909151150543355

Earmarking this to think it through more carefully & for others' input.

MichaelChirico:
Closes #6352 as discussed.

Not a big fan of the need for `suppressWarnings()` in the R body; `...length()` is the way to avoid this, but that's not available till R3.5.0.


tdhock (approved):
looks good, nice NEWS item.
 you may want to ask @DavisVaughan for review too?

MichaelChirico:
I'm not sure he's reviewed code here before so it might take some effort to parse through it. I think the high-level feedback he gave leading to this was clear enough & implemented here. But Davis, do feel free to chime in!

Issue #6350, opened 2024-08-03
encoding-related test failures on Alpine Linux


bastistician:
(Only reporting now, seeing that **data.table** is being developed again.)

Checking released **data.table** 1.15.4, my Alpine Linux server gives
```
Error: 3 error(s) out of 11070. Search tests/tests.Rraw.bz2 for test number(s) 1590.05, 1590.06, 1997.14. Duration: 34.4s elapsed (34.9s cpu).
```
but at this point it is probably more useful to look at the development version of **data.table**.

So in a vanilla Alpine Linux container,

```sh
docker run --rm -it alpine
```

running

```sh
export TZ=UTC
apk add R R-dev R-doc
## get data.table (devel) and suggested packages
R -s -e 'install.packages("data.table", repos = "https://rdatatable.gitlab.io/data.table", dependencies = TRUE, destdir = "/tmp")'
export _R_CHECK_TESTS_NLINES_=0
R CMD check --extra-arch /tmp/data.table_*.tar.gz
```

gives only 2 failures for [test numbers 1590.05 and 1590.06](https://github.com/Rdatatable/data.table/blob/2dac309515d3d3ec9f645a050c425085082a224a/inst/tests/tests.Rraw#L8400-L8420):

<details><summary>Error in test.data.table()</summary>

```
* using R version 4.4.0 (2024-04-24)
* using platform: x86_64-pc-linux-musl
* R was compiled by
    gcc (Alpine 13.2.1_git20240309) 13.2.1 20240309
    GNU Fortran (Alpine 13.2.1_git20240309) 13.2.1 20240309
* running under: Alpine Linux v3.20
* using session charset: UTF-8
[...]
Running the tests in ‘tests/main.R’ failed.
Complete output:
  > require(data.table)
  Loading required package: data.table
  > 
  > test.data.table()  # runs the main test suite of 5,000+ tests in /inst/tests/tests.Rraw
  getDTthreads(verbose=TRUE):
    OpenMP version (_OPENMP)       201511
    omp_get_num_procs()            12
    R_DATATABLE_NUM_PROCS_PERCENT  unset (default 50)
    R_DATATABLE_NUM_THREADS        unset
    R_DATATABLE_THROTTLE           unset (default 1024)
    omp_get_thread_limit()         2147483647
    omp_get_max_threads()          12
    OMP_THREAD_LIMIT               unset
    OMP_NUM_THREADS                unset
    RestoreAfterFork               true
[...]

PR #6366, opened 2024-08-13
Workaround failing test on Alpine Linux

MichaelChirico:
Stop-gap for #6350 not to block CRAN release.

Idea to check if `/etc/alpine-release` exists taken from https://serverfault.com/questions/805389/which-os-is-running-in-my-docker-container#comment1290522_805390.

Here's some more references to that file:

 - https://endoflife.date/alpine
 - https://superuser.com/a/1163235/454554
 - https://github.com/search?q=%2F%5C%2Fetc%5C%2Falpine%5B-%5Drelease%2F&type=code

cc @bastistician

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -8416,8 +8416,11 @@ baseR = base::order(c(x2,x1,x1,x2))
   # are so relaxed now that they barely testing anything. It appears base R behaviour is undefined in this rare case of identical strings in different encodings.
 test(1590.04, identical(baseR, INT(1,4,2,3)) || identical(baseR, INT(2,3,1,4)) || identical(baseR, 1:4))
 Encoding(x2) = "unknown"
-test(1590.05, x1!=x2)
-test(1590.06, forderv(    c(x2,x1,x1,x2)), INT(1,4,2,3))  # consistent with Windows-1252 result, tested further below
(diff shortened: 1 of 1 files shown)


bastistician:
I'd suggest to fix this for these C-locale tests just like in the tests 1590.12/14 below:
if x1==x2 (as seen on Alpine Linux), then test that `forderv(c(x2,x1,x1,x2))` gives `integer()`, otherwise `INT(1,4,2,3)`.

tdhock:
I'm not sure I really understand the core issue this fixes, but I guess it is OK to skip a couple of tests on one platform, temporarily. I would suggest keeping the issue open and hopefully @bastistician can help suggest a longer term fix/PR?

MichaelChirico:
it really takes some heavy focus time for me to wrap my head around encoding issues, so I'd rather kick the can here/not hold up release waiting for a fix. leaving the issue open to revisit for 1.17.0

bastistician:
> hopefully @bastistician can help suggest a longer term fix/PR?

Sure, see my suggestion above, i.e., test the forderv() result depending on x1==x2.
Conditioning on Alpine Linux specifically will make my check system happy but won't help other musl-based distros (where I'd expect the same test failures).

MichaelChirico:
I agree it's better, but it's not a priority at the moment. That code has been in data.table since 1.11.4 (2018: https://github.com/Rdatatable/data.table/commit/2b4d73613), with yours the first user report.

PR #6368, opened 2024-08-16
add translation teams to CODEOWNERS

tdhock:
hi @Rdatatable/translators
I propose that we add lines to CODEOWNERS so that a translation team will be notified when there are PRs that modify the files related to their language.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -40,6 +40,7 @@
 /po/ @michaelchirico
 /R/translation.R @michaelchirico
 /src/po.h @michaelchirico
+/po/*.pot @Rdatatable/translators
 /po/*zh_CN.po @Rdatatable/chinese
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Fantastic, great idea!

CODEOWNERS
```diff
@@ -40,6 +40,9 @@
 /po/ @michaelchirico
 /R/translation.R @michaelchirico
 /src/po.h @michaelchirico
+/po/*zh_CN.po @Rdatatable/chinese
```
MichaelChirico, line 43: WDYT about this, to notify all translators whenever the templates are updated?

```suggestion
/po/*.pot @Rdatatable/translators
/po/*zh_CN.po @Rdatatable/chinese
```

tdhock: great idea

PR #6375, opened 2024-08-18
non-ASCII test robustness

### LINKED ISSUE #6339: Non-ascii symbols on systems without UTF-8
Test of #4711 does not work in systems without `UTF-8` encoding as e.g. our `test-lin-rel-vanilla` container.

Output of spinning up a new container with the image `registry.gitlab.com/jangorecki/dockerfiles/r-base-gcc`
```r
DT = data.table(a = rep(1:3, 2))
setnames(DT, "a", "a\U00F1o")
DT[ , .N, 'a<U+00F1>o']
#>    a<U+00F1>o     N
#>         <int> <int>
#> 1:          1     2
#> 2:          2     2
#> 3:          3     2
#> Warning message:
#> In eval(bysub, x, parent.frame()) :
#>   unable to translate 'a<U+00F1>o' to native encoding
DT[ , .N, a<U+00F1>o]
#> Error: unexpected symbol in "DT[ , .N, a<U+00F1"
```

MichaelChirico:
Closes #6339

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18724,10 +18724,10 @@ if (test_bit64) local({
 })
 
 # non-ASCII plain symbol in by, #4708
+# NB: recall we can't use non-ASCII symbols in the test script. The text is a-<n-tilde>-o (year in Spanish)
 native_ano = iconv("a\U00F1o", "UTF-8", "")
 if (!is.na(native_ano)) { # #6339: symbol must be represented in native encoding
   DT = data.table(a = rep(1:3, 2))
-  # NB: recall we can't use non-ASCII symbols here. the text is a-<n-tilde>-o (year in Spanish)
   setnames(DT, "a", native_ano)
   test(2266, eval(parse(text=sprintf("DT[ , .N, %s]$N[1L]", native_ano))), 2L)
   # sub-key can also be retained in plain query, part of #4498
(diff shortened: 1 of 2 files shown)


inst/tests/tests.Rraw
```diff
@@ -18724,16 +18724,21 @@ if (test_bit64) local({
...
+  setnames(DT, "a", native_ano)
+  test(2266, eval(parse(text=sprintf("DT[ , .N, %s]$N[1L]", native_ano))), 2L)
+  # sub-key can also be retained in plain query, part of #4498
+  DT = data.table(id = rep(1:10, 2L), grp = rep(1:2, each=10L), V = 1:20/13, key=c('id', 'grp'))
+  test(2266.1, key(DT[ , .(id)]), 'id')
+  test(2266.2, key(DT[ , .(grp)]), NULL)
+  ## renaming also caught
+  test(2266.3, key(DT[ , .(newid = id, newgrp = grp)]), c('newid', 'newgrp'))
+} else {
+  cat("Tests 2266* skipped. n-tilde cannot be represented in the native encoding on this system. Please report\n")
```
ben-schwen, line 18740: Maybe we should mention the bug tracker if we want to have the reports there. However, I'm not sure if we really should want reports about that?

Besides that, LGTM

ben-schwen (approved):

inst/tests/tests.Rraw
MichaelChirico: > However, I'm not sure if we really should want reports about that?

If there's any system where conversion fails, maybe the code is also not WAI there too? Or we just need to use a different non-ASCII character that's more universal? Hard to know. Doesn't hurt much to keep the call-out here.

PR #6378, opened 2024-08-19
use \n line endings

MichaelChirico:
Closes #6369

Approach:

```r
for (f in list.files(recursive = TRUE)) {
  ch <- tryCatch(readChar(f, file.size(f)), condition=identity)
  if (inherits(ch, "condition")) next
  nch <- tryCatch(nchar(ch), condition=identity)
  if (inherits(nch, "condition")) next
  if (grepl("\r", ch, fixed = TRUE)) {
    cat(sprintf("Overwriting \\r line endings in %s\n", f))
    writeLines(readLines(f), f)
  }
}
```

PR #6379, opened 2024-08-19
Fix new_linter & config so that it actually runs, and apply the changes

MichaelChirico:
Turns out this GHA was not actually doing anything since without the actions/checkout step, `list.files()` doesn't find anything & then the step finishes without doing anything.

Fixed that, it revealed some errors, which are now fixed.


tdhock:
great idea to add a linter to check for line endings, but shouldn't this fail currently? (since CODEOWNERS has DOS line endings?)
I see a green build (all checks have passed, including new lint-misc)

tdhock:
ok so this looks great thanks.
I love the NEWS linter idea!!
does the lint-misc have green/pass build now because #6378 was merged into master and this PR has already taken in those changes?

MichaelChirico:
> does the lint-misc have green/pass build now because https://github.com/Rdatatable/data.table/pull/6378 was merged into master and this PR has already taken in those changes?

No, something is still wrong. Locally, I still see issues with CODEOWNERS, and it should be catching issues in po/{R-es,es}.po. Not sure why, still debugging...

MichaelChirico:
This is deeply frustrating. I can't seem to get GHA to _stop_ removing CRLF itself, before our linter ever has a chance to inspect the files.

tdhock:
you may need to configure git itself to allow the line endings to propagate? https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf

MichaelChirico:
I have that set here:

https://github.com/Rdatatable/data.table/pull/6379/files#diff-4a2765c2cfcbd3804a66aab805cb92ddda74de1730923cc5bf53671d0beccf06R86

We also have this set in .gitattributes:

https://github.com/Rdatatable/data.table/blob/b785de67d05520fc5a332677e7934a9f2ea27ee7/.gitattributes#L1

Which as I understand it, `-text` means "not text" (i.e. binary), which is _supposed_ to mean git doesn't try and do anything for the line endings of that file.

So, no idea what GHA is doing.

MichaelChirico:
This PR is winding up to only touch the NEWS files+associated linter. Starting a separate PR w.r.t. line endings in a moment.

tdhock:
ok that is fine

PR #6381, opened 2024-08-19
Add a new devcontainer for ancient R

MichaelChirico:
To allow us to test code on our oldest R version interactively, more easily


ben-schwen:
Nice!

PR #6382, opened 2024-08-19
Use recommended R_ prefixed versions of Calloc,Realloc,Free

### LINKED ISSUE #6380: CRAN Strict Issue regarding headers
Received this from CRAN today:

```
Please see the problems shown on <https://cran.r-project.org/web/checks/check_results_data.table.html>. 
Specifically, please see the *Strict* additional issue.

Compilation fails with _R_USE_STRICT_R_HEADERS_=true, which defines
STRICT_R_HEADERS to 1 which removes

- the legacy definition of PI (use POSIX's M_PI, available in R fer ever).
- the RS.h declarations for Calloc, Realloc, Free (use R_ forms i
  available since R 3.4.0).

The aim is to clean the namespace: in particular having a definition
for Free has conflicted with some packages' C++ code.

It is planned that STRICT_R_HEADERS=1 will become the default for 4.5.0,
which in particular makes it necesssary that all CRAN packages with many
strong reverse dependencies compile/install ok with the new default.

Your package is among the ones with many strong reverse dependencies.
We would thus really appreciate if you could provide a new version of
your package as soon as possible which checks ok with
STRICT_R_HEADERS=1.

You can verify that your package checks ok with STRICT_R_HEADERS=1 via R
CMD check --as-cran using a current version of R-devel.

Please correct before 2024-09-20 to safely retain your package on CRAN.
```

I didn't have time to look closely yet but this will need to be included in the 1.16.0 release.

MichaelChirico:
Closes #6380

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -13178,7 +13178,7 @@ options(datatable.use.index=FALSE)
 test(1942.10, DT[,sum(v),keyby=id1,verbose=TRUE], data.table(id1=c("D","A","C"), V1=INT(1,6,8), key="id1"), output="Finding groups using forderv")
 options(datatable.use.index=TRUE)
 
-# test coverage in uniqlist.c of Realloc when over initial 1000 uniqs, and use double to cover numeric tolerance (both in one-column case and >1 column branch)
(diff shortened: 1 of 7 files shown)


MichaelChirico:
Re: removing the `!defined(R_VERSION)` guard:

https://bugs.r-project.org/show_bug.cgi?id=18783

TysonStanley:
LGTM! Quick response to CRAN 👏

PR #6383, opened 2024-08-20
Various fixes for R 3.3.0

MichaelChirico:
Besides the test changes for `.Last.updated`, AFAICT all of these changes are dev-only, hence no NEWS item.


R/utils.R
```diff
@@ -21,6 +21,15 @@ nan_is_na = function(x) {
   stopf("Argument 'nan' must be NA or NaN")
 }
 
+# In R 3.3.0, rep_len() strips attributes --> breaks data.table()'s internal recycle() helper.
+if (inherits(rep_len(Sys.Date(), 1L), "Date")) {
+  # NB: safe_rep_len=rep_len throws an R CMD check error because it _appears_ to the AST
+  #   walker that we've used .Internal ourselves (which is not true, but codetools can't tell).
```
tdhock, line 27: nice explanation

inst/tests/tests.Rraw
```diff
@@ -14787,49 +14787,54 @@ test(2029.2, fread(txt, quote=""), data.table(A=1:2, B=4:5, C=7:8), warning="Dis
...
-d[2:3, z:=7L]
-test(2030.03, .Last.updated, 2L) # sub assign
-d[integer(), z:=8L]
-test(2030.04, .Last.updated, 0L) # empty sub-assign
-d[-1L, z:=9L]
-test(2030.05, .Last.updated, 3L) # inverse sub-assign
-d[-(1:4), z:=10L]
-test(2030.06, .Last.updated, 0L) # inverse empty sub-assign
+# On R 3.3.0, .Last.updated gets touched by test() itself and this cascades even to re-assignments
+#   of .Last.updated like n_updated=.Last.updated, so really force a deep copy by round-tripping from string.
```
tdhock, line 14794: again thanks for the clear explanation

tdhock:
looks good thanks for the updates.
just wondering why we are only seeing them now?
is it because the R-3.3.0 CI is only run occasionally?

ben-schwen:
> looks good thanks for the updates. just wondering why we are only seeing them now? is it because the R-3.3.0 CI is only run occasionally?

Whole CI was broken until #6211 and the dependency error with `xts` is long standing and will only be fixed after #6362

MichaelChirico:
GHA R-cmd-check-occasional is also not working. I've seen dune

R/utils.R
```diff
@@ -21,6 +21,15 @@ nan_is_na = function(x) {
   stopf("Argument 'nan' must be NA or NaN")
 }
 
+# In R 3.3.0, rep_len() strips attributes --> breaks data.table()'s internal recycle() helper.
+if (inherits(rep_len(Sys.Date(), 1L), "Date")) {
```
ben-schwen, line 25: I just checked and this seems still a problem at `3.6.0` so we might need to hold on to this for quite some time, but it would be best to check at which version this is fixed in `R` and add a comment

MichaelChirico: Thanks for checking. I suspected that might be the case -- I checked the R NEWS and NEWS.3 files and didn't see anything relevant here (searching 'attrib', 'rep_len'), and glanced at the `rep_len()` blame to see if anything about attributes popped out, no luck.

MichaelChirico: Actually looking again, I think it's this:

https://github.com/r-devel/r-svn/commit/70e8a3fd55b8f4e5b87d5fbbe7256e6f59154961

Then `rep_len()` on `Date` dispatches to `rep.Date()`.

That puts us in R 4.0.0:

https://github.com/r-devel/r-svn/blob/874fa1008e6cb090c7c898c422d703ca0026d600/doc/NEWS.Rd#L4885-L4887

PR #6384, opened 2024-08-20
macro redefined in fwrite and fread

ben-schwen:
Currently both `freadR.h` and `fwrite.h` have the line `#define STRICT_R_HEADERS` (see also at [CI](https://gitlab.com/Rdatatable/data.table/-/jobs/7617683486#L145) but happens at test-lin-dev-clang-cran and test-lin-dev-gcc-strict-cran)

This PR adds a directive to only define once, avoiding redefinition.


MichaelChirico:
Seems harmless, thanks!

MichaelChirico (approved):

Issue #6369, opened 2024-08-16
CODEOWNERS has DOS line endings


tdhock:
But it looks like most other files have unix line endings, at least in my clone on ubuntu. Does anybody else notice that?
![image](https://github.com/user-attachments/assets/de0f7774-14b7-4f90-9891-32185966376f)
Notice above (DOS) in mode line bottom left CODEOWNERS buffer, but not other 3 buffers.

I can see the line ending in git diff:
```
(base) tdhock@tdhock-MacBook:~/R/data.table[CODEOWNERS-translation-teams*]$ git diff
diff --git a/CODEOWNERS b/CODEOWNERS
index 61277276..9a360821 100644
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -40,6 +40,9 @@
 /po/ @michaelchirico
 /R/translation.R @michaelchirico
 /src/po.h @michaelchirico
+po/*zh_CN.po @Rdatatable/chinese^M
+po/*pt_BR.po @Rdatatable/brazil^M
+po/*es.po @Rdatatable/spanish^M
 
 # printing
 /R/print.data.table.R @michaelchirico
```

rikivillalba:
Yes 🤷
![image](https://github.com/user-attachments/assets/17a665ff-229a-49bb-aa50-6735ada6ea70)

PR #6386, opened 2024-08-20
Update gitattributes to force LF

MichaelChirico:
Follow-up to #6378 and #6369

My understanding is this works as follows:

 1. Every file is treated as a plain text file, with LF line endings (i.e. `\n`)
 2. Except for certain binary file types: .ico, .mo, .png, .RData
 3. And except in the inst/tests directory, where files are binary (`-text`) by default
 4. Except for .Rraw files in inst/tests, which should also be treated as plain text LF


Anirban166 (approved):
LGTM in terms of the syntax! (can just keep adding to the list of binaries to be excluded from being treated as text files)

Issue #6242, opened 2024-07-11
GitLab open source program


ben-schwen:
We could use the already available beta runner for macos if we apply for the Gitlab open source program.

_Originally posted by @ben-schwen in https://github.com/Rdatatable/data.table/issues/3326#issuecomment-2200975001_
            
I just got the notification that we are accepted to the open source program.

To finish we need to do the following:
- [ ] estimate the number of users for our GitLab repo (currently we are Michael, Jan, Matt and me), but I would add at least Toby.
- [ ] set my role to owner at the GitLab repo @jangorecki

We also need to renew the application every year, so we might want to introduce a role for that. For now, I'm happy to handle it.

MichaelChirico:
Great news!

tdhock:
cool thanks for the invite 
I almost never go on gitlab but I still have access to user id tdhock

jangorecki:
By the end of the July I can get into gitlab and set an owner so this issue has to wait a bit.

tdhock:
related: Aren't dev binaries built on gitlab? 
Is there any documentation for that anywhere? can people who know about how that works please write docs on this wiki page https://github.com/Rdatatable/data.table/wiki/Binary-managers 
Also I created a team https://github.com/orgs/Rdatatable/teams/binary-managers and I recommend adding people who know about that to this team for easy mentioning `@Rdatatable/binary-managers`

ben-schwen:
> related: Aren't dev binaries built on gitlab? Is there any documentation for that anywhere? can people who know about how that works please write docs on this wiki page https://github.com/Rdatatable/data.table/wiki/Binary-managers Also I created a team https://github.com/orgs/Rdatatable/teams/binary-managers and I recommend adding people who know about that to this team for easy mentioning `@Rdatatable/binary-managers`

Yes. AFAIK the only documentation is the code in `.gitlab-ci.yml` and our issues here. Basially, the binaries get built before, copied in the integration step and published in the publish step. Afterwards they can be downloaded with the appropriate filepath e.g. for [4.2](https://rdatatable.gitlab.io/data.table/bin/windows/contrib/4.2/data.table_1.15.99.zip)

jangorecki:
There is documentation inside git repo, not github (wiki)

jangorecki:
your role set to owner

ben-schwen:
Regarding the number of members @ Gitlab: currently we would be 5 including Toby, lets set it to 10?

[...]

PR #6389, opened 2024-08-22
add macosx runner

ben-schwen:
Since we are now in the Gitlab open source program #6242 we can also activate the macos runner.


.gitlab-ci.yml
```diff
@@ -275,7 +275,7 @@ test-win-old:
 
 ## R-release on MacOS
 # no macosx runner set yet
```
ben-schwen, line 277:
```suggestion

```

ben-schwen (commented):
I would only add it to need once we see it runs through with CI

`needs: ["mirror-packages","build","test-lin-rel","test-lin-rel-cran","test-lin-dev-gcc-strict-cran","test-lin-dev-clang-cran","test-lin-rel-vanilla","test-lin-310-cran","test-win-rel","test-win-dev" ,"test-win-old", "test-mac-rel"]`

MichaelChirico (approved):
nice!

PR #6393, opened 2024-08-23
Fix fwrite length for gzip output

### LINKED ISSUE #5506: Set compression level for gzipped output in fwrite

### LINKED ISSUE #6356: fwrite with compress="gzip" produces gz files with incorrect uncompressed file sizes
data.table::fwrite with compress="gzip" produces slightly incompatible gz files with multiple independent chunks.

It means that browsers cannot receive the compressed .gz files using `Content-Encoding: gzip` and `Transfer-Encoding: chunked` because a browser only processes the first chunk of a gzipped file, in this case the csv header chunk.

My code currently looks like this:

```R
      filename <- "c:/test.csv.gz"
      # data.table::fwrite(df, file=filename, row.names=FALSE, na="", compress="gzip")
      # ^ inbuilt data.table compression gives incorrect uncompressed size
      # need to intentionally use R.utils for gzip
      interim <- paste(filename, "_raw", sep="")
      data.table::fwrite(df, file=interim, row.names=FALSE, na="", compress="none")
      R.utils::gzip(interim, destname=filename)
```
I'm running it on Windows 10, with Rscript (R) version 4.4.1 (2024-06-14) and the latest release of data.table version 1.15.2.

Please let me know if there's anything else you need from me.

philippechataignon:
Closes #6356. Closes #5506.

This PR is an attempt to create a better gzip file with fwrite. Its an important rewrite because it includes some refactoring of actual code.

## zlib

* use Z_SYNC_FLUSH    instead of Z_FINISH
* create manual heading and write crc and len in tail
* calc len and crc in thread and summarize in main thread
* len in gzip specification is 32 bits and then is modulo 2 ^ 32 for uncompressed size > 4GiB

## C code

* simplify the implementation with only a `#pragma omp parallel for` for chunk loop and `#pragma omp ordered` for the writing and summarizing part.
* Matt Dowle introduces the use of pool of buffers : the idea is generalized. The pools are created at the beginning and then uses for writing headers and rows. All the `malloc` occur early and no need for an header buffer.
* Deobfuscate some part, especially if ( ) is followed by a new line, no `=-` or `=*`. Lot of work remains. Use of `indent` command ?
* Remove some old debug code (msg)


oliverfoster:
> len in gzip specification is 32 bits and then is false for uncompressed size > 4GiB

Is it not meant to be length modulo 4GiB rather than false? The remainder of dividing by 4GiB?

Screenshot_20240823-194218

philippechataignon:
You're right and this PR version stores the modulo 2**32 as requested but its not the right size.
Note that in new versions of gzip (after 1.12), the length stored in file is no more used. See note on `gzip -lv` on https://savannah.gnu.org/news/?id=10121

```
7z l # (use header)
    Date      Time    Attr         Size   Compressed  Name
                     .....   1298954309   2212092276  mtcars.csv

% gzip --version                                                                                                                                                    
gzip 1.12
% gzip -lv mtcars.csv.gz  # (decompress and takes much more time)
method  crc     date  time           compressed        uncompressed  ratio uncompressed_name
defla 5cd79282 Aug 24 10:19          2212092276          9888888901  77.6% mtcars.csv
```

philippechataignon:
Put PR #5513 in this PR with new param compressLevel.

src/fwrite.c
```diff
@@ -641,10 +647,10 @@ void fwriteMain(fwriteMainArgs args)
   // could be console output) and writing column names to it.
 
   double t0 = wallclock();
-  size_t maxLineLen = eolLen + args.ncol*(2*(doQuote!=0) + sepLen);
+  size_t maxLineLen = eolLen + args.ncol * (2 * (doQuote!=0) + sepLen);
   if (args.doRowNames) {
-    maxLineLen += args.rowNames==NULL ? 1+(int)log10(args.nrow)   // the width of the row number
-                  : (args.rowNameFun==WF_String ? getMaxStringLen(args.rowNames, args.nrow)*2  // *2 in case longest row name is all quotes (!) and all get escaped
+    maxLineLen += args.rowNames==NULL ? 1 + (int) log10(args.nrow)   // the width of the row number
```
MichaelChirico, line 652:
```suggestion
    maxLineLen += args.rowNames==NULL ? 1 + (int)log10(args.nrow)   // the width of the row number
```

src/fwrite.c
```diff
@@ -641,10 +647,10 @@ void fwriteMain(fwriteMainArgs args)
   // could be console output) and writing column names to it.
 
   double t0 = wallclock();
-  size_t maxLineLen = eolLen + args.ncol*(2*(doQuote!=0) + sepLen);
+  size_t maxLineLen = eolLen + args.ncol * (2 * (doQuote!=0) + sepLen);
```
MichaelChirico, line 650:
```suggestion
  size_t maxLineLen = eolLen + args.ncol * (2*(doQuote!=0) + sepLen);
```

man/fwrite.Rd
```diff
@@ -58,6 +59,7 @@ fwrite(x, file = "", append = FALSE, quote = "auto",
   \item{nThread}{The number of threads to use. Experiment to see what works best for your data on your hardware.}
   \item{showProgress}{ Display a progress meter on the console? Ignored when \code{file==""}. }
   \item{compress}{If \code{compress = "auto"} and if \code{file} ends in \code{.gz} then output format is gzipped csv else csv. If \code{compress = "none"}, output format is always csv. If \code{compress = "gzip"} then format is gzipped csv. Output to the console is never gzipped even if \code{compress = "gzip"}. By default, \code{compress = "auto"}.}
+  \item{compressLevel}{Level of compression between 1 (fastest method but less compression) and 9 (slowest compression method). The default compression level is 6 which is a good compromise.}
```
MichaelChirico, line 62: Let's link the gzip man page instead?

https://linux.die.net/man/1/gzip

MichaelChirico (commented):
could you please add some tests? Also a NEWS entry.

I don't understand gzip internals very well, so I'll largely be reviewing based on tests & a non-exhaustive read of the code.

philippechataignon:
Thanks for review. I know that gzip internals are very specific but I think that this PR simplifies the code. I've added a test for compressLevel and the URL for the documentation as suggests. For the fix, it's important to not have a regression. All the tests are actually OK.

NEWS.md
```diff
@@ -111,6 +111,10 @@
 
 14. Passing functions programmatically with `env=` doesn't produce an opaque error, e.g. `DT[, f(b), env = list(f=sum)]`, [#6026](https://github.com/Rdatatable/data.table/issues/6026). Note that it's much better to pass functions like `f="sum"` instead. Thanks to @MichaelChirico for the bug report and fix.
 
+15. `fwrite()` with compress="gzip" produces slightly incompatible gz files with multiple independent chunks [#6356](https://github.com/Rdatatable/data.table/issues/6356). This has been fixed by an internal refactoring of the fwrite function. Thanks to @olivierfoster for report and @philippechataignon for the fix.
+
+16. `fwrite()` gains a new parameter `compressLevel` to control compression level for gzip [#5506](https://github.com/Rdatatable/data.table/issues/5506)] [#5513](https://github.com/Rdatatable/data.table/issues/5513)
```
MichaelChirico, line 116:
```suggestion
16. `fwrite()` gains a new parameter `compressLevel` to control compression level for gzip, [#5506](https://github.com/Rdatatable/data.table/issues/5506). This parameter balances compression speed and total compression, and corresponds directly to the analogous command-line parameter, e.g. `compressLevel=4` corresponds to passing `-4`; the default, `6`, matches the command-line default. Thanks @mgarbuzov for the request and @philippechataignon for implementing.
```

NEWS.md
```diff
@@ -111,6 +111,10 @@
 
 14. Passing functions programmatically with `env=` doesn't produce an opaque error, e.g. `DT[, f(b), env = list(f=sum)]`, [#6026](https://github.com/Rdatatable/data.table/issues/6026). Note that it's much better to pass functions like `f="sum"` instead. Thanks to @MichaelChirico for the bug report and fix.
 
+15. `fwrite()` with compress="gzip" produces slightly incompatible gz files with multiple independent chunks [#6356](https://github.com/Rdatatable/data.table/issues/6356). This has been fixed by an internal refactoring of the fwrite function. Thanks to @olivierfoster for report and @philippechataignon for the fix.
```
MichaelChirico, line 114:
```suggestion
15. `fwrite()` with `compress="gzip"` produces compatible gz files when composed of multiple independent chunks owing to parallelization, [#6356](https://github.com/Rdatatable/data.table/issues/6356).  Earlier `fwrite()` versions could have issues with HTTP upload using `Content-Encoding: gzip` and `Transfer-Encoding: chunked`. Thanks to @olivierfoster for report and @philippechataignon for the fix.
```

Does that sound right? I'm not sure I've described the OP issue correctly as being related to "HTTP".

inst/tests/tests.Rraw
```diff
@@ -10015,8 +10015,12 @@ if (!haszlib()) {
   test(1658.423, file.info(f1)$size < file.info(f2)$size)  # 74 < 804  (file.size() isn't available in R 3.1.0)
   if (test_R.utils) test(1658.43, fread(f1), DT)  # use fread to decompress gz (works cross-platform)
   fwrite(DT, file=f3<-tempfile(), compress="gzip")   # compress to filename not ending .gz
+  fwrite(DT, file=f4<-tempfile(), compress="gzip", compressLevel=1)   # test compressLevel
+  fwrite(DT, file=f5<-tempfile(), compress="gzip", compressLevel=9)
   test(1658.441, file.info(f3)$size, file.info(f1)$size)
-  unlink(c(f1,f2,f3))
+  test(1658.442, file.info(f4)$size > file.info(f1)$size)
```
MichaelChirico, line 10021: Should we use `>=` instead here? Are there any systems/situations where the file size ordering is not strict?

NEWS.md
oliverfoster: It is clear enough, thank you. My username is wrong though.

src/fwrite.c
```diff
@@ -560,14 +560,11 @@ void writeCategString(const void *col, int64_t row, char **pch)
...
 int init_stream(z_stream *stream) {
-  memset(stream, 0, sizeof(z_stream)); // shouldn't be needed, done as part of #4099 to be sure
   stream->next_in = Z_NULL;
   stream->zalloc = Z_NULL;
   stream->zfree = Z_NULL;
   stream->opaque = Z_NULL;
-
-  // 31 comes from : windows bits 15 | 16 gzip format
-  int err = deflateInit2(stream, Z_DEFAULT_COMPRESSION, Z_DEFLATED, 31, 8, Z_DEFAULT_STRATEGY);
+  int err = deflateInit2(stream, gzip_level, Z_DEFLATED, -15, 8, Z_DEFAULT_STRATEGY);
```
MichaelChirico, line 567: why the change 31 --> -15? `-15` looks pretty magical to me, we should add a comment on how to think of it.

src/fwrite.c
```diff
@@ -665,17 +671,20 @@ void fwriteMain(fwriteMainArgs args)
         STOP(_("Internal error: type %d has no max length method implemented"), args.whichFun[j]);  // # nocov
       }
     }
-    if (args.whichFun[j]==WF_Float64 && args.scipen>0) width+=MIN(args.scipen,350); // clamp width to IEEE754 max to avoid scipen=99999 allocating buffer larger than can ever be written
```
MichaelChirico, line 668: `MIN` macro seems useful here for readability, any reason to remove it?

MichaelChirico:
@Anirban166 / @tdhock, I'm wondering if this is an appropriate place for a performance regression test?

In my mind so far we're always writing tests like (1) here's code we found slow (2) here's where it's sped up.

Here we have (1) code that currently is fast (as far as we know) (2) here it is refactored pretty significantly, we want to make sure it hasn't slowed down.

Anirban166:
It definitely is appropriate! 

> we want to make sure it hasn't slowed down.

For me, this is the main point of the regression tests (i.e. not strictly aiming to target historical cases, but future-proofing the current performance; i.e. for things to not to be slower in the long run) so not having a 'Slow' label at present sounds acceptable

inst/tests/tests.Rraw
MichaelChirico: Looks like "yes", at least for 9 vs. 6

MichaelChirico:
> not having a 'Slow' version/label at present sounds acceptable

OK, I think that's where I was stuck -- which labels are appropriate for such a use case.

I guess I thought some of them were required, but looking at `atime_test()` help I guess all of them are optional:

https://github.com/tdhock/atime/blob/main/man/atime_versions.Rd

What labels are best here, then, `Before` and `After`? But in 2 years it becomes "before and after what?"

NEWS.md
```diff
@@ -6,6 +6,12 @@
 
 1. In `DT[, variable := value]`, when value is class `POSIXlt`, we automatically coerce it to class `POSIXct` instead, [#1724](https://github.com/Rdatatable/data.table/issues/1724). Thanks to @linzhp for the report, and Benjamin Schwendinger for the fix.
 
+## NEW FEATURES
+
+1. `fwrite()` with `compress="gzip"` produces compatible gz files when composed of multiple independent chunks owing to parallelization, [#6356](https://github.com/Rdatatable/data.table/issues/6356).  Earlier `fwrite()` versions could have issues with HTTP upload using `Content-Encoding: gzip` and `Transfer-Encoding: chunked`. Thanks to @oliverfoster for report and @philippechataignon for the fix.
```
MichaelChirico, line 11: I wonder if this is new feature or bug fix. Not totally clear to me from the issue discussion -- it's not like decompression had any issues before. Leaving as a "feature" for now.

oliverfoster: I think the browser/web server HTTP gzip stuff is just a narrower implementation of the gzip compression algorithm than was utilised by fwrite. Such that some external utilities could decompress without issue and some could not. It would say fix, but that would depend on the initial intent of the compress flag, if it was only meant for internal use in R, reading and writing csv.gz files, the it is a feature to extend the implementation to support a wider ecosystem. I'm also happy to call it a feature (or fix).

Anirban166:
I suppose we'll have to see if this makes it faster or just retains the same performance - if it's the former, then clearly 'Slow'/'Fast', but if it's the latter (which is what we're at right now), then I'd say maybe 'Before' and 'Current'?

> But in 2 years it becomes "before and after what?"
Yup 'After' will become redundant soon in context (even right now I think it would not clearly indicate the version for which performance is being measured)

Anirban166:
I suppose we'll have to see if this makes it faster or just retains the same performance - if it's the former, then clearly 'Slow'/'Fast', but if it's the latter (which is what we're at right now), then I'd say maybe 'Before' and 'Current'?

> But in 2 years it becomes "before and after what?"

Yup 'After' will become redundant soon in context

Anirban166:
`Before` too would become redundant as you noted, and since the main reason for having these tests would be for a significant refactor and we could any number of those in the long run, I think adding in more context to the label names would make sense (albeit makes them longer but I think we'll be able to find a fine line between readability and good enough context) 

Currently, I'm thinking 'Baseline' and 'Post-gzip'

Anirban166:
`Before` too would become redundant as you noted, and since the main reason for having these tests would be for a significant refactor and we could have any number of those in the long run, I think adding in more context to the label names would make sense (makes them longer but I think we'll be able to find a fine line between readability and good enough context) 

Currently, I'm thinking 'Baseline' and 'Post-gzip'

Anirban166:
`Before` too would become redundant as you noted, and since the main reason for having these tests would be for a significant refactor and we could have any number of those in the long run, I think adding in more context to the label names would make sense (makes them longer but I think we'll be able to find a fine line between readability and good enough context to follow up) 

Currently, I'm thinking 'Baseline' and 'Post-gzip refactor'

tdhock:
> What labels are best here, then, Before and After? But in 2 years it becomes "before and after what?"

I documented the current process here, https://github.com/Rdatatable/data.table/wiki/Performance-testing
it says "make sure to use the names Slow/Fast, or Before/Regression/Fixed, which are assigned special colors, whereas any other names are shown in grey." 

So in this case, since there is no regression, Before/Regression/Fixed should not be used, to avoid confusion with other cases that actually are real historical regressions.

I would suggest Fast for the old commit before this PR. (no Slow necessary)

Anirban166:
Sounds reasonable for this PR (to compare 'Fast' and 'HEAD'), but then will we be only running the old commit for the regression test we make for this case after this PR has been merged?

MichaelChirico:
> I documented the current process here,

Thanks Toby, I had looked at the .ci/atime/tests.R script and some {atime} documentation directly and didn't think to check the Wiki. Should we maybe (1) migrate that documentation into .ci/atime directly (2) add .ci/atime/README.md pointing to the Wiki (3) Point to the Wiki from the first line of .ci/atime/tests.R?

NEWS.md
MichaelChirico: Thanks for the extra context. Yes, I think I'm most comfortable with "new feature" here -- it's extending `fwrite()` to "play better" with a wider set of tools for gzip files, where previously it could best be described as a more "pure" companion to `zlib`. Hopefully that makes sense :)

MichaelChirico:
@philippechataignon do you want to have a go at adding a atime performance regression test? Totally fine if not -- what would help at least would be to write a simple benchmark of gzipped fwrite that you think would capture the important pieces of what's changed here, does that make sense?

tdhock:
yes that would be great to Point to the Wiki from the first line of .ci/atime/tests.R

I would suggest keeping docs on the wiki, which is easier to update, and include screenshots/graphics.

src/fwrite.c
philippechataignon: OK. Add a comment in 117ab456

src/fwrite.c
```diff
@@ -577,16 +574,14 @@ int compressbuff(z_stream *stream, void* dest, size_t *destLen, const void* sour
...
-  if (err == Z_OK) {
-    // with Z_FINISH, deflate must return Z_STREAM_END if correct, otherwise it's an error and we shouldn't return Z_OK (0)
-    err = -9;  // # nocov
-  }
-  *destLen = stream->total_out;
-  return err == Z_STREAM_END ? Z_OK : err;
+  int err = deflate(stream, Z_SYNC_FLUSH);
+  *destLen = *destLen - stream->avail_out;
+  // *destLen = stream->total_out;
+  return err != Z_STREAM_ERROR ? Z_OK : err;
```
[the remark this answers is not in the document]
philippechataignon: OK 5687a0ce

src/fwrite.c
philippechataignon: https://lwn.net/Articles/983965 shows limit with MIN macro function. No problem if you restore MIN macro but I prefer direct test.

philippechataignon:
> @philippechataignon do you want to have a go at adding a atime performance regression test? Totally fine if not -- what would help at least would be to write a simple benchmark of gzipped fwrite that you think would capture the important pieces of what's changed here, does that make sense?

OK for testing regression but notice that the core of fwrite hasn't change : same buffer sizes, same number of jobs, same number of rows per job. Personally I observe similar timings that previous version.

One point of discussion : I notice that #2020 introduces a change that I never realized before this PR. By default `scipen` fwrite parameter uses the value of R option scipen.  Like a lot of persons,  I have a options(scipen = 999) in my Rprofile because I don't like sci output on screen but I never realize that fwrite was penalized. Why ? Because of the maxLineLen which can be very high and gives a lot of batches with few lines and little chunk. In fwrite, scipen has an maximum of 350 but it's a very high limit. And maxLineLen is used for determining number and sizes of  fwrite chunks.

For testing impact, I have this little program :
```
n = 10000
ncol = 1000
dt <- data.table(i=1:n)
dt[, paste0("V", 1:ncol) := lapply(1:ncol, function(x) as.numeric(sample(1:n, replace=T)))]
print(sessionInfo())
system.time(fwrite(dt, "/dev/null", compress="gzip", verbose=T, scipen=0))
system.time(fwrite(dt, "/dev/null", compress="gzip", verbose=T, scipen=999))
```

With scipen = 0
```
maxLineLen=61026. Found in 0.000s
Writing bom (false), yaml (0 characters) and column names (true)
Writing 10000 rows in 73 batches of 137 rows, each buffer size 8388608 bytes (8 MiB), showProgress=1, nth=4
zlib: uncompressed length=48947760 (46 MiB), compressed length=21719798 (20 MiB), ratio=44.4%, crc=26a23b0a
Written 10000 rows in 1.569 secs using 4 threads. MaxBuffUsed=7%
```
With scipen = 999
```
maxLineLen=761026. Found in 0.000s
Writing bom (false), yaml (0 characters) and column names (true)
Writing 10000 rows in 910 batches of 11 rows, each buffer size 8388608 bytes (8 MiB), showProgress=1, nth=4
zlib: uncompressed length=48947760 (46 MiB), compressed length=22745125 (21 MiB), ratio=46.5%, crc=26a23b0a
Written 10000 rows in 1.202 secs using 4 threads. MaxBuffUsed=0%
 ```
In last case real mean line length is ~ 5000 but estimated to 761026. Compression ratio is higher because the buffers are very little used. Surprisingly timing is better despite openmp number of threads overhead.

In my opinion, `scipen` parameter should be 0 in fwrite and not `scipen` option witch is related to `digits` option not present in fwrite and digits is higher in fwrite output (20 ?).

I use this little bench for scipen impact and I think it can be used for atime. I've tried to add this :
```
  # Issue with fwrite length for gzip output, fixed in: https://github.com/Rdatatable/data.table/pull/6393
  # No regression timing test
  "No regression fwrite" = atime::atime_test(
    N = 10^seq(2, 8),
    setup = {
      set.seed(1)
      ncol = 1000
      L <- data.table(i=1:N)
      L[, paste0("V", 1:ncol) := lapply(1:ncol, function(x) rnorm(N))]
    },
    expr = {
      fwrite(dt, "/dev/null", compress="gzip")
    },
    Fast = "117ab45674f1e56304abca83f9f0df50ab0274be", # Close-to-last merge commit in the PR
    Slow = "e73c2c849f921cf4ef51e3809842e0fee9b9f52c"),
```
but I'm not sure that /dev/null is portable and if we write a real file, that's made the timing. 

OK for another one to continue and test that there is not time regression.

**Comment by Anirban166** [comment]
> > I documented the current process here,
> 
> Should we maybe (1) migrate that documentation into .ci/atime directly (2) add .ci/atime/README.md pointing to the Wiki (3) Point to the Wiki from the first line of .ci/atime/tests.R?

2 and 3 sounds good to me

> yes that would be great to Point to the Wiki from the first line of .ci/atime/tests.R

Should I go ahead and make a PR for this quick addition?

> I would suggest keeping docs on the wiki, which is easier to update, and include screenshots/graphics.

I agree, both for being able to include images and in case we miss out on something that other people notice, they should be able to fill in points quickly

**Comment by MichaelChirico** [comment]
> I notice that https://github.com/Rdatatable/data.table/issues/2020 introduces a change that I never realized before this PR. By default scipen fwrite parameter uses the value of R option scipen.

Moved this to #6457 for further discussion, I think it's out of scope here. Thanks!

File: src/fwrite.c [hunk 11]
```diff
@@ -560,14 +560,18 @@ void writeCategString(const void *col, int64_t row, char **pch)
...
   stream->next_in = Z_NULL;
   stream->zalloc = Z_NULL;
   stream->zfree = Z_NULL;
   stream->opaque = Z_NULL;
-
-  // 31 comes from : windows bits 15 | 16 gzip format
-  int err = deflateInit2(stream, Z_DEFAULT_COMPRESSION, Z_DEFLATED, 31, 8, Z_DEFAULT_STRATEGY);
+  // In deflateInit2, windowBits can be –8..–15 for raw deflate, 15 is the bigger window size.
+  // With -15, deflate() will generate raw deflate data with no zlib header or trailer, and will not compute a check value.
+  // Previously this parameter was 31 baecause windowBits can be greater than 15 for optional gzip encoding.
```
MichaelChirico, line 569:
```suggestion
  // Previously this parameter was 31 because windowBits can be greater than 15 for optional gzip encoding.
```

src/fwrite.c
```diff
@@ -577,16 +581,14 @@ int compressbuff(z_stream *stream, void* dest, size_t *destLen, const void* sour
...
-  if (err == Z_OK) {
-    // with Z_FINISH, deflate must return Z_STREAM_END if correct, otherwise it's an error and we shouldn't return Z_OK (0)
-    err = -9;  // # nocov
-  }
-  *destLen = stream->total_out;
-  return err == Z_STREAM_END ? Z_OK : err;
+  int err = deflate(stream, Z_SYNC_FLUSH);
+  *destLen = *destLen - stream->avail_out;
+  // *destLen = stream->total_out;
+  return err != Z_STREAM_ERROR ? Z_OK : err;
```
MichaelChirico, line 587: Thanks for 5687a0ce45e7e4051a1b1a06ebde6bc708f0024b. I think the corresponding change here is:

```suggestion
  return (err != Z_STREAM_ERROR) ? Z_OK : err;
```

src/fwrite.c
MichaelChirico: Excellent reference, wow! I think our case falls under the "obvious const" heading, and we've stuck with trivially ignoring concerns about matching types etc. But I'm swayed it's better to avoid macros -- certainly better to avoid the temptation to bring on further complexity in the future by leaving it out now.

tdhock:
> but I'm not sure that /dev/null is portable 

this only has to run on github actions ubuntu vm, so /dev/null is ok.

Thanks for sharing your code for scipen benchmarking. I adapted it to get the following atime result, which indicates little to no impact on computation time, but a small constant factor increase in memory usage.

image

```r
edit.data.table = function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}
out.csv <- tempfile()
issue6393 <- atime::atime_versions(
  "~/R/data.table",
  N = 2^seq(1, 20),
  pkg.edit.fun=edit.data.table,
  setup = {
    set.seed(1)
    NC = 10
    L <- data.table(i=1:N)
    L[, paste0("V", 1:NC) := replicate(NC, rnorm(N), simplify=FALSE)]
  },
  expr = {
    data.table::fwrite(L, out.csv, compress="gzip")
  },
  Fast="f339aa64c426a9cd7cf2fcb13d91fc4ed353cd31", # Parent of the first commit https://github.com/Rdatatable/data.table/commit/fcc10d73a20837d0f1ad3278ee9168473afa5ff1 in the PR https://github.com/Rdatatable/data.table/pull/6393/commits with major change to fwrite with gzip.
  PR = "117ab45674f1e56304abca83f9f0df50ab0274be") # Close-to-last merge commit in the PR.
plot(issue6393)
```

MichaelChirico:
Thanks again @philippechataignon! I still can't say I've understood the C changes thoroughly, but they pass the existing suite and we have user reports it is working. I am happy to submit now and see if revdep checks tell us anything new. 🚀

MichaelChirico (approved):

src/fwrite.c
```diff
@@ -706,42 +723,153 @@ void fwriteMain(fwriteMainArgs args)
...
+  size_t zbuffSize = 0;
+  size_t compress_len = 0;
+  if (args.is_gzip) {
+  // alloc zlib streams
+    thread_streams = (z_stream*) malloc(nth * sizeof(z_stream));
+    if (verbose) {
+        DTPRINT(_("Allocate %zu bytes for thread_streams\n"), nth * sizeof(z_stream));
+    }
+    if (!thread_streams)
+        STOP(_("Failed to allocated %d bytes for threads_streams."), (int)(nth * sizeof(z_stream)));
```
MichaelChirico, line 814:
```suggestion
        STOP(_("Failed to allocated %d bytes for threads_streams."), (int)(nth * sizeof(z_stream))); // # nocov
```

src/fwrite.c
```diff
@@ -706,42 +723,153 @@ void fwriteMain(fwriteMainArgs args)
...
+    // VLA on stack should be fine for nth structs; in zlib v1.2.11 sizeof(struct)==112 on 64bit
+    // not declared inside the parallel region because solaris appears to move the struct in
+    // memory when the #pragma omp for is entered, which causes zlib's internal self reference
+    // pointer to mismatch, #4099
+
+  // compute zbuffSize which is the same for each thread
+    z_stream *stream = thread_streams;
+    if (init_stream(stream) != Z_OK)
+        STOP(_("Can't init stream structure for deflateBound"));
```
MichaelChirico, line 823:
```suggestion
        STOP(_("Can't init stream structure for deflateBound")); // # nocov
```

src/fwrite.c
```diff
@@ -750,150 +878,95 @@ void fwriteMain(fwriteMainArgs args)
...
-        // takes the max with headerLen size in case of very long header
-        size_t zbuffSize = deflateBound(&stream, headerLen > buffSize ? headerLen : buffSize);
-        char *zbuff = malloc(zbuffSize);
-        if (!zbuff) {
-          free(buff);                                                                                   // # nocov
-          STOP(_("Unable to allocate %zu MiB for zbuffer: %s"), zbuffSize / 1024 / 1024, strerror(errno));  // # nocov
-        }
+        z_stream *stream = thread_streams;
+        if (init_stream(stream) != Z_OK)
+            STOP(_("Can't init stream structure for writing header"));
```
MichaelChirico, line 890:
```suggestion
            STOP(_("Can't init stream structure for writing header")); // # nocov
```

MichaelChirico:
Ah, apologies @philippechataignon I may draw this out just a wee bit longer 🙃

I think our code coverage suite was not up & running when you first posted this PR -- it's back now.

Would you mind taking a look through the report and suggesting which lines could reasonably be covered by new tests, and which should just get `# nocov` treatment? I've suggested a few likely `# nocov` candidates already.

MichaelChirico:
Happy new year @philippechataignon! Checking if you'll find time in the next few weeks to add some coverage tests to this PR. Thanks again!

philippechataignon:
> Happy new year @philippechataignon! Checking if you'll find time in the next few weeks to add some coverage tests to this PR. Thanks again!

Happy new year too. Not a coverage expert code but will have a look.

philippechataignon:
Hi, all checks are now OK for coverage. Review can continue. Thanks @MichaelChirico for advise.

MichaelChirico:
Thanks so much for the extra effort!!

MichaelChirico (approved):
Thanks again! I found a few of the new `# nocov` extraneous and managed to come up with some simple tests to get coverage to 100% -- PTAL at my tail commits if you're interested.

tdhock:
restoring branch so that atime tests can find the commit 3630413ae493a5a61b06c50e80d166924d2ef89a

MichaelChirico:
FYI @philippechataignon, we need to retain the branch for any PR involved in an {atime} performance test:

https://github.com/Rdatatable/data.table/blob/f6dca2d3247b1ca460b59d1eb13b2ebfeab2e259/.ci/atime/tests.R#L247

I'm not super jazzed about this setup, but nothing better has emerged as yet.

PR #6398, opened 2024-08-26
install R on osx gitlab runner

ben-schwen:
Current GL CI runner indicates that R might not be available

`bash: line 152: Rscript: command not found`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -289,7 +289,7 @@ test-mac-rel:
   variables:
     R_VERSION: "$R_REL_VERSION"
   script:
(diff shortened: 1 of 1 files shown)


jangorecki:
Was this change tested?
When working with CI I found it best to push to branch directly on gl so pipeline can run and test it. Submitting PR in gh and waiting for gl to pull recent master is not a nice way to go. Please try pushing directly to GL and see if it works. If it does, then we merge to master. So we don't use master for trying out.

ben-schwen:
Thanks for the info! And the pipeline is also working like a charm for this :)

jangorecki (approved):

PR #6401, opened 2024-08-26
Add an extra condition for checking showProgress value

MichaelChirico:
As noted here:

https://github.com/Rdatatable/data.table/issues/6166#issuecomment-2310735251

This isn't _strictly_ necessary, but I did find it slightly off that the code that triggers the (phantom) error is a query like `DT[,nm,with=FALSE]` -- `showProgress` is never needed for such a case. We only need to check `showProgress` if it might be used, i.e., if `by` is present.

Of course, we could delay checking `showProgress` even further -- it's not used in GForce queries -- but I think that's not very beneficial vs. having the check fail early on in `[`.


tdhock:
great thanks

PR #6419, opened 2024-08-28
add error to prevent segfault when setDT and attr<-

### LINKED ISSUE #6410: data.table randomly crashes R when using `set` on a DT with attributes
### Minimal reproducible example

I found that `set()` crashes R when an attribute is present in a DT. The code below may reproduce the error, but it should be run multiple times. I have no idea why.

```r
mtcars <- data.table::setDT(data.table::copy(mtcars))
attr(mtcars, "bullet") <- "some test"
data.table::set(mtcars, NULL, "test", NA_character_)

 *** caught segfault ***
address 0xffffffffffffffff, cause 'invalid permissions'
```

### Output of sessionInfo()
```
R version 4.4.0 (2024-04-24)
Platform: aarch64-apple-darwin20
Running under: macOS Sonoma 14.6.1

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRblas
.0.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRlapa
ck.dylib;  LAPACK version 3.12.0

locale:

time zone: Asia/Shanghai
tzcode source: internal

attached base packages:

other attached packages:
[1] nvimcom_0.9.45

 [1] vctrs_0.6.5       cli_3.6.2         knitr_1.46        rlang_1.1.3      
 [5] xfun_0.44         processx_3.8.4    purrr_1.0.2       styler_1.10.3    
 [9] glue_1.7.0        clipr_0.8.0       htmltools_0.5.8.1 ps_1.7.6         
[13] fansi_1.0.6       rmarkdown_2.27    R.cache_0.16.0    evaluate_0.23    
[17] tibble_3.2.1      fastmap_1.2.0     yaml_2.3.8        lifecycle_1.0.4  
[21] compiler_4.4.0    fs_1.6.4          pkgconfig_2.0.3   rstudioapi_0.16.0
[25] R.oo_1.26.0       R.utils_2.12.3    digest_0.6.35     R6_2.5.1         
[29] reprex_2.1.0      utf8_1.2.4        pillar_1.9.0      parallel_4.4.0   
[33] callr_3.7.6       magrittr_2.0.3    R.methodsS3_1.8.2 tools_4.4.0      
[37] withr_3.0.0      

```

ben-schwen:
Closes #6410


tdhock:
great thanks for the fix. is there a test case that can be added, that would fail on current master? or is it too difficult to create a reliable test case, because it was a heisenbug?

src/assign.c
```diff
@@ -520,6 +520,8 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
     else if (TRUELENGTH(names) != oldtncol)
       // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768, PRId64 didn't work
       internal_error(__func__, "selfrefnames is ok but tl names [%lld] != tl [%d]", (long long)TRUELENGTH(names), oldtncol);  // # nocov
+    if (!selfrefok(dt, verbose)) // #6410 setDT(dt) and subsequent attr<- can lead to invalid selfref
+      error(_("It appears that at some ealier point, attributes of this data.table have been reassigned. Please ensure to use setattr() rather than attr<-. Otherwise, please report to data.table issue tracker.")); // # nocov
```
MichaelChirico, line 524:
```suggestion
      error(_("It appears that at some earlier point, attributes of this data.table have been reassigned. Please use setattr(DT, name, value) rather than attr(DT, name) <- value. If that doesn't apply to you, please report your case to the data.table issue tracker.")); // # nocov
```

MichaelChirico:
This comment seems to consistently generate the error for me -- we could add it as a regression test. (also NEWS)

https://github.com/Rdatatable/data.table/issues/6410#issuecomment-2312206724

As far as implementation, I would have preferred something like `attr<-.data.table` able to work around the issue by inducing the necessary copies/calls to `setattr()` with warning if possible, but AFAICT it's not possible to turn `attr<-` into an S3 generic...

src/assign.c
aitap: Should we list a specific way to cause the problem ("use `setattr` instead of `attr<-`") or a more cryptic mechanism of how the problem is caused ("It appears that at some earlier point, this data.table has been shallow-copied")? I think it's `shallow_duplicate` that requires catching the wrong `TRUELENGTH` afterwards:

```
R -q -s -e '
 mtcars <- data.table::setDT(data.table::copy(datasets::mtcars));
 x <- rlang::duplicate(mtcars, FALSE);
 data.table::set(x, NULL, "test", NA_character_)
'
# Error in data.table::set(x, NULL, "test", NA_character_) : 
#   It appears that at some earlier point, names of this data.table have been reassigned. Please ensure to use setnames() rather than names<- or colnames<-. Otherwise, please report to data.table issue tracker.
# Execution halted
R -q -s -e '
 mtcars <- data.table::setDT(data.table::copy(datasets::mtcars));
 x <- rlang::duplicate(mtcars, TRUE);
 data.table::set(x, NULL, "test", NA_character_)
'
# 
#  *** caught segfault ***
# address 0x60000006e, cause 'memory not mapped'
# 
# Traceback:
#  1: data.table::set(x, NULL, "test", NA_character_)
# An irrecoverable exception occurred. R is aborting now ...
```

HughParsonage: I think the error message remains unclear. Currently, the error message seems to suggest that the user's mistake occurred "at some earlier point". Ideally we should catch it there. I think 'earlier' is redundant and may actually cause confusion.

Is it possible to create a `attr<-.data.table` method which just raises an error. Or is the issue more broad than that.

aitap: `attr<-` is a primitive function. Unlike [`@<-`](https://github.com/r-devel/r-svn/blob/dff9236d24ca02e4d6f87185bbb2b4e28a9c31cf/src/main/attrib.c#L1617-L1619), it [does not dispatch](https://github.com/r-devel/r-svn/blob/dff9236d24ca02e4d6f87185bbb2b4e28a9c31cf/src/main/attrib.c#L1634-L1664) to S3 methods.

The issue could be more broad than `attr<-` because it is caused by the `data.table` being subject to `shallow_duplicate`. The example above uses `rlang` to cause the issue manually. Don't know whether other, more normal R functions could call `shallow_duplicate` in a similar manner.

tdhock:
> This comment seems to consistently generate the error for me -- we could add it as a regression test

Is there some other way to test this? On my windows the test() in https://github.com/Rdatatable/data.table/issues/6410#issuecomment-2312206724 does not consistently crash R. Also wouldn't it be better if we could somehow get a test for this that fails using current master, without crashing R? (don't know if that is possible though)

MichaelChirico:
LGTM, please add NEWS+test

ben-schwen:
I think we need to distinguish here between the segfault and the `selfref` not being ok. AFAICT the `selfref` will always be not ok, while the segfault might not happen every time on windows. @tdhock can you check if the added test works under windows?

NEWS.md
```diff
@@ -65,6 +65,8 @@ rowwiseDT(
 
 9. `rbindlist` and `rbind` binding `bit64::integer64` columns with `character`/`complex`/`list` columns now works, [#5504](https://github.com/Rdatatable/data.table/issues/5504). Thanks to @MichaelChirico for the request and @ben-schwen for the PR.
 
+10. `setDT(df)` and subsequent adding new columns with `set` could lead to segfault when object was attributes were added in between with `attr(df) <-`, [#6410](https://github.com/Rdatatable/data.table/issues/6410). Thanks to @hongyuanjia for the report and @ben-schwen for the PR.
```
MichaelChirico, line 68:
```suggestion
10. Fixed possible segfault in `setDT(df); attr(df, key) <- value; set(df, ...)`, i.e. adding columns to an object with `set()` that was converted to data.table and later had attributes add with `attr<-`, [#6410](https://github.com/Rdatatable/data.table/issues/6410). Thanks to @hongyuanjia for the report and @ben-schwen for the PR. Note that `setattr()` should be preferred for adding attributes to a data.table.
```

src/assign.c
```diff
@@ -529,6 +529,8 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
     else if (TRUELENGTH(names) != oldtncol)
       // Use (long long) to cast R_xlen_t to a fixed type to robustly avoid -Wformat compiler warnings, see #5768, PRId64 didn't work
       internal_error(__func__, "selfrefnames is ok but tl names [%lld] != tl [%d]", (long long)TRUELENGTH(names), oldtncol);  // # nocov
+    if (!selfrefok(dt, verbose)) // #6410 setDT(dt) and subsequent attr<- can lead to invalid selfref
```
MichaelChirico, line 532: Change looks reasonable, one lingering question: have we gotten the root cause here? Is `attr<-` the _only_ way to reach this error? e.g. what about `class<-` or `attributes<-` or `names<-`? I don't see anything in the code nearby here that speaks specifically to attributes, so if this error can trigger in some other way, it may be confusing to the user to be admonished about `attr<-`.

aitap: [`rlang::duplicate(shallow = TRUE)`](https://github.com/Rdatatable/data.table/pull/6419#discussion_r1735658681) may be contrived, so here's an example without it:

```r
mtcars <- data.table::setDT(data.table::copy(datasets::mtcars))
mtcars[[length(mtcars)]] <- 42
data.table::set(mtcars, NULL, paste0("test", 1:10), NA_character_)
```

The root cause is the same, a `shallow_duplicate` done by R. (Not entirely sure why.)

src/assign.c
MichaelChirico: > I think 'earlier' is redundant and may actually cause confusion.

That wording is shared with a similar error:

https://github.com/Rdatatable/data.table/blob/038ce2b79cc0bbf90cc9f7012dcac82029cad100/src/assign.c#L526

So I don't think we should block the PR based on that. We can file a follow-up issue to improve clarity IMO.

> attr<- is a primitive function. 

Somehow I missed this thread 🙃 https://github.com/Rdatatable/data.table/pull/6419#issuecomment-2316684212

src/assign.c
ben-schwen: Good point. I also thought about other cases but all your examples basically manipulate underlying attributes. As Ivan pointed out, the error boils down to R doing a shallow copy

MichaelChirico: Got it. Because it's R doing the copy, I also doubt that we'll be able to intercept every possible way for this to happen, too -- checking `selrefok()` in a crucial region is the best we can do as I see it.

MichaelChirico (approved):
Filed #6494 for follow-up.

MichaelChirico (approved):

tdhock:
> can you check if the added test works under windows?

yes I get an error the first time

https://github.com/Rdatatable/data.table/issues/6410#issuecomment-2346878143

PR #6420, opened 2024-08-28
Drop direct use of NAMED and LEVELS

aitap:
These are the remaining easy pickings for #6180.

Costs:

* With `verbose = TRUE`, `assign` now only prints whether `REFCNT()` is `>= 1` and `>= 2`, not its value directly.
* `Rf_charIsASCII` is only an "eapi" function, not a bona-fide "api" function, and it is currently only in R-devel. But it doesn't raise NOTEs.

`NAMED`/`REFCNT` version information gathered [here](https://aitap.codeberg.page/R-api/#MARK_005fNOT_005fMUTABLE) (shameless plug).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 1. Tests run again when some Suggests packages are missing, [#6411](https://github.com/Rdatatable/data.table/issues/6411). Thanks @aadler for the note and @MichaelChirico for the fix.
 
+2. All direct use of `NAMED` and some direct use of `LEVELS` has been removed from the package, reducing the use of non-API functions, see [#6420](https://github.com/Rdatatable/data.table/pull/6420) by Ivan Krylov. The work is on the overall issue [#6180](https://github.com/Rdatatable/data.table/issues/6180) reported by Michael Chirico.
+
 # data.table [v1.16.0](https://github.com/Rdatatable/data.table/milestone/30)  (25 August 2024)
 
 ## BREAKING CHANGES
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 1. Tests run again when some Suggests packages are missing, [#6411](https://github.com/Rdatatable/data.table/issues/6411). Thanks @aadler for the note and @MichaelChirico for the fix.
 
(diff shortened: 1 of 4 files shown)


MichaelChirico:
thanks for the great resource!

would you mind

 1. splitting this into two PRs? id rather wait for Rf_isASCII to make a release before depending on it
 2. adding a NOTE to NEWS & citing yourself?

re:LEVELS, is there a performance penalty for the new approach?

aitap:
> re:LEVELS, is there a performance penalty for the new approach?

Currently, `IS_ASCII(x)` expands to `LEVELS(x) & 64`, so a function call and a bitmask test. With marked encodings, a call to `getCharCE()` and an equality test should be similar in performance. `Rf_charIsASCII` performs the same bitmask test inside a function call (where `LEVELS` expands to access internal `SEXPREC` fields). I think it should all be similar. If you have any pointers, I can try to write a performance test.

Will split and update the PR soon-ish. Thanks for the comments!

NEWS.md
```diff
@@ -10,6 +10,8 @@
 
 1. Tests run again when some Suggests packages are missing, [#6411](https://github.com/Rdatatable/data.table/issues/6411). Thanks @aadler for the note and @MichaelChirico for the fix.
 
+2. All direct use of `NAMED` and some direct use of `LEVELS` has been removed from the package, reducing the use of non-API functions, see [#6420](https://github.com/Rdatatable/data.table/pull/6420) by Ivan Krylov. The work is on the overall issue [#6180](https://github.com/Rdatatable/data.table/issues/6180) reported by Michael Chirico.
```
MichaelChirico, line 13:
```suggestion
2. Continued work to remove non-API C functions, [#6180](https://github.com/Rdatatable/data.table/issues/6180). Thanks Ivan Krylov for the PR and for writing a clear and concise guide about the R API: https://aitap.codeberg.page/R-api/.
```

MichaelChirico (approved):

tdhock:
nice NEWS item!
This is a non-trivial contribution that merits adding Ivan to Authors@R in DESCRIPTION as role="ctb" 
Ivan can you please do that in a new PR?

PR #6421, opened 2024-08-28
Macos binaries

ben-schwen:
Follows #6398 and adds binaries for `macosx`

@jangorecki Is there also a way to pre-test the publishing mechanism without master?

Building + testing on different R versions on `arm64` worked fine on CI


.ci/ci.R
```diff
@@ -169,7 +169,8 @@ function(pkgs,
     newpkgs <- newpkgs[availpkgs]
   }
 
-  pkgsext <- switch(type,
+  typeshort <- if (substr(type, 1L, 11L) == "mac.binary.") "mac.binary" else type
```
MichaelChirico, line 172:
```suggestion
  typeshort <- if (startsWith(type, "mac.binary.")) "mac.binary" else type
```

Or `gsub("(mac.binary)[.].*", "\\1", type)`

ben-schwen: Corresponding line is taken from `contrib.url(repos, type)` which does the same for `type`

jangorecki:
Nice to see your active contribution in this field. I felt I am the only one that can do glci.

Regarding testing.
What I used in this case was to add to integration stage extra branch name of THIS branch, next to the master. So then integration also kicks in for this branch.
If this branch is to date with master you could probably try to add thus branch name also to deploy stage, but then please don't leave it for a night or so, but confirm after the pipeline finishes, and shorty after finishing testing rerun pipeline for master. I y is easy from web ui to run pipeline on demand, without need for any new commits.
Ultimately what you can do for testing is to self host gitlab, this is what I did in the distant past, using docker was pretty easy. Although not sure if public runners will work on self hosted instance.

ben-schwen:
[CI results](https://gitlab.com/Rdatatable/data.table/-/commit/b586b5a382d9c6a184be4f5dc0aad7cd70d3040b) look good.

[Cran-like checks](https://rdatatable.gitlab.io/-/data.table/-/jobs/8134116048/artifacts/bus/integration/cran/web/checks/check_results_data.table.html)

ben-schwen:
> No binaries for Mac R-devel?

There is no .pkg for Mac for R-devel and I'm not familiar enough with osx for building it myself

jangorecki:
Fair enough. Therefore that most likely implies much less users of r-devel on Mac.

jangorecki (approved):

PR #6424, opened 2024-08-28
Remove "internal error" translations

MichaelChirico:
This also updates messages since the last {R,src} ➡️ .pot extraction; we don't need to update translations yet.


rffontenelle:
I've spotted at least one untranslated string. Should it be translated in here or after this PR gets merged?

MichaelChirico:
> I've spotted at least one untranslated string. Should it be translated in here or after this PR gets merged?

Let's wait until nearer release. There will be more messages to translate then. The new untranslated strings here are kind of incidental to the removal of no-longer-translated messages.

rivaquiroga (approved):

rffontenelle (approved):

joshhwuu (approved):

PR #6445, opened 2024-08-30
Reset dec='\0' to detect numbers while dec is undecided

### LINKED ISSUE #6440: Parsing ISO8601 with microseconds in fread() broken in 1.16.0
Since 1.16.0, some data containing microsecond-precision ISO 8601 timestamps fails to parse in `fread`.

If parsed without class hints, this results in a verbatim `char` string.
If parsed with `"POSIXct"` in `colClasses`, it takes a significantly longer time on larger datasets and results in a date without time information.

```r
library("data.table")
fread(text="Datetime\n2023-10-12T06:53:53.123Z")
fread(text="DatetimeExplicit\n2023-10-12T06:53:53.123Z", colClasses="POSIXct")
```

Expected output (1.15.4):
```
              Datetime
                <POSc>
1: 2023-10-12 06:53:53
      DatetimeExplicit
                <POSc>
1: 2023-10-12 06:53:53
```

Output in 1.16.0 and 1.16.99:
```
                   Datetime
                     <char>
1: 2023-10-12T06:53:53.123Z
   DatetimeExplicit
             <POSc>
1:       2023-10-12
```

# Output of `sessionInfo()`
```
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 22.04.4 LTS

Matrix products: default

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8

time zone: Etc/UTC
tzcode source: system (glibc)

attached base packages:

other attached packages:
[1] data.table_1.16.99

[1] compiler_4.4.1 tools_4.4.1
```

# Additional observation

Not all data fails to parse. The following example works in 1.16.99, and I am not sure why:

```r
fread(text="A,B\n2023-10-12T06:53:53.123Z,2023-10-12T06:53:53.123Z")
```
```
                     A                   B
                <POSc>              <POSc>
1: 2023-10-12 06:53:53 2023-10-12 06:53:53
```

This might be why this slipped past regression tests.
In my code that broke with the update, the timestamps were a single isolated column in the middle of other types of columns.

MichaelChirico:
Closes #6440.

`dec='\0'` is reset each time when moving on to the next _column_:

https://github.com/Rdatatable/data.table/blob/80c46a05b0f859c39c756777c207732422877a81/src/fread.c#L1250

The issue here really is isolated to timestamps with milliseconds: _within_ a column, once `dec=','` is tried & fails to parse the timestamp (as a real number), `dec=','` is left dangling --> the timestamp fails to parse. This only matters for "higher" types than real numbers where double parsing is attempted, i.e. only timestamps.

Because `IS_DEC_TYPE()` excludes timestamp on master, and hence whether a timestamp parses with `,` doesn't count to the "race" vs. `dec='.'`, the timestamp also fails to parse with `,` as the microsecond separator:

```r
fread(text="Datetime\n2023-10-12T06:53:53,123Z", sep="\t") # sep= required otherwise `,` is seen as sep in this simple example
#                    Datetime
#                      <char>
# 1: 2023-10-12T06:53:53,123Z
```

But this can be nudged to work in files with fields recognized as real numbers on `master`:
```
fread(text="x\tDatetime\n1,0\t2023-10-12T06:53:53,123Z", sep="\t")
#        x            Datetime
#    <num>              <POSc>
# 1:     1 2023-10-12 06:53:53
```

One incidental change of the PR as of now is that ISO timestamps will parse correctly with `,` as the millisecond separator; @kav2k confirmed this is OK and in fact recognized by the standard.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -1072,7 +1072,6 @@ static void parse_iso8601_timestamp(FieldParseContext *ctx)
 
   date_only:
 
-  //Rprintf("date=%d\thour=%d\tz_hour=%d\tminute=%d\ttz_minute=%d\tsecond=%.1f\n", date, hour, tz_hour, minute, tz_minute, second);
   // cast upfront needed to prevent silent overflow
   *target = 86400*(double)date + 3600*(hour - tz_hour) + 60*(minute - tz_minute) + second;
 
@@ -1233,9 +1232,13 @@ static int detect_types( const char **pch, int8_t type[], int ncol, bool *bumped
         }
       }
       ch = fieldStart;
-      if (autoDec && IS_DEC_TYPE(tmpType[field]) && dec == '.') { // . didn't parse a double; try ,
-        dec = ',';
-        continue;
+      if (autoDec && IS_DEC_TYPE(tmpType[field])) {
+        if (dec == '.') { // '.' didn't parse a double; try ','
+          dec = ',';
+          continue; // i.e., restart since tmpType not incremented
+        } else if (dec == ',') { // ',' didn't parse a double either; reset
+          dec = '\0';
+        }
       }
       while (++tmpType[field]<CT_STRING && disabled_parsers[tmpType[field]]) {};
       *bumped = true;
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,6 +67,8 @@ rowwiseDT(
 
 10. Fixed possible segfault in `setDT(df); attr(df, key) <- value; set(df, ...)`, i.e. adding columns to an object with `set()` that was converted to data.table with `setDT()` and later had attributes add with `attr<-`, [#6410](https://github.com/Rdatatable/data.table/issues/6410). Thanks to @hongyuanjia for the report and @ben-schwen for the PR. Note that `setattr()` should be preferred for adding attributes to a data.table.
 
+11. `fread()` automatically detects timestamps with sub-second accuracy again, [#6440](https://github.com/Rdatatable/data.table/issues/6440). This was a regression due to interference with new `dec='auto'` support. Thanks @kav2k for the concise report and @MichaelChirico for the fix.
+
 ## NOTES
 
 1. Tests run again when some Suggests packages are missing, [#6411](https://github.com/Rdatatable/data.table/issues/6411). Thanks @aadler for the note and @MichaelChirico for the fix.
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18549,21 +18549,38 @@ test(2255, as.data.table(DF), output="DF1.V1.*DF1.V2.*DF2.V3.*DF2.V4.*V5")
 DT = data.table(a = letters, b = 1:26/6, c = 1:26)
 ## auto-detect dec=','
 fwrite(DT, f <- tempfile(), dec=',', sep=';')
-test(2256.1, fread(f), DT)
+test(2256.01, fread(f), DT)
 
 fwrite(DT, f, dec=',', sep='|')
-test(2256.2, fread(f), DT)
+test(2256.02, fread(f), DT)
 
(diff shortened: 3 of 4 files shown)


MichaelChirico:
> My concern is data that potentially mixes separators, because even preferring , for decimal, software will likely output . for ISO timestamps (data.table itself is an example):

Oh, nice example. I think that's a follow-up issue for `fwrite()` to support.

`fread()`, in the auto-detect step, "runs a race" between `dec='.'` and `dec=','`; whichever successfully parses more fields is chosen (tie goes to `'.'`):

https://github.com/Rdatatable/data.table/blob/80c46a05b0f859c39c756777c207732422877a81/src/fread.c#L1922

I think this is "good enough". We could possibly support `dec = <vector>` to allow specifying a `dec` for each column, but I'd rather not go there, and fix `fwrite()` to avoid producing such mixtures in the first place.

> Probably not a good idea to support this though. 

Yea, no thanks 😂 

TBH it would not be _that_ hard to support (switch to double parsing here, and don't fail for missing some components), but I doubt it's common enough to be worth maintaining:

https://github.com/Rdatatable/data.table/blob/80c46a05b0f859c39c756777c207732422877a81/src/fread.c#L1032

> So this fix (which, if I understand correctly, is trying to reparse with the other separator if one fails)

Some nuance, this fix is only about the "auto-detect" phase of `fread()` which runs on a sample of lines to decide the column types. Previously (since `dec='auto'` support was added) there was no re-try on timestamps, and only `dec=','` was tried (by mistake).

MichaelChirico:
> My concern is data that potentially mixes separators, because even preferring , for decimal, software will likely output . for ISO timestamps (data.table itself is an example):

From #6454, this point is moot from data.table -- `fwrite(dec=',')` will write timestamps with `,` :)

MichaelChirico:
Checking in @ben-schwen, got time for a review? Would be nice to get this last part of the 1.16.2 patch submitted.

src/fread.c
```diff
@@ -1038,7 +1038,7 @@ static void parse_iso8601_timestamp(FieldParseContext *ctx)
   if (second == NA_FLOAT64 || second < 0 || second >= 60)
     goto fail;
 
-  if (*ch == 'Z') {
+  if (*ch == 'Z') {
```
ben-schwen, line 1041:
```suggestion
  if (*ch == 'Z') {
```

ben-schwen:
fread(text="t\n2023-10-12T06:53:53.123Z", colClasses="nanotime") # works
fread(text="t\n2023-10-12T06:53:53,123Z", colClasses="nanotime") # does not

MichaelChirico:
Got another "refuse to merge unrelated histories" thing here, strange. Reinstated the changes but lost the history again :\

MichaelChirico:
> ```r
> fread(text="t\n2023-10-12T06:53:53.123Z", colClasses="nanotime") # works
> fread(text="t\n2023-10-12T06:53:53,123Z", colClasses="nanotime", sep=";") # does not
> ```

That's an R side issue:

```r
as.nanotime("2023-10-12T06:53:53,123Z")
# Error: Parse error on 2023-10-12T06:53:53,123Z
```

But I'm reminded of #6107, there may be a way to get it returned to R as `POSIXct` and then it will work:

```
as.nanotime(fread(text="t\n2023-10-12T06:53:53.123Z", colClasses="nanotime")$t)
# [1] 2023-10-12T06:53:53.123+00:00
```

Out of scope here though, filing for follow-up.

ben-schwen:
LGTM, ty!

PR #6449, opened 2024-08-31
Initialize 'ties' enum in frank

MichaelChirico:
Observed this from the compiler on the clang-ubsan codespace:

```
frank.c:118:12: warning: variable 'ties' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
  else if (!strcmp(pties, "last")) ties = LAST;
           ^~~~~~~~~~~~~~~~~~~~~~
frank.c:125:7: note: uninitialized use occurs here
  if (ties==MEAN) dans=REAL(ans); else ians=INTEGER(ans);
      ^~~~
frank.c:118:8: note: remove the 'if' if its condition is always true
  else if (!strcmp(pties, "last")) ties = LAST;
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~           ~
frank.c:110:3: note: variable 'ties' is declared here
  enum {MEAN, MAX, MIN, DENSE, SEQUENCE, LAST} ties; // RUNLENGTH
  ^
1 warning generated.
```

That's because `internal_error()`, our new custom error function, is not recognized by the compiler as stopping. AFAIU the way to proceed here is to use `NORET`; here are similar routines in the R sources:

https://github.com/search?q=repo%3Ar-devel%2Fr-svn%20NORET&type=code

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -1,5 +1,7 @@
 # A list of performance tests.
 #
+# See documentation in https://github.com/Rdatatable/data.table/wiki/Performance-testing for best practices.
+#
 # Each entry in this list corresponds to a performance test and contains a sublist with three mandatory arguments:
 # - N: A numeric sequence of data sizes to vary.
 # - setup: An expression evaluated for every data size before measuring time/memory.
--- a/.ci/linters/po/msgfmt_linter.R
+++ b/.ci/linters/po/msgfmt_linter.R
@@ -4,7 +4,7 @@ msgfmt_linter <- function(po_file) {
   mo_tmp <- tempfile()
   on.exit(unlink(mo_tmp))
 
-  res = system2("msgfmt", c("--statistics", po_file, "-o", mo_tmp)), stdout=TRUE, stderr=TRUE)
+  res = system2("msgfmt", c("--statistics", po_file, "-o", mo_tmp), stdout=TRUE, stderr=TRUE)
   if (any(grepl("untranslated message|fuzzy translation", res))) {
     cat(sprintf("In %s, found incomplete translations:\n%s\n", po_file, paste(res, collapse="\n")))
     stop("Please fix.")
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -60,3 +60,6 @@
 # performance testing
 /.ci/atime/tests.R @tdhock @Anirban166
 /.github/workflows/performance-tests.yaml @Anirban166
+
+# docs
+/man/openmp-utils.Rd @Anirban166
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 1. Using `print.data.table()` with character truncation using `datatable.prettyprint.char` no longer errors with `NA` entries, [#6441](https://github.com/Rdatatable/data.table/issues/6441). Thanks to @r2evans for the bug report, and @joshhwuu for the fix.
 
+2. `fwrite()` respects `dec=','` for timestamp columns (`POSIXct` or `nanotime`) with sub-second accuracy, [#6446](https://github.com/Rdatatable/data.table/issues/6446). Thanks @kav2k for pointing out the inconsistency and @MichaelChirico for the PR.
+
 ## NOTES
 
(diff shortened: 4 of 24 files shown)


MichaelChirico:
Looking around, it seems there's an approach that signals to the compiler that `internal_error` never returns; R itself uses the `NORET` macro for this IINM:

https://github.com/r-devel/r-svn/blob/0e1629a4224a1b5a2589859182b1d535da53d97d/src/include/R_ext/Error.h#L42-L58

```
#if defined NORET
#elif (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202301L)
# define NORET [[noreturn]]
#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201102L
# define NORET _Noreturn
#elif defined(__GNUC__) && __GNUC__ >= 3
// LLVM and Apple clang identify themselves as 4.
// But Mandriva (or OpenMandriva) is said to patch clang to 11.
// Boost also uses this for __SUNPRO_CC >= 0x590
# define NORET __attribute__((noreturn))
#else
# define NORET
#endif
```

That's some pretty complex logic, I'd love to just re-use it instead of re-implementing it, but I don't think we can. Mostly I see other packages re-implementing it:

https://github.com/search?q=org%3Acran%20lang%3AC%20NORET&type=code

MichaelChirico:
Hmm hail mary to @kevinushey -- I see your https://github.com/RcppCore/Rcpp/commit/a2e45fa5a3e95125938f3865a661e42c00877900.

For whatever reason just using `NORET` did not work as intended, and even `#include <R_ext/Error.h>` didn't suffice. If I copy-paste the way to define `NORET`, it does work. This is on R 4.1+, I see all that is about `NORET` not defined prior to 3.2 (we depend on 3.3).

Any idea why `NORET` wouldn't be defined? I don't see this documented at all in R-exts/R-ints, besides a use without comment here:

https://github.com/r-devel/r-svn/blob/982288a5640ba584ff0d7e6ba85bf566a4184686/doc/manual/R-exts.texi#L16610

It's also never come up on any mailing list:

https://github.com/search?q=repo%3AMichaelChirico%2Fr-mailing-list-archive%20NORET&type=code

Don't spend more than a minute or two on this, I'll go to r-package-devel if it's not immediately obvious to you. Thanks!

MichaelChirico:
Hmm, even if I get `NORET` correctly defined (on the compiler here, I see `void _Noreturn internal_error` under appropriate `clang -E` invocation), I still get the `maybe-uninitialized` compiler warning. Not sure why.

MichaelChirico:
OK, going back to just "initialize `ties=0`", probably over-thinking it trying to make the `NORET` thing work.

HughParsonage:
Note that `NORET` (an alias for the `_Noreturn` specifier) is not a synonym for `void` even if it raises an error. It's undefined behaviour to specify `NORET` for a function if the entire body of the function is executed (even if the last line is a `NORET` function).

MichaelChirico:
> It's undefined behaviour to specify NORET for a function if the entire body of the function is executed (even if the last line is a NORET function).

Could you elaborate what you mean here?

> What happens if you put an error or return R_NilValue after the internal_error call (within the last else branch)?

I assume you mean "in the terminal `else` branch inside `frank`". Not tested, but this is the first I've ever seen this compiler warning; previously it was just `error()` there:

https://github.com/Rdatatable/data.table/blob/18f4ab610a0124eb3b8aa8162117944ad36c4e38/src/frank.c#L120

My goal with the `NORET` stuff is to make the compiler treat `internal_error` exactly as it treats `error`.

MichaelChirico:
> After trying to reproduce I wasn't even able to reproduce your warning!

I'm also not seeing it everywhere. the clang-ubsan codespace is where I got it. maybe it's a clang thing, otherwise version-specific.

kevinushey:
Coming here kind of late, but IMHO initializing `ties` as done in this PR is probably the simplest solution; it seems like the issue here is compiler-specific wonkiness (technical term) but working around it in a non-invasive way seems reasonable.

MichaelChirico:
I agree. It may make sense to revisit if it keeps coming up.

PR #6451, opened 2024-09-02
Fix fcase() segfault (#6452)

MichaelChirico:


MichaelChirico:
* **#6452** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6451** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

src/fifelse.c
```diff
@@ -209,12 +209,12 @@ SEXP fcaseR(SEXP rho, SEXP args) {
             "Note that the default argument must be named explicitly, e.g., default=0"), narg - 2);
   }
   int nprotect=0, l;
-  int64_t len0=0, len1=0, len2=0;
+  int64_t n_ans=0, n_this_arg=0, n_undecided=0;
```
Anirban166, line 212: `n_undecided` sounds good to indicate the count of the unresolved elements/cases. An alternative I can think of is `n_pending` (as in tracking how many elements are left to be evaluated)

Is `n_this_arg` meant to represent the length of the `thens` argument that is being processed in the current iteration? (I only see it being assigned to it at 285 below, so I'm wondering if it could be renamed specific to it)

Anirban166 (approved):
I like how this change separates the variables of the same names above so as to not be confused with the the test, yes, and no arguments!

tdhock:
this is great thanks for refactoring variable names

src/fifelse.c
MichaelChirico: Thanks! Yes you're right about `n_this_arg`. I'm roughly indifferent between `n_undecided` and `n_pending`; both are way better than `len2` :)

MichaelChirico:
Made the mistake of merging #6452 into this branch instead of `master` -- I should have managed the stack from graphite.dev directly but hurriedly clicked merge from my phone :)

I'll just merge this branch to master, rename the PR & merge here as if nothing happened :)

PR #6452, opened 2024-09-02
Fix fcase() segfault

### LINKED ISSUE #6448: intermittent issue w/ fcase in 1.16
im getting a frequent, but intermittent issue with `fcase` with 1.16

The offending code has not changed in weeks
The issue only started after upgrading to 1.16
The issues does not surface after reverting to 1.15.4
The issue is intermittent. it cannot reliably be reproduced with the same input
The problem code does not use `data.table`, but uses `fcase` and `fcoalesce`

Im am trying to isolate the issue to produce an MRE, but due to the intermittent nature, i cant even reliably trap it in the debugger start investigating

MichaelChirico:
Closes #6448.

@aitap I didn't add any test -- any ideas for what would make sense as a regression test here? Technically this could have been caught by existing tests, under a stricter operating environment.

The test from OP did reliably segfault, but it's harder to turn that into a test for the _absence_ of a segfault -- experimenting, I was getting it to run 1500+ times before segfaulting.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 1. Using `print.data.table()` with character truncation using `datatable.prettyprint.char` no longer errors with `NA` entries, [#6441](https://github.com/Rdatatable/data.table/issues/6441). Thanks to @r2evans for the bug report, and @joshhwuu for the fix.
 
+2. Fixed a segfault in `fcase()`, [#6448](https://github.com/Rdatatable/data.table/issues/6448). Thanks @ethanbsmith for reporting with reprex, @aitap for finding the root cause, and @MichaelChirico for the PR.
+
 ## NOTES
 
 1. Tests run again when some Suggests packages are missing, [#6411](https://github.com/Rdatatable/data.table/issues/6411). Thanks @aadler for the note and @MichaelChirico for the fix.
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -19059,3 +19059,14 @@ test(2280.1, internal_error("broken"), error="Internal error.*broken")
 test(2280.2, internal_error("broken %d%s", 1, "2"), error="Internal error.*broken 12")
 foo = function(...) internal_error("broken")
 test(2280.3, foo(), error="Internal error in foo: broken")
+
+# ensure proper PROTECT() within fcase, #6448
+x <- 1:3
+test(2281,
+     fcase(
+       x<2, structure(list(1), class = "foo"),
+       x<3, structure(list(2), class = "foo"),
+       # Force gc() and some allocations which have a good chance at landing in the region that was earlier left unprotected
+       { gc(full = TRUE); replicate(10, FALSE); x<4 },
+       `attr<-`(list(3), "class", "foo")),
+     structure(list(1, 2, 3), class = "foo"))
--- a/src/fifelse.c
+++ b/src/fifelse.c
@@ -210,13 +210,15 @@ SEXP fcaseR(SEXP rho, SEXP args) {
   }
   int nprotect=0, l;
   int64_t n_ans=0, n_this_arg=0, n_undecided=0;
-  SEXP ans=R_NilValue, value0=R_NilValue, tracker=R_NilValue, whens=R_NilValue, thens=R_NilValue;
+  SEXP ans=R_NilValue, tracker=R_NilValue, whens=R_NilValue, thens=R_NilValue;
+  SEXP ans_class, ans_levels;
   PROTECT_INDEX Iwhens, Ithens;
   PROTECT_WITH_INDEX(whens, &Iwhens); nprotect++;
   PROTECT_WITH_INDEX(thens, &Ithens); nprotect++;
   SEXPTYPE ans_type=NILSXP;
   // naout means if the output is scalar logic na
   bool imask = true, naout = false, idefault = false;
+  bool ans_is_factor;
   int *restrict p = NULL;
   const int n = narg/2;
   for (int i=0; i<n; ++i) {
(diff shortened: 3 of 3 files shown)


MichaelChirico:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6452?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6452** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6451** <a href="https://app.graphite.dev/github/pr/Rdatatable/data.table/6451?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

MichaelChirico:
What @aitap suggested is a much simpler one-line fix here:

https://github.com/Rdatatable/data.table/blob/02ced80a1812e014b8c4e215167b6e23819b340e/src/fifelse.c#L238

```diff
-      value0 = thens;
+      value0 = PROTECT(thens); nprotect++;
```

Looking again though, it does look wasteful to me that we repeatedly run `getAttrib()` on `value0` -- and the only reason we keep `value0` around to begin with is to make sure subsequent `thens` match the first one (re-named `value0`) on class & factor levels. So the approach taken here is to record those attributes instead, and not keep all of `value0` protected.

It makes the code a bit lengthier, but should be more memory-efficient.

HughParsonage:
Are we sure this is the fix? I am surprised that a miscalculated UNPROTECT would not have triggered a stack imbalance warning message.

MichaelChirico:
> Are we sure this is the fix? I am surprised that a miscalculated UNPROTECT would not have triggered a stack imbalance warning message.

there's no missed unprotect -- because thens is REPROTECT'd.

I confirmed the examples from the issue which reliably default pretty quickly, no longer do so.

aitap:
> any ideas for what would make sense as a regression test here

Can we afford to enable `gctorture` for one test? With it enabled, it's possible to force `value0` to point to memory allocated by one of the `eval()` calls and thus fail the `class` test:

```
library(data.table)
x <- 1:5
gctorture(TRUE)
system.time(local({
 on.exit(gctorture(FALSE))
 fcase(
  x<2, structure(list(1), class = "foo"),
  x<3, structure(list(2), class = "foo"),
  # after i=1, `value0` has been unprotected;
  # one of the temporary values from the following eval()s
  # has a chance to show up at its former address
  x<4, `attr<-`(list(3), "class", "foo"),
  x<5, structure(list(4), class = "foo"),
  x<6, structure(list(5), class = "foo")
 )
}))
# Error in fcase(x < 2, structure(list(1), class = "foo"), x < 3, structure(list(2),  :
#   Argument #8 has different class than argument #2, Please make sure all output values have the same class.
# Calls: system.time ... eval.parent -> eval -> eval -> eval -> eval -> fcase
# Timing stopped at: 0.819 0 0.82
```

Unfortunately, it takes four to six columns depending on the heap layout (I get complaints about argument 8 on Linux but 10 on Windows) _and_ the better part of a second to reproduce. Without `gctorture`, `fcase` returns the right answer (a list of `1`...`5` with the class set) because garbage collection and memory reuse doesn't naturally happen that frequently.

MichaelChirico:
could you test making the input a factor with other attributes too? that will trigger the max # of branches in fcase with a getAttrib call

MichaelChirico:
Really slick idea to get `gc()` to happen in the middle of `fcase()` execution. Not clear what breadth of systems we can expect such a test to actually work as intended, but it's definitely better than nothing!

aitap:
Here's a test with factors that errors for me on R-devel/data.table from Git on Linux and R-4.4.0/data.table-1.16.0 on Windows, all x86_64:

```
x <- 1:3
y <- I(as.factor(LETTERS[1:3]))
fcase(
  x<2, y[3],
  x<3, y[2],
  { gc(full = TRUE); for (i in 1:1000) list(factor(letters[3:6])); x<4 }, y[1]
)
```

It doesn't fail in a more interesting manner because `value0` ends up not being a factor:

```
#2  0x00007ffff44cb764 in fcaseR (rho=0x555558b43390, args=0x555558884098) at fifelse.c:279
279                 error(_("Argument #2 and argument #%d are both factor but their levels are different."), i*2+2);
(gdb) p R_inspect(value0)
@555557edc868 16 STRSXP g0c1 [] (len=1, tl=0)
  @555555a6e668 09 CHARSXP g1c2 [MARK,REF(2109),gp=0x61] [ASCII] [cached] "unique.default"
```

Somehow, without the lists in the outputs, it takes more effort to get this particular `SEXP` reused, but it is never `free`'d, only given out by `allocVector`.

tdhock (approved):
looks reasonable to me, thanks!
~would we have caught this if we were using rchk? #3618~
edit: looks like we do have an rchk ci job, why didn't it catch this? isn't rchk supposed to check the PROTECT etc?

MichaelChirico:
rchk is a static analyzer, it can't catch everything. possibly it doesn't know how to handle PROTECT_WITH_INDEX.

tdhock:
@kalibera you are the main author of rchk right?
This example may be interesting for you to investigate, for improving rchk?

kalibera:
> @kalibera you are the main author of rchk right? This example may be interesting for you to investigate, for improving rchk?

Unfortunately it is not feasible examples like this would be covered. The main challenge for rchk here is, in rchk terminology, that something gets implicitly protected, but later implicitly unprotected. rchk gets the first part, but isn't able to collect enough information to track the second part (`value0` is implicitly protected by getting a value from protected `thens`, but later `thens` loses its protection in an additional iteration of the loop). It is not realistic for cases like this to be handled and in principle one should not assume that a lack of report from rchk means that the code is correct (or even lacks some kind of bugs) - it's just a bug finding tool based on static analysis, not a formal verification tool.

A higher-level explanation could be that rchk is modelling which "variables" are protected, but it is an approximation (often also of how we think about the code, but still an approximation) - in fact R objects on the heap are protected or not. For this abstraction to work better, one should not copy pointers to SEXPs. When that is still needed, one needs to be careful and perhaps could add a comment to the code about why it does not break protection.

Otherwise I confirm the reason of the bug in the code discussed above: `value0` holds to an object initially protected via `thens`, but then no longer protected.

PR #6454, opened 2024-09-02
fwrite respects dec=',' for time columns when needed

### LINKED ISSUE #6446: fwrite(dec=',') should use it for timestamps with milliseconds for consistency
Okay, I just looked at the discussions of the standard, and it actually allows the comma separator (which is surprising).

My concern is data that potentially mixes separators, because even preferring `,` for decimal, software will likely output `.` for ISO timestamps (`data.table` itself is an example):
```r
t <- fread(text="x\tDatetime\n1,1\t2023-10-12T06:53:53,123Z", sep="\t")
fwrite(t, file="", dec=",", sep=";")
```
```
x;Datetime
1,1;2023-10-12T06:53:53.123Z
```

So this fix (which, if I understand correctly, is trying to reparse with the other separator if one fails) should be OK?

Here's a sample of my real data that tripped this issue, which was output by data.table's `fwrite()`:
```
CoilID;AntennaID;Time;TagID;Pen;Side;Position;Location;Coil_Y;Coil_X
1;16403160;2023-10-12T10:30:55.270Z;DA2C6411;1;AKB;Litter central;middle;1;6
3;16403160;2023-10-12T10:30:55.270Z;DA459D86;1;AKB;Litter central;middle;1;4
15;16402963;2023-10-12T10:31:00.900Z;DA459D86;1;AKB;Litter central;right;2;3
14;16402963;2023-10-12T10:31:02.240Z;DA2C6411;1;AKB;Litter central;right;2;1
11;16403160;2023-10-12T10:31:02.650Z;DA2C6411;1;AKB;Litter central;middle;2;6
```

----

P.S. Datetimes are weird. Were you aware that ISO 8601 allows fractional values other than seconds? E.g. `10:20,5` is valid and means `10:20:30`... Probably not a good idea to support this though. Something more sane like RFC 3339 draft (which `fwrite` seems to follow) is already supported and that's likely good enough.

_Originally posted by @kav2k in https://github.com/Rdatatable/data.table/issues/6445#issuecomment-2320260096_

MichaelChirico:
Closes #6446

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -19059,3 +19059,7 @@ test(2280.1, internal_error("broken"), error="Internal error.*broken")
 test(2280.2, internal_error("broken %d%s", 1, "2"), error="Internal error.*broken 12")
 foo = function(...) internal_error("broken")
 test(2280.3, foo(), error="Internal error in foo: broken")
+
+# fwrite respects dec=',' for sub-second timestamps, #6446
+test(2281.1, fwrite(data.table(a=.POSIXct(0.001)), dec=',', sep=';'), output="1970-01-01T00:00:00,001Z")
+test(2281.2, fwrite(data.table(a=.POSIXct(0.0001)), dec=',', sep=';'), output="1970-01-01T00:00:00,000100Z")
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 1. Using `print.data.table()` with character truncation using `datatable.prettyprint.char` no longer errors with `NA` entries, [#6441](https://github.com/Rdatatable/data.table/issues/6441). Thanks to @r2evans for the bug report, and @joshhwuu for the fix.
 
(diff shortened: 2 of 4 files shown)


inst/tests/tests.Rraw
```diff
@@ -19059,3 +19059,7 @@ test(2280.1, internal_error("broken"), error="Internal error.*broken")
 test(2280.2, internal_error("broken %d%s", 1, "2"), error="Internal error.*broken 12")
 foo = function(...) internal_error("broken")
 test(2280.3, foo(), error="Internal error in foo: broken")
+
+# fwrite respects dec=',' for sub-second timestamps, #6446
+test(2281.1, fwrite(data.table(a=.POSIXct(0.001)), dec=',', sep=';'), output="1970-01-01T00:00:00,001Z")
+test(2281.2, fwrite(data.table(a=.POSIXct(0.0001)), dec=',', sep=';'), output="1970-01-01T00:00:00,000100Z")
```
Anirban166, line 19065: Tested, LGTM!

Anirban166 (approved):

PR #6459, opened 2024-09-04
Don't set selfref on data.frame

### LINKED ISSUE #5286: `shallow` adds .internal.selfref to an input data.frame

MichaelChirico:
Closes #5286. Restoration of #5556.

@OfekShilon you are a project member so you should be able to push edits directly to this branch.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -192,7 +192,9 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   SET_TRUELENGTH(newnames,n);
   SETLENGTH(newdt,l);
   SET_TRUELENGTH(newdt,n);
-  setselfref(newdt);
+  if( INHERITS(dt, char_datatable) )  { // #5286
+      setselfref(newdt);
+  }
   UNPROTECT(protecti);
   return(newdt);
 }
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -14158,7 +14158,7 @@ test(1984.242, na.omit(data.table(A=c(1,NA,2)), cols=character()), data.table(A=
 test(1984.25, rbindlist(list(DT[1L], DT[2L]), idcol = TRUE), data.table(.id=1:2, a=1:2))
 test(1984.26, setalloccol(`*tmp*`), error='setalloccol attempting to modify `*tmp*`')
 DF = as.data.frame(DT)
-test(1984.27, shallow(DF), DF)  # shallow (which is not exported) works on DF from v1.14.2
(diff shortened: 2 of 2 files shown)


MichaelChirico:
Please add a NEWS entry.

I am wondering if `setselfref()` itself should just be a no-op on non-data.table. `setselfref()` is also used in subset:

https://github.com/Rdatatable/data.table/blob/568c938584991d44579e03f7b9d67d3966613bbf/src/subset.c#L361

Is it possible to wind up running that on a data.frame?

OfekShilon:
Added a NEWS item. Agreed that making `setselfref` a no-op for data frames is safer - fixed.

OfekShilon:
Not sure why a conflict resolution was needed (didn't change anything). Anyway should be good to go now.

src/assign.c
```diff
@@ -193,6 +194,7 @@ static SEXP shallow(SEXP dt, SEXP cols, R_len_t n)
   SETLENGTH(newdt,l);
   SET_TRUELENGTH(newdt,n);
   setselfref(newdt);
+
```
MichaelChirico, line 197:
```suggestion
```

src/assign.c
```diff
@@ -24,6 +24,7 @@ static void finalizer(SEXP p)
 }
 
 void setselfref(SEXP x) {
+  if( !INHERITS(x, char_datatable) )  return; // #5286
```
MichaelChirico, line 27:
```suggestion
  if(!INHERITS(x, char_datatable))  return; // #5286
```

MichaelChirico:
Sorry it took so long to review such a tiny change 🙏

OfekShilon:
> Sorry it took so long to review such a tiny change 🙏

Thanks for merging, that was very fast.

PR #6288, opened 2024-07-17
adding an atime test case; groupby with dogroups (R expression) #PR4558

DorisAmoakohene:
This test case discusses the issue reported on performing group computations, specifically when running R's C eval on each group (q7 and q8) in the db-benchmark, indicating a slowness in the implementation of the code. https://github.com/Rdatatable/data.table/issues/4200

This is the https://github.com/Rdatatable/data.table/pull/4558 that discusses the Cause of the Regression: https://github.com/Rdatatable/data.table/issues/4200#issue-555186870 which appears that the regression occurred during the evaluation of C code within these particular groups, indicating a performance issue or slowness in the implementation of the code.

[The regression was fixed Regression by the addition of const int nth = getDTthreads](https://github.com/Rdatatable/data.table/pull/4558/files)

PR #6464, opened 2024-09-04
Added an atime test for performance improvement when using by with verbose set to TRUE

### LINKED ISSUE #6463: Add an atime test for verbose=TRUE speed-up
this looks like a good start for performance test, https://github.com/Anirban166/data.table/pull/13#issuecomment-2235046139 because we can clearly see the difference between Slow and HEAD but please add a Fast commit from this PR

_Originally posted by @tdhock in https://github.com/Rdatatable/data.table/issues/6296#issuecomment-2329681126_

Anirban166:
Closes #6463

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -167,7 +167,7 @@ test.list <- atime::atime_test_list(
       dt_mod <- copy(dt)
     },
     expr = data.table:::`[.data.table`(dt_mod, , 1, by = a, verbose = TRUE),
-    Slow = "a01f00f7438daf4612280d6886e6929fa8c8f76e" # Parent of the first commit (https://github.com/Rdatatable/data.table/commit/fc0c1e76408c34a8482f16f7421d262c7f1bde32) in the PR (https://github.com/Rdatatable/data.table/pull/6296/commits) that fixes the issue
+    Slow = "a01f00f7438daf4612280d6886e6929fa8c8f76e", # Parent of the first commit (https://github.com/Rdatatable/data.table/commit/fc0c1e76408c34a8482f16f7421d262c7f1bde32) in the PR (https://github.com/Rdatatable/data.table/pull/6296/commits) that fixes the issue
     Fast = "efe631a167818b48f80e15132638fa625768c668"), # Last commit in the PR (https://github.com/Rdatatable/data.table/pull/6296/commits) that fixes the issue
 
   NULL)
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -168,7 +168,7 @@ test.list <- atime::atime_test_list(
     },
     expr = data.table:::`[.data.table`(dt_mod, , 1, by = a, verbose = TRUE),
     Slow = "a9331b649a21532b83db8d8518cd2c8ca26d002e", # Parent of the commit that fixes the issue (https://github.com/Rdatatable/data.table/commit/f248bbe6d1204dfc8def62328788eaadcc8e17a1)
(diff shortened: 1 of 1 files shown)


Anirban166:
I noticed some minor inconsistencies in the file, for e.g. double `expr` from #6288: (also, using curly braces for only multiple lines might be good for consistency)
https://github.com/Rdatatable/data.table/blob/687de9d09bacde0d349c2d07222da2514ca39803/.ci/atime/tests.R#L137-L139

I can fix them within this PR if needed.

.ci/atime/tests.R
```diff
@@ -158,5 +158,17 @@ test.list <- atime::atime_test_list(
...
+  # Fixed in: https://github.com/Rdatatable/data.table/pull/6296
+  "DT[by, verbose = TRUE] improved in #6296" = atime::atime_test(
+    N = 10^seq(1, 9),
+    setup = {
+      dt = data.table(a = 1:N)
+      dt_mod <- copy(dt)
+    },
+    expr = data.table:::`[.data.table`(dt_mod, , 1, by = a, verbose = TRUE),
+    Slow = "a9331b649a21532b83db8d8518cd2c8ca26d002e", # Parent of the commit that fixes the issue (https://github.com/Rdatatable/data.table/commit/f248bbe6d1204dfc8def62328788eaadcc8e17a1)
+    Fast = "f248bbe6d1204dfc8def62328788eaadcc8e17a1"), # Merge commit of the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/6296)
```
tdhock, line 171: in https://github.com/Rdatatable/data.table/wiki/Performance-testing#documentation-of-historical-commits standard comment forms are suggested, can you please use them here?

Parent of first commit (COMMIT_LINK) in the PR (PR_LINK) which VERBS the issue

Last commit in the PR (PR_LINK) that fixes the issue

tdhock:
performance test figure looks good

Anirban166:
> please file another PR to fix the issues your found in tests.R

Sure

.ci/atime/tests.R
Anirban166: The first sentence makes sense and should work in all cases assuming the PR branched off from master, but the 'Last commit ... ' one doesn't fit when we are using the merge commit for a PR (not the last commit, and the commits tab page doesn't show the merge commit so `https://github.com/Rdatatable/data.table/pull/<PR_Number>/commits` should not be used in such cases).

Since the merge commit is the only way this will work for our case here (commits in the PR/#6296 got squashed or combined into one as is being done for every PR, so there are no other commits in the master branch to fetch from, aside for the merge commit: [@master](https://github.com/Rdatatable/data.table/pull/4501/commits)).

So here is what I went with:
```
Slow = "a01f00f7438daf4612280d6886e6929fa8c8f76e", # Parent of the first commit (https://github.com/Rdatatable/data.table/commit/fc0c1e76408c34a8482f16f7421d262c7f1bde32) in the PR (https://github.com/Rdatatable/data.table/pull/6296/commits) that fixes the issue
Fast = "f248bbe6d1204dfc8def62328788eaadcc8e17a1"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/6296) that fixes the issue
```
Let me know if that works or sounds reasonable to you.

MichaelChirico:
Please use a more self-contained name for the PR -- ideally, we can glance at the name of the PR in, say, 5 years' time and glean the basic idea of what the PR is doing.

Anirban166:
> Please use a more self-contained name for the PR -- ideally, we can glance at the name of the PR in, say, 5 years' time and glean the basic idea of what the PR is doing.

Thanks for pointing it out!
(I submitted the PR quickly without naming it so it went with the name of the commit, my bad)

tdhock:
great thanks

PR #6469, opened 2024-09-05
French translation (po files)

phgrosjean:
Here are the `fr.po` and `R-fr.po`for {data.table}.


MichaelChirico:
There we go, finally :)

Please copy over the compiled .mo files to this branch as well

aitap:
> ```shell
> # msgid "skip=%llu but the input only has %llu line%s"
> # msgid "skip=%lu but the input only has %lu line%s"
> # msgid "skip=%<PRIu64> but the input only has %<PRIu64> line%s"
> ```

I think `msgfmt` generates these on order to make `PRI<x>64` work in a portable manner. They come up with all possible `printf` specifiers that could be substituted in place of `PRI<x>64` and [generate a translation for each](https://www.gnu.org/software/gettext/manual/html_node/Preparing-Strings.html#index-inttypes_002eh):

> The PO file will contain the string "The amount is %0&lt;PRId64>\n". The translators will provide a translation containing "%0&lt;PRId64>" as well, and at runtime the gettext function’s result will contain the appropriate constant string, "d" or "ld" or "lld".

MichaelChirico:
@aitap interesting/nice spot... but why has that only happened for this particular .mo file?

aitap:
It's a relatively new feature ([June 2023](https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commit;h=c46f4cfca1400395e1f04b357118bff0214205be)). An `msgfmt` that doesn't have the `--no-redundancy` flag will not be expanding the specifiers.

MichaelChirico:
thanks! asking since I'm afk -- is there a corresponding option for msgunfmt?

I'm thinking of how to adjust our CI check for .mo-match-.po to account for the new feature:

 1. only require "all translated messages in .po are found in .mo"
 2. use msgfmt and compare the resulting .mo (instead of currently we do the reverse)
 3. use a msgunfmt option to better match the files
 4. ...

aitap:
@MichaelChirico looks like (3) is not an option. On [libera.chat](https://libera.chat)'s `#gnu` I was recommended (2) as the simplest option, but to keep track of `msgfmt`'s options `--alignment`, `--endianness`, `--no-hash`, which may result in semantivally equivalent (and equally portable) but different binaries. Unless someone goes out of their way to translate on a big-endian machine (or we get another update to deal with non-GNU `gettext` functions), it looks like the `*.mo` binaries should be stable. `--no-hash` seems to save 173 kB in total for `es`, `fr`, `pt_BR`, `zh_CN`, but I didn't try to measure any slowdowns from binary search vs hash lookup.

MichaelChirico:
Thanks for tracking down all the nuance here... maybe option 4 is "translators don't contribute .mo at all, leave that to data.table maintainers" is the right approach. We could also look into having GHA do that automatically.

MichaelChirico:
(also just realized I tagged @philippechataignon by mistake, very sorry for the stray ping!! I meant @phgrosjean 😬)

phgrosjean:
@MichaelChirico translators don't contribute .mo files is for me the best. My .mo file was done using poEdit 3.5 (latest version at that time). I guess that .mo files may slightly differ, depending on the software used to build it.

If it is OK, I will eliminate the .mo files from the pull request.

MichaelChirico:
CI currently requires the .mo files -- leave them in for now. But you don't need to worry about fixing the issue yourself.

MichaelChirico (approved):

MichaelChirico:
Thanks @phgrosjean!! I'll deal with getting the GHA for ensuring .po matches .mo separately.

aitap:
This pull request may have needed a `CODEOWNERS` change:
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/CODEOWNERS#L42-L50

Issue #6194, opened 2024-06-19
Make plural alternatives as plural-forms messages


rffontenelle:
Some messages embed plural alternatives e.g. "column(s)" or "column%s" (where %s can be empty "" or "s") and this makes harder for translators to properly translate this message.

Some languages have plural-form expressions far more complex than English e.g. while English alternate between singular or plural, Slovenian has 4 plurals. Even for those that have the same plural-form expression as English, appending multiple "(s)" to denote plural alternative makes harder to read, mainly for screen readers used by people who suffers for visually impairment.

A better solution is to split such messages in plural-forms using ngettext.

Some references https://michaelchirico.github.io/potools/articles/developers.html#plurals and (one personal favorite) https://wiki.gnome.org/TranslationProject/DevGuidelines/Plurals

For me, as Brazilian Portuguese translator, there are cases that I can work around the situation, but there are times I need to add way more "(s)" than the original message. Hence, using plural-forms messages would be much more cleaner.

Here is a non-exhaustive list of messages I found in the source code with such problem:

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/src/assign.c#L394

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/src/rbindlist.c#L57

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/src/utils.c#L289

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/R/data.table.R#L721

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/R/data.table.R#L735

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/R/foverlaps.R#L157

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/R/print.data.table.R#L270-L273

MichaelChirico:
Thanks for tracking these down! Yes please switch to `ngettext()`. So far they haven't been sussed out because Mandarin doesn't really have plural forms, so translating there is easy without using `ngettext()`.

MichaelChirico:
[...]

PR #6475, opened 2024-09-05
Use dngettext() for some plural messages in C

MichaelChirico:
Part of #6194. R-side messages are in #6476


src/po.h
```diff
@@ -1,6 +1,9 @@
 #ifdef ENABLE_NLS
 #include <libintl.h>
 #define _(String) dgettext("data.table", String)
+// NB: flip argument order to match that of R's ngettext()
+#define Pl_(n, String1, StringPlural) dngettext("data.table", String1, StringPlural, n)
```
aitap, line 5: `dngettext` accepts an `unsigned long`. I think that all the lengths given to it so far are `int`s and have good reason to stay nonnegative `int`s. So it's worth mentioning `((n) > ULONG_MAX ? ((n) % 1000000) + 1000000 : (n)))` from the [gettext documentation](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html), but we probably don't need it.

aitap (approved):

src/po.h
MichaelChirico: Let's leave it be for now, great flag & find though.

PR #6476, opened 2024-09-05
Use ngettext() for some plural messages in R

### LINKED ISSUE #6194: Make plural alternatives as plural-forms messages

MichaelChirico:
Closes #6194. C-side messages were in #6475.

@rffontenelle for one case, I can't tell what variable determines whether this should be plural or not. I'm tempted to just leave it be, translators can just do anything? WDYT?

https://github.com/Rdatatable/data.table/blob/97980d9a0795095e8b32d5bbf315da7e087e39fd/R/foverlaps.R#L157


tdhock (approved):
looks fine to me. I didn't know about ngettext, that is neat!

MichaelChirico:
> I didn't know about ngettext, that is neat!

It's crucial for generic translation -- once you leave romance languages + CJK pluralization gets pretty wild:

https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

Issue #6482, opened 2024-09-06
Fragmented sentences in translatable strings


aitap:
Splitting a sentence into fragments may make it harder to translate because the parts and the whole have different grammatical cases. 

For example, [here](https://github.com/Rdatatable/data.table/blob/cf05b4a0951958d2b6f9f7a4819e5b2cd851ceca/src/assign.c#L739) "target vector" would be "целево**й** вектор", but [later](https://github.com/Rdatatable/data.table/blob/cf05b4a0951958d2b6f9f7a4819e5b2cd851ceca/src/assign.c#L787) "Assigning factor numbers to `%s`", "target vector" must be "Присваиваю фактор целево**му** вектор**у**". I think that this is limited to `targetDesc()` in `assign.c`, all such cases are in the dative case and so can be translated into Russian pre-inflected, but I'm not sure this would remain the case for Arabic and Hindi.

Also a problem for languages with strict word order might be concatenation of messages in `fread.c` [[1](https://github.com/Rdatatable/data.table/blob/cf05b4a0951958d2b6f9f7a4819e5b2cd851ceca/src/fread.c#L446), [2](https://github.com/Rdatatable/data.table/blob/cf05b4a0951958d2b6f9f7a4819e5b2cd851ceca/src/fread.c#L451)], used [here](https://github.com/Rdatatable/data.table/blob/cf05b4a0951958d2b6f9f7a4819e5b2cd851ceca/src/fread.c#L1537). [This](https://github.com/Rdatatable/data.table/blob/cf05b4a0951958d2b6f9f7a4819e5b2cd851ceca/src/fread.c#L1804), on the other hand, is already slightly broken because `msg` is a full sentence.

_Originally posted by @aitap in https://github.com/Rdatatable/data.table/issues/6194#issuecomment-2333894555_

rikivillalba:
Can those be enumerated in some site i.e. a branch in which those are flagged with a comment // [fragmented] or alike?

MichaelChirico:
The `targetDesc()` case in assign.c is tough. Basically, there are a series of operations that can either affect a simple vector, or a "special" vector, namely one where we have extra info that it's a certain column in a data.frame. When we have that extra info, it's nice to present to the user, otherwise generic "target vector" is the description. It's used for 7 messages now:

https://github.com/Rdatatable/data.table/blob/c616cb9c9a37ad7d3adfcd1e3fedda2c3b1e527a/src/assign.c#L789
[...]

PR #6483, opened 2024-09-06
Refactor copyFile() message fragment for ease of translation

MichaelChirico:
One part of #6482. The complication here is that we _always_ show the before- and after- message if `verbose=True`; otherwise, we take care to only flag this to the user if the required file copy is particularly time-consuming (>.5 seconds).

It's also a tad complicated because there's a separation of logic to several places in the code...

That inspired refactoring `copyFile()` to return the time taken -- now all the messaging logic surrounding `copyFile()` is all done in one place --> easier to follow.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -1536,7 +1536,7 @@ int freadMain(freadMainArgs _args) {
           DTPRINT(_("  File ends abruptly with '%c'. Final end-of-line is missing. Copying file in RAM. %s.\n"), eof[-1], msg);
         // In future, we may discover a way to mmap fileSize+1 on all OS when fileSize%4096==0, reliably. If and when, this clause can be updated with no code impact elsewhere.
         double time_taken = copyFile(fileSize);
-        if (time_taken < 0) {
+        if (time_taken == -1.0) {
           if (!verbose)
             DTPRINT("%s. Attempt to copy file in RAM failed.", msg);
           STOP(_("Unable to allocate %s of contiguous virtual RAM."), filesize_to_str(fileSize));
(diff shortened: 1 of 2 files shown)


MichaelChirico:
This approach (pull out the message generation logic from `copyFile()` helper feels a bit more natural to me, though I do wonder if it's a bit repetitive.

OTOH, it's only called in two places, so it's probably not worth over-thinking how to best modularize here.

src/fread.c
```diff
@@ -1534,10 +1531,20 @@ int freadMain(freadMainArgs _args) {
...
       } else {
-        const char *msg = _("This file is very unusual: it ends abruptly without a final newline, and also its size is a multiple of 4096 bytes. Please properly end the last row with a newline using for example 'echo >> file' to avoid this ");
-        if (verbose) DTPRINT(_("  File ends abruptly with '%c'. Copying file in RAM. %s copy.\n"), eof[-1], msg);
+        const char *msg = _("This file is very unusual: it ends abruptly without a final newline, and also its size is a multiple of 4096 bytes. Please properly end the last row with a newline using for example 'echo >> file'");
+        if (verbose)
+          DTPRINT(_("  File ends abruptly with '%c'. Final end-of-line is missing. Copying file in RAM. %s.\n"), eof[-1], msg);
         // In future, we may discover a way to mmap fileSize+1 on all OS when fileSize%4096==0, reliably. If and when, this clause can be updated with no code impact elsewhere.
-        copyFile(fileSize, msg, verbose);
+        double time_taken = copyFile(fileSize);
+        if (time_taken < 0) {
```
aitap, line 1539: In your experience as a professional software developer, does it pay to account for very unlikely corner cases, such as the system clock being very coincidentally adjusted backwards right as `copyFile` is running? If you think it's too unlikely to care about, that's probably fine.

Maybe we could return `NaN` on error, but that's a weird error return code.

aitap (approved):

src/fread.c
MichaelChirico: I was vaguely musing if that's possible as I wrote this.

I think we can wait for a user report before fussing here

copyFile should almost never be called. the two cases are (a) one-column files with multiple trailing missing rows and (b) files without trailing newline, in both cases only if the file is an exact multiple of 4096 bytes.

for all that to be true AND the clock to run backwards, it feels like the user is just messing with us :)

MichaelChirico: one easy way to make this even more impossible to trigger by mistake would be to only check copy file==-1.0 instead of just being negative. then the user would somehow have to have all of the above with an impossibly exact regression of the system time.

aitap: If someone manages to trigger exactly that, well, it would be too late to buy lottery tickets.

PR #6485, opened 2024-09-08
Add performance regression test for fread Date coercion

MichaelChirico:
Follow-up to #6107


MichaelChirico:
1. Tests look like they're WAI -- clear performance improvement in all 3 cases, yay!
2. I think we're starting to run into "image is too wide with the # of tests we have"
3. I would have expected the 3 tests I added to be first or last in the image, but they're nonconsecutive? Is that intended?

tdhock:
yes performance test results look good

the panels are sorted by p-value, most significant difference between HEAD and min on the left

tdhock:
for historical documentation I see 
```r
Slow = "e9087ce9860bac77c51467b19e92cf4b72ca78c7",  # PR #6107 merged to master
Fast = "a77e8c22e44e904835d7b34b047df2eff069d1f2"), # Commit prior to #6107 merged to master
```
which I think is incorrect.
* on https://github.com/Rdatatable/data.table/pull/6107 I see that a77e8c22e44e904835d7b34b047df2eff069d1f2 is the merge commit for #6107
* and on https://github.com/Rdatatable/data.table/commit/a77e8c22e44e904835d7b34b047df2eff069d1f2 I see that e9087ce9860bac77c51467b19e92cf4b72ca78c7 is the parent of that merge commit

so I would suggest changing to the standard format suggested on https://github.com/Rdatatable/data.table/wiki/Performance-testing#documentation-of-historical-commits

"Last commit in the PR (PR_LINK) that fixes the issue" and  "Parent of first commit (COMMIT_LINK) in the PR (PR_LINK) which VERBS the issue"


```r
Slow = "e9087ce9860bac77c51467b19e92cf4b72ca78c7", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/a77e8c22e44e904835d7b34b047df2eff069d1f2) of the PR (https://github.com/Rdatatable/data.table/pull/6107) that fixes the issue
Fast = "a77e8c22e44e904835d7b34b047df2eff069d1f2"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/6107) that fixes the issue
```

tdhock:
I believe that the two commits were a good choice, but the comments/docs were unclear/incorrect.
* "PR #6107 merged to master" comment was for `Slow = "e9087ce9860bac77c51467b19e92cf4b72ca78c7"` but should be for `Fast = "a77e8c22e44e904835d7b34b047df2eff069d1f2"`
* and vice versa

MichaelChirico:
PTAL, thanks!

tdhock:
great thanks for the revision.
before merging I would like to attempt a revision that emphasizes the similarity between these three test cases. 
everything in these three test cases is the same, except for the select/colClasses arguments to fread, right?

MichaelChirico:
That's right -- corresponding to the 3 places that needed a code change in the fix PR: https://github.com/Rdatatable/data.table/pull/6107/files#diff-452d6ffc9ba079fa770f997eada9b232f513c449912e85b47e7c09f7f43a037c

(removing duplication across those branches is left as a future exercise 🙃)

.ci/atime/tests.R
```diff
@@ -171,4 +171,25 @@ test.list <- atime::atime_test_list(
     Fast = "f248bbe6d1204dfc8def62328788eaadcc8e17a1"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/6296) that fixes the issue
 
   NULL)
+
+extra.args.6107 <- c(
+  "colClasses=list(Date='date')",
+  "colClasses='Date'",
+  "select=list(Date='date')")
+for(extra.arg in extra.args.6107){
```
tdhock, line 179: @MichaelChirico what do you think about this refactor using a for loop?
I think it clearly emphasizes the similarities and differences between the three proposed test cases.
is it ok with you?

.ci/atime/tests.R
```diff
@@ -171,4 +171,25 @@ test.list <- atime::atime_test_list(
...
+    pkg.edit.fun=test.list[[1]]$pkg.edit.fun,
+    setup = {
+      set.seed(1)
+      DT = data.table(date=.Date(sample(20000, N, replace=TRUE)))
+      tmp_csv = tempfile()
+      fwrite(DT, tmp_csv)
+    },
+    Slow = "e9087ce9860bac77c51467b19e92cf4b72ca78c7", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/a77e8c22e44e904835d7b34b047df2eff069d1f2) of the PR (https://github.com/Rdatatable/data.table/pull/6107) that fixes the issue
+    Fast = "a77e8c22e44e904835d7b34b047df2eff069d1f2") # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/6107) that fixes the issue
+  this.test$expr = parse(text=sprintf("data.table::fread(tmp_csv, %s)", extra.arg))[[1]]
```
MichaelChirico, line 191:
```suggestion
  this.test$expr = str2lang(sprintf("data.table::fread(tmp_csv, %s)", extra.arg))
```

.ci/atime/tests.R
MichaelChirico, line 179:
```suggestion
for (extra.arg in extra.args.6107){
```

.ci/atime/tests.R
```diff
@@ -171,4 +171,25 @@ test.list <- atime::atime_test_list(
     Fast = "f248bbe6d1204dfc8def62328788eaadcc8e17a1"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/6296) that fixes the issue
 
   NULL)
+
+extra.args.6107 <- c(
```
MichaelChirico, line 175:
```suggestion
# #6107 fixed performance across 3 ways to specify a column as Date, test each individually
extra.args.6107 <- c(
```

.ci/atime/tests.R
MichaelChirico: Yes, very nice idea!

Anirban166 (approved):

MichaelChirico:
Looks like it failed @tdhock... does `test.list` have to be the last object of the script or something?

tdhock:
The error in CI is "test.list not found" which I think is coming from lazy evaluation of this line, `pkg.edit.fun=test.list[[1]]$pkg.edit.fun` which should be fixed by https://github.com/tdhock/atime/commit/418f39e56d4f748a073bcba6f5485488e165f2aa which is currently only on github, so I think this PR will have to wait until I push a new atime to CRAN, is that OK?

MichaelChirico:
Sure, are you able to confirm the fix in the meantime?

tdhock:
yes it runs on my local machine with atime from github

tdhock:
I figured out a fix that uses the existing atime functionality, so no need to wait for me to push new atime on CRAN.
please review.

MichaelChirico:
Thanks @tdhock!

Issue #6486, opened 2024-09-08
revdeps broken by DT[, j=<expr>:<expr] being with=TRUE


MichaelChirico:
A number of revdeps were broken by db3cbb47212590ef1d1b0e2ce797178fc2e1baaf (#4460):

```
MetaIntegrator
SOMnmR
VoxR
categoryEncodings
flextable
hesim
iml
immunarch
kalmanfilter
polypharmacy
sentometrics
sjtable2df
```

I haven't looked into the root cause(s) yet.

MichaelChirico:
OK, here's an MRE (not sure if it covers everything):

```r
DT=data.table(a=1,b=2)
DT[,1:ncol(DT)]

# ** on CRAN
#        a     b
#    <num> <num>
# 1:     1     2

# ** on master
DT[, 1:ncol(DT)]
# [1] 1 2
```

At a glance, we need to be more selective & make sure `all(all.vars(jsub) %in% names(x))` before switching to `with=TRUE`.

MichaelChirico:
#6487 solved everything, nice 😎

jangorecki:
@MichaelChirico If this fix is also going to be in 1.16.2 (I haven't see it merged to patch branch, but feels like it fits there, no?) then we should assign it to that milestone instead.

MichaelChirico:
no, original breakage #6487 was only in dev

PR #6487, opened 2024-09-08
Retain with=FALSE behavior in more j=call:call cases like 1:ncol(DT)

MichaelChirico:
#6486 -- not closing, I haven't verified this fixes every case. I'll wait for revdeps report to formally close.

I removed these tests `DT[1L, t:max(t)]` + `DT[1L, min(t):t]` because this change broke them, and it's not clear to me such usage is practical anyway (it only makes sense on a 1-row subset). We can revisit later IMO.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,7 +55,7 @@ rowwiseDT(
     # [1] "V1" "b" "c"
     ```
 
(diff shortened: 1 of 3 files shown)


jangorecki (commented):
Yes this removed tests doesn't look very useful.
When moving to 2.0 we should drop a lot of support in those NSE tricks IMO. Code was already pretty complex and it only getting worse now.
But +1 for better escaping NSE with with=F

PR #6489, opened 2024-09-09
Remove targetDesc helper in assign.c in favor of splitting cases for i18n

MichaelChirico:
Gave up on the grammatical case-switching approach suggested here:

https://github.com/Rdatatable/data.table/issues/6482#issuecomment-2334652431

In the end dropping `targetDesc()` altogether means about a dozen extra lines, I think the added clarity is worth it. As seen here, it also unconstrains us from shoehorning both branches into the same format, potentially adding clarity to the message.


tdhock (approved):
looks fine. i wonder if it is worth adding a comment about the history with targetDesc? (ok if you dont think it is necessary)

MichaelChirico:
> adding a comment about the history with targetDesc

In the code somewhere? (as a comment on the line where it used to be declared?) or here in the PR?

MichaelChirico:
I don't think a code comment is needed (it was not a particularly important part of the code, no "historic" relevance, and git is here for future archeologists), and we can edit the PR if needed later; submitting now.

PR #6492, opened 2024-09-10
Fixed inconsistencies in .ci/atime/tests.R 

Anirban166:
> I noticed some minor inconsistencies in the file, for e.g. double `expr` from #6288: (also, using curly braces for only multiple lines might be good for consistency)
> 
> https://github.com/Rdatatable/data.table/blob/687de9d09bacde0d349c2d07222da2514ca39803/.ci/atime/tests.R#L137-L139
> 
>
_Originally posted by @Anirban166 in [#6464 (comment)](https://github.com/Rdatatable/data.table/pull/6464#issuecomment-2329913140)_

Also, some comments needed corrections, such as:

https://github.com/Rdatatable/data.table/blob/365d2a35cdb0bf251456d5ab231d750b9c9ad759/.ci/atime/tests.R#L124

(third-last commit, not last)

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -78,7 +78,7 @@ test.list <- atime::atime_test_list(
   # Performance regression discussed in: https://github.com/Rdatatable/data.table/issues/4311
   # Fixed in: https://github.com/Rdatatable/data.table/pull/4440
   "shallow regression fixed in #4440" = atime::atime_test(
-    N = 10^seq(3,8),
+    N = 10^seq(3, 8),
     setup = {
       set.seed(1L)
       dt <- data.table(a = sample.int(N))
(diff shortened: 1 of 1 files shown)


Anirban166:
@MichaelChirico @tdhock Seems like the setup time decrease is worthwhile/significant! :D
(This is showing results that are consistent with what [I'm getting on my fork now](https://github.com/Anirban166/data.table/pull/18#issuecomment-2339404653), for e.g. the first run there was v1.2.0 taking ~ 12 minutes for setup vs v1.2.1 now taking just 3 and a half minutes)

.ci/atime/tests.R
```diff
@@ -120,23 +120,21 @@ test.list <- atime::atime_test_list(
       data.table:::setattr(L, "class", NULL)
       data.table:::setDT(L)
     },
-    Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801)
-    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"), # Last commit in the PR that fixes the issue (https://github.com/Rdatatable/data.table/pull/5427/commits)
+    Slow = "c4a2085e35689a108d67dacb2f8261e4964d7e12", # Parent of the first commit (https://github.com/Rdatatable/data.table/commit/7cc4da4c1c8e568f655ab5167922dcdb75953801) in the PR (https://github.com/Rdatatable/data.table/pull/5427/commits) that fixes the issue
+    Fast = "1872f473b20fdcddc5c1b35d79fe9229cd9a1d15"), # Antepenultimate commit in the PR (https://github.com/Rdatatable/data.table/pull/5427/commits) that fixes the issue
```
tdhock, line 124: I don't understand Antepenultimate, why did you change from the simpler "last" ?

Anirban166: It's not really the last commit (you can check the link and verify), but I agree that it's better to avoid such a jargon - will 'third-last' work? (since that seems to be the case)

Anirban166: The SHA is not associated with the last commit (you can check the link and verify), but I agree that it's better to avoid such a jargon - will 'third-last' or 'close-to-last' work?

MichaelChirico: Is there something wrong with the last SHA in the PR, then just "last commit" will be correct? otherwise, third-to-last seems best to me (as much as I hate wasting such a perfectly valid word!)

Anirban166: Great point! No issues with the SHA, just didn't get updated at the end of #6094 unfortunately (I should've tried it with the last one before changing the comment)

Thank you both!

PR #6502, opened 2024-09-16
Don't attempt reuseSorting on non-data.table in bmerge

### LINKED ISSUE #6501: Internal error trying to use "lazy" indexing on non-data.table in anti-join i
Wordy title, simple example:

```r
x = data.table(id = "aaa")
y = data.frame(id = "bbb")
class(y) = c("tbl_df", "tbl", "data.frame")
x[!y, on = "id"]
#   Internal error in forderReuseSorting: reuseSorting set to TRUE but DT is not a data.table. Please report to the data.table issues tracker.
```

No error on 1.15.4. No error if `y` is just plain `data.frame`.

It's the `bmerge()` code here that tries to call the new routine on the non-data.frame:

https://github.com/Rdatatable/data.table/blob/8f505d58b01f9b1ce90ee2e67b1e3819a39162f4/src/bmerge.c#L163

MichaelChirico:
Closes #6501.

I still don't see why having an extended class like `tbl_df` is necessary, i.e. why plain `data.frame` doesn't wind up here. Maybe Jan sees a more robust diagnosis.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -69,6 +69,8 @@ rowwiseDT(
 
 11. `fread()` automatically detects timestamps with sub-second accuracy again, [#6440](https://github.com/Rdatatable/data.table/issues/6440). This was a regression due to interference with new `dec='auto'` support. Thanks @kav2k for the concise report and @MichaelChirico for the fix.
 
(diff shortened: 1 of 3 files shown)


tdhock (approved):
great thanks

jangorecki:
Interesting one, I don't have any better fix for now, we should probably start with understanding the issue better to figure out better fix.

Issue #6503, opened 2024-09-17
Multiple small issues with messages in the C code


aitap:
This is a continuation of #6482. I think I won't find any more issues in the C code messages. Some of these are questions, a few indicate real translation hurdles.

* This one is actually possible for me to translate in parts (so we can keep it as is if needed), but I don't like translating individual words and relying on the rest of the code not to need a different translation:
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/fread.c#L1933-L1934
Translation in parts could be done reliably with [`pgettext`](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html), which seems to be currently unavailable even to C code in R's built-in copy of `gettext` on Windows.
* This one is also translatable in context, but isn't too painful to recombine into a single sentence:
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/fsort.c#L255-L259
* This one is very awkward to translate, and we're again relying on `_("no")` not to mean something else in a different part of the code:
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/fread.c#L2038
Unfortunately, expanding this sentence will duplicate the full line of the code.

The rest are questions unrelated to translations:

* I think we shouldn't be calling it ASCII if its `/` is not immediately before `0`. How about "The character '/' is not just before '0' in the [source character set](https://port70.net/%7Ensz/c/c11/n1570.html#5.2.1)"?
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/init.c#L225
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/init.c#L227
* This one probably used to be correct in the past, but now should refer to a different variable, `.Last.updated`:
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/init.c#L367
* This message is technically correct because the code does expect only symbols at this point, but `data.table:::list2lang` converts character strings to symbols. Would it be more helpful to mention character strings in the error message too?
https://github.com/Rdatatable/data.table/blob/944340901d5b37cfdac43c9cc5b8695c5c12f5a0/src/programming.c#L16
* Should this mention the issue tracker instead?
[...]

PR #6504, opened 2024-09-17
Message fixes for the C code

aitap:
This solves some of the obvious issues in #6503. I can expand on the more tough or less obvious ones when we decide which ones to fix and how.

- [ ] regenerate `po/data.table.pot`

### DIFF
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -63,7 +63,8 @@ SEXP whichwrapper(SEXP x, SEXP val) {
 
 static const char *concat(SEXP vec, SEXP idx) {
   if (!isString(vec)) error(_("concat: 'vec' must be a character vector"));
-  if (!isInteger(idx)) error(_("concat: 'idx' must be an integer vector"));
+  if (!isInteger(idx))
+    error(_("concat: 'idx' must be an integer vector"));
 
   static char ans[1024];  // so only one call to concat() per calling warning/error
   int nidx=length(idx), nvec=length(vec);
@@ -802,7 +803,8 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP 
   }
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
-  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to the data.table issue tracker"));
+  if (isNull(dtnames))
+    error(_("names(data) is NULL. Please report to the data.table issue tracker"));
   if (LOGICAL(narmArg)[0] == TRUE) narm = TRUE;
   if (LOGICAL(verboseArg)[0] == TRUE) verbose = TRUE;
   struct processData data;
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -804,7 +804,7 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP 
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
   if (isNull(dtnames))
-    error(_("names(data) is NULL. Please report to the data.table issue tracker"));
+    internal_error("names(data) is NULL");
   if (LOGICAL(narmArg)[0] == TRUE) narm = TRUE;
   if (LOGICAL(verboseArg)[0] == TRUE) verbose = TRUE;
   struct processData data;
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -802,7 +802,7 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP 
   }
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
-  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to data.table-help"));
+  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to the data.table issue tracker"));
   if (LOGICAL(narmArg)[0] == TRUE) narm = TRUE;
   if (LOGICAL(verboseArg)[0] == TRUE) verbose = TRUE;
   struct processData data;
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -804,7 +804,7 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP 
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
   if (isNull(dtnames))
-    internal_error("names(data) is NULL");
+    internal_error(__func__, "names(data) is NULL");
   if (LOGICAL(narmArg)[0] == TRUE) narm = TRUE;
   if (LOGICAL(verboseArg)[0] == TRUE) verbose = TRUE;
   struct processData data;
--- a/src/fmelt.c
+++ b/src/fmelt.c
@@ -63,7 +63,7 @@ SEXP whichwrapper(SEXP x, SEXP val) {
 
 static const char *concat(SEXP vec, SEXP idx) {
   if (!isString(vec)) error(_("concat: 'vec' must be a character vector"));
-  if (!isInteger(idx) || length(idx) < 0) error(_("concat: 'idx' must be an integer vector of length >= 0"));
+  if (!isInteger(idx)) error(_("concat: 'idx' must be an integer vector"));
 
   static char ans[1024];  // so only one call to concat() per calling warning/error
   int nidx=length(idx), nvec=length(vec);
--- a/src/fread.c
+++ b/src/fread.c
@@ -2036,9 +2036,9 @@ int freadMain(freadMainArgs _args) {
     if (verbose) {
       if (sampleLines==0) {
         if (args.header)
-          DTPRINT(_("  'header' determined to be true because there are no number fields in the first and only row\n"));
+          DTPRINT(_("  'header' determined to be TRUE because there are no number fields in the first and only row\n"));
         else
-          DTPRINT(_("  'header' determined to be false because there are number fields in the first and only row\n"));
+          DTPRINT(_("  'header' determined to be FALSE because there are number fields in the first and only row\n"));
       } else {
         if (args.header)
           DTPRINT(_("  'header' determined to be true because all columns are type string and a better guess is not possible\n"));
--- a/src/fread.c
+++ b/src/fread.c
@@ -2035,7 +2035,10 @@ int freadMain(freadMainArgs _args) {
     }
     if (verbose) {
       if (sampleLines==0) {
-        DTPRINT(_("  'header' determined to be %s because there are%s number fields in the first and only row\n"), args.header?"TRUE":"FALSE", args.header?_(" no"):"");
+        if (args.header)
+          DTPRINT(_("  'header' determined to be %s because there are no number fields in the first and only row\n"), args.header?"TRUE":"FALSE");
+        else
+          DTPRINT(_("  'header' determined to be %s because there are number fields in the first and only row\n"), args.header?"TRUE":"FALSE");
       } else {
         if (args.header)
           DTPRINT(_("  'header' determined to be true because all columns are type string and a better guess is not possible\n"));
--- a/src/fread.c
+++ b/src/fread.c
@@ -2626,7 +2626,7 @@ int freadMain(freadMainArgs _args) {
 
   if (stopTeam) {
     if (internalErr[0]!='\0') {
-      STOP("%s %s: %s. %s", _("Internal error in"), __func__, internalErr, _("Please report to the data.table issues tracker"));  // # nocov
+      STOP(_("Internal error in %s: %s. Please report to the data.table issues tracker"), __func__, internalErr); // # nocov
     }
     stopTeam = false;
 
--- a/src/fread.c
+++ b/src/fread.c
@@ -2036,9 +2036,9 @@ int freadMain(freadMainArgs _args) {
     if (verbose) {
       if (sampleLines==0) {
         if (args.header)
-          DTPRINT(_("  'header' determined to be %s because there are no number fields in the first and only row\n"), args.header?"TRUE":"FALSE");
+          DTPRINT(_("  'header' determined to be true because there are no number fields in the first and only row\n"));
         else
-          DTPRINT(_("  'header' determined to be %s because there are number fields in the first and only row\n"), args.header?"TRUE":"FALSE");
+          DTPRINT(_("  'header' determined to be false because there are number fields in the first and only row\n"));
       } else {
         if (args.header)
           DTPRINT(_("  'header' determined to be true because all columns are type string and a better guess is not possible\n"));
--- a/src/fread.c
+++ b/src/fread.c
@@ -1930,8 +1930,10 @@ int freadMain(freadMainArgs _args) {
       }
       if ( (thisNcol<ncol && ncol>1 && !fill) ||
            (!eol(&ch) && ch!=eof) ) {
-        if (verbose) DTPRINT(_("  A line with too-%s fields (%d/%d) was found on line %d of sample jump %d. %s\n"),
-                             thisNcol<ncol ? _("few") : _("many"), thisNcol, ncol, jumpLine, jump, jump>0 ? _("Most likely this jump landed awkwardly so type bumps here will be skipped.") : "");
+        if (verbose)
+          DTPRINT(thisNcol<ncol ? _("  A line with too-few fields (%d/%d) was found on line %d of sample jump %d. %s\n")
+                                : _("  A line with too-many fields (%d/%d) was found on line %d of sample jump %d. %s\n"),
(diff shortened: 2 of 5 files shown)


src/fread.c
```diff
@@ -2035,7 +2035,10 @@ int freadMain(freadMainArgs _args) {
     }
     if (verbose) {
       if (sampleLines==0) {
-        DTPRINT(_("  'header' determined to be %s because there are%s number fields in the first and only row\n"), args.header?"TRUE":"FALSE", args.header?_(" no"):"");
+        if (args.header)
```
MichaelChirico, line 2038: Looks odd to me to branch on `args.header` and then template on `args.header ? T : F`. Let's just use `TRUE` / `FALSE` directly.

src/fsort.c
```diff
@@ -252,11 +252,13 @@ SEXP fsort(SEXP x, SEXP verboseArg) {
 
     if (verbose) {
       Rprintf(_("Top 20 MSB counts: ")); for(int i=0; i<MIN(MSBsize,20); i++) Rprintf(_("%"PRId64" "), (int64_t)msbCounts[order[i]]); Rprintf(_("\n"));
-      Rprintf(_("Reduced MSBsize from %zu to "), MSBsize);
     }
-    while (MSBsize>0 && msbCounts[order[MSBsize-1]] < 2) MSBsize--;
-    if (verbose) {
-      Rprintf(_("%zu by excluding 0 and 1 counts\n"), MSBsize);
+    {
```
MichaelChirico, line 256: what does a "naked" braced statement like this do in C? Is it just restricting the scope of `oldMSBsize`? Does the variable then get freed right afterward?

MichaelChirico (commented):
Thanks! A few small questions. Fixes basically look good.

src/fsort.c
aitap: Exactly, this is to limit the scope of `oldMSBsize`. In particular, changes to `MSBsize` still propagate because it's declared in a different scope. A reasonable C compiler would probably adjust the stack pointer after the closing brace to make `oldMSBsize` disappear.

src/fread.c
```diff
@@ -2035,7 +2035,10 @@ int freadMain(freadMainArgs _args) {
...
+          DTPRINT(_("  'header' determined to be true because there are no number fields in the first and only row\n"));
+        else
+          DTPRINT(_("  'header' determined to be false because there are number fields in the first and only row\n"));
```
MichaelChirico, line 2041:
```suggestion
          DTPRINT(_("  'header' determined to be TRUE because there are no number fields in the first and only row\n"));
        else
          DTPRINT(_("  'header' determined to be FALSE because there are number fields in the first and only row\n"));
```

src/fmelt.c
```diff
@@ -63,7 +63,7 @@ SEXP whichwrapper(SEXP x, SEXP val) {
 
 static const char *concat(SEXP vec, SEXP idx) {
   if (!isString(vec)) error(_("concat: 'vec' must be a character vector"));
-  if (!isInteger(idx) || length(idx) < 0) error(_("concat: 'idx' must be an integer vector of length >= 0"));
+  if (!isInteger(idx)) error(_("concat: 'idx' must be an integer vector"));
```
MichaelChirico, line 66:
```suggestion
  if (!isInteger(idx))
    error(_("concat: 'idx' must be an integer vector"));
```

src/fmelt.c
```diff
@@ -802,7 +802,7 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP
   }
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
-  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to data.table-help"));
+  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to the data.table issue tracker"));
```
MichaelChirico, line 805:
```suggestion
  if (isNull(dtnames))
    error(_("names(data) is NULL. Please report to the data.table issue tracker"));
```

src/fmelt.c
```diff
@@ -802,7 +803,8 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP
   }
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
-  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to data.table-help"));
+  if (isNull(dtnames))
+    error(_("names(data) is NULL. Please report to the data.table issue tracker"));
```
MichaelChirico, line 807: actually this should probably just be

```suggestion
    internal_error("names(data) is NULL");
```

src/fsort.c
```diff
@@ -252,11 +252,13 @@ SEXP fsort(SEXP x, SEXP verboseArg) {
...
+      size_t oldMSBsize = MSBsize;
+      while (MSBsize>0 && msbCounts[order[MSBsize-1]] < 2) MSBsize--;
+      if (verbose) {
+        Rprintf(_("Reduced MSBsize from %zu to %zu by excluding 0 and 1 counts\n"), oldMSBsize, MSBsize);
```
MichaelChirico, line 260: is it possible `oldMSBsize == MSBsize`? If so, maybe we also branch here on `verbose && oldMSBsize != MSBsize`?

WDYT of using a helper here too, `void shrinkMSB(MSBsize, msbCounts, order, verbose)`, that way we don't need to add an indent level to get the restricted scope of `oldMSBsize` too?

src/fread.c
```diff
@@ -2035,7 +2035,10 @@ int freadMain(freadMainArgs _args) {
     }
     if (verbose) {
       if (sampleLines==0) {
-        DTPRINT(_("  'header' determined to be %s because there are%s number fields in the first and only row\n"), args.header?"TRUE":"FALSE", args.header?_(" no"):"");
+        if (args.header)
+          DTPRINT(_("  'header' determined to be TRUE because there are no number fields in the first and only row\n"));
+        else
+          DTPRINT(_("  'header' determined to be FALSE because there are number fields in the first and only row\n"));
```
Anirban166, line 2041: Curious to know if the extra space before 'header' is required for something

src/fmelt.c
```diff
@@ -802,7 +803,8 @@ SEXP fmelt(SEXP DT, SEXP id, SEXP measure, SEXP varfactor, SEXP valfactor, SEXP
   }
   int protecti=0;
   dtnames = PROTECT(getAttrib(DT, R_NamesSymbol)); protecti++;
-  if (isNull(dtnames)) error(_("names(data) is NULL. Please report to data.table-help"));
+  if (isNull(dtnames))
+    internal_error("names(data) is NULL");
```
MichaelChirico, line 807:
```suggestion
    internal_error(__func__, "names(data) is NULL");
```

src/fread.c
MichaelChirico: Not "required", but it's consistent; `verbose=TRUE` output uses "heading/indented" to organize the large amount of output:

```
1301:    if (verbose) DTPRINT(_("  Using %d threads (omp_get_max_threads()=%d, nth=%d)\n"), nth, maxth, args.nth);
1341:      DTPRINT(_("  No NAstrings provided.\n"));
1343:      DTPRINT(_("  NAstrings = ["));
1348:        DTPRINT(_("  One or more of the NAstrings looks like a number.\n"));
1350:        DTPRINT(_("  None of the NAstrings look like numbers.\n"));
1352:    if (args.skipNrow >= 0) DTPRINT(_("  skip num lines = %"PRId64"\n"), (int64_t)args.skipNrow);
1353:    if (args.skipString) DTPRINT(_("  skip to string = <<%s>>\n"), args.skipString);
1354:    DTPRINT(_("  show progress = %d\n"), args.showProgress);
1355:    DTPRINT(_("  0/1 column will be read as %s\n"), args.logical01? "boolean" : "integer");
1389:    if (verbose) DTPRINT(_("  `input` argument is provided rather than a file name, interpreting as raw text to read\n"));
1396:    if (verbose) DTPRINT(_("  Opening file %s\n"), args.filename);
1408:      if (verbose) DTPRINT(_("  File opened, size = %s.\n"), filesize_to_str(fileSize));
1440:      if (verbose) DTPRINT(_("  File opened, size = %s.\n"), filesize_to_str(fileSize));
1453:    if (verbose) DTPRINT(_("  Memory mapped ok\n"));
1476:    if (verbose) DTPRINT(_("  UTF-8 byte order mark EF BB BF found at the start of the file and skipped.\n"));
1489:    if (verbose) DTPRINT(_("  Last byte(s) of input found to be %s and removed.\n"),
1506:    _("  No \\n exists in the file at all, so single \\r (if any) will be taken as one line ending. This is unusual but will happen normally when there is no \\r either; e.g. a single line missing its end of line.\n") :
1507:    _("  \\n has been found in the input and different lines can end with different line endings (e.g. mixed \\n and \\r\\n in one file). This is common and ideal.\n"));
1531:        if (verbose) DTPRINT(_("  File ends abruptly with '%c'. Final end-of-line is missing. Using cow page to write 0 to the last byte.\n"), eof[-1]);
1539:          DTPRINT(_("  File ends abruptly with '%c'. Final end-of-line is missing. Copying file in RAM. %s.\n"), eof[-1], msg);
1548:          DTPRINT(_("  File copy in RAM took %.3f seconds.\n"), time_taken);
1608:      DTPRINT(_("  Skipped to line %"PRIu64" in the file"), (uint64_t)row1line);
1624:    if (lineStart>ch) DTPRINT(_("  Moved forward to first non-blank line (%d)\n"), row1line);
1625:    DTPRINT(_("  Positioned on line %d starting: <<%s>>\n"), row1line, strlim(lineStart, 30));
1651:    if (verbose) DTPRINT(_("  sep='\\n' passed in meaning read lines as single character column\n"));
1670:      if (verbose) DTPRINT(_("  Detecting sep automatically ...\n"));
1677:      if (verbose) DTPRINT(_("  Using supplied sep '%s'\n"), args.sep=='\t' ? "\\t" : seps);
1692:        // if (verbose) DTPRINT(_("  Trying sep='%c' with quoteRule %d ...\n"), sep, quoteRule);
1710:              DTPRINT((unsigned)sep<32 ? "  sep=%#02x" : "  sep='%c'", sep);
1711:              DTPRINT(_("  with %d fields using quote rule %d\n"), topNumFields, quoteRule);
1760:              DTPRINT((unsigned)sep<32 ? "  sep=%#02x" : "  sep='%c'", sep);
1761:              DTPRINT(_("  with %d lines of %d fields using quote rule %d\n"), topNumLines, topNumFields, topQuoteRule);
1768:      if (verbose) DTPRINT(_("  No sep and quote rule found a block of 2x2 or greater. Single column input.\n"));
1810:    DTPRINT(_("  Detected %d columns on line %d. This line is either column names or first data row. Line starts as: <<%s>>\n"),
1812:    DTPRINT(_("  Quote rule picked = %d\n"), quoteRule);
1813:    DTPRINT(_("  fill=%s and the most number of columns found is %d\n"), fill?"true":"false", ncol);
1821:        DTPRINT(_("  Copying file in RAM. %s\n"), msg);
1830:        DTPRINT(_("  File copy in RAM took %.3f seconds.\n"), time_taken);
1836:      if (verbose) DTPRINT(_("  1-column file ends with 2 or more end-of-line. Restoring last eol using extra byte in cow page.\n"));
1856:  if (verbose && args.header!=NA_BOOL8) DTPRINT(_("  'header' changed by user from 'auto' to %s\n"), args.header?"true":"false");
1867:      DTPRINT(_("  sep=',' so dec set to '.'\n"));
1891:    DTPRINT(_("  Number of sampling jump points = %d because "), nJumps);
1933:        if (verbose) DTPRINT(_("  A line with too-%s fields (%d/%d) was found on line %d of sample jump %d. %s\n"),
1957:        DTPRINT(_("  dec='%c' detected based on a balance of %d parsed fields\n"), dec, abs(linesForDecDot));
1967:      DTPRINT(_("  Type codes (jump %03d)    : %s  Quote rule %d\n"), jump, typesAsString(ncol), quoteRule);
1983:        DTPRINT(_("  dec='%c' detected based on a balance of %d parsed fields\n"), dec, abs(linesForDecDot));
1989:        if (verbose) DTPRINT(_("  'header' determined to be true due to column %d containing a string on row 1 and a lower type (%s) in the rest of the %"PRId64" sample rows\n"),
2038:        DTPRINT(_("  'header' determined to be %s because there are%s number fields in the first and only row\n"), args.header?"TRUE":"FALSE", args.header?_(" no"):"");
2041:          DTPRINT(_("  'header' determined to be true because all columns are type string and a better guess is not possible\n"));
2043:          DTPRINT(_("  'header' determined to be false because there are some number columns and those columns do not have a string field at the top of them\n"));
2059:    if (verbose && bumped) DTPRINT(_("  Type codes (first row)   : %s  Quote rule %d\n"), typesAsString(ncol), quoteRule);
2068:    if (verbose) DTPRINT(_("  All rows were sampled since file is small so we know nrow=%"PRIu64" exactly\n"), (uint64_t)sampleLines);
2080:      DTPRINT(_("  =====\n"));
2081:      DTPRINT(_("  Sampled %"PRIu64" rows (handled \\n inside quoted fields) at %d jump points\n"), (uint64_t)sampleLines, nJumps);
2082:      DTPRINT(_("  Bytes from first data row on line %d to the end of last row: %"PRIu64"\n"), row1line, (uint64_t)bytesRead);
2083:      DTPRINT(_("  Line length: mean=%.2f sd=%.2f min=%d max=%d\n"), meanLineLen, sd, minLen, maxLen);
2084:      DTPRINT(_("  Estimated number of rows: %"PRIu64" / %.2f = %"PRIu64"\n"), (uint64_t)bytesRead, meanLineLen, (uint64_t)estnrow);
2085:      DTPRINT(_("  Initial alloc = %"PRIu64" rows (%"PRIu64" + %d%%) using bytes/max(mean-2*sd,min) clamped between [1.1*estn, 2.0*estn]\n"),
2087:      DTPRINT(_("  =====\n"));
2093:    if (verbose) DTPRINT(_("  Alloc limited to lower nrows=%"PRIu64" passed in.\n"), (uint64_t)nrowLimit);
2158:    if (verbose) DTPRINT(_("  Cancelled by user: userOverride() returned false.")); // # nocov
2191:  if (verbose) DTPRINT(_("  After %d type and %d drop user overrides : %s\n"), nUserBumped, ndrop, typesAsString(ncol));
2200:    DTPRINT(_("  Allocating %d column slots (%d - %d dropped) with %"PRIu64" rows\n"),
2261:  if (verbose) DTPRINT(_("  jumps=[%d..%d), chunk_size=%"PRIu64", total_size=%"PRIu64"\n"),
2605:      DTPRINT(_("  Provided number of fill columns: %d but only found %d\n"), ncol, max_col);
2606:      DTPRINT(_("  Dropping %d overallocated columns\n"), ndropFill);
2631:      if (verbose) DTPRINT(_("  Too few rows allocated. Allocating additional %"PRIu64" rows (now nrows=%"PRIu64") and continue reading from jump %d\n"),
2638:      if (verbose) DTPRINT(_("  Restarting team from jump %d. nSwept==%d quoteRule==%d\n"), jump0, nSwept, quoteRule);
2658:      if (verbose) DTPRINT(_("  %d out-of-sample type bumps: %s\n"), nTypeBump, typesAsString(ncol));
2701:    DTPRINT(_("  Type counts:\n"));
```

Anirban166: Noted!

src/fsort.c
aitap: Made it into a helper in [014a3ae](https://github.com/Rdatatable/data.table/pull/6504/commits/014a3aea423d9d799e33d697b9030ac53eccf9c4).

MichaelChirico:
Don't regenerate the .pot file in this PR, save it for a follow-up

src/fsort.c
```diff
@@ -98,6 +98,15 @@ int qsort_cmp(const void *a, const void *b) {
   return (x<y)-(x>y);   // largest first in a safe branchless way casting long to int
 }
 
+static size_t shrinkMSB(size_t MSBsize, uint64_t *msbCounts, int *order, Rboolean verbose) {
+  size_t oldMSBsize = MSBsize;
+  while (MSBsize>0 && msbCounts[order[MSBsize-1]] < 2) MSBsize--;
```
MichaelChirico, line 103:
```suggestion
  while (MSBsize>0 && msbCounts[order[MSBsize-1]] < 2)
    MSBsize--;
```

src/fsort.c
```diff
@@ -252,12 +262,8 @@ SEXP fsort(SEXP x, SEXP verboseArg) {
 
     if (verbose) {
       Rprintf(_("Top 20 MSB counts: ")); for(int i=0; i<MIN(MSBsize,20); i++) Rprintf(_("%"PRId64" "), (int64_t)msbCounts[order[i]]); Rprintf(_("\n"));
-      Rprintf(_("Reduced MSBsize from %zu to "), MSBsize);
-    }
-    while (MSBsize>0 && msbCounts[order[MSBsize-1]] < 2) MSBsize--;
-    if (verbose) {
-      Rprintf(_("%zu by excluding 0 and 1 counts\n"), MSBsize);
     }
+    MSBsize = shrinkMSB(MSBsize, msbCounts, order, verbose);
```
MichaelChirico, line 266: I might have assumed this helper would be by-reference on `MSBsize`; OTOH, it's a single `size_t` and this is only done once in `fsort()`. The impact should not be noticed.

MichaelChirico (approved):
Great work + progress! I think this is ready to merge. We can always do more PRs with more fixes :)

Issue #6498, opened 2024-09-15
`patterns()` did not consider the position of provided  `cols` in the original DT


hongyuanjia:
Please see the MRE below. `patterns()` returns the matched positions of input `cols,` which are `1` and `2`, resulting using `id_1` and `id_2` as the `measure.vars`:

``` r
library(data.table)

dt <- data.table(
    id_1 = 1:3,
    id_2 = letters[1:3],
    letter_1 = LETTERS[1:3],
    letter_2 = LETTERS[4:6]
)

melt(dt, measure.vars = patterns("_\\d$", cols = paste0("letter_", 1:2)))
#> Warning in melt.data.table(dt, measure.vars = patterns("_\\d$", cols =
#> paste0("letter_", : 'measure.vars' [id_1, id_2] are not all of the same type.
#> By order of hierarchy, the molten data value column will be of type
#> 'character'. All measure variables not of type 'character' will be coerced too.
#> Check DETAILS in ?melt.data.table for more on coercion.
#>    letter_1 letter_2 variable  value
#>      <char>   <char>   <fctr> <char>
#> 1:        A        D     id_1      1
#> 2:        B        E     id_1      2
#> 3:        C        F     id_1      3
#> 4:        A        D     id_2      a
#> 5:        B        E     id_2      b
#> 6:        C        F     id_2      c
```

Anirban166:
Can confirm the MRE, but I wonder if it's by intended design that `patterns()` is the primary driver of column selection, while `cols` just acts as a filter without getting precedence?

Anirban166:
Can confirm the MRE, but I wonder if it's by intended design that `patterns()` is the primary driver of column selection, while `cols` just acts as a filter on top of it without getting precedence?

tdhock:
or
```r
> melt(dt, measure.vars=measure(value.name, number=as.integer, pattern="(.*)_([12])"))
   number     id letter
    <int> <char> <char>
1:      1      1      A
2:      1      2      B
3:      1      3      C
4:      2      a      D
5:      2      b      E
6:      2      c      F
Message d'avis :
Dans melt.data.table(dt, measure.vars = measure(value.name, number = as.integer,  :
  'measure.vars' [id_1, id_2] are not all of the same type. By order of hierarchy, the molten data value column will be of type 'character'. All measure variables not of type 'character' will be coerced too. Check DETAILS in ?melt.data.table for more on coercion.
```

hongyuanjia:
[...]

PR #6508, opened 2024-09-20
two more links to ?measure

tdhock:
in response to #6498 I re-read ?melt and added a couple of references to ?measure


MichaelChirico (approved):
Thanks!

PR #6510, opened 2024-09-20
Add support for patterns(cols=user_provided)

### LINKED ISSUE #6498: `patterns()` did not consider the position of provided  `cols` in the original DT

tdhock:
Closes #6498

Before this PR:
```r
> melt(data.table(x1=1,x2=2,y1=3,y2=4),measure.vars=patterns("2",cols=c("y1","y2")))
      x1    y1    y2 variable value
   <num> <num> <num>   <fctr> <num>
1:     1     3     4       x2     2
```
above result is unexpected since variable=x2 which is not present in cols.

After this PR:
```r
> melt(data.table(x1=1,x2=2,y1=3,y2=4),measure.vars=patterns("2",cols=c("y1","y2")))
      x1    x2    y1 variable value
   <num> <num> <num>   <fctr> <num>
1:     1     2     3       y2     4
```
above result is better since y2 variable is present in cols.

this provides functionality similar to #6077


tdhock:
on my local windows machine, all tests pass.
this could be a "breaking change" since patterns now outputs character instead of integer. 
But when patterns() is used in the context of melt() or .SDcols, it gives the same result before/after this PR. (so no breakage for the typical use case)
And I don't think people should be using patterns() otherwise (although it is exported).

hongyuanjia:
Thanks @tdhock. I love this change. For me, and I believe most DT users, the primary use of patterns is to extract columns in melt and .SDcols, so no break changes in this aspect.

Anirban166 (approved):
LGTM!

NEWS.md
```diff
@@ -63,6 +63,8 @@ rowwiseDT(
     # 15: All values  Total  Total   999   999   NaN    10
     ```
 
+4. `patterns()` returns character instead of integer vector(s), to support user-defined `cols`, which can be useful to specify a subset of columns to reshape without having to use a regex, [#6498](https://github.com/Rdatatable/data.table/issues/6498). Thanks to @hongyuanjia for the report, and to @tdhock for the PR.
```
MichaelChirico, line 66: Since `patterns()` is kind-of "internal", I don't think we need to explain the difference of returning character vs. integer (it's sort of an implementation detail). I think it suffices to just explain the user-facing behavior change.

Implementation change could be described in PR description as a hint to code reviewer.
```suggestion
4. `patterns()` in `melt()` combines correctly with user-defined `cols=`, which can be useful to specify a subset of columns to reshape without having to use a regex, for example `patterns("2", cols=c("y1", "y2"))` will only give `y2` even if there are other columns in the input matching `2`, [#6498](https://github.com/Rdatatable/data.table/issues/6498). Thanks to @hongyuanjia for the report, and to @tdhock for the PR.
```

MichaelChirico (approved):

PR #6511, opened 2024-09-20
patterns() throws new informative error when cols is non-character or has NA

tdhock:
before this PR
```r
> melt(data.table(x1=1,x2=2,y1=3,y2=4),measure.vars=patterns("2",cols=NA_character_))
Error in (function (..., cols = character(0L), ignore.case = FALSE, perl = FALSE,  : 
  Pattern not found: [[2]]
```
above is not very informative for the user.

after this PR
```r
> melt(data.table(x1=1,x2=2,y1=3,y2=4),measure.vars=patterns("2",cols=NA_character_))
Error in (function (..., cols = character(0L), ignore.case = FALSE, perl = FALSE,  : 
  cols must be a character vector of column names
```
above is more helpful for the user, and consistent with the error message in measure().


R/fmelt.R
```diff
@@ -22,6 +22,9 @@ melt.default = function(data, ..., na.rm = FALSE, value.name = "value") {
 patterns = function(..., cols=character(0L), ignore.case=FALSE, perl=FALSE, fixed=FALSE, useBytes=FALSE) {
   # if ... has no names, names(list(...)) will be "";
   #   this assures they'll be NULL instead
+  if(!is.character(cols) || anyNA(cols)){
```
MichaelChirico, line 25:
```suggestion
  if (!is.character(cols) || anyNA(cols)) {
```

inst/tests/tests.Rraw
```diff
@@ -12428,6 +12428,10 @@ DTout = data.table(
 )
 test(1866.6, melt(DT, measure.vars = patterns("^x", "^y", cols=names(DT))), DTout)
 
+# informative errors for bad user-provided cols arg to patterns
+test(1866.80, melt(data.table(x1=1,x2=2,y1=3,y2=4),measure.vars=patterns("2",cols=NULL)), error="cols must be a character vector of column names")
```
MichaelChirico, line 12432: I think the test is easier to read if we factor out the repeated part so the reader can focus on what changes from .80 to .81

MichaelChirico (approved):
Great find & fix!

Anirban166 (approved):

tdhock:
thanks!

PR #6222, opened 2024-07-05
brazil and fr logos

tdhock:
related to #6221 and #6206 
I created logos for translation teams (which are displayed, for example, on https://github.com/orgs/Rdatatable/teams/french) by putting the DT hex sticker on top of flags.
This PR adds the source SVG files along with the PNG exports to the .graphics directory.

PR #6519, opened 2024-09-21
Adding translation team logos

jmrung:
I am providing both an .svg and a high resolution .png of translation team-specific data.table logos. The differences between the main logo and translation team-specific logos are the inclusion of a two- to four-letter language code as superscripted text at the end of the package name within the logo.  [Full code list and additional information on codes here](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes). I used lowercase letters since this seems most common with these codes.

I am not sure if more specific country-based versions of the language codes are needed for any of the teams (the four-letter versions of the codes, [see list here](https://www.andiamo.co.uk/resources/iso-language-codes/)). I've made one for the Brazilian team, based on recent discussions (the "pt-br" superscript), which provides a nice example of what this would look like.

Please let me know if any general changes are needed and if additional, country-specific language code versions would better serve any of the teams.

Related to #6222


aitap:
Do you think the logos could benefit from making the language code larger? I've [tried on](https://github.com/user-attachments/assets/52ab3934-7220-4741-9e6f-5a6ede97ec19) the avatar of the Brazilian Portuguese team by pasting the image URL into the Web developer toolbar, and the language code is just 9 pixels tall.

MichaelChirico:
I think @tdhock is most familiar with the discussion here. At a glance I wonder if the code would fit better centered underneath 'data.table' rather than superscripted next to it. Maybe a slight design difference could help distinguish it, e.g. italics or a slightly different font or color. Just a thought.

ben-schwen:
Isn't there a mismatch between the .png and .svg for FR?

CODEOWNERS
```diff
@@ -67,3 +67,6 @@
 
 # docs
 /man/openmp-utils.Rd @Anirban166
+
+# meta
+/.graphics/translation-team-logos/*.png @tdhock
```
tdhock, line 72: @jmrung could you please volunteer as reviewer under .graphics, since you created the new logos?

jmrung:
> Isn't there a mismatch between the .png and .svg for FR?

Good catch @ben-schwen! I missed changing the text for the language code to a path, which was causing the problem. This is fixed now in a recent commit. 

@aitap & @MichaelChirico Thank you for the suggestions! I'll play around with a few possibilities to help with the readability of the language codes and post again soon to get everyone's thoughts. Moving the language code outside of the hex might be the most effective since the space inside is pretty tight.

MichaelChirico:
Hi @jmrung any update here? Thanks for your time!

jmrung:
Hey @MichaelChirico, yes! 

How does everyone feel about this style? (full size)
datatable-pt_v2

And the approximate size as it would appear on translation team pages
datatable-pt2_resize

I can fiddle with the line properties a bit on the seal if needed so it stays crisper when downsizing, though the resized version shown here was scaled after saving to png (so may be an artifact of that).

aitap:
It's hard to make the logo recognisable in all cases. Consider the reviewers pane for #6767 with the Brazilian Portuguese logo added manually:
reviewers

For maximum legibility, I'd go with a large language code in `data.table` colours, something like [this](https://github.com/user-attachments/assets/5bc87651-1dcd-404d-964d-0f428337bbe8):
reviewers2

But I'm not a designer and this is still far from pretty.

Maybe uniformity is good. Perhaps it's more important to see that the pending reviewer is a `data.table` team than to see which team it is.

tdhock:
I like the idea of using the two letter language codes. that would fix the issue one person had with the flags (countries != languages)

MichaelChirico:
I like Ivan's suggestion. Obviously the data.table logo is mostly obscured, but that's not what's important in this case -- the language code is. We can still basically infer what's behind the letters.

Is it possible we'll ever put the regional codes on a separate team (e.g. `@Rdatatable/portugese-brazilian` vs `@Rdatatable/portugese-portugal`)? I would guess no, in which case we don't need to think about how it'd look with 5 characters.

jmrung:
Great context on the use and suggestion, @aitap. Thank you!

Full, ~120px square, and ~20px square:

datatable-pt_full_ns
datatable-pt_120sq_ns
datatable-ptbr_20sq_ns

The text can be increased a bit still if the readability leaves a bit to be desired at ~20px, though it seems okay at the current size (at least to me--other eyes needed!). I left a little breathing room for balance with seal visibility at the larger dimensions, for now.

MichaelChirico:
Hey @jmrung I think we are in agreement about the latest version. I think you just need to upload them into this branch, right?

jmrung:
Ah! I had been waiting on a verbal confirmation before making all the individual versions, but in hindsight realize I also didn’t state that. Apologies! I will prep and export the remaining and upload soon. I will try to get that done this weekend. Best,Jill

jmrung:
@MichaelChirico, I all the new ones are uploaded. 😊 I put them in a separate directory (translation-team-codelogos) so that the package logo look-a-likes are still retained in case something that has both the package name and the language reference is helpful.

If ultimately not needed, they can be deleted of course! New versions below, shown at ~200x200 and ~20x20

MichaelChirico:
Looks fine to me, awesome!

@tdhock could you weigh in on the file structure?

What is the procedure for adding new languages in the future?

MichaelChirico:
I'll go ahead and merge, we can refine later if needed. Thank you!!!

PR #4129, opened 2019-12-18
Adding plike

KyleHaynes:
Added %plike%. 

Closes #3702

PR #6521, opened 2024-09-21
Inconsistent code formatting / minor fixes to vignettes

KyleHaynes:
Noticed 1 inconsistent code formatting issue and a few broken links in RMarkdown Vignettes.


vignettes/datatable-importing.Rmd
```diff
@@ -138,7 +138,7 @@ The option mechanism in R is _global_. Meaning that if a user sets a `data.table
...
+Some of the most common issues developers are facing are usually related to helper tools that are meant to automate some package development tasks, for example, using `roxygen` to generate your `NAMESPACE` file from metadata in the R code files. Others are related to helpers that build and check the package. Unfortunately, these helpers sometimes have unintended/hidden side effects which can obscure the source of your troubles. As such, be sure to double check using R console (run R on the command line) and ensure the import is defined in the `DESCRIPTION` and `NAMESPACE` files following the [instructions](#description-file-description) [above](#namespace-file-namespace).
```
aitap, line 141: These sections were intended to have a certain class name:
https://github.com/Rdatatable/data.table/blob/ca904cb4c87257b38540b7c7cc24da802e44241a/vignettes/datatable-importing.Rmd#L30
https://github.com/Rdatatable/data.table/blob/ca904cb4c87257b38540b7c7cc24da802e44241a/vignettes/datatable-importing.Rmd#L34

It doesn't really work because the correct syntax would have been `{#DESCRIPTION}` and `{#NAMESPACE}`, respectively. It is important to have fixed `id` values for headings that are linked to because the auto-generated `id`s will change as a result of translation.

vignettes/datatable-reference-semantics.Rmd
```diff
@@ -68,7 +68,7 @@ DF$c <- 18:13               # (1) -- replace entire column
 DF$c[DF$ID == "b"] <- 15:13 # (2) -- subassign in column 'c'
 ```
 
-both (1) and (2) resulted in deep copy of the entire data.frame in versions of `R` versions `< 3.1`. [It copied more than once](https://stackoverflow.com/q/23898969/559784). To improve performance by avoiding these redundant copies, *data.table* utilised the [available but unused `:=` operator in R](https://stackoverflow.com/q/7033106/559784).
+both (1) and (2) resulted in deep copy of the entire data.frame in versions of `R` `< 3.1`. [It copied more than once](https://stackoverflow.com/q/23898969/559784). To improve performance by avoiding these redundant copies, *data.table* utilised the [available but unused `:=` operator in R](https://stackoverflow.com/q/7033106/559784).
```
Anirban166, line 71: `R` `< 3.1` probably should be combined to `R < 3.1` now that there is nothing between them

I'd personally not consider the version number to be something within backticks, but to maintain consistency with cases like `R 3.1+` below:
https://github.com/Rdatatable/data.table/blob/a673460dd83fa499edee71d7eee7fa1f1b5f3fa6/vignettes/datatable-reference-semantics.Rmd#L73

vignettes/datatable-importing.Rmd
MichaelChirico: seconded -- the `id`s are also better than the silly auto-generated names.

PS to clarify since I misread this comment: we need to fix the IDs to be `{#DESCRIPTION}`, not `{DESCRIPTION}`; the linking form `(#DESCRIPTION)` is fine:

https://www.markdownguide.org/extended-syntax/#heading-ids

vignettes/datatable-importing.Rmd
```diff
@@ -15,7 +15,7 @@ h2 {
 }
 </style>
 
-This document is focused on using `data.table` as a dependency in other R packages. If you are interested in using `data.table` C code from a non-R application, or in calling its C functions directly, jump to the [last section](#non-r-API) of this vignette.
+This document is focused on using `data.table` as a dependency in other R packages. If you are interested in using `data.table` C code from a non-R application, or in calling its C functions directly, jump to the [last section](#importing-from-non-r-applications-non-r-api) of this vignette.
```
MichaelChirico, line 18:
```suggestion
This document is focused on using `data.table` as a dependency in other R packages. If you are interested in using `data.table` C code from a non-R application, or in calling its C functions directly, jump to the [last section](#non-r-api) of this vignette.
```

vignettes/datatable-importing.Rmd
MichaelChirico, line 141:
```suggestion
Some of the most common issues developers are facing are usually related to helper tools that are meant to automate some package development tasks, for example, using `roxygen` to generate your `NAMESPACE` file from metadata in the R code files. Others are related to helpers that build and check the package. Unfortunately, these helpers sometimes have unintended/hidden side effects which can obscure the source of your troubles. As such, be sure to double check using R console (run R on the command line) and ensure the import is defined in the `DESCRIPTION` and `NAMESPACE` files following the [instructions](#DESCRIPTION) [above](#NAMESPACE).
```

vignettes/datatable-reference-semantics.Rmd
MichaelChirico, line 71:
```suggestion
both (1) and (2) resulted in deep copy of the entire data.frame in versions of `R < 3.1`. [It copied more than once](https://stackoverflow.com/q/23898969/559784). To improve performance by avoiding these redundant copies, *data.table* utilised the [available but unused `:=` operator in R](https://stackoverflow.com/q/7033106/559784).
```

vignettes/datatable-reshape.Rmd
```diff
@@ -157,7 +157,7 @@ DT.m1[, c("variable", "child") := tstrsplit(variable, "_", fixed = TRUE)]
 DT.c1 = dcast(DT.m1, family_id + age_mother + child ~ variable, value.var = "value")
 DT.c1
 
-str(DT.c1) ## gender column is character type now!
+str(DT.c1) ## gender column is character IDate now!
```
MichaelChirico, line 160:
```suggestion
str(DT.c1) ## gender column is class IDate now!
```

MichaelChirico: Actually this is an interesting point, this example was designed for when `dob_child1` etc. would be read by `fread()` as type `character`, but now they're recognized as dates and read as `IDate`.

Before this example illustrated `gender` being cast from `integer` to `character`, now it only goes `integer`->`integer`, but gains the class attribute. Maybe we should revisit this.

MichaelChirico (approved):

MichaelChirico:
Nice finds all! Thanks for filing.

Our rule is to add as Members anyone making _code_ contributions; this current PR (while certainly highly valued!) doesn't qualify (doc-only), however, your earlier #4129 does (and I see you were already added to DESCRIPTION then). Therefore I invited you to join as a member! That'll make things easier e.g. by allowing you to create branches directly on this repo. Thank you!

MichaelChirico (approved):

PR #6525, opened 2024-09-23
Add class1/classes1 helpers for common operation class(x)[1]

### LINKED ISSUE #6522: New internal class1 helper
We use `function(x) class(x)[1]` a few times, including one new time in #5973:

https://github.com/Rdatatable/data.table/blob/a673460dd83fa499edee71d7eee7fa1f1b5f3fa6/R/print.data.table.R#L106

https://github.com/Rdatatable/data.table/blob/a673460dd83fa499edee71d7eee7fa1f1b5f3fa6/R/test.data.table.R#L293

https://github.com/Rdatatable/data.table/pull/5973/files#diff-8bde12f7775c5541aa8cce5665991eb0f9e0c1cf94ad222a7d6889fcc9101909R79

Better to use a helper to wrap this, e.g. named `class1`.

<details>

Found with {lintr}:

```r
l=lintr::make_linter_from_xpath("//FUNCTION/following-sibling::expr[OP-LEFT-BRACKET]/expr/expr/SYMBOL_FUNCTION_CALL[text() = 'class']", "xxx")
lintr::lint_dir("R", linters=l())
```

</details>

MichaelChirico:
Closes #6522

### DIFF
--- a/R/fcast.R
+++ b/R/fcast.R
@@ -17,8 +17,8 @@ dcast <- function(
   else {
     data_name = deparse(substitute(data))
     ns = tryCatch(getNamespace("reshape2"), error=function(e)
-      stopf("The %1$s generic in data.table has been passed a %2$s, but data.table::%1$s currently only has a method for data.tables. Please confirm your input is a data.table, with setDT(%3$s) or as.data.table(%3$s). If you intend to use a method from reshape2, try installing that package first, but do note that reshape2 is superseded and is no longer actively developed.", "dcast", class(data)[1L], data_name))
-    warningf("The %1$s generic in data.table has been passed a %2$s and will attempt to redirect to the relevant reshape2 method; please note that reshape2 is superseded and is no longer actively developed, and this redirection is now deprecated. Please do this redirection yourself like reshape2::%1$s(%3$s). In the next version, this warning will become an error.", "dcast", class(data)[1L], data_name)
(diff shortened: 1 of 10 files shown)


tdhock (approved):
great thanks, what was your motivation? more easily readable?

MichaelChirico:
Yep -- readability, conciseness, consistency. Generally I think it's good practice to make such helpers for operations done 3+ times in the code base. We see here some small improvement from making `use.names=FALSE` the default too (faster + nice to make the `use.names=TRUE` behavior explicit where needed).

The downside would be if it's clearer to see the function body written out vs. infer what it does from the name; I think it's clear enough in this case.

R/fcast.R
```diff
@@ -17,8 +17,8 @@ dcast <- function(
...
+      stopf("The %1$s generic in data.table has been passed a %2$s, but data.table::%1$s currently only has a method for data.tables. Please confirm your input is a data.table, with setDT(%3$s) or as.data.table(%3$s). If you intend to use a method from reshape2, try installing that package first, but do note that reshape2 is superseded and is no longer actively developed.", "dcast", class1(data), data_name))
+    warningf("The %1$s generic in data.table has been passed a %2$s and will attempt to redirect to the relevant reshape2 method; please note that reshape2 is superseded and is no longer actively developed, and this redirection is now deprecated. Please do this redirection yourself like reshape2::%1$s(%3$s). In the next version, this warning will become an error.", "dcast", class1(data), data_name)
```
jangorecki, line 21: This two lines is not meant to be updated. We don't export the function

PR #6526, opened 2024-09-23
Added an atime test for performance improvement in melt (when measure.vars has NAs) brought by #5054

Anirban166:
Tested [here](https://github.com/Anirban166/data.table/pull/20)


tdhock:
great thanks

PR #6527, opened 2024-09-24
performance test cleanup: link original benchmarks, use common N

tdhock:
Hi!
While reviewing/adding a performance test recently, https://github.com/Rdatatable/data.table/pull/6526 I noticed that most other performance tests originated from issues/prs in which users reported `system.time` benchmarks, so this PR adds links to these comments.
And I updated the wiki to include instructions to do this for future performance tests: "Also if there is some issue/PR comment with benchmark code that was used to derive the atime test, please include a link, as above `test case adapted from https://github.com/Rdatatable/data.table/pull/4501#issue-625311918 which is the fix PR.`"

This PR also adds a common N which should be reasonable for any test, `N = 10^seq(1, 7, by=0.25),`

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -115,11 +115,11 @@ test.list <- atime::atime_test_list(
   # test case adapted from https://github.com/Rdatatable/data.table/pull/5463#issue-1373642456 which is the fix PR.
   "memrecycle regression fixed in #5463" = atime::atime_test(
     setup = {
-      n <- N/100
+      bigN <- N*100
       set.seed(2L)
       dt <- data.table(
-        g = sample(seq_len(n), N, TRUE),
-        x = runif(N),
+        g = sample(seq_len(N), bigN, TRUE),
+        x = runif(bigN),
         key = "g")
       dt_mod <- copy(dt)
     },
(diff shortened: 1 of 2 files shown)


tdhock:
this PR also shortens the name of the CI job to "atime performance tests" (previous name was too long to read status/time on PR page)

.ci/atime/tests.R
```diff
@@ -40,6 +40,7 @@ for (extra.arg in extra.args.6107){
 # @note Please check https://github.com/tdhock/atime/blob/main/vignettes/data.table.Rmd for more information.
 # nolint start: undesirable_operator_linter. ':::' needed+appropriate here.
 test.list <- atime::atime_test_list(
+  N = as.integer(10^seq(1, 7, by=0.25)),
```
MichaelChirico, line 43: WDYT about naming like `N_grid` or `commonN` to make it more obvious that this is a constant defined & re-used here? I sorry with just plain `N` that it might take a bit longer to catch on to that.

MichaelChirico (approved):
LGTM, feel free to submit now or after the suggested change!

.ci/atime/tests.R
tdhock: I'll add a comment

Anirban166 (approved):
LGTM!

tdhock:
thanks!

PR #6530, opened 2024-09-24
Refactor CHECK_RANGE to be i18n friendly

MichaelChirico:
Another part of #6503.

I don't see a way to avoid introducing the new `columnDesc()` helper + associated "sub-buffer"; but I never liked that we're running `snprintf()` with two "wasted" inputs when `colnum==0` previously anyway.

An approach like `snprintf(..., strcat(FMT, colnum==0?_(" (target vector)"):_(" column %d named '%s')")), ...)` was attempted, but we get the `-Wformat-security` issue because the formatting template is no longer a string literal -- it'd have to be `"%s (%s)"` and then another sub-buffer approach IINM, roughly equivalent to what's done here.

I also tried to only create the `column_desc` buffer inside `columnDesc` but that would require returning a pointer to a local variable IINM: https://stackoverflow.com/q/12380758/3576984, hence another `static char` array. I made it "large" to accommodate wide column names generously.

This refactor also exposed the issue where for an `Rcomplex` value was being passed to `%f`; for now I'm just using `val.r` and punting to follow-up to handle this more "properly": #6531.

I think this is in the same spirit as @aitap's comment: https://github.com/Rdatatable/data.table/issues/6503#issuecomment-2368523744

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -744,6 +744,12 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
 #define MSGSIZE 1000
 static char memrecycle_message[MSGSIZE+1]; // returned to rbindlist so it can prefix with which one of the list of data.table-like objects
 
+static char column_desc[MSGSIZE+1]; // can contain column names, hence relatively large allocation.
+const char *columnDesc(int colnum, const char *colname) {
+  snprintf(column_desc, MSGSIZE, _("column %d named '%s'"), colnum, colname);
+  return column_desc;
+}
+
 const char *memrecycle(const SEXP target, const SEXP where, const int start, const int len, SEXP source, const int sourceStart, const int sourceLen, const int colnum, const char *colname)
 // like memcpy but recycles single-item source
 // 'where' a 1-based INTEGER vector subset of target to assign to, or NULL or integer()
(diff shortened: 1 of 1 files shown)


src/assign.c
```diff
@@ -744,6 +744,12 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
 #define MSGSIZE 1000
 static char memrecycle_message[MSGSIZE+1]; // returned to rbindlist so it can prefix with which one of the list of data.table-like objects
 
+static char column_desc[MSGSIZE+1]; // can contain column names, hence relatively large allocation.
+const char *columnDesc(int colnum, const char *colname) {
+  snprintf(column_desc, MSGSIZE, _("column %d named '%s'"), colnum, colname);
```
aitap, line 749: Can we add a comment [to be extracted by `xgettext -cTRANSLATORS`](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html#index-_002dc_002c-xgettext-option) saying

```c
/* TRANSLATORS: use nominative case; string will be used inside parentheses */
```

src/assign.c
```diff
@@ -932,18 +938,16 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
...
           const char *tType = targetIsI64 ? "integer64" : type2char(TYPEOF(target));                                        \
-          snprintf(memrecycle_message, MSGSIZE,                                                                             \
-            colnum == 0                                                                                                     \
-            ? _("%"FMT" (type '%s') at RHS position %d "TO" when assigning to type '%s' (target vector)")                   \
-            : _("%"FMT" (type '%s') at RHS position %d "TO" when assigning to type '%s' (column %d named '%s')"),           \
-            FMTVAL, sType, i+1, tType, colnum, colname);                                                                    \
+          snprintf(memrecycle_message, MSGSIZE, FMT,                                                                        \
+            FMTVAL, sType, i+1, tType,                                                                                      \
+            colnum == 0 ? _("target vector") : columnDesc(colnum, colname));                                                \
```
aitap, line 950: And a similar comment about nominative case here. The comments should just work when put on the preceding line; `xgettext` doesn't care that `_("...")` is not on the start of the line.

aitap (commented):
Sorry about the back and forth between what was originally in the code and a lot of extra strings (and extra work for the translators).

aitap:
> I also tried to only create the column_desc buffer inside columnDesc but that would require returning a pointer to a local variable

The buffer could be declared as `static` inside `columnDesc` and thus be guaranteed to live on after the function returns. R's own `EncodeChar` [works in a similar way](https://github.com/r-devel/r-svn/blob/31340c871c7df54e45bfc7c4f49d09bb5806ec70/src/main/printutils.c#L868-L873) (except the `static` part points to a dynamically allocated buffer).

src/assign.c
MichaelChirico: WDYT about using ` %s` instead of ` (%s)` in the longer string, then `columnDesc` returns `(column %d names '%s')`, then the parenthetical won't need the comment?

MichaelChirico:
> The buffer could be declared as static inside columnDesc and thus be guaranteed to live on after the function returns.

Nice, done! CMIIW but this is preferable to a `malloc()` approach which would either need to intentionally leak memory or complicate the `CHECK_RANGE()`/surrounding logic to ensure the memory is `free()`'d. The approach here is more like an "intentional leak of a fixed, small amount of memory". The downside is that column names could be arbitrarily long; we might add a check for extremely long names but it seems like overkill until requested by a user.

src/assign.c
aitap: Makes perfect sense, this should be even better!

MichaelChirico:
> Sorry about the back and forth between what was originally in the code and a lot of extra strings (and extra work for the translators).

I think no problem at all! Most of the extra work is copy-pasting and they should all be one-after-the-other in the .po file. I view it as a big plus to translators who are empowered to make much better translations without having to awkwardly kludge their languages to fit what we provide.

If anything the downside is on the code maintainers, but I think it's perfectly manageable.

aitap (approved):

Issue #6532, opened 2024-09-24
Issues in translatable messages in R code


aitap:
### Fragmented sentences

* [x] `%s` is `elements of fun.list` or ...`arguments to measure` or maybe `measured columns`, untranslated, coming from the `group.desc` argument, which is a documented ("for internal use") interface. Another `%s` is `number of capture groups in pattern` or `max number of items after splitting column names`, also untranslated.
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/fmelt.R#L91
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/fmelt.R#L97
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/fmelt.R#L102
* [x] `%1$s` is `keys` or `indices`, untranslated.
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/setops.R#L160-L177
* [x] `%3$s` is either `error` or `warning`, untranslated. That's relatively minor and could be worked around in translation (`condition of class %s`):
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/fread.R#L314

### Minor issues with correctness

* [x] This should have said `POSIXct`, not just `POSIX` (which could be read as `POSIXt` or `POSIXlt`, and we don't like `POSIXlt`):
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/between.R#L13
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/between.R#L15
* [x] Missing `]` at the end:
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/data.table.R#L295
* [x] The last `}` needs to be replaced with `]`
https://github.com/Rdatatable/data.table/blob/745160d7247170b7910324ae7c2f25c8d5cc4cf6/R/data.table.R#L412
* [x] [PR17478](https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17478) is now `FIXED CLOSED`:
[...]

PR #6534, opened 2024-09-24
Refactor internal helpers which generate fragmented messages

MichaelChirico:
Noted in #6532. We should also address the `group.desc` issue, but in principle this PR could be submitted first. cc @aitap


tdhock (approved):
helper function duplicated_values looks ok.
I think we should remove group.desc argument, and add a new argument, `called.from.measure` default FALSE.
* If TRUE then we use stopf("... arguments to measure should be uniquely named, problems: %s")
* If FALSE then we use stopf("elements of fun.list should be uniquely named, problems: %s")
or is there some better way of handling the translation issue?

MichaelChirico:
I went with "just drop `group.desc` altogether", I think the error will be clear enough. `measurev()` is mentioned, which is documented next to `measure()` with the argument `fun.list`. If we really wanted to keep the distinction, I'd suggest looking at `sys.calls()` or something like that, rather than adding a new argument which is kind of useless when `measurev()` is exported.

tdhock:
yes that is a good point that we shouldn't be adding arguments that the user can not provide

MichaelChirico:
OK, I'll merge here. If we decide to decide internally that parent is `...` and change the message, we can do so as a follow-up :)

PR #6536, opened 2024-09-24
Missed/partial translations in setops/fread

MichaelChirico:
Towards #6532; partner: #6534

 - While we _could_ use "condition of class %s" in `fread()`, I think that's less user-friendly. I suspect many (most) users will not be comfortable with that terminology, it's a relatively advanced topic in R and `fread()` might be the first function someone ever uses when learning R.
 - I was a bit indifferent between the approach taken here for `all.equal.data.table()` (change to `sprintf()` and `gettext()` a full sentence) vs. just `gettext("indices")`; I chose the former since sentences are typically easier to translate as a whole. (FFR: a @aitap pointed out, this is also better for translators of gendered languages, where individual words with different genders might affect translation of nearby words)

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -311,7 +311,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
              methods::as(v, new_class))
       },
       warning = fun <- function(e) {
-        warningf("Column '%s' was requested to be '%s' but fread encountered the following %s:\n\t%s\nso the column has been left as type '%s'", names(ans)[j], new_class, if (inherits(e, "error")) "error" else "warning", e$message, typeof(v))
+        warningf("Column '%s' was requested to be '%s' but fread encountered the following %s:\n\t%s\nso the column has been left as type '%s'", names(ans)[j], new_class, if (inherits(e, "error")) gettext("error") else gettext("warning"), e$message, typeof(v))
         v
       },
       error = fun)
--- a/R/fread.R
+++ b/R/fread.R
@@ -310,8 +310,14 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
              # finally:
              methods::as(v, new_class))
       },
-      warning = fun <- function(e) {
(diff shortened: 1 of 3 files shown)


R/fread.R
```diff
@@ -311,7 +311,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
              methods::as(v, new_class))
       },
       warning = fun <- function(e) {
-        warningf("Column '%s' was requested to be '%s' but fread encountered the following %s:\n\t%s\nso the column has been left as type '%s'", names(ans)[j], new_class, if (inherits(e, "error")) "error" else "warning", e$message, typeof(v))
+        warningf("Column '%s' was requested to be '%s' but fread encountered the following %s:\n\t%s\nso the column has been left as type '%s'", names(ans)[j], new_class, if (inherits(e, "error")) gettext("error") else gettext("warning"), e$message, typeof(v))
```
aitap, line 314:
```suggestion
        warningf(if (inherits(e, "error")) gettext("Column '%s' was requested to be '%s' but fread encountered the following error:\n\t%s\nso the column has been left as type '%s'") else gettext("Column '%s' was requested to be '%s' but fread encountered the following warning:\n\t%s\nso the column has been left as type '%s'"), names(ans)[j], new_class, e$message, typeof(v))
```
Can we branch this too? This will also help with inflection, both due to grammatical case (nominative "ошибк**а**" but "encountered" "ошибк**у**", accusative) and grammatical gender ("данн**ое** предупреждение", warning, but "данн**ую** ошибку", error)

MichaelChirico: OK, I moved it outside the `warningf()` call though for readability.

aitap:
I'm sorry, what does it mean that I'm assigned here? I think you've fixed all the issues, at least I'm not seeing any fragmented sentences after `potools::po_extract()` and `msgmerge`. If we want to translsate [the message mentioned by @rikivillalba](https://github.com/Rdatatable/data.table/issues/6532#issuecomment-2377920535), should I push the commit here?

MichaelChirico:
Since there's not really any code/logic change here, I think it's OK to get review from non-committer in this case, just hoping from a LGTM from you since you have the translator's eye for whether the changes look good. Thanks!

PR #6540, opened 2024-09-25
Restore covr

MichaelChirico:
Reverts Rdatatable/data.table#6122 with change to pin R version to R<4.4.0 (the still unfixed source of the segfault).

Closes #6479.

PR #6541, opened 2024-09-26
# nocov for recent missed lines

MichaelChirico:
Follow-up to #6540, mainly to mark recently-missed lines with `# nocov` as appropriate. There are three main sources of recent false positives:

 1. We stopped testing certain Suggests packages in our main suite, relegating such tests to other.Rraw e.g. #5518
 2. We added a lot of safety checks on whether `malloc()` allocations succeeded: #6271
 3. I think {covr} itself started marking `default:` lines from `switch()` as "uncovered". We _might_ want to just do `default: internal_error() // # nocov` (i.e., all in one line vs. two consecutive `# nocov` lines)

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -158,6 +158,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     }
     # whitespace at the beginning or end of na.strings is checked at C level and is an error there; test 1804
   }
+  # nocov start. Tested in other.Rraw tests 16, not in the main suite.
   if (yaml) {
     if (!requireNamespace('yaml', quietly = TRUE))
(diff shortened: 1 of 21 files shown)


R/fread.R
```diff
@@ -158,6 +158,7 @@ yaml=FALSE, autostart=NA, tmpdir=tempdir(), tz="UTC")
     }
     # whitespace at the beginning or end of na.strings is checked at C level and is an error there; test 1804
   }
+  # nocov start. Tested in other.Rraw tests 16, not in the main suite.
```
tdhock, line 161: these kind of comments with references to test numbers in other files is great

tdhock (approved):

tdhock:
thanks for your detailed analysis Michael

Issue #37, opened 2014-06-08
[R-Forge #5435] Using "digits" in in "print" function has no effect on data.table objects


arunsrinivasan:
Submitted by: Matthew Beckers; Assigned to: Arun ; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5435&group_id=240&atid=975)

I came across this issue using version 1.8.10, and it is also discussed here by others http://lists.r-forge.r-project.org/pipermail/datatable-help/2013-June/001882.html

Issue #36, opened 2014-06-08
[R-Forge #5437] By operations with factors 


arunsrinivasan:
Submitted by: Christophe Dervieux; Assigned to: Arun ; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5437&group_id=240&atid=975)

Hi, 

I have updated data.table package to 1.9.2 recently from 1.8.10 and I found errors on my previous code. 

See reproductible example below: 

On 1.8.10 :

``` S
DT <- data.table(X = factor(2006:2012), Y = rep(1:7, 2))
DT[, Z := paste(X, .N, sep = " - "), by = list(X)][]

       X     Y        Z
 1: 2006 1 2006 - 2
 2: 2007 2 2007 - 2
 3: 2008 3 2008 - 2
 4: 2009 4 2009 - 2
 5: 2010 5 2010 - 2
 6: 2011 6 2011 - 2
 7: 2012 7 2012 - 2
 8: 2006 1 2006 - 2
 9: 2007 2 2007 - 2
10: 2008 3 2008 - 2
11: 2009 4 2009 - 2
12: 2010 5 2010 - 2
13: 2011 6 2011 - 2
14: 2012 7 2012 - 2
```

In column Z, I get the level of the factor column X 
pasted with count '.N' as expected

However, in the 1.9.2, with same code :

``` S
DT<-data.table(X=factor(2006:2012),Y=rep(1:7,2))
DT[,Z:=paste(X,.N,sep=" - "),by=list(X)][]

        X     Y  Z
 1: 2006 1 1 - 2
 2: 2007 2 2 - 2
 3: 2008 3 3 - 2
 4: 2009 4 4 - 2
 5: 2010 5 5 - 2
 6: 2011 6 6 - 2
 7: 2012 7 7 - 2
 8: 2006 1 1 - 2
 9: 2007 2 2 - 2
10: 2008 3 3 - 2
11: 2009 4 4 - 2
12: 2010 5 5 - 2
13: 2011 6 6 - 2
14: 2012 7 7 - 2
```

as results, I do not get levels of factor column X but the numeric values associated with the level. 

is this working normally? Why has it changed? Is that a bug? 

I use this kind of procedure to make labels for `ggplot`. All my previous code is not working anymore. It's kind of annoying.

Thanks

Christophe

PR #6545, opened 2024-09-28
Bumped performance-tests GHA to v1.3.1

Anirban166:
Changelog:
- Previously, the time taken to run `atime::atime_pkg` was reported, which consisted of two tasks - Installing the different (git) versions of `data.table`, and running/plotting the test cases. Both of these are now considered separately for a clear breakdown.
- The plot shown in the PR threads will be one generated for preview, meaning it is condensed to only show the top 4 tests (the number can be configured using the `N.tests.preview` variable in our [tests.R](https://github.com/Rdatatable/data.table/blob/master/.ci/atime/tests.R)) based on having the most significant differences between HEAD and min. The full version (all tests) will be shown when we click/tap on the plot. 
(Note: Currently, the width of the preview plot is based on the total number of test cases as opposed to the number considered for generating the image, but it has been fixed and we just need to wait for that updated version of `atime` to be on CRAN, as that issue should automatically be resolved - Needs no updating on the GHA end)
- The timings information have been organized in a table. (I like this in particular as it avoids the repeated 'Time taken to ... ' lines)
- Only the second count is shown if a task we are measuring falls under 60 seconds.

This stemmed from [#36](https://github.com/Anirban166/Autocomment-atime-results/issues/36) and [#37](https://github.com/Anirban166/Autocomment-atime-results/issues/37) - Thanks @tdhock!

### DIFF
--- a/.github/workflows/performance-tests.yml
+++ b/.github/workflows/performance-tests.yml
@@ -22,4 +22,4 @@ jobs:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       repo_token: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: Anirban166/Autocomment-atime-results@v1.2.1
+      - uses: Anirban166/Autocomment-atime-results@v1.3.1


MichaelChirico:
Great update, thanks!

PR #6555, opened 2024-10-03
Added an atime test for performance improvement in forderv when reusing existing key and index attributes

### LINKED ISSUE #6320: Add an atime performance regression test for forder caching
@Anirban166 this is a good candidate for a performance regression test, would you like to take a stab at writing one? I don't think Jan will get to it quickly :)

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/issues/4386#issuecomment-2224114772_
            
It will be good to have a test in the performance suite showing the improvement + preventing regression.

Anirban166:
Closes #6320

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -23,7 +23,7 @@ for (extra.arg in extra.args.6107){
 retGrp_values <- c("T","F")
 for(retGrp_setup in retGrp_values){
   for(retGrp_expr in retGrp_values){
-    test.name <- sprintf("forderv(retGrp=%s->%s) improved in #4386", retGrp_setup, retGrp_expr)
+    test.name <- sprintf("forderv(retGrp=%s-%s) improved in #4386", retGrp_setup, retGrp_expr)
     extra.test.list[[test.name]] <- list(
     setup = substitute({
       options(datatable.forder.auto.index = TRUE)
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -19,6 +19,28 @@ for (extra.arg in extra.args.6107){
   extra.test.list[[sprintf("fread(%s) improved in #6107", extra.arg)]] <- this.test
 }
 
+# Test case adapted from https://github.com/Rdatatable/data.table/pull/4386#issue-602528139 which is where the performance was improved.
+retGrp_values <- c("T","F")
+for(retGrp_setup in retGrp_values){
+  for(retGrp_expr in retGrp_values){
+    test.name <- sprintf("forderv(retGrp=%s->%s) improved in #4386", retGrp_setup, retGrp_expr)
+    extra.test.list[[test.name]] <- list(
+    setup = substitute({
+      options(datatable.forder.auto.index = TRUE)
+      set.seed(1)
+      dt <- data.table(index = sample(N), values = sample(N))
+      data.table:::forderv(dt, "index", retGrp = RETGRP) # Initial sort to create the index and initialize caching
+    }, list(RETGRP=str2lang(retGrp_setup))),
+    expr = substitute({
+      data.table:::forderv(dt, "index", retGrp = RETGRP) # Reusing the index and computing group info.
+    }, list(RETGRP=str2lang(retGrp_expr))),
+    Slow = "c152ced0e5799acee1589910c69c1a2c6586b95d", # Parent of the merge commit of the PR (https://github.com/Rdatatable/data.table/pull/4386/commits) that fixes the regression
+    Fast = "1a84514f6d20ff1f9cc614ea9b92ccdee5541506") # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/4386/commits) that fixes the regression
+  }
+}
+  
+
+
 # A list of performance tests.
 #
 # See documentation in https://github.com/Rdatatable/data.table/wiki/Performance-testing for best practices.
@@ -207,21 +229,6 @@ test.list <- atime::atime_test_list(
     expr = data.table:::melt(DT, measure.vars = measure.vars),
     Slow = "fd24a3105953f7785ea7414678ed8e04524e6955", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/ed72e398df76a0fcfd134a4ad92356690e4210ea) of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
     Fast = "ed72e398df76a0fcfd134a4ad92356690e4210ea"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
-
-# Improvement discussed in and brought by https://github.com/Rdatatable/data.table/pull/4386
-"forderv improved in #4386" = atime::atime_test(
-  setup = {
-    options(datatable.forder.auto.index = TRUE)
-    set.seed(1)
-    dt <- data.table(index = sample(N), values = sample(N))
-    data.table:::forderv(dt, "index") # Initial sort to create the index and initialize caching
-  },
-  expr = {
-    data.table:::forderv(dt, "index", retGrp = FALSE) # Reusing the cached index (no group information required)
-    data.table:::forderv(dt, "index", retGrp = TRUE) # Reusing the index and computing group info.
-  },
-  Slow = "c152ced0e5799acee1589910c69c1a2c6586b95d", # Parent of the merge commit of the PR (https://github.com/Rdatatable/data.table/pull/4386/commits) that fixes the regression
-  Fast = "1a84514f6d20ff1f9cc614ea9b92ccdee5541506"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/4386/commits) that fixes the regression
   
   tests=extra.test.list)
 # nolint end: undesirable_operator_linter.
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -208,5 +208,20 @@ test.list <- atime::atime_test_list(
     Slow = "fd24a3105953f7785ea7414678ed8e04524e6955", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/ed72e398df76a0fcfd134a4ad92356690e4210ea) of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
     Fast = "ed72e398df76a0fcfd134a4ad92356690e4210ea"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
 
+# Improvement discussed in and brought by https://github.com/Rdatatable/data.table/pull/4386
+"forderv improved in #4386" = atime::atime_test(
+  N = 10^seq(3, 8), # 1e9 exceeds the runner's memory (process gets killed)
+  setup = {
+    options(datatable.forder.auto.index = TRUE, datatable.forder.reuse.sorting = TRUE)
+    dt <- data.table(index = sample(N), values = sample(N))
+    data.table:::forderv(dt, "index") # Initial sort to create the index and initialize caching
+  },
+  expr = {
+    data.table:::forderv(dt, "index", retGrp = FALSE) # Reusing the cached index (no group information required)
+    data.table:::forderv(dt, "index", retGrp = TRUE) # Reusing the index and computing group info.
+  },
+  Slow = "c152ced0e5799acee1589910c69c1a2c6586b95d", # Parent of the merge commit of the PR (https://github.com/Rdatatable/data.table/pull/4386/commits) that fixes the regression
+  Fast = "1a84514f6d20ff1f9cc614ea9b92ccdee5541506"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/4386/commits) that fixes the regression
+  
   tests=extra.test.list)
 # nolint end: undesirable_operator_linter.
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -24,11 +24,12 @@ for(retGrp_chr in c("T","F"))extra.test.list[[sprintf(
   "forderv(retGrp=%s) improved in #4386", retGrp_chr
 )]] <- list(
   setup = quote({
-    options(datatable.forder.auto.index = TRUE)
     dt <- data.table(group = rep(1:2, l=N))
   }),
   expr = substitute({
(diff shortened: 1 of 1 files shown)


.ci/atime/tests.R
```diff
@@ -210,7 +210,7 @@ test.list <- atime::atime_test_list(
 
 # Improvement discussed in and brought by https://github.com/Rdatatable/data.table/pull/4386
 "forderv improved in #4386" = atime::atime_test(
-  N = 10^seq(3, 8), # 1e9 exceeds the runner's memory (process gets killed)
+  N = 10^seq(3, 7),
```
tdhock, line 213: default above is `N = as.integer(10^seq(1, 7, by=0.25))` so `N` does not have to be specified here, so can you please delete?

.ci/atime/tests.R
```diff
@@ -208,5 +208,20 @@ test.list <- atime::atime_test_list(
     Slow = "fd24a3105953f7785ea7414678ed8e04524e6955", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/ed72e398df76a0fcfd134a4ad92356690e4210ea) of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
     Fast = "ed72e398df76a0fcfd134a4ad92356690e4210ea"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
 
+# Improvement discussed in and brought by https://github.com/Rdatatable/data.table/pull/4386
+"forderv improved in #4386" = atime::atime_test(
+  N = 10^seq(3, 8), # 1e9 exceeds the runner's memory (process gets killed)
+  setup = {
+    options(datatable.forder.auto.index = TRUE, datatable.forder.reuse.sorting = TRUE)
+    dt <- data.table(index = sample(N), values = sample(N))
```
tdhock, line 216: please add set.seed(1) for reproducibility in setup, whenever you use random generation

Anirban166: Done! (and noted)

.ci/atime/tests.R
```diff
@@ -208,5 +208,20 @@ test.list <- atime::atime_test_list(
     Slow = "fd24a3105953f7785ea7414678ed8e04524e6955", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/ed72e398df76a0fcfd134a4ad92356690e4210ea) of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
     Fast = "ed72e398df76a0fcfd134a4ad92356690e4210ea"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
 
+# Improvement discussed in and brought by https://github.com/Rdatatable/data.table/pull/4386
+"forderv improved in #4386" = atime::atime_test(
+  N = 10^seq(3, 8), # 1e9 exceeds the runner's memory (process gets killed)
+  setup = {
+    options(datatable.forder.auto.index = TRUE, datatable.forder.reuse.sorting = TRUE)
```
tdhock, line 214: I wonder if setting these options will affect other tests? 
I checked where this is defined / documented and I found for `auto.index` :

```
grep  -nH --null "auto[.]index" ../*/*
../R/data.table.R:3309:    if (!getOption("datatable.auto.index")) return(NULL)
../R/onLoad.R:87:       "datatable.auto.index"="TRUE",          # DT[col=="val"] to auto add index so 2nd time faster
../R/setkey.R:144:  if (isTRUE(getOption("datatable.forder.auto.index"))) return(invisible())
grep: ../inst/include: Is a directory
grep: ../inst/po: Is a directory
grep: ../inst/tests: Is a directory
../man/datatable-optimize.Rd:102:\code{options(datatable.auto.index = FALSE)}. To switch off using existing
Binary file ../src/data_table.dll matches
../src/forder.c:1628:// isTRUE(getOption("datatable.auto.index"))
../src/forder.c:1630:  // for now temporarily 'forder.auto.index' not 'auto.index' to disabled it by default
../src/forder.c:1633:  SEXP opt = GetOption(install("datatable.forder.auto.index"), R_NilValue);
../src/forder.c:1637:    error("'datatable.forder.auto.index' option must be TRUE or FALSE"); // # nocov
../src/forder.c:1771:        GetAutoIndex()) { // disabled by default, use datatable.forder.auto.index=T to enable, do not export/document, use for debugging only
Binary file ../src/forder.o matches
grep: ../vignettes/css: Is a directory
../vignettes/datatable-benchmarking.Rmd:67:options(datatable.auto.index=TRUE)
../vignettes/datatable-benchmarking.Rmd:72:- `auto.index=FALSE` disables building index automatically when doing subset on non-indexed data, but if indices were created before this option was set, or explicitly by calling `setindex` they still will be used for optimization.
../vignettes/datatable-secondary-indices-and-auto-indexing.Rmd:318:* Auto indexing can be disabled by setting the global argument `options(datatable.auto.index = FALSE)`.
grep: ../vignettes/plots: Is a directory
```
above I see two options, `datatable.forder.auto.index` and `datatable.auto.index`
* `datatable.auto.index` is documented in Rd and two vignettes
* forder.c comments say "datatable.forder.auto.index=T to enable, do not export/document, use for debugging only"

which option does your test require setting? can you please add a comment to clarify why?


and I found for `reuse.sorting`:
```
grep  -nH --null "reuse[.]sorting" ../*/*
../R/setkey.R:151:forderv = function(x, by=seq_along(x), retGrp=FALSE, retStats=retGrp, sort=TRUE, order=1L, na.last=FALSE, reuseSorting=getOption("datatable.reuse.sorting", NA)) {
```

in your test you have `datatable.forder.reuse.sorting` but in the code there is no forder in the option name, `datatable.reuse.sorting` ... so I guess this option is not required to get your test result... where did you get that option name?

Anirban166: Yes I don't need `datatable.reuse.sorting`, so do you want me to add a comment as to why `datatable.forder.auto.index` is required?

Anirban166: > I wonder if setting these options will affect other tests?

I hope not! (we'll have to see from the results we get here)

Anirban166: Doesn't look like its affecting the other tests

tdhock:
your original test had 3 calls to forderv.
* first in setup with retGrp default(=FALSE)
* then in expr with retGrp=FALSE
* then again in expr with retGrp=TRUE
I was wondering if we need both retGrp=TRUE and FALSE in expr?
If we have TRUE only (or FALSE only) I get this result (constant instead of linear for Fast)
image
I coded a for loop over doing retGrp=T or F in setup, and also in expr. (4 test cases total)
For three of these we see constant Fast, but for retGrp=T in setup then retGrp=F in expr, we see linear Fast. is this expected @jangorecki @MichaelChirico ?
looking at datatable.verbose=T in this case I see "using existing index" which to me suggests that this should not constant (not linear as observed), so is this a performance bug??
I think we need explain in the comments what is expected to happen (when is expr expected to be linear vs constant).
I wonder if we need to use something like `data.table:::setattr(L, "index", NULL)`  ??

tdhock:
so here is a CI result that shows 3 constant Fast vs 1 linear Fast https://asset.cml.dev/f6abccc722845026bf081f0c9aa93b071bef4a4a?cml=png&cache-bypass=5e8ae8d9-fc0e-4b9b-9831-12442334b5d5

.ci/atime/tests.R
```diff
@@ -25,15 +25,23 @@ for(retGrp_setup in retGrp_values){
...
+      }
+    }),
     expr = substitute({
-      data.table:::forderv(dt, "index", retGrp = RETGRP) # Reusing the index and computing group info.
-    }, list(RETGRP=str2lang(retGrp_expr))),
+      setattr(dt, "index", index.list[[retGrp_setup]]) 
+      data.table:::forderv(dt, "index", retGrp = retGrp_expr) # Reusing the index and computing group info.
+    }, list(
+      retGrp_setup=retGrp_setup,
+      retGrp_expr=str2lang(retGrp_expr)
```
tdhock, line 43: in this refactor I compute both retGrp=T and F in setup (using CRAN data.table), and store the "index" attribute in `index.list`

then in expr I set the "index" attribute to one of the two indices we computed in setup (using either retGrp=T or F).
and then we call forderv using a git version of data.table...
and I observe a different result than before --- Fast is always linear (for any combination of retGrp=T or F), never constant.

I don't understand what is happening, or which version of this test we should prefer (with or without setattr).
Can someone who knows forderv better please investigate/explain what is expected time complexity here?

jangorecki:
Not sure what exactly question is about.

tdhock:
Hi @jangorecki we are trying to adapt your examples here https://github.com/Rdatatable/data.table/pull/4386#issue-602528139  to be a performance test, but we are not sure about what is the right way to test the new functionality vs the old.
Could you please fill in the TODOs below with "linear" or "constant" ? 
(is it true that we expect constant time if old index is used, and linear time if new index must be computed?)
* If we first run forderv(retGrp=TRUE) then running forderv(retGrp=TRUE) should take TODO time. (TDH guess TODO=constant?)
* If we first run forderv(retGrp=TRUE) then running forderv(retGrp=FALSE) should take TODO time. (TDH guess TODO=constant?)
* If we first run forderv(retGrp=FALSE) then running forderv(retGrp=TRUE) should take TODO time. (TDH guess TODO=linear?)
* If we first run forderv(retGrp=FALSE) then running forderv(retGrp=FALSE) should take TODO time. (TDH guess TODO=constant?)

I guess I don't understand the difference between retGrp=TRUE and FALSE, is that documented somewhere?

jangorecki:
Rerunning same should be constant. As for the other two cases I cannot look so much in details into it at the moment, but possibly when computing retGrp=F when having retGrp=T we may still need to run forder on the index itself to get the original order from groups. In one of PR was fdistinct() which is very small function that was using those so the exact logic can be looked up there easily.

tdhock:
> I think we need explain in the comments what is expected to happen (when is expr expected to be linear vs constant).

I added a comment.

> I wonder if we need to use something like data.table:::setattr(L, "index", NULL) ??

We don't, and in fact we don't even need to run forderv in setup, because the first memory measurement run sets the index, for use in subsequent time measurements.

> Rerunning same should be constant

Thanks, that helps! I edited the PR so that it only has these two test cases. (instead of all four described above for which we do not have a clear expectation).

I think this PR should be ready to merge now. @Anirban166 can you please review?

The atime results are missing Slow/Fast for some test cases, https://github.com/tdhock/atime/issues/64 which is fixed in github atime (to send to CRAN later this week).

Anirban166:
> I think this PR should be ready to merge now. @Anirban166 can you please review?

Something that I see [here](https://github.com/Rdatatable/data.table/pull/6555#issuecomment-2392212863), in my [fork](https://github.com/Anirban166/data.table/pull/24#issuecomment-2395683152), and also when I run the tests locally on my machine (sharing the generated plot below, it uses the fixed/latest `atime` version from GitHub) is that the lines for these two cases in terms of memory isn't being shown for versions other than 'Slow'/'Fast': (base, HEAD, CRAN, and merge-base are stuck below along the x-axis line)

tests_all_facet

Their individual plots:

| retGrp=T | retGrp=F |
|---|---|
| forderv(retGrp=T)_improved_in_#4386  |  forderv(retGrp=F)_improved_in_#4386  |

Is that expected or within reason? (asking since for the other tests I don't notice such; also are these results consistent with what you've been observing locally?)
Otherwise, it looks great to me! (and the remaining tests look unaffected/normal)

> The atime results are missing Slow/Fast for some test cases

As per the CI results we're getting, 'some test cases' more specifically would be all the other tests aside from the `forderv` cases we are introducing here, as they are missing labels for Slow/Fast/Before/Fixed/Regression if you look at both plots (preview and full version)

> [tdhock/atime#64](https://github.com/tdhock/atime/issues/64) which is fixed in github atime (to send to CRAN later this week).

Sweet!

tdhock:
> the lines for these two cases in terms of memory isn't being shown for versions other than 'Slow'/'Fast': (base, HEAD, CRAN, and merge-base are stuck below along the x-axis line)

thanks for sharing. This result implies that no memory is used/measurable during the test for those versions. I don't understand why though.

tdhock:
> the lines for these two cases in terms of memory isn't being shown for versions other than 'Slow'/'Fast': (base, HEAD, CRAN, and merge-base are stuck below along the x-axis line)

thanks for sharing. This result implies that no memory is used/measurable during the test for those versions. I don't understand why though.

Anirban166:
Right, and I expected Fast to be beside HEAD (along with base, CRAN, and merge-base for CI/PRs)

.ci/atime/tests.R
```diff
@@ -207,6 +232,6 @@ test.list <- atime::atime_test_list(
     expr = data.table:::melt(DT, measure.vars = measure.vars),
     Slow = "fd24a3105953f7785ea7414678ed8e04524e6955", # Parent of the merge commit (https://github.com/Rdatatable/data.table/commit/ed72e398df76a0fcfd134a4ad92356690e4210ea) of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
     Fast = "ed72e398df76a0fcfd134a4ad92356690e4210ea"), # Merge commit of the PR (https://github.com/Rdatatable/data.table/pull/5054) that fixes the issue
-
+
```
Anirban166, line 235: whitespaces

.ci/atime/tests.R
```diff
@@ -19,6 +19,31 @@ for (extra.arg in extra.args.6107){
...
 
+# Test case adapted from https://github.com/Rdatatable/data.table/pull/4386#issue-602528139 which is where the performance was improved.
+for(retGrp_chr in c("T","F"))extra.test.list[[sprintf(
+  "forderv(retGrp=%s) improved in #4386", retGrp_chr
+)]] <- list(
+  setup = quote({
+    dt <- data.table(group = rep(1:2, l=N))
+  }),
+  expr = substitute({
+    old.opt <- options(datatable.forder.auto.index = TRUE) # required for test, un-documented, comments in forder.c say it is for debugging only.
```
Anirban166, line 30: I think Jan mentioned that it is only for 'debugging' since it was not exported (coming from his [comment here](https://github.com/Rdatatable/data.table/pull/4386#issue-602528139)), but we can probably ask him for more information if we want to document it or have more details (otherwise yes, not having this option set to `TRUE` will not show performance improvement)

Anirban166: Btw, good move to save the old options here and then restore them after the call to `forderv`!

Anirban166: Good move to save the old options here and then restore them after the call to `forderv`!

PR #6557, opened 2024-10-04
add export measurev to NAMESPACE

### LINKED ISSUE #6535: Export measurev()
since measure is exported, we should export measurev too. both are documented because they are meant to be used as the `measure.vars` argument to `melt`.
when I wrote measure/measurev, I was not aware of the issues created by breaking strings apart when translating messages.
so we should probably get rid of the `group.desc` argument, I will look at #6534

_Originally posted by @tdhock in https://github.com/Rdatatable/data.table/issues/6532#issuecomment-2371871773_

`measurev()` is analogous to `setorderv()`, `setkeyv()`, `setindexv()`.

tdhock:
closes #6535

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -28,7 +28,7 @@ export(tstrsplit)
 export(frank)
 export(frankv)
 export(address)
-export(.SD,.N,.I,.GRP,.NGRP,.BY,.EACHI, measure, patterns)
+export(.SD,.N,.I,.GRP,.NGRP,.BY,.EACHI, measure, measurev, patterns)
 # TODO(#6197): Export these.
 # export(., J)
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Does this warrant a NEWS entry?

MichaelChirico (approved):

MichaelChirico:
For future reference: no changes need to man/, it's already an `\alias{}` there:

https://github.com/Rdatatable/data.table/blob/036fb48f6d7b039e485b1a39d47204fddcfc6a87/man/measure.Rd#L3

Anirban166 (approved):

tdhock:
I added a NEWS item, thanks!

PR #6558, opened 2024-10-05
Update datatable-intro.Rmd

gurbuxanink:
A powerful feature of data.table is the ability to store lists as elements (a facility not available in data.frame).  Here is a simple illustration of storing a fitted model as data.table element.

When statistical models are fit to different groups within a data.table, these models can be stored in a result data.table.  Now this resulting data.table can be used to access fitted models.

### DIFF
--- a/vignettes/datatable-intro.Rmd
+++ b/vignettes/datatable-intro.Rmd
@@ -639,7 +639,7 @@ DT[, print(list(c(a,b))), by = ID] # (2)
 
 In (1), for each group, a vector is returned, with length = 6,4,2 here. However, (2) returns a list of length 1 for each group, with its first element holding vectors of length 6,4,2. Therefore, (1) results in a length of ` 6+4+2 = `r 6+4+2``, whereas (2) returns `1+1+1=`r 1+1+1``.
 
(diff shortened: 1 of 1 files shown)


tdeenes:
note that this is not exactly true, `data.frame`s __can__ hold list elements:
```
datfr <- data.frame(idx = 1:3)
datfr$lst <- lapply(datfr$idx, seq_len)
str(datfr)
# 'data.frame':	3 obs. of  2 variables:
#  $ idx: int  1 2 3
#  $ lst:List of 3
#   ..$ : int 1
#   ..$ : int  1 2
#   ..$ : int  1 2 3
```

Having said that, of course it is much more convenient to use `data.table`s for working with this kind of objects. E.g.:
```
lst <- lapply(1:3, seq_len)
## this works
data.table(idx = seq_along(lst), lst = lst)
## this fails
data.frame(idx = seq_along(lst), lst = lst)
```

tdhock:
please clarify your proposed addition by mentioning that it is possible in frame but requires a bit more typing, since you have to wrap lists with `I()`
```r
> data.table(L=list(1:2))
        L
   <list>
1:    1,2
> data.frame(L=list(1:2))
  X1.2
1    1
2    2
> data.frame(L=I(list(1:2)))
     L
1 1, 2
```

tdhock:
that looks better thanks

tdhock (approved):
great thanks!

Issue #6509, opened 2024-09-20
tests/knitr.R mismatch


jangorecki:
reported on CRAN now. https://cran.r-project.org/web/checks/check_results_data.table.html
none of the commits mentioned in 1.16.2 issue seems to be providing the fix, so likely the issue is still there
```
    Comparing ‘knitr.Rout’ to ‘knitr.Rout.save’ ...28,38d27
  < ```
  < 
  < ```
  < ##        x     y     z
  < ##    <int> <int> <int>
  < ## 1:     1     4     7
  < ## 2:     2     5     8
  < ## 3:     3     6     9
  < ```
  < 
  < ``` r
  52,62d40
  < ```
  < 
  < ```
  < ##        x     y     z     a
  < ##    <int> <int> <int> <int>
  < ## 1:     1     4    10     1
  < ## 2:     2     5    11     1
  < ## 3:     3     6    12     1
  < ```
  < 
  < ``` r
```

MichaelChirico:
Copying @TysonStanley 's [comment](https://github.com/Rdatatable/data.table/issues/6444#issuecomment-2380056832) here:

Looks like the difference is that `DT[, z := 7:9]` is printing in the `knitr` output.

```
  < ##        x     y     z
  < ##    <int> <int> <int>
  < ## 1:     1     4     7
  < ## 2:     2     5     8
  < ## 3:     3     6     9
```

MichaelChirico:
Copying Tyson's [next comment](https://github.com/Rdatatable/data.table/issues/6444#issuecomment-2380097469) here too:

Not sure why it's not showing up in CI checks at all. It's a note on Linux and windows on CRAN and it showed up on MacOS for me here.

MichaelChirico:
I'm not reproducing it locally, and we don't see it on GHA nor on GitLab CI logs. Makes it a lot tougher to debug.

@TysonStanley, can you confirm your R, knitr and data.table versions where you're seeing the issue? And maybe try doing `trace(data.table:::shouldPrint, at=3, quote(message("ret=", ret)))` and check the output of the knitr test again?

TysonStanley:
Only happens when I'm accessing R directly in the terminal but this is what happened when I inserted that code into the .Rmd before knitting:

````
``` r
require(data.table)              # print?
trace(data.table:::shouldPrint, at=3, quote(message("ret=", ret)))
```

```
## [1] "shouldPrint"
```

``` r
[...]

PR #6563, opened 2024-10-09
print.data.table: check for knitr at sys.calls()-6

aitap:
This is a band-aid for #6509. Previous check for `mimicsAutoPrint` (i.e. `knit_print.default`) at `sys.calls()[[length(sys.calls())-3L]]` [no longer works after a knitr upgrade](https://github.com/Rdatatable/data.table/issues/6509#issuecomment-2400148516). Currently, the call to the autoprint-like function resides at position -6.

**Edit**: unfortunately there doesn't seem to be a right way to solve this problem. Returning `invisible(anything)` from `[.data.table` sets it back to visible because `PRIMPRINT(.Primitive("["))` says so, and `knitr` respects visibility of the evaluation results.

### DIFF
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -34,6 +34,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
         ( length(SYS) >= 3L && is.symbol(thisSYS <- SYS[[length(SYS)-2L]][[1L]]) &&
           as.character(thisSYS) == 'source') || # suppress printing from source(echo = TRUE) calls, #2369
         ( length(SYS) > 3L && is.symbol(thisSYS <- SYS[[length(SYS)-3L]][[1L]]) &&
+          as.character(thisSYS) %chin% mimicsAutoPrint ) ||
+        ( length(SYS) > 6L && is.symbol(thisSYS <- SYS[[length(SYS)-6L]][[1L]]) &&
           as.character(thisSYS) %chin% mimicsAutoPrint ) )  {
       return(invisible(x))
       # is.symbol() temp fix for #1758.
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -34,7 +34,15 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
         ( length(SYS) >= 3L && is.symbol(thisSYS <- SYS[[length(SYS)-2L]][[1L]]) &&
           as.character(thisSYS) == 'source') || # suppress printing from source(echo = TRUE) calls, #2369
         ( length(SYS) > 3L && is.symbol(thisSYS <- SYS[[length(SYS)-3L]][[1L]]) &&
-          as.character(thisSYS) %chin% mimicsAutoPrint ) ||
+          as.character(thisSYS) %chin% mimicsAutoPrint ) || # suppress printing from knitr, #6509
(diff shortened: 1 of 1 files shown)


R/print.data.table.R
```diff
@@ -34,6 +34,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
         ( length(SYS) >= 3L && is.symbol(thisSYS <- SYS[[length(SYS)-2L]][[1L]]) &&
           as.character(thisSYS) == 'source') || # suppress printing from source(echo = TRUE) calls, #2369
         ( length(SYS) > 3L && is.symbol(thisSYS <- SYS[[length(SYS)-3L]][[1L]]) &&
+          as.character(thisSYS) %chin% mimicsAutoPrint ) ||
```
tdhock, line 37: I don't really understand the issue/fix, so comments would be super helpful here.
Could you please add somethign like `# suppress printing in knitr, #6509 `
and maybe some explanation about where the magic number 6 comes from?

aitap: Does ffeb415 help?

The original problem is that `[.data.table` would like to return normally for `DT[<any read-only operation>]` but _invisibly_ for `DT[,<var> := <assignment by reference>]`.

Unfortunately, when evaluating calls to primitives, [R checks a special flag for whether to restore the visibility](https://github.com/r-devel/r-svn/blob/734e1434d3b229a008e73e3dfd4a27d733564e0e/src/main/eval.c#L1228-L1242), and the primitive `` `[` `` has [this flag](https://github.com/r-devel/r-svn/blob/734e1434d3b229a008e73e3dfd4a27d733564e0e/src/main/names.c#L64) [set to 0](https://github.com/r-devel/r-svn/blob/734e1434d3b229a008e73e3dfd4a27d733564e0e/src/main/names.c#L107) (force `R_Visible = TRUE`), so there doesn't seem to be a way to return `invisible(anything)` from `[.data.table`.

To achieve something like that, `data.table` remembers the address of the last data.table on which a by-reference operation has been computed. `print.data.table` checks if the address of the argument matches the remembered one and, if it suspects it's being called for auto-printing (call stack very shallow *or* called by a specific `knitr` function), it refuses to print anything. After something in the `knitr` stack got upgraded, the check for "am I being called for auto-printing by `knitr`" no longer works. We could also change the check to "is `knit_print.default` _anywhere_ in the call stack, not just at positions -3 or -6" and risk false positives.

We could try asking R core to change the flag on `.Primitive("[")`, but almost always forcing visibility on for results of `` `[` `` is the right thing and it'll take years before we can enjoy one less workaround in the codebase.

tdhock (approved):
great comments thanks!

for a medium-term solution I wonder if rendering vignettes using litedown would fix this, or cause other problems? https://yihui.org/litedown/#sec-package-vignettes

haha

MichaelChirico:
> for a medium-term solution I wonder if rendering vignettes using litedown would fix this, or cause other problems?

Not really. The test is here to ensure that _anyone_ using {knitr}+{data.table} gets the expected behavior when knitting their vignettes, regardless of what engine we use.

R/print.data.table.R
MichaelChirico: Thanks for the really good summary, really this should be in the codebase somewhere.

Issue #6559, opened 2024-10-07
1.16.0 testunit fails


nunotexbsd:
testunit from previous version 1.15.4 is OK.

```
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: amd64-portbld-freebsd15.0
Running under: FreeBSD 15amd64-main 15.0-CURRENT FreeBSD 15.0-CURRENT 1500023 amd64

Matrix products: default
LAPACK: /usr/local/lib/R/lib/libRlapack.so;  LAPACK version 3.12.0

locale:
[1] C.UTF-8/C.UTF-8/C.UTF-8/C/C.UTF-8/C.UTF-8

time zone: Europe/Lisbon
tzcode source: system (FreeBSD)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
[1] compiler_4.4.1
```
```
<snip>
* checking examples ... OK
* checking for unstated dependencies in 'tests' ... OK
* checking tests ...
  Running 'S4.R'
  Running 'autoprint.R'
  Comparing 'autoprint.Rout' to 'autoprint.Rout.save' ... OK
  Running 'froll.R'
  Running 'knitr.R'
  Comparing 'knitr.Rout' to 'knitr.Rout.save' ... OK
  Running 'main.R'
 ERROR
Running the tests in 'tests/main.R' failed.
Last 13 lines of output:
     0.000s ( 24%) Rereading 1 columns due to out-of-sample type exceptions
     0.000s        Total
  Column 1 <<A>> bumped from 'int32' to 'int64' due to <<123456789123456>> on row 114
            A           B
  "integer64"   "integer"
  Test 2266 produced 1 errors but expected 0
  Expected:
  Observed: no documentation of type 'a' and topic 'o' (or error in processing help)
  Error in eval(exprs[i], envir) : object 'tmp2' not found

[...]

PR #6567, opened 2024-10-10
tests: Detect session encoding incapable of U+00F1

aitap:
`iconv()` does not necessarily fail to convert `U+00F1` to ASCII and return `NA`. For example, FreeBSD `iconv()` succeeds and returns `?` instead of the character in question. Use `identical()` to compare the result of the conversion back to the original (which internally converts both to UTF-8).

Closes: #6339

Should with the part of #6559 that's about encodings, but not OpenMP

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18768,7 +18768,7 @@ if (test_bit64) local({
 # non-ASCII plain symbol in by, #4708
 # NB: recall we can't use non-ASCII symbols in the test script. The text is a-<n-tilde>-o (year in Spanish)
 native_ano = iconv("a\U00F1o", "UTF-8", "")
-if (!is.na(native_ano)) { # #6339: symbol must be represented in native encoding
+if (identical(native_ano, "a\U00F1o")) { # #6339: symbol must be represented in native encoding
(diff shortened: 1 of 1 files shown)


nunotexbsd:
Hello,

This change fixes test https://github.com/Rdatatable/data.table/issues/6559

Thanks

inst/tests/tests.Rraw
```diff
@@ -18768,7 +18768,7 @@ if (test_bit64) local({
 # non-ASCII plain symbol in by, #4708
 # NB: recall we can't use non-ASCII symbols in the test script. The text is a-<n-tilde>-o (year in Spanish)
 native_ano = iconv("a\U00F1o", "UTF-8", "")
-if (!is.na(native_ano)) { # #6339: symbol must be represented in native encoding
+if (identical(native_ano, "a\U00F1o")) { # #6339: symbol must be represented in native encoding
```
MichaelChirico, line 18771: Hmm, thanks. It still leaves us with the `else` branch here on FreeBSD where this is unanticipated / the behavior of `DT[,.N, by=año]` goes untested there. But let's start with un-breaking the test.

MichaelChirico: Follow-up is #6573

PR #6580, opened 2024-10-18
add clang-tidy to GLCI

### LINKED ISSUE #2781: Add clang-tidy to CI
As suggested by @michaelquinn32 here: https://github.com/Rdatatable/data.table/pull/2738#issuecomment-380888101

ben-schwen:
Closes #2781

Right now I would only start with the `readability-inconsistent-declaration-parameter` check and move forward with adding additional checks when needed.

edit: switched to C99 instead of C11 since C11 gave us a huge amount of false-positives

[CI Results](https://gitlab.com/Rdatatable/data.table/-/jobs/8126395916)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -198,7 +198,7 @@ test-lin-dev-clang-cran:
     - echo 'CFLAGS=-g -O2 -fno-common -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
     - echo 'CXXFLAGS=-g -O2 -fno-common -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' >> ~/.R/Makevars
     - *install-deps
-    - clang-tidy -extra-arg=-I/usr/local/lib/R/include -checks='readability-inconsistent-declaration-parameter' src/*.c -- -std=c11
+    - clang-tidy -extra-arg=-I/usr/local/lib/R/include -checks='readability-inconsistent-declaration-parameter' src/*.c -- -std=c99
     - R CMD check --as-cran $(ls -1t data.table_*.tar.gz | head -n 1)
     - (! grep "warning:" data.table.Rcheck/00install.out)
     - >-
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -198,6 +198,7 @@ test-lin-dev-clang-cran:
     - echo 'CFLAGS=-g -O2 -fno-common -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
     - echo 'CXXFLAGS=-g -O2 -fno-common -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' >> ~/.R/Makevars
     - *install-deps
+    - clang-tidy -extra-arg=-I/usr/local/lib/R/include -checks='readability-inconsistent-declaration-parameter' src/*.c -- -std=c11
```
MichaelChirico: what will this look like for manually specifying, say, dozens of checks? Will it just be

```shell
clang-tidy -extra-arg=... \
  -checks='...' \
  src/*.c
```

Is it possible to keep this as a list in a file (say in .ci)?

(doesn't need to be done in this PR since we only have one check thus far, just trying to think ahead before it gets much harder to refactor)

MichaelChirico (approved):

.gitlab-ci.yml
ben-schwen: Yes. We can also globbing for checks e.g. `-check='readability-*'` or multiple checks with `-checks='bugprone-*,readability-*'`

.gitlab-ci.yml
```diff
@@ -198,6 +198,7 @@ test-lin-dev-clang-cran:
     - echo 'CFLAGS=-g -O2 -fno-common -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
     - echo 'CXXFLAGS=-g -O2 -fno-common -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' >> ~/.R/Makevars
     - *install-deps
+    - clang-tidy -extra-arg=-I/usr/local/lib/R/include -checks='readability-inconsistent-declaration-parameter' src/*.c -- -std=c99
```
jangorecki, line 201: Comments describing this job, here in the file, could mention about this check as well. Also .ci/README.MD could be updated for that.

ben-schwen: Updating the .ci/README.md and adding some best practices like developing CI changes on Gitlab and then fetching them here is on my long standing to-do list

MichaelChirico:
Thanks for getting this started! My instinct is that eventually we'd like to move to "run all `clang-tidy` checks, except for a smaller sample that we disable" (e.g. that's also the plan for R {lintr} settings). In any case we should set up a tracking bug for inching this forward into an eventual steady state.

PR #6583, opened 2024-10-19
use litedown for rendering vignettes

ben-schwen:
Closes #6394

Issue #2369, opened 2017-09-18
[bug] assigning using := prints during source( , echo=TRUE) (but not from console)


jan-glx:
```
> dt <- data.table(a=1)
> dt[,a:=1]
> source('~/.active-rstudio-document', echo=TRUE)

> dt <- data.table(a=1)

> dt[,a:=1]
   a
1: 1
```
Related to #2319.

`sessionInfo()`:
```
R version 3.4.1 RC (2017-06-22 r72854)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.10.5

loaded via a namespace (and not attached):
[1] compiler_3.4.1 tools_3.4.1   
```

MichaelChirico:
Confirming this is not RStudio-specific; command line code:

```
echo "library(data.table)
DT = data.table(a = 1)
DT[ , a := 1]
" > test.R
R
source('test.R', echo = TRUE)
#    a
# 1: 1
DT[ , a := 1]
```

Also related to #933

PR #6589, opened 2024-10-21
Solve the `knitr` auto-printing problem by registering a method for `knit_print`

### LINKED ISSUE #6566: medium- and long-term plans for tests/knitr.R
- [x] I slightly botched #6563 by aiming it at `patch-1.16.2` instead of `master`. Should we cherry-pick 9de25675aa9f6efe982f4faff0ceb6ca55563563 into `master` too? See also: #6587
 - [x] Do we want a more reliable secret handshake with `knitr`? For example,
   - <del>Have `knitr` set `options(knitr.is.auto.printing = TRUE)` and detect that in `print.data.table`</del>
   - <del>`signalCondition(<please_dont_autoprint_me>)` from `print.data.table` and have `knitr` detect that</del>
   - > If you want the exact printing behavior of `data.table:::print.data.table` and want the result to be a table, you can register an S3 method on `xfun::record_print` (data.table doesn't have to import xfun).
     
      _Originally posted by @yihui in https://github.com/yihui/litedown/issues/38#issuecomment-2425149397_
      
      See #6589
 - [ ] Do we want to petition R developers to [allow returning invisibly from `[`](https://github.com/Rdatatable/data.table/pull/6563#discussion_r1793959120)? The change is just [one line](https://github.com/r-devel/r-svn/blob/db840e6da4cac9dd685a6e20237ec088c85da5fb/src/main/names.c#L107), eval = 0 to eval = 200, but it does break R's own `tests/reg-tests-2.Rout.save`, making `matrix(5:-6, 3)[2, invisible(3)]` also invisible, and might break a lot of other stuff.

aitap:
It turns out that there is a better solution for #6566 after all. It's documented in [`vignette('knit_print')`](https://cran.r-project.org/web/packages/knitr/vignettes/knit_print.html#for-package-authors) that packages can override `knitr`'s (auto-)printing behaviour by registering a method for it. R ≥ 3.6 can perform "delayed registration" of S3 methods for a generic in a package without requiring it to be loaded, hence keeping `knitr` in `Suggests:`. In R &in; [3.3, 3.6), the same thing can be done manually: I've reused a piece of code from `base::registerS3methods` to register `knit_print.data.table` if the package is already loaded and set a hook to register it later if needed.

The method itself just checks `shouldPrint(x)` and returns `invisible(x)` if not, otherwise delegates to the default method.

This can close #6566 if we don't want to change the behaviour of `.Primitive("[")` with regards to visibility.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -104,6 +104,8 @@ if (getRversion() >= "4.0.0") {
   # version of R (and that is checked in .onLoad with error if not).
   export(.rbind.data.table) # only export in R<4.0.0 where it is still used; R-devel now detects it is missing doc, #5600
 }
+if (getRversion() >= "3.6.0") S3method(knitr::knit_print, data.table)
+# else manual delayed registration from the onLoad hook
 S3method(dim, data.table)
 S3method(dimnames, data.table)
 S3method("dimnames<-", data.table)
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -104,8 +104,7 @@ if (getRversion() >= "4.0.0") {
   # version of R (and that is checked in .onLoad with error if not).
   export(.rbind.data.table) # only export in R<4.0.0 where it is still used; R-devel now detects it is missing doc, #5600
 }
-if (getRversion() >= "3.6.0") S3method(knitr::knit_print, data.table)
-# else manual delayed registration from the onLoad hook
+if (getRversion() >= "3.6.0") S3method(knitr::knit_print, data.table) # else manual delayed registration from the onLoad hook
 S3method(dim, data.table)
 S3method(dimnames, data.table)
 S3method("dimnames<-", data.table)
--- a/R/onLoad.R
+++ b/R/onLoad.R
@@ -66,7 +66,7 @@
       lockBinding("rbind.data.frame",baseenv())
     }
   }
-  if (session_r_version < "3.6.0") {
+  if (session_r_version < "3.6.0") { # corresponds to S3method() directive in NAMESPACE
     # no delayed registration support for NAMESPACE; perform it manually
     if (isNamespaceLoaded("knitr")) {
       registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
--- a/R/onLoad.R
+++ b/R/onLoad.R
@@ -69,10 +69,10 @@
   if (session_r_version < "3.6.0") {
     # no delayed registration support for NAMESPACE; perform it manually
     if (isNamespaceLoaded("knitr")) {
-        registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
+      registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
     }
     setHook(packageEvent("knitr", "onLoad"), function(...) {
-        registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
+      registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
     })
   }
 
--- a/R/onLoad.R
+++ b/R/onLoad.R
@@ -70,11 +70,10 @@
     # no delayed registration support for NAMESPACE; perform it manually
     if (isNamespaceLoaded("knitr")) {
       registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
-    } else {
-      setHook(packageEvent("knitr", "onLoad"), function(...) {
-        registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
-      })
     }
+    setHook(packageEvent("knitr", "onLoad"), function(...) {
+      registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
+    })
   }
 
   # Set options for the speed boost in v1.8.0 by avoiding 'default' arg of getOption(,default=)
--- a/R/onLoad.R
+++ b/R/onLoad.R
@@ -66,6 +66,15 @@
       lockBinding("rbind.data.frame",baseenv())
     }
   }
+  if (session_r_version < "3.6.0") {
+    # no delayed registration support for NAMESPACE; perform it manually
+    if (isNamespaceLoaded("knitr")) {
+        registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
+    }
+    setHook(packageEvent("knitr", "onLoad"), function(...) {
+        registerS3method("knit_print", "data.table", knit_print.data.table, envir = asNamespace("knitr"))
+    })
+  }
 
   # Set options for the speed boost in v1.8.0 by avoiding 'default' arg of getOption(,default=)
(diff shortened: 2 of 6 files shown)


NAMESPACE
```diff
@@ -104,6 +104,8 @@ if (getRversion() >= "4.0.0") {
   # version of R (and that is checked in .onLoad with error if not).
   export(.rbind.data.table) # only export in R<4.0.0 where it is still used; R-devel now detects it is missing doc, #5600
 }
+if (getRversion() >= "3.6.0") S3method(knitr::knit_print, data.table)
```
MichaelChirico, line 107: what happens if knitr is not installed?

IIUC this is an Enhances dependency; weve done something similar for {lintr}

https://github.com/r-lib/lintr/blob/1db7cc70655a41c77fcc89dbed5e40d5dd8cf74a/R/zzz.R#L337

aitap: `data.table` still loads with no warnings. `R CMD check` fails even with `_R_CHECK_FORCE_SUGGESTS_=FALSE` because it's the `VignetteBuilder`.

[WRE](https://cran.r-project.org/doc/manuals/R-exts.html#Registering-S3-methods) says:

> function `gen.cls` will get registered as an S3 method for class `cls` and generic `gen` from package `pkg` only when the namespace of pkg is loaded

Here's the code:

https://github.com/r-devel/r-svn/blob/320c6deb8bb07f9f83b3aa6cb3ca9b572c5d0b00/src/library/base/R/namespace.R#L1730-L1753

MichaelChirico: > even with _R_CHECK_FORCE_SUGGESTS_=FALSE because it's the VignetteBuilder.

Right, we need the flags to skip vignettes too, e.g. `--no-build-vignettes` `--no-vignettes`

https://github.com/r-devel/r-svn/blob/e1cbaee23a2c12c8faa85f91aa1cc19fe709ca03/src/library/tools/R/check.R#L6851-L6852

Anyway, my concern was about loading {data.table}, thanks for the WRE cite!

I still think we should put {knitr} into `Enhances:`, WDYT?

aitap: On the one hand, WRE says

> the `Enhances` field lists packages “enhanced” <...> by providing methods for classes from these packages

...which is a good argument for `Enhances`, but then it says

> A package should be listed in only one of these fields. 

> The `VignetteBuilder` <...> may include the current package, or ones listed in `Depends`, `Suggests` or `Imports`.

If we switch to `litedown` as the vignette builder, `knitr` should definitely go into `Enhances`. Currently, it looks like `knitr` and `markdown` should be in both `Suggests` and `VignetteBuilder`, although the latter doesn't seem to be tested. (E.g. [`cmocean` had `VignetteBuilder: knitr` for 5 years](https://github.com/aitap/cmocean/commit/7a8a907954b590455c41c5c38a1c4a98f4800fca) despite having [`%\VignetteEngine{knitr::rmarkdown}`](https://github.com/aitap/cmocean/blame/47b75447ed7ba4552952e5fc43e37261c72eb58a/vignettes/cmocean.Rmd#L7) and not a single test flagged it.)

Although it's of course possible to read WRE differently, in which case `knitr` could move from `Suggests` to `Enhances`.

MichaelChirico: You're 100% right, I'm not sure what I was thinking. {knitr} is already in `Suggests`, which is a stronger dependency than `Enhances`. #6583 is to use {litedown} instead, I added a comment there about being sure to register `Enhances` if {knitr} is dropped from `Suggests`.

NAMESPACE
```diff
@@ -104,6 +104,8 @@ if (getRversion() >= "4.0.0") {
...
+# else manual delayed registration from the onLoad hook
```
MichaelChirico, line 108: is this comment attached to the `knitr::knit_print` directive? If so either move to the end of the line or the previous line.

R/onLoad.R
```diff
@@ -66,6 +66,15 @@
       lockBinding("rbind.data.frame",baseenv())
     }
   }
+  if (session_r_version < "3.6.0") {
+    # no delayed registration support for NAMESPACE; perform it manually
+    if (isNamespaceLoaded("knitr")) {
```
MichaelChirico, line 71: this is a really nice defensive tactic for avoiding loading the `Enhances` namespace if we can 💪

MichaelChirico: IINM we don't need to do the `setHook()` step when `isNamespaceLoaded()`, so your original code was a no-op in that case; I put `setHook()` inside a branch for some clarity/to reduce potential confusion.

R/print.data.table.R
```diff
@@ -30,13 +30,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     # Other options investigated (could revisit): Cstack_info(), .Last.value gets set first before autoprint, history(), sys.status(),
     #   topenv(), inspecting next statement in caller, using clock() at C level to timeout suppression after some number of cycles
     SYS = sys.calls()
-    if (length(SYS) <= 2L ||  # "> DT" auto-print or "> print(DT)" explicit print (cannot distinguish from R 3.2.0 but that's ok)
-        ( length(SYS) >= 3L && is.symbol(thisSYS <- SYS[[length(SYS)-2L]][[1L]]) &&
-          as.character(thisSYS) == 'source') || # suppress printing from source(echo = TRUE) calls, #2369
```
MichaelChirico, line 35: could you confirm you've tested this PR is a no-op for the #2369 case as well?

aitap: Sorry, I shouldn't have removed that condition. Restored it now:

```
> .libPaths(c('data.table.Rcheck', .libPaths()))
writeLines('library(data.table)
DT = data.table(a = 1)
DT[ , a := 1]', 'test.R')
source("test.R", echo=TRUE)

> library(data.table)
data.table 1.16.99 IN DEVELOPMENT built 2024-12-03 19:02:16 UTC; aitap using 4 threads (see ?getDTthreads).  Latest news: r-datatable.com

> DT = data.table(a = 1)

> DT[ , a := 1]
>
```

MichaelChirico: Great! Would it be easy to add this as a test to autoprint.R?

https://github.com/Rdatatable/data.table/blob/98cf24e635ad47ba02617929d2237fca263e2b0c/tests/autoprint.R

R/onLoad.R
MichaelChirico: Good point. So not quite a no-op, even if it'll only show up in extremely odd use cases.

R/onLoad.R
```diff
@@ -66,6 +66,16 @@
       lockBinding("rbind.data.frame",baseenv())
     }
   }
+  if (session_r_version < "3.6.0") {
```
MichaelChirico, line 69: This should help keep the two intimately-linked pieces of logic in sync despite residing in totally different parts of the codebase

```suggestion
  if (session_r_version < "3.6.0") { # corresponds to S3method() directive in NAMESPACE
```

MichaelChirico (commented):
BTW please add a NEWS entry

aitap:
Good $localtime @MichaelChirico, the NEWS entry is now done too. Looks like were making commits at the same time.

MichaelChirico (approved):
Excellent, thanks!

(greetings from 🇸🇬 😄)

aitap:
Something happened in `knitr` again, see https://github.com/yihui/knitr/issues/2427. Thanks @simon-delage for the report and initial investigation.

PR #6595, opened 2024-10-31
Don't warn about numeric coercion in set(j=)

### LINKED ISSUE #6594: Don't warn about coercing j to integer
https://github.com/Rdatatable/data.table/blob/afa87e1eeb61922cdef512fefd86e3ba5847eef9/src/assign.c#L428

I find this warning a bit annoying. It's very common in R to have double "just work" as integer. Why this of all places is there an efficiency concern?

I can't imagine the efficiency concern is real anyway -- unless we are coercing 1e7+ column numbers to integer I can't imagine there's a real performance implication.

I also see two cases where this might come up:

1. You are passing `j` as a variable to `set()`:

    ```r
    setDT(DT, , col_idx, val)
    ```

    In this case, the recommendation is to do `as.integer(col_idx)`, which is what `assign.c` is doing implicitly. Is there really any difference in efficiency?

2. You are doing interactive work (as I just was when I came across this) and lazily omitting the `L`:

    ```r
    setDT(DT, , c(1, 4), NULL)
    ```

    OK, I _could_ write `1L`/`4L`, and in package/production code, I probably should. But to throw a _warning_ seems a bit heavy-handed. Users really under such use cases should be using [`lintr::implicit_integer_linter()`](https://lintr.r-lib.org/reference/implicit_integer_linter.html) to catch such cases.

Originally added a long time ago: e2cf7e2d8e6660562debe37f11e6998dc3e98f23

MichaelChirico:
Closes #6594

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -133,6 +133,8 @@ rowwiseDT(
 
 8. `measurev()` was implemented and documented in v1.15.0, for use within `melt()`, and it is now exported (dependent packages can now use without a NOTE from CRAN check).
 
+9. Using a double vector in `set()`'s `j=` no longer throws a warning about preferring integer, [#6594](https://github.com/Rdatatable/data.table/issues/6594). While it may improve efficiency to use integer, there's no guarantee it's an improvement and the difference is likely to be minimal. The coercion will still be reported under `datatable.verbose=TRUE`. For package/production use cases, static analyzers such as `lintr::implicit_integer_linter()` can also report when numeric literals should be rewritten as integer literals.
+
 # data.table [v1.16.0](https://github.com/Rdatatable/data.table/milestone/30)  (25 August 2024)
 
 ## BREAKING CHANGES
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Should we also cover the `i=` case here?

`i` is slightly more likely to have benefits from avoiding (e.g. 1e8 rows), but I think the same argument still basically applies.

If you received `row_idx` from out of your control (e.g. as output from another function), you'll still need to run `as.integer()` to overcome this warning. A wash.

MichaelChirico:
NB: the rchk error is unrelated to the current PR, probably an upstream change (though rchk has no recent changes, maybe the docker image was updated recently, or maybe there's a new LLVM version catching new issues):

Here's a PR run without this PR applied showing the same error; the change happened in the last 5 days since that commit to `master` shows clean.

https://github.com/Rdatatable/data.table/actions/runs/11616342257/job/32349061905

ben-schwen:
I would either change both i and j or none of them. I agree that it's strange that we warn about this, but having a verbose message seems appropriate. Educating ppl if they are eager to learn 😁

PR #6596, opened 2024-10-31
Fixes for new rchk issues

MichaelChirico:
These just started appearing. Either a new LLVM version, or possibly an update to the docker image we're using that is pulling in a more recent version of rchk (rchk itself has no recent source code updates).

Report:

https://github.com/Rdatatable/data.table/actions/runs/11616716077/job/32350299506?pr=6595

All look related to `getAttrib()` usage, which may be newly marked as requiring `PROTECT()`.

PR #6598, opened 2024-11-03
update github action versions

ben-schwen:
Current GH CI runs warn about the deprecation of certain old versions, e.g. for `upload-pages-artifact` we are currently using v1 although v3 is already deprecated.

### DIFF
--- a/.github/workflows/R-CMD-check-occasional.yaml
+++ b/.github/workflows/R-CMD-check-occasional.yaml
@@ -65,7 +65,7 @@ jobs:
           echo "LC_ALL=lv_LV.utf8" >> $GITHUB_ENV
           echo "LANGUAGE=lv_LV" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


Anirban166 (approved):
LGTM!

ben-schwen:
> are the CI failures on windows and rchk expected?

Those are other separate issues. Windows is #6602. rchk has no issue, but a PR #6596

tdhock:
ok great

MichaelChirico:
We might consider enabling Dependabot to automatically send PRs whenever our GHA versions are out-of-date:

https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts

MichaelChirico:
Also, just generally, our GHA are quite out of sync with what would be generated by [`usethis::use_github_action()`](https://usethis.r-lib.org/reference/use_github_action.html), without a strong reason to do so I'd recommend against it as that is designed in close collaboration with the authors of r-lib/actions & thus reliable as the best source of canonical GHA. If we need to customize I'd do so as patches on top of what {usethis} generates.

PR #6600, opened 2024-11-03
Fix divergence between CRAN version and master

### LINKED ISSUE #6587: Master diverges from patch 1.16.2
Currently, we have commits in our last CRAN version which are not part of current master, e.g. [9de2567](https://github.com/Rdatatable/data.table/commit/9de25675aa9f6efe982f4faff0ceb6ca55563563)

ben-schwen:
Closes #6587

This PR merge the `patch-1.16.2` branch into curent `master`. After resolving the conflicts it shows that we indeed are only missing #6563

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,13 +67,11 @@ rowwiseDT(
 
 ## BUG FIXES
 
-1. Using `print.data.table()` with character truncation using `datatable.prettyprint.char` no longer errors with `NA` entries, [#6441](https://github.com/Rdatatable/data.table/issues/6441). Thanks to @r2evans for the bug report, and @joshhwuu for the fix.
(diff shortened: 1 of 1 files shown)


MichaelChirico:
I believe I saw the 1.16.2 NEWS missing from master too, would you mind merging that in this PR as well?

ben-schwen:
> I believe I saw the 1.16.2 NEWS missing from master too, would you mind merging that in this PR as well?

Done, unfortunately I had to renumber the NEWS items manually.

TysonStanley:
Sorry about that, oversight on my part. Can PR the NEWS as soon as it's accepted to avoid having to renumber too much. Is there a good way to get the rest of the stuff in there or just a larger PR like this one with the relevant commits to the patch?

MichaelChirico:
> Is there a good way to get the rest of the stuff in there or just a larger PR like this one with the relevant commits to the patch?

I think this is new territory for all the maintainers, open to ideas. It definitely feels a bit clunky.

Issue #6554, opened 2024-09-30
Normalizing verbose messages in bmerge.R


rikivillalba:
Hello.
These are a some observations in verbose messages in bmerge.R I wish to share. 
- Some fragmented messages i.e. a `" (which contains no fractions)"` could be unified.
- The class/data types are parameterized arguments in some cases or have been harcoded in others, leading to inconsistency if someone uses to translate the data types.
- Hope if we parameterize some message the number of translatable strings may be reduced in two or three lines!

Here is the code with example proposals:

```
@@ -56,7 +56,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
         next
       } else {
         if (xclass=="character") {
-          if (verbose) catf("Coercing factor column %s to type character to match type of %s.\n", iname, xname)
+          if (verbose) catf("Coercing %s column %s to type %s to match type of %s.\n", "factor", iname, "character", xname)
           set(i, j=ic, value=val<-as.character(i[[ic]]))
           set(callersi, j=ic, value=val)  # factor in i joining to character in x will return character and not keep x's factor; e.g. for antaresRead #3581
           next
@@ -93,26 +93,30 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       nm = c(iname, xname)
       if (xclass=="integer64") { w=i; wc=ic; wclass=iclass; } else { w=x; wc=xc; wclass=xclass; nm=rev(nm) }  # w is which to coerce
       if (wclass=="integer" || (wclass=="double" && !isReallyReal(w[[wc]]))) {
-        if (verbose) catf("Coercing %s column %s%s to type integer64 to match type of %s.\n", wclass, nm[1L], if (wclass=="double") " (which contains no fractions)" else "", nm[2L])
+        if (verbose)
+          if (wclass=="double")
+            catf("Coercing %s column %s (which contains no fractions) to type %s to match type of %s.\n", wclass, nm[1L], "integer64", nm[2L])
+          else
+            catf("Coercing %s column %s to type %s to match type of %s.\n", wclass, nm[1L], "integer64", nm[2L])
         set(w, j=wc, value=bit64::as.integer64(w[[wc]]))
-      } else stopf("Incompatible join types: %s is type integer64 but %s is type double and contains fractions", nm[2L], nm[1L])
[...]

Issue #3506, opened 2019-04-15
Keyed subset of Date to character should "just work"


MichaelChirico:
```
DT = data.table(d = as.Date('2019-01-01'), key = 'd')

# fails since d is actually numeric
DT[.('2019-01-01')]

# this works since R coerces type for ==
DT[d == '2019-01-01']
```

Since we're already used to `==` implicitly coercing dates-as-strings to dates, I think there's reasonable expectation that join subsets would work similarly.

Logic is: if `x`'s join key `inherit`s from `Date` & `i`'s join key is `character`, try `as.IDate`.

Type laxness presents a weird edge case -- what if `x` has `.Date(0.1)`, should we `round` it first?

MichaelChirico:
@mattdowle maybe take a look at this one while #2784 is still fresh?

PR #6603, opened 2024-11-03
fix type coercion in bmerge

### LINKED ISSUE #6602: as.Date can result in different underlying types
R CMD check is currently failing on windows-devel with `typeof x.START_DATE (integer) != typeof i.DATE (double)`

The responsible code is explicitly casting to `as.Date`, so I guess this a bug in `R-devel`?

Seems related to this recent change in R-devel:
https://bugs.r-project.org/show_bug.cgi?id=18782

ben-schwen:
Closes #6602

Base does not encounter this problem since one join column can not be in multiple join conditions.

What do we do:
In bmerge we check if the types to merge x and i in `x[i]` are compatible. This also does some type coercion for example when the responsible columns are integer and double.
If a single column of i is in multiple conditions with columns of x, all of the to be joined columns need to coerce, this is why we introduce the extra check `length(ic_idx)>1L` and then iterate over the corresponding columns of x.


ben-schwen:
@MichaelChirico is it worth to make the `if` more specific to this case e.g. `(length(unique(icols))!= length(icols) || length(unique(xcols))!=length(xcols))` ensuring that either `icols` has a double condition or `xcols` has a double condition?

R/bmerge.R
```diff
@@ -71,7 +71,13 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, xclass, iname, iclass)
     }
     if (xclass == iclass) {
-      if (verbose) catf("%s has same type (%s) as %s. No coercion needed.\n", iname, xclass, xname)
+      if (length(icols)>1 && is(x[[xc]], "Date") && is(i[[ic]], "Date")) {
```
iago-pssjd, line 74: To modify only when strictly needed, could be added the next condition?
```r
&& identical(storage.mode(x[[xc]]), storage.mode(i[[ic]]))
```
(following https://bugs.r-project.org/show_bug.cgi?id=18782#c9)

MichaelChirico:
Sorry, I just saw that this is a CRAN requirement, checking now.

R/bmerge.R
MichaelChirico: I would use `inherits()`, `is()` is for S4 classes IMO (even though it happens to work for S3 classes).

MichaelChirico: BTW, IINM `POSIXct` can have the same concern -- {base} doesn't care about storing `POSIXct` as `integer` sometimes if possible.

R/bmerge.R
```diff
@@ -71,7 +71,13 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, xclass, iname, iclass)
     }
     if (xclass == iclass) {
```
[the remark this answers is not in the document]
MichaelChirico: another aside: It looks like we do `x[[xc]]` and `i[[ic]]` extraction at least one and up to several times. Should we just store the extracted column locally?

R/bmerge.R
ben-schwen: > I would use `inherits()`, `is()` is for S4 classes IMO (even though it happens to work for S3 classes).

Previously I checked if `class(x[[xc]])=="Date"` are equal and our linter nudged me towards `is`, so we might also want to adjust the linter :D

R/bmerge.R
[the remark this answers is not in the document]
ben-schwen: > aside: `{x,i}class` seem like poor name choices here, given that they typically come from `typeof()` (and `class()` is never called, except maybe internally by `inherits()`).

Yes but didn't want to change more than needed for the quick fix

ben-schwen:
> IMO `as.Date()` should _always_ coerce to `double` storage, even if it's possible to get `Date` objects with `integer` storage. Therefore I would hold off until (1) R-core confirms that yes, `as.Date(x)` might return `integer` sometimes; and/or (2) I get a clearer understanding of the underlying issue here.

Unfortunately this "contract" does not hold. I have a Windows dev version installed here and get the following
```r
> typeof(.Date(0L))
[1] "integer"
> typeof(as.Date(.Date(0L)))
[1] "integer"
```

R/bmerge.R
MichaelChirico: `class(x)=="Date"` is bad because `class()` can be `length>1`, c.f. `class(Sys.time())`, hence `==` is rarely intended.

The linter mentions `is()` as well as `inherits()` to cover all bases since there's no `is.Date()` and we didn't hand-code anything to cover base S3 classes:

```r
lintr::lint("class(x) == 'Date'\n", lintr::class_equals_linter())
# <text>:1:1: warning: [class_equals_linter] Instead of comparing class(x) with ==, use inherits(x, 'class-name') or is.<class> or is(x, 'class').
# class(x) == 'Date'
# ^~~~~~~~~~~~~~~~~~
```

MichaelChirico:
> I have a Windows dev version installed here and get the following

Thanks, also confirmed that on 4.4.1

TysonStanley:
This is strange as Kurt mentioned there wasn't an intention to do this but I figure it would have been fixed by now if that were the case.

R/bmerge.R
ben-schwen: My bad, seems that I overlooked the `inherits` part.

ben-schwen:
Update this fix now can convert into one direction from integer to double

```r
# this works
x = data.table(a=1L)
y = data.table(c=1L, d=2)
y[x, on=.(c==a, d==a)]
y[x, on=.(d==a, c==a)]
# this still needs to fixed
x = data.table(a=1)
y = data.table(c=1, d=2L)
y[x, on=.(c==a, d==a)]
y[x, on=.(d==a, c==a)]
```

MichaelChirico:
> this still needs to fixed

Should I consider this PR in-progress for now?

The diff has grown enough that it would help to add a brief overview of the changes to the PR description to orient reading

R/bmerge.R
```diff
@@ -34,86 +34,100 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
-          if (verbose) catf("Coercing factor column %s to type character to match type of %s.\n", iname, xname)
-          set(i, j=ic, value=val<-as.character(i[[ic]]))
-          set(callersi, j=ic, value=val)  # factor in i joining to character in x will return character and not keep x's factor; e.g. for antaresRead #3581
-          next
-        } else if (iclass=="character") {
-          if (verbose) catf("Matching character column %s to factor levels in %s.\n", iname, xname)
-          newvalue = chmatch(i[[ic]], levels(x[[xc]]), nomatch=0L)
-          if (anyNA(i[[ic]])) newvalue[is.na(i[[ic]])] = NA_integer_  # NA_character_ should match to NA in factor, #3809
-          set(i, j=ic, value=newvalue)
+  if (nrow(i)) {
```
MichaelChirico, line 37: Haven't read carefully yet, but given the description of the problem as I understand it now, I had expected something like this to be needed:

 (1) A pass over the join columns to determine the "join type" of each column
 (2) A second pass to actually do the coercions as needed

Does that make sense to you? If so, each step could be a helper function which would improve readability & modularity.

TysonStanley:
Just received the "deadline" on this from Kurt:

> Please correct before 2024-12-09 to safely retain your package on CRAN. Note that this will be the *final* reminder.

ben-schwen:
@MichaelChirico the PR is ready for review. What I'm unsure about is if we always want to try to copy at `cast_with_atts` or maybe want to introduce an extra argument for that. Moreover, I tried to keep everything quite modular.

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
+    if (x_merge_type == i_merge_type) {
+      if (verbose) catf("%s has same type (%s) as %s. No coercion needed.\n", iname, x_merge_type, xname)
       next
     }
-    if (xclass=="character" || iclass=="character" ||
-        xclass=="logical" || iclass=="logical" ||
-        xclass=="factor" || iclass=="factor") {
+    cfl = c("character", "logical", "factor")
+    if (x_merge_type %chin% cfl || i_merge_type %chin% cfl) {
+      msg = "Coercing all-NA %s column %s to type %s to match type of %s.\n"
```
MichaelChirico, line 90: nit: for translations we'll want to do

```r
msg = gettext(...)
...
catf(fmt, domain=NULL)
```

The run-time difference is 0, but we rely on static analysis to pull out the messages eligible for translation, which my suggestion will facilitate.

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
     ans
   }
 
+  cast_with_atts = function(x, as.f) {
```
MichaelChirico, line 37: should these helpers be declared in the {data.table} namespace instead (not exported)?

if `bmerge()` gets called dozens/hundreds of times, these helpers would have to be re-created each time. my approach is usually to put such helpers outside function bodies unless it's strongly beneficial for them to inherit directly from the function environment (not the case here AFAICT).

MichaelChirico: oh, I see `coerce_col` inherits `verbose`.

Suggestion: `verbose_msg = NULL`, then `if (!is.null(verbose_msg))` to replace `if (verbose)`

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
+      if (verbose) catf("%s has same type (%s) as %s. No coercion needed.\n", iname, x_merge_type, xname)
       next
     }
-    if (xclass=="character" || iclass=="character" ||
-        xclass=="logical" || iclass=="logical" ||
-        xclass=="factor" || iclass=="factor") {
-      if (anyNA(i[[ic]]) && allNA(i[[ic]])) {
-        if (verbose) catf("Coercing all-NA %s (%s) to type %s to match type of %s.\n", iname, iclass, xclass, xname)
-        set(i, j=ic, value=match.fun(paste0("as.", xclass))(i[[ic]]))
+    cfl = c("character", "logical", "factor")
```
MichaelChirico, line 88: weren't `factor` columns handled in the previous branch?

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
           next
         }
       }
-      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, xclass, iname, iclass)
+      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
     }
-    if (xclass == iclass) {
-      if (verbose) catf("%s has same type (%s) as %s. No coercion needed.\n", iname, xclass, xname)
+    if (x_merge_type == i_merge_type) {
+      if (verbose) catf("%s has same type (%s) as %s. No coercion needed.\n", iname, x_merge_type, xname)
```
MichaelChirico, line 85: Aside, probably for follow-up: it's a bit odd for this message to follow the `x_merge_type == "factor" || i_merge_type == "factor"` branch, since we get "no coercion needed" after coercion was already possibly done...

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
+    if (x_merge_type=="integer64" || i_merge_type=="integer64") {
       nm = c(iname, xname)
-      if (xclass=="integer64") { w=i; wc=ic; wclass=iclass; } else { w=x; wc=xc; wclass=xclass; nm=rev(nm) }  # w is which to coerce
+      if (x_merge_type=="integer64") { w=i; wc=icol; wclass=i_merge_type; } else { w=x; wc=xcol; wclass=x_merge_type; nm=rev(nm) }  # w is which to coerce
       if (wclass=="integer" || (wclass=="double" && !isReallyReal(w[[wc]]))) {
         if (verbose) catf("Coercing %s column %s%s to type integer64 to match type of %s.\n", wclass, nm[1L], if (wclass=="double") " (which contains no fractions)" else "", nm[2L])
         set(w, j=wc, value=bit64::as.integer64(w[[wc]]))
       } else stopf("Incompatible join types: %s is type integer64 but %s is type double and contains fractions", nm[2L], nm[1L])
     } else {
       # just integer and double left
```
MichaelChirico, line 109: Aside, again almost certainly for follow-up: that's not true, we haven't covered `complex` yet either.

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
-      if (iclass=="double") {
-        if (!isReallyReal(i[[ic]])) {
+      ic_idx = which(icol == icols) # check if on is joined on multiple conditions, #6602
```
MichaelChirico, line 110: On first read I think "this can be doing a lot of duplicate work since it's basically `for (icol in icols) icol==icols`", but actually it only applies to the subset `icols[icols != xcols & i_merge_type %in% c("integer", "double", "complex")]`.

`length(icols)` will not really be large in general (almost always `<10`), the subset there is even smaller, I doubt there's any real concern about the "lost efficiency" here unless you see something obvious (my first instinct is `duplicated()` but that's not quite right, everything else that comes to mind harms readability).

but it might be worth calling out the intentional redundancy in a comment.

R/bmerge.R
MichaelChirico: #6627

R/bmerge.R
MichaelChirico: Oh, nevermind, it's not possible to encounter this if we enter the previous branch -- it's always `next`'d. Still, it might make sense to move this step first as logically it's the simplest case & improves the flow of the code:

(1) if `i_merge_type == x_merge_type`, done, just move to next step
(2) now start checking what coercion is needed or it's impossible

MichaelChirico:
IINM this also closes #6554 (cc @rikivillalba)

MichaelChirico:
Aside for follow-up:

I think we should implement `mergeType` returning `I?Date`, `POSIXct` as separate cases. That would for example solve #3506 and could make the solution here clearer.

R/bmerge.R
```diff
@@ -34,86 +34,120 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
-          val = as.integer(i[[ic]])
-          if (!is.null(attributes(i[[ic]]))) attributes(val) = attributes(i[[ic]])  # to retain Date for example; 3679
-          set(i, j=ic, value=val)
-          set(callersi, j=ic, value=val)       # change the shallow copy of i up in [.data.table to reflect in the result, too.
-        } else {
-          if (verbose) catf("Coercing integer column %s to type double to match type of %s which contains fractions.\n", xname, iname)
-          set(x, j=xc, value=as.double(x[[xc]]))
+          if (length(ic_idx)>1L) {
+            xc_idx = xcols[ic_idx]
+            for (xb in xc_idx[which(vapply_1c(x[0L, xc_idx, with=FALSE], mergeType) == "double")]) {
```
MichaelChirico, line 119: nit: `vapply_1c(.shallow(x, xc_idx), mergeType)`? I'm wary of `[` overhead even if the resulting table is tiny...

ditto below

inst/tests/tests.Rraw
```diff
@@ -15727,7 +15727,8 @@ DT = data.table(z = 1i)
 test(2069.33, DT[DT, on = 'z'], error = "Type 'complex' is not supported for joining/merging")
 
 # forder verbose message when !isReallyReal Date, #1738
-DT = data.table(d=sample(seq(as.Date("2015-01-01"), as.Date("2015-01-05"), by="days"), 20, replace=TRUE))
+date_dbl = as.Date(as.double(seq(as.Date("2015-01-01"), as.Date("2015-01-05"), by="days")), origin="1970-01-01")
```
MichaelChirico, line 15730: do we still need to edit this test? I'd've thought one effect of this PR is to make the old test keep working as intended...

inst/tests/tests.Rraw
```diff
@@ -20596,3 +20597,29 @@ test(2295.3, is.data.table(d2))
 
 # #6588: .checkTypos used to give arbitrary strings to stopf as the first argument
 test(2296, d2[x %no such operator% 1], error = '%no such operator%')
+
+# coerce Dates to double if join on multiple columns, #6602
+x = data.table(a=1L)
+y = data.table(c=1L, d=1)
+test(2297.01, options=c(datatable.verbose=TRUE), y[x, on=.(c == a, d == a)], data.table(c=1L, d=1L), output="Coercing .*c to type double")
```
MichaelChirico, line 20604: nit: WDYT about using `verbose=TRUE` argument to `[` instead?

MichaelChirico: `a` also gets coerced, right? should that be reflected in the `output=` test?

MichaelChirico: (also, I might expect `d` to be the coerced column here...)

inst/tests/tests.Rraw
```diff
@@ -20596,3 +20597,29 @@ test(2295.3, is.data.table(d2))
...
+test(2297.02, options=c(datatable.verbose=TRUE), y[x, on=.(d == a, c == a)], data.table(c=1L, d=1L), output="Coercing .*c to type double")
+x = data.table(a=1)
+y = data.table(c=1, d=1L)
+test(2297.03, options=c(datatable.verbose=TRUE), y[x, on=.(c == a, d == a)], data.table(c=1L, d=1L), output="Coercing double column x.c (which contains no fractions) to type integer")
```
MichaelChirico, line 20608: Ditto about `a`'s coercion appearing in the output

inst/tests/tests.Rraw
```diff
@@ -20596,3 +20597,29 @@ test(2295.3, is.data.table(d2))
...
+test(2297.11, y[x, on=.(c == a, d == a)], data.table(c=d_int, d=d_int))
+test(2297.12, y[x, on=.(d == a, c == a)], data.table(c=d_int, d=d_int))
+x = data.table(a=d_dbl)
+y = data.table(c=d_dbl, d=d_int)
+test(2297.13, y[x, on=.(c == a, d == a)], data.table(c=d_int, d=d_int))
+test(2297.14, y[x, on=.(d == a, c == a)], data.table(c=d_int, d=d_int))
+# real double
+x = data.table(a=1)
+y = data.table(c=1.5, d=1L)
+test(2297.21, y[x, on=.(c == a, d == a)], data.table(c=1, d=1))
```
MichaelChirico, line 20624: It might help to include a non-join column in these tests, here at first glance it's surprising, but this is the right-join-by-default effect at play (i.e. there'd be no rows if `nomatch=NULL`).

R/bmerge.R
ben-schwen: Only if `x_merge_type` and `i_merge_type` are both `factor` "like". This branch is for handling coercing all NA values (mostly ofc coercing NA `logical`)

R/bmerge.R
ben-schwen: I agree that `i_merge_type == x_merge_type` would be best at top but then we would have doubled logic for rolling joins on `factors`

MichaelChirico: who on earth is doing rolling joins on factors 😂

thanks for explaining, it seems worthy of a comment

R/bmerge.R
ben-schwen: done

inst/tests/tests.Rraw
ben-schwen: Yes, we alter this test to become future proof and ensure that we have a indeed a `double`, no matter what `R` decides what the underlying type should be

inst/tests/tests.Rraw
ben-schwen: done

inst/tests/tests.Rraw
ben-schwen: done

inst/tests/tests.Rraw
ben-schwen: done

R/bmerge.R
ben-schwen: done

R/bmerge.R
ben-schwen: Also thought that it might be better to just iterate over `unique(icols)`, but this only applies on this uncommon case of merging a single column on multiple conditions. I'll file a separate issue to follow up on this.

NEWS.md
```diff
@@ -107,6 +107,8 @@ rowwiseDT(
 
 11. `tables()` now returns the correct size for data.tables over 2GiB, [#6607](https://github.com/Rdatatable/data.table/issues/6607). Thanks to @vlulla for the report and the PR.
 
+12. Joins on multiple columns, such as `x[y, on=c("x1==y1", "x2==y1")]`, could fail during type coercions if `x1` and `x2` had different but still compatible types, [#6602](https://github.com/Rdatatable/data.table/issues/6602). This was particularly unexpected when all columns `x1, x2, and y1` were of the same class e.g. `Date` but differed in their underlying types. Thanks to Benjamin Schwendinger for the report and the fix.
```
MichaelChirico, line 110:
```suggestion
12. Joins on multiple columns, such as `x[y, on=c("x1==y1", "x2==y1")]`, could fail during implicit type coercions if `x1` and `x2` had different but still compatible types, [#6602](https://github.com/Rdatatable/data.table/issues/6602). This was particularly unexpected when columns `x1`, `x2`, and `y1` were all of the same class, e.g. `Date`, but differed in their underlying storage types. Thanks to Benjamin Schwendinger for the report and the fix.
```

MichaelChirico (approved):
Awesome!

PR #6610, opened 2024-11-10
close #6605, joins on date/time do not warn

### LINKED ISSUE #6605: Joins do not warn user when using POSc and Date comparisons
## Issue
There is a possible silent warning/error that occurs when joining on POSc and Date columns. It is unclear if this is expected behaviour based on my readings.

The example below shows how the same tables (dt1 and dt2) either join or fail to (left) join based upon the format of the date/time columns in dt1 (either POSc or Date). Although the simple <= and >= comparisons on vectors pass a warning, there is no warning passed when performing a join, which led to confusion as to the cause of join difference.

## Example

```r
library(data.table)

# POSc
dt1_posc <- data.table(id = 1,
		date_time1 = as.POSIXct('2017-01-01'),
		date_time2 = as.POSIXct('2020-01-01'), 
		infocol = 'test')
dt1 <- data.table(id = 1,
		date_time1 = as.Date('2017-01-01'),
		date_time2 = as.Date('2020-01-01'), 
		infocol = 'test')

dt2 <- data.table(id = c(1, 1, 2),
		date_1=as.Date(c('2018-01-01', '2019-01-01', '2010-01-01')),
		date_2=as.Date(c('2018-06-01', '2020-11-01', '2010-01-01')))

# Works
dt1[dt2, on = .(id, date_time1 <= date_2, date_time2 >= date_1)]
#>      id date_time1 date_time2 infocol
#>   <num>     <Date>     <Date>  <char>
#>1:     1 2018-06-01 2018-01-01    test
#>2:     1 2020-11-01 2019-01-01    test
#>3:     2 2010-01-01 2010-01-01    <NA>

# Doesnt work
dt1_posc[dt2, on = .(id, date_time1 <= date_2, date_time2 >= date_1)]
#>      id date_time1 date_time2 infocol
#>   <num>     <Date>     <Date>  <char>
#>1:     1 2018-06-01 2018-01-01    <NA>
#>2:     1 2020-11-01 2019-01-01    <NA>
#>3:     2 2010-01-01 2010-01-01    <NA>

# Check simple comparison
dt1_posc$date_time1 <= dt2$date_2 & dt1_posc$date_time2 >= dt2$date_1
#> [1] FALSE FALSE FALSE
#> Warning message:
#> Incompatible methods ("Ops.POSIXt", "Ops.Date") for "<=" 
dt1$date_time1 <= dt2$date_2 & dt1$date_time2 >= dt2$date_1
#> [1]  TRUE  TRUE FALSE

```

## Session info
I used the latest version of data.table on CRAN (1.16.2) in the following session

```
R version 4.3.1 (2023-06-16 ucrt)

Matrix products: default


locale:
[1] LC_COLLATE=English_Canada.utf8  LC_CTYPE=English_Canada.utf8    LC_MONETARY=English_Canada.utf8 LC_NUMERIC=C                    LC_TIME=English_Canada.utf8    

time zone: America/Edmonton
tzcode source: internal

attached base packages:

other attached packages:
[1] data.table_1.16.2

[1] compiler_4.3.1    tools_4.3.1       rstudioapi_0.15.0 
```

r2evans:


NEWS.md
```diff
@@ -107,6 +107,8 @@ rowwiseDT(
 
 11. `tables()` now returns the correct size for data.tables over 2GiB, [#6607](https://github.com/Rdatatable/data.table/issues/6607). Thanks to @vlulla for the report and the PR.
 
+16. Joining with incompatible column times (e.g., `Date` with `POSIXt`) now provides a clear warning, [#6605](https://github.com/Rdatatable/data.table/issues/6605). Thanks to @al-obrien for the report and @r2evans for the PR.
```
ben-schwen, line 110:
```suggestion
12. Joining with incompatible column times (e.g., `Date` with `POSIXt`) now provides a clear warning, [#6605](https://github.com/Rdatatable/data.table/issues/6605). Thanks to @al-obrien for the report and @r2evans for the PR.
```

inst/tests/tests.Rraw
```diff
@@ -20596,3 +20596,6 @@ test(2295.3, is.data.table(d2))
 
 # #6588: .checkTypos used to give arbitrary strings to stopf as the first argument
 test(2296, d2[x %no such operator% 1], error = '%no such operator%')
+
+# #6605: Joins do not warn user when using POSc and Date comparisons
+test(2297.1, data.table(a = Sys.time(), v1 = 1)[data.table(a = Sys.Date(), v2 = 2), on = "a"], output = ".+", warning = "incompatible")
```
ben-schwen, line 20601:
```suggestion
test(2297.1, data.table(a = .POSIXct(20000L), v1 = 1)[data.table(a = .Date(20000L), v2 = 2), on = "a"], output = ".+", warning = "incompatible")
```

ben-schwen:
Please also add yourself to the DESCRIPTION file.

MichaelChirico:
I actually think it would be even better to try and coerce among these classes where possible on the user's behalf, but anyway this PR is an improvement that we can always iterate on.

MichaelChirico:
#6728 touches very similar code, I would merge that one first and then re-work this to use `coerce_col()` as there.

r2evans:
> I actually think it would be even better to try and coerce among these classes where possible on the user's behalf, but anyway this PR is an improvement that we can always iterate on.

First, `coerce_col` isn't going to work without some changes:

```r
data.table(d=Sys.Date()) |>
  data.table:::coerce_col("d", "POSIXct", "Date") |>
  _[]
#             d
#        <Date>
# 1: 2025-04-07
```

Second, just to be clear: since dates are all UTC, "midnight" alignment is likely to be different ... for instance, it might be counter-intuitive to users that the bottom two rows are a different date:

```r
(tms <- as.POSIXct(Sys.Date() + 0:1) |> `attr<-`("tzone", "America/Denver"))
# [1] "2025-04-06 18:00:00 MDT" "2025-04-07 18:00:00 MDT"
data.table(tm = seq(as.POSIXct("2025-04-07 00:00:00", tz = "America/Denver"), length.out=12, by="2 hour")) |>
  _[, z := between(tm, tms[1], tms[2])] |>
  _[]
#                      tm      z
#                  <POSc> <lgcl>
#  1: 2025-04-07 00:00:00   TRUE
#  2: 2025-04-07 02:00:00   TRUE
#  3: 2025-04-07 04:00:00   TRUE
#  4: 2025-04-07 06:00:00   TRUE
#  5: 2025-04-07 08:00:00   TRUE
#  6: 2025-04-07 10:00:00   TRUE
#  7: 2025-04-07 12:00:00   TRUE
#  8: 2025-04-07 14:00:00   TRUE
#  9: 2025-04-07 16:00:00   TRUE
# 10: 2025-04-07 18:00:00   TRUE
# 11: 2025-04-07 20:00:00  FALSE
# 12: 2025-04-07 22:00:00  FALSE
```

I've been trying to wrap my brain around TZ issues, looking for something that is robust, user-intuitive, and data-safe, and I cannot find peace with it.

I suggest for now we stick with this PR as-is, just warn to the user that they are responsible for converting and ensuring compliance with their intuition.

R/bmerge.R
```diff
@@ -92,6 +92,14 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       }
       stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
     }
+    # data.table::as.ITime, chron::times, nanotime::nanotime
+    time_classes = c("Date", "POSIXt", "ITime", "times", "nanotime")
+    x_class_time = intersect(class(x[[xcol]]), time_classes)
+    i_class_time = intersect(class(i[[icol]]), time_classes)
+    if (length(x_class_time) > 0L && length(i_class_time) > 0L && !identical(x_class_time, i_class_time)) {
```
MichaelChirico: I removed the `sort()` check of the classes -- I figure if someone's using some weird ordering of their S3 class, throwing a warning can't hurt. It may well be that having mismatched class ordering results in some unexpected behavior anyway if the "wrong" dispatch eventually happens.

R/bmerge.R
```diff
@@ -25,6 +25,21 @@ coerce_col = function(dt, col, from_type, to_type, from_name, to_name, from_deta
...
+.maybe_warn_mismatched_time_types = function(x_class, i_class, x_name, i_name) {
+  x_class_time = intersect(class(x[[xcol]]), .known_time_classes)
+  if (!length(x_class_time)) return(invisible())
+
+  i_class_time = intersect(class(i[[icol]]), .known_time_classes)
+  if (!length(i_class_time)) return(invisible())
+
+  if (identical(x_class_time, i_class_time)) return(invisible())
+
+  warningf("Attempting to join column %s (%s) with column %s (%s). They are likely to be incompatible, so we suggest you convert one to the other's class.",
```
MichaelChirico, line 40: "they are likely to be incompatible" feels like not the best wording. What's meant by "incompatible" here?

One thing I have in mind is, it'd be nice to recommend the user which conversion to do, e.g. they should only do POSIXct-->Date if they're OK with some precision loss, whereas they can do Date-->POSIXct WLOG.

OTOH, nanotime and POSIXt may be incompatible in a much deeper sense, that there are times mutually non-representable between 64-bit storage epoch nanosecond precision and 64-bit epoch second precision.

inst/tests/tests.Rraw
```diff
@@ -20775,6 +20775,9 @@ x = data.table(a=1, b=2L)
 y = data.table(c=1.5, d=1L)
 test(2297.31, y[x, on=.(c == a, d == a), nomatch=NULL], output="Empty data.table (0 rows and 3 cols): c,d,b")
 
+# Ensure a warning when trying problematic join on mismatched time types, #6605
+test(2297.41, data.table(a = .POSIXct(20000L), v1 = 1)[data.table(a = .Date(20000L), v2 = 2), on = "a"], output = ".+", warning = "incompatible")
```
MichaelChirico, line 20779: what's the purpose of the `output=".+"` check?

R/bmerge.R
r2evans: I don't disagree, though while it's always nice to suggest a remedy, with several less-unambiguous class-comparisons I didn't spend the time to think through all of them. Perhaps that was me being a little lazy at the time.

R/bmerge.R
```diff
@@ -92,6 +108,7 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
       }
       stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
     }
+
```
MichaelChirico, line 111:
```suggestion
```

R/bmerge.R
```diff
@@ -160,6 +176,8 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
     }
   }
 
+  .maybe_warn_mismatched_time_types(class(x[[xcol]]), class(i[[icol]]), x_name, i_name)
```
MichaelChirico: moved the check later in the body -- I think it's possible to run a join where it would have thrown this warning, only to immediately then error because of mismatched 64-bit and 32-bit classes. better to only warn once the compatibility of the storage types has been confirmed.

MichaelChirico: Should add a test about this in fact, #7123

Issue #6613, opened 2024-11-12
translated vignette proposal : all pages link back to EN root vignette


ChristianWia:
The EN vignette holds links towards translated pages (for example https://rdatatable.gitlab.io/data.table/articles/datatable-intro.html leads to  FR transltion in -> https://rdatatable.gitlab.io/data.table/articles/fr/datatable-intro.html 

In the translated page, there is no longer useful to reference the page itself, nor all other existing translations (for page maintenance reasons when new language is added) 

So my proposal would be that all translated pages display the same equivalent of  :

Translations of this document is available in

    [English](https://rdatatable.gitlab.io/data.table/articles/datatable-intro.html)

rikivillalba:
Some chunk could be provided that generate the link list, provided the corresponding files and directories do exist
I.e.
```
```{r echo = FALSE}
  url = "https://rdatatable.gitlab.io/data.table/articles"
  path = dirname(knitr::current_input(TRUE))
  if (basename(path) == "vignettes") {
    lang = "en"
  } else {
    lang = basename(path)
    path = dirname(path)
  }

  translation = dir(path,
    recursive = TRUE,
    pattern = glob2rx(knitr::current_input(FALSE))
  )
  transl_lang = ifelse(dirname(translation) == ".", "en", dirname(translation))

  block = sprintf("%s\n", paste(sprintf(
    " [%s](%s) ",
    transl_lang[transl_lang != lang],
    file.path(url, sub("(?i)\\.Rmd$", ".html", translation[transl_lang != lang]))
  )))

  knitr::asis_output(block)
``` #
```

rikivillalba:
Relevant: #6455

tdhock:
the idea of automatically generated links between languages is interesting, but where would you put that code? (to avoid duplicated logic in each vignette)

Issue #6394, opened 2024-08-23
Explore {litedown} for rendering vignettes


MichaelChirico:
{knitr} requires R>=3.6.0; that means we can't render vignettes on our oldest supported versions of R (3.3, 3.4, 3.5).

@jangorecki sugegsts trying to render them instead with {litedown}:

https://github.com/Rdatatable/data.table/commit/d66e12bd884ee37d46952d65fe808a6a2a43b6c7#r145771409

https://cran.r-project.org/web/packages/litedown/index.html

https://github.com/yihui/litedown

{litedown} is only 5 months old:

https://github.com/yihui/litedown/commit/03db9970b290e3194942d5817140387ef5aebb6d

So we can expect to be among the first "major" users of the package if we make the switch -- I expect that to generate a decent amount of work filing/even fixing issues with upstream.

@jangorecki do you want to take the lead on owning the potential switch (exploring/validating/maintaining)?

jangorecki:
IMO we should as well move away from pkgdown to new https://github.com/yihui/litedown/issues/24#issuecomment-2384933793
Pkgdown was causing extra burden with dependencies (new and many OS deps) and was at some point cached in a docker image, so it is not even updated in our CI for a longer while now.

ben-schwen:
> IMO we could as well move away from pkgdown to new [yihui/litedown#24 (comment)](https://github.com/yihui/litedown/issues/24#issuecomment-2384933793) Pkgdown was causing extra burden with dependencies (new and many OS deps) and was at some point cached in a docker image, so it is not even updated in our CI for a longer while now.

I would keep this as an extra issue, one step at a time!

PR #6617, opened 2024-11-15
Using hyperlinks instead of vignette() calls for readability; doing the same for vignette titles without links for consistency

Anirban166:
Closes #6612

PR #6618, opened 2024-11-16
Automate lang links

rikivillalba:
In response to #6613

This is my proposal to automate link generation:

- The logic is in .translation_links.R in the vignettes directory, and it is sourced in each vignette
- The message "Translations of this document are available in the following languages" is translated in the same source (there is a switch which defaults to english)
- Currently the output format is   en | fr | es | ..., that is, not a list but all inline. 

Note that currently, in documents with a TOC (datatable-faq.Rmd) the link list is below the toc in the packaged vignettes. That is unchanged as the TOC is built automatically by commonmark.
Also, #6394 should be considered.


vignettes/.translation_links.R
```diff
@@ -0,0 +1,32 @@
...
+  recursive = TRUE,
+  pattern = glob2rx(knitr::current_input(FALSE))
+)
+
+transl_lang = ifelse(dirname(translation) == ".", "en", dirname(translation))
+
+block = sprintf(
+  "%s: %s\n", 
+  switch(lang,
+    fr = "Des traductions de ce document sont disponibles dans les langues suivantes",
```
tdhock, line 23: This method for defining translations is non-standard. I think this would be better as gettext(...)
or to simplify, just use English. Only list the other languages at the top of the English vignette. and then the other vignettes can just link back to English.

rikivillalba: Yes, I think in gettext however I don't figure out how to get the string for the appropriate language to compile the vignettes in all languages, other than set `Sys.setLanguage()` in each call.
Another option is to hardcode the header and let the code generate only the links.
But I think your second option is ok.

rikivillalba:
I made the following changes:
- The script is now is a function to which the "translations available" message is an arg, so it is not harcoded but passed from the vignette text.
- The message only prints if there are any translations other than current language (as now french version of datatable-joins.Rmd is missing)
- renamed .translation_links.R to _translation_links.R so ls can see it (it is part of the vignettes after all)
- NOTE: **I manually merge** datatable-keys-fast-subset.Rmd, datatable-reference-semantics.Rmd and datatable-secondary-indices-and-auto-indexing.Rmd because #6617 modified lines adjacent to this branch. 

_translation_links.R must be bundled in the package so r CMD check can build the vignettes. However that source is 'knitted' and present in each .R file generated from .Rmd and installed in doc.

French vignettes are no modified at all. An equivalent to "from English *original* vignette" message linking back to english might be added to translated vignettes, or code similar to that of the english ones, it is up to the owner to decide.  In the script case it will add all available links excluding that of the current vignette language determined from the folder name. 

As part of spanish translators there is work in progres in https://github.com/cienciadedatos/traduccion-vignettes-datatable 
tks

vignettes/_translation_links.R
```diff
@@ -0,0 +1,25 @@
+# build a link list of alternative languages (may be character(0))
+.write.translation.links <- function(fmt) {
+    url = "https://rdatatable.gitlab.io/data.table/articles"
+    path = dirname(knitr::current_input(TRUE))
```
MichaelChirico, line 4: is there a litedown equivalent (#6583)?

vignettes/_translation_links.R
```diff
@@ -0,0 +1,25 @@
...
+    } else {
+      lang = basename(path)
+      path = dirname(path)
+    }
+    translation = dir(path,
+      recursive = TRUE,
+      pattern = glob2rx(knitr::current_input(FALSE))
+    )
+    transl_lang = ifelse(dirname(translation) == ".", "en", dirname(translation))
+    block = if (length(which(transl_lang != lang))) {
```
MichaelChirico, line 16:
```suggestion
    block = if (!all(transl_lang == lang)) {
```

vignettes/_translation_links.R
```diff
@@ -0,0 +1,25 @@
+# build a link list of alternative languages (may be character(0))
```
MichaelChirico, line 1:
```suggestion
# build a link list of alternative languages (may be character(0))
# idea is to look like 'Other languages: en | fr | de'
```

Just commenting on expected output to facilitate reading/future refactoring

vignettes/_translation_links.R
```diff
@@ -0,0 +1,25 @@
...
+    )
+    transl_lang = ifelse(dirname(translation) == ".", "en", dirname(translation))
+    block = if (length(which(transl_lang != lang))) {
+      sprintf(
+        fmt, paste(collapse = " | ", sprintf(
+          "[%s](%s)",
+          transl_lang[transl_lang != lang],
+          file.path(url, sub("(?i)\\.Rmd$", ".html", translation[transl_lang != lang]))
+      )))
+    } else ""
```
MichaelChirico, line 23: WDYT about leaving unlinked text in for the current language instead?

Available languages: [fr](http://google.com) | [es](http://google.com)

vs

Available languages: en | [fr](http://google.com) | [es](http://google.com)

vignettes/datatable-benchmarking.Rmd
```diff
@@ -25,9 +25,9 @@ h2 {
 }
 </style>
 
-Translations of this document are available in
-
-* [French](https://rdatatable.gitlab.io/data.table/articles/fr/datatable-benchmarking.html)
+```{r echo=FALSE, file='_translation_links.R'}
+```
+`r .write.translation.links("Translations of this document are available in: %s")`
```
MichaelChirico, line 30: IIUC the idea is the translation team will just translate this string themselves directly in the .Rmd file, is that right? (seems fine, just want to confirm my understanding)

vignettes/_translation_links.R
MichaelChirico, line 1: we also have this helper script, let's keep a consistent naming style (personally, I like your snake case style here):

https://github.com/Rdatatable/data.table/blob/master/vignettes/.check.translations.R

Also note that this should probably be added to `.Rbuildignore`? Maybe just ignore all `vignettes/_.*` files?

vignettes/datatable-benchmarking.Rmd
rikivillalba: Exactly.

vignettes/_translation_links.R
rikivillalba: The only alternative I found was `litedown:::.env$input`, that stores the file name (the input path to `litedown::fuse()`). Tested to work building the package from . and from ..

vignettes/_translation_links.R
rikivillalba: Sadly, adding it to .Rbuildignore causes R check to fail. It checks the vignettes to be able to be built from the tarball bundle alone.

MichaelChirico: oh got it. I guess that's why it must be prefixed with `_` to not be "hidden", which in turn is why there's no R CMD check failure.

I am only concerned about R CMD check passing, it is fine to include it.

rikivillalba:
This is the language-link printing function, with the propositions, adapted to litedown. Must be after #6583

```r
# build a link list of alternative languages (may be character(0))
# idea is to look like 'Other languages: en | fr | de'
.write.translation.links <- function(fmt) {
  url = "https://rdatatable.gitlab.io/data.table/articles"
  # TODO: better litedown alternative to knitr::current_input()
  # fuse() setwd the .Rmd dir before evaluating chunks
  input_file = basename(litedown:::.env$input)
  path = dirname(normalizePath(input_file, mustWork = TRUE))
  
  if (basename(path) == "vignettes") {
    lang = "en"
  } else {
    lang = basename(path)
    path = dirname(path)
  }
  translation = dir(path,
    recursive = TRUE,
    pattern = glob2rx(input_file)
  )
  transl_lang = ifelse(dirname(translation) == ".", "en", dirname(translation))
  block = if (!all(transl_lang == lang)) {
    sprintf(fmt, paste(collapse = " | ", ifelse(
      transl_lang != lang,
      sprintf("[%s](%s)", transl_lang,  file.path(url, sub("(?i)\\.Rmd$", ".html", translation))),
      transl_lang     # print no link, only language code.
    )))
  } else ""
  block
}
```

MichaelChirico:
Filed https://github.com/yihui/litedown/issues/67 requesting a public interface equivalent to `knitr::current_input()`

vignettes/_translation_links.R
```diff
@@ -0,0 +1,26 @@
...
+      pattern = glob2rx(knitr::current_input(FALSE))
+    )
+    transl_lang = ifelse(dirname(translation) == ".", "en", dirname(translation))
+    block = if (!all(transl_lang == lang)) {
+      sprintf(
+        fmt, paste(collapse = " | ", sprintf(
+          "[%s](%s)",
+          transl_lang[transl_lang != lang],
+          file.path(url, sub("(?i)\\.Rmd$", ".html", translation[transl_lang != lang]))
+      )))
```
MichaelChirico, line 23: (untested) I think this will get the effect of including the current language unlinked

```suggestion
      linked_transl = sprintf("[%s](%s)", transl_lang, file.path(url, sub("(?i)\\.Rmd$", ".html", translation))
      linked_transl[transl_lang == lang] = lang
      sprintf(fmt, paste(linked_transl, collapse = " | "))
```

rikivillalba: I tested your suggestion succesfully, this is what looks like (pkgdown build)
image
image
image

Link's urls are absolute, as they are expected to be followed from vignettes in a path where translations are not available. I don't know whether it is possible to specify to generate relative paths when built on pkgdown but keep absolute paths otherwise. 

Also: if there is no translation available, no message is generate at all (which I think it's ok).

Note. Missing ) in line 18.
```suggestion
      linked_transl = sprintf("[%s](%s)", transl_lang, file.path(url, sub("(?i)\\.Rmd$", ".html", translation)))
      linked_transl[transl_lang == lang] = lang
      sprintf(fmt, paste(linked_transl, collapse = " | "))
```

aitap:
My apologies to everyone involved for pushing e7e966ea1cd9497318c007f3b51855e033647665 straight to master. I forgot to create a branch first.

Does this PR need any additional work, besides cleaning up the merge conflicts, before it can be merged?

rikivillalba:
> My apologies to everyone involved for pushing [e7e966e](https://github.com/Rdatatable/data.table/commit/e7e966ea1cd9497318c007f3b51855e033647665) straight to master. I forgot to create a branch first.
> 
> Does this PR need any additional work, besides cleaning up the merge conflicts, before it can be merged?

I think this pr is linked to #6583, and to the litedown feature akin to `knitr::current_input()`: `litedown::get_context("input")` not yet in CRAN.

MichaelChirico:
> I think this pr is linked to #6583, and to the litedown feature akin to `knitr::current_input()`: `litedown::get_context("input")` not yet in CRAN.

I don't think that's required to merge here. If you think this is ready in its current form, let's merge, we can mark the litedown translation part for follow-up.

rikivillalba:
> > I think this pr is linked to #6583, and to the litedown feature akin to `knitr::current_input()`: `litedown::get_context("input")` not yet in CRAN.
> 
> I don't think that's required to merge here. If you think this is ready in its current form, let's merge, we can mark the litedown translation part for follow-up.

Added the code sections to generate the translation links. this is the status:
- All vignettes currently on master are updated, I merged the most recent content and added the code to russian and french vignettes for ease of merging.
- Building the articles via `pkgdown::build_articles()` works as expected.
Pending:
- The message header for translation links *is not translated yet* (i'm handing off to translators). The english is `"Translations of this document are available in: %s"`. Using gettext here would imply changing locale before running each vignette, someting I think is not worth and I didn' tested.
Observations:
- Merging should probably be manually solved for the code printing the links being just aside the original links.

MichaelChirico (approved):
I think this is good to go :shipit:

We can iterate more later.

PR #6621, opened 2024-11-19
Minor fix on text redaction

### LINKED ISSUE #5456: Documentation bug in datatable-reshape vignette
In the current `datatable-reshape` vignette in CRAN, https://cran.r-project.org/web/packages/data.table/vignettes/datatable-reshape.html, in the section corresponding to
https://github.com/Rdatatable/data.table/blob/c4a2085e35689a108d67dacb2f8261e4964d7e12/vignettes/datatable-reshape.Rmd#L157-L174

the variables `dob` and `gender` become `IDate` format in `DT.c1`, so the types `character` and `integer` mentioned at point 2 (L174) seem to be wrong.

iago-pssjd:
Updating the redaction, which currently disagrees with given example,

Closes #5456

Further, updates heading which is a bit outdated.


tdhock (approved):
great thanks

PR #6626, opened 2024-11-26
Allow double-integer64 joins when double is in (integer32 , integer64] range

### LINKED ISSUE #4167: bmerge errors on integer64 to double join
Just came across this... haven't had time to minimize the example but here goes:

```
library(DBI)
library(RSQLite)
library(data.table)
# download and unzip file from here:
#   https://ankiweb.net/shared/info/867291675
card_db = dbConnect(SQLite(),  'collection.anki2')
cards = setDT(dbGetQuery(card_db, 'select * from cards'))
cards[id == 1419644220828]
#   Incompatible join types: x.id is type integer64 but i.id is type double and contains fractions
```

### LINKED ISSUE #6625: Misleading (and incorrect?) error about incompatible column types in integer64 join
https://github.com/Rdatatable/data.table/blob/546259ddaba0e8ab1506729113688f85ca2986fd/R/bmerge.R#L98

The "and contains fractions" part really threw off a debugging effort -- typically when that comes up in my experience, it's because the "integer64" class got dropped and we wound up trying to join (integer64, unclassed-integer64-which-is-tiny-double).

Observe:

```r
library(bit64)
i64_val = .Machine$integer.max + 1
DT1 = data.table(id = as.integer64(i64_val))
DT2 = data.table(id = i64_val)
DT1[DT2, on='id']
#   Incompatible join types: x.id is type integer64 but i.id is type double and contains fractions
```

But there's no fraction! It's just that `i64_val` is not integer-representable.

At a minimum we should correct the error message. But even better would be to recognize that we are perfectly capable of joining these two tables _if we coerce to integer64_.

(actually, it's a bit more complicated than that in general -- we need to ensure that the two values can be represented in the same class, since there are integer64 that cannot be represented in double and vice versa)

MichaelChirico:
Closes #4167.

Probably has some conflict with #6602 -- that one is the priority since it's a CRAN blocker.

I decided to totally ditch `isReallyReal()` -- it's only ever used in `if()` statements, so the feature to return the first true-real index is not useful. That leads to the IMO much more readable `isRealReallyInt{32,64}` approach here.


ben-schwen:
LGTM, besides that I find the current naming of `isRealReally` not ideal

inst/tests/tests.Rraw
```diff
@@ -7359,6 +7359,7 @@ test(1514.6, isRealReallyInt32(x), FALSE)
 x = numeric()
 test(1514.7, isRealReallyInt32(x), TRUE)
 test(1514.8, isRealReallyInt32(9L), FALSE) # b/c not double input
+test(1514.9, isRealReallyInt64(9L), FALSE)
```
MichaelChirico, line 7362: thanks! no harm keeping it now that you've already gone to the trouble, but for future reference these days I prefer not to test private API directly, especially not just for codecov sake.

ben-schwen: OK noted

src/utils.c
```diff
@@ -11,30 +11,40 @@ bool within_int64_repres(double x) {
   return R_FINITE(x) && x <= (double)INT64_MAX && x >= (double)INT64_MIN;
 }
 
-static R_xlen_t firstNonInt(SEXP x) {
+// used to error if not passed type double but this needed extra is.double() calls in calling R code
+// which needed a repeat of the argument. Hence simpler and more robust to return false when not type double.
+bool isRealReallyInt32(SEXP x) {
```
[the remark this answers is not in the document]
MichaelChirico: Hmm I'm rethinking. That naming implies a symmetry, e.g. suggesting it could accept `integer` inputs also.

WDYT about `isInt32AsReal`? `realFitsInInt32`/`fitsInInt32`? That highlights we are only interested in whether something currently stored as `double` could be stored as `int` instead.

(for the `int` (32) case, the symmetry part is kinda trivial, because `int` is a strict subset of `double`, but the symmetry _does not_ apply to the `int64_t` case, hence why I'm hesitating)

MichaelChirico:
I'll go ahead and merge since the routine name is just a minor detail. We can easily file a follow-up PR to fine-tune.

Issue #1122, opened 2015-04-19
`:=` not suppressing print in R 3.2.0


jangorecki:
Worth to note that `test.data.table` passed successfully.
```r
library(data.table)
# data.table 1.9.5  For help type: ?data.table
# *** NB: by=.EACHI is now explicit. See README to restore previous behaviour.
DT <- data.table(a=1L)
DT[, a:=2L]
#    a
# 1: 2
sessionInfo()
# R version 3.2.0 (2015-04-16)
# Platform: x86_64-pc-linux-gnu (64-bit)
# Running under: Ubuntu 14.04.2 LTS
# 
# locale:
#  [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
#  [3] LC_TIME=en_DK.UTF-8        LC_COLLATE=en_US.UTF-8    
#  [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C             
#  [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
#  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
# [11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       
# 
# attached base packages:
# [1] stats     graphics  grDevices utils     datasets  methods   base     
# 
# other attached packages:
# [1] data.table_1.9.5
# 
# loaded via a namespace (and not attached):
# [1] chron_2.3-45
```

shrektan:
I encounter the same problem .
Here is my `sessioninfo`:

```r
R version 3.2.0 (2015-04-16)
Platform: i386-w64-mingw32/i386 (32-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

locale:
[1] LC_COLLATE=Chinese (Simplified)_People's Republic of China.936 
[2] LC_CTYPE=Chinese (Simplified)_People's Republic of China.936   
[3] LC_MONETARY=Chinese (Simplified)_People's Republic of China.936
[4] LC_NUMERIC=C                                                   
[5] LC_TIME=Chinese (Simplified)_People's Republic of China.936    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] data.table_1.9.5

loaded via a namespace (and not attached):
[1] tools_3.2.0  chron_2.3-45
```

jangorecki:
[...]

PR #6631, opened 2024-12-04
Suppress autoprint during := assignment on subclasses of data.table

### LINKED ISSUE #3029: Print method for subclass makes `:=` return visible output
I'm extending `data.table` class and I'm having a problem. When defining a custom print method operations with `:=` print the object instead of returning it invisibly. 

``` r
library(data.table)

dt <- data.table(x = 1)

dt[, y := 1]  # No output. Correct!

setattr(dt, "class", c("someclass", class(dt)))

dt[, y := 1] # still no output. Yey!

print.someclass <- function(x, ...) {
  NextMethod("print")
  # maybe other things
}

dt[, y := 1]   # there's output, Noooo!
```

(The output doesn't seem to appear when reprexed, but it's there in interactive mode, please test)

I don't know what I'm doing wrong and I cannot find much documentation on this.

MichaelChirico:
Closes #3029.

NB: This is currently based to #6589 given it touches a very similar aspect of `print.data.table()` (hence cc @aitap too).

This constitutes a small breaking change by restoring behavior of `print(DT)` _and_ `print(DT[, a:=b])` to _always_ print. I don't really see a downside of this -- the other tests in autoprint continue to WAI.

Does anyone see a reason to maintain the old behavior? @jangorecki @tdhock @ben-schwen. It should be feasible to adjust this PR to make the code a tiny bit more complicated while still achieving the desired fix for #3029 -- I just am not seeing the benefit of that (besides back-compatibility on this pretty minor aspect of the package).

Here's the relevant commit that introduced this behavior: 5d95da3a750476d27a11177569a64c0b6967b9d5 fixing #1122.

It seems like this PR achieves what that PR couldn't (possibly due to an underlying change in R itself), namely to have `if (TRUE) DT[, a:=b]` and `print(DT)` _both_ not print.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,4 +1,5 @@
 .dir-locals.el
+.check.translations.R
 ^\.Rprofile$
 ^data\.table_.*\.tar\.gz$
 ^config\.log$
--- a/.ci/.lintr.R
+++ b/.ci/.lintr.R
@@ -84,7 +84,7 @@ exclusions = c(local({
       infix_spaces_linter = Inf,
       undesirable_function_linter = Inf
     )),
-    exclusion_for_dir("vignettes", list(
+    exclusion_for_dir(c("vignettes", "vignettes/fr"), list(
       quotes_linter = Inf,
       sample_int_linter = Inf
       # strings_as_factors_linter = Inf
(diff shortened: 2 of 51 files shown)


MichaelChirico:
cc @eliocamp @nikosbosse @seabbs @sbfnk in case you want to test out the proposed fix for your use case.

ben-schwen:
Old behavior seems unwanted to me. Forcing that `print` indeed prints to output seems legit!

MichaelChirico:
SG! maybe revdeps will see something but I doubt it (as this should only really affect interactive sessions)

PR #6632, opened 2024-12-04
Explain comment in tests previously marked as OBSCURE

MichaelChirico:
h/t @rikivillalba for explaining. Managed to find a source for it in English too.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -10543,7 +10543,9 @@ test(1702.1, isoweek(test_cases), test_values)
 #   but calculating from Date brings these into play, #2407
 test(1702.2, isoweek(as.Date(test_cases)), test_values)
 
-# *** OBSCURE ERROR WHEN Sys.timezone() = 'America/Argentina/Buenos_Aires' ***
+# *** DST ERROR WHEN Sys.timezone() = 'America/Argentina/Buenos_Aires' ***
+# *** 00:00 does not exist, h/t rikivillalba@ who worked through this  ***
+# *** https://techcommunity.microsoft.com/blog/dstblog/argentina-is-changing-their-daylight-saving-time-on-december-30/311020 ***
 test(1702.3, isoweek(as.POSIXct(test_cases)), test_values)
 
 # 1% sample of a 400-year cycle of dates for extra robustness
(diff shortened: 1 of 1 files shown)


MichaelChirico:
@rikivillalba actually, I tried to reproduce the error so that we could skip the test if needed, no luck, here's what I saw:

```r
as.POSIXct(as.POSIXct(c("2007-12-30 02:00:00", "2007-12-30 03:00:00"), tz = "UTC"), tz = "America/Buenos_aires")
# [1] "2007-12-30 02:00:00 America" "2007-12-30 03:00:00 America"
Sys.setenv(TZ="America/Buenos_aires")
Sys.time()
# [1] "2024-12-04 07:50:48 America"
Sys.timezone()
# [1] "America/Buenos_aires"
test(1702.1, isoweek(test_cases), test_values)
test(1702.2, isoweek(as.Date(test_cases)), test_values)
test(1702.3, isoweek(as.POSIXct(test_cases)), test_values)

# (i.e. no error)
```

Do you still reproduce an error in America/Buenos_aires time zone?

rikivillalba:
Hello. I test [this](https://github.com/rikivillalba/reportes_api_mem/blob/120caccdcd87b3feb2e1e7a5792404676bcb2e8f/.github/workflows/America_Buenos_Aires.yml) workflow and the results were annoying:

In Ubuntu:
```
[1] "UTC"
[1] "2007-12-30 02:00:00 America" "2007-12-30 03:00:00 America"
[1] "America/Buenos_aires"
[1] "2007-12-30 02:00:00 America" "2007-12-30 03:00:00 America"
```

In Windows:
```
[1] "UTC"
[1] "2007-12-29 23:00:00 -03" "2007-12-30 01:00:00 -02"
[1] "America/Buenos_aires"
[1] "2007-12-29 23:00:00 -03" "2007-12-30 01:00:00 -02"
```
That is, the results are different between oses.

My at-work's notebook is windows, that i use the most. I can see the same outcome than in the workflow. I have another with linux in which i don't tested yet.
I also checked the tz database from [IANA](https://www.iana.org/time-zones), lines 195-197 have these rules:

?|?|?|?|?|?|?|?|?|?
----|---|----|----|-|---|--|----|----|-
Rule|Arg|2007|only|-|Dec|30|0:00|1:00|-
Rule|Arg|2008|2009|-|Mar|Sun>=15|0:00|0|-
Rule|Arg|2008|only|-|Oct|Sun>=15|0:00|1:00|-

Perhaps an R issue? I'll try test in another way wether the same outocome is reproducible. Nevertheless, I think the term OBSCURE till remains appropriate.  
[southamerica.txt](https://github.com/user-attachments/files/18012627/southamerica.txt)

MichaelChirico:
What a mess 😅 thanks for digging into the database file, the comments there are pretty fun :)

PR #6637, opened 2024-12-06
simplify format_col.default, allowing e.g. vctrs_list_of columns to print well

### LINKED ISSUE #5948: `list` sub-class with `format()` method prints full contents
I ran into this when converting a nested `tibble` (i.e. with a list column of `tibble`s) to a `data.table`. Normally, `list` columns print with something like `<tibble[3x1]>` in `data.table`; however, these columns printed as a string displaying the entirety of the nested `tibble`s contents (in my case, ~20k rows of data per `tibble`). This appears to be due to a check for the existence of a `format()` method for the column type in `data.table:::format_col()`. In this case, the list columns were `vctrs_list_of` class, which implements its own `format.vctrs_list_of()` method. See below reprex for an example.

Fixing this is easy enough (see below reprex for proposed solution), but it would change the default printing behavior of list-cols. Any list subclass would then have to implement a `format_list_item()` method to get special treatment. On the other hand, any list subclass *could* then implement that method and get special treatment (I think this is basically what `tibble` does to get special printing of `vctrs_list_of` and other custom classes).

To me, defaulting to the standard list print behavior is an improvement, given that `format()` methods for list-like classes are generally not going to product output suitable for a column in a `data.table` (which is why `format_list_item()` is needed in the first place).

Reprex is below, along with proposed solution. I'm happy to file a PR but wanted to make sure the change in defaults is acceptable in principle first. Thanks!

``` r
# Setup -----------------------------------------------------------------------

# Use development version of {data.table}
data.table::update_dev_pkg()
#> R data.table package is up-to-date at 8f8ef9343dcabefa9e4cb0af4251cbb74dae9f55 (1.15.99)
# Attach internals
ns <- asNamespace("data.table")
name <- format(ns)
attach(ns, name = name, warn.conflicts = FALSE)

# Create example data
dt <- data.table(
  list_col = list(data.table(a = 1:3), data.table(a = 4:6)),
  list_of_col = vctrs::list_of(data.table(a = 1:3), data.table(a = 4:6))
)


# Problem ---------------------------------------------------------------------
# `format_col()` does not format `list_of` columns from {vctrs} as lists

# Print `data.table`
print(dt)
#>             list_col     list_of_col
#>               <list> <vctrs_list_of>
#> 1: <data.table[3x1]>         1, 2, 3
#> 2: <data.table[3x1]>         4, 5, 6

# Format individually
format_col(dt$list_col)
#> [1] "<data.table[3x1]>" "<data.table[3x1]>"
format_col(dt$list_of_col)
#> [1] "1, 2, 3" "4, 5, 6"


# Solution --------------------------------------------------------------------

# Current function definition:
# format_col.default = function(x, ...) {
#   if (!is.null(dim(x)))
#     "<multi-column>"
#   else if (has_format_method(x) && length(formatted<-format(x, ...))==length(x))
#     formatted
#   else if (is.list(x))
#     vapply_1c(x, format_list_item, ...)
#   else
#     format(char.trunc(x), ...)
# }

# Swapping the order of the `else if` statements in `format_col()` will fix the issue
# `format_col_updated()` formats `list_of` columns from {vctrs} as lists
format_col_updated = function(x, ...) {
  if (!is.null(dim(x)))
    "<multi-column>"
  else if (is.list(x)) # Now comes before `has_format_method()` check
    vapply_1c(x, format_list_item, ...)
  else if (has_format_method(x) && length(formatted<-format(x, ...))==length(x))
    formatted
  else
    format(char.trunc(x), ...)
}

# Replace as default method
registerS3method("format_col", "default", format_col_updated)

# Print `data.table`
print(dt)
#>             list_col       list_of_col
#>               <list>   <vctrs_list_of>
#> 1: <data.table[3x1]> <data.table[3x1]>
#> 2: <data.table[3x1]> <data.table[3x1]>

# Print individually
# Still formatted as list
format_col_updated(dt$list_col)
#> [1] "<data.table[3x1]>" "<data.table[3x1]>"
# Now also formatted as list
format_col_updated(dt$list_of_col)
#> [1] "<data.table[3x1]>" "<data.table[3x1]>"

# Clean up --------------------------------------------------------------------
detach(name, character.only = TRUE)
rm(list = ls())
```

<sup>Created on 2024-02-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.2 (2023-10-31 ucrt)
#>  os       Windows 11 x64 (build 22631)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  English_United States.utf8
#>  ctype    English_United States.utf8
#>  tz       America/Chicago
#>  date     2024-02-21
#>  pandoc   3.1.1 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  cli           3.6.2   2023-12-11 [1] CRAN (R 4.3.2)
#>  data.table    1.15.99 2024-02-21 [1] local
#>  digest        0.6.33  2023-07-07 [1] CRAN (R 4.3.2)
#>  evaluate      0.23    2023-11-01 [1] CRAN (R 4.3.2)
#>  fastmap       1.1.1   2023-02-24 [1] CRAN (R 4.3.2)
#>  fs            1.6.3   2023-07-20 [1] CRAN (R 4.3.2)
#>  glue          1.7.0   2024-01-09 [1] CRAN (R 4.3.2)
#>  htmltools     0.5.7   2023-11-03 [1] CRAN (R 4.3.2)
#>  knitr         1.45    2023-10-30 [1] CRAN (R 4.3.2)
#>  lifecycle     1.0.4   2023-11-07 [1] CRAN (R 4.3.2)
#>  magrittr      2.0.3   2022-03-30 [1] CRAN (R 4.3.2)
#>  purrr         1.0.2   2023-08-10 [1] CRAN (R 4.3.2)
#>  R.cache       0.16.0  2022-07-21 [1] CRAN (R 4.3.2)
#>  R.methodsS3   1.8.2   2022-06-13 [1] CRAN (R 4.3.1)
#>  R.oo          1.25.0  2022-06-12 [1] CRAN (R 4.3.1)
#>  R.utils       2.12.3  2023-11-18 [1] CRAN (R 4.3.2)
#>  reprex        2.0.2   2022-08-17 [1] CRAN (R 4.3.2)
#>  rlang         1.1.3   2024-01-10 [1] CRAN (R 4.3.2)
#>  rmarkdown     2.25    2023-09-18 [1] CRAN (R 4.3.2)
#>  rstudioapi    0.15.0  2023-07-07 [1] CRAN (R 4.3.2)
#>  sessioninfo   1.2.2   2021-12-06 [1] CRAN (R 4.3.2)
#>  styler        1.10.2  2023-08-29 [1] CRAN (R 4.3.2)
#>  vctrs         0.6.5   2023-12-01 [1] CRAN (R 4.3.2)
#>  withr         3.0.0   2024-01-16 [1] CRAN (R 4.3.2)
#>  xfun          0.41    2023-11-01 [1] CRAN (R 4.3.2)
#>  yaml          2.3.7   2023-01-23 [1] CRAN (R 4.3.2)
#> 
#>  [1] D:/ProgramFiles/R/R-4.3.2/library
#> 
#> ──────────────────────────────────────────────────────────────────────────────
```

</details>

MichaelChirico:
Closes #5948

Turns out, this code in support of #2273 (#5224) is not needed -- I don't notice any difference before/after this change and we pass the related test:

https://github.com/Rdatatable/data.table/blob/a36caac455d4d269d13d9bcb192dd9740face676/inst/tests/other.Rraw#L220-L226

It's possible our testing is just not extensive enough.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -117,6 +117,8 @@ rowwiseDT(
 
 15. The auto-printing suppression in `knitr` documents is now done by implementing a method for `knit_print` instead of looking up the call stack, [#6589](https://github.com/Rdatatable/data.table/pull/6589). Thanks to @jangorecki for the report [#6509](https://github.com/Rdatatable/data.table/issues/6509) and @aitap for the fix.
 
+16. A data.table with a column of class `vctrs_list_of` (from package {vctrs}) prints as expected, [#5948](https://github.com/Rdatatable/data.table/issues/5948). Before, they could be printed messily, e.g. printing every entry in a nested data.frame. Thanks @jesse-smith for the report, @DavisVaughan and @r2evans for contributing, and @MichaelChirico for the PR.
+
 ## NOTES
 
 1. There is a new vignette on joins! See `vignette("datatable-joins")`. Thanks to Angel Feliz for authoring it! Feedback welcome. This vignette has been highly requested since 2017: [#2181](https://github.com/Rdatatable/data.table/issues/2181).
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -193,8 +193,6 @@ has_format_method = function(x) {
 format_col.default = function(x, ...) {
   if (!is.null(dim(x)))
     "<multi-column>"
-  else if (has_format_method(x) && length(formatted<-format(x, ...))==length(x))
-    formatted  #PR5224 motivated by package sf where column class is c("sfc_MULTIPOLYGON","sfc") and sf:::format.sfc exists
   else if (is.list(x))
     vapply_1c(x, format_list_item, ...)
   else
@@ -235,6 +233,8 @@ format_list_item.default = function(x, ...) {
 format_list_item.data.frame = function(x, ...) {
   paste0("<", class1(x), paste_dims(x), ">")
 }
+
+# format_list_item.vctrs_list_of = function(, .)
 
 # FR #1091 for pretty printing of character
 # TODO: maybe instead of doing "this is...", we could do "this ... test"?
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -234,8 +234,6 @@ format_list_item.data.frame = function(x, ...) {
   paste0("<", class1(x), paste_dims(x), ">")
 }
 
-# format_list_item.vctrs_list_of = function(, .)
-
 # FR #1091 for pretty printing of character
 # TODO: maybe instead of doing "this is...", we could do "this ... test"?
 # Current implementation may have issues when dealing with strings that have combinations of full-width and half-width characters,
--- a/inst/tests/other.Rraw
+++ b/inst/tests/other.Rraw
@@ -1,4 +1,4 @@
-pkgs = c("ggplot2", "hexbin", "plyr", "dplyr", "caret", "zoo", "xts", "gdata", "nlme", "bit64", "knitr", "parallel", "sf", "nanotime", "R.utils", "yaml")
+pkgs = c("bit64", "caret", "dplyr", "gdata", "ggplot2", "hexbin", "knitr", "nanotime", "nlme", "parallel", "plyr", "R.utils", "sf", "vctrs", "xts", "yaml", "zoo")
 # First expression of this file must be as above: .gitlab-ci.yml uses parse(,n=1L) to read one expression from this file and installs pkgs.
 # So that these dependencies of other.Rraw are maintained in a single place.
 # TEST_DATA_TABLE_WITH_OTHER_PACKAGES is off by default so this other.Rraw doesn't run on CRAN. It is run by GLCI, locally in dev, and by
@@ -773,3 +773,9 @@ if (loaded[["nanotime"]]) {
 res <- tables(env=.e)
 test(32, res[, .(NAME,NROW,NCOL,MB)], data.table(NAME="DT",NROW=20000000L,NCOL=15L,MB=2288.0))
 rm(.e, res)
+
+if (loaded[["vctrs"]]) {
+  # vctrs::list_of() columns are treated the same as other list() columns
+  DT = data.table(a = 1, b = list_of(mtcars))
+  test(33, DT, output="<vctrs_list_of>.*<data\\.frame\\[32x11\\]>")
+}


MichaelChirico:
This test fails:

```r
registerS3method("format", "foo2130", function(x, ...) rep("All hail foo",length(x)))
test(2130.15, print(DT), output="All hail foo")  # e.g. sf:::format.sfc rather than sf:::format.sfg on each item
```

My sense is that if we only fail a toy example, we should just break it. In general I am thinking the better solution here is to add `format_col` and/or `format_list_item` methods as needed.

I think given the potential for breaking change, it's best to save this PR for 1.18.0.

aitap (approved):
Let's compare with other `data.frame`s.

 * `tibble` refuses classed lists that don't come from their own packages:

```
> tibble(a = structure(1, class = 'foo2130'))
# A tibble: 1 × 1
  a
  <foo2130>
1 All hail foo
> tibble(a = structure(list(1), class = 'foo2130'))
Error in `tibble()`:
! All columns in a tibble must be vectors.
✖ Column `a` is a `foo2130` object.
Run `rlang::last_trace()` to see where the error occurred.
> tibble(a = list(structure(1, class = 'foo2130')))
# A tibble: 1 × 1
  a
  <list>
1 <foo2130 [1]>
> tibble(a = 1, b = list(mtcars))
# A tibble: 1 × 2
      a b
  <dbl> <list>
1     1 <df [32 × 11]>
> tibble(a = 1, b = list_of(mtcars))
# A tibble: 1 × 2
      a               b
  <dbl> <list<df[,11]>>
1     1       [32 × 11]
```

* `data.frame` is consistent in saying that `format(<data.frame>)` is whatever `format(...)` returns for the individual columns, but doesn't support list columns that well. By default list elements are converted to `data.frame` columns. Constructing one with a list verbatim requires `I()`, which overrides `format`. A list inserted manually into an existing `data.frame` will be formatted using the usual `format` method:

```
> data.frame(a = 1, b = list_of(mtcars)) # NB: uses as.data.frame.vctrs_vctr to make the list into a column
# omitted: similar to current data.table behaviour, i.e., formats the whole `mtcars`
> x <- data.frame(a = 1)
> x$b <- list(mtcars)
> x
# same
> data.frame(a = 1, b = I(list(mtcars))) # stores list as is but with format() overriden
  a            b
1 1 c(21, 21....
> data.frame(a = structure(list(1), class = 'foo2130')) # tries to convert a list to data.frame
Error in as.data.frame.default(x[[i]], optional = TRUE, stringsAsFactors = stringsAsFactors) :
  cannot coerce class ‘"foo2130"’ to a data.frame
> traceback()
4: stop(gettextf("cannot coerce class %s to a data.frame", sQuote(deparse(class(x))[1L])),
       domain = NA)
3: as.data.frame.default(x[[i]], optional = TRUE, stringsAsFactors = stringsAsFactors)
2: as.data.frame(x[[i]], optional = TRUE, stringsAsFactors = stringsAsFactors)
1: data.frame(a = structure(list(1), class = "foo2130"))
> x <- data.frame(a = 1)
> x$b <- structure(list(1), class = 'foo2130') # uses format()
> x
  a            b
1 1 All hail foo
> data.frame(a = I(structure(list(1), class = 'foo2130'))) # uses format.AsIs()
  a
1 1
```

* `DataFrame` allows classed lists but expands unclassed lists or `data.frame`s into columns. Seems to ignore `format` methods altogether:

```
> DataFrame(a = structure(list(1,10), class = 'foo2130'))
DataFrame with 2 rows and 1 column
          a
  <foo2130>
1         1
2        10
> DataFrame(a = list(1,2))
DataFrame with 1 row and 2 columns
       a.X1      a.X2
  <numeric> <numeric>
1         1         2
> DataFrame(a = list(mtcars))
DataFrame with 32 rows and 11 columns
                      a.mpg     a.cyl    a.disp      a.hp    a.drat      a.wt    a.qsec      a.vs      a.am    a.gear
                  <numeric> <numeric> <numeric> <numeric> <numeric> <numeric> <numeric> <numeric> <numeric> <numeric>
Mazda RX4              21.0         6       160       110      3.90     2.620     16.46         0         1         4
Mazda RX4 Wag          21.0         6       160       110      3.90     2.875     17.02         0         1         4
# omitted
> DataFrame(a = structure(1, class = 'foo2130'))
DataFrame with 1 row and 1 column
          a
  <foo2130>
1         1
```

* `data.table` formats lists specially by default, but defers to a format method if it exists:

```
> data.table(a = 1, b = list(mtcars))
       a                   b
   <num>              <list>
1:     1 <data.frame[32x11]>
> data.table(a = 1, b = list_of(mtcars))
# omitted: formats whole `mtcars` into a single string
```

Since we allow list columns of any class without applying `as.data.(frame|table)` first, we might as well format all lists, even classed ones, in a special compact form, despite a few methods, such as `sf:::format.sfc`, exist to do the right thing. And the [`sf` classes still do print specially](https://github.com/Rdatatable/data.table/pull/5224) because `data.table:::format_list_item.default` still checks for a `format` method and eventually calls `sf:::format.sfg`.

We could also register `format_col.sfc` and set `Enhances: sf`. No need for conditional registration here since we own the generic.

MichaelChirico:
Thanks Ivan! For future reference, I think you omitted your `format.foo2130` method. Also I'm not sure which package's `DataFrame` that is, is it maybe [{S4Vectors}](https://github.com/Bioconductor/S4Vectors/blob/4564bfa6eecd3c2538f01adff6a11d79c63cf8a0/NAMESPACE#L375)?

aitap:
Correct, that's the [S4 class from Bioconductor](https://bioconductor.org/packages/release/bioc/vignettes/S4Vectors/inst/doc/S4VectorsOverview.html#list-like-objects). I must have used `registerS3method("format", "foo2130", function(x, ...) rep("All hail foo",length(x)))` from [the previous comment](https://github.com/Rdatatable/data.table/pull/6637#issuecomment-2522012204), but I wouldn't have to remember it if I had copied it in my message.

PR #6640, opened 2024-12-07
Postpone the Non-API Problem.

SebKrantz:
In case you want to postpone your solution to #6180. This is all I can offer for now. Runs perfectly without error or notes on all plattforms and would immediately get CRAN off your back and allow you (us) to think about how to properly solve the non-API issues (e.g., full blown ALTREP overhaul a la @aitap + new character match or something else). Seems to me the proper solution is not going to be here by tomorrow. If CRAN is not pushing you then of course you can ignore this. They were not responding anymore to {collapse} submissions with non-API calls.

### DIFF
--- a/src/R_Defn.h
+++ b/src/R_Defn.h
@@ -1,0 +1,86 @@
+#include <R.h>
+#include <Rinternals.h>
+
+// NOTE: All of this is copied from Defn.h: https://github.com/wch/r-source/blob/28de75af0541f93832c5899139b969d290bf422e/src/include/Defn.h
+// We intend to gradually remove the need for this header file
+
+// Writable vector/string pointer
+#define SEXPPTR(x) ((SEXP *)DATAPTR(x))
+
+#ifndef NAMED_BITS
+# define NAMED_BITS 16
+#endif
+
+#ifndef SEXPREC_HEADER
+
+// https://github.com/wch/r-source/blob/2640a203d13473f95c9c7508eb2976fefb5c931c/src/include/Defn.h#L123
+struct sxpinfo_struct {
+  SEXPTYPE type      :  TYPE_BITS;
+  /* ==> (FUNSXP == 99) %% 2^5 == 3 == CLOSXP
+   * -> warning: `type' is narrower than values
+   *              of its type
+   * when SEXPTYPE was an enum */
+  unsigned int scalar:  1;
+  unsigned int obj   :  1;
+  unsigned int alt   :  1;
+  unsigned int gp    : 16;
+  unsigned int mark  :  1;
+  unsigned int debug :  1;
+  unsigned int trace :  1;  /* functions and memory tracing */
+  unsigned int spare :  1;  /* used on closures and when REFCNT is defined */
+  unsigned int gcgen :  1;  /* old generation number */
+  unsigned int gccls :  3;  /* node class */
+  unsigned int named : NAMED_BITS;
+  unsigned int extra : 32 - NAMED_BITS; /* used for immediate bindings */
+}; /*		    Tot: 64 */
+  
+// https://github.com/wch/r-source/blob/2640a203d13473f95c9c7508eb2976fefb5c931c/src/include/Defn.h#L190  
+#define SEXPREC_HEADER           \
+  struct sxpinfo_struct sxpinfo; \
+  struct SEXPREC *attrib;        \
+  struct SEXPREC *gengc_next_node, *gengc_prev_node
+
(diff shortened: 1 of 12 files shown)


TysonStanley:
Thanks! They mentioned on the latest patch submission that was accepted that we need this fixed so this is great timing.

SebKrantz:
Should have only minor performance implications. One thing I could add to compensate/make `data.table` faster is to replace `TRULEN` (`#define TRULEN(x) (ALTREP(x) ? 0 : STDVEC_TRUELENGTH(x))`) with `STDVEC_TRUELENGTH` in `forder.c` and `chmatch.c`. You are only using it for character strings here which are not ALTREP. I have done this in {collapse}'s version of forder for like 3 years and never encountered an issue. More fundamentally, your algorithm wouldn't work anymore with ALTREP strings. So if you like I'd make that replacement in `forder.c` and `chmatch.c` which should boost performance.

SebKrantz:
Maybe if you want to be very correct you need to adjust the license, see, e.g. the [{collapse} LICENSE](https://github.com/SebKrantz/collapse/blob/master/LICENSE). But I'd not make a big deal about it. If we are doing it right, this should just pass through the autochecks without the need for human engagement. Only need to get rid of some NOTES still: 

Found the following (possibly) invalid URLs:
  URL: https://cran.r-project.org/web/checks/check_results_data.table.html
    From: man/openmp-utils.Rd
          README.md
    Status: Error
    Message: Connection died, tried 5 times before giving up
  URL: https://twitter.com/hashtag/rdatatable
    From: README.md
    Status: 403
    Message: Forbidden
  URL: https://x.com/r_data_table
    From: README.md
    Status: 403
    Message: Forbidden

Found the following (possibly) invalid file URI:
  URI: Seal_of_Approval.md
    From: README.md

tdhock:
> Should have only minor performance implications.  

do you have any idea where those minor performance implications should be? It is difficult for me to tell, because there are so many files changed. It would be great to add a performance test.

@Rdatatable/performance-testers do you know why the atime CI job failed here?

ben-schwen:
@SebKrantz thank you very much for the effort you put into this! superseded by #7451 and #6694
If you need a 2nd or 3rd pair of eyes for transitioning collapse away from the Defn.h header hack feel free to tag us!

SebKrantz:
Thanks @ben-schwen! And congratulations on the work! I think the more immediate need for *collapse* in this regard is to transition the data.table creation code: https://github.com/fastverse/collapse/blob/master/src/data.table_subset.c. This may cause issues on CRAN once you push your update. I wanted to look into that, but haven't found time yet. Some additional eyes on it would certainly be appreciated.

aitap:
Internally, the resizable vectors API uses the same SEXP fields as
before: GROWABLE_BIT set, allocation length stored in TRUELENGTH. Since
'collapse' already requires R above 3.4, the main thing is to mark the
data.tables with the GROWABLE_BIT (LEVELS(x) & 0x20) so that
R_resizeVector doesn't refuse to change the number of columns. Once
they are growable, the finalizer can be removed. Its purpose was to
compensate for the GC mis-counting allocated memory when collecting
over-allocated vectors without the GROWABLE_BIT set.

(This way you'll also avoid the situation where
unloadNamespace('collapse') with data.tables created by your code still
in memory dooms your process to crash as soon as one of them is
garbage-collected.)

SebKrantz:
Thanks @aitap. Will get back to this once I find time.

SebKrantz:
Are there any backwards comparibility challenges @aitap? I presume removing the finalizer would break things. Also: when are you planning to push this to CRAN?

aitap:
The finalizer really is mostly cosmetic. [It was needed on R < 3.4](https://rdatatable-community.github.io/The-Raft/posts/2025-01-13-non-api-use/index.html#:~:text=When%20deallocating%20shortened%20objects%20without%20the%20GROWABLE_BIT%20set) where no `GROWABLE_BIT` existed and the garbage collector forgot to count the memory taken by the invisible parts of the over-allocated vectors. No code checks whether the finalizer is set on a given table and it's bound individually to every `.internal.selfref` attribute, so even with `data.table` < 1.18.0 loaded things should just work.

Having the growable bit set is more important. `R_resizeVector` will refuse to resize vectors that don't have enough of a `TRUELENGTH` or the `GROWABLE_BIT` set; the latter is the new requirement. With `data.table` 1.18.0 loaded, adding or removing (but not changing) columns by reference will stop working until `setDT` recreates the column pointer list.

Tyson is submitting the package now. I'm sorry I didn't prepare anything for you in advance; I spent most of the year writing my thesis and then had to fix everything in firefighting mode during the past couple of weeks. I hope fastverse/collapse#809 will be of some help.

PR #6647, opened 2024-12-10
Remove traces of abandoned plan to change logical01 default

MichaelChirico:
Should have been part of #5883.


MichaelChirico:
Comment-only change, merging

PR #6650, opened 2024-12-10
Remove obsolete deprecation comment in as.data.table.matrix

MichaelChirico:
The lines just below show this is already an error. Checking the CRAN blame, it's been an error since v1.15.0


MichaelChirico:
Comment-only change. Merging.

PR #6652, opened 2024-12-10
Progress deprecation of in.place argument to droplevels.data.table

MichaelChirico:


R/fdroplevels.R
```diff
@@ -8,10 +8,9 @@ fdroplevels = function(x, exclude = if (anyNA(levels(x))) NULL else NA, ...) {
   return(ans)
 }
 
-droplevels.data.table = function(x, except=NULL, exclude, in.place=FALSE, ...){
-  stopifnot(is.logical(in.place))
-  if (isTRUE(in.place)) warningf("droplevels() with in.place=TRUE is deprecated. Use setdroplevels() instead.")
-  if (!in.place) x = copy(x)
+droplevels.data.table = function(x, except=NULL, exclude, in.place=NULL, ...){
```
ben-schwen, line 11: I guess we can even remove the `in.place=NULL` from the signature

MichaelChirico: I think realistically we can just completely deprecate the argument already -- literally 0 usage on GitHub:

https://github.com/search?q=lang%3AR+%2Fdroplevels%2F+%2Fin%5B.%5Dplace%2F+-path%3Afdroplevels&type=code

But in keeping with strict adherence to approaching deprecation slowly+carefully, I think we should keep it in the signature for now -- it'll give a friendlier error for the imagined user whose code calling `droplevels(in.place=TRUE)` starts to break.

PR #6654, opened 2024-12-10
Version-guard STRING_PTR definition

MichaelChirico:
Part of #6180. I'm not quite sure why the `#ifndef` approach failed us, but I confirmed this works, as well as confirming R 3.5.0 is the first version where `STRING_PTR_RO()` is in a release:

https://github.com/wch/r-source/blob/tags/R-3-4-4/src/include/Rinternals.h
https://github.com/wch/r-source/blob/tags/R-3-5-0/src/include/Rinternals.h

### DIFF
--- a/src/data.table.h
+++ b/src/data.table.h
@@ -6,6 +6,7 @@
 #  define ALTREP(x) 0     // #2866
 #  define USE_RINTERNALS  // #3301
 #  define DATAPTR_RO(x) ((const void *)DATAPTR(x))
+#  define STRING_PTR_RO STRING_PTR
 #  define R_Calloc(x, y) Calloc(x, y)         // #6380
 #  define R_Realloc(x, y, z) Realloc(x, y, z)
 #  define R_Free(x) Free(x)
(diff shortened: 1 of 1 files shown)


aitap:
Good catch! `STRING_PTR_RO` used to be [available as a macro](https://github.com/wch/r-source/blob/88de3c2999d72e257670deb5216acf3a426586b9/src/include/Rinternals.h#L424) with [`USE_RINTERNALS` defined](https://github.com/wch/r-source/blob/88de3c2999d72e257670deb5216acf3a426586b9/src/include/Rinternals.h#L187), otherwise as the [exported](https://github.com/wch/r-source/blob/88de3c2999d72e257670deb5216acf3a426586b9/src/include/Rinternals.h#L658) function-[wrapper](https://github.com/wch/r-source/blob/88de3c2999d72e257670deb5216acf3a426586b9/src/main/memory.c#L3656-L3662) with slightly better safety. Most of `USE_RINTERNALS` [disappeared in R-4.2.0](https://github.com/wch/r-source/commit/86b4a0641e889808ce204e28dd8bd19bbd1c8cbb).

MichaelChirico:
Ah, well that makes sense! #ifndef doesn't work because simply it's not a macro! 🤪

PR #4458, opened 2020-05-18
order optimizes to forderv, uses key or index, forder more consistent to base order

jangorecki:
- [x] `order` could utilize existing index #3023 (once #4386 merged)

using main.Rraw as of current master (not counting tests in this PR), optimization DT[order->forderv] from this PR kicks in 46 times

- [x] `DT[order(.), ...]` could create index on DT #3261 (once #4386 merged, and `setindex` from C forder enabled)

- [x] order optimization does recognize all order args properly #4456
- [ ] forder should not accept list, but forderv should #4346

PR #6655, opened 2024-12-11
better match base::order(method=) and decreasing=

### LINKED ISSUE #4456: order optimization does not recognize all order args properly
current devel

- rather uncommon case, a non missing `method` argument
```r
data.table(x=2:1)[base::order(x, method="auto")]
#   x
#1: 1
#2: 2
data.table(x=2:1)[order(x, method="auto")]
#Error: Column 2 is length 1 which differs from length of column 1 (2)
```

- length 1+ `decreasing` argument

```r
data.table(x=2:1, y=2L)[order(x, y, decreasing=c(FALSE,FALSE))]
#Error: isTRUEorFALSE(decreasing) is not TRUE
```

```r
d = data.table(x=3:1, y=c(2L,1L,3L), z=1:3)
with(d, order(x, y, decreasing=c(FALSE,FALSE)))
#[1] 3 2 1
d[with(d, order(x, y, decreasing=c(FALSE,FALSE)))] ## order is masked with forder inside outer `d` thus has to be compatible
#Error: isTRUEorFALSE(decreasing) is not TRUE
d[with(d, base::order(x, y, decreasing=c(FALSE,FALSE)))]
#   x y z
#1: 1 3 3
#2: 2 1 2
#3: 3 2 1
```

MichaelChirico:
Closes #4456


MichaelChirico:
Somehow I'd missed #4458. This PR is narrower, so I suggest merging it independently.

tdhock:
lgtm

Issue #5981, opened 2024-03-05
[BUG] Cannot print a data.table with column of type `rvar` from posterior package


SteveBronder:
# Summary

When printing a `data.table` that contains an [`rvar`](https://mc-stan.org/posterior/articles/rvar.html) type from the `posterior` package an empty `data.table` is printed instead. See the mrp below for a full case. Looking into it more it appears `dim.data.table` returns a value of zero for the rows where there should be a nonzero value. `dim` returning a zero causes [line 55](https://github.com/Rdatatable/data.table/blob/46ee52571214b135b645e367ebacd02de02aff52/R/print.data.table.R#L55) of `print.data.table` to go to the empty `data.table` message.

I tried looking into the code for [`dim`](https://github.com/Rdatatable/data.table/blob/46ee52571214b135b645e367ebacd02de02aff52/src/wrappers.c#L101) (pasted below). The `data.table` has a `length(x)` equal to 1 so the last branch is chosen.

```
SEXP dim(SEXP x)
{
  // fast implementation of dim.data.table

  if (TYPEOF(x) != VECSXP) {
    error(_("dim.data.table expects a data.table as input (which is a list), but seems to be of type %s"),
        type2char(TYPEOF(x)));
  }

  SEXP ans = PROTECT(allocVector(INTSXP, 2));
  if(length(x) == 0) {
    INTEGER(ans)[0] = 0;
    INTEGER(ans)[1] = 0;
  } else {
// This branch chosen
    INTEGER(ans)[0] = length(VECTOR_ELT(x, 0));
    INTEGER(ans)[1] = length(x);
  }
  UNPROTECT(1);
  return ans;
}
```

Calling `dim.data.frame` with the data.table gives the correct output below of (4, 1). I think this is because `.row_names_info` used in `dim.data.frame` is just asking for the length of the row names via `.Internal(shortRowNames(x, type))`.

## Minimal reproducible example

```r
library(data.table)
library(posterior)
n <- 4   # length of output vector
x_rvar <- rvar(array(rnorm(n*n, mean = 1, sd = 1), dim = c(n, n)))
x_rvar
# rvar<4000>[4] mean ± sd:
#   [1] 0.98 ± 1.00  1.00 ± 1.02  1.00 ± 0.99  0.99 ± 0.99 

# does not work :(
ex_dt = data.table(ex = x_rvar)
ex_dt
# Empty data.table (0 rows and 1 cols): ex

ex_df = data.frame(ex = x_rvar)

[...]

PR #6658, opened 2024-12-12
More consistent handling of POSIXlt input to setDT

### LINKED ISSUE #4800: Print a warning/... when creating data.table with POSIXlt

MichaelChirico:
Closes #4800.

Still needed:

 - Figure out what to do about the inconsistency of `setDT()` for list vs. data.frame input with matrices: https://github.com/Rdatatable/data.table/issues/4800#issuecomment-2537560759
 - Split out the `length_with_dispatch()` part which is a vestige of an earlier attempt to get `setDT()` working on POSIXlt input, but which also happens to be useful (and I think solves?) #5981
 - NEWS, tests


MichaelChirico:
This is easiest to start from scratch at current `master` given ample recent changes here.

PR #6659, opened 2024-12-12
Fix issues mixing 'const'-ness of STRING_PTR_RO

### LINKED ISSUE #6656: New compiler warnings about modifying 'const' object (STRING_PTR_RO)
https://github.com/Rdatatable/data.table/blob/a5995579d8512bfa51e35c056a992406f074d4e6/src/coalesce.c#L159

https://github.com/Rdatatable/data.table/blob/a5995579d8512bfa51e35c056a992406f074d4e6/src/chmatch.c#L43

The compiler (on GitHub codespaces) warns at these two places:

```
chmatch.c: In function ‘chmatchMain’:
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include    -fpic  -fopenmp -std=c11 -O3 -pipe -Wall -pedantic -Wstrict-prototypes -isystem /usr/share/R/include  -fno-common -c subset.c -o subset.o
coalesce.c: In function ‘coalesce’:
chmatch.c:43:8: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
   43 |     xd = STRING_PTR_RO(PROTECT(coerceUtf8IfNeeded(x))); nprotect++;
      |        ^
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include    -fpic  -fopenmp -std=c11 -O3 -pipe -Wall -pedantic -Wstrict-prototypes -isystem /usr/share/R/include  -fno-common -c transpose.c -o transpose.o
coalesce.c:159:17: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  159 |       valP[k++] = STRING_PTR_RO(item);
      |                 ^
```

The chmatch.c one is easy to fix (by instantiating `xd` as `const`), I'm less clear on how to fix the coalesce issue.

MichaelChirico:
Closes #6656

Thanks for looking into this @aitap! I _think_ the PR here is the best option.

 - for chmatch.c, `xd` is not `const` because of this:
https://github.com/Rdatatable/data.table/blob/a5995579d8512bfa51e35c056a992406f074d4e6/src/chmatch.c#L117
 - for coalesce.c, I think just switching to the `_RO` accessors for the other `SEXPTYPE`s is correct -- `valP` is only ever read from (and its value coerced before writing to `val`).

### DIFF
--- a/src/chmatch.c
+++ b/src/chmatch.c
@@ -35,12 +35,13 @@ static SEXP chmatchMain(SEXP x, SEXP table, int nomatch, bool chin, bool chmatch
     return ans;
   }
   // Since non-ASCII strings may be marked with different encodings, it only make sense to compare
-  // the bytes under a same encoding (UTF-8) #3844 #3850
+  // the bytes under a same encoding (UTF-8) #3844 #3850.
+  // Not 'const' because we might SET_TRUELENGTH() below.
   SEXP *xd;
   if (isSymbol(x)) {
     xd = &sym;
   } else {
-    xd = STRING_PTR_RO(PROTECT(coerceUtf8IfNeeded(x))); nprotect++;
+    xd = (SEXP *)STRING_PTR_RO(PROTECT(coerceUtf8IfNeeded(x))); nprotect++;
   }
   const SEXP *td = STRING_PTR_RO(PROTECT(coerceUtf8IfNeeded(table))); nprotect++;
   if (xlen==1) {
--- a/src/coalesce.c
+++ b/src/coalesce.c
@@ -53,7 +53,7 @@ SEXP coalesce(SEXP x, SEXP inplaceArg) {
     first = PROTECT(copyAsPlain(first)); nprotect++;
     if (verbose) Rprintf(_("coalesce copied first item (inplace=FALSE)\n"));
   }
-  void **valP = (void **)R_alloc(nval, sizeof(void *));
+  const void **valP = (const void **)R_alloc(nval, sizeof(void *));
   switch(TYPEOF(first)) {
   case LGLSXP:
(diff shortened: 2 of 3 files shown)


aitap:
Counter-intuitively, a `SEXP` being `const` doesn't preclude modification of the underlying value. This is because `SEXP` is an alias for a pointer type: `const SEXP * xd` is equivalent to `struct SEXPREC * const * xd`, so while the pointer `xd[i]` is `const`, the contents, `*(xd[i])`, are not.

We're not allowed to `*(xd[i])` because we're no longer told what `struct SEXPREC` is, but `SET_TRUELENGTH` getting a copy of a `const SEXP` pointer as the function argument and modifying the pointed-to buffer doesn't break any rules.

MichaelChirico:
Thanks for the explanation (which I'm still processing)!

I had noticed that the compiler doesn't care if we used `const SEXP *xd`, however I decided to take the current approach based on the "spirit" of `const`, namely it's telling the reader (as well as the compiler) 'this is "read-only"' (even though in practice the actual thing that is read-only is somewhat complex).

IMO understanding why `const` is _technically_ allowed here requires too much reasoning about internal/implementation details of `SEXP`, and I am not seeing much downside to using `(SEXP *)` instead of making `xd` `const`.

aitap:
That's reasonable. Besides, the memory is writeable anyway, so the cast will not introduce undefined behaviour.

A better explanation may be that the garbage collector cares about the pointed-to object remaining at the same address but not about its contents changing (we can't guarantee the latter at the C level).

aitap (approved):

PR #6660, opened 2024-12-12
data.frame methods for melt(), dcast()

MichaelChirico:
`dcast()` is proving a bit tricker to get working, I think owing to several NSE facets in its API.

TODO:

 - Get `dcast()` WAI
 - Add further tests to ensure the NSE aspects of `melt()` are also respected in the new method
 - NEWS


MichaelChirico:
Hm, I've forgotten the context here.

PR #6666, opened 2024-12-19
missing newline after verbose message

MichaelChirico:
Observed under `verbose=TRUE`:

```
Joining but 'x' has no key, natural join using: [V1, size]i.V1 has same type (character) as x.V1. No coercion needed.
i.size has same type (integer) as x.size. No coercion needed.
```


MichaelChirico:
Going ahead to merge the trivial change

PR #6678, opened 2024-12-20
Add example of cols=<list> to ?setindexv

### LINKED ISSUE #6665: ?setindexv lacks example of "cols=<list>" behavior
In `?setindexv` we see:

https://github.com/Rdatatable/data.table/blob/a9cd0bf6371fedba780853d630dce2a39aae5791/man/setkey.Rd#L58

I would like an example of how exactly that should be specified, but there are none in `Examples`.

venom1204:
Closes #6665 

In this i have added the example as requested 
@MichaelChirico can you please verify it ,and merge it if its correct ,
thank you.

### DIFF
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -116,6 +116,16 @@ reference.
 \examples{
 # Type 'example(setkey)' to run these at the prompt and browse output
 
+# Example: Single character vector of column names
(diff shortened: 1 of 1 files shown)


MichaelChirico (approved):
Thanks!!

PR #6694, opened 2024-12-26
Start replacing `TRUELENGTH` markers with a hash

aitap:
With apologies to Matt Dowle, who had poured a lot of effort into making `data.table` go fast.

Ongoing work towards #6180. Unfortunately, doesn't completely remove any uses of non-API entry points by itself. Detailed motivation [here](https://github.com/rdatatable-community/The-Raft/blob/56ccfaad8251839d4ed17136274f67abe4c3924a/posts/2024-12-02-non-api-use/index.qmd#L348-L407) in a pending blog post. Can't start implementing [stretchy ALTREP vectors](https://github.com/rdatatable-community/The-Raft/blob/56ccfaad8251839d4ed17136274f67abe4c3924a/posts/2024-12-02-non-api-use/index.qmd#L307-L344) until `data.table` stops using `TRUELENGTH` to mark them.

Currently implemented:

- An open-addressing hash table with Knuth's linear multiplication hashing
  - Deliberately no removal operation, only insertion and update
  - Can be made more performant (collisions much less likely) by switching to tables of power-of-two length and double hashing
- Almost all uses of `TRUELENGTH` to mark `CHARSXP`s or columns replaced with a hash

Needs more work:
- [ ] one use in `rbindlist()` pre-allocates a lot of unneeded memory
  - will probably require a dynamically growing hash table
- [ ] `forder.c`, the last remaining user of `savetl`
  - `SET_TRUELENGTH` is atomic, `hash_set` is not, will need additional care in multi-threaded environment
- [ ] `savetl` machinery in `assign.c`

Let's just see how much worse is the performance going to get.


tdhock:
hi, thanks for this. Can you please propose one or two performance test cases that you think may be adversely affected by these changes? Is it when we create a table with one column, and then use `:=` or `set` to add a lot more columns?

aitap:
The `rbindlist()` code pre-allocates `O(sum(lengths(lapply(l, names)))` memory. In the best case, only `O(length(names(l[[1]])))` is needed, `length(l)` times less. In the worst case, all names are different and the code needs all that memory for `fill = TRUE`.

The `forder()` code pre-allocates `O(nrow(x))` memory.

I'll try giving many `data.table`s to `rbindlist()` and long `data.table`s with string columns to `setorder()`.

SebKrantz:
Perhaps you can also try a fast 3rd party hash map: https://martin.ankerl.com/2019/04/01/hashmap-benchmarks-01-overview/

I particular Google's Abseil hash is pretty fast: https://abseil.io/docs/cpp/guides/container https://abseil.io/docs/cpp/guides/hash

aitap:
It's pretty bad. For typical cases, the current hash table eats >1 order of magnitude more memory, and it's similarly slower in `forder()`:
forder, typical case
rbindlist, typical case

<details>

The hash table is only on par by time in the worst case for `forder()` (all strings different) and might be on par in the worst case for `rbindlist(fill = TRUE)` (all column names different), but the latter is completely unrealistic as a use case.

forder, worst case
rbindlist, worst case

```r
```

tdhock:
thanks for proposing the performance test cases and sharing the atime benchmark graphs. I agree that we should try to avoid an order of magnitude constant factor increase in time/memory usage.

aitap:
Since profiling has shown that a noticeable amount of time is wasted initialising the giant pre-allocated hash tables, I was able to make the slowdown factor closer to 2 by dynamically re-allocating the hash table:
dynamic_hash

The memory use is significantly reduced (except for the worst cases), but cannot be measured with `atime` (just like other uses of `calloc`/`malloc`/`realloc` in `data.table` are invisible to `Rprofmem`).

<details>

```r
library(atime)

pkg.path <- '.'
limit <- 1
# taken from .ci/atime/tests.R
pkg.edit.fun <- function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}

versions <- c(
  master   = '70c64ac08c6becae5847cd59ab1efcb4c46437ac',
  static_hash = '24e81785669e70caac31501bf4424ba14dbc90f9',
  dynamic_hash = 'd7a9a1707ec94ec4f2bd86a5dfb5609207029ba4'
)

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
forderv1_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
forderv1 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(forderv1_work); gc(full = TRUE)

# worst case: all strings different
# (a challenge for the allocator too due to many small immovable objects)
N <- 10^seq(2, 7.5, .25)
forderv2_work <- lapply(setNames(nm = N), \(N)
  format(runif(N), digits = 16)
)
forderv2 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv2_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(forderv2_work); gc(full = TRUE)

# expected case: all columns named the same
N <- 10^seq(1, 5.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist1_work <- lapply(setNames(nm = N), \(N)
  rep(list(setNames(as.list(1:k), letters[1:k])), N)
)
rbindlist1 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(rbindlist1_work); gc(full = TRUE)

# worst case: all columns different
N <- 10^seq(1, 4.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist2_work <- lapply(setNames(nm = N), \(N)
  replicate(N, setNames(as.list(1:k), format(runif(k), digits = 16)), FALSE)
)
rbindlist2 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist2_work[[as.character(N)]], fill = TRUE),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(rbindlist2_work); gc(full = TRUE)

#save(forderv1, forderv2, rbindlist1, rbindlist2, file = 'times.rda')
```

</details>

The main problem with the current approach is that since the parallel loop in `src/forder.c:range_str(...)` needs to read the hash table from outside the critical section while hash elements are being set inside the critical section, it's possible that `dhash_set()` will need to reallocate the table while `dhash_lookup()` is using it. I have already tried using [pthread R/W locks](https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_rwlock_unlock.html) and OpenMP critical sections; they were too slow. I wasn't able to use OpenMP atomics to make atomic reads/writes through a pointer (maybe that's asking for too much). Any third-party hash table will also need to be outfitted with appropriate locking.

The current code keeps one previous hash table until the next reallocation cycle and hopes that `dhash_enlarge` writes the new table pointer into the object _roughly_ at the same time as the metadata, so that when `dhash_lookup` reads the object, it will _hopefully_ receive a consistent view of the object. This is not good enough for production because (1) the race window still exists and (2) there's no proof that there _won't_ be a stale reader still using the previous table array when the current writer deallocates it. What can be done with it:

 - Don't deallocate the previous hash tables until the function is completely done with them. This will double the memory use (because the code doubles the buffer when re-allocating it and $\sum_{i=0}^{n} 2^i = 2^{n+1} - 1$). We'll be able to switch back from `malloc` to `R_alloc` and make the allocations visible to R's allocator.
 - Devise a [read-copy-update](https://en.wikipedia.org/wiki/Read-copy-update#Analogy_with_reader%E2%80%93writer_locking) scheme (see also: [1](https://github.com/nicowilliams/ctp/blob/master/thread_safe_global.c) [2](http://liburcu.org/)) to properly expire the previous versions of the hash table. These may require synchronisation primitives not easily available through OpenMP. Reference-counting the hash tables may result in contention and slowdowns similar to pthread R/W locks.

SebKrantz:
In case it helps, {collapse}'s hash functions (https://github.com/SebKrantz/collapse/blob/master/src/kit_dup.c and https://github.com/SebKrantz/collapse/blob/master/src/match.c) are pretty fast as well - inspired by base R -> multiplication hash using unsigned int. It's bloody fast but requires a large table. But Calloc() is quite efficient. Anyway, would be great if you'd test the Google Hash function, curious to see it it can do much better.

aitap:
The abseil hash function is very slightly slower in my tests, although the difference may be not significant. Perhaps that's because [my C port](https://github.com/Rdatatable/data.table/blob/159e1d48926b72af9f212b8c645a8bc8ab6b20be/src/hash.c#L5-L62) fails to inline some of the things that naturally inline in [the original C++ with templates](https://github.com/abseil/abseil-cpp/blob/65b7da13bf05228cfffc0db7c72f8343866b9c57/absl/hash/internal/hash.h#L1066-L1077). I can try harder, but that's a lot of extra code to bring in properly.

forder() performance on a load with a few unique strings

forder() performance on a pathological load with all strings unique

<details>

```r
library(atime)

pkg.path <- '.'
limit <- 1
# taken from .ci/atime/tests.R
pkg.edit.fun <- function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}

versions <- c(
  master = '70c64ac08c6becae5847cd59ab1efcb4c46437ac',
  'Knuth_hash' = 'd7a9a1707ec94ec4f2bd86a5dfb5609207029ba4',
  'abseil_hash' = '159e1d48926b72af9f212b8c645a8bc8ab6b20be'
)

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
forderv1_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
forderv1 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(forderv1_work); gc(full = TRUE)

# worst case: all strings different
# (a challenge for the allocator too due to many small immovable objects)
N <- 10^seq(2, 7.5, .25)
forderv2_work <- lapply(setNames(nm = N), \(N)
  format(runif(N), digits = 16)
)
forderv2 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv2_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(forderv2_work); gc(full = TRUE)

# expected case: all columns named the same
N <- 10^seq(1, 5.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist1_work <- lapply(setNames(nm = N), \(N)
  rep(list(setNames(as.list(1:k), letters[1:k])), N)
)
rbindlist1 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(rbindlist1_work); gc(full = TRUE)

# worst case: all columns different
N <- 10^seq(1, 4.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist2_work <- lapply(setNames(nm = N), \(N)
  replicate(N, setNames(as.list(1:k), format(runif(k), digits = 16)), FALSE)
)
rbindlist2 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist2_work[[as.character(N)]], fill = TRUE),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
rm(rbindlist2_work); gc(full = TRUE)

save(forderv1, forderv2, rbindlist1, rbindlist2, file = 'times.rda')
```

</details>

`collapse::fmatch()` works very well, faster than `data.table::chmatch()` even in its current form:

chmatch() vs. fmatch() on a best case load

<details>

```r
library(atime)
library(data.table)

limit <- 1

# assumes that atime_versions() had pre-installed the packages
# master = '70c64ac08c6becae5847cd59ab1efcb4c46437ac',
# 'Knuth_hash' = 'd7a9a1707ec94ec4f2bd86a5dfb5609207029ba4',

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chmatch1 <- atime(
  N,
  seconds.limit = limit, verbose = TRUE,
  master = data.table.70c64ac08c6becae5847cd59ab1efcb4c46437ac::chmatch(chmatch_work1[[as.character(N)]], letters),
  Knuth_hash = data.table.d7a9a1707ec94ec4f2bd86a5dfb5609207029ba4::chmatch(chmatch_work1[[as.character(N)]], letters),
  collapse = collapse::fmatch(chmatch_work1[[as.character(N)]], letters)
)
rm(chmatch_work1); gc(full = TRUE)

save(chmatch1, file = 'times_collapse.rda')
```

</details>

And the real memory cost isn't even that large:
memory use as returned by getrusage()

<details>

```r
library(atime)
library(data.table)
# assumes that atime_versions() had pre-installed the packages
# master = '70c64ac08c6becae5847cd59ab1efcb4c46437ac',
# 'Knuth_hash' = 'd7a9a1707ec94ec4f2bd86a5dfb5609207029ba4',
library(data.table.70c64ac08c6becae5847cd59ab1efcb4c46437ac)
library(data.table.d7a9a1707ec94ec4f2bd86a5dfb5609207029ba4)
library(parallel)

# only tested on a recent Linux system
# measures the _maximal_ amount of memory in kB used by the current process
writeLines('
#include <sys/resource.h>
void maxrss(double * kb) {
  struct rusage ru;
  int ret = getrusage(RUSAGE_SELF, &ru);
  *kb = ret ? -1 : ru.ru_maxrss;
}
', 'maxrss.c')
tools::Rcmd('SHLIB maxrss.c')
dyn.load(paste0('maxrss', .Platform$dynlib.ext))

limit <- 1

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
versions <- expression(
  master = data.table.70c64ac08c6becae5847cd59ab1efcb4c46437ac::chmatch(chmatch_work1[[as.character(N)]], letters),
  Knuth_hash = data.table.d7a9a1707ec94ec4f2bd86a5dfb5609207029ba4::chmatch(chmatch_work1[[as.character(N)]], letters),
  collapse = collapse::fmatch(chmatch_work1[[as.character(N)]], letters)
)

plan <- expand.grid(N = N, version = names(versions))
chmatch1 <- lapply(seq_len(nrow(plan)), \(i) {
  # use a disposable child process
  mccollect(mcparallel({
    eval(versions[[plan$version[[i]]]], list(N = plan$N[[i]]))
    .C('maxrss', kb = double(1))$kb
  }))
})

rm(chmatch_work1); gc(full = TRUE)

save(chmatch1, file = 'times_collapse.rda')
chmatch1p <- lattice::xyplot(
  maxrss_kb ~ N, cbind(plan, maxrss_kb = unlist(chmatch1)), group = version,
  auto.key = TRUE, scales = list(log = 10),
  par.settings=list(superpose.symbol=list(pch=19))
)
```

</details>

SebKrantz:
Nice! Thanks. I always wondered about this tradeoff between the size of the table and the quality of the hash function. Looks like speed + large table still wins. Anyway, if you want to adopt, feel free to copy it under your MPL license. Just mention me in the top of the file and as a contributor.

SebKrantz:
PS: I believe it also depends on the size of the `table` in R. `letters` is small. I believe for much larger tables `chmatch()` should be faster. Would be interesting to also check those hash functions.

tdhock:
excellent work thank you very much

aitap:
The `chmatch()` performance problem is three-dimensional.

There's `length(x)`; we'd like the lookup time to be linear in that and spend as few instructions per lookup as possible. `TRUELENGTH()` is a very short function call.

Initialising the hash currently takes time and memory proportional to `length(table)`, but we really want it to be closer to `length(unique(table))`. See the benchmark where the `table` is 10 times as large as `x`:
https://github.com/Rdatatable/data.table/blob/9fe1b8d01701d2b23adcff3085ea0290da5df022/src/chmatch.c#L163-L170

A good `TRUELENGTH()` replacement needs to perform well when `length(x)` and `length(table)` are large and `length(unique(table))` is still reasonable. (`length(unique(table))` becomes unreasonable when it starts to strain the `CHARSXP` cache.)

I still have a few ideas how to speed this up.

src/chmatch.c
```diff
@@ -56,33 +55,14 @@ static SEXP chmatchMain(SEXP x, SEXP table, int nomatch, bool chin, bool chmatch
...
-  }
+  hashtab * marks = hash_create(tablelen);
   int nuniq=0;
   for (int i=0; i<tablelen; ++i) {
     const SEXP s = td[i];
-    int tl = TRUELENGTH(s);
-    if (tl>0) { savetl(s); tl=0; }
-    if (tl==0) SET_TRUELENGTH(s, chmatchdup ? -(++nuniq) : -i-1); // first time seen this string in table
+    int tl = hash_lookup(marks, s, 0);
+    if (tl==0) hash_set(marks, s, chmatchdup ? -(++nuniq) : -i-1); // first time seen this string in table
```
aitap, line 63: `hash_set` wastes time performing the lookup again. Need to expose a "lookup or create" operation somehow to avoid that.

MichaelChirico:
WDYT about the possibility of proposing our %chin% just be up streamed into R itself? Our version is "well-tested & battle-hardened" for many years, and seems naturally suited to r-devel given it's reliance on internal behavior.

src/hash.c
```diff
@@ -0,0 +1,226 @@
+#include <pthread.h>
+
+#include "data.table.h"
+
+struct hash_pair {
+    SEXP key;
+    R_xlen_t value;
```
aitap, line 7: Can halve the memory requirements by storing `int` indices into an array instead of `SEXP` pointers and `int` mark values instead of `R_xlen_t` (thanks to Sebastian Krantz for the idea). Cost: this makes the hash table incompatible with long vectors. Since `int` dimensions for matrices and `data.frame`s are baked into almost all of R, it's probably worth using `int`s here too.

src/forder.c
```diff
@@ -307,20 +302,22 @@ static void range_str(const SEXP *x, int n, uint64_t *out_min, uint64_t *out_max
       na_count++;
       continue;
     }
-    if (TRUELENGTH(s)<0) continue;  // seen this group before
-    #pragma omp critical
-    if (TRUELENGTH(s)>=0) {  // another thread may have set it while I was waiting, so check it again
-      if (TRUELENGTH(s)>0)   // save any of R's own usage of tl (assumed positive, so we can both count and save in one scan), to restore
-        savetl(s);           // afterwards. From R 2.14.0, tl is initialized to 0, prior to that it was random so this step saved too much.
-      // now save unique SEXP in ustr so i) we can loop through them afterwards and reset TRUELENGTH to 0 and ii) sort uniques when sorting too
+    // Why is it acceptable to call dhash_lookup when marks can be shared between threads?
```
aitap, line 305: So far the only way I know to make this provably safe is to throw away old hash table buffers (to be garbage collected by R).

Fast, thread-safe, dynamically growing, memory efficient hash table: choose two out of four, maybe three if you're lucky.

aitap:
@MichaelChirico, it may be worth a try. I suppose this would mean falling back to `TRUELENGTH` code on R < 4.6? Although the hash table by Sebastian Krantz is pretty fast and sometimes beats `data.table::chmatch` (need to explore all three performance dimensions to understand it better); we can have one like that too.

The `forder.c` problem is also quite hair-raising. We can't just replace it with `sort(method = "radix")`, can we?

The remaining uses of `TRUELENGTH` are much less performance-critical, for things that people usually don't have too many of (e.g. factor levels, columns).

jangorecki:
> Can't start implementing stretchy ALTREP vectors until data.table stops using TRUELENGTH to mark them.

It is worth noting that there is an ongoing effort, expected to start from May 2025 to work on R API growable vector API. Then we may expect easier, or more performant, ways to address current problems.

MichaelChirico:
If `%chin%` were accepted into r-devel, we could just mark `%chin%` as deprecated and meanwhile mask `` `%chin%` = `%in%`; chmatch = match ``, leaving users on older R with worse performance, WDYT

src/chmatch.c
ben-schwen: I guess we can do that analagous to `hash_set`?

```C
R_xlen_t hash_lookup_or_insert(hashtab *h, SEXP key, R_xlen_t value) {
  struct hash_pair *cell = h->tb + hash_index(key, h->multiplier) % h->size, *end = h->tb + h->size - 1;
  for (size_t i = 0; i < h->size; ++i, cell = (cell == end ? h->tb : cell + 1)) {
    if (cell->key == key) {
      return cell->value;  // Key exists, don't update, return value
    } else if (!cell->key) {
      if (!h->free) internal_error(
        __func__, "no free slots left (full size=%zu)", h->size
      );
      --h->free;
      *cell = (struct hash_pair){.key = key, .value = value};
      return value;  // insert here
    }
  }
  
  internal_error( // # nocov
    __func__, "did not find a free slot for key %p; size=%zu, free=%zu",
    (void*)key, h->size, h->free
  );
  // Should be impossible, but just in case:
  return value;
}

```

ben-schwen:
I extended Ivans benchmark a little further e.g. when our table gets bigger and also included NO matches. We definitely have to take a closer look at this with varying table size etc. (might also try other hashing methods like double, robin hood or cuckoo hashing)




<details>

```r
library(atime)
library(data.table)

# master = '70c64ac08c6becae5847cd59ab1efcb4c46437ac',
# 'Knuth_hash' = 'd7a9a1707ec94ec4f2bd86a5dfb5609207029ba4',

sample_strings = function(N=10, len=4) {
   do.call(paste0, replicate(len, sample(LETTERS, N, TRUE), FALSE))
}

N <- 10^seq(2, 7.5, .25)
tab_full = sample_strings(1e6, 10)
tab_small = sample(tab_full, 9e5)
# expected case: a few distinct strings
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(tab_full, N, TRUE)
)
chmatch1 <- atime(
  N,
  seconds.limit = limit, verbose = TRUE,
  master = data.table.70c64ac08c6becae5847cd59ab1efcb4c46437ac::chmatch(chmatch_work1[[as.character(N)]], tab_small),
  Knuth_hash = data.table.d7a9a1707ec94ec4f2bd86a5dfb5609207029ba4::chmatch(chmatch_work1[[as.character(N)]], tab_small),
  collapse = collapse::fmatch(chmatch_work1[[as.character(N)]], tab_small)
)
plot(chmatch1)

rm(chmatch_work1); gc(full = TRUE)

save(chmatch1, file = 'fmatch_missings.rda')
```
</details>

src/chmatch.c
ben-schwen: It can be found in 337a0c2d508a31c59885416d7929ff6d6a4b0bda. Gives a little speedup but not too much.

ben-schwen:
Another low hanging fruit:
When probing in `chmatch` we currently do not parallize after building the hash table, but probing could be essentially parallized (at least for the `chmatch` and `chin` cases).

``` c
#pragma omp parallel for num_threads(getDTthreads(xlen, true))
  for (int i=0; i<xlen; i++) {
    const int m = hash_lookup(marks,xd[i],0);
    ansd[i] = (m<0) ? -m : nomatch;
  }
```

edit: Without checking I suppose that `collapse` is essentially already doing this

times_collapse

``` r
library(atime)
library(data.table)
library(ggplot2)

limit <- 1

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chmatch1 <- atime(
  N,
  seconds.limit = limit, verbose = TRUE,
  master = data.table.3c90b0f80e1d5b54ea97f3b56f28df07b93e820a::chmatch(chmatch_work1[[as.character(N)]], letters),
  Knuth_hash = data.table.d7a9a1707ec94ec4f2bd86a5dfb5609207029ba4::chmatch(chmatch_work1[[as.character(N)]], letters),
  double_noxor = data.table.48b19422dd331a9d7eaddff507563bcf9643ebdd::chmatch(chmatch_work1[[as.character(N)]], letters),
  khash = data.table.3178f10af7860223b7bea0fba519104d6a9d55ca::chmatch(chmatch_work1[[as.character(N)]], letters),
  collapse = collapse::fmatch(chmatch_work1[[as.character(N)]], letters)
)
rm(chmatch_work1); gc(full = TRUE)

ggsave(plot(chmatch1), filename = 'times_collapse.png')

save(chmatch1, file = 'times_collapse.rda')

```

</details>

ben-schwen:
I updated the benchmarks to the current head without `dhashtab`. Timings look good, but we might think about a smarter approach for initializing large hashtables in `chmatch`?

`double_hashing` is the head of this PR `66ac645`
`linear_probing` was the first version of the hashtable approach in `56dcbf9`
`master` is the current dev-version

image

image

image


Also added some comparison of `match` and `chmatch`

image


<details>

``` r
library(atime)
library(data.table)

pkg.path <- '.'
limit <- 1
# taken from .ci/atime/tests.R
pkg.edit.fun <- function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}

versions <- c(
  master = '3c90b0f80e1d5b54ea97f3b56f28df07b93e820a',
  double_hashing = '66ac645b56cf676a9a0808e74feb6d89ccf203aa',
  linear_probing = '56dcbf9834d063fbaf81e9b19d24a5450fb52adc'
)

set.seed(3)
sample_strings = function(N=10, len=4) {
   do.call(paste0, replicate(len, sample(LETTERS, N, TRUE), FALSE))
}

N <- 10^seq(2, 7.5, .25)

# Benchmark 1: Large table 
tab_full = sample_strings(1e6, 10)
tab_small = sample(tab_full, 9e5)
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(tab_full, N, TRUE)
)

chmatch1 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work1[[as.character(N)]], tab_small),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch1)

# Benchmark 2: Small table (expected case: a few distinct strings - 26 letters)
chmatch_work2 <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chmatch2 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work2[[as.character(N)]], letters),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch2)

# Benchmark 3: Medium table (10K unique strings)
medium_table <- sample_strings(10000, len=4)
chmatch_work3 <- lapply(setNames(nm = N), \(N)
  sample(medium_table, N, TRUE)
)
chmatch3 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work3[[as.character(N)]], medium_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch3)

# Benchmark 4: High miss rate (most items not found)
miss_table <- sample_strings(100, len=4)
chmatch_work4 <- lapply(setNames(nm = N), \(N)
  sample_strings(N, len=4)
)
chmatch4 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work4[[as.character(N)]], miss_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch4)

# Benchmark 5: All hits (every item found)
hit_table <- sample_strings(1000, len=4)
chmatch_work5 <- lapply(setNames(nm = N), \(N)
  sample(hit_table, N, TRUE)
)
chmatch5 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work5[[as.character(N)]], hit_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch5)

# Benchmark 6: chin (logical return) with small table
chin_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chin_bench <- atime_versions(
  pkg.path, N,
  expr = data.table::`%chin%`(chin_work[[as.character(N)]], letters),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chin_bench)

# Benchmark 7: Long strings (16 chars)
long_table <- sample_strings(1000, len=16)
chmatch_work7 <- lapply(setNames(nm = N), \(N)
  sample(long_table, N, TRUE)
)
chmatch7 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work7[[as.character(N)]], long_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch7)

# Benchmark 8: Realistic join keys (90% hit rate)
join_table <- sample_strings(10000, len=8)
chmatch_work8 <- lapply(setNames(nm = N), \(N) {
  # 90% from join_table (will be found), 10% new (won't be found)
  c(
    sample(join_table, as.integer(N * 0.9), TRUE),
    sample_strings(as.integer(N * 0.1), len=8)
  )
})
chmatch8 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work8[[as.character(N)]], join_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch8)

# Compare chmatch vs base::match across common scenarios
compare_N <- 10^seq(2, 6, .5)
chmatch_match_work_small <- lapply(setNames(nm = compare_N), \(N)
  sample(letters, N, TRUE)
)
chmatch_match_small <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_small[[as.character(N)]], letters)),
    match = quote(base::match(chmatch_match_work_small[[as.character(N)]], letters))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_small)

chmatch_match_work_medium <- lapply(setNames(nm = compare_N), \(N)
  sample(medium_table, N, TRUE)
)
chmatch_match_medium <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_medium[[as.character(N)]], medium_table)),
    match = quote(base::match(chmatch_match_work_medium[[as.character(N)]], medium_table))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_medium)

chmatch_match_work_large <- lapply(setNames(nm = compare_N), \(N)
  sample(tab_full, N, TRUE)
)
chmatch_match_large <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_large[[as.character(N)]], tab_small)),
    match = quote(base::match(chmatch_match_work_large[[as.character(N)]], tab_small))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_large)

chmatch_match_work_miss <- lapply(setNames(nm = compare_N), \(N)
  sample_strings(N, len = 4)
)
chmatch_match_miss <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_miss[[as.character(N)]], miss_table)),
    match = quote(base::match(chmatch_match_work_miss[[as.character(N)]], miss_table))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_miss)


N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
forderv1_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
forderv1 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(forderv1)
rm(forderv1_work); gc(full = TRUE)

# worst case: all strings different
# (a challenge for the allocator too due to many small immovable objects)
N <- 10^seq(2, 7.5, .25)
forderv2_work <- lapply(setNames(nm = N), \(N)
  format(runif(N), digits = 16)
)
forderv2 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv2_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(forderv2)
rm(forderv2_work); gc(full = TRUE)

# expected case: all columns named the same
N <- 10^seq(1, 5.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist1_work <- lapply(setNames(nm = N), \(N)
  rep(list(setNames(as.list(1:k), letters[1:k])), N)
)
rbindlist1 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(rbindlist1)
rm(rbindlist1_work); gc(full = TRUE)

# worst case: all columns different
N <- 10^seq(1, 4.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist2_work <- lapply(setNames(nm = N), \(N)
  replicate(N, setNames(as.list(1:k), format(runif(k), digits = 16)), FALSE)
)
rbindlist2 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist2_work[[as.character(N)]], fill = TRUE),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(rbindlist2)
rm(rbindlist2_work); gc(full = TRUE)

# ===== SAVE BENCHMARK RESULTS =====
results_dir <- "benchmark_results"
dir.create(results_dir, showWarnings = FALSE)

# Save R objects
save(
  chmatch1, chmatch2, chmatch3, chmatch4, chmatch5,
  chin_bench, chmatch7, chmatch8,
  chmatch_match_small, chmatch_match_medium,
  chmatch_match_large, chmatch_match_miss,
  forderv1, forderv2,
  rbindlist1, rbindlist2,
  file = file.path(results_dir, "benchmarks_hash.rda")
)

library(atime)
library(data.table)
results_dir <- "benchmark_results"
load(file.path(results_dir, "benchmarks_hash.rda"))

library(ggplot2)
library(patchwork)   # install.packages("patchwork") if needed

# helper: ensure we get a ggplot and add a title
g <- function(p, title) p + ggtitle(title)

p <- list(
  g(plot(chmatch1), "Large table (900K unique)"),
  g(plot(chmatch2), "Small table (26 letters)"),
  g(plot(chmatch3), "Medium table (10K unique)"),
  g(plot(chmatch4), "High miss rate"),
  g(plot(chmatch5), "All hits"),
  g(plot(chin_bench), "chin (logical return)"),
  g(plot(chmatch7), "Long strings (16 chars)"),
  g(plot(chmatch8), "Join keys (90% hit rate)"),
  g(plot(forderv1), "forderv: few distinct strings"),
  g(plot(forderv2), "forderv: all unique strings"),
  g(plot(rbindlist1), "rbindlist: same columns"),
  g(plot(rbindlist2), "rbindlist: different columns"),
  g(plot(chmatch_match_small), "chmatch vs match: small table"),
  g(plot(chmatch_match_medium), "chmatch vs match: medium table"),
  g(plot(chmatch_match_large), "chmatch vs match: large table"),
  g(plot(chmatch_match_miss), "chmatch vs match: high miss rate")
)

# write a multipage PDF with 2x2 panels per page
pdf("benchmark_results.pdf", width = 10, height = 8)
for (i in seq(1, length(p), by = 4)) {
  pg <- wrap_plots(p[i:min(i+3, length(p))], ncol = 2, nrow = 2, guides = "collect")
  print(pg)
}
dev.off()
```

</details>

MichaelChirico:
I never got around to writing out a case for R-core to adopt our `chmatch()` implementation in `match()` itself given they have access to the private API. Does this benchmark work in support of that argument? I guess `chmatch`-on-`master` is not compared to `match`-in-r-devel directly...

ben-schwen:
@MichaelChirico the benchmarks from comparing `match` with `chmatch` don't seem convincing enough for me for R-core to introduce extra code they need to maintain

image

<details>

```r
library(atime)
library(data.table)

set.seed(3)
sample_strings = function(N=10, len=4) {
   do.call(paste0, replicate(len, sample(LETTERS, N, TRUE), FALSE))
}

compare_N <- 10^seq(2, 6, .5)
limit <- 1

tab_full = sample_strings(1e6, 10)
tab_small = sample(tab_full, 9e5)
medium_table <- sample_strings(10000, len=4)
miss_table <- sample_strings(100, len=4)

# Benchmark 1: Small table comparison (26 letters)
chmatch_match_work_small <- lapply(setNames(nm = compare_N), \(N)
  sample(letters, N, TRUE)
)
chmatch_match_small <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_small[[as.character(N)]], letters)),
    match = quote(base::match(chmatch_match_work_small[[as.character(N)]], letters))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_small)

# Benchmark 2: Medium table comparison (10K unique)
chmatch_match_work_medium <- lapply(setNames(nm = compare_N), \(N)
  sample(medium_table, N, TRUE)
)
chmatch_match_medium <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_medium[[as.character(N)]], medium_table)),
    match = quote(base::match(chmatch_match_work_medium[[as.character(N)]], medium_table))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_medium)

# Benchmark 3: Large table comparison (900K unique)
chmatch_match_work_large <- lapply(setNames(nm = compare_N), \(N)
  sample(tab_full, N, TRUE)
)
chmatch_match_large <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_large[[as.character(N)]], tab_small)),
    match = quote(base::match(chmatch_match_work_large[[as.character(N)]], tab_small))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_large)

# Benchmark 4: High miss rate comparison
chmatch_match_work_miss <- lapply(setNames(nm = compare_N), \(N)
  sample_strings(N, len = 4)
)
chmatch_match_miss <- atime(
  N = compare_N,
  expr.list = list(
    chmatch = quote(data.table::chmatch(chmatch_match_work_miss[[as.character(N)]], miss_table)),
    match = quote(base::match(chmatch_match_work_miss[[as.character(N)]], miss_table))
  ),
  seconds.limit = limit, verbose = TRUE
)
plot(chmatch_match_miss)

results_dir <- "benchmark_results"
dir.create(results_dir, showWarnings = FALSE)

save(
  chmatch_match_small, chmatch_match_medium,
  chmatch_match_large, chmatch_match_miss,
  file = file.path(results_dir, "benchmarks_chmatch_vs_match.rda")
)

library(ggplot2)
library(patchwork)

g <- function(p, title) p + ggtitle(title)

p <- list(
  g(plot(chmatch_match_small), "chmatch vs match: small table (26 letters)"),
  g(plot(chmatch_match_medium), "chmatch vs match: medium table (10K unique)"),
  g(plot(chmatch_match_large), "chmatch vs match: large table (900K unique)"),
  g(plot(chmatch_match_miss), "chmatch vs match: high miss rate")
)

pdf("benchmark_chmatch_vs_match.pdf", width = 10, height = 8)
pg <- wrap_plots(p, ncol = 2, nrow = 2, guides = "collect")
print(pg)
dev.off()

```

</details>

src/hash.c
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // need to rehash
+  return false;
+}
+
+void hash_set(hashtab *h, SEXP key, R_xlen_t value) {
+  if (!hash_set_(h, key, value))
+    *h = *hash_rehash(h);
+  (void)hash_set_(h, key, value); // must succeed on the second try
```
HughParsonage: Should this be 

```c
void hash_set(hashtab *h, SEXP key, R_xlen_t value) {
  if (!hash_set_(h, key, value)) {
    *h = *hash_rehash(h);
    if (!hash_set_(h, key, value)) internal_error(__func__, "rehash failed"); 
  }
}
```

i.e. only call `hash_set_` twice if the first attempt failed?

(the call to `internal_error` is optional)

src/hash.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+hashtab *hash_set_shared(hashtab *h, SEXP key, R_xlen_t value) {
+  if (!hash_set_(h, key, value))
+    h = hash_rehash(h);
+  (void)hash_set_(h, key, value);
+  return h;
+}
```
HughParsonage, line 171: Similar issue as in `hash_set`

src/hash.c
```diff
@@ -0,0 +1,141 @@
...
+  else if ((h2 & 1) == 0) h2 |= 1;
+
+  for (size_t i = 0; i < h->size; ++i) {
+    size_t idx = (h1 + i * h2) & mask;
+
+    if (!h->table[idx].key) {
+      // Empty slot found
+      h->table[idx].key = key;
+      h->table[idx].value = value;
+      h->free--;
```
HughParsonage, line 116: Is there a risk of underflow here (with `h->free`)?

src/forder.c
```diff
@@ -295,36 +291,35 @@ static void cradix(SEXP *x, int n)
...
         ustr_alloc = (ustr_alloc==0) ? 16384 : ustr_alloc*2;  // small initial guess, negligible time to alloc 128KiB (32 pages)
         if (ustr_alloc>n) ustr_alloc = n;  // clamp at n. Reaches n when fully unique (no dups)
         ustr = realloc(ustr, sizeof(SEXP) * ustr_alloc);
         if (ustr==NULL) STOP(_("Unable to realloc %d * %d bytes in range_str"), ustr_alloc, (int)sizeof(SEXP));  // # nocov
       }
       ustr[ustr_n++] = s;
-      SET_TRUELENGTH(s, -ustr_n);  // unique in any order is fine. first-appearance order is achieved later in count_group
+      // Under the OpenMP memory model, if the hash table is expanded, we have to explicitly update the pointer.
+      #pragma omp atomic write
+      marks = hash_set_shared(marks, s, -ustr_n); // unique in any order is fine. first-appearance order is achieved later in count_group
```
HughParsonage: Is there a way to avoid this atomic? Can it be hoisted out of the parallel region? It looks to me like there is a (potential) call to `R_alloc` through this function, in which case `#pragma omp atomic write` won't be sufficient.

src/rbindlist.c
```diff
@@ -74,7 +74,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
     SEXP *uniq = malloc(sizeof(*uniq) * upperBoundUniqueNames);  // upperBoundUniqueNames was initialized with 1 to ensure this is defined (otherwise 0 when no item has names)
     if (!uniq)
       error(_("Failed to allocate upper bound of %"PRId64" unique column names [sum(lapply(l,ncol))]"), (int64_t)upperBoundUniqueNames); // # nocov
-    savetl_init();
+    hashtab * marks = hash_create(1 + (LENGTH(l) ? (2 * LENGTH(getAttrib(VECTOR_ELT(l, 0), R_NamesSymbol))) : 0));
```
HughParsonage: The choice to use the first name to size the hash table seems arbitrary. Is there a better way? Using `upperBoundUniqueNames`?

src/forder.c
aitap: Sigh. You're right. `fread` already has this problem (and blows up with `--enable-strict-barrier` when a non-main thread happens to call `mkChar()`). Without the strict barrier checking, this is "safe" thanks to the critical section and the atomic write, but it's far from perfect.

The problem is that inside this critical section, the hash table may _sometimes_ need to grow. Other threads reading from the hash table can see stale content, which is fine, but we shouldn't `free` the table from under them or let them see it in an inconsistent state.

I can make this into a chain of `malloc`ed objects (to avoid `R_alloc` on a non-main thread), reintroduce the [memory protection machinery](https://github.com/Rdatatable/data.table/pull/6694/commits/66ac645b56cf676a9a0808e74feb6d89ccf203aa) and perform the atomic write only if `hash_set_shared` returns a different pointer.

src/hash.c
aitap: Actually, yes. We fill the table completely, exceeding the load factor, before we rehash. This currently doesn't break anything because we don't use `free` besides decrementing it here. Thanks for spotting this!

ben-schwen:
Regarding speed, it seems that we have reached point where this PR is on par or faster than master!

NB: `chmatchdup` case is not considered yet.

image

image

image

<details>

```r
library(atime)
library(data.table)

pkg.path <- '.'
limit <- 1
# taken from .ci/atime/tests.R
pkg.edit.fun <- function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}

versions <- c(
  master = '585ab23004d9b331dc2d53843c0a970fae123dd7',
  thisPR = 'c93324719ed2aa36f9748268179535db5366b981'
)

set.seed(3)
sample_strings = function(N=10, len=4) {
   do.call(paste0, replicate(len, sample(LETTERS, N, TRUE), FALSE))
}

N <- 10^seq(2, 7.5, .25)

# Benchmark 1: Large table 
tab_full = sample_strings(1e6, 10)
tab_small = sample(tab_full, 9e5)
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(tab_full, N, TRUE)
)

chmatch1 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work1[[as.character(N)]], tab_small),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch1)

# Benchmark 2: Small table (expected case: a few distinct strings - 26 letters)
chmatch_work2 <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chmatch2 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work2[[as.character(N)]], letters),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch2)

# Benchmark 3: Medium table (10K unique strings)
medium_table <- sample_strings(10000, len=4)
chmatch_work3 <- lapply(setNames(nm = N), \(N)
  sample(medium_table, N, TRUE)
)
chmatch3 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work3[[as.character(N)]], medium_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch3)

# Benchmark 4: High miss rate (most items not found)
miss_table <- sample_strings(100, len=4)
chmatch_work4 <- lapply(setNames(nm = N), \(N)
  sample_strings(N, len=4)
)
chmatch4 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work4[[as.character(N)]], miss_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch4)

# Benchmark 5: All hits (every item found)
hit_table <- sample_strings(1000, len=4)
chmatch_work5 <- lapply(setNames(nm = N), \(N)
  sample(hit_table, N, TRUE)
)
chmatch5 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work5[[as.character(N)]], hit_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch5)

# Benchmark 6: chin (logical return) with small table
chin_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chin_bench <- atime_versions(
  pkg.path, N,
  expr = data.table::`%chin%`(chin_work[[as.character(N)]], letters),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chin_bench)

# Benchmark 7: Long strings (16 chars)
long_table <- sample_strings(1000, len=16)
chmatch_work7 <- lapply(setNames(nm = N), \(N)
  sample(long_table, N, TRUE)
)
chmatch7 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work7[[as.character(N)]], long_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch7)

# Benchmark 8: Realistic join keys (90% hit rate)
join_table <- sample_strings(10000, len=8)
chmatch_work8 <- lapply(setNames(nm = N), \(N) {
  # 90% from join_table (will be found), 10% new (won't be found)
  c(
    sample(join_table, as.integer(N * 0.9), TRUE),
    sample_strings(as.integer(N * 0.1), len=8)
  )
})
chmatch8 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work8[[as.character(N)]], join_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch8)

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
forderv1_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
forderv1 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(forderv1)
rm(forderv1_work); gc(full = TRUE)

# worst case: all strings different
# (a challenge for the allocator too due to many small immovable objects)
N <- 10^seq(2, 7.5, .25)
forderv2_work <- lapply(setNames(nm = N), \(N)
  format(runif(N), digits = 16)
)
forderv2 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv2_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(forderv2)
rm(forderv2_work); gc(full = TRUE)

# expected case: all columns named the same
N <- 10^seq(1, 5.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist1_work <- lapply(setNames(nm = N), \(N)
  rep(list(setNames(as.list(1:k), letters[1:k])), N)
)
rbindlist1 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(rbindlist1)
rm(rbindlist1_work); gc(full = TRUE)

# worst case: all columns different
N <- 10^seq(1, 4.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist2_work <- lapply(setNames(nm = N), \(N)
  replicate(N, setNames(as.list(1:k), format(runif(k), digits = 16)), FALSE)
)
rbindlist2 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist2_work[[as.character(N)]], fill = TRUE),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(rbindlist2)
rm(rbindlist2_work); gc(full = TRUE)

# ===== SAVE BENCHMARK RESULTS =====
results_dir <- "benchmark_results"
dir.create(results_dir, showWarnings = FALSE)

# Save R objects
save(
  chmatch1, chmatch2, chmatch3, chmatch4, chmatch5,
  chin_bench, chmatch7, chmatch8,
  forderv1, forderv2,
  rbindlist1, rbindlist2,
  file = file.path(results_dir, "benchmarks_hash.rda")
)

library(atime)
library(data.table)
results_dir <- "benchmark_results"
load(file.path(results_dir, "benchmarks_hash.rda"))

library(ggplot2)
library(patchwork)   # install.packages("patchwork") if needed

# helper: ensure we get a ggplot and add a title
g <- function(p, title) p + ggtitle(title)

p <- list(
  g(plot(chmatch1), "Large table (900K unique)"),
  g(plot(chmatch2), "Small table (26 letters)"),
  g(plot(chmatch3), "Medium table (10K unique)"),
  g(plot(chmatch4), "High miss rate"),
  g(plot(chmatch5), "All hits"),
  g(plot(chin_bench), "chin (logical return)"),
  g(plot(chmatch7), "Long strings (16 chars)"),
  g(plot(chmatch8), "Join keys (90% hit rate)"),
  g(plot(forderv1), "forderv: few distinct strings"),
  g(plot(forderv2), "forderv: all unique strings"),
  g(plot(rbindlist1), "rbindlist: same columns"),
  g(plot(rbindlist2), "rbindlist: different columns")
)

# write a multipage PDF with 2x2 panels per page
pdf("benchmark_results.pdf", width = 10, height = 8)
for (i in seq(1, length(p), by = 4)) {
  pg <- wrap_plots(p[i:min(i+3, length(p))], ncol = 2, nrow = 2, guides = "collect")
  print(pg)
}
dev.off()
```

</details>

src/assign.c
```diff
@@ -855,45 +844,41 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
...
           const SEXP *sourceD = STRING_PTR_RO(source);
           for (int i=0; i<nSource; ++i) {  // convert source integers to refer to target levels
             const SEXP val = sourceD[i];
-            newSourceD[i] = val==NA_STRING ? NA_INTEGER : -TRUELENGTH(val);
+            newSourceD[i] = val==NA_STRING ? NA_INTEGER : -hash_lookup(marks, val, 0);
           }
         }
         source = newSource;
-        for (int k=0; k<nTargetLevels; ++k) SET_TRUELENGTH(targetLevelsD[k], 0);  // don't need those anymore
+        for (int k=0; k<nTargetLevels; ++k) hash_set(marks, targetLevelsD[k], 0);  // don't need those anymore
```
ben-schwen, line 795: do we even need this cleanup step now?

aitap: Empirically, yes. There is a later check for a non-zero mark, and if this line is omitted, the check doesn't fire the expected number of times and raises an internal error (with a small memory leak):
https://github.com/Rdatatable/data.table/blob/41e112348bb13d08132f50ab58b9d376c6343d07/src/assign.c#L882-L884

src/forder.c
aitap: The hash table now uses the C heap and offers an option to check for allocation failures instead of singalling an R error. In most places, I've switched temporary heap allocations to `R_alloc` (using `vmaxset` to unprotect them when they were no longer used), but there was one more place in `rbindlist` where manual allocation checking was the least invasive option. (It could definitely be rewritten to use the R heap instead.)

src/hash.c
```diff
@@ -45,38 +47,65 @@ static R_INLINE size_t get_full_size(size_t n_elements, double load_factor) {
...
+static struct hash_tab * hash_create_(size_t n, double load_factor) {
   size_t n_full = get_full_size(n, load_factor);
-  hashtab *ret = (hashtab *)R_alloc(sizeof(hashtab), 1);
+
+  struct hash_tab *ret = malloc(sizeof *ret);
+  if (!ret)
+    return NULL; // #nocov
+
+  // Not strictly portable but faster than an explicit loop setting keys to NULL
+  struct hash_pair *table = calloc(n_full, sizeof(*table));
```
ben-schwen, line 58: OK but `calloc` zeroes keys and values, while we only need to NULL `keys`?

ben-schwen: edit: I just tested all three options with `inline` and there is no noticeable different between `calloc`, `malloc` + `loop`, `malloc` + `memset`

ben-schwen:
We should probably also add a NOTE regarding all the changes

Internal use of declared non-API R functions `SETLENGTH`, `TRUELENGTH`, `SET_TRUELENGTH`, and `SET_GROWABLE_BIT` has been eliminated. Most usages have been migrated to R's experimental resizable vectors API (thanks to @ltierney, introduced in R 4.6.0, backported for older R versions), [#7451](https://github.com/Rdatatable/data.table/pull/7451). Uses of `TRUELENGTH` for marking seen items during grouping and binding operations (aka free hash table trick) have been replaced with proper hash tables, [#6694](https://github.com/Rdatatable/data.table/pull/6694). The new hash table implementation uses linear probing with power of 2 tables and automatic resizing. Additionally, `chmatch()` now hashes the needle (`x`) instead of the haystack (`table`) when `length(table) >> length(x)`, significantly improving performance for lookups into large tables. TODO ADD MORE INFO about tables?
Thanks to @aitap, @ben-schwen, @jangorecki and @HughParsonage for implementation and reviews.


probably also last benchmark (since we look on par with `1.17.8`)

image
image
image


<details>

```r
library(atime)
library(data.table)

pkg.path <- '.'
limit <- 1
# taken from .ci/atime/tests.R
pkg.edit.fun <- function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  pkg_find_replace(
    "NAMESPACE",
}

versions <- c(
  '1.17.8' = '4152c6cff80c60896e86b80f8b1c901a4ae93354',
  'truehash' = 'a7f5c6c0c0edd098793837525205244f49c41aa2'
)

set.seed(3)
sample_strings = function(N=10, len=4) {
   do.call(paste0, replicate(len, sample(LETTERS, N, TRUE), FALSE))
}

N <- 10^seq(2, 7.5, .25)

# Benchmark 1: Large table 
tab_full = sample_strings(1e6, 10)
tab_small = sample(tab_full, 9e5)
chmatch_work1 <- lapply(setNames(nm = N), \(N)
  sample(tab_full, N, TRUE)
)

chmatch1 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work1[[as.character(N)]], tab_small),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch1)

# Benchmark 2: Small table (expected case: a few distinct strings - 26 letters)
chmatch_work2 <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chmatch2 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work2[[as.character(N)]], letters),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch2)

# Benchmark 3: Medium table (10K unique strings)
medium_table <- sample_strings(10000, len=4)
chmatch_work3 <- lapply(setNames(nm = N), \(N)
  sample(medium_table, N, TRUE)
)
chmatch3 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work3[[as.character(N)]], medium_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch3)

# Benchmark 4: High miss rate (most items not found)
miss_table <- sample_strings(100, len=4)
chmatch_work4 <- lapply(setNames(nm = N), \(N)
  sample_strings(N, len=4)
)
chmatch4 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work4[[as.character(N)]], miss_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch4)

# Benchmark 5: All hits (every item found)
hit_table <- sample_strings(1000, len=4)
chmatch_work5 <- lapply(setNames(nm = N), \(N)
  sample(hit_table, N, TRUE)
)
chmatch5 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work5[[as.character(N)]], hit_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch5)

# Benchmark 6: chin (logical return) with small table
chin_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
chin_bench <- atime_versions(
  pkg.path, N,
  expr = data.table::`%chin%`(chin_work[[as.character(N)]], letters),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chin_bench)

# Benchmark 7: Long strings (16 chars)
long_table <- sample_strings(1000, len=16)
chmatch_work7 <- lapply(setNames(nm = N), \(N)
  sample(long_table, N, TRUE)
)
chmatch7 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work7[[as.character(N)]], long_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch7)

# Benchmark 8: Realistic join keys (90% hit rate)
join_table <- sample_strings(10000, len=8)
chmatch_work8 <- lapply(setNames(nm = N), \(N) {
  # 90% from join_table (will be found), 10% new (won't be found)
  c(
    sample(join_table, as.integer(N * 0.9), TRUE),
    sample_strings(as.integer(N * 0.1), len=8)
  )
})
chmatch8 <- atime_versions(
  pkg.path, N,
  expr = data.table::chmatch(chmatch_work8[[as.character(N)]], join_table),
  seconds.limit = limit, verbose = TRUE, sha.vec = versions,
  pkg.edit.fun = pkg.edit.fun
)
plot(chmatch8)

N <- 10^seq(2, 7.5, .25)
# expected case: a few distinct strings
forderv1_work <- lapply(setNames(nm = N), \(N)
  sample(letters, N, TRUE)
)
forderv1 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(forderv1)
rm(forderv1_work); gc(full = TRUE)

# worst case: all strings different
# (a challenge for the allocator too due to many small immovable objects)
N <- 10^seq(2, 7.5, .25)
forderv2_work <- lapply(setNames(nm = N), \(N)
  format(runif(N), digits = 16)
)
forderv2 <- atime_versions(
  pkg.path, N,
  expr = data.table:::forderv(forderv2_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(forderv2)
rm(forderv2_work); gc(full = TRUE)

# expected case: all columns named the same
N <- 10^seq(1, 5.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist1_work <- lapply(setNames(nm = N), \(N)
  rep(list(setNames(as.list(1:k), letters[1:k])), N)
)
rbindlist1 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist1_work[[as.character(N)]]),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(rbindlist1)
rm(rbindlist1_work); gc(full = TRUE)

# worst case: all columns different
N <- 10^seq(1, 4.5, .25) # number of data.tables in the list
k <- 10 # number of columns per data.table
rbindlist2_work <- lapply(setNames(nm = N), \(N)
  replicate(N, setNames(as.list(1:k), format(runif(k), digits = 16)), FALSE)
)
rbindlist2 <- atime_versions(
  pkg.path, N,
  expr = data.table::rbindlist(rbindlist2_work[[as.character(N)]], fill = TRUE),
  sha.vec = versions, seconds.limit = limit, verbose = TRUE,
  pkg.edit.fun = pkg.edit.fun
)
plot(rbindlist2)
rm(rbindlist2_work); gc(full = TRUE)

# ===== SAVE BENCHMARK RESULTS =====
results_dir <- "benchmark_results"
dir.create(results_dir, showWarnings = FALSE)

# Save R objects
save(
  chmatch1, chmatch2, chmatch3, chmatch4, chmatch5,
  chin_bench, chmatch7, chmatch8,
  forderv1, forderv2,
  rbindlist1, rbindlist2,
  file = file.path(results_dir, "benchmarks_hash.rda")
)

library(atime)
library(data.table)
results_dir <- "benchmark_results"
load(file.path(results_dir, "benchmarks_hash.rda"))

library(ggplot2)
library(patchwork)   # install.packages("patchwork") if needed

# helper: ensure we get a ggplot and add a title
g <- function(p, title) p + ggtitle(title)

p <- list(
  g(plot(chmatch1), "Large table (900K unique)"),
  g(plot(chmatch2), "Small table (26 letters)"),
  g(plot(chmatch3), "Medium table (10K unique)"),
  g(plot(chmatch4), "High miss rate"),
  g(plot(chmatch5), "All hits"),
  g(plot(chin_bench), "chin (logical return)"),
  g(plot(chmatch7), "Long strings (16 chars)"),
  g(plot(chmatch8), "Join keys (90% hit rate)"),
  g(plot(forderv1), "forderv: few distinct strings"),
  g(plot(forderv2), "forderv: all unique strings"),
  g(plot(rbindlist1), "rbindlist: same columns"),
  g(plot(rbindlist2), "rbindlist: different columns")
)

# write a multipage PDF with 2x2 panels per page
pdf("benchmark_results.pdf", width = 10, height = 8)
for (i in seq(1, length(p), by = 4)) {
  pg <- wrap_plots(p[i:min(i+3, length(p))], ncol = 2, nrow = 2, guides = "collect")
  print(pg)
}
dev.off()

```

</details>

MichaelChirico:
Probably also prudent to add something along the lines of "We've benchmarked the refactored code and find the performance satisfactory, but do report any edge case performance regressions we may have missed."

src/hash.c
aitap: `calloc` also saves one overflow check, while an explicit loop is more formally correct. Thank you, by the way, for all the benchmarking!

aitap:
Oh no.

```r
# use 6144 elements instead of 6145 and it doesn't trigger
strings = apply(outer(1:6145, 95^(0:2), function(x, y) as.raw(0xa1 + floor(x / y) %% 95)), 1, rawToChar)
Encoding(strings) = "latin1"
DT = data.table(x = strings)
setorder(DT, x)
plot(match(DT$x, sort.int(strings, method='radix')))
```
it's not a diagonal line at all

Bisect points at c93324719ed2aa36f9748268179535db5366b981. I'm also investigating, but does anyone else have any pointers?

***

6145 is, of course, `8192*.75+1`, where 8192 is the initial true size of the hash table allocated for 4096 elements here:

https://github.com/Rdatatable/data.table/blob/c93324719ed2aa36f9748268179535db5366b981/src/forder.c#L611

...and `0.75` is its load factor. The 6145th element triggers a rehash in `range_str` → `hash_set_shared`:

https://github.com/Rdatatable/data.table/blob/c93324719ed2aa36f9748268179535db5366b981/src/forder.c#L322

(And according to coverage data, we never rehash in place via `hash_set`.)

***

Rehashing itself seems completely fine. The newly created table is empty, as it should be. Once the elements are set, I can walk both tables and verify that the other one looks up the same value

It doesn't seem to be due to threading as well, because this still happens after `setDTthreads(1)`.

Naturally, neither Valgrind nor the sanitizers see any problems.

aitap:
Found the problem: changes to `marks` that happened inside `range_str` did not propagate to its caller. My bad. Sorry.

TysonStanley:
Really incredible work here. Looks like some conflicts to resolve before merging. Believe this is the last *necessary* merge before 1.18.0?

aitap:
Yes, this is the last PR needed to get rid of the non-API NOTE (for now).

Looks like rehashing is unreachable from `hash_set` because we always either pre-allocate a large enough table, or (in case of `forder`) pre-populate it with all the elements it needs. What to do, `# nocov` it for now?

aitap:
Changes to GLCI config tested on GitLab:
https://gitlab.com/Rdatatable/data.table/-/pipelines/2214728801

UBSan report also addressed:
https://gitlab.com/Rdatatable/data.table/-/jobs/12435568788

We should probably specify the number of arguments when registering our
callMethods to avoid surprises like this from UBSan.

ben-schwen:
> Looks like rehashing is unreachable from `hash_set` because we always either pre-allocate a large enough table, or (in case of `forder`) pre-populate it with all the elements it needs. What to do, `# nocov` it for now?

I would not cover it with `# nocov` so we become aware when we finally cover it (in the future), but I guess thats just personal preference

jangorecki (approved):

PR #6696, opened 2024-12-27
closes #4519 added a reference page for .internal.selfref

### LINKED ISSUE #4519: Need a reference page for .internal.selfref
Currently barely mentioned in the manual:

```
grep -Flr ".internal.selfref" man
man/all.equal.data.table.Rd
```

Would be useful to have some brief mention about what exactly this is, preferably available at `?".internal.selfref"`

venom1204:
closes https://github.com/Rdatatable/data.table/issues/4519
i added a seperate file internal.selfref.rd
containing its

1) Purpose and Functionality
2) Behavior and Use
3) User Interaction
4) Examples
5) Reference to Related Functions

@MichaelChirico can you please review it.


venom1204:
I introduced the changes requested by you 
and for the reference of the internal.selfref i have added its guide path in the NEWS.md and the NEWS0.md 
as they were the places where i find the internal.selfref .
is it fine now @aitap @MichaelChirico can you review it or there are any more changes i need to do ?

aitap:
Thank you very much for the fixes!

When I mentioned linking to `internal.selfref` from other help pages, I meant the help pages in `man/*.Rd`, not `inst/NEWS*md`. I don't think that a link from `NEWS.md` to an Rd file will work in an installed R package; but `\link{...}`s will work if you run (for example) `help(data.table, data.table, help_type='html')`. The idea is to make the new help page [a part of the same graph](https://llrs.dev/post/2024/09/23/cran-files-3/), reachable by links from other _help_ pages.

venom1204:
Yha the link will not be working from .md file to .rd file that whi I provided the directory path for the internal.selfref .
And I did not include it in the man.r as I did not find the use of internal.selfref much .
Should I make change again ?

aitap:
Writing good documentation is hard. When reading the help pages for
data.table, where would the user benefit the most from knowing about
the .internal.selfref attribute? I think it would be appropriate to
link to the new help page from data.table.Rd, assign.Rd, and copy.Rd.

man/assign.Rd
```diff
@@ -92,12 +92,12 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
...
 \code{DT} is modified by reference and returned invisibly. If you require a copy, take a \code{\link{copy}} first (using \code{DT2 = copy(DT)}).
 }
-\seealso{ \code{\link{data.table}}, \code{\link{copy}}, \code{\link{setalloccol}}, \code{\link{truelength}}, \code{\link{set}}, \code{\link{.Last.updated}}
+\seealso{ \code{\link{data.table}}, \code{\link{copy}}, \code{\link{setalloccol}}, \code{\link{truelength}}, \code{\link{set}}, \code{\link{.Last.updated}},\code{\link{internal.selfref}}
 }
 \examples{
 DT = data.table(a = LETTERS[c(3L,1:3)], b = 4:7)
 DT[, c := 8]                # add a numeric column, 8 for all rows
-DT[, d := 9L]               # add an integer column, 9L for all rows
+DT[, d := 9L]               # add an integer column, 9L for all rows\code{\link{.Last.updated}}
```
aitap, line 100: This syntax doesn't work inside `\examples`. Unfortunately, the Rd format [has a lot of little details](https://developer.r-project.org/parseRd.pdf). Try rendering the help pages using `R CMD Rdconv` or installing the package and browsing its help pages before pushing commits.

NEWS.0.md
```diff
@@ -1072,6 +1072,8 @@
      query once and will never have noticed these, but anyone looping calls to grouping (such as when running in parallel, or benchmarking) may have suffered. Tests added. Thanks to many including vc273 and Y T for reporting [here](https://stackoverflow.com/questions/20349159/memory-leak-in-data-table-grouped-assignment-by-reference) and [here](https://stackoverflow.com/questions/15651515/slow-memory-leak-in-data-table-when-returning-named-lists-in-j-trying-to-reshap) on SO.
 
   2.  In long running computations where data.table is called many times repetitively the following error could sometimes occur, #2647: *"Internal error: .internal.selfref prot is not itself an extptr"*. Now fixed. Thanks to theEricStone, StevieP and JasonB for (difficult) reproducible examples [here](https://stackoverflow.com/questions/15342227/getting-a-random-internal-selfref-error-in-data-table-for-r).
+  for more info about internal.selfref  Refer to [internal.selfref](../man/internal.selfref.Rd) for additional information.
```
aitap, line 1075: Let's revert this one.

NEWS.md
```diff
@@ -744,6 +744,7 @@ rowwiseDT(
...
+Refer to [internal.selfref](../man/internal.selfref.Rd) for additional information.
```
aitap, line 747: This will also need to be reverted.

man/internal.selfref.rd
```diff
@@ -0,0 +1,37 @@
...
+}
+\value{
+  The \code{.internal.selfref} attribute is an internal implementation detail and does not produce a value that users would typically interact with. It is invisible during regular \code{data.table} operations.
+}
+\seealso{
+  \code{\link{data.table}}, \code{\link{setkey}}, \code{\link{merge}}, \code{\link{[.data.table}}
+}
+\examples{
+  library(data.table)
+
```
aitap, line 24: Let's not add `library(data.table)` to the examples. It's okay to assume that the package is already attached. Both `R CMD check` and `example()` will attach the package for us.

aitap (dismissed):
This is better, thank you for the improvements. A few more changes and then we'll ask someone with commit rights to merge.

man/assign.Rd
```diff
@@ -92,12 +92,12 @@ Since \code{[.data.table} incurs overhead to check the existence and type of arg
 \value{
 \code{DT} is modified by reference and returned invisibly. If you require a copy, take a \code{\link{copy}} first (using \code{DT2 = copy(DT)}).
 }
-\seealso{ \code{\link{data.table}}, \code{\link{copy}}, \code{\link{setalloccol}}, \code{\link{truelength}}, \code{\link{set}}, \code{\link{.Last.updated}}
+\seealso{ \code{\link{data.table}}, \code{\link{copy}}, \code{\link{setalloccol}}, \code{\link{truelength}}, \code{\link{set}}, \code{\link{.Last.updated}},\code{\link{internal.selfref}}
```
aitap, line 95: When linking to a help page, [use the `\alias{...}` name of the help page (i.e. `.internal.selfref` here), not its file name (i.e. not `internal.selfref`)](https://cran.r-project.org/doc/manuals/R-exts.html#Cross_002dreferences). All other occurrences will need to be fixed too.

man/internal.selfref.rd
aitap, line 1: For consistency with other help files, let's rename the file to `internal.selfref.Rd`, with capital R in the extension.

man/internal.selfref.rd
```diff
@@ -0,0 +1,37 @@
+\name{.internal.selfref}
+\alias{.internal.selfref}
+\title{Internal Self-Reference Attribute in data.table}
+\description{
+  The \code{.internal.selfref} attribute is an internal mechanism used by \code{data.table} to optimize memory management and performance. It acts as a pointer that allows \code{data.table} objects to reference their own memory location. While the \code{.internal.selfref} attribute may appear to always point to \code{NULL} when inspected directly, this is a result of its implementation in R's memory management system. The true significance of this attribute lies in its role in supporting reference semantics, which enables efficient in-place modification of \code{data.table} objects without unnecessary copying.
```
aitap, line 5: It's not that the external pointer _appears_ to point at `NULL`; that's [done deliberately](https://github.com/Rdatatable/data.table/blob/9875906f445eddfb2a67b6326f6ce66537dcd374/src/assign.c#L33) so that `identical()` would be `TRUE` for such attributes from two different `data.table`s. (Why is that needed? So that `identical()` would work on `data.table`s with otherwise identical contents.) The actual self-reference pointer lives in the `prot` part of the external pointer, visible using `.Internal(inspect(x))`, wrapped into another external pointer to avoid creating a reference loop visible to R. Once the `data.table` is duplicated, its address will be different from what is stored in the self-reference attribute, making it possible to detect such copies.

Try summarising the [comments and the code of `_selfrefok`](https://github.com/Rdatatable/data.table/blob/9875906f445eddfb2a67b6326f6ce66537dcd374/src/assign.c#L45-L139) to document what it actually does.

man/assign.Rd
venom1204: thanx for the guidance i will introduce all the changes and submit it till toady or tomorrow evening

venom1204:
@aitap can you now review this and plss tell me if you have any more suggestions.

MichaelChirico:
Thanks for all this!

Honestly, I am not sure it's the best idea to write this page. To me, `.internal.selfref` is really an private implementation detail of the data.table class; over-documenting it effectively turns it into a public part of the API & prevents us from making back-end changes*

Perhaps a better approach here is to completely remove mention of `.internal.selfref` from user-facing messages. We can describe what's gone wrong without anchoring on `.internal.selfref` _per se_. I reckon doing so would also make the messages more helpful as an average user seeing `.internal.selfref` is more likely to be confused by this obscurely-named object, even if `?.internal.selfref` works.

WDYT @jangorecki / @tdhock / @ben-schwen?

*Well, certainly there are already downstreams making assumptions about these internal details, but I'm more comfortable with breaking changes of undocumented code.

ben-schwen:
I agree with Michael that a manpage actually puts more burden on us in maintaining the described functionality than in helping the average user so in favor of closing #4519 and replacing the mentioning of `.internal.selfref` with more user-appropriate information.

However, I think that this implementation detail is still worth either a wiki entry or a blog post on the Raft for the interested reader (maybe as a part of a broader topic like reference semantics).

aitap:
I defer to @MichaelChirico and @ben-schwen. Sorry for making you do extra work, @venom1204.

venom1204:
@aitap No worries! I’ve gained a lot of valuable knowledge throughout this process, and I truly appreciate the guidance and insights you’ve provided. Thank you!

MichaelChirico:
Thanks for your cooperation @venom1204! Glad you found the process nevertheless insightful & hope you could learn about data.table more deeply --> more contributions in the future :)

BTW, it can't hurt to check in on an issue's discussion, especially for older issues, e.g. "I am thinking of working on this". That might save effort if we're able to raise concerns about whether our minds have changed about actually fixing a given issue :)   (the issue backlog is quite large and old, so certainly there are some issues in there that should just be closed 😢)

PR #6697, opened 2024-12-29
Growable ALTREP vectors

aitap:
Ongoing work towards #6180. Depends on #6694. Detailed motivation in a [pending blog post](https://github.com/rdatatable-community/The-Raft/blob/56ccfaad8251839d4ed17136274f67abe4c3924a/posts/2024-12-02-non-api-use/index.qmd#L247-L346).

Good news:

```
* checking compiled code ... NOTE
File ‘data.table/libs/data_table.so’:
  Found non-API call to R: ‘LEVELS’ # <-- and that's it!
```

Bad news:

```
Test 2289.2 ran without errors but failed check that x equals y:
> x = dt[, b, a] 
   a b  [Key= Types=int,lis Classes=int,lis]
1: 1 1                                      
2: 2 2                                      
> y = dt 
   a b  [Key= Types=int,exp Classes=int,exp]
1: 1 1                                      
2: 2 2                                      
Datasets have different column modes. First 3: b(list!=expression)
```

Unfortunately, since there is no `altexpression` ALTREP class, best I was able to support `EXPRSXP` columns was by creating `VECSXP` lists for their contents.

How long do we have and does anyone have a plan B?

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -19303,12 +19303,6 @@ test(2290.3, DT[, `:=`(a, c := 3)], error="It looks like you re-used `:=` in arg
 # partially-named `:=`(...) --> different branch, same error
 test(2290.4, DT[, `:=`(a = 2, c := 3)], error="It looks like you re-used `:=` in argument 2")
 
-# segfault when selfref is not ok before set #6410
-df = data.frame(a=1:3)
-setDT(df)
-attr(df, "att") = 1
-test(2291.1, set(df, NULL, "new", "new"), error="attributes .* have been reassigned")
-
 # ns-qualified bysub error, #6493
 DT = data.table(a = 1)
 test(2292.1, DT[, .N, by=base::mget("a")], data.table(a = 1, N = 1L))
--- a/src/assign.c
+++ b/src/assign.c
@@ -1,5 +1,6 @@
 #include "data.table.h"
 
+#if R_VERSION < R_Version(3,4,0) // not needed with GROWABLE_BIT
 static void finalizer(SEXP p)
 {
   SEXP x;
@@ -22,6 +23,7 @@ static void finalizer(SEXP p)
   UNPROTECT(1);
   return;
 }
+#endif
 
 void setselfref(SEXP x) {
   if(!INHERITS(x, char_datatable))  return; // #5286
@@ -38,7 +40,9 @@ void setselfref(SEXP x) {
       R_NilValue
     ))
   ));
+#if R_VERSION < R_Version(3,4,0) // not needed with GROWABLE_BIT
   R_RegisterCFinalizerEx(p, finalizer, FALSE);
+#endif
   UNPROTECT(2);
 
 /*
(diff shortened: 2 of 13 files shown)


ben-schwen:
We did not support expressions prior to #5596 and still not fully it support. My guess is that nobody really uses this anyway

src/assign.c
```diff
@@ -514,7 +520,7 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
   // modify DT by reference. Other than if new columns are being added and the allocVec() fails with
   // out-of-memory. In that case the user will receive hard halt and know to rerun.
   if (length(newcolnames)) {
-    oldtncol = TRUELENGTH(dt);   // TO DO: oldtncol can be just called tl now, as we won't realloc here any more.
+    oldtncol = is_growable(dt) ? growable_max_size(dt) : 0;   // TO DO: oldtncol can be just called tl now, as we won't realloc here any more.
```
ben-schwen, line 523:
```suggestion
    oldtncol = truelength(x);   // TO DO: oldtncol can be just called tl now, as we won't realloc here any more.
```

src/assign.c
```diff
@@ -527,13 +533,13 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
       error(_("It appears that at some earlier point, names of this data.table have been reassigned. Please ensure to use setnames() rather than names<- or colnames<-. Otherwise, please report to data.table issue tracker."));  // # nocov
       // Can growVector at this point easily enough, but it shouldn't happen in first place so leave it as
       // strong error message for now.
-    else if (TRUELENGTH(names) != oldtncol)
+    else if (growable_max_size(names) != oldtncol)
```
ben-schwen, line 536: Maybe also use `truelength(names)` here?

inst/tests/tests.Rraw
```diff
@@ -19303,12 +19303,6 @@ test(2290.3, DT[, `:=`(a, c := 3)], error="It looks like you re-used `:=` in arg
 # partially-named `:=`(...) --> different branch, same error
 test(2290.4, DT[, `:=`(a = 2, c := 3)], error="It looks like you re-used `:=` in argument 2")
 
-# segfault when selfref is not ok before set #6410
-df = data.frame(a=1:3)
-setDT(df)
-attr(df, "att") = 1
-test(2291.1, set(df, NULL, "new", "new"), error="attributes .* have been reassigned")
-
```
ben-schwen, line 19311: So this will only bite us on R < `4.3.0` and will be detected earlier on the ALTREP version?

src/growable.c
```diff
@@ -0,0 +1,276 @@
...
+#if R_VERSION >= R_Version(3, 4, 0)
+    && IS_GROWABLE(x)
+#endif
+  ;
+}
+
+// Assuming no ALTREP columns
+SEXP make_growable(SEXP x) {
+  if (TRUELENGTH(x) < XLENGTH(x)) SET_TRUELENGTH(x, XLENGTH(x));
+  SET_GROWABLE_BIT(x);
```
ben-schwen, line 40:
```suggestion
#if R_VERSION >= R_Version(3, 4, 0)
  SET_GROWABLE_BIT(ret);
#endif //
```

src/growable.c
```diff
@@ -0,0 +1,276 @@
...
+  dta_grow_list = R_make_altlist_class("growable_list_v0", "data.table", info);
+  R_set_altrep_Length_method(dta_grow_list, altall_Length);
+  R_set_altrep_Inspect_method(dta_grow_list, altlist_Inspect);
+  R_set_altvec_Dataptr_method(dta_grow_list, altlist_Dataptr);
+  R_set_altvec_Dataptr_or_null_method(dta_grow_list, altall_Dataptr_or_null);
+  R_set_altlist_Elt_method(dta_grow_list, altlist_Elt);
+  R_set_altlist_Set_elt_method(dta_grow_list, altlist_Set_elt);
+}
+
+static R_altrep_class_t dta_grow_string, dta_grow_integer, dta_grow_logical, dta_grow_real, dta_grow_complex, dta_grow_raw, dta_grow_list;
```
ben-schwen, line 207: We already have the same line in line 48?

aitap: Good catch, thanks!

src/growable.c
aitap: We've got a dummy `#define` here:
https://github.com/Rdatatable/data.table/blob/9fe1b8d01701d2b23adcff3085ea0290da5df022/src/data.table.h#L17-L19

...but it may be cleaner to check the version at the point of use. Which may become moot with the move to R > 3.4.

inst/tests/tests.Rraw
aitap: By setting `GROWABLE_BIT` on R ≥ 3.4.0 for `data.table` lists, we prevent the situation where a shallow duplicate of a `data.table` has a misleading `TRUELENGTH`. When duplicating a _growable_ list, R will notice the bit and reset `TRUELENGTH` to 0. This causes `set` to fail an earlier check for non-zero `TRUELENGTH` instead of failing the `.internal.selfref` check, so the error is different ("This data.table has either been loaded from disk...").

Perhaps the test should account for that, especially if we're moving from R-3.3 and thus will no longer hit the old behaviour (where "attributes .* have been reassigned" is still reachable).

src/growable.c
```diff
@@ -0,0 +1,274 @@
...
+
+static const void * altall_Dataptr_or_null(SEXP x) { return DATAPTR_RO(x); }
+
+// lots of boilerplate, but R calling *_ELT one by one would be far too slow
+#define make_extract_subset_method(class, type, accessor, NA) \
+  static SEXP alt##class##_Extract_subset(SEXP x, SEXP indx, SEXP call) { \
+    (void)call; \
+    indx = PROTECT(coerceVector(indx, REALSXP)); \
+    double * ii = REAL(indx); \
+    R_xlen_t rlen = XLENGTH(indx), mylen = XLENGTH(x); \
```
ben-schwen, line 103: maybe I'm misunderstanding extract_subset but shouldn't this only be valid for the actual size and not the full overallocated length?

Hence, `mylen = (R_xlen_t)REAL(R_altrep_data2(x))[0];` ?

src/growable.c
ben-schwen: I see, TY!

src/growable.c
aitap: `XLENGTH(x)` goes through the ALTREP dispatch and lands in `altall_Length`, but we could definitely save a function call and make the code clearer.

Issue #6671, opened 2024-12-19
revdep dbi.table broken by with=FALSE


tdhock:
revdep check machine
says #6487 broke `dbi.table` tests:
```
* checking tests ...
  Running 'testthat.R'
 ERROR
Running the tests in 'tests/testthat.R' failed.
Complete output:
  > # This file is part of the standard setup for testthat.
  > # It is recommended that you do not modify it.
  > #
  > # Where should you do additional test configuration?
  > # Learn more about the roles of various files in:
  > # * https://r-pkgs.org/testing-design.html#sec-tests-files-overview
  > # * https://testthat.r-lib.org/articles/special-files.html
  > 
  > library(testthat)
  > library(dbi.table)
  Loading required package: data.table
  > 
  > test_check("dbi.table")
  [ FAIL 2 | WARN 332 | SKIP 0 | PASS 264 ]
  
  == Failed tests ================================================================
  -- Failure ('test-dbi.table.R:77:5'): j syntax consistent with data.table ------
  reference.test(Track[, 1:MediaTypeId], verbose = FALSE) is not TRUE
  
  `actual`:   FALSE
  `expected`: TRUE 
  -- Failure ('test-dbi.table.R:77:5'): j syntax consistent with data.table ------
  reference.test(Track[, 1:MediaTypeId], verbose = FALSE) is not TRUE
  
  `actual`:   FALSE
  `expected`: TRUE 
  
  [ FAIL 2 | WARN 332 | SKIP 0 | PASS 264 ]
  Error: Test failures
  Execution halted
* checking for unstated dependencies in vignettes ... OK
```
@MichaelChirico @jangorecki you commented on that PR so can you please investigate?

MichaelChirico:
OK, here's a reprex:

```r
DT = data.table(a=1, b=2:3)
DT[, 1:b]

## master
# [1] 1 2
# Warning message:
# In 1:b : numerical expression has 2 elements: only the first used

## 1.16.4
#        a     b
#    <num> <num>
# 1:     1     2
# 2:     1     3
```

I think `with=FALSE` should be retained in this case, i.e. we should fix the regression.

tdhock:
fixed using 6ad0524f

PR #6700, opened 2025-01-02
Retain with=FALSE in j=lit:var cases

MichaelChirico:
For #6671 (not closing because we should wait for the revdep checker to confirm the fix).

This also fixes a mistake where `is.name(e[[1]])` was used. `e[[1]]` is always `:` in this case, and therefore is always a `name`. `e[[2]]` and `e[[3]]` are the arguments to `:`, which is what we really want to inspect.

The added tests help cover the mistaken logic there too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -95,7 +95,7 @@ rowwiseDT(
     # [1] "V1" "b" "c"
     ```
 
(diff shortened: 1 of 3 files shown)


tdhock:
lgtm thanks

PR #6710, opened 2025-01-06
Add a devcontainer on alpine

MichaelChirico:
This is for #6350.

After the investigation, I am OK kicking the can on this one a bit further. I wanted to investigate since on first glance, `forder()` being length-0 looked more concerning. But in actuality, it just means "input is sorted", which is the same as `base::order()` finds, hence not show-stopping -- this is pretty deeply edge-case behavior anyway.

I don't know enough about how `forder()` handles strings (especially w.r.t. encoding); AFAICT this will be a lot easier to debug after #6669 (since we can then use print-debugging to see how different parts of the routine give different results on different platforms).


ben-schwen (approved):
LGTM

Issue #6556, opened 2024-10-04
[feature request] diagnostic for merge.data.table when by = key is not present in dt being merged


zhizhongpu:
`#` [`Minimal reproducible example`](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example); please be sure to set `verbose=TRUE` where possible!

When merging two dt using merge.data.table and the key is not present in one of the two dts, data.table throws out an uninformative error:

```
Error in merge.data.table(test, manual, by = c("iso3c", "year"), all = T) : 
  Elements listed in `by` must be valid column names in x and y
```
which is not helpful as the user would then have to debug which key is not present in which dt. A more informative error would list all keys not present in all dts. For example:
```
> combined = test |> 
+   left_join(manual, by = c('iso3c', 'year')) 
Error in `left_join()`:
! Join columns in `y` must be present in the
  data.
✖ Problem with `iso3c`.
```

`#` `Output of sessionInfo()`

MichaelChirico:
Great idea, thanks! Here's the relevant error:

https://github.com/Rdatatable/data.table/blob/b4538a0a1fb6dbd6453a00bc45bc90eb3ada3f19/R/merge.R#L53-L54

Do you want to have a go at writing a PR?

PR #6713, opened 2025-01-08
Improve merge.data.table error messages for missing keys

venom1204:
In this i have made the error popping out to be more informative by two things

Which column(s) are missing.
Which data.table is missing the column(s).
Closes #6556


venom1204:
Hi @aitap,
I hope you’re doing well! I’m currently facing some challenges with the atime performance test, and I could use your guidance.
Sorry for the disturbance, but if you have a moment, could you please help me figure out how to resolve this? I’d really appreciate your insights.
Thank you so much for your support!

venom1204:
Hi @MichaelChirico,
Apologies for the interruption, but could you kindly take a look at the issue with the atime performance tests? I’m running into an error, and it seems to be related to how the tests are triggered for pull requests from external forks.
If you could provide any guidance on how to resolve this, it would be greatly appreciated. Your help would mean a lot as I navigate this issue!

venom1204:
@MichaelChirico thanks for the clarification 
can you please review the changes in the pr.

inst/tests/tests.Rraw
```diff
@@ -13520,14 +13528,28 @@ test(1962.016, merge(DT1, DT2, by.x = 'a', by.y = c('a', 'V')),
...
+  merge(DT1, DT2, by.x = 'a', by.y = 'z')
+}, error = 'Elements listed in `by.y` are missing from y: z')
+
 test(1962.0201, merge(DT1, DT2, by=character(0L)), ans)  # was error before PR#5183
 test(1962.0202, merge(DT1, DT2, by=NULL),          ans)  # test explicit NULL too as missing() could be used inside merge()
-test(1962.021, merge(DT1, DT2, by = 'z'),
-     error = 'must be valid column names in x and y')
+test(1962.021, {
+  if (!"z" %in% colnames(DT1) || !"z" %in% colnames(DT2)) {
+    stop("The columns listed in `by` are missing from either x or y: z")
```
tdhock, line 13549: please remove backticks, which are for markdown, not error messages

inst/tests/tests.Rraw
```diff
@@ -8563,16 +8563,24 @@ test(1600.2, names(DT1[DT2, .(id1=id1, val=val, bla=sum(z1, na.rm=TRUE)), on="id
...
 test(1601.3, merge(DT0, DT1, by="a"),
      warning="Input data.table 'x' has no columns.",
-     error="Elements listed in `by`")
+     error="The following columns are missing:\n - From `x`: a")
+
+# Test 1601.4: Merge DT0 with DT0 on column 'a'
 test(1601.4, merge(DT0, DT0, by="a"),
      warning="Neither of the input data.tables to join have columns.",
-     error="Elements listed in `by`")
+     error="The following columns are missing:\n - From `x`: a\n - From `y`: a")
```
tdhock, line 8583: please remove newlines in error messages

R/merge.R
```diff
@@ -35,10 +35,16 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
     if (length(by.x)==0L || !is.character(by.x) || !is.character(by.y))
       stopf("A non-empty vector of column names is required for `by.x` and `by.y`.")
-    if (!all(by.x %chin% nm_x))
-      stopf("Elements listed in `by.x` must be valid column names in x.")
-    if (!all(by.y %chin% nm_y))
-      stopf("Elements listed in `by.y` must be valid column names in y.")
+    if (!all(by.x %chin% nm_x)) {
+      missing_in_x <- setdiff(by.x, nm_x)
+      stopf("The following columns listed in `by.x` are missing from `x`: %s",
+            toString(missing_in_x))
```
tdhock, line 41: brackify instead of toString?

R/merge.R
```diff
@@ -50,8 +56,13 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
     if (length(by) == 0L || !is.character(by))
       stopf("A non-empty vector of column names for `by` is required.")
-    if (!all(by %chin% intersect(nm_x, nm_y)))
-      stopf("Elements listed in `by` must be valid column names in x and y")
+    missing_in_x <- setdiff(by, nm_x)
+    missing_in_y <- setdiff(by, nm_y)
+    if (length(missing_in_x) > 0 || length(missing_in_y) > 0) {
+      stopf("The following columns are missing:\n%s%s",
+            if (length(missing_in_x) > 0) sprintf(" - From `x`: %s\n", toString(missing_in_x)) else "",
+            if (length(missing_in_y) > 0) sprintf(" - From `y`: %s\n", toString(missing_in_y)) else "")
```
tdhock, line 64: please refactor to remove repetition

and also please remove line breaks inside function calls: use `stopf(some, code)` instead of 
```r
stopf(some,
  code)
```

tdhock (changes requested):

tdhock:
I added documentation which explains that atime failure is normal in forks, https://github.com/Rdatatable/data.table/wiki/Performance-testing#can-not-be-run-from-forks

> If this is not by design (e.g. performance tests are relatively expensive and therefore must be run only from local branches),

I believe the problem is not the "relatively expensive part" but rather that the action requires permission to upload artifacts to rdatatable/data.table repo.

inst/tests/tests.Rraw
```diff
@@ -13520,14 +13528,28 @@ test(1962.016, merge(DT1, DT2, by.x = 'a', by.y = c('a', 'V')),
...
      data.table(a = 2:3, V.x = c("a", "a"), V.y = c("b", "b"), key = 'a'),
      warning = 'Supplied both.*argument will be ignored')
-test(1962.018, merge(DT1, DT2, by.x = 'z', by.y = 'a'),
-     error = 'Elements listed in `by.x`')
-test(1962.019, merge(DT1, DT2, by.x = 'a', by.y = 'z'),
-     error = 'Elements listed in `by.y`')
+test(1962.018, {
+  if (!"z" %in% colnames(DT1)) {
+    stop("Elements listed in `by.x` are missing from x: z")
+  }
```
aitap, line 13534: This is a creative solution to the problem of the failing test (would you mind letting us know how you came up with it?), but, unfortunately, not the right one.

The idea here and below is to test the error raised by the following `merge()` call, not to manually raise the error expected by the previous test code. Instead of calling `stop()` in the test expression, set the `error` argument to make sure it matches.

R/merge.R
venom1204: The brackify function adds brackets around column names in error messages, which may not align with the expected format in your test cases.should i change the format of the test case ?

inst/tests/tests.Rraw
venom1204: Hi @aitap,
Thank you for the feedback. I apologize for the confusion. You are absolutely right that the goal is to test the error raised by the merge() function call, not to manually trigger the error with a stop() call in the test expression.
I had initially added the stop() to handle the missing column case manually as a quick fix, I was about to make this change but missed it during my initial implementation.

R/merge.R
venom1204: @tdhock I've implemented the changes you suggested. Please review them and let me know if there's anything else that needs to be addressed or improved.

R/merge.R
```diff
@@ -17,9 +17,9 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
     if (x0 && y0)
       warningf("Neither of the input data.tables to join have columns.")
     else if (x0)
-      warningf("Input data.table '%s' has no columns.", "x")
+      warningf("Input data.table x has no columns.")
```
MichaelChirico, line 20: Why this change? This code is structured intentionally so that there are not two nearly-but-not-quite identical messages for translators.

R/merge.R
```diff
@@ -29,16 +29,22 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
 
   ## set up 'by'/'by.x'/'by.y'
   if ( (!is.null(by.x) || !is.null(by.y)) && length(by.x)!=length(by.y) )
-    stopf("`by.x` and `by.y` must be of same length.")
+    stopf("by.x and by.y must be of same length.")
```
MichaelChirico, line 32: why this change? backticks serve to highlight that these are code objects, not plain English terms.

inst/tests/tests.Rraw
MichaelChirico: Hi Toby, sorry, I disagree.

The backticks serve to highlight that this is a code object, and not a plain English word. Without them, a reader can easily be confused into thinking there's some grammatical mistake "in by", or otherwise struggle to parse the message they're given.

Of course, we could choose some other convention (single/double quotes, e.g.), and we should try and pick one and stick to it throughout the codebase... but that's a separate issue.

Personally, these days I am using `` `arg=` `` for function arguments to highlight that (1) it's code with the backticks and (2) it's a keyword argument with `=`.

R/merge.R
MichaelChirico: Sorry, I see this was requested elsewhere. Let's continue discussion there: https://github.com/Rdatatable/data.table/pull/6713#discussion_r1912605078

R/merge.R
MichaelChirico: Yes, use brackify. It provides nice formatting and also some simple truncation mechanism in case `missing_in_x` happens to have 10s or dozens of elements.

R/merge.R
```diff
@@ -49,9 +55,14 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
       by = intersect(nm_x, nm_y)
     if (length(by) == 0L || !is.character(by))
-      stopf("A non-empty vector of column names for `by` is required.")
-    if (!all(by %chin% intersect(nm_x, nm_y)))
-      stopf("Elements listed in `by` must be valid column names in x and y")
+      stopf("A non-empty vector of column names for by is required.")
+    missing_in_x <- setdiff(by, nm_x)
+    missing_in_y <- setdiff(by, nm_y)
+    if (length(missing_in_x) > 0 || length(missing_in_y) > 0) {
+      stopf("The following columns are missing:%s%s",
```
MichaelChirico, line 62: This looks to be missing newlines.

R/merge.R
```diff
@@ -49,9 +55,14 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+            if (length(missing_in_x) > 0) sprintf(" - From x: %s", toString(missing_in_x)) else "",
```
MichaelChirico, line 63: Use `gettextf()` to enable translation

inst/tests/tests.Rraw
MichaelChirico: Remove, or add? I find the current output hard to read:

```
The following columns are missing: - From x: a
```

I would find this much more readable (possibly indenting the second line):

```
The following columns are missing:
- From x: a
```

At a higher level, I wonder if translation would be easier if we instead structured the message like so:

```
The following columns are missing from x: ...
The following columns are missing from y: ...
```

inst/tests/tests.Rraw
```diff
@@ -8564,16 +8564,24 @@ test(1600.2, names(DT1[DT2, .(id1=id1, val=val, bla=sum(z1, na.rm=TRUE)), on="id
 # warn when merge empty data.table #597
 DT0 = data.table(NULL)
 DT1 = data.table(a=1)
+
+# Test 1601.1: Merge DT1 with itself on column 'a'
```
MichaelChirico, line 8568: You shouldn't need comments like this unless the test is really obscure. You definitely don't need to write 'Test 1601.1' when that is very obvious from the next line.

Such comments are very easy to fall out of sync with the actual code as it changes over time. See for example https://swimm.io/learn/code-collaboration/comments-in-code-best-practices-and-mistakes-to-avoid.

If the test case's purpose is not obvious from the written code, often that's a sign that the test is poorly designed -- typically we should strive for the purpose of the test to be immediately apparent, only rarely needing small clarifying comments.

venom1204:
Hi @MichaelChirico and @tdhock,
I have implemented all the changes you suggested. Could you please review the updates and let me know if there's anything else I can improve? 
Thank you!

inst/tests/tests.Rraw
```diff
@@ -13542,14 +13545,20 @@ test(1962.016, merge(DT1, DT2, by.x = 'a', by.y = c('a', 'V')),
...
+     merge(DT1, DT2, by.x = 'a', by.y = 'z'), 
+     error = "The following columns listed in 'by.y' are missing from y: [z]")
 test(1962.0201, merge(DT1, DT2, by=character(0L)), ans)  # was error before PR#5183
 test(1962.0202, merge(DT1, DT2, by=NULL),          ans)  # test explicit NULL too as missing() could be used inside merge()
-test(1962.021, merge(DT1, DT2, by = 'z'),
-     error = 'must be valid column names in x and y')
+test(1962.021, {
+  if (!"z" %in% colnames(DT1) || !"z" %in% colnames(DT2)) {
+    stop("The columns listed in `by` are missing from either x or y: [z]")
+  }
```
aitap, line 13559: One last occurrence of a test raising the error and then testing for it instead of letting `merge()` do that.

venom1204: Sorry by mistake I missed it 
Now I have made changes .

MichaelChirico:
@venom1204 would you mind editing the PR title to be more self-contained?

> made the error more informative for #6556

Ideally we can glance now and far into the future at the title of a PR and have a good idea what part of the codebase it's about. See for example https://blog.montrealanalytics.com/4-tips-for-effective-pull-request-naming-f60793998f04.

Especially as `#6556` does not get auto-linked to the issue, it is not very useful to include in the title (a bit unfortunately, I do wish GitHub would support links there).

venom1204:
> @venom1204 would you mind editing the PR title to be more self-contained?
> 
> > made the error more informative for #6556
> 
> Ideally we can glance now and far into the future at the title of a PR and have a good idea what part of the codebase it's about. See for example https://blog.montrealanalytics.com/4-tips-for-effective-pull-request-naming-f60793998f04.
> 
> Especially as `#6556` does not get auto-linked to the issue, it is not very useful to include in the title (a bit unfortunately, I do wish GitHub would support links there).

venom1204:
@MichaelChirico i changed the title of the pr.

R/merge.R
```diff
@@ -11,36 +11,41 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
   }
-  check_duplicate_names(x)
-  check_duplicate_names(y)
+check_duplicate_names(x)
+check_duplicate_names(y)
 
-  nm_x = names(x)
-  nm_y = names(y)
+nm_x = names(x)
+nm_y = names(y)
```
aitap, line 29: Indentation is important. While R cares very little about the spaces at the beginning of the lines (or about spaces between the keywords in general), the people reading the code find it easier when the lines [are aligned](https://en.wikipedia.org/wiki/Indentation_style) according to the `function` / `if` / `while` blocks they belong to. In `data.table` we increase the indentation by two spaces every time a new `{` block begins and decrease it again when it ends. There are other styles out there, but this one is what we're using.

Removing those spaces above makes it look as if the line `nm_x = names(x)` is outside the enclosing `function` block, which is makes it misleading. Commits that change the spaces (or, in general, the looks) in the code without changing the substance (i.e. what actually gets run) are also detrimental for a less obvious reason. We store the change history in Git, more than 5000 changes dating back to 2008. This lets us quickly find the sources of problems using the power of [bisect](https://git-scm.com/docs/git-bisect). If a change is small and only touches the code, it's easy to understand what's broken. When the change also moves around the text without changing its meaning, the challenge becomes harder.

While some whitespace changes can be good because they make the code more readable, we should definitely not break the indentation.

venom1204:
hi @MichaelChirico any updates on this ?

inst/tests/tests.Rraw
aitap: The longer sentence would definitely be easier to translate.

Vinit-Thakur-v1:
> hi @MichaelChirico @tdhock any updates on this ?
@venom1204 Some checks were not successful !! try to fix this

venom1204:
> > hi @MichaelChirico @tdhock any updates on this ?
> > @venom1204 Some checks were not successful !! try to fix this

hi @Vinit-Thakur-v1 they told me to ignore that test failure.

R/merge.R
```diff
@@ -28,34 +29,42 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
   nm_y = names(y)
 
   ## set up 'by'/'by.x'/'by.y'
-  if ( (!is.null(by.x) || !is.null(by.y)) && length(by.x)!=length(by.y) )
-    stopf("`by.x` and `by.y` must be of same length.")
+  if ((!is.null(by.x) || !is.null(by.y)) && length(by.x) != length(by.y))
+    stopf("'by.x' and 'by.y' must be of same length.")
```
MichaelChirico, line 33:
```suggestion
    stopf("`by.x` and `by.y` must be of same length.")
```

R/merge.R
```diff
@@ -28,34 +29,42 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
   if (!missing(by) && !missing(by.x))
-    warningf("Supplied both `by` and `by.x/by.y`. `by` argument will be ignored.")
+    warningf("Supplied both by and 'by.x/by.y.' by argument will be ignored.")
```
MichaelChirico, line 35:
```suggestion
    warningf("Supplied both `by` and `by.x`/`by.y`. `by` argument will be ignored.")
```

R/merge.R
```diff
@@ -28,34 +29,42 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
-      stopf("Elements listed in `by.y` must be valid column names in y.")
+    if (length(by.x) == 0L || !is.character(by.x) || !is.character(by.y))
+      stopf("A non-empty vector of column names is required for 'by.x' and 'by.y'.")
+    if (!all(by.x %chin% nm_x)) {
+      missing_in_x <- setdiff(by.x, nm_x)
+      stopf("The following columns listed in 'by.x' are missing from x: %s", brackify(missing_in_x))
+    }
+    if (!all(by.y %chin% nm_y)) {
+      missing_in_y <- setdiff(by.y, nm_y)
+      stopf("The following columns listed in 'by.y' are missing from y: %s", brackify(missing_in_y))
```
MichaelChirico, line 45:
```suggestion
      stopf("The following columns listed in `by.y` are missing from y: %s", brackify(missing_in_y))
```

R/merge.R
```diff
@@ -28,34 +29,42 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    }
     by = by.x
     names(by) = by.y
   } else {
     if (is.null(by))
       by = intersect(key(x), key(y))
-    if (!length(by))   # was is.null() before PR#5183  changed to !length()
+    if (!length(by)) # was is.null() before PR#5183 changed to !length()
```
MichaelChirico, line 52:
```suggestion
    if (!length(by))   # was is.null() before PR#5183  changed to !length()
```

R/merge.R
```diff
@@ -109,7 +118,7 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
   }
 
   # Throw warning if there are duplicate column names in 'dt' (i.e. if
-  # `suffixes=c("","")`, to match behaviour in base:::merge.data.frame)
+  # `suffixes=c("",""), to match behaviour in base:::merge.data.frame)
```
MichaelChirico, line 121:
```suggestion
  # `suffixes=c("","")`, to match behaviour in base:::merge.data.frame)
```

R/merge.R
```diff
@@ -11,15 +11,16 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+  x0 = length(x) == 0L
+  y0 = length(y) == 0L
+  if (x0 || y0){
+    if (x0 && y0) {
       warningf("Neither of the input data.tables to join have columns.")
-    else if (x0)
+    } else if (x0) {
       warningf("Input data.table '%s' has no columns.", "x")
-    else
+    } else if (y0) {
```
MichaelChirico, line 21:
```suggestion
    } else {
```

R/merge.R
```diff
@@ -28,17 +29,21 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    warningf("Supplied both `by` and `by.x`/`by.y`. `by` argument will be ignored.")
   if (!is.null(by.x)) {
-    if (length(by.x)==0L || !is.character(by.x) || !is.character(by.y))
-      stopf("A non-empty vector of column names is required for `by.x` and `by.y`.")
-    if (!all(by.x %chin% nm_x))
-      stopf("Elements listed in `by.x` must be valid column names in x.")
-    if (!all(by.y %chin% nm_y))
-      stopf("Elements listed in `by.y` must be valid column names in y.")
+    if (length(by.x) == 0L || !is.character(by.x) || !is.character(by.y))
+      stopf("A non-empty vector of column names is required for 'by.x' and 'by.y'.")
```
MichaelChirico, line 38:
```suggestion
      stopf("A non-empty vector of column names is required for `by.x` and `by.y`.")
```

R/merge.R
```diff
@@ -28,17 +29,21 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    if (!all(by.x %chin% nm_x)) {
+      missing_in_x <- setdiff(by.x, nm_x)
+      stopf("The following columns listed in `by.x` are missing from x: %s", brackify(missing_in_x))
```
MichaelChirico, line 41:
```suggestion
      stopf("The following columns listed in `%s` are missing from %s: %s", "by.x", "x", brackify(missing_in_x))
```

R/merge.R
```diff
@@ -28,17 +29,21 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    }
+    if (!all(by.y %chin% nm_y)) {
+      missing_in_y <- setdiff(by.y, nm_y)
+      stopf("The following columns listed in `by.y` are missing from y: %s", brackify(missing_in_y))
```
MichaelChirico, line 45:
```suggestion
      stopf("The following columns listed in `%s` are missing from %s: %s", "by.y", "y", brackify(missing_in_y))
```

R/merge.R
```diff
@@ -49,13 +54,17 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
     if (!length(by))
       by = intersect(nm_x, nm_y)
     if (length(by) == 0L || !is.character(by))
-      stopf("A non-empty vector of column names for `by` is required.")
-    if (!all(by %chin% intersect(nm_x, nm_y)))
-      stopf("Elements listed in `by` must be valid column names in x and y")
+      stopf("A non-empty vector of column names for by is required.")
```
MichaelChirico, line 57:
```suggestion
      stopf("A non-empty vector of column names for `by` is required.")
```

R/merge.R
```diff
@@ -49,13 +54,17 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    missing_in_x <- setdiff(by, nm_x)
+    missing_in_y <- setdiff(by, nm_y)
+    if (length(missing_in_x) > 0 || length(missing_in_y) > 0) {
+      stopf(gettextf("The following columns are missing:\n%s%s",
+        if (length(missing_in_x) > 0) gettextf(" - From x: %s\n", brackify(missing_in_x)) else "",
+        if (length(missing_in_y) > 0) gettextf(" - From y: %s\n", brackify(missing_in_y)) else ""))
```
MichaelChirico, line 63:
```suggestion
      stopf("The following columns are missing:\n%s%s",
        if (length(missing_in_x) > 0L) gettextf(" - From x: %s\n", brackify(missing_in_x)) else "",
        if (length(missing_in_y) > 0L) gettextf(" - From y: %s\n", brackify(missing_in_y)) else "")
```

R/merge.R
```diff
@@ -49,13 +54,17 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
     if (!length(by))
       by = intersect(nm_x, nm_y)
     if (length(by) == 0L || !is.character(by))
-      stopf("A non-empty vector of column names for `by` is required.")
-    if (!all(by %chin% intersect(nm_x, nm_y)))
-      stopf("Elements listed in `by` must be valid column names in x and y")
+      stopf("A non-empty vector of column names for by is required.")
+    missing_in_x <- setdiff(by, nm_x)
+    missing_in_y <- setdiff(by, nm_y)
+    if (length(missing_in_x) > 0 || length(missing_in_y) > 0) {
```
MichaelChirico, line 60:
```suggestion
    if (length(missing_in_x) > 0L || length(missing_in_y) > 0L) {
```

R/merge.R
```diff
@@ -50,12 +55,16 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
       stopf("A non-empty vector of column names for `by` is required.")
-    if (!all(by %chin% intersect(nm_x, nm_y)))
-      stopf("Elements listed in `by` must be valid column names in x and y")
+    missing_in_x <- setdiff(by, nm_x)
+    missing_in_y <- setdiff(by, nm_y)
+    if (length(missing_in_x) > 0L || length(missing_in_y) > 0L) {
+      stopf("The following columns are missing:\n%s%s",
+        if (length(missing_in_x) > 0L) gettextf(" - From x: %s\n", brackify(missing_in_x)) else "",
+        if (length(missing_in_y) > 0L) gettextf(" - From y: %s\n", brackify(missing_in_y)) else "")
+    }
```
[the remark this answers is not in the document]
MichaelChirico: I think this is closer to the earlier code by checking `x` then `y`, and allows us to re-use the same message template 2 more times.

MichaelChirico:
Thank you for the PR! Please be sure to take a look at the final touch-up edits I've just made. A lot of it is minor points, but it does make review a lot easier if everything is "neat as a pin".

R/merge.R
MichaelChirico: To clarify, I meant to swap `sprintf()` for `gettextf()`. `stopf()` is already equipped to run `gettext()`:

https://github.com/Rdatatable/data.table/blob/2714cb43262bc4dc2c25810030175db8727964f1/R/translation.R#L20-L22

MichaelChirico (approved):

R/merge.R
```diff
@@ -11,15 +11,16 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
-  y0 = length(y)==0L
-  if (x0 || y0) {
-    if (x0 && y0)
+  x0 = length(x) == 0L
+  y0 = length(y) == 0L
+  if (x0 || y0){
+    if (x0 && y0) {
       warningf("Neither of the input data.tables to join have columns.")
-    else if (x0)
+    } else if (x0) {
```
MichaelChirico, line 19:
```suggestion
    else if (x0)
```

R/merge.R
```diff
@@ -11,15 +11,16 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
       warningf("Input data.table '%s' has no columns.", "x")
-    else
+    } else {
```
MichaelChirico, line 21:
```suggestion
    else
```

R/merge.R
```diff
@@ -11,15 +11,16 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
       warningf("Input data.table '%s' has no columns.", "y")
+    }
```
MichaelChirico, line 23:
```suggestion
```

R/merge.R
```diff
@@ -11,10 +11,10 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
     }
   }
-  x0 = length(x)==0L
-  y0 = length(y)==0L
-  if (x0 || y0) {
-    if (x0 && y0)
+  x0 = length(x) == 0L
+  y0 = length(y) == 0L
+  if (x0 || y0){
+    if (x0 && y0) {
```
MichaelChirico, line 17:
```suggestion
    if (x0 && y0)
```

R/merge.R
```diff
@@ -11,9 +11,9 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
       by = key(x)
     }
   }
-  x0 = length(x)==0L
-  y0 = length(y)==0L
-  if (x0 || y0) {
+  x0 = length(x) == 0L
+  y0 = length(y) == 0L
+  if (x0 || y0){
```
MichaelChirico, line 16:
```suggestion
  if (x0 || y0) {
```

venom1204:
@MichaelChirico 
Thank you for reviewing the PR and making those touch-up edits—I sincerely apologize for not having everything as polished as it should have been. I realize this stems from the coding style I’ve been accustomed to since I started coding, but I understand now how important it is to maintain a cleaner, more consistent style to make the review process smoother. I’ll be more mindful of these details moving forward to prevent causing unnecessary effort on your part. Thank you for your patience and understanding!

PR #6724, opened 2025-01-16
Need setup-r step for lint-md

MichaelChirico:
This should fix the failure noted recently. I'm not sure how this ever worked TBH.


MichaelChirico:
Merging since this un-breaks CI and is otherwise trivial.

PR #6727, opened 2025-01-16
Adapt to changed RHOME in runner for pkgup GHA job

MichaelChirico:
This GHA is failing on `master`:

https://github.com/Rdatatable/data.table/actions/runs/12817202760

Based on the error it looks like `R.home()` has changed on these machines. Instead we should prefer the much more robust approach taken in this PR. Based on documentation:

https://stat.ethz.ch/R-manual/R-devel/library/base/html/Rhome.html

### DIFF
--- a/.github/workflows/pkgup.yaml
+++ b/.github/workflows/pkgup.yaml
@@ -9,8 +9,8 @@ concurrency:
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


ben-schwen:
LGTM TY

Issue #6729, opened 2025-01-17
fread signed integer overflow on gcc-UBSAN and clang-UBSAN checks


TaikiSan21:
My package on CRAN (PAMscapes) recently got flagged because there is an error triggered in both the gcc-UBSAN and clang-UBSAN additional CRAN checks. Here is a [link](https://cran.r-project.org/web/checks/check_results_PAMscapes.html) to the check page, here is some version info from the check log:
* using R Under development (unstable) (2025-01-15 r87580)
* using platform: x86_64-pc-linux-gnu
* R was compiled by
    clang version 19.1.6
    flang-new version 19.1.6
* running under: Fedora Linux 36 (Workstation Edition)
* using session charset: UTF-8
* using option ‘--no-stop-on-test-error’

And here is  copy of the error message from the Ex-out:

fread.c:1919:30: runtime error: signed integer overflow: 237281 * 237281 cannot be represented in type 'int'
    #0 0x7fee3c6f9969 in freadMain /tmp/Rtmp8p0OKk/R.INSTALLecbd63f70502d/data.table/src/fread.c:1919:30
    #1 0x7fee3c70e547 in freadR /tmp/Rtmp8p0OKk/R.INSTALLecbd63f70502d/data.table/src/freadR.c:218:3
    #2 0x55b4d6c2342b in R_doDotCall (/data/gannet/ripley/R/R-clang/bin/exec/R+0xf342b)
    #3 0x55b4d6c23d1d in do_dotcall (/data/gannet/ripley/R/R-clang/bin/exec/R+0xf3d1d)
    #4 0x55b4d6c5ffe2 in bcEval_loop eval.c
    #5 0x55b4d6c5971b in bcEval eval.c
    #6 0x55b4d6c58ea4 in Rf_eval (/data/gannet/ripley/R/R-clang/bin/exec/R+0x128ea4)
    #7 0x55b4d6c719c1 in R_execClosure eval.c
    #8 0x55b4d6c70eab in applyClosure_core eval.c
    #9 0x55b4d6c592f5 in Rf_eval (/data/gannet/ripley/R/R-clang/bin/exec/R+0x1292f5)
    #10 0x55b4d6c77006 in do_set (/data/gannet/ripley/R/R-clang/bin/exec/R+0x147006)
    #11 0x55b4d6c590cf in Rf_eval (/data/gannet/ripley/R/R-clang/bin/exec/R+0x1290cf)
[...]

PR #6731, opened 2025-01-17
fread: avoid integer overflow in sumLenSq

aitap:
Cast to double earlier to avoid multiplying in integer arithmetic.

Fixes: #6729


src/fread.c
```diff
@@ -1971,7 +1971,7 @@ int freadMain(freadMainArgs _args) {
       int thisLineLen = (int)(ch-lineStart);  // ch is now on start of next line so this includes line ending already
       sampleLines++;
       sumLen += thisLineLen;
-      sumLenSq += thisLineLen*thisLineLen;
+      sumLenSq += (double)thisLineLen*thisLineLen; // avoid integer overflow, #6729
```
MichaelChirico, line 1974: this can be tested without _too_ much pain right? 

IINM we need a line with width ≥46341 in a file that does line sampling.

If you can write such a test in 10 minutes or so please go ahead (i.e. don't slave away too much here)

MichaelChirico (commented):
LGTM, thanks for the quick find! don't forget NEWS

src/fread.c
aitap: Can it be a test that only fails with UBSan enabled? Even if we overflow here, end up with a negative `sumLenSq` and `sd` equal to `NaN`, later we use `fmax(meanLineLen - 2*sd, minLen)` when calculating `allocnrow`, and `fmax` will return `minLen` if the other argument is `NaN`, so the overflow doesn't spread. Even making `meanLineLen - 2*sd` equal to 0 won't cause any damage because `minLen` is at least 1 (line lengths include the line terminators).

Without a sanitizer, I think this is only noticeable as a wrong initial allocation.

MichaelChirico: Yea, I think we have a number of tests in the suite that similarly only fail in certain settings. As long as the test is not too expensive to run I think it's fine to leave it in there.

aitap: I confirm that I see the "cannot be represented" message from UBSan in 28e9ed0b4ca704b74718069795050b7532c3c186 but that it's absent in 3d0a592a4e5e4609ca5e084467489f7bd6aacc77.

tdhock (approved):
lgtm

Issue #6733, opened 2025-01-18
Memory leak in `fwrite`


aitap:
Found while testing the fix for #6731.

Setting aside the leaks from `zlib` and `fontconfig`, we've got the following for `tests/main.R`:

```
Direct leak of 125829120 byte(s) in 15 object(s) allocated from:
    #1 0x7f43e5ea224a in fwriteMain src/fwrite.c:792
    #2 0x7f43e5ea7638 in fwriteR src/fwriteR.c:310
```

plus a few tens of megabytes from other objects allocated in the same place.

Looks like this allocation:
https://github.com/Rdatatable/data.table/blob/1aa92bc4ab3206b5e9f8accdc5b205edbb76c73a/src/fwrite.c#L792
are not always matched with this deallocation:
https://github.com/Rdatatable/data.table/blob/1aa92bc4ab3206b5e9f8accdc5b205edbb76c73a/src/fwrite.c#L1096

(Some allocations done internally by `zlib` are also leaks, but LeakSanitizer doesn't trace them to a specific line in `data.table`.)

We've got quite a few calls to `STOP()` between the two lines, and most of those don't release the memory allocated for `buffPool`. This bisects to e0abdfcd79ba30efcf679e33cbb8eba897a46f9c, i.e., #6393. Before this commit, `tests/main.R` finishes without any notes from LeakSanitizer.

**Edit**: the title used to say "leak in `fread`", which was not correct at all.

tdhock:
wow this is great detective work, thanks!
how did you realize there may be a leak?
what code did you use to do the leak check and bisect?
do you think it may be possible to adapt the code to run on GitHub actions?
and maybe for functions other than fwrite?

aitap:
I was testing #6731 on a [sanitized build of R](https://cran.r-project.org/doc/manuals/R-exts.html#Using-Undefined-Behaviour-Sanitizer). I needed UBSan (`-fsanitize=undefined`) to see the integer overflow, but I also had [ASan (`-fsanitize=address`)](https://cran.r-project.org/doc/manuals/R-exts.html#Using-Address-Sanitizer) enabled. Nowadays ASan includes [LeakSanitizer](https://cran.r-project.org/doc/manuals/R-exts.html#Using-Leak-Sanitizer), which alerted me to the leak by ending `tests/main.R` with a non-zero exit code.

[...]

PR #6746, opened 2025-01-21
GitLab CI workflow with sanitizers

aitap:
There are multiple ways to write the workflow in order to detect the issues like #6733.

One is to be quite close to the reference R-hub workflow description, which is in the `HEAD` of the current pull request. [Here](https://github.com/aitap/data.table/actions/runs/12885575751/job/35924326119)'s how it fails on a recent `master`.

The other is to write a more custom workflow, still reusing the container maintained by Gábor Csárdi and the infrastructure providing binary packages linked with `libc++` for the clang+sanitizers check. It's in the parent of the current pull request or [the GHA-sanitizers branch](https://github.com/Rdatatable/data.table/tree/GHA-sanitizers). [Here](https://github.com/Rdatatable/data.table/actions/runs/12885971483/job/35925545125)'s what the failure looks like.

(The supported option is to use `rhub::rhub_setup()`.)

Comments from people experienced with GHA are very welcome; this is the third time I'm doing any CI and the first time I'm writing a non-throwaway GitHub workflow.


MichaelChirico:
Did you have a look at the `clang-ubsan` .devcontainer we have? I have used that in the past. I am not sure about performance in GHA.

I'm not sure why the most recent run was cancelled, I'll close & re-open to re-trigger:

https://github.com/Rdatatable/data.table/actions/runs/12886173136

At a higher level, if performance is bad (say >=20% slower than any current PR-level action), I would recommend only running such an action on `master`.

To date, `master`-only checks have been done in GLCI -- @ben-schwen / @jangorecki do you know if we've ever had a UBSan (or MSan etc) action set up on GitLab?

aitap:
Thank you for pointing me towards the devcontainer! Unlike [`rocker/r-devel-ubsan`](https://github.com/rocker-org/r-devel-san/blob/e358f431db0d49b5d4c40d084a64649d7ee4f8e1/Dockerfile#L79), [`rocker/r-devel-san-clang`](https://github.com/rocker-org/r-devel-san-clang/blob/f794085e92d3bf738f778c9da05feeb2020cc1cf/Dockerfile#L94) does have AddressSanitizer enabled, and we can try to additionally enable LeakSanitizer like here.

I don't think it we'll get it significantly faster than 4 minutes (or anywhere close to 40 seconds), especially if optional dependencies are introduced, so let's leave it on `master`.

MichaelChirico:
GLCI is _probably_ the right way to go (cc @jangorecki & @ben-schwen).

Just want to note that because of our minimal approach to dependencies, the cache approach should be OK in general -- I guess the most important package to check with the sanitizer will be bit64, which is relatively stable / won't change noticeably more often than R. The other packages are probably not important for testing with the sanitizer _per se_, so if they are slightly behind CRAN it's unlikely to cause issues.

ben-schwen:
In GLCI we also cache the pkgs and install them but it seems that in this GHA we also install all imports/suggests for these pkgs?

```
Installing package(s) 'BiocVersion', 'bit64', 'bit', 'R.utils', 'xts', 'zoo',
  'yaml', 'knitr', 'markdown', 'ggplot2', 'hexbin', 'plyr', 'dplyr', 'caret',
  'gdata', 'nlme', 'sf', 'nanotime'
also installing the dependencies ‘listenv’, ‘parallelly’, ‘future’, ‘globals’, ‘digest’, ‘shape’, ‘future.apply’, ‘numDeriv’, ‘progressr’, ‘SQUAREM’, ‘diagram’, ‘lava’, ‘colorspace’, ‘tzdb’, ‘cpp11’, ‘prodlim’, ‘timechange’, ‘stringi’, ‘farver’, ‘labeling’, ‘munsell’, ‘RColorBrewer’, ‘viridisLite’, ‘fansi’, ‘pkgconfig’, ‘utf8’, ‘proxy’, ‘iterators’, ‘data.table’, ‘clock’, ‘gower’, ‘hardhat’, ‘ipred’, ‘lubridate’, ‘purrr’, ‘tidyr’, ‘timeDate’, ‘stringr’, ‘wk’, ‘R.oo’, ‘R.methodsS3’, ‘evaluate’, ‘highr’, ‘xfun’, ‘commonmark’, ‘cli’, ‘glue’, ‘gtable’, ‘isoband’, ‘lifecycle’, ‘rlang’, ‘scales’, ‘tibble’, ‘vctrs’, ‘withr’, ‘Rcpp’, ‘generics’, ‘magrittr’, ‘pillar’, ‘R6’, ‘tidyselect’, ‘e1071’, ‘foreach’, ‘ModelMetrics’, ‘pROC’, ‘recipes’, ‘reshape2’, ‘gtools’, ‘classInt’, ‘DBI’, ‘s2’, ‘units’, ‘RcppCCTZ’, ‘RcppDate’
```

aitap:
These are the strong dependencies of the packages from `inst/tests/other.Rraw`:
```
parse('inst/tests/other.Rraw')[[1]][[3]] |> eval() |>
 tools::package_dependencies(which = 'strong', recursive=TRUE) |>
 unlist() |> unique()
  [1] "cli"          "glue"         "grDevices"    "grid"         "gtable"
  [6] "isoband"      "lifecycle"    "MASS"         "mgcv"         "rlang"
 [11] "scales"       "stats"        "tibble"       "vctrs"        "withr"
 [16] "utils"        "methods"      "nlme"         "graphics"     "Matrix"
 [21] "splines"      "farver"       "labeling"     "munsell"      "R6"
 [26] "RColorBrewer" "viridisLite"  "fansi"        "magrittr"     "pillar"
 [31] "pkgconfig"    "colorspace"   "lattice"      "utf8"         "Rcpp"
 [36] "generics"     "tidyselect"   "ggplot2"      "e1071"        "foreach"
 [41] "ModelMetrics" "plyr"         "pROC"         "recipes"      "reshape2"
 [46] "stats4"       "class"        "proxy"        "codetools"    "iterators"
 [51] "data.table"   "dplyr"        "clock"        "gower"        "hardhat"
 [56] "ipred"        "lubridate"    "purrr"        "tidyr"        "timeDate"
 [61] "stringr"      "tzdb"         "cpp11"        "rpart"        "survival"
 [66] "nnet"         "prodlim"      "timechange"   "stringi"      "diagram"
 [71] "KernSmooth"   "lava"         "tools"        "shape"        "future.apply"
 [76] "numDeriv"     "progressr"    "SQUAREM"      "future"       "globals"
 [81] "parallel"     "digest"       "listenv"      "parallelly"   "zoo"
 [86] "gtools"       "bit"          "evaluate"     "highr"        "xfun"
 [91] "yaml"         "classInt"     "DBI"          "s2"           "units"
 [96] "wk"           "bit64"        "RcppCCTZ"     "RcppDate"     "R.oo"
[101] "R.methodsS3"
```
(Including `data.table` itself, which is interesting.)

Not installing these will indeed save a lot of time.

MichaelChirico:
oh, I definitely would not concern myself with getting other.Rraw working in the sanitizer GHA :)

aitap:
Tested on GitLab. The existing workflow is quite comprehensive, I didn't have to do much. The sanitized checks add slightly more than 10 minutes to the CI time.

.dev/ubsan.supp
```diff
@@ -0,0 +1,2 @@
+# TODO(bit>4.5.0.1): remove after a new 'bit' version is on CRAN
```
MichaelChirico, line 1: planning to do that next after data.table release :)

.gitlab-ci.yml
```diff
@@ -212,6 +212,25 @@ test-lin-ancient-cran:
     # Restore checking vignettes if upgrading our R dependency means knitr can be installed.
     - R CMD check --no-manual --no-build-vignettes --ignore-vignettes $(ls -1t data.table_*.tar.gz | head -n 1)
 
+# run the main checks with Address(+Leak),UBSanitizer enabled
+test-lin-san:
+  <<: *test-lin
+  image: docker.io/rocker/r-devel-ubsan-clang
```
MichaelChirico, line 218: is this `r-devel-gcc-strict` image equivalent?

https://gitlab.com/jangorecki/dockerfiles/-/blame/master/r-devel/r-devel-gcc-strict?ref_type=heads#L43

(thinking it may be preferable to consistently use Jan's repo)

MichaelChirico (commented):
LGTM, cc @ben-schwen / @jangorecki though who are more familiar with GLCI. Thanks!

.gitlab-ci.yml
aitap: This way the entire R and (almost) all packages are compiled with `-fsanitize=address,undefined`. I think there is a way compile just `data.table` with sanitizers by setting `PKG_CFLAGS`, `PKG_LIBS`, but then either the sanitizers must be manually preloaded using `LD_PRELOAD` or R must be linked with custom `MAIN_LD`(`FLAGS`) set.

jangorecki:
I would rather avoid using not ours docker images due to, as it happened in the past already, some of the issues in our tests might be caused by specifics of those images. It also happens they pulled in extra dependencies. Therefore having an own copy of that images already prevents breaking changes or new dependencies being pulled in, which can break some tests. Ultimately would be to move our images from my namespace to Rdatatable namespace.

ben-schwen:
I would also favor using our own images to not carry the extra dependency burden.
Otherwise LGTM.

[GLCI results](https://gitlab.com/Rdatatable/data.table/-/jobs/9156689429)

MichaelChirico:
^ merged the above. I think there's a follow-up step we need to publish the image? We should really write the procedure down...

MichaelChirico:
@ben-schwen / @jangorecki what would it take to move the images into the Rdatatable account?

ben-schwen:
Not much. Move or fork the repo (depending on if Jan wants to maintain some versions). Then we might have to initially push the images (but this might can be skipped if we move the repo)

jangorecki:
Yes, there should be not much extra effort. The only purpose for those images was data.ravle so there is no need to maintain them in my namespace if they will be moved to Rdatatable.

MichaelChirico:
Great! Jan I think you need to initiate the move, I didn't think I have high enough permissions

jangorecki:
Transfer does not work due to

> Project cannot be transferred to a different top-level namespace, because image tags are present in its container registry

It will be easier to just take a subset of images that are actually in use and clone them to Rdatatable.

MichaelChirico:
It was easiest to just `git push` everything to the new remote:

https://gitlab.com/Rdatatable/dockerfiles

We can prune as follow-ups. History & everything is retained since it's just a copy of the same `.git` repo.

What I don't know is what's next, e.g. re-registering the images.

aitap:
The `!1` merge request at the new location of the containers repo should build and publish the new container.

jangorecki:
FYI R devel image was set to be build on schedule, each night, not only on pushes

MichaelChirico:
I may just be bad at GitLab, but it _looks_ like as `Maintainer` I can't add @aitap as a "group member" over there, I think @jangorecki / @ben-schwen as `Owner` would need to do so.

ben-schwen:
> I may just be bad at GitLab, but it _looks_ like as `Maintainer` I can't add @aitap as a "group member" over there, I think @jangorecki / @ben-schwen as `Owner` would need to do so.

I think you need to do it at a group level (because of our open source license which has a number of limited seats) and not at a repo level. At a repo level, I also can't add members. In any case, I added @aitap

aitap:
Now [tested](https://gitlab.com/Rdatatable/data.table/-/jobs/9307006944) on GitLab, takes 10 minutes.

MichaelChirico (approved):
looks like we're in a lot better shape, thanks!

let's leave this sit another day. I started pre release stuff for bit last night, might actually submit it today in which case we can knock off that TODO already.

aitap:
Two problems remaining:
1. Need to properly fail the test if `R CMD check` succeeds but UBSan complains about something
2. Need to make sure the new `bit` is installed for the check

ben-schwen (approved):
LGTM TY!

PR #6748, opened 2025-01-21
Clarify proper nouns for role names

MichaelChirico:
cc @Rdatatable/committers

I think this improves clarity by making it clearer the word is specifically a role in each case.


tdhock:
great, in addition to "proper" please write "capitalized" or "upper-case" in the change log

PR #6757, opened 2025-01-23
Fix up some potential memory leaks in fwrite

### LINKED ISSUE #6733: Memory leak in `fwrite`

MichaelChirico:
Closes #6733.

Looked for any `STOP` or `return` before `free()` is called on any `malloc()` object, and made sure to `free()` before those exits if not already done & that object has successfully `malloc()`'d.

Also

 - Moved the `#ifndef` guards for `NOZLIB` outside the `args.is_gzip` checks, so that gzip-disabled installations don't even bother checking `is_gzip` (we error if `is_gzip` is attempted in gzip-disabled installations above)
 - Moved some `DTPRINT()` calls alerting about allocations to _after_ we checked that the allocation succeeded (the error message is informative enough in case of failure)
 
Thanks @aitap for pointing out the `mystream` issue too. cc @philippechataignon.


src/fwrite.c
```diff
@@ -829,14 +820,18 @@ void fwriteMain(fwriteMainArgs args)
       DTPRINT(_("Allocate %zu bytes (%zu MiB) for zbuffPool\n"), alloc_size, alloc_size / MEGA);
     }
     zbuffPool = malloc(alloc_size);
-    if (!zbuffPool) {
+    if (zbuffPool == NULL) {
```
MichaelChirico, line 823: We use this throughout the codebase and our "linter" checks for `!` usage specifically:

https://github.com/Rdatatable/data.table/blob/f72e46bbedc3f3f8c2abe0ea779d80e1e034ce32/.ci/linters/c/alloc_linter.R#L23

```suggestion
    if (!zbuffPool) {
```

philippechataignon:
A lot of allocation stuff in fwrite was written by Matt. I remove allocation for zlib streams and declare them as z_stream, not z_stream*. I've had an malloc(0) for freeing zbuffPool in all case (except if NOZLIB is defined).

src/fwrite.c
philippechataignon: OK, sorry for this useless change.

src/fwrite.c
```diff
@@ -836,7 +827,11 @@ void fwriteMain(fwriteMainArgs args)
            zbuffSize / MEGA, nth, errno, strerror(errno));
       // # nocov end
     }
+  } else {
+    // if no is_gzip, malloc 0 for allowing freeing zbuffPool
+    zbuffPool = malloc(0);
```
aitap, line 833: The C standard (even C89) allows giving null pointers to `free` (and allows `malloc(0)` to return a null pointer). Circa-2016 (R-3.3.0 release year) C standard libraries should all handle `free(NULL)` without a problem.

Initialising `char * zbuffPool = NULL` on line 801 above and skipping the `else` branch here would have been sufficient.

src/fwrite.c
```diff
@@ -1067,36 +1073,29 @@ void fwriteMain(fwriteMainArgs args)
...
+  if (args.is_gzip) {
+    unsigned char tail[10];
+    tail[0] = 3;
+    tail[1] = 0;
+    PUT4(tail + 2, crc);
+    PUT4(tail + 6, len);
+    int ret = WRITE(f, tail, 10);
+    compress_len += 10;
+    if (ret == -1)
+      STOP("Error: can't write gzip tailer"); // # nocov
```
aitap, line 1098:
```suggestion
      STOP("Error: can't write gzip trailer"); // # nocov
```
Does this message need translation like other calls to `STOP(...)`?

aitap (commented):
Thank you! We needed those `deflateEnd`s in addition to the `free`s. One remaining leak was visible through testing (zero-row `fwrite`) and one more leak on the error path is not easily testable but was confirmed manually in the debugger. There are no remaining complaints from LeakSanitizer.

src/fwrite.c
MichaelChirico: I thought 'tailer' was a special word 😂 

Yes, added `_()`

src/fwrite.c
MichaelChirico: NP, it's of course equivalent, just a matter of style (and some guides recommend your approach as being more transparent). Only using `!` for consistency.

PR #6760, opened 2025-01-23
don't check malloc(0)

MichaelChirico:
Should fix #6757


MichaelChirico:
Bypassing review for trivial ci-only change

PR #6766, opened 2025-01-25
sep2 can be an empty string

### LINKED ISSUE #5414: Feature Request: Use of sep2=''
Hello

Is there any reason why ```sep2=''``` is not allowed in ```fwrite```? 

Several R packages are used to manipulate genomic data (such as SNP) and do a good job filtering out those with bad quality. However, when comes to performing genomic analyses to estimate "Breeding Values" for example, on large scale, we have to do that outside of R. A common layout when writing the SNP to a file is the following, where the first column is the individual identification and the second is the SNP (where all SNPs are collapsed without any space between them).

```data.table``` is the fastest package I've benchmarked to read, write, and filter rows/columns. Because of that, I wonder if there is any way to use ```sep2=''```, as shown in the following example.

```
library("data.table")
geno <- data.table(
  IID = 1:10,
  SNP = lapply(1:10, function(i) sample(0:2, 10, replace = TRUE))
)
```

According to ```?fwrite```, ```sep2[2]``` must be a single character. Therefore I have to collapse the list, rather than use ```sep2```. Hence, if I try to run ```fwrite(geno, "Geno.txt", col.names = FALSE, sep = " ", sep2 = c("","",""))``` I get the following error message:

```
Error in fwrite(geno, "Geno.txt", col.names = FALSE, row.names = FALSE,  : 
  is.character(sep2) && length(sep2) == 3L && nchar(sep2[2L]) ==  .... is not TRUE
```

I would like to have the following result, without having to collapse all values before writing it to a file.

```
1 2221210202
2 0020010221
3 1010022212
4 0120121221
5 1212211202
6 2100002010
7 1110011210
8 1212012121
9 2221121021
10 1122220101
```

Thank you.

### PR DESCRIPTION
**Pull request by philippechataignon**
This PR permits to have an empty string in sep2. Test added. Closes #5414

### DIFF
--- a/R/fwrite.R
+++ b/R/fwrite.R
@@ -46,7 +46,7 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
     is.list(x),
     identical(quote,"auto") || isTRUEorFALSE(quote),
     is.character(sep) && length(sep)==1L && (nchar(sep) == 1L || identical(sep, "")),
-    is.character(sep2) && length(sep2)==3L && nchar(sep2[2L])==1L,
+    is.character(sep2) && length(sep2)==3L && nchar(sep2[2L])<=1L,
     is.character(dec) && length(dec)==1L && nchar(dec) == 1L,
     dec != sep,  # sep2!=dec and sep2!=sep checked at C level when we know if list columns are present
     is.character(eol) && length(eol)==1L,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by philippechataignon** [comment]
test.data.table() sometimes (not always) fails with message :
```
Test 1736.11 produced 1 errors but expected 0
Expected: 
Observed: Row 3 of list column is type 'NULL' - not yet implemented. fwrite() can write list columns containing items which are atomic vectors of type logical, integer, integer64, double, complex and character.
```

Issue #6616, opened 2024-11-15
Release data.table 1.17.0


MichaelChirico:
We are nearing our planned 6-month cadence for a non-patch release.

Filing this issue to track progress and as a call to flag issues for inclusion in / exclusion from the milestone:

https://github.com/Rdatatable/data.table/milestone/34

Thanks for everyone's work :)

Let's aim for what can be coded up & reviewed by the end of January?

TysonStanley:
That works, will get it live before the 20th deadline too.

TysonStanley:
How are we feeling about meeting the Feb 20 deadline for the whole release? A TON has been done but looks like a few lingering ones. The only one remaining that seems particularly messy is the C API changes. I imagine since we are technically submitting a release in response to a fix on CRAN, they may allow us one more release without having the full C API fixed.

MichaelChirico:
@TysonStanley yes, I think the investment required for the non-API stuff remains large. I suggest trying one more submission with the `NOTES` intact (especially, since a fix requires R 4.5.0 which is not yet officially released, or even scheduled).

I do want to fix up the last translation stuff, #6774 and #6776 have translation contributions that we should merge in, but it requires a bit of time I haven't found yet.

TysonStanley:
Sounds good, we still will ship as `1.17.0` and will communicate with CRAN about the process we are undertaking for the non-API stuff. Let me know when the translation stuff is all complete and will get this off.

MichaelChirico:
@TysonStanley GTG!

TysonStanley:
@MichaelChirico will start the process tonight

PR #6516, opened 2024-09-21
Russian translation

aitap:
First of all, big thanks to @Meeko28 and @Kate-lz! This would have never been possible without your work. I hope that you've received your grant payouts, or at least that the paperwork is on the way.

The `*.po` files are prepared against the current `*.pot` files, so a few of the strings containing fragmented sentences have been left untranslated.

Questions:
- Should this pull request be WIP until all the `*.pot` files are updated? I've got to submit some edits for the R message templates too, and the vignettes and the help pages may need their own rounds of editing too.
- Where do I put the [`data.table.ru` package](https://codeberg.org/aitap/data.table.ru/src/branch/translated-package) with the three vignettes (intro, reshape, importing) and (currently) 25% of the help pages translated? I ended up using a [po4a](https://po4a.org/)-based workflow for the vignettes (thank you very much @phgrosjean for the hint) and a similar but more [handmade workflow](https://codeberg.org/aitap/Rd2gettext) for the help files. The resulting package (when built) depends on only `data.table` and suggests a few of its dependencies (to build vignettes and run examples).

PR #6773, opened 2025-01-30
Re-generate .pot template files

MichaelChirico:
Hello all, in anticipation of #6616, here are the updated .pot files. I will merge this and then open PRs for the individual languages with the new/deprecated/fuzzy translations.

cc @aitap w.r.t. #6516


aitap:
Hopefully the only typo: `%d (type '%s' at` needs a closing parenthesis after `%s`.
https://github.com/Rdatatable/data.table/blob/79f5923acc8da7b3cbc72bda7ee663980b54abcb/src/assign.c#L970

PR #6774, opened 2025-01-30
Updated Spanish .po files

MichaelChirico:
Thanks in advance!


rikivillalba:
Pushed translation updates.

rikivillalba (commented):
- https://github.com/Rdatatable/data.table/pull/6774/commits/c9eff633da5741715fe6dccf7491709e4e39cfc5
- https://github.com/Rdatatable/data.table/pull/6774/commits/46e0084ee52bc8620755a16ef514805905b6212a

MichaelChirico:
Argh, it appears my update re-fuzzed your messages. The same happend in #6776, I had quite a manual fix. Any ideas for something more automated here?

rikivillalba:
Added new translations and restored re-fuzzied ones. based on _master_ .pot
master
- [61c1857a](https://github.com/Rdatatable/data.table/commit/61c1857a87a9a0580632eca63faa6bf5bb84b98d)

MichaelChirico:
> there are some entries that I do not see in the .pot in this branch. But i find the entries in in the .po Not sure if those were added later or should not be here, for case:

I've been a bit sloppy about updating this branch, probably I just updated the .po file (definitely so here: 0d8ab0caf05494c5e8c9adc32f7a0b5d36981adc). I think you're right it will make it easier for translators if they can just work from a single branch instead of working back and forth vs. master to find the `.pot` file.

MichaelChirico:
I added the `msgfmt`-compiled .mo files, and checked that all `msgid` in the current .po are indeed translated.

Thank you!!

Issue #6779, opened 2025-01-30
New CRAN Issue: boolean.h creating errors on r-devel Debian gcc


TysonStanley:
Just got this message from CRAN and co:

```
That seems to be from the change to src/include/R_ext/Boolean.h of

r87656 | ripley | 2025-01-28 13:27:50 +0100 (Tue, 28 Jan 2025) | 1 line
add section on C23doc/manual/R-exts.texi

Please investigate.
```

Error is on r-devel for Debian gcc <https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/data.table-00check.html> tied to the change shown for Jan 29 <https://developer.r-project.org/blosxom.cgi/R-devel/NEWS>.

What I'm seeing as the diff is:

```
Index: src/include/R_ext/Boolean.h
===================================================================
--- src/include/R_ext/Boolean.h	(revision 87655)
+++ src/include/R_ext/Boolean.h	(revision 87656)
@@ -1,6 +1,6 @@
 /*
  *  R : A Computer Language for Statistical Data Analysis
- *  Copyright (C) 2000, 2001 The R Core Team.
+ *  Copyright (C) 2000, 2025 The R Core Team.
  *
  *  This header file is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU Lesser General Public License as published by
@@ -29,13 +29,56 @@
 #undef FALSE
 #undef TRUE
 
+/* This can eventually be simplified to
+#if (!defined __cplusplus) && !(defined __STDC_VERSION__ && __STDC_VERSION__ > 201710L)
+# include <stdbool.h>
+#endif
+
 #ifdef  __cplusplus
 extern "C" {
+typedef bool Rboolean;
+#define FALSE false
+#define TRUE true
+}
+#define _R_RBOOLEAN_IS_BOOL_ 1
 #endif
-typedef enum { FALSE = 0, TRUE /*, MAYBE */ } Rboolean;
+*/
 
 #ifdef  __cplusplus
+
+extern "C" {
+    /* once cp11 is soerted
+typedef bool Rboolean;
+#define FALSE false
+#define TRUE true
+    */
+typedef enum { FALSE = 0, TRUE } Rboolean;
 }
+
+# else
+
+// Also include standard C boolean type
+#if defined __STDC_VERSION__ && __STDC_VERSION__ > 201710L
+// C23 so these are keywords
+#else
+// stdbool.h is C99, so available everywhere
[...]

PR #6781, opened 2025-01-30
Replace Rboolean type with int for LOGICAL

MichaelChirico:
Closes #6779. h/t @aitap for pinpointing the issue. I even read this line earlier and didn't make the connection :)

PR #6782, opened 2025-01-30
fread(): avoid casting int* to Rboolean*

aitap:
This used to work because the compilers were very likely to use `int` as the underlying type of the `enum Rboolean`. R-devel r87656 changes the definition of `Rboolean` from an `enum` to `bool`, so the sizes and the semantics of the types are very different now. Since `LOGICAL()` returns `int*`, and `NA_INTEGER` is an `int`, keep using that to access the vector contents.

Fixes: #6779. It may be worth looking further, but this one makes the tests pass on my PC.

It's unfortunate that we don't get a compiler warning for this; the sanitizer is also silent. Perhaps the compiler lets it go because accessing objects using a pointer to a byte-sized variable must be fine?

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -1061,7 +1061,7 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
     case RAWSXP:    BODY(Rbyte, RAW,    int, val!=0,                                    td[i]=cval)
     case LGLSXP:
       if (mc) {
-                    memcpy(td, LOGICAL(source), slen*sizeof(Rboolean)); break;
+                    memcpy(td, LOGICAL(source), slen*sizeof(int)); break;
       } else        BODY(int, LOGICAL,  int, val,                                       td[i]=cval)
     case INTSXP:    BODY(int, INTEGER,  int, val==NA_INTEGER ? NA_LOGICAL : val!=0,     td[i]=cval)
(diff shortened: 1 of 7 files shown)


aitap:
See #6781.

MichaelChirico:
Looks like there's other (untested?) places to fix too:

https://github.com/Rdatatable/data.table/blob/874e571ea193e2e3c4b71d6539c5ec04cc4fdaa3/src/negate.c#L8

MichaelChirico:
Here's another one, I guess `sizeof(bool)` matches `sizeof(int)`? for now?

https://github.com/Rdatatable/data.table/blob/874e571ea193e2e3c4b71d6539c5ec04cc4fdaa3/src/assign.c#L1064

MichaelChirico:
Another one here I am surprised is not failing at the compiler level:

https://github.com/Rdatatable/data.table/blob/874e571ea193e2e3c4b71d6539c5ec04cc4fdaa3/src/assign.c#L1210-L1211

MichaelChirico:
Thanks again for narrowing down the issue!

MichaelChirico (approved):

MichaelChirico:
I am wavering on whether NEWS is warranted here. Feels like a purely-internal change, so I am merging without a NOTE.

PR #6783, opened 2025-01-30
Place useFancyQuotes setting in test.data.table

### LINKED ISSUE #6772: Set useFancyQuotes=FALSE in test.data.table
Currently, this is only done in one place:

https://github.com/Rdatatable/data.table/blob/b739ab2d9150777e50f730b4147623733188bb2f/inst/tests/tests.Rraw#L162

But then each test that encounters this would also need to set the same option:

https://github.com/Rdatatable/data.table/blob/b739ab2d9150777e50f730b4147623733188bb2f/inst/tests/tests.Rraw#L10997-L10998

Better to just disable it everywhere.

Intended to do so in #6728, but it got mixed up with #6771. However, to be sure that it indeed causes no issues to set `useFancyQuotes=FALSE`, we should wait for CI to be green again before attempting a clean PR with this change.

MichaelChirico:
Closes #6772 now that #6779 is fixed

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -95,12 +95,13 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
     scipen = 0L,  # fwrite now respects scipen
     datatable.optimize = Inf,
     datatable.alloccol = 1024L,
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Pretty trivial test-only change

PR #6785, opened 2025-01-31
Avoid casts that don't change the pointer type

aitap:
In C, explicit pointer casts are mostly needed to work with the byte representation of a larger type. In other cases, either the implicit cast (e.g. to and from `void*`) is already fine and won't cause any warnings, or the code is doing something suspicious (e.g. `Rboolean*` ↔ `int*`) and we'd rather have the warning from the implicit cast.

For example, adding an explicit cast from `LOGICAL(...)` to `int*` doesn't prevent a warning, because `LOGICAL(...)` already returns an `int*`. If the pointer type returned by `LOGICAL(...)` ever changes, the _absence_ of the cast will cause the compiler to issue a diagnostic: then it would be an implicit cast to a different pointer type, likely a violation of the "strict aliasing" rule, something bad to warn about. An explicit cast, on the other hand, tells the compiler to silence the warning and do as written.

Initially found while working on #6782. More examples located using the following [Coccinelle](https://coccinelle.gitlabpages.inria.fr/website/) script:

```
@@
type T;
T E;
@@
- (T)
  E
```

and the following command line:

    spatch --sp-file identity_casts.cocci --dir $DATA_TABLE/src --recursive-includes -I $R_HOME/include -I $DATA_TABLE/src

The remaining examples found by `spatch` are all in `snprintf(...)` calls casting non-pointer values like an `uint64_t` corresponding to the `PRIu64` conversion specifier. It's probably fine to keep the code defensive and leave these casts in. If the non-pointer type does change, the cast should prevent most problems, unless the cast causes a signed overflow (to be caught by UBSan).


MichaelChirico (approved):
Great finds!

aitap:
Coccinelle installation adds 11 seconds to the `lint-c` job. Overall, it's still close to the other checks we have.

On a broader note, in addition to pre-written data.table-specific linters, might we benefit from more generic static checking for the C code? A [`cppcheck`](https://cppcheck.sourceforge.io/) run (3.5 minutes, could be sped up by limiting the combinatorial explosion of `#ifdef`s that it considers) found only a few very minor problems, such as `ptr = realloc(ptr, newsize)` (the original `ptr` would leak if `realloc` fails) or a forgotten `va_end` (I think it's a no-op on most platforms?). A commercial analyser could also be an option, like [PVS-Studio](https://pvs-studio.com/en/docs/manual/6579/) (downside: their conditions amount to [an ad in the README](https://pvs-studio.com/en/order/open-source-license/)).

PR #6792, opened 2025-02-02
Improve clarity of fwrite errors

### LINKED ISSUE #6791: Confusing message in fwrite.c
Reading fwrite.c code, it seems `ret0` should store the gzip header. If my understanding is correct, the message should be changed to either `Can't write gzip header` or `Write gzip header error`. I see the trailing "error" is similar to the next message, but with "Can't" added.

https://github.com/Rdatatable/data.table/blob/f9c2824b2ce1820c04713e10ba8d4d482cdc2e91/src/fwrite.c#L909-L911

MichaelChirico:
Closes #6791 

cc @rffontenelle

### DIFF
--- a/po/data.table.pot
+++ b/po/data.table.pot
@@ -2692,12 +2692,12 @@ msgstr ""
 
 #: fwrite.c:909
 #, c-format
-msgid "Can't write gzip header error: %d"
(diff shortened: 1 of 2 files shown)


rffontenelle (approved):
LGTM

PR #6800, opened 2025-02-07
Fix forder segfault

### LINKED ISSUE #6797: Potential segfault in forder
I've been hitting an issue running ```dt[, .(a = mean(a)), keyby = .(b, c, d)]``` leading to an error ```INTEGER() can only be applied to a 'integer', not a 'NULL'```. Unfortunately it's difficult to reproduce without sharing proprietary data and the error only happens intermittently, but I think I've traced it back to this line https://github.com/Rdatatable/data.table/blob/1.16.4/src/forder.c#L545.  If I try to run `Rf_PrintValue(ascArg);` immediately before this line, I eventually hit a segfault. My knowledge of R internals is limited, but it may be related to unprotecting the value in line 469.

It's worth noting that I've only hit this issue when running my code in a `foreach` loop with `doParallel`, but this may be leading to increased garbage collection.

MichaelChirico:
Closes #6797 

Not testable AFAIK.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -133,6 +133,8 @@ rowwiseDT(
 
 19. An integer overflow in `fread()` with lines longer than `2^(31/2)` bytes is prevented, [#6729](https://github.com/Rdatatable/data.table/issues/6729). The typical impact was no worse than a wrong initial allocation size, corrected later. Thanks to @TaikiSan21 for the report and @aitap for the fix.
 
+20. Fixed a memory issue causing segfaults in `forder`, [#6797](https://github.com/Rdatatable/data.table/issues/6797). Thanks @dkutner for the report and @MichaelChirico for the fix.
+
 ## NOTES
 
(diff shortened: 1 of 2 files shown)


aitap:
I'm not seeing a way to add a unit test, but building R with
--enable-strict-barrier, I can see (with master):

library(data.table)
dt <- data.table(a = runif(100), b = sample(1:5, 100, TRUE), c =
sample(1:4, 100, TRUE), d = sample(1:3, 100, TRUE))
gctorture2(1,1,TRUE) # <-- need inhibit_release=TRUE
setkey(dt, b, c, d)
# Error in forderv(x, cols, sort = TRUE, retGrp = !physical,
# reuseSorting = TRUE) :
# unprotected object (0x5652ef2eca28) encountered (was INTSXP)
# Calls: setkey -> setkeyv -> forderv

and the HEAD of this pull request doesn't raise any errors.

MichaelChirico:
OK I think we have consensus to review from a few committees between comments/reactions here & on issue. Going ahead.

PR #6807, opened 2025-02-09
tests.Rraw: only call setlocale(en_US.UTF-8) if not already UTF-8

aitap:
This avoids the `Sys.setlocale()` call on systems where the current locale is already UTF-8-compatible (almost all modern systems). If the `en_US.UTF-8` locale is missing, the call will fail, producing a warning. Mostly harmless and rare, but is annoying when it happens.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -18525,7 +18525,9 @@ rm(.datatable.aware)
 # tests for trunc.char handling wide characters # 5096
 local({
   lc_ctype = Sys.getlocale('LC_CTYPE')
-  Sys.setlocale('LC_CTYPE', "en_US.UTF-8") # Japanese multibyte characters require utf8
+  # Japanese multibyte characters require utf8. As of 2025, we're likely to be already running in a UTF-8 locale, but if not, try this setlocale() call as a last chance.
+  # Unfortunately, there is no guaranteed, portable way of switching to UTF-8 US English.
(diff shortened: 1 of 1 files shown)


inst/tests/tests.Rraw
```diff
@@ -18525,7 +18525,9 @@ rm(.datatable.aware)
 # tests for trunc.char handling wide characters # 5096
 local({
   lc_ctype = Sys.getlocale('LC_CTYPE')
-  Sys.setlocale('LC_CTYPE', "en_US.UTF-8") # Japanese multibyte characters require utf8
+  # Japanese multibyte characters require utf8. As of 2025, we're likely to be already running in a UTF-8 locale, but if not, try this setlocale() call as a last chance.
+  # Unfortunately, there is no guaranteed, portable way of switching to UTF-8 US English.
+  if (!l10n_info()$`UTF-8`) Sys.setlocale('LC_CTYPE', "en_US.UTF-8")
```
MichaelChirico, line 18530: My instinct is to make this `isTRUE()` for robustness, but I think we can be pretty confident this will only be `TRUE`/`FALSE`:

https://github.com/r-devel/r-svn/blob/9e4828a2c5bf3f3a3e0c6b7a66f8cb058c2fdf1f/src/main/platform.c#L3278
https://github.com/r-devel/r-svn/blob/9e4828a2c5bf3f3a3e0c6b7a66f8cb058c2fdf1f/src/include/Defn.h#L1569

(it's `Rboolean` which doesn't support `NA`)

MichaelChirico (approved):

PR #6808, opened 2025-02-09
Fixes for interactive `test.data.table()`

aitap:
The `catf()` issue causes warnings and eventually fails due to `options(warn=2)` set somewhere during `tests.Rraw`. Sorry for missing that in #6793. The second commit is mostly to improve the quality of life when running `test.data.table()` without an English locale set or translation disabled. There are still 37 failed tests that look directly at the output or condition objects, will see what can be done about those.


inst/tests/tests.Rraw
```diff
@@ -1361,43 +1361,20 @@ if (test_bit64) {
...
-err4 = try(data.table(a=1:3), silent=TRUE)
-options(datatable.alloccol=-1)
-err5 = try(data.table(a=1:3), silent=TRUE)
-options(datatable.alloccol=1024L)   # otherwise test() itself fails in its internals with the alloc.col error
-test(432.1, inherits(err1,"try-error") && grep("Has getOption[(]'datatable.alloccol'[)] somehow become unset?", err1))
-test(432.2, inherits(err2,"try-error") && grep("getOption[(]'datatable.alloccol'[)] should be a number, by default 1024. But its type is 'character'.", err2))
-test(432.3, inherits(err3,"try-error") && grep("is a numeric vector ok but its length is 2. Its length should be 1.", err3))
-test(432.4, inherits(err4,"try-error") && grep("It must be >=0 and not NA.", err4))
-test(432.5, inherits(err5,"try-error") && grep("It must be >=0 and not NA.", err5))
+# Test that unsetting datatable.alloccol is caught, #2014
```
MichaelChirico, line 1364: great fix & modernization!

R/test.data.table.R
```diff
@@ -478,7 +482,8 @@ test = function(num,x,y=TRUE,error=NULL,warning=NULL,message=NULL,output=NULL,no
     if (out[length(out)] == "NULL") out = out[-length(out)]
     out = paste(out, collapse="\n")
     output = paste(output, collapse="\n")  # so that output= can be either a \n separated string, or a vector of strings.
-    if (length(output) && !string_match(output, out)) {
+    # it also happens to turn off the 'y' checking branch below
+    if (length(output) && !string_match(output, out) && !foreign) {
```
MichaelChirico, line 486:
```suggestion
    if (length(output) && !foreign && !string_match(output, out)) {
```

R/test.data.table.R
```diff
@@ -490,7 +495,7 @@ test = function(num,x,y=TRUE,error=NULL,warning=NULL,message=NULL,output=NULL,no
       fail = TRUE
       # nocov end
     }
-    if (length(notOutput) && string_match(notOutput, out, ignore.case=TRUE)) {
+    if (length(notOutput) && string_match(notOutput, out, ignore.case=TRUE) && !foreign) {
```
MichaelChirico, line 498:
```suggestion
    if (length(notOutput) && !foreign && string_match(notOutput, out, ignore.case=TRUE)) {
```

R/test.data.table.R
```diff
@@ -425,13 +421,21 @@ test = function(num,x,y=TRUE,error=NULL,warning=NULL,message=NULL,output=NULL,no
...
+  }
   if (is.null(output) && is.null(notOutput)) {
     x = suppressMessages(withCallingHandlers(tryCatch(x, error=eHandler), warning=wHandler, message=mHandler))
     # save the overhead of capture.output() since there are a lot of tests, often called in loops
     # Thanks to tryCatch2 by Jan here : https://github.com/jangorecki/logR/blob/master/R/logR.R#L21
   } else {
     out = capture.output(print(x <- suppressMessages(withCallingHandlers(tryCatch(x, error=eHandler), warning=wHandler, message=mHandler))))
   }
+  if (!is.null(options)) {
+    base::options(old_options)
```
MichaelChirico, line 436: maybe worth a comment on why we need to reset options here

MichaelChirico (approved):
Thanks! For posterity please also tidy up the PR description which still discusses the `catf()` part that was split off to #6811.

R/test.data.table.R
aitap: Would you use a different idiom here? Something like

```r
e <- environment()
local({
  if (!is.null(options)) {
    old_options <- do.call(base::options, as.list(options)) # as.list(): allow passing named character vector for convenience
    on.exit(base::options(old_options))
  }
  if (is.null(output) && is.null(notOutput)) {
    e$x = suppressMessages(withCallingHandlers(tryCatch(x, error=eHandler), warning=wHandler, message=mHandler))
    # save the overhead of capture.output() since there are a lot of tests, often called in loops
    # Thanks to tryCatch2 by Jan here : https://github.com/jangorecki/logR/blob/master/R/logR.R#L21
  } else {
    e$out = capture.output(print(e$x <- suppressMessages(withCallingHandlers(tryCatch(x, error=eHandler), warning=wHandler, message=mHandler))))
  }
})
```

It's inconvenient to use `on.exit` while returning more than one value from `local`.

MichaelChirico: No, I think the comment here is good and the code is clean enough. Thanks!

Issue #6806, opened 2025-02-09
`print.data.table` may warn when key and index are both present


aitap:
Observed while reading the `source("data.table-Ex")` output for translation purposes; taken from `example(print.data.table)`:

```r
options(warn = 2)
DT = data.table(grp1 = sample(100, NN, TRUE),grp2 = sample(90, NN, TRUE),grp3 = sample(80, NN, TRUE))
setkey(DT, grp1, grp2)
setindex(DT, grp1, grp3) # will warn during printing after this but not before
DT
# Key: <grp1, grp2>
# Index: <grp1__grp3>
# Error in rbind(abbs, toprint) : 
#   (converted from warning) number of columns of result is not a multiple of vector length (arg 1)
```

The warning most likely originates here:
https://github.com/Rdatatable/data.table/blob/b7f2106efe038d93577f427f34c06d9c00b4c486/R/print.data.table.R#L106

May be a good exercise in finding the emergency exit from [R Inferno](https://www.burns-stat.com/documents/books/the-r-inferno/). If not, will debug myself soon.

PradeepFSTdhane123:
@aitap  I am excited to get involved and help with this bug issue. Could you assign it to me? I am looking forward to contributing and collaborating with the team. Thank you!"

aitap:
@PradeepFSTdhane123 You are absolutely welcome to try your hand at preventing the warning. Good luck! Start by reproducing the warning with `options(warn=2, error=recover)` set.

Mukulyadav2004:
Hi , i was checking out this issue , i am not able to reproduce it .


```
> NN = 200
> options(warn = 2, error = recover)
> DT = data.table(grp1 = sample(100, NN, TRUE),grp2 = sample(90, NN, TRUE),grp3 = sample(80, NN, TRUE))
> setkey(DT, grp1, grp2)
> setindex(DT, grp1, grp3) 
> DT
```
Key: <grp1, grp2>
Index: <grp1__grp3>
      grp1  grp2  grp3
     <int> <int> <int>
  1:     1    72    17
  2:     1    78    59
  3:     2    60    14
  4:     3    15    12
  5:     3    48     6
 ---                  
196:    98    57    25
197:    99    55    80
198:    99    65    70
199:   100     2    80
200:   100     9    46


could you please explain.

aitap:
It looks like you also need options(datatable.show.indices=TRUE). So
the full reproducer is either library(data.table); options(warn = 2,
[...]

PR #6816, opened 2025-02-14
Fix index printing by adding index info to header (#6806)

Mukulyadav2004:
Problem:
Currently, when options(datatable.show.indices = TRUE), print.data.table() tries to add index info to toprint. However, toprint may have a different number of columns, leading to the error:

`Error in rbind(abbs, toprint) : number of columns of result is not a multiple of vector length (arg 1)`

Fix:
Instead of modifying toprint, this PR adds the index information directly to header, ensuring a cleaner and safer display.

Changes:
  Extracts index names from the "index" attribute.
  Formats them (removes __ and replaces _ with , ).
  Creates a "Indices: ..." header string.
  Appends this to header.  

File changed: print.data.table R


R/print.data.table.R
```diff
@@ -63,14 +63,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
   if (show.indices) {
-    if (is.null(indices(x))) {
-      show.indices = FALSE
-    } else {
-      index_dt <- as.data.table(attributes(attr(x, 'index')))
-      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt)))
-      setnames(index_dt, print_names)
+    indices <- names(attr(x, "index", exact = TRUE))
+    if (length(indices)) {
+      cleaned_indices <- gsub("^__|_", ", ", indices)
```
aitap, line 68: `gsub` with `_` as a complete subexpression will match `_` anywhere, including in the middle of the column name. Why does the code match it here?

R/print.data.table.R
```diff
@@ -63,14 +63,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
-    if (is.null(indices(x))) {
-      show.indices = FALSE
-    } else {
-      index_dt <- as.data.table(attributes(attr(x, 'index')))
-      print_names <- paste0("index", if (ncol(index_dt) > 1L) seq_len(ncol(index_dt)) else "", ":", sub("^__", "", names(index_dt)))
-      setnames(index_dt, print_names)
+    indices <- names(attr(x, "index", exact = TRUE))
+    if (length(indices)) {
+      cleaned_indices <- gsub("^__|_", ", ", indices) 
+      cleaned_indices <- sub(", $", "", cleaned_indices)
```
aitap, line 69: There's probably a way to fix the code above so that it doesn't create a separator at the end of the string in the first place.

R/print.data.table.R
```diff
@@ -63,14 +63,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+    if (length(indices)) {
+      cleaned_indices <- gsub("^__|_", ", ", indices) 
+      cleaned_indices <- sub(", $", "", cleaned_indices) 
+    
+      # Create header string (matches key display style)
+      header <- paste0("Indices: ", paste(cleaned_indices, collapse = ", "))
+      if (exists("header", envir = parent.frame(), inherits = FALSE)) {
+        # Match data.table's existing header handling
+        assign("header", c(get("header", envir = parent.frame()), header), 
+               envir = parent.frame())
```
aitap, line 76: What is this code intended to do? Where is the `header` variable used? Use of `parent.frame()` in this context is almost certainly a mistake because [`parent.frame()` is the environment of the function that calls `print(x)`](http://adv-r.had.co.nz/Environments.html#calling-environments). Why is the `exists()` check needed? Why `assign()`? In R, `if` doesn't create a new lexical scope (unlike a function call), so your changes to the variables will be retained outside the block.

aitap (changes requested):
The guide that you and [your colleagues](https://github.com/Rdatatable/data.table/pull/6805#pullrequestreview-2604332710) are following has some advice that is counter-productive. Where does it come from? Perhaps it should be changed?

Putting the issue number that you would like to fix into the pull request _title_ doesn't help: GitHub ignores it. Instead you need to put the number into the _text_ (i.e. the _comment_) of the pull request. Then GitHub will link the two together.

Let's try to stick to the [code formatting style](https://en.wikipedia.org/wiki/Indentation_style) used by this project: when closing an `if` block and following it with an `else` block, put both braces on the same line as the word `else`:

```r
if (whatever) {
  # some code
} else { # <-- like here
  # some more code
}
```

It's usually best to minimise the changes you're introducing to a code base. This is both easier to review and has less chance of introducing bugs. Since R is mostly ambivalent about spaces, let's not introduce extra space at the end of some but not all of the lines. On the [Files tab](https://github.com/Rdatatable/data.table/pull/6816/files) of this pull request, you can see "lint-r" complaining about extra spaces. Could you please remove them?

But the main problem is the suggested solution. `print(x)` shouldn't have to change variables above it in the function call stack. The main issue here is that `abbs` is constructed from `classes1(x)` without taking into account that `cbind(x[...,], index_dt)` may be printed instead of just a subset of `x`. There must be a simpler way of either (1) extracting the classes from both data.tables or (2) padding `abbs` with a string for each index.

R/print.data.table.R
Mukulyadav2004: Thank you, @aitap, for your valuable feedback.
I now understand the issues you pointed out and will try to make the necessary changes accordingly. Initially, I declared header because an error occurred when an index existed, and since the index is metadata, I created header to store it. My intent was to update header dynamically when printing a data.table, ensuring index information is included.         
However after you tell that  if statements do not create a new lexical scope , I came to know about it. This led to header being modified in the parent environment instead of being managed locally within print.data.table().

Mukulyadav2004:
Hi @aitap 
As you mentioned, in the modified data.table, where we add columns from both x and index_dt via cbind, abbs was initially constructed only from x and not index_dt. I have made the necessary changes to address this.

However, after implementing these modifications, three tests from tests.Rraw are failing because the expected output differs from the observed output. Specifically, in these tests, the expected output does not include an extra column for index_dt. The failing tests are as follows:
```
DT2 <- data.table(a = 1:3, b = 4:6)
setindexv(DT2, c("b","a"))
test(1775.2, print(DT2, print.keys = TRUE),
     output=c("Index: <b__a>", "   a b", "1: 1 4", "2: 2 5", "3: 3 6"))

setindexv(DT2, "b")
test(1775.3, print(DT2, print.keys = TRUE),
     output=c("Indices: <b__a>, <b>", "   a b", "1: 1 4", "2: 2 5", "3: 3 6"))

setkey(DT2, a)
setindexv(DT2, c("b","a"))
test(1775.4, print(DT2, print.keys = TRUE),
     output=c("Key: <a>", "Indices: <b__a>, <b>", "   a b", "1: 1 4", "2: 2 5", "3: 3 6"))
```
Could you please clarify this.

aitap:
Hi @Mukulyadav2004

Could you please push your changes to this pull request? It's very hard
to diagnose code only knowing that it produces unwanted output but
without seeing the changes to the output code.

The expected output in tests 1775.2, 1775.3, 1775.4 does not contain
the index columns because show.indices= is not set, only print.keys= is.

R/print.data.table.R
aitap: Is there a place in the code where the `header` variable is used after being created by this block?

R/print.data.table.R
```diff
@@ -101,8 +110,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
       IDate = "<IDat>", integer64 = "<i64>", raw = "<raw>",
       expression = "<expr>", ordered = "<ord>")
     classes = classes1(x)
+    col_names <- colnames(toprint)
+    classes <- sapply(col_names, function(col_name) {
+      if (grepl("^index:", col_name)) {
```
aitap, line 115: This check is not very reliable. A used can create a column whose name starts with `index:`: ``data.table(`index:foo`='bar')``. Moreover, `grepl('^static-string')` can usually be replaced by `startsWith()`, which avoids the cost of having to compile a regular expression.

R/print.data.table.R
```diff
@@ -101,8 +110,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+        cls <- class(x[[col_name]])
+        if (is.list(cls)) cls <- unlist(cls)  # Ensure it's a character vector
+        if (length(cls) == 0) cls <- "unknown"  # Handle empty cases
+        cls[1]  # Take only the first class name
+      } else {
+        "unknown"
+      }
+    })
     abbs = unname(class_abb[classes])
+    abbs[classes == "index"] <- "<index>"
```
aitap, line 127: What if an R package provides an S3 class called `"index"`? There's more than 20000 CRAN packages, plus some more Bioconductor packages, plus a lot of packages only published on GitHub or even not published at all.

R/print.data.table.R
```diff
@@ -101,8 +110,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
       IDate = "<IDat>", integer64 = "<i64>", raw = "<raw>",
       expression = "<expr>", ordered = "<ord>")
     classes = classes1(x)
+    col_names <- colnames(toprint)
+    classes <- sapply(col_names, function(col_name) {
```
aitap, line 114: Instead of overwriting `classes` by re-creating it anew, can you append the index markers to the variable `classes` created two lines above?

R/print.data.table.R
```diff
@@ -101,8 +110,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+        if (length(cls) == 0) cls <- "unknown"  # Handle empty cases
+        cls[1]  # Take only the first class name
+      } else {
+        "unknown"
+      }
+    })
     abbs = unname(class_abb[classes])
+    abbs[classes == "index"] <- "<index>"
     if ( length(idx <- which(is.na(abbs))) ) abbs[idx] = paste0("<", classes[idx], ">")
+    stopifnot(length(abbs) == ncol(toprint))
```
aitap, line 129: Defensive programming is good, but it's better to add a test to `inst/tests/tests.Rraw` to verify that printing a `data.table` with indices does not raise any warnings due to invalid `rbind()`.

aitap (changes requested):
Thank you for making the improvements! You are approaching a good solution.

R/print.data.table.R
Mukulyadav2004: Thank you for your feedback. I have reviewed the code, and currently, the header variable is being assigned within this block. However, I do not see any explicit usage of header after its creation.  I believe the header might be printed due to the following code snippet: 
```
if (print.keys) {
    if (!is.null(ky <- key(x)))
    catf("Key: <%s>\n", toString(ky))
    if (!is.null(ixs <- indices(x)))
    cat(sprintf(
      ngettext(length(ixs), "Index: %s\n", "Indices: %s\n"),
      paste0("<", ixs, ">", collapse = ", ")
    ))
  }
```
  However, I am not entirely certain, and I would appreciate any clarification if I have misunderstood

R/print.data.table.R
Mukulyadav2004: To avoid such conflicts, I will ensure that only actual index columns are modified by explicitly tracking them using indices(x). This way, we only modify columns that are confirmed to be indices, rather than relying on a generic "index" class check.

Mukulyadav2004:
Hi @aitap,
I have incorporated all the changes you suggested, but I am still encountering 6 test failures out of 11,760 tests. The failing tests are: 2264.1, 2264.2, 2264.3, 2264.4, 2264.7, and 2264.8, all from inst/tests/tests.Rraw.
The issue seems to be that the observed output does not match the expected output.

Expected Output:
 ```
   grp1 grp2 grp3 index1:grp1__grp3 index2:grp3__grp1
 1:   77   61   53                 3                 5
 2:   80   66   37                 8                 4
 3:   27   42    8                 5                 3
 4:   66   37    7                 4                 7
 5:   38   69    5                 6                 2
 6:   72   89   69                 1                10
 7:   86   52   16                 2                 1
 8:   28   35   62                10                 8
 9:   95   82   80                 7                 6
10:   83   64   41                 9                 9
```
Observed Output:
    grp1 grp2 grp3
 1:   77   61   53
 2:   80   66   37
 3:   27   42    8
 4:   66   37    7
 5:   38   69    5
 6:   72   89   69
 7:   86   52   16
 8:   28   35   62
 9:   95   82   80
10:   83   64   41

All failures appear to be of the same nature—the index columns are missing in the observed output. Could you please provide guidance on what might be causing 
this and how best to resolve it?

File: R/print.data.table.R (hunk 4 above)
Comment (aitap) [reply R/print.data.table.R]: Let me rephrase that. When _you_ wrote the code that introduced the `header` variable, where did you _intend_ it to be used? I'm asking this question because there is no other use of this variable in `R/print.data.table.R`, and no uses of variables named `header` in other `*.R` files are applicable.

If you are using a machine learning model to write the code or the comments for you, please stop doing that, because it's doing us both a disservice. Relying on machine output deprives you from learning opportunities and prevents you from accumulating skill. (If machine learning models start writing good code by themselves, what use will there be for human programmers like you and I?) Having to review code that superficially looks like it would work but ultimately proves useless wastes maintainer time and morale.

File: inst/tests/tests.Rraw [hunk 8]
```diff
@@ -18784,6 +18784,11 @@ ans = c(
  "10:   83   64   41                 9                 9")
 # test where topn isn't necessary
 test(2264.8, print(DT, show.indices=TRUE), output=ans)
+# printing does not fail when indices are present
+test(2264.9, {
+  suppressWarnings( print(DT, show.indices=TRUE) )
```
aitap, line 18789: Instead of suppressing warnings, test that they don't occur. Try capturing the output from a `print(DT, show.indices=TRUE)` with the warning fixed and call `test(2264.9, print(DT, show.indices=TRUE), output = c(...example output goes here...))`.

aitap:
The current patch both removes the code that creates `index_dt` and then seems to set `show.indices` to `FALSE`, which prevents an error due to the `index_dt` being undefined. Try providing the variables that are needed by the `show.indices == TRUE` branches in `print.data.table` and then preserving the `show.indices` argument being true.

Mukulyadav2004:
Hi @aitap,

I have incorporated all the changes as per your suggestions, and this time, all tests from inst/tests/tests.Rraw have passed successfully. However, the following checks have failed:
atime performance tests / comment (pull_request)
code-quality / lint-r (pull_request)

Could you please advise on how to proceed with these.

MichaelChirico:
- for lintr, scroll through the Files tab, you will see line annotations telling you what's wrong
 - same for codecov
 - for atime, that's a known issue, you can ignore

Mukulyadav2004:
I have added tests for the lines in 'inst/tests/tests/tests.Rraw' that were identified by Codecov, but it seems they are still not being recognized, as I am encountering the same error as before. Could you please advise on how to resolve this issue.

R/print.data.table.R
```diff
@@ -22,23 +22,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+#  := in [.data.table sets .global$print=address(x) to suppress the next print i.e., like <- does. See FAQ 2.22 and README item in v1.9.5
     # The issue is distinguishing "> DT" (after a previous := in a function) from "> DT[,foo:=1]". To print.data.table(), there
     # is no difference. Now from R 3.2.0 a side effect of the very welcome and requested change to avoid silent deep copy is that
     # there is now no longer a difference between > DT and > print(DT). So decided that DT[] is now needed to guarantee print; simpler.
     # This applies just at the prompt. Inside functions, print(DT) will of course print.
     # Other options investigated (could revisit): Cstack_info(), .Last.value gets set first before autoprint, history(), sys.status(),
     #   topenv(), inspecting next statement in caller, using clock() at C level to timeout suppression after some number of cycles
     SYS = sys.calls()
-    if (identical(SYS[[1L]][[1L]], print) || # this is what auto-print looks like, i.e. '> DT' and '> DT[, a:=b]' in the terminal; see #3029.
```
MichaelChirico, line 33: why remove these comments?

R/print.data.table.R
```diff
@@ -22,23 +22,23 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
   if (col.names == "none" && class)
     warningf("Column classes will be suppressed when col.names is 'none'")
   if (!shouldPrint(x)) {
-    #  := in [.data.table sets .global$print=address(x) to suppress the next print i.e., like <- does. See FAQ 2.22 and README item in v1.9.5
+#  := in [.data.table sets .global$print=address(x) to suppress the next print i.e., like <- does. See FAQ 2.22 and README item in v1.9.5
```
MichaelChirico, line 25: why this change?

R/print.data.table.R
```diff
@@ -57,8 +57,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
       catf("Null data.%s (0 rows and 0 cols)\n", class)  # See FAQ 2.5 and NEWS item in v1.8.9
     } else {
       catf("Empty data.%s (%d rows and %d cols)", class, NROW(x), NCOL(x))
-      if (length(x)>0L) cat(": ",paste(head(names(x),6L),collapse=","),if(length(x)>6L)"...",sep="") # notranslate
-      cat("\n") # notranslate
+      if (length(x)>0L) cat(": ",paste(head(names(x),6L),collapse=","),if(length(x)>6L)"...",sep="")  # notranslate
```
MichaelChirico, line 60: why this change?

inst/tests/tests.Rraw
```diff
@@ -21041,3 +21041,20 @@ test(2304.100, set(copy(DT), i=2L, j=c("L1", "L2"), value=list(list(NULL), list(
...
+# Test for covering if (is.list(cls)) cls <- unlist(cls)
+DT = data.table(A = 1:3, B = list(NULL, NULL, NULL))
+test(2306.2, DT, data.table(A = 1:3, B = list(NULL, NULL, NULL)))
+# Test for covering if (length(cls) == 0) cls <- "unknown"
+DT = data.table(A = 1:3, B = I(list()))
+test(2306.3, DT, data.table(A = 1:3, B = I(list())))
+# Test for covering classes[col_name] <- "unknown"
+DT = data.table(A = 1:3, B = 4:6, C = 7:9)
+if ("D" %in% colnames(DT)) DT[, D := NULL]
+test(2306.4, DT, data.table(A = 1:3, B = 4:6, C = 7:9))
```
MichaelChirico, line 21060: missing final newline

MichaelChirico:
> I have added tests for the lines in 'inst/tests/tests/tests.Rraw' that were identified by Codecov, but it seems they are still not being recognized, as I am encountering the same error as before. Could you please advise on how to resolve this issue.

You are writing code about data.table's `print()` method, but none of your tests involve printing (just input/output equality).

You want to use `test(output=)` to ensure printing is invoked and you're actually testing the behavior you're changing.

R/print.data.table.R
Mukulyadav2004: I noticed these changes were made while re-checking the code. I will revert them to maintain consistency.

Mukulyadav2004:
Hi @MichaelChirico 
Apologies for the interruption. Could you kindly review the test I’ve written. Despite making changes to include printing, it still fails at the same checks. I would greatly appreciate your guidance on how to correct it.

inst/tests/tests.Rraw
```diff
@@ -21041,3 +21041,37 @@ test(2304.100, set(copy(DT), i=2L, j=c("L1", "L2"), value=list(list(NULL), list(
...
+       "2: 2 [NULL]",
+       "3: 3 [NULL]"))
+# Test for covering classes[col_name] <- "unknown"
+DT <- data.table(A = 1:3, B = 4:6, C = 7:9)
+if ("D" %in% colnames(DT)) DT[, D := NULL]
+test(2306.4, {print(DT); capture.output(print(DT))}, 
+     c("   A B C",
+       "1: 1 4 7",
+       "2: 2 5 8",
+       "3: 3 6 9"))
```
aitap, line 21076: In order to exercise the index-printing code, make sure to create an index using `setindex(...)` and use `print(..., datatable.show.indices=TRUE)` or `test(..., output = ..., options = c(datatable.show.indices=TRUE))`. In order to reproduce the warning, make sure to create more than one column, otherwise `rbind()` silently recycles the length-1 vector `abbs`.

R/print.data.table.R
```diff
@@ -101,7 +101,22 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
-    abbs = unname(class_abb[classes])
+    col_names <- colnames(toprint)
+    index_cols <- paste0("index:", indices(x))
+    for (col_name in col_names) {
+      if (col_name %in% index_cols) {
+        classes[col_name] <- "index"
+      } else if (col_name %in% names(x)) {
+        cls <- class(x[[col_name]])
+        if (is.list(cls)) cls <- unlist(cls)
+        if (length(cls) == 0) cls <- "unknown"
```
aitap, line 112: I don't think R will allow making the class vector a list. Also, `class()` will return the name of the primitive type of the value if the class attribute is set empty or removed altogether, so both tests here are impossible.

R/print.data.table.R
```diff
@@ -101,7 +101,22 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+        classes[col_name] <- cls[1]
+      } else {
+        classes[col_name] <- "unknown"
```
aitap, line 115: This is likely to be impossible due to the way `toprint` is constructed from `x`.

aitap (changes requested):
Better, but the changes to unrelated code need to be reverted, and it's better to reuse the `classes1` logic instead of reimplementing it.

Mukulyadav2004:
Hi @aitap,
Could you please review the changes when you have some time?
To summarize, I’ve made the following modifications:
1: Ensured that toprint correctly inherits from "data.table" and "data.frame" if it does not already.
2: Adjusted how column classes are determined, appending "<index>" when show.indices is enabled and indices are present.
Currently, two checks are failing: atime performance and lint -r. The lint check suggests replacing explicit returns with implicit returns across multiple R files. Would you recommend making these changes as part of this PR, or should they be handled separately?

R/print.data.table.R
```diff
@@ -86,6 +86,14 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     if (show.indices) toprint = cbind(toprint, index_dt)
   }
   require_bit64_if_needed(x)
+  if (!inherits(toprint, "data.frame")) {
+    class(toprint) <- c("data.table", "data.frame")
```
aitap, line 90: Could you please explain why this is needed?

aitap (changes requested):
Almost there. Here's a test:

```r
library(data.table)
DT <- data.table(a = 1:2, b = 2:1)
setindex(DT, b)
data.table:::test(
 0,
 { print(DT); TRUE },
 options = c(datatable.show.indices = TRUE)
)
```

Currently it fails with "produced 1 warnings but expected 0". In order to get it to pass, you need to produce the correct `classes` variable. Currently, you add the indices to the column headers twice. Once when you call `classes <- classes1(toprint)`:

```
debug: classes <- classes1(toprint)
# ...
Browse[1]> print(classes)
[1] "integer" "integer" "integer"
```

...and for the second time when you append `rep("<index">, length(indices(x)))`:

```
debug: classes <- c(classes, rep("<index>", length(indices(x))))
# <...>
Browse[1]> print(classes)
[1] "integer" "integer" "integer" "<index>"
```

Use one technique but not the other.

R/print.data.table.R
aitap: This error is signalled by `colnames(toprint)`. Since the `col_names` variables isn't used anywhere, it's better to remove both it and the `inherits(toprint, ...)` check.

aitap (approved):
Thank you very much for your persistence. The two failing CI checks are currently expected. The test fails on `master` (checked manually) but passes here.

PR #6817, opened 2025-02-15
add committer emails

tdhock:
to support #6803 I would suggest that committer add their emails to DESCRIPTION.

PR #6818, opened 2025-02-15
move Ivan up with other committers in DESCRIPTION

tdhock:
While doing #6817 I noticed that Ivan's name is still at the middle of the author list, so I moved it up with the other Committers, for easy human discoverability.


aitap:
Thanks! Adding information for #6817 here to avoid merge conflicts.

tdhock:
thanks

Issue #6819, opened 2025-02-17
`growVector`, `copyAsPlain` may formally cause undefined behaviour with 0-length vectors


aitap:
Found while working on a custom Debian testing-based container (`using C compiler: ‘Debian clang version 19.1.7 (1+b1)’`) with Clang sanitizers enabled for #6746:

When trying to access the contents of a zero-length vector using `INTEGER(...)` or `REAL(...)` or other accessor, R may return an invalid pointer (`0x1`). The C standard [says that giving an invalid pointer to `memcpy()` is undefined behaviour](https://stackoverflow.com/a/5243068), even though in practice nothing breaks (`memcpy` sees `n=0` and doesn't dereference it).

If nothing breaks, what's the risk? One of the CRAN special checks (`clang-UBSAN` or `0len`) might pick this up too. Very far-fetched, a compiler might optimize away a chunk of code deemed to cause undefined behaviour.

```
Running test id 173.1

dogroups.c:541:39: runtime error: load of misaligned address 0x000000000001 for type 'int *', which requires 4 byte alignment
0x000000000001: note: pointer points here
<memory cannot be printed>
    #0 0x7f2efbaec116 in growVector /work/data.table.Rcheck/00_pkg_src/data.table/src/dogroups.c
    #1 0x7f2efbae90d5 in dogroups /work/data.table.Rcheck/00_pkg_src/data.table/src/dogroups.c:409:66

(gdb) frame 4
#4  0x00007fc6d4aec117 in growVector (x=0x52500661b038, newlen=newlen@entry=3) at dogroups.c:543
543       case CPLXSXP: memcpy(COMPLEX(newx), COMPLEX(x), len*SIZEOF(x)); break;
(gdb) p Rf_xlength(x)
$4 = 0
(gdb) p Rf_xlength(newx)
$5 = 3
(gdb) call Rf_PrintValue(R_GlobalContext->call)
`[.data.table`(DT, , B[B > 3], by = A)
```

```
Running test id 893.5

utils.c:233:12: runtime error: store to misaligned address 0x000000000001 for type 'int *', which requires 4 byte alignment
0x000000000001: note: pointer points here
<memory cannot be printed>
    #0 0x7f2efbbfd5f9 in copyAsPlain /work/data.table.Rcheck/00_pkg_src/data.table/src/utils.c:233:5
[...]

PR #6820, opened 2025-02-17
Avoid `memcpy()`-ing 0-length vectors

aitap:
The C standard (before C23) says that giving invalid pointers, even with length=0, to `memcpy()` is undefined behaviour. R returns an invalid pointer for vectors of length 0, so avoid calls to `memcpy()` altogether in such cases.

Fixes: #6819

### DIFF
--- a/src/dogroups.c
+++ b/src/dogroups.c
@@ -535,7 +535,7 @@ SEXP growVector(SEXP x, const R_len_t newlen)
   if (isNull(x)) error(_("growVector passed NULL"));
   PROTECT(newx = allocVector(TYPEOF(x), newlen));   // TO DO: R_realloc(?) here?
   if (newlen < len) len=newlen;   // i.e. shrink
-  switch (TYPEOF(x)) {
+  if (len) switch (TYPEOF(x)) {
   case RAWSXP:  memcpy(RAW(newx),     RAW(x),     len*SIZEOF(x)); break;
   case LGLSXP:  memcpy(LOGICAL(newx), LOGICAL(x), len*SIZEOF(x)); break;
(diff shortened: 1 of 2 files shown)


src/dogroups.c
```diff
@@ -535,7 +535,7 @@ SEXP growVector(SEXP x, const R_len_t newlen)
   if (isNull(x)) error(_("growVector passed NULL"));
   PROTECT(newx = allocVector(TYPEOF(x), newlen));   // TO DO: R_realloc(?) here?
   if (newlen < len) len=newlen;   // i.e. shrink
-  switch (TYPEOF(x)) {
+  if (len) switch (TYPEOF(x)) {
```
MichaelChirico, line 538: I wonder if it'll be clearer to be more explicit here, WDYT?

```suggestion
  if (!len) { // cannot memcpy invalid pointer, #6819
    keepattr(newx, x)
    UNPROTECT(1);
    return newx;
  }
  switch (TYPEOF(x)) {
```

MichaelChirico (approved):
LGTM, feel free to merge whether or not you want to make my suggested change.

Do we have any way of preventing this from recurring in the future?

aitap:
Maybe not literally preventing, but a sanitizer-enabled CI check should be able to catch similar problems as soon as they land on `master`.

Then why does `rocker/r-devel-san-clang` built using the same Debian Testing `clang` miss this, while [our own container](https://gitlab.com/jangorecki/dockerfiles/-/raw/master/r-devel/r-devel-clang-san) doesn't? Not sure. Maybe due to [`-O3` optimizing some of it away](https://github.com/rocker-org/r-devel-san-clang/blob/f794085e92d3bf738f778c9da05feeb2020cc1cf/Dockerfile#L96)?

PR #6821, opened 2025-02-18
Set French team as OWNERS

MichaelChirico:
I just noticed @Rdatatable/french was not automatically assigned to review #6775 :\

This should fix that going forward.


tdhock:
great thanks

PR #6822, opened 2025-02-18
Remove accidental fuzzy flag in R-pt_BR.po

rffontenelle:


leofontenelle (changes requested):
Que tal trocar "Subconjunto otimizado ..." por "Obtendo subconjunto de forma otimizada com ..."?

aitap:
Does potools use msgfmt -f? If not, fuzzy translations don't make it
into the *.mo. I think that -f is off by default, and the documentation
recommends against using it:
https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html#index-_002df_002c-msgfmt-option

leofontenelle:
> I _think_ we don't need to update the .mo file, my understanding is `fuzzy` has no impact on the po->mo compilation.

IIRC fuzzy translations are not included in the MO.

leofontenelle (approved):

rffontenelle:
> I _think_ we don't need to update the .mo file, my understanding is `fuzzy` has no impact on the po->mo compilation.

If fuzzy flag has no impact on the mo file, something is not correct. If the string has fuzzy, mo should not contain the translation (i.e. software should not show it translated).

MichaelChirico:
I wound up running `msgfmt` manually, I need to update {potools} to be able to specify which languages to compile, it will fit our work flow better:

I copied from {potools}, anyway -- no `-f`:

https://github.com/MichaelChirico/potools/blob/0dc529285c4f54a86d0755317d9304d735c3858f/R/msgmerge.R#L43

> IIRC fuzzy translations are not included in the MO.

This was my memory too, but the main gettext docs hit for this is not helpful :)

https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html

(aside, something odd is happening with your messages @aitap, they come through with HTML mark-up to e-mail?)

image

MichaelChirico (approved):

aitap:
@MichaelChirico This is what I get for being too lazy to log in and replying by e-mail. Apparently, GitHub has a plain-text-email-to-Markdown translator which outputs HTML &lt;br&gt; tags. They then apparently send the resulting "Markdown" (a superset of HTML) as the plain text part of the e-mail.

MichaelChirico:
> @MichaelChirico This is what I get for being too lazy to log in and replying by e-mail. Apparently, GitHub has a plain-text-email-to-Markdown translator which outputs HTML <br> tags. They then apparently send the resulting "Markdown" (a superset of HTML) as the plain text part of the e-mail.

Fun. I suspected that might be at play... between this and plain-text-required for the r-devel lists... 😭

PR #6824, opened 2025-02-19
Use seq_along() where appropriate

MichaelChirico:
This is newly exposed by updates to {lintr}:

https://github.com/r-lib/lintr/pull/2758


MichaelChirico:
Merging this since it's a no-op and the existing failure is causing some confusion on other PR reviews.

PR #6825, opened 2025-02-19
consistently use "=" for assignment

MichaelChirico:
There are few enough `<-` usages that I went ahead and made the switch, not that upstream {lintr} accommodates our choice of default operator.

There is still a pretty major issue with `assignment_linter()`:

https://github.com/r-lib/lintr/issues/2765

So I am still holding off activating it generally.


tdhock:
ok for the change from <- to = but I don't understand what is the issue with linting? and why you want to hold off activating it generally?
I see lint-r failures that say we should use <- instead of = why? (I expect those failures should be suppressed because we want to use = instead of <- right?)

.ci/.lintr.R
```diff
@@ -26,8 +26,8 @@ linters = c(dt_linters, all_linters(
   #   setwd = NULL
   # )),
   undesirable_operator_linter(),
-  # TODO(lintr#2441): Use upstream implementation.
-  assignment_linter = NULL,
+  # TODO(lintr#2765): Use upstream implementation.
+  # assignment_linter(operator = "="),
```
MichaelChirico, line 30:
```suggestion
  # assignment_linter(operator = "="),
  assignment_linter = NULL,
```

MichaelChirico:
> what is the issue with linting?

Whoops, we still need to actively disable the default `assignment_linter()` (which requires `<-` or `<<-`). Fixed: 3af2c2a211d0561b2358d4fc3426d60c713d613d

We still comment out `assignment_linter(operator="=")` because of the linked {lintr} issue: we use `<-` a lot for implicit assignments, e.g.

https://github.com/Rdatatable/data.table/blob/770e80b6d83bad5870d1df2df252d5b00ddd5134/R/merge.R#L38

The _intended_ behavior is that `implicit_assignment_linter()` governs the style for implicit assignments:

https://github.com/Rdatatable/data.table/blob/770e80b6d83bad5870d1df2df252d5b00ddd5134/.ci/.lintr.R#L59

MichaelChirico:
The remaining lints here are those fixed in #6284

tdhock:
ok, so I understand that there is some issue in lintr about "implicit" assignment versus not. (explicit?)
since the remaining lints are unrelated, I guess it is ok to merge?
```
Warning: file=R/data.table.R,line=1186,col=33,[seq_linter] Use seq_along(x) instead of seq_len(length(x)).
Warning: file=R/data.table.R,line=2383,col=30,[seq_linter] Use seq_along(x) instead of seq_len(length(x)).
Warning: file=R/data.table.R,line=2870,col=38,[seq_linter] Use seq_along(x) instead of seq_len(length(x)).
```

MichaelChirico:
> since the remaining lints are unrelated, I guess it is ok to merge?

See comment above :)

PR #6826, opened 2025-02-20
Add a mask for future %||%, use it

MichaelChirico:
Uncovered by the new `coalesce_linter():

https://github.com/r-lib/lintr/pull/2767

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -22,7 +22,7 @@ nan_is_na = function(x) {
 }
 
 # R 4.4.0
-if (!exists("%||%", "package:base")) `%||%` <- function(x, y) if (is.null(x)) y else x # nolint: null_coalescing_linter.
+if (!exists("%||%", "package:base")) `%||%` <- function(x, y) if (is.null(x)) y else x # nolint: coalesce_linter.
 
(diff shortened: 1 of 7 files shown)


R/utils.R
```diff
@@ -21,6 +21,9 @@ nan_is_na = function(x) {
   stopf("Argument 'nan' must be NA or NaN")
 }
 
+# R 4.4.0
+if (!exists("%||%", "package:base")) `%||%` <- function(x, y) if (is.null(x)) y else x # nolint: null_coalescing_linter.
```
MichaelChirico, line 25:
```suggestion
if (!exists("%||%", "package:base")) `%||%` <- function(x, y) if (is.null(x)) y else x # nolint: coalesce_linter.
```

MichaelChirico:
Are you thinking to replace `get0()` with `%||%`, something like `get("done_units_report", parent.frame()) %||% FALSE`? I don't have an immediate preference either way... presumably `get0()` is a bit optimized...

tdhock (approved):
cool, I did not know about `%||%` but that is a nice trick

PR #6829, opened 2025-02-22
Minimal set of changes to permit master to be 1.17.99

eddelbuettel:
Congratulations on getting 1.17.0 onto CRAN.

As with previous releases, `master` lags as is still at 1.16.99 in DESCRIPTION and src/init.c (plus some .po files I left alone).  Feel free to add to the PR as needed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: data.table
-Version: 1.16.99
+Version: 1.17.99
 Title: Extension of `data.frame`
 Depends: R (>= 3.3.0)
 Imports: methods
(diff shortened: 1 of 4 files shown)


TysonStanley:
Thanks @eddelbuettel, didn't get this in last night after it was accepted, thanks for getting it started. @MichaelChirico this should be all the files that are needed to bump to `1.17.99`

MichaelChirico:
atime & lintr failures are OK. Latter is due to some improved logic just landed upstream, will fix that separately.

thanks!

MichaelChirico (approved):

MichaelChirico:
@TysonStanley maybe the procedure should be to push an "as-submitted" release branch, then anyone who sees it hit CRAN can merge async, WDYT?

of course if it's rejected, the PR can be deleted

Issue #6841, opened 2025-02-25
Alias .formula2varlist for old R


MichaelChirico:
I've tried \method{sort_by}{data.table}(x, y, \dots) together with
\alias{sort_by.data.table} and, surprisingly, it just worked on both
R-4.2.2 and R-3.3.0. There's no need for scary-looking kludges like
\if{\Sexpr[stage=install,results=rd]{as.character(exists("sort_by",
"package:base"))}{\method{...}}. R CMD check only verifies that all
exported symbols are documented, not that all documented aliases are
exported (or even exist in the namespace).

The old versions of R do complain about missing .formula2varlist.

_Originally posted by @aitap in https://github.com/Rdatatable/data.table/issues/6679#issuecomment-2683352970_

MichaelChirico:
Follow-up to #6679

PR #6830, opened 2025-02-23
Remove more explicit return()s thanks to lintr update

MichaelChirico:
{lintr} removed a lot of false negatives in `return_linter()`:

https://github.com/r-lib/lintr/pull/2779

Basically any function with any argument with defaults could be affected.

PR #6842, opened 2025-02-26
Provide our own `.formula2varlist` implementation

aitap:
Since [`base::.formula2varlist` is not an API](https://search.r-project.org/R/refmans/base/html/base-internal.html) and is now needed in two places, provide our own implementation.

Fixes: #6841. Tested manually that an older R version no longer complains about "no visible binding for `.formula2varlist`" during `R CMD check`.

Alternatively, we could try some tricks to convince the static analysis tools on R < 4.3 that `.formula2varlist` exists (despite `sort_by.data.table` is not used there).


aitap:
Hmm, right, [there is no coverage for new enough versions of R](https://github.com/Rdatatable/data.table/pull/6679#issuecomment-2680831434).

MichaelChirico (approved):
Works for me!

Ignoring codecov issue as noted, and lintr issue is still #6830

Issue #6839, opened 2025-02-24
Adding Months to IDate Generates Incompatible classes with Period


emallickhossain:
I cannot add months to an IDate object when the date is at the end of a month (29th, 30th, or 31st). However, when I use a Date object, it works just fine. 

```
library(data.table)
library(lubridate)

> as.IDate("2020-01-01") %m+% months(1)
[1] "2020-02-01"
> as.Date("2020-01-01") %m+% months(1)
[1] "2020-02-01"
> as.IDate("2020-01-30") %m+% months(1)
Error: Incompatible classes: <IDate> - <Period>
> as.Date("2020-01-30") %m+% months(1)
[1] "2020-02-29"
```

# Session Info
```
> sessionInfo()
R version 4.4.2 (2024-10-31)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 24.10

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.12.0 
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.12.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

time zone: America/New_York
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] lubridate_1.9.4   data.table_1.17.0

loaded via a namespace (and not attached):
[1] compiler_4.4.2   generics_0.1.3   tools_4.4.2      timechange_0.3.0
```

emallickhossain:
Unfortunately, it doesn't. Same error along with another one about multiple classes in the cache.

```
> as.IDate(Sys.Date()) - days(1)
Error: Incompatible classes: <IDate> - <Period>
> setClass("IDate", contains = "Date")
> as.IDate(Sys.Date()) - days(1)
Found more than one class "IDate" in cache; using the first, from namespace 'data.table'
[...]

PR #6844, opened 2025-02-28
Register the full IDate S3 class for S4 dispatch

aitap:
As recommended by `?setOldClass`, register `c("IDate", "Date")`, not just `"IDate"`. The unit test doesn't involve `lubridate`, but also tested with `lubridate` manually:
```
> as.IDate(Sys.Date()) - days(1)
[1] "2025-02-27"
> as.IDate("2020-01-30") %m+% months(1)
[1] "2020-02-29"
```

Fixes: #6839

### DIFF
--- a/R/AllS4.R
+++ b/R/AllS4.R
@@ -16,7 +16,7 @@ methods::setAs("data.table", "data.frame", function(from) {
   as.data.frame(from)
 })
 
-methods::setOldClass("IDate")
+methods::setOldClass(c("IDate", "Date"))
 methods::setOldClass("ITime")
 
 methods::setAs("character", "IDate", function(from) as.IDate(from))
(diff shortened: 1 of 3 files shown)


R/AllS4.R
```diff
@@ -16,7 +16,7 @@ methods::setAs("data.table", "data.frame", function(from) {
   as.data.frame(from)
 })
 
-methods::setOldClass("IDate")
+methods::setOldClass(c("IDate", "Date"))
```
MichaelChirico, line 19: up on L6 I see multiple calls for a similar case on the data.table/data.frame class. should we remove the one-class call there, or is there benefit from having both calls here too?

MichaelChirico: based on the description of the test= argument I guess we should remove L6

MichaelChirico:
looks great! missing NEWS

aitap:
Hopefully no reverse dependency actually imported the `data.frame` S4 class from us.

PR #6847, opened 2025-03-03
Simplified and Extended "Updating by Reference" Section in Joins Vignette

### LINKED ISSUE #6846: Joins in data.table vignette. Simplify and extend the "update by reference section"
The new [Joins in data.table](https://rdatatable.gitlab.io/data.table/articles/datatable-joins.html)vignette is great, thank you! 

I think the [updating by reference](https://rdatatable.gitlab.io/data.table/articles/datatable-joins.html#updating-by-reference) section could improve by:

1.  provinding a simpler example with simple one-to-one match example. In the current example the beauty of the NO-COPY solutioun is a bit obscured by the complicated grouping ("last element selection", "tail", etc) used. 
2. provide an efficient solution for a RIGH JOIN that minimizes copying. At least in my workflow, that is the main use case. I have a main DT, on which I performe several join operations `a[DT,on=.(idA)], b[DT,on=.(idB)],...`. I assume this is a super common usecase and would be great to discuss the canonical solution to it (even if the answer is that the `x[i,on=, j =]` syntax does not allow for updating `i` by reference)

In this [SO question](https://stackoverflow.com/questions/79478214/r-data-table-update-join-by-reference-the-but-updating-the-right-table), I provided sample code for 1., with a simpler join example, and 2., with my current solution to avoid the expensive copy.

UPDATE1: the solution proposed in SO is super elegant and performant:

```
x = data.table(id = c(1:5,8), newvar1=c(LETTERS[1:5],'h'),newvar2=c(5:1,-2))
#In practice x would have more vars: newvar2, ..., newvarN 
i = data.table(id = 1:7,var1 = c('bla','ble','bli','blo','blu','blA','blS'),var2=7:1,var3=2*(7:1),var4=1)
cols <- setdiff(names(x), 'id')
i[, (cols) := x[.SD, on = "id", .SD, .SDcols = cols]]
```

This should definetivelly go on the tutorial I wish I had known this ealier. On my application ( i: 220m rows, 40ish columns. x: 130m and 5 cols), the performance gain was massive. This method is 3x faster and uses only 1/5th to 1/6th of the RAM increase relative to `i <- x[i,on=.(id)]`

[I could not find an issue specific to this vignette, just the general #944]

venom1204:
Simplified and extended the Updating by Reference section in the Joins in data.table vignette.
closes #6846
in this i 
- Added a simpler one-to-one update example for clarity.
- Provided an efficient solution for updating multiple columns using mget(cols), reducing unnecessary copies.
- Clarified RIGHT JOIN limitations and how := modifies x in place but does not update i.
- Mentioned performance improvements observed in large datasets.

### DIFF
--- a/vignettes/datatable-joins.Rmd
+++ b/vignettes/datatable-joins.Rmd
@@ -698,23 +698,38 @@ Products[!"popcorn",
 
 The `:=` operator in `data.table` is used for updating or adding columns by reference. This means it modifies the original `data.table` without creating a copy, which is very memory-efficient, especially for large datasets. When used inside a `data.table`, `:=` allows you to **add new columns** or **modify existing ones** as part of your query.
 
-Let's update our `Products` table with the latest price from `ProductPriceHistory`:
-
-```{r}
-copy(Products)[ProductPriceHistory,
-               on = .(id = product_id),
-               j = `:=`(price = tail(i.price, 1),
-                        last_updated = tail(i.date, 1)),
-               by = .EACHI][]
+1) Let's update our `Products` table with the latest price from `ProductPriceHistory`:
+```{r Simple One-to-One Update}
+Products[ProductPriceHistory, on = .(id = product_id), price := i.price]
 ```
+- The price column in Products is updated using the price column from ProductPriceHistory.
+- The on = .(id = product_id) ensures that updates happen based on matching IDs.
+- This method modifies Products in place, avoiding unnecessary copies.
 
-In this operation:
+2) If we need to get the latest price and date (instead of all matches), we can still use := efficiently:
+```{r Updating with the Latest Record}
+Products[ProductPriceHistory, 
+         on = .(id = product_id),
+         `:=`(price = last(i.price), last_updated = last(i.date)), 
+         by = .EACHI]
+```
+- last(i.price) ensures that only the latest price is selected.
+- last_updated column is added to track the last update date.
+- by = .EACHI ensures that the last price is picked for each product.
 
-- The function copy creates a ***deep*** copy of the `Products` table, preventing modifications made by `:=` from changing the original table by reference.
-- We join `Products` with `ProductPriceHistory` based on `id` and `product_id`.
-- We update the `price` column with the latest price from `ProductPriceHistory`.
-- We add a new `last_updated` column to track when the price was last changed.
-- The `by = .EACHI` ensures that the `tail` function is applied for each product in `ProductPriceHistory`.
+3) When we need to update Products with multiple columns from ProductPriceHistory
+```{r Efficient Right Join Update }
+cols <- setdiff(names(ProductPriceHistory), 'product_id')
+Products[ProductPriceHistory, 
+         on = .(id = product_id), 
+         (cols) := mget(cols)]
+
+```
+- Efficiently updates multiple columns in Products from ProductPriceHistory.
+- mget(cols) retrieves multiple matching columns dynamically.
+- This method is faster and more memory-efficient than Products <- ProductPriceHistory[Products, on=...].
+- Note: := updates Products in place, but does not modify ProductPriceHistory.
+   - Unlike traditional RIGHT JOIN, data.table does not allow i (right table) to be updated directly.
 
 ***
 


jangorecki:
I think it is best to ping author of this vignette. I haven't had yet opportunity to read the original version.

vignettes/datatable-joins.Rmd
```diff
@@ -698,23 +698,37 @@ Products[!"popcorn",
...
-- The `by = .EACHI` ensures that the `tail` function is applied for each product in `ProductPriceHistory`.
+1) Let's update our `Products` table with the latest price from `ProductPriceHistory`:
+```{r Simple One-to-One Update}
+Products[ProductPriceHistory, on = .(id = product_id), price := i.price]
+```
+- The price column in Products is updated using the price column from ProductPriceHistory.
+- The on = .(id = product_id) ensures that updates happen based on matching IDs.
+- This method modifies Products in place, avoiding unnecessary copies.
+
+2) If we need to get the latest price and date (instead of all matches), we can still use := efficiently:
```
MichaelChirico, line 709: I'm not sure we've cleanly explained the difference between this join and the previous one. The only difference I see is `tail()` vs. `last()`, am I missing something?

IIUC the difference between `tail()` and `last()` would be that `last()` can skip `NA` values, right?

venom1204: You're right, and thank you for pointing this out. The key differences are:

- The second example updates two columns (price and last_updated) instead of just price.
- It uses last(), which skips NA values, instead of directly assigning values.
 -The by = .EACHI clause ensures the operation is performed for each unique product in ProductPriceHistory.

These differences make the second approach more suitable for retrieving the latest valid (non-NA) price and date, especially when dealing with time-series data or multiple price updates per product. I'll update the vignette to clarify these points explicitly.

vignettes/datatable-joins.Rmd
```diff
@@ -702,23 +702,45 @@ Products[!"popcorn",
...
-- The function copy creates a ***deep*** copy of the `Products` table, preventing modifications made by `:=` from changing the original table by reference.
-- We join `Products` with `ProductPriceHistory` based on `id` and `product_id`.
-- We update the `price` column with the latest price from `ProductPriceHistory`.
-- We add a new `last_updated` column to track when the price was last changed.
-- The `by = .EACHI` ensures that the `tail` function is applied for each product in `ProductPriceHistory`.
+#### Let's update our `Products` table with the latest price from `ProductPriceHistory`:
+```{r Simple One-to-One Update}
+Products[ProductPriceHistory, on = .(id = product_id), price := i.price]
+```
+- The price column in Products is updated using the price column from ProductPriceHistory.
```
MichaelChirico, line 709: please remember to wrap code objects in backticks (`` ` ``)

vignettes/datatable-joins.Rmd
```diff
@@ -702,23 +702,45 @@ Products[!"popcorn",
...
+         by = .EACHI]
+```
+- last(i.price) ensures that only the latest price is selected.
+- last_updated column is added to track the last update date.
+- by = .EACHI ensures that the last price is picked for each product.
+
+#### Understanding last() vs. tail()
+
+- The key difference between last() and tail() is:
+- last(x): Returns the last element of x. Skips NAs when used on a data.table column.
```
MichaelChirico, line 727: this is not true, please test code yourself before writing it

```r
data.table::last(c(1, NA))
# [1] NA

data.table::data.table(a = c(1, 1, 2, 2), b = c(1, NA, 2, NA))[, last(b), by=a]
#        a    V1
#    <num> <num>
# 1:     1    NA
# 2:     2    NA
```

venom1204: Thanks for pointing that out, and I apologize for the mistake. You’re absolutely right—last() does not skip NA values, 
i will Clarifying the difference between last() and tail():
- last(x): Returns the last element of x, including NA if it's the last value.
- tail(x, 1): Also returns the last element but works more consistently across different object types.

and also i am planning to include examples for them like 
- Handling vectors with NA values.
- Grouped operations in data.table.
- Working with lists.
- Behavior with empty vectors.

```{r}
# Test 1: Simple vector with NA at the end
x <- c(1, 2, 3, NA)
last(x)  # Returns NA
tail(x, 1)  # Returns NA

# Test 2: data.table grouping behavior
dt <- data.table(group = c(1,1,2,2), value = c(10, NA, 20, NA))
dt[, .(last_value = last(value)), by = group]  # last() does not skip NA
dt[, .(tail_value = tail(value, 1)), by = group]  # tail() behaves similarly

# Test 3: Working with lists
l <- list(a = 1, b = 2, c = 3)
last(l)  # Returns 3
tail(l, 1)  # Returns a list of length 1

# Test 4: Empty vector behavior
z <- numeric(0)
length(last(z))  # Returns 0
length(tail(z, 1))  # Returns 0

```


should i procedd with these changes ?

venom1204: You're right, and thank you for pointing this out. The key differences are:

- The second example updates two columns (price and last_updated) instead of just price.
- It uses last(), which skips NA values, instead of directly assigning values.
 -The by = .EACHI clause ensures the operation is performed for each unique product in ProductPriceHistory.

These differences make the second approach more suitable for retrieving the latest valid (non-NA) price and date, especially when dealing with time-series data or multiple price updates per product. I'll update the vignette to clarify these points explicitly.

venom1204: @MichaelChirico any suggestions ?

venom1204:
@lucasmation @MichaelChirico
I've refined and modified the section to align with the issue requirements. Kindly review and share any suggestions when you have time.

thanks.

Advaitgaur004:
@venom1204, what is the update on this, have you tested the code .
```
data.table::last(c(1, NA))
# [1] NA

data.table::data.table(a = c(1, 1, 2, 2), b = c(1, NA, 2, NA))[, last(b), by=a]
#        a    V1
#    <num> <num>
# 1:     1    NA
# 2:     2    NA
```
If completed, it is better for @MichaelChirico to look at it

venom1204:
@MichaelChirico  I’ve minimized the changes I introduced while ensuring they contain sufficient information and remain easy to understand. Could you review them and let me know if any further simplification is needed, or if they are good as they are?

venom1204:
@iagogv3  I’ve incorporated your suggestions into my latest version. Could you review the updates and let me know if any further improvements are needed? Your feedback would be greatly appreciated! 
thank you.

vignettes/datatable-joins.Rmd
```diff
@@ -700,25 +700,77 @@ Products[!"popcorn",
...
-In this operation:
+**Grouped Updates with `.EACHI`**
+Get last price/date for each product:
+```{r Updating_with_the_Latest_Record}
+Products[ProductPriceHistory,
+         on = .(id = product_id),
+         `:=`(price = last(i.price), last_updated = last(i.date)),
+         by = .EACHI]
+```
+- `by = .EACHI` groups by i's rows (1 group per Products row).
```
iagogv3, line 724: Wouldn't be 1 group per `ProductPriceHistory` row?

vignettes/datatable-joins.Rmd
```diff
@@ -700,25 +700,77 @@ Products[!"popcorn",
...
+- Ensures unmatched values do not propagate `NA` unintentionally.
+
+**Dynamic Column Selection and Updates**
+Columns can be dynamically updated based on variable names:
+```{r}
+my_var_name <- "price"
+Products[ProductPriceHistory, on = .(id = product_id), 
+         (my_var_name) := i.price]
+```
+- This approach allows flexibility in specifying columns programmatically.
```
iagogv3, line 758: The examples in https://stackoverflow.com/a/44592473 include what I believe is another interesting point, that the `on =` option may also depend on the `my_var_name` (`on = paste0(my_var_name, "==from")`)

vignettes/datatable-joins.Rmd
```diff
@@ -700,25 +700,77 @@ Products[!"popcorn",
...
+
+**Iterating Through Multiple Columns for Updates**
+Dynamically updating multiple columns from `ProductPriceHistory`:
+```{r}
+update_cols <- intersect(c("price", "category", "stock"), names(ProductPriceHistory))
+
+for (col in update_cols) {
+  Products[ProductPriceHistory, on = .(id = product_id), (col) := get(paste0("i.", col))]}
```
iagogv3, line 766: Here may be better to use `env` than `get` as explained in https://rdatatable.gitlab.io/data.table/articles/datatable-programming.html#retired-interfaces

iagogv3:
Thank you! @venom1204 I added some comments inline.

venom1204:
@iagogv3, @MichaelChirico

I’ve updated the entire section accordingly.
When you have some time, could you please review it?

Apologies for being away for a while — I was engaged elsewhere.

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
-copy(Products)[ProductPriceHistory,
-               on = .(id = product_id),
-               j = `:=`(price = tail(i.price, 1),
-                        last_updated = tail(i.date, 1)),
-               by = .EACHI][]
+Products[ProductPriceHistory, 
+         on = .(id = product_id), 
+         price := i.price]
 ```
+- `i.price` refers to price from `i` `(ProductPriceHistory)`.
```
MichaelChirico:
```suggestion
- `i.price` refers to price from `ProductPriceHistory`.
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
+- Modifies Products in-place.
```
MichaelChirico:
```suggestion
- Modifies `Products` in-place.
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
+**Grouped Updates with `.EACHI`**
+Get last price/date for each product:
+```{r Updating_with_the_Latest_Record}
+Products[ProductPriceHistory,
+         on = .(id = product_id),
+         `:=`(price = last(i.price), last_updated = last(i.date)),
+         by = .EACHI]
+```
+- `by = .EACHI` groups by i's rows (1 group per ProductPriceHistory row).
+- `last()` returns last value including `NA`:
```
MichaelChirico:
```suggestion
- `last()` returns last value
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
 
-- The function copy creates a ***deep*** copy of the `Products` table, preventing modifications made by `:=` from changing the original table by reference.
-- We join `Products` with `ProductPriceHistory` based on `id` and `product_id`.
-- We update the `price` column with the latest price from `ProductPriceHistory`.
-- We add a new `last_updated` column to track when the price was last changed.
-- The `by = .EACHI` ensures that the `tail` function is applied for each product in `ProductPriceHistory`.
+```{r}
+data.table::last(c(1, NA))  # NA
+```
```
MichaelChirico:
```suggestion
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
+**Efficient Right Join Update**
+Add product details to `ProductPriceHistory` without copying:
+
+```{r}
+cols <- setdiff(names(Products), "id")
+ProductPriceHistory[, (cols) := 
+  Products[.SD, on = .(id = product_id), .SD, .SDcols = cols]]
+```
+- `.SD` refers to `ProductPriceHistory` during the join.
```
MichaelChirico:
```suggestion
- In `i`, `.SD` refers to `ProductPriceHistory`.
- In `j`, `.SD` refers to `Products`.
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
+- Updates `ProductPriceHistory` by reference.
+
+**Handling Edge Cases and Dynamic Column Updates**
+To dynamically update columns and handle missing values:
+```{r}
+cols <- setdiff(names(Products), "id")
+ProductPriceHistory[, (cols) := 
+  Products[.SD, on = .(id = product_id), .SD, .SDcols = cols]]
+ProductPriceHistory[is.na(price), price := 0]  # Handle missing values
```
MichaelChirico:
```suggestion
setnafill(ProductPriceHistory, fill=0, cols="price") # Handle missing values
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
+Products[ProductPriceHistory, on = .(id = product_id), 
+         (my_var_name) := i.price]
+```
+- This approach allows flexibility in specifying columns programmatically.
+
+**Iterating Through Multiple Columns for Updates**
+Dynamically updating multiple columns from `ProductPriceHistory`:
+```{r}
+update_cols <- intersect(c("price", "category", "stock"), names(ProductPriceHistory))
+
```
MichaelChirico:
```suggestion
```

vignettes/datatable-joins.Rmd
```diff
@@ -700,26 +700,82 @@ Products[!"popcorn",
...
+           (col) := i[[col]],
+           env = list(col = col)]}
+```
+- Ensures multiple columns are updated efficiently in a loop.
 
+**Summary**
+- `last(x)` vs `tail(x,1)`: Both return last element, but `tail()` returns list for lists.
+- `:=` always modifies `x`, never `i`. For right joins, update `i` directly via `i[, ... := x[.SD]]`.
+- `.EACHI` is crucial for per-row operations; simple joins use first match.
+- Note: Older functions like `mapvalues()` from the deprecated `plyr` package were previously used for recoding values. It is recommended to use data.table’s native update-join methods for efficient and future-proof code.
```
MichaelChirico:
```suggestion
```

trobx:
@MichaelChirico Sorry to chime in - I don't know whether it escaped notice (possibly down to my hopeless github skills) but I'd commented that the `by=.EACHI` update example might be better left out because:

1) the example just does what `mult` is for but very much (500x) slower
2) it's confusing imo to introduce `by=.EACHI` in the context of an update join as it ends up being effectively "`by=.EACHX`"

Mentioned at https://github.com/Rdatatable/data.table/issues/6846#issuecomment-2897501591 and the benchmark at #6997 showing a `mult`/`fifelse` solution that takes ~0.5 secs vs ~2.5 mins for `tail`/`by=.EACHI`.

MichaelChirico:
> @MichaelChirico Sorry to chime in - I don't know whether it escaped notice (possibly down to my hopeless github skills) but I'd commented that the `by=.EACHI` update example might be better left out because:
> 
> 1. the example just does what `mult` is for but very much (500x) slower
> 2. it's confusing imo to introduce `by=.EACHI` in the context of an update join as it ends up being effectively "`by=.EACHX`"
> 
> Mentioned at [#6846 (comment)](https://github.com/Rdatatable/data.table/issues/6846#issuecomment-2897501591) and the benchmark at #6997 showing a `mult`/`fifelse` solution that takes ~0.5 secs vs ~2.5 mins for `tail`/`by=.EACHI`.

could you file a new issue & ideally a follow-up PR? thank you!

Issue #6852, opened 2025-03-05
fwrite() emits invalid gzips when `col.names = FALSE` as of 1.17.0


maxscheiber:
This reproduces on both macOS 14.7.1 and on Ubuntu 22.04.

### [`Minimal reproducible example`](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example); please be sure to set `verbose=TRUE` where possible!
```
> library(data.table)
data.table 1.17.0 using 6 threads (see ?getDTthreads).  Latest news: r-datatable.com
> options(datatable.verbose = TRUE)
> state_dt = data.table(state = state.name, code = state.abb)
> data.table::fwrite(state_dt, '~/Desktop/state.gzip', col.names = FALSE, compress = 'gzip')
  OpenMP version (_OPENMP)       201811
  omp_get_num_procs()            12
  R_DATATABLE_NUM_PROCS_PERCENT  unset (default 50)
  R_DATATABLE_NUM_THREADS        unset
  R_DATATABLE_THROTTLE           unset (default 1024)
  omp_get_thread_limit()         2147483647
  omp_get_max_threads()          12
  OMP_THREAD_LIMIT               unset
  OMP_NUM_THREADS                unset
  RestoreAfterFork               true
  data.table is using 6 threads with throttle==1024. See ?setDTthreads.
No list columns are present. Setting sep2='' otherwise quote='auto' would quote fields containing sep2.
Column writers: 12 12
args.doRowNames=0 args.rowNames=0x0 args.rowNameFun=0 doQuote=-128 args.nrow=50 args.ncol=2 eolLen=1
maxLineLen=39. Found in 0.000s
Writing bom (false), yaml (0 characters) and column names (false)
Writing 50 rows in 1 batches of 50 rows, each buffer size 8388608 bytes (8 MiB), showProgress=1, nth=1
Allocate 8388608 bytes (8 MiB) for buffPool
zbuffSize=8391175 returned from deflateBound
Allocate 8391175 bytes (8 MiB) for zbuffPool
Initialization done in 0.001s
zlib: uncompressed length=6103596894 (5820 MiB), compressed length=424 (0 MiB), ratio=0.0%, crc=1b165bf5
Wrote 50 rows in 0.000 secs using 1 thread. MaxBuffUsed=0%
> invisible(data.table::fread('~/Desktop/state.gzip'))
  OpenMP version (_OPENMP)       201811
  omp_get_num_procs()            12
[...]

PR #6853, opened 2025-03-06
`fwrite()`: when `col.names=FALSE`, produce a `gzip` header, avoid a leak

aitap:
`gzip` header generation was previously tied to header printing, together with some initialization (like `crc`, `len`) and de-initialization (like `deflateEnd(&strm)`). Moved it outside the `headerLen` branch.

Fixes: #6852.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -10024,9 +10024,12 @@ if (!haszlib()) {
   fwrite(DT, file=f3<-tempfile(), compress="gzip")   # compress to filename not ending .gz
   fwrite(DT, file=f4<-tempfile(), compress="gzip", compressLevel=1)   # test compressLevel
   fwrite(DT, file=f5<-tempfile(), compress="gzip", compressLevel=9)
+  # col.names=FALSE must not disable gzip header, #6852
+  fwrite(DT, file=f6<-tempfile(), compress="gzip", col.names=FALSE)
   test(1658.441, file.info(f3)$size, file.info(f1)$size)
   test(1658.442, file.info(f4)$size >= file.info(f1)$size)
   test(1658.443, file.info(f1)$size >= file.info(f5)$size)
+  test(1658.444, fread(f6, col.names = c("a", "b")), DT)
   unlink(c(f1,f2,f3,f4,f5))
 }
 DT = data.table(a=1:3, b=list(1:4, c(3.14, 100e10), c("foo", "bar", "baz")))
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -10030,7 +10030,7 @@ if (!haszlib()) {
   test(1658.442, file.info(f4)$size >= file.info(f1)$size)
   test(1658.443, file.info(f1)$size >= file.info(f5)$size)
   test(1658.444, fread(f6, col.names = c("a", "b")), DT)
-  unlink(c(f1,f2,f3,f4,f5))
+  unlink(c(f1,f2,f3,f4,f5,f6))
 }
 DT = data.table(a=1:3, b=list(1:4, c(3.14, 100e10), c("foo", "bar", "baz")))
 test(1658.45, fwrite(DT), output=c("a,b","1,1|2|3|4","2,3.14|1e+12","3,foo|bar|baz"))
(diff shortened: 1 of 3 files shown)


MichaelChirico:
NB: this will be patched into 1.17.2:

https://github.com/Rdatatable/data.table/blob/6a8634e0ff8700f1a037c107b8507dce14e5da7d/.dev/CRAN_Release.cmd#L620-L628

inst/tests/tests.Rraw
```diff
@@ -10024,9 +10024,12 @@ if (!haszlib()) {
   fwrite(DT, file=f3<-tempfile(), compress="gzip")   # compress to filename not ending .gz
   fwrite(DT, file=f4<-tempfile(), compress="gzip", compressLevel=1)   # test compressLevel
   fwrite(DT, file=f5<-tempfile(), compress="gzip", compressLevel=9)
+  # col.names=FALSE must not disable gzip header, #6852
+  fwrite(DT, file=f6<-tempfile(), compress="gzip", col.names=FALSE)
   test(1658.441, file.info(f3)$size, file.info(f1)$size)
   test(1658.442, file.info(f4)$size >= file.info(f1)$size)
   test(1658.443, file.info(f1)$size >= file.info(f5)$size)
+  test(1658.444, fread(f6, col.names = c("a", "b")), DT)
   unlink(c(f1,f2,f3,f4,f5))
```
MichaelChirico, line 10033:
```suggestion
  unlink(c(f1,f2,f3,f4,f5,f6))
```

MichaelChirico: (GitHub doesn't like applying suggestions to this huge file)

src/fwrite.c
```diff
@@ -829,8 +829,27 @@ void fwriteMain(fwriteMainArgs args)
...
+        CLOSE(f);
+        free(buffPool);
+        free(zbuffPool);
+        deflateEnd(&strm);
+        STOP(_("Failed to write gzip header. Write returned %d"), ret0);
+        // # nocov end
+      }
+    }
+  }
 #endif
```
MichaelChirico, line 853:
```suggestion
#endif // #NOZLIB
```

src/fwrite.c
```diff
@@ -829,9 +829,28 @@ void fwriteMain(fwriteMainArgs args)
...
-#endif
+    if (f != -1) {
+      // write minimal gzip header, but not on the console
+      static const char header[] = "\037\213\10\0\0\0\0\0\0\3";
+      int ret0 = WRITE(f, header, (sizeof header) - 1);
+      compress_len += (sizeof header) - 1;
+
+      if (ret0 == -1) {
+        // # nocov start
+        int errwrite = errno; // capture write errno now in case close fails with a different errno
```
MichaelChirico, line 843: we can remove this right?

```suggestion
```

aitap: Good catch! Made use of the `errno` instead of just saying that `write()` returned -1. Alternatively we may change the translation template and make use of `strerror()` as below:
https://github.com/Rdatatable/data.table/blob/551aa8737f9dc4a7780c707de443422467a7322b/src/fwrite.c#L924

MichaelChirico: LGTM for now

MichaelChirico (approved):
Thanks for getting this done!

src/fwrite.c
```diff
@@ -898,21 +912,26 @@ void fwriteMain(fwriteMainArgs args)
...
 #endif
-        if (ret0 == -1) STOP(_("Failed to write gzip header. Write returned %d"), ret0);
-        else if (ret1) STOP(_("Failed to compress gzip. compressbuff() returned %d"), ret1);
+        if (ret1) STOP(_("Failed to compress gzip. compressbuff() returned %d"), ret1);
         else STOP(_("%s: '%s'"), strerror(errwrite), args.filename);
         // # nocov end
       }
     }
   }
+#ifndef NOZLIB
```
ben-schwen, line 929: Interestingly #6857 is not an issue for me when deleting this block 😕

ben-schwen: And if `args.is_gzip` is not `true` then we never call `init_stream` so this is actually not surprising anymore.

PR #6856, opened 2025-03-08
fread parse dates with leading zeros and keepLeadingZeros=TRUE

### LINKED ISSUE #6851: fread() ignores colClasses Date type and read column as character instead.
Issue in essence
```
df = fread("bad.csv", colClasses = c("lc_score_date" = "Date"), sep = "\t", quote = "", verbose = T)
class(df$lc_score_date)
[1] "character"
```

This example code in older versions of data.table returns dataframe with column of type `Date`. 
`is.numeric(unclass(df$lc_score_date))` returns `TRUE` in older data.table packages.
In `data.table_1.17.0` fread returns dataframe with character column instead of data. 
```
library(data.table)

options(datatable.na.strings = "\\N")
options(datatable.fread.datatable=FALSE)
options(datatable.keepLeadingZeros=TRUE)
options(stringsAsFactors=FALSE)

df = fread("bad.csv", colClasses = c("lc_score_date" = "Date"), sep = "\t", quote = "", verbose = T)

print(paste0("is.numeric(unclass(df$lc_score_date)): ", is.numeric(unclass(df$lc_score_date))))
```
output of verbose fread
```
data.table 1.17.0 using 2 threads (see ?getDTthreads).  Latest news: r-datatable.com
  OpenMP version (_OPENMP)       201511
  omp_get_num_procs()            4
  omp_get_thread_limit()         2147483647
  omp_get_max_threads()          4
  data.table is using 2 threads with throttle==1024. See ?setDTthreads.
freadR.c has been passed a filename: bad.csv
[01] Check arguments
  Using 2 threads (omp_get_max_threads()=4, nth=2)
  NAstrings = [<<\N>>]
  show progress = 1
  0/1 column will be read as integer
  Y/N column will be read as character
[02] Opening the file
  Opening file bad.csv
  File opened, size = 5.406KB (5536 bytes).
  Memory mapped ok
[03] Detect and skip BOM
  \n has been found in the input and different lines can end with different line endings (e.g. mixed \n and \r\n in one file). This is common and ideal.
  Positioned on line 1 starting: <<lc_score_date>>
  Using supplied sep '\t'
  No sep and quote rule found a block of 2x2 or greater. Single column input.
  Detected 1 columns on line 1. This line is either column names or first data row. Line starts as: <<lc_score_date>>
  Quote rule picked = 0
  fill=false and the most number of columns found is 1
[07] Detect column types, dec, good nrow estimate and whether first row is column names
  Number of sampling jump points = 1 because (5535 bytes from row 1 to eof) / (2 * 316 jump0size) == 8
  dec='.' detected based on a balance of 0 parsed fields
  Type codes (jump 000)    : C  Quote rule 0
  Type codes (jump 001)    : C  Quote rule 0
  'header' determined to be true due to column 1 containing a string on row 1 and a lower type (int32) in the rest of the 152 sample rows
  =====
  Sampled 152 rows (handled \n inside quoted fields) at 2 jump points
  Bytes from first data row on line 2 to the end of last row: 5521
  Line length: mean=3.05 sd=0.65 min=2 max=11
  Estimated number of rows: 5521 / 3.05 = 1813
  Initial alloc = 2760 rows (1813 + 52%) using bytes/max(mean-2*sd,min) clamped between [1.1*estn, 2.0*estn]
  =====
[08] Assign column names
  After 0 type and 0 drop user overrides : C
  Allocating 1 column slots (1 - 0 dropped) with 2760 rows
[11] Read the data
  jumps=[0..1), chunk_size=1048576, total_size=5521
  1 out-of-sample type bumps: E
  jumps=[0..1), chunk_size=1048576, total_size=5521
Read 1798 rows x 1 columns from 5.406KB (5536 bytes) file in 00:00.000 wall clock time
[12] Finalizing the datatable
  Type counts:
         1 : string    'E'
=============================
   0.000s ( 24%) Memory map 0.000GB file
   0.000s ( 34%) sep='' ncol=1 and header detection
   0.000s (  4%) Column type detection using 152 sample rows
   0.000s (  3%) Allocation of 2760 rows x 1 cols (0.000GB) of which 1798 ( 65%) rows used
   0.000s ( 36%) Reading 1 chunks (0 swept) of 1.000MB (each chunk 1798 rows) using 1 threads
   +    0.000s ( 24%) Parse to row-major thread buffers (grown 0 times)
   +    0.000s (  3%) Transpose
   +    0.000s (  8%) Waiting
   0.000s ( 16%) Rereading 1 columns due to out-of-sample type exceptions
   0.000s        Total
Column 1 <<lc_score_date>> bumped from 'int32' to 'string' due to <<2014-12-05>> on row 118
> class(df$lc_score_date)
[1] "character"
```
Line `bumped from 'int32' to 'string' due to <<2014-12-05>> on row 118` in verbose output pretty much tells fread decided to read it as string. 

sessionInfo() output
```
sessionInfo()
R version 4.2.2 Patched (2022-11-10 r83330)
Running under: Ubuntu 22.04.1 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so

locale:
 [1] LC_CTYPE=C.UTF-8       LC_NUMERIC=C           LC_TIME=C.UTF-8
 [4] LC_COLLATE=C.UTF-8     LC_MONETARY=C.UTF-8    LC_MESSAGES=C.UTF-8
 [7] LC_PAPER=C.UTF-8       LC_NAME=C              LC_ADDRESS=C
[10] LC_TELEPHONE=C         LC_MEASUREMENT=C.UTF-8 LC_IDENTIFICATION=C

attached base packages:

other attached packages:
[1] data.table_1.17.0

[1] compiler_4.2.2
```

There is nothing special in tsv file, its just NA values as `\N` and dates. It's sparse though. 

[bad.csv](https://github.com/user-attachments/files/19095858/bad.csv)

ben-schwen:
Closes #6851

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21079,3 +21079,8 @@ setindex(DT, b)
 # make sure that print(DT) doesn't warn due to the header missing index column types, #6806
 # can't use output= here because the print() call is outside withCallingHandlers(...)
 test(2307, { capture.output(print(DT, class = TRUE, show.indices = TRUE)); TRUE })
+
+# fread with colClasses and keepLeadingZeros=TRUE #6851
+dt = data.table(date=as.IDate(c(NA, "2014-12-05")))
+test(2308.01, fread("date\nNA\n2014-12-05", keepLeadingZeros=TRUE), dt)
+test(2308.02, fread("date\nNA\n2014-12-05", keepLeadingZeros=FALSE), dt)
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,7 @@
 
 2. `fwrite(compress="gzip")` once again produces a gzip header when the column names are missing or disabled, [@6852](https://github.com/Rdatatable/data.table/issues/6852). Thanks @maxscheiber for the report and @aitap for the fix.
 
-3. `fread(keepLeadingZeros=TRUE)` parsed dates with a leading zero as strings although `fread` correctly detected them as dates. Thanks @TurnaevEvgeny for the report and @ben-schwen for the fix.
+3. `fread(keepLeadingZeros=TRUE)` now correctly parses dates with leading zeros as dates instead of strings, [#6851](https://github.com/Rdatatable/data.table/issues/6851). Thanks @TurnaevEvgeny for the report and @ben-schwen for the fix.
 
 ## NOTES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,8 @@
 
 2. `fwrite(compress="gzip")` once again produces a gzip header when the column names are missing or disabled, [@6852](https://github.com/Rdatatable/data.table/issues/6852). Thanks @maxscheiber for the report and @aitap for the fix.
 
+3. `fread(keepLeadingZeros=TRUE)` parsed dates with a leading zero as strings although `fread` correctly detected them as dates. Thanks @TurnaevEvgeny for the report and @ben-schwen for the fix.
+
 ## NOTES
 
 1. Continued work to remove non-API C functions, [#6180](https://github.com/Rdatatable/data.table/issues/6180). Thanks Ivan Krylov for the PRs and for writing a clear and concise guide about the R API: https://aitap.codeberg.page/R-api/.
--- a/src/fread.c
+++ b/src/fread.c
@@ -580,11 +580,11 @@ static void Field(FieldParseContext *ctx)
   }
 }
 
-static void str_to_i32_core(const char **pch, int32_t *target)
+static void str_to_i32_core(const char **pch, int32_t *target, bool parse_date)
 {
   const char *ch = *pch;
 
-  if (*ch=='0' && args.keepLeadingZeros && IS_DIGIT(ch[1])) return;
+  if (*ch=='0' && args.keepLeadingZeros && IS_DIGIT(ch[1]) && !parse_date) return;
   bool neg = *ch=='-';
   ch += (neg || *ch=='+');
   const char *start = ch;  // to know if at least one digit is present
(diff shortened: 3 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -21079,3 +21079,8 @@ setindex(DT, b)
 # make sure that print(DT) doesn't warn due to the header missing index column types, #6806
 # can't use output= here because the print() call is outside withCallingHandlers(...)
 test(2307, { capture.output(print(DT, class = TRUE, show.indices = TRUE)); TRUE })
+
+# fread with colClasses and keepLeadingZeros=TRUE #6851
+dt = data.table(date=as.IDate(c(NA, "2014-12-05")))
+test(2308.01, fread("date\nNA\n2014-12-05", keepLeadingZeros=TRUE), dt)
```
aitap, line 21085: Interesting that this test request fails the same way in 6c69c15801474eb5a550c6f3233e50b9a835002f (immediately after `1.16.0`) as in `master`, but the problem was masked for the example from #6851 until #6107.

aitap (approved):

inst/tests/tests.Rraw
MichaelChirico: Do you think 'regression' is still the right way to categorize this issue?

BTW, this test and the NEWS bullet don't add up for me -- neither of these tests have any leading zeroes.

Maybe the bullet needs some refinement, or perhaps an example?

ben-schwen: They have leading zeros since year, month and date are parsed separately.

This is also the reason why the first date in the bad.csv in the issue isn't triggering it

Issue #6857, opened 2025-03-09
revdeps segfault after fwrite modifications


tdhock:
revdep check machine says that https://github.com/Rdatatable/data.table/pull/6853 is causing a segfault in a few different packages.

https://rcdata.nau.edu/genomic-ml/data.table-revdeps/analyze/2025-03-09/PeakSegDisk.txt
```
* checking tests ...
  Running 'testthat.R'/home/th798/R/R-devel/bin/BATCH: line 60: 370750 Segmentation fault      (core dumped) ${R_HOME}/bin/R -f ${in} ${opts} ${R_BATCH_OPTIONS} > ${out} 2>&1

 ERROR
Running the tests in 'tests/testthat.R' failed.
Complete output:
  > if(require("testthat")){
  +   suite <- Sys.getenv("TEST_SUITE")
  +   if(suite=="")suite <- NULL
  +   test_check("PeakSegDisk", filter=suite)
  + }
  Loading required package: testthat
  Loading required package: PeakSegDisk
  
   *** caught segfault ***
  address 0x17f2785d0, cause 'memory not mapped'
  
  Traceback:
   1: fwrite(cov.dt, coverage.bedGraph, sep = "\t", row.names = FALSE,     col.names = FALSE)
   2: eval(code, test_env)
   3: eval(code, test_env)
   4: withCallingHandlers({    eval(code, test_env)    if (!handled && !is.null(test)) {        skip_empty()    }}, expectation = handle_expectation, skip = handle_skip, warning = handle_warning,     message = handle_message, error = handle_error)
   5: doTryCatch(return(expr), name, parentenv, handler)
   6: tryCatchOne(expr, names, parentenv, handlers[[1L]])
   7: tryCatchList(expr, names[-nh], parentenv, handlers[-nh])
   8: doTryCatch(return(expr), name, parentenv, handler)
   9: tryCatchOne(tryCatchList(expr, names[-nh], parentenv, handlers[-nh]),     names[nh], parentenv, handlers[[nh]])
  10: tryCatchList(expr, classes, parentenv, handlers)
  11: tryCatch(withCallingHandlers({    eval(code, test_env)    if (!handled && !is.null(test)) {        skip_empty()    }}, expectation = handle_expectation, skip = handle_skip, warning = handle_warning,     message = handle_message, error = handle_error), error = handle_fatal,     skip = function(e) {    })
  12: test_code(test = NULL, code = exprs, env = env, default_reporter = StopReporter$new())
[...]

PR #6858, opened 2025-03-10
fwrite: avoid deflateEnd() for uninitialized strm

aitap:
As diagnosed by @ben-schwen, when `!args.is_gzip`, `strm` is not initialized, so calling `deflateEnd()` on it is a mistake. Sorry about it.

Fixes: #6857

Tested using `R CMD check PeakSegDisk*.tar.gz` after observing the crash without the fix and by running `fwrite(list(1:100), "/dev/null")` under Valgrind (which knows about undefined values and diagnoses the problem immediately). Unfortunately, Valgrind is too slow to run everything in it.

### DIFF
--- a/src/fwrite.c
+++ b/src/fwrite.c
@@ -927,8 +927,8 @@ void fwriteMain(fwriteMainArgs args)
     }
   }
 #ifndef NOZLIB
-  else {
-    // was unconditionally initialized for zbuffSize, not used for header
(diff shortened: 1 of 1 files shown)


ben-schwen:
LGTM TY!

MichaelChirico:
> Sorry about it.

no problem at all, thank you to @tdhock for the revdep system to catch this long before release!

PR #6865, opened 2025-03-12
Ensure Consistent Key Handling in as.data.table S3 Methods

### LINKED ISSUE #6859: `key(as.data.table(x, key = <key>))` is inconsistent, varies based on `x`
- `key(as.data.table(x, key = <key>))` is not `<key>` if `x` is
  - a tibble
  - a `data.table` not already keyed by `<key>`.
- `key(as.data.table(x))` is not `NULL` if `x` is a keyed `data.table`.

``` r
suppressPackageStartupMessages({
  library(data.table)
  library(dplyr)
})

# One probably expects key="t" to produce a key of "t":
data.frame(t = c(3:1,4:5), y = 1:5) %>% as.data.table(key = "t") %>% key()
#> [1] "t"
tibble(t = c(3:1,4:5), y = 1:5) %>% as.data.table(key = "t") %>% key()
#> NULL
data.table(t = c(3:1,4:5), y = 1:5) %>% as.data.table(key = "t") %>% key()
#> NULL

# One may also expect omitting `key =` arg to produce unkeyed output; however, 
# that's also not always the case.  The "proper"/expected output is probably
# more up for debate here than above, though.
data.table(t = c(3:1,4:5), key = "t") %>%
  as.data.table() %>%
  key()
#> [1] "t"
```

`#` `Output of sessionInfo()`

```r
sessionInfo()
#> R version 4.4.2 (2024-10-31)
#> Platform: x86_64-pc-linux-gnu
#> Running under: openSUSE Tumbleweed
#> 
#> Matrix products: default
#> BLAS:   /home/fullname/files/sw2/R-4.4.2/lib/libRblas.so 
#> LAPACK: /home/fullname/files/sw2/R-4.4.2/lib/libRlapack.so;  LAPACK version 3.12.0
#> 
#> locale:
#> 
#> time zone: America/Los_Angeles
#> tzcode source: system (glibc)
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] dplyr_1.1.4        data.table_1.17.99
#> 
#>  [1] digest_0.6.37     R6_2.6.1          fastmap_1.2.0     tidyselect_1.2.1 
#>  [5] xfun_0.51         magrittr_2.0.3    glue_1.8.0        tibble_3.2.1     
#>  [9] knitr_1.49        pkgconfig_2.0.3   htmltools_0.5.8.1 generics_0.1.3   
#> [13] rmarkdown_2.29    lifecycle_1.0.4   cli_3.6.4         vctrs_0.6.5      
#> [17] reprex_2.1.1      withr_3.0.2       compiler_4.4.2    tools_4.4.2      
#> [21] pillar_1.10.1     evaluate_1.0.3    yaml_2.3.10       rlang_1.1.5      
#> [25] fs_1.6.5
```

<sup>Created on 2025-03-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

`#` Related issues:
- https://github.com/Rdatatable/data.table/issues/5699 : this was about excluding attributes from non-data.table data.frame extensions in the result of as.data.table().  The fix for that issue likely introduced the as.data.table() behavior on tibbles above; [this line](https://github.com/Rdatatable/data.table/blob/11aef2fbc82d305c26540ae28adb085431f9595d/R/as.data.table.R#L218) may just need to forward args to the `as.data.table()` call.

Mukulyadav2004:
Closes #6859 
This PR standardizes key assignment in as.data.table(), ensuring:

->key(as.data.table(x, key = <key>)) always returns <key>.
->Existing keys are cleared when key = NULL.

Changes:
->Added key = key in as.data.table.data.frame for nested conversions.
->Applied keys only when explicitly requested (if (!is.null(key)) setkeyv(ans, key)).
->Ensured key application to rn before renaming.
->Standardized as.data.table.data.table: key = NULL, keep.rownames = FALSE to prevent unexpected key retention.
->Ensured explicit key assignment: setkeyv(x, key) when provided, setkeyv(x, NULL) when NULL.

@ben-schwen , please review when you have time. Thanks!


R/as.data.table.R
```diff
@@ -214,21 +214,22 @@ as.data.table.list = function(x,
 }
 
 as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
-  if (is.data.table(x)) return(as.data.table.data.table(x)) # S3 is weird, #6739. Also # nocov; this is tested in 2302.{2,3}, not sure why it doesn't show up in coverage.
-  if (!identical(class(x), "data.frame")) return(as.data.table(as.data.frame(x)))
+  if (is.data.table(x)) return(as.data.table.data.table(x,keep.rownames = keep.rownames, key = key, ...)) # S3 is weird, #6739. Also # nocov; this is tested in 2302.{2,3}, not sure why it doesn't show up in coverage.
```
ben-schwen, line 217: data.table don't have rownames so we probably don't need an argument for keeping them

R/as.data.table.R
```diff
@@ -214,21 +214,22 @@ as.data.table.list = function(x,
...
+  if (!identical(class(x), "data.frame")) return(as.data.table(as.data.frame(x), keep.rownames=keep.rownames, key=key, ...))
   if (!isFALSE(keep.rownames)) {
     # can specify col name to keep.rownames, #575; if it's the same as key,
     #   kludge it to 'rn' since we only apply the new name afterwards, #4468
     if (is.character(keep.rownames) && identical(keep.rownames, key)) key='rn'
-    ans = data.table(rn=rownames(x), x, keep.rownames=FALSE, key=key)
+    ans = data.table(rn=rownames(x), x, keep.rownames=FALSE)
+    if (!is.null(key)) setkeyv(ans, key)
```
ben-schwen, line 224: why change the original behavior?

R/as.data.table.R
```diff
@@ -241,17 +242,22 @@ as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
...
+    return(as.data.table.list(x, keep.rownames = keep.rownames, key = key, ...))
   }
   x = copy(x) # #1681
   # fix for #1078 and #1128, see .resetclass() for explanation.
   setattr(x, 'class', .resetclass(x, "data.table"))
+  if (!is.null(key)) {
+    setkeyv(x, key)
+  } else {
+    setkeyv(x, NULL)
+  }
```
ben-schwen, line 261:
```suggestion
  if (!missing(key)) setkeyv(x, key)
```
is much simpler and should have the same behavior

R/as.data.table.R
```diff
@@ -241,17 +242,22 @@ as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   # fix for #1078 and #1128, see .resetclass() for explanation.
   setattr(ans, "class", .resetclass(x, "data.frame"))
   setalloccol(ans)
-  setkeyv(ans, key)
+  if (!is.null(key)) setkeyv(ans, key)
   ans
 }
 
-as.data.table.data.table = function(x, ...) {
+as.data.table.data.table = function(x, keep.rownames = FALSE, key = NULL, ...) {
```
ben-schwen, line 249: Given a default value to `key` might introduce breaking behavior.

Ideally `as.data.table.data.table` follows the following simple rules:
`as.data.table.data.table(keyed-dt)` returns a keyed data.table
`as.data.table.data.table(unkeyed-dt` returns a data.table without a key

Mukulyadav2004:
Hi @ben-schwen ,
I sincerely apologize for the excess commits on this pull request, especially around the test cases. I’ve cleaned up any unnecessary changes and would greatly appreciate your review when you have time. Thank you for your patience and understanding.

R/as.data.table.R
```diff
@@ -241,17 +241,19 @@ as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   # fix for #1078 and #1128, see .resetclass() for explanation.
   setattr(ans, "class", .resetclass(x, "data.frame"))
   setalloccol(ans)
-  setkeyv(ans, key)
+  if (!is.null(key)) setkeyv(ans, key)
   ans
 }
 
 as.data.table.data.table = function(x, ...) {
+  key = list(...)$key
```
ben-schwen, line 249: why no extra argument for `key` but packing it into ellipsis?

inst/tests/tests.Rraw
```diff
@@ -21084,3 +21084,23 @@ test(2307, { capture.output(print(DT, class = TRUE, show.indices = TRUE)); TRUE
...
 test(2308.01, fread("date\nNA\n2014-12-05", keepLeadingZeros=TRUE), dt)
 test(2308.02, fread("date\nNA\n2014-12-05", keepLeadingZeros=FALSE), dt)
+
+# Test that as.data.table.data.table preserves key when explicitly specified but not when omitted
+# data.frame to data.table with key
+test(2309.01, {
+  key(as.data.table(data.frame(t = c(3:1, 4:5), y = 1:5), key="t"))
+}, "t")
+# tibble to data.table with key
+DF <- data.frame(t = c(3:1, 4:5), y = 1:5)
```
ben-schwen, line 21094: DF = data.frame(t = c(3:1, 4:5), y = 1:5)
test(2309.01, key(as.data.table(DF, key="t")), "t")

ben-schwen, line 21094:
```suggestion
DF = data.frame(t = c(3:1, 4:5), y = 1:5)
test(2309.01, key(as.data.table(DF, key="t")), "t")
```

Please keep to our style of tests, e.g. setting it up and then test on a single line. Also keep in mind that we do not use `<-` when not absolutely necessary.

R/as.data.table.R
Mukulyadav2004: I had initially considered adding an explicit argument for key, but in our previous discussion, you mentioned that giving a default value to key might introduce breaking behavior. Based on that feedback, I packed it into ... to avoid any unintended issues. However, I may have misunderstood your intent—if adding an explicit key argument is preferable after all, I’ll adjust accordingly. Could you please clarify?"

ben-schwen: Previously you had `key=NULL` giving it the default value of NULL and making it hard to distinguished if key is NULL because someone provided NULL as argument or because it is the default value.

NEWS.md
```diff
@@ -14,6 +14,8 @@
 
 3. `fread(keepLeadingZeros=TRUE)` now correctly parses dates with components with leading zeros as dates instead of strings, [#6851](https://github.com/Rdatatable/data.table/issues/6851). Thanks @TurnaevEvgeny for the report and @ben-schwen for the fix.
 
+4. `as.data.table()` now properly handles keys: explicitly specified keys are set, while unspecified keys are cleared (returning NULL from `key()`), [#6859](https://github.com/Rdatatable/data.table/issues/6859). Thanks @brookslogan for the report and @Mukulyadav2004 for the fix.
```
ben-schwen, line 17:
```suggestion
4. `as.data.table()` now properly handles keys: specifying keys sets them, omitting keys preserves existing ones, and setting `key=NULL` clears them, [#6859](https://github.com/Rdatatable/data.table/issues/6859). Thanks @brookslogan for the report and @Mukulyadav2004 for the fix.
```

R/as.data.table.R
```diff
@@ -245,13 +245,15 @@ as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   ans
 }
 
-as.data.table.data.table = function(x, ...) {
+as.data.table.data.table = function(x, keep.rownames, key, ...) {
+  if (missing(key))  key = NULL
```
ben-schwen, line 249:
```suggestion
as.data.table.data.table = function(x, keep.rownames, key=NULL, ...) {

```

R/as.data.table.R
```diff
@@ -245,13 +245,15 @@ as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   ans
 }
 
-as.data.table.data.table = function(x, ...) {
+as.data.table.data.table = function(x, keep.rownames, key=NULL, ...) {
+
```
ben-schwen, line 249:
```suggestion
```

inst/tests/tests.Rraw
```diff
@@ -21084,3 +21084,18 @@ test(2307, { capture.output(print(DT, class = TRUE, show.indices = TRUE)); TRUE
...
 test(2308.01, fread("date\nNA\n2014-12-05", keepLeadingZeros=TRUE), dt)
 test(2308.02, fread("date\nNA\n2014-12-05", keepLeadingZeros=FALSE), dt)
+
+# Test that as.data.table.data.table preserves key when explicitly specified but not when omitted
+# data.frame to data.table with key
+DF = data.frame(t = c(3:1, 4:5), y = 1:5)
+test(2309.01, key(as.data.table(DF, key="t")), "t")
+# tibble to data.table with key
+DF <- data.frame(t = c(3:1, 4:5), y = 1:5)
+class(DF) <- c("tbl_df", "tbl", "data.frame")
```
ben-schwen, line 21094:
```suggestion
class(DF) = c("tbl_df", "tbl", "data.frame")
```

inst/tests/tests.Rraw
```diff
@@ -21084,3 +21084,18 @@ test(2307, { capture.output(print(DT, class = TRUE, show.indices = TRUE)); TRUE
 dt = data.table(date=as.IDate(c(NA, "2014-12-05")))
 test(2308.01, fread("date\nNA\n2014-12-05", keepLeadingZeros=TRUE), dt)
 test(2308.02, fread("date\nNA\n2014-12-05", keepLeadingZeros=FALSE), dt)
+
+# Test that as.data.table.data.table preserves key when explicitly specified but not when omitted
+# data.frame to data.table with key
+DF = data.frame(t = c(3:1, 4:5), y = 1:5)
+test(2309.01, key(as.data.table(DF, key="t")), "t")
+# tibble to data.table with key
+DF <- data.frame(t = c(3:1, 4:5), y = 1:5)
```
ben-schwen, line 21093:
```suggestion
```

Unnecessary line, since we already create DF before

inst/tests/tests.Rraw
```diff
@@ -21084,3 +21084,18 @@ test(2307, { capture.output(print(DT, class = TRUE, show.indices = TRUE)); TRUE
...
+class(DF) <- c("tbl_df", "tbl", "data.frame")
+test(2309.02, key(as.data.table(DF, key="t")), "t")
+# data.table to data.table with key
+DT = data.table(t = c(3:1, 4:5), y = 1:5)
+test(2309.03, key(as.data.table(DT, key="t")), "t")
+# data.table to data.table without key
+DT = data.table(t = c(3:1, 4:5), key="t")
+test(2309.04, key(as.data.table(DT)), NULL)
```
ben-schwen, line 21101: Actually we have three cases:
For data.table keyed with "b":
* `as.data.table(key="a")` should set "a" as key
* `as.data.table()` should leave "b" as key intact
* `as.data.table(key=NULL)` should remove the key
For a non-keyed data.table:
* `as.data.table(key="a")` should set "a" as key
* `as.data.table()` should leave the data.table unkeyed
* `as.data.table(key=NULL)` should leave the data.table unkeyed

R/as.data.table.R
```diff
@@ -245,13 +245,14 @@ as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   ans
 }
 
-as.data.table.data.table = function(x, ...) {
+as.data.table.data.table = function(x, keep.rownames, key=NULL, ...) {
```
ben-schwen, line 248:
```suggestion
as.data.table.data.table = function(x, key=NULL, ...) {
```
As mentioned above we don't need a keep.rownames argument for data.table. While you could theoretically set them, we don't even print them

Mukulyadav2004: I understand the reasoning behind simplifying the signature. However, removing keep.rownames from as.data.table.data.table() creates an S3 method mismatch with as.data.table(), which formally declares ...(x, keep.rownames,...). This inconsistency leads to a 'formal argument' warning in R due to the differing signatures. While keep.rownames may not be commonly used with a data.table, it remains necessary to maintain correct dispatch consistency

Mukulyadav2004: I haven't figured that out yet. I'll run the tests again to double-check whether the warning occurs as I described, and I'll get back to you with my findings

Mukulyadav2004: I rechecked this, and the warning does occur as I described. Based on my research, the mismatch was previously masked by the use of ..., which prevented R’s checks from flagging it. Once keep.rownames was made explicit, R recognized the discrepancy between the S3 generic and method, as well as the documentation mismatch where the argument name and default values don’t align with the generic.

aitap: It's probably [due to argument order](https://github.com/r-devel/r-svn/blob/2c9a39e7d042f31c468da8f82ab7a4edb475b8dd/src/library/tools/R/QC.R#L2675-L2680). The `as.data.table.data.table = function(x, ...)` method signature fits the `as.data.table = function(x, keep.rownames=FALSE, ...)` generic because `...` doesn't contradict the `keep.rownames` argument. On the other hand, `function(x, key, ...)` would contradict the generic signature because `key` goes first.

MichaelChirico: That sounds right. Moreover, adding new arguments before `...` could constitute a breaking change for any downstreams relying on positional matching, c.f.

```r
as.data.table(data.table(a=1), '2')
```

Previously, the `'2'`is ignored, now it will be matched to `keep.rownames=`. Obviously pretty silly toy example, but it's more believable for this to happen accidentally in the case of more generic dispatch examples winding up with arguments in `...` perhaps unintentionally.

The upshot is -- new arguments should go after `...`. That requires them to be named, which I think is a good practice anyway, see e.g.

https://design.tidyverse.org/dots-after-required.html

ben-schwen:
One last change (in code + docs) of changing the signature `as.data.table.data.table` to `as.data.table.data.table = function(x, ..., key=NULL)` as suggested by @MichaelChirico and we are good to go!

ben-schwen:
Thanks @Mukulyadav2004!

PR #6866, opened 2025-03-14
Rename the `test-lin-san` GitLab CI job

aitap:
As it turns out, #6476 [broke GitLab CI](https://gitlab.com/Rdatatable/data.table/-/pipelines/1716214169) because test stage names are expected to follow a pattern of `test-<platform>-<R version>[-<suffixes...>]`:
https://github.com/Rdatatable/data.table/blob/3afc750115abad122a27089605837c5b5aa881c9/.ci/publish.R#L162-L168
(And if `<R version>` is not `rel`, `dev`, or `old`, it must be a series of digits to be joined by dots.)

Perhaps this needs a comment in `.gitlab-ci.yml`, but I'm not sure where to place it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -213,7 +213,7 @@ test-lin-ancient-cran:
     - R CMD check --no-manual --no-build-vignettes --ignore-vignettes $(ls -1t data.table_*.tar.gz | head -n 1)
 
 # run the main checks with Address(+Leak),UBSanitizer enabled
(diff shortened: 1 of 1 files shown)


MichaelChirico (approved):

jangorecki:
https://github.com/Rdatatable/data.table/blob/master/.ci/README.md
This may be best place to mention such pattern, and possibly mention the reason for that as well.

PR #6870, opened 2025-03-16
Remove partial argument usage #6855

### LINKED ISSUE #6855: Typo in vignette on reshaping
I am currently working through the vignette on efficient reshaping and I think I came across one point, that is a bit unclear to me (or maybe wrong):
In section 3 (Enhanced (new) functionality), you say “We pass a list of columns to measure.vars”, but then your code passes a list to “measure” as “measure = list(colA, colB)”. To me, there is a difference between the measure() function, and the measure.vars argument of the melt function. I think, the code still works and the result is identical (probably, because the argument “measure” is automatically interpreted as “measure.vars”), but for clarity, I would change the melt function call to have the arguement “measure.vars = list(colA, colB)”, instead of “melt”, or am I misinterpreting something?

Mukulyadav2004:
Closes #6855 

This PR replaces partial argument names (id, fun.agg, measure) with their full names (id.vars, fun.aggregate, measure.vars) in the following vignettes for improved clarity and consistency:

->vignettes/data.table-reshape.Rmd
->vignettes/fr/data.table-reshape.Rmd
->vignettes/ru/data.table-reshape.Rmd
Additionally, these changes ensure better alignment with the surrounding text for improved coherence.

@MichaelChirico, please review when you have time. Thank you!

### DIFF
--- a/vignettes/datatable-reshape.Rmd
+++ b/vignettes/datatable-reshape.Rmd
@@ -130,7 +130,7 @@ dcast(DT.m1, family_id + age_mother ~ child, value.var = "dob")
 You can also pass a function to aggregate by in `dcast` with the argument `fun.aggregate`. This is particularly essential when the formula provided does not identify single observation for each cell.
 
 ```{r}
-dcast(DT.m1, family_id ~ ., fun.agg = function(x) sum(!is.na(x)), value.var = "dob")
+dcast(DT.m1, family_id ~ ., fun.aggregate = function(x) sum(!is.na(x)), value.var = "dob")
 ```
 
 Check `?dcast` for other useful arguments and additional examples.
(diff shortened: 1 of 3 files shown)


aitap (approved):
Tested by re-rendering with `options(warnPartialMatchArgs=TRUE, warn=2)`.

ben-schwen:
Ideally, we remove partial arguments from all vignettes in one go, right?

aitap:
There could be more in non-executable chunks, but in my tests, only
data.table-reshape.Rmd previously failed with
options(warnPartialMatchArgs=TRUE, warn=2).

ben-schwen:
> There could be more in non-executable chunks, but in my tests, only data.table-reshape.Rmd previously failed with options(warnPartialMatchArgs=TRUE, warn=2).

You are right, I just manually build all vignettes with `tools::buildVignette()` and only the reshape vignette failed

Mukulyadav2004:
I have removed partial matching in the reshaping vignette since it was causing build failures with options(warnPartialMatchArgs=TRUE, warn=2). The other instances in comments and tests are either intentional or not causing any issues.
I kept this PR focused on resolving that issue. However, I’m open to addressing partial matching in other non-executable chunks if needed.

MichaelChirico (approved):

MichaelChirico:
Thank you @Mukulyadav2004!!

Issue #1633, opened 2016-04-06
No realloc warning when adding many columns in a loop, inside a function


mattdowle:
```
> packageVersion("data.table")
[1] ‘1.9.7’
> x = data.table(a=1:3)
> truelength(x)
[1] 100
> f = function(DT) {
+  for (i in 1:200) {
+    DT[,paste0("new",i):=1:3]
+  }
+ }
> f(x)            # I thought there was a warning here that := on i==100 did a realloc
> dim(x)
[1]   3 100       # wrong result
> f = function(DT) {
+  for (i in 1:200) {
+    DT[,paste0("new",i):=1:3]
+  }
+  DT                            # needed in this case
+ }
> x = f(x)                       # x= needed in this case
> dim(x)
[1]   3 201                      # correct result
```

I seem to remember there was a warning. Maybe it isn't happening in this case for some reason. Need to reinstate the warning and add a test for it.

ProjectsByFerm:
Hi there, I am currently in an Open Source Software class, and my team and I are interested in this issue. Is this issue still open/important?

venom1204:
hi @tdhock @joshhwuu 
i looked into the issue and planning to restore a clear warning when reaclloc happens inside functions.

if this sounds good can i start working on it .
thanks

PR #6875, opened 2025-03-18
force class reset in as.data.table.data.frame

### LINKED ISSUE #6874: data.table fails with emmeans objects
This issue is fairly recent and used to work perfectly fine in the past. It may be partly related to the R version.  

Using `data.table` with `emmeans` objects now produces errors, while equivalent operations with base `data.frame` succeed.  

### **Reproducible Example:**  

```r
library(data.table) # v1.17.0
library(emmeans) # v1.10.7

mdl <- lm(dist ~ speed, data = cars)
em_res <- emmeans(mdl, ~ 1)

data.frame(em_res)
#        X1 emmean       SE df lower.CL upper.CL
# 1 overall  42.98 2.175002 48 38.60687 47.35313

data.table(em_res)
# Error in setDT(ans, key = key) : LENGTH or similar applied to S4 object

as.data.table(em_res)
# Error: C stack usage 15924944 is too close to the limit
```

### **Session Info:**  
```
R version 4.4.3 (2025-02-28 ucrt)
Platform: x86_64-w64-mingw32/x64

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.utf8  
[2] LC_CTYPE=English_United States.utf8    
[3] LC_MONETARY=English_United States.utf8 
[4] LC_NUMERIC=C                           
[5] LC_TIME=English_United States.utf8    

time zone: Europe/Berlin
tzcode source: internal

attached base packages:

other attached packages:
[1] emmeans_1.10.7-100003 data.table_1.17.0    

 [1] codetools_0.2-20   multcomp_1.4-28    Matrix_1.7-3       
 [4] lattice_0.22-6     coda_0.19-4.1      TH.data_1.1-3      
 [7] splines_4.4.3      estimability_1.5.1 zoo_1.8-13         
[10] mvtnorm_1.3-3      xtable_1.8-4       grid_4.4.3        
[13] sandwich_3.1-1     compiler_4.4.3     tools_4.4.3       
[16] survival_3.8-3     MASS_7.3-65       
```

ben-schwen:
Closes #6874 

- [x] data.table(s4)
- [x] as.data.table(no.reset)


ben-schwen:
Not sure how to best mention the regression of #5427 in NEWS 😭

inst/tests/S4.Rraw
```diff
@@ -126,3 +126,8 @@ setMethod(`%foo%`, c("Date", "CustomDurationClass"), function (e1, e2) e1 - e2@.
 test(8, as.IDate("2025-03-01") %foo% CustomDurationClass(1), as.IDate("2025-02-28"))
 removeGeneric("%foo%")
 removeClass("CustomDurationClass")
+
+# data.table(s4) #6874 should work
+s4cl = setClass("s4cl", slots = list(x="integer"))
+DT = setalloccol(structure(list(a = new("s4cl", x = 1L)), row.names = c(NA, -1L), class = c("data.table", "data.frame")))
+test(9, data.table(a=s4cl(x=1L)), DT)
```
MichaelChirico, line 133:
```suggestion
test(9, data.table(a=s4cl(x=1L)), DT)
removeClass("s4cl")
```

inst/tests/S4.Rraw
```diff
@@ -126,3 +126,8 @@ setMethod(`%foo%`, c("Date", "CustomDurationClass"), function (e1, e2) e1 - e2@.
 test(8, as.IDate("2025-03-01") %foo% CustomDurationClass(1), as.IDate("2025-02-28"))
 removeGeneric("%foo%")
 removeClass("CustomDurationClass")
+
+# data.table(s4) #6874 should work
+s4cl = setClass("s4cl", slots = list(x="integer"))
+DT = setalloccol(structure(list(a = new("s4cl", x = 1L)), row.names = c(NA, -1L), class = c("data.table", "data.frame")))
```
MichaelChirico, line 132:
```suggestion
s4cl = setClass("s4cl", slots=list(x="integer"))
DT = setalloccol(structure(list(a=new("s4cl", x=1L)), row.names=c(NA, -1L), class=c("data.table", "data.frame")))
```

R/as.data.table.R
```diff
@@ -215,7 +215,11 @@ as.data.table.list = function(x,
 
 as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   if (is.data.table(x)) return(as.data.table.data.table(x, key=key)) # S3 is weird, #6739. Also # nocov; this is tested in 2302.{2,3}, not sure why it doesn't show up in coverage.
-  if (!identical(class(x), "data.frame")) return(as.data.table(as.data.frame(x), keep.rownames=keep.rownames, key=key, ...))
+  if (!identical(class(x), "data.frame")) {
+    x = as.data.frame(x)
+    setattr(x,"class","data.frame") # cater for cases when as.data.frame is overridden #6874
```
MichaelChirico, line 220:
```suggestion
    setattr(x, "class", "data.frame") # cater for cases when as.data.frame can generate a loop #6874
```

NEWS.md
```diff
@@ -16,6 +16,8 @@
 
 4. `as.data.table()` now properly handles keys: specifying keys sets them, omitting keys preserves existing ones, and setting `key=NULL` clears them, [#6859](https://github.com/Rdatatable/data.table/issues/6859). Thanks @brookslogan for the report and @Mukulyadav2004 for the fix.
 
+5. `as.data.table.data.frame(x)` could enter an infinity loop if the S3 method for `as.data.frame(x)` was overridden without ensuring `class(as.data.frame(x)) == "data.frame"`, [#6874](https://github.com/Rdatatable/data.table/issues/6874). Thanks @matschmitz for the report and @ben-schwen for the fix.
```
MichaelChirico, line 19:
```suggestion
5. `as.data.table()` on `x` avoids an infinite loop if the output of the corresponding `as.data.frame()` method has the same class as the input, [#6874](https://github.com/Rdatatable/data.table/issues/6874). Concretely, we had `class(x) = c('foo', 'data.frame')` and `class(as.data.frame(x)) = c('foo', 'data.frame')`, so `as.data.frame.foo` wound up getting called repeatedly. Thanks @matschmitz for the report and @ben-schwen for the fix.
```

R/as.data.table.R
```diff
@@ -215,7 +215,11 @@ as.data.table.list = function(x,
 
 as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   if (is.data.table(x)) return(as.data.table.data.table(x, key=key)) # S3 is weird, #6739. Also # nocov; this is tested in 2302.{2,3}, not sure why it doesn't show up in coverage.
-  if (!identical(class(x), "data.frame")) return(as.data.table(as.data.frame(x), keep.rownames=keep.rownames, key=key, ...))
+  if (!identical(class(x), "data.frame")) {
+    x = as.data.frame(x)
+    setattr(x, "class", "data.frame") # cater for cases when as.data.frame can generate a loop #6874
```
MichaelChirico, line 220: This might not be ideal, e.g. if there's a "proper" S3 class where we have

```r
class(x) = c('class1', 'class2', 'data.frame')
class(as.data.frame(x)) = c('class2', 'data.frame')
class(as.data.frame(as.data.frame(x))) = 'data.frame'
```

Then we exit the loop, and possibly apply one additional layer of `as.data.frame()` that is required for correct results.

Unfortunately I'm not sure right workaround? Maybe record `class_orig = class(x); x=as.data.frame(x); if (identical(class(x), class_orig)) setattr(x, 'class', 'data.frame')`?

Maybe moot in practice, let's monitor the revdeps.

src/assign.c
```diff
@@ -232,7 +232,7 @@ SEXP setdt_nrows(SEXP x)
       }
       len_xi = INTEGER(dim_xi)[0];
     } else {
-      len_xi = LENGTH(xi);
+      len_xi = length(xi);
```
MichaelChirico, line 235: Are any of the other >200 places we use `LENGTH()` worth changing too? (maybe glance through now if anything looks like an obvious fix to put in the patch release, and file a follow-up to go through more carefully in the future)

ben-schwen: Apparently, we already had this case before item 3 of `1.7.2`

> A crash bug is fixed when j returns a (strictly) NULL column next
        to a non-empty column, #1633. This case was anticipated and coded
        for but an errant LENGTH() should have been length(). Thanks
        to Dennis Murphy for reporting.

Might be worth to check carefully

R/as.data.table.R
ben-schwen: I added the check. Better to be safe than sorry and as we saw some objects apparently like this nested `as.data.frame` conversion.

MichaelChirico:
maybe we should just directly call as.data.table.data.frame() after we reun as.data.frame(), just like we run as.data.table.data.table() in the line before, WDYT?

src/assign.c
ben-schwen: I filed #6877 for this

ben-schwen:
> maybe we should just directly call as.data.table.data.frame() after we reun as.data.frame(), just like we run as.data.table.data.table() in the line before, WDYT?

Good point. We save a dispatch this way :)

R/as.data.table.R
```diff
@@ -215,7 +215,12 @@ as.data.table.list = function(x,
 
 as.data.table.data.frame = function(x, keep.rownames=FALSE, key=NULL, ...) {
   if (is.data.table(x)) return(as.data.table.data.table(x, key=key)) # S3 is weird, #6739. Also # nocov; this is tested in 2302.{2,3}, not sure why it doesn't show up in coverage.
-  if (!identical(class(x), "data.frame")) return(as.data.table(as.data.frame(x), keep.rownames=keep.rownames, key=key, ...))
+  if (!identical(class(x), "data.frame")) {
+    class_orig = class(x)
+    x = as.data.frame(x)
+    if (identical(class(x), class_orig)) setattr(x, "class", "data.frame") # cater for cases when as.data.frame can generate a loop #6874
+    return(as.data.table.data.frame(x, keep.rownames=keep.rownames, key=key, ...))
+  }
```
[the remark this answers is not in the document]
ben-schwen: but this would put us back to the infinity recursion?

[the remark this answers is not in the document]
MichaelChirico: true true. for my suggestion we'd have to separate the rest of the body into an implementation function and call that to skip the input check at the top. not sure it's that beneficial actually.

MichaelChirico (approved):
Please help to cherry-pick this into the patch branch

PR #6879, opened 2025-03-21
Document jj scoping behavior in groupingsets() with practical examples

### LINKED ISSUE #5560: groupingsets() wrong scoping logic
```R
library(data.table)
irisDT <- data.table(iris)

foo = function(w) {
  bv = "Species"
  groupingsets(
    irisDT, 
    j  = lapply(.SD, \(y) sum(y > w)),
    by = bv,
    sets = as.list(bv),
    .SDcols = c("Sepal.Length", "Petal.Length")
  ) |> print()
  irisDT[, lapply(.SD, \(y) sum(y > w)), .SDcols = c("Sepal.Length", "Petal.Length"), by=bv]
}
foo(5)
# Error in ..FUN1(Sepal.Length) : object 'w' not found

w = 4
foo(5)
#       Species Sepal.Length Petal.Length
#        <fctr>        <int>        <int>
# 1:     setosa           50            0
# 2: versicolor           50           34
# 3:  virginica           50           50
#       Species Sepal.Length Petal.Length
#        <fctr>        <int>        <int>
# 1:     setosa           22            0
# 2: versicolor           47            1
# 3:  virginica           49           41
```

I'm using data.table ‘1.14.7’ (just updated with `data.table::update_dev_pkg()`) with R 4.2.1 on Windows 10.

Mukulyadav2004:
closes #5560 
In this PR I improved documentation for the jj argument in groupingsets(), explaining how jj = substitute(expr) preserves both expression and environment context when j would fail with outer-environment variables.
The documentation now explains why users encounter "object not found" errors when using variables from outer environments with j, and how jj solves this problem.

Key changes in groupingsets.Rd:
1.)Added detailed explanation of scoping behavior in the \details section
2.)Added three practical examples demonstrating different jj usage patterns:
   Basic usage with list and lapply
   Simple aggregation
   Named results with direct column references
   
These changes directly address the request for "more examples on the use of the jj argument" and will help users understand when and how to use this parameter in their own functions.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,8 @@
    + Argument `in.place` to `droplevels` has been removed.
    + It's now an error to set `datatable.nomatch`, which has been warning since 1.15.0.
 
+3. The groupingsets() documentation now explicitly addresses scoping issues when using variables from outer environments (e.g., function parameters) in aggregation expressions. Previously, using such variables in j could lead to "object not found" errors due to evaluation environment mismatches. This is resolved by using jj = substitute(expr) instead, which captures both the expression and its environment per R’s scoping rules (§6). Three new examples demonstrate proper usage, including function parameter references (e.g., sum(value > threshold)) and multi-aggregation patterns.
+
 # data.table [v1.17.0](https://github.com/Rdatatable/data.table/milestone/34)  (20 Feb 2025)
 
 ## POTENTIALLY BREAKING CHANGES
--- a/man/groupingsets.Rd
+++ b/man/groupingsets.Rd
@@ -34,6 +34,8 @@ groupingsets(x, \dots)
     The \code{label} argument can be a named list of scalars, or a scalar, or \code{NULL}. When \code{label} is a list, each element name must be (1) a variable name in \code{by}, or (2) the first element of the class in the data.table \code{x} of a variable in \code{by}, or (3) one of 'character', 'integer', 'numeric', 'factor', 'Date', 'IDate'. The order of the list elements is not important. A label specified by variable name will apply only to that variable, while a label specified by first element of a class will apply to all variables in \code{by} for which the first element of the class of the variable in \code{x} matches the \code{label} element name, except for variables that have a label specified by variable name (that is, specification by variable name takes precedence over specification by class). For \code{label} elements with name in \code{by}, the class of the label value must be the same as the class of the variable in \code{x}. For \code{label} elements with name not in \code{by}, the first element of the class of the label value must be the same as the \code{label} element name. For example, \code{label = list(integer = 999, IDate = as.Date("3000-01-01"))} would produce an error because \code{class(999)[1]} is not \code{"integer"} and \code{class(as.Date("3000-01-01"))[1]} is not \code{"IDate"}. A corrected specification would be \code{label = list(integer = 999L, IDate = as.IDate("3000-01-01"))}.
 
     The \code{label = <scalar>} option provides a shorter alternative in the case where only one class of grouping variable requires a label. For example, \code{label = list(character = "Total")} can be shortened to \code{label = "Total"}. When this option is used, the label will be applied to all variables in \code{by} for which the first element of the class of the variable in \code{x} matches the first element of the class of the scalar.
+
+    The \code{jj} argument provides a mechanism to handle scoping issues when working with variables from outer environments. When using \code{j} directly with variables not defined in the columns of the data.table, R's evaluation rules may not find these variables, resulting in "object not found" errors. The \code{jj} argument accepts a quoted expression created with \code{substitute()}, which captures both the expression and its environment. This approach is particularly useful in functions where parameters need to be referenced within aggregation expressions. When \code{jj} is provided, the function will ignore any value specified in the \code{j} argument.
 }
 \value{
     A data.table with various aggregates.
@@ -81,5 +83,12 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
 # groupingsets
 groupingsets(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"),
              sets = list("color", c("year","status"), character()), id=TRUE)
+# Using jj argument(enables access to variables from outer environments)
+groupingsets(DT, jj = substitute(c(list(count=.N), lapply(.SD, sum))), by = c("color","year","status"),  # Basic jj usage with substitute()
+             sets = list("color", c("year","status"), character()), id = TRUE, .SDcols = "value") 
+groupingsets(DT, jj = substitute(lapply(.SD, sum)), by = c("color","year","status"),   # Simple aggregation with jj
+             sets = list("color", c("year","status"), character()), id = TRUE, .SDcols = "value")
+groupingsets(DT, jj = substitute(list(count = .N, total = sum(value))), by = c("color","year","status"), # jj with direct column references         
+             sets = list("color", c("year","status"), character()), id = TRUE)      
 }
 \keyword{ data }
--- a/man/groupingsets.Rd
+++ b/man/groupingsets.Rd
@@ -84,11 +84,14 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
 groupingsets(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"),
              sets = list("color", c("year","status"), character()), id=TRUE)
 # Using jj argument(enables access to variables from outer environments)
-groupingsets(DT, jj = substitute(c(list(count=.N), lapply(.SD, sum))), by = c("color","year","status"),  # Basic jj usage with substitute()
+groupingsets(DT, jj = substitute(c(list(count=.N), lapply(.SD, sum))), 
+             by = c("color","year","status"),  # Basic jj usage with substitute()
              sets = list("color", c("year","status"), character()), id = TRUE, .SDcols = "value") 
-groupingsets(DT, jj = substitute(lapply(.SD, sum)), by = c("color","year","status"),   # Simple aggregation with jj
+groupingsets(DT, jj = substitute(lapply(.SD, sum)),
+             by = c("color","year","status"),   # Simple aggregation with jj
              sets = list("color", c("year","status"), character()), id = TRUE, .SDcols = "value")
-groupingsets(DT, jj = substitute(list(count = .N, total = sum(value))), by = c("color","year","status"), # jj with direct column references         
+groupingsets(DT, jj = substitute(list(count = .N, total = sum(value))),
+             by = c("color","year","status"), # jj with direct column references         
              sets = list("color", c("year","status"), character()), id = TRUE)      
 }
 \keyword{ data }


man/groupingsets.Rd
```diff
@@ -81,5 +83,15 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
 # groupingsets
 groupingsets(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"),
              sets = list("color", c("year","status"), character()), id=TRUE)
+# Using jj argument(enables access to variables from outer environments)
```
jangorecki, line 86: Space is missing

man/groupingsets.Rd
```diff
@@ -81,5 +83,15 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
...
+groupingsets(DT, jj = substitute(c(list(count=.N), lapply(.SD, sum))),
```
jangorecki, line 87: Won't quote() be sufficient here? Substitute will be good when using from inside a function

jangorecki (commented):
Looks good, please see if quote() won't be sufficient where substitute was used

man/groupingsets.Rd
Mukulyadav2004: you are right that substitute will be good when using from inside a function and as substitution is necessary when capturing variables from the function's environment but here all variables are within data.table(like column) 
But after taking a closer look at  groupingsets() implementation to check whether quote() would work instead of substitute() in which after reviewing the code, I think substitute() is indeed the right approach as the whole codebase consistently use substitute()

here are snippets in which they behave the same way whether direcly calling groupingsets() or using functions like rollup() or cube()

jj = if (!missing(jj)) jj else substitute(j)          and         jj = substitute(j)

this code evaluates eval(jj) without specifying environment , which works correctly as it brings both the expression and its environment.
So I  think using substitute() keeps things consistent with how the functions are built to work, even if quote() might be sufficient here .
But I'm open to use quote() in place of substitute() in examples if you feel it would be clearer for users.

NEWS.md
```diff
@@ -27,6 +27,8 @@
    + Argument `in.place` to `droplevels` has been removed.
    + It's now an error to set `datatable.nomatch`, which has been warning since 1.15.0.
 
+3. The groupingsets() documentation now explicitly addresses scoping issues when using variables from outer environments (e.g., function parameters) in aggregation expressions. Previously, using such variables in j could lead to "object not found" errors due to evaluation environment mismatches. This is resolved by using jj = substitute(expr) instead, which captures both the expression and its environment per R’s scoping rules (§6). Three new examples demonstrate proper usage, including function parameter references (e.g., sum(value > threshold)) and multi-aggregation patterns.
```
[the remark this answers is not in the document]
Mukulyadav2004: It represents chapter , I'll update it to ch for better readability

man/groupingsets.Rd
jangorecki: If quote() will be sufficient then it should be used instead of substitute because it will be as simple as necessary, substitute is more powerful and may create impression that it is necessary there. While it is necessary only from inside a function, that is why is used in rollup and cube. One example of function wrapping groupingsets and use of substitute would therefore be useful as well.

NEWS.md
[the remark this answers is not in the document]
jangorecki: This chapter 6 is from R language manual, one of the few official R manuals, therefore specifying that would be desired as well.

man/groupingsets.Rd
Mukulyadav2004: Thanks for your clarification. I updated the description of jj usage - include when to use quote() and when substitute() and examples to use quote() for the direct calls to groupingsets().
I also added an example showing when substitute() is actually needed - inside a wrapper function where variables from the outer environment need to be captured.

Could you please review these changes and let me know if any further modifications are needed? Thank you!

NEWS.md
```diff
@@ -27,6 +27,8 @@
    + Argument `in.place` to `droplevels` has been removed.
    + It's now an error to set `datatable.nomatch`, which has been warning since 1.15.0.
 
+3. The groupingsets() documentation now explicitly addresses scoping issues when using variables from outer environments (e.g., function parameters) in aggregation expressions. Previously, using such variables in j could lead to "object not found" errors due to evaluation environment mismatches. This is resolved by using jj = substitute(expr) instead, which captures both the expression and its environment per R’s scoping rules (Chapter 6 of the R Language Definition manual). Three new examples demonstrate proper usage, including function parameter references (e.g., sum(value > threshold)) and multi-aggregation patterns.
```
aitap, line 30: `substitute()` does not capture the environment (that's the problem that [quosures](https://www.r-project.org/dsc/2017/slides/tidyeval-hygienic-fexprs.pdf) are intended to solve). What it does is _access_ the list or environment given to `substitute()` as an argument (defaults to the caller's environment) at the time of the call. Any symbol that can be directly found in that environment (ignoring the parent chain) is substituted: promises are replaced with the promise expression (losing the promise environment in the process!), other values are substituted directly.

man/groupingsets.Rd
```diff
@@ -81,5 +83,26 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
...
+             by = c("color","year","status"),   # Simple aggregation with jj
+             sets = list("color", c("year","status"), character()), id = TRUE, .SDcols = "value")
+groupingsets(DT, jj = quote(list(count = .N, total = sum(value))),
+             by = c("color","year","status"), # jj with direct column references         
+             sets = list("color", c("year","status"), character()), id = TRUE)      
+
+# Example showing when substitute() is necessary - in a wrapper function
+custom_grouping <- function(DT, multiplier = 1) {
+  # Here substitute() is needed to capture 'multiplier' from function environment
+  groupingsets(DT, jj = substitute(list(count = .N, total = sum(value) * multiplier)),
```
aitap, line 101: Unfortunately, a plain `substitute()` is not a reliable solution here. Consider:

```r
local({ foo <- 3; custom_grouping(DT, multiplier=foo) })
```

`substitute(list(count = .N, total = sum(value) * multiplier))` expands the `multiplier` promise into `foo`, which is not a variable that `custom_grouping` is supposed to know about, and it doesn't (since it lives in a local environment that `environment(custom_grouping)` does not inherit from).

What does work is `substitute(list(count = .N, total = sum(value) * multiplier), list(multiplier = multiplier))`, because with a `list` argument to `substitute`, `multiplier` is a plain value, not an argument promise.

Mukulyadav2004: Thank you for explaining the nuance about substitute() with nested environments. I'll implement your suggested approach using substitute(expr, list(param=value)) in the example to properly handle the multiplier parameter regardless of call context.

Mukulyadav2004:
Hi @aitap @jangorecki,
I have implemented the changes as suggested. Could you please review them and let me know if there are any further suggestions? I am open to feedback.
Thank you

man/groupingsets.Rd
```diff
@@ -81,5 +83,28 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
...
+# Note: Use substitute(expr, list(param=value)) to bind values, not symbols
+custom_grouping <- function(DT, multiplier = 1) {
+  # use substitute() with environment to ensure variable are properly evaluated
+  groupingsets(DT, jj = substitute(list(count = .N, total = sum(value) * multiplier),
+               list(multiplier = multiplier)),
+               by = c("color", "year", "status"),
+               sets = list("color", c("year", "status"), character()), id = TRUE)
+}
+# Using the wrapper function
+custom_grouping(DT, multiplier = 2)
```
jangorecki, line 108: There is not much gain providing value to be substituted. Shouldn't this be a symbol to be substituted? For current example j argument will work well enough, isn't it?

Mukulyadav2004: you're right , here symbol from outer environment should be used . I updated this example to use this .

man/groupingsets.Rd
```diff
@@ -97,14 +97,14 @@ groupingsets(DT, jj = quote(list(count = .N, total = sum(value))),
...
-  groupingsets(DT, jj = substitute(list(count = .N, total = sum(value) * multiplier),
-               list(multiplier = multiplier)),
+  groupingsets(DT, jj = substitute(list(count = .N, total = sum(value) * m),
+               list(m = multiplier)),
                by = c("color", "year", "status"),
                sets = list("color", c("year", "status"), character()), id = TRUE)
 }
 # Using the wrapper function
-custom_grouping(DT, multiplier = 2)
+local({ foo <- 3; custom_grouping(DT, multiplier = foo) })
```
jangorecki, line 108: I think it is overly complex now. We don't need local(), just something similar to cube/rollup, where user can pass column names as symbols to his/her wrapper function.

Mukulyadav2004: so does by simply replacing multiplier = 2, to a parameter holding a symbol works here ? ,  by converting amount to symbol in groupingsets .
Here amount is defined early before the examples as one of the parameter like color , value and status in data.table.
For eg .
custom_grouping <- function(DT, value_col) {
  groupingsets(DT, jj = substitute(list(count = .N, total = sum(col)), 
                                 list(col = as.name(value_col))),
               by = c("color", "year", "status"),
               sets = list("color", c("year", "status"), character()), 
               id = TRUE)
}
custom_grouping(DT, value_col = "amount")

Mukulyadav2004: Yes, I'd like to provide an example as I believe that since j evaluates expressions directly in groupingsets()'s environment, it can't resolve dynamic symbols like as.name(value_col). Using substitute() for jj replaces the placeholder with the actual column name (amount), ensuring that sum(amount) correctly operates on the data.

iagogv3: @Mukulyadav2004 It would be great!
I was going to ask for such an example, just using `substitute2` instead of `substitute` and making explicit that `list(col = as.name(value_col))` is the `env` argument: `env = list(col = as.name(value_col))`.

Thank you for this work!

jangorecki (commented):
finally got some time to elaborate a bit more

man/groupingsets.Rd
jangorecki: You need to provide more complete example as it is not clear what `...` is meant to pass to substitute2().

As for use of substitute:
AFAIR when you use substitute2() then as.name() is not needed, if you use base R substitute() then it is needed.

If function does not have NSE, like this one, then substitute may not be needed, easier just as.call(list(as.name(value_col))) or bquote(). I am not sure here as I don't know the desired computation.

Mukulyadav2004: Thanks for the clarification. You're right that this function doesn't use NSE since the input is a string. But as per requirement of example that uses `substitute2()` with an explicit `env = list(col = as.name(value_col))`, here below is complete example-
```
custom_grouping <- function(DT, value_col) {
  groupingsets(DT, jj = substitute2(list(count = .N, total = sum(col)), env = list(col = as.name(value_col))),
    by = c("color", "year", "status"),
    sets = list("color", c("year", "status"), character()),
    id = TRUE
  )
}
custom_grouping(DT, "amount")
```
I hope this aligns with your intent, but if there's a better way you’d suggest approaching this, I’d really appreciate your guidance.

jangorecki: I really like the use of `substitute2()` here, one function call less.

Mukulyadav2004: Thanks for the clarification. I’ve added example that accepts an arbitrary j and forwards it with` substitute(j)` to `groupingsets()`, now that [#6899](https://github.com/Rdatatable/data.table/pull/6899) handles the environment correctly, this removes need for relying on `substitute2()` or `env `workarounds. Let me know if you'd like any further adjustments.

man/groupingsets.Rd
```diff
@@ -81,5 +81,25 @@ cube(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"
 # groupingsets
 groupingsets(DT, j = c(list(count=.N), lapply(.SD, sum)), by = c("color","year","status"),
              sets = list("color", c("year","status"), character()), id=TRUE)
+
+# Using jj argument with quote() for direct calls
+groupingsets(DT, jj = quote(c(list(count=.N), lapply(.SD, sum))),
```
jangorecki, line 86: what is the advantage of using `jj` in this example over `j`?

if we use `j` we have one function call less: quote()
examples like this may be confusing as I don't see good reasons to use it like that. If `jj` is constructed programatically, or passed from outer function, then yes `jj` is useful.

NEWS.md
```diff
@@ -27,6 +27,8 @@
    + Argument `in.place` to `droplevels` has been removed.
    + It's now an error to set `datatable.nomatch`, which has been warning since 1.15.0.
 
+3. The groupingsets() documentation now explicitly addresses scoping issues when using variables from outer environments (e.g., function parameters) in aggregation expressions. Previously, using such variables in j could lead to "object not found" errors due to evaluation environment mismatches. This is resolved by using jj = substitute(expr, list(param=value)) instead, which creates an expression with values directly substituted at call time, following R's evaluation rules (Chapter 6 of the R Language Definition manual). Three new examples demonstrate proper usage, including function parameter references (e.g., sum(value > threshold)) and multi-aggregation patterns.
```
jangorecki, line 30: considering changes in this PR, is this news description still valid?

man/groupingsets.Rd
Mukulyadav2004: Sorry for the late response; our practical exams are ongoing. We initially introduced `jj` to highlight how expressions referencing variables in an outer environment can be effectively captured.  But if you feel no need for this as of now so keeping one wrapper-function example demonstrating how `jj` and `substitute()` handle outer-environment variables, and remove the extra `jj` examples for direct calls.

NEWS.md
Mukulyadav2004: I will update news after you approve the final example

man/groupingsets.Rd
jangorecki: "highlight how expressions referencing variables in an outer environment can be effectively captured" complete example of it should make it more clear for me and other readers

Mukulyadav2004: Apologies for the confusion in my previous message. I misspoke – the `jj=quote(...)` examples don't actually capture outer variables. The wrapper function example using `jj=substitute(j)` is the one that correctly demonstrates capturing variables(`multiplier`) from the outer environment.
I'll remove the confusing `jj=quote(...)` examples and keep only the wrapper function example, as that clearly shows the intended use case for handling outer variables.

NEWS.md
MichaelChirico: Ping @jangorecki for if there's any further changes needed besides refining the NEWS

jangorecki:
Yes, description in the news is not very clear. Much better would be to provide example there, one of a problem, and second of a solution.

MichaelChirico:
> Yes, description in the news is not very clear. Much better would be to provide example there, one of a problem, and second of a solution.

Sorry, let me be clearer: Mukul had asked about whether the current state of the examples is satisfactory. He was waiting to update the NEWS until the rest of the PR is in a good state.

jangorecki:
Yes, example is perfect as simple as possible and still express the use case.

Mukulyadav2004:
Sorry, there was internet issue.  Does this line explaining issue and solution example seems right to you - `mygrp(DT, sum(value) * multiplier)` would fail with "object not found" when `multiplier` is defined outside the function, but using `jj = substitute(j)` instead of `j = j` resolves this by properly capturing the expression and its environment.

PR #6880, opened 2025-03-22
add support for melt(id.vars=measure())

### LINKED ISSUE #6867: melt could support patterns for id.vars
Today I tried this:
```r
> DT=data.table(x_long=0, x_short=0, z=0, y1=1, y2=2)
> melt(DT, measure.vars=patterns("y"), id.vars=patterns("x"))
Erreur dans patterns("x") : Pattern not found: [[x]]
```
And I expected this:
```r
> melt(DT, measure.vars=patterns("y"), id.vars=patterns("x", cols=names(DT)))
   x_long x_short variable value
    <num>   <num>   <fctr> <num>
1:      0       0       y1     1
2:      0       0       y2     2
> melt(DT, measure.vars=patterns("y"), id.vars=c("x_long","x_short"))
   x_long x_short variable value
    <num>   <num>   <fctr> <num>
1:      0       0       y1     1
2:      0       0       y2     2
```
Should be easy to implement, just need to turn on NSE (eval_with_cols) for id.vars in addition to measure.vars.

tdhock:
Closes #6867

### DIFF
--- a/R/fmelt.R
+++ b/R/fmelt.R
@@ -183,6 +183,13 @@ melt.data.table = function(data, id.vars, measure.vars, variable.name = "variabl
        verbose = getOption("datatable.verbose")) {
   if (!is.data.table(data)) stopf("'data' must be a data.table")
   if (missing(id.vars)) id.vars=NULL
+  id.sub = substitute(id.vars)
+  if (is.call(id.sub)) {
+    eval.result = eval_with_cols(id.sub, names(data))
+    if (!is.null(eval.result)) {
+      id.vars = eval.result
+    }
+  }
   if (missing(measure.vars)) measure.vars = NULL
   measure.sub = substitute(measure.vars)
   if (is.call(measure.sub)) {
--- a/R/fmelt.R
+++ b/R/fmelt.R
@@ -182,20 +182,17 @@ melt.data.table = function(data, id.vars, measure.vars, variable.name = "variabl
        value.name = "value", ..., na.rm = FALSE, variable.factor = TRUE, value.factor = FALSE,
        verbose = getOption("datatable.verbose")) {
   if (!is.data.table(data)) stopf("'data' must be a data.table")
-  if (missing(id.vars)) id.vars=NULL
-  id.sub = substitute(id.vars)
-  if (is.call(id.sub)) {
-    eval.result = eval_with_cols(id.sub, names(data))
-    if (!is.null(eval.result)) {
-      id.vars = eval.result
-    }
-  }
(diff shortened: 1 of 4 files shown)


R/fmelt.R
```diff
@@ -183,6 +183,13 @@ melt.data.table = function(data, id.vars, measure.vars, variable.name = "variabl
        verbose = getOption("datatable.verbose")) {
   if (!is.data.table(data)) stopf("'data' must be a data.table")
   if (missing(id.vars)) id.vars=NULL
+  id.sub = substitute(id.vars)
```
tdhock, line 186: I just copied the code for measure.vars and changed to id.vars -- do you think we should try a for loop instead? or is this ok?

tdhock: (for loop over measure.vars and id.vars to emphasize they are handled via the same logic)

MichaelChirico:
I can take a look, wondering if anyone else is more suitable, I still don't feel very familiar/facile with the "new" `melt()` API. WDYT?

tdhock:
I'm the main melt API person, maybe @jangorecki could review?

R/fmelt.R
MichaelChirico: might be preferable to use a helper rather than copy-paste to ensure they stay in sync going forward

inst/tests/tests.Rraw
```diff
@@ -3221,6 +3221,10 @@ test(1034, as.data.table(x<-as.character(sample(letters, 5))), data.table(V1=x))
   test(1035.12, attr(melt(DT, id.vars=1:2)$x, "foo"), "bla1")
   test(1035.13, attr(melt(DT, id.vars=1:2)$y, "bar"), 1:4)
 
+  # issue #6867 - id.vars=patterns().
+  DT=data.table(x_long=0, x_short=0, z=0, y1=1, y2=2)
+  test(1035.131, melt(DT, measure.vars=patterns("y"), id.vars=patterns("x")), data.table(x_long=0, x_short=0, variable=factor(c("y1","y2")), value=c(1,2)))
```
MichaelChirico, line 3226: should this get more tests, e.g. error cases, cases where `id.vars`+`measure.vars` interact, or do you think that's well-covered enough by existing tests already?

(especially if using a helper, we are more confident tests applying to edge cases of `measure.vars` apply as well to `id.vars`).

tdhock:
I added a for loop and updated the man page (doc for id.vars argument + example).

tdhock:
thanks for your review!

PR #6884, opened 2025-03-25
add test for `as.data.table.data.frame(x, keep.rownames=TRUE)`

ben-schwen:
Follows #6865


MichaelChirico:
Oh, NVM, I should have clicked the link :) I thought this was a follow-up to #6875, not #6865

PR #6887, opened 2025-03-26
Don't mention 14-year-old version info in ?alloc.col

MichaelChirico:
v1.7.2 is from 2011, I don't think it's relevant to detail a before/after 2011 distinction any more.

R 2.14.0 is also from 2011.

https://cran.r-project.org/src/contrib/Archive/data.table/
https://github.com/wch/r-source/tree/tags/R-2-14-0

Also wrapped some of the text, not all IDEs do text wrapping so these very-very wide lines are a bit hard to work with.


ben-schwen:
LGTM. It's gonna be 19 years of DT on April 15th!

PR #6889, opened 2025-03-31
`fwrite`: pre-encode strings and factor levels

aitap:
Previously, `fwrite()` deferred encoding of the strings to `fwriteR.c`:[`getString`](https://github.com/Rdatatable/data.table/blob/2cb03162a21328cc5f68a8c3b0e554f5edfcb5b9/src/fwriteR.c#L22-L25),[`getCategString`](https://github.com/Rdatatable/data.table/blob/2cb03162a21328cc5f68a8c3b0e554f5edfcb5b9/src/fwriteR.c#L50-L54) called from OpenMP threads. Calling translateChar(UTF8) to encode a string results in memory allocation unless it is already in the desired encoding, which is unsafe to perform on a non-main thread.

Fixes: #6883


MichaelChirico:
LGTM, it looks like @ben-schwen looked more closely in the issue so I'll assign him for final approval

ben-schwen:
NEWS is missing.

Current test case also does not run into an error on my system with `1.17.0` but needs to be adjusted to 
`DF = list(rep(iconv("\uf8", from = "UTF-8", to = "latin1"), 2e6))`

aitap:
Thanks for finding another reproducer, it crashes for me too. Does it crash for you with `nThread` explicitly set to 2?

ben-schwen (approved):
LGTM, TY!

Issue #6886, opened 2025-03-26
Strange behaviours when updating by reference a factor


iagogv3:
Look below:
```r
> library(data.table)
data.table 1.17.0 using 6 threads (see ?getDTthreads).  Latest news: r-datatable.com
> load("../Tresorit/aux_.rdata")
> aux
        seqn                        area
       <int>                      <fctr>
    1:     1     Pontevedra-Ourense-Vigo
    2:     2 Coruña-Santiago-Ferrol-Lugo
    3:     3 Coruña-Santiago-Ferrol-Lugo
    4:     4     Pontevedra-Ourense-Vigo
    5:     5     Pontevedra-Ourense-Vigo
   ---                                  
14472: 14472     Pontevedra-Ourense-Vigo
14473: 14473     Pontevedra-Ourense-Vigo
14474: 14474 Coruña-Santiago-Ferrol-Lugo
14475: 14475 Coruña-Santiago-Ferrol-Lugo
14476: 14476     Pontevedra-Ourense-Vigo
> str(aux)
Classes ‘data.table’ and 'data.frame':  14476 obs. of  2 variables:
 $ seqn: int  1 2 3 4 5 6 7 8 9 10 ...
 $ area: Factor w/ 3 levels "","Coruña-Santiago-Ferrol-Lugo",..: 3 2 2 3 3 2 3 2 2 2 ...
 - attr(*, ".internal.selfref")=<externalptr> 
> levels(aux$area)
[1] ""                            "Coruña-Santiago-Ferrol-Lugo"
[3] "Pontevedra-Ourense-Vigo"    
> aux[area==""]
    seqn   area
   <int> <fctr>
1:  4207       
2:  6514       
> aux[seqn==6514, area := "Coruña-Santiago-Ferrol-Lugo"]
> levels(aux$area)
[1] ""                            "Coruña-Santiago-Ferrol-Lugo" "Pontevedra-Ourense-Vigo"     "Coruña-Santiago-Ferrol-Lugo"
> unique(levels(aux$area))
[1] ""                            "Coruña-Santiago-Ferrol-Lugo" "Pontevedra-Ourense-Vigo"    
> sessionInfo()
R version 4.4.2 (2024-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows 11 x64 (build 26100)

Matrix products: default


locale:
[...]

PR #6890, opened 2025-03-31
`memrecycle()`: match factor levels in UTF-8

aitap:
Previously, by-reference sub-assignment to a factor column could fail to match strings with identical content if they had different encoding bits (even `CE_NATIVE` in a UTF-8 locale vs. `CE_UTF8`), causing duplicate levels.

Fixes: #6886

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21114,6 +21114,10 @@ DF = structure(list(a = 1:2), class = c("data.frame", "no.reset"), row.names = c
 test(2310.01, as.data.table(DF), data.table(a=1:2))
 
 # memrecycle() did not consider string encodings for factor levels #6886
-DT = data.table(factor(rep(enc2utf8("\uf8"), 3)))
-DT[1, V1 := iconv(levels(V1), from = "UTF-8", to = "latin1")]
-test(2311, nlevels(DT$V1), 1L)
+DT = data.table(factor(rep("\uf8", 3)))
+# identical() to V1's only level but stored in a different CHARSXP
+samelevel = iconv(levels(DT$V1), from = "UTF-8", to = "latin1")
+DT[1, V1 := samelevel]
+test(2311.1, nlevels(DT$V1), 1L) # used to be 2
+DT[1, V1 := factor("a", levels = c("a", samelevel))]
+test(2311.2, nlevels(DT$V1), 2L) # used to be 3
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21115,6 +21115,6 @@ DF = structure(list(a = 1:2), class = c("data.frame", "no.reset"), row.names = c
 test(2310.01, as.data.table(DF), data.table(a=1:2))
 
 # memrecycle() did not consider string encodings for factor levels #6886
-DT = data.table(factor(rep(enc2utf8('ø'), 3)))
-DT[1,V1 := iconv('ø', to='latin1')]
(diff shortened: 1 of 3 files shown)


inst/tests/tests.Rraw
```diff
@@ -21113,3 +21113,8 @@ test(2309.09, as.data.table(df, keep.rownames=TRUE), data.table(rn = c("a","b"),
 as.data.frame.no.reset = function(x) x
 DF = structure(list(a = 1:2), class = c("data.frame", "no.reset"), row.names = c(NA, -2L))
 test(2310.01, as.data.table(DF), data.table(a=1:2))
+
+# memrecycle() did not consider string encodings for factor levels #6886
+DT = data.table(factor(rep(enc2utf8('ø'), 3)))
```
MichaelChirico, line 21118: should this be a UTF escape instead of a literal non-ASCII character?

inst/tests/tests.Rraw
```diff
@@ -21113,3 +21113,8 @@ test(2309.09, as.data.table(df, keep.rownames=TRUE), data.table(rn = c("a","b"),
...
+DT[1,V1 := iconv('ø', to='latin1')]
+test(2311, length(levels(DT$V1)), 1L)
```
MichaelChirico, line 21120: nlevels() is available on very old R right?

inst/tests/tests.Rraw
```diff
@@ -21112,3 +21112,8 @@ test(2309.09, as.data.table(df, keep.rownames=TRUE), data.table(rn = c("a","b"),
 as.data.frame.no.reset = function(x) x
 DF = structure(list(a = 1:2), class = c("data.frame", "no.reset"), row.names = c(NA, -2L))
 test(2310.01, as.data.table(DF), data.table(a=1:2))
+
+# memrecycle() did not consider string encodings for factor levels #6886
+DT = data.table(factor(rep(enc2utf8("\uf8"), 3)))
+DT[1, V1 := iconv(levels(V1), from = "UTF-8", to = "latin1")]
```
MichaelChirico, line 21118: this test is _somewhat_ confusing, since the end result is basically "do nothing", maybe it'll help to be more practical?

e.g. `new_val = factor('a', levels = c('a', iconv(...)))`, then check that all we get in the update is an `a` level?

Re-reading the issue, I see your example is actually quite true to the original bug report. So maybe add this second example as another test instead?

MichaelChirico (approved):
LGTM, just a minor comment to see if we can't further improve the test a little. Feel free to submit when you see fit.

Thanks!

inst/tests/tests.Rraw
aitap: It's also a good idea to test both scenarios: assigning from a string and assigning from a factor.

ben-schwen:
@aitap apparently this is the regression for #7404 since the by columns in db-benchmark are factors.

```r
set.seed(42)

n_groups = 1e5
rows_per_group = 100
N = n_groups * rows_per_group

dt = data.table(
  id3 = as.factor(rep(seq_len(n_groups), each = rows_per_group)),
  v1  = rnorm(N),
  v2  = rnorm(N)
)

dt[sample.int(N, N * 0.05), v1 := NA_real_]
dt[sample.int(N, N * 0.05), v2 := NA_real_]

# 20 seconds after #6890
# 0.1 seconds before #6890
system.time(dt[, .(range_v1_v2 = max(v1, na.rm = TRUE) - min(v2, na.rm = TRUE)), by = id3])
```

PR #6891, opened 2025-04-01
tests: `structure(NULL, *)` is now an error

aitap:
As of R-devel [r88075](https://github.com/r-devel/r-svn/commit/942709c7ca5fca31f1663b127558fd17d4c0010f), `structure(NULL, ...)` signals an error instead of setting attributes on a new list. Test either behaviour depending on the R version and the SVN revision number.

We could also just remove the test, as `data.table` doesn't seem to be using `structure(NULL, ...)` in any of the tested code paths.


MichaelChirico (commented):
yea I'd just delete the test it seems ancient & not really useful.

our practice when deleting tests is to replace them with a gravestone comment: test X was removed because ...

PR #6895, opened 2025-04-02
removed goto statements, clarified control flow

badasahog:
all the pointers passed to the function are allocated with malloc/calloc and already restrict qualified, plus the recursive calls are all just offset versions of the parameters from the parent call

### DIFF
--- a/man/datatable-optimize.Rd
+++ b/man/datatable-optimize.Rd
@@ -31,8 +31,7 @@ it does so in "C-locale". This has some subtle implications; see Examples.
 out what level of optimisation is required. The default value \code{Inf}
 activates \emph{all} available optimisations.
 
-At optimisation level \code{>= 1}, i.e., \code{getOption("datatable.optimize")}
->= 1, these are the optimisations:
+For \code{getOption("datatable.optimize") >= 1}, these are the optimisations:
 
 \itemize{
     \item The base function \code{order} is internally replaced with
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Thanks for the change! What is the motivation here? it would help to elaborate the impact of this PR.

badasahog:
usually I wouldn't suggest a change like this because the compiler can infer these optimizations on its own, but it felt natural from a consistency point of view, because all the arguments being passed in were restrict-qualified, even though they came directly from malloc/calloc.

badasahog:
I've updated the pull request with an additional syntactic improvement for the function pointer table.

In c, all function designators are effectively just pointers, and derferencing a function pointer just yields the same function pointer, so the dereference only makes the code harder to read.

example:
```c
#include <stdio.h>

void pfunc()
{
    printf("Hello");
}

int main()
{
    if(pfunc == *pfunc && pfunc == **********pfunc)
    {
        printf("pointers are the same\n");
    }
}
```

badasahog:
I've improved the read logic to remove the `read` label and associated goto statements.

Currently, it's implemented with a while(true) loop, which isn't super, but I want to make sure maintainers are okay with this incremental improvement before I progress to move intrusive adjustments.

MichaelChirico:
Thanks for the change @badasahog. May I ask why they are included in this otherwise unrelated PR? It would seem best to me to add these changes into an independent PR split from current `master`.

badasahog:
@MichaelChirico that was an accident. I'm not sure how that happened, actually.

I thought I wiped my local repo and created a new fork, but I guess not.

\o/

badasahog:
would you be willing to merge if I separated out the changes?

I apologize, I'm not familiar with open source etiquette

MichaelChirico:
> would you be willing to merge if I separated out the changes?

well, I can never guarantee a merge ex-ante :)

but yes, i'm willing to review. both proposed changes (being better about `restrict` and avoiding `goto`) are good in principle.

> I apologize, I'm not familiar with open source etiquette

I'm not sure there's anything specific to FOSS here, just encouraging general best practices for code review

PR #6899, opened 2025-04-07
`groupingsets()`: introduce and make use of the `enclos` argument

aitap:
When forwarding the `j` argument for non-standard evaluation via `jj=`, also forward the environment containing the symbols that it may be referencing. Now `groupingsets()` `substitute`s all the local arguments into the `x[, jj, ...]` call and then evaluates it in the specified environment, making it possible for `jj` to refer to local symbols without confusing them to variables belonging to `groupingsets()` or its caller.

To avoid `cedta()` problems, set `.datatable.aware = TRUE` in the environment where the call is evaluated. (A column named `.datatable.aware` would be shadowed. So it goes.)

An alternative solution would give the `enclos=` argument to `[.data.table` instead of evaluating the `[` call inside `enclos`, but that would require carefully replacing a lot of `parent.frame()` calls. Also, `[.data.table` already has an `env=` argument, which operates on the expression directly, not provides an enclosing environment.

Another alternative solution could construct [Brodie Gaslam's DIY quosure](https://www.brodieg.com/2020/08/11/quosures/#quosures-au-naturel), but `[.data.table` doesn't seem to "see" inside such calls (for example, to replace the `.(...)` special form with `list(...)`) when given one as the `j` argument and I didn't investigate it further.

Fixes: #5560

### DIFF
--- a/R/groupingsets.R
+++ b/R/groupingsets.R
@@ -13,7 +13,7 @@ rollup.data.table = function(x, j, by, .SDcols, id = FALSE, label = NULL, ...) {
   sets = lapply(length(by):0L, function(i) by[0L:i])
   # redirect to workhorse function
   jj = substitute(j)
-  groupingsets.data.table(x, by=by, sets=sets, .SDcols=.SDcols, id=id, jj=jj, label=label)
+  groupingsets.data.table(x, by=by, sets=sets, .SDcols=.SDcols, id=id, jj=jj, label=label, enclos = parent.frame())
 }
 
 cube = function(x, ...) {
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Basically makes sense to me, though I'm not all that fluent with `groupingsets()` & friends -- please add a NEWS entry.

I am taking @jangorecki 's removal as reviewer to be "no opposition" to the extended API.

jangorecki:
Go ahead, will look and understand later on...

aitap:
This does need fixing, right? Do we want to discourage referring to variables outside the `data.table` in `j`-expressions?

jangorecki:
Yes, definitely. Thank you Ivan.

jangorecki:
@aitap news entry needs reference to issue or PR

MichaelChirico:
We could potentially add a lint rule for this, but it requires some more careful approach:

 1. There are plenty of NEWS entries lacking a GitHub reference now, e.g. https://github.com/Rdatatable/data.table/blob/7c59daaed1836db57747d92494b1ce96612bbf80/NEWS.md?plain=1#L98

    We'd need to come up with a way to ignore those (just ignore the `NOTES` section? Only require in the `BUG FIXES` section?), or decide to edit in a bunch of ex-post PR numbers. We'd probably want to exclude `NEWS.{0,1}` wholesale -- some of those entries predate GitHub hosting altogether.

 2. We'd probably want to use {commonmark} to parse the NEWS file, so have to be aware of a new dependency (more minor).

jangorecki:
I would leave the old ones as is and just use it now. We can test some linters but then maybe rather to switch then on after next NEWS.n file archive.

PR #6901, opened 2025-04-07
Set Jan as owner of some files

MichaelChirico:


jangorecki (commented):
Although I authored the code in those files I would rather not call myself file owner due to my limited resources for support of those...

MichaelChirico:
Thanks, makes sense. I still think you are the best default owner -- you can always unassign yourself/reassign if you're not available. I only see a handful of "real" changes in these files over the last 5 years.

PR #6915, opened 2025-04-17
Edit 'Calloc' et al on old versions for atime

MichaelChirico:
@tdhock it looks like atime is failing to compile older {data.table} commits with current versions of R, e.g.

```
bmerge.c: In function ‘bmerge’:
bmerge.c:107:16: warning: implicit declaration of function ‘Calloc’; did you mean ‘valloc’? [-Wimplicit-function-declaration]
  107 |     retFirst = Calloc(anslen, int); // anslen is set above
      |                ^~~~~~
      |                valloc
```

What's the right way to deal with that?

I took a stab at it here.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -117,6 +117,11 @@ test.list <- atime::atime_test_list(
       file.path("src", "init.c"),
       paste0("R_init_", Package_regex),
       paste0("R_init_", gsub("[.]", "_", new.Package_)))
+    # allow compilation on new R versions where 'Calloc' is not defined
+    pkg_find_replace(
+      file.path("src", "*.c"),
+      "\\b(Calloc|Free|Realloc)",
+      "R_\\1")
     pkg_find_replace(
       "NAMESPACE",
       sprintf('useDynLib\\("?%s"?', Package_regex),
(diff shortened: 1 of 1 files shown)


tdhock:
yes I think putting some code in pkg.edit.fun is the right approach

MichaelChirico:
OK I think this is WAI now

.ci/atime/tests.R
```diff
@@ -117,6 +117,11 @@ test.list <- atime::atime_test_list(
       file.path("src", "init.c"),
       paste0("R_init_", Package_regex),
       paste0("R_init_", gsub("[.]", "_", new.Package_)))
+    # allow compilation on new R versions where 'Calloc' is not defined
```
tdhock, line 120: maybe a a link to the PR#6915?

tdhock (approved):

.ci/atime/tests.R
MichaelChirico, line 120:
```suggestion
    # allow compilation on new R versions where 'Calloc' is not defined (i.e., apply #6382)
```

MichaelChirico: how about a reference to the original PR where this was applied?

tdhock:
lgtm with the reference to https://github.com/Rdatatable/data.table/pull/6382

PR #6916, opened 2025-04-17
Fix: Ensure column widths are data-driven when col.names = "none" in print.data.table 

### LINKED ISSUE #6882: print col.names='none' does not affect column width
In `print.data.table` when setting `col.names='none'` I expected column widths to be sized according to their contents. Instead, the column widths are unchanged. Incidentally, when class is suppressed the column widths are different. I expect the behavior to be consistent between those two options.

``` r
library(data.table)
dt<-data.table(x=3:1,sillylongcolumnname=1:3)
print(dt)
#>        x sillylongcolumnname
#>    <int>               <int>
#> 1:     3                   1
#> 2:     2                   2
#> 3:     1                   3
```

``` r
print(dt,col.names='none')
#> Warning: Column classes will be suppressed when col.names is 'none'
#> 1: 3                   1
#> 2: 2                   2
#> 3: 1                   3
```

<sup>Created on 2025-03-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

venom1204:
This PR resolves #6882 by ensuring that when col.names = "none" is specified in print.data.table, the column widths are determined solely by the data content, not by the (hidden) column names.

What was changed
Added the following logic to print.data.table:

```
if (col.names == "none") 
  colnames(toprint) <- rep.int("", ncol(toprint))
```
This is placed right after toprint is constructed, but before any formatting or width calculations.

This ensures that all downstream formatting and width calculations use empty column names, so column widths are always based on the data, fully resolving the issue where hidden long column names could cause excessive column width.

@MichaelChirico have a review of this when you have time
this is the change i was suggesting should i change it 
or it works fine let me know 

thank you.

### DIFF
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -261,8 +261,8 @@ char.trunc = function(x, trunc.char = getOption("datatable.prettyprint.char")) {
 dt_width = function(x, nrow, class, row.names, col.names) {
   widths = apply(nchar(x, type='width'), 2L, max)
   if (class) widths = pmax(widths, 6L)
-  names_widths = if (col.names == "none") rep(0L, ncol(x)) else sapply(colnames(x), nchar, type = "width")
-  dt_widths = pmax(widths, names_widths)
+  if (col.names != "none") names = sapply(colnames(x), nchar, type="width") else names = 0L
+  dt_widths = pmax(widths, names)
   rownum_width = if (row.names) as.integer(ceiling(log10(nrow))+2.0) else 0L
   cumsum(dt_widths + 1L) + rownum_width
 }
--- a/R/print.data.table.R
+++ b/R/print.data.table.R
@@ -139,6 +139,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     print_default(toprint)
     return(invisible(x))
   }
+  if (col.names == "none") 
+  colnames(toprint) <- rep.int("", ncol(toprint))
   if (nrow(toprint)>20L && col.names == "auto")
     # repeat colnames at the bottom if over 20 rows so you don't have to scroll up to see them
     #   option to shut this off per request of Oleg Bondar on SO, #1482
@@ -259,8 +261,8 @@ char.trunc = function(x, trunc.char = getOption("datatable.prettyprint.char")) {
 dt_width = function(x, nrow, class, row.names, col.names) {
   widths = apply(nchar(x, type='width'), 2L, max)
   if (class) widths = pmax(widths, 6L)
-  if (col.names != "none") names = sapply(colnames(x), nchar, type="width") else names = 0L
-  dt_widths = pmax(widths, names)
+  names_widths = if (col.names == "none") rep(0L, ncol(x)) else sapply(colnames(x), nchar, type = "width")
+  dt_widths = pmax(widths, names_widths)
   rownum_width = if (row.names) as.integer(ceiling(log10(nrow))+2.0) else 0L
   cumsum(dt_widths + 1L) + rownum_width
 }
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21130,7 +21130,7 @@ test(2311.2, nlevels(DT$V1), 2L) # used to be 3
 DF = list(rep(iconv("\uf8", from = "UTF-8", to = "latin1"), 2e6))
 test(2312, fwrite(DF, nullfile(), encoding = "UTF-8", nThread = 2L), NULL)
 
-# New tests for GitHub #6882 (col.names='none' column widths)
+# tests for #6882 (col.names='none' column widths)
 test(2313.1, {
   dt <- data.table(short=1:3, verylongcolumnname=4:6)
   print(dt, col.names="none")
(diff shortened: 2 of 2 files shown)


R/print.data.table.R
```diff
@@ -139,6 +139,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     print_default(toprint)
     return(invisible(x))
   }
+  if (col.names == "none")
```
MichaelChirico, line 142: did you try making this change on lines 94-96 above instead?

venom1204: sorry i forgot to include i will try it

venom1204: @MichaelChirico I tried implementing it 
However, when I tried to integrate this with the existing column name clearing logic as you 
Tests 2125.06 and 2125.07 fail, which involve truncation messages when col.names="none".

The issue appears to be related to how the truncation message logic works in trunc_cols_message(). When column names are cleared early in the combined condition, the truncation logic can no longer access the original column names, resulting in empty brackets in messages:
- Expected: 1 variable not shown: [d]
- Observed: 1 variable not shown: []

With the standalone implementation, the column names are likely cleared after the truncation message logic has already determined which columns to display, maintaining the correct behavior.

venom1204:
Hi @MichaelChirico ,
I tested the change you suggested, but encountered the error mentioned above. I’d appreciate your guidance on how to proceed further whenever you have the time.
Thank you!

R/print.data.table.R
```diff
@@ -139,6 +139,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
...
+  colnames(toprint) <- rep.int("", ncol(toprint))
```
joshhwuu, line 143: Use `=` instead of `<-` here:

https://github.com/Rdatatable/data.table/blob/master/.github/CONTRIBUTING.md#coding-style

R/print.data.table.R
```diff
@@ -292,4 +294,4 @@ trunc_cols_message = function(not_printed, abbs, class, col.names){
 knit_print.data.table = function(x, ...) {
   if (!shouldPrint(x)) return(invisible(x))
   NextMethod()
-}
+}
```
joshhwuu, line 297: Need a terminal newline

inst/tests/tests.Rraw
```diff
@@ -21129,3 +21129,17 @@ test(2311.2, nlevels(DT$V1), 2L) # used to be 3
...
+  print(dt, col.names="none")
+}, output="1: 1 4\n2: 2 5\n3: 3 6")
+test(2313.2, {
+  dt <- data.table(x=123456, y="wide_string")
+  print(dt, col.names="none")
+}, output="1: 123456 wide_string")
+test(2313.3, {
+  dt <- data.table(a=NA_integer_, b=NaN)
+  print(dt, col.names="none")
+}, output="1: NA NaN")
```
joshhwuu, line 21145: I believe this also needs a terminal newline

venom1204:
hi
@joshhwuu  i included your suggested imporvements ,can you please have a final review when you have time.
thanks...

R/print.data.table.R
venom1204: @joshhwuu  Thanks for the suggestion! Looking at the test outputs, I agree with your point about user expectations - if someone uses col.names="none", they're explicitly indicating they don't want to see column names.

Based on these test results and your feedback, I think the simplest approach would be to modify the truncation message logic to show a simpler message (without column names) when col.names="none" is used:

```
if (col.names == "none") {
  # Simpler message without column names
  catf("%d variable%s not shown\n", n, if(n==1L) "" else "s", domain=NA)
} else {
  # Original message with column names
  # ... existing code ...
}
```
This approach:
- Maintains consistency with user expectations
- Avoids the complexity of storing/restoring column names temporarily
- Keeps the consolidated condition as suggested by @MichaelChirico

Would this work for everyone? I can implement this change alongside the fix for the column width issue.

venom1204:
hi @joshhwuu ,
When you have a moment, could you please review and clarify what the next steps should be?
I’d like to ensure everything is on track so I can complete this PR.

inst/tests/tests.Rraw
```diff
@@ -21141,3 +21141,16 @@ dt = data.table(id = 1:25)
...
+  print(dt, col.names="none")
+}, output="1: 1 4\n2: 2 5\n3: 3 6")
+test(2315.2, {
+  dt <- data.table(x=123456, y="wide_string")
+  print(dt, col.names="none")
+}, output="1: 123456 wide_string")
+test(2315.3, {
+  dt <- data.table(a=NA_integer_, b=NaN)
+  print(dt, col.names="none")
+}, output="1: NA NaN\n")
```
Mukulyadav2004, line 21156: Just a couple of small suggestions to align with the style used in inst/tests/tests.Rraw:

- Initialize data.table outside the test() call.
- No need for {}, since each test() contains only a single expression.
- Add a comment above the test block like:
`# Tests for print() output with fixed index header (PR #6806)`

venom1204:
Hi @tdhock , @joshhwuu, @MichaelChirico,

Could you please guide me further on this?
1) Should the current placement and approach I've used be retained?
2) Or would you recommend an alternative implementation?

Looking forward to your feedback. Thanks!

tdhock:
I don't really know this code

MichaelChirico:
@venom1204 Mukul gave a comment in your tests, could you please address that first?

PR #6920, opened 2025-04-22
fix #6898, disable checking timezone in `between()`

### LINKED ISSUE #6898: fix tzone _error_
I know there is precedence for discussions on tzone, and I feel as if I asked something similar and was told that keeping the warnings was intentional for some reason. This is amplified because it's an unnecessary _error_.

Up front, my request is to add `options(datatable.ignore_tzone=TRUE)` to preempt all warnings and now this error. Comparisons of `POSIXt` values with different or missing time zones is perfectly valid, the only risk is _interpretation_ of the results when the user is under-informed on time zones. (R's representation of timestamps leaves a bit to be desired as well, different topic.)

# Reprex

```r
library(data.table)
# data.table 1.17.0 using 8 threads (see ?getDTthreads).  Latest news: r-datatable.com
tm2 <- tm1 <- Sys.time()
attr(tm1, "tzone")
# NULL
attr(tm2, "tzone") <- "America/Denver"
DT <- data.table(when=tm1 + -3:3)
DT[between(when, tm1-1, tm2+2),]
# Error in .checkTypos(e, names_x) : 
#   'between' lower= and upper= are both POSIXct but have different tzone attributes: ['', 'America/Denver']. Please align their time zones.
```

It is most definitely not an error in code. It _may_ be a mistake in the data if the user imported it incorrectly, but that's the case in anything. There is [precedence](https://github.com/Rdatatable/data.table/issues/4773#issuecomment-721749212) to not over-concerning ourselves (I'm speaking as a package dev, even if I'm just a contributor here) with some mistakes the user may make.

My request:

- At a minimum, convert this to a warning.
- Better, add `options(datatable.ignore_tzone=TRUE)` so that all protective warnings and errors can be avoided. This option may best protect against all of the existing unit-tests and in case any revdep actually relies on the warnings/errors being present.
- My preferred: do away with warnings and errors associated with different time zones.
- Last option: convert this to a warning, and emit each warning only once per R session.

I can write a PR for any of the above.

# Session info

```r
> sessionInfo()
R version 4.4.3 (2025-02-28)
Platform: aarch64-apple-darwin20
Running under: macOS Sequoia 15.3.2

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRblas.0.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.0

locale:

time zone: America/New_York
tzcode source: internal

attached base packages:

other attached packages:
[1] data.table_1.17.0

[1] compiler_4.4.3 tools_4.4.3    tcltk_4.4.3   
```

r2evans:
(Admittedly I'm having trouble compiling this on MacOS due to libomp symbol problems, unrelated to this PR. I believe the change and tests are straight-forward but have not fully tested it locally.)


r2evans:
I was about to also change the `stopf(.)` to `warningf(.)`, but the next line in one of the tz checks blocks says

```r 
  # otherwise the check in between.c that lower<=upper can (correctly) fail for this reason
```

Looking back at the history (around Sep 2019), it appears jangorecki added that comment with some changes to `between.c`, but I'm not following how differences in timezones would produce problems. I suspect that if a difference in zone _may_ produce mis-comparisons (or warnings/errors) in `between.c`, then this PR will re-trigger those problems.

I suggest @jangorecki review this or at least chime in with thoughts on (1) changing the error to a warning, and (2) how bypassing the check can possibly cause problems in `between.c`. Thanks!

R/between.R
```diff
@@ -16,7 +16,7 @@ between = function(x, lower, upper, incbounds=TRUE, NAbounds=TRUE, check=FALSE)
     stopifnot(is.px(x), is.px(lower), is.px(upper)) # nocov # internal
   }
   # POSIX check timezone match
-  if (is.px(x) && is.px(lower) && is.px(upper)) {
+  if (!ignore_tzone && is.px(x) && is.px(lower) && is.px(upper)) {
     tzs = sapply(list(x,lower,upper), function(x) {
```
MichaelChirico, line 20:
```suggestion
    tzs = vapply_1c(list(x,lower,upper), function(x) {
```

man/between.Rd
```diff
@@ -35,6 +35,8 @@ interpreted as \code{lower} and \code{y[[2]]} as \code{upper}.}
 It is set to \code{TRUE} by default for infix notations.}
 \item{NAbounds}{ If \code{lower} (\code{upper}) contains an \code{NA} what should \code{lower<=x} (\code{x<=upper}) return? By default \code{TRUE} so that a missing bound is interpreted as unlimited. }
 \item{check}{ Produce error if \code{any(lower>upper)}? \code{FALSE} by default for efficiency, in particular type \code{character}. }
+\item{ignore_tzone}{ \code{TRUE} means skip timezone checks among
+\code{x}, \code{lower}, and \code{upper}. }
```
MichaelChirico, line 39:
```suggestion
\item{ignore_tzone}{ \code{TRUE} means skip timezone checks among \code{x}, \code{lower}, and \code{upper}. }
```

inst/tests/tests.Rraw
```diff
@@ -21141,3 +21141,13 @@ dt = data.table(id = 1:25)
 test(2314.1, any(grepl("<int>", tail(capture.output(print(dt, class = TRUE)), 2))), TRUE)
 # Test that class=TRUE with col.names="top" doesn't show classes at bottom
 test(2314.2, !any(grepl("<int>", tail(capture.output(print(dt, class = TRUE, col.names = "top")), 2))), TRUE)
+
+# Test that tzone behavior stops warning with ignore_tzone=TRUE
+tm1 = .POSIXct(1, "UTC")
+dt1 = data.table(a = .POSIXct(0), b = .POSIXct(2, "UTC"))
```
MichaelChirico, line 21147: nit style comments:

 - no space around `=` in keyword args
 - use `=` for assignment
 
 (I'd offer suggested changes but it doesn't work in GitHub on the huge tests.Rraw file)

inst/tests/tests.Rraw
```diff
@@ -21141,3 +21141,13 @@ dt = data.table(id = 1:25)
...
+test(2315.1, dt1[, between(tm1, a, b)], error = "different tzone attributes")
+test(2315.2, dt1[, between(tm1, a, b, ignore_tzone=TRUE)])
+tm2 <- .POSIXct(1)
+dt2 = data.table(a = .POSIXct(0, "UTC"), b = .POSIXct(2, "UTC"))
+test(2315.3, dt2[, between(tm2, a, b)], message = "mismatched tzone attributes")
+test(2315.4, dt2[, between(tm2, a, b)])
```
MichaelChirico, line 21153: what's the difference vs. 2315.3?

inst/tests/tests.Rraw
```diff
@@ -21141,3 +21141,13 @@ dt = data.table(id = 1:25)
 test(2314.1, any(grepl("<int>", tail(capture.output(print(dt, class = TRUE)), 2))), TRUE)
 # Test that class=TRUE with col.names="top" doesn't show classes at bottom
 test(2314.2, !any(grepl("<int>", tail(capture.output(print(dt, class = TRUE, col.names = "top")), 2))), TRUE)
+
+# Test that tzone behavior stops warning with ignore_tzone=TRUE
+tm1 = .POSIXct(1, "UTC")
```
MichaelChirico, line 21146: nit: the structure of the test is odd, why put `a`/`b` in a data.table, but not `tm1`?

I would put either all or no `between()` inputs in the same table.

MichaelChirico:
Thanks! Basically LGTM. Agree about the need for Jan to weigh in, I also don't see how it could cause issues. I don't see anything in between.c at that point (4d355d65353eabfb2a5b58549d96b66e2266a951) that suggests the issue either:

https://github.com/Rdatatable/data.table/blob/4d355d65353eabfb2a5b58549d96b66e2266a951/src/between.c

inst/tests/tests.Rraw
r2evans: typo
(side note: I cannot check/compile on mac, it's failing with OMP missing symbols)

r2evans:
All fixed (and I'm able to build it locally again, after s-u recommended downgrading to XCode-16.2).

jangorecki:
Unfortunately it is too far in the past... I don't see now how that could be causing troubles.

R/between.R
```diff
@@ -16,15 +16,14 @@ between = function(x, lower, upper, incbounds=TRUE, NAbounds=TRUE, check=FALSE)
...
-  if (is.px(x) && is.px(lower) && is.px(upper)) {
-    tzs = sapply(list(x,lower,upper), function(x) {
+  if (!ignore_tzone && is.px(x) && is.px(lower) && is.px(upper)) {
+    tzs = vapply_1c(list(x,lower,upper), function(x) {
       attr(x, "tzone", exact=TRUE) %||% ""
     })
     # lower/upper should be more tightly linked than x/lower, so error
     #   if the former don't match but only inform if they latter don't
     if (tzs[2L]!=tzs[3L]) {
       stopf("'between' lower= and upper= are both POSIXct but have different tzone attributes: %s. Please align their time zones.", brackify(tzs[2:3], quote=TRUE))
```
MichaelChirico, line 26: I'm OK changing this to a warning if you'd like -- please just add a number of stress tests that the code WAI when time zones don't match.

R/between.R
```diff
@@ -16,15 +16,14 @@ between = function(x, lower, upper, incbounds=TRUE, NAbounds=TRUE, check=FALSE)
     stopifnot(is.px(x), is.px(lower), is.px(upper)) # nocov # internal
   }
   # POSIX check timezone match
-  if (is.px(x) && is.px(lower) && is.px(upper)) {
-    tzs = sapply(list(x,lower,upper), function(x) {
+  if (!ignore_tzone && is.px(x) && is.px(lower) && is.px(upper)) {
+    tzs = vapply_1c(list(x,lower,upper), function(x) {
       attr(x, "tzone", exact=TRUE) %||% ""
     })
```
MichaelChirico, line 22:
```suggestion
    tzs = vapply_1c(list(x,lower,upper), function(x) attr(x, "tzone", exact=TRUE) %||% "")
```

R/between.R
MichaelChirico: Hi @r2evans WDYT, do you want to demote this to `warning()`? Or perhaps leave as a separate PR?

PR #6913, opened 2025-04-17
Weaken testing of Date, POSIXct objects

MichaelChirico:
Closes #6912

PR #6940, opened 2025-05-01
Don't coerce to numeric when coercing IDate to Date

### LINKED ISSUE #6922: as.Date() and as.POSIXct() methods could leave integer storage
Recently, R is much more lax about the storage mode for `Date`, `POSIXct` objects.

I reckon we could take advantage of that and keep the object as integer when Date is requested.

This may break some downstream with overly-strict tests, but I don't think it will break any practical code.

MichaelChirico:
Closes #6922


MichaelChirico:
All of our tests pass, presumably because of #6913. I don't think any new test is warranted.

Issue #6860, opened 2025-03-10
Release data.table 1.17.2


MichaelChirico:
There are two regressions we should put into a patch:

 - #6851
 - #6852

We can leave this sit for at least another week to see if anything else comes up.

MichaelChirico:
Started the patch branch: [`patch-1.17.2`](https://github.com/Rdatatable/data.table/tree/patch-1.17.2). I also included the non-API tweaks: #6422, #6849, #6850

TysonStanley:
This should have what we need (including the https://github.com/Rdatatable/data.table/commit/476de7e399c925c3bd71deac91de0ea3899fb495 cherry-picked). Checks are passing (other than a note about the C-API still of course).

TysonStanley:
I didn't bring those into this patch branch. If we want those included, we can. I have submitted the current patch-1.17.2 to CRAN but it is easy enough to resubmit after including these.

TysonStanley:
That is, I can withdraw the current submission to CRAN and upload version with those fixes.

TysonStanley:
No problem, will do that shortly and update the CRAN submission.

TysonStanley:
There were some conflicts but both changes should be included now and in the NEWS. Going to submit updated version after running all checks tonight.

TysonStanley:
I'm getting errors in my checks, wondering if I messed anything up in the cherry-picking. @aitap @MichaelChirico ideas?

The errors are for tests/tests.Rraw.bz2 for test numbers 2309.03, 2309.05, 2309.06, 2311.1, 2311.2 

```
Test 2309.03 ran without errors but failed check that x equals y:
> x = key(as.data.table(DT, key = "a")) 
First 1 of 1 (type 'character'): 
[1] "b"
> y = "a" 
First 1 of 1 (type 'character'): 
[1] "a"
1 string mismatch
Test 2309.05 ran without errors but failed check that x equals y:
> x = key(as.data.table(DT, key = NULL)) 
First 1 of 1 (type 'character'): 
[1] "b"
> y = NULL 
First 0 of 0 (type 'NULL'): 
NULL
Test 2309.06 ran without errors but failed check that x equals y:
> x = key(as.data.table(DT, key = "a")) 
First 0 of 0 (type 'NULL'): 
NULL
> y = "a" 
First 1 of 1 (type 'character'): 
[1] "a"
Test 2311.1 ran without errors but failed check that x equals y:
> x = nlevels(DT$V1) 
[...]

PR #6941, opened 2025-05-01
Replace links that fail URL checks

aitap:
Current candidate for 1.17.2 (#6860) has two URLs that get flagged by `R CMD check --as-cran`. One redirects to a 404 error page, another is a Cloudflare false positive but still could be replaced by the official documentation.

Replace these links with ones that don't fail `curl -I`. If this is a good idea, consider cherry-picking into `patch-1.17.2`.

### DIFF
--- a/man/fread.Rd
+++ b/man/fread.Rd
@@ -139,7 +139,7 @@ Background :\cr
 \url{https://stackoverflow.com/a/9818473/403310}\cr
 \url{https://stackoverflow.com/questions/9608950/reading-huge-files-using-memory-mapped-files}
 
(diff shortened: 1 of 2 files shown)


TysonStanley:
Yes, was going to include that in our bigger release as CRAN hasn't pushed back on those (I thought both were false positives but I see the one that just I sent leading to an actual answer). But seems to be something we should do now. Going to cherry pick.

MichaelChirico:
Merging this to `master`, regardless of what ends up on CRAN (we can re-sync later if needed).

PR #6943, opened 2025-05-04
made `typeLetter` global string constant

badasahog:
making global objects const, including string literals, is considered good practice and improves both readability and compiler optimizations


MichaelChirico:
Thanks @badasahog! I've sent you an invite to become a project member. That should make future collaboration slightly easier by allowing you to write branches on this repo directly. Please do so going forward, and please add yourself to the DESCRIPTION in a follow-up PR.

PR #6948, opened 2025-05-07
Removed scoped macro definition

badasahog:
Scoped macro definitions are typically discouraged, as they don't respect scope boundaries. This can dilute the namespace of other functions are create readability issues. I think the all caps name suitably indicates its intent.

This change cannot be equally applied to the other inline macro definitions because they are used for array declarations and thus need to evaluate to constant expressions.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 3. `print.data.table()` now shows column classes at the bottom of large tables when `class=TRUE` and `col.names="auto"` (default) for tables with more than 20 rows, [#6902](https://github.com/Rdatatable/data.table/issues/6902). This follows the same behavior as column names at the bottom, making it easier to see column types for large tables without scrolling back to the top. Thanks to @TimTaylor for the suggestion and @Mukulyadav2004 for the PR.
 
+4. `as.Date()` method for `IDate` no longer coerces to `double` [#6922](https://github.com/Rdatatable/data.table/issues/6922). Thanks @MichaelChirico for the report and PR. The only effect should be on overly-strict tests that assert `Date` objects have `double` storage, which is not in general true, especially from R 4.5.0.
+
 ## BUG FIXES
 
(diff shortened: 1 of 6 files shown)


MichaelChirico:
Thanks, LGTM. Could you comment on the choice of `int_fast32_t` specifically?

What about just moving the macro out of "scope"? Would a header guard help?

https://github.com/Rdatatable/data.table/blob/cafe337f39daec6d67cdb51128ffaba1bbc9d8b4/src/fread.c#L109-L112

> This change cannot be equally applied to the other inline macro definitions

Just to be clear, this is the type of macro you're referring to here, right?

https://github.com/Rdatatable/data.table/blob/cafe337f39daec6d67cdb51128ffaba1bbc9d8b4/src/fread.c#L412-L413

MichaelChirico:
Thanks. Could you also comment on any performance tradeoff?

The macro will just be substituted at compile time, quite appealing; how will that translate here?

I think because it's `static const` it will just be defined once at run-time and then live on indefinitely? So there's not really much disadvantage?

Or will this be re-instantiated on every invocation of `parse_double_regular_core`, so that we have to worry about whether that's optimized away and/or `parse_double_regular_core` is invoked a very large number of times (I think it will be).

If you think it's worthwhile to confirm the performance aspect, you could re-open this PR on the repo directly, then the `atime` GHA will function as intended to do performance regression testing.

badasahog:
There's no performance loss.

Both clang and gcc will inline `static const`-qualified objects, and the object itself will then be abstracted away if the address is never taken.

This is technically also true if the object is neither static nor const, but putting this in explicitly improves clarity.

proof here:

https://godbolt.org/z/4n3GsPWEj

In terms of speed, even if the backing storage for the object was not eliminated, adding 4 bytes to the binary would not create a measurable difference on any modern system.

MichaelChirico:
Thanks for the patient explanation!

MichaelChirico (approved):

MichaelChirico:
> This change cannot be equally applied to the other inline macro definitions

Do you think we should explicitly `#undef` in those cases?

badasahog:
`#undef` might make sense. I want a little more time to assess.

I will have a pr ready hopefully tomorrow.

PR #6952, opened 2025-05-08
Use new grepv internally (with backport)

MichaelChirico:
`grepv()` is a nice new wrapper of `grep()` for subsetting a vector. I think it's a very nice addition -- it saves some characters and makes `grep()` less of a Swiss army knife (`grep()` --> integer output, `grepv()` --> string output).


jangorecki (approved):
LGTM

PR #6953, opened 2025-05-08
False negatives in alloc check linter

MichaelChirico:
As discovered in #6951 -- by omitting the valid whitespace checks, we missed a few cases in the earlier sweep.

Also make clear in the failure message that our convention is `!ptr`, not `ptr==NULL`.

Split off for clarity.

### DIFF
--- a/.ci/linters/c/alloc_linter.R
+++ b/.ci/linters/c/alloc_linter.R
@@ -6,7 +6,7 @@ alloc_linter = function(c_obj) {
   lines = c_obj$lines
   # Be a bit more precise to avoid mentions in comments, and allow
   #   malloc(0) to be used for convenience (e.g. #6757)
-  alloc_lines = grep(R"{=\s*([(]\w+\s*[*][)])?\s*[mc]alloc[(][^0]}", lines)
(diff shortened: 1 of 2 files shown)


.ci/linters/c/alloc_linter.R
```diff
@@ -6,7 +6,7 @@ alloc_linter = function(c_obj) {
   lines = c_obj$lines
   # Be a bit more precise to avoid mentions in comments, and allow
   #   malloc(0) to be used for convenience (e.g. #6757)
-  alloc_lines = grep(R"{=\s*([(]\w+\s*[*][)])?[mc]alloc[(][^0]}", lines)
+  alloc_lines = grep(R"{=\s*([(]\w+\s*[*][)])?\s*[mc]alloc[(][^0]}", lines)
```
MichaelChirico, line 9:
```suggestion
  alloc_lines = grep(R"{=\s*([(]\s*\w+\s*[*]\s*[)])?\s*[mc]alloc[(][^0]}", lines)
```

MichaelChirico:
Merging as effectively a GHA-only change.

PR #6956, opened 2025-05-09
Removed macros, improved code readability

badasahog:
Follow-up to #6948.

I felt that the `NSUFFIXES` macro was unnecessary and unhelpful.

The whole point of macros like this are to improve readability and specific the meaning behind numbers. In this case, I felt that `NSUFFIXES` was not more clear than `sizeof(suffixes)`.

In addition, `static char output[BUFFSIZE];` is not more clear or readable than just putting the size in directly. The reader already knows that the number in the subscript is the buffer size. It just serves to hide the actual size to a few lines up.

Later, using `sizeof(output)` is also more clear than `BUFFSIZE` because it makes it clear exactly which buffer the size is referring to.

Last but not least, storing `"1 byte"` in a temporary object is pointless, the reader will understand the meaning of the string literal `"1 byte"` without giving it a name.

This patch reduces the size of the function by 2 lines, and removes the remaining 2 scoped macro definitions in this file.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -409,35 +409,33 @@ double wallclock(void)
  */
 static const char* filesize_to_str(size_t fsize)
 {
-  #define NSUFFIXES 4
-  #define BUFFSIZE 100
-  static char suffixes[NSUFFIXES] = {'T', 'G', 'M', 'K'};
-  static char output[BUFFSIZE];
-  static const char one_byte[] = "1 byte";
+  static const char suffixes[] = {'T', 'G', 'M', 'K'};
+  static char output[100];
   size_t lsize = fsize;
-  for (int i = 0; i <= NSUFFIXES; i++) {
-    int shift = (NSUFFIXES - i) * 10;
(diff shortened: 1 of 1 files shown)


badasahog:
@aitap I'll work on that

badasahog:
you want it added to this pr or can I work on another one? I'm at EOD

MichaelChirico:
> Since we're rewriting this function, how about using [proper binary SI prefixes](https://bugs.r-project.org/show_bug.cgi?id=18297) too?

I would do so in a separate PR, it has some potential back-compatibility implications, so it'll be easier to isolate in `bisect` if it's committed independently.

MichaelChirico (approved):
LGTM. I think learning that `sizeof()` is not a normal function is going to take me a while to get used to :)

MichaelChirico:
@badasahog reminder to please try and accept the invite to become a project Member so you can create branches directly on this repo, thanks again!

PR #6958, opened 2025-05-11
removed redundant header `stdbool.h`

badasahog:


MichaelChirico:
Thanks for the change -- could you add more context to the PR description? At a minimum, why it's redundant, but perhaps also (for edification) how you found this, or any other useful details for future readers.

badasahog:
@MichaelChirico It's redundant because it's included multiple times. I didn't really know what to say past that. Seems like a pretty simple change

MichaelChirico:
for example, linking to the "canonical" inclusion that makes these two redundant, would be helpful

badasahog:
@MichaelChirico that's not the issue.

There's already a canonical inclusion outside the `ifdef`

https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L28

MichaelChirico:
thanks, that's the one I missed :)

MichaelChirico:
BTW if you click for the perma-link, GitHub will auto-render a preview of the file lines highlighted.

It doesn't do so for `master` references since the those specific lines will generally change over time.

MichaelChirico (approved):

PR #6959, opened 2025-05-12
removed `size_t` cast

badasahog:
In many places, there was an explicit cast to `size_t`, but this is clunky and unnecessary because the sizeof operator always evaluates to type `size_t`

https://godbolt.org/z/54x1dMsWP


MichaelChirico:
LGTM, is this something Coccinelle can help with @aitap?

PR #6963, opened 2025-05-12
removed redundant extra object lsize

badasahog:
There was an object called `lsize` that was assigned to using the `fsize` parameter but never modified. I just combined them together into one.


MichaelChirico (approved):
LGTM, thanks!

PR #6965, opened 2025-05-12
Remove Unused HAVE_LONG_DOUBLE Conditional and Use sqrtl Directly 

### LINKED ISSUE #6938: HAVE_LONG_DOUBLE is conditioned on but never set
In gsumm.c, `HAVE_LONG_DOUBLE` is queried but it is never set (it is an internal value of R that seems to have come over from some copy-pasta).

https://github.com/Rdatatable/data.table/blob/a850eed4a469b1217ed920e703d876596a237188/src/gsumm.c#L24-L29

I think you can safely replace SQRTL by direct calls to sqrtl.

See [thread from r-package-devel](https://stat.ethz.ch/pipermail/r-package-devel/2025q2/011627.html) which was motivated by a similarish issue in the matrixStats package.

venom1204:
Closes #6938

This PR addresses the unused HAVE_LONG_DOUBLE conditional in src/gsumm.c and simplifies the handling of square root calculations for variance and standard deviation. The following changes have been implemented:
- Removed Unused Conditional Compilation Block:
- The preprocessor block checking for HAVE_LONG_DOUBLE (which was never defined in the package) has been removed to eliminate dead code and potential confusion.

Direct Use of sqrtl:
- All occurrences of the SQRTL macro have been replaced with direct calls to sqrtl. This ensures that calculations use the higher-precision long double square root function, aligning with R's original intent for statistical computations.

All tests pass locally, and there are no regressions in numerical results.

Kindly review when you have time.
Thank you!

### DIFF
--- a/src/gsumm.c
+++ b/src/gsumm.c
@@ -21,13 +21,6 @@ static int *oo = NULL;
 static int *ff = NULL;
 static int isunsorted = 0;
 
-// from R's src/cov.c (for variance / sd)
-#ifdef HAVE_LONG_DOUBLE
-# define SQRTL sqrtl
-#else
-# define SQRTL sqrt
-#endif
-
 static int nbit(int n)
 {
   // returns position of biggest bit; i.e. floor(log2(n))+1 without using fpa
@@ -1060,7 +1053,7 @@ static SEXP gvarsd1(SEXP x, SEXP narmArg, bool isSD)
           v += (subd[j]-(double)m) * (subd[j]-(double)m);
         }
         ansd[i] = (double)v/(nna-1);
-        if (isSD) ansd[i] = SQRTL(ansd[i]);
+        if (isSD) ansd[i] = sqrtl(ansd[i]);
       }
     }}
     break;
@@ -1091,7 +1084,7 @@ static SEXP gvarsd1(SEXP x, SEXP narmArg, bool isSD)
           v += (subd[j]-(double)m) * (subd[j]-(double)m);
         }
         ansd[i] = (double)v/(nna-1);
-        if (isSD) ansd[i] = SQRTL(ansd[i]);
+        if (isSD) ansd[i] = sqrtl(ansd[i]);
       }
     }}
     break;
--- a/src/gsumm.c
+++ b/src/gsumm.c
@@ -1053,7 +1053,7 @@ static SEXP gvarsd1(SEXP x, SEXP narmArg, bool isSD)
           v += (subd[j]-(double)m) * (subd[j]-(double)m);
         }
         ansd[i] = (double)v/(nna-1);
-        if (isSD) ansd[i] = sqrtl(ansd[i]);
+        if (isSD) ansd[i] = (double)sqrtl((long double)ansd[i]);
       }
     }}
     break;
@@ -1084,7 +1084,7 @@ static SEXP gvarsd1(SEXP x, SEXP narmArg, bool isSD)
           v += (subd[j]-(double)m) * (subd[j]-(double)m);
         }
         ansd[i] = (double)v/(nna-1);
-        if (isSD) ansd[i] = sqrtl(ansd[i]);
+        if (isSD) ansd[i] = (double)sqrtl((long double)ansd[i]);
       }
     }}
     break;


MichaelChirico:
Shouldn't we just use `sqrt()`? cc @TimTaylor

 - `sqrt()` has signature `double sqrt(double arg);`
 - that matches the type of `ansd`:
 
 https://github.com/Rdatatable/data.table/blob/b0ef41eccd105c5ff51e700aa7e28fc68f934297/src/gsumm.c#L1032

I'm not sure our case maps perfectly well onto that of {matrixStats}, [as noted in the linked issue](https://github.com/HenrikBengtsson/matrixStats/issues/278#issuecomment-2841321469), the `HAVE_LONG_DOUBLE` use there only concerns summed doubles, which is a lot less finicky operation as compared to `sqrt()`... I guess the thinking in base R is that we can _maybe_ achieve a tiny bit better accuracy by doing `(double)sqrtl((long double)x)` as opposed to just `sqrt(x)`. Could be.

R itself apparently keeps going back and forth on using `sqrt()` vs `sqrtl()`, e.g.

https://github.com/r-devel/r-svn/commit/cbdae1e498681337a20be4bfeaccbeadf80dfbcb
https://github.com/r-devel/r-svn/commit/109c6cb94200850aebb75bcfad00f7b82b314cae
https://github.com/r-devel/r-svn/commit/0286a2e213869cb2cd8465c6fc4ea3bd0e5a4473
https://github.com/r-devel/r-svn/commit/a916f3fbe57d86eb958c6699cecbf28d0d35a0c5

At a minimum, if we use `sqrtl()` it would be appropriate IMO to have the cast to `double` explicit (as it is in cov.c).

venom1204:
Hi @MichaelChirico , @TimTaylor ,
Thank you for the insightful discussion. Based on your feedback about precision considerations and R's implementation patterns, I've modified the approach to include explicit type casting in gsumm.c:

```
ansd[i] = (double)sqrtl((long double)val);
```
Rationale for this approach:

- Precision Preservation: Maintains long double precision during sqrt calculation before downcasting to double, aligning with R's practice in src/library/stats/src/cov.c.
- Type Safety: Explicit casting avoids implicit conversion warnings and ensures compatibility with REAL(ans) type requirements.

If this balance doesn't align with data.table's design goals, I'm happy to:
a) Revert to plain sqrt() with your confirmation
b) Explore alternative precision-preserving strategies

Please let me know your preferred direction. I appreciate your guidance on balancing precision and performance in this critical path.

MichaelChirico:
What I'd like to see is a numeric study of the differences.

venom1204:
Thanks for your feedback. I’ll try to work on a numeric study comparing the precision and performance and will share the results here as soon as possible.

venom1204:
Hi @MichaelChirico,

Thanks for the direction regarding precision. I’ve completed a preliminary numeric and performance study comparing sqrt() and (double)sqrtl((long double)x) as applied in gsumm.c. Here's a quick summary:
Precision Comparison
I ran tests on a wide range of values including edge cases (very small and very large numbers). Results:
-For typical values (e.g., 2.0, 1e-8, 1e15), no precision difference was observed. Hex representations matched exactly.
- For very large values like 1.797693e+308, there was a measurable difference:
- sqrt() returned a slightly smaller result than (double)sqrtl(...), with a difference of ~1.49e+138.

Performance Comparison (performance_test.R)
Using microbenchmark in R:
```
Unit: milliseconds
      expr      min       lq     mean   median       uq       max neval
    R_sqrt   39.54    40.97    43.15    42.93    44.71    48.79    10
 Cpp_sqrtl 49.57    53.12    64.91    59.54    62.69   128.75    10
```
On average, the sqrtl version was ~50-60% slower, though still well within reasonable performance for typical datasets.

MichaelChirico:
please include a repro script

venom1204:
Thanks for the follow-up. Here's the full reproducible setup I used for both the precision and performance comparisons.

- Precision Test (precision_test.c) 
 ```
 #include <stdio.h>
#include <math.h>
#include <float.h>

int main() {
    double test_values[] = {2.0, 1e-8, 1e+15, 1.23456789e-5, DBL_MIN, DBL_MAX};

    printf("Precision Comparison:\n");
    for(int i=0; i<6; i++) {
        double val = test_values[i];
        double sqrt_std = sqrt(val);
        double sqrtl_cast = (double)sqrtl((long double)val);

        printf("\nValue: %e\n", val);
        printf("Standard sqrt: %.20f\n", sqrt_std);
        printf("sqrtl cast: %.20f\n", sqrtl_cast);
        printf("Difference: %e\n", sqrtl_cast - sqrt_std);
        printf("sqrt_hex: %a\n", sqrt_std);
        printf("sqrtl_hex: %a\n", sqrtl_cast);
    }
    return 0;
}
```

Performance Benchmark (sqrt_test.cpp + performance_test.R)
- sqrt_test.cpp
```
#include <Rcpp.h>
#include <cmath>

// [[Rcpp::export]]
Rcpp::NumericVector cpp_sqrt(Rcpp::NumericVector x) {
    return Rcpp::sqrt(x);
}

// [[Rcpp::export]]
Rcpp::NumericVector cpp_sqrtl(Rcpp::NumericVector x) {
    Rcpp::NumericVector out(x.size());
    for(int i = 0; i < x.size(); i++) {
        out[i] = (double)sqrtl((long double)x[i]);
    }
    return out;
}
```
- performance_test.R
```
library(Rcpp)
library(microbenchmark)

sourceCpp("sqrt_test.cpp")

n <- 1e7
x <- runif(n)

mb <- microbenchmark(
  R_sqrt = sqrt(x),
  Cpp_sqrtl = cpp_sqrtl(x),
  times = 10
)

print(mb)
```

PR #6969, opened 2025-05-13
Type change from `size_t` to `ptrdiff_t`

badasahog:
The intended purpose of the `size_t` type is to denote the size of an object, or a multiple thereof. It is the evaluated type of the `sizeof` operator.

The `ptrdiff_t` type denotes the difference between two memory addresses.

While they are *usually* identical, they are not intended to be used interchangeably.


badasahog:
I merged this on top of a different pr. That's my fault.

PR #6970, opened 2025-05-13
Type change from `size_t` to `ptrdiff_t`

badasahog:
The intended purpose of the `size_t` type is to denote the size of an object, or a multiple thereof. It is the evaluated type of the `sizeof` operator.

The `ptrdiff_t` type denotes the difference between two memory addresses.

While they are *usually* identical, they are not intended to be used interchangeably.


src/fread.c
```diff
@@ -2798,4 +2798,4 @@ int freadMain(freadMainArgs _args) {
   }
   freadCleanup();
   return 1;
-}
+}
```
MichaelChirico, line 2801:
```suggestion
}

```

src/fread.c
```diff
@@ -1918,7 +1918,7 @@ int freadMain(freadMainArgs _args) {
       DTPRINT(_("  Number of sampling jump points = %d because jump0size==0\n"), nJumps);
     } else {
       DTPRINT(_("  Number of sampling jump points = %d because (%"PRIu64" bytes from row 1 to eof) / (2 * %"PRIu64" jump0size) == %"PRIu64"\n"),
-              nJumps, (uint64_t)sz, (uint64_t)jump0size, (uint64_t)(sz/(2*jump0size)));
+              nJumps, sz, jump0size, sz/(2*jump0size));
```
MichaelChirico, line 1921: Shouldn't this use `%td`?

https://stackoverflow.com/a/7954455/3576984

badasahog: I don't see any reference to `%td` in the language standard, including the latest version.

https://www.open-std.org/jtc1/sc22/wg14/www/docs/[n3088.pdf](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3088.pdf)

MichaelChirico: it's there:

text on ptrdiff_t from "The length modifiers and their meanings are:"

MichaelChirico: Read more carefully -- `t` specifies that the following `d` conversion specifier applies to a `ptrdiff_t`.

Gemini also spits out a bunch more elaboration:

----

For `printf`, the specifier used for `ptrdiff_t` is `t` combined with an integer conversion specifier. `ptrdiff_t` is a signed integer type, so the most common specifiers are:

 - `%td`: For printing `ptrdiff_t` as a signed decimal integer.
 - `%ti`: Also for printing `ptrdiff_t` as a signed decimal integer.

You can also use other integer conversion specifiers with the `t` length modifier:

 - `%to`: For octal output.
 - `%tu`: For unsigned decimal output (though `ptrdiff_t` is signed, so this would represent the bit pattern as unsigned).
 - `%tx` or `%tX`: For hexadecimal output (lowercase or uppercase).

Example:

```c
#include <stdio.h>
#include <stddef.h> // For ptrdiff_t

int main() {
    int arr[10];
    int *p1 = &arr[2];
    int *p2 = &arr[7];
    ptrdiff_t diff = p2 - p1;

    printf("The difference is: %td\n", diff); // Prints "The difference is: 5"
    return 0;
}
```

It's important to note that the `t` length modifier was introduced in C99. While widely supported now, very old compilers might not recognize it. In such legacy environments, you might see `ptrdiff_t` values cast to a long and printed with `%ld`. However, for modern C programming, `%td` is the correct and portable way to print a `ptrdiff_t`.

badasahog: Got it. Makes sense.

src/fread.c
```diff
@@ -1917,8 +1917,8 @@ int freadMain(freadMainArgs _args) {
     } else if (jump0size==0) {
       DTPRINT(_("  Number of sampling jump points = %d because jump0size==0\n"), nJumps);
     } else {
-      DTPRINT(_("  Number of sampling jump points = %d because (%"PRIu64" bytes from row 1 to eof) / (2 * %"PRIu64" jump0size) == %"PRIu64"\n"),
-              nJumps, (uint64_t)sz, (uint64_t)jump0size, (uint64_t)(sz/(2*jump0size)));
+      DTPRINT(_("  Number of sampling jump points = %d because (%td bytes from row 1 to eof) / (2 * %td jump0size) == %td\n"),
+              nJumps, sz, jump0size, sz/(2*jump0size));
```
MichaelChirico, line 1921: One more note to self -- `sz / (2*jump0size)` will do integer division --> therefore is also `ptrdiff_t` --> `%td` is right.

MichaelChirico (approved):

PR #6972, opened 2025-05-13
removed `goto` labels

badasahog:
I'm working on getting the `goto` statements and associated labels.

The ones that are part of the type bump system are more complex, so I'm gonna make a separate make a separate PR for it.

I'm also saving `parse_double_extended` for later since I intend to rewrite that function to be more readable and efficient.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -681,6 +681,8 @@ static void parse_double_regular_core(const char **pch, double *target)
   static const int_fast32_t FLOAT_MAX_DIGITS = 18;
   const char *ch = *pch;
 
+  *target = NA_FLOAT64;
+
   if (*ch=='0' && args.keepLeadingZeros && IS_DIGIT(ch[1])) return;
   bool neg, Eneg;
   ch += (neg = *ch=='-') + (*ch=='+');
@@ -710,7 +712,7 @@ static void parse_double_regular_core(const char **pch, double *target)
       ch++;
       e++;
     }
-    if (*ch!=dec && *ch!='e' && *ch!='E') goto fail;
+    if (*ch!=dec && *ch!='e' && *ch!='E') return;
   }
 
   // Read the fractional part of the number, if it's present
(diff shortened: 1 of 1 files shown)


src/fread.c
```diff
@@ -1009,27 +1003,34 @@ static void parse_iso8601_timestamp(FieldParseContext *ctx)
...
+    *(ctx->ch) = ch;
+    return;
+  }
+  // allows date-only field in a column with UTC-marked datetimes to be parsed as UTC too; test 2150.13
   ch++;
 
   str_to_i32_core(&ch, &hour, true);
   if (hour == NA_INT32 || hour < 0 || hour > 23 || *ch != ':')
-    goto fail;
+    return;
```
MichaelChirico, line 1023: looks like these are previously uncovered too:

https://app.codecov.io/gh/Rdatatable/data.table/blob/master/src%2Ffread.c#L1022

Would you mind adding some unit tests while we're here?

src/fread.c
```diff
@@ -1009,27 +1003,34 @@ static void parse_iso8601_timestamp(FieldParseContext *ctx)
...
+  *target = NA_FLOAT64;
+
   parse_iso8601_date_core(&ch, &date);
   if (date == NA_INT32)
-    goto fail;
+    return;
   if (*ch != ' ' && *ch != 'T')
-    goto date_only;
-    // allows date-only field in a column with UTC-marked datetimes to be parsed as UTC too; test 2150.13
+  {
```
MichaelChirico, line 1012: nit: move `{` to the previous line

badasahog: Sure.

Did you do this, or are you asking me to fix it?

Sorry, I'm new to open source.

MichaelChirico: I tried to do it but GitHub doesn't like suggested changes that mix newly-added lines & existing lines, it seems...

src/fread.c
MichaelChirico: This is more of a boyscout rule request:

https://iamjosephmj.medium.com/the-boy-scout-rule-leaving-code-better-than-you-found-it-fc21857ef8a1

MichaelChirico (approved):

Issue #6967, opened 2025-05-13
Feature request: `GET_UNBUMPED` and `MAKE_BUMPED`


badasahog:
Right now the way we denote type bumps is with negative numbers, which is fine, but I want to make that behavior more clear.

The meaning behind something like `abs(type)` doesn't clearly communicate the intent. Something like `GET_UNBUMPED` does.

In addition, the meaning of `-type` is not nearly as clear of `MAKE_BUMPED`

The type deduction/type bump system is a common source of bugs, so any improvement in clarity would seem to me to be well worth it.

I'm very open to suggestions for what the macros should be called

MichaelChirico:
Sounds good to me, let's discuss precise details in a PR

badasahog:
Do you have an notes about what the macros should be called?

MichaelChirico:
Your suggestion looks roughly right, it will be easier to make a final call by assessing within the context of the actual code that's being updated.

badasahog:
@MichaelChirico sure. I'll get to work.

This this will includes dozens of lines I'm gonna wait until my other PRs are either merged or blocked.

PR #6976, opened 2025-05-13
added type bump macros

badasahog:
Satisfies #6967

I think I got all the instances of these mechanisms being used, but they're hard to spot (that's why I suggested using macros)

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -109,6 +109,9 @@ static void Field(FieldParseContext *ctx);
 #define AS_DIGIT(x) (uint_fast8_t)(x - '0')
 #define IS_DIGIT(x) AS_DIGIT(x) < 10
 
+#define GET_UNBUMPED(x) abs(x)
+#define MAKE_BUMPED(x) (-x)
+
 //=================================================================================================
 //
 //   Utility functions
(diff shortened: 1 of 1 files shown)


src/fread.c
```diff
@@ -109,6 +109,9 @@ static void Field(FieldParseContext *ctx);
 #define AS_DIGIT(x) (uint_fast8_t)(x - '0')
 #define IS_DIGIT(x) AS_DIGIT(x) < 10
 
+#define GET_UNBUMPED(x) abs(x)
```
MichaelChirico, line 112: WDYT about `MARK_BUMPED` and `UNMARK_BUMPED`?

badasahog: I'm not sure if that's better or worse. Up to you

badasahog: On second thought, using the term mark is incorrect imo.

When you mark something, you would presume that it would continue to work in the same way, but you can't use a type bumped value without unbumping it or you'll create a mess.

Still, it's up to you.

MichaelChirico: I see what you mean (`MARK` hinting at some side-effect), let's do `BUMP`/`UNBUMP` then.

badasahog: I don't like `BUMP` and `UNBUMP` because those (again, in my opinion) hint at external state, such as placement in a vector. The identifier `MAKE_BUMPED` clearly communicates that it only operates on its input argument.

MichaelChirico: OK, one more try :)

```c
#define BUMP_INDEPENDENT(x) abs(x) // OR maybe IGNORE_BUMP()?
#define TOGGLE_BUMP(x) (-x)
```

 - `abs(x)` is not really "get unbumped" -- it's used on both bumped and un-bumped inputs. The idea is to get the type regardless of if it's been "flipped".
 - Similarly, `-x` is not strictly about "making x bumped" -- the `-` operation applies (and is used) equally well for types already bumped, i.e., it's used both to "bump" and "unbump".

badasahog: I apologize if this comes across as oppositional, but I do think that the meaning here is important.

I don't think this makes sense, because while your proposed names do more accurately describe the behavior of these macros, it gives a misleading notion of how they're actually used.

`UNBUMP` is always used to get the unbumped version of a type, while `MAKE_BUMPED` specifies that the output is always bumped (which is the case in the code). It better indicates the intent, better than a toggle function that is only ever toggled in one direction.

badasahog: I'm not sure that's more clear. It is clearly referred to type bumping elsewhere in the code. Is there something wrong with my original naming structure.

It would seem odd to me to nitpick on this so hard when so many identifiers are unclear or follow inconsistent naming structures.

PR #6978, opened 2025-05-13
Use %ld, not %td

MichaelChirico:
Follow-up addressing https://github.com/Rdatatable/data.table/pull/6970#issuecomment-2877636459

### DIFF
--- a/.dev/CRAN_Release.cmd
+++ b/.dev/CRAN_Release.cmd
@@ -95,16 +95,18 @@ grep "PROTECT_PTR" ./src/*.c
 # No use of long long, instead use int64_t. TODO
 # grep "long long" ./src/*.c
 
-// No use of llu, lld, zd or zu
-grep -nE "(llu|lld|zd|zu)" src/*.[hc]
+// No use of llu, lld, td or tu
+grep -nE "%(ll|t)[ud]\b" src/*.[hc]
 // Comment moved here from fread.c on 19 Nov 2019
 // [Moved from fread.c on 19 Nov 2019] On Windows variables of type `size_t` cannot be printed
-// with "%zu" in the `snprintf()` function. For those variables we used to cast them into
-// `unsigned long long int` before printing, and defined (llu) to make the cast shorter.
-// We're now observing warnings from gcc-8 with -Wformat-extra-args, #4062. So
-// now we're more strict and cast to [u]int64_t and use PRIu64/PRId64 from <inttypes.h>
-// In many cases the format specifier is passed to our own macro (e.g. DTPRINT) or to Rprintf(),
-// error() etc, and even if they don't call sprintf() now, they could in future.
+//   with "%zu" in the `snprintf()` function. For those variables we used to cast them into
+//   `unsigned long long int` before printing, and defined (llu) to make the cast shorter.
+//   We're now observing warnings from gcc-8 with -Wformat-extra-args, #4062. So
+//   now we're more strict and cast to [u]int64_t and use PRIu64/PRId64 from <inttypes.h>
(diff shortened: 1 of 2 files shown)


src/fread.c
```diff
@@ -1911,8 +1911,9 @@ int freadMain(freadMainArgs _args) {
     } else if (jump0size==0) {
       DTPRINT(_("  Number of sampling jump points = %d because jump0size==0\n"), nJumps);
     } else {
-      DTPRINT(_("  Number of sampling jump points = %d because (%td bytes from row 1 to eof) / (2 * %td jump0size) == %td\n"),
-              nJumps, sz, jump0size, sz/(2*jump0size));
+      DTPRINT(_("  Number of sampling jump points = %d because (%ld bytes from row 1 to eof) / (2 * %ld jump0size) == %ld\n"),
+              // NB: ptrdiff_t should use '%td', but that's a C99 addition, which may not work with old Rtools (pre-Rtools42)
```
MichaelChirico, line 1915:
```suggestion
```

.dev/CRAN_Release.cmd
```diff
@@ -95,16 +95,18 @@ grep "PROTECT_PTR" ./src/*.c
...
+grep -nE "%(ll|t)[ud]\b" src/*.[hc]
 // Comment moved here from fread.c on 19 Nov 2019
 // [Moved from fread.c on 19 Nov 2019] On Windows variables of type `size_t` cannot be printed
-// with "%zu" in the `snprintf()` function. For those variables we used to cast them into
-// `unsigned long long int` before printing, and defined (llu) to make the cast shorter.
-// We're now observing warnings from gcc-8 with -Wformat-extra-args, #4062. So
-// now we're more strict and cast to [u]int64_t and use PRIu64/PRId64 from <inttypes.h>
-// In many cases the format specifier is passed to our own macro (e.g. DTPRINT) or to Rprintf(),
-// error() etc, and even if they don't call sprintf() now, they could in future.
+//   with "%zu" in the `snprintf()` function. For those variables we used to cast them into
```
MichaelChirico, line 102: Hmm, it seems we've already let plenty of `%zu` slip through, e.g.

https://github.com/Rdatatable/data.table/blob/7620b469f7c9f24eb95aebc362d11fe1b90f9f7e/src/reorder.c#L22

So I've just dropped that check. I also see som `%lld` have slipped through, e.g.

https://github.com/Rdatatable/data.table/blob/7620b469f7c9f24eb95aebc362d11fe1b90f9f7e/src/fifelse.c#L260

Is this check just obsolete?

MichaelChirico:
So it looks like basically, users building data.table from source on very old R (correspondingly very old Rtools) may get compiler warnings, but no functionality issue?

Maybe we should just let this slide -- possibly with some sort of log/warning/error from `./configure` about the potential issue?

MichaelChirico:
> When data.table calls snprintf to buffer a message from an OpenMP thread for future printing, C99 specifiers like %zu don't work, but %ll[udx] seem to do:

To clarify, that means that users (1) triggering these errors formatted through `snprintf()`; and (2) on versions of data.table built with Rtools<4.2; will experience weird errors because `%zu` is just not recognized, either `snprintf()` will error due to e.g. `%d` getting matched to an incompatible input, or the error string will just wind up a bit mangled. In particular, this is a run-time issue, and data.table will _not_ fail to compile on old Rtools.

Is that the right understanding?

MichaelChirico:
OK, I think this is good enough for now. What remains is if we use these specifiers outside the R API, which more troubling to deal with (vs. simple `grep` check), and not clear there's any payoff anyway. Let's revisit if there is a user report.

PR #6983, opened 2025-05-15
Add helper macros for the type bump system

MichaelChirico:
Supersedes @badasahog 's #6976 -- I don't have write permissions to that branch. Now that you're a Project Member, you should create branches directly on the Rdatatable/data.table repo to avoid this issue, as well as enable {atime} performance regression testing.

Thanks for the discussion; I made an executive decision on the names. I found `GET_UNBUMPED` and `MAKE_BUMPED` awkward; I think the choices here are the most preferable among what's discussed. I also combed over the file to identify one or two more applications.


MichaelChirico:
Merging since I am not quite the "author", really I am "Approving" the original PR.

PR #6985, opened 2025-05-16
Define RO accessors for RAW, LOGICAL

aitap:
Like other `_RO` accessors, these are also not available on R < 3.5. We make use of them since #6907.


MichaelChirico (approved):
thanks!

Issue #6984, opened 2025-05-16
Review and update GitLab CI failure rules


jangorecki:
I feel that GitLab CI needs a bit of attention.

- We should ensure that url links are being checked in at least one of the jobs, to avoid repeatedly coming back issue with links redirection, that recently caused difficulties in communication with CRAN team during recent package submission.

- It is likely that new NOTEs appeared in R CMD check, and those NOTEs could be escaped from causing failures - if they are known are acceptable.

```
Status: 1 NOTE
...
$ Rscript -e 'l=tail(readLines("data.table.Rcheck/00check.log"), 1L); if (!identical(l, "Status: OK")) stop("Last line of ", shQuote("00check.log"), " is not ", shQuote("Status: OK"), " but ", shQuote(l)) else q("no")'
Error: Last line of '00check.log' is not 'Status: OK' but 'Status: 1 NOTE'
```

```
Status: 3 NOTEs
...
$ (! grep "warning:" data.table.Rcheck/00install.out)
$ Rscript -e 'l=tail(readLines("data.table.Rcheck/00check.log"), 1L); notes<-"Status: 2 NOTEs"; if (!identical(l, notes)) stop("Last line of ", shQuote("00check.log"), " is not ", shQuote(notes), " (size of tarball, non-API calls) but ", shQuote(l)) else q("no")'
Error: Last line of '00check.log' is not 'Status: 2 NOTEs' (size of tarball, non-API calls) but 'Status: 3 NOTEs'
```

- Moreover new WARNINGs appeared in some jobs and they do not causes `Log` status to be `failed`. Possibly they never did, but they should.

```
* checking whether package ‘data.table’ can be installed ... WARNING
Found the following significant warnings:
  fread.c:422:18: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  fread.c:422:49: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  fread.c:427:70: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  fread.c:432:55: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
```

- yet another failure comes from the installation failure, R-devel cran mirror seems to be unavailable

```
[...]

PR #6987, opened 2025-05-16
Fix warnings and truncations around `filesize_to_str()`

aitap:
<ul><li> Seen on 32-bit Windows and Linux:

```r
library(data.table)
fwrite(mtcars, "mtcars")
data.table::fread("mtcars", verbose=TRUE) |> capture.output() |> grepv("bytes", x = _, value = TRUE)
# [1] "  File opened, size = 0.000TB (1281 bytes)."
# [2] "  Number of sampling jump points = 1 because (1280 bytes from row 1 to eof) / (2 * 1280 jump0size) == 0"
# [3] "Read 32 rows x 11 columns from 0.000TB (1281 bytes) file in 00:00.000 wall clock time"
```

Performing a 40-bit shift on a 32-bit `size_t` is undefined behaviour; things go downhill from there. Accept `uint64_t` instead of `size_t` for a file size, since it's still possible for a file to exceed the address space in size.
</li><li>
This will also fix the warnings noted in #6984, where a <code>size_t</code> was passed to the <code>PRIu64</code> size specifier:

```
  fread.c:422:18: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  fread.c:422:49: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  fread.c:427:70: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  fread.c:432:55: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
```
This warning ends up being harmless even on 32-bit systems, but it's still undefined behaviour.
</li><li>
Make use of this on both Windows and non-Windows. Instead of blindly reading the file size modulo 4 GiB, check that the file at least has a chance to fit into the address space. On Linux with <code>glibc</code>, we default to 32-bit <code>off_t</code> and thus fail to open files that are too large, but guard against this anyway in case some other POSIX system behaves differently.
</li><li>
Since we already make use of powers of two in the size printouts, print the <a href="https://en.wikipedia.org/wiki/Binary_prefix">binary SI prefixes</a>.
</li></ul>

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -1425,9 +1425,9 @@ int freadMain(freadMainArgs _args) {
         close(fd);                              // # nocov
         STOP(_("File is too large [%s > %s]: %s"), filesize_to_str(stat_buf.st_size), filesize_to_str(SIZE_MAX), fnam); // # nocov
       }
-      if (verbose) DTPRINT(_("  File opened, size = %s.\n"), filesize_to_str(stat_buf.st_size));
       fileSize = (size_t) stat_buf.st_size;
       if (fileSize == 0) {close(fd); STOP(_("File is empty: %s"), fnam);}
+      if (verbose) DTPRINT(_("  File opened, size = %s.\n"), filesize_to_str(fileSize));
 
       // No MAP_POPULATE for faster nrows=10 and to make possible earlier progress bar in row count stage
       // Mac doesn't appear to support MAP_POPULATE anyway (failed on CRAN when I tried).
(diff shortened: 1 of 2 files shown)


jangorecki:
Before merging this we could possibly add check to CI which detects compilation warnings also on those jobs, now we have one for Linux.

MichaelChirico:
I think this warrants a NEWS entry since it could be user-facing

src/fread.c
```diff
@@ -426,12 +426,12 @@ static const char* filesize_to_str(const size_t fsize)
     }
     if (ndigits == 0 || (fsize == (fsize >> shift << shift))) {
       if (i < sizeof(suffixes)) {
-        snprintf(output, sizeof(output), "%"PRIu64"%cB (%"PRIu64" bytes)", // # notranslate
-                 (fsize >> shift), suffixes[i], fsize);
+        snprintf(output, sizeof(output), "%"PRIu64"%ciB (%"PRIu64" bytes)", // # notranslate
```
MichaelChirico, line 429: tempted to say we should "just" put the full prefix in `suffixes`, but it means moving from simple one-byte-each `char` array to 3-byte version, probably much more trouble than it's worth.

src/fread.c
```diff
@@ -1415,15 +1415,19 @@ int freadMain(freadMainArgs _args) {
...
       struct stat stat_buf;
       if (fstat(fd, &stat_buf) == -1) {
         close(fd);                                                     // # nocov
         STOP(_("Opened file ok but couldn't obtain its size: %s"), fnam); // # nocov
       }
+      if (verbose) DTPRINT(_("  File opened, size = %s.\n"), filesize_to_str(stat_buf.st_size));
+      if (stat_buf.st_size > SIZE_MAX) {
+        close(fd);                              // # nocov
+        STOP(_("File is too large: %s"), fnam); // # nocov
+      }
```
MichaelChirico, line 1428: maybe better

```suggestion
      if (stat_buf.st_size > SIZE_MAX) {
        close(fd);                              // # nocov
        STOP(_("File is too large [%s > %s]: %s"), filesize_to_str(stat_buf.st_size), filesize_to_str(SIZE_MAX), fnam); // # nocov
      }
      if (verbose) DTPRINT(_("  File opened, size = %s.\n"), filesize_to_str(stat_buf.st_size));
```

MichaelChirico:
> Before merging this we could possibly add check to CI which detects compilation warnings also on Mac, now we have one for Linux.

I want to avoid piling in unrelated edits to PRs, it makes the git history a lot harder to understand. Just this week we had to try and parse out why a certain change was made in an unrelated 7+-year-old PR & guess based on PR comments:

https://github.com/Rdatatable/data.table/pull/6982#discussion_r2089351366

Since I don't think it's a blocking issue, I filed #6990.

src/fread.c
MichaelChirico: Let's leave it be for now, not worth it unless there's a user report IMO.

PR #6988, opened 2025-05-16
Compress tarball to <5000000 bytes

aitap:
This yields a 4.56-megabyte tarball, getting rid of one NOTE.

Binary diffs are scary, especially for `*.RData` files. To reproduce:

```
bzip2 -9 vignettes/flights14.csv
tools::resaveRdaFiles('vignettes', compression_level = 9, version = 2)
sed -i 's|\.\./flights14\.csv|../flights14.csv.bz2|' vignettes/*/*.Rmd
sed -i 's|"flights14\.csv|"flights14.csv.bz2|' vignettes/*.Rmd
```

One possible downside: this makes the vignettes dependent on a package that we `Suggest:`, `R.utils`. I don't think this will automatically fail the [noSuggests](https://www.stats.ox.ac.uk/pub/bdr/noSuggests/README.txt) special check for us, but it'll be one more thing to keep in mind.


vignettes/datatable-intro.Rmd
```diff
@@ -52,8 +52,8 @@ options(width = 100L)
 ```
 
 ```{r}
-input <- if (file.exists("flights14.csv")) {
-   "flights14.csv"
+input <- if (file.exists("flights14.csv.bz2")) {
```
jangorecki, line 55: I would rather prefer to keep intro vignette simple and very basic, without need for suggested deps. Extra note is not really a problem as long as we handle it well in CI.

PR #6989, opened 2025-05-16
Mark base signal mechanisms as undesirable

MichaelChirico:
Triggered by https://github.com/jangorecki/data.table/pull/9; but there's been a few other times where `stop()` & friends may slip through. Better to just start linting.


MichaelChirico:
Merging: no already passing

PR #6993, opened 2025-05-17
Header cleanup

aitap:
- [x] Drop the compatibility definition of `SET_GROWABLE_BIT` for R < 3.4.0.
 - [x] Drop the definitions of `R_(Calloc|Realloc|Free)`. The macros cause a lot of re-definition warnings on R < 3.5.0 because the definitions already exist, at least as of 3.4.0.
 - [x] Drop the SVN revision checks we needed during the 4.5.0 release cycle.
 - [x] `Rinternals.h` used to omit declarations for remapped functions implemented as macros until [`r80644`](https://github.com/r-devel/r-svn/commit/b38c8bc04a967cb886715b0d413acdd7fc8aec96) (4.2.0) when `USE_RINTERNALS` was defined:
      ```sh
      svn cat src/include/Rinternals.h -r80643 | cpp -DUSE_RINTERNALS -I src/include/ | grep isNull
      # empty!
      svn cat src/include/Rinternals.h -r80644 | cpp -DUSE_RINTERNALS -I src/include/ | grep isNull
      # Rboolean (Rf_isNull)(SEXP s);
      ```
   We define it for R < 3.5.0 (#3301):
   https://github.com/Rdatatable/data.table/blob/c4ea09e099b1833fae0eacb50552bce6f051ed82/src/data.table.h#L5-L7
   and get this implicit declaration warning:
   ```
   assign.c: In function 'setdt_nrows':
   assign.c:218:5: warning: implicit declaration of function 'Rf_isNull' [-Wimplicit-function-declaration]
        if (Rf_isNull(xi)) continue;
        ^
   ```
   This code path isn't actually broken; the implicit prototype happens to be compatible enough. The recommended practice nowadays is, indeed, to use the full `Rf_...` names of the API entry points instead of relying on them being remapped. We could ignore the warning, use `isNull` instead of `Rf_isNull` until we depend on R ≥ 3.5, or drop `USE_RINTERNALS` altogether.

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -215,7 +215,7 @@ SEXP setdt_nrows(SEXP x)
      *   many operations still work in the presence of NULL columns and it might be convenient
      *   e.g. in package eplusr which calls setDT on a list when parsing JSON. Operations which
      *   fail for NULL columns will give helpful error at that point, #3480 and #3471 */
-    if (Rf_isNull(xi)) continue;
+    if (isNull(xi)) continue;
     if (Rf_inherits(xi, "POSIXlt")) {
       error(_("Column %d has class 'POSIXlt'. Please convert it to POSIXct (using as.POSIXct) and run setDT() again. We do not recommend the use of POSIXlt at all because it uses 40 bytes to store one date."), i+1);
     }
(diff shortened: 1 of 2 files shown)


tdhock:
lgtm thanks!

Issue #6980, opened 2025-05-14
`_const_cast` is de facto undefined behavior


badasahog:
I am aware that the usage of sidestepping constness is used to invoke copy on write behavior.

This is still undefined behavior.

I'm going to investigate to see if I can find a decent solution to this problem, and I also want to hear back if anyone disagrees with my assessment.

badasahog:
Why even const qualify the pointer, then? Is there any genuine performance benefit?

MichaelChirico:
For context it was added in #2153; there's not much exposition there, but ostensibly they were trying to quiet some compiler warnings.

It's used twice:

https://github.com/Rdatatable/data.table/blob/9c54bd073decac61dbd284c7a22d65d0f719aa19/src/fread.c#L1573

https://github.com/Rdatatable/data.table/blob/9c54bd073decac61dbd284c7a22d65d0f719aa19/src/fread.c#L1865-L1866

Agree about proceeding cautiously if there's a big performance impact (particularly looking at the `args.filename` branch).

badasahog:
we should at least rename it to just `const_cast`, because starting a global identifier with an underscore is also undefined behavior

badasahog:
quick update!

we may be able to sidestep this issue by keeping a non const pointer to use for the '\0' write.

https://stackoverflow.com/questions/79650846/does-const-char-alias-with-char

I still think that removing the cow page is a good idea. It's cleaner and will result in performance gain.

C11's safe functions like `strnlen_s` would be ideal for this, but we're compiling with c99.

If maintainers find this to be a significant enough incentive to upgrade, I'm ready to update the code accordingly (I know that's a serious decision to be made, not something that can happen all of a sudden).

PR #7001, opened 2025-05-21
renamed `_const_cast` to `const_cast`

badasahog:
starting a global identifier with an underscore is undefined behavior.

related to #6980


badasahog:
@TimTaylor you might be right, I'm not 100%. Either way, there's no need to tow the line.

This code segment is also in the process of being removed anyway.

badasahog:
It's not completely pointless, it's a stopgap. I don't know how long it will take to get a proper solution merged. I'll add `__halt`.

badasahog:
I'll get on that

badasahog:
replaced by #7011

PR #7005, opened 2025-05-21
copied commit from local branch

badasahog:


badasahog:
@aitap I moved this to a local branch as requested. Can we get it merged?

tdhock:
this may have performance implications.
what kind of data would be able to show any performance differences of this pr?

badasahog:
@tkhock the performance gain would be tiny because it doesn't happen in a loop.

I still insist this is best practice, and severely minimizes the potential for bugs in the future while still being much simpler. Fewer lines.

tdhock:
removing static means new allocations for every function call, and maybe a performance hit (not benefit).
sometimes in data.table we want to do things which in other contexts may be considered "unsafe" and that is for a performance benefit.
the existing fread performance tests do not show any slow downs.

also it is normal to wait some time before code review.

MichaelChirico:
I would prefer Ivan to weigh in here, I would need to do a lot of background reading to review competently.

We have no review time SLO. For a volunteer project, a review period of ~1 week is pretty acceptable IMO. We do not meet this in general at the moment, something we are working to improve.

We are also a review team of primarily R authors -- changes like this touch on deeper aspects of C coding that are outside my (I won't speak directly for others here) comfort zone. They are very welcome improvements, but also take more time to review. Thank you for your contributions & for your patience.

badasahog:
This codebase already relies on features that aren't in c89
-fixed size integers
-restrict qualified pointers
-inline functions
-designated initializers
-data/code is desegregated
-integer division
-//comments

badasahog:
I don't have long standing or specific safety/style guidelines for the code.

What I will say is that I'm not a fan of the `x=y;` style, gives me no room to breathe. It's also not super consistent, so I would suggest a final decision be made to unify the style. I prefer `x = y;`.

tdhock:
the only data.table style guide I know of is https://github.com/Rdatatable/data.table/blob/master/.github/CONTRIBUTING.md#coding-style

MichaelChirico:
This is producing new compiler warnings, e.g.:

```
freadR.h:27:51: note: in definition of macro ‘DTWARN’
   27 | #define DTWARN(...) warningsAreErrors ? halt__(1, __VA_ARGS__) : warning(__VA_ARGS__)
      |                                                   ^~~~~~~~~~~
fread.c:2766:71: warning: ISO C forbids empty initializer braces before C23 [-Wpedantic]
 2766 |           DTi + row1line, ncol, tt, strlim(skippedFooter, (char[500]) {}, 500));
      |                                                                       ^
```

Is there any workaround?

Issue #6919, opened 2025-04-22
Excessive `goto` usage.


badasahog:
the `fread.c` file uses goto 43 times, making it hard to read and reason about, likely slowing down development.

I'm happy to work on this if other contributors agree it would be a step in the right direction.

I'm new to this project, and I don't want to step on any toes

MichaelChirico:
Very open to clean-up, streamlining, and modernization of the code.

Some of the code is highly optimized, so do be aware of benchmarking your changes. The blast footprint of changes can be quite large as well, so please be ready for breaking changes to even small refactors.

badasahog:
I forgot to close this.

PR #7009, opened 2025-05-22
removed 2 more goto statements

badasahog:
part of an ongoing effort to improve control flow clarity #6919

This is not the end. Even though the goto statements have been removed, there's still a lot of ground to cover.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -2290,439 +2290,446 @@ int freadMain(freadMainArgs _args) {
   nth = imin(nJumps, nth);
 
   if (verbose) DTPRINT(_("[11] Read the data\n"));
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Thanks for the change. The diff size is quite large, but it _looks_ like that's mostly about whitespace changes:

https://github.com/Rdatatable/data.table/pull/7009/files?w=1

It would help if you could quickly describe what was done. It looks like a `read` --> `goto read` "loop" was replaced by a `while (true)` repeat loop with some guarantees it will `break`. Is there anything else noteworthy in how this change was done?

badasahog:
@MichaelChirico not really. That's about it.

PR #7010, opened 2025-05-22
Removes unnecessary `data.table` call from as.data.table.array

eliocamp:
I noticed that as.data.table() was quite slow and after a quick look at the code I noticed that there's a line that calls `data.table()` on the output of `CJ()`, which seems to be adding some extra overhead. Removing that call and adding that extra column by reference gives some small improvement

``` r
dims <- c(50, 50, 300, 10)
arr <- array(1:prod(dims), dim = dims)

bench::mark(as.data.table.array(arr),
            old_as.data.table.array(arr)
          )
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression                        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 as.data.table.array(arr)        1.29s    1.29s     0.773     503MB     2.32
#> 2 old_as.data.table.array(arr)    1.87s    1.87s     0.536     610MB     2.14
```

It’s still much slower than this simpler implementation, but maybe it’s an unfair comparison because as.data.table.array does more stuff. There is a bunch of reordering in `as.data.table.array()`, for example. 

``` r
.melt_array <- function(array, dims, value.name = "V1", sorted = TRUE) {

  dims <- c(dims[length(dims):1], sorted = sorted)
  grid <- do.call(data.table::CJ, dims)
  grid[, c(value.name) := c(array)][]

  return(grid)
}
          

bench::mark(as.data.table.array(arr, na.rm = FALSE),
            .melt_array(arr, lapply(dims, seq_len), value.name = "N"),
          check = FALSE)
#> # A tibble: 2 × 6
#>   expression                            min  median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 "as.data.table.array(arr, na.rm = … 991ms 990.9ms      1.01     207MB      0  
#> 2 ".melt_array(arr, lapply(dims, seq…  55ms  56.8ms     17.6      143MB     23.5
```

<sup>Created on 2025-05-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,8 @@
 
 4. `as.Date()` method for `IDate` no longer coerces to `double` [#6922](https://github.com/Rdatatable/data.table/issues/6922). Thanks @MichaelChirico for the report and PR. The only effect should be on overly-strict tests that assert `Date` objects have `double` storage, which is not in general true, especially from R 4.5.0.
 
+5. `as.data.table()` is slightly faster at converting arrays to data.tables. Thanks @eliocamp. 
+
 ### BUG FIXES
 
 1. Custom binary operators from the `lubridate` package now work with objects of class `IDate` as with a `Date` subclass, [#6839](https://github.com/Rdatatable/data.table/issues/6839). Thanks @emallickhossain for the report and @aitap for the fix.
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -251,19 +251,17 @@ test.list <- atime::atime_test_list(
     Before = "f339aa64c426a9cd7cf2fcb13d91fc4ed353cd31", # Parent of the first commit https://github.com/Rdatatable/data.table/commit/fcc10d73a20837d0f1ad3278ee9168473afa5ff1 in the PR https://github.com/Rdatatable/data.table/pull/6393/commits with major change to fwrite with gzip.
     PR = "3630413ae493a5a61b06c50e80d166924d2ef89a"), # Close-to-last merge commit in the PR.
 
-  tests=extra.test.list)
-
-
-  # Test case created directly using the atime code below (not adapted from any other benchmark), based on the issue/fix PR https://github.com/Rdatatable/data.table/pull/5054#issue-930603663 "melt should be more efficient when there are missing input columns."
+  # Test case created directly using the atime code below (not adapted from any other benchmark), based on the PR  Removes unnecessary data.table call from as.data.table.array https://github.com/Rdatatable/data.table/pull/7010 
   "as.data.table.array improved in #7010" = atime::atime_test(
     setup = {
       dims = c(50, 50, 300, 10)
       arr = array(1:prod(dims), dim = dims)
     },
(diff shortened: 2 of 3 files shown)


ben-schwen:
Good find, since output of `CJ` is ensured to be a `data.table`. What about the recycling capablity of `data.table()`.

In any case, we are missing tests and a NEWS item!

MichaelChirico:
great find, want to take a stab at adding an atime performance regression test as well?

eliocamp:
> great find, want to take a stab at adding an atime performance regression test as well?

Sure. Is there documentation on how to do it? I read through the contributing document and didn't find any.

MichaelChirico:
Check https://github.com/Rdatatable/data.table/wiki/Performance-testing; in most cases you can grok what to do from reading a few examples in the existing suite: https://github.com/Rdatatable/data.table/blob/master/.ci/atime/tests.R

if you think more documentation would help, don't hesitate! it's still in pretty early stages.

eliocamp:
The performance-testing documentation is good and clear. Maybe it would need to be linked to from the CONTRIBUTING.md file. 

It seems that the atime test is failing because the "fast" commit points to a different repo? Seems that the "fast" sha needs to point to the commit in which this PR is merged? Is there a way of adding the test before this PR is merged?

NEWS.md
```diff
@@ -14,6 +14,8 @@
 
 4. `as.Date()` method for `IDate` no longer coerces to `double` [#6922](https://github.com/Rdatatable/data.table/issues/6922). Thanks @MichaelChirico for the report and PR. The only effect should be on overly-strict tests that assert `Date` objects have `double` storage, which is not in general true, especially from R 4.5.0.
 
+5. `as.data.table()` is slightly faster at converting arrays to data.tables. Thanks @eliocamp.
```
jangorecki, line 17: Link to gh issue or PR would be nice here

tdhock:
Hi Elio I was just at Rencontres R meeting last week, there were people who mentioned your translation work in their talks (Maelle Salmon and Christophe Regouby).

> It seems that the atime test is failing because the "fast" commit points to a different repo? Seems that the "fast" sha needs to point to the commit in which this PR is merged? Is there a way of adding the test before this PR is merged?

Right, our performance testing system only works with commits in this repo. So 

* you are a project member https://github.com/orgs/Rdatatable/teams/project-members?query=elio so you do have permission to create new branches in this repo, so please do that instead of using your fork next time.
* for this case we can just merge this PR without the performance test, and then add it in another PR later (with references to the commit from this PR as Fast, and the previous commit as Slow)

MichaelChirico:
I went ahead and re-pushed the changes directly to a branch on this repo. Let's continue any discussion there.

PR #7016, opened 2025-05-25
link performance testing wiki page from CONTRIBUTING.md

tdhock:
Hi @eliocamp in you https://github.com/Rdatatable/data.table/pull/7010#issuecomment-2907530647 you suggested adding a link from CONTRIBUTING.md to the performance testing wiki page docs.
Does this PR add the link in a place where you would find it useful?


MichaelChirico (approved):
Thanks! I moved it after the description of `test()`, since atime testing is less common.

PR #7019, opened 2025-05-26
Removes unnecessary `data.table` call from as.data.table.array

MichaelChirico:
Re-push of the changes in #7010 (authored by @eliocamp) onto `Rdatatable/data.table` so that we get {atime} results directly here instead of waiting for follow-up.


tdhock:
on CI we do see a small improvement in speed
image

PR #6925, opened 2025-04-25
Updating fread's file.info argument to avoid uname and udomain lookup.

ProfFancyPants:
Addresses issue: https://github.com/Rdatatable/data.table/issues/6918

PR #7022, opened 2025-05-26
Faster commit for fread test

tdhock:
@MichaelChirico suggested adding a performance test for the improvement in https://github.com/Rdatatable/data.table/pull/6925#issuecomment-2908879750
the test cases already existed; I added a Faster commit.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -13,7 +13,7 @@ for (extra.arg in extra.args.6107){
       tmp_csv = tempfile()
       fwrite(DT, tmp_csv)
     },
-    Faster = "60a01fa65191c44d7997de1843e9a1dfe5be9f72", # First commit of the PR (https://github.com/Rdatatable/data.table/pull/6925/commits) that reduced time usage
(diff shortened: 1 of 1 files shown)


tdhock:
I don't see any difference on CI 
image

I do see is locally though (see below) so I guess it is highly system-dependent (disk response time).
image

.ci/atime/tests.R
```diff
@@ -13,6 +13,7 @@ for (extra.arg in extra.args.6107){
       tmp_csv = tempfile()
       fwrite(DT, tmp_csv)
```
MichaelChirico, line 14: this test is only of `fwrite()`, where the claimed improvement is for `fread()`, am I missing something?

tdhock: yes you missed something.
fwrite is used in setup, but setup is not timed.
the timing happens on expr which uses fread below.

MichaelChirico: :facepalm: thanks :)

tdhock:
This is the result I get on my system, we can see a constant factor difference.
image

tdhock:
still difficult to see on CI
image

MichaelChirico (approved):
Great, this looks more faithful as a test -- scaling the number of `fread()` calls --> scale the number of `file.info()` calls. Even if this doesn't catch the specific improvement we were after, having something like this will still be a useful performance regression test.

Issue #5564, opened 2022-12-15
i column not found in rolling join with by


tdhock:
hi all! I'm not an expert in rolling joins, but I thought it should be possible to use arguments `by` and `roll` at the same time to do rolling join and summarization efficiently in one step. But I observed a "not found" error when I tried that. 
Here is a complete example:
```r
library(data.table)
N = 4
X.vec = c(0.05, 0.15, 0.75, 0.95)
Y.vec = 10*X.vec
(irregular = data.table(X=X.vec, Y=Y.vec, key="X"))
grid.space = 0.2
times = 1/grid.space
offset = grid.space/2
grid = seq(offset, 1-offset, by=grid.space)
(regular <- data.table(grid, X=grid, min=grid-offset, max=grid+offset, key="X"))
(joined = regular[irregular, roll="nearest"])
joined[, .(N=.N, mean.Y=mean(Y)), by=grid]
irregular[regular, .(grid, N=.N, mean.Y=mean(Y)), on=.(X<max, X>min), by=.EACHI, nomatch=0L]
regular[irregular, .(N=.N), roll="nearest", by=grid]
regular[irregular, .(N=.N, mean.Y=mean(Y)), roll="nearest", by=grid]
```
The goal is a rolling join between `regular` and `irregular`, with matching each `irregular$X` to the closest `regular$grid` value, then computing the mean of `irregular$Y` for each value of `regular$grid`.  
```r
> (irregular = data.table(X=X.vec, Y=Y.vec, key="X"))
      X   Y
1: 0.05 0.5
2: 0.15 1.5
3: 0.75 7.5
4: 0.95 9.5
> (regular <- data.table(grid, X=grid, min=grid-offset, max=grid+offset, key="X"))
   grid   X min max
1:  0.1 0.1 0.0 0.2
2:  0.3 0.3 0.2 0.4
3:  0.5 0.5 0.4 0.6
4:  0.7 0.7 0.6 0.8
5:  0.9 0.9 0.8 1.0
```
One way to do that is by first materializing the `joined` table, then summarizing it:
```r
> (joined = regular[irregular, roll="nearest"])
   grid    X min max   Y
1:  0.1 0.05 0.0 0.2 0.5
2:  0.1 0.15 0.0 0.2 1.5
3:  0.7 0.75 0.6 0.8 7.5
4:  0.9 0.95 0.8 1.0 9.5
> joined[, .(N=.N, mean.Y=mean(Y)), by=grid]
[...]

PR #7025, opened 2025-05-26
Allow tbl in i to use 'i.' and 'x.' prefixes in joins

### LINKED ISSUE #6998: DT[TBL, a:=i.b] update join with 'i.' prefix fails when 'i' is a tibble
```r
library(data.table)
library(tibble)

DT1 = data.table(a=1, b=2)
DT2 = data.table(a=1, c=3)
DF = data.frame(a=1, d=4)
TBL = tibble(a=1, e=5)

# fine
DT1[DT2, on='a', c := i.c]

# also fine
DT1[DF, on='a', d := i.d]

# error
DT1[TBL, on='a', e := i.e]
# Error in eval(jsub, SDenv, parent.frame()) : object 'i.e' not found

DT1
#        a     b     c     d
#    <num> <num> <num> <num>
# 1:     1     2     3     4
```

This is inconsistent -- either we should require `i` to be a data.table (so `DT1[DF, ...]` would fail), or we should treat all `is.data.frame(i) && !is.data.table(i)` cases the same.

MichaelChirico:
Closes #6998


tdhock:
cool!
I noticed a similar issue, https://github.com/Rdatatable/data.table/issues/5564 which I thought may be fixed by this PR, but I just tested it, and it seems to still not work.

tdhock:
THanks, I approve but I don't see the approve review button for some reason.

PR #7027, opened 2025-05-27
improved consistency and clarity around allocation functions

badasahog:
This is a follow up to a previous pr that was merged. I just missed some stuff which I missed the first time around. I also updated the linter.

To recap:

We agreed on the syntax `a = malloc(sizeof(*a) * count);` for allocation functions.
No casting the result, as that could hide bugs and hinder readability.

Unfortunately, the R allocation function (like `R_Alloc`) must be explicitly casted because they return `char*` instead of `void*`, so implicit casting would cause a warning.


tdhock:
It is no problem to have lots of little branches, I think that is one of the main advantages of git.

badasahog:
understood. I will migrate to a local branch

PR #7030, opened 2025-05-27
basic preprocessor formatting improvements

badasahog:
Improvements for translation phases 1-3

1. misaligned or unnecessary trailing backslashes fixed
2. include statement whitespace realigned
3. source files are only required to have a single new line at the end, not two


src/assign.c
```diff
@@ -1335,4 +1335,3 @@ SEXP setcharvec(SEXP x, SEXP which, SEXP newx)
   }
   return R_NilValue;
 }
-
```
MichaelChirico, line 1338: Weirdly the web UI is hiding the duplicate newline:

image

Confirmed it's there on the command line:

```shell
tail -c 3 src/assign.c | cat -A
# }$
# $
```

MichaelChirico (approved):
Thanks! I don't think it's likely to cause any troublesome conflicts.

PR #7032, opened 2025-05-28
Respect integer64= selection when out-of-sample type bumps happen in fread()

### LINKED ISSUE #2749: in fread integer64 argument is ignored if int64 type-bump occurs out-of-sample
```
> dt0 = data.table(a=seq(10000), b="100")
> dt0[111, b := "1000000000000"]
> f = tempfile()
> fwrite(dt0, f)

> fread(f, integer64="numeric")
           a     b
       <int> <i64>
    1:     1   100
    2:     2   100
    3:     3   100
    4:     4   100
   ---            
 9996:  9996   100
 9997:  9997   100
 9998:  9998   100
 9999:  9999   100
10000: 10000   100

> fread(f, integer64="character")
           a     b
       <int> <i64>
    1:     1   100
    2:     2   100
    3:     3   100
    4:     4   100
   ---            
 9996:  9996   100
 9997:  9997   100
 9998:  9998   100
 9999:  9999   100
10000: 10000   100
```

badasahog:
Closes #2749 

I also changed `while(true)` to `for(;;)` for consistency. Technically should've been its own pr, hope you don't mind.

### DIFF
--- a/src/freadR.c
+++ b/src/freadR.c
@@ -28,7 +28,7 @@ Secondary separator for list() columns, such as columns 11 and 12 in BED (no nee
 static int  typeSxp[NUT] =     {NILSXP,  LGLSXP,    LGLSXP,     LGLSXP,     LGLSXP,     LGLSXP,     LGLSXP,     INTSXP,    REALSXP,     REALSXP,    REALSXP,        REALSXP,        INTSXP,          REALSXP,         STRSXP,      REALSXP,    STRSXP   };
 static char typeRName[NUT][10]={"NULL",  "logical", "logical",  "logical",  "logical",  "logical",  "logical",  "integer", "integer64", "double",   "double",       "double",       "IDate",         "POSIXct",       "character", "numeric",  "CLASS"  };
 static int  typeEnum[NUT] =    {CT_DROP, CT_EMPTY,  CT_BOOL8_N, CT_BOOL8_U, CT_BOOL8_T, CT_BOOL8_L, CT_BOOL8_Y, CT_INT32,  CT_INT64,    CT_FLOAT64, CT_FLOAT64_HEX, CT_FLOAT64_EXT, CT_ISO8601_DATE, CT_ISO8601_TIME, CT_STRING,   CT_FLOAT64, CT_STRING};
-colType readInt64As=CT_INT64;
+static colType readInt64As=CT_INT64;
 static SEXP selectSxp;
 static SEXP dropSxp;
 static SEXP colClassesSxp;


MichaelChirico (commented):
Please add some test(s) in inst/tests/tests.Rraw, and be sure to write a NEWS entry.

@oleksiyskononenko are you able to confirm if pydatatable is still maintained / whether this change to `readInt64As` affects that implementation?

badasahog:
I'm sorry, I don't know how to add tests, I'm not familiar with the testing system.

oleksiyskononenko:
@MichaelChirico I'm not involved in `datatable` development anymore, so can't confirm. However, it doesn't share any code with `data.table` for a long time, so your changes to `fread()` wouldn't affect anything on the python side.

badasahog:
Sorry for the late response. I don't know how to make tests, I'm not proficient in R, and I don't know how the testing system works.

MichaelChirico:
Writing tests is a required part of OSS contribution. By not writing tests, you're burdening other contributors/the maintainers with doing so on your behalf.

I pointed above to the documentation to `test()`. The README for `cc()` might help more -- it provides some simple steps for iterating through tests locally:

https://github.com/Rdatatable/data.table/tree/master/.dev

You should also feel free to just use the GitHub CI suite -- make changes to tests, push to GitHub, check for GHA failures, iterate.

If you're stuck somewhere, we are happy to help, but simply opting out of writing tests won't work.

src/fread.c
```diff
@@ -26,6 +26,8 @@
 #include <stdbool.h>
 #include "freadLookups.h"
 
+extern colType readInt64As;
```
aitap: Let's make it part of `FREAD_MAIN_ARGS_EXTRA_FIELDS` in `freadR.h` and pass through the `args` parameter instead of declaring a variable in the non-header file.

inst/tests/tests.Rraw
```diff
@@ -3070,6 +3070,18 @@ if (test_bit64) test(1017.1, fread(f), copy(DT)[,A:=as.integer64(A)])
...
+  dt0 = data.table(a=seq(10000), b="100")
+  dt0[111, b := "1000000000000"]
+  f = tempfile()
+  fwrite(dt0, f)
+
+  test(1017.3, fread(f, integer64="numeric"), fread(f, colClasses=c("integer", "numeric")))
+  test(1017.4, fread(f, integer64="character"), fread(f, colClasses=c("integer", "character")))
+
+  unlink(f)
+}
```
aitap: Good test, thank you! It fails for me on `master` and passes on the PR branch. One thing: since R doesn't have variable declarations, the `{` braces `}` don't create a new scope; any variables created inside them are fully visible in the outer environment. (The only thing they do is combine multiple expressions, returning the value of the last expression in the group.) The code might as well not create an indentation level at all.

MichaelChirico: If you _do_ want to enforce a new scope, you can use `local({ ... })`. Though we don't really do that in our tests (we should, but that's another issue)

PR #7034, opened 2025-05-29
Upgrade atime action to 1.4.3

tdhock:
follow-up to #7033 which did not work as expected (no bullet points) -- I had tested the main branch, not 1.4.2 which was supposed to be the most recent version, but in fact was an older tag https://github.com/Anirban166/Autocomment-atime-results/pull/43#issuecomment-2918372627

Also reduces p-value threshold from 0.01 (atime default) to 0.001, to reduce likelihood of false positive messages about HEAD being slow.


tdhock:
@Anirban166 there seems to be some issue downloading packages, /usr/bin/sudo DEBIAN_FRONTEND=noninteractive apt-get install -y gdebi-core qpdf devscripts ghostscript errors with server timeout. I think we can just wait until the servers go back up (nothing for us to change).

PR #7037, opened 2025-05-29
removed const qualifier

badasahog:
I removed the const qualifier, as well as the `const_cast` function.

Since it was a read-mostly object, the performance hit should be minimal, while removing all undefined behavior.

We can also experiment with restrict qualified pointers, but I want to get this merged first to avoid a longer approval process.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -1,3 +1,5 @@
+pval.thresh <- 0.001 # to reduce false positives.
+
 # Test case adapted from https://github.com/Rdatatable/data.table/issues/6105#issue-2268691745 which is where the issue was reported.
 # https://github.com/Rdatatable/data.table/pull/6107 fixed performance across 3 ways to specify a column as Date, and we test each individually.
 extra.args.6107 <- c(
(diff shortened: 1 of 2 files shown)


badasahog:
I don't know what's wrong with the benchmark system, I had the same problem on my other PRs

MichaelChirico:
> the performance hit should be minimal

It would be good to quantify this, exactly. Are we talking % difference, fraction-of-%? Let's see if the atime suite tells us anything; would there be an easy way to measure it more directly, too?

I'm particularly looking at the first case in the `args.filename` branch, which is very common.

badasahog:
It's like christmas lights. If I untangle these warnings the entire thing falls apart.

Why do we even need the cow page? I'm not sure I understand its purpose.

badasahog:
@aitap I can't find that. can you point me to it?

If so, that could be a silver bullet.

aitap:
It's not just one place, the code has been assuming this for 8 years. Consider `eol()`: if there's not a `\0` at the end and the buffer consists of ASCII carriage returns, it'll be overrun. Consider the callers of `eol`: sometimes they check `ch == eof` _after_ dereferencing `*ch`. It'll take a very careful reading of the 2064 lines of code to untangle.

PR #7029, opened 2025-05-27
improved consistency and clarity around allocation functions

badasahog:
This is a follow up to a previous pr that was merged. I just missed some stuff the first time around. I also updated the linter.

To recap:

We agreed on the syntax `a = malloc(sizeof(*a) * count);` for allocation functions.
No casting the result, as that could hide bugs and hinder readability.

Unfortunately, the R allocation function (like `R_Alloc`) must be explicitly casted because they return `char*` instead of `void*`, so implicit casting would cause a warning.

PR #7039, opened 2025-05-29
Fix Coccinelle allocation cast linter

aitap:
The allocation cast linter didn't parse properly after #7029. Got it back in working order at the cost of feeding it preprocessed code; now it's not necessarily possible to just apply what it prints as a patch in order to fix a problem.

What would you say about extracting the C-linter code into a script and making print all errors even after the first `stop()` (in a separate pull request)? This should make it easier to test changes to linters before committing them.


MichaelChirico:
> What would you say about extracting the C-linter code into a script and making print all errors even after the first `stop()` (in a separate pull request)? This should make it easier to test changes to linters before committing them.

Yes, absolutely, it's consistent with how other linters work & much easier to work with generally. Something I always wind up kicking the can on when I notice it...

Issue #6995, opened 2025-05-18
awful one liners don't improve readability


badasahog:
This is a pattern I've seen in various places in the code. Collapsing everything into one line doesn't make the code run faster, the compiler is smart enough to sort it out.

examples:
https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L1257
https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L1801
https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L2497

I think it's especially important to separate operations like the increment (`++`) operator away from control flow statements. Again, this doesn't cost any performance.

MichaelChirico:
Please refrain from describing long-standing and fully-functional code as 'awful' :)

Also, note that you used line references on `master`, now they are incorrect. If you use a permalink, they will continue to match your original reference, and GitHub will even be so kind as to preview the code directly in the issue. I am taking a guess as to what you were originally citing in my edit.

MichaelChirico:
Of all the difficult-to-read code in the codebase, I don't particularly find these to stand out. Personally I find it "nice" to have all the logic in-line.

If anything, I might suggest wrapping those in a helper routine to make it clearer what the loop is doing, but I'm not sure the implementation needs to change. I would happily review a PR with a specific alternative proposal, but consider if it's more impactful to put effort elsewhere.

PR #7045, opened 2025-06-04
implemented const_cast as a macro

badasahog:
I changed `const_cast` to be a type agnostic macro, much more in line with the C++ version.


aitap:
This seems to crash `fread`. We only have one use for casting away the `const`: to fiddle with the memory-mapped file. There is no need for the generic version. If you would like to get rid of `union` type punning, let's launder the non-`const` pointer through `uintptr_t`. If you have time, try to find out which compiler emits warnings when casting a `const char *` to `char *` - what if we don't have to worry about it?

src/fread.c
```diff
@@ -126,13 +126,11 @@ static void Field(FieldParseContext *ctx);
...
- * Drops `const` qualifier from a `const char*` variable, equivalent of
- * `const_cast<char*>` in C++.
+ * Drops `const` qualifier from any scalar object, equivalent of
+ * `const_cast` in C++.
  */
-static char* const_cast(const char *ptr) {
-  union { const char *a; char *b; } tmp = { ptr };
-  return tmp.b;
-}
+#define const_cast(x) (((union{typeof(x) a; typeof(+(*(x)))* b;}){.a=(x)}).b)
```
Anirban166, line 147: In the issue titled 'awful one liners don't improve readability or performance' (https://github.com/Rdatatable/data.table/issues/6995) I see you wrote:

> Collapsing everything into one line doesn't make the code run faster, the compiler is smart enough to sort it out.

 I wonder what category this belongs to then 🤔

badasahog: A fair criticism!

The problem is, I don't think it's even possible to implement it differently since its a macro, and I think const_cast already has a clear enough meaning that the one liner isn't too bad here. It's clear what the code does.

PR #7046, opened 2025-06-04
removed macro `internalErrSize`

badasahog:
In the interest of continued improvements to readability, I've removed the unnecessary `internalErrSize` macro constant.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -2238,7 +2238,7 @@ int freadMain(freadMainArgs _args) {
   double tRead = 0, tReread = 0;
   double thRead = 0, thPush = 0;  // reductions of timings within the parallel region
   int max_col = 0;
-  char *typeBumpMsg = NULL;  size_t typeBumpMsgSize=0;
+  char *typeBumpMsg = NULL;  size_t typeBumpMsgSize = 0;
   int typeCounts[NUMTYPE];  // used for verbose output; needs populating after first read and before reread (if any) -- see later comment
   char internalErr[256] = "";  // must be compile time size: the message is generated and we can't free before STOP
 
--- a/src/fread.c
+++ b/src/fread.c
@@ -2257,8 +2257,7 @@ int freadMain(freadMainArgs _args) {
   int max_col=0;
   char *typeBumpMsg=NULL;  size_t typeBumpMsgSize=0;
   int typeCounts[NUMTYPE];  // used for verbose output; needs populating after first read and before reread (if any) -- see later comment
-  #define internalErrSize 1000
-  char internalErr[internalErrSize+1]="";  // must be compile time size: the message is generated and we can't free before STOP
+  char internalErr[1001]="";  // must be compile time size: the message is generated and we can't free before STOP
   int64_t DTi = 0;                  // the current row number in DT that we are writing to
   const char *headPos = pos;       // the jump start corresponding to DTi
   int nSwept = 0;                  // count the number of dirty jumps that were swept
@@ -2314,7 +2313,7 @@ int freadMain(freadMainArgs _args) {
         nth = omp_get_num_threads();
         if (me!=0) {
           // # nocov start
-          snprintf(internalErr, internalErrSize, "Master thread is not thread 0 but thread %d.\n", me); // # notranslate
+          snprintf(internalErr, sizeof(internalErr) - 1, "Master thread is not thread 0 but thread %d.\n", me); // # notranslate
           stopTeam = true;
           // # nocov end
         }
@@ -2583,7 +2582,7 @@ int freadMain(freadMainArgs _args) {
           }
           else if (headPos!=thisJumpStart && nrowLimit>0) { // do not care for dirty jumps since we do not read data and only want to know types
              // # nocov start
-            snprintf(internalErr, internalErrSize, "invalid head position. jump=%d, headPos=%p, thisJumpStart=%p, sof=%p", jump, headPos, thisJumpStart, sof); // # notranslate
+            snprintf(internalErr, sizeof(internalErr) - 1, "invalid head position. jump=%d, headPos=%p, thisJumpStart=%p, sof=%p", jump, headPos, thisJumpStart, sof); // # notranslate
             stopTeam = true;
             // # nocov end
           }
--- a/src/fread.c
+++ b/src/fread.c
@@ -2257,7 +2257,7 @@ int freadMain(freadMainArgs _args) {
   int max_col=0;
   char *typeBumpMsg=NULL;  size_t typeBumpMsgSize=0;
   int typeCounts[NUMTYPE];  // used for verbose output; needs populating after first read and before reread (if any) -- see later comment
-  char internalErr[1001]="";  // must be compile time size: the message is generated and we can't free before STOP
+  char internalErr[256]="";  // must be compile time size: the message is generated and we can't free before STOP
   int64_t DTi = 0;                  // the current row number in DT that we are writing to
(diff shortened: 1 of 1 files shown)


src/fread.c
```diff
@@ -2314,7 +2313,7 @@ int freadMain(freadMainArgs _args) {
         nth = omp_get_num_threads();
         if (me!=0) {
           // # nocov start
-          snprintf(internalErr, internalErrSize, "Master thread is not thread 0 but thread %d.\n", me); // # notranslate
+          snprintf(internalErr, sizeof(internalErr) - 1, "Master thread is not thread 0 but thread %d.\n", me); // # notranslate
```
aitap: Let's fix an off-by-one mistake while we're at it. `snprintf` accepts the total size of the buffer, including the place for the zero terminator. It's safe to give it the full `sizeof internalErr`.

src/fread.c
```diff
@@ -2257,8 +2257,7 @@ int freadMain(freadMainArgs _args) {
   int max_col=0;
   char *typeBumpMsg=NULL;  size_t typeBumpMsgSize=0;
   int typeCounts[NUMTYPE];  // used for verbose output; needs populating after first read and before reread (if any) -- see later comment
-  #define internalErrSize 1000
-  char internalErr[internalErrSize+1]="";  // must be compile time size: the message is generated and we can't free before STOP
+  char internalErr[1001]="";  // must be compile time size: the message is generated and we can't free before STOP
```
aitap: Can the buffer size be lowered to 256 without truncating any messages?

badasahog: I don't know. That's an improvement I'm not comfortable making, I'd rather get this merged first, then try that later for easier rollback and regression tracking.

badasahog:
I don't know why the tests are failing. Everything works fine on my end

src/fread.c
```diff
@@ -2531,18 +2530,18 @@ int freadMain(freadMainArgs _args) {
...
                       typeName[IGNORE_BUMP(joldType)], typeName[IGNORE_BUMP(thisType)],
                       (int)(tch-fieldStart), fieldStart, (int64_t)(ctx.DTi+myNrow));
-                    if (len > 1000) len = 1000;
-                    if (len > 0) {
-                      typeBumpMsg = realloc(typeBumpMsg, typeBumpMsgSize + len + 1);
-                      strcpy(typeBumpMsg+typeBumpMsgSize, temp);
-                      typeBumpMsgSize += len;
-                    }
+
+                    len = min(len, sizeof(buffer));
```
MichaelChirico: where is min defined? 

> undefined symbol: min

aitap:
In the Microsoft Windows SDK, when compiling C code, <stdlib.h>
#defines the macros min() and max() in addition to the documented
Microsoft-specific macros __min() and __max() [1], unless __STDC__ is
pre-defined to a non-zero value (it's not) or _POSIX_ is defined
(neither it is). In MinGW, the SDK used by R on Windows as part of
Rtools [2], these definitions are disabled [3]; they are absent on
other systems.

[1]
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/min?view=msvc-170

[2]
https://cran.r-project.org/bin/windows/Rtools/

[3]
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/e1ff62d7f9acfbcb748e636f097fd2e6cc21d249/tree/mingw-w64-headers/crt/stdlib.h#l641

src/fread.c
```diff
@@ -2536,18 +2536,18 @@ int freadMain(freadMainArgs _args) {
...
-                    if (len > 1000) len = 1000;
-                    if (len > 0) {
-                      typeBumpMsg = realloc(typeBumpMsg, typeBumpMsgSize + len + 1);
-                      strcpy(typeBumpMsg+typeBumpMsgSize, temp);
-                      typeBumpMsgSize += len;
-                    }
+
+                    len = iminInt(len, sizeof(buffer));
+
+                    typeBumpMsg = realloc(typeBumpMsg, typeBumpMsgSize + len + 1);
```
MichaelChirico, line 2531: why move this out of the branch?

(it's in a `verbose` branch, so performance doesn't matter much, just curious your thinking)

MichaelChirico (approved):

src/fread.c
badasahog: I'm not sure I understand the question

MichaelChirico: Previously, `typeBumpMsg` only undergoes `realloc()` when `len>0`. Now it's done unconditionally.

badasahog: I don't think the 0 case is possible? how can snprintf return 0?

If it is possible, its extremely rare.

Fwiw, I don't think it's able to overshoot the buffer you give it, either.

MichaelChirico: Thanks

src/fread.c
```diff
@@ -2237,10 +2238,10 @@ int freadMain(freadMainArgs _args) {
   double tRead = 0, tReread = 0;
   double thRead = 0, thPush = 0;  // reductions of timings within the parallel region
   int max_col = 0;
-  char *typeBumpMsg = NULL; size_t typeBumpMsgSize = 0;
-  int typeCounts[NUMTYPE]; // used for verbose output; needs populating after first read and before reread (if any) -- see later comment
-  #define internalErrSize 1000
-  char internalErr[internalErrSize+1]="";  // must be compile time size: the message is generated and we can't free before STOP
+  char *typeBumpMsg = NULL;  size_t typeBumpMsgSize=0;
```
```suggestion
  char *typeBumpMsg = NULL;  size_t typeBumpMsgSize = 0;
```

src/fread.c
```diff
@@ -2237,10 +2238,10 @@ int freadMain(freadMainArgs _args) {
   double tRead = 0, tReread = 0;
   double thRead = 0, thPush = 0;  // reductions of timings within the parallel region
   int max_col = 0;
-  char *typeBumpMsg = NULL; size_t typeBumpMsgSize = 0;
-  int typeCounts[NUMTYPE]; // used for verbose output; needs populating after first read and before reread (if any) -- see later comment
-  #define internalErrSize 1000
-  char internalErr[internalErrSize+1]="";  // must be compile time size: the message is generated and we can't free before STOP
+  char *typeBumpMsg = NULL;  size_t typeBumpMsgSize = 0;
```
```suggestion
  char *typeBumpMsg = NULL; size_t typeBumpMsgSize = 0;
```

src/fread.c
```diff
@@ -2518,18 +2519,19 @@ int freadMain(freadMainArgs _args) {
...
                       j + 1, colNames ? " <<" : "", colNames ? (colNames[j].len) : 0, colNames ? (colNamesAnchor + colNames[j].off) : "", colNames ? ">>" : "",
                       typeName[IGNORE_BUMP(joldType)], typeName[IGNORE_BUMP(thisType)],
-                      (int)(tch - fieldStart), fieldStart, (int64_t)(ctx.DTi + myNrow));
-                    if (len > 1000) len = 1000;
-                    if (len > 0) {
-                      typeBumpMsg = realloc(typeBumpMsg, typeBumpMsgSize + len + 1);
-                      strcpy(typeBumpMsg + typeBumpMsgSize, temp);
-                      typeBumpMsgSize += len;
-                    }
+                      (int)(tch-fieldStart), fieldStart, (int64_t)(ctx.DTi + myNrow));
```
```suggestion
                      (int)(tch - fieldStart), fieldStart, (int64_t)(ctx.DTi + myNrow));
```

MichaelChirico (approved):

PR #7047, opened 2025-06-04
Clarified setindex vs setkey subsetting behavior

### LINKED ISSUE #5321: unexpected effects of setindex vs. setkey
> DT <- data.table(a = c(TRUE, FALSE), b = 1:2)

> setindex(DT, a)
DT[ .(TRUE), ] # ERROR

> setkey(DT, a)
DT[ .(TRUE), ] # 1: TRUE

Same with column `b`. Neither in the documentation nor in the vignettes exists any hint on the varying behavior. `setindex()` works only with character valued columns? Please add some hints in the documentation of your great package.

> sessionInfo()
R version 4.0.4 (2021-02-15)
Running under: Debian GNU/Linux 11 (bullseye)

> Matrix products: default

> locale:
 [1] LC_CTYPE=de_DE.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=de_DE.UTF-8        LC_COLLATE=de_DE.UTF-8    
 [5] LC_MONETARY=de_DE.UTF-8    LC_MESSAGES=de_DE.UTF-8   
 [7] LC_PAPER=de_DE.UTF-8       LC_NAME=C                 

> attached base packages:

> loaded via a namespace (and not attached):
[1] compiler_4.0.4

venom1204:
closes #5321 

This PR improves documentation to clarify the differences in subsetting behavior when using `setindex` versus `setkey`, particularly regarding the use of the `on=` argument with `DT[.(value)]` syntax.

in this pr i did the following things

- `datatable-secondary-indices-and-auto-indexing.Rmd`: Added a new section "Keyed vs. Indexed Subsetting" with examples and an explanation of the lookup hierarchy.
- `setkey.Rd` : Added a subsection "Key vs. Index Subsetting" to highlight usage differences, particularly the need for `on=` with `setindex` for join-like subsets.

hi @jangorecki ,can you please review this when you have time.
thank you.

### DIFF
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -155,23 +155,6 @@ setindex(DT, A)
 setindex(DT, B)
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
-
-# Keyed subsetting (no 'on' needed)
-DT = data.table(A=5:1,B=letters[5:1])
-setkey(DT, B)
-DT[.("e")]  # Returns row where B = "e"
-
-# Indexed subsetting (requires 'on')
-DT = data.table(A=5:1,B=letters[5:1])
-setindex(DT, B)
-DT[.("e"), on = "B"]  # Works
-# DT[.("e")]           # Would error (missing 'on')
-
-DT <- data.table(a = c(1,2,1,2), b = c("x", "x", "y", "y"), val = 1:4)
-# Set an index on a single column
-setindex(DT, a)
-# Set multiple indices at once using setindexv
-setindexv(DT, list("a", c("a", "b")))
   
 # Setting multiple indices at once
 DT = data.table(A = 5:1, B = letters[5:1], C = 10:6)
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -39,6 +39,20 @@ ordering vector of the dataset's rows according to the provided columns. This or
 is stored as an attribute of the \code{data.table} and the dataset retains the original order
 of rows in memory. See the \href{../doc/datatable-secondary-indices-and-auto-indexing.html}{\code{vignette("datatable-secondary-indices-and-auto-indexing")}} for more details.
 
+\subsection{Key vs. Index Subsetting}{
+When using \code{setkey}:
+\itemize{
+  \item Subsetting can omit \code{on} (e.g., \code{DT[.(value)]})
+  \item Data is physically reordered in RAM
+}
+When using \code{setindex}:
+\itemize{
+  \item Must specify \code{on} (e.g., \code{DT[.(value), on = "col"]})
+  \item Multiple indices can coexist via \code{setindexv(x, list(cols))}
+  \item Original row order is preserved
+}
+}
+
 \code{key} returns the \code{data.table}'s key if it exists; \code{NULL} if none exists.
 
 \code{haskey} returns \code{TRUE}/\code{FALSE} if the \code{data.table} has a key.
@@ -141,6 +155,15 @@ setindex(DT, A)
 setindex(DT, B)
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
+
+# Keyed subsetting (no 'on' needed)
+setkey(DT, B)
+DT[.("e")]  # Returns row where B = "e"
+
+# Indexed subsetting (requires 'on')
+setindex(DT, B)
+DT[.("e"), on = "B"]  # Works
+# DT[.("e")]           # Would error (missing 'on')
 
 # Setting multiple indices at once
 DT = data.table(A = 5:1, B = letters[5:1], C = 10:6)
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -149,7 +149,6 @@ setindex(DT, A)
 setindex(DT, B)
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
-  
 # Setting multiple indices at once
 DT = data.table(A = 5:1, B = letters[5:1], C = 10:6)
 setindexv(DT, list(c("A", "B"), c("B", "C")))
--- a/man/setkey.Rd
+++ b/man/setkey.Rd
@@ -39,20 +39,6 @@ ordering vector of the dataset's rows according to the provided columns. This or
 is stored as an attribute of the \code{data.table} and the dataset retains the original order
 of rows in memory. See the \href{../doc/datatable-secondary-indices-and-auto-indexing.html}{\code{vignette("datatable-secondary-indices-and-auto-indexing")}} for more details.
 
-\subsection{Key vs. Index Subsetting}{
-When using \code{setkey}:
-\itemize{
-  \item Subsetting can omit \code{on} (e.g., \code{DT[.(value)]})
-  \item Data is physically reordered in RAM
-}
-When using \code{setindex}:
-\itemize{
-  \item Must specify \code{on} (e.g., \code{DT[.(value), on = "col"]})
-  \item Multiple indices can coexist via \code{setindexv(x, list(cols))}
-  \item Original row order is preserved
-}
-}
-
 \code{key} returns the \code{data.table}'s key if it exists; \code{NULL} if none exists.
 
 \code{haskey} returns \code{TRUE}/\code{FALSE} if the \code{data.table} has a key.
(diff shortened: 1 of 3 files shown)


man/setkey.Rd
```diff
@@ -39,6 +39,20 @@ ordering vector of the dataset's rows according to the provided columns. This or
...
+\subsection{Key vs. Index Subsetting}{
+When using \code{setkey}:
+\itemize{
+  \item Subsetting can omit \code{on} (e.g., \code{DT[.(value)]})
+  \item Data is physically reordered in RAM
+}
+When using \code{setindex}:
+\itemize{
+  \item Must specify \code{on} (e.g., \code{DT[.(value), on = "col"]})
+  \item Multiple indices can coexist via \code{setindexv(x, list(cols))}
```
jangorecki: Not sure about this code here. Just here in comment could you make a reproducible example? Note that keys and indices can be composed of multiple columns

man/setkey.Rd
```diff
@@ -39,6 +39,20 @@ ordering vector of the dataset's rows according to the provided columns. This or
 is stored as an attribute of the \code{data.table} and the dataset retains the original order
 of rows in memory. See the \href{../doc/datatable-secondary-indices-and-auto-indexing.html}{\code{vignette("datatable-secondary-indices-and-auto-indexing")}} for more details.
 
+\subsection{Key vs. Index Subsetting}{
```
jangorecki: Did we ever used "subsection"? If not then I would stick just to "section*

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
```diff
@@ -62,6 +62,49 @@ Secondary indices are similar to `keys` in *data.table*, except for two major di
 
 * There can be more than one secondary index for a data.table (as we will see below).
 
+#### 1.1 Keyed vs. Indexed Subsetting
+
+While both **keys** and **indices** enable fast binary search subsetting, they differ critically in usage:
```
jangorecki: Word critically is not a great fit here. They differ, for good reason, nothing critical here :)

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
```diff
@@ -62,6 +62,49 @@ Secondary indices are similar to `keys` in *data.table*, except for two major di
...
+
+**Indexed subsetting** (explicit column specification)
+
+```{r}
+DT <- data.table(a = c(TRUE, FALSE), b = 1:2)
+setindex(DT, a)              # Set index only (no reorder)
+DT[.(TRUE), on = "a"]        # Must specify 'on'
+```
+
+Always use `on`with indices to avoid errors. Keys establish permanent implicit matching, while indices require explicit direction for each operation.
```
jangorecki: "Always use..." - there is no other way so the advice is rather not very practical. Whole line doesn't really make clear message here.

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
```diff
@@ -62,6 +62,49 @@ Secondary indices are similar to `keys` in *data.table*, except for two major di
 
 * There can be more than one secondary index for a data.table (as we will see below).
 
+#### 1.1 Keyed vs. Indexed Subsetting
+
+While both **keys** and **indices** enable fast binary search subsetting, they differ critically in usage:
+
+**Keyed subsetting** (implicit column matching)
+```{r}
+DT <- data.table(a = c(TRUE, FALSE), b = 1:2)
```
jangorecki: If we use = for assignment in this vignette then good to stick to that style, otherwise ignore

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
```diff
@@ -62,6 +62,49 @@ Secondary indices are similar to `keys` in *data.table*, except for two major di
...
+
+```{r}
+DT <- data.table(a = c(TRUE, FALSE), b = 1:2)
+setindex(DT, a)              # Set index only (no reorder)
+DT[.(TRUE), on = "a"]        # Must specify 'on'
+```
+
+Always use `on`with indices to avoid errors. Keys establish permanent implicit matching, while indices require explicit direction for each operation.
+
+#### **Hierarchy of Lookup Operations**
```
jangorecki: Header I think doesn't need to be bold **

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
```diff
@@ -62,6 +62,49 @@ Secondary indices are similar to `keys` in *data.table*, except for two major di
...
+
+When `on` is specified, `data.table` uses this lookup order:
+
+- Existing keys matching `on` columns
+
+- Pre-built secondary indices matching `on` columns
+
+- Temporary auto-generated index (if none exist)
```
jangorecki: This index won't be temporary

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
```diff
@@ -62,6 +62,49 @@ Secondary indices are similar to `keys` in *data.table*, except for two major di
...
+
+**Example**: Error When Omitting on with Indices
+
+```{r}
+DT <- data.table(a = c(TRUE, FALSE), b = 1:2)
```
jangorecki: It feels like repeating what was already explained above...

aitap:
On Wed, 04 Jun 2025 12:27:20 -0700
venom1204 wrote:

> - `setkey.Rd` (or `setindex.Rd` - *specify which one you primarily
> edited for this point*):

Is that instruction for the reviewer, or for you?

venom1204:
Thanks for pointing that out! That instruction ,it is unrelated to the review; it was a note for myself that I forgot to remove.

vignettes/datatable-secondary-indices-and-auto-indexing.Rmd
venom1204: hi 
i updated the sections as you suggested kindly have a view when you have time 
thanks.

jangorecki:
> Thanks for pointing that out! That instruction ,it is unrelated to the review; it was a note for myself that I forgot to remove.

Reading output of `git diff --cached` just before `git commit` is very useful step in contribution process that helps to spot issues like that.

venom1204:
hi @MichaelChirico 
When you have a moment, could you please review this? I've made some touch-ups and would appreciate your feedback.
Thanks!

man/setkey.Rd
```diff
@@ -74,6 +74,20 @@ The sort is \emph{stable}; i.e., the order of ties (if any) is preserved.
...
+  \item Subsetting can omit \code{on} (e.g., \code{DT[.(value)]})
+  \item Data is physically reordered in RAM
+}
+When using \code{setindex}:
+\itemize{
+  \item Must specify \code{on} (e.g., \code{DT[.(value), on = "col"]})
+  \item Multiple indices can coexist via \code{setindexv(x, list(cols))}
+  \item Original row order is preserved
+}
+}
```
MichaelChirico: I find it a bit confusing to mix topics here (operation of `setkey()` vs. `setindex()`, and differences of `[` join subsetting on keyed vs. indexed tables).

```suggestion
\section{Keys vs. Indices}{
Setting a key (with \code{setkey}) and an index (with \code{setindex}) are similar, but have very important distinctions.

Setting a key physically reorders the data in RAM.

Setting an index computes the sort order, but instead of applying the reordering, simply \emph{stores} this computed ordering. That means that multiple indices can coexist, and that the original row order is preserved.
}
```

man/setkey.Rd
```diff
@@ -142,6 +156,23 @@ setindex(DT, B)
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
 
+# Keyed subsetting (no 'on' needed)
```
MichaelChirico: I would instead put these examples in `?data.table`, which is where `[` is covered.

venom1204: hi, I did the modification ,kindly have a view .
thanks ..

man/setkey.Rd
```diff
@@ -141,7 +149,7 @@ setindex(DT, A)
 setindex(DT, B)
 indices(DT)           # get indices single vector
 indices(DT, vectors = TRUE) # get indices list
-
+
```
MichaelChirico:
```suggestion

```

Issue #6641, opened 2024-12-09
Error messages from merge() somewhat confusingly "swap" 'x' and 'i' prefixes


MichaelChirico:
```r
DT1=data.table(a=factor('a'))
DT2=data.table(a=1L)
merge(DT1, DT2, by='a')
# Error in bmerge(i, x, leftcols, rightcols, roll, rollends, nomatch, mult,  : 
#   Incompatible join types: x.a (integer) and i.a (factor). Factor columns must join to factor or character columns.
```

Note that the `x.` prefix applies to `DT2`, while the `i.` prefix applies to `DT1`, while the user provided them in order `(DT1, DT2)`.

At root is that under the hood, `merge(x,y, ...)` is constructed as a join with `y[x, ...]`.

It would be _most_ consistent with users' expectations of `merge()` if the prefixes were `DT1 : x.`, `DT2: y.`, or perhaps even better if the errors matched `suffixes=`. A really subtle (and back-incompatible) way to thread the needle here would be to change the `merge.data.table()` defaults to be `suffixes=c(".i", ".x")`.

A different approach (and probably the most practical at this point) is just to emphasize this consideration in the docs.

rikivillalba:
Perhaps one can raise an errorCondition with a custom 'call' argument such that  'bmerge(i, x, leftcols,...' becomes  'bmerge(i = DT2, x = DT1, leftcols,...'

MichaelChirico:
Great idea! I think we can populate custom attributes of the error object raised on the custom class and use those more reliably & transparently than messing with the call.

Abhishek2634:
Hi @MichaelChirico, I would like to work on this issue. can you assign me this issue, As i have figure out what to do.

MichaelChirico:
Thanks @Abhishek2634, the issue with your proposed fix is it's incorrect for the "primary" merge interface, namely `DT1[DT2, on='a']`.

The root issue here is that `merge(DT1, DT2, by='a')` eventually runs `DT2[DT1, on='a']`, i.e., in the reverse order of the arguments to `merge()`. And moreover, `DT2[DT1, on='a']` uses `x` prefix to refer to the "outer" table, `i` prefix for the inner table (matching the names of the arguments to `[`), whereas the `merge()` arguments are named `x` and `y`, respectively.

The goal of this issue is to match the output to users' expectations in both cases (joins with `x[i]` and `merge(x, y)`).

It will probably require some understanding of condition objects in R:

[...]

PR #7048, opened 2025-06-05
Clarify x/i error prefixes in merge() for incompatible factor joins

Mukulyadav2004:
fixes: #6641 

This PR addresses confusing error messages from merge(x, y) when an incompatible factor join occurs ( factor with integer). The original error message used x. and i. prefixes based on the internal y[x] join structure, which was misleading for merge() users as it swapped the references to merge's x and y arguments.

So following are the changes to fix the x/i confusion for merge() users:

`R/bmerge.R`:
- For this incompatible factor join, `bmerge` now signals a custom error object  `bmerge_incompatible_type_error`.
- This error object includes attributes with the column names and types from bmerge's inputs.

`R/merge.R` :
- A `tryCatch` is added around the main internal join call (y[x]) to specifically handle `bmerge_incompatible_type_error`.
- The handler uses the attributes from the caught error object to construct a new error message. This message correctly uses x.colname to refer to merge's first argument (x) and y.colname for its second argument (y). 

@MichaelChirico @jangorecki @tdhock can you please review the modifications.

### DIFF
--- a/R/bmerge.R
+++ b/R/bmerge.R
@@ -84,7 +84,18 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           next
         }
       }
-      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
+      condition_message = sprintf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
+      condition <- structure(
+        list(
+          message = condition_message,
+          bmerge_x_arg_col_name      = names(x)[xcol],
+          bmerge_x_arg_type          = x_merge_type,
+          bmerge_i_arg_col_name      = names(i)[icol],
+          bmerge_i_arg_type          = i_merge_type
+        ),
+        class = c("bmerge_incompatible_type_error", "data.table_error", "error", "condition")
+      )
+      stop(condition)
     }
     # we check factors first to cater for the case when trying to do rolling joins on factors
     if (x_merge_type == i_merge_type) {
--- a/R/merge.R
+++ b/R/merge.R
@@ -92,13 +92,25 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
   if (!is.null(incomparables)) {
     # %fin% to be replaced when #5232 is implemented/closed
     "%fin%" = function(x, table) if (is.character(x) && is.character(table)) x %chin% table else x %in% table
-    xind = rowSums(x[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.x]) == length(by)
-    yind = rowSums(y[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.y]) == length(by)
+    xind = rowSums(x[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.x]) == length(by.x)
+    yind = rowSums(y[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.y]) == length(by.y)
     # subset both so later steps still work
     x = x[xind]
     y = y[yind]
   }
-  dt = y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]   # includes JIS columns (with a i. prefix if conflict with x names)
+  dt = tryCatch({
+      y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]
+    },
+    bmerge_incompatible_type_error = function(e) {
+      x_part_col_name = e$bmerge_i_arg_col_name
+      x_part_type     = e$bmerge_i_arg_type
+      y_part_col_name = e$bmerge_x_arg_col_name
+      y_part_type     = e$bmerge_x_arg_type
+
+      msg = sprintf("Incompatible join types: x.%s (%s) and i.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type)
+      stop(errorCondition(message = msg, class = c("datatable_merge_type_error", "data.table_error", "error", "condition")))
+    }
+  )
 
   if (all.y && nrow(y)) {  # If y does not have any rows, no need to proceed
     # Perhaps not very commonly used, so not a huge deal that the join is redone here.


R/bmerge.R
```diff
@@ -84,7 +84,18 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
           next
         }
       }
-      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
+      condition_message = sprintf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
```
MichaelChirico:
```suggestion
      condition_message = gettextf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", xname, x_merge_type, iname, i_merge_type)
```

R/bmerge.R
```diff
@@ -84,7 +84,18 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
+      condition <- structure(
+        list(
+          message = condition_message,
+          bmerge_x_arg_col_name      = names(x)[xcol],
+          bmerge_x_arg_type          = x_merge_type,
+          bmerge_i_arg_col_name      = names(i)[icol],
+          bmerge_i_arg_type          = i_merge_type
+        ),
+        class = c("bmerge_incompatible_type_error", "data.table_error", "error", "condition")
```
MichaelChirico: Elsewhere in the code base we've used the "namespace" `dt_` for our error class, let's stick with that:

https://github.com/Rdatatable/data.table/blob/5bbc4d516eb9a05471e7b5221c8c23a540af97a4/R/data.table.R#L2834
https://github.com/Rdatatable/data.table/blob/5bbc4d516eb9a05471e7b5221c8c23a540af97a4/R/fcast.R#L178

```suggestion
        class = c("dt_bmerge_incompatible_type_error", "error", "condition")
```

R/bmerge.R
```diff
@@ -84,7 +84,18 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
+      )
+      stop(condition)
```
MichaelChirico, line 97: `stopf()` adds a `call` entry to the error object, too:

https://github.com/Rdatatable/data.table/blob/5bbc4d516eb9a05471e7b5221c8c23a540af97a4/R/translation.R#L8

That will let this new error object play well with `conditionCall()`

R/merge.R
```diff
@@ -92,13 +92,25 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    yind = rowSums(y[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.y]) == length(by.y)
     # subset both so later steps still work
     x = x[xind]
     y = y[yind]
   }
-  dt = y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]   # includes JIS columns (with a i. prefix if conflict with x names)
+  dt = tryCatch({
+      y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]
+    },
+    bmerge_incompatible_type_error = function(e) {
```
MichaelChirico: please add a unit test in inst/tests/tests.Rraw.

R/merge.R
```diff
@@ -92,13 +92,25 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+      x_part_col_name = e$bmerge_i_arg_col_name
+      x_part_type     = e$bmerge_i_arg_type
+      y_part_col_name = e$bmerge_x_arg_col_name
+      y_part_type     = e$bmerge_x_arg_type
+
+      msg = sprintf("Incompatible join types: x.%s (%s) and i.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type)
```
MichaelChirico:
```suggestion
      msg = gettextf("Incompatible join types: x.%s (%s) and i.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type)
```

R/merge.R
```diff
@@ -92,13 +92,25 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+      stop(errorCondition(message = msg, class = c("datatable_merge_type_error", "data.table_error", "error", "condition")))
```
MichaelChirico: Here, we can just use `stopf()` since we don't need to pass along any further attributes:

```suggestion
      stopf("Incompatible join types: x.%s (%s) and i.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
```

Also, I'm not sure there's any need as of now for _this_ error to signal its _own_ custom condition class. I think we can just leave it with the default class for now unless there's a future request.

Mukulyadav2004: Hi @MichaelChirico 
I have made the changes as per you suggestion. But when I run this, then it gives - `Error in value[3L] :`  I can't figure out whether classes passed to `stopf `are not correctly handled, or if there's an issue with the call stack that `stopf` tries to report.
```
> DT1=data.table(a=factor('a'))
> DT2=data.table(a=1L)
> merge(DT1, DT2, by='a')
Error in value[3L] :
Incompatible join types: x.a (factor) and i.a (integer). Factor columns must join to factor or character columns.
```

R/merge.R
```diff
@@ -124,4 +133,4 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
   # retain custom classes of first argument that resulted in dispatch to this method, #1378
   setattr(dt, "class", class_x)
   dt
-}
+}
```
MichaelChirico:
```suggestion
}

```

R/bmerge.R
```diff
@@ -84,7 +84,19 @@ bmerge = function(i, x, icols, xcols, roll, rollends, nomatch, mult, ops, verbos
...
+        list(
+          message = condition_message,
+          call = sys.call(sys.nframe()-1L),
+          bmerge_x_arg_col_name      = names(x)[xcol],
+          bmerge_x_arg_type          = x_merge_type,
+          bmerge_i_arg_col_name      = names(i)[icol],
+          bmerge_i_arg_type          = i_merge_type
+        ),
+        class = c("dt_bmerge_incompatible_type_error", "error", "condition")
+      )
```
MichaelChirico: `structure()` is weirdly slow; better to just add `class()` as a follow-up step

```suggestion
      condition = list(
        message = condition_message,
        call = sys.call(sys.nframe() - 1L),
        bmerge_x_arg_col_name      = names(x)[xcol],
        bmerge_x_arg_type          = x_merge_type,
        bmerge_i_arg_col_name      = names(i)[icol],
        bmerge_i_arg_type          = i_merge_type
      )
      class(condition) = c("dt_bmerge_incompatible_type_error", "error", "condition")
```

R/merge.R
```diff
@@ -90,18 +90,27 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+  dt = tryCatch({
+      y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]
+    },
+    dt_bmerge_incompatible_type_error = function(e) {
+      x_part_col_name = e$bmerge_i_arg_col_name
+      x_part_type     = e$bmerge_i_arg_type
+      y_part_col_name = e$bmerge_x_arg_col_name
+      y_part_type     = e$bmerge_x_arg_type
+
+      stopf("Incompatible join types: x.%s (%s) and i.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
```
MichaelChirico:
```suggestion
      stopf("Incompatible join types: x.%s (%s) and y.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
```

R/merge.R
```diff
@@ -90,18 +90,27 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
     x = x[xind]
     y = y[yind]
   }
-  dt = y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]   # includes JIS columns (with a i. prefix if conflict with x names)
 
-  if (all.y && nrow(y)) {  # If y does not have any rows, no need to proceed
-    # Perhaps not very commonly used, so not a huge deal that the join is redone here.
+  dt = tryCatch({
+      y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]
+    },
```
MichaelChirico:
```suggestion
  dt = tryCatch(
      y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian],
```

R/merge.R
```diff
@@ -90,18 +90,26 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+    yind = rowSums(y[, lapply(.SD, function(y_col_val) !(y_col_val %fin% incomparables)), .SDcols=by.y]) == length(by.y)
     x = x[xind]
     y = y[yind]
   }
-  dt = y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian]   # includes JIS columns (with a i. prefix if conflict with x names)
 
-  if (all.y && nrow(y)) {  # If y does not have any rows, no need to proceed
-    # Perhaps not very commonly used, so not a huge deal that the join is redone here.
+  dt = tryCatch(
+      y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian],
```
MichaelChirico:
```suggestion
    y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian],
```

R/merge.R
```diff
@@ -90,18 +90,26 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
-    # Perhaps not very commonly used, so not a huge deal that the join is redone here.
+  dt = tryCatch(
+    y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian],
+    dt_bmerge_incompatible_type_error = function(e) {
+      x_part_col_name = e$bmerge_i_arg_col_name
+      x_part_type     = e$bmerge_i_arg_type
+      y_part_col_name = e$bmerge_x_arg_col_name
+      y_part_type     = e$bmerge_x_arg_type
+
+      stopf("Incompatible join types: x.%s (%s) and y.%s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
```
MichaelChirico: In terms of translation, we'll now have two _very_ similar messages:

```
Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.
Incompatible join types: x.%s (%s) and y.%s (%s). Factor columns must join to factor or character columns.
```

That makes translation a bit harder to maintain -- if different translators (or the same translator with a sufficiently large gap in time) translates these two messages, there's a decent chance they'll use subtly different wording, whereas they are exactly the same in English. Better to use identical messages, so that there's only one thing to translate.

MichaelChirico:
```suggestion
      x_part_col_name = paste0("x.", e$bmerge_i_arg_col_name)
      x_part_type     = e$bmerge_i_arg_type
      y_part_col_name = paste0("y.", e$bmerge_x_arg_col_name)
      y_part_type     = e$bmerge_x_arg_type

      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
```

R/merge.R
```diff
@@ -90,18 +90,26 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
   # implement incomparables argument #2587
   if (!is.null(incomparables)) {
-    # %fin% to be replaced when #5232 is implemented/closed
-    "%fin%" = function(x, table) if (is.character(x) && is.character(table)) x %chin% table else x %in% table
-    xind = rowSums(x[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.x]) == length(by)
-    yind = rowSums(y[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.y]) == length(by)
-    # subset both so later steps still work
+    "%fin%" = function(x_val, table_val) if (is.character(x_val) && is.character(table_val)) x_val %chin% table_val else x_val %in% table_val
+    xind = rowSums(x[, lapply(.SD, function(x_col_val) !(x_col_val %fin% incomparables)), .SDcols=by.x]) == length(by.x)
+    yind = rowSums(y[, lapply(.SD, function(y_col_val) !(y_col_val %fin% incomparables)), .SDcols=by.y]) == length(by.y)
```
MichaelChirico, line 95:
```suggestion
    # %fin% to be replaced when #5232 is implemented/closed
    "%fin%" = function(x, table) if (is.character(x) && is.character(table)) x %chin% table else x %in% table
    xind = rowSums(x[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.x]) == length(by)
    yind = rowSums(y[, lapply(.SD, function(x) !(x %fin% incomparables)), .SDcols=by.y]) == length(by)
    # subset both so later steps still work
```

MichaelChirico: I see, this is to disambiguate: lambda argument `x` collides with parent function `merge()`'s own `x` argument, which is confusing.

(for future reference, calling out the purpose of unrelated changes in the PR description is helpful to the reviewer)

R/merge.R
```diff
@@ -90,18 +90,26 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+      x_part_col_name = paste0("x.", e$bmerge_i_arg_col_name)
+      x_part_type     = e$bmerge_i_arg_type
+      y_part_col_name = paste0("y.", e$bmerge_x_arg_col_name)
+      y_part_type     = e$bmerge_x_arg_type
+
+      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
+    }
+  )
+
+  if (all.y && nrow(y)) {
```
MichaelChirico:
```suggestion
  if (all.y && nrow(y)) {  # If y does not have any rows, no need to proceed
    # Perhaps not very commonly used, so not a huge deal that the join is redone here.
```

R/merge.R
```diff
@@ -90,15 +90,24 @@ merge.data.table = function(x, y, by = NULL, by.x = NULL, by.y = NULL, all = FAL
...
+
+  dt = tryCatch(
+    y[x, nomatch=if (all.x) NA else NULL, on=by, allow.cartesian=allow.cartesian],
+    dt_bmerge_incompatible_type_error = function(e) {
+      x_part_col_name = paste0("x.", e$bmerge_i_arg_col_name)
+      x_part_type     = e$bmerge_i_arg_type
+      y_part_col_name = paste0("y.", e$bmerge_x_arg_col_name)
+      y_part_type     = e$bmerge_x_arg_type
+
+      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type, class = "dt_merge_incompatible_type_error")
```
MichaelChirico:
```suggestion
      stopf("Incompatible join types: %s (%s) and %s (%s). Factor columns must join to factor or character columns.", x_part_col_name, x_part_type, y_part_col_name, y_part_type)
```

MichaelChirico (commented):
This is a good start, but it doesn't close the issue -- this only addresses one possible error coming from `bmerge()` where the order gets mangled. Consider:

```r
DT1 = data.table(a = TRUE)
DT2 = data.table(a = as.raw(2))
merge(DT1, DT2, by='a')
#   x.a is type raw which is not supported by data.table join
```

Here again we get `x.` prefix on the `y=` input.

Ideally, all such errors that could be raised to `bmerge()` are caught and transposed.

R/merge.R
MichaelChirico: OK, let's just use `stop()` for simplicity.

Issue #7049, opened 2025-06-05
Status on spacing in formatting (and other formatting rules)


badasahog:
It is clear that datatable uses very unusual formatting rules, like `x=y;` instead of `x = y;`.

It appears that these have been mostly abandoned creating an uncomfortable mix between the two styles.

Are we willing to commit to one or the other? If so I could have a pr done in a fairly quick manner.

jangorecki:
I would say to agree that any new code should be x = y; and eventually one day when there will be not many PRs left in the queue, we can think about changing existing code.

badasahog:
I can do it right now for fread, there aren't *that man* outstanding pulls.

Or I can do it for the whole project, and break it into a separate pr for each file.

Would like maintainer feedback on this.

MichaelChirico:
I agree there are not many pending changes in fread -- I am the author of most of the others, which are marked as `Draft` and will need to be rebased before review anyway. Go ahead and make the change for fread if it is bugging you :)

Generally, our approach has been to make incremental changes -- minor edits near the code you are actually changing.

aitap:
The problem with the "let me fix your code style for you" patches is
their high effort/reward ratio. At best the code will work as well as
it used to; more likely there will be even more problems to fix as a
result of the fix (e.g. #7039, #7040, #7042).

The results of the style cleanup will erode over time unless enforced
by a style linter. If you do want the style to be fixed, please stir
the discussion in #5982 on which exact style choice data.table needs to
adopt and how to enforce it automatically.

Issue #7023, opened 2025-05-26
Inconsistent use of postfix/prefix operators


badasahog:
`x++;` is the same as `++x;` when used in isolation (and the evaluated value is not used).

The data.table codebase appears to use both forms, both in isolation and in for loops. I prefer postfix (`x++`), but either way it would be better to make it consistent across all source files.

MichaelChirico:
I agree there are more productive uses of time at the moment. @badasahog, you may be interested to follow #5982 as well.

PR #7052, opened 2025-06-06
Formatting improvements

badasahog:
Partial progress toward closing #7049 and #7023

I want to make sure maintainers are on board or at least provide feedback if they thing this style is bad.

After this is done, the best way to improve readability (in my opinion) is to rename identifiers, and create well named utility functions.


badasahog:
Also, looks like some extra files from visual studio made their way into the pr somehow. If anyone knows how to remove them, be my guest.

.vs/data.table/v17/DocumentLayout.json
```diff
@@ -0,0 +1,41 @@
...
+            },
+            {
+              "$type": "Document",
+              "DocumentIndex": 0,
+              "Title": "bmerge.c",
+              "DocumentMoniker": "E:\\data.table\\src\\bmerge.c",
+              "RelativeDocumentMoniker": "src\\bmerge.c",
+              "ToolTip": "E:\\data.table\\src\\bmerge.c",
+              "RelativeToolTip": "src\\bmerge.c",
+              "ViewState": "AgIAAFIBAAAAAAAAAAAAANEBAAAJAAAAAAAAAA==",
```
tdhock: please undo addition of these vscode files

MichaelChirico: you can add `.vs` to .gitignore if you'd rather not be bothered having to delete these files locally

aitap:
At the very least, let's not touch the indentation for now (see
assign.c).

jangorecki (changes requested):
Wasn't that meant to be for fread only at the moment?

I would appreciate at least an analysis how many conflicts it will cause to open PRs before deciding on merging that.
As well said by @aitap 
> high effort/reward ratio. At best the code will work as well as
it used to; more likely there will be even more problems to fix as a
result of the fix

Is it so difficult to understand that we have limited resources on pursuing some of PRs that are pending for 2-3 years, and this single PR does not resolve any problem but create a risk that those resources will run out before those pending PRs will be merged?

badasahog:
@jangorecki I respect your feedback. I'll localize to fread instead in a separate pr, but I don't think there's any harm in keeping this open.

In terms of your assertion that big pull requests like this slow down development, I think resolving these collisions would be fairly straight forward, and will help make development smoother in the future.

Again, I can tell that this may have been over zealous, so I'll try to read the room, and stick with smaller PRs, like ones just localized to fread.

MichaelChirico:
> I think resolving these collisions would be fairly straight forward

I think in isolation, yes, but they quickly become entangled with real changes and the mental load to tease apart "real" vs superficial changes increases something like exponentially.

TysonStanley:
I certainly don't mind the style, I'm a fan of well formatted code like this. However, I have to agree with @MichaelChirico and @jangorecki  knowing how many collisions will occur with already existing PRs and other large upcoming PRs.

jangorecki:
Here is an example of a tiny change and a conflict that it created, just from 2 weeks ago, resolved last week: https://github.com/Rdatatable/data.table/commit/82e66e7abe6f74b14023677706560ebd9a079df4#r158650259
It is only on one of the PRs, same and then also more complex conflicts (because other PRs modified that code again) will have to be resolved again, and again, in other PRs that are waiting in the queue.
This just doesn't make sense. If there is code formatting to be applied on a file, let's first merge (or close) all PRs that are attempting to modify that file, and then apply formatting change. As simple as that.

MichaelChirico:
Following up once more:

 - We _really do_ appreciate such changes. It's with regrets that we don't accept them at the moment due to a slowly-clearing PR backlog (our fault, not yours). We recognize the value of code quality changes which just generally make it less tedious + reduce cognitive load to do new development/bug fixing work.
 - You're absolutely right there's no harm to leaving this open, please do. Thank you for all your recent contributions!

Issue #7051, opened 2025-06-06
gcc-UBSAN on CRAN errors on `forder.c:137:3: runtime error: null pointer passed as argument 2`


helske:
I got a message from CRAN regarding [`seqHMM`](https://cran.r-project.org/package=seqHMM) package that there is a test failuri with gcc-UBSAN stemming from `data.table`: https://www.stats.ox.ac.uk/pub/bdr/memtests/gcc-UBSAN/seqHMM/tests/testthat.Rout, namely `forder.c`. 

I'll paste the trimmed output here in case the log files disappear:

```
> library(testthat)
> library(seqHMM)
Please cite seqHMM appropriately, see `citation('seqHMM')` for details.
> 
> test_check("seqHMM")
forder.c:137:3: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 0x7b6289002f38 in flush /tmp/Rtmp8svXjO/R.INSTALL29ad3d450313e2/data.table/src/forder.c:137
    #1 0x7b628907a303 in flush /tmp/Rtmp8svXjO/R.INSTALL29ad3d450313e2/data.table/src/forder.c:128
    #2 0x7b628907a303 in radix_r /tmp/Rtmp8svXjO/R.INSTALL29ad3d450313e2/data.table/src/forder.c:1327
    #3 0x7b6289085235 in forder /tmp/Rtmp8svXjO/R.INSTALL29ad3d450313e2/data.table/src/forder.c:804
    #4 0x7b628908a830 in forderReuseSorting /tmp/Rtmp8svXjO/R.INSTALL29ad3d450313e2/data.table/src/forder.c:1764
    #5 0x000000742222 in R_doDotCall /data/gannet/ripley/R/svn/R-devel/src/main/dotcode.c:780
    #6 0x0000008d6f6a in bcEval_loop /data/gannet/ripley/R/svn/R-devel/src/main/eval.c:8668
    #7 0x0000008ad7ab in bcEval /data/gannet/ripley/R/svn/R-devel/src/main/eval.c:7501
    #8 0x00000084e0f2 in Rf_eval /data/gannet/ripley/R/svn/R-devel/src/main/eval.c:1167
    #9 0x000000865b0a in R_execClosure /data/gannet/ripley/R/svn/R-devel/src/main/eval.c:2393
    #10 0x000000869d1a in applyClosure_core /data/gannet/ripley/R/svn/R-devel/src/main/eval.c:2306
[...]

Issue #7054, opened 2025-06-08
`frollapply` doesn't handle zero-length output


aitap:
Found while looking for unprotected calls to `memcpy` for #7051:

```r
data.table::frollapply(list(integer()), 0, \(x) 1)
```
```
froll.c:436:15: runtime error: pointer index expression with base 0x000000000001 overflowed to 0xfffffffffffffff9
froll.c:436:21: runtime error: store to misaligned address 0xfffffffffffffff9 for type 'double', which requires 8 byte alignment
0xfffffffffffffff9: note: pointer points here
<memory cannot be printed>

 *** caught segfault ***
address (nil), cause 'unknown'

Traceback:
 1: data.table::frollapply(list(integer()), 0, function(x) 1)
An irrecoverable exception occurred. R is aborting now ...
```

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff3f2341c in frollapply (x=0x1, nx=0, w=w@entry=0x1, k=0, ans=ans@entry=0x55555618c070, align=align@entry=1, fill=fill@entry=nan(0x0000007a2), call=0x555558b09ae0, 
    rho=0x555558b09ed0, verbose=false) at froll.c:436
436         ans->dbl_v[k-1] = REAL(eval0)[0];
(gdb) bt
#0  0x00007ffff3f2341c in frollapply (x=0x1, nx=0, w=w@entry=0x1, k=0, ans=ans@entry=0x55555618c070, align=align@entry=1, fill=fill@entry=nan(0x0000007a2), 
    call=0x555558b09ae0, rho=0x555558b09ed0, verbose=false) at froll.c:436
#1  0x00007ffff3f24bf6 in frollapplyR (fun=0x555558b075d8, obj=0x5555585c8448, k=<optimized out>, fill=<optimized out>, align=<optimized out>, rho=0x555558b09ed0)
    at frollR.c:292
```

I was looking for calls of the form `memcpy(<destination>, <invalid pointer>, 0)` (which typically don't show any symptoms until additional sanitizer checks at CRAN find it; I think it also happens here), but `frollapply()` actually dereferences the invalid pointer and crashes the process.

jangorecki:
It will be resolved by #5575 which moved frollapply implementation from C to R.

---

Actually it will be resolved even sooner, as PR #5441, preceding PR #5575, already adds an extra check:
```r
data.table::frollapply(list(integer()), 0, function(x) 1)
[...]

PR #7055, opened 2025-06-08
Avoid `memcpy(dst, NULL, 0)` in `forder.c`

aitap:
Closes: #7051. Tested manually by running the reproducer in the same configuration that previously showed the warning. The NEWS item is technically correct but may be overly detailed.

Looking at the remaining `memcpy()` calls, I wasn't able to find any definite examples of the same thing happening; there's usually a way to prove that the size is above 0. Possible issues in the new `froll` code are tracked in #7054.

### DIFF
--- a/src/forder.c
+++ b/src/forder.c
@@ -128,6 +128,7 @@ static void flush(void) {
   if (!retgrp) return;
   int me = omp_get_thread_num();
   int n = gs_thread_n[me];
+  // normally doesn't happen, can be encountered under heavy load, #7051
   if (!n) return; // # nocov
   int newn = gs_n + n;
   if (gs_alloc < newn) {


MichaelChirico:
LGTM (needs a test); I'll let Jan review based on how it will fit into the ongoing froll workstream.

jangorecki (approved):
LGTM, could have nocov marker because codecov now complaints about added line

aitap:
Thanks! Currently I don't know any way besides adding the # nocov. I
can reproduce the issue either by running cat /dev/urandom > /dev/null
on every CPU core or by pinning all data.table threads to one CPU core
prior to running the R script, and neither seems appropriate for an R
package test.

Issue #2487, opened 2017-11-15
first(named_vector) removes names


franknarf1:
``` r
library(data.table)
x = setNames(, "A")
x[1]
#>   A 
#> "A"
first(x)
#> [1] "A"
```

In `example(first)`, it suggests `first(x)` and `x[1]` should be interchangeable, but it doesn't hold here.

MichaelChirico:
To be clear, you aren't requesting that `first(x)` keeps the names, just that the documentation should clarify this distinction, right?

franknarf1:
@MichaelChirico Yeah, a very minor request, but it bit me when I was using `lapply` and then accessing the results by name. Maybe `first`/`last` also strips other attributes; I didn't check and don't know the internals, so I'm not sure exactly what the doc should say about it.

Henrik-P:
Somewhat related: [Documentation: first(x) # [not always the] same as x[1]](https://github.com/Rdatatable/data.table/issues/2002)

jangorecki:
related https://github.com/Rdatatable/data.table/pull/4065#issuecomment-557763540

PR #7056, opened 2025-06-09
Clarify behavior of first() and last()

### LINKED ISSUE #2002: Documentation: first(x) # [not always the] same as x[1]
From `?first`: `first(x)` # same as `x[1]`.

However, because of this...
```
if (!length(x)) 
                return(x)
```
...in the code of `first`, the result of `first(x)` is not the same as `x[1]` when `x` is a zero-length vector: `x` vs. `NA`. The behaviour of `first` is fine with me, but this difference may be clarified in the help text.

------

Just to show this difference, here's a small (possibly contrieved) small example. Find the first index of `x == 1`, within each group, where group "b" have no `1`s
```
d <- data.table(grp = c("a", "a", "a", "b", "b", "c", "c"),
                x = c(0, 1, 1, 0, 0, 1, 1))

#    grp x
# 1:   a 0
# 2:   a 1
# 3:   a 1
# 4:   b 0
# 5:   b 0
# 6:   c 1
# 7:   c 1

# select first element using [1]
d[ , .(first.true = which(x == 1)[1]), by = grp]
#    grp first.true
# 1:   a          2
# 2:   b         NA <~~ zero-length result of grp b appears as NA
# 3:   c          1

# select first element using `first`
d[ , .(first.true = first(which(x == 1))), by = grp]
#    grp first.true
# 1:   a          2 <~~ zero-length result of grp b is "gone" (still zero-length)
# 2:   c          1
```

Mukulyadav2004:
Closes #2002 and #2487 

Adds a note to the documentation for `first()` and `last()` to clarify they are not direct replacements for base R subsetting (i.e., `x[1]` and `x[length(x)]`).

The note explains the different handling of zero-length vectors and the stripping of attributes. This helps prevent common user errors.

@tdhock @jangorecki @Anirban166 could you please review these docs update.

### DIFF
--- a/man/last.Rd
+++ b/man/last.Rd
@@ -19,7 +19,7 @@ of \code{xts::first} is deployed. }
 are passed through to \code{xts}'s first/last. }
 }
 \details{
-Note: \code{first(x)} is not an exact replacement for \code{x[1]}. It differs in its handling of zero-length vectors (returning a zero-length vector instead of \code{NA}) and does not preserve attributes like names.
+Note: \code{first(x)} and \code{last(x)} are not exact replacements for \code{x[1]} and \code{x[length(x)]}. They differ in handling of zero-length vectors (returning a zero-length vector instead of \code{NA}) and do not preserve attributes like names.
 }
 \value{
 If no other arguments are supplied it depends on the type of \code{x}. The first/last item
--- a/man/last.Rd
+++ b/man/last.Rd
@@ -19,7 +19,7 @@ of \code{xts::first} is deployed. }
 are passed through to \code{xts}'s first/last. }
 }
 \details{
-Note: \code{first(x)} and \code{last(x)} are designed to behave like \code{head(x, 1)} and \code{tail(x, 1)}, respectively. Consequently, they differ from base R subsetting (\code{x[1]}): they return a zero-length vector for empty inputs (instead of \code{NA}) and do not preserve attributes like names.
+Note: \code{first(x)} and \code{last(x)} differ from base R subsetting (e.g. \code{x[1]}). In its handling of zero-length vectors, \code{first(x)} is designed to behave like \code{head(x, 1)}, returning an empty vector instead of \code{NA}. However, unlike both \code{head(x, 1)} and \code{x[1]}, \code{first(x)} strips attributes like names for performance.
 }
 \value{
 If no other arguments are supplied it depends on the type of \code{x}. The first/last item
--- a/man/last.Rd
+++ b/man/last.Rd
@@ -18,6 +18,9 @@ of \code{xts::first} is deployed. }
 \item{\dots}{ Not applicable for \code{data.table} first/last. Any arguments here
 are passed through to \code{xts}'s first/last. }
 }
+\details{
+Note: \code{first(x)} is not an exact replacement for \code{x[1]}. It differs in its handling of zero-length vectors (returning a zero-length vector instead of \code{NA}) and does not preserve attributes like names.
+}
 \value{
 If no other arguments are supplied it depends on the type of \code{x}. The first/last item
 of a vector or list. The first/last row of a \code{data.frame} or \code{data.table}.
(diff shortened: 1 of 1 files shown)


man/last.Rd
```diff
@@ -18,6 +18,9 @@ of \code{xts::first} is deployed. }
 \item{\dots}{ Not applicable for \code{data.table} first/last. Any arguments here
 are passed through to \code{xts}'s first/last. }
 }
+\details{
+Note: \code{first(x)} and \code{last(x)} are not exact replacements for \code{x[1]} and \code{x[length(x)]}. They differ in handling of zero-length vectors (returning a zero-length vector instead of \code{NA}) and do not preserve attributes like names.
```
tdhock: good!
perhaps mention that this is intentionally similar to head(,1) ?
```r
> x <- integer(0)
> first(x)
integer(0)
> head(x,1)
integer(0)
> x[1]
[1] NA
```

man/last.Rd
```diff
@@ -19,7 +19,7 @@ of \code{xts::first} is deployed. }
 are passed through to \code{xts}'s first/last. }
 }
 \details{
-Note: \code{first(x)} and \code{last(x)} are not exact replacements for \code{x[1]} and \code{x[length(x)]}. They differ in handling of zero-length vectors (returning a zero-length vector instead of \code{NA}) and do not preserve attributes like names.
+Note: \code{first(x)} and \code{last(x)} are designed to behave like \code{head(x, 1)} and \code{tail(x, 1)}, respectively. Consequently, they differ from base R subsetting (\code{x[1]}): they return a zero-length vector for empty inputs (instead of \code{NA}) and do not preserve attributes like names.
```
Anirban166: The way its read can imply that `head()` and `tail()` do not preserve attributes (since the current text emphasizes that `first` and `last` are designed to behave like them), when in fact they do, including names - The analogy holds for zero-length vectors only

Mukulyadav2004: Thanks! Does below this seems right to you? Can you please rephrase it with your suggestions.
```
first(x) and last(x) are not exact replacements for base R subsetting (e.g. x[1]). For zero-length vectors, they are designed to behave like head(x, 1) and tail(x, 1), returning an empty vector instead of NA. However, unlike both head(x, 1) and base subsetting, they do not preserve attributes like names.
```

Anirban166: Factually accurate, so I think what you wrote here sounds good!

I would just make it slightly more concise:
'For zero-length vectors, first(x) and last(x) mimic head(x, 1) and tail(x, 1) by returning an empty vector instead of NA. However, unlike head()/tail() and base R subsetting (e.g., x[1]), they do not preserve attributes like names.'

Anirban166: (Avoid repetition wherever possible)

tdhock:
thanks!

Mukulyadav2004:
Thanks @tdhock for merging the PR! Just wanted to say that I believe this also resolves the documentation point raised in issue #2487 .

man/last.Rd
```diff
@@ -18,6 +18,9 @@ of \code{xts::first} is deployed. }
 \item{\dots}{ Not applicable for \code{data.table} first/last. Any arguments here
 are passed through to \code{xts}'s first/last. }
 }
+\details{
+Note: For zero-length vectors, \code{first(x)} and \code{last(x)} mimic \code{head(x, 1)} and \code{tail(x, 1)} by returning an empty vector instead of \code{NA}. However, unlike \code{head()}/\code{tail()} and base R subsetting (e.g., \code{x[1]}), they do not preserve attributes like names.
```
jangorecki, line 22: there is a Rd keyword `note` that seems more feasible here rather than `details`, considering sentence starts with "Note:".
example: https://github.com/Rdatatable/data.table/blob/ca65511cc0f91c3a73ecf129a4b8e444609a14ac/man/like.Rd#L33

Mukulyadav2004: Thanks! so do I have to submit a new PR?

jangorecki: Yes

PR #7059, opened 2025-06-11
Update README.md

kbodwin:
Fix donate link


MichaelChirico (approved):

jangorecki:
what about size of the numfocus logo on r-datatable.com? can we have it the same size as visible on github.com?

TysonStanley:
Yeah, that is huge on r-datatable.com. Wonder why it isn't using the properties in the README (e.g., `height = '60px'`). After playing around with it, if we use something like `width = '30%'` that may be processed better by `pkgdown` with its default properties.

Issue #6974, opened 2025-05-13
`fread()` does not report `""` as `NA` even when specified in `na.strings`


AngelFelizR:
#### Description

When reading a CSV file that contains empty quoted strings (`""`), I expect these values to be parsed as `NA_character_` if explicitly defined in the `na.strings` argument. However, `fread()` does not display the `NA`s accordingly. Instead, the fields are shown as empty strings in the output.

#### Minimal reproducible example

``` r
temp_file <- tempfile(fileext = ".csv")

writeLines(
  text = c(
    '"Date Example","Question 1","Question 2","Site: Country"',
    '"12/5/2012","Yes","Yes","Chile"',
    '"","","","Virgin Islands, British"'
  ),
  con = temp_file
)

# Missing values not shown as NA by default
data.table::fread(temp_file)
#>    Date Example Question 1 Question 2           Site: Country
#>          <char>     <char>     <char>                  <char>
#> 1:    12/5/2012        Yes        Yes                   Chile
#> 2:                                    Virgin Islands, British

# Missing values still not shown as NA, even with na.strings = '""'
data.table::fread(temp_file, na.strings = '""')
#>    Date Example Question 1 Question 2           Site: Country
#>          <char>     <char>     <char>                  <char>
#> 1:    12/5/2012        Yes        Yes                   Chile
#> 2:                                    Virgin Islands, British

unlink(temp_file)

sessionInfo()
#> R version 4.2.3 (2023-03-15 ucrt)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 26100)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.utf8 
#> [2] LC_CTYPE=English_United States.utf8   
#> [3] LC_MONETARY=English_United States.utf8
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.utf8    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] digest_0.6.31      withr_2.5.2        R.methodsS3_1.8.2  lifecycle_1.0.4   
#>  [5] magrittr_2.0.3     reprex_2.0.2       evaluate_0.23      rlang_1.1.3       
#>  [9] cli_3.6.2          data.table_1.17.99 rstudioapi_0.15.0  fs_1.6.3          
[...]

PR #7068, opened 2025-06-16
fread: consider quoted `na.strings` in text columns

aitap:
Previously, `Field()` only called `end_NA_string()` for non-quoted fields, making it impossible to set `na.strings='""'` and parse empty quoted strings as missing.

Fixes: #6974

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,6 +42,8 @@
 
 11. Out of sample type bumps now respect `integer64=` selection, [#7032](https://github.com/Rdatatable/data.table/pull/7032).
 
+12. `fread()` now handles the `na.strings` argument for quoted text columns, making it possible to specify `na.strings = '""'` and read empty quoted strings as `NA`s, [#6974](https://github.com/Rdatatable/data.table/issues/6974). Thanks to @AngelFelizR for the report and @aitap for the PR.
+
 ### NOTES
 
 1. Continued work to remove non-API C functions, [#6180](https://github.com/Rdatatable/data.table/issues/6180). Thanks Ivan Krylov for the PRs and for writing a clear and concise guide about the R API: https://aitap.codeberg.page/R-api/.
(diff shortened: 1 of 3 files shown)


MichaelChirico:
I haven't quite wrapped my head around this, but it seems like it's preferable not to need the quotes inside `na.strings`? That `na.strings` should apply to the "contents" of each field, where "contents" is applied after handling quotes?

E.g. in this file:

```
"foo","bar","baz",
"a","b","c"
```

All of the string fields are clearly (to us humans) quoted, therefore the "contents" on the first line are `foo`, `bar`, `baz`, which is what we should compare against `na.strings`.

Similarly for OP's example:

```
"Date Example","Question 1","Question 2","Site: Country"
"12/5/2012","Yes","Yes","Chile"
"","","","Virgin Islands, British"
```

I would expect `na.strings=""` to match the first 3 entries on the 3rd line.

I think there are a few related issues, e.g. #2586?

aitap:
The documented behaviour for `na.strings` seems to consider them raw, as they are present in the file:

> When they occur in the file, the strings in `na.strings` should not appear quoted since that is how the string literal `,"NA",` is distinguished from `,NA,`, for example, when `na.strings="NA"`.

Currently, `na.strings = ""` only considers a field missing if there's nothing except whitespace between the field separators. This is done deliberately to make it possible to make ` ` missing while reading `""` as empty string, except in this case the user wants `""` to be `NA` as well.

It seems that according to the current rules, a quoted empty string should be marked as missing using `na.strings = '""'`, except this is not implemented for strings. Non-strings do already work like this:

```
> fread('foo\n42\n0\n"missing"')
       foo
    <char>
1:      42
2:       0
3: missing
> fread('foo\n42\n0\n"missing"', na.strings='missing')
       foo
    <char>
1:      42
2:       0
3: missing
> fread('foo\n42\n0\n"missing"', na.strings='"missing"')
     foo
   <int>
1:    42
2:     0
3:    NA
```

MichaelChirico:
Thanks... yea, I guess we can imagine a world where we only attempt to disambiguate `` from `""` when both are present, but this change is consistent with how things are now.

PR #7072, opened 2025-06-17
Fix: prevent frank() from mutating non-data.table inputs by deep copying atomic and list objects

### LINKED ISSUE #5617: frank modifies non-data.table inputs as a side effect
[Here](https://github.com/Rdatatable/data.table/blob/f84fb935114124748c08d6786012e0dc6435421d/R/frank.R#L26-L27) frank function creates a shallow copy before calling `setDT`. `setDT` is works as intended to change the input by reference but since the input is a shallow copy of frank's input, if it is a list, the components are altered by the function.

# [Minimal reproducible example](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example)
```r
# a list with named vectors as components
ls = list(a = c(a = 1, b = 2),
          b = c(a = 2, b = 3))

frank(ls)

ls # altered list that lost the names of the vector due to setDT
```
```
$a
[1] 1 2

$b
[1] 2 3
```

# Output of sessionInfo()

R version 4.2.2 Patched (2022-11-10 r83330)
Running under: Ubuntu 20.04.2 LTS

Matrix products: default

locale:
 [1] LC_CTYPE=en_US.UTF-8      
 [2] LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8   
 [6] LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8      
 [8] LC_NAME=C                 
 [9] LC_ADDRESS=C              
[10] LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8
[12] LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils    
[5] datasets  methods   base     

other attached packages:
[1] data.table_1.14.9

[1] compiler_4.2.2  cli_3.6.0      
[3] tools_4.2.2     rstudioapi_0.14
[5] rlang_1.0.6

venom1204:
closes #5617

This PR resolves the long-standing issue where frank() modified non-data.table inputs as a side effect — particularly named atomic vectors and lists with named components

Changes made:

Atomic inputs (e.g., named vectors, factors):
- Now explicitly deep copied using copy(x) before converting to list form with as_list(x) to avoid modifying the original input in-place.

List inputs (e.g., list(a = c(a = 1, b = 2), ...)) that are not data.frames:
- Are now deep copied before further processing, ensuring that internal modifications (like setDT) do not mutate the original list or strip names from its components.

Benefits:
- Keeps names and attributes on vectors and lists intact.

Hi @MichaelChirico, @tdhock, @joshhwuu, @jangorecki  — please take a look when you have time. Thanks!

### DIFF
--- a/R/frank.R
+++ b/R/frank.R
@@ -16,11 +16,14 @@ frankv = function(x, cols=seq_along(x), order=1L, na.last=TRUE, ties.method=c("a
     if (!missing(cols) && !is.null(cols))
       stopf("x is a single vector, non-NULL 'cols' doesn't make sense")
     cols = 1L
+    x = copy(x)  # Deep copy atomic vectors
     x = as_list(x)
   } else {
     cols = colnamesInt(x, cols, check_dups=TRUE)
     if (!length(cols))
       stopf("x is a list, 'cols' can not be 0-length")
+    if (!is.data.frame(x))
+      x = copy(x)
   }
   # need to unlock for #4429
   x = .shallow(x, cols, unlock = TRUE) # shallow copy even if list..
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21167,3 +21167,38 @@ test(2317.6, DT1[DF1, on='a', .(d = x.a + i.d)]$d, 5)
 test(2317.7, DT1[DF2, on='a', e := i.e]$e, 5)
 test(2317.8, DT1[DF2, on='a', e2 := x.a + i.e]$e2, 6)
 test(2317.9, DT1[DF2, on='a', .(e = x.a + i.e)]$e, 6)
+
+# frank() should not strip names from named vector input
+x = c(a = "a", b = "b", c = "a")
+test(2318.01, { frank(x); names(x) }, c("a", "b", "c"))
+
+# frank() should not modify names inside a list of named vectors
+ls = list(a = c(a = 1, b = 2), b = c(a = 2, b = 3))
+frank(ls)
+test(2318.02, names(ls[[1]]), c("a", "b"))
+test(2318.03, names(ls[[2]]), c("a", "b"))
+
+# frank() should not mutate original list (deep equality)
+ls = list(a = c(a = 10, b = 20), b = c(a = 30, b = 40))
+original_ls = list(a = c(a = 10, b = 20), b = c(a = 30, b = 40))
+frank(ls)
+test(2318.04, identical(ls, original_ls), TRUE)
+
+# frank() should not mutate list of unnamed vectors (control)
+ls = list(a = c(1, 2), b = c(2, 3))
+original_ls = list(a = c(1, 2), b = c(2, 3))
+frank(ls)
+test(2318.05, identical(ls, original_ls), TRUE)
+
+# frank() should return correct ranks on data.table column without side effect
+DT = data.table(x = c(3, 2, 1, 3))
+original_x = DT$x
+ranks = frank(DT$x)
+test(2318.06, ranks, c(3.5, 2, 1, 3.5))
+test(2318.07, identical(DT$x, original_x), TRUE)
+
+# frank() should preserve names on factors
+x = factor(c("low", "medium", "high"), levels = c("low", "medium", "high"))
+names(x) = c("a", "b", "c")
+frank(x)
+test(2318.08, names(x), c("a", "b", "c"))


jangorecki:
Would be good to see worst case scenario benchmark so we can see how severe can performance implications be.
I wonder if it is not more correct to look at the code, where names are being lost, and try to fix it there, rather than copy() upfront.

tdhock:
would be nice to create an atime performance test case so we can see if there are any performance differences.

venom1204:
> Would be good to see worst case scenario benchmark so we can see how severe can performance implications be. I wonder if it is not more correct to look at the code, where names are being lost, and try to fix it there, rather than copy() upfront.

Following the suggestion, I've run the benchmark to evaluate the performance impact of the change.
The tests confirm that for worst-case inputs like large lists, there is a performance drop. The fixed version is approximately 1.6 to 1.8 times slower than the baseline.

joshhwuu:
Could you include the benchmarking script? Does the performance drop happen with both data.table and non-data.table inputs?

venom1204:
> Could you include the benchmarking script? Does the performance drop happen with both data.table and non-data.table inputs?

Here is the script I used for the following results. It tests both a list and a data.table as input to show the specific impact of the fix.
```R
library(data.table)
library(microbenchmark)

N <- 1e6
M <- 10

large_ls <- replicate(M, setNames(runif(N), paste0("name_", 1:N)), simplify = FALSE)
names(large_ls) <- paste0("col_", 1:M)

large_dt <- as.data.table(large_ls)

results <- microbenchmark(
  "List_Input"       = frank(large_ls),
  "data.table_Input" = frank(large_dt),
  times = 10,
  unit = "ms"
)

print(results)
```

tdhock:
what versions are you comparing to see the 1.6-1.8x difference? I compared current master with this PR and I see no time difference but I see a difference in memory usage.
image
```r
pkg.edit.fun <- function(old.Package, new.Package, sha, new.pkg.path) {
  }
  pkg_find_replace(
    "DESCRIPTION",
  pkg_find_replace(
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
    Package_regex,
    new.Package_)
  pkg_find_replace(
    file.path("R", "onLoad.R"),
  pkg_find_replace(
    file.path("src", "init.c"),
  # allow compilation on new R versions where 'Calloc' is not defined
  pkg_find_replace(
    file.path("src", "*.c"),
    "\\b(Calloc|Free|Realloc)\\b",
    "R_\\1")
  pkg_find_replace(
    "NAMESPACE",
}

expr_list <- list()
for(suffix in c("dt", "ls")){
  var_name <- paste0("large_", suffix)
  suffix_call <- substitute(atime::atime_versions_exprs(
    "~/R/data.table",
    pkg.edit.fun = pkg.edit.fun,
    expr = data.table::frank(VAR),
    master="f5ef6b8da8bf9d60c0c4ae19ad54e06647229593",
    PR7072="eea8cff7dec9e3197f2d6613ad8cd0642f72e4fe"
  ), list(VAR=as.symbol(var_name)))
  some_expr_list <- eval(suffix_call)
  expr_list[paste(suffix, names(some_expr_list))] <- some_expr_list
}
expr_list

library(data.table)
frank_result <- atime::atime(
  setup={
    M <- 10
    large_ls <- replicate(M, setNames(runif(N), paste0("name_", 1:N)), simplify = FALSE)
    names(large_ls) <- paste0("col_", 1:M)
    large_dt <- as.data.table(large_ls)
  },
  expr.list = expr_list)
plot(frank_result)
```

venom1204:
> what versions are you comparing to see the 1.6-1.8x difference

I was comparing the latest master branch against my PR branch . In my local runs, I observed a time difference for list inputs, with my patched version being slower.

Given that your more robust atime analysis shows no significant timing difference but increased memory usage instead, I’m wondering if there’s something off in my benchmarking script, or perhaps a mistake on my end.

In any case, if there is indeed no performance regression, that is reassuring. Kindly let me know if we can move forward from here, or if there are any further checks you would recommend.

venom1204:
Hi @tdhock, @joshhwuu,

Thanks for investigating the performance impact. Since your benchmarking shows no significant timing difference, but there is a difference in memory usage, I wanted to check in:

What could be the next step for this ?


Happy to make any further changes if needed — just let me know!

Thanks again!

joshhwuu:
> What could be the next step for this ?

I agree with @jangorecki, ideally we go through the code and fix the problem at it's root rather than delegating to `copy()`.

tdhock:
Hi @venom1204 can you please resolve conflicts? or close and open another branch/PR?

venom1204:
Hi @tdhock i will create a new PR,just give me e week, right now I am engaged somewhere will soon work on this issue

Issue #6971, opened 2025-05-13
c89-style data/executable code segregation policy


badasahog:
c89 requires data and executable code to be segregated, similar to assemblers like MASM.

It appears that a lot of code still uses similar patterns (albeit not at the beginning of the function), likely an artifact of an earlier language target before this project was updated to c99.

for example: https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L2241

This can hamper readability, because objects are declared miles away from where they're actually used.

I know some people prefer this older style, so I wanted to get a vote of confidence before I started migrating to c99 syntax.

MichaelChirico:
What is the proposed change, exactly? Personally I find that function body far too long & would break it up into more helpers. Do you have something different in mind?

badasahog:
I was going to move the object declarations generally closer to where they're actually used.

This change would also make it easier to extract stuff into helper functions if we choose to do that

MichaelChirico:
I guess the effort would be better spent jumping straight to designing helpers, but if you want to write a PR I will review :)

PR #7074, opened 2025-06-18
formatting improvements in fread

badasahog:
The last time I tried to do this, I got yelled at for creating conflicts. Hopefully `fread` is less actively worked on, so it won't create a gridlock.

Once this is done, I'll work on #6995 and #6971, as well as the creation of helper functions to reduce the overall complexity.

Open to feedback, of course.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -51,11 +51,11 @@ static bool eol_one_r;  // only true very rarely for \r-only files
 //
 static int quoteRule;
 static const char* const* NAstrings;
-static bool any_number_like_NAstrings=false;
-static bool blank_is_a_NAstring=false;
-static bool stripWhite=true;  // only applies to character columns; numeric fields always stripped
-static bool skipEmptyLines=false;
-static int fill=0;
+static bool any_number_like_NAstrings = false;
+static bool blank_is_a_NAstring = false;
+static bool stripWhite = true;  // only applies to character columns; numeric fields always stripped
+static bool skipEmptyLines = false;
+static int fill = 0;
 static int *dropFill = NULL;
 
 static double NA_FLOAT64;  // takes fread.h:NA_FLOAT64_VALUE
@@ -66,7 +66,7 @@ static void *mmp_copy = NULL;
 static size_t fileSize;
 static int8_t *type = NULL, *tmpType = NULL, *size = NULL;
 static lenOff *colNames = NULL;
-static freadMainArgs args = {0};  // global for use by DTPRINT; static implies ={0} but include the ={0} anyway just in case for valgrind #4639
+static freadMainArgs args = { 0 };  // global for use by DTPRINT; static implies ={ 0 } but include the ={ 0 } anyway just in case for valgrind #4639
 #ifdef __EMSCRIPTEN__
   // Under Wasm we must keep the fd open until we're done with the mmap, due to an Emscripten bug (#5969).
   // TODO(emscripten-core/emscripten#20459): revert this
@@ -74,8 +74,8 @@ static freadMainArgs args = {0};  // global for use by DTPRINT; static implies =
 #endif
 
 // See header for more explanation.
-const char typeName[NUMTYPE][10] = {"drop", "bool8", "bool8", "bool8", "bool8", "bool8", "bool8", "int32", "int64", "float64", "float64", "float64", "int32", "float64", "string"};
-int8_t     typeSize[NUMTYPE]     = { 0,      1,       1,       1,       1,       1,      1,       4,       8,       8,         8,         8,         4,       8       ,  8      };
+const char typeName[NUMTYPE][10] = { "drop", "bool8", "bool8", "bool8", "bool8", "bool8", "bool8", "int32", "int64", "float64", "float64", "float64", "int32", "float64", "string" };
+int8_t     typeSize[NUMTYPE]     = { 0,      1,       1,       1,       1,       1,       1,       4,       8,       8,         8,         8,         4,       8       ,  8      };
 
 // In AIX, NAN and INFINITY don't qualify as constant literals. Refer: PR #3043
 // So we assign them through below init function.
@@ -215,7 +215,7 @@ static const char* strlim(const char *ch, size_t limit) {
   char *ch2 = ptr;
   limit = imin(limit, 500);
   size_t width = 0;
-  while ((*ch>'\r' || (*ch!='\0' && *ch!='\r' && *ch!='\n')) && width++<limit) {
+  while ((*ch > '\r' || (*ch != '\0' && *ch != '\r' && *ch != '\n')) && width++ < limit) {
     *ch2++ = *ch++;
   }
   *ch2 = '\0';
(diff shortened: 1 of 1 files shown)


TysonStanley:
Thanks, @badasahog! Very much appreciate you doing this for fread. Over time we'll be able to expand this to the rest of the code set.

badasahog:
My main concern is if this takes too long to merge. That's the main bottleneck for me.

jangorecki:
Merging PR in this repo used to be 3 weeks to 3 months process for couple years. Considering there is no funding dedicated for reviewing and merging PR the process goes pretty smooth. Of course there is always space for improvement. It is far from how it could look like if there would be a full time dev in the project.

src/fread.c
```diff
@@ -577,11 +568,11 @@ static void Field(FieldParseContext *ctx)
   if (*ch==quote) {   // quote=='\0' (user set quote="") would have returned earlier above in the same branch as quoteRule 3
     ch++;
     skip_white(&ch);
-    *(ctx->ch) = ch;
+    *ctx->ch = ch;
```
MichaelChirico, line 574: is there any harm to the extra `()` here? I would imagine the preprocessor strips away any difference.

knowing it's `*(ctx->ch)` and not `(*ctx)->ch` might require having the operator precedence rules tattooed on the backs of your eyelids 😄

src/fread.c
```diff
@@ -785,7 +773,7 @@ static void parse_double_regular_core(const char **pch, double *target)
     }
     e += Eneg? -E : E;
   }
-  if (e<-350 || e>350) return;
+  if (e < -350 || e > 350) return;
```
MichaelChirico, line 779: this one is particularly welcome 🙏

src/fread.c
```diff
@@ -1511,51 +1495,49 @@ int freadMain(freadMainArgs _args) {
...
     STOP(_("File is encoded in UTF-16, this encoding is not supported by fread(). Please recode the file to UTF-8."));
   }
-  if (eof>sof && (eof[-1]=='\x1A' || eof[-1]=='\0')) {
+  if (eof > sof && (eof[-1] == '\x1A' || eof[-1] == '\0')) {
     char c = eof[-1];
-    while (eof>sof && eof[-1]==c) eof--;
-    if (verbose) DTPRINT(_("  Last byte(s) of input found to be %s and removed.\n"),
-                         c ? "0x1A (Ctrl+Z)" : "0x00 (NUL)");
+    while (eof > sof && eof[-1] == c) eof--;
+    if (verbose) DTPRINT(_("  Last byte(s) of input found to be %s and removed.\n"), c ? "0x1A (Ctrl+Z)" : "0x00 (NUL)");
```
MichaelChirico: why remove the extra line here? is it because the `if()` doesn't use `{ ... }`?

generally I find the

```
"formatting string"
entries
```

style nice for long strings; here, I find it extra clear since we have the ternary entry, IMO slightly harder to read that on a long-and-slightly-cluttered line.

so checking your thoughts here.

(ditto elsewhere we see this type of edit)

src/fread.c
```diff
@@ -1666,215 +1647,216 @@ int freadMain(freadMainArgs _args) {
...
   const char *prevStart = NULL;  // the start of the non-empty line before the first not-ignored row (for warning message later, or taking as column names)
-  int jumpLines = nrowLimit==0 ? 100 : (int)umin(100, nrowLimit);   // how many lines from each jump point to use. If nrows>0 is supplied, nJumps is later set to 1. #4029
-  if (fill==INT_MAX) { // if user provides fill=INT_MAX then full file should be sampled #2727
+  int jumpLines = nrowLimit == 0 ? 100 : (int)umin(100, nrowLimit);   // how many lines from each jump point to use. If nrows>0 is supplied, nJumps is later set to 1. #4029
+  if (fill == INT_MAX) { // if user provides fill=INT_MAX then full file should be sampled #2727
     jumpLines = INT_MAX;
     fill = 1; // set fill to true value to not overallocate
   }
+
   {
```
MichaelChirico, line 1663: maybe just remove the `{` "sectioning" instead to keep the indentation level lower & reduce the diff in this region, WDYT?

Ditto for below ~L2149

src/fread.c
```diff
@@ -2250,43 +2232,43 @@ int freadMain(freadMainArgs _args) {
...
   int64_t quoteRuleBumpedLine = -1;
-  int buffGrown=0;
+  int buffGrown = 0;
   // chunkBytes is the distance between each jump point; it decides the number of jumps
   // We may want each chunk to write to its own page of the final column, hence 1000*maxLen
   // For the 44GB file with 12875 columns, the max line len is 108,497. We may want each chunk to write to its
   // own page (4k) of the final column, hence 1000 rows of the smallest type (4 byte int) is just
   // under 4096 to leave space for R's header + malloc's header.
-  size_t chunkBytes = umax((uint64_t)(1000*meanLineLen), 1ULL/*MB*/ *1024*1024);
+  size_t chunkBytes = umax((uint64_t)(1000 * meanLineLen), 1ULL * 1024 * 1024/*MB*/);
```
MichaelChirico, line 2260: ~pedantic suggestion while we're here 😉~

```suggestion
  size_t chunkBytes = umax((uint64_t)(1000 * meanLineLen), 1ULL * 1024 * 1024/*MiB*/);
```

src/fread.c
```diff
@@ -2745,72 +2727,72 @@ int freadMain(freadMainArgs _args) {
...
   }
 
   if (verbose) {
     DTPRINT("=============================\n"); // # notranslate
-    if (tTot<0.000001) tTot=0.000001;  // to avoid nan% output in some trivially small tests where tot==0.000s
-    DTPRINT(_("%8.3fs (%3.0f%%) Memory map %.3fGB file\n"), tMap-t0, 100.0*(tMap-t0)/tTot, 1.0*fileSize/(1024*1024*1024));
-    DTPRINT(_("%8.3fs (%3.0f%%) sep="), tLayout-tMap, 100.0*(tLayout-tMap)/tTot);
-      DTPRINT(sep=='\t' ? "'\\t'" : (sep=='\n' ? "'\\n'" : "'%c'"), sep); // # notranslate
+    if (tTot < 0.000001) tTot = 0.000001;  // to avoid nan% output in some trivially small tests where tot==0.000s
+    DTPRINT(_("%8.3fs (%3.0f%%) Memory map %.3fGB file\n"), tMap - t0, 100.0 * (tMap - t0) / tTot, 1.0 * fileSize / (1024 * 1024 * 1024));
```
MichaelChirico, line 2784: Similar pedantic suggestion

```suggestion
    DTPRINT(_("%8.3fs (%3.0f%%) Memory map %.3fGiB file\n"), tMap - t0, 100.0 * (tMap - t0) / tTot, 1.0 * fileSize / (1024 * 1024 * 1024));
```

MichaelChirico: Actually, let's just do that separately: #7079

MichaelChirico (commented):
LGTM, thank you!

Main thing is about whether we should remove the precedence-clarifying parentheses in `*(obj->elt)`.

MichaelChirico:
Reviewed in two passes: Once in the plain 'files' pane, then again with whitespace differences ignored:

https://github.com/Rdatatable/data.table/pull/7074/files?w=1

src/fread.c
jangorecki: Yeah, good one! This what Matt always mentioned when asked why he uses = vs <- in R

src/fread.c
badasahog: Maybe I've just been using c for too long, but I think this should feel natural? I can reverse, but I believe this is a common idiom.

src/fread.c
badasahog: the place where the line breaks is inconsistent throughout the code. Is there an agreed upon line wrap point? If there is I can apply that.

src/fread.c
badasahog: generally a good idea to use blocks when possible to reduce the namespace dilution and avoid naming conflicts of shadowing. This will also be useful later once we section things down into smaller functions.

src/fread.c
MichaelChirico: No, nor do I think it's possible to pursue that generally. The only version of that rule that I've seen "enforce-able" automatically is "fill up the line as much as possible to a fixed width and always wrap if and only if the next "word" goes past the width".

IMO, using line breaks for readability is the best practice, though it's the hardest to automate since it's a little bit up to taste. The "maximum fill" rule produces less readable code, IMHO.

src/fread.c
MichaelChirico: OK. Yea, at root, not having such massive function bodies is more important & can be fixed later.

src/fread.c
badasahog: How many characters is the maximum? My monitor is probably not the same as everyone else's

MichaelChirico: Generally we don't adhere to a maximum. Traditionally, Matt wrote in an IDE that did wrapping so he never bothered with it.

Personally, I tend to target ~80 characters, with up to 120 characters if it improves readability (again a matter of taste).

If we were going to enforce something, I would use 120.

badasahog: Done.

src/fread.c
badasahog: Do we have a final decision? I say we go without them, but you're in charge

src/fread.c
```diff
@@ -1514,51 +1498,49 @@ int freadMain(freadMainArgs _args) {
...
     STOP(_("File is encoded in UTF-16, this encoding is not supported by fread(). Please recode the file to UTF-8."));
   }
-  if (eof>sof && (eof[-1]=='\x1A' || eof[-1]=='\0')) {
+  if (eof > sof && (eof[-1] == '\x1A' || eof[-1] == '\0')) {
     char c = eof[-1];
-    while (eof>sof && eof[-1]==c) eof--;
-    if (verbose) DTPRINT(_("  Last byte(s) of input found to be %s and removed.\n"),
-                         c ? "0x1A (Ctrl+Z)" : "0x00 (NUL)");
+    while (eof > sof && eof[-1] == c) eof--;
+    if (verbose) DTPRINT(_("  Last byte(s) of input found to be %s and removed.\n"), c ? "0x1A (Ctrl+Z)" : "0x00 (NUL)");
```
```suggestion
    if (verbose) DTPRINT(_("  Last byte(s) of input found to be %s and removed.\n"),
                         c ? "0x1A (Ctrl+Z)" : "0x00 (NUL)");
```

src/fread.c
```diff
@@ -1620,12 +1602,11 @@ int freadMain(freadMainArgs _args) {
...
                   args.skipString);
-    while (ch>sof && ch[-1]!='\n') ch--;  // move to beginning of line
+    while (ch > sof && ch[-1] != '\n') ch--;  // move to beginning of line
     pos = ch;
     ch = sof;
-    while (ch<pos) row1line+=(*ch++=='\n');
-    if (verbose) DTPRINT(_("Found skip='%s' on line %d. Taking this to be header row or first row of data.\n"),
-                         args.skipString, row1line);
+    while (ch < pos) row1line += (*ch++ == '\n');
+    if (verbose) DTPRINT(_("Found skip='%s' on line %d. Taking this to be header row or first row of data.\n"), args.skipString, row1line);
```
```suggestion
    if (verbose) DTPRINT(_("Found skip='%s' on line %d. Taking this to be header row or first row of data.\n"),
                         args.skipString, row1line);
```

src/fread.c
MichaelChirico: The only place besides fread.c I see something similar is in fmelt.c:

https://github.com/Rdatatable/data.table/blob/d03357f03d8456944e1e7c1fe64eed3b01ca5afa/src/fmelt.c#L510

Where `*...->` is using `*` for multiplication... fun

https://github.com/search?q=repo%3ARdatatable%2Fdata.table%20%2F%5B*%5D%5B(%5D.*-%3E%2F&type=code

Since it's localized to fread.c, I am fine for either way.

PR #7077, opened 2025-06-19
Unify the linters a bit

aitap:
Addressing https://github.com/Rdatatable/data.table/pull/7039#issuecomment-2920661331:

Provide a runner for linters in `.ci/lint.R` that would:
1. `source()` everything in `.ci/linters/${given_subdirectory}/*.R`
2. collect files according to the given mask and optionally preprocess them, which may mean `gcc -E` or skipping some of the files
3. run every function from (1) on every file, collect their errors but let the process continue
4. if caught at least one error, complain and fail the process when done

The individual linters can now focus on their job and just `stop()` when convenient. This is fairly general and could be used for Rd linting in #7075.

The "preprocessing" part prevents the linter calls from being one-liners. Should there be wrapper scripts for every kind of linter? What do you think, @MichaelChirico?

### DIFF
--- a/.ci/lint.R
+++ b/.ci/lint.R
@@ -7,13 +7,12 @@ if (identical(args, '--help')) {
     'KIND must name the directory containing the *.R files defining the linter functions.',
     'WHERE must name the directory containing the files to lint, e.g. "po", or "src".',
     "WHAT must contain the regular expression matching the files to lint, e.g., '[.]po$', or '[.][ch]$'.",
-    'PREPROCESS is an optional R function that accepts a file path and returns an R object accepted by the lint functions, or NULL to skip the file.'
   ))
   q('no')
 }
-stopifnot(`Invalid arguments, see .ci/lint.R --help` = length(args) %in% 3:4)
+stopifnot(`Invalid arguments, see .ci/lint.R --help` = length(args) == 3)
 
-linter_env = new.env()
+linter_env = list2env(list(.preprocess = identity))
 for (f in list.files(args[[1]], full.names=TRUE)) sys.source(f, linter_env)
 if (!length(ls(linter_env))) stop(
   "No linters found after sourcing files in ", dQuote(args[[1]])
@@ -23,8 +22,7 @@ sources = list.files(args[[2]], pattern = args[[3]], full.names = TRUE, recursiv
 if (!length(sources)) stop(
   "No files to lint found in directory ", dQuote(args[[2]]), " for mask ", dQuote(args[[3]])
 )
-preprocess = if (length(args) == 4) eval(str2lang(args[[4]])) else identity
-sources = Filter(Negate(is.null), lapply(setNames(nm = sources), preprocess))
+sources = Filter(Negate(is.null), lapply(setNames(nm = sources), linter_env$.preprocess))
 
 okay = TRUE
 for (src in names(sources))
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -55,25 +55,15 @@ Base R runner for the manual (non-`lintr`) lint checks to be run from GitHub Act
 1. Path to the directory containing files defining the linters. A linter is a function that accepts one argument (typically the path to the file) and signals an error if it fails the lint check.
 2. Path to the directory containing files to check.
 3. A regular expression matching the files to check.
-4. An R expression evaluating to a function that processes a single file path into a form understood by the lint functions. Additionally, the function may return `NULL` to indicate that the file must be skipped.
+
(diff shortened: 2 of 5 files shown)


.ci/lint.R
```diff
@@ -0,0 +1,38 @@
...
+    "WHAT must contain the regular expression matching the files to lint, e.g., '[.]po$', or '[.][ch]$'.",
+    'PREPROCESS is an optional R function that accepts a file path and returns an R object accepted by the lint functions, or NULL to skip the file.'
+  ))
+  q('no')
+}
+stopifnot(`Invalid arguments, see .ci/lint.R --help` = length(args) %in% 3:4)
+
+linter_env = new.env()
+for (f in list.files(args[[1]], full.names=TRUE)) sys.source(f, linter_env)
+if (!length(ls(linter_env))) stop(
```
MichaelChirico, line 17: Just `length(linter_env)`? Or do we intentionally want to exclude private names?

```r
e <- new.env()
length(e)
# [1] 0
length(ls(e))
# [1] 0
e$x = 1
length(e)
# [1] 1
length(ls(e))
# [1] 1
e$.y = 2
length(e)
# [1] 2
length(ls(e))
# [1] 1
```

(e.g. maybe we _do_ want to exclude the `.preprocessor` name discussed below...)

.ci/lint.R
```diff
@@ -0,0 +1,38 @@
...
+  "No linters found after sourcing files in ", dQuote(args[[1]])
+)
+
+sources = list.files(args[[2]], pattern = args[[3]], full.names = TRUE, recursive = TRUE)
+if (!length(sources)) stop(
+  "No files to lint found in directory ", dQuote(args[[2]]), " for mask ", dQuote(args[[3]])
+)
+preprocess = if (length(args) == 4) eval(parse(text = args[[4]])) else identity
```
MichaelChirico: I think it's safe to assume this is run on R>=3.6.0

```suggestion
preprocess = if (length(args) == 4) eval(str2lang(args[[4]])) else identity
```

.ci/linters/md/heading_id_linter.R
```diff
@@ -1,8 +1,8 @@
-any_mismatch = FALSE
-
 # ensure that ids are limited to alphanumerics and dashes
 # (in particular, dots and underscores break the links)
 check_header_ids = function(md) {
+  if (!grepl('[.]Rmd$', md)) return()
```
MichaelChirico:
```suggestion
  if (!grepl('[.]Rmd$', md)) return(invisible())
```

.ci/linters/md/news_linter.R
```diff
@@ -1,13 +1,13 @@
-any_mismatch = FALSE
-
 # ensure that numbered list in each section is in sequence
 check_section_numbering = function(news) {
+  if (!grepl("NEWS", news)) return()
```
MichaelChirico:
```suggestion
  if (!grepl("NEWS", news)) return(invisible())
```

.ci/linters/md/news_linter.R
```diff
@@ -22,11 +22,12 @@ check_section_numbering = function(news) {
       paste0("  [", section_entries[!matched], " --> ", intended_value[!matched], "]", collapse="\n")
     ))
   }
-  return(any_mismatch)
+  stopifnot("Please fix the issues above" = !any_mismatch)
```
MichaelChirico:
```suggestion
  stopifnot("Please fix the NEWS issues above" = !any_mismatch)
```

.ci/linters/md/news_linter.R
```diff
@@ -48,14 +49,5 @@ check_gh_links = function(news) {
...
-  return(TRUE)
-}
-
-any_error = FALSE
-for (news in list.files(pattern = "NEWS")) {
-  cat(sprintf("Checking NEWS file %s...\n", news))
-  news_lines = readLines(news)
-  any_error = check_section_numbering(news_lines) || any_error
-  any_error = check_gh_links(news_lines) || any_error
+  stop("Please fix the issues above.")
```
MichaelChirico:
```suggestion
  stop("Please fix the NEWS issues above.")
```

.ci/lint.R
aitap: This does feel a bit like how R packages worked before namespaces. Running everything from the environment would be simpler, cleaner design, but common helper functions that start with a dot could be convenient for the linter writer.

The current scripts run `for (linter in ls(linter_env))`. Let's also use this for preprocessing.

Issue #7082, opened 2025-06-22
Vignette build fails with: there is no package called 'litedown'


jangorecki:
win-dev
mac-rel
https://rdatatable.gitlab.io/data.table/web/checks/check_results_data.table.html

aitap:
The older CRAN snapshots in the cache do work ([source](https://gitlab.com/Rdatatable/data.table/-/jobs/10429872550/artifacts/browse/bus/mirror-packages/cran/src/contrib/), [win-rel](https://gitlab.com/Rdatatable/data.table/-/jobs/10429872550/artifacts/browse/bus/mirror-packages/cran/bin/windows/contrib/4.5/)), while the newer versions do not ([win-dev](https://gitlab.com/Rdatatable/data.table/-/jobs/10429872550/artifacts/browse/bus/mirror-packages/cran/bin/windows/contrib/4.6/)). `litedown` should be installed because `dcf.dependencies('DESCRIPTION', which='all')` include `markdown`, `markdown` `Imports: litedown (>= 0.6)`, and `litedown_0.7` is present in the cache. And yet it's not installed:

```
> install.packages('markdown', lib=l1, repos='file:z:/dta/bus/mirror-packages/cran')
also installing the dependencies 'commonmark', 'xfun'

package 'commonmark' successfully unpacked and MD5 sums checked
package 'xfun' successfully unpacked and MD5 sums checked
package 'markdown' successfully unpacked and MD5 sums checked
> packageVersion('markdown')
[1] '2.0' # ← 2.0 means that the newer binary (not the older source) was installed
```

It looks like `install.packages()` uses `available.packages()` somewhere with the default `type = getOption('pkgType')` argument, i.e. `both`, which (somehow) gives dependency information about the _source_ packages. But then `install.packages()` interprets `type = "both"` as `win.binary`, and binaries are installed according to wrong (source) dependency information. I suppose it hasn't been encountered in the wild because source packages and binaries typically have matching versions.

To work around the problem:

1. Specify the type explicitly so that `install.packages` uses the right dependency information:
```
> install.packages('markdown', lib=l1, repos='file:z:/dta/bus/mirror-packages/cran', type = 'win.binary')
also installing the dependencies 'commonmark', 'xfun', 'litedown'

package 'commonmark' successfully unpacked and MD5 sums checked
package 'xfun' successfully unpacked and MD5 sums checked
package 'litedown' successfully unpacked and MD5 sums checked # ← !
[...]

PR #7089, opened 2025-06-24
`mirror.packages`: download newer package versions

aitap:
Only skip downloading a package if it already has the right version and MD5sum (in case of binary rebuilds). When downloading a new version of the package, remove the old one.

This should help ensure that the source packages mirrored from CRAN are as new as binaries (or newer), which `install.packages(type = "both")` assumes when resolving dependencies.

The updated [`mirror_packages` step](https://gitlab.com/Rdatatable/data.table/-/jobs/10448007607) of [the pipeline](https://gitlab.com/Rdatatable/data.table/-/pipelines/1886467568) downloads a new version of `evaluate` and keeps the rest of the packages intact; the Windows and macOS jobs pass their checks.

Fixes: #7082


jangorecki (approved):
LGTM

PR #7097, opened 2025-06-26
Add error handling for gzipped file decompression and external command failures in fread()

### LINKED ISSUE #5415: feature request: fread should raise an error instead of a warning when reading a gzipped file that does not fit in temporary storage
Hi,

Thanks for developing this awesome package!

fread has supported gzipped files for a while now and it usually works great
#717

Recently, I have been working with containers more often. These often have limited temporary storage. When reading a gzipped file with fread, I believe it first unzips it to temporary storage, and then reads the file from there into memory (I am aware of the `tmpdir` argument that controls this). If the file is very large, the temporary storage can fill up. fread, instead of raising an error, reads the truncated file and prints a cryptic warning (telling you that the file was truncated). This warning is easily overlooked when working non-interactively.

I think it would be better if instead of reading a truncated file, fread would actually raise an error in these cases. The error could then be caught early on. Only reading part of the data often leads to downstream errors, which then harder to de-bug. I don't think many users want to read just the first x lines that happen to fit into their temporary storage at random.

A flag could still force reading a truncated gzipped file, if necessary.

thank you for your time

Mukulyadav2004:
fixes #5415 

This PR enhances `fread()`'s error handling to provide actionable error messages when system commands fail or file decompression encounters issues (like disk full scenarios). Previously, these failures could result in warnings or silent truncation.

The current implementation of `fread()` has two silent failure modes that can lead to data corruption and hard-to-debug issues:

- **Gzipped file decompression failures:** When `R.utils::decompressFile()` fails due to insufficient disk space or other issues, the function silently continues with a truncated file, leading to partial data reads and warnings that are easily missed in non-interactive sessions.

- **External command failures:** When using the `cmd` parameter, failed external commands (non-zero exit codes) are not detected, potentially leading to processing of empty or incomplete files.

**Solution-**
- As `R.utils::decompressFile()` is called without error handling, allowing silent failures so I wrapped `R.utils::decompressFile()` in `tryCatch()` to catch decompression failures and added comprehensive error message that explains the likely cause (disk full) and mention disc space and `tmpdir` argument.

- And System command exit codes are not checked thus missing command failures, so to fix - Captured the return status from `system()` command and added exit code validation with error message.

- Also in both modifications - moved `on.exit(unlink(tmpFile), add=TRUE)` before command execution to prevent storage leaks.

@tdhock @jangorecki @Anirban166 @joshhwuu can you please review.

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -70,9 +70,13 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     }
   }
   if (!is.null(cmd)) {
-    (if (.Platform$OS.type == "unix") system else shell)(paste0('(', cmd, ') > ', tmpFile<-tempfile(tmpdir=tmpdir)))
+    tmpFile = tempfile(tmpdir=tmpdir)
+    on.exit(unlink(tmpFile), add=TRUE)  
+    status = suppressWarnings((if (.Platform$OS.type == "unix") system else shell)(paste0('(', cmd, ') > ', tmpFile)))
+    if (status != 0) {
+      stopf("External command failed with exit code %d. This can happen when the disk is full in the temporary directory ('%s'). See ?fread for the tmpdir argument.", status, tmpdir)
+    }
     file = tmpFile
-    on.exit(unlink(tmpFile), add=TRUE)
   }
   if (!is.null(file)) {
     if (!is.character(file) || length(file)!=1L)
@@ -116,9 +120,14 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
       if (!requireNamespace("R.utils", quietly = TRUE))
         stopf("To read %s files directly, fread() requires 'R.utils' package which cannot be found. Please install 'R.utils' using 'install.packages('R.utils')'.", if (w<=2L || gzsig) "gz" else "bz2") # nocov
       FUN = if (w<=2L || gzsig) gzfile else bzfile
-      R.utils::decompressFile(file, decompFile<-tempfile(tmpdir=tmpdir), ext=NULL, FUN=FUN, remove=FALSE)   # ext is not used by decompressFile when destname is supplied, but isn't optional
-      file = decompFile   # don't use 'tmpFile' symbol again, as tmpFile might be the http://domain.org/file.csv.gz download
-      on.exit(unlink(decompFile), add=TRUE)
+      decompFile = tempfile(tmpdir=tmpdir)
+      on.exit(unlink(decompFile), add=TRUE)  
+      tryCatch({
+        R.utils::decompressFile(file, decompFile, ext=NULL, FUN=FUN, remove=FALSE)
+      }, error = function(e) {
+        stopf("Failed to decompress file '%s'. This can happen when the disk is full in the temporary directory ('%s'). See ?fread for the tmpdir argument.", file, tmpdir)
+      })
+      file = decompFile
     }
     file = enc2native(file) # CfreadR cannot handle UTF-8 if that is not the native encoding, see #3078.
 


R/fread.R
```diff
@@ -70,9 +70,13 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     }
   }
   if (!is.null(cmd)) {
-    (if (.Platform$OS.type == "unix") system else shell)(paste0('(', cmd, ') > ', tmpFile<-tempfile(tmpdir=tmpdir)))
+    tmpFile = tempfile(tmpdir=tmpdir)
+    on.exit(unlink(tmpFile), add=TRUE)  
+    status = suppressWarnings((if (.Platform$OS.type == "unix") system else shell)(paste0('(', cmd, ') > ', tmpFile)))
```
jangorecki: Printing warning can be still good, just before raising error, then more info will be provided to user

R/fread.R
```diff
@@ -70,9 +70,13 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
...
+    if (status != 0) {
+      stopf("External command failed with exit code %d. This can happen when the disk is full in the temporary directory ('%s'). See ?fread for the tmpdir argument.", status, tmpdir)
```
jangorecki, line 77: nocov market here will be needed

R/fread.R
```diff
@@ -116,9 +120,14 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
       if (!requireNamespace("R.utils", quietly = TRUE))
         stopf("To read %s files directly, fread() requires 'R.utils' package which cannot be found. Please install 'R.utils' using 'install.packages('R.utils')'.", if (w<=2L || gzsig) "gz" else "bz2") # nocov
       FUN = if (w<=2L || gzsig) gzfile else bzfile
-      R.utils::decompressFile(file, decompFile<-tempfile(tmpdir=tmpdir), ext=NULL, FUN=FUN, remove=FALSE)   # ext is not used by decompressFile when destname is supplied, but isn't optional
-      file = decompFile   # don't use 'tmpFile' symbol again, as tmpFile might be the http://domain.org/file.csv.gz download
-      on.exit(unlink(decompFile), add=TRUE)
+      decompFile = tempfile(tmpdir=tmpdir)
+      on.exit(unlink(decompFile), add=TRUE)  
+      tryCatch({
+        R.utils::decompressFile(file, decompFile, ext=NULL, FUN=FUN, remove=FALSE)
```
jangorecki, line 126: Does this empty line has any purpose here?

R/fread.R
Mukulyadav2004: Can you please tell do I need to add this to decompress part also? And by adding this do I still need to add tests to test the changes?

R/fread.R
```diff
@@ -70,9 +70,13 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     }
   }
   if (!is.null(cmd)) {
-    (if (.Platform$OS.type == "unix") system else shell)(paste0('(', cmd, ') > ', tmpFile<-tempfile(tmpdir=tmpdir)))
+    tmpFile = tempfile(tmpdir=tmpdir)
+    on.exit(unlink(tmpFile), add=TRUE)
```
MichaelChirico:
```suggestion
    on.exit(unlink(tmpFile), add=TRUE)
```

R/fread.R
```diff
@@ -116,9 +120,14 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
       if (!requireNamespace("R.utils", quietly = TRUE))
         stopf("To read %s files directly, fread() requires 'R.utils' package which cannot be found. Please install 'R.utils' using 'install.packages('R.utils')'.", if (w<=2L || gzsig) "gz" else "bz2") # nocov
       FUN = if (w<=2L || gzsig) gzfile else bzfile
-      R.utils::decompressFile(file, decompFile<-tempfile(tmpdir=tmpdir), ext=NULL, FUN=FUN, remove=FALSE)   # ext is not used by decompressFile when destname is supplied, but isn't optional
-      file = decompFile   # don't use 'tmpFile' symbol again, as tmpFile might be the http://domain.org/file.csv.gz download
-      on.exit(unlink(decompFile), add=TRUE)
+      decompFile = tempfile(tmpdir=tmpdir)
+      on.exit(unlink(decompFile), add=TRUE)
```
MichaelChirico:
```suggestion
      on.exit(unlink(decompFile), add=TRUE)
```

R/fread.R
```diff
@@ -116,9 +120,14 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
       if (!requireNamespace("R.utils", quietly = TRUE))
         stopf("To read %s files directly, fread() requires 'R.utils' package which cannot be found. Please install 'R.utils' using 'install.packages('R.utils')'.", if (w<=2L || gzsig) "gz" else "bz2") # nocov
       FUN = if (w<=2L || gzsig) gzfile else bzfile
-      R.utils::decompressFile(file, decompFile<-tempfile(tmpdir=tmpdir), ext=NULL, FUN=FUN, remove=FALSE)   # ext is not used by decompressFile when destname is supplied, but isn't optional
-      file = decompFile   # don't use 'tmpFile' symbol again, as tmpFile might be the http://domain.org/file.csv.gz download
```
MichaelChirico, line 120: why drop the comments?

R/fread.R
```diff
@@ -116,9 +120,14 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
...
+      decompFile = tempfile(tmpdir=tmpdir)
       on.exit(unlink(decompFile), add=TRUE)
+      tryCatch({
+        R.utils::decompressFile(file, decompFile, ext=NULL, FUN=FUN, remove=FALSE)
+      }, error = function(e) {
+        stopf("Failed to decompress file '%s'. This can happen when the disk is full in the temporary directory ('%s'). See ?fread for the tmpdir argument.", file, tmpdir)
```
MichaelChirico: We should also report the error that {R.utils} does, maybe?

```suggestion
        stopf("R.utils::decompressFile failed to decompress file '%s':\n  %s\n. This can happen when the disk is full in the temporary directory ('%s'). See ?fread for the tmpdir argument.", file, conditionMessage(e), tmpdir)
```

> See ?fread for the tmpdir argument.

R/fread.R
MichaelChirico: We add `# nocov` if _and only if_ it is prohibitively difficult (most typically, impossible) to test a change in a CI setting.

There are ways to mock up the CI, e.g. "hiding" resources from the runner, but IMO that counts as "prohibitively difficult" / not worth it.

We _could_ (should?) also just have a test like `fread(cmd = 'not a command')` just to get coverage here, but it won't actually test the original issue you're fixing. It's not a totally useless test, just not covering the actual intent of this branch.

Similarly below, we _could_ (should?) test it by making a faked compressed file with the expected '.gz' extension that will error out decompressFile(), which again won't test anything about resource pressure.

Mukulyadav2004: Thanks for clarification. I added tests: fread(cmd = 'not a command') for external command failures and a fake .gz file for decompression failures. And should I proceed using # nocov for the decompression test as appears little complex to implement.

MichaelChirico:
I still see the tests failing, PTAL & ping if you're stuck

Mukulyadav2004:
Thanks! I tried making the file in 2nd test more corrupted cuz might be the corrupted gzip file not corrupted enough. But its still failing, can you please review the second test.

MichaelChirico:
> corrupted gzip file not corrupted enough

Yea, I was also having some trouble generating an error when I was playing around. @HenrikBengtsson if you have a second could you suggest a way to quickly generate a file that will break `R.utils::decompressFile()`? Don't spend more than a few seconds -- we'll just `#nocov` if it's not something readily apparent. Glancing at the implementation, most of the logic is owned by R anyway.

MichaelChirico:
Oh, I managed to do it:

```r
tmp <- tempfile()
file.create(tmp)
conn <- file(tmp, 'wb')
# this is data.table:::known_signatures$gzip
writeBin(as.raw(c(31, 139)), conn)
close(conn)

fread(tmp)
# Error in readBin(inn, what = raw(0L), size = 1L, n = BFR.SIZE) : 
#   error reading from the connection
# In addition: Warning message:
# In readBin(inn, what = raw(0L), size = 1L, n = BFR.SIZE) :
#   invalid or incomplete compressed data
```

Mukulyadav2004:
Thank you for the working example. I've updated the test to use your approach

inst/tests/tests.Rraw
```diff
@@ -21192,3 +21192,16 @@ test(2319.1, !is.null(attr(dt_get, ".internal.selfref")))
...
 test(2319.2, !is.null(attr(dt_get0, ".internal.selfref")))
+
+# Improved fread error handling for cmd exe and decompression #5415
+test(2320.1, fread(cmd="false"), error="External command failed with exit code") # external command execution error handling
+
+if (test_R.utils) {
+  tmp <- tempfile(fileext=".csv.gz")
+  file.create(tmp)
+  conn <- file(tmp, 'wb')
+  writeBin(as.raw(c(31, 139)), conn)
```
MichaelChirico, line 21203: one last thing: please comment about these seemingly arbitrary numbers for future readers

inst/tests/tests.Rraw
```diff
@@ -21192,3 +21192,16 @@ test(2319.1, !is.null(attr(dt_get, ".internal.selfref")))
...
+# Improved fread error handling for cmd exe and decompression #5415
+test(2320.1, fread(cmd="false"), error="External command failed with exit code") # external command execution error handling
+
+if (test_R.utils) {
+  tmp <- tempfile(fileext=".csv.gz")
+  file.create(tmp)
+  conn <- file(tmp, 'wb')
+  writeBin(as.raw(c(31, 139)), conn) # Gzip header starts with magic numbers: 0x1F 0x8B (decimal 31, 139) - these identify the file as gzip
+  close(conn)
+  test(2320.2, fread(tmp), error="R.utils::decompressFile failed to decompress") # Compressed file decompression error handling
```
MichaelChirico, line 21205: looks like my example throws 2 conditions -- warning, then error. We can:

 - Add `withCallingHandlers()` to the implementation to also capture warnings
 - Error on warning by including it in `tryCatch()`
 - Leave the implementation unchanged and only catch the warning in `test()`

I don't have a strong opinion -- if I picked one, it would probably be #3.

Mukulyadav2004: I'm not sure which one is most worthy but I believe third one is quick to implement. WDYT?

MichaelChirico:
Slightly worried the warnings/errors we observe could be more platform-dependent than we've detected here. Let's see if anything comes out of the more thorough GLCI suite.

Issue #7099, opened 2025-06-26
tests on R 3.4 raises extra warning


jangorecki:
https://rdatatable.gitlab.io/data.table/web/checks/data.table/test-lin-ancient-cran/00check.log

latest master

```
* checking tests ... OK
  Running ‘autoprint.R’
  Comparing ‘autoprint.Rout’ to ‘autoprint.Rout.save’ ...17,20d16
<        a
<    <int>
< 1:     1
< 2:     3
27,30d22
<        a
<    <int>
< 1:     1
< 2:     3
32,36d23
< Index: <a>
<        a
<    <int>
< 1:     1
< 2:     3
38,42d24
< Index: <a>
<        a
<    <int>
< 1:     1
< 2:     3
60,63d41
<        a
<    <int>
< 1:     1
< 2:     5
65,68d42
<        a
<    <int>
< 1:     1
< 2:     6
72,75d45
<        a
<    <int>
< 1:     1
< 2:     5
142,144d111
<        a
<    <int>
< 1:     9
150,153d116
<        a
<    <int>
< 1:    10
< 2:    10
155,158d117
<        a
<    <int>
< 1:    10
< 2:    10
160,163d118
<        a
<    <int>
< 1:    10
< 2:    10
209,214d163
<        x     y
<    <num> <num>
< 1:     1     1
< Warning message:
< In `[.data.table`(dt, , `:=`(y, 1)) :
<   A shallow copy of this data.table was taken so that := can add or remove 1 columns by reference. At an earlier point, this data.table was copied by R (or was created manually using structure() or similar). Avoid names<- and attr<- which in R currently (and oddly) may copy the whole data.table. Use set* syntax instead to avoid copying: ?set, ?setnames and ?setattr. It's also not unusual for data.table-agnostic packages to produce tables affected by this issue. If this message doesn't help, please report your use case to the data.table issue tracker so the root cause can be fixed or this message improved.
```

MichaelChirico:
First guess is it's due to #6631. Unfortunately codespaces is not cooperating so I won't be able to install 3.4 to debug further.

I guess here:

```
dt = data.table(x = 1)
class(dt) = c("foo", "data.table", "data.frame")
dt[, y := 1]
```

`class<-` has created the shallow copy -- isn't this warning expected in that case? Possibly we should just ignore this issue as WAI.

jangorecki:
[...]

PR #7102, opened 2025-06-26
Use setattr to avoid shallow copy in autoprint test setup

MichaelChirico:
Part of #7099

### DIFF
--- a/tests/autoprint.Rout.save
+++ b/tests/autoprint.Rout.save
@@ -172,7 +172,7 @@ NULL
 > 
 > # child class of data.table doesn't induce unintended print, #3029
 > dt = data.table(x = 1)
-> class(dt) = c("foo", "data.table", "data.frame")
+> setattr(dt, "class", c("foo", "data.table", "data.frame"))
 > print.foo = function(x, ...) {
 +   NextMethod("print")
 + }


jangorecki:
```sh
git push gl print-setattr
```

MichaelChirico:
Is that right? I pushed the branch and even opened an MR but don't see any pipeline

https://gitlab.com/Rdatatable/data.table/-/merge_requests/4

image

jangorecki:
MR should not be needed...

MichaelChirico:
I saw you created something, is it something we have to do manually?

 1. Push to GL
 2. Go to pipelines flow & create from branch

jangorecki:
I run CI for this branch manually via https://gitlab.com/Rdatatable/data.table/-/pipelines/new
but in general it should start itself, just after push to branch

jangorecki:
Two of dependencies could not be installed in this job
```
Warning message:
packages ‘xts’, ‘knitr’ are not available (for R version 3.4.0) 
```
warning is gone but problem with matching output is still there, maybe for 3.4.0 we need different mockup file.
https://gitlab.com/Rdatatable/data.table/-/jobs/10484872441

https://gitlab.com/Rdatatable/data.table/-/jobs/10484872441/artifacts/browse/bus/test-lin-ancient-cran/data.table.Rcheck/tests/

jangorecki:
Looks like
```r
DT[2,a:=3L]
```
(amongst others) prints in 3.4.0

MichaelChirico:
Thanks. Looks like this is an improvement, but doesn't fix the issue. I think worth merging anyway.

jangorecki:
yes, worth merging. Assuming R 3.5.0 is not being affected I think we could ignore the printing issue for R 3.4.0. At least till nobody reported that till now.

PR #7108, opened 2025-06-30
Add optional= argument to as.data.frame.ITime to interact better with data.frame()

### LINKED ISSUE #4673: create data.frame with ITime will lose the colname

MichaelChirico:
Closes #4673. Supersedes #5301.

I'm somewhat confident we don't need this method anymore, but keeping it here precautiously.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -70,6 +70,8 @@
 
 14. `data.table()` function is now more aligned with `data.frame()` with respect to the names of the output when one of its inputs is a single-column matrix object, [#4124](https://github.com/Rdatatable/data.table/issues/4124). Thanks @PavoDive for the report and @jangorecki for the PR.
 
+15. Including an `ITime` object as a named input to `data.frame()` respects the provided name, i.e. `data.frame(a = as.ITime(...))` will have column `a`, [#4673](https://github.com/Rdatatable/data.table/issues/4673). Thanks @shrektan for the report and @MichaelChirico for the fix.
+
 ### NOTES
 
 1. Continued work to remove non-API C functions, [#6180](https://github.com/Rdatatable/data.table/issues/6180). Thanks Ivan Krylov for the PRs and for writing a clear and concise guide about the R API: https://aitap.codeberg.page/R-api/.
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -70,6 +70,8 @@
 
 14. `data.table()` function is now more aligned with `data.frame()` with respect to the names of the output when one of its inputs is a single-column matrix object, [#4124](https://github.com/Rdatatable/data.table/issues/4124). Thanks @PavoDive for the report and @jangorecki for the PR.
 
+15. Including an `ITime` object as a named input to `data.frame()` respects the provided name, i.e. `data.frame(a = as.ITime(...))` will have column `a`, [#4673](https://github.com/Rdatatable/data.table/issues/4673). Thanks @shrektan for the report and @MichaelChirico for the fix.
```
jangorecki, line 73: if that's the breaking change then worth to mention how to retain previous behavior

inst/tests/tests.Rraw
```diff
@@ -21246,3 +21246,7 @@ test(2322.12, levels(fctr(c("b","a","c"), rev=NA)), error="TRUE or FALSE")
 test(2322.21, levels(fctr(c("b","a","c"), sort=TRUE)), c("a","b","c"))
 test(2322.22, levels(fctr(c("b","a","c"), sort=NA)), error="TRUE or FALSE")
 test(2322.31, levels(fctr(c("b","a","c"), rev=TRUE, sort=TRUE)), c("c","b","a"))
+
+# data.frame() uses provided names of ITime inputs
+test(2323.1, names(data.frame(COL = as.ITime('00:00:00'))), "COL")
+test(2323.2, names(data.frame(b = 1, COL = as.ITime('00:00:00'))), c("b", "COL"))
```
jangorecki: maybe one tests using `optional` arg?

NEWS.md
MichaelChirico: I wouldn't call it a breaking change to rely on an undocumented bug (we violated the data.frame() contract as I see it). let's see if revdep shows anything.

jangorecki (approved):

PR #7110, opened 2025-06-30
Suppressed spurious min/max warnings in cube() internal structure-seeding eval()

### LINKED ISSUE #6964: Strange warning about no non-missing arguments in min or max when used within cube
Consider the following minimal reproducible example:

```
library(data.table)
SimData <- data.table(var = rep(c("a", "b", "c"), each = 3),
                      value = 1:9)
cube(SimData, .(min(value)), "var")
```

This gives the following warning: `In min(value) : no non-missing arguments to min; returning Inf`.

I don't understand this warning: there are no combinations for which there are no non-missing arguments! There are non-missing arguments both for each `var` and globally. The result actually reflects this (no `Inf` appears), so I really don't understand the message. (The same happens with `max`.)

`#` `Output of sessionInfo()`

`R version 4.5.0 (2025-04-11 ucrt) Platform: x86_64-w64-mingw32/x64 Running under: Windows 10 x64 (build 19045)  Matrix products: default   LAPACK version 3.12.1  locale: [1] LC_COLLATE=Hungarian_Hungary.utf8  LC_CTYPE=Hungarian_Hungary.utf8    LC_MONETARY=Hungarian_Hungary.utf8 [4] LC_NUMERIC=C                       LC_TIME=Hungarian_Hungary.utf8      time zone: Europe/Budapest tzcode source: internal  attached base packages: [1] stats     graphics  grDevices utils     datasets  methods   base       other attached packages: [1] data.table_1.17.0  loaded via a namespace (and not attached): [1] compiler_4.5.0    tools_4.5.0       rstudioapi_0.17.1`

venom1204:
Closes #6964


R/groupingsets.R
```diff
@@ -112,7 +112,14 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, labe
     .SDcols = if (".SD" %chin% av) setdiff(names(x), by) else NULL
   if (length(names(by))) by = unname(by)
   # 0 rows template data.table to keep colorder and type
-  empty = if (length(.SDcols)) x[0L, eval(jj), by, .SDcols=.SDcols] else x[0L, eval(jj), by]
+  empty = withCallingHandlers(
+  if (length(.SDcols)) x[0L, eval(jj), by, .SDcols=.SDcols] else x[0L, eval(jj), by],
+  simpleWarning = function(w) {
+    if (grepl("no non-missing arguments to min|max", w$message, ignore.case = TRUE)) {
```
MichaelChirico: This assumes the warning message is in English:

```r
LANGUAGE=zh_CN Rscript -e "min(integer())"
# [1] Inf
# 警告信息:
# In min(integer()) : min里所有的参数都不存在; 返回Inf
```

Anyway, it's too specific -- the solution here should not be unique to `min()`/`max()`, but _any_ warning raised by code we evaluate on user's behalf.

venom1204: Thanks for the helpful feedback — you're absolutely right about the language dependency and the need for a more general solution.
Should I go ahead and replace the current handler with a tightly scoped suppressWarnings()? It would avoid locale issues and handle all similar cases.

MichaelChirico: I suspect that's the best option, yes

jangorecki:
I belive merge conflicts have not been properly resolved. Last merge to master seem to lost recent changes and it does suppresses all warnings now.

If the change was intentional it should not be named as a merge branch to master, but another commit.

R/groupingsets.R
jangorecki: If we go with suppressing all warnings then it should be documented because it is not what user can expect.

venom1204: i updated it kindly have a look
thanks.

MichaelChirico: > If we go with suppressing all warnings then it should be documented because it is not what user can expect.

I still don't understand why we'd do that -- this `eval()` is one we generate internally:

https://github.com/Rdatatable/data.table/blob/ae2b8155d1fca6ba62f317ee3273dfd4d02bdde0/R/groupingsets.R#L116-L118

https://github.com/Rdatatable/data.table/blob/ae2b8155d1fca6ba62f317ee3273dfd4d02bdde0/R/groupingsets.R#L178-L181

Why would the user want to know about that?

jangorecki: AFAIU, because if user will construct those grouping calls, in this particular case including those that have 0 elements in a group, manually then user would get those warnings.
There might be various kind of warnings in UDF which we cannot predict, not only those related to grouping combinations that have no values in the data. Imagine
```r
sum_at_least_2 = function (x) {
  if (length(x)<2) warning("less elements then expected")
  sum(x)
}
```
Then user naturally will expect possible warnings.

jangorecki: Or calling `log` on a data which will happen to have negative values, warning will be lost where it would be quiet useful

jangorecki:
github. com/Rdatatable/data.table/pull/ 7110/commits/807b0607e0383b535b471ae637599508bac90a66

Please copy paste as hot link on github redirects it to another commit diff. This is, I believe, result of a force push.

This commit has change to suppress warnings and as well has merging to master changes in NEWS file

venom1204:
https://github.com/Rdatatable/data.table/commit/0f1b434ea48d1fef713731486da21cef58ce6543

This commit contains only the intended changes: suppression of spurious warnings in cube() internal eval, and a corresponding NEWS.md entry.

jangorecki:
The commit I was referring to disappeared, due to force push, it is best to avoid those.

Anyway, the question about suppressing all warnings stays valid, if we do so it has to be very well documented.

Consider the following case, different than min/max where warning comes from 0 length input for a group combinations that are not existing for `a * b` product, below we have all combinations for `a * b`, and seeing this warning is very much desired:
```r
dt = data.table(a=1:2, b=1:2, v=1:4-2)
rollup(dt, log(v), by=c("a","b"))
```

R/groupingsets.R
MichaelChirico: > Or calling `log` on a data which will happen to have negative values

But such warning will still be surfaced, e.g. using the `DT` from `?cube`:

```r
DT[, value := -value]
# running this on current 'master'
nrow(cube(DT, j = sum(log(value)), by = c("color","year"), id=TRUE))
# [1] 21
# There were 21 warnings (use warnings() to see them)
```

Note that in fact we get one warning for each output row, which means there's _not_ the warning suppressed in this PR -- because `log(double())` has no warning.

MichaelChirico: > Then user naturally will expect possible warnings.

I'm still not convinced. Sure, if user knows some groups have 1-2 rows, they will expect such warnings. And they'll still get them!

But there is no guarantee they'll see the warnings when constructing the equivalent manually -- because presumably, they know the output structure, they don't need to do the structure-seeding query. We only need to do this as an implementation detail in order to support more generic queries.

MichaelChirico:
> The commit I was referring to disappeared, due to force push, it is best to avoid those.
> 
> Anyway, the question about suppressing all warnings stays valid, if we do so it has to be very well documented.
> 
> Consider the following case, different than min/max where warning comes from 0 length input for a group combinations that are not existing for `a * b` product, below we have all combinations for `a * b`, and seeing this warning is very much desired:
> 
> ```r
> dt = data.table(a=1:2, b=1:2, v=1:4-2)
> rollup(dt, log(v), by=c("a","b"))
> ```

Here again, this PR doesn't change the surfaced warnings, of which there are 3, one for each of the sets `c("a", "b")`, `"a"`, and `character()`. The `dt[0]` seeding query again does not generate a warning.

R/groupingsets.R
jangorecki: I looked now more carefully at the code and realized it does suppress warnings only for a template row. From the PR description I understood it just simply suppresses all the warnings in j.
Then of course issue about it is not existing. Sorry for a false alarm.

NEWS.md
```diff
@@ -78,6 +78,8 @@
 
 17. A data.table with a column of class `vctrs_list_of` (from package {vctrs}) prints as expected, [#5948](https://github.com/Rdatatable/data.table/issues/5948). Before, they could be printed messily, e.g. printing every entry in a nested data.frame. Thanks @jesse-smith for the report, @DavisVaughan and @r2evans for contributing, and @MichaelChirico for the PR.
 
+18. `cube()`, `rollup()`, and `groupingsets()` no longer produce a spurious warning when using `min()` or `max()` as aggregations, [#6964](https://github.com/Rdatatable/data.table/issues/6964). Thanks @ferenci-tamas for the report and @venom1204 for the fix.
```
```suggestion
18. Spurious warnings from internal code in `cube()`, `rollup()`, and `groupingsets()` are no longer surfaced to the caller, [#6964](https://github.com/Rdatatable/data.table/issues/6964). Thanks @ferenci-tamas for the report and @venom1204 for the fix.
```

R/groupingsets.R
```diff
@@ -115,7 +115,7 @@ groupingsets.data.table = function(x, j, by, sets, .SDcols, id = FALSE, jj, labe
   # inline all arguments that might clash with enclosing environment
   pcall = substitute(x[0L, jj, by], list(x = x, jj = jj, by = by))
   if (length(.SDcols)) pcall$.SDcols = .SDcols
-  empty = eval(pcall, list(.datatable.aware = TRUE), enclos)
+  empty = suppressWarnings(eval(pcall, list(.datatable.aware = TRUE), enclos))
```
MichaelChirico, line 119:
```suggestion
  # suppress e.g. the min(double()) warning, #6964
  empty = suppressWarnings(eval(pcall, list(.datatable.aware = TRUE), enclos))
```

MichaelChirico (approved):
Thanks!

R/groupingsets.R
MichaelChirico: OK, great. I had a go at amending the working of the PR title, description and NEWS entry to make all that more obvious.

inst/tests/tests.Rraw
```diff
@@ -21303,3 +21303,21 @@ local({
   test(2326.2, key(d), "x")
   test(2326.3, indices(d), c("y", "z"))
 })
+
+# Test for spurious min/max warnings in cube(), #6964
```
MichaelChirico: I think the tests here are overkill. Note that we don't need separate tests for "no warning" and "correct value" -- `test()` will report an unchecked warning as needed, e.g. `test(1, {warning('hi'); 1}, 1)` will fail.

I think we are fine with just two tests -- the rest will be covered by existing tests of `cube()`:

 - Test that there's no warning from `min()` with simple data
 - Test that there _is_ a warning from `min()` in data where one would be produced "naturall", e.g. `min(x, na.rm=TRUE)` including some grouping with no non-`NA` values.

venom1204: i updated test kindly have a look when you have time 
thanks

MichaelChirico (approved):

PR #6111, opened 2024-04-30
Forder segfault

ben-schwen:
Closes #4300

- [x] fix 
- [ ] tests
- [ ] news

PR #7113, opened 2025-07-01
made todo labels consistent

badasahog:
I think todo labels are the thing you're you're most likely to search code for, so making them all appear the same would seem pertinant.

It was very inconsistent, so I settled on `todo:`


MichaelChirico:
Related: #678

badasahog:
@MichaelChirico this should make that work, right? or do you prefer a different syntax?

MichaelChirico:
We can worry about the exact syntax later -- once we have something consistent, at least, a find-and-replace version becomes a lot easier.

That said, this PR will also sit idle for a while in the spirit of avoiding merge conflicts 🫤

MichaelChirico:
Here are the unique, open, non-draft PRs touching files also touched by this PR: #6965, #6640, #6595, #6111, #5907, #5898, #5896, #5895, #5893, #5892, #5890, #5692, #5682, #5576, #5575, #5269, #5168, #5167, #4978, #4585, #4372, #4370

PR #7121, opened 2025-07-02
removing temporary *ch object

badasahog:
I need feedback on this;

I think using a temporary, which gets reassigned to the original is bad practice, but using the double dereference is ugly, too.

In addition, requiring *ch to be copied back to **pch could be dangerous, and cause bugs if the function returns somewhere you didn't expect.

It may also be reasonable to take in a char*, move it, then return it to avoid a double pointer.

I want to know what you guys think is appropriate before I start renovating all the relevant functions.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -242,13 +242,11 @@ static char *typesAsString(int ncol) {
 static inline void skip_white(const char **pch) {
   // skip space so long as sep isn't space and skip tab so long as sep isn't tab
   // always skip any \0 (NUL) that occur before end of file, #3400
-  const char *ch = *pch;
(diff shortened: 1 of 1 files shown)


badasahog:
@MichaelChirico can you let me know?

MichaelChirico:
I really don't know how to weigh in here, sorry. Let's let this simmer a bit longer.

badasahog:
@aitap can you weigh in?

aitap:
Let's try `diff`ing the assembly output from compiling two different versions of the code? Seeing the assembly change (or not) will be instructive.

Issue #7079, opened 2025-06-20
Use *iB for power-of-2 bytes units consistently


MichaelChirico:
Split off from #7074. I believe @aitap did a related fix recently (not easy to search for it). But we should try and be consistent about: 1024 bytes: kibibyte=KiB, _not_ KB (which is 1000 bytes), and so on.

Examples found quickly:

https://github.com/Rdatatable/data.table/blob/0afb20f1a9e6720c66c7b0d45c983fe4d273345d/src/fread.c#L2797
https://github.com/Rdatatable/data.table/blob/0afb20f1a9e6720c66c7b0d45c983fe4d273345d/src/fwrite.c#L794

More to search through (lots of false positives since GH search is not case-sensitive, but a local `grep` will do better)

https://github.com/search?q=repo%3ARdatatable%2Fdata.table+%2F%5Bkmg%5Db+%2F+-path%3A%2F%5C.po%2F&type=code

MichaelChirico:
Two particularly problematic things:

The use of `buffMB` as a parameter in `fwrite()`:

https://github.com/Rdatatable/data.table/blob/4773ef9d7d15c6d2611ec892aa1caa1b0a8e37e0/man/fwrite.Rd#L17

The use of `mb` as a parameter and output column name `MB` in `tables()`:

https://github.com/Rdatatable/data.table/blob/4773ef9d7d15c6d2611ec892aa1caa1b0a8e37e0/man/tables.Rd#L12

I'm not sure it's worth a breaking change for renaming any of those, but something to consider. We can at least correctly describe whether they actually refer to `MB` or `MiB` (IINM it's always actually `MiB`).

PR #7126, opened 2025-07-04
Use IEC memory unit abbreviations

MichaelChirico:
For #7079

I spot-checked most of these, but was not fully precise for numbers in comments where the 2% difference is not meaningful.

Here's the regex I applied: `[^A-Z][KMGTP]B($|[^\w-])`. I excluded these files (glob): `*{.po,NEWS,rchk}*`.

We could probably set this up as a `code-quality` check

What's left are all `buffMB` or `tables()`-related. Whether to proceed with a breaking change there should be addressed separately.

Some external references:

 - [`man du`](https://man7.org/linux/man-pages/man1/du.1.html)
 - [{BSgenome.Hsapiens.UCSC.hg19}](https://www.bioconductor.org/packages/release/data/annotation/html/BSgenome.Hsapiens.UCSC.hg19.html)
 - [`_R_CHECK_PKG_SIZES_THRESHOLD_`](https://github.com/r-devel/r-svn/blob/96c3f260416e9754145d304a6a4ab38f1d80f5fa/doc/manual/R-ints.texi#L4008-L4010)
 - [`man ps`](https://man7.org/linux/man-pages/man1/ps.1.html) (as returned by internal helper `rss()`)
 - [`?gc`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/gc.html) (in "R legacy units")

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -4,7 +4,7 @@ Title: Extension of `data.frame`
 Depends: R (>= 3.4.0)
 Imports: methods
 Suggests: bit64 (>= 4.0.0), bit (>= 4.0.4), R.utils, xts, zoo (>= 1.8-1), yaml, knitr, markdown
-Description: Fast aggregation of large data (e.g. 100GB in RAM), fast ordered joins, fast add/modify/delete of columns by group using no copies at all, list columns, friendly and fast character-separated-value read/write. Offers a natural and flexible syntax, for faster development.
+Description: Fast aggregation of large data (e.g. 100GiB in RAM), fast ordered joins, fast add/modify/delete of columns by group using no copies at all, list columns, friendly and fast character-separated-value read/write. Offers a natural and flexible syntax, for faster development.
 License: MPL-2.0 | file LICENSE
 URL: https://r-datatable.com, https://Rdatatable.gitlab.io/data.table, https://github.com/Rdatatable/data.table
 BugReports: https://github.com/Rdatatable/data.table/issues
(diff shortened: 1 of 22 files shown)


DESCRIPTION
```diff
@@ -4,7 +4,7 @@ Title: Extension of `data.frame`
 Depends: R (>= 3.4.0)
 Imports: methods
 Suggests: bit64 (>= 4.0.0), bit (>= 4.0.4), R.utils, xts, zoo (>= 1.8-1), yaml, knitr, markdown
-Description: Fast aggregation of large data (e.g. 100GB in RAM), fast ordered joins, fast add/modify/delete of columns by group using no copies at all, list columns, friendly and fast character-separated-value read/write. Offers a natural and flexible syntax, for faster development.
+Description: Fast aggregation of large data (e.g. 100GiB in RAM), fast ordered joins, fast add/modify/delete of columns by group using no copies at all, list columns, friendly and fast character-separated-value read/write. Offers a natural and flexible syntax, for faster development.
```
jangorecki: Oh, it feels too much, is it so much significant difference that we need to update description that was there for a decade(?) (or maybe even more?)

jangorecki: To be fair, I believe for most readers GiB will be a bit confusing comparing to simply GB.

MichaelChirico: funny use of "bit" :joy:

11 years: c2c8ed4936d89d1ac1d1656c6fa5cca565f15f7d

You're right this doesn't really matter (both 100GB and 100GiB are "large data"), this change is more with the thinking to enforce using IEC units everywhere.

But, for now at least, we'll already need carve-outs for `fwrite(buffMB=)` and `tables(mb=)`, so one more small carve-out makes no difference. Reverting.

jangorecki (approved):

PR #7127, opened 2025-07-04
Avoid backticks in .Rd files where \code was intended

### LINKED ISSUE #7087: New code quality check: don't use markdown '`' notation in Rd files
Something that's tough -- remembering that `` ` `` is not supported like markdown in .Rd files. In fact, that should be a second Rd linter, but let's mark that for follow-up. It's not trivial since we have to exclude example code (and other common sources of false positives).

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/pull/7075#discussion_r2162630833_

MichaelChirico:
Closes #7087 by including a new Rd linter for preventing backsliding.


man/IDateTime.Rd
```diff
@@ -217,7 +217,7 @@ accounting for both year transitions and varying day counts per week.
 }
 \references{
 
-  G. Grothendieck and T. Petzoldt, ``Date and Time Classes in R,''
+  G. Grothendieck and T. Petzoldt, "Date and Time Classes in R",
```
aitap: This is a TeX idiom for typographically appropriate double quotes, Rd has `\dQuote{}` for it (and `\sQuote{...}` for `` `single-quoted' `` text).
```suggestion
  G. Grothendieck and T. Petzoldt, \dQuote{Date and Time Classes in R},
```

man/assign.Rd
```diff
@@ -75,7 +75,7 @@ For additional resources, please read \href{../doc/datatable-faq.html}{\code{vig
...
+Unlike \code{<-} for \code{data.frame}, the (potentially large) LHS is not coerced to match the type of the (often small) RHS. Instead the RHS is coerced to match the type of the LHS, if necessary. Where this involves double precision values being coerced to an integer column, a warning is given when fractional data is truncated. It is best to get the column types correct up front and stick to them. Changing a column type is possible but deliberately harder: provide a whole column as the RHS. This RHS is then \emph{plonked} into that column slot and we call this \emph{plonk syntax}, or \emph{replace column syntax} if you prefer. By needing to construct a full length vector of a new type, you as the user are more aware of what is happening and it is clearer to readers of your code that you really do intend to change the column type; e.g., \code{DT[, colA:=as.integer(colA)]}. A plonk occurs whenever you provide a RHS value to \code{:=} which is \code{nrow} long. When a column is \emph{plonked}, the original column is not updated by reference because that would entail updating every single element of that column whereas the plonk is just one column pointer update.
```
aitap: `\code{...}` is specifically for "R-like" text that could go through `parse(text = ...)` ([parseRd.pdf, page 7](https://developer.r-project.org/parseRd.pdf)). For pieces of code that cannot be parsed by R as-is, use [`\samp`](https://cran.r-project.org/doc/manuals/R-exts.html#index-_005csamp) if it's acceptable to wrap and reformat the text fragment or [`\verb`](https://cran.r-project.org/doc/manuals/R-exts.html#index-_005cverb) if it's not.
```suggestion
Unlike \samp{<-} for \code{data.frame}, the (potentially large) LHS is not coerced to match the type of the (often small) RHS. Instead the RHS is coerced to match the type of the LHS, if necessary. Where this involves double precision values being coerced to an integer column, a warning is given when fractional data is truncated. It is best to get the column types correct up front and stick to them. Changing a column type is possible but deliberately harder: provide a whole column as the RHS. This RHS is then \emph{plonked} into that column slot and we call this \emph{plonk syntax}, or \emph{replace column syntax} if you prefer. By needing to construct a full length vector of a new type, you as the user are more aware of what is happening and it is clearer to readers of your code that you really do intend to change the column type; e.g., \code{DT[, colA:=as.integer(colA)]}. A plonk occurs whenever you provide a RHS value to \samp{:=} which is \code{nrow} long. When a column is \emph{plonked}, the original column is not updated by reference because that would entail updating every single element of that column whereas the plonk is just one column pointer update.
```

man/setattr.Rd
```diff
@@ -24,7 +24,7 @@ setnames(x, old, new, skip_absent=FALSE)
 
   \code{setattr} is a more general function that allows setting of any attribute to an object \emph{by reference}.
 
-  A very welcome change in R 3.1+ was that `names<-` and `colnames<-` no longer copy the \emph{entire} object as they used to (up to 4 times), see examples below. They now take a shallow copy. The `set*` functions in data.table are still useful because they don't even take a shallow copy. This allows changing names and attributes of a (usually very large) \code{data.table} in the global environment \emph{from within functions}. Like a database.
+  A very welcome change in R 3.1+ was that \code{\link[base]{names<-}} and \code{\link[base]{colnames<-}} no longer copy the \emph{entire} object as they used to (up to 4 times), see examples below. They now take a shallow copy. The \code{set*} functions in data.table are still useful because they don't even take a shallow copy. This allows changing names and attributes of a (usually very large) \code{data.table} in the global environment \emph{from within functions}. Like a database.
```
aitap:
```suggestion
  A very welcome change in R 3.1+ was that \code{\link[base]{names<-}} and \code{\link[base]{colnames<-}} no longer copy the \emph{entire} object as they used to (up to 4 times), see examples below. They now take a shallow copy. The \samp{set*} functions in data.table are still useful because they don't even take a shallow copy. This allows changing names and attributes of a (usually very large) \code{data.table} in the global environment \emph{from within functions}. Like a database.
```

PR #7128, opened 2025-07-04
Exclude '..1' from up-a-level search

### LINKED ISSUE #5460: Using ..1 with data.table gives an error
The following results in an error:

```
BODdt <- data.table(BOD)
BODdt[, X := sapply(demand, function(...) sqrt(..1))]
## Error in `[.data.table`(BODdt, , `:=`(X, sapply(demand, function(...) sqrt(..1)))) : 
##   Variable '1' is not found in calling scope. Looking in calling scope because this symbol was prefixed with .. in the j= parameter.
```

If we use dot dot dot instead of ..1 then there is no error so it seems it is specifically related to the use of ..1 .

```
BODdt <- data.table(BOD)
BODdt[, X := sapply(demand, function(...) sqrt(...))]
# no error
```

Also if we hide the ..1 in the body of a function defined externally it produces no error.

```
SQRT <- function(...) sqrt(..1)
BODdt <- data.table(BOD)
BODdt[, X := sapply(demand, SQRT)]
```

MichaelChirico:
Closes #5460


tdhock:
lgtm

PR #7130, opened 2025-07-04
Fix: Ensure column widths are data-driven when col.names = "none" in print.data.table

### LINKED ISSUE #6882: print col.names='none' does not affect column width

venom1204:
closes #6882 

Hi @MichaelChirico,
I've now implemented the suggestions provided by Mukul — thank you for pointing them out. They were on my mind earlier as well, but I had initially planned to incorporate them alongside some broader changes. Apologies for the delay, and also for creating a new PR for the same issue due to some complications.

Would appreciate it if you could take a moment to review the updates when convenient. Thanks again!


R/print.data.table.R
```diff
@@ -141,6 +141,8 @@ print.data.table = function(x, topn=getOption("datatable.print.topn"),
     print_default(toprint)
     return(invisible(x))
   }
+  if (col.names == "none")
+  colnames(toprint) = rep.int("", ncol(toprint))
```
MichaelChirico:
```suggestion
    colnames(toprint) = rep.int("", ncol(toprint))
```

MichaelChirico:
I think it looks good, thanks! Please add a NEWS entry

venom1204:
> I think it looks good, thanks! Please add a NEWS entry

done.👍

Issue #6968, opened 2025-05-13
Inconsistent naming structure


badasahog:
The names of identifiers often use different formats, like camel case / snake case / dromedary case.

Is there a consensus on how things should be named?

In addition, should I fix it? I know people can get irritated when identifiers change under their feet.

MichaelChirico:
in user-facing code we shouldn't change anything. often (e.g. with fread parameters) it's done for consistency with base equivalents (read.table), and anyway I don't find it worth breaking peoples' code over.

in purely internal code I'm ok with some minor refactoring to use snake_case more consistently

jangorecki:
Conclusion in the past to apply style to newly written code but do not change the old code for this purpose.

badasahog:
what about internal identifiers that are just a single a few letters long?

PR #7112, opened 2025-07-01
Third round of formatting improvements in fread

badasahog:
One of these days I'll catch em all

As an aside, which other files should I do this for next? I'm not aware of which ones have outstanding PRs.

PR #7135, opened 2025-07-07
Updated formatting

badasahog:
Hopefully no conflicts?

Once this is merged I'll get more aggressive with #6968 and #6971

### DIFF
--- a/src/fwrite.c
+++ b/src/fwrite.c
@@ -791,7 +791,7 @@ void fwriteMain(fwriteMainArgs args)
   }
   char *buffPool = malloc(alloc_size);
   if (!buffPool) {
-    STOP(_("Unable to allocate %zu MB * %d thread buffers; '%d: %s'. Please read ?fwrite for nThread, buffMB and verbose options."), // # nocov
+    STOP(_("Unable to allocate %zu MiB * %d thread buffers; '%d: %s'. Please read ?fwrite for nThread, buffMB and verbose options."), // # nocov
          buffSize / MEGA, nth, errno, strerror(errno)); // # nocov
   }
 


src/fwrite.c
```diff
@@ -226,59 +226,59 @@ void writeFloat64(const void *col, int64_t row, char **pch)
...
+        else { dl0 = 1; width = dr + 1 + dl0; }            // 0.1234, 0.0001234
       }
       // So:  3.1416 => l=31416, sf=5, exp=0     dr=4; dl0=0; width=6
       //      30460  => l=3046, sf=4, exp=4      dr=0; dl0=1; width=5
       //      0.0072 => l=72, sf=2, exp=-3       dr=4; dl0=1; width=6
-      if (width <= sf + (sf>1) + 2 + (abs(exp)>99?3:2) + scipen) {
-        //               ^^^^ to not include 1 char for dec in -7e-04 where sf==1
-        //                       ^ 2 for 'e+'/'e-'
+      if (width <= sf + (sf > 1) + 2 + (abs(exp) > 99 ? 3 : 2) + scipen) {
+        //                   ^^^^ to not include 1 char for dec in -7e-04 where sf==1
```
MichaelChirico, line 264:
```suggestion
        //               ^^^^^^ to not include 1 char for dec in -7e-04 where sf==1
```

MichaelChirico:
```suggestion
        //               ^^^^^^ to not include 1 char for dec in -7e-04 where sf==1
```

src/fwrite.c
```diff
@@ -581,12 +581,12 @@ int compressbuff(z_stream *stream, void* dest, size_t *destLen, const void* sour
 {
   stream->next_out = dest;
   stream->avail_out = *destLen;
-  stream->next_in = (Bytef *)source; // don't use z_const anywhere; #3939
+  stream->next_in = (Bytef*)source; // don't use z_const anywhere; #3939
   stream->avail_in = sourceLen;
   int err = deflate(stream, Z_SYNC_FLUSH);
-  *destLen = *destLen - stream->avail_out;
+  *destLen -= stream->avail_out;
```
MichaelChirico, line 587: nice!

src/fwrite.c
```diff
@@ -791,7 +791,7 @@ void fwriteMain(fwriteMainArgs args)
   }
   char *buffPool = malloc(alloc_size);
   if (!buffPool) {
-    STOP(_("Unable to allocate %zu MiB * %d thread buffers; '%d: %s'. Please read ?fwrite for nThread, buffMB and verbose options."), // # nocov
+    STOP(_("Unable to allocate %zu MB * %d thread buffers; '%d: %s'. Please read ?fwrite for nThread, buffMB and verbose options."), // # nocov
```
MichaelChirico: `MiB` is correct, I guess this is just a faulty merge to `master`:

https://github.com/Rdatatable/data.table/blob/c446f2e2ba5f9b2c5712e88441ffd7d8a838f90d/src/fwrite.c#L29

MichaelChirico:
```suggestion
    STOP(_("Unable to allocate %zu MiB * %d thread buffers; '%d: %s'. Please read ?fwrite for nThread, buffMB and verbose options."), // # nocov
```

MichaelChirico (approved):
Thanks! What was done this round to avoid the need for follow-up PRs? It makes the `git` history cleaner if we can avoid multiple "style-only" edits

src/fwrite.c
badasahog: sorry about that. did you change it or should I?

badasahog:
@MichaelChirico I thought it was better to separate style out. I can add more in on top, in future PRs. Up to you.

src/fwrite.c
MichaelChirico: Yes, it's done (FYI GitHub writes `Outdated` on comments where the underlying source is already changed)

MichaelChirico:
What I have in mind is #7106 / #7112, where we wound up with 3 commits doing ~the same formatting-only changes. Can we avoid that by using tooling to identify the changes needed, or otherwise taking another pass at the file now before merging?

badasahog:
I think this one is simpler, so that won't be needed.

I'll make sure to include other improvements with future PRs

PR #7136, opened 2025-07-07
Row Name Extraction for data.table() with keep.rownames

### LINKED ISSUE #1916: data.table(keep.rownames = TRUE) should preserve names from vectors
`data.frame`s will form row names from the names of vector inputs.

```{r}
data.frame(
  x = setNames((1:12) ^ 3, month.abb)
)
```

I expected `data.table` to do something similar when `keep.rownames = TRUE`.  That is I expected that these two pieces of code would be equivalent:

```{r}
data.table(
  x = setNames((1:12) ^ 3, month.abb),
  keep.rownames = TRUE
)

data.table(
  rn = month.abb,
  x = (1:12) ^ 3
)
```

Currently `keep.rownames = TRUE` seems to only apply to matrix or data.frame inputs, rather than vectors.  Changing the behavior so that `keep.rownames = TRUE` mimics `data.frame` behavior seems desirable.

Mukulyadav2004:
fixes #1916 
Implements row name extraction from named vectors in `data.table()` and `as.data.table()` calls when `keep.rownames=TRUE` or `keep.rownames="column_name"`. This matches the behavior of `data.frame()` and `as.data.frame.list()` by extracting names from the first named atomic vector in the input.

Problem : Currently, `data.table()` and `as.data.table()` do not extract row names from named vectors like `data.frame()` . 

So added logic to `as.data.table.list()` (the common path for both `data.table()` and `as.data.table()`) to:
- Iterate through input vectors to locate the first atomic vector with valid names.
- Store the names and remove them from the source vector to avoid duplication.
- Prepend row names as a new column with name "rn" or custom name.

@MichaelChirico @tdhock can you please review.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -277,5 +277,14 @@ test.list <- atime::atime_test_list(
     Slow = "73d79edf8ff8c55163e90631072192301056e336",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8397dc3c993b61a07a81c786ca68c22bc589befc)
     Fast = "8397dc3c993b61a07a81c786ca68c22bc589befc"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7019/commits) that removes inefficiency
 
+  "isoweek improved in #7144" = atime::atime_test(
+    setup = {
+      set.seed(349)
+      x = sample(Sys.Date() - 0:5000, N, replace=TRUE)
+    },
+    expr = data.table::isoweek(x),
+    Slow = "548410d23dd74b625e8ea9aeb1a5d2e9dddd2927",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/548410d23dd74b625e8ea9aeb1a5d2e9dddd2927)
+    Fast = "c0b32a60466bed0e63420ec105bc75c34590865e"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7144/commits) that uses a much faster implementation
+
     tests=extra.test.list)
 # nolint end: undesirable_operator_linter.
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,7 +42,9 @@
 
 8. `groupingsets()` gets a new argument `enclos` for use together with the `jj` argument in functions wrapping `groupingsets()`, including the existing wrappers `rollup()` and `cube()`, [#5560](https://github.com/Rdatatable/data.table/issues/5560). When forwarding a `j`-expression as `groupingsets(jj = substitute(j))`, make sure to pass `enclos = parent.frame()` as well, so that the `j`-expression will be evaluated in the right context. This makes it possible for `j` to refer to variables outside the `data.table`. Thanks @sindribaldur for the report and @aitap for the fix.
 
-9. `data.table()` and `as.data.table()` with `keep.rownames=TRUE` now extract row names from named vectors, matching `data.frame()` behavior. Names from the first named vector in the input are used to create the row names column (default name `"rn"` or custom name via `keep.rownames="column_name"`), [#1916](https://github.com/Rdatatable/data.table/issues/1916). Thanks to @richierocks for the feature request and @Mukulyadav2004 for the implementation.
+9. `isoweek()` is much faster (e.g. 20x) by re-using an implementation from {base}, [#5111](https://github.com/Rdatatable/data.table/issues/5111). Thanks @MichaelChirico for the report and PR.
+
+10. `data.table()` and `as.data.table()` with `keep.rownames=TRUE` now extract row names from named vectors, matching `data.frame()` behavior. Names from the first named vector in the input are used to create the row names column (default name `"rn"` or custom name via `keep.rownames="column_name"`), [#1916](https://github.com/Rdatatable/data.table/issues/1916). Thanks to @richierocks for the feature request and @Mukulyadav2004 for the implementation.
 
 ### BUG FIXES
 
--- a/R/IDateTime.R
+++ b/R/IDateTime.R
@@ -342,7 +342,8 @@ yday    = function(x) convertDate(as.IDate(x), "yday")
 wday    = function(x) convertDate(as.IDate(x), "wday")
 mday    = function(x) convertDate(as.IDate(x), "mday")
 week    = function(x) convertDate(as.IDate(x), "week")
-isoweek = function(x) {
+# TODO(#3279): Investigate if improved as.IDate() makes our below implementation faster than this
+isoweek = function(x) as.integer(format(as.IDate(x), "%V"))
   # ISO 8601-conformant week, as described at
   #   https://en.wikipedia.org/wiki/ISO_week_date
   # Approach:
@@ -350,11 +351,11 @@ isoweek = function(x) {
   # * Find the number of weeks having passed between
   #   January 1st of the year of the nearest Thursdays and x
 
-  x = as.IDate(x)   # number of days since 1 Jan 1970 (a Thurs)
-  nearest_thurs = as.IDate(7L * (as.integer(x + 3L) %/% 7L))
-  year_start = as.IDate(format(nearest_thurs, '%Y-01-01'))
-  1L + (nearest_thurs - year_start) %/% 7L
-}
+#  x = as.IDate(x)   # number of days since 1 Jan 1970 (a Thurs)
+#  nearest_thurs = as.IDate(7L * (as.integer(x + 3L) %/% 7L))
+#  year_start = as.IDate(format(nearest_thurs, '%Y-01-01'))
+#  1L + (nearest_thurs - year_start) %/% 7L
+
 
 month   = function(x) convertDate(as.IDate(x), "month")
 quarter = function(x) convertDate(as.IDate(x), "quarter")


MichaelChirico:
Hey @Mukulyadav2004, please ping when tests are passing, or if you're stuck & need an extra pair of eyes. Thanks!

Mukulyadav2004:
Thanks @MichaelChirico ! These both tests are passing when I run on Rstudio but can't figure out why they are failing here.

inst/tests/tests.Rraw
```diff
@@ -21365,3 +21365,10 @@ test(2328.1, levels(droplevels(DT)$f), character())
 DT[, i := integer()]
 DT[, f2 := factor()]
 test(2328.2, droplevels(DT), data.table(f=factor(), i=integer(), f2=factor()))
+
+# Row name extraction from multiple vectors, #7136
+x <- c(1, 2, 3)  
+y <- setNames(c(4, 5, 6), c("A", "B", "C"))  
+test(2329.1, as.data.table.list(list(x, y), keep.rownames=TRUE), data.table(rn=c("A", "B", "C"), V1=c(1, 2, 3), V2=c(4, 5, 6)))
+test(2329.2, as.data.table.list(list(x, y), keep.rownames="custom"), data.table(custom=c("A", "B", "C"), V1=c(1, 2, 3), V2=c(4, 5, 6)))
```
MichaelChirico: don't use the S3 method directly (it's not exported); use as.data.table and let dispatch work.

as for why it's working locally, I assume you're using pkgload::load_all()? or maybe our cc()? either of which would make all private functions available to call in your session

Mukulyadav2004: Thank you! yes I was using devtools::load(). Can you please tell how in future I get to know these kind of fail locally

MichaelChirico: you can R CMD INSTALL the package and run the tests in a clean session with library(data.table); test.data.table().

eventually, you can get more comfortable with which functions are exported vs. private & typically avoid this kind of issue.

imo it's also a useful skill to be able to read the CI failure logs and quickly guess the issue, another thing that just takes practice.

thanks!

R/as.data.table.R
```diff
@@ -133,6 +133,22 @@ as.data.table.list = function(x,
   missing.check.names = missing(check.names)
   origListNames = if (missing(.named)) names(x) else NULL  # as.data.table called directly, not from inside data.table() which provides .named, #3854
   empty_atomic = FALSE
+
+  #Handle keep.rownames for vectors (mimicking data.frame behavior)
+  vector_rownames = NULL
+  if(!identical(keep.rownames, FALSE)) {
```
MichaelChirico: use isFALSE instead

R/as.data.table.R
```diff
@@ -133,6 +133,22 @@ as.data.table.list = function(x,
   missing.check.names = missing(check.names)
   origListNames = if (missing(.named)) names(x) else NULL  # as.data.table called directly, not from inside data.table() which provides .named, #3854
   empty_atomic = FALSE
+
+  #Handle keep.rownames for vectors (mimicking data.frame behavior)
+  vector_rownames = NULL
+  if(!isFALSE(keep.rownames)) {
+    for(i in seq_len(n)){
```
MichaelChirico:
```suggestion
    for (i in seq_len(n)) {
```

R/as.data.table.R
```diff
@@ -133,6 +133,22 @@ as.data.table.list = function(x,
...
+      xi = x[[i]]
+      if (!is.null(xi) && is.atomic(xi) && !is.null(names(xi)) && is.null(dim(xi)) && length(names(xi)) > 0) {
```
MichaelChirico: IINM the condition is redundant

```suggestion
      if (!is.null(xi) && is.atomic(xi) && !is.null(names(xi)) && is.null(dim(xi))) {
```

R/as.data.table.R
```diff
@@ -133,6 +133,22 @@ as.data.table.list = function(x,
...
+        valid_names = names(xi)
+        if(any(nzchar(valid_names))) {
```
MichaelChirico:
```suggestion
        if (any(nzchar(valid_names))) {
```

R/as.data.table.R
```diff
@@ -200,6 +216,13 @@ as.data.table.list = function(x,
   }
   if (any(vnames==".SD")) stopf("A column may not be called .SD. That has special meaning.")
   if (check.names) vnames = make.names(vnames, unique=TRUE)
+
+  # Add rownames column when vector names were found
+  if(!is.null(vector_rownames)){
```
MichaelChirico:
```suggestion
  if (!is.null(vector_rownames)) {
```

R/as.data.table.R
```diff
@@ -133,6 +133,22 @@ as.data.table.list = function(x,
   missing.check.names = missing(check.names)
   origListNames = if (missing(.named)) names(x) else NULL  # as.data.table called directly, not from inside data.table() which provides .named, #3854
   empty_atomic = FALSE
+
+  #Handle keep.rownames for vectors (mimicking data.frame behavior)
+  vector_rownames = NULL
+  if(!isFALSE(keep.rownames)) {
```
MichaelChirico:
```suggestion
  if (!isFALSE(keep.rownames)) {
```

R/as.data.table.R
```diff
@@ -133,6 +133,22 @@ as.data.table.list = function(x,
...
+  vector_rownames = NULL
+  if (!isFALSE(keep.rownames)) {
+    for (i in seq_len(n)) {
+      xi = x[[i]]
+      if (!is.null(xi) && is.atomic(xi) && !is.null(names(xi)) && is.null(dim(xi))) {
+        valid_names = names(xi)
+        if (any(nzchar(valid_names))) {
+          vector_rownames = valid_names
+          x[[i]] = unname(xi)
+          break
```
MichaelChirico: I am leaning towards merging this logic into the below loop. We can write `check_rownames = !isFALSE(keep.rownames)` and then the marginal cost of checking `if (check_rownames && ...)` is low. We can replace the early `break` with checking `is.null(vector_rownames)`. WDYT?

inst/tests/tests.Rraw
```diff
@@ -21365,3 +21365,10 @@ test(2328.1, levels(droplevels(DT)$f), character())
 DT[, i := integer()]
 DT[, f2 := factor()]
 test(2328.2, droplevels(DT), data.table(f=factor(), i=integer(), f2=factor()))
+
+# Row name extraction from multiple vectors, #7136
+x <- c(1, 2, 3)  
+y <- setNames(c(4, 5, 6), c("A", "B", "C"))  
+test(2329.1, as.data.table(list(x, y), keep.rownames=TRUE), data.table(rn=c("A", "B", "C"), V1=c(1, 2, 3), V2=c(4, 5, 6)))
+test(2329.2, as.data.table(list(x, y), keep.rownames="custom"), data.table(custom=c("A", "B", "C"), V1=c(1, 2, 3), V2=c(4, 5, 6)))
```
MichaelChirico: Thanks! Let's make the test suite more extensive:

 - Test also `list(y, x)`
 - Test the behavior under `data.frame()`, not just `as.data.table`
 - Test your condition about `any(nzchar(valid_names))`

I also don't think we've matched data.frame behavior yet, c.f.

```r
DF = data.frame(row.names = letters, V = 1:26)
head(rownames(data.frame(a = 26:1, DF)))
# [1] "a" "b" "c" "d" "e" "f"
```

R/as.data.table.R
Mukulyadav2004: agree

Mukulyadav2004:
@MichaelChirico @tdhock can you please review.

MichaelChirico:
> @MichaelChirico @tdhock can you please review.

You were affected by an underlying change in `master`, see #7149 -- sorry about that.

R/as.data.table.R
```diff
@@ -148,7 +165,7 @@ as.data.table.list = function(x,
         if (is.matrix(xi) && NCOL(xi)<=1L && is.null(colnames(xi))) { # 1 column matrix naming #4124
           xi = x[[i]] = c(xi)
         } else {
-          xi = x[[i]] = as.data.table(xi, keep.rownames=keep.rownames)  # we will never allow a matrix to be a column; always unpack the columns
+          xi = x[[i]] = as.data.table(xi, keep.rownames=FALSE)  # we will never allow a matrix to be a column; always unpack the columns
```
MichaelChirico:
```suggestion
          xi = x[[i]] = as.data.table(xi, keep.rownames=keep.rownames)  # we will never allow a matrix to be a column; always unpack the columns
```

MichaelChirico:
@Mukulyadav2004 I'm curious about the `any(nzchar(nm))` condition -- where did it come from?

Observe that we already can produce `""`-only `rn` column from a matrix today:

```r
M = cbind(1:3)
rownames(M) = rep("", 3L)
as.data.table(M, keep.rownames='blank')
#    blank V1
# 1:        1
# 2:        2
# 3:        3
```

Maybe we should drop that condition here too?

Mukulyadav2004:
I included the `any(nzchar(nm))` condition because I was trying to mimic the behavior of base R's `data.frame()` function, which typically only uses non-empty names as row names.  
However, you're right we should preserve empty string rownames.

MichaelChirico:
Thanks! The data.frame behavior makes sense, but yes, let's break from that. consistency within the data.table API is more important.

Mukulyadav2004:
learned a lot from your changes, thank you.

inst/tests/tests.Rraw
```diff
@@ -21406,3 +21406,17 @@ dt = data.table(x = 123456, y = "wide_string")
...
+test(2330.1, as.data.table(list(x, y), keep.rownames=TRUE), data.table(rn=c("A", "B", "C"), V1=1:3, V2=4:6))
+test(2330.2, as.data.table(list(x, y), keep.rownames="custom"), data.table(custom=c("A", "B", "C"), V1=1:3, V2=4:6))
+test(2330.3, as.data.table(list(y, x), keep.rownames=TRUE), data.table(rn=c("A", "B", "C"), V1=4:6, V2=1:3)) 
+
+# Behavior under data.frame()
+test(2330.4, as.data.table(data.frame(x, y), keep.rownames=TRUE), data.table(rn=c("A", "B", "C"), x=1:3, y=4:6))
+test(2330.5, as.data.table(data.frame(y, x), keep.rownames=TRUE), data.table(rn=c("A", "B", "C"), y=4:6, x=1:3))
+
+DF <- data.frame(row.names = letters[1:6], V = 1:6)     # Test data.frame with explicit rownames
+test(2330.6, as.data.table(list(a = 6:1, DF), keep.rownames=TRUE), data.table(rn=letters[1:6], a=6:1, V=1:6))
```
MichaelChirico, line 21422: rather than remove the rest with all-empty names, let's test the expected behavior in that case as well.

Please also add a test of list(M) for empty-rowname'd matrix input

MichaelChirico (commented):
I think it's basically ready to go, please don't forget a NEWS entry

PR #7138, opened 2025-07-08
various object placement and constness improvements

badasahog:
I was told previously to work on bigger PRs.

This one focuses primarily on object declarations, placements, and constness to further clarify intent earlier in the code.

After this I want to redo the naming scheme for some of the parsing functions and redo the parameters to return objects instead of using return pointers and structs.

I also want to arrange all the benchmarking data into a single struct for clarity. Often the timestamps are held in objects with unclear names.


MichaelChirico:
Please check the CI failure

src/fread.c
```diff
@@ -625,10 +632,14 @@ static void StrtoI32(FieldParseContext *ctx)
...
   const char *ch = *ctx->ch;
-  int64_t *target = (int64_t*) ctx->targets[sizeof(int64_t)];
+  int64_t *target = ctx->targets[sizeof(int64_t)];
   if (*ch == '0' && args.keepLeadingZeros && IS_DIGIT(ch[1])) return;
-  bool neg = *ch == '-';
-  ch += (neg || *ch=='+');
+
+  const bool neg = (*ch == '-');
+
+  if (neg) ch++;
```
MichaelChirico, line 640: why not `if (neg || *ch=='+') ch++;`?

src/fread.c
```diff
@@ -2509,9 +2544,7 @@ int freadMain(freadMainArgs _args) {
             if (thisType != joldType) {             // rare out-of-sample type exception.
               if (!checkedNumberOfFields && !fill) {
                 // check this line has the correct number of fields. If not, don't apply the bump from this invalid line. Instead fall through to myStopEarly below.
-                const char *tt = fieldStart;
-                int fieldsRemaining = countfields(&tt);
-                if (j + fieldsRemaining != ncol) break;
+                if (j + countfields(&(const char*) { fieldStart }) != ncol) break;
```
MichaelChirico: could you explain this change a bit more?

src/fread.c
```diff
@@ -2530,8 +2563,6 @@ int freadMain(freadMainArgs _args) {
                       typeName[IGNORE_BUMP(joldType)], typeName[IGNORE_BUMP(thisType)],
                       (int)(tch - fieldStart), fieldStart, (int64_t)(ctx.DTi + myNrow));
                     
-                    len = iminInt(len, sizeof(buffer));
```
MichaelChirico, line 2533: could you explain why this is no longer needed?

MichaelChirico:
Basically looks good, could you please add some commentary for my edification & posterity?

src/fread.c
badasahog: a matter of taste. That actually looks fine if you prefer it.

src/fread.c
badasahog: it doesn't use temporary variable (which I think is good because it clarifies intent directly.

By doing address-of-temporary you can input the pointer without allowing the function to change it.

src/fread.c
badasahog: the size of the buffer is inputted into snprintf directly.

src/fread.c
MichaelChirico: I think I see it better now. Could we restore the `fieldsRemaining` temporary variable though? I think that improves readability, unless there's a reason against doing so.

src/fread.c
aitap: When truncation occurs, `snprintf` returns the _intended_ number of characters, not counting the terminator, it needs to format the string fully; not the number of characters actually written. Removing `iminInt` means that the next type bump warning may be written way past the current type bump warning, and that `typeBumpMsg` will be over-allocated. With this PR, giant column names together with type bumps could be used to fail the `realloc()` and crash the R session.

The current `master` also has a bug: `len = iminInt(len, sizeof(buffer))` needs to be `len = iminInt(len, sizeof(buffer) - 1)`. Otherwise, when truncation occurs, `len` includes the terminator, and the next `strcpy()` places the string _past_ the terminator of the previous string, so all type bump messages after the first one are invisible to the user. (Since the error messages are truncated, it's probably not feasible to trigger a `realloc()` failure here on current `master`; other allocations fail first.)

This should be a matter of calling `snprintf(typeBumpMsg + offset, 0, format, ...)`, finding out the allocation length, extending the buffer, and then calling `snprintf(typeBumpMsg + offset, remaining_length, format, ...)`, without the temporary buffer on the stack and potential for truncation. And the allocation should probably be checked.

src/fread.c
```diff
@@ -894,29 +916,35 @@ static void parse_double_extended(FieldParseContext *ctx)
...
 
   if (ch[0] == '0' && (ch[1] == 'x' || ch[1] == 'X') &&
-      (ch[2] == '1' || (subnormal = ch[2] == '0')) && ch[3] == '.') {
+      (ch[2] == '1' || (subnormal)) && ch[3] == '.') {
     ch += 4;
     uint64_t acc = 0;
     uint8_t digit;
     const char *ch0 = ch;
-    while ((digit = hexdigits[(uint8_t)(*ch)]) < 16) {
+    while ((digit = hexdigits[(int)*ch]) < 16) {
```
aitap: Won't this read out of bounds on systems where `char` is signed when `*ch < 0`? This includes GNU/Linux on amd64 and Windows:

```r
f <- file('baz.txt', 'w')
writeLines('a', f)
writeLines(rep('0x1.ffffp0', 10000), f)
writeLines('0x1.ff\x9fp0', f)
writeLines(rep('0x1.ffffp0', 20000), f)
close(f)
data.table::fread("baz.txt", verbose = TRUE)
```
```
[11] Read the data                                                                     
  jumps=[0..1), chunk_size=330009, total_size=330009             
fread.c:938:30: runtime error: index -97 out of bounds for type 'uint8_t [256]'
=================================================================   
==13089==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7f18cf14e91f at pc 0x7f18cf03ce2d bp 0x7ffdd82a9af0 sp 0x7ffdd82a9ae8
READ of size 1 at 0x7f18cf14e91f thread T0                                             
    #0 0x7f18cf03ce2c in parse_double_hexadecimal /tmp/RtmpdeCcHp/R.INSTALL2f942700ff5b/data.table/src/fread.c:938
```

src/fread.c
badasahog: this is the only place it's used.

Isn't it more clear to have it inlined so you can see exactly what it is instead of guessing by the identifier?

MichaelChirico: IMO no

badasahog: ok. I will revert

src/fread.c
```diff
@@ -894,29 +916,35 @@ static void parse_double_extended(FieldParseContext *ctx)
...
+  const char *ch = *ctx->ch;
+  double *target = ctx->targets[sizeof(double)];
   *target = NA_FLOAT64;
+  
+  const bool neg = (*ch == '-');
+  
+  if (neg) ch++;
+  else if (*ch == '+') ch++;
+
+  const bool subnormal = ch[2] == '0';
```
MichaelChirico, line 930: Can you help me understand the tradeoffs here?

 - Before, we have a "dynamic" (whatever the opposite of `const` is) `bool`. It is only ever overwritten with `ch[2]=='0'` on inputs where `ch[0] == '0'` and `lower(ch[1] == 'x')`.
 - Now, we are able to declare `const`, but we _always_ read and compare `ch[2]`, even for cases where `ch[1] != '0'` or `lower(ch[1]) != 'x'`

Does the `const` declaration offer enough benefit to justify the additional reads? Some benchmarking on real files would help.

badasahog: It's already on an active cache line, so I think there won't actually be any difference. In addition, I think the compiler should be able to work it out?

I think it's useful to add const to clarify control flow.

I am happy to revert if you insist.

MichaelChirico: Not insisting, I just don't understand C and the compiler that well, I think. I need you to correct my mental model 😃

To me it looks like a major run-time change since `ch[2]` can't be known in advance, which would produce potentially millions of extra `ch[2]` reads on large files with doubles in hex format.

But I think you're saying the compiler will be smart enough not to do the actual `ch[2]` read if it's not needed, is that right?

MichaelChirico: I asked Gemini and it basically called me an idiot 😂

Compilers are magic 🪄

src/fread.c
```diff
@@ -925,7 +953,7 @@ static void parse_double_hexadecimal(FieldParseContext *ctx)
     E = 1023 + (Eneg ? -E : E) - subnormal;
     if (subnormal ? E : (E < 1 || E > 2046)) return;
 
-    *((uint64_t*)target) = (neg << 63) | (E << 52) | (acc);
+    *((uint64_t*)target) = ((1ULL * neg) << 63) | (E << 52) | (acc);
```
MichaelChirico, line 956: nice change to avoid `uint64_t` for `neg` which is naturally `bool` storage

src/fread.c
```diff
@@ -2509,9 +2544,8 @@ int freadMain(freadMainArgs _args) {
             if (thisType != joldType) {             // rare out-of-sample type exception.
               if (!checkedNumberOfFields && !fill) {
                 // check this line has the correct number of fields. If not, don't apply the bump from this invalid line. Instead fall through to myStopEarly below.
-                const char *tt = fieldStart;
-                int fieldsRemaining = countfields(&tt);
-                if (j + fieldsRemaining != ncol) break;
+                  const char* fieldsRemaining = fieldStart;
```
```suggestion
                const char* fieldsRemaining = fieldStart;
```

src/fread.c
MichaelChirico: @badasahog why mark resolved without comment?

src/fread.c
MichaelChirico: Ditto here... if it's fixed by your later changes, it will help to write something here to make it easier to track the lineage.

src/fread.c
badasahog: I thought I fixed it?

src/fread.c
MichaelChirico: Would this work? (and be better?)

```c
int fieldsRemaining = countfields(&(const char*){ fieldStart });
if (j + fieldsRemaining != ncol) break;
```

badasahog: yeah, that would work. better how?

src/fread.c
badasahog: Well it's fixed /:

src/fread.c
MichaelChirico: Purely on style, I suppose, or if it's preferably somehow for the one-time temp variable to be `const` vs. not.

src/fread.c
MichaelChirico: I guess that's 13f1db88b4723c71622d19062743eaac8e277c1a?

src/fread.c
MichaelChirico: I guess that's 19b4e11e4640317703be055a4c3df51541e8041d?

src/fread.c
badasahog: yup

src/fread.c
MichaelChirico: Even better than just reverting would be to add a new regression test that would have detected the error in the first place. That will prevent future authors (including ourselves!) from making the same mistake by encountering a failing test if so.

src/fread.c
MichaelChirico: Could you try and devise a regression test for this case too?

src/fread.c
badasahog: Would you like me to do that? I'm not super good with R.

Besides that are there any other problems with this PR or are you ready to merge?

MichaelChirico: The code LGTM.

For this one, I believe Ivan's example above provides something you could drop into a test: inst/tests/tests.Rraw. I'm less sure about the other one, hopefully Ivan's example gives you a good starting point.

src/fread.c
aitap: Thank you for fixing the immediate problem. The type bump warnings should no longer be lost when the message is truncated. Would you like to work on a follow-up PR to avoid the temporary buffer altogether?

src/fread.c
aitap: Thank you for fixing this problem, the out-of-bounds read is averted. Creating a <400-kilobyte file on the fly for test purposes shouldn't cause us problems.

MichaelChirico:
This change is still pending new tests as requested

badasahog:
Am I expected to write these tests? I'm not proficient in R.

MichaelChirico:
No time like the present to improve your proficiency.

src/fread.c
badasahog: Yeah, I'll get on that

badasahog:
I added @aitap 's test to the testing suite. It's giving a warning, I'm not sure why.

inst/tests/tests.Rraw
```diff
@@ -3106,6 +3106,15 @@ test(1017.4, fread(f, integer64="character"), fread(f, colClasses=c("integer", "
 
 unlink(f)
 
+f = tempfile()
+writeLines('a', f)
+writeLines(rep('0x1.ffffp0', 10000), f)
+writeLines('0x1.ff\x9fp0', f)
+writeLines(rep('0x1.ffffp0', 20000), f)
+fread(f)
```
MichaelChirico: thanks!

please add a comment here describing the purpose of the test and cite the PR number.

also this should invoke test() with an associated number (which must be in increasing order).  maybe easiest to dump these lines at the bottom of the file instead.

inst/tests/tests.Rraw
```diff
@@ -21523,3 +21523,12 @@ test(2332.81, {M1[]=frev(M1); M1}, {M2[]=rev(M2); M2})
 
 # regression test of edge case report #4964
 test(2333, as.expression(data.table(a = 1))[["a"]], 1)
+
+f = tempfile()
+writeLines('a', f)
+writeLines(rep('0x1.ffffp0', 10000), f)
+writeLines('0x1.ff\x9fp0', f)
+writeLines(rep('0x1.ffffp0', 20000), f)
+test(2334, fread(f))
```
MichaelChirico: thanks, three last things:

 1. test() expects something to test in its 3rd argument (default is TRUE). we can just check names(fread(f)) is "a". really what we care about is "no error", which _could_ be expressed more directly, but at the cost of code clarity.
 2. file needs a terminal newline
 3. still need a comment for future reference

MichaelChirico:
Thank you!

MichaelChirico (approved):

src/fread.c
```diff
@@ -279,18 +284,20 @@ static inline bool eol(const char **pch) {
...
+{
   // \r is 13, \n is 10, and \0 is 0. The second part is optimized based on the
   // fact that the characters in the ASCII range 0..13 are very rare, so a
   // single check `ch<=13` is almost equivalent to checking whether `ch` is one
   // of \r, \n, \0. We cast to unsigned first because `char` type is signed by
   // default, and therefore characters in the range 0x80-0xFF are negative.
   // We use eol() because that looks at eol_one_r inside it w.r.t. \r
   // \0 (maybe more than one) before eof are part of field and do not end it; eol() returns false for \0 but the ch==eof will return true for the \0 at eof.
-  return *ch == sep || ((uint8_t)*ch <= 13 && (ch == eof || eol(&ch)));
+  return *ch == sep || (*ch <= 13 && (ch == eof || eol(&ch)));
```
aitap, line 296: Sorry for missing this the first time, but the comment is now wrong. `eol()` is now called for $\mathtt{ch} \in [-128, 13]$ instead of $\mathtt{ch} \in [0, 13]$. At least it's harmless and probably won't show show up in profiling.

src/freadR.c
```diff
@@ -453,20 +457,28 @@ size_t allocateDT(int8_t *typeArg, int8_t *sizeArg, int ncolArg, int ndrop, size
       if (colClassesAs) setAttrib(DT, sym_colClassesAs, colClassesAs);
     } else {
       int nprotect = 0;
-      SEXP tt, ss = R_NilValue;
-      setAttrib(DT, R_NamesSymbol, tt = PROTECT(allocVector(STRSXP, ncol - ndrop))); nprotect++;
+      SEXP tt = PROTECT(allocVector(STRSXP, ncol - ndrop));
+      setAttrib(DT, R_NamesSymbol, tt); nprotect++;
```
aitap, line 461: Stylistic: better keep the `nprotect++` on the same line as the `PROTECT()` it's marking. `setAttrib()`, on the other hand, does not touch the [protection stack](https://cran.r-project.org/doc/manuals/R-exts.html#Garbage-Collection).

PR #7141, opened 2025-07-08
tables() supports recursive search for nested data.tables via recursive=TRUE argument 

### LINKED ISSUE #2606: tables could look for en-list-ed data.tables as well

```
      data.table(a = 2, b = 7:10))
```

venom1204:
closes #2606 

This PR adds a new recursive=TRUE argument to the tables() function to detect data.table objects nested within plain lists,

Changes introduced:

- Added recursive=FALSE (default) to preserve backward compatibility.

*When recursive=TRUE, tables():*
- Recursively traverses plain lists in the environment to find nested data.tables.
- Skips data.frame and data.table containers to avoid descending into list-columns.
- Handles cyclic list references safely using an internal visited environment.
- Generates user-friendly names for nested objects (e.g., mylist$dt, nested[[1]][[2]]).

The recursive logic is implemented iteratively for performance and stack safety.

hi @MichaelChirico can you please have a look when you have time.
thanks..

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -286,7 +286,7 @@ test.list <- atime::atime_test_list(
     Slow = "548410d23dd74b625e8ea9aeb1a5d2e9dddd2927",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/548410d23dd74b625e8ea9aeb1a5d2e9dddd2927)
     Fast = "c0b32a60466bed0e63420ec105bc75c34590865e"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7144/commits) that uses a much faster implementation
 
-  "tables(recursive=FALSE) performance unchanged in #7141" = atime::atime_test(
+  "tables() default performance unchanged in #7141" = atime::atime_test(
     N = as.integer(10^seq(1, 4)),
     setup = {
       test_env <- new.env()
@@ -295,9 +295,10 @@ test.list <- atime::atime_test_list(
         assign(paste0("vec_", i), 1:10, envir = test_env)
       }
     },
-    expr = data.table::tables(env = test_env, recursive = FALSE, silent = TRUE),
+    expr = data.table::tables(env = test_env, silent = TRUE),
 
-    Slow = "7c59daaed1836db57747d92494b1ce96612bbf80",
-    Fast = "6fee38c89200e10dcc10a6f2057ab784f9a011e7"),
+    Slow = "7c59daaed1836db57747d92494b1ce96612bbf80",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/7c59daaed1836db57747d92494b1ce96612bbf80)
+    Fast = "6fee38c89200e10dcc10a6f2057ab784f9a011e7"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7141/commits)
+
     tests=extra.test.list)
 # nolint end: undesirable_operator_linter.


venom1204:
hi @MichaelChirico ,can you please review it if has met the required conditions.

working of the fix -
- Activation: The new search logic only runs if recursive is TRUE. Otherwise, the function behaves exactly as it did before.

- The "To-Do List": It creates a to-do list (an agenda) that starts with all the objects in your current workspace.

The Work Loop: The code then loops through this agenda one item at a time until it's empty. Inside the loop:
  - If it's a data.table: It's a match! The item is saved to a final results list.
  - If it's a list: This is where it goes deeper. All of the list's contents are added to the front of the to-do list, each with a descriptive name (like my_list$item).
  - If it's anything else: It's simply ignored.
- The Safety Net (Cycle Detection): This is the most important part. Before exploring a list, the code checks a high-speed "memory" to see if it has visited that exact list before. If so, it skips it. This prevents the code from ever getting stuck in an infinite loop.

- Displaying Results: After the search is complete, the function takes all the data.tables it found and formats them into the same neat summary table as before.

MichaelChirico:
I will take some time to review as this PR is non-trivial, thanks for your patience.

inst/tests/tests.Rraw
```diff
@@ -21365,3 +21365,28 @@ test(2328.1, levels(droplevels(DT)$f), character())
 DT[, i := integer()]
 DT[, f2 := factor()]
 test(2328.2, droplevels(DT), data.table(f=factor(), i=integer(), f2=factor()))
+
+#2606
+test(2329.1, {
```
MichaelChirico: please avoid using many expressions inside `test()`; also, note that the default for `y` is `TRUE`, so omit it if that's the target value you're testing.

For `tables()`, it's a bit troublesome to test because it's specific to the current environment. I recommend trying something like this for these tests:

```r
local({
  dt1 <- data.table(a=1)
  tables()
})
```

That is assured to only pick up objects defined within the `local({...})` scope.

R/tables.R
```diff
@@ -19,21 +19,62 @@ type_size = function(DT) {
...
+        agenda = c(rev(children_to_add), agenda)
+      }
+    }
+  } else {
+    w = which(vapply_1b(obj, is.data.table))
+    if (length(w)) {
+      found_items = lapply(w, function(i) list(name=names[i], obj=obj[[i]]))
+    }
+  }
+  if (!length(found_items)) { # MODIFIED: Check `found_items` instead of `w`
```
MichaelChirico: we don't need this comment

inst/tests/tests.Rraw
venom1204: I updated it 👍

inst/tests/tests.Rraw
```diff
@@ -21426,3 +21426,27 @@ test(2330.7, as.data.table(list(z), keep.rownames=TRUE), data.table(rn=rep("", 3
 
 M <- matrix(1:6, nrow=3, dimnames=list(rep("", 3), c("V1", "V2")))   #  test of list(M) for empty-rowname'd matrix input
 test(2330.8, as.data.table(list(M), keep.rownames=TRUE), data.table(rn=rep("", 3), V1=1:3, V2=4:6))
+
+#2606 Recursive tables() naming convention
+test(2331.1, local({
```
MichaelChirico: I would combine these three into one test:

```r
local({
  lst_named = list(...)
  lst_unnamed = list(...)
  nested = list(...)
  test(, tables(recursive=TRUE)$NAME, c(...))
})
```

Please also include tests that cover:

 - Mixed named/unnamed simple lists: `list(data.table(..), b=data.table(...))`
 - Mixing named/unnamed in a nested setting too

inst/tests/tests.Rraw
```diff
@@ -21426,3 +21426,27 @@ test(2330.7, as.data.table(list(z), keep.rownames=TRUE), data.table(rn=rep("", 3
...
+test(2331.2, local({
+  lst <- list(data.table(b = 2))
+  "lst[[1]]" %in% tables(recursive = TRUE)$NAME
+}))
+test(2331.3, local({
+  nested <- list(l1 = list(l2 = data.table(c = 3)))
+  "nested$l1$l2" %in% tables(recursive = TRUE)$NAME
+}))
+test(2331.4, local({
+  cycle <- list()
```
MichaelChirico: for cycles, I don't think `list()` is the case we should care about, AFAIK it's not possible to create an address cycle in a list() because of R's copy-on-write semantics. Rather, we should be testing the behavior of an environment with such a cycle, c.f.

```r
l = list(1)
l[[2]] = l
address(l) == address(l[[2]])
# [1] FALSE

e =list2env(list(a = 1))
e$cpy = e
address(e) == address(e$cpy)
# [1] TRUE
```

venom1204: I’ve updated the tests to align with the requirements. Kindly review them at your convenience.

inst/tests/tests.Rraw
```diff
@@ -21489,3 +21489,45 @@ if (test_bit64) {
 # support rotate idiom
 M1 = M2 = matrix(1:4, nrow=2)
 test(2332.81, {M1[]=frev(M1); M1}, {M2[]=rev(M2); M2})
+
+#2606 Recursive tables() naming convention
+test(2333.1, local({
```
MichaelChirico: please read my response more carefully, this is not what I recommended

venom1204: I’ve merged the three tests into a single test case as suggested. Could you please let me know if there’s anything that might still be missing ? —thanks again!

inst/tests/tests.Rraw
```diff
@@ -21523,3 +21523,40 @@ test(2332.81, {M1[]=frev(M1); M1}, {M2[]=rev(M2); M2})
 
 # regression test of edge case report #4964
 test(2333, as.expression(data.table(a = 1))[["a"]], 1)
+
+#2606 Recursive tables() naming convention
+test(2334.1, local({
```
MichaelChirico: again: don't use `local()` inside `test()`, but the other way around.

venom1204: i have submitted the modified test do they align with the requirements .
thankks

R/tables.R
```diff
@@ -19,21 +19,62 @@ type_size = function(DT) {
...
+  if (!length(found_items)) {
+    if (!silent) catf("No objects of class data.table exist in %s%s\n",
+                      if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env),
+                      if (recursive) " (recursively)" else "")
     return(invisible(data.table(NULL)))
   }
-  info = data.table(NAME=names[w], NROW=0L, NCOL=0L, MB=0.0, COLS=list(), KEY=list(), INDICES=list())
-  for (i in seq_along(w)) {  # avoid rbindlist(lapply(DT_names)) in case of a large number of tables
-    DT = obj[[w[i]]]
+  info = data.table(NAME=vapply(found_items, `[[`, "name", FUN.VALUE=character(1L)),
```
MichaelChirico:
```suggestion
  info = data.table(NAME=vapply_1c(found_items, `[[`, "name"),
```

R/tables.R
```diff
@@ -19,21 +19,62 @@ type_size = function(DT) {
...
+  } else {
+    w = which(vapply_1b(objs, is.data.table))
+    if (length(w)) {
+      found_items = lapply(w, function(i) list(name=names[i], obj=obj[[i]]))
+    }
+  }
+  if (!length(found_items)) {
+    if (!silent) catf("No objects of class data.table exist in %s%s\n",
+                      if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env),
+                      if (recursive) " (recursively)" else "")
```
MichaelChirico: this will not translate well. I would just omit this.

```suggestion
    if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
```

R/tables.R
```diff
@@ -19,21 +19,60 @@ type_size = function(DT) {
...
+  }
+  if (!length(found_items)) {
     if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
     return(invisible(data.table(NULL)))
   }
-  info = data.table(NAME=names[w], NROW=0L, NCOL=0L, MB=0.0, COLS=list(), KEY=list(), INDICES=list())
-  for (i in seq_along(w)) {  # avoid rbindlist(lapply(DT_names)) in case of a large number of tables
-    DT = obj[[w[i]]]
+  info = data.table(NAME=vapply_1c(found_items, `[[`, "name"),
+                    NROW=0L, NCOL=0L, MB=0.0, COLS=list(), KEY=list(), INDICES=list())
```
MichaelChirico:
```suggestion
  info = data.table(NAME=vapply_1c(found_items, `[[`, "name"), NROW=0L, NCOL=0L, MB=0.0, COLS=list(), KEY=list(), INDICES=list())
```

R/tables.R
```diff
@@ -19,21 +19,59 @@ type_size = function(DT) {
...
+          }
+          children_to_add[[i]] = list(obj=x[[i]], name=child_name)
+        }
+        agenda = c(rev(children_to_add), agenda)
+      }
+    }
+  } else {
+    w = which(vapply_1b(objs, is.data.table))
+    if (length(w)) {
+      found_items = lapply(w, function(i) list(name=names[i], obj=obj[[i]]))
```
MichaelChirico: does this lead to new copies of the data.table? please check the performance, e.g. consider adding an {atime} test here that confirms `tables()` performs the same under `!recursive` before and after this change.

venom1204: No copies are made. found_items stores a list of names and references (pointers) to the original data.tables.

.ci/atime/tests.R
```diff
@@ -286,5 +286,18 @@ test.list <- atime::atime_test_list(
     Slow = "548410d23dd74b625e8ea9aeb1a5d2e9dddd2927",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/548410d23dd74b625e8ea9aeb1a5d2e9dddd2927)
     Fast = "c0b32a60466bed0e63420ec105bc75c34590865e"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7144/commits) that uses a much faster implementation
 
+  "tables() !recursive refactor in #2606" = atime::atime_test(
+    N = as.integer(10^seq(1, 4, by=0.5)),
```
tdhock: you should omit N unless you need it to be different from the global definition above (and I don't think you do in this case).
```r
  N = as.integer(10^seq(1, 7, by=0.25)),
```

.ci/atime/tests.R
```diff
@@ -286,5 +286,18 @@ test.list <- atime::atime_test_list(
...
+    setup = {
+      test_env <- new.env()
+      for (i in 1:N) {
+        assign(paste0("dt_perf_test", i), data.table(a=1), envir = test_env)
+        assign(paste0("vec_perf_test", i), 1, envir = test_env)
+      }
+    },
+    expr = {data.table::tables(env = test_env, silent = TRUE, index = TRUE); NULL},
+    "before" = "5bb645082aa5c4a295cdd211a5a75c849d590b75",
+    "after" = "8978cf201d8d228506e1e96d3eda7e542471720a"),
```
tdhock: please remove double quotes from "before" and "after" (they are only necessary for non-syntactic names)

tdhock:
atime job errors with this output
```r
* DONE (data.table.8978cf201d8d228506e1e96d3eda7e542471720a)
Registered S3 methods overwritten by 'data.table.5bb645082aa5c4a295cdd211a5a75c849d590b75':
  method                   from                                               
  all.equal.data.table     data.table.c0b32a60466bed0e63420ec105bc75c34590865e
...
  sort_by.data.table       data.table.c0b32a60466bed0e63420ec105bc75c34590865e
Error: Error in `[.data.frame`(x, i) : undefined columns selected
Calls: <Anonymous> ... <Anonymous> -> system.time -> do.call -> <Anonymous>
Timing stopped at: 0.06 0.002 0.062
Execution halted
Error: Process completed with exit code 1.
```
this suggests there is some issue with the version of data.table it is trying to run, 5bb645082aa5c4a295cdd211a5a75c849d590b75, which is the "before" historical version introduced in the new atime test.

.ci/atime/tests.R
tdhock: also please add comments to explain where these version numbers come from.

.ci/atime/tests.R
```diff
@@ -286,5 +286,22 @@ test.list <- atime::atime_test_list(
...
+          assign(paste0("vec_perf_test", i), 1, envir = test_env)
+        }
+      })
+    },
+    expr = {
+      data.table::tables(env = test_env, silent = TRUE, index = TRUE)
+      NULL
+    },
+    before = "2b191aec3df51675a7ab5e6701384a8b89470af6",  # The merge-base parent commit for PR #2606
+    after =  "66024e6ea5304ae059b18c71de9973715ddb77ce"),   # The final commit of PR #2606
```
tdhock: please use the same template for comments as the other test cases (inlcude URLs)
also "final commit" does not make sense for a PR which has not yet been merged

.ci/atime/tests.R
```diff
@@ -300,8 +300,8 @@ test.list <- atime::atime_test_list(
       data.table::tables(env = test_env, silent = TRUE, index = TRUE)
       NULL
     },
-    before = "2b191aec3df51675a7ab5e6701384a8b89470af6",  # The merge-base parent commit for PR #2606
-    after =  "66024e6ea5304ae059b18c71de9973715ddb77ce"),   # The final commit of PR #2606
+    before = "2b191aec3df51675a7ab5e6701384a8b89470af6",  # Merge-base parent commit for PR #2606 (https://github.com/Rdatatable/data.table/commit/2b191aec3df51675a7ab5e6701384a8b89470af6
```
tdhock: please read https://github.com/Rdatatable/data.table/wiki/Performance-testing#documentation-of-historical-commits
the link should be to a PR page (on a page where we can see how the commit id is related to the PR), not to the commit page (on which we do not see anything related to PR2606).

also there is no PR2606, but there is an issue 2606.

venom1204: > also there is no PR2606, but there is an issue 2606.
that was a type error it was meant to be 7141

.ci/atime/tests.R
```diff
@@ -295,13 +295,14 @@ test.list <- atime::atime_test_list(
...
     },
     expr = {
-      data.table::tables(env = test_env, silent = TRUE, index = TRUE)
+      data.table::tables(env = test_env, silent = TRUE, index = TRUE);
       NULL
     },
-    before = "2b191aec3df51675a7ab5e6701384a8b89470af6",  # Merge-base parent commit for PR #2606 (https://github.com/Rdatatable/data.table/commit/2b191aec3df51675a7ab5e6701384a8b89470af6
-    after  = "66024e6ea5304ae059b18c71de9973715ddb77ce"),   # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7141/commits) 
+    before = "2b191aec3df51675a7ab5e6701384a8b89470af6", # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/2b191aec3df51675a7ab5e6701384a8b89470af6)
+    after =  "66024e6ea5304ae059b1e96d3eda7e542471720a"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7141/commits) that uses a much faster implementation
```
tdhock: If this is much faster then please change before to Slow and after to Fast

venom1204: It was only for checking, to ensure there’s no performance regression after adding the recursive argument.

Issue #2611, opened 2018-02-07
bug in week means first week only has 6 days?


MichaelChirico:
```
dates = structure(0:7, class = 'Date')
dates
# [1] "1970-01-01" "1970-01-02" "1970-01-03"
# [4] "1970-01-04" "1970-01-05" "1970-01-06"
# [7] "1970-01-07" "1970-01-08"
```
I think we'd expect `week(dates)` to be `1L` for the first _seven_ days, but this is not the case:

```
week(dates)
# [1] 1 1 1 1 1 1 2 2
```

The fix is trivial -- the problem is `yday` returns `1`-based days, but relying on `%/%` needs `0`-based days:

```
week = function(x) (yday(x) - 1L) %/% 7L + 1L
# -or better, given what yday does- 
week = function(x) as.POSIXlt(x)$yday %/% 7L + 1L
```

Just want to make sure this is not expected behavior? And if this is expected, we should document.

MichaelChirico:
maybe we just add a 'type' argument covering all the cases here?

On Feb 8, 2018 2:31 AM, "Frank" <notifications@github.com> wrote:

Yeah, based on the docs...

These values are all taken directly from the POSIXlt representation of x

I think it was meant to match %U (while data.table::isoweek matches %V).
Eg, January 1st in the example is the fifth day of the week in POSIXlt

> unlist(as.POSIXlt(as.IDate("1970-01-01")))
  sec   min  hour  mday   mon  year  wday  yday isdst
    0     0     0     1     0    70     4     0     0

(A new week starts when wday == 0.)

So the first three days of that year belong to the first week (being its
fifth, sixth and seventh days).

I might find the version Michael has in mind more useful though (where the
first week always has seven days). If switching to that, I guess the docs
should say everything "is derived from the POSIXlt yday", not "taken
directly from the POSIXlt representation".

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Rdatatable/data.table/issues/2611#issuecomment-363865116>,
or mute the thread

UweBlock:
According to `?strptime`, R recognises 3 different conventions for numbering the week of the year:

1. ISO 8601: Week of the year as decimal number (01–53) as defined in ISO 8601. If the week (starting on Monday) containing 1 January has four or more days in the new year, then it is considered week 1. Otherwise, it is the last week of the previous year, and the next week is week 1. (`%V`) 
The first week of the year contains the first Thursday of the year.
[...]

PR #7144, opened 2025-07-08
Use faster base implementation for isoweek

### LINKED ISSUE #5111: Just use base functionality for `isoweek`?
Happened to notice that `format((Date), "%V")` gives identical results to our `isoweek`, and it's a decent amount faster. Should we just migrate our backend to use base?

e.g.

```
x = rep(Sys.Date() - 0:5000, 1000)

Sys.time(out1 <- isoweek(x))
#    user  system elapsed 
# 12.428  21.093  34.376 
Sys.time(out2 <- as.integer(format(x, '%V')))
#    user  system elapsed 
#   4.346   0.221   4.685 

identical(out1, out2)
# [1] TRUE
```

Or should we try and optimize it better?

MichaelChirico:
Closes #5111.

Matt's comment about trying to make `isoweek()` _even faster_ than the base implementation is nice, but in the meantime, we shouldn't miss out on the 20x speed-up.

I'm fairly confident #3279 will help a lot in this case -- the slowest step in the `master` implementation is converting all the Jan. 1 in the input from character with `as.IDate()`. Presumably there will only be a few unique Jan. 1 in the input, meaning we only really need to convert a few unique dates with the slow `strptime()` path.

But anyway, that can be done later for a second, further speed-up.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -277,5 +277,14 @@ test.list <- atime::atime_test_list(
     Slow = "73d79edf8ff8c55163e90631072192301056e336",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/8397dc3c993b61a07a81c786ca68c22bc589befc)
     Fast = "8397dc3c993b61a07a81c786ca68c22bc589befc"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7019/commits) that removes inefficiency
 
+  "isoweek improved in #7144" = atime::atime_test(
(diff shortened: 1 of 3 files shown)


ben-schwen:
LGTM, I only question if we should also add `isoyear = function(x) as.integer(format(as.IDate(x), "%G"))`?

Since `isoyear("2019-12-30")` is `2020`.

MichaelChirico:
Good point -- it also came up in #2611. But as we don't have it yet, I'll mark it for follow-up: #7154.

Issue #7145, opened 2025-07-08
revdeps fail after names handling change


tdhock:
https://github.com/Rdatatable/data.table/commit/131af20abaa9400ae0604ea91f42f33e065f603e is responsible for several new revdep issues:

https://rcdata.nau.edu/genomic-ml/data.table-revdeps/analyze/2025-07-08/

dtts tests (output too large to be useful here) search for "checking tests" in https://rcdata.nau.edu/genomic-ml/data.table-revdeps/analyze/2025-07-08/dtts.txt

tidytable
```
* checking tests ...
  Running 'testthat.R'
 ERROR
Running the tests in 'tests/testthat.R' failed.
Complete output:
  > library(testthat)
  > library(tidytable)
  
  Attaching package: 'tidytable'
  
  The following object is masked from 'package:testthat':
  
      matches
  
  The following objects are masked from 'package:stats':
  
      dt, filter, lag
  
  The following object is masked from 'package:base':
  
      %in%
  
  > 
  > test_check("tidytable")
  [ FAIL 1 | WARN 2158 | SKIP 0 | PASS 1335 ]
  
  == Failed tests ================================================================
  -- Failure ('test-as_tidytable.R:23:3'): works ---------------------------------
  Names of `.m` ('id', 'x') don't match 'id', 'V1'
  
  [ FAIL 1 | WARN 2158 | SKIP 0 | PASS 1335 ]
  Error: Test failures
  Execution halted
```

vardpoor
```

* checking examples ... ERROR
Running examples in 'vardpoor-Ex.R' failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: vardannual
> ### Title: Variance estimation for measures of annual net change or annual
> ###   for single and multistage stage cluster sampling designs
> ### Aliases: vardannual
> ### Keywords: vardannual
> 
> ### ** Examples
> 
>  
> ### Example
> library("data.table")
> 
> set.seed(1)
> 
> data("eusilc", package = "laeken")
> eusilc1 <- eusilc[1:20, ]
> rm(eusilc)
> 
> dataset1 <- data.table(rbind(eusilc1, eusilc1),
+                        year = c(rep(2010, nrow(eusilc1)),
+                                 rep(2011, nrow(eusilc1))))
> rm(eusilc1)
> 
> dataset1[, country := "AT"]
> dataset1[, half := .I - 2 * trunc((.I - 1) / 2)]
[...]

PR #7149, opened 2025-07-08
Don't use auto-inferred name in as.data.table.matrix under new rule

MichaelChirico:
For #7145. This probably doesn't fix everything; I've only checked on {tidytable}.


tdhock:
great, thanks, let's see how this works.

PR #7158, opened 2025-07-11
Add an option for enabling new data.table(<1-column matrix>) auto-naming behavior

### LINKED ISSUE #3193: data.table(vector, matrix) can return duplicate column names
```
data.table(1:5, matrix(6:15, nrow = 5L))
#    V1 V1 V2
# 1:  1  6 11
# 2:  2  7 12
# 3:  3  8 13
# 4:  4  9 14
# 5:  5 10 15
```

Not ideal to be creating `data.table`s with duplicate names from the get-go... those names should be `V2_1` and `V2_2` i guess.

Fixes with `check.names` but I'm not sure creating duplicates like this by default is the right behavior; my only hesitation is I'm sure there are other more complicated cases that are still creating duplicates even if we fix this.

The fix to this specific issue can be very simple, however, just add 7 characters to this line:

https://github.com/Rdatatable/data.table/blob/master/R/data.table.R#L98

```
if (any(tt)) namesi[tt] = paste0("V", i, '_', which(tt))
```

### LINKED ISSUE #7159: 0-column matrix results in 0-row table in data.table()
This is another breakage owing to #4363 -- this broke {latrend}.

```r
M = cbind(1:4)

# current master
data.table(a=1, M[,0])
Empty data.table (0 rows and 2 cols): a,V2

# before #4363
data.table(a=1, M[,0])
#        a
#    <num>
# 1:     1
```

Ignoring such input does seem like the better behavior.

MichaelChirico:
Follow-up to #7149 for #7145. Confirmed all of {dtts}, {latrend}, {vardpoor} pass `R CMD check` under this branch.

Closes #7159; Closes #3193

NB: this has a conflict with #7156.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21455,9 +21455,9 @@ z <- setNames(1:3, rep("", 3))  # vector with all-empty names     # behaviour wi
 test(2330.7, as.data.table(list(z), keep.rownames=TRUE), data.table(rn="", V1=1:3))
 
 M <- matrix(1:6, nrow=3, dimnames=list(rep("", 3L), c("V1", "V2")))   #  test of list(M) for empty-rowname'd matrix input
-test(2332.8, as.data.table(list(M), keep.rownames=TRUE), data.table(rn="", V1=1:3, V2=4:6))
+test(2330.8, as.data.table(list(M), keep.rownames=TRUE), data.table(rn="", V1=1:3, V2=4:6))
 # 0-column input can still provide rownames
-test(2332.9, as.data.table(list(M[, 0L], 1:3), keep.rownames=TRUE), data.table(rn="", V2=1:3))
+test(2330.9, as.data.table(list(M[, 0L], 1:3), keep.rownames=TRUE), data.table(rn="", V2=1:3))
 
 # .SD reference in '...' passed to lapply(FUN=) is recognized as data.table
 test(2331, lapply(list(data.table(a=1:2)), `[`, j=.SD[1L]), list(data.table(a=1L)))


NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 ## data.table [v1.17.99](https://github.com/Rdatatable/data.table/milestone/35)  (in development)
 
+### POTENTIALLY BREAKING CHANGES
+
+1. `data.table(x=1, <expr>)`, where `<expr>` is an expression resulting in a 1-column matrix without column names, will eventually have names `x` and `V2`, not `x` and `V1`, consistent with `data.table(x=1, <expr>)` where `<expr>` results in an atomic vector, for example `data.table(x=1, cbind(1))` and `data.table(x=1, 1)` will both have columns named `x` and `V2`. In this release, the matrix case continues to be named `V1`, but the new behavior can be activated by setting `options(datatable.old.matrix.autoname)` to `FALSE`. See point 14 under Bug Fixes for more context; this change will provide more internal consistency as well as more consistency with `data.frame()`.
```
jangorecki: Nice to keep it backward compatible for this release. Shouldn't we call this section "Coming breaking changes" rather "potentially"?

MichaelChirico: Thanks, I was thinking like that but couldn't find where else we'd used the wording, turns out, not since 2018:

https://github.com/Rdatatable/data.table/blob/40303ce222017d0395ba90bbda2ad8b949f52eb7/NEWS.1.md?plain=1#L1166

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 ## data.table [v1.17.99](https://github.com/Rdatatable/data.table/milestone/35)  (in development)
 
+### POTENTIALLY BREAKING CHANGES
```
```suggestion
### NOTICE OF INTENDED FUTURE POTENTIAL BREAKING CHANGES 
```

PR #7162, opened 2025-07-11
Extend cedta for more lapply() usages

### LINKED ISSUE #2982: lapply(list_o_DTs, `[`, ...) dispatches `[` incorrectly
This is probably an unfixable base R problem, but I figured I'd report just in case:

```
library(data.table)
# data.table 1.11.5 IN DEVELOPMENT built 2018-07-11 03:11:00 UTC; travis  Latest news: http://r-datatable.com
s = list(data.table(a = 1:2))

# okay
s[[1]][, .SD[1L]]
#    a
# 1: 1

# okay
lapply(s, function(x) x[,  .SD[1L]])
# [[1]]
#    a
# 1: 1

# not okay
lapply(s, `[`, j = .SD[1L])
# [[1]]
#    a
# 1: 1
# 2: 2
```

So, `lapply` uses the correct `[` on the parent data.table `s[[1]]`, but uses `[.data.frame` on `.SD`. Even if I try to force usage of the correct function, no dice:

```
lapply(s, `[`, j = data.table:::`[.data.table`(as.data.table(.SD), 1L))
# [[1]]
#    a
# 1: 1
# 2: 2
```

For what it's worth, I arrived at this problem when writing code interactively with magrittr like....

```
split(DT, by="id") %>% lapply(`[`, j = yada yada)
```

I've used this pattern for a long time, so maybe it's a recent bug. (Or maybe I've never mixed `.SD` into it before and so didn't notice this behavior.)

One more, turns out it's not just .SD: 

```
> lapply(s, `[`, j = data.table(a = 3:4)[1L])
[[1]]
   a
1: 3
2: 4
```

MichaelChirico:
Closes #2982

Also refactors the main `||` chain in `cedta()` for readability


jangorecki:
This looks like a risky change now when approaching major release. What do you think about merging it just after 1.18.0 release, so it has more time to mature sitting in devel.

MichaelChirico:
you think the refactoring part is risky, or the new rule? possibly we could tease them apart. I am fine either way, this is just bug triage, not particularly crucial.

(btw, I'm not sure how close we actually are to release)

jangorecki:
Just a heavy rework of the internal mechanism that packages depends on... I haven't look into the changes much but doing such changes before release is like asking for troubles.

MichaelChirico:
IMO the change is pretty simple, low-risk -- just changing a bunch of `||` checks into individual early returns inside a helper. And any possible issue would very likely be caught by revdep checks since `[` must be called O(1M) times in downstreams.

Anyway, I've separated the refactor from the actual bug-fix: #7166.

R/cedta.R
```diff
@@ -47,6 +44,42 @@ cedta.pkgEvalsUserCode = c("gWidgetsWWW","statET","FastRWeb","slidify","rmarkdow
...
+  }
+
+  nsname = getNamespaceName(ns)
+  if (nsname == "data.table") return(TRUE)
+
+  if ("data.table" %chin% names(getNamespaceImports(ns))) return(TRUE) # nocov
+
+  if (isTRUE(ns$.datatable.aware)) return(TRUE) # nocov
+
+  sc <- sys.calls()
```
ben-schwen:
```suggestion
  sc = sys.calls()
```

R/cedta.R
```diff
@@ -47,6 +44,42 @@ cedta.pkgEvalsUserCode = c("gWidgetsWWW","statET","FastRWeb","slidify","rmarkdow
   FALSE
 }
 
+# in a helper to promote readability
+# NB: put the most common and recommended cases first for speed
+.cedta_impl_ <- function(ns, n) {
```
ben-schwen:
```suggestion
.cedta_impl_ = function(ns, n) {
```

PR #7173, opened 2025-07-14
Add a regression test of already-fixed issue converting data.table to expression

### LINKED ISSUE #4964: as.expression
Click preview tab ^^^ above!

By continuing to file this new issue / feature request, I confirm I have :
1. searched the [live NEWS file](https://github.com/Rdatatable/data.table/blob/master/NEWS.md) to see if it has been fixed in dev already. If so, I tried the [latest dev version](https://github.com/Rdatatable/data.table/wiki/Installation#windows).
2. looked at the titles of all the issues in the [current milestones](https://github.com/Rdatatable/data.table/milestones) and am aware of those. (Adding new information to existing issues is very helpful and appreciated.)
3. [searched all issues](https://github.com/Rdatatable/data.table/issues) (i.e. not in a milestone yet) for similar issues to mine and will include links to them explaining why mine is different.
4. searched on [Stack Overflow's data.table tag](http://stackoverflow.com/questions/tagged/data.table) and there is nothing similar there.
5. read the [Support](https://github.com/Rdatatable/data.table/wiki/Support) and [Contributing](https://github.com/Rdatatable/data.table/wiki/Contributing) guides.
6. please don't tag your issue with text in the title; project members will add the appropriate tags later.

#### Thanks! Please remove the text above and include the two items below.

`#` `Output of sessionInfo()`

MichaelChirico:
Closes #4964

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21489,3 +21489,6 @@ if (test_bit64) {
 # support rotate idiom
 M1 = M2 = matrix(1:4, nrow=2)
 test(2332.81, {M1[]=frev(M1); M1}, {M2[]=rev(M2); M2})
+
+# regression test of edge case report #4964
+test(2333, as.expression(data.table(a = 1))[["a"]], 1)


jangorecki:
I wonder if there are any real use cases for that. Being consistent to data.frame is good but in case like this I am not sure if it's worth. Is data.frame behavior documented anywhere?

jangorecki:
Anyway it is already implemented so no sense to try to undo it

jangorecki (approved):

MichaelChirico:
yea, that's why I didn't add any NEWS or make a full equivalence test (.internal.selfref attribute is retained). not erroring seems good enough. I guess the data.frame behavior comes through the list type.

MichaelChirico:
?as.expression has

> lists are changed type to an expression vector (keeping all attributes). Other types [besides atomic vectors and lists] are not currently supported.

so the other way to close this would be to add as.expression.data.table to throw an error directing the user to as.list explicitly first.

PR #7174, opened 2025-07-15
Add shift() error for Date/POSIXct mismatch

### LINKED ISSUE #5218: `shift` converts date to 1970 when converting to timestamp
I realise this is down to using a date for filling the shift rather than a timestamp, but the behaviour is a bit weird and thought best to flag. When using `shift` with timestamps and set a date as the `fill` argument (not necessarily `Sys.Date()`), the filled elements come back as 1st Jan 1970.

```
library(data.table)
dates <- seq(Sys.time() - 5, Sys.time(), 1)
data.table::shift(dates, fill = Sys.Date())
[1] "1970-01-01 06:15:13 CET"  "2021-10-13 12:17:23 CEST" "2021-10-13 12:17:24 CEST" "2021-10-13 12:17:25 CEST"
[5] "2021-10-13 12:17:26 CEST" "2021-10-13 12:17:27 CEST"
```

`#` `Output of sessionInfo()`
```
R version 3.5.2 (2018-12-20)
Running under: Red Hat Enterprise Linux Server 7.4 (Maipo)

Matrix products: default
BLAS/LAPACK: /usr/lib64/R/lib/libRblas.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] data.table_1.14.2

[1] compiler_3.5.2 tools_3.5.2    yaml_2.2.1  
```

ben-schwen:
Closes #5218

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -6854,10 +6854,9 @@ test(1463.80, shift(x,-(1:2),type="cyclic"), list(as.raw(c(2:5, 1)), as.raw(c(3:
 # shift incompatible types (e.g. Date and POSIXct)
 d = .Date(0:4)
 p = .POSIXct(1:5)
-i = as.IDate(2000)
-test(1463.81, shift(d, fill=p[1]), error="Filling Date with POSIXct .* unsupported.*")
-test(1463.82, shift(p, fill=d[1]), error="Filling POSIXct with Date .* unsupported.*")
-test(1463.83, shift(d, fill=as.IDate(2000)), .Date(c(2000, 0:3)))
+test(1463.81, shift(d, fill=p[1L]), error="Filling Date with POSIXct .* unsupported.*")
+test(1463.82, shift(p, fill=d[1L]), error="Filling POSIXct with Date .* unsupported.*")
+test(1463.83, shift(d, fill=as.IDate(2000L)), .Date(c(2000L, 0:3)))
 
 # FR #686
 DT = data.table(a=rep(c("A", "B", "C", "A", "B"), c(2,2,3,1,2)), foo=1:10)
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -6851,6 +6851,12 @@ test(1463.78, shift(x,1:2,   type="cyclic"), list(as.raw(c(5, 1:4)), as.raw(c(4:
 test(1463.79, shift(x,-1L,   type="cyclic"), as.raw(c(2:5, 1)))
 test(1463.80, shift(x,-(1:2),type="cyclic"), list(as.raw(c(2:5, 1)), as.raw(c(3:5,1:2))))
 
+# shift incompatible types (e.g. Date and POSIXct)
+d = .Date(1:5)
+p = .POSIXct(1:5)
+test(1463.81, shift(d, fill=p[1]), error="Filling Date with POSIXct .* unsupported.*")
+test(1463.82, shift(p, fill=d[1]), error="Filling POSIXct with Date .* unsupported.*")
+
 # FR #686
 DT = data.table(a=rep(c("A", "B", "C", "A", "B"), c(2,2,3,1,2)), foo=1:10)
 # Seemingly superfluous 'foo' is needed to test fix for #1942
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -6852,10 +6852,12 @@ test(1463.79, shift(x,-1L,   type="cyclic"), as.raw(c(2:5, 1)))
 test(1463.80, shift(x,-(1:2),type="cyclic"), list(as.raw(c(2:5, 1)), as.raw(c(3:5,1:2))))
 
 # shift incompatible types (e.g. Date and POSIXct)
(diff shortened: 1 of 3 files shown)


MichaelChirico:
The Title is "improve error" but there was no error before. I wonder if others may be relying on the behavior? Seems doubtful, but at least the title should be more accurate.

How does this compare with other scenarios where we do "filling"? I can think of fcase(.default=), dcast(fill=), fifelse(na=), ...

I also wonder if this fix is too narrow, don't we have a warning in bmerge() covering mixed time class types that's at least a little broader?

ben-schwen:
Didn't check all the cases, but seems to be a common problem for all comparisons of `Date` and `POSIXct`. For joins I suspect it to be a common problem but user figure it out because the result is too small so they notice they have done something that they should not.

``` r
library(data.table)
# bmerge goes wrong here
x = data.table(a=.Date(1), b=1)
y = data.table(a=.POSIXct(1), c="a")
x[y, on="a"]
#>                      a     b      c
#>                 <POSc> <num> <char>
#> 1: 1970-01-01 01:00:01     1      a

# dcast goes wrong here
DT = data.table(a=1:10, b=.Date(sample(3, 10L, TRUE)))
dcast(DT, a~b, drop=FALSE, fill=.POSIXct(1000))
#> Using 'b' as value column. Use 'value.var' to override
#> Key: <a>
#>         a 1970-01-02 1970-01-03 1970-01-04
#>     <int>     <Date>     <Date>     <Date>
#>  1:     1 1972-09-27 1970-01-03 1972-09-27
#>  2:     2 1970-01-02 1972-09-27 1972-09-27
#>  3:     3 1972-09-27 1972-09-27 1970-01-04
#>  4:     4 1972-09-27 1972-09-27 1970-01-04
#>  5:     5 1972-09-27 1970-01-03 1972-09-27
#>  6:     6 1972-09-27 1972-09-27 1970-01-04
#>  7:     7 1972-09-27 1972-09-27 1970-01-04
#>  8:     8 1972-09-27 1970-01-03 1972-09-27
#>  9:     9 1972-09-27 1970-01-03 1972-09-27
#> 10:    10 1972-09-27 1970-01-03 1972-09-27
```

MichaelChirico:
Maybe a warning is more appropriate here, WDYT?

We should also try and have a single helper function here to share across the functions (as a follow-up: #7176 ).

inst/tests/tests.Rraw
```diff
@@ -6851,6 +6851,14 @@ test(1463.78, shift(x,1:2,   type="cyclic"), list(as.raw(c(5, 1:4)), as.raw(c(4:
 test(1463.79, shift(x,-1L,   type="cyclic"), as.raw(c(2:5, 1)))
 test(1463.80, shift(x,-(1:2),type="cyclic"), list(as.raw(c(2:5, 1)), as.raw(c(3:5,1:2))))
 
+# shift incompatible types (e.g. Date and POSIXct)
+d = .Date(0:4)
+p = .POSIXct(1:5)
+i = as.IDate(2000)
+test(1463.81, shift(d, fill=p[1]), error="Filling Date with POSIXct .* unsupported.*")
+test(1463.82, shift(p, fill=d[1]), error="Filling POSIXct with Date .* unsupported.*")
+test(1463.83, shift(d, fill=as.IDate(2000)), .Date(c(2000, 0:3)))
```
MichaelChirico, line 6860: I guess this is meant to be `fill=i`? Otherwise delete the `i` temp variable.

ben-schwen: Yes!

ben-schwen:
> Maybe a warning is more appropriate here, WDYT?
> 
> We should also try and have a single helper function here to share across the functions (as a follow-up: #7176 ).

True. I thought about implementing it in coerceAs but didn't want to bloat/slower it for a single case.

MichaelChirico (approved):

src/shift.c
```diff
@@ -40,6 +40,12 @@ SEXP shift(SEXP obj, SEXP k, SEXP fill, SEXP type)
     SEXP elem  = VECTOR_ELT(x, i);
     size_t size  = RTYPE_SIZEOF(elem);
     R_xlen_t xrows = xlength(elem);
+    if (INHERITS(elem, char_Date) && INHERITS(fill, char_POSIXct) || INHERITS(elem, char_POSIXct) && INHERITS(fill, char_Date)) {
```
MichaelChirico, line 43: I think some compiler doesn't like the potential to rely on operator precedence of `||` vs. `&&` here?

https://gitlab.com/Rdatatable/data.table/-/jobs/10696837243

ben-schwen: That's what happens when you think you outsmart the compiler 🤓

PR #7179, opened 2025-07-15
Christian wia patch 1

ChristianWia:

PR #7151, opened 2025-07-10
add french version of datatable-joins.Rmd

ChristianWia:
aligned on last version of June23, 2025 data.table/vignettes/datatable-joins.Rmd

PR #7184, opened 2025-07-16
Use \n in vignette, not \r\n

MichaelChirico:
Somewhat annoyed that GitHub didn't flag this the first time around, I suspect the parent company is to blame 🙃

My local git copy _really_ doesn't like there being a CRLF file around.


ChristianWia:
I dont know from where is this problem. Reports just said that my branch was no longer aligned with the master and requested to solve the conflicts. Web interface just proposed the complete file selected in the modification - which I founded rather strange, but must go through to complete the step.  Any idea ?

MichaelChirico:
@ChristianWia I think your IDE/editor is using CRLF (`\r\n`) line endings, please change the setting so that it uses LF only (`\n`), if you search it should be easy to find.

From within R, you can do `f <- 'vignettes/...'; writeLines(readLines(f), f)` to overwrite all the existing files with `\r`.

ChristianWia:
> @ChristianWia I think your IDE/editor is using CRLF (`\r\n`) line endings, please change the setting so that it uses LF only (`\n`), if you search it should be easy to find.
> 
> From within R, you can do `f <- 'vignettes/...'; writeLines(readLines(f), f)` to overwrite all the existing files with `\r`.

As for the file corrections, I just used the web site editor proposed online by github in a standard way to solve the merge conflict (the pen icon). I have had no control  on the EOL chars definition .   Really strange. ....

MichaelChirico:
Here's my understanding:

 - In #7151, the vignette was submitted with CRLF endings. This was unnoticed because it was only `git add` a new file.
   + `master` has CRLF
 - You created #7179 branching from `master`
   + You inherit CRLF from `master`
 - This PR changed the vignette to use LF
   + `master` has LF
 - Now there is a conflict between your branch (CRLF) and `master` (LF).

GitHub "simple" editor is not well-suited for an end-of-line character resolution, so it's become awkward. Options I see are (1) make the edits locally instead (2) use the in-browser VScode editor https://docs.github.com/en/codespaces/the-githubdev-web-based-editor

ChristianWia:
so the worm was in the apple. Will try githubdev editor on next updates. very clear explaination, i like! thanks.

PR #7187, opened 2025-07-17
Suggest fcoalesce() as an alternative for some nafill() errors

MichaelChirico:
`fcoalesce(x, y)` is equivalent to `nafill(x, type="const", fill=y)` when `length(y)==1`; the former is more general, though, in supporting `length(x)==length(y)` as well.

This change just points the user to the former when possibly misusing `nafill()`.


ben-schwen:
@MichaelChirico I edited the description to include `type=const`. PR LGTM

PR #7189, opened 2025-07-17
Add nan parameter to fcoalesce for NaN/NA distinction control

### LINKED ISSUE #4567: fcoalesce will not replace NaN with NA_real_
```
> fcoalesce(c(NaN, NaN), c(0,NA_real_))
[1]   0 NaN
```

I realize this is  a very edge case, but the current implementation doesn't support simple consistent processing. since `NA_real_` and `NaN` are not equal, i.e. `is.nan(NA_real)` is `FALSE`, i think this should apply the conversion. As it currently stands, the only workaround I can think of is to explicitly do a second pass through the data to explicitly check for and handle this condition

Mukulyadav2004:
fixes #4567 

Currently fcoalesce always treats NaN as missing (same as NA_real_), providing no user control over NaN handling. This inconsistency with nafill, which offers a nan parameter for controlling whether NaN should be treated distinctly from NA.

This PR adds a nan parameter to fcoalesce with default behavior nan=NA (treating NaN as missing, default) and option nan=NaN (preserving NaN values), ensuring fcoalesce(x, y, nan=Z) behaves consistently with nafill(x, fill=y, nan=Z) when y is length-1.

Implemetation:
In `wraper.R` - Added nan parameter with default behaviour - `nan = NA` means treat `NaN` as `NA` - in both fcoalesce as well setcoalesce 
Then update header - `data.table.h` to match new function signature.

In `coalesce.c`
- Add third parameter - `nan_is_na_arg` to support user control
- After then extract and validate the nan control parameter
- Apply user's nan preference when checking scalar fill values, for the final value checking and in main replacement loop.

The logic is :
nan_is_na(NA) -> TRUE - then use ISNAN(val) - thus both NaN and NA_real treated as missing
nan_is_na(NaN) -> FALSE - then use ISNA(val) - thus only NA_real treated as missing, NaN preserved

In Docs change - `coalesce.Rd` : 
- Added nan parameter to both function signatures
- Added explanation of how nan = NA vs nan = NaN affects behaviour.
- Demonstrate the difference between default and explicit nan = NaN behaviour 

@MichaelChirico @jangorecki can you please review

### DIFF
--- a/.ci/.lintr.R
+++ b/.ci/.lintr.R
@@ -21,6 +21,7 @@ linters = c(dt_linters, all_linters(
     message = "Use messagef to avoid fragmented translations.",
     warning = "Use warningf to avoid fragmented translations.",
     stop = "Use stopf to avoid fragmented translations.",
+    rev = "Use frev internally, or setfrev if by-reference is safe.",
     NULL
   )),
   # undesirable_function_linter(modify_defaults(
--- a/.ci/linters/r/eval_parse_linter.R
+++ b/.ci/linters/r/eval_parse_linter.R
@@ -1,0 +1,8 @@
+eval_parse_linter = make_linter_from_xpath(
+  "//SYMBOL_FUNCTION_CALL[text() = 'parse']
+     /ancestor::expr
+     /preceding-sibling::expr[SYMBOL_FUNCTION_CALL[text() = 'eval']]
+     /parent::expr
+  ",
+  "Avoid eval(parse()); build the language directly, possibly using substitute2()."
+)
--- a/.ci/linters/rd/condition_class_docs_check.R
+++ b/.ci/linters/rd/condition_class_docs_check.R
@@ -1,0 +1,51 @@
+# Ensure that data.table condition classes in code match documentation
+condition_classes_documentation_linter = function(rd_file) {
+  if (!grepl("\\name{data.table-condition-classes}", readChar(rd_file, 100L), fixed = TRUE)) return(invisible())
+
+  # Find condition classes in R code 
+  walk_r_ast_for_classes = function(expr) {
+    if (is.call(expr) && is.name(e <- expr[[1L]]) && as.character(e) %in% c("stopf", "warningf", "messagef", "packageStartupMessagef") && is.character(class_arg <- expr[["class"]]) && startsWith(class_arg, "dt_")) {
+      class_arg
+    } else if (is.recursive(expr)) {
+      unlist(lapply(expr, walk_r_ast_for_classes))
+    }
+  }
+
+  # Find condition classes in documentation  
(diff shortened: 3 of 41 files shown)


man/coalesce.Rd
```diff
@@ -7,10 +7,11 @@ Fill in missing values in a vector by successively pulling from candidate vector
 Written in C, and multithreaded for numeric and factor types.
 }
 \usage{
-  fcoalesce(\dots)
+  fcoalesce(\dots, nan = NA)
```
MichaelChirico:
```suggestion
  fcoalesce(\dots, nan=NA)
```

src/coalesce.c
```diff
@@ -116,9 +118,9 @@ SEXP coalesce(SEXP x, SEXP inplaceArg) {
       #pragma omp parallel for num_threads(getDTthreads(nrow, true))
       for (int i=0; i<nrow; ++i) {
         double val=xP[i];
-        if (!ISNAN(val)) continue;
-        int j=0; while (ISNAN(val) && j<k) val=((double *)valP[j++])[i];
-        if (!ISNAN(val)) xP[i]=val; else if (final) xP[i]=finalVal;
+        if (!(nan_is_na ? ISNAN(val) : ISNA(val))) continue;
```
MichaelChirico: Would you mind benchmarking if this has a performance impact? Ideally the compiler optimizes this away, but it is potentially worrisome to see this ternary operator being introduced inside the parallel region.

Something like vector of size `5e7`.

Mukulyadav2004: Yes there is a significant performance impact and your concern about the ternary operator in the parallel region is right
As the 5e7 elements benchmark result has median execution times as - 
original: 168.57 ms
nan=NA:   195.95 ms  
nan=NaN:  165.67 ms

MichaelChirico: In that case, we typically just duplicate the logic -- two copy-pasted loops, one using `ISNA`, the other `ISNAN`.

PS, don't miss the `ISNAN` usage above on L115[117]. It looks like you should add a new test corresponding to this omission, too.

Mukulyadav2004: Hi @MichaelChirico,
I’ve duplicated the loop to separately handle both the nan = NA and nan = NAN paths. This seems to work well:
Original: 343.09 ms
nan=NA: 270.27 ms
nan=NaN: 293.64 ms 
Although it did add quite a few lines. Just wanted to check if this direction looks okay to you, or if you’d suggest a more concise alternative.

MichaelChirico (approved):
LGTM! Please add a NEWS entry

MichaelChirico:
sorry, last step is to merge master to resolve the conflict

Mukulyadav2004:
Done, sorry for the delay, my laptop's been quite slow but the merge is complete now.

Issue #7164, opened 2025-07-12
`between()` gives wrong answers for some `integer64` inputs


aitap:
`#` `Minimal reproducible example`

Let's take these two perfectly normal (if somewhat large) 64-bit integers:
```
library(bit64)
a <- as.integer64('4630967054332067840')
b <- as.integer64('4631248529308778496')
```
...and test whether another integer lies between them:
```
library(data.table)
between(42L, a, b)
# [1] TRUE
```
That doesn't seem right.

When interpreted as doubles, these numbers turn out to be small integers:
```
unclass(c(a,b))
# [1] 41 43
```
...and so they pass the test for `within_int32_repres(REAL(x)[i])`:

https://github.com/Rdatatable/data.table/blob/2f0d12fbbbb02bc017862a6b26b5b06845cf0a7c/src/utils.c#L20-L23

...which is enough to have them converted to integers, ignoring the `integer64` class:

https://github.com/Rdatatable/data.table/blob/2f0d12fbbbb02bc017862a6b26b5b06845cf0a7c/src/between.c#L33-L39

Should `fitsInInt32` return `false` when its argument inherits from `integer64`?

`#` `Output of sessionInfo()`

```
R Under development (unstable) (2025-07-11 r88405)
Platform: x86_64-pc-linux-gnu
Running under: Debian GNU/Linux 12 (bookworm)

Matrix products: default
BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.21.so;  LAPACK version 3.11.0

locale:
 [1] LC_CTYPE=ru_RU.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=ru_RU.UTF-8        LC_COLLATE=ru_RU.UTF-8
 [5] LC_MONETARY=ru_RU.UTF-8    LC_MESSAGES=ru_RU.UTF-8
 [7] LC_PAPER=ru_RU.UTF-8       LC_NAME=C
 [9] LC_ADDRESS=C               LC_TELEPHONE=C
[11] LC_MEASUREMENT=ru_RU.UTF-8 LC_IDENTIFICATION=C

time zone: Europe/Moscow
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] data.table_1.17.99 bit64_4.6.0-1      bit_4.6.0

loaded via a namespace (and not attached):
[1] compiler_4.6.0
```

[...]

PR #7193, opened 2025-07-19
Fixes for `integer64` issues in `between`

aitap:
Fixes: #7164

* Move the check for `integer64` mismatch between `x` and `lower`/`upper` above the potential type conversions.
   * Optionally: fail `integer64` vectors in `fitsInInt32` and `fitsInInt64`. The additional calls to `inherits()` may result in a slowdown?
* Do not auto-convert the boundaries from `numeric` to `integer64`. This might break if someone relied on `between(<integer64>, <numeric fraction>, <numeric fraction>)` automatically rounding to `integer64`. May have to check the range instead of fully checking `fitsInInt64`?

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -533,10 +533,8 @@ bool isRectangularList(SEXP x) {
   return isRectangular(x);
 }
 
-// TODO: use isDataFrame (when included in any R release).
-// isDataTable(x) || isFrame(x) || isRectangularList(x)
 bool perhapsDataTable(SEXP x) {
-  return isDataTable(x) || isFrame(x) || isRectangularList(x);
+  return isDataTable(x) || isDataFrame(x) || isRectangularList(x);
 }
 SEXP perhapsDataTableR(SEXP x) {
   return ScalarLogical(perhapsDataTable(x));
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -153,7 +153,7 @@ if (getRversion() >= "3.6.0") {
 
 # IDateTime support:
 export(as.IDate,as.ITime,IDateTime)
-export(second,minute,hour,yday,wday,mday,week,isoweek,month,quarter,year,yearmon,yearqtr)
+export(second,minute,hour,yday,wday,mday,week,isoweek,isoyear,month,quarter,year,yearmon,yearqtr)
 
 S3method("[", ITime)
 S3method("+", IDate)
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,16 @@
 
 ### NEW FEATURES
 
-1. New `sort_by()` method for data.tables, [#6662](https://github.com/Rdatatable/data.table/issues/6662). It uses `forder()` to improve upon the data.frame method and also match `DT[order(...)]` behavior with respect to locale. Thanks @rikivillalba for the suggestion and PR.
+1. New `sort_by()` method for data.tables, [#6662](https://github.com/Rdatatable/data.table/issues/6662). It uses `forder()` to improve upon the data.frame method and also matches `DT[order(...)]` behavior with respect to locale. Thanks @rikivillalba for the suggestion and PR.
+
+    ```r
+    DT = data.table(a=c(1L, 2L, 1L), b=c(3L, 1L, 2L))
+    sort_by(DT, ~a + b)
+    #    a b
+    # 1: 1 2
+    # 2: 1 3
+    # 3: 2 1
+    ```
 
 2. `melt()` now supports using `patterns()` with `id.vars`, [#6867](https://github.com/Rdatatable/data.table/issues/6867). Thanks to Toby Dylan Hocking for the suggestion and PR.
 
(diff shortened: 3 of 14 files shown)


src/between.c
```diff
@@ -29,6 +29,13 @@ SEXP between(SEXP x, SEXP lower, SEXP upper, SEXP incbounds, SEXP NAboundsArg, S
   const bool check = LOGICAL(checkArg)[0];
   const bool verbose = GetVerbose();
 
+  // check before potential coercion which ignores methods, #7164
+  if (INHERITS(x, char_integer64)) {
+    if (!INHERITS(lower, char_integer64) || !INHERITS(upper, char_integer64))
+      error(_("x is integer64 but lower and/or upper are not.")); // e.g. between(int64, character, character)
```
MichaelChirico:
```suggestion
    if (!INHERITS(lower, char_integer64))
      error(_("x is integer64 but %s is not."), "lower"); // e.g. between(int64, character, character)
    if (!INHERITS(upper, char_integer64))
      error(_("x is integer64 but %s is not."), "upper"); // e.g. between(int64, character, character)
```

src/between.c
```diff
@@ -29,6 +29,13 @@ SEXP between(SEXP x, SEXP lower, SEXP upper, SEXP incbounds, SEXP NAboundsArg, S
...
+  } else if (INHERITS(lower, char_integer64) || INHERITS(upper, char_integer64))
+    error(_("x is not integer64 but lower and/or upper is integer64. Please align classes."));
```
MichaelChirico:
```suggestion
  } else {
    if (INHERITS(lower, char_integer64))
      error(_("x is not integer64 but %s is. Please align classes."), "lower");
    if (INHERITS(upper, char_integer64))
      error(_("x is not integer64 but %s is. Please align classes."), "upper");
  }
```

src/utils.c
```diff
@@ -14,7 +14,7 @@ bool within_int64_repres(double x) {
 // used to error if not passed type double but this needed extra is.double() calls in calling R code
 // which needed a repeat of the argument. Hence simpler and more robust to return false when not type double.
 bool fitsInInt32(SEXP x) {
-  if (!isReal(x))
+  if (!isReal(x) || INHERITS(x, char_integer64))
```
MichaelChirico, line 17: Could you elaborate here? It seems redundant, I thought `isReal()` is just testing `TYPEOF()`?

https://github.com/r-devel/r-svn/blob/cc4aa7f99b0107e42b368a405f77bfb1fd385299/src/include/Defn.h#L770-L771

So that this is just an added test for a potentially malformed integer64?

aitap: The `integer64` class is a very clever invention. There's no 64-bit integer `SEXPTYPE` in R, and it's impossible for a CRAN package to create a new one, so `integer64` takes the vector `SEXPTYPE` that (de-facto) consists of 8-byte elements (`REALSXP`) and then overrides as many methods as it can. The methods know to cast the `double *` pointer returned by `REAL()` into `int64_t *`. This mostly works well, but breaks down when R is allowed to ignore the class and look at `REAL()` by itself:

```
> .Internal(inspect( bit64::as.integer64(42) ))
@555cf8f271d0 14 REALSXP g0c1 [OBJ,ATT] (len=1, tl=0) 2.07508e-322 # ← stored as a very small double
ATTRIB:
  @555cf8da5f10 02 LISTSXP g0c0 [REF(1)]
    TAG: @555cf6d73b80 01 SYMSXP g1c0 [MARK,REF(34912),LCK,gp=0x4000] "class" (has value)
    @555cf8f26f30 16 STRSXP g0c1 [REF(65535)] (len=1, tl=0)
      @555cf7d7b5e8 09 CHARSXP g0c2 [MARK,REF(1015),gp=0x61] [ASCII] [cached] "integer64"
> bit64::as.integer64(42) |> deparse(control = c('hexNumeric')) # remove the class and look at the double
[1] "0x0.000000000002ap-1022"
> structure(0x0.000000000002ap-1022, class = 'integer64') # reattach the class to the double
integer64
[1] 42 # ← in C that would be 42 == *(int64_t *)&(double){0x0.000000000002ap-1022}
```

The same approach is taken by the [`float`](https://cran.r-project.org/package=float) package, except it stores 4-byte `float`s in `INTSXP` vectors:

```
> float::as.float(42)
# A float32 vector: 1
[1] 42
> float::as.float(42)@Data
[1] 1109917696 # ← *(int *)&(float){42.F}
> writeBin(1109917696L, raw()) |> readBin(numeric(), 1, size = 4) # R doesn't have floats, but it can read them
[1] 42
```

src/between.c
```diff
@@ -29,6 +29,19 @@ SEXP between(SEXP x, SEXP lower, SEXP upper, SEXP incbounds, SEXP NAboundsArg, S
   const bool check = LOGICAL(checkArg)[0];
   const bool verbose = GetVerbose();
 
+  // check before potential coercion which ignores methods, #7164
+  if (INHERITS(x, char_integer64)) {
+    if (!INHERITS(lower, char_integer64))
+      error(_("x is integer64 but %s is not."), "lower"); // e.g. between(int64, character, character)
+    if (!INHERITS(upper, char_integer64))
+      error(_("x is integer64 but %s is not."), "upper"); // e.g. between(int64, character, character)
```
MichaelChirico: now that they're all side-by-side... add "Please align classes" here too? It looks somewhat oddly asymmetric now.

src/utils.c
MichaelChirico: I should know, I'm the maintainer of {bit64} =D

I see now my mistake, I was mis-reading / ignored the `!`, I thought this was `isReal(x) || inherits(x, "integer64")`, which _would_ be redundant.

aitap:
Does the change in behaviour regarding `integer64` boundaries need a news entry? I don't see it documented anywhere.

MichaelChirico:
Ah, yes, that's a bug fix based on the issue you filed.

PR #7196, opened 2025-07-21
added example of sort_by in the news.

### LINKED ISSUE #7140: Add sort_by() example in NEWS
This PR would benefit from example of usage in NEWS file

_Originally posted by @jangorecki in https://github.com/Rdatatable/data.table/issues/6679#issuecomment-3048408891_

venom1204:
closes #7140 

I’ve added a brief example of `sort_by()` to `NEWS.md` for clarity.

Hi @MichaelChirico, could you take a look when you have a moment?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,17 @@
 
 ### NEW FEATURES
 
-1. New `sort_by()` method for data.tables, [#6662](https://github.com/Rdatatable/data.table/issues/6662). It uses `forder()` to improve upon the data.frame method and also match `DT[order(...)]` behavior with respect to locale. Thanks @rikivillalba for the suggestion and PR.
+1. New `sort_by()` method for data.tables, [#6662](https://github.com/Rdatatable/data.table/issues/6662). It uses `forder()` to improve upon the data.frame method and also match `DT[order(...)]` behavior with respect to locale, and allows sorting by expressions or functions. Thanks @rikivillalba for the suggestion and PR.
+
+```r
+    DT = data.table(a = c(1, 2, 1), b = c(3, 1, 2))
+    # sort by column 'b'
+    sort_by(DT, ~b)
+    #    a b
+    # 1: 2 1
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -10,7 +10,17 @@
 
 ### NEW FEATURES
 
-1. New `sort_by()` method for data.tables, [#6662](https://github.com/Rdatatable/data.table/issues/6662). It uses `forder()` to improve upon the data.frame method and also match `DT[order(...)]` behavior with respect to locale. Thanks @rikivillalba for the suggestion and PR.
+1. New `sort_by()` method for data.tables, [#6662](https://github.com/Rdatatable/data.table/issues/6662). It uses `forder()` to improve upon the data.frame method and also match `DT[order(...)]` behavior with respect to locale, and allows sorting by expressions or functions. Thanks @rikivillalba for the suggestion and PR.
```
MichaelChirico: > , and allows sorting by expressions or functions

What's the purpose of this?

NEWS.md
```diff
@@ -10,7 +10,17 @@
...
+
+```r
+    DT = data.table(a = c(1, 2, 1), b = c(3, 1, 2))
```
MichaelChirico: I would do an example that uses two sorting keys, `~a+b`, to demonstrate how sorting by >1 column works.

NEWS.md
venom1204: it highlights that sort_by() can take tilde‐formulas (e.g. ~ x + y, ~ log(z)) or functions as sort keys, evaluating them per row for on‐the‐fly sorting. I’ll revert this.

MichaelChirico: I see, as written it looked to mean `expression` or `function` types. Something like ", and allows sorting by generic expressions of columns" would be clearer, but I think it's best to just point the user to `?base::sort_by` to understand the S3 generic generally. I would omit it in this PR.

Issue #7183, opened 2025-07-16
ASAN error in fread.c


MichaelChirico:
Here are the logs of the error:

https://gitlab.com/Rdatatable/data.table/-/jobs/10696837289

I think we should fix the bug (presumably introduced by #7138) before continuing to dig a hole with further edits to the file.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/issues/7181#issuecomment-3079228687_

aitap:
The failing test is 1869.6, for files whose size is a multiple of the
memory page size in the operating system:

https://github.com/Rdatatable/data.table/blob/32e5aea1681398335f1bcedb4fe1563906701634/inst/tests/tests.Rraw#L12582-L12587

1869.7 may fail as well.

aitap:
The out-of-bounds read can be turned into a real crash if a file that
fits in a large but exact number of pages is left mapped into memory,
not copied like in test 1869.6, and detect_types() tries
parse_double_hexadecimal(), which reads memory past the last page
containing the file contents:

# edited below: make sure the crash happens on Windows too
f <- file("foo.txt", "wb")
writeLines(
 c("a234567", rep("0x1.56p", 4096*1000/8-3), "0x1.5p", "0x1.5p", "0"),
 f
)
close(f)
data.table::fread("foo.txt", verbose = TRUE)
<...>
[02] Opening the file
  Opening file foo.txt
  File opened, size = 3.906MiB (4096000 bytes).
  Memory mapped ok
[03] Detect and skip BOM
[04] Arrange mmap to be \0 terminated
  \n has been found in the input and different lines can end with
  different line endings (e.g. mixed \n and \r\n in one file). This is
  common and ideal.
<...>
[07] Detect column types, dec, good nrow estimate and whether first row
is column names
  Number of sampling jump points = 100 because (4095999 bytes from row
  1 to eof) / (2 * 792 jump0size) == 2585
  dec='.' detected based on a balance of 0 parsed fields
  Type codes (jump 000)    : B  Quote rule 0
  
 *** caught segfault ***
address 0x7f46627e8000, cause 'memory not mapped'

Traceback:
 1: data.table::fread("foo.txt", verbose = TRUE)
An irrecoverable exception occurred. R is aborting now ...

This is one of the places where the code implicitly relies on the
invariant that *eof is always addressable and equal to '\0'
<https://github.com/Rdatatable/data.table/pull/7037#issuecomment-2927740711>.
[...]

PR #7198, opened 2025-07-22
reformatting in more files

badasahog:
As far as I can tell these files haven't been touched in a while, so there shouldn't be any major conflicts.

Only one change was functional (switching to memset). Other than that, everything here is purely cosmetic.

### DIFF
--- a/src/freadR.c
+++ b/src/freadR.c
@@ -572,18 +572,6 @@ void dropFilledCols(int* dropArg, int ndelete)
 
 void pushBuffer(ThreadLocalFreadParsingContext *ctx)
 {
-  const void *buff8 = ctx->buff8;
-  const void *buff4 = ctx->buff4;
-  const void *buff1 = ctx->buff1;
-  const char *anchor = ctx->anchor;
-  const size_t nRows = ctx->nRows;
-  const size_t DTi = ctx->DTi;
-  const size_t rowSize8 = ctx->rowSize8;
-  const size_t rowSize4 = ctx->rowSize4;
-  const size_t rowSize1 = ctx->rowSize1;
-  const int nStringCols = ctx->nStringCols;
-  const int nNonStringCols = ctx->nNonStringCols;
-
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Hi @badasahog I would prefer to focus on the new segfault #7183. We'll have to revert the breaking change if the new issue can't be fixed.

badasahog:
I think that's in a separate part of the code?

Also, do you have a minimal example I could use to recreate the segfault?

All tests cleared on my end.

MichaelChirico:
Did you test under ASAN? The issue is shown clearly in the logs linked in the issue. Let's please continue the discussion there.

badasahog:
now that that's done, is this ready to go?

MichaelChirico:
Revert the changes to nafill.c and uniqlist.c which might conflict with pending PRs: #5692, #5682, #5576, #5575, #4372

src/nqrecreateindices.c
```diff
@@ -17,26 +17,27 @@ SEXP nqRecreateIndices(SEXP xo, SEXP len, SEXP indices, SEXP nArg, SEXP nomatch)
   const int inomatch = isNull(nomatch) ? 0 : INTEGER(nomatch)[0];
   int *inewstarts = INTEGER(newstarts);
 
-  for (int i=0; i<n; ++i) inewlen[i] = 0;
+  memset(inewlen, 0, n * sizeof(*inewlen));
```
MichaelChirico: let's apply the `memset()` change in a clean PR. It is a somewhat risky function, e.g. https://pvs-studio.com/en/blog/posts/cpp/0360/, therefore I'd prefer it to be more cleanly identified in the `git` history.

badasahog: gotcha!

badasahog:
are there any other files that don't have outstanding PRs?

MichaelChirico:
You can check: https://gist.github.com/MichaelChirico/64e981b4359a36d2279bf55c39c7d3dd

PR #7200, opened 2025-07-22
Fix fread segfault from faulty assumption on compilers

### LINKED ISSUE #7183: ASAN error in fread.c

badasahog:
Closed #7183 cc @MichaelChirico

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -929,6 +929,9 @@ static void parse_double_hexadecimal(FieldParseContext *ctx)
 
   bool subnormal = false;
 
+  // Important!
+  // Keep in mind that only ch[0] is guaranteed to be mapped.
+  // Rearanging these checks will lead to segfault in rare cases.
   if (ch[0] == '0' && (ch[1] == 'x' || ch[1] == 'X') &&
       (ch[2] == '1' || (subnormal = (ch[2] == '0'))) && ch[3] == '.') {
     ch += 4;
--- a/src/fread.c
+++ b/src/fread.c
@@ -927,10 +927,10 @@ static void parse_double_hexadecimal(FieldParseContext *ctx)
   if (neg) ch++;
   else if (*ch == '+') ch++;
 
-  bool subnormal;
+  bool subnormal = false;
 
   if (ch[0] == '0' && (ch[1] == 'x' || ch[1] == 'X') &&
-      (ch[2] == '1' || (subnormal = ch[2] == '0')) && ch[3] == '.') {
+      (ch[2] == '1' || (subnormal = (ch[2] == '0'))) && ch[3] == '.') {
     ch += 4;
     uint64_t acc = 0;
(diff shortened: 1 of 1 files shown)


MichaelChirico:
thanks! would you mind

(1) adding a helpful comment for any intrepid future devs that have the same thoughts about making it const
(2) re-read the older PR again with fresh eyes, if memory serves this type of change was made more than once [it could just mean the other instances are not covered by tests & this not showing up]

src/fread.c
```diff
@@ -927,10 +927,13 @@ static void parse_double_hexadecimal(FieldParseContext *ctx)
   if (neg) ch++;
   else if (*ch == '+') ch++;
 
-  const bool subnormal = ch[2] == '0';
+  bool subnormal = false;
 
+  // Important!
+  // Keep in mind that only ch[0] is guaranteed to be mapped.
+  // Rearanging these checks will lead to segfaults in rare cases.
```
```suggestion
  // Rearranging these checks (e.g. to make 'subnormal' const) will lead to segfaults in rare cases.
```

MichaelChirico:
I believe these are also at risk:

https://github.com/Rdatatable/data.table/blob/c8bbb580dd5f54710583a3477382e949d093b302/src/fread.c#L775
https://github.com/Rdatatable/data.table/blob/c8bbb580dd5f54710583a3477382e949d093b302/src/fread.c#L946

Can you think of an input file that would trigger the segfault? If so we should add them as regression tests.

badasahog:
Those wouldn't create a problem because ch[1] is tested first anyway (I'm not rearranging the order of access as far as I can tell)

MichaelChirico:
Oh, got it, `ch[0]` is already implicitly tested before in both cases:

https://github.com/Rdatatable/data.table/blob/c8bbb580dd5f54710583a3477382e949d093b302/src/fread.c#L772

https://github.com/Rdatatable/data.table/blob/c8bbb580dd5f54710583a3477382e949d093b302/src/fread.c#L932

MichaelChirico:
I suppose we could retain the `const` label by adding a layer of nesting:

```c
if (ch[0] == '0' && (ch[1] == 'x' || ch[1] == 'X')) {
  const bool subnormal = ch[2] == '0';
  if ((subnormal || ch[2] == '1') && ch[3] == '.') {
    // ...
  }
}
```

But I'm not sure the `const` offers that much benefit to be worth the slightly worse readability of the code.

badasahog:
the whole point of adding const is to improve readability and clarify intent. adding would seem counterintuitive.

PR #7201, opened 2025-07-22
context= parameter to test()

### LINKED ISSUE #7119: Add a context= or info= parameter to test() for looped/functional tests
Whenever we need to loop `test()`, the error becomes hard to determine from the failure itself:

```r
for (ii in 0:4 / 5) test(1 + ii, ii < 0.5)
# Test 1.6 ran without errors but failed check that x equals y:
# > x = ii < 0.5 
# First 1 of 1 (type 'logical'): 
# [1] FALSE
# > y = TRUE 
# First 1 of 1 (type 'logical'): 
# [1] TRUE
# 1 element mismatch
# Test 1.8 ran without errors but failed check that x equals y:
# > x = ii < 0.5 
# First 1 of 1 (type 'logical'): 
# [1] FALSE
# > y = TRUE 
# First 1 of 1 (type 'logical'): 
# [1] TRUE
# 1 element mismatch
```

It would be a lot better to be able to insert some string into the `test()` call to capture the context of the loop, e.g.

```r
for (ii in 0:4 / 5) test(1 + ii, ii < 0.5, context=sprintf("ii=%.1f", ii))
```

Which could then be read directly off of the error.

MichaelChirico:
Closes #7119

Should probably sit for a while to avoid conflicts in tests.Rraw.

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -323,7 +323,10 @@ gc_mem = function() {
   # nocov end
 }
 
-test = function(num,x,y=TRUE,error=NULL,warning=NULL,message=NULL,output=NULL,notOutput=NULL,ignore.warning=NULL,options=NULL,env=NULL) {
+test = function(num, x, y=TRUE,
+                error=NULL, warning=NULL, message=NULL, output=NULL, notOutput=NULL, ignore.warning=NULL,
+                options=NULL, env=NULL,
+                context=NULL) {
   if (!is.null(env)) {
     old = Sys.getenv(names(env), names=TRUE, unset=NA)
     to_unset = !lengths(env)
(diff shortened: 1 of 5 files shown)


R/test.data.table.R
```diff
@@ -570,6 +573,9 @@ test = function(num,x,y=TRUE,error=NULL,warning=NULL,message=NULL,output=NULL,no
     }
     # nocov end
   }
+  if (fail && !is.null(context)) {
+    catf("Test context: %s\n", context)
```
MichaelChirico:
```suggestion
    catf("Test context: %s\n", context) # nocov
```

jangorecki:
I suspect froll.Rraw will also have conflicts

jangorecki (approved):

Issue #5747, opened 2023-11-21
manual `\item` must not use braces inside `\itemize` and `\enumerate`


jangorecki:
> CHANGES IN R-devel UTILITIES
> When checking Rd files, ‘R CMD check --as-cran’ now notes some of the “lost braces” that ‘tools::checkRd()’ finds. Typical problems are Rd macros missing the initial backslash (e.g., ‘code{...}’), in-text set notation (e.g., ‘{1, 2}’, where the braces need escaping), and \itemize lists with _description_-like entries of the form \item{label}{description}.

Fix already submitted (vs master branch) inside #5746, thanks to @mmaechler 

We need to include it in 1.14.10 (branched from 1.14.8) coming patch release as well.

We need to ensure it is included in CI because multiple of pending PRs uses `\item{}` inside `\itemize`

mmaechler:
You are welcome.  I think you should edit the title (and part of the above):
In `Rd`, as in LaTeX, 
- `\item{.}` *must* be used in some cases (in LaTeX inside `\description{ .. }`), in Rd, inside `\arguments{ ..}`  and `\describe{ .. }`  *and* inside `\value{..}` where it has the special meaning for list-like results
- but makes *no* sense  (and may be misleading) inside  `\itemize{..}` and `\enumerate{..}`

PR #7202, opened 2025-07-22
Revise ?mergelist Details re. join-from/join-to #7190

### LINKED ISSUE #7190: Revise explanation of 'join-to'/'join-from' in `mergelist` documentation
As far I can see `mergelist` works (beautifully!) as one would hope/expect: it is (or at least resembles) a recursive `x[i]` join across a list of tables, where the column order reflects the left-to-right order of the tables. In other words each merge is at least analogous to:

`mergelist(list(LHS,RHS), how="right", ...)` <---> `LHS[i=RHS, ...)]`

`mergelist(list(LHS,RHS), how="left", ...)`  <---> `RHS[i=LHS, j=<LHS-then-RHS-cols>, ...)]`

All other cases (`how`s) are like `how="left"`. I'm abstracting here from details like having to complete the full join with `how="full"`, and the fact that `mult` seems to have some specialised behaviour in inner/outer joins (see note at the end of the documentation).

This basic description of things (even if it's only an analogy to whatever the internals are) accounts for column and row order, the "direction of action" of `mult`, and the `key` that gets used with implicit `on`.

```
library(data.table)

# ________________________
# row and column order
# ________________________
LHS <- data.table(id=1:3)[, a:=paste0("LHS",.I)]
RHS <- data.table(id=4:2)[, b:=paste0("RHS",.I)]
LHS
RHS

# all TRUE
identical(
  mergelist(list(LHS,RHS), on="id", how="left"),
  RHS[LHS, on="id", c("id","a","b")]
)
identical(
  mergelist(list(LHS,RHS),on="id",how="inner"),
  RHS[LHS, on="id", c("id","a","b"), nomatch=NULL]
)
identical(
  mergelist(list(LHS,RHS), on="id", how="right"),
  LHS[RHS, on="id"]
)

# ________________________
# mult
# ________________________
LHSm <- data.table(id=rep(1:3, each=2))[, a:=paste0("LHS",.I)][]
RHSm <- data.table(id=rep(4:2, each=2))[, b:=paste0("RHS",.I)][]
LHSm
RHSm

# all TRUE
identical(
  mergelist(list(LHSm,RHSm), on="id", mult="first", how="left"),
  RHSm[LHSm, on="id", c("id","a","b"), mult="first"]
)
identical(
  mergelist(list(LHSm,RHSm), on="id", mult="first", how="right"),
  LHSm[RHSm, on="id", mult="first"]
)

# ________________________
# key, implicit `on`
# ________________________
LHSk <- setkey(copy(LHS), "id")

# errors (DT/join-to not keyed)
try(mergelist(list(LHSk,RHS), how="left"))
try(RHS[LHSk, c("id","a","b")])

# TRUE
identical(
  mergelist(list(LHSk,RHS), on="id", mult="first", how="right"),
  LHSk[RHS]
)

```
## Current documentation and "join-from"/"join-to"

The documentation uses "join-from" and "join-to" to refer to (the equivalents of) `i` and `x` respectively. I have some worries about this. First, to a data.table user this seems like obfuscation - why not just say `i` and `x` (or e.g. the "`i`-equivalent" and "`x`-equivalent" tables)? Second, to an outside user (who might be dipping into data.table precisely to use `mergelist`), there is a risk that these terms will mean nothing at all and just end up being mysterious intermediate variables that obscure the documentation. Why can't we just say what goes on (with respect to row order, direction of `mult`, `on`), pointing out in each case that things are the other way round in a right-join? Also, for me at least, the terms don't make sense: in my mind it is `i` that is joined to from `DT` (because `i` presents an ordered list of demands that `DT` supplies where it can). This is not worth arguing about because it's so subjective, but the point is that neither the terms themselves nor the words "baseline" or "authoritative" can hope to convey very much.

I'm also puzzled about the content of the Details section that explains this distinction:

> The terms _join-to_ and _join-from_ indicate which in a pair of tables is the "baseline" or "authoritative" source – this governs the ordering of rows and columns. Whether each refers to the "left" or "right" table of a pair depends on the `how` argument:
>
> 1. `how %in% c("left", "semi", "anti")`: _join-to_ is _RHS_, _join-from_ is _LHS_.
> 2. `how %in% c("inner", "full", "cross")`: _LHS_ and _RHS_ tables are treated equally, so that the terms are interchangeable.
> 3. `how == "right"`: _join-to_ is _LHS_, _join-from_ is _RHS_.

This clearly says that the identity of the "join-from"/"join-to" tables determines column and row order in `mergelist`. But

- with respect to column order, it doesn't, because column order is always LHS-then-RHS (see above).
- with respect to row order, case 2 is logically impossible . `i` retains its row order because it is iterated over, but `DT` will only do so if:
  - the join columns have the same order in both tables
  - a row of `i` never joins with multiple rows of `DT` (we mustn't have a pattern of `DT`'s row indices like 1,2,3,2,3,4,...)
  - in a full join, the value of the join columns in all of `DT`'s non-matching rows (which get appended at the bottom) is above the maximum in any of `i`s rows

In fact, inner, full and cross joins behave like `how="left"` (*LHS* as the "join-from"/`i`-equivalent table).

## Proposed improvement

If I have got all this right (and I may well be missing something), my suggestion is:

- Eliminate "join-from"/"join-to" and just say what happens in relation to implicit `on`, row order, and `mult` in `how="right` vs the other cases (after fixing the problems I've noted). This directly documents how `mergelist` works in these respects without these baffling intermediate variables.

- Add a note for the benefit of data.table users explicitly spelling out the connection (or analogy, whichever it is) between `mergelist` and `x[i]`.

I think this documentation would be much clearer and more direct to both "audiences".

Delete the "join-from"/"join-to" section in Details and replace with:

> The columns of the result reflect the left-right order of the tables in `l`, i.e. columns of *LHS* are placed before columns of *RHS* in the output of each merge. The rows reflect the row orders of *LHS* then *RHS* in all cases except `how="right"`, where they reflect *RHS* then *LHS*.

Then remove all references to "join-from"/"join-to" as follows:

- Definition of `on`:

> `on`  character vector of column names to merge on; when missing, the `key` of the *RHS* table is used (*LHS* table if `how="right"`).

- Definition of `mult` (first sentence):

> mult  Character string, controls how to proceed when a row in the *LHS* table encounters multiple matching rows in the *RHS* table (vice versa with `how="right"`). 

Details re. `mult`:

> Using `mult="error"` will throw an error when a row in the *LHS* table encounters multiple matching rows in the *RHS* table (vice versa with `how="right"`).

(NB I've slightly altered the baseline wording of these two.)

Add a separate section, something like:

> Users familiar with data.table `x[i]` join syntax may find it helpful to think of each merge in `mergelist` as being equivalent to `LHS[RHS, ...]` when `how="right"`, and `RHS[LHS, j=<LHS-then-RHS-cols>, ...]` in all other cases. The row order, direction of action of `mult`, and the `key` used in implicit `on`, all reflect the "usual rules" of `x[i]`. Please note, however, that `mergelist` differs from `x[i]` joins in the specific behaviour of `mult`.

trobx:
closes #7190 @MichaelChirico @jangorecki

### DIFF
--- a/man/mergelist.Rd
+++ b/man/mergelist.Rd
@@ -26,17 +26,22 @@
 
   Merging is performed sequentially from "left to right", so that for \code{l} of 3 tables, it will do something like \code{merge(merge(l[[1L]], l[[2L]]), l[[3L]])}. \emph{Non-equi joins} are not supported. Column names to merge on must be common in both tables on each merge.
 
-  Arguments \code{on}, \code{how}, \code{mult}, \code{join.many} could be lists as well, each of length \code{length(l)-1L}, to provide argument to be used for each single tables pair to merge, see examples.
+  Arguments \code{on}, \code{how}, \code{mult}, \code{join.many} may also be lists, each of length \code{length(l)-1L}, providing the argument to be used at each successive merge; see examples.
 
-  The terms \emph{join-to} and \emph{join-from} indicate which in a pair of tables is the "baseline" or "authoritative" source -- this governs the ordering of rows and columns. 
+  Heuristically speaking, the \emph{join-from} table searches the \emph{join-to} table. More precisely:
+  \itemize{
+    \item{ \code{mult} determines the policy when a row of \emph{join-from} finds multiple matches in \emph{join-to}. }
+    \item{ When \code{on} is missing, the key of \emph{join-to} is used as the join column(s). }
+  }
   Whether each refers to the "left" or "right" table of a pair depends on the \code{how} argument:
   \enumerate{
-    \item{ \code{how \%in\% c("left", "semi", "anti")}: \emph{join-to} is \emph{RHS}, \emph{join-from} is \emph{LHS}. }
-    \item{ \code{how \%in\% c("inner", "full", "cross")}: \emph{LHS} and \emph{RHS} tables are treated equally, so that the terms are interchangeable. }
-    \item{ \code{how == "right"}: \emph{join-to} is \emph{LHS}, \emph{join-from} is \emph{RHS}. }
+    \item{ \code{how \%in\% c("left", "semi", "anti")}: \emph{join-from} is \emph{LHS}, \emph{join-to} is \emph{RHS}. }
+    \item{ \code{how == "right"}: \emph{join-from} is \emph{RHS}, \emph{join-to} is \emph{LHS}. }
+    \item{ \code{how \%in\% c("inner", "full")}: \emph{LHS} and \emph{RHS} are treated equally, so that each is both \emph{join-from} and \emph{join-to}. }
+    \item{ \code{how == "cross"}: \code{mult} must be \code{"all"} and \code{on} is not used, so the terms are not relevant. }
   }
 
-  Using \code{mult="error"} will throw an error when multiple rows in \emph{join-to} table match to the row in \emph{join-from} table. It should not be used just to detect duplicates, which might not have matching row, and thus would silently be missed.
+  Using \code{mult="error"} will throw an error when a row in the \emph{join-from} table finds multiple matching rows in the \emph{join-to} table. It should not be used just to detect duplicates in \emph{join-to}, as these might not have a matching row in \emph{join-from}, and thus silently be missed.
 
   When not specified, \code{mult} takes its default depending on the \code{how} argument:
   \enumerate{
@@ -45,10 +50,10 @@
     \item{ When \code{how == "cross"}, \code{mult="all"}. }
   }
 
-  When the \code{on} argument is missing, it will be determined based \code{how} argument:
-  \enumerate{
-    \item{ When \code{how \%in\% c("left", right", "semi", "anti")}, \code{on} becomes the key column(s) of the \emph{join-to} table. }
-    \item{ When \code{how \%in\% c("inner", full")}, if only one table has a key, then this key is used; if both tables have keys, then \code{on = intersect(key(lhs), key(rhs))}, having its order aligned to shorter key. }
+  Symmetrical \emph{join-from}/\emph{join-to} treatment of \emph{LHS} and \emph{RHS} when \code{how \%in\% c("inner", "full")} is as follows:
+  \itemize{
+    \item{ When \code{mult \%in\% c("first", "last", "error")}, then at distinct each value of the join column(s) the rows joined are respectively first-to-first, last-to-last, and only-to-only (or else an error). }
+    \item{ If only one table has a key, then this key is used; if both tables have keys, then \code{on = intersect(key(lhs), key(rhs))}, having its order aligned to the shorter key. }
   }
 
   When joining tables that are not directly linked to a single table, e.g. a snowflake schema (see References), a \emph{right} outer join can be used to optimize the sequence of merges, see Examples.


trobx:
Quick explanation:

- quick heuristic metaphor about join-from searching join-to, but define by substance (relevance to mult and on)
- use join-from first in all phrasing for clarity, e.g. join-from finds matches in join-to
- repurpose the bit about `on` (since now partially already covered) to be about the symmetric treatment in inner/full (new bullet on "mutual `mult`")

No mention of column or row order at all, per Jan's comments.

man/mergelist.Rd
```diff
@@ -26,17 +26,22 @@
...
 
-  Arguments \code{on}, \code{how}, \code{mult}, \code{join.many} could be lists as well, each of length \code{length(l)-1L}, to provide argument to be used for each single tables pair to merge, see examples.
+  Arguments \code{on}, \code{how}, \code{mult}, \code{join.many} may also be lists, each of length \code{length(l)-1L}, providing the argument to be used at each successive merge; see examples.
 
-  The terms \emph{join-to} and \emph{join-from} indicate which in a pair of tables is the "baseline" or "authoritative" source -- this governs the ordering of rows and columns. 
+  Heuristically speaking, the \emph{join-from} table searches the \emph{join-to} table. More precisely:
+  \itemize{
+    \item{ \code{mult} determines the policy when a row of \emph{join-from} finds multiple matches in \emph{join-to}. }
+    \item{ When \code{on} is missing, the key of \emph{join-to} is used as the join column(s). }
```
MichaelChirico:
```suggestion
    \item{ When \code{on} is missing, the join column(s) are determined by the key of \emph{join-to}. }
```

man/mergelist.Rd
```diff
@@ -26,17 +26,22 @@
...
+  }
   Whether each refers to the "left" or "right" table of a pair depends on the \code{how} argument:
   \enumerate{
-    \item{ \code{how \%in\% c("left", "semi", "anti")}: \emph{join-to} is \emph{RHS}, \emph{join-from} is \emph{LHS}. }
-    \item{ \code{how \%in\% c("inner", "full", "cross")}: \emph{LHS} and \emph{RHS} tables are treated equally, so that the terms are interchangeable. }
-    \item{ \code{how == "right"}: \emph{join-to} is \emph{LHS}, \emph{join-from} is \emph{RHS}. }
+    \item{ \code{how \%in\% c("left", "semi", "anti")}: \emph{join-from} is \emph{LHS}, \emph{join-to} is \emph{RHS}. }
+    \item{ \code{how == "right"}: \emph{join-from} is \emph{RHS}, \emph{join-to} is \emph{LHS}. }
+    \item{ \code{how \%in\% c("inner", "full")}: \emph{LHS} and \emph{RHS} are treated equally, so that each is both \emph{join-from} and \emph{join-to}. }
```
MichaelChirico:
```suggestion
    \item{ \code{how \%in\% c("inner", "full")}: \emph{LHS} and \emph{RHS} are treated equally, so that the \emph{join-from}/\emph{join-to} designation is not so instructive. }
```

man/mergelist.Rd
```diff
@@ -45,10 +50,10 @@
     \item{ When \code{how == "cross"}, \code{mult="all"}. }
   }
 
-  When the \code{on} argument is missing, it will be determined based \code{how} argument:
-  \enumerate{
-    \item{ When \code{how \%in\% c("left", right", "semi", "anti")}, \code{on} becomes the key column(s) of the \emph{join-to} table. }
-    \item{ When \code{how \%in\% c("inner", full")}, if only one table has a key, then this key is used; if both tables have keys, then \code{on = intersect(key(lhs), key(rhs))}, having its order aligned to shorter key. }
+  Symmetrical \emph{join-from}/\emph{join-to} treatment of \emph{LHS} and \emph{RHS} when \code{how \%in\% c("inner", "full")} is as follows:
+  \itemize{
+    \item{ When \code{mult \%in\% c("first", "last", "error")}, then at distinct each value of the join column(s) the rows joined are respectively first-to-first, last-to-last, and only-to-only (or else an error). }
```
MichaelChirico: sentence not computing, PTAL

trobx: Thanks - inserted "distinct" in the wrong place. Will address.

man/mergelist.Rd
trobx: @MichaelChirico apologies - I did not want to commit this but to discuss. (I'm gitHub-inept and this is my first PR). Have changed it back in a new commit.

trobx: @MichaelChirico   The instructive bit is supposed to be that we have a blended/mutual case that `x[i]` doesn't, but evidently that isn't coming across. Please see what you think of the latest - trying to consolidate/nail the definition in one place.

man/mergelist.Rd
```diff
@@ -26,17 +26,28 @@
 
   Merging is performed sequentially from "left to right", so that for \code{l} of 3 tables, it will do something like \code{merge(merge(l[[1L]], l[[2L]]), l[[3L]])}. \emph{Non-equi joins} are not supported. Column names to merge on must be common in both tables on each merge.
 
-  Arguments \code{on}, \code{how}, \code{mult}, \code{join.many} could be lists as well, each of length \code{length(l)-1L}, to provide argument to be used for each single tables pair to merge, see examples.
+  Arguments \code{on}, \code{how}, \code{mult}, \code{join.many} may also be lists, each of length \code{length(l)-1L}, providing the argument to be used at each successive merge; see examples.
 
-  The terms \emph{join-to} and \emph{join-from} indicate which in a pair of tables is the "baseline" or "authoritative" source -- this governs the ordering of rows and columns. 
+  Heuristically speaking, the \emph{join-from} table searches the \emph{join-to} table. More precisely:
```
MichaelChirico: I guess it would help to point out to regular users of {data.table} that the familiar `x[i, on=...]` join has "join-from: i, join-to: x", WDYT?

jangorecki: by design it cannot be constant if we have `how` arg left/right, in [ we just swap tables to achieve that

trobx: If we mention it, it should be to point out the difference (@MichaelChirico I think is what you are saying). Maybe, after the "symmetry" point with `how="inner"`:

> Note the difference between `mergelist` with `how="inner"` and `x[i]` with `nomatch=NULL`, where only `i` is _join-from_ and only `x` is _join-to_.

Potentially that could help people twig why this different terminology is needed?

Btw explaining it as two intersections could be useful in e.g. a vignette, though too long for the documentation.

```
on <- intersect(key(RHS), key(LHS)) # (but also align order to the shorter key)
fintersect(
  RHS[LHS[, on=on, nomatch=NULL, mult="last", <lhs-cols-then-rhs-non-join-cols> ]
  LHS[RHS[, on=on, nomatch=NULL, mult="last"]
)
```
Though as the note at the bottom says, the second line would be more efficiently done as something like:

`RHS[LHS[, last(.SD), by=on], on=on, nomatch=NULL, mult="last", <lhs-cols-then-rhs-non-join-cols> ]`

man/mergelist.Rd
jangorecki: I think \item should not have { after it since recent version of R

man/mergelist.Rd
jangorecki: In inner join it is not relevant which is x or i

trobx: Yes exactly - that's the intended point

man/mergelist.Rd
aitap: Correct! I wonder why it's not caught by any of our `R CMD check`s:

```
> tools::checkRd('man/mergelist.Rd')
checkRd: (-3) mergelist.Rd:33: Lost braces
    33 |     \item{ \code{mult} determines the policy when a row of \emph{join-from} finds multiple matches in \emph{join-to}. }
       |          ^
checkRd: (-3) mergelist.Rd:34: Lost braces
    34 |     \item{ When \code{on} is missing, the join column(s) are determined by the key of \emph{join-to}. }
       |          ^
checkRd: (-3) mergelist.Rd:38: Lost braces
    38 |     \item{ \code{how \%in\% c("left", "semi", "anti")}: \emph{join-from} is \emph{LHS}, \emph{join-to} is \emph{RHS}. }
       |          ^
checkRd: (-3) mergelist.Rd:39: Lost braces
    39 |     \item{ \code{how == "right"}: \emph{join-from} is \emph{RHS}, \emph{join-to} is \emph{LHS}. }
       |          ^
checkRd: (-3) mergelist.Rd:40: Lost braces
    40 |     \item{ \code{how \%in\% c("inner", "full")}: \emph{LHS} and \emph{RHS} are treated symmetrically, so that each is both \emph{join-from} and \emph{join-to}; see below. }
       |          ^
checkRd: (-3) mergelist.Rd:41: Lost braces
    41 |     \item{ \code{how == "cross"}: \code{mult} must be \code{"all"} and \code{on} is not used, so the terms are not relevant. }
       |          ^
checkRd: (-3) mergelist.Rd:46: Lost braces
    46 |     \item{ When \code{mult \%in\% c("first", "last", "error")}, then (respectively) the first, last, or only matching row on each side binds with the same on the other. \code{mult} is satisfied mutually and the merge is one-to-one. }
       |          ^
checkRd: (-3) mergelist.Rd:47: Lost braces
    47 |     \item{ If only one table has a key, then this key is used; if both tables have keys, then \code{on = intersect(key(lhs), key(rhs))}, having its order aligned to the shorter key. }
       |          ^
checkRd: (-3) mergelist.Rd:54: Lost braces
    54 |     \item{ When \code{how \%in\% c("left", "inner", "full", "right")}, \code{mult="error"}. }
       |          ^
checkRd: (-3) mergelist.Rd:55: Lost braces
    55 |     \item{ When \code{how \%in\% c("semi", "anti")}, \code{mult="last"}, although this is equivalent to \code{mult="first"}. }
       |          ^
checkRd: (-3) mergelist.Rd:56: Lost braces
    56 |     \item{ When \code{how == "cross"}, \code{mult="all"}. }
       |          ^
```

(Answer: this is a weak problem of severity "-3", and `R CMD check` prints errors at severity "-1" and above by default.)

jangorecki: related issue: https://github.com/Rdatatable/data.table/issues/5747

PR #7203, opened 2025-07-22
New flow for dropping a CRAN release into a branch

### LINKED ISSUE #6837: Fix up CI so that CRAN release can be pushed as a PR
In #6829 @TysonStanley and I discussed pushing the as-submitted CRAN release to GitHub to allow other Committers to merge it once the update is accepted by CRAN.

That would help minimize the time between a CRAN acceptance and an update here, without requiring the CRAN maintainer to operate with 24/7 availability.

However, we do have this note in the CRAN release today:

https://github.com/Rdatatable/data.table/blob/299d44d54f096d815e19fc6bf31a279c74befb2f/.dev/CRAN_Release.cmd#L601

IINM, we should be able to update the CI configs so that certain branches/rules _don't_ trigger the generation of binaries, etc. & thus avoid the concern mentioned here.

E.g., the rule could be "push to a branch specifically named `cran-release`", and our configs actively exclude that specific name.

MichaelChirico:
Closes #6837

I hope this captures the discussion well.

### DIFF
--- a/.dev/CRAN_Release.cmd
+++ b/.dev/CRAN_Release.cmd
@@ -584,8 +584,7 @@ bunzip2 inst/tests/*.Rraw.bz2  # decompress *.Rraw again so as not to commit com
 # Many thanks!
 # Best, Tyson
 # ------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


TysonStanley (approved):
LGTM

PR #7205, opened 2025-07-23
Trigger error not only on verbose in fread

### LINKED ISSUE #5252: fread can error if not verbose
Taking a look at #4029 brought my attention to https://github.com/Rdatatable/data.table/blob/e88826e94127991754a8821ac8149d15f1f9c3c2/src/fread.c#L1975-L1986

Code errors on a slightly modified example from #4029

``` r
tmp <- tempfile(fileext = ".csv")
fwrite(iris, tmp, col.names=FALSE)
fread(tmp, nrows=0)
#> Error in fread(tmp, nrows = 0): Internal error: sampleLines(1) > allocnrow(0)
```

Bug was introduced by https://github.com/Rdatatable/data.table/commit/a00d43f0fc7ac060b2c0c960145e2a05cca3cce4#

Possibly related to #4355

ben-schwen:
Closes #5252

But couldn't come up with an example to trigger it

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -2153,9 +2153,8 @@ int freadMain(freadMainArgs _args)
       DTPRINT(_("  Initial alloc = %"PRId64" rows (%"PRId64" + %d%%) using bytes/max(mean-2*sd,min) clamped between [1.1*estn, 2.0*estn]\n"),
               allocnrow, estnrow, (int)(100.0 * allocnrow / estnrow - 100.0));
       DTPRINT("  =====\n"); // # notranslate
-    } else {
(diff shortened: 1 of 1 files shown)


aitap:
I think that `allocnrow` is always at least `1.1 * ceil(bytesRead / meanLineLen)` = $1.1 \lceil \frac{\mathtt{eof} - \mathtt{firstRowStart}}{\mathtt{sumLen}} \mathtt{sampleLines} \rceil$. Indeed, it shouldn't be possible to sample more bytes (`sumLen`) than there is in the whole text file (`eof - firstRowStart`) without overcounting, so this internal error should be unreachable.

PR #7206, opened 2025-07-23
removed python compatibility

### LINKED ISSUE #7069: Remnants of compatibility with python datatable library dilute code
There are still old code paths that relate to shared files between this project and (I believe) a python project called `datatable`.

This can be seen in macros like `DTPY` which no longer do anything.

If compatibility is well and truly dead, I say we get rid of these.

badasahog:
closes #7069

There's a lot I can do from here to remove static/global objects and speed up the code.

More to come.


MichaelChirico:
Thanks! I will allow this to dwell for a while (say 3-6 months after the issue was posted) in case the original authors/pydatatable maintainers decide to weigh in

badasahog:
I think it can be confirmed by just looking at the code? It's pretty clear from the datatable codebase that the paths have diverged for some time now.

MichaelChirico:
I'm not sure if there's another repo (possibly private?) that has been split off

ben-schwen (approved):
LGTM TY

tdhock:
Hi! Is the following text (currently on https://r-datatable.com/articles/datatable-importing.html) now out of date since this PR was merged?

Importing from non-r Applications

Some tiny parts of data.table C code were isolated from the R C API and can now be used from non-R applications by linking to .so / .dll files. More concrete details about this will be provided later; for now you can study the C code that was isolated from the R C API in [src/fread.c](https://github.com/Rdatatable/data.table/blob/master/src/fread.c) and [src/fwrite.c](https://github.com/Rdatatable/data.table/blob/master/src/fwrite.c).

ben-schwen:
> Hi! Is the following text (currently on https://r-datatable.com/articles/datatable-importing.html) now out of date since this PR was merged?

Nope. This part of the vignette refers to [inst/include/datatableAPI.h](https://github.com/Rdatatable/data.table/blob/master/inst/include/datatableAPI.h)

Not sure if @eddelbuettel or others are still using this API

eddelbuettel:
I do not see a use that would break.  The header file was always an idea to demonstrate how other packages could use parts of the inner API of data.table when it is cleanly exposed, as the subsetter is here.   This was slow moving, nothing was ever built that relies on it, and nothing was added to the demo API either.  I would find removing it quite sad, but this is not my call to make.

aitap:
"Importing data.table C routines" is still valid; "Importing from non-r
Applications" is not. Both fread.c and fwrite.c now need R headers to
compile.

PR #7208, opened 2025-07-24
NEWS entry for the `between` `integer64` fix

aitap:
Add the missing news entry for #7193, since it fixed a bug report, #7164.

If a behaviour is not documented anywhere, fails <del>in a forest</del> with no user to notice it and then is fixed, was it a bug? Should we file a documentation bug for `?data.table::between` so that it says that `integer64` arguments are accepted?


MichaelChirico:
> Should we file a documentation bug for ?data.table::between so that it says that integer64 arguments are accepted?

As a rule I think {data.table} is expected to support integer64 columns/inputs unless otherwise noted...

aitap:
Understood, thank you!

PR #7209, opened 2025-07-24
tests: avoid invalid UTF-8 warning

aitap:
It turns out that my test case from https://github.com/Rdatatable/data.table/pull/7138#discussion_r2193521976 causes a warning, but only if run under `test.data.table()`. The underlying cause is `options(encoding = "UTF-8")` being in effect for the duration of the test:
https://github.com/Rdatatable/data.table/blob/63339e1bf4a65b1821c0c5a7019270d34e63346e/R/test.data.table.R#L92-L94

We need to write a byte with the most significant bit set, and `file(encoding = getOption('encoding'))` opened by `writeLines()` trying to convert that byte from the current encoding to UTF-8 doesn't help. Avoid the warning by explicitly opening a file connection with `encoding=""` (no conversion) and marking the payload line as `"bytes"`.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21554,12 +21554,18 @@ test(2332.81, {M1[]=frev(M1); M1}, {M2[]=rev(M2); M2})
 test(2333, as.expression(data.table(a = 1))[["a"]], 1)
 
 # regression test for hexdigits subscript overrun (uint8_t wraps over 255, unsigned overflow is well defined in c)
-f = tempfile()
-ff = file(f, encoding = "")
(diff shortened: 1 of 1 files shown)


inst/tests/tests.Rraw
```diff
@@ -21555,7 +21555,9 @@ test(2333, as.expression(data.table(a = 1))[["a"]], 1)
 
 # regression test for hexdigits subscript overrun (uint8_t wraps over 255, unsigned overflow is well defined in c)
 f = tempfile()
-writeLines(c('a', rep('0x1.ffffp0', 10000L), '0x1.ff\x9fp0', rep('0x1.ffffp0', 20000L)), f)
+ff = file(f, encoding = "")
```
MichaelChirico, line 21558: might as well put the whole thing in `local()` to `on.exit()` closing & deleting.

Also a comment pointing to this PR would be good for posterity sake.

MichaelChirico (approved):

PR #7210, opened 2025-07-24
RFC: fail `test.data.table()` if non-test code produces warnings

aitap:
Related to #7209. I happened to notice the warning while looking at unrelated failures. Should `test.data.table()` explicitly go looking for warnings produced by code outside the `test()` calls and fail `R CMD check` because of them?

Pros: sometimes there's a good reason to fix the warning (like #7209).

Cons: sometimes warnings happen for stupid reasons outside our direct control. For example, a system that neither currently has a UTF-8 locale set nor has `en_US.UTF-8` installed will cause a warning between tests 2252.2 and 2253.01, because the `setlocale()` call will fail:

https://github.com/Rdatatable/data.table/blob/63339e1bf4a65b1821c0c5a7019270d34e63346e/inst/tests/tests.Rraw#L18619-L18623

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -205,14 +205,14 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
   }
   # nocov end
 
-  warnings = list()
+  env$warnings = list()
   err = try(
     withCallingHandlers(
       sys.source(fn, envir=env),
       warning=function(w) {
         # nocov start
         if (!silent && showProgress) print(w)
-        warnings <<- c(warnings, list(list(env$prevtest, toString(w))))
+        env$warnings = c(env$warnings, list(list(env$prevtest, toString(w))))
         invokeRestart("muffleWarning")
         # nocov end
       }
@@ -275,9 +275,9 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
     # important to stopf() here, so that 'R CMD check' fails
     # nocov end
   }
-  if (length(warnings)) {
+  if (length(env$warnings)) {
     # nocov start
-    warnings = rbindlist(warnings)
+    warnings = rbindlist(env$warnings)
     setnames(warnings, c("after test", "warning"))
     catf(
       ngettext(length(warnings),
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -290,7 +290,7 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
     warnings = rbindlist(env$warnings)
     setnames(warnings, c("after test", "warning", "calls"))
     catf(
-      ngettext(length(warnings),
+      ngettext(nrow(warnings),
         "Caught %d warning outside the test() calls:\n",
         "Caught %d warnings outside the test() calls:\n"
       ),
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -284,7 +284,7 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
         "Caught %d warning outside the test() calls:\n",
         "Caught %d warnings outside the test() calls:\n"
       ),
-      length(warnings)
+      nrow(warnings)
     )
     print(warnings)
     stopf("Tests succeeded, but non-test code caused warnings. Search %s for tests shown above.", names(fn))
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -213,10 +213,10 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
         # nocov start
         if (!silent && showProgress) print(w)
         env$warnings = c(env$warnings, list(list(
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Personally I like requiring warnings to be handled properly everywhere, I'm a fan of setting `options(warn=2)` in my own {testthat} suites, for example.

If it's really a stupid/spurious warning, just `suppressWarnings()` and move on.

That `setlocale()` example is a case in point -- we should maybe skip some tests where we assume we're allowed to change locales, but can't, that could have unanticipated results.

jangorecki (approved):

inst/tests/tests.Rraw
```diff
@@ -18615,12 +18615,17 @@ test(2252.2, dt[, let(b=2L)], error = "\\[ was called on a data.table.*not data.
...
   # Japanese multibyte characters require utf8. As of 2025, we're likely to be already running in a UTF-8 locale, but if not, try this setlocale() call as a last chance.
   # Unfortunately, there is no guaranteed, portable way of switching to UTF-8 US English.
-  if (!l10n_info()$`UTF-8`) Sys.setlocale('LC_CTYPE', "en_US.UTF-8")
-  on.exit(Sys.setlocale('LC_CTYPE', lc_ctype))
+  lc_newctype = suppressWarnings(Sys.setlocale('LC_CTYPE', lc_wantctype))
+  if (identical(lc_newctype, lc_wantctype)) {
+    on.exit(Sys.setlocale('LC_CTYPE', lc_ctype))
+    TRUE
+  } else FALSE
+}) {
```
aitap: Seeking suggestions for how to improve this test. The logic is as follows:
1. If `` l10n_info()$`UTF-8` ``, that's fine, proceed with the tests.
2. Otherwise, try to set the UTF-8 charset from the `en_US.UTF-8` locale.
   * If succeeded, use `on.exit()` to restore the locale afterwards and proceed with the tests.
3. Otherwise, skip the tests.

The current code works, but putting `if` _with side effects_ inside another `if` doesn't feel like a good idea.

R/test.data.table.R
```diff
@@ -205,7 +205,30 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
...
-  err = try(sys.source(fn, envir=env), silent=silent)
+  env$warnings = list()
+  err = try(
+    withCallingHandlers(
+      sys.source(fn, envir=env),
+      warning=function(w) {
+        # nocov start
+        if (!silent && showProgress) print(w)
+        env$warnings = c(env$warnings, list(list(
+          env$prevtest, toString(w),
```
MichaelChirico: why not `conditionMessage()`?

~should we give this object a bit more structure, e.g. name the elements for clarity?~ I see it's done below

R/test.data.table.R
```diff
@@ -205,7 +205,30 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
...
+          paste(
+            vapply_1c(sys.calls(), function(call) {
+              if (length(call) && is.name(call[[1]])) {
```
MichaelChirico: is `!length(call)` possible?

R/test.data.table.R
```diff
@@ -262,6 +285,21 @@ test.data.table = function(script="tests.Rraw", verbose=FALSE, pkg=".", silent=F
     # important to stopf() here, so that 'R CMD check' fails
     # nocov end
   }
+  if (length(env$warnings)) {
+    # nocov start
+    warnings = rbindlist(env$warnings)
+    setnames(warnings, c("after test", "warning", "calls"))
+    catf(
```
MichaelChirico, line 291: just `cat()` here, or `domain=NA`?

inst/tests/tests.Rraw
MichaelChirico: it's ugly, but it looks sound.

if anything, maybe we would just move it inside `test.data.table()` like `foreign`:

https://github.com/Rdatatable/data.table/blob/39e2875a17fdf1ef6fe05cf79f48ffd6613808de/R/test.data.table.R#L119-L125

Then we just do, say, `if (supports_en_utf) local(...)`.

It would also facilitate if this sort of check would make other tests easier, we don't need to duplicate this complicated logic more than once.

aitap: I think it would be simpler but no less reliable if we skipped these tests when an `enc2native()` check fails, but I opened #7217 because there are other tests in a similar position, outside the scope of this PR.

R/test.data.table.R
aitap: Now that I've looked into it, that shouldn't be possible, thanks for pointing this out. `sys.calls()` returns only `CTXT_FUNCTION` calls of type `LANGSXP`, which always have a non-zero length. (An empty one would be `NULL`, but that won't happen.)

MichaelChirico (approved):

aitap:
Should this perhaps be changed to fail in our CI but keep going otherwise, similar to how `RUN_ALL_DATATABLE_TESTS` lets us skip some tests? I think this PR gave us at least one CRAN deadline. (Edit: now two.)

PR #7212, opened 2025-07-25
Add forceDecimal parameter to fwrite

### LINKED ISSUE #2864: Need an argument in fwrite to force writing a decimal point to numeric columns
One downside of CSV format is that the columns do not preserve types. A numeric column that happens to be fully represented by integers is written as integer in CSV. Next time the file is read, the column becomes `integer` unless `colClasses` is specified to force reading the column as numeric. 

I'm wondering if it makes sense to provide an argument in `fwrite` to force numeric columns to end up with decimal point or numbers (e.g. `1` is written as `1.0`) so that next time the file is read, the type of the column is still numeric.

ben-schwen:
Closes #2864


inst/tests/tests.Rraw
```diff
@@ -21566,3 +21566,6 @@ test(2335.3, isoyear(as.IDate("2023-08-15")), 2023L)  # Normal mid-year case
 test(2335.4, isoyear(as.IDate(c("2019-12-30", "2016-01-01", "2023-08-15"))),c(2020L, 2015L, 2023L))
 test(2335.5, isoyear("2019-12-30"), 2020L)
 test(2335.6, isoyear(as.Date("2019-12-30")), 2020L)
+
+# 2864 force decimal points for whole numbers in numeric columns
+test(2336.1, capture.output(fwrite(data.table(x=c(1, 2, 3)), forceDecimal=TRUE)), c("x", "1.", "2.", "3."))
```
MichaelChirico: be sure to also test with a non-`.` decimal spec

MichaelChirico: also, be sure to test that `fread()` indeed understands these are double on the round-trip. (in case `fread()` has / might eventually consider fraction-less numeric column as integer)

src/fwrite.c
```diff
@@ -257,6 +264,11 @@ void writeFloat64(const void *col, int64_t row, char **pch)
         if (sf > dr) width = sf + 1;                     // 1.234 and 123.4
         else { dl0 = 1; width = dr + 1 + dl0; }            // 0.1234, 0.0001234
       }
+      bool isWholeNumber = (dr == 0 && dl0 >= 0);
```
MichaelChirico: `const`?

src/fwrite.c
```diff
@@ -198,6 +199,9 @@ void writeFloat64(const void *col, int64_t row, char **pch)
     }
   } else if (x == 0.0) {
     *ch++ = '0';   // and we're done.  so much easier rather than passing back special cases
+    if (forceDecimal) {
```
MichaelChirico, line 202: is there a way to use a flag like `needs_dec=false;` up-front to only add `*ch++=dec;` once just before returning? i was surprised to see we needed 4 similar invocations.

inst/tests/tests.Rraw
MichaelChirico: also test that integer columns do _not_ get the decimal under this option

src/fwrite.c
ben-schwen: But iinm we write backwards into the buffer, so adding at the end will place it after the whole field, not the last digit

man/fwrite.Rd
```diff
@@ -62,6 +63,7 @@ fwrite(x, file = "", append = FALSE, quote = "auto",
   \item{bom}{If \code{TRUE} a BOM (Byte Order Mark) sequence (EF BB BF) is added at the beginning of the file; format 'UTF-8 with BOM'.}
   \item{verbose}{Be chatty and report timings?}
   \item{encoding}{ The encoding of the strings written to the CSV file. Default is \code{""}, which means writing raw bytes without considering the encoding. Other possible options are \code{"UTF-8"} and \code{"native"}. }
+  \item{forceDecimal}{ Should decimal points be forced for whole numbers in numeric columns? When \code{TRUE}, whole numbers like \code{c(1.0, 2.0, 3.0)} will be written as \code{1., 2., 3.} instead of \code{1, 2, 3}. Default is \code{FALSE}. }
```
MichaelChirico: I guess the description needs to avoid assuming `dec='.'`. Also, `1., 2., 3.` is not valid R code, so we need `\samp`.

```suggestion
  \item{forceDecimal}{ Should decimal points be forced for whole numbers in numeric columns? When \code{FALSE}, the default, whole numbers like \code{c(1.0, 2.0, 3.0)} will be written as \samp{1, 2, 3} i.e., dropping \code{dec}. }
```

src/fwrite.c
MichaelChirico: OK, I guess it would require a broader refactoring to accomplish that... something to keep in mind...

src/fwrite.c
```diff
@@ -239,6 +243,9 @@ void writeFloat64(const void *col, int64_t row, char **pch)
     if (l == 0) {
       if (*(ch - 1) == '-') ch--;
       *ch++ = '0';
+      if (forceDecimal) {
+        *ch++ = dec;
```
MichaelChirico, line 248: this one isn't covered by tests yet

ben-schwen: this is not reachable from within R since everything below `2^-1074` is rounded to zero, `.Internal(inspect( 2^-1075))`. We might not even need the `*ch++='0';` but it feels safe to be future proof.

MichaelChirico: I see, that branch is already not covered:

image

src/fwrite.c
```diff
@@ -237,8 +241,13 @@ void writeFloat64(const void *col, int64_t row, char **pch)
     if (l % 10 >= 5) l += 10; // use the last digit to round
     l /= 10;
     if (l == 0) {
+      # nocov start. Very likely not needed as such numbers (e.g. 2^-1075) likely not representable in R.
```
MichaelChirico:
```suggestion
      // # nocov start. Very likely not needed as such numbers (e.g. 2^-1075) likely not representable in R.
```

src/fwrite.c
```diff
@@ -237,8 +241,13 @@ void writeFloat64(const void *col, int64_t row, char **pch)
...
       if (*(ch - 1) == '-') ch--;
       *ch++ = '0';
+      if (forceDecimal) {
+        *ch++ = dec;
+      }
+      # nocov end
```
MichaelChirico:
```suggestion
      // # nocov end
```

src/fwrite.c
ben-schwen: FWIW I also tried a log-linear search in pure C and didn't find a number to hit this

MichaelChirico: Maybe share that script here? and we could revisit dropping this branch later.

Proposal:

 1. Current release don't do anything
 2. Next release make this an `internal_error()` to see if we get any reports
 3. Subsequently drop the branch

PR #7214, opened 2025-07-25
Use a less generic name than 'init' for routine to initialize Inf & NaN

### LINKED ISSUE #4369: fread.c init() causes segfault
This visible symbol causes a segfault during fread in my environment- could it be changed to something like: 

static void freadInit(void)

https://github.com/Rdatatable/data.table/blob/b1b1832b0d2d4032b46477d9fe6efb15006664f4/src/fread.c#L78

MichaelChirico:


aitap:
I'm not sure that this should help. The problem from #4369 should no longer be happening on current `master` thanks to the function already being `static`. `data_table.so` exports a lot of symbols, but `init` isn't one of them, even at the object file level:

```r
fread(
 'nm -P fread.o', fill = TRUE,
 col.names = c('name', 'type', 'value', 'size')
)[type %in% c('B', 'D', 'R', 'T')][,
        type := fcase(
                type == 'B', 'variable',
                type == 'D', 'data',
                type == 'R', 'read-only data',
                type == 'T', 'function'
        )
][]
#                       name           type  value   size
#                     <char>         <char> <char> <char>
#  1:               copyFile       function   4600     69
#  2: cumDaysCycleMonthsLeap read-only data   19f0     18
#  3: cumDaysCycleMonthsNorm read-only data   1a20     18
#  4:      cumDaysCycleYears read-only data   13a0    644
#  5:           freadCleanup       function   28e0    20f
#  6:              freadMain       function   4670   6816
#  7:              hexdigits read-only data   1a60    100
#  8:           leapYearDays read-only data   1a08      c
#  9:           normYearDays read-only data   1a38      c
# 10:            pow10lookup read-only data     c0   12d0
# 11:               typeName read-only data     20     96
# 12:               typeSize           data      0      f
# 13:              wallclock       function   2af0     61
```

Still, the change is harmless, and maybe, if your environment uses link-time optimisation, it will help anyway due to the LTO process enforcing something like the one-definition rule a little over-zealously? We can try to investigate failures in your environment in more detail.

Here's what we can use to systematically avoid this kind of bugs: https://cran.r-project.org/doc/manuals/R-exts.html#Controlling-visibility

1. Add `$(C_VISIBILITY)` to `PKG_CFLAGS`
2. For the only function we want to export, `R_init_data_table`, add the `attribute_visible` annotation.
3. Use a `*.def` file on Windows, also exporting only one function.

All the other functions should be known to R thanks to them being registered, not by name through `dlsym()`. Not sure when it was introduced, but WRE for R-3.4.0 already mentions it.

MichaelChirico:
I agree there's no bug fix here, but init () _does_ seem inappropriately generic a name for this routine

aitap:
I agree, the new name is better.

PR #7222, opened 2025-07-28
turned quote rule into enumeration + minor formatting improvements

badasahog:


aitap:
You seem to have started from an earlier master, before #7214 and
#7205, because this PR reverts them. On the other hand, your branch
stems from cfa9f49bd27195962573ad493a31600d173abc5c (which is master as
of time of writing). Did a merge go wrong? Maybe we can help you fix
your process?

badasahog:
yikes, I know exactly what happened. Didn't realize this file was being played with.

src/fread.c
```diff
@@ -34,22 +34,33 @@ static char quote, dec;
...
+    // For example: <<...,"hello \"world\"",...>>
+    QUOTE_RULE_EMBEDDED_QUOTES_ESCAPED,
+
+    // Fields may be quoted, but any quotes inside will appear verbatim and
+    // not escaped in any way. It is not always possible to parse the file
+    // unambiguously, but we give it a try anyways. A quote will be presumed
+    // to mark the end of the field iff it is followed by the field separator.
+    // Under this rule eol characters cannot appear inside the field.
+    // For example: <<...,"hello "world"",...>>
+    QUOTE_RULE_HYBRID,
```
MichaelChirico: HYBRID is not quite evocative to me, maybe `QUOTE_RULE_EMBEDDED_QUOTES_NOT_ESCAPED` or just `QUOTE_RULE_NOT_ESCAPED` is better?

src/fread.c
```diff
@@ -34,22 +34,33 @@ static char quote, dec;
...
+
+    // Fields are not quoted at all.Any quote characters appearing anywhere
```
```suggestion
    // Fields are not quoted at all. Any quote characters appearing anywhere
```

src/fread.c
```diff
@@ -34,22 +34,33 @@ static char quote, dec;
...
+    // inside the field will be treated as any other regular characters.
+    // Example: <<...,hello "world",...>>
+    QUOTE_RULE_IGNORE_QUOTES,
+
+    QUOTE_RULE_COUNT
+};
+
+static enum quoteRule quoteRule;
```
MichaelChirico: This is a bit confusing, should it be `typedef` instead, or perhaps the `enum` should have a slightly different name like `enum quoteRuleEnum { ... }; static enum quoteRuleEnum quoteRule;`?

src/fread.c
```diff
@@ -75,10 +86,10 @@ static freadMainArgs args = { 0 };  // global for use by DTPRINT; static implies
 
 // See header for more explanation.
 const char typeName[NUMTYPE][10] = { "drop", "bool8", "bool8", "bool8", "bool8", "bool8", "bool8", "int32", "int64", "float64", "float64", "float64", "int32", "float64", "string" };
-int8_t     typeSize[NUMTYPE]     = { 0,      1,       1,       1,       1,       1,       1,       4,       8,       8,         8,         8,         4,       8       ,  8      };
+int8_t     typeSize[NUMTYPE]     = { 0,      1,       1,       1,       1,       1,       1,       4,       8,       8,         8,         8,         4,       8,         8        };
 
 // In AIX, NAN and INFINITY don't qualify as constant literals. Refer: PR #3043
-// So we assign them through below init_const_literals function.
+// So we assign them through below init function.
```
MichaelChirico: revert

src/fread.c
```diff
@@ -1280,7 +1291,7 @@ static int detect_types(const char **pch, int ncol, bool *bumped)
           dec = '\0';
         }
       }
-      while (++tmpType[field] < CT_STRING && disabled_parsers[tmpType[field]]) {};
+      while (++tmpType[field] < CT_STRING && disabled_parsers[tmpType[field]]);
```
MichaelChirico: IMO it's slightly clearer to leave the `{}`, without it we have to read more carefully about whether the following `*bumped=true;` is part of the same statement

badasahog: I disagree, but you're the boss. Should I revert?

I think the best option would be a do/while loop with `++tmpType[field]` in the body, if you're open to that.

src/fread.c
badasahog: yeah, missed that one. my bad.

src/fread.c
MichaelChirico: Yes, revert. We can discuss alternatives in a separate/more focused PR.

MichaelChirico (approved):

PR #7225, opened 2025-07-28
Follow strict Rd rules

MichaelChirico:
Follow-up to #7218 to fix the issues raised by strict `checkRd()`.

Example log with issues: https://gitlab.com/Rdatatable/data.table/-/jobs/10826193197


tdhock:
great thanks Michael

PR #7226, opened 2025-07-28
move timestamps to dedicated structure to be more organized

badasahog:
This is a work-in-progress pr.

For some reason it won't compile (relating to an openmp pragma).

If anyone knows what this is about, I would love to be enlightened.


src/fread.c
```diff
@@ -2384,8 +2396,10 @@ int freadMain(freadMainArgs _args)
         stopTeam = true;
       }
       prepareThreadContext(&ctx);
-    
-      #pragma omp for ordered schedule(dynamic) reduction(+:thRead,thPush) reduction(max:max_col)
+      
+      double th_read_openmp = timestamps.th_read;//pragma workaround
+      double th_push_openmp = timestamps.th_push;
+      #pragma omp for ordered schedule(dynamic) reduction(+:th_read_openmp,th_push_openmp) reduction(max:max_col)
```
Anirban166: > For some reason it won't compile (relating to an openmp pragma).
>
> If anyone knows what this is about, I would love to be enlightened.

The reduction clause here is being applied to local variables that are private to each thread, not shared variables.

src/fread.c
```diff
@@ -2609,7 +2623,7 @@ int freadMain(freadMainArgs _args)
           if (tch != eof) tch++;
           myNrow++;
         }
-        if (verbose) { double now = wallclock(); thRead += now - tLast; tLast = now; }
+        if (verbose) { double now = wallclock(); timestamps.th_read += now - tLast; tLast = now; }
```
Anirban166, line 2626: And even if it did compile, this change would create a mess as `timestamps.th_read` is shared by all threads. Inside the parallel loop, multiple threads would be trying to execute this line, reading and writing to the *same* shared variable at the *same* time. The reduction was supposed to prevent this race condition, but with this change it's being applied to the wrong variables.

Anirban166: And even if it did compile, this would create a mess as `timestamps.th_read` will then be shared by all threads. 

Inside the parallel loop, multiple threads would be trying to execute this line, reading and writing to the *same* shared variable at the *same* time. The reduction was supposed to prevent this race condition, but with this change it's being applied to the wrong variables.

badasahog:
I got it to at least compile, but I still don't get why thRead and thPush can't be put in a struct.

A problem for a future pr, I suppose.

MichaelChirico:
I'm not sure I see the point of this change

badasahog:
organization and clarity.

the purpose of an object like `tMap` isn't obvious, nor is it obvious why it's at function scope.

`timestamps.map` is much more clear and self documenting.

badasahog:
are you willing to consider merge? or is this DOA?

MichaelChirico:
Could you describe how you fixed the race condition issue? It's quite off-putting to evaluate a low-value-add change that touches thread-sensitive code.

badasahog:
I just reverted back to how it was before. This sort of thing is very finnicky, I understand your concern.

PR #7231, opened 2025-07-31
switched to memset

badasahog:
This was moved from a separate PR.

Funny enough, I thought there would be more cases like this (hence the branch name) but apparently not.


MichaelChirico (approved):
Great!

MichaelChirico (changes requested):
Please double-check -- this appears to be triggering a segfault.

badasahog:
It appears the problem has been resolved

MichaelChirico (approved):

PR #7236, opened 2025-08-04
Add select parameter to fwrite() to avoid temporary objects

### LINKED ISSUE #4177: fwrite to gain select column(s) argument without creating a temporary object
I am on my last step of my R session, where I got a huge data.table object to be passed on to my 2 clients in CSV format. The 2 clients only need several columns of my data.table object. But, I am running tight on RAM and couldn't afford to create another R object, I also can't delete the unnecessary variable as the other client need it. 

Currently, my implementation is:
```
set.seed(1)
dt <- data.table(
  a = rnorm(1e8),
  b = runif(1e8),
  c = sample(letters,1e8,T)
)
fwrite(dt)
fwrite(dt[,.(a,c)]) # data requested by the first client
fwrite(dt[,.(b,c)]) # data requested by the second client
```
From my inspection and understanding, the second and the third fwrite, will create a throwaway object that requires additional RAM. Is it possible to have select argument for fwrite that doesn't require such temporary object? Something like, `fwrite(dt,select=c("a","c")`?

`#` `Output of sessionInfo()`
```
R version 3.6.1 (2019-07-05)

Matrix products: default

locale:
[1] LC_COLLATE=English_Singapore.1252  LC_CTYPE=English_Singapore.1252   
[3] LC_MONETARY=English_Singapore.1252 LC_NUMERIC=C                      
[5] LC_TIME=English_Singapore.1252    

attached base packages:

other attached packages:
[1] data.table_1.12.9

[1] compiler_3.6.1 tools_3.6.1    curl_4.2
```

Mukulyadav2004:
closes #4177 

Problem: 
When writing only specific columns requires creating temporary objects
`fwrite(DT[,.(col1,col3)])`  

Solution:
Add select parameter that uses .shallow() for data.table objects to create shallow copy without data duplication.

`fwrite(DT, select = c("col1", "col3"))`

@MichaelChirico @tdhock please review.


Mukulyadav2004:
@MichaelChirico can I add # nocov to this line - `x = x[select]` as this is an edge case and only executes when select parameter is provided, and input x is a data.frame.

R/fwrite.R
```diff
@@ -13,7 +13,8 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
            yaml = FALSE,
            bom = FALSE,
            verbose=getOption("datatable.verbose", FALSE),
-           encoding = "") {
+           encoding = "",
+           select) {
```
jangorecki: I would use default value as NULL, same as fread

jangorecki (commented):
Needs manual update and NEWS entry

R/fwrite.R
```diff
@@ -26,6 +27,16 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
...
+
+  # Handle select argument using .shallow()
+  if (!null(select)) {
+    if (is.data.table(x)) {
+      cols = colnamesInt(x, select)
+      x = .shallow(x, cols)
+    } else {
+      x = x[select]
+    }
+  }
```
ben-schwen: Is the distinction between `data.table` and non data.table the best one here, since `colnamesInt` basically also work with `data.frames` and `list`.

Moreover, we also need to handle matrix input!

tdhock:
also would be good to add an atime performance test so we can see the reduction in memory usage.

R/fwrite.R
Mukulyadav2004: Sure, I've placed our select logic after matrix conversion to `data.table` thus it should work for matrix input also.

Mukulyadav2004:
I surely will. But can you please tell where should `atime` test should be placed - .ci/atime/tests.R ?

inst/tests/tests.Rraw
```diff
@@ -21602,3 +21602,10 @@ if (getRversion() >= "4.0.0") { # rely on stopifnot(named = ...) for correct mes
 test(2337.3, is.null(fwrite(data.table(c(0.1, 0.2)), dec=",", sep="\t")))
 test(2337.4, is.null(fwrite(data.table(a=numeric(), b=numeric()), dec=",", sep=",")))
 test(2337.5, is.null(fwrite(data.table(a=numeric()), dec=",", sep=",")))
+
+# test for select parameter #4177
+DT = data.table(a=1:2, b=3:4)
```
ben-schwen, line 21624: please add tests for other classes e.g.
```r
df = as.data.frame(DT)
l = as.list(DT)
m = as.matrix(DT)
```

Mukulyadav2004:
Hi @tdhock I've applied all the changes as per suggestion, now it's ready for review from my side , please have a look.

tdhock:
please fix CI  failures

Mukulyadav2004:
Could you please guide to which ci check failure your are referring

MichaelChirico:
image

.ci/atime/tests.R
```diff
@@ -286,5 +286,19 @@ test.list <- atime::atime_test_list(
...
+    # Fixed in  https://github.com/Rdatatable/data.table/pull/7236
+    "fwrite select parameter improved in #4177" = atime::atime_test(
+    setup = {
+      set.seed(1)
+      DT = data.table(a = rnorm(N), b = rnorm(N), c = rnorm(N), d = rnorm(N), e = rnorm(N))
+      temp_file = tempfile()
+      has_select = "select" %chin% names(formals(data.table::fwrite))
+    },
+    expr = if (has_select)
+      data.table::fwrite(DT, temp_file, select=c("a","b","c")),
```
tdhock: this test case will not work as intended.

only expr is modified based on the data.table version. (not setup)

here is a modification in which I observe a big difference on windows. (in both time and memory, is that expected??)

```r
alist <- atime::atime_versions(
  pkg.path="~/R/data.table",
  seconds.limit=Inf,
  pkg.edit.fun = function(old.Package, new.Package, sha, new.pkg.path) {
    }
    pkg_find_replace(
      "DESCRIPTION",
    pkg_find_replace(
      Package_regex,
      new.Package_)
    pkg_find_replace(
      file.path("R", "onLoad.R"),
      Package_regex,
      new.Package_)
    pkg_find_replace(
      file.path("R", "onLoad.R"),
    pkg_find_replace(
      file.path("src", "init.c"),
    # allow compilation on new R versions where 'Calloc' is not defined
    pkg_find_replace(
      file.path("src", "*.c"),
      "\\b(Calloc|Free|Realloc)\\b",
      "R_\\1")
    pkg_find_replace(
      "NAMESPACE",
  },
  setup = {
    set.seed(1)
    DT = data.table(a = rnorm(N), b = rnorm(N), c = rnorm(N), d = rnorm(N), e = rnorm(N))
    temp_file = tempfile()
    select_vec <- c("a","b","c")
  },
  expr = {
    has_select = "select" %chin% names(formals(data.table::fwrite))
    if(has_select){
      data.table::fwrite(DT, temp_file, select=select_vec)
    }else{
      data.table::fwrite(DT[, select_vec, with=FALSE], temp_file)
    }
  },
  Slow = "66cb6d2393cef30083b444346a7600a079207806",    # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/66cb6d2393cef30083b444346a7600a079207806)
  Fast = "1887699fe965b5aa1fb8cb16b5507b7a5cbf5c85")
plot(alist)
```
image
We observe:

* Fast is constant memory and linear time
* Slow is linear time and memory (with a discontinuity between 1e2 and 1e3, why?)

tdhock (changes requested):

.ci/atime/tests.R
tdhock: versus current CI shows no difference
image

tdhock: this is a different test case:
```r
## write first column from DT with 1 row and N columns.
alist <- atime::atime_versions(
  pkg.path="~/R/data.table",
  pkg.edit.fun = function(old.Package, new.Package, sha, new.pkg.path) {
    }
    pkg_find_replace(
      "DESCRIPTION",
    pkg_find_replace(
      Package_regex,
      new.Package_)
    pkg_find_replace(
      file.path("R", "onLoad.R"),
      Package_regex,
      new.Package_)
    pkg_find_replace(
      file.path("R", "onLoad.R"),
    pkg_find_replace(
      file.path("src", "init.c"),
    # allow compilation on new R versions where 'Calloc' is not defined
    pkg_find_replace(
      file.path("src", "*.c"),
      "\\b(Calloc|Free|Realloc)\\b",
      "R_\\1")
    pkg_find_replace(
      "NAMESPACE",
  },
  setup = {
    DT = data.table(t(1:N))
    temp_file = tempfile()
    select_vec = "V1"
  },
  expr = {
    has_select = "select" %chin% names(formals(data.table::fwrite))
    if(has_select){
      data.table::fwrite(DT, temp_file, select=select_vec)
    }else{
      data.table::fwrite(DT[, select_vec, with=FALSE], temp_file)
    }
  },
  Slow = "66cb6d2393cef30083b444346a7600a079207806",    # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/66cb6d2393cef30083b444346a7600a079207806)
  Fast = "1887699fe965b5aa1fb8cb16b5507b7a5cbf5c85")
plot(alist)
```
image
I expected:

* using select arg (Fast) with one column name in a table of N entries should be constant time and memory
* using square brackets (Slow) invokes a copy, only 1 column though, so constant time and memory.

However we observe in above figure:

* Fast is actually linear time and space (why not constant?)
* Slow is constant space but linear time (why not constant?)

@Mukulyadav2004 can you investigate, explain, and fix please?

tdhock: perhaps the linear time and space of Fast can be explained by shallow() ?

Mukulyadav2004: Thank you @tdhock for the corrected atime_versions harness. I can reproduce your Windows deltas.
In the second case with single row and N columns - you are right to your point to expect constant time and memory but I find that it holds only if select is already numeric.
So (Fast) is linear time and space because select was provided as character. Resolving a name to an index is O(#cols) and allocates helper vectors. With 1 row, that O(#cols) dominates which make time grow with columns thus helper allocations show as memory growth.
And same also 'Slow' is constant space but linear time as `DT[, "V1", with=FALSE]` reuses the column vector, so memory is roughly constant. But it still performs the O(#cols) name lookup, so time is linear.


So name lookup cost is O(#cols) when using character select-
```
library(data.table)
for (p in 2:6) {
  N <- 10^p
  DT <- data.table(t(seq_len(N)))
  print(system.time(colnamesInt(DT, "V1"))["elapsed"])
}
```

But passing numeric indices avoids the O(#cols) name lookup-
```
idx <- 1L
system.time(idx)
```

So the fix I believe should be to accept numeric select directly thus skips O(#cols) name matching and avoid creating a shallow view when it’s a no‑op thus only build a shallow view when columns are actually reduced or reordered
```
if (length(cols) < NCOL(x) || !identical(cols, seq_len(NCOL(x)))) {
        x = .shallow(x, cols)
}
```

For case 1, the big windows gap is expected as `select=` keeps R memory constant as work is in C, while pre-subset allocates an R object proportional to the row count, hence memory and time are linear. The small step between ~1e2–1e3 rows is a Windows allocator threshold where the heap switches bucket strategy thus I would say it’s platform behavior rather than a correctness issue. Any thoughts about this?
I’ll also update the atime harness to branch only in expr, per your suggestion.

tdhock: ok thanks for the detailed analysis, that makes sense. 

if the windows allocator threshold is true then we should see different behavior when running that test case on github actions (ubuntu). 
```r
## N=number of rows, fwrite 1 column of DT with 1000 columns.
alist <- atime::atime_versions(
  pkg.path="~/R/data.table",
  pkg.edit.fun = function(old.Package, new.Package, sha, new.pkg.path) {
    }
    pkg_find_replace(
      "DESCRIPTION",
    pkg_find_replace(
      Package_regex,
      new.Package_)
    pkg_find_replace(
      file.path("R", "onLoad.R"),
      Package_regex,
      new.Package_)
    pkg_find_replace(
      file.path("R", "onLoad.R"),
    pkg_find_replace(
      file.path("src", "init.c"),
    # allow compilation on new R versions where 'Calloc' is not defined
    pkg_find_replace(
      file.path("src", "*.c"),
      "\\b(Calloc|Free|Realloc)\\b",
      "R_\\1")
    pkg_find_replace(
      "NAMESPACE",
  },
  setup = {
    N.col = 1e3
    M = matrix(0, N, N.col)
    DT = as.data.table(M)
    temp_file = tempfile()
    select_vec = 1
  },
  expr = {
    has_select = "select" %chin% names(formals(data.table::fwrite))
    if(has_select){
      data.table::fwrite(DT, temp_file, select=select_vec)
    }else{
      data.table::fwrite(DT[, select_vec, with=FALSE], temp_file)
    }
  },
  Slow = "66cb6d2393cef30083b444346a7600a079207806",    # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/66cb6d2393cef30083b444346a7600a079207806)
  Fast = "1887699fe965b5aa1fb8cb16b5507b7a5cbf5c85")
plot(alist)
```
on my ubuntu machine I get the result below:
image

* time is linear for both methods
* Slow is linear memory whereas Fast is constant, but the constant is larger, why?

in the test case where N is the number of columns, I tried changing select_vec from "V1" to 1 but I still observe linear memory usage for Fast, why?
image

Mukulyadav2004: Thanks @tdhock for providing the ubuntu run. As per I find, 'Fast' allocates a small constant-size I/O buffer in c, so it shows a constant but higher baseline. Slow’s baseline starts lower but grows linearly because DT[, sel, with=FALSE] materializes an N×1 temporary before fwrite.
And in case when N is column and select_vec is 1, current fwrite, select always goes through colnamesInt(x, select) and we build a shallow view unconditionally. Even when select is numeric (1), colnamesInt does O(#cols) allocations, which is why memory rises with N. So probably this is the reason for 'Fast' to still show linear memory.

tdhock: > we build a shallow view unconditionally. Even when select is numeric (1), colnamesInt does O(#cols) allocations, which is why memory rises with N. So probably this is the reason for 'Fast' to still show linear memory.

can this be avoided? it seems desirable to have constant time/memory in this case.

Mukulyadav2004: Ig yess! If we bypass name resolution for numeric select and only build a shallow view then time/memory become constant.

tdhock: ok good news thanks.
please commit those changes, add these two test cases, and resolve this conversation after.

Mukulyadav2004: So the key changes I understand after your suggestions are-
Directly access the column without using colnamesInt to avoid O(#cols) allocations. And only build  shallow view when columns are actually reduced or reordered.
Should I proceed with these?

vignettes/datatable-fread-and-fwrite.Rmd
```diff
@@ -270,14 +270,23 @@ if (requireNamespace("bit64", quietly = TRUE)) {
 
 ### 2.4 Column Order and Subset Control
 
-To control the order and subset of columns written to file, subset the data.table before calling `fwrite()`. The `col.names` argument in `fwrite()` is a logical (TRUE/FALSE) that controls whether the header row is written, not which columns are written.
+To control the order and subset of columns written to file, you can use `[.data.table` to make a new table before calling `fwrite()`, but it is more efficient to use the `select` argument, which avoids making a copy.
```
tdhock, line 273: @Mukulyadav2004 please proof-read and modify if you like.

R/fwrite.R
```diff
@@ -43,11 +43,17 @@ fwrite = function(x, file="", append=FALSE, quote="auto",
   }
   # Handle select argument using .shallow()
   if (!is.null(select)) {
-    cols = colnamesInt(x, select)
+    if (is.integer(select) || is.numeric(select)) {  # numeric/integer avoids O(#cols) name-match overhead
+        cols = as.integer(select)
+    } else {
+        cols = colnamesInt(x, select)
+    }
```
tdhock:
```suggestion
    cols = if (is.integer(select) || is.numeric(select)) {  # numeric/integer avoids O(#cols) name-match overhead
        as.integer(select)
    } else {
        colnamesInt(x, select)
    }
```

vignettes/datatable-fread-and-fwrite.Rmd
Mukulyadav2004: Sure, also I've replaced select with cols in the last line so that x remains list or data.table

Mukulyadav2004: @tdhock can you please guide why the error in atime even after seperating both the test cases

tdhock:
current error in atime CI is
```
 Error in parse(test.env$tests.R) : 
  /__w/data.table/data.table/.ci/atime/tests.R:328:0: unexpected end of input
326:     tests=extra.test.list)
327: # nolint end: undesirable_operator_linter.
```
this means that the file does not contain valid R code (open paren/brance without matching closer).
You should double check by trying to source/run all code in tests.R

.ci/atime/tests.R
```diff
@@ -285,6 +285,43 @@ test.list <- atime::atime_test_list(
...
+      expr = {
+        has_select = "select" %chin% names(formals(data.table::fwrite))
+        if (has_select) {
+          data.table::fwrite(DT, temp_file, select = select_idx)
+        } else {
+          data.table::fwrite(DT[, select_idx, with = FALSE], temp_file)
+        }
+      },
+      Slow = "66cb6d2393cef30083b444346a7600a079207806",     # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/66cb6d2393cef30083b444346a7600a079207806)
+      Fast = "1887699fe965b5aa1fb8cb16b5507b7a5cbf5c85",     # Commit in the PR (https://github.com/Rdatatable/data.table/pull/4177/commits) that adds select parameter
```
tdhock: both test cases are missing closing paren )

tdhock (changes requested):

.ci/atime/tests.R
Mukulyadav2004: sorry! I will recheck.

.ci/atime/tests.R
```diff
@@ -285,6 +285,43 @@ test.list <- atime::atime_test_list(
     expr = data.table::isoweek(x),
     Slow = "548410d23dd74b625e8ea9aeb1a5d2e9dddd2927",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/548410d23dd74b625e8ea9aeb1a5d2e9dddd2927)
     Fast = "c0b32a60466bed0e63420ec105bc75c34590865e"),  # Commit in the PR (https://github.com/Rdatatable/data.table/pull/7144/commits) that uses a much faster implementation
+  
+  # Test case adapted from https://github.com/Rdatatable/data.table/issues/4177 which is where the issue was reported.
+  # Fixed in https://github.com/Rdatatable/data.table/pull/7236
+  "fwrite select parameter improved in #4177 (N rows and 5 columns, select a,b,c)" = atime::atime_test(
```
tdhock:
```suggestion
  "fwrite(select) #4177 Nx5" = atime::atime_test(
```

.ci/atime/tests.R
```diff
@@ -285,6 +285,43 @@ test.list <- atime::atime_test_list(
...
+        data.table::fwrite(DT, temp_file, select = c("a","b","c"))
+      } else {
+        data.table::fwrite(DT[, c("a","b","c"), with = FALSE], temp_file)
+      }
+    },
+    Slow = "66cb6d2393cef30083b444346a7600a079207806",   # Parent of the first commit in the PR (https://github.com/Rdatatable/data.table/commit/66cb6d2393cef30083b444346a7600a079207806)
+    Fast = "1887699fe965b5aa1fb8cb16b5507b7a5cbf5c85"),   # Commit in the PR (https://github.com/Rdatatable/data.table/pull/4177/commits) that adds select parameter
+
+  # Wide case: 1 row, N columns; select a single column
+  "fwrite select parameter improved in #4177 (N columns and 1 row, select 1)" = atime::atime_test(
```
tdhock:
```suggestion
  "fwrite(select) #4177 1xN" = atime::atime_test(
```

tdhock:
the 1xN performance test is still failing, can you please investigate, explain/fix?

Mukulyadav2004:
@tdhock I find out that when select doesn't exist then we must subset first - so we've to emulatate select via subsetting before fwrite. And that subsetting must also be version pinned, so we've to use data.table:::[.data.table so that atime rewrites both [ and fwrite to same SHA.
Also can you please tell how to test atime test locally

tdhock:
https://github.com/Rdatatable/data.table/wiki/Performance-testing#running-locally

I don't understand you comment "I find out that when select doesn't exist then we must subset first - so we've to emulatate select via subsetting before fwrite. And that subsetting must also be version pinned, so we've to use data.table:::[.data.table so that atime rewrites both [ and fwrite to same SHA." can you please clarify?

Current atime result is shown below.
image
We see that master=CRAN=slow is constant whereas HEAD/Fast is linear.
Can you please figure out why HEAD is linear, and change it to be constant?
Thanks!

Mukulyadav2004:
> I don't understand you comment "I find out that when select doesn't exist then we must subset first - so we've to emulatate select via subsetting before fwrite. And that subsetting must also be version pinned, so we've to use data.table:::[.data.table so that atime rewrites both [ and fwrite to same SHA." can you please clarify?

Offcourse, so atime rewrites only namespace qualified calls. agree? So when we write DT[, j], that uses the currently attached [.data.table and is not rewritten. Thus for slow(no select), we emulate select by first subsetting and that subsetting must be pinned so both fwrite and [ come from the same SHA. Hence data.table:::[.data.table(...).
Its my understanding, maybe I'm wrong here.

Mukulyadav2004:
@tdhock thank you telling the atime CRAN/HEAD diffrence here! the problem I can identify is probably using - `cols = if (is.numeric(x))` rather than - `cols = if (is.numeric(select))` , I think this forces the name - resolution path for every run thus failure in intended O(1) path.

tdhock:
HEAD is still linear

Mukulyadav2004:
@tdhoch can you please tell how to check atime test locally so that I recheck before commits as I just added subsetting to data.table:::[.data.table.

.ci/atime/tests.R
```diff
@@ -317,7 +317,7 @@ test.list <- atime::atime_test_list(
       if (has_select) {
         data.table::fwrite(DT, temp_file, select = select_idx)
       } else {
-        data.table::fwrite(DT[, select_idx, with = FALSE], temp_file)
+        data.table::fwrite(data.table:::`[.data.table`(DT, , select_idx, with = FALSE), temp_file)
```
tdhock, line 320: this is a good idea for correctness, thanks!
but I don't think it will affect this PR, because there are no modifications to `[.data.table`

Mukulyadav2004: thanks for review! can you please guide how can I improve it then to get head constant

tdhock:
```r
tinfo <- atime::atime_pkg_test_info("~/R/data.table")
alist <- eval(tinfo$test.call[["fwrite(select) #4177 1xN"]])
plot(alist)
```
image

Mukulyadav2004: @tdhock I can't able to figure out why plot is linear at high N, its clear to me that memory plot grows with N as out test builds a 1*N table but for time graph, for most N its flat and rises at very high N - likely some overhead  when N is big maybe OS or disk effect, not sure.
Could you please spare some time to explain me the likely cause so that I can work upon it

tdhock: you must have added some code which is linear in the number of columns. please look in the additions you made.

Mukulyadav2004:
Hi @tdhock 

The code which is linear in the number of columns might be 'seq_len(NCOL(x))' - allocate linear number of columns at large N, and instead we can assign x as .shallow(x, cols) when we have length(cols) == 1L or a strict subset or a re-order.
Below is the plot-
image

Does this seems right to you?

tdhock:
that seems linear in the number of cols but your plot still shows linear for HEAD=CRAN=Fast. (ideally should be constant like master=Slow)

Mukulyadav2004:
@tdhock  As far as I can understand is in fwrite(select) disc I/O caused the big N spikes as when I reproduced the 1*N case with a null device, with a shallow 1 column view and OS null device, both memory and time plots are constant across N
image

tdhock:
thanks for your analysis.
can you please share the code you used to get that result?
in that case, can you suggest changes to either the code or test?

Mukulyadav2004:
Sure, I added this test to test.R to check complexity without disc variance using shallow subset
```
fwrite(select) #4177 1xN (null, shallow)" = atime::atime_test(
  setup = {
    DT <- data.table(t(1:N))
    shallow_fun <- getFromNamespace("shallow", "data.table")
    DT1 <- shallow_fun(DT, 1L)
    nullfile <- if (.Platform$OS.type == "windows") "NUL" else "/dev/null"
  },
  expr = data.table::fwrite(DT1, nullfile, buffMB = 1L, nThread = 1L, col.names = FALSE),
  times = 10L,
  seconds.limit = 0.05,
  Slow = "66cb6d2393cef30083b444346a7600a079207806",
  Fast = "1887699fe965b5aa1fb8cb16b5507b7a5cbf5c85"
)
```

tdhock:
why did you do `shallow_fun <- getFromNamespace("shallow", "data.table")` ?
can you please change to data.table::shallow?
Also can you please remove times and seconds.limit argument?
Also can you please add comments to explain the Slow/Fast commit ids?
Finally can you please explain why this test would be relevant? we don't see any difference between Fast and Slow, so it does not seem worth adding. (Fast and Slow should be obvious different)

Mukulyadav2004:
I added this new test - 1xN (null, shallow) as it removes disk cost -null device and column width as shallow 1 col view so that both Slow - pre select and Fast - with select already pass only one column to c, thus they are expected to look identical and flat whose result tells us the spikes we saw in the disk version come from file I/O, not per column processing. 
Thank you!

PR #7238, opened 2025-08-07
replace temporary objects with subscript operator

badasahog:
Temporary objects are unnecessary, especially when derived from a const pointer.

I think this creates an improvement in readability, as the code becomes more self documenting.

I would also add that *potentially* we should make the pointer `idxp` a restrict qualification (if safe) and a more descript name.

I don't know exactly what it's for tho, so I'm useless there.

### DIFF
--- a/src/subset.c
+++ b/src/subset.c
@@ -25,13 +25,11 @@ void subsetVectorRaw(SEXP ans, SEXP source, SEXP idx, const bool anyNA)
     if (nth>1) {                                                  \
       _Pragma("omp parallel for num_threads(nth)")                \
       for (int i=0; i<n; ++i) {                                   \
-        int elem = idxp[i];                                       \
-        ap[i] = elem==NA_INTEGER ? _NAVAL_ : sp[elem-1];          \
+        ap[i] = idxp[i]==NA_INTEGER ? _NAVAL_ : sp[idxp[i]-1];    \
       }                                                           \
     } else {                                                      \
       for (int i=0; i<n; ++i) {                                   \
-        int elem = idxp[i];                                       \
-        ap[i] = elem==NA_INTEGER ? _NAVAL_ : sp[elem-1];          \
+        ap[i] = idxp[i]==NA_INTEGER ? _NAVAL_ : sp[idxp[i]-1];    \
       }                                                           \
     }                                                             \
   } else {                                                        \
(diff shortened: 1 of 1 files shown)


src/subset.c
```diff
@@ -112,15 +110,14 @@ const char *check_idx(SEXP idx, int max, bool *anyNA_out, bool *orderedSubset_ou
   int last = INT32_MIN;
   int *idxp = INTEGER(idx), n=LENGTH(idx);
   for (int i=0; i<n; i++) {
-    int elem = idxp[i];
-    if (elem<=0 && elem!=NA_INTEGER) return "Internal inefficiency: idx contains negatives or zeros. Should have been dealt with earlier.";  // e.g. test 762  (TODO-fix)
-    if (elem>max) return "Internal inefficiency: idx contains an item out-of-range. Should have been dealt with earlier.";                   // e.g. test 1639.64
-    anyNA |= elem==NA_INTEGER;
-    anyLess |= elem<last;
-    last = elem;
+    if (idxp[i]<=0 && idxp[i]!=NA_INTEGER) return "Internal inefficiency: idx contains negatives or zeros. Should have been dealt with earlier.";  // e.g. test 762  (TODO-fix)
```
MichaelChirico, line 113: this one in particular looks useful as a local variable, rather than reading `idxp[i]` 6 more times.

Is there an advantage of `idxp[i]` over just adding `const` here?

MichaelChirico:
> I think this creates an improvement in readability, as the code becomes more self documenting.

I'm not sure about this, could you elaborate? Any external reference you can point to?

src/subset.c
badasahog: `idxp` is already a pointer to const, so `idxp[i]` would likely be held in a register.

badasahog:
It's more obvious where the data comes from, instead of adding a temporary object which adds another level of unnecessary abstraction.

src/subset.c
jangorecki: "Likely" may not be best reason for change that doesn't really do much. If it is more readable is also questionable... x[y[i]] IMO is less readable than x[yi]

MichaelChirico:
> a temporary object which adds another level of unnecessary abstraction.

No, the necessity is the improved readability.

PR #7240, opened 2025-08-07
made types more explicit to be consistent with the wider codebase

badasahog:
hopefully uncontroversial.

The format as it exists today doesn't even save space and only serves to make the code more difficult to read quickly.


MichaelChirico (approved):
LGTM

PR #7242, opened 2025-08-07
replace R_FINITE with c99's isfinite

badasahog:
`isfinite` is likely faster, and also not opaque to the compiler, allowing for other optimizations.


HughParsonage:
The R_ macro expands to isfinite if it’s available or provides a fallback
if not. So this doesn’t really do anything

MichaelChirico:
+1

https://github.com/r-devel/r-svn/blame/3a7dcc4b1b9c356c0fb2816bce8054ee8b318020/src/nmath/mlutils.c#L32-L39

badasahog:
I wasn't aware of that, but we are compiling with c99, no?

I think it's better to use standard C when possible, up to you tho.

jangorecki:
I think it is better to use R API, in case behavior changes in future we are protected. Also this change unnecessarily will introduce many conflicts...

MichaelChirico:
I think this is not needed

PR #7244, opened 2025-08-07
replaced 0 with NULL for null pointers (best practice for C)

badasahog:


MichaelChirico:
Could you please provide a reference for this

badasahog:
A reference for the fact that `NULL` is preferred over 0? I think that's common knowledge.

I'm not sure where I would look for the specific guideline that specifies this, but `NULL` exists for precisely this purpose.

MichaelChirico:
in that case I am uncommonly ignorant :)

aitap:
Gustedt even advocates against NULL (and for C23 nullptr), but that
depends on the context (this isn't a variadic function call) and we're
not C23 yet:
https://inria.hal.science/hal-02383654v1/preview/ModernC.pdf#page=149

badasahog:
I'm not familiar with this document, but I do know that NULL is used elsewhere is the code.

I still affirm my belief that using `NULL` is a good idea, but if you're confident in this document, would you like me to update the rest of the code?

It would seem hard to believe that `0` is better than `NULL`, because `NULL` is more specific, while often evaluating to `0` anyway.

It specifies behavior and intent, while `0` only specifies behavior.

MichaelChirico:
In fact the Google Guide recommends `nullptr` but I understood that's more of a C++ thing:

https://google.github.io/styleguide/cppguide.html#0_and_nullptr/NULL

badasahog:
Yeah, that's not available in c99

MichaelChirico:
I'm happy with this SO Q&A as a reference:

https://stackoverflow.com/a/1296865/3576984

Gemini also pointed me to:

https://www.mathworks.com/help/bugfinder/ref/misrac2012rule11.9.html

> The macro NULL shall be the only permitted form of integer null pointer constant

MichaelChirico (approved):

PR #7247, opened 2025-08-10
pointer type changes + replace if with switch

badasahog:
the `char*` was always casted to another type anyway, so `void*` denotes this pattern better.

### DIFF
--- a/src/reorder.c
+++ b/src/reorder.c
@@ -63,14 +63,17 @@ SEXP reorder(SEXP x, SEXP order)
     // This check is once up front, and then idx is applied to all the columns which is where the most time is spent.
   }
 
-  char *TMP = R_alloc(nmid, maxSize);
+  void *TMP = R_alloc(nmid, maxSize);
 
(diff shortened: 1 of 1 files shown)


src/reorder.c
```diff
@@ -63,14 +63,17 @@ SEXP reorder(SEXP x, SEXP order)
...
     const size_t size = RTYPE_SIZEOF(v);    // size_t, otherwise #61 (integer overflow in memcpy)
-    if (size==4) {
-      const int *restrict vd = DATAPTR_RO(v);
-      int *restrict tmp = (int *)TMP;
+
+    switch (size)
+    {
+    case 4: {
+      const int* restrict vd = DATAPTR_RO(v);
+      int* restrict tmp = TMP;
```
ben-schwen:
```suggestion
      const int *restrict vd = DATAPTR_RO(v);
      int *restrict tmp = TMP;
```

leave this at is

ben-schwen:
LGTM TY!

PR #7248, opened 2025-08-11
renamed parser functions

badasahog:
I renamed some of the parser functions to be more consistent with the wider codebase, as well as more descriptive.

These functions aren't called directly, so the diff is relatively small.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -635,12 +635,12 @@ static void str_to_i32_core(const char **pch, int32_t *target, bool parse_date)
   }
 }
 
-static void StrtoI32(FieldParseContext *ctx)
+static void parse_i32(FieldParseContext *ctx)
 {
   str_to_i32_core(ctx->ch, (int32_t*) ctx->targets[sizeof(int32_t)], false);
 }
 
-static void StrtoI64(FieldParseContext *ctx)
+static void parse_i64(FieldParseContext *ctx)
 {
   const char *ch = *ctx->ch;
   int64_t *target = ctx->targets[sizeof(int64_t)];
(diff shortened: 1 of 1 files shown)


badasahog:
@MichaelChirico I have 5 PRs outstanding, please review.

src/fread.c
```diff
@@ -635,12 +635,12 @@ static void str_to_i32_core(const char **pch, int32_t *target, bool parse_date)
   }
 }
 
-static void StrtoI32(FieldParseContext *ctx)
+static void parse_i32(FieldParseContext *ctx)
```
MichaelChirico, line 639: wouldn't the consistent naming here be `parse_int32_str` (`parse_${DEST_TYPE}_${SOURCE_DESCRIPTION}`)?

src/fread.c
```diff
@@ -1123,7 +1123,7 @@ static void parse_empty(FieldParseContext *ctx)
 }
 
 /* Parse numbers 0 | 1 as boolean and ,, as NA (fwrite's default) */
-static void parse_bool_numeric(FieldParseContext *ctx)
+static void parse_bool_10(FieldParseContext *ctx)
```
MichaelChirico, line 1127: I prefer the old name, right now it's only `1`/`0` but I'm not sure we would add another parser if we wanted to, say, include `00`,`01`. numeric->bool seems the best description here.

src/fread.c
```diff
@@ -1189,7 +1189,7 @@ static void parse_bool_lowercase(FieldParseContext *ctx)
 }
 
 /* Parse Y | y | N | n as boolean */
-static void parse_bool_yesno(FieldParseContext *ctx)
+static void parse_bool_yn(FieldParseContext *ctx)
```
MichaelChirico, line 1193: same here, "yes/no" is the most general description for what this parser _could_ do, anchoring on only `y`/`n` in a sense exposes the implementation detail that this currently only applies to those two characters -- in principle eventually `y e s` | `n o` could be parsed here.

PR #7249, opened 2025-08-11
skip .lintr file during R build

jangorecki:
PR resolves a NOTE from R CHECK on my configuration

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -19,6 +19,7 @@
 ^\.github$
 ^\.vscode$
 ^\.zed$
+^\.lintr$
 
 ^\.gitlab-ci\.yml$
 


aitap:
Maybe we should skip all hidden files and directories? The following
regex in .Rbuildignore would implement that:

(^|/)\.

jangorecki:
> Maybe we should skip all hidden files and directories? The following regex in .Rbuildignore would implement that: (^|/)\.

it is nice idea but then would be good to ensure if there are valid use cases to have dotfiles included in the package build. Keeping it like this guarantee less surprises in future

aitap:
FWIW, the current version of data.table has no hidden files inside the
tarball. The only hidden paths inside CRAN tarballs [*] are:

.aspell/*
.Rinstignore
vignettes/.install_extras
.Rbuildignore # only in 'popgenr', likely by mistake

[*]
for f in *_*.tar.gz; do
 tar tf "$f" | sed -En '/(^|\/)[.]/s/^[^\/]*\///p'
done | sort -u

PR #7250, opened 2025-08-12
rowwise(): Detect and reject non-atomic, non-list column values with clear error message

### LINKED ISSUE #7219: `rowwiseDT` with column of functions
I'm trying to create a column with functions using `rowwiseDT`. This creates an error. 

Debugging, I see the current implementation checks `needs_list = lengths(body) != 1L`. This check, while simple, misses out on functions (or more generally anything that needs to be wrapped in `list`). 

Reprex:
``` r
data.table::rowwiseDT(
  x=, func=,
  1e3, \(x) x + 1,
  1e4, \(x) x + 1,
  1e5, \(x) x + 1,
  1e6, \(x) x + 1
)
#> Error in rbindlist(body): invalid type/length (closure/4) in vector allocation
```

venom1204:
closes #7219 

This PR enhances rowwise() to detect all non-atomic, non-list column values (e.g., language objects, expressions, pairlists, environments, S4 objects) instead of only functions. It adds a clear error message with the offending column name and type, plus guidance to wrap values in list(...) if intended. Includes updated tests for both allowed and rejected cases.

hi @tdhock @MichaelChirico @aitap can you please have a look when you got time .

thanks.

### DIFF
--- a/R/rowwiseDT.R
+++ b/R/rowwiseDT.R
@@ -13,10 +13,7 @@ rowwiseDT = function(...) {
   nrows = length(body) %/% ncols
   if (length(body) != nrows * ncols)
     stopf("There are %d columns but the number of cells is %d, which is not an integer multiple of the columns", ncols, length(body))
-  is_problematic = vapply_1b(
-    body,
-    function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list")
-  )
+  is_problematic = vapply_1b(body, function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list"))
   if (any(is_problematic)) {
     idx = which(is_problematic)[1L]
     col_idx = (idx - 1L) %% ncols + 1L


R/rowwiseDT.R
```diff
@@ -13,6 +13,24 @@ rowwiseDT = function(...) {
   nrows = length(body) %/% ncols
   if (length(body) != nrows * ncols)
     stopf("There are %d columns but the number of cells is %d, which is not an integer multiple of the columns", ncols, length(body))
+
```
tdhock: please undo addition of empty lines

venom1204:
hi @tdhock I did the modifications can you please have a look when you got time.
thanks

R/rowwiseDT.R
```diff
@@ -13,6 +13,22 @@ rowwiseDT = function(...) {
   nrows = length(body) %/% ncols
   if (length(body) != nrows * ncols)
     stopf("There are %d columns but the number of cells is %d, which is not an integer multiple of the columns", ncols, length(body))
+  is_problematic = vapply(
+    body,
+    function(v) !is.atomic(v) && !is.null(v) && typeof(v) != "list",
```
ben-schwen: is this check really what @aitap wrote about catering for that `is.atomic(NULL)=TRUE` on older R versions?

it should also be faster to check `!(is.atomic(v) || is.null(v) || typeof(v) == "list")`

Note that we have our internal versions of vapply e.g. `vapply_1b`

R/rowwiseDT.R
```diff
@@ -13,6 +13,21 @@ rowwiseDT = function(...) {
...
     stopf("There are %d columns but the number of cells is %d, which is not an integer multiple of the columns", ncols, length(body))
+  is_problematic = vapply_1b(
+    body,
+    function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list")
+  )
+  if (any(is_problematic)) {
+    first_problem_idx = which(is_problematic)[1L]
+    col_idx = (first_problem_idx - 1L) %% ncols + 1L
+    col_name = header[col_idx]
+    obj_type = typeof(body[[first_problem_idx]])
```
ben-schwen: Not sure if `typeof` is the right choice here. For the problem raised in #7219 `class` would be a better choice since `class` would be "function" but `typeof` would be still "closure".

R/rowwiseDT.R
```diff
@@ -13,6 +13,21 @@ rowwiseDT = function(...) {
...
+    stopf(
+      "In column '%s', received an object of type '%s'.\nComplex objects (like functions, models, etc.) must be wrapped in list() to be stored in a data.table column.\nPlease use `list(...)` for this value.",
```
ben-schwen: Is this really true? Many `models` are indeed lists, e.g. `lm(mpg~., data=mtcars)`.

Other thing on top of my head would be `expression`s which we do not fully support. Storing environments into a `data.table` seems uncommon to me.

R/rowwiseDT.R
```diff
@@ -13,6 +13,21 @@ rowwiseDT = function(...) {
...
+    body,
+    function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list")
+  )
+  if (any(is_problematic)) {
+    first_problem_idx = which(is_problematic)[1L]
+    col_idx = (first_problem_idx - 1L) %% ncols + 1L
+    col_name = header[col_idx]
+    obj_type = class1(body[[first_problem_idx]])
+    stopf(
+      "In column '%s', received an object of type '%s'.\nComplex objects (like functions, formulas, or calls) must be wrapped in list() to be stored in a data.table column.\nPlease use `list(...)` for this value.",
```
ben-schwen: The indent is off, see e.g.

```R
> rowwiseDT(x =, func =, 1, \(x) x + 1)
Error in rowwiseDT(x = , func = , 1, function(x) x + 1) : 
  In column 'func', received an object of type 'function'.
Complex objects (like functions, formulas, or calls) must be wrapped in list() to be stored in a data.table column.
Please use `list(...)` for this value.
```

R/rowwiseDT.R
```diff
@@ -13,6 +13,21 @@ rowwiseDT = function(...) {
...
     stopf("There are %d columns but the number of cells is %d, which is not an integer multiple of the columns", ncols, length(body))
+  is_problematic = vapply_1b(
+    body,
+    function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list")
+  )
+  if (any(is_problematic)) {
+    first_problem_idx = which(is_problematic)[1L]
+    col_idx = (first_problem_idx - 1L) %% ncols + 1L
+    col_name = header[col_idx]
+    obj_type = class1(body[[first_problem_idx]])
```
ben-schwen:
```suggestion
    idx = which(is_problematic)[1L]
    col_idx = (idx - 1L) %% ncols + 1L
    col_name = header[col_idx]
    obj_type = class1(body[[idx]])
```

Would feel more consistent to our codebase/

NEWS.md
```diff
@@ -197,6 +197,8 @@
 
 19. Ellipsis elements like `..1` are correctly excluded when searching for variables in "up-a-level" syntax inside `[`, [#5460](https://github.com/Rdatatable/data.table/issues/5460). Thanks @ggrothendieck for the report and @MichaelChirico for the fix.
 
+20. `rowwiseDT()` now provides a helpful error message when a complex object that is not a list (e.g., a function) is provided as a cell value, instructing the user to wrap it in `list()`. [#7219](https://github.com/Rdatatable/data.table/issues/7219). Thanks @kylebutts for the report and @venom1204 for the fix.
```
ben-schwen:
```suggestion
20. `rowwiseDT()` now provides a helpful error message when a complex object that is not a list (e.g., a function) is provided as a cell value, instructing the user to wrap it in `list()`, [#7219](https://github.com/Rdatatable/data.table/issues/7219). Thanks @kylebutts for the report and @venom1204 for the fix.
```

ben-schwen: Ofc, it must also be moved to the `1.18.99` NEWS section

R/rowwiseDT.R
ben-schwen: Maybe also something more in the direction of 


```r
stopf("Column '%s' is type '%s'. Non-atomic, non-list objects must be wrapped in list(), e.g., list(f) instead of f", col_name, obj_type)
```

venom1204:
> Sorry for the late review. Some minor things, but overall close to finish.

hi @ben-schwen 
sorry for the late replies, was engaged somewhere else .
Have a look and let me know if any changes are needed.

R/rowwiseDT.R
```diff
@@ -13,6 +13,17 @@ rowwiseDT = function(...) {
   nrows = length(body) %/% ncols
   if (length(body) != nrows * ncols)
     stopf("There are %d columns but the number of cells is %d, which is not an integer multiple of the columns", ncols, length(body))
+  is_problematic = vapply_1b(
+    body,
+    function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list")
+  )
```
ben-schwen:
```suggestion
  is_problematic = vapply_1b(body, function(v) !(is.atomic(v) || is.null(v) || typeof(v) == "list"))
```

ben-schwen:
TY. Good work!

PR #6333, opened 2024-08-01
melt warns for measure.vars=list of length=1

tdhock:
Closes #6071 

This PR reverts the functionality introduced in https://github.com/Rdatatable/data.table/pull/5247 and adds a new warning.

Old behavior is restored, to avoid a breaking change in the next release, which affected at least one revdep.

PR #7257, opened 2025-08-21
Align melt with docs for list measure.vars

### LINKED ISSUE #6629: move from warning to breaking change in melt/dcast
v1.16.0 NEWS says that two new warnings may be changed in a future release into backwards-incompatible changes, for consistency:

18. `?melt` has long documented that the returned `variable` column should contain integer column indices when `measure.vars` is a list, but when the list length is 1, `variable` is actually a character column name, which is inconsistent with the documentation, [#5209](https://github.com/Rdatatable/data.table/issues/5209). To increase consistency in the next release, we plan to change `variable` to integer, so users who were relying on this behavior should change `measure.vars=list("col_name")` (`variable` currently is a column name but will be a column index/integer after this planned change) to `measure.vars="col_name"` (`variable` is column name before and after the planned change). For now, relying on this undocumented behavior throws a new warning.

19. `?dcast` has always required `fun.aggregate` to return a single value, and when `fill=NULL`, `dcast` would indeed error if a vector with `length!=1` was returned, but an undefined result was silently returned for non-`NULL` fill. Now `dcast()` will additionally warn that this is undefined behavior when fill is not `NULL`, [#6032](https://github.com/Rdatatable/data.table/issues/6032). In particular, this will warn for `fun.aggregate=identity`, which was observed in several revdeps. We may change this to an error in a future release, so revdeps should fix their code as soon as possible. Thanks to @tdhock for the PR, and @MichaelChirico for analysis of GitHub revdeps.

Mukulyadav2004:
closes #6629 

In melt, docs stated that when measure.vars is a list, the variable column contains integer indices. Actual behavior was inconsistent when the list length was 1 (character name), with a “warn now, break later” warning [#5209](https://github.com/Rdatatable/data.table/issues/5209).

So in this PR
melt breaking change (consistency with docs)(fmelt.c)
- When measure.vars is a list (including length 1), the variable column now contains integer indices and the transitional warning is removed.
- When measure.vars is a character or integer vector,behavior is unchanged(variable shows column names).
- If variable.factor=TRUE, levels are '1', '2', .. when measure.vars is a list,unchaged behaviour for vector measure.vars

In test 2182.6, 2182.71, 2182.73, 2182.75, expectations no swiched from names to integer and also removed the warning. And in 1037.404 and 1037.407 - Int

@tdhock @joshhwuu @Anirban166  please review.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -17421,9 +17421,9 @@ test(2182.6, melt(DT.wide, measure.vars=list(b=c("b1","b2"))), data.table(a2=2, 
 # consistency between measure.vars=list with length=1 and length>1 now uses indices for list case, #5209
 test(2182.71, melt(DT.wide, measure.vars=list("a2"), variable.factor=TRUE), data.table(b1=1, b2=2, variable=factor(1), value=2))
 test(2182.72, melt(DT.wide, measure.vars=c("a2"), variable.factor=TRUE), data.table(b1=1, b2=2, variable=factor("a2"), value=2))
-test(2182.73, melt(DT.wide, measure.vars=list("a2"), variable.factor=FALSE), data.table(b1=1, b2=2, variable="1", value=2))
+test(2182.73, melt(DT.wide, measure.vars=list("a2"), variable.factor=FALSE), data.table(b1=1, b2=2, variable=1L, value=2))
 test(2182.74, melt(DT.wide, measure.vars=c("a2"), variable.factor=FALSE), data.table(b1=1, b2=2, variable="a2", value=2))
-test(2182.75, melt(data.table(a=10, b=20), measure.vars=list(n="a"), variable.factor=FALSE), data.table(b=20, variable="1", n=10))#thanks @mnazarov
+test(2182.75, melt(data.table(a=10, b=20), measure.vars=list(n="a"), variable.factor=FALSE), data.table(b=20, variable=1L, n=10))#thanks @mnazarov
 
 ### First block testing measurev
 # new variable_table attribute for measure.vars, PR#4731 for multiple issues


src/fmelt.c
```diff
@@ -598,62 +598,84 @@ SEXP getvarcols(SEXP DT, SEXP dtnames, Rboolean varfactor, Rboolean verbose, str
...
-          const int thislen = data->narm ? length(VECTOR_ELT(data->not_NA_indices, j)) : data->nrow;
-          char buff[20];
-          snprintf(buff, sizeof(buff), "%d", level++); // # notranslate
-          for (int k=0; k<thislen; ++k) SET_STRING_ELT(target, ansloc++, mkChar(buff));
+      } else {
+        // same behavior for vector measure.vars: variable is column names
+        SET_VECTOR_ELT(ansvars, 0, target=allocVector(STRSXP, data->totlen));
+        if (data->lvalues == 1) {
+          const int *thisvaluecols = INTEGER(VECTOR_ELT(data->valuecols, 0));
+          for (int j=0, ansloc=0; j<data->lmax; ++j) {
```
tdhock: there should be no such addition here, why did you do this?

tdhock:
can you please separate dcast and melt into two separate PRs? (currently they are combined in this PR) That would be much easier to review.

src/fmelt.c
Mukulyadav2004: You’re right the non-list path in getvarcols should not change. I’ve reverted that addition so the only functional change is confined to the list branch

Mukulyadav2004:
> can you please separate dcast and melt into two separate PRs? (currently they are combined in this PR) That would be much easier to review.

Sure

src/fmelt.c
```diff
@@ -598,62 +598,75 @@ SEXP getvarcols(SEXP DT, SEXP dtnames, Rboolean varfactor, Rboolean verbose, str
...
-        }
       }
     } else {// varfactor==TRUE
       SET_VECTOR_ELT(ansvars, 0, target=allocVector(INTSXP, data->totlen));
       SEXP levels;
       int *td = INTEGER(target);
-      if (data->lvalues == 1) {//one value column to output. TODO #5247 change to !data->measure_is_list
-        SEXP thisvaluecols = VECTOR_ELT(data->valuecols, 0);
-        int len = length(thisvaluecols);
-        levels = PROTECT(allocVector(STRSXP, len)); protecti++;
```
tdhock, line 628: please revert these changes and instead use the approach here https://github.com/Rdatatable/data.table/pull/5247/files

Mukulyadav2004:
Thanks! I reverted non-list changes and followed from PR #5247 - branch on data->measure_is_list inside getvarcols. For list measure.vars, variable returns integer indices and 1..lmax levels when variable.factor=TRUE. Non-list paths are unchanged.

Does this seems right to you?

Mukulyadav2004:
@tdhock what all I'm understanding and doing here is that when variable.factor is FALSE, list measure.vars give INTSXP indices 1to lmax while non-list remains STRSXP of column names and when variable.factor is TRUE, list give a factor with integer codes 1 to lmax and levels "1" to "lmax", and non list remains the original name based factor logic, keeping the if (data->lvalues == 1).
Could you please clarify if I'm lagging somewhere.

inst/tests/tests.Rraw
```diff
@@ -17417,13 +17417,13 @@ exid = data.table(id=1, expected)
 test(2182.3, melt(DTid, measure.vars=list(a=c(NA,1), b=2:3), id.vars="id"), exid)
 test(2182.4, melt(DTid, measure.vars=list(a=c(NA,"a2"), b=c("b1","b2")), id.vars="id"), exid)
 test(2182.5, melt(DT.wide, measure.vars=list(a=c(NA,1), b=2:3), na.rm=TRUE), data.table(variable=factor(2), a=2, b=2))
-test(2182.6, melt(DT.wide, measure.vars=list(b=c("b1","b2"))), data.table(a2=2, variable=factor(c("b1","b2")), b=c(1,2)), warning="measure.vars is a list with length=1") # measure.vars named list length=1, #5065
-# consistency between measure.vars=list with length=1 and length>1, #5209
-test(2182.71, melt(DT.wide, measure.vars=list("a2"), variable.factor=TRUE), data.table(b1=1, b2=2, variable=factor("a2"), value=2), warning="measure.vars is a list with length=1")
+test(2182.6, melt(DT.wide, measure.vars=list(b=c("b1","b2"))), data.table(a2=2, variable=factor(1:2), b=c(1,2))) # list yields indices
```
tdhock, line 17420: please add these tests from the other pr
```r
test(2182.6, melt(DT.wide, measure.vars=list(b=c("b1","b2"))), data.table(a2=2, variable=factor(c("1","2")), b=c(1,2))) # measure.vars named list length=1, #5065
# consistency between measure.vars=list with length=1 and length>1, #5209
test(2182.71, melt(DT.wide, measure.vars=list("a2"), variable.factor=TRUE), data.table(b1=1, b2=2, variable=factor(1), value=2))
test(2182.72, melt(DT.wide, measure.vars=c("a2"), variable.factor=TRUE), data.table(b1=1, b2=2, variable=factor("a2"), value=2))
test(2182.73, melt(DT.wide, measure.vars=list("a2"), variable.factor=FALSE), data.table(b1=1, b2=2, variable="1", value=2))
test(2182.74, melt(DT.wide, measure.vars=c("a2"), variable.factor=FALSE), data.table(b1=1, b2=2, variable="a2", value=2))
test(2182.75, melt(data.table(a=10, b=20), measure.vars=list(n="a"), variable.factor=FALSE), data.table(b=20, variable="1", n=10))#thanks @mnazarov
```

src/fmelt.c
```diff
@@ -598,62 +598,75 @@ SEXP getvarcols(SEXP DT, SEXP dtnames, Rboolean varfactor, Rboolean verbose, str
...
+            levels = newlevels;
+          }
+          for (int j=0, ansloc=0; j<data->lmax; ++j) {
+            const int thislen = data->narm ? length(VECTOR_ELT(data->not_NA_indices, j)) : data->nrow;
+            for (int k=0; k<thislen; ++k) td[ansloc++] = md[j];
+          }
+        } else {//multiple output columns.
+          int nlevel=0;
+          levels = PROTECT(allocVector(STRSXP, data->lmax)); protecti++;
+          for (int j=0, ansloc=0; j<data->lmax; ++j) {
```
tdhock: all of this code is un-necessary. where did it come from?
all you need to do is change the test in the if statements, as in https://github.com/Rdatatable/data.table/pull/5247/files#diff-1727bfee28bf4521c354c6f4463e6fcaca14a4d62eda781552a489a749b5ec59

tdhock:
this pr is confusing to me, because it has a lot of irrelevant changes/commits.
I am closing. can you please create a new one?
an acceptable PR will just undo what this PR did:
https://github.com/Rdatatable/data.table/pull/6333/files

Mukulyadav2004:
Thank you @tdhock for the guidance! I've opened a new PR in which only revert the changes from #6333 as you said. Please review.

Issue #3452, opened 2019-03-11
With by=.EACHI and unmatched i, can we set nomatch= to get .SD[0] instead of .SD[NA]?


franknarf1:
Example from https://stackoverflow.com/q/25956178

```
library(data.table)
DT = data.table(
    x = c("A", "A", "A", "B", "B", "B"),
    y = c("i", "ii", "ii", "i", "i", "i"),
    z = 1:6
)

# .N correctly displays 0 for unmatched group
DT[CJ(x, y, unique=TRUE), on=.(x, y), .N, by=.EACHI]
#    x  y N
# 1: A  i 1
# 2: A ii 2
# 3: B  i 3
# 4: B ii 0

# .I correctly (?) displays 0 for unmatched group
DT[CJ(x, y, unique=TRUE), on=.(x, y), .I, by=.EACHI]
#    x  y I
# 1: A  i 1
# 2: A ii 2
# 3: A ii 3
# 4: B  i 4
# 5: B  i 5
# 6: B  i 6
# 7: B ii 0

# z incorrectly is NA for unmatched group
DT[CJ(x, y, unique=TRUE), on=.(x, y), sum(z), by=.EACHI]
#    x  y V1
# 1: A  i  1
# 2: A ii  5
# 3: B  i 15
# 4: B ii NA
```

Intuitively, I expect the latter to be 0, with `sum(z)` operating on `.SD[0]$z` for the unmatched group. Re https://github.com/Rdatatable/data.table/issues/857, maybe this can also be included as an option (passing `.SD[0]` to `j` when `.N == 0L`), eg like `nomatch = integer(0)`, though that's kind of esoteric notation.

PR #7258, opened 2025-08-21
Clarify .I behavior for non-matching rows in joins with by = .EACHI

### LINKED ISSUE #5457: Understanding the value of .I for non-matching rows when using .EACHI
I join two data.tables, use `.I` in `j`, and `by = .EACHI`. When a row in `i` has no match to `x` the result is `0`. I wish to understand why this is the case.

Some toy data:

```
d1 = data.table(v = c("A", "B", "C", "A", "C"))

# add column identical (value-wise) to .I
d1[ , i := .I]

d2 = data.table(v = c("D", "A", "G", "C"))

d1
#    v i
# 1: A 1
# 2: B 2
# 3: C 3
# 4: A 4
# 5: C 5

d2
#    v
# 1: D
# 2: A
# 3: G
3 4: C
```

Join the two tables on 'v'. In `j`, call either "i" or `.I`. Use `by = .EACHI` ("evaluates `j` for the groups in 'DT' that each row in `i` joins to"). 

When `j` is "i" (which at least "looks the same" as `.I`), non-matched rows evaluates to `NA`. To me, this seems consistent with the default `nomatch` behaviour: "When a row in `i` has no match to `x`, `nomatch=NA` (default) means `NA` is returned":
```
d1[d2, on = .(v), i, by = .EACHI]
#    v  i
# 1: D NA # unmatched row in `i` evaluates to NA
# 2: A  1
# 3: A  4
# 4: G NA # unmatched row in `i` evaluates to NA
# 5: C  3
# 6: C  5
```
On the other hand, when `j` is `.I`, non-matched rows evaluates to `0`:

```
d1[d2, on = .(v), .I, by = .EACHI]
#    v I
# 1: D 0 # unmatched row in `i` evaluates to 0
# 2: A 1
# 3: A 4
# 4: G 0 # unmatched row in `i` evaluates to 0
# 5: C 3
# 6: C 5
``` 

From `?.I`:
>While grouping, it holds for each item in the group, its row location in `x`

However, I fail to find documentation on how _unmatched rows_ in `i` evaluate to 0 when `j = .I`. Can someone help me understand this behaviour? 

------

R version 4.2.1 (2022-06-23 ucrt)

data.table_1.14.2

venom1204:
closes #5457 

this pr 

Clarifies `.I` docs in joins with `by = .EACHI`:

Column lookup = NA

`.I `/ `.N `on unmatched groups → 0

Docs updated, no behavior change.

hi @MichaelChirico @tdhock can you please have a look when you got time
thanks.

### DIFF
--- a/vignettes/datatable-joins.Rmd
+++ b/vignettes/datatable-joins.Rmd
@@ -259,6 +259,42 @@ dt2 = ProductReceived[
 identical(dt1, dt2)
 ```
 
+##### Understanding `j` Evaluation with `by=.EACHI` for Non-Matches
+
+A common point of confusion arises when using special symbols like `.I` in `j` with `by=.EACHI`. The behavior for non-matching rows differs from what you might expect when selecting a regular column.
+
+Let's illustrate with a simple example:
+```{r by-eachi-special-symbols}
+d1 = data.table(v = c("A", "B", "C", "A", "C"), i_col = 1:5)
+d2 = data.table(v = c("D", "A", "G", "C"))
+```
+
+*Case 1: Selecting a regular column*
+
+When we select a column from `x (d1)`, non-matching rows from `i (d2)` result in `NA`. This is the standard behavior governed by `nomatch = NA`.
+```{r}
+d1[d2, on = .(v), .(i_col), by = .EACHI]
+```
+
+For the rows `D` and `G` in `d2`, there is no matching row in `d1`, so the value for `i_col` is missing `(NA)`.
+
+*Case 2: Evaluating the special symbol `.I`*
+
+However, when we use the special symbol `.I`, non-matching rows evaluate to `0`.
+```{r}
+d1[d2, on = .(v), .I, by = .EACHI]
+```
+
+The reason for this difference is crucial:
+- In Case 1, we are performing a value lookup. A failed lookup results in a missing value (`NA`).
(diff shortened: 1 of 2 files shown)


vignettes/datatable-joins.Rmd
```diff
@@ -259,6 +259,42 @@ dt2 = ProductReceived[
...
+*Case 2: Evaluating the special symbol `.I`*
+
+However, when we use the special symbol `.I`, non-matching rows evaluate to `0`.
+```{r}
+d1[d2, on = .(v), .I, by = .EACHI]
+```
+
+The reason for this difference is crucial:
+- In Case 1, we are performing a value lookup. A failed lookup results in a missing value (`NA`).
+- In Case 2, we are performing an evaluation. The symbol `.I` is defined as "the row indices in `x` for the current group". For non-matching rows like `D`, the group of matching rows in d1 is empty. The set of indices for an empty group is integer(0). data.table represents this zero-length result as a single `0` in the output.
```
ben-schwen, line 290: where does this info come from?

ben-schwen:
Not sure `data.table.Rd` is the right place. Would rather put it into `special-symbols.Rd`

ben-schwen:
Tbh I would also rather fix it than adopting documentation.

edit: this should then also close #3452

Issue #6032, opened 2024-03-29
revdep failure from dcast error if fun.aggregate returns length>1


tdhock:
after merging https://github.com/Rdatatable/data.table/pull/5549 we get the following two new revdep check errors, not sure why, still need to investigate.
```
* checking examples ... ERROR
Running examples in 'mschart-Ex.R' failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: chart_data_smooth
> ### Title: Smooth series
> ### Aliases: chart_data_smooth
> 
> ### ** Examples
> 
> linec <- ms_linechart(data = iris, x = "Sepal.Length",
+   y = "Sepal.Width", group = "Species")
Error: Aggregating function(s) should take vector inputs and return a single value (length=1). However, function(s) returns length!=1. This value will have to be used to fill any missing combinations, and therefore must be length=1. Either override by setting the 'fill' argument explicitly or modify your function to handle this case appropriately.
Execution halted
```
and
```

* checking examples ... ERROR
Running examples in 'maditr-Ex.R' failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: to_long
> ### Title: Convert data to long or to wide form
> ### Aliases: to_long to_wide
> 
> ### ** Examples
> 
> data(iris)
> 
> # 'to_long'
> 
> long_iris = iris %>%
+     to_long(keep = Species)
> 
> long_iris
       Species     variable value
        <fctr>       <fctr> <num>
  1:    setosa Sepal.Length   5.1
  2:    setosa Sepal.Length   4.9
  3:    setosa Sepal.Length   4.7
  4:    setosa Sepal.Length   4.6
  5:    setosa Sepal.Length   5.0
 ---                             
596: virginica  Petal.Width   2.3
597: virginica  Petal.Width   1.9
598: virginica  Petal.Width   2.0
599: virginica  Petal.Width   2.3
600: virginica  Petal.Width   1.8
> 
> iris_with_stat = long_iris %>%
+     take(mean = mean(value),
+          sd = sd(value),
+          n = .N*1.0,
+          by = .(Species, variable)
+     ) %>%
+     to_long(columns = c(mean, sd, n), names_in = "stat")
> 
> # 'to_wide' - table with multiple stats
> iris_with_stat %>%
+     to_wide()
Key: <Species, stat>
[...]

PR #7260, opened 2025-08-22
Error on non scalar aggregates regardless of fill

### LINKED ISSUE #6629: move from warning to breaking change in melt/dcast

Mukulyadav2004:
closes #6629 

In dcast, fun.aggregate must return length 1. Previously, with fill != NULL, dcast warned but still produced an undefined result [#6032](https://github.com/Rdatatable/data.table/issues/6032). Plan is to turn that into an error

So in this PR
Dcast breaking change(fcast.R)
- fun.aggregate must return length 1. Previously, with fill != NULL, dcast warned but still produced an undefined result (#6032). Now this unconditionally errors and the warn branch is removed.

@tdhock @joshhwuu @Anirban166  please review.

### DIFF
--- a/R/fcast.R
+++ b/R/fcast.R
@@ -186,11 +186,7 @@ dcast.data.table = function(data, formula, fun.aggregate = NULL, sep = "_", ...,
     maybe_err = function(list.of.columns) {
       if (!all(lengths(list.of.columns) == 1L)) {
         msg = gettext("Aggregating functions should take a vector as input and return a single value (length=1), but they do not, so the result is undefined. Please fix by modifying your function so that a single value is always returned.")
-        if (is.null(fill)) { # TODO change to always stopf #6329
-          stop(msg, domain=NA, call. = FALSE)
-        } else {
-          warning(msg, domain=NA, call. = FALSE)
-        }
+        stop(msg, domain=NA, call. = FALSE)
       }
       list.of.columns
     }


tdhock:
sorry I merged too quickly. please submit another pr with a news item

Mukulyadav2004:
Sure

Issue #6662, opened 2024-12-13
base R generic sort_by for data.tables


rikivillalba:
I wonder whether it is worth to implement `sort_by.data.table` (R 4.4.0) with internal `forder` instead of base `order`, and possibly an 'in_place' argument

Some improvement can be seen in my little test:
``` r
library(data.table)
set.seed(1234)
DT <- data.table(x=runif(1e7),y=runif(1e7))

# sort_by.data.frame with "forder" instead of "order"
sort_by2 <- function (x, y, ...) 
{
  if (inherits(y, "formula")) 
    y <- .formula2varlist(y, x)
  if (!is.list(y)) 
    y <- list(y)
  o <- do.call(data.table:::forder, c(unname(y), list(...)))
  x[o, , drop = FALSE]
}

microbenchmark::microbenchmark(times = 1, 
  sort_by2(DT, ~ x + y),
  sort_by(DT, ~ x + y)
)
#> Unit: seconds
#>                  expr      min       lq     mean   median       uq      max
#>  sort_by2(DT, ~x + y) 1.068101 1.068101 1.068101 1.068101 1.068101 1.068101
#>   sort_by(DT, ~x + y) 2.317193 2.317193 2.317193 2.317193 2.317193 2.317193
```

<sup>Created on 2024-12-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

MichaelChirico:
Good idea. I wouldn't offer `in_place` (you can use `setorder()` for that, with a slightly different API).

It also looks like `.formula2varlist()` is key for allowing much more complicated formulas to WAI:

```r
mtcars |> sort_by(~I(am + mpg))
```

It will be a lot harder to try and translate generic formulas into, e.g., a `setorderv()` call.

aitap:
Can this be fixed by now-closed #6679?

PR #7267, opened 2025-08-28
add performance test for sort_by.data.table

tdhock:
Hi @rikivillalba can you please review this PR?
It adds a performance test for the new sort_by method added in #6679 
It can be run via the code:
```r
tinfo=atime::atime_pkg_test_info("~/R/data.table")
v_res <- eval(tinfo$test.call[["sort_by.data.table created in #6679"]])
plot(v_res)
```
which on my windows machine gives this result:
image

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -131,6 +131,25 @@ test.list <- atime::atime_test_list(
       paste0('useDynLib(', new.Package_))
   },
 
+  # New sort_by.data.table method implemented https://github.com/Rdatatable/data.table/pull/6679
+  # Test case adapted from https://github.com/Rdatatable/data.table/issues/6662#issue-2737165196
+  "sort_by.data.table created in #6679" = atime::atime_test(
+    seconds.limit=0.1,
+    Fast="ee44ef45814115003d1499284227af6f5e487ad3",# Last commit in the PR (https://github.com/Rdatatable/data.table/pull/6679/commits) that implemented the new feature.
+    Slow="4a2474b59637aad9e032b1eaee6e9bdcfc4df949",# Parent of the first commit (https://github.com/Rdatatable/data.table/commit/60828522cb1dbf696ce32a7323464d9d8870b9f6) in the PR (https://github.com/Rdatatable/data.table/pull/6679/commits) that implemented the new feature.
+    expr={
+      sort_by_lang <- str2lang("data.table:::sort_by.data.table")
+      tryCatch(eval(sort_by_lang), error=function(e){
+        sort_by_lang <<- str2lang("sort_by.data.frame")
+      })
+      my_call <- substitute(FUN(DT, ~ x + y), list(FUN=sort_by_lang))
+      eval(my_call)
+    },
+    setup={
+      set.seed(1234)
+      DT <- data.table(x=runif(N),y=runif(N))
+    }),
+
   # Constant overhead improvement https://github.com/Rdatatable/data.table/pull/6925
   # Test case adapted from https://github.com/Rdatatable/data.table/pull/7022#discussion_r2107900643
   "fread disk overhead improved in #6925" = atime::atime_test(


tdhock:
Also I converted your microbenchmark https://github.com/Rdatatable/data.table/issues/6662#issue-2737165196
to the atime code below
```r
data.table::update_dev_pkg()
library(data.table)
# sort_by.data.frame with "forder" instead of "order"
sort_by2 <- function (x, y, ...) 
{
  if (inherits(y, "formula")) 
    y <- .formula2varlist(y, x)
  if (!is.list(y)) 
    y <- list(y)
  o <- do.call(data.table:::forder, c(unname(y), list(...)))
  x[o, , drop = FALSE]
}
a_res <- atime::atime(
  setup={
    set.seed(1234)
    DT <- data.table(x=runif(N),y=runif(N))
  },
  seconds.limit=0.1,
  sort_by=sort_by(DT, ~ x + y),
  sort_by2=sort_by2(DT, ~ x + y),
  sort_by.data.table=data.table:::sort_by.data.table(DT, ~ x + y),
  sort_by.data.frame=sort_by.data.frame(DT, ~ x + y))
plot(a_res)
```
image
the result above on my windows machine shows that `sort_by.data.frame` is somewhat different from the others (which are all using the new `sort_by.data.table` presumably).
I was expecting this result to be similar to the performance test result https://github.com/Rdatatable/data.table/pull/7267#issue-3363754835
but it seems different (no constant time difference for small N), and I'm not sure why.

tdhock:
This test includes `str2lang("data.table:::sort_by.data.table")` which is modified for the different versions, for example:
```r
tinfo=atime::atime_pkg_test_info("~/R/data.table")
do.call(atime::atime_versions_exprs, tinfo$test.list[["sort_by.data.table created in #6679"]][c("pkg.path","Slow","Fast","expr")])
```
the code above gives the result below,
```r
> do.call(atime::atime_versions_exprs, tinfo$test.list[["sort_by.data.table created in #6679"]][c("pkg.path","Slow","Fast","expr")])
$Slow
{
    sort_by_lang <- str2lang("data.table.4a2474b59637aad9e032b1eaee6e9bdcfc4df949:::sort_by.data.table")
    tryCatch(eval(sort_by_lang), error = function(e) {
        sort_by_lang <<- str2lang("sort_by.data.frame")
    })
    my_call <- substitute(FUN(DT, ~x + y), list(FUN = sort_by_lang))
    eval(my_call)
}

$Fast
{
    sort_by_lang <- str2lang("data.table.ee44ef45814115003d1499284227af6f5e487ad3:::sort_by.data.table")
    tryCatch(eval(sort_by_lang), error = function(e) {
        sort_by_lang <<- str2lang("sort_by.data.frame")
    })
    my_call <- substitute(FUN(DT, ~x + y), list(FUN = sort_by_lang))
    eval(my_call)
}
```
we see in the result above that there are two different data table package versions used (Slow=4a2474b59637aad9e032b1eaee6e9bdcfc4df949 and Fast=ee44ef45814115003d1499284227af6f5e487ad3).
The effect is that if `sort_by.data.table` does not exist (as in Slow/old versions), we use `sort_by.data.frame` instead, which we see in the demo below:
```r
> N=2
> set.seed(1234)
> DT <- data.table(x=runif(N),y=runif(N))
> sort_by_lang <- str2lang("data.table.4a2474b59637aad9e032b1eaee6e9bdcfc4df949:::sort_by.data.table")
> tryCatch(eval(sort_by_lang), error = function(e) {
+ sort_by_lang <<- str2lang("sort_by.data.frame")
+ })
> my_call <- substitute(FUN(DT, ~x + y), list(FUN = sort_by_lang))
> my_call
sort_by.data.frame(DT, ~x + y)
> eval(my_call)
           x         y
       <num>     <num>
1: 0.1137034 0.6092747
2: 0.6222994 0.6233794
> sort_by_lang <- str2lang("data.table.ee44ef45814115003d1499284227af6f5e487ad3:::sort_by.data.table")
> tryCatch(eval(sort_by_lang), error = function(e) {
+ sort_by_lang <<- str2lang("sort_by.data.frame")
+ })
function (x, y, ...) 
{
    if (!cedta()) 
        return(NextMethod())
    if (inherits(y, "formula")) 
        y <- .formula2varlist(y, x)
    if (!is.list(y)) 
        y <- list(y)
    o <- do.call(forder, c(unname(y), list(...)))
    x[o, , drop = FALSE]
}
<bytecode: 0x0000021e73f8a970>
<environment: namespace:data.table.ee44ef45814115003d1499284227af6f5e487ad3>
> my_call <- substitute(FUN(DT, ~x + y), list(FUN = sort_by_lang))
> my_call
data.table.ee44ef45814115003d1499284227af6f5e487ad3:::sort_by.data.table(DT, 
    ~x + y)
> eval(my_call)
           x         y
       <num>     <num>
1: 0.1137034 0.6092747
2: 0.6222994 0.6233794
```

tdhock:
on ubuntu github actions CI we see the following result in which Fast is actually on top (takes more time) and Slow is on bottom.
image
there seems to be a constant factor difference, why?
Is this reproducible on others machines?
```r
tinfo=atime::atime_pkg_test_info("~/R/data.table")
v_res <- eval(tinfo$test.call[["sort_by.data.table created in #6679"]])
plot(v_res)
```

rikivillalba:
Hello Toby, excuse me for the delay in answering.

- First of all I was able to reproduce your test in my windows machine with the same outcomes overall. For the observed difference in the small N in the former example but not the later I'm not sure too but it may have to be related with the substitution in expr, or the error catching? who knows. If I paste the "measurements" data.table of each atime test individually in clean R session with the data.table.<version> packages in each case, i see no difference.  
- More important: the motivation of the data.table specific sort_by method is not only efficience but also the fact that data.table does c ordering while base R does local collated ordering, so if sort_by is not implemented then there is going to be a discrepancy between `order` and `sort_by` when applied to data.tables.
  - In the same line, it might be worth to atime the sorting  of a string vector given that forder specially optimizes character vector match and sorting.  So i generate some random text in the ascii range.

If I modify the  test such that  (and perhaps increment seconds.limit)
````
    DT <- data.table(
      x=runif(sqrt(N))[sample(sqrt(N),N,TRUE)],
      y=substring(intToUtf8(sample(33:126, 6 * N, TRUE), TRUE), 0:(N-1)*6 + 1, 1:N*6)
    ) 
````
I obtain this:
{ED5D5F61-803D-4A71-83EE-204B137C6DB2}
This is my only suggestion.
PD later today can test in Ubuntu.

tdhock:
thanks! that is interesting to observe that it is faster for character data, compared to double/numeric data.

tdhock:
we could close this PR if performance is not an issue for sort_by?

rikivillalba:
I Agree.

PR #7269, opened 2025-08-30
internal helper make.roll.names - split from frollapply rewrite

jangorecki:
This PR is a split from [frollapply rewritten, supports by.column=F #5575](https://github.com/Rdatatable/data.table/pull/5575) which is not strictly related to `frollapply` rewrite.
It includes internal helper function `make.roll.names` for naming output when `give.names=TRUE`.


jangorecki:
merging as of now to continue with https://github.com/Rdatatable/data.table/pull/5575
if anything important will be still spotted here I will amend in the next PR.

PR #7272, opened 2025-08-30
frollapply rewritten, by.column=F, parallel, any type

### LINKED ISSUE #4887: add by.column=F argument in frollapply
I might be fully knowledgable about the use of `frollapply`, but as far as I have experimented I was not successful to run rolling custom functions that requires multiple columns. 

I found zoo::rollapply function which has `by.column=F` argument that allowed me to do the job. Thing is, it is not fully compatible with `data.table[,by=]` arguments so I had to loop manually.  Could this `by.column` argument, or similar be implemented in the future? Or if I'm missing existing feature or workaround, please let me know. Thank you.

### LINKED ISSUE #7054: `frollapply` doesn't handle zero-length output

jangorecki:
This PR rewrites completely `frollapply()` function. It decouples `frollapply()` from other common rolling functions, which redirect to optimized C routines.  

The legacy implementation in C code was re-using once allocated window across all iterations, and then calling `Rf_eval()` on a user defined function supplied to `FUN` - therefore could not have been parallelized with OpenMP as we normally do in data.table due to lack of thread safety of arbitrary R function.  
Moreover it supported only `real` type on input and on output to match the input/output type of C "optimized" rolling functions.  

The implementation proposed here is written mostly in R. Similarly as legacy C implementation it also re-uses allocated memory across iterations. Additionally it uses multiple CPU threads so iterations are computed in parallel (on a decent OSes). Unlike any other multithreaded code in data.table till date it uses base R `parallel` package rather than OpenMP. OpenMP could not be used due to reason stated above.  
It now supports any atomic type and data.table/data.frame/list on input and any type output.
Closes #4887, #7054.

**Regarding PR reviews**: any code formatting or code reorganization changes which are not fixing any issue in this PR I would like to put in a follow up PR, rather than here. As it will reduce git merge conflicts in upcoming 3 PRs that builds on top of this one.

This PR supersede #5575.

---

## PR vs master

Using 4 threads:
  - rolling first, down from 2.4s to 1.7s
  - rolling median, down from 27s to 8.7s

```r
library(data.table)
set.seed(108)
setDTthreads(4)
x = rnorm(1e6)

## master
system.time(frollapply(x, 500, first))
#   user  system elapsed 
#  2.469   0.005   2.477
system.time(frollapply(x, 500, median))
#   user  system elapsed 
# 27.372   0.027  27.405

## now
system.time(frollapply(x, 500, first, simplify=unlist))
#   user  system elapsed 
#  5.827   0.360   1.715
system.time(frollapply(x, 500, median, simplify=unlist))
#   user  system elapsed 
# 33.878   0.403   8.782
```


jangorecki:
Test coverage job looks like a false positive. Adding `parallel` to Suggests in DESCRIPTION file should not have breaking impact, and the previous commit passed codecov test job.

R/frollapply.R
```diff
@@ -0,0 +1,379 @@
...
+        }
+      }
+    }
+  }
+  ## prepare templates for errors and warnings
+  msg.collect = "frollapply calling parallel::mccollect to collect results from forked processes raised %s.\n%s"
+  msg.simplify = if (is.function(simplify))
+    "frollapply completed successfully but raised %s when attempting to simplify results using user specified function in 'simplify' argument. Be sure to provide 'fill' argument matching the type and shape of results returned by the your function. Use simplify=FALSE to obtain a list instead.\n%s"
+  else if (isTRUE(simplify))
+    "frollapply completed successfully but raised %s when attempting to simplify results using our internal 'simplifylist' function. Be sure to provide 'fill' argument matching the type and shape of results returned by the your function. Use simplify=FALSE to obtain a list instead. If you believe your results could be automatically simplified please submit your use case as new issue in our issue tracker.\n%s"
```
aitap: I think these strings will need to be wrapped in `gettext()` so that `potools::po_extract` would notice them and add them to the `*.pot` file.

R/frollapply.R
```diff
@@ -0,0 +1,379 @@
...
+        ans[[thisansi]] = tryCatch(
+          simplify(ans[[thisansi]]),
+          error = function(e) stopf(msg.simplify, "an error", e[["message"]]),
+          warning = function(w) warningf(msg.simplify, "a warning", w[["message"]])
+        )
+      } else if (isTRUE(simplify)) {
+        ans[[thisansi]] = tryCatch(
+          simplifylist(ans[[thisansi]], fill, ansmask),
+          error = function(e) stopf(msg.simplify, "an error", e[["message"]]),
+          warning = function(w) warningf(msg.simplify, "a warning", w[["message"]])
```
aitap: We might have to duplicate the `msg.simplify` string to have separate cases for `an error` and `a warning` to make it nicely translatable.

R/frollapply.R
```diff
@@ -0,0 +1,379 @@
...
+      thisn = N[[j]]
+      w = allocWindow(thisx, thisn) ## prepare window, handles adaptive
+      ansmask = ansMask(thislen, thisn)
+      ansi = which(ansmask)
+      if (use.fork) { ## !windows && getDTthreads()>1L
+        ths = min(DTths, length(ansi))
+        ii = split(ansi, sort(rep_len(seq_len(ths), length(ansi)))) ## assign row indexes to threads
+        jobs = vector("integer", ths)
+        for (th in seq_len(ths)) {
+          jobs[th] = parallel::mcparallel({
```
aitap, line 296: Ambitious. And definitely better than running strictly single-threaded.

The warnings in `?frollapply` about avoiding side effects in the `FUN` are _very_ appropriate. For example, a Positron bug: `frollapply(1:1e6, 42, print)` hangs the child processes (and further experiments crashed the session for me). At least you can still interrupt the parent and continue when the children are hung. Can a macOS user test this under R.app? Any front-end that internally uses threads is a potential source of troubles here.

For a future fix: do we want to handle interrupts more gracefully than right now (leaving zombies-to-be to eat CPU time)? It might be possible to wrap this block in `withCallingHandlers(..., interrupt = \(e) ...)` to send a `SIGINT` to the worker processes before jumping out of the function and abandoning the handles.

src/frollapply.c
```diff
@@ -0,0 +1,100 @@
...
+  }
+  return dest;
+}
+// # nocov end
+
+// needed in adaptive=TRUE
+SEXP setgrowable(SEXP x) {
+  if (!isNewList(x)) {
+    SET_GROWABLE_BIT(x);
+    SET_TRUELENGTH(x, LENGTH(x)); // we just reset length, doesn't seems to be necessary but for future compatibility to R it is more safe to set it up
```
aitap: The garbage collector uses `TRUELENGTH` instead of `LENGTH` of growable vectors to count released memory, so this line is important to keep the counts (e.g. those visible in the `gc()` return value) accurate.

R/frollapply.R
```diff
@@ -0,0 +1,380 @@
...
+  if (!adaptive) {
+    cpy = copy
+    ansMask = function(len, n) {
+      mask = rep(TRUE, len)
+      mask[seq_len(n-1L)] = FALSE
+      mask
+    }
+    if (by.column) {
+      allocWindow = function(x, n) x[seq_len(n)]
+      tight = function(i, dest, src, n) FUN(.Call(CmemcpyVector, dest, src, i, n), ...)
```
aitap, line 217: Here's how the sharing problem (`setDTthreads(1); frollapply(c(1, 9), N=1L, FUN=identity)`) can be avoided for the `!adaptive && by.column` case:
1. Hide the newly-allocated window vector inside a list to control its reference count. The `w` will be referenced by `frollapply` (+1) and by `lapply` (+1) for a total reference count of 2 (and we can't count higher than that), but `w[[1]]` must stay with a reference count of 1 (only referenced by `w`). This needs raw `Ccopy` (i.e. `duplicate()`), not R-level `copy` (with special processing for lists) to work.
2. In `memcpyVector`, check whether `REFCNT(w[[1]])` is 1. If not, someone must have kept a reference to it, so repair the situation by allocating a new vector. Return the window itself, not the list-container.

```diff
diff --git a/R/frollapply.R b/R/frollapply.R
index d1c30842..527208a6 100644
--- a/R/frollapply.R
+++ b/R/frollapply.R
@@ -213,7 +213,7 @@ frollapply = function(X, N, FUN, ..., by.column=TRUE, fill=NA, align=c("right","
       mask
     }
     if (by.column) {
-      allocWindow = function(x, n) x[seq_len(n)]
+      allocWindow = function(x, n) list(x[seq_len(n)])
       tight = function(i, dest, src, n) FUN(.Call(CmemcpyVector, dest, src, i, n), ...)
     } else {
       if (!list.df) {
@@ -306,7 +306,7 @@ frollapply = function(X, N, FUN, ..., by.column=TRUE, fill=NA, align=c("right","
         oldDTthreads = setDTthreads(1L) ## for consistency, anyway window size is unlikely to be big enough to benefit any parallelism
         withCallingHandlers(
           tryCatch(
-            thisans <- lapply(ansi, FUN = tight, dest = cpy(w), src = thisx, n = thisn),
+            thisans <- lapply(ansi, FUN = tight, dest = .Call(Ccopy, w), src = thisx, n = thisn),
             error = function(e) h$err = conditionMessage(e)
           ), warning = function(w) {h$warn = c(h$warn, conditionMessage(w)); invokeRestart("muffleWarning")}
         )
diff --git a/src/frollapply.c b/src/frollapply.c
index 3b07cca3..9b0e6483 100644
--- a/src/frollapply.c
+++ b/src/frollapply.c
@@ -34,10 +34,14 @@ default: internal_error(__func__, "column type not supported in memcpyVector or
  */
 SEXP memcpyVector(SEXP dest, SEXP src, SEXP offset, SEXP size) {
   size_t o = INTEGER(offset)[0] - INTEGER(size)[0];
-  int nrow = LENGTH(dest);
-  SEXP d = dest, s = src;
+  SEXP d = VECTOR_ELT(dest, 0), s = src;
+  if (MAYBE_SHARED(d)) {
+    d = duplicate(d); // something else besides 'dest' was referencing 'd'
+    SET_VECTOR_ELT(dest, 0, d);
+  }
+  int nrow = LENGTH(d);
   MEMCPY
-  return dest;
+  return d;
 }
 // # nocov start ## does not seem to be reported to codecov most likely due to running in a fork, I manually debugged that it is being called when running froll.Rraw
 SEXP memcpyDT(SEXP dest, SEXP src, SEXP offset, SEXP size) {
```

This breaks tests 6010.011, 6010.012, 6010.025, 6010.026, 6010.027, but they are testing the currently broken behaviour. This may be worth trying to generalise to the other three(?) cases.

R/frollapply.R
```diff
@@ -0,0 +1,380 @@
...
+        ii = split(ansi, sort(rep_len(seq_len(ths), length(ansi)))) ## assign row indexes to threads
+        jobs = vector("integer", ths)
+        for (th in seq_len(ths)) {
+          jobs[th] = parallel::mcparallel({
+            #catf("%d\n", 4, "")
+            # nocov start ## fork processes seem not to be tracked by codecov
+            setDTthreads(1L)       ## disable nested parallelism
+            lapply(ii[[th]],       ## loops over indexes for that thread
+                   FUN = tight,    ## handles adaptive and by.column
+                   dest = cpy(w),  ## allocate own window for each thread
```
aitap: Why make an extra copy of `w` here? The `parallel::mcparallel` expression is already running in a separate child process, overwriting `w` in it won't affect the parent.

man/frollapply.Rd
```diff
@@ -0,0 +1,274 @@
...
+#[1] 1 9
+frollapply(c(1, 5, 9), n=1L, FUN=identity) ## unexpected again
+#[1] 5 5 9
+}
+    Problem occurs, in rather unlikely scenarios for rolling computations, when objects returned from a function can be its input (i.e. \code{identity}), or a reference to it (i.e. \code{list}), then one has to add extra \code{copy} call:
+\preformatted{
+setDTthreads(1)
+frollapply(c(1, 9), n=1L, FUN=function(x) copy(identity(x))) ## only 'copy' would be equivalent here
+#[1] 1 9
+frollapply(c(1, 9), n=1L, FUN=function(x) copy(list(x)))
```
aitap: These examples probably predate the switch from `n=` to `N=`.

aitap (commented):
Currently, most concerning are the `covr` failures. Interesting that the builds started failing after adding `Suggests: parallel`, because that's [how `covr` decides whether to apply fixes for `parallel:::mcexit`](https://cran.r-project.org/web/packages/covr/refman/covr.html#package_coverage). Not yet sure what could be broken as a result.

R/frollapply.R
jangorecki: > overwriting `w` in it won't affect the parent.

`tight` function writes to `w` by reference. If you say that it won't affect the parent (and other child processes) then you probably mean that copy-on-write feature of `fork` is being used.
This I was no aware of before. I thought that if I copy() from within code that is run by each thread, then it will be private per-thread `w`. Indeed copy-on-write makes it redundant.
Note that it does not copy nrow(x) times but only single time copy() for each thread, so the overhead is close to zero.

Considering that copy of the memory will happen anyway, but instead of being explicitly called by us, it will be handled by `fork` during copy-on-write, I believe we will not have any gain by removing that.

R/frollapply.R
jangorecki: Yes, that make sense. The question is what will be the performance impact of such change.

R function `tight` as well as `memcpy*` C functions have been designed to be critically light, as they have to run for each single observation.
Adding VECTOR_ELT() and MAYBE_SHARED() is probably not much, but still called 1e6 times can add up.

Therefore I am thinking the more elegant solution could be to check this only for first observation and based on that route call to different branches (memcpyVector vs memcpyVectorMaybeShared). Branching would obviously have to happen outside of `tight` function to avoid branching 1e6 times. WDYT?

Considering that
1. it would be significant change to the existing code (and documentation),
2. current code is well behaving as documented,
3. new behavior will be backward compatible,
4. frollapply is declared as experimental at the beginning of its manual

I believe we could add this in a follow up PR.

R/frollapply.R
jangorecki: I have committed now handling interrupts.
As for Positron I don't have it installed to test myself. Should probably be best to provide minimal example using `parallel` package only and submit as a bug report then.

R/frollapply.R
jangorecki: Those strings are passed to stopf / warningf which calls gettextf under the hood. Isn't that sufficient?

jangorecki:
Thank you for the review. I addressed your feedback either with commits or a replies in-line.

> Currently, most concerning are the `covr` failures. Interesting that the builds started failing after adding `Suggests: parallel`, because that's [how `covr` decides whether to apply fixes for `parallel:::mcexit`](https://cran.r-project.org/web/packages/covr/refman/covr.html#package_coverage). Not yet sure what could be broken as a result.

Good find. For now I removed `parallel` from suggests. I know it should be there.
I noticed we have been using `tools::file_ext` in fread (now also added for interrupt handling), but `tools` was not declared in suggests.
1. we should add `parallel` and `tools` to suggests
2. it may not be that critical considering lack of `tools` in suggests was never reported by CRAN

R/frollapply.R
MichaelChirico: That's sufficient at run-time, but calling `gettext()` here helps the other tools to earmark them for translation in the first place (that's the .pot file).

Only strings in the .pot file can be translated, and only strings marked for translation wind up in the .pot file.

jangorecki: done

jangorecki:
@tdhock @Anirban166 there seems to be an issue with atime configuration
```
Installing R system requirements
  /usr/bin/sudo DEBIAN_FRONTEND=noninteractive apt-get install -y gdebi-core qpdf devscripts ghostscript
  Reading package lists...
  Building dependency tree...
  Reading state information...
  E: Unable to locate package gdebi-core
Error: Failed to get R release: Failed to get R 4.5.1: Failed to install R: Error: The process '/usr/bin/sudo' failed with exit code 100
```

tdhock:
the atime failure appears to be a missing package in setup-r step, but it is not happening in any other PRs, so I will restart the job to see if it was spurious.

If it happens again I think we may file an issue upstream, https://github.com/r-lib/actions/issues
````
**Describe the bug**
We are using setup-r action in data.table github actions. A step in setup-r has this line https://github.com/r-lib/actions/blob/v2-branch/setup-r/src/installer.ts#L332
```
/usr/bin/sudo DEBIAN_FRONTEND=noninteractive apt-get install -y gdebi-core qpdf devscripts ghostscript
```
it appears gdebi-core package is not found, because we observe this error:
```
Error: Failed to get R release: Failed to get R 4.5.1: Failed to install R: Error: The process '/usr/bin/sudo' failed with exit code 100
```

**To Reproduce**
https://github.com/Rdatatable/data.table/actions/runs/17485610303/job/49664271994?pr=7272
https://github.com/Rdatatable/data.table/pull/7272#issuecomment-3257246424

**Expected behavior**
install R.

**Additional context**
````

**Comment by tdhock** [comment]
I have a 6-core windows machine on which I ran the timings, but I see different results
```r
#master
> system.time(frollapply(x, 500, first))
   user  system elapsed 
   2.86    0.04    2.92 
> system.time(frollapply(x, 500, median))
   user  system elapsed 
  42.39    2.19   44.69 
> system.time(frollapply(x, 500, first, simplify=unlist))
   user  system elapsed 
  12.36    0.77   13.15 
> system.time(frollapply(x, 500, median, simplify=unlist))
   user  system elapsed 
  42.94    2.23   45.51 

#this pr
> system.time(frollapply(x, 500, first))
   user  system elapsed 
   7.60    0.04    7.63 
> system.time(frollapply(x, 500, median))
   user  system elapsed 
  50.20    2.34   52.61 
> system.time(frollapply(x, 500, first, simplify=unlist))
   user  system elapsed 
   5.67    0.03    5.72 
> system.time(frollapply(x, 500, median, simplify=unlist))
   user  system elapsed 
  49.17    2.30   51.48 
```
3 timings are slower using this PR, and 1 is faster (FUN=first with simplify=unlist)

.ci/atime/tests.R
```diff
@@ -143,6 +143,20 @@ test.list <- atime::atime_test_list(
...
+  # Performance test based on system.time() code from https://github.com/Rdatatable/data.table/pull/7272#issue-3368806267
+  "frollapply improved in #7272" = atime::atime_test(
+    N=as.integer(10^seq(2, 7, by=0.2)),
+    setup={
+      set.seed(108)
+      setDTthreads(4)
+      x = rnorm(N)
+    },
+    seconds.limit=0.1,
+    expr=data.table::frollapply(x, as.integer(N/10), median),
```
tdhock: this performance test case gives the result below on my machine
image
we see little difference between versions, but HEAD takes most time, and master/Slow is slightly faster.
```
> tinfo=atime::atime_pkg_test_info("~/R/data.table")
> ares=eval(tinfo$test.call[["frollapply improved in #7272"]])
> plot(ares)
> ares$measurements[N==max(N), .(N, expr.name, median)][order(-median)]
       N           expr.name    median
   <int>              <char>     <num>
1:  3981 HEAD=frollapply2025 0.1865908
2:  3981         CRAN=1.17.8 0.1861979
3:  3981                Fast 0.1857601
4:  3981         base=master 0.1715435
5:  3981                Slow 0.1654733
```
Is it expected that this PR would not help performance on windows?

tdhock: this atime test case gives the error below on github actions CI
```r
Error: Error: Memory profiling failed.
  If you are benchmarking parallel code you must set `memory = FALSE`.
Timing stopped at: 0.369 0.052 0.405
Execution halted
```
I'm not sure why I don't get this on my local windows machine, and I'm not sure how to fix. So OK to remove the performance test if you don't think it is helpful.

jangorecki: > Is it expected that this PR would not help performance on windows?

no, it will actually be slower on windows due to overhead caused by new features added.

jangorecki:
@tdhock on Windows machine new frollapply will be slower. It is so because previous frollapply did not support any other types than double on input and output. So results from each operation could have been written directly into single answer double vector, single upfront known allocation. It can also be slower on Linux platform, when function to compute is fast and code is single threaded. It is because new implementation support any type, therefore results from each iteration are being written in list elements, and then `unlist`-ing list needs to make a copy. So at minimum there is an extra copy of results when unlisting. The cheaper the actual computation, the more noticeable the impact of copy.

----

> I have a 6-core windows machine on which I ran the timings, but I see different results
> 
> ```r
> #master
> > system.time(frollapply(x, 500, first))
> user system elapsed 
> 2.86 0.04 2.92 
> > system.time(frollapply(x, 500, median))
> user system elapsed 
> 42.39 2.19 44.69 
> > system.time(frollapply(x, 500, first, simplify=unlist))
> user system elapsed 
> 12.36 0.77 13.15 
> > system.time(frollapply(x, 500, median, simplify=unlist))
> user system elapsed 
> 42.94 2.23 45.51 
> 
> #this pr
> > system.time(frollapply(x, 500, first))
> user system elapsed 
> 7.60 0.04 7.63 
> > system.time(frollapply(x, 500, median))
> user system elapsed 
> 50.20 2.34 52.61 
> > system.time(frollapply(x, 500, first, simplify=unlist))
> user system elapsed 
> 5.67 0.03 5.72 
> > system.time(frollapply(x, 500, median, simplify=unlist))
> user system elapsed 
> 49.17 2.30 51.48 
> ```
> 
> 3 timings are slower using this PR, and 1 is faster (FUN=first with simplify=unlist)

few comments:
- windows does not support `parallel` so performance-wise advantages of this PR are absent on this platform.
- running master with `simplify=unlist` makes no sense, there was no `simplify` argument in frollapply as of master. It ends up being eated by `...` but still it is passed on each iteration and not used, therefore it has so big impact: 2.9s vs 7.6s (on windows)
- windows is not an OS for a performance computing
  - does not support `fork` (`parallel` package)
  - suffers more severely from various scenarios (openmp, or even this redundant `simplify` argument: comparing to Linux when I use `simplify` on master I have 3.6s vs 6.9s)

please fix the atime job to keep PR in ready-to-merge state.

tdhock:
ok I removed the perforamnce test.

jangorecki:
anyone else would like to have a look and submit review before I will merge?

jangorecki:
I am merging this PR to unblock further development that depends on this one. I am still very much looking forward for @aitap responses to my comments. Anyone else is of course welcome to review even after merge. If needed I will submit follow up PR(s).

R/frollapply.R
aitap: The Positron bug turned out to be posit-dev/positron#3817.

R/frollapply.R
aitap: Performing a trial run (like `empty` in `groupingsets`) and choosing the `tight` function based on that sounds like an even better idea, thank you!

R/frollapply.R
aitap: I see, you'd rather allocate a new buffer upfront than have the `parallel` processes take page faults when they try to modify the copy-on-write memory pages. Makes sense and doesn't cost much compared to the rest of the overhead.

PR #7274, opened 2025-08-31
Fix week() calculation to start week 1 at days 1–7 (resolves #2611)

### LINKED ISSUE #2611: bug in week means first week only has 6 days?

venom1204:
closes #2611 

This PR fixes the long-standing bug in week() (#2611) by defining weeks sequentially: days 1–7 are week 1, days 8–14 week 2, etc.

hi @MichaelChirico @tdhock can you please have a review when you got time .
thanks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -197,6 +197,8 @@
 
 19. Ellipsis elements like `..1` are correctly excluded when searching for variables in "up-a-level" syntax inside `[`, [#5460](https://github.com/Rdatatable/data.table/issues/5460). Thanks @ggrothendieck for the report and @MichaelChirico for the fix.
 
+20. `week()` now calculates the week of the year sequentially, where days 1-7 are always week 1. This fixes a long-standing bug where the first week of the year was incorrectly calculated as having only 6 days, [#2611](https://github.com/Rdatatable/data.table/issues/2611). Thanks to @MichaelChirico for the report and @venom1204 for the fix.
+
 ### NOTES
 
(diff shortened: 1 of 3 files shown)


tdhock:
@venom1204 please resolve conflicts

@MichaelChirico can you please review because you posted the issue this fixes?

venom1204:
@tdhock done

jangorecki:
Doesn't LGTM because it is a breaking change, not even documented. Maybe we can detect breaking behavior and raise warning that can be suppressed with a temporary global option?

ben-schwen:
@venom1204 Can you please add this breaking and raise the warning when mixed weeks arise? E.g. calculate both weeks (old + new formula) and warn when there is a mismatch?

PR #7276, opened 2025-09-01
Indent code block to fix `NEWS.md` rendering

jdblischak:
Very minor fix. I was reading the latest developments in `NEWS.md` today and noticed that the rendering was slightly off. The bullet points after number 13 were all rendered as code.

xref: https://github.com/Rdatatable/data.table/commit/f7263a30fcc6b5daddc8fcbc85ae7dbbe31fc787, https://github.com/Rdatatable/data.table/pull/7172

PR #7280, opened 2025-09-03
Add more path filters and manual triggers to GitHub Actions workflows

jdblischak:
I noticed that in my previous PR to fix a minor formatting typo that a lot of CI was run (https://github.com/Rdatatable/data.table/pull/7276). To reduce the required resources and also increase the speed of development, I added some more path-based filters to the GitHub Actions workflow triggers. Specifically:

* Only run `rchk` if a file in `src/` is updated
* Only run coverage if a file in `inst/tests/`, `R/`, `src/`, or `tests/` is updated

I also added the manual trigger `workflow_dispatch`. This is convenient when preparing a PR. As a developer, I can manually run a workflow on my feature branch in my feedstock (ie not the master branch), prior to opening the PR.

I didn't add any path filters to `code-quality.yaml` or `R-CMD-check.yaml` so that a baseline of CI is still run for every PR, but I could add them if requested.

### DIFF
--- a/.github/workflows/rchk.yaml
+++ b/.github/workflows/rchk.yaml
@@ -18,7 +18,14 @@
 on:
   push:
     branches: [master]
+    paths:
+      - '.github/workflows/rchk.yaml'
+      - 'src/**'
   pull_request:
+    paths:
+      - '.github/workflows/rchk.yaml'
+      - 'src/**'
+  workflow_dispatch:
 
 name: 'rchk'
 
--- a/.github/workflows/R-CMD-check-occasional.yaml
+++ b/.github/workflows/R-CMD-check-occasional.yaml
@@ -1,6 +1,7 @@
 on:
   schedule:
    - cron: '17 13 23 * *' # 23rd of month at 13:17 UTC
+  workflow_dispatch:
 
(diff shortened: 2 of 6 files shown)


.github/workflows/rchk.yaml
```diff
@@ -18,7 +18,14 @@
 on:
   push:
     branches: [master]
+    paths:
+      - '.github/workflows/rchk.yaml'
+      - 'src/**'
```
jangorecki, line 23: I think we need also R/\*.R, DESCRIPTION, NAMESPACE, man/\*.Rd, NEWS.md, README.md
Changes in any of them may eventually impact R CMD check.

jdblischak: But this workflow only runs [`rchk`](https://github.com/kalibera/rchk), which checks for memory protection errors in C code. I didn't add any path filters to `R-CMD-check.yaml`, which runs `R CMD check`

jdblischak:
Thank you for the reviews @jangorecki and @Anirban166. Are there any additional steps you would like me to take before merging this PR?

jangorecki:
Probably nothing, it is waiting for someone who has commit rights and feels fine to take responsibility for fixing problems if something will go wrong...

ben-schwen:
LGTM TY

PR #7291, opened 2025-09-07
froll support for unevenly spaced time series

### LINKED ISSUE #3241: rolling funs / shift could support logical window
I am filling this issue as a placeholder to evaluate users demand for such feature, at present there are no plans for incorporating it, so if you would need it be sure to upvote.
Extension of #2778.

Rolling functions and `shift` has been implemented to operate on physical order of data, which means that they do not handle "gaps" in, for example, time/date fields. If one wants to shift an `IDate` type vector by one day, one has to ensure that every single day is included in vector. If it isn't then one has to expand vector (or eventually a data.table) and perform shift afterwards. This can be flexibly and time efficiently solved using "rolling join" but the problem is memory consumption, especially for very sparse data. In an ideal world we would prefer to isolate `roll` functionality of rolling joins into helper function and re-use it in those cases.
Some examples of expected output for input `x`:
```r
library(data.table)
id = c(0L,1L,2L,5L,6L,8L)
x = data.table(date=as.IDate(id), value=c(1,2,3,4,5,6))
x
#         date value
#1: 1970-01-01     1
#2: 1970-01-02     2
#3: 1970-01-03     3
#4: 1970-01-06     4
#5: 1970-01-07     5
#6: 1970-01-09     6

## shift value by 1 date
cbind(x, data.table(s_date=as.IDate(id-1), s_value=c(NA,1,2,NA,4,NA)))
#         date value     s_date s_value
#1: 1970-01-01     1 1969-12-31      NA
#2: 1970-01-02     2 1970-01-01       1
#3: 1970-01-03     3 1970-01-02       2
#4: 1970-01-06     4 1970-01-05      NA
#5: 1970-01-07     5 1970-01-06       4
#6: 1970-01-09     6 1970-01-08      NA

## shift value by 1 date locf
cbind(x, data.table(s_date=as.IDate(id-1), s_value=c(NA,1,2,3,4,5)))
#         date value     s_date s_value
#1: 1970-01-01     1 1969-12-31      NA
#2: 1970-01-02     2 1970-01-01       1
#3: 1970-01-03     3 1970-01-02       2
#4: 1970-01-06     4 1970-01-05       3
#5: 1970-01-07     5 1970-01-06       4
#6: 1970-01-09     6 1970-01-08       5

## shift value by 1 date nocb
cbind(x, data.table(s_date=as.IDate(id-1), s_value=c(1,1,2,4,4,6)))
#         date value     s_date s_value
#1: 1970-01-01     1 1969-12-31       1
#2: 1970-01-02     2 1970-01-01       1
#3: 1970-01-03     3 1970-01-02       2
#4: 1970-01-06     4 1970-01-05       4
#5: 1970-01-07     5 1970-01-06       4
#6: 1970-01-09     6 1970-01-08       6

## shift value by -1 date
cbind(x, data.table(s_date=as.IDate(id+1), s_value=c(2,3,NA,5,NA,NA)))
#         date value     s_date s_value
#1: 1970-01-01     1 1970-01-02       2
#2: 1970-01-02     2 1970-01-03       3
#3: 1970-01-03     3 1970-01-04      NA
#4: 1970-01-06     4 1970-01-07       5
#5: 1970-01-07     5 1970-01-08      NA
#6: 1970-01-09     6 1970-01-10      NA

## shift value by -1 date locf
cbind(x, data.table(s_date=as.IDate(id+1), s_value=c(2,3,3,5,5,6)))
#         date value     s_date s_value
#1: 1970-01-01     1 1970-01-02       2
#2: 1970-01-02     2 1970-01-03       3
#3: 1970-01-03     3 1970-01-04       3
#4: 1970-01-06     4 1970-01-07       5
#5: 1970-01-07     5 1970-01-08       5
#6: 1970-01-09     6 1970-01-10       6

## shift value by -1 date nocb
cbind(x, data.table(s_date=as.IDate(id+1), s_value=c(2,3,4,5,6,NA)))
#         date value     s_date s_value
#1: 1970-01-01     1 1970-01-02       2
#2: 1970-01-02     2 1970-01-03       3
#3: 1970-01-03     3 1970-01-04       4
#4: 1970-01-06     4 1970-01-07       5
#5: 1970-01-07     5 1970-01-08       6
#6: 1970-01-09     6 1970-01-10      NA

## rollsum value by 3 date
cbind(x, data.table(w_date=sapply(as.IDate(id), function(x) paste(x+((-2):0), collapse=",")), w_value=c(sum(NA,NA,1),sum(NA,1,2),sum(1,2,3),sum(NA,NA,4),sum(NA,4,5),sum(5,NA,6))))
#         date value                           w_date w_value
#1: 1970-01-01     1 1969-12-30,1969-12-31,1970-01-01      NA
#2: 1970-01-02     2 1969-12-31,1970-01-01,1970-01-02      NA
#3: 1970-01-03     3 1970-01-01,1970-01-02,1970-01-03       6
#4: 1970-01-06     4 1970-01-04,1970-01-05,1970-01-06      NA
#5: 1970-01-07     5 1970-01-05,1970-01-06,1970-01-07      NA
#6: 1970-01-09     6 1970-01-07,1970-01-08,1970-01-09      NA

## rollsum value by 3 date locf
cbind(x, data.table(w_date=sapply(as.IDate(id), function(x) paste(x+((-2):0), collapse=",")), w_value=c(sum(NA,NA,1),sum(NA,1,2),sum(1,2,3),sum(3,3,4),sum(3,4,5),sum(5,5,6))))
#         date value                           w_date w_value
#1: 1970-01-01     1 1969-12-30,1969-12-31,1970-01-01      NA
#2: 1970-01-02     2 1969-12-31,1970-01-01,1970-01-02      NA
#3: 1970-01-03     3 1970-01-01,1970-01-02,1970-01-03       6
#4: 1970-01-06     4 1970-01-04,1970-01-05,1970-01-06      10
#5: 1970-01-07     5 1970-01-05,1970-01-06,1970-01-07      12
#6: 1970-01-09     6 1970-01-07,1970-01-08,1970-01-09      16

## rollsum value by 3 date nocb
cbind(x, data.table(w_date=sapply(as.IDate(id), function(x) paste(x+((-2):0), collapse=",")), w_value=c(sum(1,1,1),sum(1,1,2),sum(1,2,3),sum(4,4,4),sum(4,4,5),sum(5,6,6))))
#         date value                           w_date w_value
#1: 1970-01-01     1 1969-12-30,1969-12-31,1970-01-01       3
#2: 1970-01-02     2 1969-12-31,1970-01-01,1970-01-02       4
#3: 1970-01-03     3 1970-01-01,1970-01-02,1970-01-03       6
#4: 1970-01-06     4 1970-01-04,1970-01-05,1970-01-06      12
#5: 1970-01-07     5 1970-01-05,1970-01-06,1970-01-07      13
#6: 1970-01-09     6 1970-01-07,1970-01-08,1970-01-09      17
```
Related issue tagged as data.table: https://stackoverflow.com/questions/33553230/calculate-moving-average-every-n-hours

---

Worth to note that pandas, as of 0.23.4, do support rolling functions by logical order when `window` argument received `offset` instead of `int`: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.rolling.html
> window : int, or offset
    Size of the moving window. This is the number of observations used for calculating the statistic. Each window will be a fixed size.
    If its an offset then this will be the time period of each window. Each window will be a variable sized based on the observations included in the time-period. This is only valid for datetimelike indexes. This is new in 0.19.0

jangorecki:
This PR adds new function `frolladapt`. It is a helper function to be used to prepare adaptive rolling window width for adaptive rolling functions.
As a result it is possible to apply rolling functions on a window that is time-aware, rather than just number of rows in data.
Closes #3241
Supersede #5576

Example below presents rolling mean applied over 3 days, rather than 3 rows.
```r
idx = as.Date("2022-10-23") + c(0,1,4,5,6,7,9,10,14)
dt = data.table(index=idx, value=seq_along(idx))
dt
#        index value
#       <Date> <int>
#1: 2022-10-23     1
#2: 2022-10-24     2
#3: 2022-10-27     3
#4: 2022-10-28     4
#5: 2022-10-29     5
#6: 2022-10-30     6
#7: 2022-11-01     7
#8: 2022-11-02     8
#9: 2022-11-06     9
dt[, c("rollmean3","rollmean3days") := list(
  frollmean(value, 3),
  frollmean(value, frolladapt(index, 3), adaptive=TRUE)
  )]
dt
#        index value rollmean3 rollmean3days
#       <Date> <int>     <num>         <num>
#1: 2022-10-23     1        NA            NA
#2: 2022-10-24     2        NA            NA
#3: 2022-10-27     3         2           3.0
#4: 2022-10-28     4         3           3.5
#5: 2022-10-29     5         4           4.0
#6: 2022-10-30     6         5           5.0
#7: 2022-11-01     7         6           6.5
#8: 2022-11-02     8         7           7.5
#9: 2022-11-06     9         8           9.0
```

---

benchmark vs `slider`

numbers shows elapsed time of data.table divided by elapsed time of slider.
```r
library(slider)
library(data.table)
set.seed(108)
N = 1e6
n = 1e3
x = rnorm(N)
IDX = list(
  "completely dense" = sort(sample(N*1.0, N)),
  "very dense" = sort(sample(N*1.1, N)),
  "moderately dense" = sort(sample(N*1.5, N)),
  "moderately sparse" = sort(sample(N*2, N)),
  "very sparse" = sort(sample(N*8, N)),
  "completely sparse" = sort(sample(N*64, N))
)
lapply(IDX, function(idx) {
  ss = system.time(s <- slide_index_dbl(x, idx, mean, .before=n-1L, .complete=TRUE))
  dd = system.time(d <- frollmean(x, frolladapt(idx, n), adaptive=TRUE))
  stopifnot(isTRUE(all.equal(d, s)))
  dd[[3L]]/ss[[3L]]
})
#$`completely dense`
#[1] 0.0006895601
#
#$`very dense`
#[1] 0.0008710801
#
#$`moderately dense`
#[1] 0.001671682
#
#$`moderately sparse`
#[1] 0.002047273
#
#$`very sparse`
#[1] 0.002767992
#
#$`completely sparse`
#[1] 0.002910361
```


MichaelChirico:
> numbers show elapsed time of data.table divided by elapsed time of slider.

nit: I think it's better to cite 1/times to read results like "data.table is 100x faster" directly from the PR

NEWS.md
```diff
@@ -32,6 +32,19 @@
     ```
     Additionally argument names in `frollapply` has been renamed from `x` to `X` and `n` to `N` to avoid conflicts with common argument names that may be passed to `...`, aligning to base R API of `lapply`. `x` and `n` continue to work with a warning, for now.
 
+5. Adaptive rolling functions no longer tolerate `NA`s and negative values passed to `n` argument.
```
MichaelChirico: wrong NEWS entry?

(I guess a mistake in merge/rebase process)

NEWS.md
```diff
@@ -210,6 +223,40 @@
...
+dt
+#        index value rollmean3 rollmean3days
+#       <Date> <int>     <num>         <num>
+#1: 2022-10-23     1        NA            NA
+#2: 2022-10-24     2        NA            NA
+#3: 2022-10-27     3         2           3.0
+#4: 2022-10-28     4         3           3.5
+#5: 2022-10-29     5         4           4.0
+#6: 2022-10-30     6         5           5.0
+#7: 2022-11-01     7         6           6.5
```
MichaelChirico: nit: understanding this requires remembering that october has 31 days :)

is there a downside to using dates all in the same month?

NEWS.md
```diff
@@ -210,6 +223,40 @@
...
+#3: 2022-10-27     3
+#4: 2022-10-28     4
+#5: 2022-10-29     5
+#6: 2022-10-30     6
+#7: 2022-11-01     7
+#8: 2022-11-02     8
+#9: 2022-11-06     9
+dt[, c("rollmean3","rollmean3days") := list(
+  frollmean(value, 3),
+  frollmean(value, frolladapt(index, 3), adaptive=TRUE)
```
MichaelChirico: in terms of naming, will this function typically be called inside `froll*` or maybe very nearby (e.g. 1-3 expressions earlier)? it looks slightly redundant to use `froll` twice. `adapt()`/`adaptive()` otoh is maybe too generic, but if we expect >90% of usage to look like your simple example, maybe it's the right choice anyway?

other things that come to mind: `make_adaptive()`/`do_adapt()`, or emphasizing it like `ADAPT()`/`Adapt()`.

MichaelChirico: next, i guess it helps to emphasize somehow that `3` is _in the units of the `index` column_, so that a POSIXct column would have `n` in _seconds_.

MichaelChirico: third, WDYT about `frolladapt()` returning a classed object (or other attribute)?

Then `frollmean()` can just look for that attribute and apply `adaptive=TRUE` automatically.

Now, using `frollmean(., frolladapt(), adaptive=TRUE)` looks pretty redundant because we have to write 'adapt' twice.

R/froll.R
```diff
@@ -111,6 +111,47 @@ make.roll.names = function(x.len, n.len, n, x.nm, n.nm, fun, adaptive) {
...
+    }
+  }
+  if (!length(n))
+    stopf("'n' must be non 0 length")
+  if (anyNA(n))
+    stopf("'n' must not have NAs")
+  if (!identical(align, "right"))
+    stopf("'align' other than 'right' has not yet been implemented")
+  if (!isTRUEorFALSE(partial))
+    stopf("'partial' must be TRUE or FALSE")
```
MichaelChirico: (possible follow-up TODO):

```suggestion
    stopf("'%s' must be TRUE or FALSE", "partial")
```

R/froll.R
```diff
@@ -111,6 +111,47 @@ make.roll.names = function(x.len, n.len, n, x.nm, n.nm, fun, adaptive) {
...
+  if (!isTRUEorFALSE(give.names))
+    stopf("'give.names' must be TRUE or FALSE")
```
MichaelChirico:
```suggestion
    stopf("'%s' must be TRUE or FALSE", "give.names")
```

src/frollR.c
```diff
@@ -208,3 +208,60 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
...
+        j++;
+      } else if (lhs < rhs+n-1L) {
+        ians[i] = !p && lhs<first ? n : an; // for i==j ans=1L, unless !partial, then ans=n
+        i++;
+      }
+    }
+  }
+  UNPROTECT(1);
+  return ans;
+}
```
MichaelChirico:
```suggestion
}

```

src/frollR.c
```diff
@@ -208,3 +208,60 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
...
 }
+
+// helper called from R to generate adaptive window for irregularly spaced time series
+SEXP frolladapt(SEXP xobj, SEXP kobj, SEXP partial) {
+
+  bool p = LOGICAL(partial)[0];
+  int n = INTEGER(kobj)[0];
+  if (n < 1L)
+    error(_("'n' must be positive integer values (>= 1)"));
+  int *x = INTEGER(xobj);
```
MichaelChirico:
```suggestion
  int *x = INTEGER_RO(xobj);
```

src/frollR.c
```diff
@@ -208,3 +208,60 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
...
+  int64_t len = XLENGTH(xobj); // can be 0
+
+  if (len && x[0] == NA_INTEGER)
+    error(_("index provided to 'x' must: be sorted, have no duplicates, have no NAs")); // error text for consistency to the one below
+  for (int64_t i=1; i<len; i++) {
+    if (x[i] <= x[i-1L])
```
MichaelChirico, line 225: should we just check `is.unsorted()` from R? haven't checked it matches behavior w.r.t. NAs but it would save us from maintaining some code

NEWS.md
jangorecki: no, it was frolladapt that originally was producing NAs for partial window. Because it was difficult (without losing performance) to make consistent handling of NAs in `n` for all cases (adaptive, non-adaptive, frollapply) I decided to avoid NAs. `frolladapt` instead of `NA` will just use full `n` size, which has same behavior for partial windows.

NEWS.md
jangorecki: 1. just `adapt` is too generic as it will only be applicable to adaptive rolling functions, so having that in the name should be fine
2. I think best to focus on that in documentation rather than in function name
3. very good idea, I will file issue for that, adding it here would require many changes in various places

src/frollR.c
jangorecki: when I look at the source of `is.unsorted` it just feels unnecessarily complex, it will also be slightly slower, as it has extra anyNA call. Here we have basically 3 lines of logic, and two lines of error messages.

MichaelChirico: Yea it's pretty simple. I have in mind if maybe they would add `ALTREP` logic, for example, that would make `is.unsorted()` way faster without us needing to do anything.

Even with complicated logic, I found `is.unsorted()` basically instantaneous:

```r
x = sort(runif(1e7))
system.time(is.unsorted(x))
#    user  system elapsed 
#       0       0       0 
is.unsorted(x)
# [1] FALSE

x[1e7-(1:0)] = x[1e7-(0:1)]
system.time(is.unsorted(x))
#    user  system elapsed 
#   0.028   0.000   0.029 
is.unsorted(x)
# [1] TRUE
```

It's up to 0.5 seconds for 1e8 inputs.

In fact on R 4.5.0 at least, I think it's already taking advantage of `ALTREP`:

```
x = sort(runif(100))
.Internal(inspect(x))
# @52adb8099670 14 REALSXP g0c0 [REF(1)]  wrapper [srt=1,no_na=1]
#   @52adbb3d5c00 14 REALSXP g0c7 [REF(1)] (len=100, tl=0) 0.0108125,0.0309309,0.0486779,0.056436,0.0564726,...

x=x[100:1]
.Internal(inspect(x))
# @52adbe8ebc00 14 REALSXP g0c7 [REF(1)] (len=100, tl=0) 0.994145,0.988786,0.9683,0.95282,0.950957,...
```

IINM that `srt=1` is marking the vector as sorted which `is.unsorted()` can read & return instantly even for `1e8`.

jangorecki: Yes, `sort()` call marks the attribute, so then results are instantaneous. The thing is in the real life use cases nobody will be calling `sort()` on the index, it will be either already sorted, most probably without the attribute, or `sort()` will be called on data.frame / data.table so the values vector is ordered according to order of the index.
As for the example timing, I think we need to call it with `strictly=TRUE` to detect duplicates as well. And to use integer rather than numeric.

MichaelChirico: > nobody will be calling sort() on the index

Maybe, but it's related to #4697 -- data.table itself should be marking this attribute where possible, e.g. `setkey()` could do so

`strictly=TRUE` makes it slower for the fully-sorted case, but still 0.1 seconds for 1e8 inputs.

jangorecki: on 1e8 timings using integer and strictly=TRUE:

- sorted, no dups, no NA:
0.156s vs 0.08s

- sorted, no dups, NA at the end:
0.078s vs 0.0s

- sorted, dup at the end, no NA:
0.156s vs 0.0s

- unsorted at the end, no dups, no NA
0.158s vs 0.0s

Three latter cases are the most pessimistic ones, but even in the first one - which is "every run penalty" when input is good - our own loop cuts the timing in half.
BTW. NA case is 0.078s rather than 0.15s, like all the others, because of early call in R anyNA which escapes early.

```c
  double tic = omp_get_wtime();
  if (len && x[0] == NA_INTEGER)
    error(_("index provided to 'x' must: be sorted, have no duplicates, have no NAs")); // error text for consistency to the one below
  for (int64_t i=1; i<len; i++) {
    if (x[i] <= x[i-1L])
      break; //error(_("index provided to 'x' must: be sorted, have no duplicates, have no NAs"));
  }
  Rprintf("checking sorted non dup no NA took %.3fs\n",  omp_get_wtime()-tic);
```

NEWS.md
MichaelChirico: > I think best to focus on that in documentation rather than in function name

Yes, that's what I intended

NEWS.md
```diff
@@ -32,6 +32,19 @@
...
+    frollsum(1:3, n, adaptive=TRUE)
+    #Error in froll(fun = "sum", x = x, n = n, fill = fill, algo = algo, align = align,  :
+    #  'n' must be non-negative integer values (>= 0)
+    ```
+    If for some reason previous `NA`s behavior is needed, it can be achieved by replacing `NA`s with a value big enough
+    ```r
+    n = nafill(c(2,NA,2), fill=.Machine$integer.max)
+    frollsum(1:3, n, adaptive=TRUE)
+    ```
+
```
MichaelChirico: I don't think we need so much exposition here. If there are some breaking changes reported by revdeps we might reconsider

````suggestion
5. Negative and missing values of `n=` in adaptive rolling functions trigger an error.
````

File: NEWS.md [hunk 10]
```diff
@@ -210,6 +223,40 @@
     #[1] TRUE
     ```
 
+18. New `frolladapt` helper function has been added to aid in preparation of adaptive length of rolling window width when dealing with _irregularly spaced ordered data_. This lets the user to apply a rolling function over a period without having to deal with gaps in a data where some periods might be missing, [#3241](https://github.com/Rdatatable/data.table/issues/3241). Thanks to @jangorecki for implementation.
```
MichaelChirico: WDYT about this phrasing

```suggestion
18. New helper `frolladapt` to facilitate applying rolling functions over windows of fixed calendar-time width in irregularly-spaced data sets, thereby bypassing the need to "augment" such data with placeholder rows, [#3241](https://github.com/Rdatatable/data.table/issues/3241). Thanks to @jangorecki for implementation.
```

NEWS.md
```diff
@@ -210,6 +211,40 @@
...
+    #8: 2025-09-18     8
+    #9: 2025-09-22     9
+    dt[, c("rollmean3","rollmean3days") := list(
+      frollmean(value, 3),
+      frollmean(value, frolladapt(index, 3), adaptive=TRUE)
+      )]
+    dt
+    #        index value rollmean3 rollmean3days
+    #       <Date> <int>     <num>         <num>
+    #1: 2025-09-08     1        NA            NA
```
MichaelChirico: QQ, not directly related to this PR: if this first date is 2025-09-05 instead of 2025-09-08, is the value in the second row 2 instead of `NA`? (maybe the example should illustrate that?)

R/froll.R
```diff
@@ -111,6 +111,47 @@ make.roll.names = function(x.len, n.len, n, x.nm, n.nm, fun, adaptive) {
   ans
 }
 
+# irregularly spaced time series, helper for creating adaptive window size
+frolladapt = function(x, n, align="right", partial=FALSE, give.names=FALSE) {
+  x = unclass(x)
+  if (!is.numeric(x))
+    stopf("'x' must be of a numeric type")
+  if (!is.integer(x))
+    x = as.integer(x)
```
MichaelChirico, line 120: Is there a TODO to support double storage `x`? What's the technical limitation here?

R/froll.R
```diff
@@ -111,6 +111,47 @@ make.roll.names = function(x.len, n.len, n, x.nm, n.nm, fun, adaptive) {
...
+  if (!is.numeric(n)) {
+    stopf("'n' must be an integer")
+  } else {
+    nms = names(n) ## only for give.names
+    if (!is.integer(n)) {
+      if (!fitsInInt32(n))
```
MichaelChirico, line 126: why do we check `fitsInInt32()` for `n` but not `x`?

man/frolladapt.Rd
```diff
@@ -0,0 +1,42 @@
...
+\alias{frolladapt}
+\title{Adapt rolling window to irregularly spaced time series}
+\description{
+  Helper function to generate adaptive window size based on the irregularly spaced time series index. Experimental. Generated adaptive window can be then used in rolling functions. See \code{\link{froll}} and \code{\link{frollapply}} for details.
+}
+\usage{
+  frolladapt(x, n, align="right", partial=FALSE, give.names=FALSE)
+}
+\arguments{
+  \item{x}{ Integer. Other objects of type numeric (including \code{Date}, \code{POSIXct} and any others numeric-based class) will be coerced to integer, which, for example, in case of \code{POSIXct} means truncating to whole seconds. Must be sorted, have no duplicate and have no missing values. }
```
MichaelChirico: "numeric" is such a weird word in R; it's definitely not a type, but nor do I want to imply it's a "mode" here (since `is.numeric(Sys.Date())` is `FALSE`). I think referring to "numeric storage" is the most concise way to get the point across accurately.

```suggestion
  \item{x}{ Integer. Must be sorted with no duplicates or missing values. Other objects with numeric storage (including most commonly \code{Date} and \code{POSIXct}) will be coerced to integer, which, for example, in case of \code{POSIXct} means truncating to whole seconds. }
```

man/frolladapt.Rd
```diff
@@ -0,0 +1,42 @@
...
+  \item{n}{ Integer, positive, vector giving rolling window size(s). This is the \emph{total} number of included values in aggregate function. Value corresponds to unit of \code{x}. When \code{x} is a \code{POSIXct} then \code{n} are seconds, when \code{x} is a \code{Date} then \code{n} are days. }
+  \item{align}{ Character, default \code{"right"}. Other alignments than the default have not yet been implemented. }
```
MichaelChirico:
```suggestion
  \item{align}{ Character, default \code{"right"}. Other alignments have not yet been implemented. }
```

man/frolladapt.Rd
```diff
@@ -0,0 +1,42 @@
...
+\title{Adapt rolling window to irregularly spaced time series}
+\description{
+  Helper function to generate adaptive window size based on the irregularly spaced time series index. Experimental. Generated adaptive window can be then used in rolling functions. See \code{\link{froll}} and \code{\link{frollapply}} for details.
+}
+\usage{
+  frolladapt(x, n, align="right", partial=FALSE, give.names=FALSE)
+}
+\arguments{
+  \item{x}{ Integer. Other objects of type numeric (including \code{Date}, \code{POSIXct} and any others numeric-based class) will be coerced to integer, which, for example, in case of \code{POSIXct} means truncating to whole seconds. Must be sorted, have no duplicate and have no missing values. }
+  \item{n}{ Integer, positive, vector giving rolling window size(s). This is the \emph{total} number of included values in aggregate function. Value corresponds to unit of \code{x}. When \code{x} is a \code{POSIXct} then \code{n} are seconds, when \code{x} is a \code{Date} then \code{n} are days. }
```
MichaelChirico: It's not clear from this description how to interpret when `length(n) > 1`

MichaelChirico: I'm struggling a bit with how to phrase this most cleanly that leads me to step back and consider a design question.

Here's where I'm at as of now:

```suggestion
  \item{n}{ Integer vector giving rolling positive window size(s). Up to \code{n} values nearest to each value of \code{x}, with distance in the units of \code{x} and according to the window implied by \code{align}, are included in each rolling aggregation window. Thus when \code{x} is a \code{POSIXct}, \code{n} are seconds, and when \code{x} is a \code{Date}, \code{n} are days. }
```

Part of the complication is I'd like to say something like "all values within `n`" but that's not really true, because for right-adaptive, it's more like "all `j` such that `x[i] - n < x[j] <= x[i]`. Is that right?

It becomes even harder to describe well for `align = "center"` -- "all `j` such that `x[i] - n/2 < x[j] < x[i] + n/2` ??

In some sense, I'm wondering if `n` might be better off redefined in units that are easier to define, e.g. if `m := n-1`, we can just say "all observations up to `m` days prior".

OTOH, the current definition maps most cleanly into how the `n` is defined for existing rolling functions, as well as having the cleanest way to switch among `align = c("left", "center", "right")` where my `m` would wind up doubling in width for `align="center"`.

How do other implementations handle this?

R/froll.R
```diff
@@ -111,6 +111,47 @@ make.roll.names = function(x.len, n.len, n, x.nm, n.nm, fun, adaptive) {
...
+    stopf("'align' other than 'right' has not yet been implemented")
+  if (!isTRUEorFALSE(partial))
+    stopf("'%s' must be TRUE or FALSE", "partial")
+  if (!isTRUEorFALSE(give.names))
+    stopf("'%s' must be TRUE or FALSE", "give.names")
+
+  if (length(n) == 1L) {
+    ans = .Call(Cfrolladapt, x, n, partial)
+  } else {
+    ans = lapply(n, function(.n) .Call(Cfrolladapt, x, .n, partial))
```
MichaelChirico, line 145: I can imagine it's more efficient to iterate over each `x` for a large number of `n`, rather than recompute the full window at each iteration, right?

just checking my understanding. we can revisit that much later if it becomes a highly-requested use case.

src/frollR.c
```diff
@@ -208,3 +208,60 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
...
+  int first;
+  if (len)
+    first = x[0]+n-1;
+  while (i < len) {
+    int lhs = x[i], rhs = x[j];
+    int an = i-j+1;                 // window we are currently looking at in this iteration
+    if (an > n) {
+      error(_("internal error: an > n, should not increment i in the first place")); // # nocov
+    } else if (an == n) {           // an is same size as n, so we either have no gaps or will need to shrink an by j++
+      if (lhs == rhs+n-1) {         // no gaps - or a k gaps and a k dups?
```
MichaelChirico: what's k?

R/froll.R
jangorecki: if we check for `x` then POSIXct will not be supported. On contrary for the `n` argument there is no reason to use non-integer values.

R/froll.R
jangorecki: double storage is not a good fit for index,
even int64 is not good, it's description says

> WARNING: do not use them as replacement for 32bit integers, integer64 are **not supported for subscripting** by R-core

we use it to calculate _subscripting_ range for each window

I am not saying it couldn't work, but I haven't look into what traps might be there.

R/froll.R
jangorecki: `n` a vector means to computer multiple window sizes, I believe usually will be 1, then sometimes 2, maybe 4. In rather unlikely cases more.
If we would want to loop over `x` elements (which is the data observations itself, so much longer than `n`), then we would have to use different implementation of C routine.
If really big `n` vector would eventually being highly requested, I would rather prefer to switch lapply to mclapply than have two C implementations.

man/frolladapt.Rd
jangorecki: I will make a follow up PR to improve this also for froll.Rd and frollapply.Rd (so not really fit in this PR) focusing on scalar input in description, and then just at the end mention it can be vectorized.

jangorecki: https://github.com/Rdatatable/data.table/pull/7297

src/frollR.c
```diff
@@ -208,3 +208,60 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
...
+
+// helper called from R to generate adaptive window for irregularly spaced time series
+SEXP frolladapt(SEXP xobj, SEXP kobj, SEXP partial) {
+
+  bool p = LOGICAL(partial)[0];
+  int n = INTEGER(kobj)[0];
+  if (n < 1L)
+    error(_("'n' must be positive integer values (>= 1)"));
+  const int *x = INTEGER_RO(xobj);
+  int64_t len = XLENGTH(xobj); // can be 0
```
aitap, line 220: The native length type for vectors in R is [`R_xlen_t`](https://cran.r-project.org/doc/manuals/R-ints.html#Long-vectors) (surprisingly mentioned only a few times in WRE). We do have a lot of code that uses `int64_t`, so it's understandable that we may need to use `int64_t` to talk to it, but for new code that only talks to R it may be worth sticking to `R_xlen_t`.

Issue #5923, opened 2024-01-28
frollmax is slow on descending sequences


davidcsterratt:
`#` [`Minimal reproducible example`](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example); please be sure to set `verbose=TRUE` where possible!

```
if (!require("devtools")) install.packages("devtools")
if (!require("RCRoll")) devtools::install_github("davidcsterratt/RCRoll")
if (!require("roll")) install.packages("roll")
if (!require("data.table")) devtools::install_github("Rdatatable/data.table", branch="frollmax")

benchmark <- function(n, k, func) {
  x = func(n)
  start_time = Sys.time(); y.RCRoll = RCRoll::rollmin(x, k) ; print(paste("RCRoll:", Sys.time() - start_time))
  start_time = Sys.time(); y.roll = roll::roll_min(x, k) ; print(paste("roll:", Sys.time() - start_time))
  start_time = Sys.time(); y.data.table = data.table::frollmax(x, k) ; print(paste("data.table:", Sys.time() - start_time))
   return(list("RCRoll"=y.RCRoll, "roll"=y.roll, "data.table"=y.data.table))
}

n <- 1E7
k <- 1E3

# Random normal
out1 <- benchmark(n, k, rnorm)
# All equal to 0
out2 <- benchmark(n, k, function(n) {rep(0, n)})
# Ascending
out3 <- benchmark(n, k, function(n) {seq(0, length.out=n)})
# Descending
out4 <- benchmark(n, k, function(n) {seq(0, by=-1, length.out=n)})
```

Output of the benchmarking code:
```
[1] "RCRoll: 0.169624805450439"
[1] "roll: 0.153290987014771"
[1] "data.table: 0.0348567962646484"
[1] "RCRoll: 0.0276196002960205"
[1] "roll: 0.0650320053100586"
[1] "data.table: 0.034125804901123"
[1] "RCRoll: 0.0440917015075684"
[1] "roll: 0.0651366710662842"
[1] "data.table: 0.0270349979400635"
[1] "RCRoll: 0.0287351608276367"
[1] "roll: 0.0672793388366699"
[1] "data.table: 3.9090781211853"
```
`data.table::frollmax()` is considerably slower for descending sequences.

[...]

PR #7299, opened 2025-09-09
new rolling functions: min and prod

jangorecki:
Two new rolling functions, towards #2778.
Supersede #5682

Optimizations
- for min are the same as used for max
- for prod are the same as used for mean

---

benchmark vs very fast `roll` package
```r
library(roll)
library(data.table)
set.seed(108)
x = rnorm(1e8)
n = 1e3

## PROD
system.time(a <- roll_prod(x, n))
#   user  system elapsed 
#  1.083   0.836   1.919
system.time(b <- frollprod(x, n))
#   user  system elapsed 
#  0.822   0.384   1.209 
all.equal(a,b)
#[1] TRUE

## MIN
system.time(a <- roll_min(x, n))
#   user  system elapsed 
#  1.896   0.840   2.739
system.time(b <- frollmin(x, n))
#   user  system elapsed 
#  0.278   0.394   0.672
all.equal(a,b)
#[1] TRUE
```
This is for `rnorm`, be aware that for specific input, data.table's implementation can be much slower, see issue reported for rolling max https://github.com/Rdatatable/data.table/issues/5923

---


src/froll.c
```diff
@@ -669,3 +683,390 @@ void frollmaxExact(double *x, uint64_t nx, ans_t *ans, int k, double fill, bool
     }
   }
 }
+
+static inline void wmin(const double * restrict x, uint64_t o, int k, double * restrict w, uint64_t *iw, bool narm) {
```
ben-schwen, line 687: not a big fan of macros but couldn't we do something like and deduplicate `wmin` and `wmax`?
```c
#define DEFINE_WINDOWFUN(fname, init_val, op)
DEFINE_WINDOWFUN(wmax, R_NegInf, >=)
DEFINE_WINDOWFUN(wmin, R_PosInf, <=)
```

jangorecki: could be, one reason not to is reduced code coverage

ben-schwen: Same holds for `frollmaxFast` and `frollmaxExact`

jangorecki: > Same holds for `frollmaxFast` and `frollmaxExact`

frollmax Fast vs. Exact does not have that much in common, you possibly meant froll Min vs Max

as for the main functions, to turn them into macros would require `__func__` handling as well. Additionally it would reduce clarify of code, and friendliness for code editors.

As for the helper functions wmin, wmax, IMO it indeed make sense, but maybe better as a follow up. So this PR will be focused more on adding new things rather than refactoring previous ones.

PR #7301, opened 2025-09-09
froll: use `const`, `internal_error`; style `malloc`

jangorecki:
tidying up:
- frolladapt's `n` validation move to C as another validation was there already
- use `const` for pointers for `x` and adaptive `n`
- use `internal_error()` rather `error("internal error: ...")`
- align to `malloc(sizeof(*isnan) * nx)` usage style

build on top of #7299
diff: https://github.com/Rdatatable/data.table/compare/froll-min-prod...froll-qualifiers


src/frollR.c
```diff
@@ -99,13 +99,13 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
...
   SEXP k = PROTECT(coerceK(kobj, badaptive)); protecti++;
   int nk = length(k);
-  int *ik = NULL; int **lk = NULL;
+  const int *ik = NULL; const int **lk = NULL;
   if (!badaptive) {
-    ik = INTEGER(k);
+    ik = INTEGER_RO(k);
   } else {
-    lk = (int**)R_alloc(nk, sizeof(int*));
+    lk = (const int **)R_alloc(nk, sizeof(**lk));
```
aitap: `lk` is a `const int **`, array of pointers to constant integers. This makes `**lk` an integer. Despite the extra star in the type of `lk`, the allocation unit size still must be `sizeof(*lk)` because `lk` needs to be an array of `nk` pointers, not `nk` integers. Currently, this is a short allocation on 64-bit systems and a correctly-sized one on 32-bit ones.

src/frollR.c
```diff
@@ -133,8 +133,8 @@ SEXP frollfunR(SEXP fun, SEXP xobj, SEXP kobj, SEXP fill, SEXP algo, SEXP align,
   if (verbose)
     Rprintf(_("%s: allocating memory for results %dx%d\n"), __func__, nx, nk);
   ans_t *dans = (ans_t *)R_alloc(nx*nk, sizeof(*dans));         // answer columns as array of ans_t struct
-  double** dx = (double**)R_alloc(nx, sizeof(*dx));         // pointers to source columns
-  uint64_t* inx = (uint64_t*)R_alloc(nx, sizeof(*inx));     // to not recalculate `length(x[[i]])` we store it in extra array
+  const double** dx = (const double**)R_alloc(nx, sizeof(**dx));  // pointers to source columns
```
aitap: Similarly to line 106 above, this needs to be `sizeof(*dx)`. Since `**dx` is a `double`, this allocates the correct amount of memory on 64-bit machines anyway and twice as much on 32-bit ones.

PR #7304, opened 2025-09-10
upgraded MEMCPY from macro to function

badasahog:
Also:
formatting improvements
inlining of temporary objects
constness improvements (both qualifiers and API functions)


jangorecki:
nice idea, if wonder if we could declare it static inline function?

if the input is 1e7 long, then this code will be calling `MEMCPY_SEXP()` 1e7 times, and then function call overhead can be noticeable.
master vs PR
```r
library(data.table)
x = rnorm(5e7)
n = 1e5
f = function(x) x[1L]+x[length(x)]
system.time(frollapply(x, n, f, simplify=unlist))
```
```
   user  system elapsed 
334.876   2.350  44.801 
   user  system elapsed 
341.572   2.262  45.801 
```
334s vs 341s
```r
library(data.table)
x = rnorm(5e7)
n = 3
f = function(x) x[1L]+x[length(x)]
system.time(frollapply(x, n, f, simplify=unlist))
```
```
   user  system elapsed 
231.111   9.655  46.468 
   user  system elapsed 
239.731  11.807  48.498
```
231s vs 239s

PR #7315, opened 2025-09-14
rolling median

jangorecki:
New rolling function: median. Towards #2778

Supersede #5692

---

Implementation of rolling median is based on a novel algorithm "sort-median" described by [@suomela](https://github.com/suomela) in 2014 in his paper [Median Filtering is Equivalent to Sorting](https://arxiv.org/abs/1406.1717). "sort-median" scales very well, not only for size of input vector but also for size of rolling window. Implementation here is extended to address following limitations present in the paper:
- size of input vector must by divisible by window size
- window size must be an uneven number

Thanks to Jukka for the paper and hints on extending the algo. 

```r
rollmedian = function(x, n) {
  ans = rep(NA_real_, nx<-length(x))
  if (n<=nx) for (i in n:nx) ans[i] = median(x[(i-n+1L):(i)])
  ans
}
library(data.table)
setDTthreads(8)
set.seed(108)
x = rnorm(1e5)

n = 100
system.time(rollmedian(x, n))
#   user  system elapsed
#  2.049   0.001   2.051
system.time(frollapply(x, n, median, simplify=unlist))
#   user  system elapsed
#  3.071   0.223   0.436
system.time(frollmedian(x, n))
#   user  system elapsed
#  0.013   0.000   0.004

n = 1000
system.time(rollmedian(x, n))
#   user  system elapsed
#  3.496   0.009   3.507
system.time(frollapply(x, n, median, simplify=unlist))
#   user  system elapsed
#  4.552   0.307   0.632
system.time(frollmedian(x, n))
#   user  system elapsed
#  0.015   0.000   0.004

n = 10000
system.time(rollmedian(x, n))
#   user  system elapsed
# 16.350   0.025  16.382
system.time(frollapply(x, n, median, simplify=unlist))
#   user  system elapsed
# 14.865   0.722   2.267
system.time(frollmedian(x, n))
#   user  system elapsed
#  0.028   0.000   0.005
```

---


jangorecki:
Updated timings in https://github.com/jangorecki/rollbench

NEWS.md
```diff
@@ -246,7 +246,51 @@
     #9: 2025-09-22     9         8           9.0
     ```
 
-19. New rolling functions, `frollmin` and `frollprod`, have been implemented, towards [#2778](https://github.com/Rdatatable/data.table/issues/2778). Thanks to @jangorecki for implementation.
+19. New rolling functions: `frollmin`, `frollprod` and `frollmedian`, have been implemented, towards [#2778](https://github.com/Rdatatable/data.table/issues/2778). Thanks to @jangorecki for implementation. Implementation of rolling median is based on a novel algorithm "sort-median" described by [@suomela](https://github.com/suomela) in 2014 in his paper [Median Filtering is Equivalent to Sorting](https://arxiv.org/abs/1406.1717). "sort-median" scales very well, not only for size of input vector but also for size of rolling window.
```
mcol:
```suggestion
19. New rolling functions: `frollmin`, `frollprod` and `frollmedian`, have been implemented, towards [#2778](https://github.com/Rdatatable/data.table/issues/2778). Thanks to @jangorecki for implementation. Implementation of rolling median is based on a novel algorithm "sort-median" described by [@suomela](https://github.com/suomela) in his 2014 paper [Median Filtering is Equivalent to Sorting](https://arxiv.org/abs/1406.1717). "sort-median" scales very well, not only for size of input vector but also for size of rolling window.
```

man/froll.Rd
```diff
@@ -30,6 +32,8 @@
     na.rm=FALSE, has.nf=NA, adaptive=FALSE, partial=FALSE, give.names=FALSE, hasNA)
   frollprod(x, n, fill=NA, algo=c("fast","exact"), align=c("right","left","center"),
     na.rm=FALSE, has.nf=NA, adaptive=FALSE, partial=FALSE, give.names=FALSE, hasNA)
+  frollmedian(x, n, fill=NA, algo=c("fast","exact"), align=c("right","left","center"),
```
mcol, line 35: Here in the docs, I would move `frollmedian` close to `frollmean`, as they seem more related.

man/froll.Rd
```diff
@@ -82,20 +86,21 @@
     \item \code{has.nf=FALSE} uses faster implementation that does not support non-finite values. Then depending on the rolling function it will either:
     \itemize{
       \item (\emph{mean, sum, prod}) detect non-finite, re-run non-finite aware.
-      \item (\emph{max, min}) does not detect non-finites and may silently give incorrect answer.
+      \item (\emph{max, min, median}) does not detect non-finites and may silently give incorrect answer.
```
mcol:
```suggestion
      \item (\emph{max, min, median}) does not detect non-finites and may silently produce an incorrect answer.
```

man/froll.Rd
```diff
@@ -82,20 +86,21 @@
...
     }
     In general \code{has.nf=FALSE && any(!is.finite(x))} should be considered as undefined behavior. Therefore \code{has.nf=FALSE} should be used with care.
   }
 }
 \section{Implementation}{
-  Each rolling function has 4 different implementations. First factor that decides which implementation is being used is \code{adaptive} argument, see setion below for details. Then for each of those two algorithms (adaptive \code{TRUE} or \code{FALSE}) there are two \code{algo} argument values.
+  Each rolling function has 4 different implementations. First factor that decides which implementation is being used is \code{adaptive} argument, see setion below for details. Then for each of those two algorithms (adaptive \code{TRUE} or \code{FALSE}) there are usually two implementations depending on the \code{algo} argument.
```
mcol:
```suggestion
  Each rolling function has 4 different implementations. First factor that decides which implementation is used is the \code{adaptive} argument (either \code{TRUE} or \code{FALSE}), see section below for details. Then for each of those two algorithms there are usually two implementations depending on the \code{algo} argument.
```

man/froll.Rd
```diff
@@ -82,20 +86,21 @@
...
-      \item Not all functions have \emph{fast} implementation available. As of now \emph{max} and \emph{min} in case of \code{adaptive=TRUE} do not have \emph{fast} implementation, therefore it will automatically fall back to \emph{exact} implementation. \code{datatable.verbose} option can be used to check that.
+      \item \emph{max} and \emph{min} rolling function will not do only a single pass but, on average, \code{length(x)/n} nested loops will be computed. The bigger the window the bigger advantage over algo \emph{exact} which computes \code{length(x)} nested loops. Note that \emph{exact} uses multiple CPUs so for a small window size and many CPUs it is possible it will be actually faster than \emph{fast} but in those cases elapsed timings will likely be far below a single second.
```
mcol:
```suggestion
      \item \emph{max} and \emph{min} rolling function will not do only a single pass but, on average, they will compute \code{length(x)/n} nested loops. The larger the window, the greater the advantage over the \emph{exact} algorithm, which computes \code{length(x)} nested loops. Note that \emph{exact} uses multiple CPUs so for a small window sizes and many CPUs it may actually be faster than \emph{fast}. However, in such cases the elapsed timings will likely be far below a single second.
```

man/froll.Rd
```diff
@@ -82,20 +86,21 @@
...
+      \item Not all functions have \emph{fast} implementation available. As of now adaptive \emph{max}, adaptive \emph{min} and adaptive \emph{median} do not have \emph{fast} implementation, therefore it will automatically fall back to \emph{exact} implementation. \code{datatable.verbose} option can be used to check that.
     }
-    \item \code{algo="exact"} will make rolling functions to use a more computationally-intensive algorithm. For each observation from input vector it will compute a function on a window from scratch (complexity \eqn{O(n^2)}).
+    \item \code{algo="exact"} will make rolling functions to use a more computationally-intensive algorithm. For each observation from input vector it will compute a function on a rolling window from scratch (complexity \eqn{O(n^2)}).
```
mcol:
```suggestion
    \item \code{algo="exact"} will make the rolling functions use a more computationally-intensive algorithm. For each observation in the input vector it will compute a function on a rolling window from scratch (complexity \eqn{O(n^2)}).
```

man/froll.Rd
jangorecki: This is not much relevant. As it now reflects the order in which was implemented (around 10 places in the code), rather than something random, I would prefer to keep it for the moment like this. Once #2778 will be closed, which needs only frollvar and frollsd for, then possibly we can think about reordering.

jangorecki:
in https://github.com/Rdatatable/data.table/pull/7315/commits/34af12ecfbd7e2bcc90a7e93870fbd3dede474c2 I have worked around missing code coverage by raising warning in case any of those lines are reached.
The warning looks like this (it is emulated as we don't know how to reach there). Number 1000 is a marker telling which of the untested lines were reached:

Warning message:
In froll(fun = "median", x = x, n = n, fill = fill, algo = algo,  :
  frollfunR: 1:
frollmedianFast: algo='fast' algorithm has reached an untested lines of code: 1000. Please report to data.table issue tracker including this error message and ideally a reproducible example. Please use algo='exact' for now.

PR #6642, opened 2024-12-09
configure: test linking with -fopenmp, test the resulting shared object

aitap:
* When testing various OpenMP-related flags, `dyn.load()` the resulting shared object and run code from it before deciding to use these flags for `data.table`
 * When testing `-fopenmp`, add it to both compiler and linker flags

Hopefully, this will fix #6622 and prevent similar problems in the future. A more conservative fix is #6633.

Upsides:
* A configuration that successfully compiles but later doesn't work is much less likely to be chosen
  * Instead of compiling a shared object that fails to load, `data.table` will install without OpenMP support
* At least on GNU/Linux and (seemingly) @dvg-p4's Homebrew toolchain, `-fopenmp` is required in both compiler and linker flags

Downsides:
* If something makes `omp_get_max_threads()` return a number ≤ 0 during `data.table` installation, the build process will consider OpenMP not working and disable it.
  * On my GNU/Linux setup, setting environment variables `OMP_NUM_THREADS=0` or `OMP_THREAD_LIMIT=0` results in warnings, but `omp_get_max_threads()` remains positive.
* `-fopenmp` is now tested unconditionally, not limited to `uname %in% c('Linux', 'Darwin')`. Something that previously built single-threaded `data.table` may now enable OpenMP and surprise someone.
* `configure` now runs an R script. This can be rewritten back into a shell script if we can guarantee `mktemp -d` available on everything that runs `data.table`.
* Only tested on GNU/Linux. Given enough free time, can test on FreeBSD, OpenBSD VMs, but not a Mac.

Issue #6409, opened 2024-08-27
Please fix OpenMP support detection on macOS


barracuda156:
The code in configure for OpenMP on macOS is incorrect: it assumes Clang with `libopm`, so the check fails with GCC, where OpenMP is supported in fact. Instead, compiler should be checked first: if GCC is detected, OpenMP should be set to supported, and only Clang needs a test, because it may or may not have `libomp`.

barracuda156:
Thank you!

PR #7318, opened 2025-09-15
fix installation using clang-17

### LINKED ISSUE #6622: linker error on mac clang-17: symbol not found in flat namespace '___kmpc_barrier'
This issue may be a result of https://github.com/Rdatatable/data.table/pull/6418 causing openmp support to be detected, and data.table to be compiled, *but not linked* with -fopenmp. 

```
dgealow@Dans-P4-MacBook-Pro ~ % R          

R version 4.2.3 (2023-03-15) -- "Shortstop Beagle"
[...]

> install.packages('data.table')
Installing package into ‘/Users/dgealow/Library/R/arm64/4.2/library’
(as ‘lib’ is unspecified)
--- Please select a CRAN mirror for use in this session ---
[selected 0-cloud]

  There is a binary version available but the source version is later:
           binary source needs_compilation
data.table 1.15.4 1.16.2              TRUE

Do you want to install from sources the package which needs compilation? (Yes/no/cancel) yes
installing the source package ‘data.table’
```

<details><summary>Full output</summary>
<p>

```
trying URL 'https://cloud.r-project.org/src/contrib/data.table_1.16.2.tar.gz'
Content type 'application/x-gzip' length 5490076 bytes (5.2 MB)
==================================================
downloaded 5.2 MB

** package ‘data.table’ successfully unpacked and MD5 sums checked
** using staged installation
zlib 1.2.11 is available ok
* checking if R installation supports OpenMP with "-Xclang -fopenmp" ... no
* checking if R installation supports OpenMP with "-fopenmp" ... yes
** libs
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c assign.c -o assign.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c between.c -o between.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c bmerge.c -o bmerge.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c chmatch.c -o chmatch.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c cj.c -o cj.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c coalesce.c -o coalesce.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c dogroups.c -o dogroups.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fastmean.c -o fastmean.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fcast.c -o fcast.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fifelse.c -o fifelse.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fmelt.c -o fmelt.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c forder.c -o forder.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c frank.c -o frank.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fread.c -o fread.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c freadR.c -o freadR.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c froll.c -o froll.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c frollR.c -o frollR.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c frolladaptive.c -o frolladaptive.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fsort.c -o fsort.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fwrite.c -o fwrite.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c fwriteR.c -o fwriteR.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c gsumm.c -o gsumm.o
gsumm.c:1173:38: warning: implicit conversion from 'long long' to 'long double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]
 1173 |       ansd[i] = (ISNAN(s[i]) || s[i]>INT64_MAX || s[i]<=INT64_MIN) ? NA_INTEGER64 : (int64_t)s[i];
      |                                     ~^~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdint.h:94:26: note: expanded from macro 'INT64_MAX'
   94 | #define INT64_MAX        9223372036854775807LL
      |                          ^~~~~~~~~~~~~~~~~~~~~
1 warning generated.
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c idatetime.c -o idatetime.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c ijoin.c -o ijoin.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c init.c -o init.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c inrange.c -o inrange.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c nafill.c -o nafill.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c negate.c -o negate.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c nqrecreateindices.c -o nqrecreateindices.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c openmp-utils.c -o openmp-utils.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c programming.c -o programming.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c quickselect.c -o quickselect.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c rbindlist.c -o rbindlist.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c reorder.c -o reorder.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c shift.c -o shift.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c snprintf.c -o snprintf.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c subset.c -o subset.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c transpose.c -o transpose.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c types.c -o types.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c uniqlist.c -o uniqlist.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c utils.c -o utils.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c vecseq.c -o vecseq.o
clang -arch arm64 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/opt/R/arm64/include  -fopenmp   -fPIC  -falign-functions=64 -Wall -g -O2  -c wrappers.c -o wrappers.o
clang -arch arm64 -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/opt/R/arm64/lib -o data.table.so assign.o between.o bmerge.o chmatch.o cj.o coalesce.o dogroups.o fastmean.o fcast.o fifelse.o fmelt.o forder.o frank.o fread.o freadR.o froll.o frollR.o frolladaptive.o fsort.o fwrite.o fwriteR.o gsumm.o idatetime.o ijoin.o init.o inrange.o nafill.o negate.o nqrecreateindices.o openmp-utils.o programming.o quickselect.o rbindlist.o reorder.o shift.o snprintf.o subset.o transpose.o types.o uniqlist.o utils.o vecseq.o wrappers.o -lz -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
ld: warning: -single_module is obsolete
ld: warning: -multiply_defined is obsolete
ld: warning: -single_module is obsolete
ld: warning: -multiply_defined is obsolete
PKG_CFLAGS = -fopenmp
PKG_LIBS = -lz
installing to /Users/dgealow/Library/R/arm64/4.2/library/00LOCK-data.table/00new/data.table/libs
** R
** inst
** help
*** installing help indices
** building package indices
** installing vignettes
Error: package or namespace load failed for ‘data.table’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/Users/dgealow/Library/R/arm64/4.2/library/00LOCK-data.table/00new/data.table/libs/data_table.so':
  dlopen(/Users/dgealow/Library/R/arm64/4.2/library/00LOCK-data.table/00new/data.table/libs/data_table.so, 0x0006): symbol not found in flat namespace '___kmpc_barrier'
Error: loading failed
Execution halted
ERROR: loading failed
* removing ‘/Users/dgealow/Library/R/arm64/4.2/library/data.table’
* restoring previous ‘/Users/dgealow/Library/R/arm64/4.2/library/data.table’

The downloaded source packages are in
	‘/private/var/folders/sr/lw2ng95s3mj1qsqtdxywzjvc0000gn/T/Rtmp0Drt3Z/downloaded_packages’
Warning message:
In install.packages("data.table") :
  installation of package ‘data.table’ had non-zero exit status
> 
```

</p>
</details> 

## Workaround

`Sys.setenv(PKG_LIBS='-fopenmp')` before installing the package, as [suggested by aitap below](https://github.com/Rdatatable/data.table/issues/6622#issuecomment-2507643756) along with a possible package-level solution.

# Output of sessionInfo()
```
R version 4.2.3 (2023-03-15)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS 15.1.1

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib

locale:

attached base packages:

[1] compiler_4.2.3
```

tdhock:
Closes #6622 

So far I have just setup CI to test the installation.

Hopefully we just need to remove "-fopenmp" somewhere in configure.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -26,7 +26,8 @@ jobs:
           # jobs (mainly test-coverage) to run on every commit in PRs so as to not slow down dev.
           # GHA does run these jobs concurrently but even so reducing the load seems like a good idea.
           - {os: windows-latest, r: 'devel'}
-          # - {os: macOS-latest, r: 'release'}    # test-coverage.yaml uses macOS
+          - {os: macos-14, r: 'release'}    
+          - {os: macos-15, r: 'release'}    
           # TODO(remotes>2.5.0): Use 24.04[noble?]
           - {os: ubuntu-22.04, r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/jammy/latest"}
           # - {os: ubuntu-22.04,   r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/jammy/latest", http-user-agent: "R/4.1.0 (ubuntu-22.04) R (4.1.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }


aitap:
Based on R-SIG-Mac posts from earlier this year, I think that clang-17
needs the very latest OpenMP runtime from
<https://mac.r-project.org/openmp/>, because the one shipped with R is
not compatible with Xcode 16.3+. Then we need to set the compilation
flags so that data.table picks up the new /usr/local/lib/libomp.dylib
installed by us instead of the -lomp shipped with R.

I think it should work if (1) our CI sets the SHLIB_OPENMP_CFLAGS Make
macro in ~/.R/Makevars (like a Mac user would do after upgrading Xcode
and before recompiling everything from scratch) and (2) data.table
checks this macro in the ./configure script, but I haven't seen this
exact configuration in the wild. Will try to get it working.

tdhock:
the github action R-CMD-check / macos-15 is failing with the same error as reported in #6622 
```
Error: Error: package or namespace load failed for ‘data.table’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/private/var/folders/sw/lqy3v8g55m76fhwckg439jjm0000gn/T/RtmpzQ41MM/Rinstfe91e48ef29/00LOCK-data.table/00new/data.table/libs/data_table.so':
  dlopen(/private/var/folders/sw/lqy3v8g55m76fhwckg439jjm0000gn/T/RtmpzQ41MM/Rinstfe91e48ef29/00LOCK-data.table/00new/data.table/libs/data_table.so, 0x0006): symbol not found in flat namespace '___kmpc_dispatch_deinit'
Error: Error: loading failed
```

.github/workflows/R-CMD-check.yaml
```diff
@@ -15,7 +15,6 @@ jobs:
     name: ${{ matrix.config.os }} (${{ matrix.config.r }})
 
     strategy:
-      fail-fast: true
```
tdhock, line 23: I deleted this line because I would like to have the macos-14 build run to completion/sucess as a positive control https://github.com/Rdatatable/data.table/actions/runs/17743367148/job/50422540824 but I still see macos-14 build gets cancelled as soon as macos-15 build fails. is there a way to configure that here? Or is the configuration from master branch overriding somehow?

aitap:
[Installing a custom OpenMP runtime](https://stat.ethz.ch/pipermail/r-sig-mac/2025-May/015192.html) seems to help, but this is somewhat fragile:
 * The `PKG_*` environment variables are set globally for all source packages. We get most of our dependencies as binary packages from CRAN, so it only affects `data.table`. If there was a standard Make macro like `SHLIB_OPENMP_LIBS` documented in WRE, we would set it in `~/.R/Makevars` when setting up the custom OpenMP runtime so that only packages that need OpenMP would be affected. Unfortunately, there's only `SHLIB_OPENMP_CFLAGS`/`SHLIB_OPENMP_FFLAGS`/`SHLIB_OPENMP_CXXFLAGS`, where the same flags are supposed to be passed to both the compiler and the linker (so `SHLIB_OPENMP_LIBS` would be counter-productive); only macOS needs different flags.
 * Strictly speaking, this setup is not compatible with CRAN binaries. If we try to run a CRAN-built package that uses OpenMP together with our own source build of `data.table`, it will probably fail. But installing from source takes time and effort, and `data.table` is the only user of OpenMP during R CMD check, so it works.

tdhock:
great please add NEWS item

aitap:
What would you suggest for the text of the news item when the change is mostly not user-visible? "The package is now also tested with Apple clang 17"?

tdhock:
sure that is fine.
I would disagree about "mostly not user-visible" -- easy installation is a really important user-visible feature.

Should the news item say that the compilation works with homebrew clang-17? (which is apparently different from xcode clang)

I have a response https://github.com/actions/runner-images/issues/13012#issuecomment-3297365182 from the github actions macos-15 image dev, who says: Clang is available as a part of Xcode and separately as an independent application instance installed from the Homebrew. Your https://github.com/Rdatatable/data.table/issues/6622 referring to clang-17 which is not a part of our default config by the way. That means that one of your steps changes the default setup somehow.

tdhock:
to finalize this pr, we should revert the deletions I made in the R CMD check CI config.
But I would suggest keeping the macos-15 build that I added (and changing it to macos-latest), is that OK?

aitap:
> easy installation is a really important user-visible feature

Then we might want to add a `configure` test in addition to the current CI fixes. When running macOS and `$(R CMD config CC) --version` says `Apple clang 17`, check whether `/usr/local/lib/libomp.dylib` exists. If it does exist, use it instead of `-lomp`. If not, suggest installing the OpenMP runtime from https://mac.r-project.org/openmp and try using `-lomp` as before.

The risk is breaking installation from source on systems with Apple clang 17 where `-lomp` works but `/usr/local/lib/libomp.dylib` is broken. This sounds weird but not impossible in theory. Another risk is breaking installation on CRAN builders in the future, when they update to Xcode ≥16.0 and a new version of bundled libomp (so that `-lomp` would be sufficient).

> Your https://github.com/Rdatatable/data.table/issues/6622 referring to clang-17 which is not a part of our default config by the way.

This is technically correct (there's probably no binary named `clang-17` in the `macos-15` image), but doesn't contradict what's stated in #6622: the C compiler we're having problems with is Apple clang version 17 from Xcode 16.4, [default in `macos-15`](https://github.com/Rdatatable/data.table/actions/runs/17779239477/job/50534539088).

aitap:
I [found a test](https://github.com/Rdatatable/data.table/actions/runs/17856985863/job/50778144482) that distinguishes between working and broken OpenMP runtime situations on macOS: it's `#pragma omp schedule(dynamic)`. Without the correct OpenMP runtime on Xcode 16.4, the resulting shared library fails to load due to missing `___kmpc_dispatch_deinit`. This can be used in the configure test (à la #6642 but less intrusive), letting us be more confident about setting `PKG_LIBS=/usr/local/lib/libomp.dylib` for Apple clang 17. Give me some more time and we'll have a proper fix and a NEWS item here.

tdhock:
wow excellent job thanks Ivan!

jeroen:
Could you also add a runner with `macos-15-intel` to the matrix? That way we also test on x86_64, see https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

.github/workflows/R-CMD-check.yaml
```diff
@@ -68,6 +69,21 @@ jobs:
             eval sudo $cmd
           done < <(Rscript -e 'writeLines(remotes::system_requirements("ubuntu", "22.04"))')
 
+      - name: Install R Package Build Dependencies on MacOS, from https://github.com/stan-dev/cmdstanr/pull/1072/files
+        if: runner.os == 'macOS'
+        uses: r-hub/actions/setup-r-sysreqs@v1
+        with:
+          type: 'minimal'
+
+      - name: Intall and configure OpenMP runtime
```
tdhock:
```suggestion
      - name: Install and configure OpenMP runtime
```

.github/workflows/R-CMD-check.yaml
```diff
@@ -26,7 +26,8 @@ jobs:
           # jobs (mainly test-coverage) to run on every commit in PRs so as to not slow down dev.
           # GHA does run these jobs concurrently but even so reducing the load seems like a good idea.
           - {os: windows-latest, r: 'devel'}
-          # - {os: macOS-latest, r: 'release'}    # test-coverage.yaml uses macOS
+          - {os: macos-14, r: 'release'}
```
tdhock:
```suggestion
          - {os: macos-15-intel, r: 'release'}    
```

tdhock:
> Could you also add a runner with `macos-15-intel` to the matrix? That way we also test on x86_64, see https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

before this pr there were no macos github actions checks, so adding 3 seems like an unusually large move.

I would propose to add either 1 macos-15 or 2 macos (macos-15 for arm and macos-15-intel for x86_64).

.github/workflows/R-CMD-check.yaml
tdhock: @aitap is this suggestion ok with you? (adding two new macos jobs to CI)

aitap: Sure! If `macos-15-intel` breaks, it should definitely be fixed.

aitap:
It turns out that our caching scheme was in conflict with having macOS 15 running on both Intel and ARM.

Coincidentally, this enables OpenMP support on FreeBSD. Previously, the `configure` script only tried it for macOS and Linux; now it's tested everywhere, with extra cases on macOS.

Do we want to fail CI if `data.table` installs without OpenMP support on systems where it should be supported?

Would it be feasible for someone to test this with a Homebrew installation of R?

tdhock:
> Do we want to fail CI if data.table installs without OpenMP support on systems where it should be supported?

No, I think it would be better to install single core. Anyway we dont see large speedups when adding cores, for many common operations https://github.com/Anirban166/data.table.threads

jeroen:
FYI, data.table builds on r-universe are broken because of this: http://cran.dev/data.table/buildlog

NEWS.md
```diff
@@ -248,6 +248,8 @@
 
 19. New rolling functions, `frollmin` and `frollprod`, have been implemented, towards [#2778](https://github.com/Rdatatable/data.table/issues/2778). Thanks to @jangorecki for implementation.
 
+20. Enhanced tests for OpenMP support, detecting incompatibilities such as R-bundled runtime _vs._ newer Xcode and testing for a manually installed runtime from <https://mac.r-project.org/openmp>, [#6622](https://github.com/Rdatatable/data.table/issues/6622). Thanks to @dvg-p4 for initial report and testing, @twitched for the pointers, @tdhock and @aitap for the fix.
```
jangorecki: I think this should go into NOTES section of news.md

aitap:
@jeroen, is there a way to get a CRAN-compatible toolchain on the `macos-15` runner? Otherwise macOS builds produced on R-Universe using this runner should (and will, after this PR) have OpenMP disabled: the new compiler requires a new OpenMP runtime, but the new OpenMP runtime is incompatible with the one shipped with base R (and presumably used by other CRAN binaries).

jeroen:
I'm not sure. Could you ask this on the `r-sig-mac` mailing list? I'll try to update the headers to the latest from https://mac.r-project.org/openmp/ but IIUC they need to ship a newer openmp dylib with base R?

jeroen:
OK I found a workaround for GitHub actions:

The MacOS 15 GHA runners (both intel and arm64) also have some older versions of Xcode installed, see: https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md

If we look at https://gist.github.com/yamaya/2924292, the last version of xcode that had clang 16 was Xcode_16.2.app. We can activate that using the following command:

```sh
sudo xcode-select -s /Applications/Xcode_16.2.app
```

After which the current master branch of `data.table` was able to build on MacOS 15 with OMP. 

Of course this doesn't solve the problem but hopefully buys us some time until R for MacOS updates their libomp.dylib.

aitap:
Thank you very much @jeroen for finding a solution. Sorry I couldn't
allocate the time to help (I need to finish this PhD before it finishes
me). I know there are plans to update the toolchain version and thus
the OpenMP runtime, but I don't remember how specific they are.
(Possibly for 4.6.0?)

aitap:
Dear @barracuda156, @kevinushey:

This rewrite of the `configure` script touches the use-cases (#6034, #6409) that I cannot easily test. Could you please help testing the new code? I'm worried about macOS configurations that aren't Xcode 14.x + the OpenMP runtime bundled with R or latest Xcode + the OpenMP runtime from https://mac.r-project.org/openmp.

The test should succeed if all exit codes are zero and the final command shows that some `_OPENMP` version is detected:
```sh
R CMD build .
mkdir -p data.table.Rcheck
R CMD INSTALL -l data.table.Rcheck data.table_1.17.99.tar.gz
R_LIBS_USER=data.table.Rcheck:%U R -q -s -e 'data.table::getDTthreads(TRUE)'
```
In particular, I'm interested in the `configure` script output shown by `R CMD INSTALL`. I think that at least some of the special cases will now be covered by our use of `$(SHLIB_OPENMP_CFLAGS)`, but I'm less sure about Homebrew (and the many other possible configurations possible on macOS).

jeroen:
The libomp.dylib version that is included with R for MacOS has been updated in R-4.6-arm64 build 2025-12-14 r89168 to a newer version, which should be compatible with xcode 16.3 and up as it includes the `___kmpc_dispatch_deinit` symbol.

For the R version 4.5 and lower, as well as all x86_64 versions the libomp.dylib does not get updated, so it is still needed to compile packages with xcode 16.2 or lower.

FYI users can completely avoid all problems by installing prebuilt binaries from https://rdatatable.r-universe.dev/data.table

configure
```diff
@@ -68,89 +68,79 @@ fi
...
-      if ${CC} ${CFLAGS} ${LIBOMP_INCLUDE} ${LIBOMP_LINK} test-omp.c >> config.log 2>&1; then
-        echo "yes"
-        export PKG_CFLAGS="${PKG_CFLAGS} ${LIBOMP_INCLUDE}"
-        export PKG_LIBS="${PKG_LIBS} ${LIBOMP_LINK}"
-        export R_OPENMP_ENABLED=1
-        return
-      else
-        echo "no"
-      fi
+    test -e "${HOMEBREW_PREFIX}/opt/libomp" ] \
```
ben-schwen:
```suggestion
    test -e "${HOMEBREW_PREFIX}/opt/libomp" \
```

ben-schwen (approved):
LGTM besides the unmatched `[` (see suggested change)

aitap:
Manually tested on a 10.15 Hackintosh with Clang 12 from Xcode CLT:

* CRAN build of R-4.2.3 + OpenMP runtime from https://mac.r-project.org/openmp/ + `pkg-config` and `zlib` stubs from https://mac.r-project.org/tools/: passes test for `LIBS=-lomp`
* MacPorts build of R-4.5.2:
  * `+openmp`: passes test for `$(OPENMP_SHLIB_CFLAGS)`
  * `-openmp`: passes test for `-fopenmp` (is this the right thing to do in a build of R where OpenMP was explicitly disabled?)
* Homebrew doesn't really support 10.15 and fails to download the `nasm` source tarball, I gave up
* GitLab CI used to pick up `libomp` from Homebrew, and now the jobs fail: [1](https://gitlab.com/Rdatatable/data.table/-/jobs/12595194227) [2](https://gitlab.com/Rdatatable/data.table/-/jobs/12595194224). The runtime is alive and well, but the `omp.h` header is missing. According to [Simon](https://stat.ethz.ch/pipermail/r-sig-mac/2025-April/015165.html), this is deliberate: if a user wants to compile OpenMP-using packages from source, they really should follow the instructions at https://mac.r-project.org/tools/. I'll fix the `configure` test to check for the header too, but won't change GLCI configuration so that the Homebrew branch is still exercised.

barracuda156:
@aitap MacPorts does not build R with Xcode clang. It picks specific LLVM clang (on most systems) and specific gcc (for gfortran on those, and as the primary compiler on legacy ones).
In both cases OpenMP _is_ enabled: https://github.com/macports/macports-ports/blob/127c979274f93fd322bfe639871143caf3cfedd5/math/R/Portfile#L290

So unless you built R in a custom way with `--enforce-variants -openmp`, it is expected that OpenMP is detected.

P. S. R packages are stale in upstream MacPorts, but maintained here: https://github.com/macos-powerpc/R-ports
This repo is a submodule, which replaces R directory in ports. If you make a local overlay and git-clone it as `R`, add the path to `sources.conf` and run port sync, you can install R ports coherently with just `port install`.

aitap:
Thank you for providing the information, @barracuda156!

@Rdatatable/committers, should we merge now despite the limited testing with Homebrew? After merging, should this change be included into the patch branch, or kept on `master` until `1.20`? FWIW both GitHub and GitLab CI passes on macOS ([1](https://gitlab.com/Rdatatable/data.table/-/jobs/13338824799) [2](https://gitlab.com/Rdatatable/data.table/-/jobs/13341825328)), although there's probably an infinite number of configurations where it still doesn't work.

ben-schwen:
> @Rdatatable/committers, should we merge now despite the limited testing with Homebrew? After merging, should this change be included into the patch branch, or kept on `master` until `1.20`? FWIW both GitHub and GitLab CI passes on macOS ([1](https://gitlab.com/Rdatatable/data.table/-/jobs/13338824799) [2](https://gitlab.com/Rdatatable/data.table/-/jobs/13341825328)), although there's probably an infinite number of configurations where it still doesn't work.

Fine for me. Since it seems to be an improvement to the current situtation I would put it into the patch branch, which might warrant some additional issues/infos about configurations that do not work.

PR #7320, opened 2025-09-15
frollapply simplify more conservative

### LINKED ISSUE #7317: review simplify=TRUE rules in frollapply()
After https://github.com/Rdatatable/data.table/pull/7314 our simplify rules got very wide.
PR meant to add handling for adaptive rolling median.

Considering `frollapply` manual says

> In future version we may improve internal \code{simplifylist} function, then \code{simplify=TRUE} may return object of a different type, breaking downstream code. If results are already simplified with \code{simplify=TRUE}, then it can be considered backward compatible.

it is risky to keep current simplify rules.

We should either:

1. change manual that even simplified results may not be considered backward compatible
2. or make simplify to work for a less cases

If we go with 2. then this issue can be good for agreeing on the scope of `simplifylist()`.

Absolute minimum are:
- coerce int to double when varies within answer results (#7314)
- coerce `fill` to answers type
- recycle `fill` to answers length
- `unlist` for atomic vector of matching types
- `rbindlist` of list/df/dt of matching dimension and types

jangorecki:
closes #7317

In general much less objects might get simplified now
but some features have been added
- matching names between `fill` and FUN results - this allowed to simplified examples
- more flexible recycling of default fill=NA

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -177,20 +177,20 @@
     ```r
     x = data.table(v1=rnorm(120), v2=rnorm(120))
     f = function(x) coef(lm(v2 ~ v1, data=x))
-    coef.fill = c("(Intercept)"=NA_real_, "v1"=NA_real_)
-    frollapply(x, 4, f, by.column=FALSE, fill=coef.fill)
+    frollapply(x, 4, f, by.column=FALSE)
     #     (Intercept)         v1
+    #           <num>      <num>
     #  1:          NA         NA
     #  2:          NA         NA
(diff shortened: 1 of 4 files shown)


jangorecki:
This PR is hanging now for 9 days, it is not long, but considering it is a polishing of tiny part of a complete rewrite of `frollapply` that happened in this dev cycle, it feels it will do more good by being merged sooner, and give more opportunity for testing by the community. Of course reviews are welcome, even after it is merged!

PR #7322, opened 2025-09-17
suspendInterrupts missing on 3.4

jangorecki:
suspendInterrupts() was used in frollapply() but it happens to be not available in R 3.4

this should do


MichaelChirico:
confirmed it was added in 3.5.0:

https://github.com/r-devel/r-svn/blob/96f23c268cf6a3983da5242f06200fd7063f8fbf/doc/NEWS.3#L1563

also noticed it's absent from {backports}

https://github.com/r-lib/backports/issues/90

PR #7324, opened 2025-09-17
testing for broken selfref pointer adapted for macos

### LINKED ISSUE #7323: R macos jobs fails
https://rdatatable.gitlab.io/data.table/web/checks/data.table/test-mac-rel/00check.log
https://rdatatable.gitlab.io/data.table/web/checks/data.table/test-mac-old/00check.log

macos jobs seem to fail on the function that is testing proper self-referencing of data.table
```r
is.ok = function(x) {
  stopifnot(is.data.table(x))
  capture.output(print(attr(x, ".internal.selfref", TRUE)))!="<pointer: (nil)>"
}
```

jangorecki:
most likely closes #7323

### DIFF
--- a/man/frollapply.Rd
+++ b/man/frollapply.Rd
@@ -84,7 +84,11 @@ setDTthreads(old)
 }
         \item Objects returned from forked processes, \code{FUN}, are serialized. This may cause problems for objects that are meant not to be serialized, like data.table. We are handling that for data.table class internally in \code{frollapply} whenever \code{FUN} is returning data.table (which is checked on the results of the first \code{FUN} call so it assumes function is type stable). If data.table is nested in another object returned from \code{FUN} then the problem may still manifest, in such case one has to call \code{setDT} on objects returned from \code{FUN}. This can be also nicely handled via \code{simplify} argument when passing a function that calls \code{setDT} on nested data.table objects returned from \code{FUN}. Anyway, returning data.table from \code{FUN} should, in majority of cases, be avoided from the performance reasons, see \emph{UDF optimization} section for details.
 \preformatted{
-is.ok = function(x) {stopifnot(is.data.table(x)); format(attr(x, ".internal.selfref", TRUE))!="<pointer: (nil)>"}
+is.ok = function(x) {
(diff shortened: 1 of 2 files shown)


man/frollapply.Rd
```diff
@@ -84,7 +84,11 @@ setDTthreads(old)
 }
         \item Objects returned from forked processes, \code{FUN}, are serialized. This may cause problems for objects that are meant not to be serialized, like data.table. We are handling that for data.table class internally in \code{frollapply} whenever \code{FUN} is returning data.table (which is checked on the results of the first \code{FUN} call so it assumes function is type stable). If data.table is nested in another object returned from \code{FUN} then the problem may still manifest, in such case one has to call \code{setDT} on objects returned from \code{FUN}. This can be also nicely handled via \code{simplify} argument when passing a function that calls \code{setDT} on nested data.table objects returned from \code{FUN}. Anyway, returning data.table from \code{FUN} should, in majority of cases, be avoided from the performance reasons, see \emph{UDF optimization} section for details.
 \preformatted{
-is.ok = function(x) {stopifnot(is.data.table(x)); format(attr(x, ".internal.selfref", TRUE))!="<pointer: (nil)>"}
+is.ok = function(x) {
```
MichaelChirico, line 87: why do we need this in the documentation?

jangorecki: To explain examples in the lines below that one.

jangorecki: We could eventually have exported function only for testing that as this may be useful for users who are not sure if DT needs to setDT after loading from disk or if it was re-serialized like here.

MichaelChirico: I think as much as possible we should avoid mentioning .internal.selfref to users; it's an implementation detail that only data.table devs should concern themselves with.

jangorecki: So do you think it make sense to add a helper function and hide this behind that function?

MichaelChirico: I think it makes sense. Let's discuss in #7329

PR #7325, opened 2025-09-17
switch to snake case + minor spelling correction

badasahog:
In many other places in the code, `if` is spelled `iff`. I didn't fix the others to avoid creating conflicts.

Also, I've updated type case to snake case. The codebase is mostly snake case, so that's what I think we should go with. If you agree, I will do so more comprehensively.

### DIFF
--- a/src/fread.c
+++ b/src/fread.c
@@ -47,7 +47,7 @@ enum quote_rule_t
     // Fields may be quoted, but any quotes inside will appear verbatim and
     // not escaped in any way. It is not always possible to parse the file
     // unambiguously, but we give it a try anyways. A quote will be presumed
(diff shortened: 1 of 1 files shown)


aitap:
> In many other places in the code, `if` is spelled `iff`. 

That's an abbreviation for
<https://en.wikipedia.org/wiki/If_and_only_if>.

badasahog:
@aitap understood. reverting.

ben-schwen:
I don't think will adapt to this since the R C interface has a lot of camelCase e.g. `allocVector`, `isString`, `getAttrib`, so we can't efficiently cover this with a linter.

In general, we could but we have at least >2k appearances of camelCase touching most of our codebase, so I would rather not.

Issue #7326, opened 2025-09-18
fread() cannot read .gz compressed files after upgrading R to 4.5


mhyx1122:
Description:

Hello,

After upgrading R from version 4.4 to 4.5, I noticed that the fread() function in the data.table package no longer works with .gz compressed files. This used to work perfectly fine in R 4.4.

The error message I am encountering is as follows:

**regexpr(as.character(pattern), text, ignore.case, perl, fixed, : 
  NA in coercion to boolean**

It seems that this issue might be related to updates in R's base functionality, which could have caused a change in how regexpr() works.

I suspect that this issue was caused by changes in the base R package, leading to some unexpected behavior in regexpr().

I would appreciate it if this could be fixed, and thank you for your help in advance.

Looking forward to the fix.

Best regards.

mhyx1122:
We found that data.table package, when handling compressed files, first attempts to decompress the file using . Upon further investigation, I discovered that the error is coming from R.utils when attempting to decompress .gz compressed files. This leads to the following error message:
regexpr(as.character(pattern), text, ignore.case, perl, fixed, :
NA in coercion to boolean
Is it possible to bypass the R.utils package and use an alternative method to decompress .gz files in ord
er to resolve this issue?
Looking forward to your suggestions. Thanks in advance!

aitap:
Try upgrading R.utils. See #6962.

mhyx1122:
> Try upgrading R.utils. See [#6962](https://github.com/Rdatatable/data.table/issues/6962).

Thank you very much! The issue has been resolved. I looked at the latest version of R.utils on GitHub, which was v2.12.3, and mistakenly assumed that it was the latest version. However, I should have downloaded it from CRAN using install.packages("R.utils") to get the most up-to-date version. I hope this information helps others as well.

mhyx1122:
> Try upgrading R.utils. See [#6962](https://github.com/Rdatatable/data.table/issues/6962).

Thank you very much! The issue has been resolved. I looked at the latest version of R.utils on GitHub, which was v2.12.3, and mistakenly assumed that it was the latest version. However, I should have downloaded it from CRAN using install.packages("R.utils") to get the most up-to-date version(2.13.0). I hope this information helps others as well.

Issue #6962, opened 2025-05-12
fread cannot open the csv.gz generated with fwrite


FrankwaP:
Hello,

I am running a code that used to work but is now having problem with `fread` on gz files generated with `fwrite`.  
I probably went from R4.4 to R4.5 between then and now…

```
Erreur dans regexpr(as.character(pattern), text, ignore.case, perl, fixed,  : 
  NA in coercion to boolean
```

I've check and the problem does not happens with csv or csv.bz files.

`#` [`Minimal reproducible example`](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example); please be sure to set `verbose=TRUE` where possible!

```{r}
filename <- 'test.csv'
data.table::fwrite(mtcars, filename)
new_data <- data.table::fread(filename)
cat(filename, "OK!\n")

filename <- 'test.csv.bz'
data.table::fwrite(mtcars, filename)
new_data <- data.table::fread(filename)
cat(filename, "OK!\n")

filename <- 'test.csv.gz'
data.table::fwrite(mtcars, filename)
new_data <- data.table::fread(filename)
cat(filename, "OK!\n")  # will not reach this line
```

`#` `Output of sessionInfo()`
```
R version 4.5.0 (2025-04-11)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 22.04.5 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3 
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so;  LAPACK version 3.10.0

locale:
 [1] LC_CTYPE=fr_FR.UTF-8       LC_NUMERIC=C               LC_TIME=fr_FR.UTF-8       
 [4] LC_COLLATE=fr_FR.UTF-8     LC_MONETARY=fr_FR.UTF-8    LC_MESSAGES=fr_FR.UTF-8   
 [7] LC_PAPER=fr_FR.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=fr_FR.UTF-8 LC_IDENTIFICATION=C       

time zone: Europe/Paris
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[...]

Issue #5601, opened 2023-02-18
Testsuite on PPC: 2 errors out of 1161


barracuda156:
```
* checking tests ...
  Running ‘autoprint.R’
  Comparing ‘autoprint.Rout’ to ‘autoprint.Rout.save’ ... OK
  Running ‘froll.R’
 ERROR
Running the tests in ‘tests/froll.R’ failed.
Last 13 lines of output:
   [9] 4.333333 4.500000 4.142857 5.000000
  
  [[3]]
   [1]       NA 0.750000 1.250000       NA 1.375000 1.400000 2.000000 2.500000
   [9] 2.250000 2.333333 2.250000 2.666667
  
  [[4]]
   [1]       NA       NA 1.000000       NA 1.200000 1.250000 1.750000 2.250000
   [9] 2.166667 2.250000 2.071429 2.500000
  
  
  Sat Feb 18 12:12:00 2023  endian==big, sizeof(long double)==16, longdouble.digits==1024, sizeof(pointer)==4, TZ==unset, Sys.timezone()=='Asia/Taipei', Sys.getlocale()=='C/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8', l10n_info()=='MBCS=TRUE; UTF-8=TRUE; Latin-1=FALSE; codeset=UTF-8', getDTthreads()=='This installation of data.table has not been compiled with OpenMP support.; omp_get_num_procs()==1; R_DATATABLE_NUM_PROCS_PERCENT==unset (default 50); R_DATATABLE_NUM_THREADS==unset; R_DATATABLE_THROTTLE==unset (default 1024); omp_get_thread_limit()==1; omp_get_max_threads()==1; OMP_THREAD_LIMIT==unset; OMP_NUM_THREADS==unset; RestoreAfterFork==true; data.table is using 1 threads with throttle==1024. See ?setDTthreads.', zlibVersion()==1.2.13 ZLIB_VERSION==1.2.13
  Error in test.data.table(script = "froll.Rraw") : 
    2 errors out of 1161. Search tests/froll.Rraw for test numbers: 6000.137, 6000.138.
  Execution halted
```

MichaelChirico:
We rely on outside help to fix something like this, PRs welcome (including possibly extending our CI suite to include this platform somehow)

barracuda156:
[...]

PR #7328, opened 2025-09-19
Add a version dependency for R.utils

MichaelChirico:
More than one user has been bitten by the bug in {R.utils} looking like it's a {data.table} bug: #7326, #6962, #5601

Since we only use {R.utils} for `decompressFile()`, I think it's good on balance to include this version requirement.

Technically it only affects users with `{R.utils} < 2.13.0` *AND* `R >= 4.5.0`, but I'm not sure it's worth quibbling given that (1) {R.utils} is in `Suggests` and (2) the new version has been on CRAN for >6 months already.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -3,7 +3,7 @@ Version: 1.17.99
 Title: Extension of `data.frame`
 Depends: R (>= 3.4.0)
 Imports: methods
-Suggests: bit64 (>= 4.0.0), bit (>= 4.0.4), R.utils, xts, zoo (>= 1.8-1), yaml, knitr, markdown
+Suggests: bit64 (>= 4.0.0), bit (>= 4.0.4), R.utils (>= 2.13.0), xts, zoo (>= 1.8-1), yaml, knitr, markdown
 Description: Fast aggregation of large data (e.g. 100GB in RAM), fast ordered joins, fast add/modify/delete of columns by group using no copies at all, list columns, friendly and fast character-separated-value read/write. Offers a natural and flexible syntax, for faster development.
 License: MPL-2.0 | file LICENSE
 URL: https://r-datatable.com, https://Rdatatable.gitlab.io/data.table, https://github.com/Rdatatable/data.table


jangorecki:
Is this version in any way enforced the way imports are?

Whenever we use it we could check for version and raise meaningful error.

MichaelChirico:
my memory is that they're not, but that it's considered something of a bug in R (can't find anything specific there, and not tested). I just wanted to at least annotate in DESCRIPTION as a minimal step to say "we told you so" :)

we could consider a runtime version check, is it worth it? I guess it comes up often enough...

jangorecki:
Added

R/fread.R
```diff
@@ -119,6 +119,8 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     if ((w <- endsWithAny(file, c(".gz", ".bgz",".bz2"))) || (gzsig <- is_gzip(file_signature)) || is_bzip(file_signature)) {
       if (!requireNamespace("R.utils", quietly = TRUE))
         stopf("To read %s files directly, fread() requires 'R.utils' package which cannot be found. Please install 'R.utils' using 'install.packages('R.utils')'.", if (w<=2L || gzsig) "gz" else "bz2") # nocov
+      if (packageVersion("R.utils") < "2.13.0")
```
MichaelChirico:
```suggestion
      if (getRversion() >= "4.5.0" && packageVersion("R.utils") < "2.13.0")
```

R/fread.R
```diff
@@ -119,6 +119,8 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     if ((w <- endsWithAny(file, c(".gz", ".bgz",".bz2"))) || (gzsig <- is_gzip(file_signature)) || is_bzip(file_signature)) {
       if (!requireNamespace("R.utils", quietly = TRUE))
         stopf("To read %s files directly, fread() requires 'R.utils' package which cannot be found. Please install 'R.utils' using 'install.packages('R.utils')'.", if (w<=2L || gzsig) "gz" else "bz2") # nocov
+      if (getRversion() >= "4.5.0" && packageVersion("R.utils") < "2.13.0")
```
MichaelChirico:
```suggestion
      # not worth doing a behavior test here, so just use getRversion().
      if (packageVersion("R.utils") < "2.13.0" && base::getRversion() >= "4.5.0")
```

Issue #2334, opened 2017-09-06
Add is.valid(DT) function


mattdowle:
Excellent PR #2332 (which removed the repeated and wasteful check that current keys are valid) triggered the thought that perhaps we need an `is.valid(DT)` function for use by users.  It would check that `DT` is indeed sorted by the key (if any) and that the indexes (if any) do indeed sort those index columns correctly. Although unlikely, it is possible for advanced users to go 'under the hood' and change attributes directly and that's when mistakes can happen. They can decide to use `is.valid(DT)` to check all is well at a moment that suits them.

MichaelChirico:
Not sure how I feel about the name (`validate`? `valid_keys`? `check_sorting`? etc), but the function itself would find use.

jangorecki:
some extra checks that could be included as well
- unique column names
- ascii only column names (see #4351)

PR #7332, opened 2025-09-20
selfref.ok() helper to close #7329

### LINKED ISSUE #7329: Add a function for validating data.tables that might need setDT
We could eventually have exported function only for testing that as this may be useful for users who are not sure if DT needs to setDT after loading from disk or if it was re-serialized like here.

_Originally posted by @jangorecki in https://github.com/Rdatatable/data.table/pull/7324#discussion_r2356325641_

jangorecki:
I put it into NOTES rather than NEW FEATURES because it is very technical function, and because reason for exporting is to make documentation of frollapply less "internal" and not really throw it onto users and saying "please use new function".

I keep the original name because this function does not check anything else. Ultimately, when implementing #2334 we should use this function for testing selfref.

closes #7329


MichaelChirico:
Not a big fan of the naming -- even as a highly technical function, it's too specific IMO. I still prefer something like 'validateDT' that conveys the actual functionality, rather then the technical implementation.

If you insist on something including 'selfref', let's please prefix it with `.` to further convey it's not a typical export & only meant for highly technical end users.

NEWS.md
```diff
@@ -311,6 +311,8 @@
 
 6. Using a double vector in `set()`'s `i=` and/or `j=` no longer throws a warning about preferring integer, [#6594](https://github.com/Rdatatable/data.table/issues/6594). While it may improve efficiency to use integer, there's no guarantee it's an improvement and the difference is likely to be minimal. The coercion will still be reported under `datatable.verbose=TRUE`. For package/production use cases, static analyzers such as `lintr::implicit_integer_linter()` can also report when numeric literals should be rewritten as integer literals.
 
+7. In rare situations a data.table object may lose its internal attribute that holds a self-reference. New helper function `selfref.ok()` tests just that. See manual for examples.
```
```suggestion
7. In rare situations a data.table object may lose its internal attribute that holds a self-reference. New helper function `.selfref.ok()` tests just that. It is only intended for technical use cases. See manual for examples.
```

man/.selfref.ok.Rd
```diff
@@ -0,0 +1,34 @@
+\name{.selfref.ok}
+\alias{.selfref.ok}
+\title{Tests self reference of a data.table}
+\description{
+  In rare situations, as presented in examples below, a data.table object may lose its internal attribute that holds a self-reference. This function tests just that.
```
MichaelChirico, line 5:
```suggestion
  In rare situations, as presented in examples below, a data.table object may lose its internal attribute that holds a self-reference. This function tests just that.

  It is not expected that many end users will have need for this highly technical function about data.table internals.
```

R/helpers.R
```diff
@@ -10,3 +10,10 @@ fctr = function(x, levels=unique(x), ..., sort=FALSE, rev=FALSE) {
   if (rev) levels = frev(levels)
   factor(x, levels=levels, ...)
 }
+
+# add a function for validating data.tables that might need setDT #7329
+.selfref.ok = function(x) {
+  if (!is.data.table(x))
+    stopf(".selfref.ok expects data.table class object.")
+  selfrefok(x, verbose=FALSE) > 0L
```
MichaelChirico, line 18: Why force `FALSE`? seems useful to respect datatable.verbose.

jangorecki: Because use of this is very narrow, we don't want even more internals to be pushed on users.

jangorecki:
> Let's remove the other selfref reference in the manual?
> 
> ```
> man/frollapply.Rd:  out = capture.output(print(attr(x, ".internal.selfref", TRUE)))
> ```

Will do in a follow up PR

jangorecki:
added in https://github.com/Rdatatable/data.table/pull/7320/commits/37423fb6d509eddded6ab3cb7c6b01a520f782dc

PR #7337, opened 2025-09-22
rename to correct name and simplify website for R 4.4+

jangorecki:
not sure why it was named pkgdown... now fixed :)
Also uses R 4.4+ new function `tools::pkg2HTML`


jangorecki:
base R html manual available at: https://rdatatable.github.io/data.table/
which is very nice for CTRL+F searching

PR #7340, opened 2025-09-25
dev version suffix

### LINKED ISSUE #7339: increment devel version number
Recently I came across on how posit package manager handles version number for development packages.
It is appending a timestamp of a last git commit to the version of the package
Simple yet brilliant idea.
Once adopted in DT then we don't need `update_dev_pkg()` anymore, or we can redirect it to `update.packages(repos=.)`.

I would say we have two options:

- `1.17.99.1758749285`
- `1.17.99.20250924232805`

That would happen just before package build, the same place where we currently save git commit sha into description file.
It will be visible only for our devel packages published on gitlab.io or github.io

----

Oh, `--as-cran` check does not allow to many dots in the version string, therefore `-` must be used
Oh2, number after dash must be parsed to int32 so `-20250101.101112` can be used as a suffix
Oh3, number must not contains leading zeroes so we must not commit before 10:00 AM UTC ;)

unix epoch time seems to be the way to go

jangorecki:
closes #7339

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -77,6 +77,8 @@ build:
     - rm -r bus
   script:
     - sed -i '/^[[:space:]]*$/d' ./DESCRIPTION ## make last line end abruptly; i.e. without a final \n
+    - |
+      sed -i "0,/^Version: [0-9.]\+$/s//&.$(TZ=UTC git log -1 --format=%cd --date=format-local:'%Y%m%d%H%M%S')/" ./DESCRIPTION
     - echo "Revision:" $CI_COMMIT_SHA >> ./DESCRIPTION
     - R CMD build .
     - mkdir -p bus/$CI_JOB_NAME/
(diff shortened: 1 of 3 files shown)


tdhock:
great thanks

.gitlab-ci.yml
```diff
@@ -77,6 +77,8 @@ build:
     - rm -r bus
   script:
     - sed -i '/^[[:space:]]*$/d' ./DESCRIPTION ## make last line end abruptly; i.e. without a final \n
+    - |
+      sed -i "0,/^Version: [0-9.]\+$/s//&-$(TZ=UTC git log -1 --format=%ct)/" ./DESCRIPTION
```
ben-schwen, line 81: @jangorecki it seems that this broke our CI pipeline @Gitlab?

jangorecki: Thanks Ben, submitted fix

PR #7347, opened 2025-09-27
fix gcc flag warnings

ben-schwen:
Currently we get various warnings when installing due to the `-Wpedantic` and `-Wmaybe-uninitialized` flags. Although those do no harm, they cluster the output with unnecessary warnings.

Example:
```c
  fread.c:2822:71: warning: ISO C forbids empty initializer braces [-Wpedantic]
   2822 |           DTi + row1line, ncol, tt, strlim(skippedFooter, (char[500]) {}, 500));
  fifelse.c:286:10: warning: ‘ans_is_factor’ may be used uninitialized in this function [-Wmaybe-uninitialized]
    286 |       if (!naout && ans_is_factor) {
```


jangorecki:
Thanks, atime job failure looks like a false positive

PR #7348, opened 2025-09-27
unit tests for already resolved sort by reference issue

### LINKED ISSUE #710: Mixing use of data.frame and data.table in R 3.1+
This is something we should basically be aware of. R v3.1.0+ has nice changes and cleverly avoids deep copies wherever possible. And therefore, scenarios like this, however rare, might happen. And we should provide a fix here.

``` R
DF1 <- data.frame(x=1:5, y=6:10, z=11:15)
DF2 <- transform(DF1, z = 16:20)

require(data.table)
setorder(setDT(DF2), -z)

DF1
#   x  y  z
#1 5 10 11
#2 4  9 12
#3 3  8 13
#4 2  7 14
#5 1  6 15
DF2
#    x  y  z
#1: 5 10 20
#2: 4  9 19
#3: 3  8 18
#4: 2  7 17
#5: 1  6 16
```

The fact that `x` and `y` are shallow copied results in `DF1`'s columns being affected as well.

On a side note, this also means that if one were to write a function to operate on `DF2` using the `C` API (by reference), then the same thing would happen.

jangorecki:
closes #710

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21685,3 +21685,10 @@ d3 = unserialize(serialize(d2, NULL))
 test(2340.05, .selfref.ok(d3), FALSE)
 setDT(d3)
 test(2340.06, .selfref.ok(d3), TRUE)
+
+# Mixing use of data.frame and data.table in R 3.1+ #710
+DF1 = data.frame(x=1:5, y=6:10, z=11:15)
+DF2 = transform(DF1, z = 16:20)
+setorder(setDT(DF2), -z)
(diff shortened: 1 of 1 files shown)


ben-schwen:
Unfortunately the problem still persists with other types than integer
```r
DF1 <- data.frame(b=letters[1:3], c=c(1.1, 2.2, 3.3))
DF2 <- transform(DF1, c = c(4.4, 5.5, 6.6))
setorder(setDT(DF2), -c)
```

MichaelChirico:
Closing as it doesn't really resolve the issue, fix would eventually be more involved, and easiest to repopulate the test from scratch

PR #7361, opened 2025-10-13
frollvar and frollsd

### LINKED ISSUE #2778: rolling functions, rolling aggregates, sliding window, moving average

jangorecki:
closes #2778 

```r
library(roll)
library(data.table)
set.seed(108)
x = rnorm(1e8)
n = 1e3
system.time(ans1 <- roll_var(x, n))
#   user  system elapsed 
#  3.296   0.561   3.858 
system.time(ans2 <- frollvar(x, n))
#   user  system elapsed 
#  1.421   0.230   1.652 
all.equal(ans1, ans2) && all.equal(ans1, ans3)
#[1] TRUE
```

### DIFF
--- a/src/froll.c
+++ b/src/froll.c
@@ -1137,8 +1137,6 @@ void frollprodExact(const double *x, uint64_t nx, ans_t *ans, int k, double fill
   }
 }
 
-#define CLAMP0(x) (((x) < 0) ? 0 : (x))
-
 /* fast rolling var - fast
   Welford's online algorithm
   numerically stable
@@ -1172,7 +1170,7 @@ void frollvarFast(const double *x, uint64_t nx, ans_t *ans, int k, double fill, 
     wmean += delta / (i + 1);
     m2 += delta * (x[i] - wmean);
     double ans_i = m2 / k0;
-    ans->dbl_v[i] = CLAMP0(ans_i);
+    ans->dbl_v[i] = MAX(0,ans_i);
     if (R_FINITE((double) m2)) {
       for (uint64_t i=k; i<nx; i++) {
           double x_out = x[i-k];
@@ -1184,7 +1182,7 @@ void frollvarFast(const double *x, uint64_t nx, ans_t *ans, int k, double fill, 
           wmean += delta_in / k;
           m2 += delta_in * (x_in - wmean);
           double ans_i = m2 / k0;
-          ans->dbl_v[i] = CLAMP0(ans_i);
+          ans->dbl_v[i] = MAX(0,ans_i);
       }
       if (!R_FINITE((double) m2)) {
         if (hasnf==-1)
@@ -1255,7 +1253,7 @@ void frollvarExact(const double *x, uint64_t nx, ans_t *ans, int k, double fill,
           wsumxi += (xi * xi);
         }
         double ans_i = wsumxi / (k - 1);
-        ans->dbl_v[i] = CLAMP0(ans_i);
+        ans->dbl_v[i] = MAX(0,ans_i);
       }
     }
     if (truehasnf) {
@@ -1291,7 +1289,7 @@ void frollvarExact(const double *x, uint64_t nx, ans_t *ans, int k, double fill,
             wsumxi += (xi * xi);
           }
           double ans_i = wsumxi / (k - 1);
-          ans->dbl_v[i] = CLAMP0(ans_i);
+          ans->dbl_v[i] = MAX(0,ans_i);
         } else if (nc < (k - 1)) { // var(scalar) is also NA thats why k-1 so at least 2 numbers must be there
           long double wmean = wsum / (k - nc);
           long double xi = 0.0;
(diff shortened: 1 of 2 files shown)


jangorecki:
@tdhock @Anirban166 it seems that atime is raising false positive again, do you know if it is possible to set this test job as "allowed to fail" as it can be in gitlab? so it will still run, but its failure won't cascade to mark general status of PR as failed

Anirban166:
You can try using the `continue-on-error` directive at the [job](https://github.com/Rdatatable/data.table/blob/master/.github/workflows/performance-tests.yml#L14) level. Maybe base it off labels (then apply that label to this PR) since using `pull_request.number` would be redundant; e.g. say we name the label 'ignore-atime-failure':
```yaml
continue-on-error: ${{ contains(github.event.pull_request.labels.*.name, 'ignore-atime-failure') }}
```

src/froll.c
```diff
@@ -1123,6 +1137,197 @@ void frollprodExact(const double *x, uint64_t nx, ans_t *ans, int k, double fill
...
+      long double wsum = 0.0;
+      for (int j=-k+1; j<=0; j++) {
+        wsum += x[i+j];
+      }
+      if (!R_FINITE((double) wsum)) {
+        if (ISNAN((double) wsum)) {
+          if (!narm) {
+            ans->dbl_v[i] = (double) wsum;
+          }
+          truehasnf = true;
```
ben-schwen, line 1243: I know I'm picky on this one, but I just checked the [openmc documentation](https://www.openmp.org/spec-html/5.2/openmpsu9.html#x18-170001.4.1) and it seems that our common used idiom setting boolean flag is UB in openmp.

> Two memory operations are considered unordered if the order in which they must complete, as seen by their aﬀected threads, is not speciﬁed by the memory consistency guarantees listed in [Section 1.4.6](https://www.openmp.org/spec-html/5.2/openmpsu14.html#x23-220001.4.6). If multiple threads write to the same memory unit (deﬁned consistently with the above access considerations) then a data race occurs if the writes are unordered. Similarly, if at least one thread reads from a memory unit and at least one thread writes to that same memory unit then a data race occurs if the read and write are unordered. If a data race occurs then the result of the program is unspeciﬁed.

ben-schwen: I posted it as follow up since we need to revisit this in our whole code!

src/froll.c
```diff
@@ -1123,6 +1137,197 @@ void frollprodExact(const double *x, uint64_t nx, ans_t *ans, int k, double fill
...
+      for (uint64_t i=k; i<nx; i++) {
+          double x_out = x[i-k];
+          long double delta_out = x_out - wmean;
+          wmean = (wmean * k - x_out) / k0;
+          m2 -= delta_out * (x_out - wmean);
+          double x_in = x[i];
+          long double delta_in = x_in - wmean;
+          wmean += delta_in / k;
+          m2 += delta_in * (x_in - wmean);
+          ans->dbl_v[i] = (double)(m2 / k0);
```
ben-schwen: shouldn't we clamp here to avoid sqrt(-0)?

Constructured example:

```r
y=c(1e8+2.980232e-8,1e8,1e8,1e8)
frollvar(y, n=3)
# [1] NA NA  0.00000000000000029604140319716698570573499 -0.00000000000000000005421010862427522170037
frollsd(y, n=3)
# [1] NA NA 0.00000001720585374798841343197 NaN

```

ben-schwen: Same ofc applies to exact.

jangorecki: also fixed for adaptive, PTAL

ben-schwen:
LGTM but I would either move `CLAMP0(x)` to `data.table.h` or use `MAX(0,x)` (which is also defined in `data.table.h`)

ben-schwen (approved):

PR #7368, opened 2025-10-16
Changed many function calls to use read only variants when appropriate

badasahog:
This should improve speed and readability

### DIFF
--- a/src/transpose.c
+++ b/src/transpose.c
@@ -9,7 +9,7 @@ SEXP transpose(SEXP l, SEXP fill, SEXP ignoreArg, SEXP keepNamesArg, SEXP listCo
     error(_("l must be a list."));
   if (!length(l))
     return(copyAsPlain(l));
-  if (!isLogical(ignoreArg) || LOGICAL(ignoreArg)[0] == NA_LOGICAL)
+  if (!isLogical(ignoreArg) || LOGICAL_RO(ignoreArg)[0] == NA_LOGICAL)
     error(_("ignore.empty should be logical TRUE/FALSE."));
   const bool ignore = LOGICAL_RO(ignoreArg)[0];
   if (!(isNull(keepNamesArg) || (isString(keepNamesArg) && LENGTH(keepNamesArg) == 1)))


src/subset.c
```diff
@@ -88,12 +88,12 @@ void subsetVectorRaw(SEXP ans, SEXP source, SEXP idx, const bool anyNA)
     }
   } break;
   case CPLXSXP : {
-    Rcomplex *sp = COMPLEX(source);
+    const Rcomplex *sp = COMPLEX_RO(source);
```
ben-schwen, line 91: why are we leaving `INTEGER(source)`, `REAL(source)` here in the subset file?

src/inrange.c
```diff
@@ -4,8 +4,8 @@
 SEXP inrange(SEXP ansArg, SEXP xoArg, SEXP startsArg, SEXP lenArg)
 {
   int *ans = INTEGER(ansArg);
-  const int *xo = INTEGER(xoArg);
-  const int *starts = INTEGER(startsArg), *len = INTEGER(lenArg);
+  const int *xo = INTEGER_RO(xoArg);
+  const int *starts = INTEGER(startsArg), *len = INTEGER_RO(lenArg);
```
ben-schwen: why does `starts` not become read only?

src/transpose.c
```diff
@@ -11,7 +11,7 @@ SEXP transpose(SEXP l, SEXP fill, SEXP ignoreArg, SEXP keepNamesArg, SEXP listCo
     return(copyAsPlain(l));
   if (!isLogical(ignoreArg) || LOGICAL(ignoreArg)[0] == NA_LOGICAL)
     error(_("ignore.empty should be logical TRUE/FALSE."));
-  bool ignore = LOGICAL(ignoreArg)[0];
+  bool ignore = LOGICAL_RO(ignoreArg)[0];
```
ben-schwen: why aren't we making ignore const?

ben-schwen:
Overall LGTM, some minor tweaks needed. AFAIU R internals the read only variants will not be faster or only in subtle cases like `ALTREP`.

jangorecki:
If something is meant to be faster it is always good to include a benchmark, even the most optimistic one, of before vs after

PR #7370, opened 2025-10-16
fix sep detection for single column quoted file

### LINKED ISSUE #7366: fread doesn't manage to read without additional pointers
```r
require(data.table) # 1.17.8, loaded on a clean R session (v4.4.2)
# Macbook pro 14 inch (2023) running Sequoia 15.7

fread('"this_that"\n"2025-01-01 00:00:01"') 
# Empty data.table (0 rows and 2 cols): V1,"this_that"

# Warning messages:
# 1: In fread("\"this_that\"\n\"2025-01-01 00:00:01\"") :
#   Found and resolved improper quoting in first 100 rows. If the fields are not quoted (e.g. field separator does not appear within any field), try quote="" to avoid this warning.
# 2: In fread("\"this_that\"\n\"2025-01-01 00:00:01\"") :
#   Detected 1 column names but the data has 2 columns (i.e. invalid file). Added an extra default column name for the first column which is guessed to be row names or an index. Use setnames() afterwards if this guess is not correct, or fix the file write command that created the file to create a valid file.
# 3: In fread("\"this_that\"\n\"2025-01-01 00:00:01\"") :
#   Discarded single-line footer: <<"2025-01-01 00:00:01">>

fread('"this_that"\n"2025-01-01 00:00:01"', nrow=1L)
#              this_that
#                 <POSc>
# 1: 2025-01-01 00:00:01

fread('"this_that"\n"2025-01-01 00:00:01"', sep=",")
#              this_that
#                 <POSc>
# 1: 2025-01-01 00:00:01
```

The entire file has 1 column header and a row. Both are quoted. I feel like this should be read in correctly. There are no mismatches in quotes here. But fread fails unless additional help is provided via `nrow` or `sep` for example. I guess it's attempting to use `sep = ' '`, but don't understand how it returns an empty table even then...

Seems like a bug to me. It's possible this was introduced somewhere before 1.12.2 (from my logs, I seem to have created a patch to workaround this around that time but forgot to file an issue). I can still reproduce in latest CRAN.

ben-schwen:
Closes #7366

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -337,6 +337,7 @@ test-win-old:
 ## R-release on MacOS
 test-mac-rel:
   <<: *test-mac
+  image: macos-14-xcode-15
   variables:
     R_VERSION: "$R_REL_VERSION"
     R_BIN: "$R_REL_MAC_BIN"
@@ -344,6 +345,7 @@ test-mac-rel:
 ## R-oldrel on MacOS
 test-mac-old:
   <<: *test-mac
+  image: macos-14-xcode-15
   variables:
     R_VERSION: "$R_OLD_VERSION"
     R_BIN: "$R_OLD_MAC_BIN"
--- a/src/idatetime.c
+++ b/src/idatetime.c
@@ -145,7 +145,7 @@ SEXP convertDate(SEXP x, SEXP type)
         SEXP ans = PROTECT(allocVector(INTSXP, n));
         int *ansp = INTEGER(ans);
 
-        SEXP opt = GetOption(install("datatable.week"), R_NilValue);
+        SEXP opt = GetOption1(install("datatable.week"));
         const char *mode = isString(opt) && length(opt) == 1 ? CHAR(STRING_ELT(opt, 0)) : "default";
 
         bool use_sequential = !strcmp(mode, "sequential");


inst/tests/tests.Rraw
```diff
@@ -21685,3 +21685,7 @@ d3 = unserialize(serialize(d2, NULL))
 test(2340.05, .selfref.ok(d3), FALSE)
 setDT(d3)
 test(2340.06, .selfref.ok(d3), TRUE)
+
+# fread with quotes and single column #7366
+str = '"this_that"\n"2025-01-01 00:00:01"'
+test(2341, fread(str), data.table(this_that = as.POSIXct("2025-01-01 00:00:01", tz="UTC")))
```
tdhock: I would suggest avoiding intermediate variables that are used only once, and also avoiding a variable named "str" which is the same as base function str() -- potentially confusing.
```suggestion
test(2341, fread('"this_that"\n"2025-01-01 00:00:01"'), data.table(this_that = as.POSIXct("2025-01-01 00:00:01", tz="UTC")))
```

ben-schwen: Done

aitap (approved):
This feels like a relatively complicated special case, but I wasn't able to simplify it.

tdhock (approved):

PR #7375, opened 2025-10-17
implement comment.char argument for fread

### LINKED ISSUE #856: Implement comment.char argument in fread

ben-schwen:
Closes #856
Supersedes #4486

Needed this on machine without grep.
It should have no performance impact since we always check for `comment.char` before entering costly checking branch.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -21771,9 +21771,9 @@ test(2341.16, fread('a#b
 # na.strings
 local({
   txt = 'v
-  #NA
-  1
-  # comment'
+#NA
+1
+# comment'
   writeLines(txt, tmp <- tempfile())
   on.exit(unlink(tmp))
   test(2341.170, fread(tmp, na.strings="#NA", comment.char='#'), data.table(v=1L))


R/fread.R
```diff
@@ -30,6 +30,10 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     isTRUEorFALSE(stringsAsFactors) || (is.double(stringsAsFactors) && length(stringsAsFactors)==1L && 0.0<=stringsAsFactors && stringsAsFactors<=1.0),
     is.numeric(nrows), length(nrows)==1L
   )
+  if (is.null(comment.char)) comment.char = ""
```
MichaelChirico: any partcular reason to support this? I don't see `read.table()` doing so, e.g.

NEWS.md
```diff
@@ -291,6 +291,7 @@
     #   user  system elapsed
     #  0.028   0.000   0.005
     ```
+20. `fread()` has now `comment.char` argument implemented to skip trailing comments or comment-only lines similar as in `read.table`, [#856](https://github.com/Rdatatable/data.table/issues/856). Thanks to @arunsrinivasan and many others for the suggestion and @ben-schwen for the implementation.
```
MichaelChirico: `?read.table` has

> Using comment.char = "" will be appreciably faster than the read.table default.

If there's no performance impact, maybe that's worth highlighting in the NEWS here?

MichaelChirico: It's also worth highlighting that we have a different default than `read.table()`, for back-compatiblity I assume.

Should we plan to eventually change the default?

inst/tests/tests.Rraw
```diff
@@ -21685,3 +21685,10 @@ d3 = unserialize(serialize(d2, NULL))
 test(2340.05, .selfref.ok(d3), FALSE)
 setDT(d3)
 test(2340.06, .selfref.ok(d3), TRUE)
+
+# implement comment.char argument in fread, #856
+test(2341.1, fread('a,b\n#a comment\n1,2\n#another comment\n3,4', comment.char='#'), data.table(a=c(1L,3L), b=c(2L,4L)))
```
MichaelChirico: I would expect a fair number more tests :)

Let's e.g. codify invalid values of `comment.char`: `comment.char %in% list(1L, factor('a'), letters, NA_character_, "##")`

MichaelChirico: I also find it a bit hard to parse the tests visually, it takes a lot of concentration to figure out what we might expect by parsing the string and newlines. Any ideas? Perhaps use literal newlines instead of `\n`?

```suggestion
test(2341.1, fread('a,b
#a comment
1,2
#another comment
3,4', comment.char='#'), data.table(a=c(1L,3L), b=c(2L,4L)))
```

pretty ugly but maybe there's something there.

inst/tests/tests.Rraw
```diff
@@ -21685,3 +21685,10 @@ d3 = unserialize(serialize(d2, NULL))
...
+test(2341.2, fread('a,b #line-trailing comment\n1,2', comment.char='#'), data.table(a=c(1L), b=c(2L)))
```
MichaelChirico: What's the rule, exactly? All spaces preceding the comment character are also dropped? Why isn't the column named `b `? Is this an interaction with `strip.white` that should also be tested?

MichaelChirico: (PS my own style is to prefer `1L` to `c(1L)`, we might lint that eventually: https://lintr.r-lib.org/reference/unnecessary_concatenation_linter.html)

inst/tests/tests.Rraw
```diff
@@ -21685,3 +21685,10 @@ d3 = unserialize(serialize(d2, NULL))
...
+test(2341.3, fread('a,b#line-trailing comment and no whitespace\n1,2', comment.char='#'), data.table(a=c(1L), b=c(2L)))
+test(2341.4, fread('a,b\n1,2 #trailing after numeric', comment.char='#'), data.table(a=c(1L), b=c(2L)))
+test(2341.5, fread('a\n"#quotes#"\n', comment.char="#"), data.table(a=c("#quotes#")))
```
MichaelChirico: more suggested tests:

 1. comments at the front / end of the file
 2. consecutive lines of comments
 3. more interactions of `comment.char` with other arguments, e.g. `sep`, `quote`, `na.strings`, `dec`, maybe `header` and `skip`
 4. any other comment character than `#`
 5. weird comment character like ` ` or `"`

src/fread.c
```diff
@@ -336,8 +338,24 @@ static inline int countfields(const char **pch)
...
   targets[8] = (void*) &trash;
   const char *ch = *pch;
-  if (sep == ' ') while (*ch == ' ') ch++;  // multiple sep==' ' at the start does not mean sep
-  skip_white(&ch);
+  for (;;) {
+    if (ch >= eof) { *pch = ch; return 0; }
+    if (sep == ' ') while (*ch == ' ') ch++;  // multiple sep==' ' at the start does not mean sep
+    skip_white(&ch);
+    if (commentChar && *ch == commentChar) {
+      while (ch < eof && *ch != '\n' && *ch != '\r') ch++;
```
MichaelChirico: should this be a helper since it's reused?

src/fread.c
```diff
@@ -2206,12 +2236,31 @@ int freadMain(freadMainArgs _args)
         ch++;
         Field(&fctx);  // stores the string length and offset as <uint,uint> in colNames[i]
         ((lenOff**) fctx.targets)[8]++;
+        if (commentChar) {
+          // skip leading whitespace to detect inline comment marker in header row
+          const char *commentPos = ch;
+          while (commentPos < eof && (*commentPos == ' ' || *commentPos == '\t' || *commentPos == '\0')) commentPos++;
```
MichaelChirico: similarly, this looks like it should be abstracted into a helper to re-use shared logic for easier maintenance

src/freadR.c
```diff
@@ -158,6 +159,13 @@ SEXP freadR(
   // here we use bool and rely on fread at R level to check these do not contain NA_LOGICAL
   args.stripWhite = LOGICAL(stripWhiteArg)[0];
   args.skipEmptyLines = LOGICAL(skipEmptyLinesArg)[0];
+  if (!isString(commentCharArg) || LENGTH(commentCharArg) != 1)
+    error(_("comment.char must be a single character vector of length 1"));  // # notranslate
```
MichaelChirico: NB: `_()` is the translation macro, I guess maybe you meant `# nocov` since this is tested on the R side?

(personally, I don't see much point in a check like this that can never be reached)

ben-schwen: Fine for me, I've removed the unreachable code here.

R/fread.R
ben-schwen: Good spot. This was a legacy issue since I started with `comment.char=NULL` before looking at `read.table`

NEWS.md
ben-schwen: Not sure we will ever change this default and in my opinion the default of `read.table()` should also be `comment.char=""` since this seems the more common case.

ben-schwen: Updated the NEWS accordingly.

inst/tests/tests.Rraw
ben-schwen: added the visual style although they are still hard to grasp then :/

inst/tests/tests.Rraw
ben-schwen: Yes we also drop whitespace before the comment char since I belive people like to put in the extra space for readability

src/fread.c
ben-schwen: we now have the `skip_to_comment_or_nonwhite` helper.

src/fread.c
ben-schwen: It's now the `skip_line` helper.

inst/tests/tests.Rraw
ben-schwen: Added a bunch of tests, but I'm sure ppl will come up with cases I couldn't think of. Using whitespace as `comment.char` is one of these :D

MichaelChirico: we don't necessarily need to support weird cases unless user asks, a test is mainly making sure it doesn't crash / possibly adding an error to prevent it if it behaves unexpectedly

inst/tests/tests.Rraw
MichaelChirico: I would let strip.white govern this, possibly the space is desired and it should be possible to get it even if it's not the default

ben-schwen: Thanks to the `skip_to_comment` helper this was only a small change.

inst/tests/tests.Rraw
```diff
@@ -21688,3 +21688,116 @@ d3 = unserialize(serialize(d2, NULL))
...
+
+# nrow not counting comments
+test(2341.15, fread('a,b
+1,2
+# cmt
+3,4
+5,6', nrows=2, comment.char='#'), data.table(a=c(1L,3L), b=c(2L,4L)))
+
+# sep and comment char same
+test(2341.16, fread('a#b
```
MichaelChirico, line 21767: hm that's interesting. I don't know what I'd expect out of reading this file. E.g. `data.table(a=1L, b=2L)` also seems like a possible answer, and even `data.table(a=c(1L, NA), b=c("2", "only comment"))` could be justified too. Maybe let's just error if `comment.char==sep`?

inst/tests/tests.Rraw
```diff
@@ -21688,3 +21688,116 @@ d3 = unserialize(serialize(d2, NULL))
...
+# CLRF
+test(2341.13, fread('a,b\r\n# cmt\r\n1,2\r\n3,4\r\n', comment.char='#'), data.table(a=c(1L,3L), b=c(2L,4L)))
+
+# header comment
+test(2341.14, fread('# hdr cmt
+x,y
+1,2', header=TRUE, comment.char='#'), data.table(x=1L, y=2L))
+
+# nrow not counting comments
+test(2341.15, fread('a,b
```
MichaelChirico, line 21760: great test, glad this works as expected!

inst/tests/tests.Rraw
```diff
@@ -21688,3 +21688,116 @@ d3 = unserialize(serialize(d2, NULL))
...
+# sep and comment char same
+test(2341.16, fread('a#b
+1#2
+# only comment', sep="#", comment.char="#"), data.table(a=1L))
+
+# na.strings
+test(2341.17, fread('v
+#NA
+1
+# comment', na.strings="#NA", comment.char='#'), data.table(v=1L))
```
MichaelChirico: Another one I'm not sure what good behavior is, does this match `read.table()`?

inst/tests/tests.Rraw
```diff
@@ -21688,3 +21688,116 @@ d3 = unserialize(serialize(d2, NULL))
...
+test(2341.20, fread('a,b      # header cmt with padding # second cmt
+1,2
+3,4', comment.char='#'), data.table(a=c(1L,3L), b=c(2L,4L)))
+
+test(2341.21, fread('# meta1 # meta2
+a,b
+1,2', comment.char = '#'), data.table(a=1L, b=2L))
+test(2341.22, fread('a,b   # inline header comment\r\n1,2\r\n', comment.char = '#'), data.table(a=1L, b=2L))
+
+# control skipping white space before comments with strip.white
```
MichaelChirico, line 21801: let's add one more here where the whitespace is in the header

man/fread.Rd
```diff
@@ -56,6 +56,7 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC"
   \item{strip.white}{ Logical, default \code{TRUE}, in which case leading and trailing whitespace is stripped from unquoted \code{"character"} fields. \code{"numeric"} fields are always stripped of leading and trailing whitespace.}
   \item{fill}{logical or integer (default is \code{FALSE}). If \code{TRUE} then in case the rows have unequal length, number of columns is estimated and blank fields are implicitly filled. If an integer is provided it is used as an upper bound for the number of columns. If \code{fill=Inf} then the whole file is read for detecting the number of columns. }
   \item{blank.lines.skip}{\code{logical}, default is \code{FALSE}. If \code{TRUE} blank lines in the input are ignored.}
+  \item{comment.char}{Character vector of length one containing a single character of an empty string. Any text after the comment character in a line is ignored. Use \code{""} to turn off the interpretation of comments altogether.}
```
MichaelChirico: Is there any interaction with `blank.lines.skip`? I'm trying to think how to describe the behavior of full-line comments. Is it "Any line starting with a comment is skipped"? Or "Any line with a comment preceded only by whitespace is skipped"? "Any line without `sep` but with a comment is skipped"?

src/fread.c
```diff
@@ -304,7 +347,8 @@ static inline bool end_of_field(const char *ch)
   // default, and therefore characters in the range 0x80-0xFF are negative.
   // We use eol() because that looks at eol_one_r inside it w.r.t. \r
   // \0 (maybe more than one) before eof are part of field and do not end it; eol() returns false for \0 but the ch==eof will return true for the \0 at eof.
-  return *ch == sep || ((uint8_t)*ch <= 13 && (ch == eof || eol(&ch)));
+  // Comment characters terminate a field immediately and take precedence over separators.
+  return *ch == sep || ((uint8_t)*ch <= 13 && (ch == eof || eol(&ch))) || (commentChar && *ch == commentChar);
```
MichaelChirico, line 351: Maybe time to split it out more, WDYT? (I don't think there is any performance implication)

```suggestion
  if (*ch == sep) return true;
  if (((uint8_t)*ch <= 13 && (ch == eof || eol(&ch))) return true;
  if (!commentChar) return false;
  return *ch == commentChar;
```

MichaelChirico:
Looks basically GTG. Main remaining thing I think is what to do about `blank.lines.skip`.

@eddelbuettel if you have any "real" files laying around you'd like to test on that would be most helpful :)

eddelbuettel:
LOL @MichaelChirico -- while I wrote _eight years ago_ 

> Bump. Needing this right now.

I am sorry to say that I have not been sitting here with baited breath and no, I do not have _that file_ around.  I do not even remember what the comment character may have been.   I am sure the PR will be fine.

inst/tests/tests.Rraw
ben-schwen: Done.

inst/tests/tests.Rraw
ben-schwen: Done.

src/fread.c
ben-schwen: Done. (There is an opening branch too much here.)

man/fread.Rd
ben-schwen: Well it depends on `strip.white`.
For `strip.white=FALSE` we have "Any line starting with a comment is skipped"
For `strip.white=TRUE` we have "Any line with a comment preceded only by whitespace is skipped"

This somehow diverges from `blank.lines.skip` since we skip no matter if `blank.lines.skip` or not.

MichaelChirico: I see, that makes it a bit complicated.

We can imagine `   # comment` not skipping if `blank.lines.skip=FALSE`, but OTOH I don't know why you'd want to do that (I'm not sure why `blank.lines.skip=FALSE` by default, even).

So, let's leave it as is for now and see if there's any user request later.

```suggestion
  \item{comment.char}{Character vector of length one containing a single character of an empty string. Any text after the comment character in a line is ignored, including skipping comment-only lines. Use \code{""} to turn off the interpretation of comments altogether.}
```

MichaelChirico:
@Anirban166 it looks like some {pak} cacheing issue failing the atime GHA, any idea?

https://github.com/Rdatatable/data.table/actions/runs/18658671941/job/53193860953

MichaelChirico (approved):

Anirban166:
Haven't seen that one before, but I think its a flaky error having to do with network/resource contention as it worked now that you re-runned (the `r-lib/actions/setup-r-dependencies@v2` action that I'm using internally uses `pak`, so it would usually be an upstream error if one, but in cases of corrupted caches or missing files, we can use a different number for `cache-version`)

MichaelChirico:
Thanks I'll try and notice if it recurs, indeed seems flaky. There were other failures related to missing branches, which clouded things. Seems restored to working now.

PR #7372, opened 2025-10-17
benchmarks for openmp parallel for skip flag

jangorecki:
Towards #7371

this code assumes you have 8+ threads
```r
cc()
do = function(what) {
  n = 1e8
  rbindlist(lapply(c("1e8+1"=1e8+1, "1e7"=1e7, "10"=10), function(half) {
    rbindlist(lapply(c("nothing"=1L,"volatile"=2L,"volatile+shared"=3L,"atomic write"=4L,"atomic read write"=5L, "reduction"=6L, "cancellation"=7L), function(variant) {
      a2<-system.time(
        a1<-omp_flags(variant, n, half, 8)
      )
      if (what == "time") as.list(a2)
      else if (what == "iters") as.list(a1)
    }), idcol="variant")
  }), idcol="halt")
}
do("iters")
do("time")
```
```
      halt           variant       V1       V2       V3       V4       V5       V6       V7       V8
    <char>            <char>    <int>    <int>    <int>    <int>    <int>    <int>    <int>    <int>
 1:  1e8+1           nothing 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 2:  1e8+1          volatile 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 3:  1e8+1   volatile+shared 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 4:  1e8+1      atomic write 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 5:  1e8+1 atomic read write 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 6:  1e8+1         reduction 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 7:  1e8+1      cancellation 12500000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 8:    1e7           nothing 10000000 12500000 12500000 12500000 12500000 12500000 12500000 12500000
 9:    1e7          volatile 10000000  9368188  2219459  2374732  2843583  2794889  2256903  2389781
10:    1e7   volatile+shared 10000000  9561233  2291372  2512503  3533221  3080863  2444200  2472036
11:    1e7      atomic write 10000000  9732664  2249224  2339365  2973864  2526774  2218488  2205693
12:    1e7 atomic read write 10000000 10005487  1710672  1566658  1930782  1774714  1555453  1497075
[...]

Issue #7371, opened 2025-10-16
data races in openmp on shared boolean flags


ben-schwen:
The [openmp documentation](https://www.openmp.org/spec-html/5.2/openmpsu9.html#x18-170001.4.1) states our common used idiom e.g. setting shared boolean flag is UB in openmp.

> Two memory operations are considered unordered if the order in which they must complete, as seen by their aﬀected threads, is not speciﬁed by the memory consistency guarantees listed in [Section 1.4.6](https://www.openmp.org/spec-html/5.2/openmpsu14.html#x23-220001.4.6). If multiple threads write to the same memory unit (deﬁned consistently with the above access considerations) then a data race occurs if the writes are unordered. Similarly, if at least one thread reads from a memory unit and at least one thread writes to that same memory unit then a data race occurs if the read and write are unordered. If a data race occurs then the result of the program is unspeciﬁed.

``` c
#pragma omp parallel for
for (int i = 0; i < n; i++) {
    shared_flag = true;   // UB: concurrent unsynchronized writes
}
```

_Originally posted by @ben-schwen in https://github.com/Rdatatable/data.table/pull/7361#discussion_r2435903089_

jangorecki:
We never update flag to false, only to true, so it is one way update.
Isn't it like this, that in the worst case some of threads will simply unnecessarily do extra iterations? (if read was happening during write of another thread, till flag refreshed from memory)
That's how I would imagine data races in this context.
In such case we don't mind.

In the past this variable was declared `volatile` and then for omp also wrapped into `shared` clause: https://github.com/Rdatatable/data.table/commit/055047b521f43223d28056be03457c38e98eba2f

as of now it seems that atomic write is the best way to address it

jangorecki:
I will work on a minimal example that will examine situation and impact of different approaches.

PR #7376, opened 2025-10-18
proper skip omp parallel loops in froll

jangorecki:
Towards #7371

I am not closing linked issue because there may be other places in code where this is still relevant.

Implementation decided based on benchmarks in #7372


jangorecki:
merging for now, aligned to same changes added in #7361 
we can revisit later on based on #7372

PR #7378, opened 2025-10-19
Set digits.secs option before tests

### LINKED ISSUE #7377: Set digits.sec before tests
Currently tests `2150.022`, `2150.023`, and `2150.024` will fail when the user has set
the `digits.sec` options himself, e.g. `options(digits.secs = 3)`.

ben-schwen:
Closes #7377

No news and additional tests since test only change.


MichaelChirico:
Does it make sense to set it for all tests instead? e.g. we already do so for `digits=`.

https://github.com/Rdatatable/data.table/blob/67129f034941b1d8ea3ed684822e04f517c6c837/R/test.data.table.R#L92-L111

Issue #563, opened 2014-06-08
[R-Forge #4903] set encoding in fread


arunsrinivasan:
Submitted by: michele de meo; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=4903&group_id=240&atid=978)

It's useful to specify the encoding before importing   a csv.
For example: 

fread(... , encoding='UTF8').

In this way we can avoid the boring use of file function in read.table (not supported in fread ):

read.table(file("mycsv.csm", encoding='UTF8'), ... ).

Michele De Meo

sirvyd:
+1

harmonica2:
+1

zeltak:
+1

MarcinKosinski:
+1

jomuller:
+1 Would be really usefull. I started to hack locally the code but it's too hard for me because it call a C function (readfile) and I don't know how (for the moment) how to handle encoding in C. If any one some knowledge in C...

grgurev:
+1

lucarno:
+1

seyedamo:
+1

jcizel:
+1

Alectoria:
+1

tophcito:
+1

clarkdk:
+1

JohnsonHsieh:
+1

whizzalan:
+1

ysgit:
+1

ZeroStack:
+1

rshmyrev:
+1

dbuijs:
+1, and in the meantime, a workaround:

You need iconv available from the console. You can check this with the following command:

    Sys.which("iconv")

As long as this gives you a path to a binary, the following will work:

    # From ISO-8859-1, To UTF-8
    new.data.table <- fread("iconv -f ISO-8859-1 -t UTF-8 mytextfile.txt")

shrektan:
+ 1

panda2727:
+1

duf59:
+1

romunov:
+1

note however that in `?Encoding`, `UTF-8` is used (not `UTF8`).

rentrop:
+1

leoluyi:
+1

arunsrinivasan:
Tl;dr 

[...]

PR #7379, opened 2025-10-19
unique and duplicated warn on other encodings than UTF8

### LINKED ISSUE #469: [R-Forge #5758] unique and duplicated should also warn on columns with mixed encodings.
Submitted by: Arun ; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=5758&group_id=240&atid=975)

Sigh! It just seems not possible to put this (mixed) encoding issue to rest. As reported [here on SO](http://stackoverflow.com/q/24085906/559784).

ben-schwen:
Closes #469 

Not exactly what Arun suggested but seems like the best option since we encode to UTF8 in `forderv`. Is a warning too much here?


jangorecki:
Let's see revdeps. If none affected then I would keep it like this.

aitap:
This may bite fread() users who rely on the default value of the
encoding= argument:

``` r
fwrite(list(x = c('a møøse', 'bit my sister once')), 'foo.txt')
fread('foo.txt', sep='\t') |> setkey(x) |> unique() |> _$x |> str()
#  chr [1:2] "a møøse" "bit my sister once"
# Warning message:
# In unique.data.table(setkey(fread("foo.txt", sep = "\t"), x)) :
#   Mixed encodings detected. Strings were coerced to UTF-8 before
#   unique(x).
```

A similar problem may happen to users of old versions of R or old
versions of Windows where non-ASCII string literals would result in
CE_NATIVE strings instead of CE_UTF8.

Should the warning recommend enc2utf8()?

Could someone please share the original context from R-Forge #5758?

MichaelChirico:
> Could someone please share the original context from R-Forge `#5758`?

Basically, it links to https://stackoverflow.com/questions/24085906/unique-data-table-do-not-handle-keys-properly

jangorecki:
Late warning about unexpected consequences is better than no warning, so still despite that I see value in it.
And what about changing the default of fread as well?

aitap:
Is lower performance meant to be the unexpected consequence? Otherwise
we seem to do the same thing as what R does in identical().

Given how most systems already speak UTF-8 as the native encoding,
changing the default fread(encoding=) argument to "UTF-8" should be
mostly harmless. Only people who read invalid UTF-8 or use a non-UTF-8
system will notice a difference.

ben-schwen:
Good points! I've added `fread(encoding="UTF-8")` as default encoding. Given how much support the addition of `encoding='UTF-8'` got in #563 this sounds also like what the community wants/needs.

I have also filed this now as breaking change.

aitap:
The warning is not caused by mixed encodings. The previous example
sorts a vector of CE_NATIVE strings, which is actually safe to perform
without conversion to UTF-8, but the warning still appears. It would be
more precise to say "Detected strings not marked as ASCII or UTF-8 and
converted them to UTF-8 for detection of duplicates. Please convert
your character columns using enc2utf8() to avoid the on-the-fly
conversion."

Could you please explain why is the result unexpected when calling
duplicated() or unique() without converting strings to UTF-8 first?
After all, identical(enc2utf8('ø'), iconv('ø', to='latin1')) is TRUE,
even if those are different CHARSXPs under the hood.

ben-schwen:
I guess what Jan meant by unexptected consequences is that `forderv` can simply take longer because user are unaware of encodings.

jangorecki:
Simply longer is not a problem, I thought that something that was giving T for == can now start to return F

PR #7381, opened 2025-10-21
Clarify $ vs [, col] behavior in docs (#7369)

### LINKED ISSUE #7369: Comparison of DT$region v DT[,region]
I ***think*** this is just a documentation request. Would it be worth clarifying the differences between accessing a vector with `$` versus the alternative DT NSE way `DT[,region]`?

Motivated by the following line which encourages the use of `$` at times:

https://github.com/Rdatatable/data.table/blob/55b0de600d1770babe63d94ad3766a36bdedc443/vignettes/datatable-faq.Rmd#L59

The problem is there is a difference to whether the result is passed by reference, or evaluated within DT which can catch you out:

``` r
library(data.table)
DT <- data.table(a = 1)

x <- DT$a
DT[1,a:=2]
x # has changed, fair enough we need to throw a copy in
#> [1] 2
z <- copy(DT$a)
DT[1,a:=3]
z
#> [1] 2
x
#> [1] 3

# but ...
y <- DT[, a]
DT[1,a:=4]
y # y hasn't changed as j is evaluated in DT and the answer returned so no need for copy
#> [1] 3
x
#> [1] 4
```

I'm sure this has been discussed before but I couldn't easily find a particular mention in the documentation. Perhaps another example for `?copy` ?

venom1204:
closes #7369 

This PR clarifies the difference between DT$col and DT[, col] when selecting columns.

- datatable-reference-semantics.Rmd:
Added a short section explaining that DT$col may return a reference, while DT[, col] always returns a copy. Included a minimal example.

- datatable-faq.Rmd:
Added a brief note and link to the semantics vignette for details.

hi @jangorecki can you please review this when you have time ,and if any improvements are needed let me know .
thank you for your time.
thanks .

### DIFF
--- a/vignettes/datatable-reference-semantics.Rmd
+++ b/vignettes/datatable-reference-semantics.Rmd
@@ -365,7 +365,7 @@ However we could improve this functionality further by *shallow* copying instead
     ```
 ### c) Selecting columns: `$` / `[[...]]` vs `[, col]`
 
-When you extract a single column as a vector, there is a subtle but important difference between standard R methods ($ and [[...]]) and data.table's j expression. DT$col and DT[['col']] may return a reference to the internal column, while DT[, col] always returns a copy.
+When you extract a single column as a vector, there is a subtle but important difference between standard R methods ($ and [[...]]) and data.table's j expression. DT$col and DT[['col']] may return a reference to the column, while DT[, col] always returns a copy.
 
 A short example:
 ```{r}


vignettes/datatable-reference-semantics.Rmd
```diff
@@ -363,6 +363,37 @@ However we could improve this functionality further by *shallow* copying instead
     ## DT_n doesn't get updated
     DT_n
     ```
+### c) Selecting columns: `$` vs `[, col]`
+
+When selecting a column as a vector, `DT$col` and `DT[, col]` have a key difference: `DT$col` may return a reference to the data, while `DT[, col]` always returns a copy.
+
+This means a variable created with `$` can change if the `data.table` is modified later, which can be unexpected. A single example demonstrates this behavior and how `copy()` provides a solution:
+
```
ben-schwen:
```suggestion
When you extract a single column as a vector there is a subtle but important difference: `DT$col` may return a reference to the internal column, while `DT[, col]` always returns a copy. Because of that, a variable created with `DT$col` can change if you later modify DT by reference.

A short example:
```

vignettes/datatable-reference-semantics.Rmd
```diff
@@ -363,6 +363,37 @@ However we could improve this functionality further by *shallow* copying instead
...
+# Now, modify the original data.table by reference
+DT[, a := a + 10L]
+
+# Check the variables:
+x_ref 
+#> [1] 11 12 13
+y_cpy
+#> [1] 1 2 3
+z_cpy 
+#> [1] 1 2 3
```
ben-schwen:
```suggestion
# three ways to get the column
x_ref = DT$a        # may be a reference
y_cpy = DT[, a]     # always a copy
z_cpy = copy(DT$a)  # forced copy

# modify DT by reference
DT[, a := a + 10L]

# observe results
x_ref   # may show 11 12 13
y_cpy   # 1 2 3
z_cpy   # 1 2 3
```

vignettes/datatable-reference-semantics.Rmd
```diff
@@ -363,6 +363,37 @@ However we could improve this functionality further by *shallow* copying instead
...
+```
+
+To select a single column as a vector, remember:
+- `DT[, mycol]` is safer as it always returns a new, independent copy.
+- `DT$mycol` is fast but may return a reference. Use `copy(DT$mycol)` to guarantee independence.
```
ben-schwen, line 390:
```suggestion
- `DT$mycol` is fast but may return a reference. 
- Use `copy(DT$mycol)` to guarantee independence.
```

vignettes/datatable-faq.Rmd
```diff
@@ -56,7 +56,7 @@ For consistency so that when you use data.table in functions that accept varying
...
 
-Say column 5 is named `"region"` and you really must extract that column as a vector not a data.table. It is more robust to use the column name and write `DT$region` or `DT[["region"]]`; i.e., the same as base R. Using base R's `$` and `[[` on data.table is encouraged. Not when combined with `<-` to assign (use `:=` instead for that) but just to select a single column by name they are encouraged.
+Say column 5 is named `"region"` and you really must extract that column as a vector not a data.table. It is more robust to use the column name and write `DT$region` or `DT[["region"]]`; i.e., the same as base R. Using base R's `$` and `[[` on data.table is encouraged. Not when combined with `<-` to assign (use `:=` instead for that) but just to select a single column by name they are encouraged.A key difference, however, is that DT$col may return a reference, while DT[, col] always returns a copy. This can have important consequences and is explained in the vignette("datatable-reference-semantics", package="data.table").
```
jangorecki: One space needed between the dot and the A letter

ben-schwen:
```suggestion
Say column 5 is named `"region"` and you really must extract that column as a vector not a data.table. It is more robust to use the column name and write `DT$region` or `DT[["region"]]`; i.e., the same as base R. Using base R's `$` and `[[` on data.table is encouraged. Not when combined with `<-` to assign (use `:=` instead for that) but just to select a single column by name they are encouraged. A key difference, however, is that DT$col may return a reference, while DT[, col] always returns a copy. This can have important consequences and is explained in the `vignette("datatable-reference-semantics", package="data.table")`.
```

vignettes/datatable-reference-semantics.Rmd
```diff
@@ -363,6 +363,37 @@ However we could improve this functionality further by *shallow* copying instead
...
     DT_n
     ```
+### c) Selecting columns: `$` vs `[, col]`
+
+When selecting a column as a vector, `DT$col` and `DT[, col]` have a key difference: `DT$col` may return a reference to the data, while `DT[, col]` always returns a copy.
+
+This means a variable created with `$` can change if the `data.table` is modified later, which can be unexpected. A single example demonstrates this behavior and how `copy()` provides a solution:
+
+```{r}
+library(data.table)
```
jangorecki: No need

jangorecki:
What about DT[["a"]]

vignettes/datatable-reference-semantics.Rmd
```diff
@@ -363,6 +363,31 @@ However we could improve this functionality further by *shallow* copying instead
     ## DT_n doesn't get updated
     DT_n
     ```
+### c) Selecting columns: `$` / `[[...]]` vs `[, col]`
+
+When you extract a single column as a vector, there is a subtle but important difference between standard R methods ($ and [[...]]) and data.table's j expression. DT$col and DT[['col']] may return a reference to the internal column, while DT[, col] always returns a copy.
```
ben-schwen:
```suggestion
When you extract a single column as a vector, there is a subtle but important difference between standard R methods ($ and [[...]]) and data.table's j expression. DT$col and DT[['col']] may return a reference to the column, while DT[, col] always returns a copy.
```

ben-schwen:
TY!

PR #7382, opened 2025-10-23
ensure no nested parallelism in froll

### LINKED ISSUE #7352: ensure no nested parallelism in rolling functions
Adaptive min/max/prod algo=fast + vectorized input is most likely ending up in nested parallelism.
It is because those three functions redirects to their algo=exact versions, which are already parallelized.
This must be avoided.
Possible solutions could be:
1. redirect early, before parallel loop on vectorized input in frollR.c - keeps inner parallelism
2. disable nested parallelism using flag (as done for roll median) passed to frolladaptive.c - keeps outer parallelism

jangorecki:
closes #7352


src/froll.c
```diff
@@ -77,14 +77,14 @@ void frollfun(rollfun_t rfun, unsigned int algo, const double *x, uint64_t nx, a
     break;
   case VAR :
     if (algo==0) {
-      frollvarFast(x, nx, ans, k, fill, narm, hasnf, verbose);
+      frollvarFast(x, nx, ans, k, fill, narm, hasnf, verbose, par); // par is used only when NAs - fallback to exact, to know if outer parallelism has been applied
     } else if (algo==1) {
-      frollvarExact(x, nx, ans, k, fill, narm, hasnf, verbose);
+      frollvarExact(x, nx, ans, k, fill, narm, hasnf, verbose, /*par=*/ true); // par=true because frollvarExact at this place was invoked directly, and not by fallback, so algo=exact have been used explicitly, then outer parallelism in frollR.c is disabled already
```
jangorecki: It may actually be better to pass par rather than hardcode value

PR #7384, opened 2025-10-23
fread failed to detect correct line endings with mixed \r and \r\n

### LINKED ISSUE #4186: fread messed up by MS Excel mac .csv
This bug is really a bug in Microsoft Excel, but manifests as a "bug" in fread.
This bug is related to Issues #2542, #2248, #1183

`#` `Summary of Issue`
When saving a CSV as a mac-type in MS Excel Office 2016 from a Windows10 PC, Excel creates a file that has \r after every line EXCEPT the final line, which ends as \r\n. fread then interprets each "line" as a separate column.

You may be able to skip many of these file creation steps...taking a well-formed csv or xls file and saving it as mac-type csv is the most critical step.

Create the following in a notepad application and save as input.csv
```
Col1,Col2
1,data1
2,data2
3,data3
``` 
Open input.csv in Excel (2016 used...but likely applies to other versions)
File->Save As->Save as type: Excel Workbook (*.xlsx)->intermed.xlsx 
Close Excel
 
Open intermed.xlsx in Excel
File->Save As->Save as type: CSV (Macintosh) (*.csv)->final.csv
Close Excel

In R...

`fread("final.csv")`
> 3,data3ata.table (0 rows and 5 cols): Col1,Col2

`fredT<-fread("final.csv",header=T)`
`dim(fredT)`
> [1] 0 5

`fredF <-fread("final.csv",header=F)`
`dim(fredF)`
> [1] 1 5

`fredF`
>     V1      V2       V3       V4    V5
>   1: Col1 Col2\r1 data1\r2 data2\r3 data3

`#` `Output of sessionInfo()`
`sessioninfo::platform_info()`

>  setting  value                       
>  version  R version 3.5.0 (2018-04-23)
>  os       Windows 8.1 x64             
>  system   x86_64, mingw32             
>  ui       RStudio                     
>  language (EN)                        
>  collate  English_United States.1252  
>  ctype    English_United States.1252  
>  tz       America/Los_Angeles         
>  date     2020-01-17            

`packageVersion("data.table")`
> [1] ‘1.12.2’

ben-schwen:
Closes #4186

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -11953,7 +11953,7 @@ tt[2, B:=gsub("\n","\r",B)]  # base R changes the \r to a \n, so restore that
 test(1778.4, tt, DT)
 unlink(f)
 # fread has problems with mixed \r and \r\n #4186
-test(1778.5, fread(text="Col1;Col2\r1;data1\r2;data2\r3;data3\r\n", verbose=TRUE), data.table(Col1=1:3, Col2=c("data1","data2","data3")), output="Single \\r")
+test(1778.5, fread(text="Col1;Col2\r1;data1\r2;data2\r3;data3\r\n", verbose=TRUE), data.table(Col1=1:3, Col2=c("data1","data2","data3")), output="An \\r")
 
 # #1392 IDate ITime new methods for faster conversion
 # conversion in-out match for UTC


src/fread.c
```diff
@@ -1628,12 +1628,35 @@ int freadMain(freadMainArgs _args)
...
+      ch++;
+    }
+  }
+  // If file has mostly \r-only line endings, treat \r as line ending
+  eol_one_r = (count_r_only > count_with_n);
   if (verbose) DTPRINT(eol_one_r ?
-    _("  No \\n exists in the file at all, so single \\r (if any) will be taken as one line ending. This is unusual but will happen normally when there is no \\r either; e.g. a single line missing its end of line.\n") :
-    _("  \\n has been found in the input and different lines can end with different line endings (e.g. mixed \\n and \\r\\n in one file). This is common and ideal.\n"));
+    _("  Single \\r (if any) will be taken as one line ending (count: %d \\r vs %d \\n). This happens with old Mac CSV or when there is no \\r either.\n") :
+    _("  \\n has been found in the input (count: %d \\r vs %d \\n) and different lines can end with different line endings (e.g. mixed \\n and \\r\\n in one file). This is common and ideal.\n"),
```
aitap:
```suggestion
    _("  An \\r by itself will be taken as one line ending (counts: %d \\r by themselves vs %d [\\r]*\\n). This happens with old Mac CSV or when there is no \\r at all.\n") :
    _("  \\n has been found in the input (counts: %d \\r by themselves vs %d [\\r]*\\n) and different lines can end with different line endings (e.g. mixed \\n and \\r\\n in one file). This is common and ideal.\n"),
```
I think that the messages could be more precise, but there's a fine line between being precise and showering the user in unnecessary details. What do you think?

aitap (approved):
The code is fine, but the messages are a bit hard to parse.

src/fread.c
ben-schwen: True. This seems to be on of the older messages and I didn't want to touch them.

PR #7387, opened 2025-10-25
run linters only when respective files are changed

ben-schwen:
Follow up to #7280

Also added the fallback that manual triggers also start them


MichaelChirico:
Very cool 😎

PR #7388, opened 2025-10-25
add requires_utf8 argument to tests

### LINKED ISSUE #1343: Finish DtNonAsciiTests package and submit to CRAN

### LINKED ISSUE #7333: Escape UTF-8 dependent tests
Currently out vanilla environment job, that does not have utf8 encoding:
```
* checking code files for non-ASCII characters ... OK
* checking R files for syntax errors ... WARNING
Warning in Sys.setlocale("LC_CTYPE", "en_US.UTF-8") :
  OS reports request to set locale to "en_US.UTF-8" cannot be honored
```
Is failing on the following tests:
```

  Test 2194.7 produced 0 errors but expected 1
  Expected: Internal error.*types or lengths incorrect
  Observed: 
  Tests 2253* skipped because they need a UTF-8 locale.
  Tests 2266* skipped. n-tilde cannot be represented in the native encoding on this system. Please report to the data.table issue tracker.
  Test 2298.1 ran without errors but failed check that x equals y:
  > x = rbindlist(list(x, y), use.names = TRUE) 
     <U+00E4> <U+00F6> <U+00FC>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:        1        2        3                                              
  2:        6        4        5                                              
  > y = data.table(`<U+00E4>` = c(1, 6), `<U+00F6>` = c(2, 4), `<U+00FC>` = c(3,      5)) 
     <U+00E4> <U+00F6> <U+00FC>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:        1        2        3                                              
  2:        6        4        5                                              
  Different column names
  Test 2298.2 ran without errors but failed check that x equals y:
  > x = rbindlist(list(y, x), use.names = TRUE) 
     <f6> <fc> <e4>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:    4    5    6                                              
  2:    2    3    1                                              
  > y = data.table(`<U+00F6>` = c(4, 2), `<U+00FC>` = c(5, 3), `<U+00E4>` = c(6,      1)) 
     <U+00F6> <U+00FC> <U+00E4>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:        4        5        6                                              
  2:        2        3        1                                              
  Different column names
  Test 2298.3 ran without errors but failed check that x equals y:
  > x = rbindlist(list(x, y), use.names = TRUE, fill = TRUE) 
     <U+00E4> <U+00F6> <U+00FC>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:        1        2        3                                              
  2:       NA        4        5                                              
  > y = data.table(`<U+00E4>` = c(1, NA), `<U+00F6>` = c(2, 4), `<U+00FC>` = c(3,      5)) 
     <U+00E4> <U+00F6> <U+00FC>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:        1        2        3                                              
  2:       NA        4        5                                              
  Different column names
  Test 2298.4 ran without errors but failed check that x equals y:
  > x = rbindlist(list(y, x), use.names = TRUE, fill = TRUE) 
     <f6> <fc> <U+00E4>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:    4    5       NA                                              
  2:    2    3        1                                              
  > y = data.table(`<U+00F6>` = c(4, 2), `<U+00FC>` = c(5, 3), `<U+00E4>` = c(NA,      1)) 
     <U+00F6> <U+00FC> <U+00E4>  [Key= Types=dou,dou,dou Classes=num,num,num]
  1:        4        5       NA                                              
  2:        2        3        1                                              
  Different column names
  V1
  1
  V1
  0,1
  0,2
  a,b
  a
  
  Sun Sep 21 04:23:35 2025  endian==little, sizeof(long double)==16, capabilities('long.double')==TRUE, longdouble.digits==64, sizeof(pointer)==8, TZ=='UTC', Sys.timezone()=='UTC', Sys.getlocale()=='C', l10n_info()=='MBCS=FALSE; UTF-8=FALSE; Latin-1=FALSE; codeset=ANSI_X3.4-1968', getDTthreads()=='This installation of data.table has not been compiled with OpenMP support.; omp_get_num_procs()==1; R_DATATABLE_NUM_PROCS_PERCENT==unset (default 50); R_DATATABLE_NUM_THREADS==unset; R_DATATABLE_THROTTLE==unset (default 1024); omp_get_thread_limit()==1; omp_get_max_threads()==1; OMP_THREAD_LIMIT==unset; OMP_NUM_THREADS==unset; RestoreAfterFork==true; data.table is using 1 threads with throttle==1024. See ?setDTthreads.', .libPaths()=='/builds/Rdatatable/data.table/data.table.Rcheck','/usr/local/lib/R/site-library','/usr/lib/R/site-library','/usr/lib/R/library', zlib header files were not found when data.table was compiled
  Error in test.data.table() : 
    5 errors out of 10961. Search tests/tests.Rraw for test numbers 2194.7, 2298.1, 2298.2, 2298.3, 2298.4. Duration: 27.5s elapsed (27.2s cpu).
```
It could use same escape as:
```
Tests 2253* skipped because they need a UTF-8 locale.
  Tests 2266* skipped. n-tilde cannot be represented in the native encoding on this system. Please report to the data.table issue tracker.
```
So we don't raise false positive failure.

### LINKED ISSUE #7336: Add argument like 'requires_utf8' to test()
Maybe it would be better as an argument to `test()`, WDYT?

`test(requires_utf8 = TRUE)` then run the `utf8_check_expr` code inside `test()`, avoids the need for `local()`+`eval()` and will be easier to remember when writing future tests.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/pull/7335#pullrequestreview-3250737037_

ben-schwen:
Closes #7336
Closes #1343
Closes #7333

### DIFF
--- a/R/test.data.table.R
+++ b/R/test.data.table.R
@@ -396,7 +396,7 @@ test = function(num, x, y=TRUE,
   }
   # Check UTF-8 requirement
   if (!isFALSE(requires_utf8)) {
-    test_str = if (isTRUE(requires_utf8)) "\u00F1\u00FC\u3093" else requires_utf8
+    test_str = if (isTRUE(requires_utf8)) "\u00F1\u00FC\u3093" else requires_utf8 # the default test_str are UTF-8 symbols we found over time, TOOD: harden this default
     if (!utf8_check(test_str)) {
       # nocov start
       last_utf8_skip = get0("last_utf8_skip", parent.frame(), ifnotfound=0, inherits=TRUE)


aitap:
Sorry if I'm late to note this, but wouldn't a more reliable test for this be the same thing as we currently use for ñ in [test 2266](https://github.com/Rdatatable/data.table/blob/c26450af1d7b2381844d003867fd7f20a25ec8d0/inst/tests/tests.Rraw#L18937-L18950)? A test may require some symbols (ñ, ü, ん) to be representable in the native encoding. The symbols may be represented using Unicode escapes (`\uXXXX`) as they currently do. If `!identical(foo, enc2native(foo))`, then the test must be skipped.

ben-schwen:
> The symbols may be represented using Unicode escapes (`\uXXXX`) as they currently do. If `!identical(foo, enc2native(foo))`, then the test must be skipped.

Good point. I have integrated this for the utf8_check.

inst/tests/tests.Rraw
```diff
@@ -3569,7 +3570,31 @@ DT[,`:=`(last.x=tail(x,1L),last.x1=tail(x1,1L)),by=y]
 test(1086, class(DT$last.x), c("POSIXct", "POSIXt"))
 test(1087, class(DT$last.x1), "ITime")
 
-# Tests 1088-1093 were non-ASCII. Now in DtNonAsciiTests
+# chmatch on 'unknown' encoding (e.g. as.character(as.symbol("\u00E4")) )falling back to match, #2538 and #4818
+x1 = c("al\u00E4", "ala", "\u00E4allc", "coep")
+x2 = c("ala", "al\u00E4")
+tstc = function(y) unlist(lapply(y, function(x) as.character(as.name(x))), use.names=FALSE)
+test(1088.1, requires_utf8="\u00E4", chmatch(x1, x2), match(x1, x2)) # should not fallback to "match"
```
MichaelChirico: maybe easier to understand/write tests as `requires_utf8=c(x1, x2)`?

also, maybe better as a `local()` check here too, to reduce the visual noise of all the identical `requires_utf8=` inputs?

inst/tests/tests.Rraw
```diff
@@ -3569,7 +3570,31 @@ DT[,`:=`(last.x=tail(x,1L),last.x1=tail(x1,1L)),by=y]
...
+# change x2 to symbol to character
+test(1090.1, requires_utf8="\u00E4", chmatch(x1,tstc(x2)), match(x1, tstc(x2))) # should fallback to match in "table"
+test(1090.2, requires_utf8="\u00E4", x1 %chin% tstc(x2), x1 %in% tstc(x2))
+# both are symbols to characters
+test(1091.1, requires_utf8="\u00E4", chmatch(tstc(x1), tstc(x2)), match(tstc(x1), tstc(x2))) # should fallback to "match" in "x" as well.
+test(1091.2, requires_utf8="\u00E4", tstc(x1) %chin% tstc(x2), tstc(x1) %in% tstc(x2))
+# for completness, include test from #2528 of non ascii LHS of := (it could feasibly fail in future due to something other than chmatch)
+
+local(if (utf8_check("\u00E4")) {
+eval(parse(text='
```
MichaelChirico: why do we need `eval(parse())`?

also, `parse(keep.source=FALSE)` for micro-improvement

inst/tests/tests.Rraw
```diff
@@ -3569,7 +3570,31 @@ DT[,`:=`(last.x=tail(x,1L),last.x1=tail(x1,1L)),by=y]
...
+  DT = data.table(pas = c(1:5, NA, 6:10), good = c(1:10, NA))
+  setnames(DT, "pas", "p\u00E4s")
+  test(1092, requires_utf8="\u00E4", eval(parse(text="DT[is.na(p\u00E4s), p\u00E4s := 99L]")), data.table("p\u00E4s" = c(1:5, 99L, 6:10), good = c(1:10,NA)))
```
MichaelChirico: nested `eval(parse())`... gnarly

inst/tests/tests.Rraw
```diff
@@ -17653,12 +17700,9 @@ test(2194.4, endsWithAny(letters, 'e'), error="Internal error.*types or lengths
 test(2194.5, endsWithAny(NA_character_, 'a'), FALSE)
 test(2194.6, endsWithAny(character(), 'a'), error="Internal error.*types or lengths incorrect")
 # file used in encoding tests
-txt = readLines(testDir("issue_563_fread.txt"))
-local(if (eval(utf8_check_expr)) {
-  test(2194.7, endsWithAny(txt, 'B'), error="Internal error.*types or lengths incorrect") # txt is length 5
-} else {
-  cat("Test 2194.7 skipped because it needs a UTF-8 locale.\n")
-})
+needed_chars = "\u0105\u017E\u016B\u012F\u0173\u0117\u0161\u0119"
```
MichaelChirico: maybe easier to read in vector form?

```suggestion
needed_chars = c("\u0105", "\u017E", "\u016B", "\u012F", "\u0173", "\u0117", "\u0161", "\u0119")
```

inst/tests/tests.Rraw
```diff
@@ -18650,59 +18694,56 @@ test(2252.2, dt[, let(b=2L)], error = "\\[ was called on a data.table.*not data.
...
-  test(2253.20, options=list(datatable.prettyprint.char = 1L), data.table(a = c("abc", NA)), output="      a\n1: a...\n2: <NA>")
-} else {
-  cat("Tests 2253* skipped because they need a UTF-8 locale.\n")
-})
+accented_a = "\u0061\u0301"
+ja_ichi = "\u4E00"
+ja_ni = "\u4E8C"
+ja_ko = "\u3053"
+ja_n = "\u3093"
+nc = paste0(accented_a, ja_ichi, ja_ni, ja_ko, ja_n)
```
MichaelChirico:
```suggestion
nc = c(accented_a, ja_ichi, ja_ni, ja_ko, ja_n)
```

NEWS.md
```diff
@@ -24,6 +24,8 @@
 
 3. Vignettes are now built using `litedown` instead of `knitr`, [#6394](https://github.com/Rdatatable/data.table/issues/6394). Thanks @jangorecki for the suggestion and @ben-schwen and @aitap for the implementation.
 
+4. `test()` gains new argument `requires_utf8` to skip tests when UTF-8 support is not available, [#7336](https://github.com/Rdatatable/data.table/issues/7336). Thanks @MichaelChirico for the suggestion and @ben-schwen for the implementation.
```
MichaelChirico: `test()` is not exported, so I'm not sure this is the right framing. Maybe

> The data.table test suite is a bit more robust to lacking UTF-8 support [...]

inst/tests/tests.Rraw
aitap: When the parser sees constructs like

```r
data.table("\u00FCber" = c(1, 0, 0, 0, 0))
```

...it needs to construct a "language" (`LANGSXP`) call object where `TAG(CADR(call))` is a symbol whose `PRINTNAME` is a `CHARSXP` saying `über`. That `CHARSXP` _must_ be in the native encoding: requiring a single encoding makes it possible to compare pointers to `SYMSXP` values for equality (unlike `CHARSXP` where we have to test `NEED2UTF8` and so on), and the native encoding was the default back before R had string encodings.

So when the parser tries to translate that Unicode string into the native encoding, it fails, emits a warning, and probably fails the following test, because the resulting string contains a substitution sequence:

```sh
LC_ALL=C R -q -s -e 'parse(text = r"{data.table("\u00FCber" = c(1, 0, 0, 0, 0))}")'
```
```
expression(data.table(`<U+00FC>ber` = c(1, 0, 0, 0, 0)))
Warning message:
In parse(text = "data.table(\"\\u00FCber\" = c(1, 0, 0, 0, 0))") :
  unable to translate '<U+00FC>ber' to native encoding
```

Without the runtime `eval(parse(...))`, this warning happens during `source()` with no way to avoid it.

MichaelChirico: it would be good to write that down somewhere as a reminder, but I'm not sure the best place to do it while being (1) discoverable and (2) not repetitive.

maybe (1) document it near `require_utf8` in R/test.data.table and (2) add a comment by each eval(parse()) like "see require_utf8 description"?

NEWS.md
ben-schwen: done

inst/tests/tests.Rraw

inst/tests/tests.Rraw

inst/tests/tests.Rraw

inst/tests/tests.Rraw

inst/tests/tests.Rraw
ben-schwen: good catch!

inst/tests/tests.Rraw
```diff
@@ -4608,7 +4640,28 @@ test(1228.4, class(DT), class(DT[, sum(b), by=a]))
 test(1228.5, class(DT), class(DT[a>1, sum(b), by=a]))
 test(1228.6, class(DT), class(DT[a>1, c:=sum(b), by=a]))
 
-# test 1229 was non-ASCII, now in package DtNonAsciiTests
+# savetl_init error after error, in v1.9.2, thanks Arun
+DT = data.table(x=1:5, y=10:6)
+test(1229.1, DT[forderv(DT, -1)], error="non-existing column")
+test(1229.2, setkey(DT), data.table(x=1:5, y=10:6, key="x,y"))
+# umlaut in column names (red herring I think, but testing anyway)
+local(if (utf8_check("\u00e4\u00f6\u00fc")) {
```
MichaelChirico: Why `\u00e4`/`\u00f6`? I only see `\u00fc` required here

R/test.data.table.R
```diff
@@ -387,6 +394,20 @@ test = function(num, x, y=TRUE,
       Sys.unsetenv(names(old)[!is_preset])
     }, add=TRUE)
   }
+  # Check UTF-8 requirement
+  if (!isFALSE(requires_utf8)) {
+    test_str = if (isTRUE(requires_utf8)) "\u00F1\u00FC\u3093" else requires_utf8
```
MichaelChirico: Any reason for these 3 particular characters? `ñüん`. A comment might be appropriate.

MichaelChirico (approved):
2 more minor points, feel free to merge. Thanks!

inst/tests/tests.Rraw
ben-schwen: good spot, removed the unrelated chars here

R/test.data.table.R
ben-schwen: thats all symbols we found over the time in issues related to different encodings. added comment about this.

PR #7394, opened 2025-10-27
improved read onlyness and constness

badasahog:
followup on #7368

Increase in speed is minimal (if even existent), but this should hopefully improve readability.

Knowing ahead of time if an object will be modified could make the code more digestable.

There's still more ground to cover, but I'll wait for this pr to go through first.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -139,7 +139,6 @@ S3method(split, data.table)
 export(dcast, melt)
 S3method(dcast, data.table)
 S3method(melt, data.table)
-S3method(melt, default)
 
 # exported for historical reasons -- if reshape2 is higher on search path,
 # dcast(DT) will not dispatch since reshape2::dcast is not generic. So users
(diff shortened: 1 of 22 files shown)


src/between.c
```diff
@@ -20,13 +20,13 @@ SEXP between(SEXP x, SEXP lower, SEXP upper, SEXP incbounds, SEXP NAboundsArg, S
   const int longestBound = MAX(nl, nu);  // just for when check=TRUE
   if (!IS_TRUE_OR_FALSE(incbounds))
     error(_("'%s' must be TRUE or FALSE"), "incbounds");
-  const bool open = !LOGICAL(incbounds)[0];
-  if (!isLogical(NAboundsArg) || LOGICAL(NAboundsArg)[0]==FALSE)
+  const bool open = !LOGICAL_RO(incbounds)[0];
```
MichaelChirico, line 23: Hi @badasahog thanks for the change. Funnily enough, I think we started down the same rabbit hole independently -- much of the changes in the original PR wound up covered by #7611 and #7615.

I think there is still some value to what's left here, but it's mixed in with something that we decided against in those PRs:

https://github.com/Rdatatable/data.table/pull/7611#issuecomment-3775101675

Basically, code like `LOGICAL(x)[0]` is functionally _idiomatic_ in R. While `LOGICAL_RO(x)[0]` is more _technically_ correct, it will be visually distracting for R developers. And anyway, these cases can easily be found by regex & restored if we decide differently in the future.

There might be some cases in very hot loops where there is a performance reason to switch, but we'd want thorough benchmarking if so.

If you wouldn't mind please reverting these changes, it will make reviewing the rest of the PR much easier.

PR #7396, opened 2025-10-27
reformat and realign function table

badasahog:
Ignore the fork name.

Just reformatting to make it more readable.


src/init.c
```diff
@@ -49,124 +49,122 @@ Rcomplex NA_CPLX;
...
+  { "C_lock",                      (DL_FUNC)&lock, -1 },  // _ for these 3 to avoid Clock as in time
+  { "C_unlock",                    (DL_FUNC)&unlock, -1 },
+  { "C_islocked",                  (DL_FUNC)&islockedR, -1 },
+  { "CtestMsgR",                   (DL_FUNC)&testMsgR, -1 },
+  { "C_allNAR",                    (DL_FUNC)&allNAR, -1 },
+  { "CcoerceAs",                   (DL_FUNC)&coerceAs, -1 },
+  { "Ctest_dt_win_snprintf",       (DL_FUNC)&test_dt_win_snprintf, -1 },
+  { "Cdt_zlib_version",            (DL_FUNC)&dt_zlib_version, -1 },
+  { "Cdt_has_zlib",                (DL_FUNC)&dt_has_zlib, -1 },
+  { "Csubstitute_call_arg_namesR", (DL_FUNC)&substitute_call_arg_namesR, -1 },
```
MichaelChirico: This looks nice now, but if ever a new routine is added that is longer than `Csubstitute_call_arg_namesR`, it will induce a very large diff touching all of these lines.

So to complete this PR I would:

 - put `static const **` on the first line
 - indent the array contents by 2 spaces

And remove extra whitespace.

There are [a few](https://github.com/search?q=org%3Acran+lang%3Ac+%2F%5Cn%5Cs*%5C%7B%5Cs%2B%22%2F+%2FR_CallMethodDef%2F&type=code) packages doing `{ "..."` instead of `{"..."`, but it's [the minority](https://github.com/search?q=org%3Acran+lang%3Ac+%2F%5Cn%5Cs*%5C%7B%22%2F+%2FR_CallMethodDef%2F&type=code). Personally I prefer `{"..."`.

PR #7399, opened 2025-10-28
added newline after function signature, before opening brace

### LINKED ISSUE #7395: Inconsistent use of newline before braces on functions
Sometimes there's a newline before the brace, sometimes not.

For example:

```c
int has_new_line()
{
}

int no_new_line() {
}
```

I suggest doing the new line, but I'll await suggestions.

badasahog:
It was actually about a 50/50 split, so this should make code more consistent.

closes #7395

### DIFF
--- a/src/assign.c
+++ b/src/assign.c
@@ -23,7 +23,8 @@ static void finalizer(SEXP p)
   return;
 }
 
-void setselfref(SEXP x) {
+void setselfref(SEXP x)
+{
   if(!INHERITS(x, char_datatable))  return; // #5286
   SEXP p;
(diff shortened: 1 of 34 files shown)


jangorecki:
So far, the conclusion from the discussion on this issue is not to proceed with the change. The change would affect the history of 34 files, and it doesn’t seem necessary. It would be nice if the style were consistent, but it isn’t, and as long as it is not less readable, then I don't think there is anything to solve here.

ben-schwen:
As Jan pointed out, we very welcome PRs but dont want to put extra work in merge only for style reasons.

PR #7400, opened 2025-10-28
melt doc fixes

tdhock:
This PR includes two small fixes to the melt documentation

* typo: missing "of" added.
* nzchar() used in ?measure instead of `==""` for consistency with vignette.

### DIFF
--- a/man/measure.Rd
+++ b/man/measure.Rd
@@ -83,7 +83,7 @@ melt(who, measure.vars = measure(diagnosis, gender, ages, pattern="new_?(.*)_(.)
 print(melt(who, measure.vars = measure(
   diagnosis, gender, ages,
   ymin=as.numeric,
-  ymax=function(y)ifelse(nzchar(y), Inf, as.numeric(y)),
+  ymax=function(y)ifelse(nzchar(y), as.numeric(y), Inf),
   pattern="new_?(.*)_(.)(([0-9]{2})([0-9]{0,2}))"
 )), class=TRUE)
 }


vignettes/datatable-reshape.Rmd
ChristianWia, line 1: not impacting FR vignette

man/measure.Rd
```diff
@@ -83,7 +83,7 @@ melt(who, measure.vars = measure(diagnosis, gender, ages, pattern="new_?(.*)_(.)
 print(melt(who, measure.vars = measure(
   diagnosis, gender, ages,
   ymin=as.numeric,
-  ymax=function(y)ifelse(y=="", Inf, as.numeric(y)),
+  ymax=function(y)ifelse(nzchar(y), Inf, as.numeric(y)),
```
MichaelChirico: this is actually changing the meaning, `y == ""` is equivalent to `!nzchar(y)` (strictly peaking, it's `!nzchar(y, keepNA=TRUE)`). Did you mean to switch `yes=` and `no=`?

(it might help to link the part of the vignette you had in mind to make review easier)

tdhock: sorry about that, the line from datatable-reshape vignette is
```r
  ymax=function(y) ifelse(nzchar(y), as.numeric(y), Inf),
```
so yes you are right they should switch, thanks for the catch.

tdhock:
```suggestion
  ymax=function(y)ifelse(nzchar(y), as.numeric(y), Inf),
```

PR #7402, opened 2025-10-29
Make litedown vignettes compatible with pkgdown

rikivillalba:
#6583 
The purpose of this PR is to allow the vignettes to be rendered with litedown while still being compatible with rmarkdown since pkgdown uses rmarkdown to build the site. Some modifications were made:

- the `{r} Sys.date()` inline codes changed to !r Sys.date() (EDIT: this works in the YAML header only). They re compatible by both litedown and knitr/rmarkdown unlike `r Sys.date()` (knitr only) and `{r} Sys.date()` (litedown only)
- the script that adds the link to other language checks whether it is being rendered with knitr and if it is not the case, switch to litedown functions to get context. 
- single (external) code block in each vignette/language to print the i18n message


jangorecki:
I thought we have some workaround implemented for that in CI, but couldn't find it now. If there is we could possibly remove if this PR is merged.

vignettes/fr/datatable-fread-and-fwrite.Rmd
ChristianWia, line 1: @rikivillalba  ? you do not change the date with     date: !r Sys.Date()
Same remark applies to the EN vignette.

ben-schwen (approved):
Using `date: !r Sys.Date()` seems to not work with `pkgdwon`. At least its not building correctly for me with `pkgdown::build_site()`

Besides that, LGTM TY!

jangorecki:
this PR might be not needed as we will move away from pkgdown hopefully: #7534

ben-schwen:
> this PR might be not needed as we will move away from pkgdown hopefully: #7534

Yes, but right now, the website does not build after switching the vignettes to litedown

rikivillalba:
This PR not longer needed as #7534 #7628 #7601 were pushed.

PR #7406, opened 2025-11-03
`convertDate`: `GetOption` -> `GetOption1`

aitap:
The older non-API entry point must have sneaked in in a recent PR. See #6849 for the context.


ben-schwen:
Good catch. TY!

PR #7414, opened 2025-11-05
Clarify when := drops keys and when to use copy()

### LINKED ISSUE #7409: Document side effects of := on keys
The use of `copy()` here is subtle, it might be good to describe how to approach side effects a bit better in the documentation.

_Originally posted by @MichaelChirico in https://github.com/Rdatatable/data.table/pull/7401#discussion_r2487569639_

This should be added to our reference semantics vignette and probably to `help(':=')` if its not already there.

ANAMASGARD:
Document why := kills keys and when to use copy()

@ben-schwen raised this in #7409 after @MichaelChirico pointed out that 
the copy() pattern was too subtle in the code. The issue is about how := 
drops keys when you modify key columns, but this wasn't documented anywhere.

The discussion made it clear that people using tests with optimization= 
need to understand that copy() gets executed for each level due to lazy 
evaluation. Also important for anyone writing reusable functions or tests 
where side effects matter.

Added examples showing:
- Modifying key column = key goes away.
- Modifying non-key column = key stays  .
- Using copy() before := = original data untouched.

IN the below screenshot :-
When we modify a key column with :=, the key drops (becomes empty).
 Using copy() before :=, the original key is preserved.
The result's key still drops (because the values were modified).

Screenshot From 2025-11-05 16-00-02

FIXES  #7409


MichaelChirico:
This issue does not currently need a PR.

jangorecki:
Please do not post screenshots when content of the screenshot is to show the code, use code blocks of markdown instead

Issue #6148, opened 2024-05-22
add support for the Chinese encoding to read or write in fread or fwrite


anticmason:
Hi,
  I'm the heavy user for this package from China.
  Could it be possible to add an option <encoding='gbk'>for the argument：encoding from fread or fwrite function,which couldn't deal with Chinese encoding file correctly,like pandas's read_csv or read_excel ?
  Thanks a lot!
  Looking forward to your reply~

tdhock:
it may be possible, but can you please provide a minimal reproducible example? ideally something like 
```r
fread(text="some chinese characters in gbk encoding")
```
or upload a small file (1 or 2 rows) which represents your issue.
Also can you please explain what behavior you get, and what behavior you expected?
you wrote "fread or fwrite function,which couldn't deal with Chinese encoding file correctly" -- what does it mean to deal with it correctly? or not?

anticmason:
Hi,
 I've attached the file,when I use fread function to read data,it comes with messy code,it can't recognize correctly，encoding argument doesn't have an option to deal with it。 However, when I use pandas's read_csv, like：
pd.read_csv('file path',encoding='gbk'), it goes well。
Thanks~
[APP原生分区域点击日报表_20240523_1716426197475.csv](https://github.com/Rdatatable/data.table/files/15428327/APP._20240523_1716426197475.csv)

anticmason:
Hi,

This is obviously the messy code, since encoding argument from fread function doesn't support gbk,correct result attached,please check~ 













在 2024-05-25 03:55:44，"Toby Dylan Hocking" ***@***.***> 写道：

I don't get any error from fread, what do you get? what did you expect to get?

>data.table::fread("~/Downloads/APP._20240523_1716426197475.csv")
   ͳ\xbc\xc6\xc8\xd5\xc6\xda \xb5\xe3\xbb\xf7ʡ\xb7ݱ\xe0\xc2\xeb<int><int>1:202405192102:20240519210
   \xb5\xe3\xbb\xf7ʡ\xb7\xdd\xc3\xfb\xb3\xc6                   Ƶ\xb5\xc0<char><char>1:                                 \xc9Ϻ\xa3 \xd3\xe0\xc1\xbf\xb2\xe9ѯH52:                                 \xc9Ϻ\xa3 \xd3\xe0\xc1\xbf\xb2\xe9ѯH5
[...]

PR #7422, opened 2025-11-08
add fread file connection support

### LINKED ISSUE #561: [R-Forge #4931] Support file connections for fread
Submitted by: Chris Neff; Assigned to: Nobody; [R-Forge link](https://r-forge.r-project.org/tracker/index.php?func=detail&aid=4931&group_id=240&atid=978)

I use a corporate internal networked file system for much of my data, and so often times i need to call read.csv with a file connection.  fread doesn't support this yet.

Namely I would like the following to work:

f = file("~/path/to/file.csv")

dt = fread(f)

### LINKED ISSUE #4329: fread tries to map memory for the entire file when using nrows
Hi,

I've been trying to use `fread()` to import a large (12GB) tab-delimited text file, which is too large for my machine to import in its entirety. I thought that I would be able to use the `nrows` parameter to import a cut-down version of the file to draft my code on, but this results in the following error:

```text
System errno 22 unmapping file: Invalid argument
Error in fread("data.tab", header = T, sep = "\t", nrows = 10L) : 
  Opened 11.63GB (12491321418 bytes) file ok but could not memory map it. This is a 64bit process. There is probably not enough contiguous virtual memory available.
```

However, if I use `head` to create a new file containing only the first 11 rows of the original file I am able to use `fread()` to import the new file without issue.

Below is a sample of the code I am running to replicate the issue:

```r
### File Size (Original File) ###
file.size("data.tab")/1e9 # Roughly 12.5 GB

### Import First 10 Rows (Original File) ###
dt <- fread("data.tab",
            header = T, sep = "\t",
            nrows = 10L) # Fails

### Create New File Using the First 11 Rows of the Existing One ###
system("head -n 11 data.tab > data_head.tab")

### File Size (New File) ###
file.size("data_head.tab")/1e3 # Roughly 330 KB

### Import (New File) ###
dt <- fread("data_head.tab",
            header = T, sep = "\t") # Succeeds
```

I have searched the current issues log for _data.table_, but the only problem I've found resembling mine is issue #2321, which was closed on 3rd March 2018. The closing messages for this issue stated that the issue had been fixed in _data.table_ version 1.10.5 through the use of lazy memory mapping. However, I'm using _data.table_ version 1.12.8 and seem to be stumbling across the same issue. Once imported, the 10-row data table is only 1.9MB - nowhere near the physical memory limit of my machine (4GB).

My output for `sessionInfo()` is below:

```text
R version 3.6.3 (2020-02-29)
Platform: aarch64-unknown-linux-gnu (64-bit)
Running under: Manjaro ARM

Matrix products: default
BLAS:   /usr/lib/libopenblasp-r0.3.9.so
LAPACK: /usr/lib/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_GB.UTF-8          LC_NUMERIC=C                 
 [3] LC_TIME=en_GB.UTF-8           LC_COLLATE=en_GB.UTF-8       
 [5] LC_MONETARY=en_GB.UTF-8       LC_MESSAGES=en_GB.UTF-8      
 [7] LC_PAPER=en_GB.UTF-8          LC_NAME=en_GB.UTF-8          
 [9] LC_ADDRESS=en_GB.UTF-8        LC_TELEPHONE=en_GB.UTF-8     
[11] LC_MEASUREMENT=en_GB.UTF-8    LC_IDENTIFICATION=en_GB.UTF-8

attached base packages:

other attached packages:
[1] data.table_1.12.8 rkward_0.7.1     

[1] compiler_3.6.3 tools_3.6.3   
```
Thank you in advance,

Ben

ben-schwen:
Closes #561
Also closes #4329 with a nice workaround by just wrapping the file to read with `file()` and use the connection interface.

Spills to file since this seems to cover almost all cases. Also respects `nrows` parameter, so for peeking it does not need to spill the whole file.

We can't magically make file connections faster since we do not have random access like with `mmap`.

For the benchmark below, note that my `tempdir` already points to my SSD (therefore we cant see a big difference) and I dont have a HDD on this PC.

fread_con_vs_readtable

Extending this to `1e8` rows instead of `1e7` and verbose also shows that half of the time spent is for spilling to disk (for large files).

``` r
Read 100000000 rows x 4 columns from 3.916GiB (4205299167 bytes) file in 00:02.151 wall clock time
[12] Finalizing the datatable
  Type counts:
         1 : int32     '7'
         2 : float64   '9'
         1 : string    'E'
=============================
   2.158s ( 50%) Spill connection to tempfile (3.916GiB)
   0.000s (  0%) Memory map 3.916GiB file
   0.002s (  0%) sep=',' ncol=4 and header detection
   0.000s (  0%) Column type detection using 10049 sample rows
   0.413s ( 10%) Allocation of 109975410 rows x 4 cols (2.868GiB) of which 100000000 ( 91%) rows used
   1.736s ( 40%) Reading 4010 chunks (0 swept) of 1.000MiB (each chunk 24937 rows) using 10 threads
   +    0.575s ( 13%) Parse to row-major thread buffers (grown 0 times)
   +    0.759s ( 18%) Transpose
   +    0.402s (  9%) Waiting
   0.000s (  0%) Rereading 0 columns due to out-of-sample type exceptions
   4.309s        Total
```

<details>

``` r
library(data.table)
library(atime)
set.seed(123)
N = 1e7
test_df <- data.frame(
    a = sample(1:1000, N, replace=TRUE),
    b = rnorm(N),
    c = sample(letters, N, replace=TRUE),
    d = runif(N)
)
f = tempfile(fileext = '.csv')
fwrite(test_df, f)

Nseq = 10^seq(2, log10(N), .25)
read = atime(N = Nseq, seconds.limit=1,
    fread_con = fread(file(f), nrows = N),
    fread_con_RAM = fread(file(f), nrows = N, tmpdir = "/dev/shm"),
    readtable = read.table(file(f), header=TRUE, sep=',', nrows = N),
    fread = fread(f, nrows = N)
)

plot(read)
```

</details>


aitap:
Looking forward to reviewing this. I was wondering who was going to be
the first brave soul to start using the connections API after they
"officially" became experimental in r88181-r88223 (April-May 2025). Not
counting the even braver souls who've been working around the non-API
checks, 'curl' and 'iotools' are the only CRAN packages that openly
link to these entry points (as they have been doing for a long time).

ben-schwen:
I guess there are some more cool kids on the [CRAN block](https://github.com/search?q=org%3Acran+R_GetConnection&type=code) using `R_GetConnection` (although not too many)

R/fread.R
```diff
@@ -81,6 +90,59 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
     }
     file = tmpFile
   }
+  connection_spill_info = NULL
+  if (input_is_con) {
+    if (verbose) {
+      catf("[00] Spill connection to tempfile\n  Connection class: %s\n  Reading connection into RAM buffer... ", toString(class(input)))
```
aitap:
```suggestion
      catf("[00] Spill connection to tempfile\n  Connection class: %s\n  Reading connection into a temporary file... ", toString(class(input)))
```

src/freadR.c
```diff
@@ -724,6 +740,82 @@ void progress(int p, int eta)
 }
 // # nocov end
 
+// Spill connection contents to a tempfile so R-level fread can treat it like a filename
+SEXP spillConnectionToFile(SEXP connection, SEXP tempfile_path, SEXP nrows_limit) {
```
aitap: The interface stayed unchanged for a long time, so it's probably not worth spending effort on handling `R_CONNECTIONS_VERSION != 1` well, but let's avoid a potential compilation error (and fail unit tests instead).

```suggestion
SEXP spillConnectionToFile(SEXP connection, SEXP tempfile_path, SEXP nrows_limit) {
#if R_CONNECTIONS_VERSION != 1
INTERNAL_STOP(_("spillConnectionToFile: unexpected R_CONNECTIONS_VERSION = %d", R_CONNECTIONS_VERSION)); // # nocov
#else
```

src/freadR.c
```diff
@@ -724,6 +740,82 @@ void progress(int p, int eta)
...
+
+    if (limit_rows && nrows_seen >= row_limit) {
+      break;
+    }
+  }
+
+  free(buffer);
+  fclose(outfile);
+  return ScalarReal((double)total_read);
+}
```
aitap, line 846:
```suggestion
#endif // was R_CONNECTIONS_VERSION not != 1?
}
```

R/fread.R
```diff
@@ -81,6 +90,59 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
...
+    }
+    spill_started.at = proc.time()
+    con_summary = summary(input)
+    con_desc = con_summary$description
+    con_class = class1(input)
+    con_open = isOpen(input)
+
+    needs_reopen = FALSE
+    if (con_open) {
+      binary_modes = c("rb", "r+b", "wb", "w+b", "ab", "a+b")
```
aitap: Use of binary modes disables both newline conversions and encoding conversions. Would a user want to use `file(encoding="...")` to decode a file not in UTF-8 or native encoding?

R/fread.R
```diff
@@ -81,6 +90,59 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
...
+
+    if (needs_reopen) {
+      close(input)
+      input = switch(con_class,
+                     "file" = file(con_desc, "rb"),
+                     "gzfile" = gzfile(con_desc, "rb"),
+                     "bzfile" = bzfile(con_desc, "rb"),
+                     "url" = url(con_desc, "rb"),
+                     "pipe" = pipe(con_desc, "rb"),
+                     stopf("Unsupported connection type: %s", con_class))
```
aitap:
```suggestion
                     stopf("Don't know how to reopen connection type '%s'. Need a connection opened in binary mode to continue.", con_class))
```

R/fread.R
```diff
@@ -81,6 +90,59 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
...
+      open(input, "rb")
+      close_con = input
+    }
+    tmpFile = tempfile(tmpdir=tmpdir)
+    on.exit(unlink(tmpFile), add=TRUE)
+    bytes_copied = .Call(CspillConnectionToFile, input, tmpFile, as.numeric(nrows))
+    spill_elapsed = (proc.time() - spill_started.at)[["elapsed"]]
+
+    if (bytes_copied == 0) {
+      warningf("Connection has size 0. Returning a NULL %s.", if (data.table) 'data.table' else 'data.frame')
```
aitap, line 162: Maybe use `on.exit(add = TRUE)` when setting `close_con` as well?
```suggestion
      warningf("Connection has size 0. Returning a NULL %s.", if (data.table) 'data.table' else 'data.frame')
      if (!is.null(close_con)) close(close_con)
```

src/freadR.c
```diff
@@ -724,6 +740,82 @@ void progress(int p, int eta)
...
+  if (!buffer) {
+    fclose(outfile);                                             // # nocov
+    STOP(_("spillConnectionToFile: failed to allocate buffer")); // # nocov
+  }
+
+  size_t total_read = 0;
+  size_t nrows_seen = 0;
+
+  while (true) {
+    size_t nread = R_ReadConnection(con, buffer, chunk_size);
```
aitap: Some read methods can signal R errors (e.g. `bzfile`, `zstdfile` for corrupted files). We can move the loop into a function to be called by [`R_ExecWithCleanup`](https://cran.r-project.org/doc/manuals/R-exts.html#index-R_005fExecWithCleanup-1) or use more static variables with deferred cleanup in case of a failed call.

src/freadR.c
```diff
@@ -724,6 +740,82 @@ void progress(int p, int eta)
...
+
+  // Read and write in chunks // TODO tune chunk size
+  size_t chunk_size = 256 * 1024;
+  char *buffer = malloc(chunk_size);
+  if (!buffer) {
+    fclose(outfile);                                             // # nocov
+    STOP(_("spillConnectionToFile: failed to allocate buffer")); // # nocov
+  }
+
+  size_t total_read = 0;
```
aitap, line 777: An R connection may produce more than 4 gigabytes of data even on a 32-bit system, because R enables Large File Support. `fread` had no use for it because it relies on the file fitting in the address space to work, so I think that for a large file, `fwrite` below would fail before it could overflow `total_read`. So it should be safe to use `size_t` here.

src/freadR.c
```diff
@@ -724,6 +740,82 @@ void progress(int p, int eta)
...
+  size_t row_limit = 0;
+  if (limit_rows) {
+    row_limit = (size_t)nrows_max;
+    if (row_limit == 0) row_limit = 100;  // read at least 100 rows if nrows==0
+    row_limit++; // cater for potential header row
+  }
+
+  FILE *outfile = fopen(filepath, "wb");
+  if (outfile == NULL) {
+    STOP(_("spillConnectionToFile: failed to open temp file '%s' for writing"), filepath); // # nocov
```
aitap:
```suggestion
    STOP(_("spillConnectionToFile: failed to open temp file '%s' for writing: %s"), filepath, strerror(errno)); // # nocov
```

src/freadR.c
```diff
@@ -724,6 +740,82 @@ void progress(int p, int eta)
...
+        }
+      }
+    }
+
+    size_t nwritten = fwrite(buffer, 1, bytes_to_write, outfile);
+    if (nwritten != bytes_to_write) {
+      // # nocov start
+      free(buffer);
+      fclose(outfile);
+      STOP(_("spillConnectionToFile: write error (wrote %zu of %zu bytes)"), nwritten, bytes_to_write);
```
aitap:
```suggestion
      STOP(_("spillConnectionToFile: write error %s (wrote %zu of %zu bytes)"), strerror(errno), nwritten, bytes_to_write);
```

aitap (commented):
Should be good after fixing the potential resource leak caused by errors in `R_ReadConnection`.

R/fread.R
```diff
@@ -81,6 +90,59 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
...
+    if (con_open) {
+      binary_modes = c("rb", "r+b", "wb", "w+b", "ab", "a+b")
+      if (!con_summary$mode %chin% binary_modes) needs_reopen = TRUE
+    }
+
+    close_con = NULL
+
+    if (needs_reopen) {
+      close(input)
+      input = switch(con_class,
```
MichaelChirico: Let's make this an S3 method with the below as exported methods, then other downstream connection types can easily define methods that "just work" on other custom connections, WDYT?

R/fread.R
MichaelChirico: We say "spill to tempfile" earlier, so the suggestion would be redundant.

I think more accurate is something like "Reading connection by chunks into RAM buffer" (and then spilling ot file), is that right?

R/fread.R
ben-schwen: I could not make `file(encoding=...)` work with the example in #6148 and certain text modes, hence, I left it as a follow up.

R/fread.R
ben-schwen: Good point, added it!

R/fread.R
```diff
@@ -1,3 +1,37 @@
+# S3 generic for reopening connections in binary mode
+reopen_connection = function(con, description, ...) {
+  UseMethod("reopen_connection")
+}
+
+reopen_connection.default = function(con, description, ...) {
+  con_class = class1(con)
+  stopf("Don't know how to reopen connection type '%s'. Need a connection opened in binary mode to continue.", con_class)
+}
```
ben-schwen: @MichaelChirico Not entirely happy with this, since we need to pass description and con is more or less only used for the S3 dispatch

R/fread.R
ben-schwen: done, see below.

src/freadR.c
ben-schwen: Nice find! We should probably do this in fread too right?

inst/tests/tests.Rraw
```diff
@@ -21858,3 +21862,13 @@ test(2344.04, key(DT[, .(V4 = c("b", "a"), V2, V5 = c("y", "x"), V1)]), c("V1",
 
 # fread with quotes and single column #7366
 test(2345, fread('"this_that"\n"2025-01-01 00:00:01"'), data.table(this_that = as.POSIXct("2025-01-01 00:00:01", tz="UTC")))
+
+# fread supports connections #561
```
MichaelChirico, line 21567: what happens if we try to `fread()` from a connection that's already been opened and a few bytes consumed?

e.g. i can imagine it would be useful as a way to read a non-standard file with a header:

```
This is metadata.
I have some other content here
Now I write '=' to signify the content begins.
==============================================
a,b,c
1,2,3
```

With the file open as a connection, consume until the last `=`, then pass the "tail" to `fread()`.

man/reopen_connection.Rd
```diff
@@ -0,0 +1,26 @@
+\name{reopen_connection}
+\alias{reopen_connection}
```
MichaelChirico: Add `\alias{}` for the implemented methods?

R/fread.R
MichaelChirico: I see what you mean... what about a function factory instead? Two ideas came to mind (the names are bad):

```r
connection_host.file <- file
connection_host.gzfile <- gzfile
...

# and then
input = connection_host(input)(con_summary$description, "rb")

# or, bundle 'rb' in the method:
connection_rb.file <- function(con) function(description) file(description, "rb")
connection_rb.gzfile <- function(con) function(description) gzfile(description, "rb")

# and then
input = connection_rb(input)(con_summary$description)
```

Basically we are trying to define an inverse mapping from connection class to the generator for that class, which is not a very well-defined concept in R (certainly not in S3).

inst/tests/tests.Rraw
ben-schwen: I added a test where we consume the first lines and compare it with `skip`

man/reopen_connection.Rd
ben-schwen: done!

R/fread.R
ben-schwen: done, although the name is not perfect :(

man/connection_opener.Rd
```diff
@@ -0,0 +1,40 @@
...
+\alias{connection_opener.bzfile}
+\alias{connection_opener.url}
+\alias{connection_opener.unz}
+\alias{connection_opener.pipe}
+\title{ Create a function to open connections in binary mode }
+\description{
+  S3 generic that returns a function to open a connection in binary read mode. Used internally by \code{fread}. Exported so packages with custom connection classes can define methods.
+}
+\usage{
+connection_opener(con, ...)
```
MichaelChirico: some other ideas... connection_generator, connection_starter, connection_source, connection_creator, connection_constructor

ben-schwen: binary_reopener? connection_reopener?

src/freadR.c
aitap: Possibly, yes. There may be even a clever way to implement that without having to teach `fread.c` about R API.

R/fread.R
aitap: Apparently, decoding the data requires [calling the `fgetc` connection method](https://github.com/r-devel/r-svn/blob/0f7760a26a9381915a52609bc18fd776eee23a41/src/main/connections.c#L4477), not the `read` method. No other method [calls `iconv`](https://github.com/r-devel/r-svn/blob/0f7760a26a9381915a52609bc18fd776eee23a41/src/main/connections.c#L646) to decode data.

`spillConnectionToFile` could choose between the two implementations depending on whether `con->text` is `TRUE`. If the connection is not yet open but `con->encname` is specified, it could perform [the `readLines` trick](https://github.com/r-devel/r-svn/blob/0f7760a26a9381915a52609bc18fd776eee23a41/src/main/connections.c#L4436) for decoding to UTF-8 instead of the native encoding.

man/connection_opener.Rd
ben-schwen: went with `binary_reopener`

R/fread.R
aitap: [That](https://stat.ethz.ch/pipermail/r-package-devel/2026q1/012288.html) idea [does not fly](https://stat.ethz.ch/pipermail/r-package-devel/2026q1/012289.html) (and would have been bad for performance). I guess this means that there's no useful way to work with text connections from C.

We could perform the re-encoding ourselves using `Riconv`, but that still requires knowing the connection encoding, which can only be read from `con->encname`.

aitap (approved):
I don't like having to introduce the `binary_reopener` as another entry point to support, but there doesn't seem to be any other way to proceed given the current API. Thank you very much!

PR #6955, opened 2025-05-08
Switch to h2/h3 hierarchy of NEWS (not h1/h2)

MichaelChirico:
See https://github.com/tidyverse/style/issues/245, https://stat.ethz.ch/pipermail/r-devel/2025-May/084009.html.

PR #7423, opened 2025-11-09
Add cheatsheets to homepage

### LINKED ISSUE #3374: Room for new data.table cheat sheet on the homepage?
Some time ago, I published a cheat sheet for data.table and I've been expanding on it quite a bit since then: https://github.com/rstudio/cheatsheets/blob/master/datatable.pdf

It's currently hosted on RStudio's homepage (and GitHub) but I was wondering if there's room for it on the data.table homepage as well?

Just to be clear, I don't think it should replace the existing cheat sheet on the data.table homepage. I'm trying to brand this as a "visual cheat sheet" mostly aimed at people who are new or casual data.table users.

ben-schwen:
Closes #3374

This also fixes current error that NEWS is not correctly rendered.

Header would look like this (note that I have clicked on Cheatsheet such that the menu is opened)
image


_pkgdown.yml
```diff
@@ -67,6 +70,13 @@ navbar:
     reference:
       text: Manual
       href: reference/index.html
+    cheatsheet:
+      text: Cheat Sheet
```
ben-schwen:
```suggestion
      text: Cheatsheet
```

NEWS.md
```diff
@@ -1,4 +1,4 @@
-# data.table news and updates
+## data.table news and updates
```
MichaelChirico, line 1: Why is this needed? The intention in #6955 was this is the unique `<h1>` header in the NEWS, then each version gets a further-indented `<h2>` heading. What went wrong?

ben-schwen: As noted in #7425 you R CMD check needs either all h1 or all h2 headers. [Pkgdown](https://pkgdown.r-lib.org/reference/build_news.html) also complains about this!
``` r
pkgdown::build_news()
# Warning message:
# In build_news_single(pkg) : In NEWS.md, no version headings found
# See `?pkgdown::build_news()` for expected structure.
```

MichaelChirico: Seems wrong to me but oh well :)

PR #7424, opened 2025-11-09
add [[<- syntax

### LINKED ISSUE #6734: Broken behaviour after `[[<-`; segfaults on CRAN version
Adding or changing a column using `[[<-` seems to bring a `data.table` into a broken state. `set()` fails afterwards, and printing the `data.table` changes its `truelength()`.

The following is with GitHub HEAD. Some versions seem to segfault here, so don't run this if you want to keep your R session:
```r
library("data.table")
#> data.table 1.16.99 IN DEVELOPMENT built 2025-01-18 17:55:52 UTC; user using 4 threads (see ?getDTthreads).  Latest news: r-datatable.com
dt <- data.table(a = 1, b = 2)
dt[[2]] <- 2
set(dt, , "c", 1)
#> Error in set(dt, , "c", 1) : 
#>   It appears that at some earlier point, attributes of this data.table have been reassigned. Please use setattr(DT, name, value) rather than attr(DT, name) <- value. If that doesn't apply to you, please report your case to the data.table issue tracker.
truelength(dt)
#> [1] 1026
dt
#>        a     b
#>    <num> <num>
#> 1:     1     2
truelength(dt)  # this changed after printing for some reason
#> [1] 2
> set(dt, , "c", 1)
#> Error in set(dt, , "c", 1) : 
#>   Internal error in assign: input dt has not been allocated enough column slots. l=2, tl=2, adding 1. Please report to the data.table issues tracker.
```
Using using a column name instead of an index in `[[` also fails:
```r
dt <- data.table(a = 1, b = 2)
dt[["b"]] <- 2
set(dt, , "c", 1)
#> Error in set(dt, , "c", 1) : 
#>   It appears that at some earlier point, attributes of this data.table have been reassigned. Please use setattr(DT, name, value) rather than attr(DT, name) <- value. If that doesn't apply to you, please report your case to the data.table issue tracker.
```
Adding a new column, instead of changing an existing one, gives a different error message:
```r
dt <- data.table(a = 1, b = 2)
dt[[3]] <- 2
set(dt, , "c", 1)
#> Error in set(dt, , "c", 1) : 
#>   This data.table has either been loaded from disk (e.g. using readRDS()/load()) or constructed manually (e.g. using structure()). Please run setDT() or setalloccol() on it first (to pre-allocate space for new columns) before assigning by reference to it.
dt
#>        a     b    V3
#>    <num> <num> <num>
#> 1:     1     2     2
set(dt, , "c", 1)
#> Error in set(dt, , "c", 1) : 
#>   Internal error in assign: input dt has not been allocated enough column slots. l=3, tl=3, adding 1. Please report to the data.table issues tracker.
```

When using CRAN `data.table`, instead of the current GitHub version, the first two examples cause segfaults:
```r
library("data.table")
#> data.table 1.16.4 using 4 threads (see ?getDTthreads).  Latest news: r-datatable.com
dt <- data.table(a = 1, b = 2)
dt[[2]] <- 2
set(dt, , "c", 1)
#> 
#>  *** caught segfault ***
#> address 0x140000014, cause 'memory not mapped'
#> 
#> Traceback:
#>  1: set(dt, , "c", 1)
#> 
#> Possible actions:
#> 1: abort (with core dump, if enabled)
#> 2: normal R exit
#> 3: exit R without saving workspace
#> 4: exit R saving workspace
```

Maybe the operation corrupts memory in some way and that the new GitHub version just happens to not segfault because memory is laid out differently for some unrelated reason? Maybe there are two underlying bugs, one of which is already fixed on GitHub.

Using `$` does not seem to cause this issues:
```r
dt <- data.table(a = 1, b = 2)
dt$b <- 2
set(dt, , "c", 1)
dt
#>        a     b     c
#>    <num> <num> <num>
#> 1:     1     2     1
```

## `sessionInfo()`:
```
R version 4.4.0 (2024-04-24)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 20.04.6 LTS

Matrix products: default

locale:

time zone: Europe/Berlin
tzcode source: system (glibc)

attached base packages:

[1] compiler_4.4.0
```
(or GitHub `data.table` version in the first example)

ben-schwen:
Closes #6734

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -1411,7 +1411,7 @@ test(438.2, truelength(DT), tl+2L)
 # also allow DT[[...]]= syntax #6734
 DT[["baz"]] = 1L
 test(438.3, truelength(DT), tl+2L)
-test(483.4, DT$baz, c(1L,1L,1L))
+test(438.4, DT$baz, c(1L,1L,1L))
 DT[[5]] = 9L
 test(438.5, truelength(DT), tl+3L)
 DT[[1,"qux"]] = 10L


MichaelChirico:
Could you explain the relationship to the bug? The discussion there wound up pointing to a documentation-only fix.

inst/tests/tests.Rraw
```diff
@@ -1405,10 +1405,15 @@ test(437.2, truelength(DT), 2044L)
...
 tl = truelength(DT)
 DT$foo = 7L
-test(438, truelength(DT), tl)
+test(438.1, truelength(DT), tl)
 DT[,"bar"] = 8L
-test(439, truelength(DT), tl+2L)
-test(440, DT, data.table(a=1:3,b=4:6,foo=7L,bar=8L))
+test(438.2, truelength(DT), tl+2L)
+# also allow DT[[...]]= syntax #6734
+DT[["baz"]] = 9L
```
MichaelChirico: add a different test for a numeric index assignment?

R/as.data.table.R
```diff
@@ -130,6 +130,10 @@ as.data.table.list = function(x,
                 # e.g. cbind(foo=DF1, bar=DF2) have .named=c(TRUE,TRUE) due to the foo= and bar= and trigger "prefix." for non-vector items
   ...)
 {
+  if (is.data.table(x)) {
```
MichaelChirico: should this get a new direct test? `as.data.table.list()` is not invoked in what's added so far (except perhaps indirectly)

NAMESPACE
```diff
@@ -75,7 +75,7 @@ export(fctr)
 S3method("[", data.table)
 S3method("[<-", data.table)
 # S3method("[[", data.table)
-# S3method("[[<-", data.table)
+S3method("[[<-", data.table)
```
MichaelChirico, line 78: Originally removed here, it appears the reasoning is basically lost to history 682dd48b76af72a36a4a75024f1d2070f5a1b3e4

NEWS.md
```diff
@@ -346,6 +346,8 @@ See [#2611](https://github.com/Rdatatable/data.table/issues/2611) for details. T
 
 23. `fread()` auto-detects separators for single-column files consisting solely of quoted values (e.g. `"this_that"\n"2025-01-01 00:00:01"`), [#7366](https://github.com/Rdatatable/data.table/issues/7366). Thanks @arunsrinivasan for the report and @ben-schwen for the fix.
 
+24. Assigning via `[[<-` now changes data.tables by reference, so forms like `DT[["col"]] = value` or `DT[[1, "col"]] = value` keep over-allocation intact instead of falling back to the data.frame method, [#6734](https://github.com/Rdatatable/data.table/issues/6734). Thanks @mb706 for the report and @ben-schwen for the implementation.
```
MichaelChirico, line 349: Should we emphasize not doing this? 😄

ben-schwen:
> Could you explain the relationship to the bug? The discussion there wound up pointing to a documentation-only fix.

I saw the issue yesterday and it felt quite strange that the `$` assignment works, while `[[` does not, since I use these interchangeable (and even prefer `[[` for package since it has no partial matching)

NAMESPACE
ben-schwen: Matt added them in 773a7d516e8c277b5dc59ef2f1936d8760bd486e and they were removed because of speed issues in c39400bc56792e67c28f1be6285bdc83049d7486

Will we loose some speed? Most definitely, but in favor of user comfort. Interestlingly there is also a break even point in memory after adding about 4k columns.

subset


<details>

```r
library(atime)
df = data.frame(V1=1) 
dt = data.table(V1=1)

subset = atime(N=unique(as.integer(10^seq(0,5,l=20))), seconds.limit = 1,
    df = {for (i in seq(N)) df[[sprintf("V%d", i)]] <- 1L},
    dt = {for (i in seq(N)) dt[[sprintf("V%d", i)]] <- 1L} 
)
```

</details>

NEWS.md
ben-schwen: If we decide to do so than we should also advocate more on not using `$`, but I guess thats the charm of `data.table`. Your code, that you have written before simply works.

NAMESPACE
ben-schwen: We could also use `shallow` instead of `copy`, both in `$` and `[[` but thats opening a slippery slope of reference semantics I do not want to open for users.

R/as.data.table.R
ben-schwen: This is added because of test `2089.2` which fails without it.

inst/tests/tests.Rraw
ben-schwen: I added one!

NEWS.md
MichaelChirico: I guess the question is whether `cedta()` is sufficient to prevent cases where `[[<-` winds up being invoked by reference where the author intended it to use "normal" R copy-on-write behavior. `$<-.data.table` does `copy(x)`.

ben-schwen: True, but in any case `$<-` and `[[<-` shouldn't diverge in their behavior, right?

Issue #7288, opened 2025-09-07
tools and parallel should be added to suggests


jangorecki:
- `tools` used in `fread` and `frollapply`
- `parallel` used in `frollapply`

adding `parallel` to suggests causes codecov GH job to get stuck
therefore I think the proper path to close this issue is reproduce problem using minimal example using only `parallel` package and submit it to covr repository.

ANAMASGARD:
@jangorecki I've submitted PR #7403 with the proposed solution:

- Added `tools` and `parallel` to DESCRIPTION Suggests field
- Added 4 tests (2345.01-2345.04) to verify functionality
- Updated NEWS.md

All 12,072 local tests pass. The PR is ready for review. 

Regarding the codecov hang issue you mentioned - the CI will reveal if this occurs, and if so, we can create a minimal example for the covr repository as you suggested.

PR #7428, opened 2025-11-13
Interruption problems in `frollapply`

aitap:
There were two problems with the way `frollapply` handled interruptions:

1. `invokeRestart("abort")` in the interrupt handler prevented any further error or calling handlers from running (but not exit handlers like `tryCatch(finally=...)` or `on.exit()`; those kept working). I think that it's better to return from the calling handler, letting R continue the dispatch to other possible handlers:

```
tryCatch(
  frollapply(1:1e6, 1, \(.) { Sys.sleep(ret <- sum(.)); ret}),
  interrupt = \(e) 'interrupted'
)
^C[1] "interrupted"
```

2. While testing on an older laptop with an i5-2520M CPU, I noticed that interrupting parallel `frollapply` sometimes left zombie processes. The only way to reap them was to unload the `parallel` parallel (or quit R). I think that the parent process was too quick to call `mccollect()` after sending the `SIGINT`, so the child process did not have enough time to process the `SIGINT`, quit, and be ready to be collected by `waitpid()`. I _think_ that the zombie processes can be avoided by forcing `mccollect()` to wait, but that opens opportunities for further problems. `mccollect()`ing a terminated process results in a warning: do we need `suppressWarnings()`? If the child processes are really hung (e.g. due to the rolling function causing a deadlock), a second interrupt seems to interrupt `mccollect` despite the `suspendInterrupts()` (???) and unblock the parent process:

```
writeLines('void foo(void) { for(;;); }', 'foo.c')
tools::Rcmd('SHLIB foo.c')
foo <- mcparallel(.C('foo')) # child process is now hung
tryCatch(
 withCallingHandlers(
  Sys.sleep(10), # some interruptible process
  interrupt = \(e) suspendInterrupts( mccollect(foo) ) # handling the interrupt will hang
 ),
 interrupt = \(e) message('interrupted') # try to handle interrupting as well
)
^C^Cinterrupted # <-- unblocks after a second interrupt
tools::pskill(foo$pid) # stop the hung process
mccollect(foo) # clean up the zombie
```


ben-schwen:
Could the zombie processes also be the reason that the codecov github job gets stuck in #7288?

ben-schwen:
Apparently there are some other packages having the same problem

https://github.com/cran/bettermc/blob/34da1f6180067ae6970d42c3297442b5e0621784/R/mclapply.R#L723-L739

aitap:
Can't say no with 100% certainty, but I don't see how `R CMD check` could be producing them: the current issue is only for when the user interrupts R. So far, the difference between hang and no hang is [`covr:::save_trace` running as part of `parallel::mcexit`](https://github.com/r-lib/covr/blob/192ed5773652cb1ea15cee12056e251af785ce41/R/parallel.R#L20-L21), which looks [quite harmless](https://github.com/r-lib/covr/blob/192ed5773652cb1ea15cee12056e251af785ce41/R/covr.R#L108-L111).

jangorecki (approved):

ben-schwen (approved):
LGTM. The previous code with `wait=FALSE` looks kind of wrong since `timeout=0`.

Still would like a NEWS item, maybe as addendum to the whole froll* NEWS

jangorecki:
News should not be needed as this is change to code added in dev. It could eventually explain current interrupt behavior but I think better to have it in manual than news.

aitap:
Need to double check whether interrupting the first `mccollect` call may leave some PIDs already `wait`ed for and reused for a different process. We don't want to terminate an unrelated process because of that.

jangorecki:
@aitap

aitap:
It turns out this is possible:

```r
library(parallel)
jobs <- list(
 mcparallel(TRUE, 'completes immediately'),
 mcparallel(repeat Sys.sleep(1), 'hangs')
)
mccollect(jobs)
```
This hangs, so interrupt R. Since the first job is complete, its process has terminated, the parent process has `wait`ed for it, and its former PID may be reused:
```r
while (system(paste('test $$ -eq', jobs[[1]]$pid)) != 0) {} # the loop eventually completes!
```

This shows that `lapply(jobs, tools::pskill)` is unsafe because it may be directed at the wrong process. The race window starts when `mccollect` receives the result from the fast worker process.

In theory, we could use `parallel:::children(jobs)` to query `parallel`'s internal, accurate list of processes and filter only those whose PIDs haven't been reused yet, but this function is not exported.

Our salvation is the POSIX [`waitid`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/waitid.html) function, which lets us check whether a given PID is a child of the current process _without_ consuming the PID in the process. Any processes that still belong to us can then be terminated so that `mccollect` can `wait` for them:

```c
siginfo_t info;
bool can_terminate = waitid(P_PID, pid, &info, WCONTINUED | WEXITED | WNOHANG | WNOWAIT | WSTOPPED) == 0; // would be -1 if pid isn't ours
```

Now we can see:

```r
trace(tools::pskill, quote(message('terminating: ', toString(pid))), print = FALSE)
setDTthreads(4)
writeLines('foo', 'foo')
frollapply(1:1e4, 1, \(.) { if (suppressWarnings(file.remove('foo'))) Sys.sleep(10); 42 } ) -> val
# frollapply running on 4 CPU threads
# ^C
# terminating: 11111
# (followed by harmless warnings)
```

The PIDs previously belonging to the three processes that were fast to return the job result were not erroneously terminated.

Sorry this took so long.

HughParsonage (approved):
utils.c: LGTM 😄

Issue #6400, opened 2024-08-26
CRAN feedback on package checking time


TysonStanley:
During package submission process, Uwe said: 

> need to reduce the check time, this time we see `* checking tests ... [16m] OK` which is really too much as the overall check time of a package should be less than 10 min and this may be killed during the checks.

IIt's much shorter on my machine (macOS) and from what I could see, this seemed to be on Windows (from my testing of the package before submission using WinBuilder). Regardless, something we need to include in our next submission to CRAN (likely patch).

MichaelChirico:
Frustrating feedback. Checks run on windows GHA in 5 minutes. I don't understand why CRAN would encourage worse package testing like this -- especially from a package like {data.table} that has 1500 reverse dependencies / 3,000 recursive revdeps. This is how we wind up with packages like "IF (on CRAN) dont_test()" -- really bad for the ecosystem overall.

Does anybody even have a machine that can reproduce these terrible times?

TysonStanley:
Agreed, strange incentives. But no I can't replicate other than when I use WinBuilder (which I assume is what they use for CRAN checks too).

MichaelChirico:
> other than when I use WinBuilder

Well, that's something at least. Not the friendliest feedback mechanism, but there _is_ a mechanism...

ben-schwen:
What about deactivating only the most time consuming tests on CRAN?

MichaelChirico:
Once we open the pandora's box of "running a different suite of tests on CRAN", there's a lot of options open to us. That one probably makes the most sense, though I'm not sure how confident we are about which are the slowest tests on the CRAN machines.

eddelbuettel:
> Once we open the pandora's box of "running a different suite of tests on CRAN", there's a lot of options open to us. 

FWIW Rcpp has done that for maybe a decade and is still alive.  The package looks at the version, and when it sees a form 'a.b.c' runs reduced / quick tests -- which is what CRAN does -- whereas developers, GitHub Actions, r-universe, ... see a.b.c.d (which is usually set during development cycles) and run the full tests.  You could even only skip on Windows.  It's all programmable after all -- and CRAN has to look after 21k packages on (essentially) one hinking big machine each so you may not have the machine to yourself.

TysonStanley:
Quick glance at `Rcpp` @eddelbuettel, do you use the version number to define when tests would be reduced?

eddelbuettel:
[...]

PR #7429, opened 2025-11-15
deactivate froll tests

ben-schwen:
Towards #6400

Current `froll` checking times would not pass CRAN.

Change is structured in such a way that releases do not run `froll.Rraw`. One can still force the execution by setting `Sys.setenv("RunAllDataTableTests"="yes")`

What still needs to be discussed is whether we want to disable main test suite on CRAN...


tests/froll.R
```diff
@@ -1,2 +1,3 @@
 require(data.table)
-test.data.table(script="froll.Rraw")
+minor_version = as.integer(strsplit(packageDescription("data.table")$Version, "\\.")[[1]][[3]])
+if (minor_version == 99L || Sys.getenv("RunAllDataTableTests") == "yes") test.data.table(script="froll.Rraw")
```
MichaelChirico: This complicates the test script a fair amount -- why not just govern _only_ by `RunAllDataTableTests` and be sure to update GHA in this PR to set that as well?

If we want to keep this "always run in dev, and then optionally via env in CRAN release versions" setup, I'd prefer to have an internal helper to hide the logic and then just leave here:

```r
if (!data.table:::run_all_tests()) q('no')
test.data.table(script='froll.Rraw')
```

MichaelChirico: Alternatively, maybe this should just be another argument to `test.data.table()`:

```r
test.data.table(script='froll.Rraw', optional=TRUE)
```

jangorecki:
Why disable the whole test script. I think that all non-batch tests should finish under 10s and they are very useful, as show by recently detected breaking changes detected by revdeps check.

ben-schwen:
> Thank you for enabling tests back. As for env bars, we, and base R always uses capital case with underscores. Any good reason to use camel case instead?

Nah that was just shamelessly copied from Dirk.

Should we maybe run batch tests only in Gitlab and RMD-check-occasional?

jangorecki:
I would run them as well for interactive session. Maybe we can detect interactive inside `test.data.table()` and set env var there?
The purpose is to avoid it on CRAN only. We could avoid it as well in `--as-cran` CI jobs, so we have a workflow that can be checked to confirm if the mechanism works as expected.

ben-schwen:
> I would run them as well for interactive session. Maybe we can detect interactive inside `test.data.table()` and set env var there? The purpose is to avoid it on CRAN only.

We could but if you run them interactively you gonna type `test.data.table(script="frollBatch.Rraw")` anyway, hence leaving the optional to its default?

PR #7441, opened 2025-11-29
extra tests for no NSE in env arg

jangorecki:
couple extra tests to be even more future proof in case anyone in future will want to add NSE to env.

BTW. @MichaelChirico PR template message is perfect!

### DIFF
--- a/inst/tests/programming.Rraw
+++ b/inst/tests/programming.Rraw
@@ -616,3 +616,10 @@ test(202.2, f(dt, params=jpar), data.table(y=2:1))
 "." = function(...) list(.j=list("x"))
 test(202.3, dt[, .j, env=.(.j=list("y"))], data.table(x=1:2))
 rm(list=".")
+x = 2L ## no NSE, tests explicitly : ! -
+y = 2L
+test(202.4, dt[, .j, env = list(.j = x:y)], data.table(y=2:1))
+col = "x"
+test(202.5, dt[, .j, env = list(.j = !col)], error = "invalid argument type")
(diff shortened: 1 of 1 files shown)


sisyphuswastaken:
I was working on issue #7354 (cube support for patterns with .SDcols) and after studying the issue, my understanding is that simply using patterns("value") leads to an empty character vector being passed by default which then leads to the error that the programmer reported. When I studied melt.data.table's code, I noticed that eval_with_cols solves this problem. I had a few questions regarding this. Firstly, Is my understanding of the issue correct? 
I also noticed that this PR includes tests related to this area. Since the contributor guidelines state:

“Every new feature or bug fix must have one or more new tests in inst/tests/tests.Rraw”

I also wanted to ask whether additional tests are still expected from my side, or whether the tests in this PR already cover the needed cases, since to my understanding NSE handling would fix the issue/inconsistency in the cube() function.

Since I am new to contributing to open-source projects, I wanted to make sure I am approaching this correctly. Any guidance would be greatly appreciated.
Thank you!

jangorecki:
That's an amazing question — although does not seem to be related to this PR — tests included in this PR are not related to the PR you mentioned.

We are very happy your read contributing documentation and we appreciate if you will follow our guidances. Please submit your post in the relevant place.

If you wish I can tell you where it is, but my time is limited these days, so possibly ask your assistant instead.

sisyphuswastaken:
Thank you for the clarification! 
And yes, I will submit future posts in the relevant topic and follow the contribution guidelines closely. I appreciate your guidance :)

PR #7446, opened 2025-12-05
Avoid shadowing message in autoprint tests

aitap:
R-devel has introduced `` base::`%notin%` ``, which `data.table` now shadows, which shows up in the "golden" output test:

```
  Comparing ‘autoprint.Rout’ to ‘autoprint.Rout.save’ ...4,10d3
<
< Attaching package: 'data.table'
<
< The following object is masked from 'package:base':
<
<     %notin%
<
```

Use `suppressPackageStartupMessages(...)` in the test to avoid the message.

@TysonStanley, I think this is needed for cherry-picking to avoid an extra NOTE.

### DIFF
--- a/tests/autoprint.Rout.save
+++ b/tests/autoprint.Rout.save
@@ -15,7 +15,7 @@ Type 'demo()' for some demos, 'help()' for on-line help, or
 'help.start()' for an HTML browser interface to help.
 Type 'q()' to quit R.
 
-> # TODO(%notin%): re-enable warn.conflicts
+> # TODO(#7453): re-enable warn.conflicts
 > require(data.table, warn.conflicts=FALSE)
 > # Tests the suppression of := output
 > # Since this tests autoprinting at the console, it needs to use the .Rout.save mechanism in R CMD check


MichaelChirico:
is the definition identical? if so we should start a deprecation cycle too

aitap:
Ours might be faster when both arguments are character vectors; otherwise identical:

```
> `%notin%`
function (x, table)
match(x, table, nomatch = 0L) == 0L
<bytecode: 0x563bba77e7f8>
<environment: namespace:base>
> data.table::`%notin%`
function (x, table)
{
    if (is.character(x) && is.character(table)) {
        .Call(Cnotchin, x, table)
    }
    else {
        match(x, table, nomatch = 0L) == 0L
    }
}
<bytecode: 0x563bbab3aea0>
<environment: namespace:data.table>
```

jangorecki:
I would definitely aim for deprecation, I don't think differences will be that big

tests/autoprint.R
```diff
@@ -1,4 +1,4 @@
-require(data.table)
+suppressPackageStartupMessages(require(data.table))
```
MichaelChirico: For this PR I'd prefer `require(data.table, exclude="%notin%")`, however that's from 3.6.0 (r76248):

https://github.com/r-devel/r-svn/commit/c220e47d6012f220f6e7b7345ded4d7facfb9282

Maybe just `require(data.table, warn.conflicts=!exists("%notin%", "package:base"))` is the most concise way to narrowly skip for only this issue?

With blanket `suppress*` I worry about missing other such issues in the future.

aitap: That's very concise compared to dynamic call manipulation and testing for `'exclude' %in% names(formals(require)))`, thank you!

MichaelChirico: yea, still not ideal, as it means all future versions turn off conflict warnings. I guess it doesn't matter much for this test anyway -- we could just statically set it to FALSE?

aitap:
Three CRAN packages import `%notin%`: `bruceR`, `tidyfst`, and `weatherOz`; the first two for re-exporting. Any clever tricks (à la `cedta`?) we might use so that interactive users who would get `` base::`%notin%` `` anyway will not get penalised? I see that some packages (e.g. `backports`) export objects conditionally on R version. Should we try to eventually use `if (getRversion() < "4.6.0")) export("%notin%")` and similarly emit deprecation warnings only on R ≥ 4.6.0?

MichaelChirico:
An `if()` in the `NAMESPACE` might work... let's explore the full implications as a follow-up issue: #7453

tests/autoprint.R
```diff
@@ -1,4 +1,5 @@
-require(data.table)
+# TODO(%notin%): re-enable warn.conflicts
```
MichaelChirico:
```suggestion
# TODO(#7453): re-enable warn.conflicts
```

tests/autoprint.Rout.save
```diff
@@ -15,8 +15,8 @@ Type 'demo()' for some demos, 'help()' for on-line help, or
 'help.start()' for an HTML browser interface to help.
 Type 'q()' to quit R.
 
-> require(data.table)
-Loading required package: data.table
+> # TODO(%notin%): re-enable warn.conflicts
```
MichaelChirico:
```suggestion
> # TODO(#7453): re-enable warn.conflicts
```

MichaelChirico (approved):

PR #7450, opened 2025-12-06
Replace the use of `SET_OBJECT`

aitap:
Instead of manually transplanting `ATTRIB` and then restoring the object bits, use `SHALLOW_DUPLICATE_ATTRIB`, which takes care of the bits. (`ATTRIB` and `SET_ATTRIB` are non-API as well, although not on the radar yet.) Unlike `asS4` in some cases, this modifies the target object by reference, although the attributes do get (shallowly) duplicated.

@TysonStanley, if merged into `master`, this will be needed for cherry-picking into the patch branch.


ben-schwen:
Towards #6180 (so we can find it later)

PR #7459, opened 2025-12-10
Linguist source files

SebKrantz:
Now – just in case you want to accurately report source code – might turn this into a C-library again.


MichaelChirico:
Just linking the documentation on these terms:

https://github.com/github-linguist/linguist/blob/main/docs/overrides.md

.gitattributes
```diff
@@ -8,3 +8,16 @@
 #   want to preserve the line endings robustly
 inst/tests/** -text
 inst/tests/*.Rraw text eol=lf linguist-language=R
+# Ensure GitHub  linguist only considers source code files
+#   for language statistics
+docs/** linguist-documentation
+man/** linguist-documentation
+tests/** linguist-vendored
+vignettes/** linguist-vendored
```
MichaelChirico, line 16: what's the idea behind skipping `tests/**` and `vignettes/**`? why do we want to "only consider source code files"?

couldn't `vignettes/**` be `linguist-documentation`?

SebKrantz: Yes, vignettes could (and perhaps should) be `linguist-documentation`. The effect will be the same, however – it will be excluded from stats. I set up my repos so the stats only reflect the core library content – not code written to test or document it. I suggested it to you as I know {data.table} was once listed as a C library and I suspected that to still be the case if we ignore tests and documentation from the stats. Turns out I was wrong – the R code seems to have increased a lot.

ben-schwen: As Sebastian also explained it is to have "correct metrics" regarding lines of code in a certain language. While it does not affect the repo directly, I thought it can not hurt to display the "correct" numbers of GitHub

MichaelChirico: thanks! not particularly invested in getting it "correct", just wondering the thought process :)

PR #7467, opened 2025-12-13
Re-generate .pot with custom rules

MichaelChirico:
I should of course not have naively just run `potools::po_extract()`, I should have followed our script :)

https://github.com/Rdatatable/data.table/blob/79e4afbdcfab6a41bdaa1165dec0ab4ebde29f57/.dev/CRAN_Release.cmd#L6-L18

That's done now, which finds a fair number of new strings for translation.


ChristianWia:
no opinion.

PR #7468, opened 2025-12-13
Fix '\r' in translation

MichaelChirico:
One last round (hopefully) of corrections: `\r` can't be in a translation ID.


ChristianWia (commented):
no remark.

PR #7470, opened 2025-12-13
New French translations

MichaelChirico:
This new .po file is updated relative to the new .pot file. There are a number of new, fuzzy, and deprecated messages eligible for translation.

Ideally, this can happen over the weekend to fit in the 1.18.0 release (which must meet a CRAN deadline). Otherwise, new translations will have to be included in the next patch release.

Thank you @Rdatatable/french!!


po/R-fr.po
```diff
@@ -47,61 +48,64 @@ msgstr ""
...
+#: as.data.table.R:91
 #, c-format
 msgid ""
 "Argument 'value.name' must be scalar character, non-NA and at least one "
 "character"
 msgstr ""
-"L'argument 'value.name' doit être un caractère scalaire, pas NA et d'au moins "
-"un caractère"
+"L'argument 'value.name' doit être un caractère scalaire, pas NA et d'au "
+"moins un caractère"
```
tdhock, line 58: there are several false positive changes like this (same content but different line breaks).
can that be avoided some how?

aitap: Looks like `fr.po` and `R-fr.po` had been previously wrapped at 79 and 81 characters respectively, so

```sh
msgcat -w 79 -o po/fr.po po/fr.po
msgcat -w 81 -o po/R-fr.po po/R-fr.po
```

should at least reduce the number of changes due to wrapping.

ChristianWia: so what concretly set in POedit ?    I have :

>  EOL: unix recommended
>  wrap at 79 char
>  x preserve the format of existing files

emaigne: I tried to do it but went too fast and misread the comments, sorry. 
For R-fr.po you can set 81 characters in poedit, and uncheck "preserve".  Then when you save the file you should have the right number of chars by line.
When you edit fr.po you can set to 79. 
(I think at one point we should have the same settings for both files - for a later commit).

ChristianWia:
Detected 4 times Inconsistent source string to translate:   **by the your** function 

#: frollapply.R:322
msgid ""
"frollapply completed successfully but raised an error when attempting to "
"simplify results using user specified function in 'simplify' argument. Be "
"sure to provide 'fill' argument matching the type and shape of results "
"returned **by the your** function. Use simplify=FALSE to obtain a list instead.\n"
"%s"

#: frollapply.R:323
#: frollapply.R:325
#: frollapply.R:326

aitap:
There's also

> adaptive windows provided in n must not <del>to</del> have different lengths

https://github.com/Rdatatable/data.table/blob/e5e10a09b32f851465790cef98526ab63d5cee3a/R/froll.R#L64

ChristianWia:
Detected in fr.po  : a single \n in Poedit id 144 and requested for translation without text  
It was referenced in : 
fsort.c:264
fwrite.c:749
dogroups.c:456

MichaelChirico:
Thanks! I still see some untranslated messages:

```
Recompiling 'fr' src translation
Running system command msgfmt -c --statistics -o './inst/po/fr/LC_MESSAGES/data.table.mo' './po/fr.po'...
568 translated messages, 95 fuzzy translations, 40 untranslated messages.
```

Feel free to file a follow-up :)

ChristianWia:
it is normal, merge has been done but comment said translation was still in progress. Now i have completed fr.po (in the new tr-fr branch) and  both FR files can be sent for rereading. Hope they will be welcome.
(PR #7483)

PR #7473, opened 2025-12-13
New Russian translations

MichaelChirico:
This new .po file is updated relative to the new .pot file. There are a number of new, fuzzy, and deprecated messages eligible for translation.

Ideally, this can happen over the weekend to fit in the 1.18.0 release (which must meet a CRAN deadline). Otherwise, new translations will have to be included in the next patch release.

Thank you @Rdatatable/russian!!


MichaelChirico:
Thanks for compiling the .mo!

PR #7476, opened 2025-12-15
Improve translation-readiness of string with subclause

### LINKED ISSUE #7475: Unable to translate "in parallel"
https://github.com/Rdatatable/data.table/blob/4f8695fa560beaa7d7d8034ee41fa6146e2998b5/src/froll.c#L1627-L1628

The second placeholder, `%s`, can be replaced with "in parallel", but that text is not made available for translation.

Please consider duplicating the string with the alternatives with and without "in parallel" text.

MichaelChirico:
Closes #7475

@rffontenelle maybe this could be `%s` resolving to `in serial` or `in parallel`? I am unclear if it's as easy as swapping the word out in other languages (probably @aitap has the best guess :smile: )

### DIFF
--- a/inst/tests/froll.Rraw
+++ b/inst/tests/froll.Rraw
@@ -1529,7 +1529,7 @@ test(6004.013, frollmedian(1:8, 4), c(NA, NA, NA, 2.5, 3.5, 4.5, 5.5, 6.5), notO
 test(6004.014, frollmedian(c(NA,2:9), 3), as.double(c(NA,NA,NA,3:8)), notOutput="running implementation as described in the paper by Jukka Suomela, for uneven window size, length of input a multiple of window size, no NAs in the input data", output="NAs detected")
 test(6004.015, frollmedian(c(1,2,3,4,NA,6), 3), c(NA, NA, 2, 3, NA, NA), output="NAs detected, fall back to frollmedianExact\nfrollmedianExact: running in parallel for input length 6, window 3, hasnf 1, narm 0")
 test(6004.016, frollmedian(c(1,2,3,4,5,6), 3), c(NA, NA, 2, 3, 4, 5), output="sequentially as there is only single rolling computation.*finding order and initializing links for 2 blocks in parallel took")
-test(6004.017, frollmedian(c(1,2,3,4,5,6), c(3,3)), list(c(NA, NA, 2, 3, 4, 5), c(NA, NA, 2, 3, 4, 5)), output="in parallel.*finding order and initializing links for 2 blocks in serial took", notOutput="finding order and initializing links for 2 blocks in parallel took")
+test(6004.017, frollmedian(c(1,2,3,4,5,6), c(3,3)), list(c(NA, NA, 2, 3, 4, 5), c(NA, NA, 2, 3, 4, 5)), output="in parallel.*finding order and initializing links for 2 blocks sequentially took", notOutput="finding order and initializing links for 2 blocks in parallel took")
 options(datatable.verbose=FALSE)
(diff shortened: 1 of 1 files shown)


aitap:
This particular sentence might actually work even if translated from small parts. Since the `%s` part is in parentheses, we can pretend that it's a small sentence. If we really wanted to build sentences from small parts, we could try [contexts](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) so that two different occurrences of `in parallel` could be translated differently if needed, but duplication as currently in this PR is more in line with how gettext is intended to be used.

src/froll.c
```diff
@@ -1625,7 +1625,8 @@ void frollmedianFast(const double *x, uint64_t nx, ans_t *ans, int k, double fil
     setlinks(&o[j*k], &next[j*(k+1)], &prev[j*(k+1)], tail);
   }
   if (verbose)
-    snprintf(end(ans->message[0]), 500, _("%s: finding order and initializing links for %d blocks %stook %.3fs\n"), "frollmedianFast", b, par ? "in parallel " : "", omp_get_wtime()-tic);
+    snprintf(end(ans->message[0]), 500, par ? _("%s: finding order and initializing links for %d blocks in parallel took %.3fs\n")
+                                            : _("%s: finding order and initializing links for %d blocks in serial took %.3fs\n"), "frollmedianFast", b, omp_get_wtime()-tic);
```
ben-schwen:
```suggestion
                                            : _("%s: finding order and initializing links for %d blocks sequentially took %.3fs\n"), "frollmedianFast", b, omp_get_wtime()-tic);
```

inst/tests/froll.Rraw
```diff
@@ -1529,7 +1529,7 @@ test(6004.013, frollmedian(1:8, 4), c(NA, NA, NA, 2.5, 3.5, 4.5, 5.5, 6.5), notO
...
 test(6004.015, frollmedian(c(1,2,3,4,NA,6), 3), c(NA, NA, 2, 3, NA, NA), output="NAs detected, fall back to frollmedianExact\nfrollmedianExact: running in parallel for input length 6, window 3, hasnf 1, narm 0")
 test(6004.016, frollmedian(c(1,2,3,4,5,6), 3), c(NA, NA, 2, 3, 4, 5), output="sequentially as there is only single rolling computation.*finding order and initializing links for 2 blocks in parallel took")
-test(6004.017, frollmedian(c(1,2,3,4,5,6), c(3,3)), list(c(NA, NA, 2, 3, 4, 5), c(NA, NA, 2, 3, 4, 5)), output="in parallel.*finding order and initializing links for 2 blocks took", notOutput="finding order and initializing links for 2 blocks in parallel took")
+test(6004.017, frollmedian(c(1,2,3,4,5,6), c(3,3)), list(c(NA, NA, 2, 3, 4, 5), c(NA, NA, 2, 3, 4, 5)), output="in parallel.*finding order and initializing links for 2 blocks in serial took", notOutput="finding order and initializing links for 2 blocks in parallel took")
```
ben-schwen:
```suggestion
test(6004.017, frollmedian(c(1,2,3,4,5,6), c(3,3)), list(c(NA, NA, 2, 3, 4, 5), c(NA, NA, 2, 3, 4, 5)), output="in parallel.*finding order and initializing links for 2 blocks sequentially took", notOutput="finding order and initializing links for 2 blocks in parallel took")
```

MichaelChirico:
Ah, ideally this would have waiting until 1.18.0 was fully on CRAN 😬 

Now we have this string in .pot that's no longer in the package:

https://github.com/Rdatatable/data.table/blob/e455714d92b73f716913d3bdb225c90ac67e9d1c/po/data.table.pot#L2570-L2573

Issue #6188, opened 2024-06-18
Don't "double-print" in output= tests


MichaelChirico:
Whenever `output=` is used in `test()`, we force `print(x)`:

https://github.com/Rdatatable/data.table/blob/f5a1e09328fef5f8a4c5e225bc5b75ca09700f76/R/test.data.table.R#L425

We can consider not doing so if `x` already contains a call to `print()`, e.g. these 27 cases:

```r
library(lintr)
l <- make_linter_from_xpath(
  "//SYMBOL_FUNCTION_CALL[text()='test']
    /parent::expr
    /following-sibling::expr[expr/SYMBOL_FUNCTION_CALL[text()='print']]
    /following-sibling::SYMBOL_SUB[text()='output']",
  "xxx"
)
lint_dir("inst/tests", l(), pattern="Rraw") |>
  as.data.frame() |>
  _[, "line"] |>
  writeLines()
```

```
test(3, print(DT), output=c("   x             y", "1: 1 <ex_class[3]>",     "2: 2 <ex_class[3]>"))
test(779, print(DT[,sum(a),by=b]), output="   b V1\n1: a  3\n2: b  3$")
test(1006, print(as.data.table(M), nrows=10), output="gear NA.*1: 21.0")
    test(1037.103, print(ans[c(1,1326)]), output="state.*income.*retailprice.*percent_15_19.*beercons.*variable.*value.*1.*9.6.*89.34.*smoking88.*smoking00.*41.6")
  test(1037.301, print(melt(DT, id.vars="month", verbose=TRUE)), output="'measure.vars' is missing.*Assigned 'measure.vars' are [[]Record high, Average high, Daily mean, Average low, ...[]].*1:.*Jan.*Record high.*12.8.*108:.*Sep.*sunshine hours.*174.1")
test(1056, print(DT, digits=2), output="   x       y logy\n1: a      18  2.9\n2: a      79  4.4\n3: a     337  5.8")
test(1057, print(DT, digits=2, big.mark=","), output="   x         y logy\n1: a        18  2.9.*6: b    26,556 10.2\n7: c   113,811 11.6")
test(1636.2, print(dt), output="1: 1 <function[1]>")
test(1636.4, print(dt), output="2: 2 <function[1]>")
test(1636.6, print(dt), output="2: 2 <function[1]>")
test(1765.1, print(IDateTime(x)), output=".*idate.*itime.*1: 2017-03-1[67]",
[...]

PR #7478, opened 2025-12-15
More test fixes for when translations are present

aitap:
Switch all tests that depend on untranslated output being produced to use `output=` or `notOutput=` so that translation doesn't break them any more. Skip test 1590.14 because it fails due to the bytes having completely different meanings on Windows when the ANSI codepage is not 1252.

This will still fail in interesting ways (e.g. `translateChar()` warnings from the parser) on systems that cannot represent Latin-1 characters in the native encoding. We still need #7388 and a few more fixes for that.

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -11887,7 +11887,7 @@ test(1774.17, as.data.table(x, na.rm='a'), error="'na.rm' must be scalar")
 
 # verify print.keys works
 DT1 <- data.table(a = 1:3, key = "a")
-test(1775.1, DT1, options = c(datatable.print.keys = TRUE),
+test(1775.1, print(DT1, print.keys = TRUE),
      output = c("Key: <a>", "   a", "1: 1", "2: 2", "3: 3"))
 DT2 <- data.table(a = 1:3, b = 4:6)
 setindexv(DT2, c("b","a"))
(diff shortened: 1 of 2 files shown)


inst/tests/tests.Rraw
```diff
@@ -8539,7 +8541,7 @@ if (x1==x2) {
   # NB: x1==x2 is a condition in base R, independent of data.table
   test(1590.12, forderv(    c(x2,x1,x1,x2)), integer())
   # don't test base ... test(1590.13, base::order(c(x2,x1,x1,x2)), 1:4)
-} else {
+} else if (identical(l10n_info()$codepage, 1252L)) {
```
MichaelChirico, line 8544: Comment what 1252 is for the uninitiated future reader?

inst/tests/tests.Rraw
```diff
@@ -11653,8 +11653,8 @@ test(1765.4, {warning("foobar1"); warning("foobar2"); warning("FOO"); 4L}, 4L, i
 test(1765.5, {warning("foobar1"); warning("foobar2"); warning("FOO"); 4L}, 4L, ignore.warning="2", warning=c("foobar1","FOO"))
 
 # print(null.data.table()) should not output NULL as well, #1852
-# use capture.output() in this case rather than output= to ensure NULL is not output
-test(1766, capture.output(print(data.table(NULL))), "Null data.table (0 rows and 0 cols)")
+# the ^, $ specifiers don't match embedded newlines (if any) so match will fail if NULL is also printed
+test(1766, data.table(NULL), output = "^Null data\\.table \\(0 rows and 0 cols\\)$")
```
MichaelChirico: Why not `notOutput="NULL"`?

MichaelChirico: I guess this also serves as a test of the full output of `data.table(NULL)`, a janky mini golden test.

inst/tests/tests.Rraw
```diff
@@ -11888,8 +11888,8 @@ test(1774.17, as.data.table(x, na.rm='a'), error="'na.rm' must be scalar")
 
 # verify print.keys works
 DT1 <- data.table(a = 1:3, key = "a")
-test(1775.1, capture.output(print(DT1, print.keys = TRUE)),
-     c("Key: <a>", "   a", "1: 1", "2: 2", "3: 3"))
+test(1775.1, DT1, options = c(datatable.print.keys = TRUE),
```
MichaelChirico: I guess this changes the test slightly, I wonder if it's worth doing something like "check `substitute(x) %iscall% "print"` when `output=` is present, and if so, don't run a simple `print()`" to allow cases where we want to check `print()`'s own behavior directly.

MichaelChirico: Not the first time this has come up 😄 

#6188

aitap: It only broke 5 other tests, some of which I already had to touch in #6808, so why not.

PR #7480, opened 2025-12-16
only coerce UTF8 factors if needed in memrecycle

### LINKED ISSUE #7404: benchmark regression
I just had a look at the [ducklabs benchmarks](https://duckdblabs.github.io/db-benchmark/) and it seems that there has been regression for:

max v1 - min v2 by id3
`x[, .(range_v1_v2=max(v1, na.rm=TRUE)-min(v2, na.rm=TRUE)), by=id3]`

See the xlarge instance (with 256 GB memory) and 0.5 GB dataset where it takes ~57 seconds while running instant on the 32 GB memory instance (on a previos data.table version).

Would be nice to sort this out #7401 will make this case a lot faster.

Update: I have run this locally and #7401 takes down the regression from 24 seconds to 0.297 seconds :)

ben-schwen:
Closes #7404

Not sure about the test since we want to essentially test for `system.time` of the test. Maybe its better to use an atime test? The added tests bombs runners by taking 2 hours with regression :(

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -350,6 +350,8 @@ See [#2611](https://github.com/Rdatatable/data.table/issues/2611) for details. T
 
 25. By-group operations on missing rows (e.g. `foo[c(i, NA), bar, by=grp]`) now avoid leaving in data from the previous groups, [#7442](https://github.com/Rdatatable/data.table/issues/7442). Thanks @aitap for the report and the fix.
 
+26. Grouping by a factor with many groups is now fast again, fixing a timing regression introduced in [#6890](https://github.com/Rdatatable/data.table/pull/6890) where UTF-8 coercion and level remapping were performed unnecessarily, [#7404](https://github.com/Rdatatable/data.table/issues/7404). Thanks @ben-schwen for the report and fix.
+
 ### NOTES
 
 1. The following in-progress deprecations have proceeded:


src/assign.c
```diff
@@ -755,10 +755,14 @@ const char *memrecycle(const SEXP target, const SEXP where, const int start, con
...
       SEXP sourceLevels = source;  // character source
       if (sourceIsFactor) { sourceLevels=PROTECT(getAttrib(source, R_LevelsSymbol)); protecti++; }
-      sourceLevels = PROTECT(coerceUtf8IfNeeded(sourceLevels)); protecti++;
-      if (!sourceIsFactor || !R_compute_identical(sourceLevels, targetLevels, 0)) {  // !sourceIsFactor for test 2115.6
-        const int nTargetLevels=length(targetLevels), nSourceLevels=length(sourceLevels);
+      bool needUtf8Coerce = !sourceIsFactor || !R_compute_identical(sourceLevels, targetLevels, 0);
+      if (needUtf8Coerce) {
+        sourceLevels = PROTECT(coerceUtf8IfNeeded(sourceLevels)); protecti++;
         targetLevels = PROTECT(coerceUtf8IfNeeded(targetLevels)); protecti++;
+        if (sourceIsFactor && R_compute_identical(sourceLevels, targetLevels, 0)) needUtf8Coerce = false;
```
aitap: Is the `needUtf8Coerce = false` assignment covered? I've tried compiling with `-Og` and setting a breakpoint on the exact instruction setting the register to 0 and it didn't fire during `test.data.table()`. I think it might be unreachable.

The results of `R_compute_identical()` shouldn't change after `coerceUtf8IfNeeded()` because `identical()` takes encodings into account:

https://github.com/r-devel/r-svn/blob/96eee1cdda590de914d48fed05d8f0783f921da4/src/main/memory.c#L4978-L4997

This is quite convenient because a factor with levels `enc2utf8('ø')` and a factor with levels `iconv('ø', to = 'latin1')` will pass the first `R_compute_identical()` already, without any other string conversions.

aitap (approved):
Excellent diagnosis, thank you. Indeed, the costly calls to `need2utf8` can be gated by the levels being non-identical (which we need to test anyway).

I think this is more suitable for an atime test than a normal regression test.

src/assign.c
ben-schwen: Good point! I removed the unreachable code. Will add an atime test as a separate PR. I guess our current "issue" with atime tests and why we have to keep these branches alive, is that we squash when merging and hence the commits will disappear

@tdhock @Anirban166 does this sound right?

aitap:
A workaround discussed in a different issue is to first merge the PR and then record the SHA1-hash of the merge commit in a follow-up commit instead of the HEAD of the PR branch.

ben-schwen:
@TysonStanley this should probably be also picked for the release (since it is a regression fix for smth included in `1.17.2`)!

PR #7487, opened 2025-12-19
Replace `ATTRIB`, `SET_ATTRIB`

aitap:
Towards #6180.

Replace remaining uses of `SET_ATTRIB` with `SHALLOW_DUPLICATE_ATTRIB` or `CLEAR_ATTRIB`. Replace remaining uses of `ATTRIB` testing with `ANY_ATTRIB`; use the new `R_mapAttrib` interface to walk the attributes.

* The most invasive fix is in `assign.c` where `attr(., 'index')` is adjusted following a by-reference operation that could invalidate some indices. What used to be a `for (cons = ATTRIB(x); !isNull(cons); cons = CDR(cons))` loop now has to construct a context `struct` to share between the caller and the map callback. On the plus side, I simplified the memory management and string operations a bit, getting rid of things like `memset(string, '\0', 1)` and making use of the new hash table to look up the indices.
* The compact `rownames` attribute in `dogroups.c` is also a bit tricky. Now `dogroups` creates a throwaway `INTSXP` `c(NA_integer, -whatever)`, gives it to `setAttrib(SD, R_RowNamesSymbol, decoy)`, and obtains the actual compact `rownames` attribute that R re-creates from the return value. I think this is simpler than walking the attribute list by hand in search of the correct attribute, and the end result is the same: we can modify `rownames` by reference without spilling it into a large string vector.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,8 @@
 
 3. Vignettes are now built using `litedown` instead of `knitr`, [#6394](https://github.com/Rdatatable/data.table/issues/6394). Thanks @jangorecki for the suggestion and @ben-schwen and @aitap for the implementation.
 
+4. Removed use of non-API macros `ATTRIB`, `SET_ATTRIB`, [#6180](https://github.com/Rdatatable/data.table/issues/6180). Thanks @aitap for the continued assiduous work here.
+
 ### BUG FIXES
 
 1. `fread()` with `skip=0` and `(header=TRUE|FALSE)` no longer skips the first row when it has fewer fields than subsequent rows, [#7463](https://github.com/Rdatatable/data.table/issues/7463). Thanks @emayerhofer for the report and @ben-schwen for the fix.


src/assign.c
```diff
@@ -256,6 +256,99 @@ SEXP selfrefokwrapper(SEXP x, SEXP verbose) {
...
+  (void)val;
+  struct attrib_name_ctx *ctx = ctx_;
+  if (ctx->indexNames)
+    hash_set(ctx->indexNames, PRINTNAME(key), 1);
+  else
+    ctx->len++;
+  return NULL;
+}
+
+static SEXP fixIndexAttrib(SEXP tag, SEXP value, void *ctx_) {
```
HughParsonage, line 279: Is `SEXP` the intended return type?

src/assign.c
```diff
@@ -256,6 +256,99 @@ SEXP selfrefokwrapper(SEXP x, SEXP verbose) {
   return ScalarInteger(_selfrefok(x,FALSE,LOGICAL(verbose)[0]));
 }
 
+struct attrib_name_ctx {
+  hashtab *indexNames;
+  SEXP index, assignedNames;
+  R_xlen_t len;
```
HughParsonage: Can this field name be changed to make it clear what length it is the length of. `R_xlen_t len_of<...>;`

src/assign.c
```diff
@@ -256,6 +256,99 @@ SEXP selfrefokwrapper(SEXP x, SEXP verbose) {
...
+      hash_set(indexNames, PRINTNAME(tag), 0);
+      setAttrib(index, tag, R_NilValue);
+      hash_set(indexNames, s4Str, 1);
+      if (verbose)
+        Rprintf(_("Shortening index '%s' to '%s' due to an update on a key column\n"), c1+2, s4+2);
+    } else { // indexLength > 0 || shortened name present already
+      // indexLength > 0 indicates reordering. Drop it to avoid spurious reordering in non-indexed columns (#2372)
+      // shortened name already present indicates that index needs to be dropped to avoid duplicate indices.
+      setAttrib(index, tag, R_NilValue);
+      hash_set(indexNames, tag, 0);
```
HughParsonage: Should this be?

```c
hash_set(indexNames, PRINTNAME(tag), 0);
```

(tag is a SYMSXP, not a CHARSXP?)

src/assign.c
```diff
@@ -265,11 +358,11 @@ SEXP assign(SEXP dt, SEXP rows, SEXP cols, SEXP newcolnames, SEXP values)
   // cols : column names or numbers corresponding to the values to set
   // rows : row numbers to assign
   R_len_t numToDo, targetlen, vlen, oldncol, oldtncol, coln, protecti=0, newcolnum, indexLength;
```
HughParsonage: Is `indexLength` used in `assign` any more?

src/utils.c
```diff
@@ -672,3 +672,20 @@ void R_resizeVector_(SEXP x, R_xlen_t newlen) {
...
+
+#ifdef BACKPORT_MAP_ATTRIB
+SEXP R_mapAttrib_(SEXP x, SEXP (*fun)(SEXP key, SEXP val, void *ctx), void *ctx) {
+  PROTECT_INDEX i;
+  SEXP a = ATTRIB(x);
+  PROTECT_WITH_INDEX(a, &i);
+
+  SEXP ret = NULL;
+  for (; !isNull(a); REPROTECT(a = CDR(a), i)) {
+    ret = fun(TAG(a), CAR(a), ctx);
```
HughParsonage: Not blocking, but if time permits consider R's structure more closely with protection of the TAG(a) and CAR(a) in teh callback.

src/dogroups.c
```diff
@@ -130,12 +135,15 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
-  for (s = ATTRIB(SD); s != R_NilValue && TAG(s)!=R_RowNamesSymbol; s = CDR(s));  // getAttrib0 basically but that's hidden in attrib.c; #loop_counter_not_local_scope_ok
-  if (s==R_NilValue) error(_("row.names attribute of .SD not found"));
-  rownames = CAR(s);
-  if (!isInteger(rownames) || LENGTH(rownames)!=2 || INTEGER(rownames)[0]!=NA_INTEGER) error(_("row.names of .SD isn't integer length 2 with NA as first item; i.e., .set_row_names(). [%s %d %d]"),type2char(TYPEOF(rownames)),LENGTH(rownames),INTEGER(rownames)[0]);
+  SEXP rownames;
+  PROTECT_INDEX rownamesi;
+  PROTECT_WITH_INDEX(rownames = allocVector(INTSXP, 2), &rownamesi); nprotect++;
+  INTEGER(rownames)[0] = NA_INTEGER;
+  INTEGER(rownames)[1] = -maxGrpSize;
+  REPROTECT(rownames = setAttrib(SD, R_RowNamesSymbol, rownames), rownamesi);
```
HughParsonage: I think this is wrong; check the return value of setAttrib

HughParsonage: Consider R_mapAttrib for this.

src/assign.c
aitap: Yes, that's following the [`R_mapAttrib`](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Working-with-attributes-1) interface. I have to return `(SEXP)NULL` to continue iteration.

src/dogroups.c
aitap: R is somewhat inconsistent about this. `setAttrib(foo, R_CommentSymbol, bar)` and `setAttrib(foo, R_ClassSymbol, bar)` return `R_NilValue`, but generic `installAttrib` and our specific case `row_names_gets` do return the freshly installed value. Since it's not documented, `R_mapAttrib` might be a safer approach.

src/utils.c
aitap: AFAIK while `a` is protected, `TAG(a)`, `CAR(a)`, and `CDR(a)` should also stay protected. What do you recommend to do instead?

aitap:
This is expected to become more of a problem [around end of January](https://stat.ethz.ch/pipermail/r-devel/2025-December/084295.html).

src/utils.c
ben-schwen: any reason why we are not just backporting the implementation of R, since it seems to have the same level of complexity/speed

https://github.com/wch/r-source/blob/f21cd0f6e791dbe35292db51f96a25f46f306748/src/main/attrib.c#L2010-L2031

src/dogroups.c
```diff
@@ -129,12 +134,15 @@ SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEX
...
```
ben-schwen:
```r
static SEXP findRowNames(SEXP key, SEXP val, void *data) {
  (void)data;
  if (key == R_RowNamesSymbol) return val;
  return NULL;
}
```

```r
SEXP rownames = R_mapAttrib(SD, findRowNames, NULL);
  if (rownames == NULL) error(_("row.names attribute of .SD not found"));
  if (!isInteger(rownames) || LENGTH(rownames)!=2 || INTEGER(rownames)[0]!=NA_INTEGER) error(_("row.names of .SD isn't integer length 2 with NA as first item; i.e., .set_row_names(). [%s %d %d]"),type2char(TYPEOF(rownames)),LENGTH(rownames),INTEGER(rownames)[0]);
```

src/dogroups.c
```diff
@@ -53,20 +55,23 @@ static bool anySpecialStatic(SEXP x, hashtab * specials) {
...
+  (void)key;
+  return anySpecialStatic(val, specials) ? R_NilValue : NULL;
+}
+
 SEXP dogroups(SEXP dt, SEXP dtcols, SEXP groups, SEXP grpcols, SEXP jiscols, SEXP xjiscols, SEXP grporder, SEXP order, SEXP starts, SEXP lens, SEXP jexp, SEXP env, SEXP lhs, SEXP newnames, SEXP on, SEXP verboseArg, SEXP showProgressArg)
 {
   R_len_t ngrp, nrowgroups, njval=0, ngrpcols, ansloc=0, maxn, estn=-1, thisansloc, grpn, thislen, igrp;
   int nprotect=0;
-  SEXP ans=NULL, jval, thiscol, BY, N, I, GRP, iSD, xSD, rownames, s, RHS, target, source;
+  SEXP ans=NULL, jval, thiscol, BY, N, I, GRP, iSD, xSD, s, RHS, target, source;
```
ben-schwen, line 80:
```suggestion
  SEXP ans=NULL, jval, thiscol, BY, N, I, GRP, iSD, xSD, RHS, target, source;
```

src/dogroups.c
ben-schwen:
```suggestion
  SEXP list_el;
```

ben-schwen (commented):
Great work!

Minor nag that NEWS is missing.

Should we add an atime test like in #2389?

src/utils.c
aitap: You've been right to tell me to consider protecting `CAR(a)`. If the callback calls `setAttrib` and replaces the current attribute, `CAR(a)` becomes unprotected. `TAG(a)` is probably protected for symmetry, or to make the code future-proof against non-persistent symbols.

I'd like the function to be distinct from R's version because the [compatibility page](https://www.gnu.org/licenses/license-list.html#MPL-2.0) says that combining MPL (2.0) and GPL (>= 2) code results in a Larger Work subject to both licenses.

ben-schwen (approved):
LGTM, TY!
(Besides missing news ofc)

Issue #7488, opened 2025-12-19
serialized data.tables from before cannot be read out of the box


ben-schwen:
Since #7451 we see quite some revdep failures which are around serialization.

```r
tmp = tempfile(fileext = ".rds")
download.file("https://raw.githubusercontent.com/cran/NMdata/master/inst/examples/data/xgxr2.rds", destfile = tmp, mode = "wb")
pk = readRDS(tmp)
pk[, trtact := NULL]
# Error in `[.data.table`(pk, , `:=`(trtact, NULL)) : 
#   Internal error in R_resizeVector_: attempt to resize a non-resizable vector. Please report to the data.table issues tracker.

pk = readRDS(tmp)
alloc.col(pk)
pk[, trtact := NULL]
```

I have not dived deeper into the issue, but I guess that these `data.table`s simply have no growable bit set and are therefore not recognized are resizable.

aitap:
<del>Yes, the old (pre-#7451) `data.table`s don't have it set at all. The new ones unserialize as "growable but TRUELENGTH=0" and work fine. Checking to see if adding a check for the growable bit to `selfrefok` will suffice.</del>

Deleting columns used to be safe even if the `data.table` was not over-allocated, because `SETLENGTH` worked on all lists. `R_resizeVector` needs a resizable list now:

https://github.com/Rdatatable/data.table/blob/be67a4db81cf7d938a463ceff2b7d2bf20a3df0f/src/assign.c#L665-L666

Unserialized data.tables come out with TRUELENGTH=0; tables from the old version don't have the growable bit set; tables from the new version do.

This `setalloccol`:

https://github.com/Rdatatable/data.table/blob/f6bbd5b3073526e554ad07bc9b0a346d72ad4e3a/R/data.table.R#L1389-L1423

...probably needs to happen when removing columns as well.

Test:

```r
mtcars |> as.data.table() |> serialize(NULL) |> unserialize() |> _[, carb := NULL]
```

PR #7492, opened 2025-12-20
`[.data.table`: call `setalloccol` for non-`selfrefok` tables when removing columns

aitap:
Before #7451, removing a column was always possible, because using `SETLENGTH` to reduce the length of a `data.table` was mostly unchecked (and didn't cause problems in practice, modulo the allocated memory count sometimes being wrong), even if it was not created by `data.table` code in the current R session. `R_resizeVector` now requires its argument to be resizable, and freshly unserialized `data.table`s don't pass the mark: ones created by older versions of `data.table` lack the growable bit, and ones created after #7451 latest version still have a zero TRUELENGTH.

Solution:
 * call `selfrefok` even if not adding new columns
 * if `jval` indicates removing a column, call `setalloccol` to prevent the error

Could the test for `:=` removing a column be improved? (What about list columns containing `NULL`s?)

Fixes: #7488

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -1191,8 +1191,9 @@ replace_dot_alias = function(e) {
           if (any(m<1L | ncol(x)<m)) stopf("LHS of := appears to be column positions but are outside [1,ncol] range. New columns can only be added by name.")
         } else
           stopf("LHS of := isn't column names ('character') or positions ('integer' or 'numeric')")
+        ok<-selfrefok(x, verbose=FALSE)
         if (!anyNA(m)) {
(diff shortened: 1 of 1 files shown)


ben-schwen:
TYVM!

ben-schwen:
@TysonStanley if still possible, we should squeeze this one into release, since otherwise we will get caught by revdeps

TysonStanley:
Great! I'm using what is in `master` by tonight if that works for the team. @ben-schwen @aitap @jangorecki @MichaelChirico

Issue #7506, opened 2025-12-23
revdep `ctsem` failure in `rbindlist`


aitap:
Not to be confused with `VECTOR_ELT` failures that 1.18.0 does address.

R-4.2 (to avoid `VECTOR_ELT` out-of-bounds access), `data.table` 1.17.8:

```r
library(ctsem)
library(data.table)
debug(rbindlist)
ctPostPredPlots(ctstantestfit)
```
```
Browse[2]> str(l)
List of 2
 $ :Classes ‘data.table’ and 'data.frame':      12000 obs. of  4 variables:
  ..$ sample: int [1:12000] 1 1 1 1 1 1 1 1 1 1 ...
  ..$ row   : num [1:12000] 1 1 10 10 100 100 101 101 102 102 ...
  ..$ V1    : chr [1:12000] "Y1" "Y2" "Y1" "Y2" ...
  ..$ value : num [1:12000] 6.583 11.218 -1.171 1.837 0.485 ...
  ..- attr(*, ".internal.selfref")=<externalptr>
  ..- attr(*, "sorted")= chr [1:3] "sample" "row" "V1"
 $ :Classes ‘data.table’ and 'data.frame':      6000 obs. of  4 variables:
  ..$ sample: int [1:6000] 1 1 1 1 1 1 1 1 1 1 ...
  ..$ row   : num [1:6000] 1 2 3 4 5 6 7 8 9 10 ...
  ..$ value : num [1:6000] -3.71 -1.35 -4.87 -3.19 -3.54 ...
  ..$ V1    : chr [1:6000] "LogLik" "LogLik" "LogLik" "LogLik" ...
  ..- attr(*, ".internal.selfref")=<externalptr>
```
The variable in both parts of the list is always `V1`.

R-devel, `data.table` from `master`, _sometimes_:

```
Browse[1]> str(l)
List of 2
 $ :Classes ‘data.table’ and 'data.frame':      12000 obs. of  4 variables:
  ..$ V3    : chr [1:12000] "Y1" "Y1" "Y1" "Y1" ...
  ..$ row   : num [1:12000] 1 1 1 1 1 1 1 1 1 1 ...
  ..$ sample: int [1:12000] 1 10 11 12 13 14 15 16 17 18 ...
  ..$ value : num [1:12000] 6.58 4.3 11.22 10 7.1 ...
  ..- attr(*, ".internal.selfref")=<pointer: 0x555680c74180>
[...]

PR #7509, opened 2025-12-23
`as.data.table.array`: copy shared `dimnames` before reversing them 

aitap:
Fixes: #7506. Tested by installing HEAD of this PR and re-running `R CMD check ctsem_*.tar.gz`. It still fails tests on my machine, but due to unrelated problems in `parallelly` cluster startup, not in `data.table`. Previously, `setfrev()` could be called on either `dimnames(x)` or a list that shared `names(dimnames(x))`, mutating `x` as a result.


MichaelChirico (approved):
was there not a place to just change setfrev() to frev() to get the copy?

aitap:
Given that it took me three tries to get both tests to pass, maybe replacing `setfrev` with `frev` was a better idea.

PR #7515, opened 2025-12-24
Bump R version dependency to 3.5.0

MichaelChirico:
2018... a lifetime ago :)


jangorecki (approved):
Ubuntu LTS gets up to 15 years of support so there is a good reasons to hold to older R but I think we all want refcnt

ben-schwen:
NAG we should update the README.md for this

```md
the R dependency is as old as possible for as long as possible, dated April 2014, and we continuously test against that version; e.g. v1.11.0 released on 5 May 2018 bumped the dependency up from 5 year old R 3.0.0 to 4 year old R 3.1.0
```

Issue #7516, opened 2025-12-25
data.table 1.18.0 could not be installed on FreeBSD


jszhao:
I tried to updated data.table from 1.17.8 to 1.18.0 in R 4.5.2 installed on FreeBSD system. However, I get the following error:
```
gcc14 -I"/usr/local/lib/R/include" -DNDEBUG   -DLIBICONV_PLUG -I/usr/local/include -isystem /usr/local/include    -fpic  -O2 -pipe  -DLIBICONV_PLUG -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing   -c utils.c -o utils.o
utils.c: In function 'is_direct_child':
utils.c:692:3: error: unknown type name 'siginfo_t'
  692 |   siginfo_t info;
      |   ^~~~~~~~~
utils.c:694:39: error: passing argument 3 of 'waitid' from incompatible pointer type [-Wincompatible-pointer-types]
  694 |     pret[i] = waitid(P_PID, ppids[i], &info, WCONTINUED | WEXITED | WNOHANG | WNOWAIT | WSTOPPED) == 0;
      |                                       ^~~~~
      |                                       |
      |                                       int *
In file included from utils.c:2:
/usr/include/sys/wait.h:160:32: note: expected 'struct __siginfo *' but argument is of type 'int *'
  160 | int     waitid(idtype_t, id_t, struct __siginfo *, int);
      |                                ^~~~~~~~~~~~~~~~~~
gmake: *** [/usr/local/lib/R/etc/Makeconf:202: utils.o] Error 1
ERROR: compilation failed for package ‘data.table’
```

Here is the output of  `sessionInfo()`:
```
> sessionInfo()
R version 4.5.2 (2025-10-31)
Platform: amd64-portbld-freebsd15.0
Running under: FreeBSD epyc 15.0-RELEASE-p1 FreeBSD 15.0-RELEASE-p1 releng/15.0-n280999-7bceec30b351 GENERIC amd64

Matrix products: default
LAPACK: /usr/local/lib/R/lib/libRlapack.so;  LAPACK version 3.12.1

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: Asia/Shanghai
tzcode source: system (FreeBSD)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
[1] compiler_4.5.2
```

[...]

Issue #5374, opened 2022-04-26
warning when compiling fread on master on windows rtools42


tdhock:
```r
> remotes::install_github("Rdatatable/data.table@e9a323de01a17af70d5316016606fa8d35b25023")
...
gcc  -I"c:/PROGRA~1/R/R-42~1.0/include" -DNDEBUG     -I"c:/rtools42/x86_64-w64-mingw32.static.posix/include"  -fopenmp   -O2 -Wall  -std=gnu99 -mfpmath=sse -msse2 -mstackrealign  -c fread.c -o fread.o
In file included from fread.h:3,
                 from fread.c:27:
dt_stdio.h:25: warning: "__USE_MINGW_ANSI_STDIO" redefined
   25 |   #define __USE_MINGW_ANSI_STDIO 1
      | 
In file included from c:\rtools42\x86_64-w64-mingw32.static.posix\include\corecrt.h:10,
                 from c:\rtools42\x86_64-w64-mingw32.static.posix\include\crtdefs.h:10,
                 from c:\rtools42\x86_64-w64-mingw32.static.posix\include\time.h:9,
                 from c:\rtools42\x86_64-w64-mingw32.static.posix\include\sys\time.h:10,
                 from fread.c:8:
c:\rtools42\x86_64-w64-mingw32.static.posix\include\_mingw.h:435: note: this is the location of the previous definition
  435 | #define __USE_MINGW_ANSI_STDIO 0      /* was not defined so it should be 0 */
      | 
```

tdhock:
It is possible to see this warning if you look on the GLCI log https://gitlab.com/Rdatatable/data.table/-/jobs/2212133398#L169 but it would be easy to miss (build is still green/pass). @jangorecki maybe it is worth considering changing the GLCI logic to return a red/fail build when there are compilation warnings, so that they would be more easily identified and fixed?

jangorecki:
We already have this logic for the most strict job: https://github.com/Rdatatable/data.table/blob/e9a323de01a17af70d5316016606fa8d35b25023/.gitlab-ci.yml#L167
We could add it for windows as well. As long as there won't be many false positives then it should be ok.

jangorecki:
@tdhock by any chance if you are on windows, could you provide powershell equivalent of
```sh
(! grep "warning:" data.table.Rcheck/00install.out) 
```
?
[...]

PR #7517, opened 2025-12-25
`utils.c`: include `<signal.h>` for `siginfo_t`

aitap:
POSIX [says](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/signal.h.html#:~:text=The%20%3Csignal.h%3E%20header%20shall%20define%20the%20siginfo_t%20type%20as%20a%20structure):

> The `<signal.h>` header shall define the **siginfo\_t** type as a structure

So `<sys/wait.h>` is not enough to see the definition (not just a forward declaration) of `siginfo_t`.

Fixes: #7516

Tested manually on FreeBSD:

<details>

```
$ R CMD check data.table_1.18.99.tar.gz --no-manual

* using R version 4.5.1 (2025-06-13)                                                   
* using platform: amd64-portbld-freebsd15.0                                            
* R was compiled by                                                                                                                                                            
    FreeBSD clang version 19.1.7 (https://github.com/llvm/llvm-project.git llvmorg-19.1.7-0-gcd708029e0b2)
    GNU Fortran (FreeBSD Ports Collection) 13.3.0                                      
* running under: FreeBSD freebsd 15.0-RELEASE FreeBSD 15.0-RELEASE releng/15.0-n280995-7aedc8de6446 GENERIC amd64
* checking whether package ‘data.table’ can be installed ... OK                        
* used C compiler: ‘FreeBSD clang version 19.1.7 (https://github.com/llvm/llvm-project.git llvmorg-19.1.7-0-gcd708029e0b2)’                                                    
* checking compiled code ... NOTE                                                      
File ‘data.table/libs/data_table.so’:   
  Found non-API calls to R: ‘IS_GROWABLE’, ‘LEVELS’, ‘SETLENGTH’,
    ‘SET_GROWABLE_BIT’, ‘SET_TRUELENGTH’, ‘TRUELENGTH’, ‘XTRUELENGTH’
Status: 1 NOTE                                                                         
```

</details>

...and OpenBSD:

<details>

```
$ LANG=C.UTF-8 R CMD check data.table_1.18.99.tar.gz --no-manual

* R was compiled by                                                                    
    egcc (GCC) 8.4.0                                                                   
    GNU Fortran (GCC) 8.4.0                                                            
* running under: OpenBSD openbsd.my.domain 7.8 GENERIC#54 amd64                        

* checking whether package ‘data.table’ can be installed ... OK
* used C compiler: ‘OpenBSD clang version 19.1.7’

Status: OK
```

</details>

Also tested on GNU/Linux with `PKG_CFLAGS=--std=c99`.

@jszhao, could you please pull from this branch (`posix_siginfo_t`) and run `R CMD build`, then install the resulting tarball?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -173,8 +173,8 @@ test-lin-dev-gcc-strict-cran:
     _R_S3_METHOD_LOOKUP_BASEENV_AFTER_GLOBALENV_: "FALSE" ## detects S3 method lookup found on search path #4777
     _R_S3_METHOD_LOOKUP_REPORT_SEARCH_PATH_USES_: "TRUE"
   script:
-    - echo 'CFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -D_POSIX_C_SOURCE=200809L' > ~/.R/Makevars
-    - echo 'CXXFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -D_POSIX_C_SOURCE=200809L' >> ~/.R/Makevars
+    - echo 'CFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
+    - echo 'CXXFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' >> ~/.R/Makevars
     - *install-deps
     - R CMD check --as-cran $(ls -1t data.table_*.tar.gz | head -n 1)
     - (! grep "warning:" data.table.Rcheck/00install.out)


NEWS.md
```diff
@@ -4,6 +4,9 @@
 
 ## data.table [v1.18.99](https://github.com/Rdatatable/data.table/milestone/37?closed=1)  (in development)
 
+### BUG FIXES
+
+1. Fixed the compilation failure on FreeBSD due to an `#include` directive missing in v1.18.0, [#7516](https://github.com/rdatatable/data.table/issues/7516). Thanks to @jszhao for the report and @aitap for the fix.
```
MichaelChirico: i guess a better description is "failure in some strict environments, e.g. FreeBSD, due..." WDYT?

Maybe worth mentioning the POSIX compliance aspect too?

src/utils.c
```diff
@@ -1,6 +1,7 @@
 #ifndef _WIN32
 #  define _POSIX_C_SOURCE 200809L // required for POSIX (not standard C) features in is_direct_child e.g. 'siginfo_t'
-#  include <sys/wait.h>
```
HughParsonage, line 3:
```
# if !defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE < 200809L
#  undef _POSIX_C_SOURCE
#  define _POSIX_C_SOURCE 200809L
# endif
```

for robustness

jszhao:
> POSIX [says](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/signal.h.html#:~:text=The%20%3Csignal.h%3E%20header%20shall%20define%20the%20siginfo_t%20type%20as%20a%20structure):
> 
> > The `<signal.h>` header shall define the **siginfo_t** type as a structure
> 
> So `<sys/wait.h>` is not enough to see the definition (not just a forward declaration) of `siginfo_t`.
> 
> Fixes: #7516
> 
> Tested manually on FreeBSD:
> Details
> 
> ...and OpenBSD:
> Details
> 
> Also tested on GNU/Linux with `PKG_CFLAGS=--std=c99`.
> 
> @jszhao, could you please pull from this branch (`posix_siginfo_t`) and run `R CMD build`, then install the resulting tarball?

I have used `R CMD INSTALL ...` to install this branch (in fact, I am not familiar with github, so I just downloaded the branch with zip file, unzip, and tar it to tarball). It works. Thanks a lot for your effect to fix it for FreeBSD. Thanks again.

MichaelChirico:
cc @aitap for extra eyes on a5a264418920e5251d00d8ac8683f0efbdea2e5e -- should we add the same to the other jobs with `-D_FORTIFY_SOURCE=2`? just doing so in the `strict` job for now.

aitap:
Ironically, setting `-D_POSIX_C_SOURCE=200809L` makes [`gettimeofday`](https://github.com/freebsd/freebsd-src/blob/9c0bedbb2064e9e02e2c20d1aa909360924033d7/sys/sys/time.h#L631-L633) invisible in [FreeBSD 15](https://github.com/freebsd/freebsd-src/blob/8ac77ea6a3c6ce99d39e1dd4e19adb444b2b4d5a/sys/sys/_visible.h#L123-L165) (indeed, POSIX.1-2008 marks it as obsolete, recommending `clock_gettime()` instead):

```
user@freebsd:~ $ echo '#include <sys/time.h>' | cpp | grep gettimeofday                                   
int gettimeofday(struct timeval *, struct timezone *);
user@freebsd:~ $ echo '#include <sys/time.h>' | cpp -D_POSIX_C_SOURCE=200809L | grep gettimeofday || echo not found
not found
```

We use `gettimeofday` in `wallclock`. (The code that wants to use `clock_gettime()` deprives itself of the opportunity by testing for `#ifdef CLOCK_REALTIME` before a header file can `#define` it. A fix to that will need to be careful not to reintroduce #5374.)

Unlike [`_POSIX_C_SOURCE`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html#tag_16_02_01_01), which controls symbol visibility, [`_FORTIFY_SOURCE`](https://www.gnu.org/software/libc/manual/html_node/Source-Fortification.html) rewrites some function calls to perform stricter checks. Is the idea to make strict checks use `#define _POSIX_C_SOURCE 200809L` because it will hide non-standard or obsolete symbols?

MichaelChirico:
I'm just looking for a way to fortify our checks such that we would have caught this issue before release

aitap:
Then the strict checks must use `--std=c99` to hide even more symbols, so that `data.table` could request some of them back using `#define _POSIX_C_SOURCE=200809L`:
https://gitlab.com/Rdatatable/data.table/-/commit/f02dcb783903727cbc043620f764b9d7f25c7af9/pipelines?ref=stricter_c99

<del>(Right now the only failure is the `-vanilla` test, due to the `C` locale.)

Should I push f02dcb783903727cbc043620f764b9d7f25c7af9 to this branch?</del>

Drat. R itself uses configure-time tests for C23 features in `R_ext/Rboolean.h`, so when we later request C99 compatibility, this causes annoying warnings (because the configure-time tests aren't updated).

ben-schwen:
Since we need to patch `1.18.2` soon, I would be in favor to merge this and check afterwards if new errors pop up (I know this is not ideal)

.gitlab-ci.yml
```diff
@@ -173,8 +173,8 @@ test-lin-dev-gcc-strict-cran:
     _R_S3_METHOD_LOOKUP_BASEENV_AFTER_GLOBALENV_: "FALSE" ## detects S3 method lookup found on search path #4777
     _R_S3_METHOD_LOOKUP_REPORT_SEARCH_PATH_USES_: "TRUE"
   script:
-    - echo 'CFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
-    - echo 'CXXFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' >> ~/.R/Makevars
+    - echo 'CFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -D_POSIX_C_SOURCE=200809L' > ~/.R/Makevars
+    - echo 'CXXFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -D_POSIX_C_SOURCE=200809L' >> ~/.R/Makevars
```
MichaelChirico:
```suggestion
    - echo 'CFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' > ~/.R/Makevars
    - echo 'CXXFLAGS=-g -O2 -flto=auto -fno-common -fopenmp -Wall -Wvla -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2' >> ~/.R/Makevars
```

MichaelChirico:
Also `cherry-pick`ed to the 1.18.2 patch: 20e25d1e9701125d9bb889b9951b655dcd0c419a

NEWS.md
```diff
@@ -32,6 +32,8 @@
 
 6. `rowwiseDT()` now provides a helpful error message when a complex object that is not a list (e.g., a function) is provided as a cell value, instructing the user to wrap it in `list()`, [#7219](https://github.com/Rdatatable/data.table/issues/7219). Thanks @kylebutts for the report and @venom1204 for the fix.
 
+7. Fixed compilation failure like "error: unknown type name 'siginfo_t'" in v1.18.0 in some strict environments, e.g., FreeBSD, where the header file declaring the POSIX function `waitid` does not transitively include the header file defining the `siginfo_t` type, [#7516](https://github.com/rdatatable/data.table/issues/7516). Thanks to @jszhao for the report and @aitap for the fix.
```
aitap, line 35: Very minor: the underscore in `"error: unknown type name 'siginfo_t'"` is an unpaired emphasis mark. This doesn't confuse Pandoc or commonmark, but makes the Vim syntax highlighter unhappy. Some variation on `"error: unknown type name 'siginfo\_t'"` or `` "error: unknown type name `siginfo_t`" `` would be slightly better.

MichaelChirico: I would prefer `` `error: unknown type name 'siginfo_t'` `` to keep it human-readable too

PR #7518, opened 2025-12-25
Fix code blocks in `NEWS.md`

### LINKED ISSUE #7511: NEWS not rendering well on CRAN
Here's what I'm seeing:

Image

aitap:
Closes #7511 

Add the missing triple-backtick separator.

Separate the indented code blocks from the preceding paragraph because otherwise Pandoc fails to realise that the whitespace followed by triple-backtick denotes a fenced code block.

Before:

<details>

before
before, no. 18

</details>

After:

<details>

after
after, no.18

</details>

Any other way we could test it?


jangorecki (approved):

MichaelChirico:
Looks great, thanks for investigating. I guess CRAN rendering is going through commonmark, and we could check that somehow... not sure it's worth spending any time though.

PR #7528, opened 2025-12-26
Drop gettext and use stopf directly

MichaelChirico:
Minor change for consistency.

This could have been part of #7260 -- there, `gettext()` was used since the message would be passed to either `stop()` or `warning()`; now, it's always passed to `stop()`, so we don't need to separate into two steps.


ben-schwen:
TY!

PR #7532, opened 2025-12-26
Refine mention of ancient version in README

MichaelChirico:
Should be more future-proof this way. When updating, I do regex searches, e.g. for `3[._][45]` in the recent upgrade, which would now catch this.


ben-schwen:
TY!

PR #7534, opened 2025-12-26
Use litedown instead of pkgdown

### LINKED ISSUE #6584: Use litedown instead of pkgdown
IMO we could as well move away from pkgdown to new https://github.com/yihui/litedown/issues/24#issuecomment-2384933793
Pkgdown was causing extra burden with dependencies (new and many OS deps) and was at some point cached in a docker image, so it is not even updated in our CI for a longer while now.

_Originally posted by @jangorecki in https://github.com/Rdatatable/data.table/issues/6394#issuecomment-2385617638_

ben-schwen:
Closes #6584 

Much appreciated if someone with more web dev skills than me can fix .css styling :D

image

Whats still missing to be fully equivalent to pkgdown
- [x] favicon
- [x] TOC in NEWS

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -354,7 +354,7 @@ test-mac-old:
 # generating litedown website
 integration:
   stage: integration
-  image: registry.gitlab.com/rdatatable/dockerfiles/r-litedown
+  image: registry.gitlab.com/rdatatable/dockerfiles/r-data.table
   tags:
     - saas-linux-medium-amd64
   rules:


ben-schwen:
We need to finish either this PR or #7402 since otherwise our `dev` version is not building in CI! cc @Rdatatable/committers

jangorecki:
We don't need docker image for litedown, it was needed for pkgdown because of maaany OS deps. For litedown we can use one of our other docker images.

site/generate_reference.R
```diff
@@ -0,0 +1,97 @@
...
+    'h2 { margin-top: 1.5em; border-bottom: 1px solid #dee2e6; padding-bottom: 0.3em; }',
+    'table { border-collapse: collapse; }',
+    'table td, table th { border: 1px solid #dee2e6; padding: 0.5rem; }',
+    '</style>',
+    '<link rel="icon" type="image/png" sizes="16x16" href="../assets/favicon/favicon-16x16.png"><link rel="icon" type="image/png" sizes="32x32" href="../assets/favicon/favicon-32x32.png"><link rel="apple-touch-icon" sizes="60x60" href="../assets/favicon/apple-touch-icon-60x60.png"><link rel="apple-touch-icon" sizes="76x76" href="../assets/favicon/apple-touch-icon-76x76.png"><link rel="apple-touch-icon" sizes="120x120" href="../assets/favicon/apple-touch-icon-120x120.png"><link rel="apple-touch-icon" sizes="152x152" href="../assets/favicon/apple-touch-icon-152x152.png"><link rel="apple-touch-icon" sizes="180x180" href="../assets/favicon/apple-touch-icon-180x180.png"><link rel="apple-touch-icon" sizes="180x180" href="../assets/favicon/apple-touch-icon.png"><link rel="icon" sizes="any" href="../assets/favicon/favicon.ico">',
+    '</head>',
+    '<body>',
+    navbar_html,
+    '<div class="body">',
+    html_content,
```
aitap: It would be great to find a way to avoid starting the HTML document from scratch here:

<blockquote>

```html
<div class="body">
<!-- DOCTYPE and <html>, <head> below are invalid -->
<!DOCTYPE html><html><head><title>R: Integer based date class</title>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=yes">
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/katex.min.css">
<script>
const macros = { "\\R": "\\textsf{R}", "\\mbox": "\\text", "\\code": "\\texttt"};
function processMathHTML() {
    var l = document.getElementsByClassName('reqn');
    for (let e of l) { katex.render(e.textContent, e, { throwOnError: false, macros }); }
    return;
}</script>
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/katex.min.js"
    onload="processMathHTML();"></script>
<link rel="stylesheet" type="text/css" href="R.css">
<!-- starting a new <body> inside a <body> is invalid -->
</head><body><div class="container"><main>
```

</blockquote>

`litedown::pkg_manual()` does it using scary regexp hacks.

aitap:
Somehow, in our case, the [indentation added by `litedown:::build_toc`](https://github.com/yihui/litedown/blob/118f89775aa046d695e187b04e78f09383903ec1/R/utils.R#L616-L619) is misinterpreted by `commonmark::markdown_html` as an indented code block, so `fuse()` outputs `<code><pre>- &lt;nav&gt;...</pre></code>` instead of the table of contents. Are we holding it wrong?

<ins>**Edit**</ins>: that's easily fixed by making the first heading of `NEWS.md` a level-1 heading instead of level-2. Now the problem is that with `css: ['@default', '@article']`, `js: ['@sidenotes']` the left-side TOC easily floats outside the window border because the main content is too wide.

NEWS.md
```diff
@@ -1,4 +1,4 @@
-## data.table news and updates
+# data.table news and updates
```
jangorecki, line 1: I think there was a reason for that, @MichaelChirico any idea?

aitap:
On Sat, 17 Jan 2026 23:46:49 -0800
Jan Gorecki wrote:

> I think there was a reason for that, @MichaelChirico any idea?

The use of <h2> instead of <h1> in the first heading of the NEWS was
required to work around https://github.com/r-lib/pkgdown/issues/2897.

ben-schwen:
From my side we are good to go (I have just added `xml2` to our litedown image). Thanks Ivan for pushing this over the edge!

.gitlab-ci.yml
```diff
@@ -436,8 +448,14 @@ integration:
     - Rscript -e 'check.index("data.table", names(test.jobs))'
     ## web/checks/check_flavors.html
     - Rscript -e 'check.flavors(names(test.jobs))'
-    ## pkgdown merge
-    - Rscript -e 'common_files<-function(path1, path2) intersect(list.files(path1, all.files=TRUE, no..=TRUE), list.files(path2, all.files=TRUE, no..=TRUE)); msg = if (length(f<-common_files("website","bus/integration/cran"))) paste(c("Following artifacts will be overwritten by pkgdown artifacts:", paste0("  ", f)), collapse="\n") else "No overlapping files from pkgdown artifacts"; message(msg); q("no")'
+    ## Add backwards compatibility redirects
+    - mkdir -p website/news website/reference website/articles
```
aitap, line 452: If I download and unpack the artifacts from [GitLab CI](https://gitlab.com/Rdatatable/data.table/-/jobs/12804204298/artifacts/browse/bus/integration/cran/), the links from the navbar to the vignettes seem to point to files that don't exist, and there's a broken symlink `articles/doc` → `library/data.table/doc`.

Looks like the intent is to make `articles` itself a symlink to `library/data.table/doc`, so:

```suggestion
    - mkdir -p website/news website/reference
```

I'm afraid that now we also need some manual scripts to render the translated vignettes.

ben-schwen:
[Gitlab CI Integration](https://gitlab.com/Rdatatable/data.table/-/jobs/12818460746) shows that it is basically working, but not fully finished yet

aitap:
Hopefully, [this pipeline](https://gitlab.com/Rdatatable/data.table/-/pipelines/2283353634) will show both vignettes in `/articles` and images in translated vignettes working well. The Spanish vignette needs a two-line change in relative file paths, currently on GitLab but not here. (I also accidentally committed and reverted it on `master`.)

What else needs fixing?

ben-schwen:
I have just downloaded the archive and it looks good locally. So lets push and see what happens

ben-schwen:
I just checked on my mobile and the result looks quite good on mobile!

Breeze-Hu:
The display results look quite nice indeed! 👍

Compared to the previous version, the functions in the example section aren't outputting their execution results. Is this intentional?

jangorecki:
That's a question to @yihui probably

yihui:
`litedown::pkg_manual()` does execute examples, but I think in this PR, the help pages are not generated that way (see `site/manual.Rmd` and `site/generate_reference.R`).

ben-schwen:
Yeah we manually generated all the sites because `litedown::pkg_manual()` creates a single site with all functions.

aitap:
It looks like the `datatable-fread-and-fwrite` vignette is missing from the navbar. We could generate it before running `fuse_site`, but most of its links are actually static.

yihui:
> Yeah we manually generated all the sites because `litedown::pkg_manual()` creates a single site with all functions.

I think I can provide an option to build separate pages for each topic. Alternatively, you may generate the single page first, and then split it into separate pages. You don't really need to write that single page, but just take the results from `litedown::pkg_manual()` (it's a character string), and process the string.

In the meantime, if you want to run examples from the HTML pages by yourself, you can probably borrow my code in `litedown::pkg_manual()`: https://github.com/yihui/litedown/blob/c194e505f913ab6e66f68c57f1e7403c1a4adb37/R/package.R#L413-L456

aitap:
The link from https://rdatatable.gitlab.io/data.table/GOVERNANCE.html to https://rdatatable.gitlab.io/data.table/.github/CODE_OF_CONDUCT.md currently gives a 404 error.

yihui:
I think replacing the relative URL with the full one (https://github.com/Rdatatable/data.table/blob/master/.github/CODE_OF_CONDUCT.md) here https://github.com/Rdatatable/data.table/blob/67e78404b827359bd3d6db869351eb9ba1837bf0/GOVERNANCE.md?plain=1#L124
should fix it.

PR #7535, opened 2025-12-27
Evict STOP from parallel region, ? fix #7519

### LINKED ISSUE #7519: segfault under MKL, ATLAS configurations on CRAN
Observed from CRAN:

https://www.stats.ox.ac.uk/pub/bdr/Rblas/ATLAS/data.table.out
https://www.stats.ox.ac.uk/pub/bdr/Rblas/MKL/data.table.out

Both have similar stack traces:

```
  Traceback:
   1: all.equal(ax, ay, ..., check.attributes = check.attributes)
   2: attr.all.equal(target, current, tolerance = tolerance, scale = scale,     ...)
   3: all.equal.numeric(x, y, check.names = !isTRUE(y))
   4: all.equal(x, y, check.names = !isTRUE(y))
   5: isTRUE(all.equal.result <- all.equal(x, y, check.names = !isTRUE(y)))
   6: test(1252 + test_no * 0.001, ans1[ans1 != 0], with(DT, eval(ll)))
```

ATLAS and MKL are different BLAS implementations:

https://www.stats.ox.ac.uk/pub/bdr/Rblas/README.txt

HughParsonage:
Removes the `STOP` from the parallel region, which I suspect (though have not proven) is the cause of the segfault on CRAN for v1.18.0.  Even if it not ultimately the cause, causing `error` inside such a region is asking for trouble. 

Existing tests trigger the segfault.

### DIFF
--- a/src/forder.c
+++ b/src/forder.c
@@ -50,11 +50,19 @@ static int ustr_alloc = 0;
 static int ustr_n = 0;
 static int ustr_maxlen = 0;
 static int sortType = 0;             // 0 just group; -1 descending, +1 ascending
-static int nalast = 0;               // 1 (true i.e. last), 0 (false i.e. first), -1 (na i.e. remove)
+
 static int nradix = 0;
 static uint8_t **key = NULL;
 static int *anso = NULL;
 static bool notFirst=false;
+
+static int lgl_arg_to_int(int lgl_arg) {
+  if (lgl_arg == NA_LOGICAL) {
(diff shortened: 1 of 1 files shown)


HughParsonage:
Noticed some other spots where STOP is called

ben-schwen:
As pointed out by Ivan, this might not be a problem of us, but of `base` (where the original code was donated by data.table QQ)

PR #7537, opened 2025-12-28
nafill logical vector support

MichaelChirico:
Part of #3992.

Low-hanging fruit for `logical` which can basically be treated as `integer`.


src/nafill.c
```diff
@@ -113,7 +113,7 @@ SEXP nafillR(SEXP obj, SEXP type, SEXP fill, SEXP nan_is_na_arg, SEXP inplace, S
   if (obj_scalar) {
     if (binplace)
       error(_("'x' argument is atomic vector, in-place update is supported only for list/data.table"));
-    else if (!isReal(obj) && !isInteger(obj))
+    else if (!isReal(obj) && !isInteger(obj) && !isLogical(obj))
       error(_("'x' argument must be numeric type, or list/data.table of numeric types"));
```
ben-schwen: should we mention logical here?

src/nafill.c
```diff
@@ -124,7 +124,7 @@ SEXP nafillR(SEXP obj, SEXP type, SEXP fill, SEXP nan_is_na_arg, SEXP inplace, S
   int *icols = INTEGER(ricols);
   for (int i=0; i<length(ricols); i++) {
     SEXP this_col = VECTOR_ELT(obj, icols[i]-1);
-    if (!isReal(this_col) && !isInteger(this_col))
+    if (!isReal(this_col) && !isInteger(this_col) && !isLogical(this_col))
       error(_("'x' argument must be numeric type, or list/data.table of numeric types"));
```
ben-schwen: same as above

src/nafill.c
MichaelChirico: True, I'd missed all documentation. PTAL

PR #7539, opened 2025-12-28
Open up factor support in nafill

MichaelChirico:
Another easy win for #3992.

The only difficulty is ensuring that if `fill` is set to something outside `levels(x)`, the output is a well-formed factor. `coerceAs()` already has some logic to that effect, we've just got to take care & actually use it.


MichaelChirico:
Strange that `coerceAs()` is already meant to handle this.

```r
coerceAs("d", factor(letters[1:3]))
# [1] d
# Levels: a b c d
```

`coerceAs()` correctly generates the levels, but then `nafill()` ignores the updated levels.

src/nafill.c
```diff
@@ -218,8 +218,17 @@ SEXP nafillR(SEXP obj, SEXP type, SEXP fill, SEXP nan_is_na_arg, SEXP inplace, S
 
   if (!binplace) {
     for (R_len_t i=0; i<nx; i++) {
-      if (!isNull(ATTRIB(VECTOR_ELT(x, i))))
-        copyMostAttrib(VECTOR_ELT(x, i), VECTOR_ELT(ans, i));
+      SEXP xi = VECTOR_ELT(x, i);
+      if (!isNull(ATTRIB(xi))) {
```
ben-schwen:
```suggestion
      if (ANY_ATTRIB(ATTRIB(xi))) {
```
probably should go after #7487 for the backport of R < `4.5.0`

MichaelChirico: Great! Done

ben-schwen:
TY!

PR #7540, opened 2025-12-28
Implementing NSE in cube

sisyphuswastaken:
Pertains to issue #7354.

I noticed that cube() can lead to errors when .SDcols is supplied as an evaluative or callable expression (for example - helpers like patterns() or measure()), since it may be forwarded unevaluated and used before being resolved.

To address this, I updated cube() to handle .SDcols using non-standard evaluation in the following way: 
the argument is captured with substitute(), if it is callable then it is resolved via eval_with_cols() prior to being passed to groupingsets().

While testing this change, I encountered cases where applying numeric aggregations through j resulted in errors when non-numeric columns were present in .SD. To address this, j is captured as an expression to check if it references .SD, if it does then .SD is restricted to numeric values thereby avoiding unintended application of numeric aggregations to non-numeric data.

During development, I’m seeing three existing test failures (1378.2, 1378.3, 1750.34). I haven’t added new unit tests yet. I’m happy to add targeted tests if there are specific cases that need to be covered, and would appreciate guidance on whether additional test coverage is expected here. I’d also welcome any help on how best to approach the currently failing tests.


sisyphuswastaken:
closing PR as there are still some bugs to be resolved.

PR #7542, opened 2025-12-28
add codetools for vignettes

ben-schwen:
As shown by [CI](https://gitlab.com/Rdatatable/data.table/-/jobs/12553219137) our images need `codetools` to build vignettes. Successful run with `codetools` is [here](https://gitlab.com/Rdatatable/data.table/-/jobs/12554716051)

We could also add it to the docker images but it feels more right to do it here.


jangorecki:
Is the change of gitlab.io urls to lowercase required? Why?

ben-schwen:
> Is the change of gitlab.io urls to lowercase required? Why?

pkgdown gave me a warning about one of them and right now we use both, so a bit of consistency seemed right.

NEWS.0.md
```diff
@@ -295,7 +295,7 @@
 
   16. `merge.data.table` will raise warning if any of data.tables to join has 0 columns. Closes [#597](https://github.com/Rdatatable/data.table/issues/597).
 
-  17. Travis-CI will now automatically deploy package to drat repository hosted on [data.table@gh-pages](https://github.com/Rdatatable/data.table/tree/gh-pages) branch allowing to install latest devel from **source** via `install.packages("data.table", repos = "https://Rdatatable.github.io/data.table", type = "source")`. Closes [#1505](https://github.com/Rdatatable/data.table/issues/1505).
+  17. Travis-CI will now automatically deploy package to drat repository hosted on [data.table@gh-pages](https://github.com/Rdatatable/data.table/tree/gh-pages) branch allowing to install latest devel from **source** via `install.packages("data.table", repos = "https://rdatatable.github.io/data.table", type = "source")`. Closes [#1505](https://github.com/Rdatatable/data.table/issues/1505).
```
MichaelChirico: this one is github; we still use `R`datatable for github in a few other places.

(generally I might put this change separately, AFAICT there is only the tiny `, codetools` diff that matters in this PR... might be a good place for `jj` stacking)

ben-schwen: Good point. Filed #7564 as follow up

aitap:
The dependency chain is `litedown::fuse` → `xfun::record` → `xfun::find_locals` -> `codetools::findLocalsList`. Arguably, `litedown` could depend on `codetools`, but this gets the job done as well.

PR #7552, opened 2025-12-29
[Refactor] Helper for reading yaml header

MichaelChirico:
This avoids the way `close()` is done. We could also use `local()` but this code is already pretty nested, better to off-load to a helper IMO.

A broader question of whether we might just deprecate `yaml=TRUE` is left unaddressed, for now.

### DIFF
--- a/R/fread.R
+++ b/R/fread.R
@@ -353,6 +353,7 @@ yaml=FALSE, tmpdir=tempdir(), tz="UTC")
   ans
 }
 
+# nocov start. Covered only in other.Rraw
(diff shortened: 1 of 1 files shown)


ben-schwen (approved):
LGTM TY!

ben-schwen:
CI found smth [here](https://gitlab.com/Rdatatable/data.table/-/jobs/12655471726)

```
fread: no visible binding for global variable ‘n_read’
Undefined global functions or variables:
  n_read
```

PR #7554, opened 2025-12-30
Refactor colnamesInt to provide context-specific error messages (#5039)

### LINKED ISSUE #5039: colnamesInt makes inscrutable errors
It's great that we've simplified internals to rely consistently on logic in `colnamesInt()`, but unfortunately it has made for some head-scratching debugging, e.g.:

```
call_neighbors[token == 'expr'][!exclude_parents, on = c('file', 'id')][grepl('/merge', file)]
Error in colnamesInt(i, unname(on), check_dups = FALSE) : 
  argument specifying columns specify non existing column(s): cols[2]='id'
```

I have enough context to basically figure out what's gone wrong, but including some context here as an extra helper would be useful, IMO.

What about adding an argument to `colnamesInt` where we can supply a hint about the source of the error?

A quick pass:

```
colnamesInt(..., source){
 ...
 stopf(call. = FALSE, "Column(s) not found in %s: cols[%d]='%s'", source, ...)
}
```

Then call `colnamesInt(..., source = "i table's columns in on= join")`

WDYT Jan? Somewhat off-the-cuff with no repro, LMK if an example is needed.

ayushsingh9720:
This PR addresses issue #5039 by refactoring the internal colnamesInt function at both the C and R levels. By introducing a mandatory source argument, the internal error messages now provide specific context (e.g., pinpointing whether a failure occurred during an on= join, setkey, or frank), making debugging significantly more intuitive for the user.

Technical Changes
1. C Internal Logic (src/)
src/utils.c: Modified the colnamesInt signature to accept a 5th argument, SEXP source. Updated all internal error() calls to include a %s placeholder, prepending the error with the provided context string.

src/data.table.h: Updated the function prototype to remain in sync with the new signature.

src/init.c: Updated the R-to-C registration from -1 (variable) to a strict 5 to ensure formal argument counting at the C interface level.

2. R Wrapper and Call Sites (R/)
R/wrappers.R: Updated the R-side wrapper for colnamesInt to accept source=NULL and pass it through the .Call interface.

Global Refactor: Performed a comprehensive audit and update of 19 call sites across the following files to ensure the new 5th argument is passed correctly with descriptive hints:

R/data.table.R (Joins, setcolorder, setnames)

R/mergelist.R (Merge operations and internal column copying)

R/setkey.R (Key assignment)

R/frank.R (Ranking)

R/duplicated.R (Unique/Duplicated checks)

R/setops.R (Set operations)

Impact
Users will no longer see generic "inscrutable" errors. For example: Old Error: Error in colnamesInt(...) : argument specifying columns received non-existing column(s): cols[2]='z' New Error: Error in colnamesInt(...) : In x table's columns in on= join, argument specifying columns received non-existing column(s): cols[2]='z'

Checklist
[x] Reproduced the original "inscrutable" error.

[x] Successfully compiled using R CMD INSTALL ..

[x] Verified that all 19 call sites are updated to avoid "unused argument" or "argument mismatch" errors.

[x] Passed manual tests for joins, setkeys, and set operations.

Closes #5039

### DIFF
--- a/R/data.table.R
+++ b/R/data.table.R
@@ -495,8 +495,9 @@ replace_dot_alias = function(e) {
           allow.cartesian = TRUE
         }
         # TODO: collect all '==' ops first to speeden up Cnestedid
-        rightcols = colnamesInt(x, names(on), check_dups=FALSE)
-        leftcols  = colnamesInt(i, unname(on), check_dups=FALSE)
+        rightcols = colnamesInt(x, names(on), check_dups=FALSE, skip_absent=FALSE, context="x table's columns in on= join")
+        leftcols = colnamesInt(i, unname(on), check_dups=FALSE,
+                                skip_absent=FALSE, context="i table's columns in on= join")
       } else {
         ## missing on
         rightcols = chmatch(key(x), names_x)   # NAs here (i.e. invalid data.table) checked in bmerge()
@@ -2490,7 +2491,7 @@ na.omit.data.table = function(object, cols = seq_along(object), invert = FALSE, 
   if (!cedta()) return(NextMethod()) # nocov
   if ( !missing(invert) && is.na(as.logical(invert)) )
     stopf("Argument 'invert' must be logical TRUE/FALSE")
-  cols = colnamesInt(object, cols, check_dups=FALSE)
+  cols = colnamesInt(object, cols, check_dups=FALSE, skip_absent=FALSE, context="setnames")
   ix = .Call(Cdt_na, object, cols)
   # forgot about invert with no NA case, #2660
   if (invert) {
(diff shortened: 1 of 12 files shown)


jangorecki:
source is not a good name for an argument as it is widely used function name, I would use context instead.

Unit tests are missing.

ayushsingh9720:
Thank you for the feedback, @jangorecki ! I will rename the argument to context across the C and R levels. I will also add a dedicated unit test file to verify that the context hints are being correctly prepended to the error messages. I'll push these updates shortly.

jangorecki:
@ayushsingh9720 did you use AI when preparing this PR?

ayushsingh9720:
Hi @jangorecki, yes, I've been using an AI as a , kind of a thought/learning partner to help me with the data.table codebase. as i am new to open source, it has been helping me understand the internl logic and debug build failures, but I am manually implementing the changes, running the local R CMD INSTALL builds, and verifying the error messages in my terminal to ensure everything works as intended.

jangorecki:
Ok. Check how we define tests and try to follow that logic, you can read manual of internal function `test`.

ayushsingh9720:
checks are passing now, i have updated the unit test(Test 2357) , to follow the internal test() , as suggested. really appreciate the advice

PR #7557, opened 2026-01-01
Fix fcoalesce Date/IDate mixing and resolve test collisions (#7463)

aksh08022006:
This PR addresses #7463 by allowing fcoalesce() and setcoalesce() to handle mixing of Date (double) and IDate (integer) classes, as well as general mixing of integer and double types. Previously, these combinations would throw a strict type/class mismatch error.

Technical Details The core of the change is in 
src/coalesce.c
. I've updated the logic to:

Relax Type Checks: Added a promote_to_real flag that triggers when fcoalesce sees an integer vector followed by a double. This allows fcoalesce to return a REALSXP vector even if the first argument was INTSXP.
Handle Date/IDate Inheritance: Modified the class-matching check to allow any two objects that both inherit from the Date class. This handles the specific pairing of Date and IDate which were previously rejected as "different classes".
Strict In-place Rules: For setcoalesce(), I've kept the strict rule that we cannot promote an integer vector to double in-place (as that would require memory reallocation). It now throws a specific, helpful error suggesting manual coercion instead of a generic "type mismatch".
Attribute Persistence: Corrected the attribute handling so that if we promote an IDate to a double Date, the class attributes are correctly propagated to the new object.
Testing I've added six new tests to the end of 
inst/tests/tests.Rraw
 (numbered 2357.1–2357.6) covering:

fcoalesce(Date, IDate)
fcoalesce(IDate, Date) (promotion check)
Numeric mixing (INTSXP + REALSXP)
setcoalesce safety checks (ensuring we don't accidentally reallocate in-place).
Verified that these fail on the current master and pass with this build. I also re-ran the full local test suite to ensure no regressions in existing fcoalesce logic.

A Note to Maintainers I've tried my best to follow the data.table coding style and memory safety principles, but I'm still getting my head around some of the C-level internals. I could be wrong about how I've handled the promotion flags or if there's a more efficient way to check for the Date class inheritance at this level. I'm very much open to learning if there's a better or more "the data.table way" to implement this!

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -15628,8 +15628,8 @@ test(2060.202, fcoalesce(fkt), fkt)
 test(2060.203, fcoalesce(bool, 1L), error='Item 2 is type integer but the first item is type logical. Please coerce before coalescing')
 test(2060.204, fcoalesce(bool, NA_integer_), error='Item 2 is type integer but the first item is type logical.')
 test(2060.205, fcoalesce(fkt, 1L), error='Item 1 is a factor but item 2 is not a factor. When factors are involved, all items must be factor')
-test(2060.206, fcoalesce(num, 3L), error='Item 2 is type integer but the first item is type double')
-test(2060.207, fcoalesce(int, 3), error='Item 2 is type double but the first item is type integer')
+test(2060.206, fcoalesce(num, 3L), num_val)
+test(2060.207, fcoalesce(int, 3), num_val)
 test(2060.208, fcoalesce(fkt, 'b'), error='Item 1 is a factor but item 2 is not a factor. When factors are involved, all items must be factor.')
 test(2060.209, fcoalesce(str, factor('b')), error='Item 2 is a factor but item 1 is not a factor. When factors are involved, all items must be factor')
 test(2060.212, fcoalesce(list(1), list(2)), error="The first argument is a list, data.table or data.frame. In this case there should be no other arguments provided.")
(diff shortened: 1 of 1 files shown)


src/coalesce.c
```diff
@@ -7,190 +7,304 @@
     - The conditional checks within parallelized loops
 */
 SEXP coalesce(SEXP x, SEXP inplaceArg, SEXP nan_is_na_arg) {
-  if (TYPEOF(x)!=VECSXP) internal_error(__func__, "input is list(...) at R level"); // # nocov
-  if (!IS_TRUE_OR_FALSE(inplaceArg)) internal_error(__func__, "argument 'inplaceArg' must be TRUE or FALSE"); // # nocov
-  if (!IS_TRUE_OR_FALSE(nan_is_na_arg)) internal_error(__func__, "argument 'nan_is_na_arg' must be TRUE or FALSE"); // # nocov
+  if (TYPEOF(x) != VECSXP)
```
ben-schwen, line 10: please dont reformat existing code. This makes reviewing quite tedious.

inst/tests/tests.Rraw
```diff
@@ -20607,6 +20607,7 @@ test(2294.59,
               VDate = fifelse(VDate == as.Date("3000-01-01"), as.Date(NA), VDate),
               VIDate = fifelse(VIDate == as.IDate("3000-01-01"), as.IDate(NA), VIDate),
               VNumA, VNumB, Count, Y_Sum)])
+
```
ben-schwen, line 20610: Please no formatting changes.

inst/tests/tests.Rraw
```diff
@@ -21968,3 +21969,21 @@ test(2356.2, options=c(datatable.alloccol=1L), set(DT, j="c", value=3), data.tab
 # ensure := and set are consistent if they need to overallocate
 DT = data.table(); DT2 = data.table()
 test(2356.3, options=c(datatable.alloccol=1L), {for (i in seq(10L)) set(DT, j = sprintf("V%d",i), value = i); DT}, {for (i in seq(10)) DT2[, sprintf("V%d",i) := i]; DT2})
+
+# fcoalesce Date and IDate mixing, #7463
+test(2357.1, fcoalesce(as.Date("2026-01-01"), as.IDate("2026-01-02")), as.Date("2026-01-01"))
+test(2357.2, {
```
ben-schwen, line 21975: please dont make tests super complicated. `{}` should only be used when you are confident that it makes the test easier to read and you necessarily need multiple state. this is probably not warranted.

ben-schwen:
I can review, once the formatting diffs are removed. Note that we already have a similar promoting logic in place for `rbindlist` where parts can definetely be reused.

PR #7560, opened 2026-01-02
Restore other.Rraw tests to working state

MichaelChirico:
Motivated as a follow-up to #6583, wound up addressing some other things that were plaguing my fix-iterate loop.

Also note the new test does not exactly match the old behavior (i.e. using knitr.Rout.save), which results from `R CMD BATCH` on the `tests/knitr.R` file, instead I'm just matching the md output contents with `tools::Rdiff()`.


ben-schwen:
LGTM, TY!

PR #7562, opened 2026-01-02
Unify messages for translation

MichaelChirico:
Gemini was used to identify the similar strings; I read over the analysis and applied its findings as appropriate.

This should help out the translation teams by minimizing duplicate work.

<details>
Exact prompt to Gemini:

```
@po/R-data.table.pot is a translation template file. Find similar strings in it that can be unified.

   - First, use the shell to remove all extraneous information from the files and only look at the distinct msgid entries, excluding the initial header entry.
   - Second, examine the corpus of strings and pull out ones that should be unified. Some examples of what we're after:
     + "hello '%s'" and "hello '%s'." could unify by normalizing punctuation
     + "x is negative" and "y is negative" could unify by adding a template: "%s is negative"
     + "bad '%s' argument" and "invalid '%s' argument" could unify by aligning on consistent language: "invalid '%s' argument"
     + "bad '%s' argument" and 'bad "%s" argument' could unify by aligning on consistent punctuation: "bad '%s' argument"
     + There are many other types of messages that could be unified -- use your judgment.
   - Some _negative_ examples that are ineligible:
     + "DLL %s not found" and "shared object %s not found" _look_ like they could unify as "%s %s not found", but there might be languages where "shared object" cannot be pulled out to the template with fidelity
     + "invalid input" and "invalid named string" _look_ like they could unify as "invalid %s", but again, this only works in English, not for generic target languages
```

</details>


R/frank.R
```diff
@@ -20,7 +20,7 @@ frankv = function(x, cols=seq_along(x), order=1L, na.last=TRUE, ties.method=c("a
   } else {
     cols = colnamesInt(x, cols, check_dups=TRUE)
     if (!length(cols))
-      stopf("x is a list, 'cols' can not be 0-length")
+      stopf("x is a list, 'cols' cannot be 0-length.")
```
ben-schwen, line 23:
```suggestion
      stopf("x is a list, 'cols' cannot be 0-length")
```

since "most" error messages dont end with ., but NIT

ben-schwen (approved):
Basically LGTM.

Currently we also two different type of styles whether error message end with . or not. From a quick look, this mostly depends on the source file.

R/frank.R
MichaelChirico: The version with `.` is found elsewhere:

```shell
grep -Fnr "cannot be 0" R
# R/data.table.R:3120:    stopf("x is a list, 'cols' cannot be 0-length.")
# R/data.table.R:3145:    stopf("x is a list, 'cols' cannot be 0-length.")
```

So merging as-is. Aligning on one style there could be a follow-up: #7577

PR #7566, opened 2026-01-05
add join statistics

### LINKED ISSUE #4677: feature request: output from merge

ben-schwen:
Closes #4677

Not happy that we always call the merge arguments `x` and `i` now, even for `merge(x,y)`, but evaluating names of passed arguments seemed overkill.


jangorecki (commented):
~~Afair we have anti join in mergelist, should that be plugged into using this one now?~~
Uh no, this is only for verbose

PR #7568, opened 2026-01-05
Adding support for one level nested data.table

### LINKED ISSUE #2606: tables could look for en-list-ed data.tables as well

manmita:
Closes #2606 

added 2 args to `tables()` one for shallow search and one to skip if the data.table is too big
if `list_search` is false then its the previous code
else we loop through list-like objects using `is.list`

added name for the nested list found
we skip list scanning beyond threshold

we merge to the nested list to the main list if found else it follows the old process

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,7 +18,7 @@
 
 2. `[,showProgress=]` and `options(datatable.showProgress)` now accept an integer to control the progress bar update interval in seconds, allowing finer control over progress reporting frequency; `TRUE` uses the default 3-second interval, [#6514](https://github.com/Rdatatable/data.table/issues/6514). Thanks @ethanbsmith for the report and @ben-schwen for the PR.
 
-2. `tables()` can now optionally report `data.table` objects stored one level deep inside list objects when `list_search=TRUE`, with `list_len_threshold` to avoid scanning extremely long lists, [#2606](https://github.com/Rdatatable/data.table/issues/2606).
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -16,6 +16,8 @@
 
 1. `nafill()`, `setnafill()` extended to work on logical vectors (part of [#3992](https://github.com/Rdatatable/data.table/issues/3992)). Thanks @jangorecki for the request and @MichaelChirico for the PR.
 
+2. `tables()` can now optionally report `data.table` objects stored one level deep inside list objects when `list_search=TRUE`, with `list_len_threshold` to avoid scanning extremely long lists, [#2606](https://github.com/Rdatatable/data.table/issues/2606).
```
jangorecki: An example of when this new feature could be useful?

manmita: Hello @jangorecki,

To support lists such as 
```
      data.table(a = 2, b = 7:10))
```
      
which can occure due to `split.data.table` or `fread` . 

The previous code supported only `data.table()` top level and this code adds support for list( data.table...)

manmita: Note that if we dont want this search we can opt out for it maintaining backward compatiblity and performance as previous.

manmita: @jangorecki 
Here is an example of previous and current versions

Reproducing the `table()` on list item and data.table

Image

Here are the results as mentioned it doesnt take list(data.table(..), data.table(..))

After the PR, adding `list_search`

Image

manmita:
Will open another PR for the same from internal branch
Closing this

PR #7570, opened 2026-01-06
fix(7407): added check for x1a in fread.c to avoid segfault

### LINKED ISSUE #7407: `fread()` allergic to `\x1a` bytes

Somehow, an `\x1a` (octal `\032`, decimal 26, ASCII SUB) somewhere in the input may cause `fread()` to crash.

At the end of the input it causes an out-of-bounds read:

```r
data.table::fread(text = paste0("foo\n", strrep("a", 4096*100), "\x1a"))
#  *** caught segfault ***
# address 0x5574bcafc000, cause 'memory not mapped'
# 
# Traceback:
#  1: data.table::fread(text = paste0("foo\n", strrep("a", 4096 * 100),     "\032"))
# An irrecoverable exception occurred. R is aborting now ...
# Segmentation fault
```

```
Program received signal SIGSEGV, Segmentation fault.
end_of_field (ch=<optimized out>) at fread.c:584
584         while(!end_of_field(ch)) ch++;  // sep, \r, \n or eof will end
(gdb) bt
#0  end_of_field (ch=<optimized out>) at fread.c:584
#1  Field (ctx=0x7fffffffa7b0) at fread.c:584
#2  0x00007ffff42c7ccd in countfields (pch=pch@entry=0x7fffffffa978) at fread.c:413
#3  0x00007ffff42cb118 in freadMain (_args=...) at fread.c:1882
#4  0x00007ffff42d1dcb in freadR (inputArg=<optimized out>, isFileNameArg=<optimized out>, sepArg=<optimized out>, decArg=<optimized out>, quoteArg=<optimized out>,
    headerArg=0x5555559c7360, nrowLimitArg=0x555557f151e8, skipArg=0x555557f7f648, NAstringsArg=0x555557f150d0, stripWhiteArg=0x555557f14ea0,
    skipEmptyLinesArg=0x555557f14e30, commentCharArg=0x555557f14df8, fillArg=0x555557f80410, showProgressArg=0x5555559c7398, nThreadArg=0x555557f80090,
    verboseArg=0x555557f712c8, warnings2errorsArg=0x5555559c7398, logical01Arg=0x555557f148f0, logicalYNArg=0x555557f14880, selectArg=0x5555559c54b0, dropArg=0x5555559c54b0,
    colClassesArg=0x5555559c54b0, integer64Arg=0x555557f14f48, encodingArg=0x555557f14ed8, keepLeadingZerosArgs=0x555557f14810, noTZasUTC=0x555557f7f5a0) at freadR.c:229
(gdb) frame 1
#1  Field (ctx=0x7fffffffa7b0) at fread.c:584
584         while(!end_of_field(ch)) ch++;  // sep, \r, \n or eof will end
(gdb) p ch
$1 = 0x555558d64fff ""
```

I've also seen memory corruption (crashes in `malloc`) when it's not at the end of the field, but those are harder to reproduce. Hopefully they stem from the same problem: field size overestimated, `fread()` correcting embedded NUL bytes and overwriting allocator metadata.

manmita:
Closes #7407 

Adding `ch ==eof` to `end_of_field` to fix seg fault at fread
Removed single lined condition code from `end_of_field`

### DIFF
--- a/inst/tests/tests.Rraw
+++ b/inst/tests/tests.Rraw
@@ -22034,7 +22034,4 @@ if (test_bit64) local({
 
 # 7407 Test for fread() handling \x1A (ASCII SUB) at end of input
 txt = paste0("foo\n", strrep("a", 4096 * 100), "\x1A")
-test(2359.1,
-  nchar(fread(txt)$foo[1]) == 4096 * 100,
-  TRUE
-)
+test(2359.1, nchar(fread(txt)$foo), 409600L)


src/fread.c
```diff
@@ -349,11 +349,7 @@ static inline bool end_of_field(const char *ch)
   // We use eol() because that looks at eol_one_r inside it w.r.t. \r
   // \0 (maybe more than one) before eof are part of field and do not end it; eol() returns false for \0 but the ch==eof will return true for the \0 at eof.
   // Comment characters terminate a field immediately and take precedence over separators.
-  return *ch == sep || ((uint8_t)*ch <= 13 && (ch == eof || eol(&ch))) || (commentChar && *ch == commentChar);
-  if (*ch == sep) return true;
-  if ((uint8_t)*ch <= 13 && (ch == eof || eol(&ch))) return true;
-  if (!commentChar) return false;
-  return *ch == commentChar;
+  return *ch == sep || *ch == '\x1A'|| ((uint8_t)*ch <= 13 && (ch == eof || eol(&ch))) || (commentChar && *ch == commentChar);
```
ben-schwen: The idea of the code above was that the complicated single line OR statement gets easier to grasp, so we would favor removing line 352 and add an case for `if (*ch == '\x1A') return true;`

ben-schwen:
```suggestion
  if (*ch == sep) return true;
  if (ch == eof) return true;  // Check eof first to avoid reading past #7407
  if ((uint8_t)*ch <= 13 && eol(&ch)) return true;
  if (!commentChar) return false;
  return *ch == commentChar;
```

While I couldn't come up with another example than `0x1A` it seems safer to check for eof instead of checking for `0x1A` specifically

NEWS.md
```diff
@@ -32,6 +32,8 @@
 
 3. `fread("file://...")` works for file URIs with spaces, [#7550](https://github.com/Rdatatable/data.table/issues/7550). Thanks @aitap for the report and @MichaelChirico for the PR.
 
+4. `fread(text = paste0("foo\n", strrep("a", 4096*100), "\x1a"))` gives seg fault  [#7407](https://github.com/Rdatatable/data.table/issues/7407)which is solved by adding check for `\x1A` at `end_of_field`. Thanks @aitap for the report.
```
ben-schwen: Maybe something like 
> `fread(text=)` could segfault when reading text input ending with a `\x1a` (ASCII SUB) character after a long line,

inst/tests/tests.Rraw
```diff
@@ -21978,3 +21978,10 @@ local({
   test(2357.1, fread(f), DT)
   test(2357.2, fread(paste0("file://", f)), DT)
 })
+
+# 7407 Test for fread() handling \x1A (ASCII SUB) at end of input
+fread_sub_test_txt = paste0("foo\n", strrep("a", 4096 * 100), "\x1A")
+test(2358.1, {
```
ben-schwen: We only use { in tests if they make the tests clearer and concise.

Setting up the str with `txt = paste0("foo\n", strrep("a", 4096 * 100), "\x1A")` is fine.
For the test maybe test for `nchar(fread(txt))` which ensure that the whole string is read?

src/fread.c
manmita: Got it, in the file that code was written after the long single line one so it was dead code, hence removed it but I'll do the reverse, 

keep those lines and remove this one

manmita: yeah, adding eof should also work, the bug was that ch==eof was in and with ch<=13 but x1A is greater

inst/tests/tests.Rraw
```diff
@@ -22030,3 +22030,10 @@ if (test_bit64) local({
     merged$gforce_mean, merged$true_mean
   )
 })
+
+# 7407 Test for fread() handling \x1A (ASCII SUB) at end of input
+txt = paste0("foo\n", strrep("a", 4096 * 100), "\x1A")
+test(2359.1,
+  nchar(fread(txt)$foo[1]) == 4096 * 100,
+  TRUE
+)
```
ben-schwen:
```suggestion
test(2359.1, nchar(fread(txt)$foo), 409600L)
```

ben-schwen: Maybe put it into a single line? Our test file is already long enough :)

ben-schwen:
GJ and TY!

PR #7572, opened 2026-01-07
fix(7571): bug fix for narm issue on gforce in int64 case

### LINKED ISSUE #7571: sum(value, na.rm = FALSE) returns unexpected 0s for integer64 field within NAs in group
**Problem:** It's not clear to me why `sum(value, na.rm = FALSE)` does not always return NA when there is an NA value in an integer64 column within a given id group: `dt_short[, result_problem := sum(value, na.rm = FALSE), by = id]`.  In the examples below, the groups each contain one value, but behavior doesn't seem look to be limited to that scenario.

**Resolved by**: Casting the value in question to an integer, rather than the initial integer64 value.

`#` [`Minimal reproducible example`]
``` r
library(data.table)
library(bit64)

# Expected that sum(., na.rm = FALSE) would return "NA" for id's 1-3 in the example below.

dt_short <- data.table(
  id = c(1:8, 9, 9),
  value = c(rep(NA_integer64_, 3), 4:10)
)

# Shouldn't the first 4 values of result_problem be 3 NA's followed by a 4?
dt_short[, result_problem := sum(value, na.rm = FALSE), by = id]
dt_short 
#>        id value result_problem
#>     <num> <i64>          <i64>
#>  1:     1  <NA>           <NA>
#>  2:     2  <NA>              0
#>  3:     3  <NA>              0
#>  4:     4     4              0
#>  5:     5     5              5
#>  6:     6     6              6
#>  7:     7     7              7
#>  8:     8     8              8
#>  9:     9     9             19
#> 10:     9    10             19

# works as expected when cast as an integer (instead of an integer64)
dt_short[, result_asinteger := sum(as.integer(value), na.rm = FALSE), by = id]
dt_short

#>        id value result_problem result_asinteger
#>     <num> <i64>          <i64>            <int>
#>  1:     1  <NA>           <NA>               NA
#>  2:     2  <NA>              0               NA
#>  3:     3  <NA>              0               NA
#>  4:     4     4              0                4
#>  5:     5     5              5                5
#>  6:     6     6              6                6
#>  7:     7     7              7                7
#>  8:     8     8              8                8
#>  9:     9     9             19               19
#> 10:     9    10             19               19

# Note: For some reason, the issue does not seem to affect the first id.
# And there seems to be some offset effect since the sum for id 4 ends up being "0" instead of "4".  

# Other scenarios
# see the issue here when just returning the by and new column
dt_short[, .(result_problem = sum(value, na.rm = FALSE)), by = id]
#>       id result_problem
#>    <num>          <i64>
#> 1:     1           <NA>
#> 2:     2              0
#> 3:     3              0
#> 4:     4              0
#> 5:     5              5
#> 6:     6              6
#> 7:     7              7
#> 8:     8              8
#> 9:     9             19

# Don't see the issue when casting the value that gets summed as an integer first
dt_short[, .(result_problem = sum(as.integer(value), na.rm = FALSE)), by = id]
#>       id result_problem
#>    <num>          <int>
#> 1:     1             NA
#> 2:     2             NA
#> 3:     3             NA
#> 4:     4              4
#> 5:     5              5
#> 6:     6              6
#> 7:     7              7
#> 8:     8              8
#> 9:     9             19

# Don't see the issue here, when including another column in the returned values
dt_short[, .(value, result_problem = sum(value, na.rm = FALSE)), by = id]
#>        id value result_problem
#>     <num> <i64>          <i64>
#>  1:     1  <NA>           <NA>
#>  2:     2  <NA>           <NA>
#>  3:     3  <NA>           <NA>
#>  4:     4     4              4
#>  5:     5     5              5
#>  6:     6     6              6
#>  7:     7     7              7
#>  8:     8     8              8
#>  9:     9     9             19
#> 10:     9    10             19
```

`#` `Output of sessionInfo()`
#> R version 4.3.1 (2023-06-16)
#> Running under: Ubuntu 20.04.6 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3 
#> LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so.3;  LAPACK version 3.9.0
#> 
#> locale:
#> 
#> time zone: America/New_York
#> tzcode source: system (glibc)
#> 
#> attached base packages:
#> 
#> other attached packages:
#> [1] bit64_4.6.0-1     bit_4.6.0         data.table_1.17.8
#> 
#>  [1] digest_0.6.37     fastmap_1.2.0     xfun_0.52         glue_1.8.0       
#>  [5] knitr_1.50        htmltools_0.5.8.1 rmarkdown_2.29    lifecycle_1.0.4  
#>  [9] cli_3.6.5         reprex_2.1.1      withr_3.0.2       compiler_4.3.1   
#> [13] rstudioapi_0.17.1 tools_4.3.1       evaluate_1.0.4    yaml_2.3.10      
#> [17] rlang_1.1.6       fs_1.6.6
```

<sup>Created on 2025-12-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### PR DESCRIPTION
**Pull request by manmita**
Closes #7571 

Added a condition to check if ans consist of NA values if element is not NA
Removed the break statement if the element is NA, as that was causing bug on consecutive NA values in the same partition but different group.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,7 +34,7 @@
 
 3. `fread("file://...")` works for file URIs with spaces, [#7550](https://github.com/Rdatatable/data.table/issues/7550). Thanks @aitap for the report and @MichaelChirico for the PR.
 
-4. `sum(<int64 column>)` by group is correct with GFoce activated ([#7571](https://github.com/Rdatatable/data.table/issues/7571)). Thanks to @rweberc for the report and @manmita for the fix. The issue was caused by a faulty early `break` that spilled between groups.
+4. `sum(<int64 column>)` by group is correct with missing entries and GForce activated ([#7571](https://github.com/Rdatatable/data.table/issues/7571)). Thanks to @rweberc for the report and @manmita for the fix. The issue was caused by a faulty early `break` that spilled between groups, and resulted in silently incorrect results!
 
 ## data.table [v1.18.0](https://github.com/Rdatatable/data.table/milestone/37?closed=1)  23 December 2025
 

### CODE REVIEW

File: NEWS.md [hunk 1]
```diff
@@ -32,6 +32,8 @@
 
 3. `fread("file://...")` works for file URIs with spaces, [#7550](https://github.com/Rdatatable/data.table/issues/7550). Thanks @aitap for the report and @MichaelChirico for the PR.
 
+4. Fixed a bug in GForce grouped sum for `integer64` columns where `sum(value, na.rm = FALSE)` did not always return `NA` when any `NA_integer64_` was present in a group. [#7571](https://github.com/Rdatatable/data.table/issues/7571). Now, groups containing any `NA_integer64_` correctly return `NA`, matching base R behavior. Thanks to @rweberc for the report and @MichaelChirico for reproducible examples.
```
jangorecki: You could as well mention yourself here, and ideally write what was the root cause of that so readers are more likely to understand if they could have been affected

MichaelChirico:
Thanks @manmita! Please ping me when tests are passing, or if you're stuck.

NEWS.md
manmita: Thanks, will do

manmita:
Hello @MichaelChirico , 

There is coverage issue in dogroups.c.

inst/tests/tests.Rraw
```diff
@@ -21978,3 +21978,27 @@ local({
...
+
+#7571 issue for na.rm on int64
+if (test_bit64) local({
+  # integer64 + GForce grouped sum with na.rm = FALSE
+  # Example 1 from issue: ids 1:8, 9, 9; three leading NAs then 4:10
+  dt_short = data.table(
+    id = c(1:8, 9, 9),
+    value = c(rep(NA_integer64_, 3L), as.integer64(4:10))
+  )
+  test(2358.1,
```
MichaelChirico: Use `test(options=c(datatable.optimize=2L))` to emphasize that this was a GForce problem

manmita: Thanks, will do

NEWS.md
```diff
@@ -32,6 +32,8 @@
 
 3. `fread("file://...")` works for file URIs with spaces, [#7550](https://github.com/Rdatatable/data.table/issues/7550). Thanks @aitap for the report and @MichaelChirico for the PR.
 
+4. - Fixed a bug in GForce grouped sum for `integer64` columns where `sum(value, na.rm = FALSE)` when consecutive NA values are present in the same partition but different group. In that case only the first group was handled ([#7571](https://github.com/Rdatatable/data.table/issues/7571)). Thanks to @rweberc for the report, @MichaelChirico for reproducible examples, and @manmita for the fix.
```
MichaelChirico: I didn't want to anchor the NEWS too much on implementation details -- user reading this likely doesn't care all that much.

It still feels a bit lacking though -- maybe there's a more concise way to explain it?

What is the exact nature of the bug from a user point of view? Is it true that it always results in a group having "sum" `0` instead of the correct `NA_integer64_`?

manmita: in the code we have a variable called `howMany` which determines how many are there in this batch
but if the group by id should have been grouped by 2 and batching is done by 4, then the sum becomes 0 for everything after the first encountered na. Although the first na sets the first id group to correct value.

`dt_short[, result_problem := sum(value, na.rm = FALSE), by = id] ` for this if id is 1,1,2,2,3,3 and has na,na,na,1,2,3
then the result is na, 0, 5 if the batch size is 4

manmita: I guess we can say that when we use group by something on na values with paralel processing then multiple groups (consisting na values) can fall into the same batch but the break will only take the first group 

from user perspective

MichaelChirico: That we do things in parallel/in batches should not matter to user who just wants correct code.

> then the result is na, 0, 5 if the batch size is 4

It looks like it confirms my diagnosis -- the error always appears as an incorrect `0`. Is it not possible that the `NA_integer64_` is encountered later after the sum has already been initialized?

manmita:
Hii @MichaelChirico the coverage is failing at src/dogroups.c which was fixed by @ben-schwen in this commit 
https://github.com/Rdatatable/data.table/commit/d6fb16c1b7038df5ec935de984bca25f8e1a0b61

The sentry report shows that file only in their report
I was wondering how to add this into this PR?

NEWS.md
manmita: actually the na can still be possible if its in a diff batch like na, na, na, na, na, na for ids 1,2,3,4,5,6 and batch size 4 then it'll give na, 0, 0, 0, na, 0

manmita: but maybe we can say that na followed by zeros kind of behavior incase of consecutive nas

MichaelChirico:
> Hii @MichaelChirico the coverage is failing at src/dogroups.c which was fixed by @ben-schwen in this commit [d6fb16c](https://github.com/Rdatatable/data.table/commit/d6fb16c1b7038df5ec935de984bca25f8e1a0b61)
> 
> The sentry report shows that file only in their report I was wondering how to add this into this PR?

you can ignore it here

MichaelChirico:
This bug is really quite serious I think.

Consider:

```r
DT = data.table(id = sample(letters, 1000, TRUE), value = as.integer64(sample(c(1:100, NA), 1000, TRUE)))

DT[,
  .(gforce_sum = sum(value)), by=id
][
  DT[, .(true_sum = base::sum(value)), by=id],
  on='id'
][
  gforce_sum != true_sum | is.na(gforce_sum) != is.na(true_sum)
]
#         id           gforce_sum true_sum
#     <char>                <i64>    <i64>
#  1:      p                 1768     2188
#  2:      g                 2489     2684
#  3:      t                 1451     1922
#  4:      h -9223372036854775096     <NA>
#  5:      n                  687     2410
#  6:      m                  266     1556
#  7:      k                  986     2101
#  8:      y                 1035     1398
#  9:      q -9223372036854774756     <NA>
# 10:      d                 1650     <NA>
# 11:      c -9223372036854774942     <NA>
# 12:      f                 1073     1674
# 13:      o                  710     1513
# 14:      v                  760     1660
# 15:      l                 1076     <NA>
# 16:      e                 1089     1736
# 17:      w -9223372036854774951     <NA>
# 18:      i                 1204     <NA>
# 19:      j                 1478     <NA>
```

MichaelChirico:
It also affects `mean()`, and `Cfastmean` appears affected too

manmita:
yeahh, seems like

Will think it through, checkout the eg cases and the code again.

Thanks

manmita:
The current eg did get fixed with the changes made in this branch

```
> DT[,
  .(gforce_sum = sum(value)), by=id
][
  DT[, .(true_sum = base::sum(value)), by=id],
  on='id'
][
  gforce_sum != true_sum | is.na(gforce_sum) != is.na(true_sum)
]
Empty data.table (0 rows and 3 cols): id,gforce_sum,true_sum
```

But I think we should add more test cases and test it thoroughly
for mean and Cfastmean too

I believe why the current fix works because
we iterate through each and every element given and we add them if the sum of that particular group is not NA.

MichaelChirico:
yes, sorry, to clarify, my example is from `master`, not this PR. I'm just showing how bad the behavior is before the fix.

It would be good to ensure the same issue does not happen elsewhere.

manmita:
yeah true, 

I think extensive testing is needed for this case, will write some from my end

and check the mean/fastmean too

manmita:
in the gmean code  - line 619

```
 _ans[my_low[i]] += my_gx[i];  // let NA propagate when !narm
 ```
 
 this is adding the NA to the sum so thats creating an issue for int64 (my_gx[i] is the data item)
 if NA sentinel int64_min is added to say 2000 then its no more int64_min so that should give wrong values.
 just spectaculations
 
 Will check it out!
 
 I was wondering if fix for this will go as a seperate PR or a part of this PR?
 @MichaelChirico

**Comment by MichaelChirico** [comment]
> in the gmean code - line 619
> 
> ```
>  _ans[my_low[i]] += my_gx[i];  // let NA propagate when !narm
> ```
> 
> this is adding the NA to the sum so thats creating an issue for int64 (my_gx[i] is the data item) if NA sentinel int64_min is added to say 2000 then its no more int64_min so that should give wrong values. just spectaculations
> 
> Will check it out!
> 
> I was wondering if fix for this will go as a seperate PR or a part of this PR? @MichaelChirico

Good question. For now, let's put it in this PR. The code change seems like it will be small enough, and the new tests feel logically connected.

**Comment by ben-schwen** [comment]
Sorry that I'm late to that party, but wouldn't it be easier (and also maybe easier to grasp) to mirror the `INTSXP` case with `int64` without the overflow part?

e.g. remove the `if (!narm)` condition and the else branch and just use
```c
            for (int i=0; i<howMany; i++) {
              const int64_t a = _ans[my_low[i]];
              if (a==INT64_MIN) continue;
              const int64_t b = my_gx[i];
              if (b==INT64_MIN) {
                if (!narm) _ans[my_low[i]]=INT64_MIN;
                continue;
              }
              _ans[my_low[i]] += b;
            }
```

inst/tests/tests.Rraw
```diff
@@ -22001,4 +22001,25 @@ if (test_bit64) local({
...
+    c(10, NA, 5)
+  )
+
+  # Randomized test: GForce sum vs base::sum for integer64
+  set.seed(7571)
+  DT = data.table(id = sample(letters, 1000, TRUE), value = as.integer64(sample(c(1:100, NA), 1000, TRUE)))
+  gforce = DT[, .(gforce_sum = sum(value)), by=id]
+  base = DT[, .(true_sum = base::sum(value)), by=id]
+  merged = merge(gforce, base, by="id", all=TRUE)
+  test(2358.4, options=c(datatable.optimize=2L),
```
MichaelChirico, line 22020: I think we can make this just

```r
test(2358.4, options=c(datatable.optimize=2L),
     merged$gforce_sum, merged$true_sum)
```

manmita:
I think the mean is working correctly as I tested with similar examples but the mean code is giving correct results
reason - INT_MIN, INT64_MIN  is used as sentinel value so adding anything to that is still NA

Also updated the gsum code to follow the int counterpart

manmita:
Thank you @MichaelChirico @ben-schwen @jangorecki for you help and suggestions <3

MichaelChirico:
Thanks @manmita. I've invited you to be a Project Member.

Once you accept, you'll be able to create branches directly on this repo (Rdatatable/data.table). Please do so going forward, as it makes collaboration a bit easier.

I'll also file a follow-up PR adding you as a contributor in the DESCRIPTION.

PR #7573, opened 2026-01-08
Add @manmita as a contributor

MichaelChirico:
@manmita please confirm this is you're preferred apellation for this citation. Thanks!


manmita:
Yeah this is the preferred one, 
Thanks

I wanted to ask if I should continue on the following PR as its still ongoing
or raise another PR for that issue?
#7570

MichaelChirico:
You can keep that PR, no need to re-push it. It's good to keep the existing review comments in one place.

PR #7575, opened 2026-01-09
Use R_getVar over findVar

### LINKED ISSUE #6180: Remove non-API calls in C

MichaelChirico:
Closes #6180, IINM.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,16 +18,6 @@
 
 2. `[,showProgress=]` and `options(datatable.showProgress)` now accept an integer to control the progress bar update interval in seconds, allowing finer control over progress reporting frequency; `TRUE` uses the default 3-second interval, [#6514](https://github.com/Rdatatable/data.table/issues/6514). Thanks @ethanbsmith for the report and @ben-schwen for the PR.
 
-### Notes
-
-1. {data.table} now depends on R 3.5.0 (2018).
-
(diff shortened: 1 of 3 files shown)


ben-schwen:
TY!

PR #7580, opened 2026-01-09
Added shallow search for data.table in tables()

### LINKED ISSUE #2606: tables could look for en-list-ed data.tables as well

manmita:
Closes #2606 

added arg `depth = 1L` to tables() one for shallow search
if `depth` is 0 then its the data.table
if  `depth` is 1, we loop through list-like objects using is.list and  which are not data.table
if `depth` > 1, we throw error

added name for the nested list found `parent[[1]]` or `parent$child`
pre-allocating info to avoid reallocation cost

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -46,11 +46,9 @@ tables = function(mb=type_size, order.col="NAME", width=80L,
     is_dt = vapply_1b(obj, is.data.table)
     # list_index is a index of list which is not data.table
     list_index = which(is_list & !is_dt)
-    # obj_list is a list of lists of data.tables found inside lists
-    obj_list = vector("list", length(list_index))
     # make a list of size list_index and add wl in it
     total_dt = 0L
-    # filling obj_list and counting total_dt
+    # counting total_dt
     for (i in seq_along(list_index)) {
       L = obj[[list_index[i]]]
       total_dt = total_dt + sum(vapply_1b(L, is.data.table))


manmita:
Hello, 

I created a new PR in replacement of #7568

Reasons: There was some git issue there and the merge became too complex and I changed the algo because I didnt know previously that rbind or cbind would cost for re-allocation

The current PR considers that part and avoids appends 

Previous PR : creating seperate data.table called info and rbind at the end
This PR: pre-allocates for a total-sized data.table and fills the info

manmita:
In reply to previous comment of @jangorecki 

An example of when this new feature could be useful?

To support lists which occur due to split.data.table or fread like the following
```
      data.table(a = 2, b = 7:10))
```

The original code supported data.table() top level and this code adds support for list(data.table) if the arg `shallow_search = TRUE`

manmita:
Example of the original code and the new feature is as follows

```
> A = list(data.table(a = 1, b = 4:6),
      data.table(a = 2, b = 7:10))
> B = list(data.table(a = 1, b = 4:6), 1:5)
> C = data.table(a = 1, b = 4:6)
> tables()
   NAME NROW NCOL MB COLS    KEY
1:    C    3    2  0  a,b [NULL]
Total: 0MB using type_size
> tables(shallow_search = TRUE)
     NAME NROW NCOL MB COLS    KEY
1: A[[1]]    3    2  0  a,b [NULL]
2: A[[2]]    4    2  0  a,b [NULL]
3: B[[1]]    3    2  0  a,b [NULL]
4:      C    3    2  0  a,b [NULL]
Total: 0MB using type_size
> D = list(d = data.table(a = 1, b = 4:6), x = 1:5)
> tables(shallow_search = TRUE)
     NAME NROW NCOL MB COLS    KEY
1: A[[1]]    3    2  0  a,b [NULL]
2: A[[2]]    4    2  0  a,b [NULL]
3: B[[1]]    3    2  0  a,b [NULL]
4:      C    3    2  0  a,b [NULL]
5:    D$d    3    2  0  a,b [NULL]
Total: 0MB using type_size
```

`tables()` work same as before and ` tables(shallow_search = TRUE)` searches 1 level

inst/tests/tests.Rraw
```diff
@@ -22035,3 +22035,25 @@ if (test_bit64) local({
...
+#2606 tables() shallow_search finds nested data.tables in lists
+# creating env so that the names are within it
+xenv2 = new.env()
+xenv2$DT = data.table(a = 1L)
+xenv2$L = list(data.table(a = 1, b = 4:6), data.table(a = 2, b = 7:10))
+xenv2$M = list(b = data.table(a = 1, b = 4:6), a = 1:5)
+xenv2$N = list(a = 1:5)
+setkey(xenv2$M$b, a)
+setindex(xenv2$M$b, b)
+test(2360.1, tables(env = xenv2, shallow_search = TRUE)$NAME, c("DT", "L[[1]]", "L[[2]]", "M$b"))
```
MichaelChirico: prefer saving the output to re-running `tables(env = xenv2, shallow_search = TRUE)` many times.

alternatively, just do one test like

```r
test(2360.1, tables(env = xenv2, shallow_search = TRUE)[, .(NAME, NROW, NCOL)], data.table(...))
```

inst/tests/tests.Rraw
```diff
@@ -22035,3 +22035,25 @@ if (test_bit64) local({
...
+
+#2606 tables() shallow_search finds nested data.tables in lists
+# creating env so that the names are within it
+xenv2 = new.env()
+xenv2$DT = data.table(a = 1L)
+xenv2$L = list(data.table(a = 1, b = 4:6), data.table(a = 2, b = 7:10))
+xenv2$M = list(b = data.table(a = 1, b = 4:6), a = 1:5)
+xenv2$N = list(a = 1:5)
+setkey(xenv2$M$b, a)
+setindex(xenv2$M$b, b)
```
MichaelChirico: nit: I would move `setindex()` closer to the `index=TRUE` tests

R/tables.R
```diff
@@ -19,19 +19,85 @@ type_size = function(DT) {
...
+          if (is.function(mb)) set(info, k, "MB", as.integer(mb(DT)/1048576L)) # i.e. 1024**2
+          if (!is.null(tt<-names(DT))) set(info, k, "COLS", tt)  # TODO: don't need these if()s when #5526 is done
+          if (!is.null(tt<-key(DT))) set(info, k, "KEY", tt)
+          if (index && !is.null(tt<-indices(DT))) set(info, k, "INDICES", tt)
+        }
+      }
+    }
+  }
+  else {
+    # the original code path when shallow_search=FALSE
```
MichaelChirico: this comment doesn't make sense outside the context of this PR, "the original code path" will be a relic within a few months

R/tables.R
```diff
@@ -19,19 +19,85 @@ type_size = function(DT) {
 }
 
 tables = function(mb=type_size, order.col="NAME", width=80L,
-                  env=parent.frame(), silent=FALSE, index=FALSE)
+                  env=parent.frame(), silent=FALSE, index=FALSE,
+                  shallow_search=FALSE)
```
MichaelChirico: I am thinking the best way to go about this is actually something like either `depth=0` or `recursive=FALSE`. Then both cases share the same logic, except that the default cuts out after the shallow search.

If the code to do a recursive walk is proving intimidating, we can do `depth=0`, and this PR can support `depth=1` and error for `depth>1` "not yet supported" and leave it for future work.

WDYT?

manmita: yeah we can do a depth = 0, 1 and error at depth >1 for this PR

inst/tests/tests.Rraw
manmita: making it one test

R/tables.R
```diff
@@ -19,19 +19,89 @@ type_size = function(DT) {
...
+    # if depth is 0
+    if (!length(w)) {
+      if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
+      return(invisible(data.table(NULL)))
+    }
+    info = data.table(NAME=names[w], NROW=0L, NCOL=0L, MB=0.0, COLS=list(), KEY=list(), INDICES=list())
+  }
+  else {
+    # for depth greater than 1L,recursion is not implemented yet
+    stop("depth > 1L is not implemented yet", call. = FALSE)
```
MichaelChirico: ah, sorry, the "correct" way to make that lint go away is to use `stopf()` instead

R/tables.R
```diff
@@ -19,19 +19,89 @@ type_size = function(DT) {
...
+  # we check if depth=1L is requested and add found tables to w
+  if (depth == 1L) {
+    is_list = vapply_1b(obj, is.list)
+    is_df = vapply_1b(obj, is.data.frame)
+    is_dt = vapply_1b(obj, is.data.table)
+    # list_index is a index of list which is not data.frame or data.table
+    list_index = which(is_list & !is_dt & !is_df)
+    # obj_list is a list of lists of data.tables found inside lists
+    obj_list = vector("list", length(list_index))
+    #make a listof size list_index and add wl in it
```
MichaelChirico:
```suggestion
    # make a list of size list_index and add wl in it
```

inst/tests/tests.Rraw
```diff
@@ -21515,3 +21515,29 @@ test(2365.1, melt(df_melt, id.vars=1:2), melt(dt_melt, id.vars=1:2))
 df_dcast = data.frame(a = c("x", "y"), b = 1:2, v = 3:4)
 dt_dcast = data.table(a = c("x", "y"), b = 1:2, v = 3:4)
 test(2365.2, dcast(df_dcast, a ~ b, value.var = "v"), dcast(dt_dcast, a ~ b, value.var = "v"))
+
+#2606 tables() depth=1 finds nested data.tables in lists
+# creating env so that the names are within it
+xenv2 = new.env()
+xenv2$DT = data.table(a = 1L)
```
MichaelChirico: nit: data.table style is to omit spaces before and after named arguments: `data.table(a=1L)`

inst/tests/tests.Rraw
```diff
@@ -21515,3 +21515,29 @@ test(2365.1, melt(df_melt, id.vars=1:2), melt(dt_melt, id.vars=1:2))
...
+  data.table(
+    NAME = c("DT", "L[[1]]", "L[[2]]", "M$b"),
+    NROW = c(1L, 3L, 4L, 3L),
+    NCOL = c(1L, 2L, 2L, 2L)
+))
+setindex(xenv2$M$b, b)
+test(2366.2, tables(env = xenv2, depth = 1L, index = TRUE)$INDICES, list(NULL, NULL, NULL, "b"))
+setkey(xenv2$M$b, a)
+test(2366.3, tables(env = xenv2, depth = 1L, index = TRUE)$KEY, list(NULL, NULL, NULL, "a"))
+test(2366.4, tryCatch(tables(env = xenv2, depth = 2L), error = function(e) e$message), "depth > 1L is not implemented yet")
```
MichaelChirico: don't use tryCatch like this in our suite; you can use the `error=` argument.

https://rdatatable.gitlab.io/data.table/reference/test.html

inst/tests/tests.Rraw
```diff
@@ -21515,3 +21515,29 @@ test(2365.1, melt(df_melt, id.vars=1:2), melt(dt_melt, id.vars=1:2))
...
+rm(xenv2)
+
+# no data.table test and depth >1 test
```
MichaelChirico: comment is maybe in the wrong place?

inst/tests/tests.Rraw
```diff
@@ -21515,3 +21515,29 @@ test(2365.1, melt(df_melt, id.vars=1:2), melt(dt_melt, id.vars=1:2))
 df_dcast = data.frame(a = c("x", "y"), b = 1:2, v = 3:4)
 dt_dcast = data.table(a = c("x", "y"), b = 1:2, v = 3:4)
 test(2365.2, dcast(df_dcast, a ~ b, value.var = "v"), dcast(dt_dcast, a ~ b, value.var = "v"))
+
+#2606 tables() depth=1 finds nested data.tables in lists
+# creating env so that the names are within it
+xenv2 = new.env()
+xenv2$DT = data.table(a = 1L)
+xenv2$L = list(data.table(a = 1, b = 4:6), data.table(a = 2, b = 7:10))
```
MichaelChirico: this test should also include a further-nested table to demonstrate that `depth=1L` is honored:

```r
xenv2$LL = list(list(data.table(a=1L, b=4:6)))
```

There, we'd need `depth=2L` to find the data.table, AIUI.

inst/tests/tests.Rraw
```diff
@@ -21515,3 +21515,29 @@ test(2365.1, melt(df_melt, id.vars=1:2), melt(dt_melt, id.vars=1:2))
 df_dcast = data.frame(a = c("x", "y"), b = 1:2, v = 3:4)
 dt_dcast = data.table(a = c("x", "y"), b = 1:2, v = 3:4)
 test(2365.2, dcast(df_dcast, a ~ b, value.var = "v"), dcast(dt_dcast, a ~ b, value.var = "v"))
+
+#2606 tables() depth=1 finds nested data.tables in lists
+# creating env so that the names are within it
+xenv2 = new.env()
```
MichaelChirico: why `xenv2`? Where is `xenv`?

man/tables.Rd
```diff
@@ -15,6 +15,7 @@ tables(mb=type_size, order.col="NAME", width=80,
   \item{env}{ An \code{environment}, typically the \code{.GlobalEnv} by default, see Details. }
   \item{silent}{ \code{logical}; should the output be printed? }
   \item{index}{ \code{logical}; if \code{TRUE}, the column \code{INDICES} is added to indicate the indices assorted with each object, see \code{\link{indices}}. }
+  \item{depth}{\code{integer}; if \code{1L}, searches for \code{data.table} objects inside top-level lists. If depth = 0L it accepts data.table and Values greater than \code{1L} are not implemented yet.}
```
MichaelChirico:
```suggestion
  \item{depth}{\code{integer}, default \code{0L}. Larger values govern the depth of search in recursive objects like lists or environments. For example, \code{depth=1L} will find all data.tables in the global environment as well as all data.tables in lists (but in lists of lists). NB: \code{depth > 1L} is not yet supported.}
```

manmita:
Hello @MichaelChirico , I have added the suggested changes.
regarding the coding style for data.table as `data.table(a=1L)`, in my previous (PRs) tests, I have missed this. 
Is it alright if I fix them in this PR?

ben-schwen:
> Hello @MichaelChirico , I have added the suggested changes. regarding the coding style for data.table as `data.table(a=1L)`, in my previous (PRs) tests, I have missed this. Is it alright if I fix them in this PR?

Yes, since we follow the Boy Scout principle here, which is the practice of leaving code slightly better than you found it.

man/tables.Rd
```diff
@@ -15,6 +15,7 @@ tables(mb=type_size, order.col="NAME", width=80,
   \item{env}{ An \code{environment}, typically the \code{.GlobalEnv} by default, see Details. }
   \item{silent}{ \code{logical}; should the output be printed? }
   \item{index}{ \code{logical}; if \code{TRUE}, the column \code{INDICES} is added to indicate the indices assorted with each object, see \code{\link{indices}}. }
+  \item{depth}{\code{integer}, default \code{0L}. Larger values govern the depth of search in recursive objects like lists or environments. For example, \code{depth=1L} will find all data.tables in the global environment as well as all data.tables in lists (but in lists of lists). NB: \code{depth > 1L} is not yet supported.}
```
MichaelChirico:
```suggestion
  \item{depth}{\code{integer}, default \code{0L}. Larger values govern the depth of search in recursive objects like lists or environments. For example, \code{depth=1L} will find all data.tables in the global environment as well as all data.tables in lists (but not in lists of lists). NB: \code{depth > 1L} is not yet supported.}
```

R/tables.R
```diff
@@ -19,19 +19,87 @@ type_size = function(DT) {
...
   names = ls(envir=env, all.names=TRUE)  # include "hidden" objects (starting with .)
   obj = mget(names, envir=env)  # doesn't copy; mget is ok with ... unlike get, #5197
   w = which(vapply_1b(obj, is.data.table))
-  if (!length(w)) {
-    if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
-    return(invisible(data.table(NULL)))
+
+  info = NULL
+  # we check if depth=1L is requested and add found tables to w
+  if (depth==1L) {
```
MichaelChirico:
```suggestion
  if (depth == 1L) {
```

R/tables.R
```diff
@@ -19,19 +19,87 @@ type_size = function(DT) {
...
+    for (i in seq_along(list_index)) {
+      L = obj[[list_index[i]]]
+      wl = which(vapply_1b(L, is.data.table))
+      total_dt = total_dt + length(wl)
+      obj_list[[i]] = L[wl]
+    }
+    name_count = length(w) + total_dt
+    # initialize info data.table with total number of data.tables found
+    if (name_count==0L) {
+      # nocov start. Requires long-running test case
```
MichaelChirico: I don't follow, why would it be slow?

R/tables.R
```diff
@@ -19,19 +19,87 @@ type_size = function(DT) {
...
   w = which(vapply_1b(obj, is.data.table))
-  if (!length(w)) {
-    if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
-    return(invisible(data.table(NULL)))
+
+  info = NULL
+  # we check if depth=1L is requested and add found tables to w
+  if (depth==1L) {
+    is_list = vapply_1b(obj, is.list)
+    is_df = vapply_1b(obj, is.data.frame)
```
MichaelChirico: note that `inherits(<data.table>, "data.frame")` is TRUE (or at least we can assume it to be), so `is_dt` is redundant

R/tables.R
```diff
@@ -19,19 +19,87 @@ type_size = function(DT) {
...
+    is_dt = vapply_1b(obj, is.data.table)
+    # list_index is a index of list which is not data.frame or data.table
+    list_index = which(is_list & !is_dt & !is_df)
```
MichaelChirico: anyway, why not also search data.frames?

```r
DF = data.frame(a = 1:2)
DF$b = data.table(c = 3:4, d = 5:6)
str(DF)
# 'data.frame':   2 obs. of  2 variables:
#  $ a: int  1 2
#  $ b:Classes ‘data.table’ and 'data.frame':     2 obs. of  2 variables:
#   ..$ c: int  3 4
#   ..$ d: int  5 6
#   ..- attr(*, ".internal.selfref")=<pointer: 0x561d6ec08e30>
```

R/tables.R
```diff
@@ -19,19 +19,87 @@ type_size = function(DT) {
...
+          if (!is.null(elem_names) && nzchar(elem_names[j])) {
+            new_name = paste0(parent_name, "$", elem_names[j])
+          } else {
+            new_name = paste0(parent_name, "[[", j, "]]")
+          }
+          DT = obj_list[[i]][[j]]
+          k = cnt + length(w) # row number in info data.table
+          cnt = cnt + 1L
+          set(info, k, "NAME", new_name)
+          set(info, k, "NROW", nrow(DT))
```
MichaelChirico: let's use a helper to share logic between here & the `depth=0` branch

R/tables.R
manmita: due to format(env) it might be slow, I had tried adding test to increase coverage for this line but atime didnt pass then. will try that again.

R/tables.R
manmita: removed is_df and kept only is_dt to count lists of df as well as the below comment says

inst/tests/tests.Rraw
```diff
@@ -21514,11 +21514,46 @@ test(2364.2, levels(x$a), levels(y$a))
...
+#2606 tables() depth=1 finds nested data.tables in lists
+# creating env so that the names are within it
+xenv = new.env()
+xenv$DT = data.table(a=1L)
+xenv$L = list(data.table(a=1, b=4:6), data.table(a=2, b=7:10))
+xenv$LL = list(list(data.table(a=1L, b=4:6)))
+xenv$M = list(b=data.table(a=1, b=4:6), a=1:5)
+xenv$N = list(a=1:5)
+test(2366.1,
+     tables(env=xenv, depth=1L, index=TRUE)[, .(NAME, NROW, NCOL, INDICES, KEY)],
```
manmita, line 21537: @MichaelChirico this test was failing as indices and key column was missing, can you please check if this is the intended testcase? I have updated it to add index.

R/tables.R
manmita: updated the code, and tested

inst/tests/tests.Rraw
MichaelChirico: That's my mistake 👍

R/tables.R
MichaelChirico: atime suite only runs these tests, not the full test suite:

https://github.com/Rdatatable/data.table/blob/master/.ci/atime/tests.R

So I guess it was unrelated

R/tables.R
```diff
@@ -18,28 +18,92 @@ type_size = function(DT) {
...
   w = which(vapply_1b(obj, is.data.table))
-  if (!length(w)) {
-    if (!silent) catf("No objects of class data.table exist in %s\n", if (identical(env, .GlobalEnv)) ".GlobalEnv" else format(env))
-    return(invisible(data.table(NULL)))
+
+  info = NULL
+  # we check if depth=1L is requested and add found tables to w
+  if (depth == 1L) {
+    is_list = vapply_1b(obj, is.list)
+    is_dt = vapply_1b(obj, is.data.table)
```
MichaelChirico, line 46: It's also possible to define data.table-as-column for data.tables:

```r
DT = data.table(a = 1:2)
DT[, b := data.table(c = 3:4, d=5:6)]
DT
#        a              b
#    <int>   <data.table>
# 1:     1 <multi-column>
# 2:     2 <multi-column>
DT$b
#    <int> <int>
# 1:     3     5
# 2:     4     6
```

(this is a bit odd because the `names()` are lost, so maybe it's not too common..., but I don't think it's difficult to allow for this here -- please flag if it makes the code uglier)

R/tables.R
```diff
@@ -18,28 +18,91 @@ type_size = function(DT) {
...
+          idx = wl[j]
+          if (!is.null(elem_names) && nzchar(elem_names[idx])) {
+            new_name = paste0(parent_name, "$", elem_names[idx])
+          } else {
+            new_name = paste0(parent_name, "[[", idx, "]]")
+          }
+          DT = L[[idx]]
+          k = cnt + length(w) # row number in info data.table
+          cnt = cnt + 1L
+          set(info, k, "NAME", new_name)
```
MichaelChirico: is this redundant with the name setting in L67 above?

manmita: yeah, moved it to the function

R/tables.R
manmita: in  this case code changes needed are
1) for every datatable, check the columns if that has datable and add this results to the list_index
this will make sure that those tables are also added into the result, but columns are null in this case as names are lost

Question: what to add as the names of the columns in this case, can we let it be null?

MichaelChirico: Hm, good point... on second thought, let's consider this type of data.table out of scope for now.

manmita: yeah, we can check this out when we look at recursion for nested dt

MichaelChirico:
LGTM thanks!

manmita:
Thanks!

PR #7585, opened 2026-01-11
Fix latest rchk issues

MichaelChirico:
Following https://github.com/kalibera/rchk/issues/40, `rchk` is showing some new uncaught `PROTECT()` issues.

### DIFF
--- a/src/rbindlist.c
+++ b/src/rbindlist.c
@@ -305,7 +305,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
         if (isOrdered(thisCol)) {
           orderedFactor = true;
           int thisLen = length(getAttrib(thisCol, R_LevelsSymbol));
-          if (thisLen > longestLen) { longestLen=thisLen; REPROTECT(longestLevels = getAttrib(thisCol, R_LevelsSymbol), ILongestLevels); /*for warnings later ...*/longestW=w; longestI=i; }
+          if (thisLen > longestLen) { longestLen=thisLen; REPROTECT(longestLevels=getAttrib(thisCol, R_LevelsSymbol), ILongestLevels); /*for warnings later ...*/longestW=w; longestI=i; }
         }
       } else if (!isString(thisCol)) anyNotStringOrFactor=true;  // even for length 0 columns for consistency; test 2113.3
       if (INHERITS(thisCol, char_integer64)) {


src/rbindlist.c
```diff
@@ -277,7 +277,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
     int maxType=LGLSXP;  // initialize with LGLSXP for test 2002.3 which has col x NULL in both lists to be filled with NA for #1871
     bool factor=false, orderedFactor=false;     // ordered factor is class c("ordered","factor"). isFactor() is true when isOrdered() is true.
     int longestLen=-1, longestW=-1, longestI=-1; // just for ordered factor; longestLen must be initialized as -1 so that rbind zero-length ordered factor could work #4795
-    SEXP longestLevels=R_NilValue;              // just for ordered factor
+    SEXP longestLevels=PROTECT(R_NilValue); nprotect++; // just for ordered factor
```
MichaelChirico: this feels dumb, but the other things I tried didn't work. Overall we really should try and modularize the internals here a bit...

src/rbindlist.c
```diff
@@ -303,7 +303,7 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
         if (isOrdered(thisCol)) {
           orderedFactor = true;
           int thisLen = length(getAttrib(thisCol, R_LevelsSymbol));
-          if (thisLen>longestLen) { longestLen=thisLen; longestLevels=getAttrib(thisCol, R_LevelsSymbol); /*for warnings later ...*/longestW=w; longestI=i; }
+          if (thisLen>longestLen) { longestLen=thisLen; UNPROTECT(1); longestLevels=PROTECT(getAttrib(thisCol, R_LevelsSymbol)); /*for warnings later ...*/longestW=w; longestI=i; }
```
ben-schwen: maybe move the assignement of longestLevels here into the else branch?

src/assign.c
```diff
@@ -70,11 +70,11 @@ static int _selfrefok(SEXP x, Rboolean checkNames, Rboolean verbose) {
   if (!isNull(p)) internal_error(__func__, ".internal.selfref ptr is neither NULL nor R_NilValue"); // # nocov
   tag = R_ExternalPtrTag(v);
   if (!(isNull(tag) || isString(tag))) internal_error(__func__, ".internal.selfref tag is neither NULL nor a character vector"); // # nocov
-  names = getAttrib(x, R_NamesSymbol);
   prot = R_ExternalPtrProtected(v);
   if (TYPEOF(prot) != EXTPTRSXP)   // Very rare. Was error(_(".internal.selfref prot is not itself an extptr")).
     return 0;                      // # nocov ; see http://stackoverflow.com/questions/15342227/getting-a-random-internal-selfref-error-in-data-table-for-r
-  return checkNames ? names==tag : x==R_ExternalPtrAddr(prot);
+  if (!checkNames) return x == R_ExternalPtrAddr(prot);
+  return getAttrib(x, R_NamesSymbol) == tag;
```
ben-schwen, line 77: We should at least add a comment that this needs to be the last call since `Rf_getAttrib(?, R_NamesSymbol)` is an allocating function

ben-schwen: So AFAICT the "problem" of `rchk` arises when we call another allocating function, if its not in a return. I don't why but this feels some kind of wrong since can't we wrap everything into an return block to trick this?

ben-schwen: Moreover, we do not have to PROTECT the return of `getAttrib` if we directly wrap another unary function around it like `length(getAttrib(?,?))` or `STRING_ELT(getAttrib(?,?), ?)` but `R_compute_identical` needs a PROTECT.

src/rbindlist.c
MichaelChirico: hm I'm not quite sure what you have in mind here, maybe I'm missing it. I have something like it, definitely an improvement.

src/assign.c
MichaelChirico: I guess this is likely just an `rchk` false positive, and I think Tomas excludes it from CRAN, but we don't have a way to match his heuristics, so this is overly-conservative.

I think you're right we _could_ shrink the diff and just put the `getAttrib()` directly inside the `return` statement, OTOH it would make it a fair bit more complicated, so I went with this early return approach instead. WDYT?

ben-schwen: `_selfrefok` looks ok.

src/rbindlist.c
```diff
@@ -303,7 +303,8 @@ SEXP rbindlist(SEXP l, SEXP usenamesArg, SEXP fillArg, SEXP idcolArg, SEXP ignor
         if (isOrdered(thisCol)) {
           orderedFactor = true;
           int thisLen = length(getAttrib(thisCol, R_LevelsSymbol));
-          if (thisLen>longestLen) { longestLen=thisLen; longestLevels=getAttrib(thisCol, R_LevelsSymbol); /*for warnings later ...*/longestW=w; longestI=i; }
+          if (longestLen == -1) longestLevels = PROTECT(R_NilValue);
```
ben-schwen: seems we really need to go with the `SEXP longestLevels = PROTECT(R_NilValue); nprotect++`

MichaelChirico:
ah, I forgot about reprotect too. let me try that later.

MichaelChirico:
Thanks for jamming on this @ben-schwen! I always forget about `REPROTECT()`, it's clearly the "right" thing to do here. I think we are good to go now.

Issue #7546, opened 2025-12-29
valgrind issue in 1.18.0


MichaelChirico:
From CRAN:

https://www.stats.ox.ac.uk/pub/bdr/memtests/valgrind/data.table/00check.log

```
  Test 6001.732 ran without errors but failed check that x equals y:
  > x = frollsd(y, 3)[4L] 
  First 1 of 1 (type 'double'): 
  [1] 1.825012e-08
  > y = 0 
  First 1 of 1 (type 'double'): 
  [1] 0
  Mean relative difference: 1
```

```
  Error in test.data.table(script = "froll.Rraw") : 
    1 error out of 1316. Search tests/froll.Rraw.bz2 for test number 6001.732. Duration: 00:03:47 elapsed (00:03:20 cpu).
  Calls: test.data.table -> stopf -> raise_condition -> signal
  Execution halted
  ==3229069== 
  ==3229069== HEAP SUMMARY:
  ==3229069==     in use at exit: 92,363,323 bytes in 16,467 blocks
  ==3229069==   total heap usage: 174,670 allocs, 158,203 frees, 332,134,037 bytes allocated
  ==3229069== 
  ==3229069== 432 bytes in 1 blocks are possibly lost in loss record 358 of 2,244
  ==3229069==    at 0x484B133: calloc (/builddir/build/BUILD/valgrind-3.24.0/coregrind/m_replacemalloc/vg_replace_malloc.c:1675)
  ==3229069==    by 0x4011F63: UnknownInlinedFun (/usr/src/debug/glibc-2.39-38.fc40.x86_64/elf/../include/rtld-malloc.h:44)
  ==3229069==    by 0x4011F63: allocate_dtv (/usr/src/debug/glibc-2.39-38.fc40.x86_64/elf/../elf/dl-tls.c:395)
  ==3229069==    by 0x4012A61: _dl_allocate_tls (/usr/src/debug/glibc-2.39-38.fc40.x86_64/elf/../elf/dl-tls.c:673)
  ==3229069==    by 0x557CC03: allocate_stack (/usr/src/debug/glibc-2.39-38.fc40.x86_64/nptl/allocatestack.c:431)
  ==3229069==    by 0x557CC03: pthread_create@@GLIBC_2.34 (/usr/src/debug/glibc-2.39-38.fc40.x86_64/nptl/pthread_create.c:660)
[...]

PR #7589, opened 2026-01-12
frollmedianFast: avoid reading uninitialised array

aitap:
The `n` array is only initialised if `even` is true, so skip the comparisons otherwise. Detected by checking with `--use-valgrind` or performing a `frollmedian()` with an odd window size under `R -d valgrind`.

Fixes: #7546

### DIFF
--- a/src/froll.c
+++ b/src/froll.c
@@ -1707,11 +1707,13 @@ void frollmedianFast(const double *x, uint64_t nx, ans_t *ans, int k, double fil
           snprintf(end(ans->message[3]), 500, _("%s: 's[A] + s[B] == h' is not true\n"), "frollmedianFast");
           return;
         }*/
-        if (n[A]!=tail && m[A] == n[A]) {
-          n[A] = tail;
-        }
-        if (n[B]!=tail && m[B] == n[B]) {
-          n[B] = tail;
+        if (even) {
+          if (n[A]!=tail && m[A] == n[A]) {
+            n[A] = tail;
+          }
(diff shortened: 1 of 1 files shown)


MichaelChirico:
LGTM, I'll let Jan give final approval. Should we add a new test? codecov says we lost a line.

src/froll.c
```diff
@@ -1707,11 +1707,13 @@ void frollmedianFast(const double *x, uint64_t nx, ans_t *ans, int k, double fil
...
           return;
         }*/
-        if (n[A]!=tail && m[A] == n[A]) {
-          n[A] = tail;
-        }
-        if (n[B]!=tail && m[B] == n[B]) {
-          n[B] = tail;
+        if (even) {
+          if (n[A]!=tail && m[A] == n[A]) {
+            n[A] = tail;
```
aitap, line 1712: I'm struggling to devise a test that would reach this line. The closest I've got was `n[A]` = `m[A]` = `tail` = `4` for `frollmedian(c(1:4, rep(1,3)), 4)`.

jangorecki:
Codecov was a tricky part here. As long as function is marked experimental, then I think it is ok to not have that line covered.

Ben's idea is the best but it requires someone very ambitious to crunch that. Possibly adding NA support too.
I got an inquiry on the email about rolling quantitle, which could possibly be considered if someone wants to have a deeper look at reworking rolling median.

jangorecki (approved):

Issue #7595, opened 2026-01-14
setattr() crashes with duplicate levels


sindribaldur:
I'm aware this is not something you should be doing but I guess crashing is not the best way out. 

```R
setattr(factor(c(1:3), levels = c("1", "2")), "levels", c("1", "1"))
```

### System info (where it consistently crashes R).
```
> version
               _
platform       x86_64-w64-mingw32
arch           x86_64
os             mingw32
crt            ucrt
system         x86_64, mingw32
status
major          4
minor          5.1
year           2025
month          06
day            13
svn rev        88306
language       R
version.string R version 4.5.1 (2025-06-13 ucrt)
nickname       Great Square Root
```

aitap:
Thank you for the report! Confirmed on R-devel / `gcc -fsanitize=address,undefined` / Linux and latest `master`:

```
wrappers.c:48:54: runtime error: signed integer overflow: -2147483648 - 1 cannot be represented in type 'int'
Error in data.table::setattr(factor(c(1:3), levels = c("1", "2")), "levels",  : 
  attempt access index 2147483647/2 in STRING_ELT
Execution halted
```

For your future reports, `sessionInfo()` would be more informative than just `version`, even with sensitive parts redacted.

PR #7596, opened 2026-01-14
`setlevels()`: avoid crash on missing factor values

aitap:
Check the index for being `NA_INTEGER` before trying to access it.

Should this also protect from invalid factors? For example, this also currently crashes:

```r
data.table::setattr(structure(integer(999), class = "factor", levels = "a"), "levels", c("b", "b"))
```

Should this go into 1.18.2?

Fixes: #7595


aitap:
Test `2361.2` could also signal an error, similar to how `` base::`levels<-` `` fails (but doesn't crash) when given such a factor.

MichaelChirico:
> Should this go into 1.18.2?

I'd say yes, anytime we eliminate a crash it's important to include it ASAP

